PR #27020, opened 2024-05-08
Rover: Notify lat and long to the seventh decimal place

muramura:
The seventh decimal place corresponds to centimeters.

### DIFF
--- a/Rover/commands.cpp
+++ b/Rover/commands.cpp
@@ -33,7 +33,7 @@ bool Rover::set_home(const Location& loc, bool lock)
     mode_auto.mission.write_home_to_storage();
 
     // send text of home position to ground stations
-    gcs().send_text(MAV_SEVERITY_INFO, "Set HOME to %.6f %.6f at %.2fm",
+    gcs().send_text(MAV_SEVERITY_INFO, "Set HOME to %.7f %.7f at %.2fm",
             static_cast<double>(loc.lat * 1.0e-7f),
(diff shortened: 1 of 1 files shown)


khancyr (approved):
Fit in a single message anyway. LGTM

peterbarker (approved):

rmackay9:
Does it actually display all the digits?  I suspect it does in which case I should do the same in the recent slung payload simulator.

peterbarker:
> Does it actually display all the digits? I suspect it does in which case I should do the same in the recent slung payload simulator.

I haven't provoked it to do so, but we may as well give it every chance.

Rover/commands.cpp
```diff
@@ -33,7 +33,7 @@ bool Rover::set_home(const Location& loc, bool lock)
     mode_auto.mission.write_home_to_storage();
 
     // send text of home position to ground stations
-    gcs().send_text(MAV_SEVERITY_INFO, "Set HOME to %.6f %.6f at %.2fm",
+    gcs().send_text(MAV_SEVERITY_INFO, "Set HOME to %.7f %.7f at %.2fm",
```
tridge, line 36: this won't work on ChibiOS at least, possibly not on SITL too, we go via a float inside the printf engine

rmackay9:
We think this doesn't actually work.  Remember to provide testing results!

We discussed on the dev call that what we need is a function somewhere that allows printing lat,lon values to the required number of decimal places

PR #27022, opened 2024-05-08
Rover: Perform an arm check first

muramura:
Performing an arm check first ensures that the fence check does not become unnecessary.

### DIFF
--- a/Rover/fence.cpp
+++ b/Rover/fence.cpp
@@ -4,16 +4,15 @@
 void Rover::fence_check()
 {
 #if AP_FENCE_ENABLED
-    uint8_t new_breaches;  // the type of fence that has been breached
-    const uint8_t orig_breaches = fence.get_breaches();
-
-    // check for a breach
-    new_breaches = fence.check();
-
     // return immediately if motors are not armed
     if (!arming.is_armed()) {
         return;
     }
+
+    const uint8_t orig_breaches = fence.get_breaches();
+
+    // check for a breach
+    uint8_t new_breaches = fence.check();
 
     // if there is a new breach take action
     if (new_breaches) {


Rover/fence.cpp
```diff
@@ -4,17 +4,16 @@
...
-
     // return immediately if motors are not armed
     if (!arming.is_armed()) {
         return;
     }
 
+    const uint8_t orig_breaches = fence.get_breaches();
+
+    // check for a breach
+    uint8_t new_breaches = fence.check();
```
IamPete1, line 15: Think this is the main "update" function for fence. So even thought the return is not used we still need to call it. Otherwise you the fence arming check might not work for example.

muramura: A fence_check is performed in the arming.pre_arm_checks method.

https://github.com/ArduPilot/ardupilot/blob/master/Rover/Rover.cpp#L450

IamPete1: Right, but it does not actually check if the vehicle is inside the fence, it just uses the result of the last check. Except there now will be no result because your not running the check.

IamPete1 (changes requested):
This is failing CI because it not correct.
```
2024-05-08T19:37:04.4497292Z Failure Summary:
2024-05-08T19:37:04.4497637Z   test.Rover:
2024-05-08T19:37:04.4498793Z     PolyFence (test fence-related functions) (Expected to be breached) (see /tmp/buildlogs/Rover-PolyFence.txt) (duration 3.3232529163360596 s)
2024-05-08T19:37:04.4500007Z FAILED 1 tests: ['test.Rover']
```

It would be much clearer if the fence `check` method was void and called `update`. Then a second call to `get_breaches` to check for new breaches.

Rover/fence.cpp
muramura: I understood.

PR #27024, opened 2024-05-09
Initialize sample buffer pointer with nullptr

reznikmm:
to avoid memory corruption on deallocating uninitialized value in `CompassCalibrator.cpp`:384:

```c
if (_sample_buffer != nullptr) {
    free(_sample_buffer);
```

### DIFF
--- a/libraries/AP_Compass/CompassCalibrator.cpp
+++ b/libraries/AP_Compass/CompassCalibrator.cpp
@@ -79,7 +79,8 @@ extern const AP_HAL::HAL& hal;
 ///////////////////// PUBLIC INTERFACE /////////////////////
 ////////////////////////////////////////////////////////////
 
-CompassCalibrator::CompassCalibrator()
+CompassCalibrator::CompassCalibrator() :
+    _sample_buffer(nullptr)
 {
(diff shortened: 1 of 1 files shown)


libraries/AP_Compass/CompassCalibrator.cpp
```diff
@@ -79,7 +79,8 @@ extern const AP_HAL::HAL& hal;
 ///////////////////// PUBLIC INTERFACE /////////////////////
 ////////////////////////////////////////////////////////////
 
-CompassCalibrator::CompassCalibrator()
+CompassCalibrator::CompassCalibrator() :
+    _sample_buffer(nullptr)
```
peterbarker, line 83: For this to be a problem we'd have to have allocated this on the stack, essentially: https://ardupilot.org/dev/docs/style-guide.html#implicit-zeroing-of-memory

Do you still see a problem here?

reznikmm: Yes, in my case it was on the stack. Thank you for the reference. Feel free to close PR then and sorry for the noise.

peterbarker:
Thanks anyway, @reznikmm

PR #27027, opened 2024-05-09
Fix signedness issue with snprintf

durka:
`AP_InternalError::errors_as_string` casts the return value of `snprintf`, which is signed `int`, to unsigned `size_t`. So the comparison `<= 0` is effectively `== 0`, ignoring errors.

I don't think snprintf actually returns an error unless the format string is wrong, which it isn't, so this is probably mostly academic. I don't believe the `== 0` case is hit anyway, since the loop exits earlier by checking `buffer_used >= len`.

Also included a micro-optimization to avoid a conditional in the loop.

Note: I am employed by Exyn Technologies

### DIFF
--- a/libraries/AP_InternalError/AP_InternalError.cpp
+++ b/libraries/AP_InternalError/AP_InternalError.cpp
@@ -85,22 +85,19 @@ void AP_InternalError::errors_as_string(uint8_t *buffer, const uint16_t len) con
 {
     buffer[0] = 0;
     uint32_t buffer_used = 0;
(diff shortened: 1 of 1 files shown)


tridge (approved):

tridge:
@durka thanks!
minor thing, our code style requires the commit to start with the subsystem, so:
  AP_InternalError: fix signedness issue with snprintf
I can fix your PR if you don't know how to reword a commit

durka:
updated the commit message

peterbarker:
Merged, thanks!

PR #26189, opened 2024-02-11
Copter: run rate loop at full filter rate in its own thread

tridge:
This adds FLIGHT_OPTIONS=8 which creates a rate loop thread, allowing rate updates at the full filtered gyro rate (2kHz on most copters), settable with existing INS_GYRO_RATE parameter.
This was test flown successfully on a 5" quad from @MichelleRos, the rate loop did run at 2kHz (I added a send text for average rate every 2s)

We've now had successful flights at 2kHz, 4kHz and 8kHz on a MatekH743-bdshot 5" quad. At 8kHz we had to disable the 2nd IMU with IMU_ENABLE_MASK or we had R/C issues

PR #27029, opened 2024-05-09
Copter: run rate loop at full filter rate in its own thread

andyp1per:
This is a redo of https://github.com/ArduPilot/ardupilot/pull/26189

I have squashed the commits, rebased and started fixing the underlying problems. There were some fundamental problems with how the original PR was handling attitude control changes so I thought it was better to just open a new PR.

Support is enabled by setting:
```
FSTRATE_ENABLE = 1
```
which gives a variable attitude rate depending on load, and:
```
FSTRATE_ENABLE = 2
```
which gives an attitude rate locked to the gyro rate, but dynamic while disarmed. For testing there is also:
```
FSTRATE_ENABLE = 3
```
which is a locked rate at all times.

The output rate can be adjusted using ```FSTRATE_DIV``` which is a gyro divisor (default 1).
Two different rates - and therefore tunes - can be compared by using the lua applet ```switch_rates.lua``` which will persistently switch the appropriate tune/rate parameters with ones saved in names beginning with ```X_```


ArduCopter/Attitude.cpp
```diff
@@ -4,18 +4,141 @@
...
+// @LoggerMessage: RTDT
+// @Description: Attitude controller time deltas
+// @Field: TimeUS: Time since system startup
+// @Field: dt: current time delta
+// @Field: dtAvg: current time delta average
+// @Field: dtMax: Max time delta since last log output
+// @Field: dtMin: Min time delta since last log output
+
+        if (now_us - last_rtdt_log_us >= 2500) {    // 400 Hz
+            AP::logger().WriteStreaming("RTDT", "TimeUS,dt,dtAvg,dtMax,dtMin", "Qffff",
```
IamPete1, line 75:
```suggestion
        if (now_us - last_rtdt_log_us >= 2500) {    // 400 Hz
            last_rtdt_log_us = now_us;
            AP::logger().WriteStreaming("RTDT", "TimeUS,dt,dtAvg,dtMax,dtMin", "Qffff",
```

libraries/AC_PID/AC_PID.cpp
```diff
@@ -275,6 +277,8 @@ float AC_PID::update_all(float target, float measurement, float dt, bool limit,
     _pid_info.D = D_out;
     _pid_info.FF = _target * _kff;
     _pid_info.DFF = _target_derivative * _kdff;
+    _pid_info.dt = dt;
+    _pid_info.max_dt = MAX(dt, _pid_info.dt);
```
IamPete1, line 281: This will presumably be the same for roll pitch and yaw? That is 4 redundant floats vs a time stamp and log header. New log message would use less log space especially if we ever wanted to expand logging of such things eg min dt.

andyp1per: Yes agree, just WIP - will remove it when I have the bugs sorted

ArduCopter/Attitude.cpp
andyp1per: Fixed

libraries/AC_PID/AC_PID.cpp
andyp1per: Removed

andyp1per:
Flow today on top of 4.5 - working well.

andyp1per:
@rmackay9 this is close to being ready and I could use your input on what the configuration should look like. The basic premise is that we run the rate controller at the same rate as the gyros which is controlled by ```INS_GYRO_RATE``` (0=1Khz,1=2Khz,2=4Khz,3=8Khz). However, this quite expensive and will not be possible on some hardware setups. In addition we need to be able to log the rate outputs at a faster rate than the main loop rate, but its unlikely that we could successfully log at the gyro rate. Finally we need to update the filters (notch filter in particular) at a fast rate than the main loop, but it does not make sense to do this any faster than at half the gyro rate. 

Currently we have the following:
```FLIGHT_OPTIONS=8``` - enables the rate thread and dynamically schedules the relationship to the gyro rate based on load
```FLIGHT_OPTIONS=16``` - fixes the rate thread at the same rate as the gyro rate
```ATT_FILTER_RATE``` - update rate of the filters in Hz or fixed at half the gyro rate if set to 0
```ATT_LOG_RATE``` - update rate of the rate logs in Hz or set at the main loop rate if set to 0
The last two are not exact numbers - they are just the closest we can get to the required frequency by counting gyro samples so need to be an integer divisor or the gyro rate.

I'm not currently convinced about the naming or the configuration. I think this is an important enough feature that it needs its own enable flag. I also think it would be worth having a rate parameter that allows you to fix the rate. I also think that ATT is a little bit misleading. Maybe FAST_RATE_ or RATET_ or something. So an alternative might be:

```FRATE_ENABLE``` - set to 1 to enable the rate thread
```FRATE_RATE``` - set to 0 to dynamically throttle, fixed Hz otherwise
```FRATE_LOG_RATE``` - as above
```FRATE_FILT_RATE``` - as above

or something. It would also be possible to not use Hz at all but simply pick the divisor. So 2 would be gyro rate / 2 etc.
Your thoughts appreciated on this important topic!

File: ArduCopter/Attitude.cpp [hunk 3]
```diff
@@ -1,21 +1,285 @@
...
+        if ((!flight_option_is_set(FlightOptions::USE_RATE_LOOP_THREAD) &&
+             !flight_option_is_set(FlightOptions::USE_FIXED_RATE_LOOP_THREAD)) ||
+            ap.motor_test) {
+            using_rate_thread = false;
+            if (was_using_rate_thread) {
+                // if we were using the rate thread, we need to
+                // setup the notch filter sample rate
+                const float loop_rate_hz = AP::scheduler().get_filtered_loop_rate_hz();
+                attitude_control->set_notch_sample_rate(loop_rate_hz);
+                motors->set_dt(1.0/loop_rate_hz);
```
IamPete1, line 70: motors dt is set lots of places, but I don't think the motors update rate is actually changed?

IamPete1 (commented):
It would be good to split this out into several smaller PRs. Things like the DAL rename would stand alone?

andyp1per:
Absolutely, but I want to make sure there is general agreement on the structure before doing thatAndySent from my iPhoneOn 16 Jun 2024, at 11:36, Peter Hall ***@***.***> wrote:﻿
@IamPete1 commented on this pull request.

It would be good to split this out into several smaller PRs. Things like the DAL rename would stand alone?

Tools/autotest/arducopter.py
```diff
@@ -6183,6 +6183,77 @@ def DynamicRpmNotches(self):
...
+            "INS_LOG_BAT_OPT": 0,
+            "INS_GYRO_FILTER": 300, # set gyro filter high so we can observe behaviour
+            "LOG_BITMASK": 959,
+            "LOG_DISARMED": 0,
+            "SIM_VIB_MOT_MAX": 350,
+            "SIM_GYR1_RND": 20,
+            "SIM_ESC_TELEM": 1,
+            "FLIGHT_OPTIONS": 8
+        })
+        self.reboot_sitl()
```
peterbarker, line 6203: We need a `context_push()` at the top of this test and `context_pop()` at the bottom followed by the existing `reboot_sitl()`

... otherwise subsequent tests will end up running with the wrong parameters in effect (not value, effect).  Assuming this reboot_sitl() is here for a reason....

libraries/AP_HAL/board/chibios.h
```diff
@@ -86,6 +88,10 @@
 #define HAL_INS_ICM20608_AM_NAME "icm20608-am"
 #define HAL_INS_ICM20608_EXT_NAME "icm20608_ext"
 
+#ifndef HAL_INS_RATE_LOOP
+#define HAL_INS_RATE_LOOP 1
+#endif
+
```
peterbarker, line 94: Maybe mve this default-to-1 code up into `AP_HAL/AP_HAL_Boards.h`?

andyp1per: Fixed

Tools/autotest/arducopter.py
andyp1per: Fixed

andyp1per:
Did some fairly comprehensive flying including checking that I could switch between rates mid-flight - all worked fine

ArduCopter/Attitude.cpp
andyp1per: Fixed

ArduCopter/Attitude.cpp
```diff
@@ -1,21 +1,345 @@
...
+#include <AP_HAL/CondMutex.h>
 
 /*************************************************************
  *  Attitude Rate controllers and timing
  ****************************************************************/
 
-// update rate controllers and output to roll, pitch and yaw actuators
-//  called at 400hz by default
-void Copter::run_rate_controller()
+#if AP_INERTIALSENSOR_RATE_LOOP_WINDOW_ENABLED
```
tridge, line 8: maybe break into fast_rate.cpp ?

tridge (changes requested):
my preference for 4.6.x would be this as a custom build server feature, but should ask @rmackay9 about this.
It should have less or preferably zero flash impact on plane.
I'd like more info on why condition variables are needed and why binary semaphores don't work as well, preferably a version of the binary semaphore test that shows the issue

andyp1per:
This is the rough cost for the refactoring required:

```
andy@Eagle:~/github/ardupilot$ ./Tools/scripts/pretty_diff_sym.py ../ardupilot-dev/build/MatekH743-bdshot/bin/arduplane build/MatekH743-bdshot/bin/arduplane
Symbol size differences between ../ardupilot-dev/build/MatekH743-bdshot/bin/arduplane and build/MatekH743-bdshot/bin/arduplane:
  18 AP_InertialSensor_Backend::_publish_gyro(unsigned 126 -> 144
   4 AC_AttitudeControl::input_thrust_vector_heading(Vector3<float> 376 -> 380
 180 AP_InertialSensor_Backend::update_gyro_filters(unsigned 0 -> 180
-190 AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_2(float, 212 -> 22
   2 AP_HAL::RCOutput::force_trigger_groups(bool) 0 -> 2
   4 ChibiOS::DeviceBus::bus_thread(void*) 188 -> 192
   6 ChibiOS::SPIDevice::set_periodic_minimum(unsigned 0 -> 6
  -8 ChibiOS::RCOutput::push_local() 408 -> 400
  20 chMtxLockS 120 -> 140
   4 AP_AHRS_get_accel(lua_State*) 84 -> 88
   4 AC_AttitudeControl::input_euler_angle_roll_pitch_yaw(float, 540 -> 544
   2 AP_InertialSensor_Backend::set_primary_gyro(unsigned 0 -> 2
  -8 AC_AttitudeControl::attitude_controller_run_quat() 508 -> 500
   4 AC_AttitudeControl::relax_attitude_controllers() 176 -> 180
   4 SLT_Transition::update() 1148 -> 1152
   4 AC_AttitudeControl::input_rate_bf_roll_pitch_yaw(float, 296 -> 300
   4 AC_AttitudeControl_TS::input_euler_rate_yaw_euler_angle_pitch_bf_roll(bool, 556 -> 560
  20 AP_AHRS_View::Write_Rate(AP_Motors 484 -> 504
   8 SRV_Channels::calc_pwm() 212 -> 220
   4 ChibiOS::DeviceBus::DeviceBus(unsigned 44 -> 48
   8 rcoutDriver 672 -> 680
  44 SRV_Channels::output_pwm_chan(unsigned 0 -> 44
  16 AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_3(float, 472 -> 488
  24 AP_InertialSensor::update() 464 -> 488
   4 AC_AttitudeControl::input_angle_step_bf_roll_pitch_yaw(float, 136 -> 140
  -4 AP_InertialSensor::write_notch_log_messages() 368 -> 364
   4 AC_AttitudeControl::input_euler_angle_roll_pitch_euler_rate_yaw(float, 420 -> 424
   4 ChibiOS::RCOutput::set_dshot_rate(unsigned 136 -> 140
   6 ChibiOS::I2CDevice::set_periodic_minimum(unsigned 0 -> 6
  24 AP_Motors::add_motor_num(signed 26 -> 50
  12 AC_AttitudeControl_Multi::AC_AttitudeControl_Multi(AP_AHRS_View&, 480 -> 492
  62 AP_MotorsMulticopter::update_servo_map() 0 -> 62
   8 AP_Vehicle::update_dynamic_notch_at_specified_rate() 92 -> 100
 300 AC_AttitudeControl_Multi::rate_controller_run_dt(float, 0 -> 300
 104 AC_AttitudeControl::input_rate_step_bf_roll_pitch_yaw(float, 0 -> 104
   4 AP_AHRS::_get_primary_accel_index() 0 -> 4
   4 AC_AttitudeControl::input_euler_rate_roll_pitch_yaw(float, 384 -> 388
   4 GCS_MAVLINK_Plane::send_attitude_target() 104 -> 108
   2 AP_Motors::rc_write(unsigned 72 -> 74
 -32 AP_InertialSensor_Backend::update_accel(unsigned 134 -> 102
   6 AP_AHRS::write_video_stabilisation() 186 -> 192
  28 CSWTCH.378 0 -> 28
   2 AP_InertialSensor_Backend::set_primary_accel(unsigned 0 -> 2
   2 AC_AttitudeControl::rate_controller_run_dt(float, 0 -> 2
   4 AP_MotorsMulticopter::output() 80 -> 84
  60 AP_InertialSensor_Backend::update_accel_filters(unsigned 0 -> 60
 228 AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_4(float, 0 -> 228
 -64 AC_AutoTune_Multi::twitch_test_run(AC_AutoTune::AxisType, 832 -> 768
-290 AC_AttitudeControl_Multi::rate_controller_run() 332 -> 42
   8 ChibiOS::I2CDeviceManager::businfo 200 -> 208
   4 AP_InertialSensor_Invensense::start() 924 -> 928
   8 ChibiOS::RCOutput::trigger_groups() 204 -> 212
  44 AC_AttitudeControl_Multi::rate_controller_target_reset() 0 -> 44
 -28 CSWTCH.368 28 -> 0
   8 Plane::set_takeoff_expected() 140 -> 148
  48 AP_InertialSensor_Invensense::_read_fifo() 524 -> 572
   6 ChibiOS::RCOutput::force_trigger_groups(bool) 0 -> 6
   2 AP_InertialSensor_Invensensev3::set_primary_gyro(unsigned 0 -> 2
  -6 AP_InertialSensor_Backend::apply_gyro_filters(unsigned 276 -> 270
-172 AP_InertialSensor_Backend::update_gyro(unsigned 274 -> 102
   2 AC_AttitudeControl::rate_controller_target_reset() 0 -> 2
   2 AP_InertialSensor_Invensense::set_primary_gyro(unsigned 0 -> 2
Total: 586
```

andyp1per:
The reason why BinarySemaphore does not work in this case is two-fold:

1. It does not use mutexes and so there is no priority inheritance between related threads, this was one reason why the original code had to increase the thread prioty of the attitude thread. The new code has no need of this - the priorities manage themselves. This makes for a safer change.
2. You if you wait after signal has been called you will timeout. This is because binary semaphores contain no history - there is no matching of wait to signal. This is easy to verify by making the first thread in the test signal and delay before creating the second thread. We are managing a queue of gyro samples of which we need to process all. If a gyro sample is queued before the attitude thread is ready to receive then signal is lost and the attitude thread will not get woken up until the *next* gyro sample arrives. Obviously not good. To try and fix this you could then have an additional mutex to control access to the number of waiting gyros, but at that point you have essentially implemented condition variables which ChibiOS has a much more low-level and efficient implementation of - plus you still don't get the benefits of point (1).

tridge:
I hope I'm not just being thick here, but I think I really do need to understand what is going on here. Potentially dumb questions below.

> 1. It does not use mutexes and so there is no priority inheritance between related threads, this was one reason why the original code had to increase the thread prioty of the attitude thread. The new code has no need of this - the priorities manage themselves. This makes for a safer change.

first off, by "attitude thread" I assume you mean the rate thread? The attitude control is still in the main thread as far as I can see, only rate update in the rate thread. I know we discussed moving the attitude update into that thread but unless I've missed it that isn't being done in this PR, and it would be a tricky thing to do as you would need an attitude update (so would need to either run EKF propagation at high rate or do your own propagation using gyro/accel data)
secondly, I can't see how priority inheritance plays any role at all. The only two threads involved with this CondMutex are the SPI threads doing the transfers and filtering and the rate thread. Both are at priority 181, so neither change priority at all. Is there some interaction I'm missing that causes priority inheritance to kick in? If so, between what threads and under what circumstances?

> 2\. This is easy to verify by making the first thread in the test signal and delay before creating the second thread.

the signal can come from any thread, the state is in the binary semaphore object itself, so I don't see how the first thread delaying makes a difference. I'd like to have this as an actual test to understand it

tridge:
> 2\. This is easy to verify by making the first thread in the test signal and delay before creating the second thread.

I've updated the BinarySem test to do what you describe and it works and behaves exactly as I expected

https://github.com/tridge/ardupilot/commits/pr-copter-rate-thread/

I used a blocking wait, which matches what you are doing in this PR, no timeouts (you can't timeout with a blocking wait, it blocks forever), and gets exactly the performance I expect:

```
tick 21068 99.681 ops/s 0.000 timeouts/s
tick 22071 99.681 ops/s 0.000 timeouts/s
tick 23075 99.681 ops/s 0.000 timeouts/s
tick 24078 99.681 ops/s 0.000 timeouts/s
tick 25081 99.681 ops/s 0.000 timeouts/s
tick 26084 99.681 ops/s 0.000 timeouts/s
tick 27087 99.681 ops/s 0.000 timeouts/s
tick 28091 99.681 ops/s 0.000 timeouts/s
tick 29094 99.681 ops/s 0.000 timeouts/s
tick 30097 99.681 ops/s 0.000 timeouts/s
tick 31100 99.681 ops/s 0.000 timeouts/s
tick 32103 99.681 ops/s 0.000 timeouts/s
tick 33107 99.681 ops/s 0.000 timeouts/s
```

tridge:
> 1. It does not use mutexes and so there is no priority inheritance between related threads, this was one reason why the original code had to increase the thread prioty of the attitude thread. The new code has no need of this - the priorities manage themselves. This makes for a safer change.

I just noticed that the rate thread is being created with AP_HAL::Scheduler::PRIORITY_RCOUT+1, so it is priority 182, and SPI threads are priority 181, so there are different priorities, but the rate thread is the one waiting, and priority inheritance only ever boosts priority of the waiting thread. As the thread that is waiting (the rate thread) is at a higher priority than the thread sending the signal there can be no priority change, so I'm still puzzled as to how priority inheritance comes into the picture at all.

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
```diff
@@ -257,6 +272,28 @@ void AP_InertialSensor_Backend::apply_gyro_filters(const uint8_t instance, const
...
         _imu._gyro_filtered[instance] = gyro_filtered;
     }
+
+#if AP_INERTIALSENSOR_RATE_LOOP_WINDOW_ENABLED
+    if (_imu.push_rate_loop_gyro(instance)) {
+        /*
+          tell the rate thread we have a new sample
+        */
+        if (++_imu.rate_decimation_count >= _imu.rate_decimation) {
+            _imu._cmutex->lock_and_signal();
```
tridge, line 282: why are you signalling before you push?? I think this may be why you are getting the odd binary sem behaviour. You have to signal after you have made the sample available or the rate thread will wake up pointlessly and you will always be one loop behind

tridge (changes requested):

tridge:
@andyp1per we had a "climb away" event with this PR on Sunday with Michelle's quad, with FSTRATE_ENABLE=2. Looks like Z controller was not updating in ALT_HOLD mode.

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
andyp1per: This makes no difference - the waiter will not be woken up until it gets the mutex which won't happen until the mutex is unlocked by the signaler.

If the waiter is in chCondWait the first thing it needs to do is re-aquire the mutex which it cannot do until the signaler has released it.

https://chibios.org/dokuwiki/lib/exe/fetch.php?media=chibios:documentation:books:rt:kernel:monitor_states.png

As the docs state this is very similar to java synchronized which I am very familiar with:

https://chibios.org/dokuwiki/doku.php?id=chibios:documentation:books:rt:kernel_mutexes

andyp1per:
> I've updated the BinarySem test to do what you describe and it works and behaves exactly as I expected

Ok I have added an example that I think illustrates the problem. With binary sem:
```
Setup threads
tick 1000 1659.794 ops/s 0.000 timeouts/s missing 13
tick 2000 1700.771 ops/s 0.000 timeouts/s missing 22
tick 3000 1586.266 ops/s 0.000 timeouts/s missing 13
tick 4001 1695.454 ops/s 0.000 timeouts/s missing 17
tick 5001 1641.411 ops/s 0.000 timeouts/s missing 23
tick 6001 1783.319 ops/s 0.000 timeouts/s missing 11
tick 7001 1772.874 ops/s 0.000 timeouts/s missing 17
```
with CondMutex:
```
Setup threads
tick 1000 1692.284 ops/s 0.000 timeouts/s missing 0
tick 2000 1685.213 ops/s 0.000 timeouts/s missing 0
tick 3001 1574.518 ops/s 0.000 timeouts/s missing 0
tick 4001 1615.891 ops/s 0.000 timeouts/s missing 0
tick 5001 1660.900 ops/s 0.000 timeouts/s missing 0
tick 6002 1733.622 ops/s 0.000 timeouts/s missing 0
```
it's obviously a bit contrived, but the point is that BinarySemphore will not work on its own and actually from a concurrency POV I am not sure it can be made to work at all because it is not a lock in the true sense of the word - you end up with all these narrow races if you start trying to use a mutex to manage a counter whereas CondMutex does not suffer from this problem because the check is completely atomic with the wait/signal. I guess one of the issues here is that I moved on from your original code that was only dealing with a single sample onto this little buffer of samples that makes the concurrency more tricky.

andyp1per:
You are correct about the name - I have been using "attitude" and "rate" interchangeably when in fact it is rate. In my defence it was all inside a file called "Attitude.cpp". I have changed the name of the thread to reflect this.

tridge:
> Ok I have added an example that I think illustrates the problem

in that example I don't see why you expect missing to be zero. With the random sleeps it could easily be that thread1 calls signal several times while thread2 is sleeping. You only want thread2 to get one event in that case, so "sent - ops" with be non-zero with quite a high chance. 
Also, get_random() returns 0 one in 256 times, so that means one in 256 times thread1 sends multiple signals with no chance for thread2 to run at all as thread1 doesn't sleep.

tridge:
> I guess one of the issues here is that I moved on from your original code that was only dealing with a single sample onto this little buffer of samples that makes the concurrency more tricky.

but you do process that whole buffer at once when you wake up, yes? If so then if that buffer ever gets more than one entry in it then you don't want to wake up every time.

I've pushed a fix for your test and it works for BinarySem. 
https://github.com/tridge/ardupilot/commits/pr-copter-rate-thread/

andyp1per:
> > I guess one of the issues here is that I moved on from your original code that was only dealing with a single sample onto this little buffer of samples that makes the concurrency more tricky.
> 
> but you do process that whole buffer at once when you wake up, yes? If so then if that buffer ever gets more than one entry in it then you don't want to wake up every time.
> 
> I've pushed a fix for your test and it works for BinarySem. https://github.com/tridge/ardupilot/commits/pr-copter-rate-thread/

This ignores the concurrency issue I was alluding to - the update to bsize is not atomic with the wait/notification. It works because in the test the race is so narrow - but its there and is more obvious when you need to do something more complicated than simply update a counter.

andyp1per:
Now only compiled in on Copter, still some cost on other platforms due to the minor refactorings:

```
andy@Eagle:~/github/ardupilot$ ./Tools/scripts/pretty_diff_sym.py ../ardupilot-dev/build/MatekH743-bdshot/bin/arduplane ./build/MatekH743-bdshot/bin/arduplane | sort --key=2
Total: 628
  -8 AC_AttitudeControl::attitude_controller_run_quat() 508 -> 500
   4 AC_AttitudeControl::input_angle_step_bf_roll_pitch_yaw(float, 136 -> 140
   4 AC_AttitudeControl::input_euler_angle_roll_pitch_euler_rate_yaw(float, 420 -> 424
   4 AC_AttitudeControl::input_euler_angle_roll_pitch_yaw(float, 540 -> 544
   4 AC_AttitudeControl::input_euler_rate_roll_pitch_yaw(float, 384 -> 388
-190 AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_2(float, 212 -> 22
  16 AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_3(float, 472 -> 488
 228 AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_4(float, 0 -> 228
   4 AC_AttitudeControl::input_rate_bf_roll_pitch_yaw(float, 296 -> 300
 104 AC_AttitudeControl::input_rate_step_bf_roll_pitch_yaw(float, 0 -> 104
   4 AC_AttitudeControl::input_thrust_vector_heading(Vector3<float> 376 -> 380
  12 AC_AttitudeControl_Multi::AC_AttitudeControl_Multi(AP_AHRS_View&, 480 -> 492
-290 AC_AttitudeControl_Multi::rate_controller_run() 332 -> 42
 300 AC_AttitudeControl_Multi::rate_controller_run_dt(float, 0 -> 300
  44 AC_AttitudeControl_Multi::rate_controller_target_reset() 0 -> 44
   2 AC_AttitudeControl::rate_controller_run_dt(float, 0 -> 2
   2 AC_AttitudeControl::rate_controller_target_reset() 0 -> 2
   4 AC_AttitudeControl::relax_attitude_controllers() 176 -> 180
   4 AC_AttitudeControl_TS::input_euler_rate_yaw_euler_angle_pitch_bf_roll(bool, 556 -> 560
 -64 AC_AutoTune_Multi::twitch_test_run(AC_AutoTune::AxisType, 832 -> 768
   4 AP_AHRS_get_accel(lua_State*) 84 -> 88
   4 AP_AHRS::_get_primary_accel_index() 0 -> 4
  20 AP_AHRS_View::Write_Rate(AP_Motors 484 -> 504
   6 AP_AHRS::write_video_stabilisation() 186 -> 192
   2 AP_HAL::RCOutput::force_trigger_groups(bool) 0 -> 2
   2 AP_InertialSensor_Backend::apply_gyro_filters(unsigned 276 -> 278
  40 AP_InertialSensor_Backend::_publish_gyro(unsigned 126 -> 166
   2 AP_InertialSensor_Backend::set_primary_accel(unsigned 0 -> 2
   2 AP_InertialSensor_Backend::set_primary_gyro(unsigned 0 -> 2
  60 AP_InertialSensor_Backend::update_accel_filters(unsigned 0 -> 60
 -32 AP_InertialSensor_Backend::update_accel(unsigned 134 -> 102
 180 AP_InertialSensor_Backend::update_gyro_filters(unsigned 0 -> 180
-172 AP_InertialSensor_Backend::update_gyro(unsigned 274 -> 102
  48 AP_InertialSensor_Invensense::_read_fifo() 524 -> 572
   2 AP_InertialSensor_Invensense::set_primary_gyro(unsigned 0 -> 2
   4 AP_InertialSensor_Invensense::start() 924 -> 928
   2 AP_InertialSensor_Invensensev3::set_primary_gyro(unsigned 0 -> 2
   2 AP_InertialSensor::push_next_gyro_sample(unsigned 0 -> 2
   4 AP_InertialSensor::push_rate_loop_gyro(unsigned 0 -> 4
  24 AP_InertialSensor::update() 464 -> 488
  -4 AP_InertialSensor::write_notch_log_messages() 368 -> 364
  24 AP_Motors::add_motor_num(signed 26 -> 50
   4 AP_MotorsMulticopter::output() 82 -> 86
  62 AP_MotorsMulticopter::update_servo_map() 0 -> 62
   2 AP_Motors::rc_write(unsigned 72 -> 74
   8 AP_Vehicle::AP_Vehicle() 468 -> 476
   4 AP_Vehicle::one_Hz_update() 44 -> 48
   8 AP_Vehicle::setup() 664 -> 672
   8 AP_Vehicle::update_dynamic_notch_at_specified_rate() 92 -> 100
   4 ChibiOS::DeviceBus::bus_thread(void*) 188 -> 192
   4 ChibiOS::DeviceBus::DeviceBus(unsigned 44 -> 48
   8 ChibiOS::I2CDeviceManager::businfo 200 -> 208
   6 ChibiOS::I2CDevice::set_periodic_minimum(unsigned 0 -> 6
   6 ChibiOS::RCOutput::force_trigger_groups(bool) 0 -> 6
  -8 ChibiOS::RCOutput::push_local() 408 -> 400
   4 ChibiOS::RCOutput::set_dshot_rate(unsigned 136 -> 140
   8 ChibiOS::RCOutput::trigger_groups() 204 -> 212
   6 ChibiOS::SPIDevice::set_periodic_minimum(unsigned 0 -> 6
  20 chMtxLockS 120 -> 140
 -28 CSWTCH.368 28 -> 0
  28 CSWTCH.378 0 -> 28
   4 GCS_MAVLINK_Plane::send_attitude_target() 104 -> 108
  -4 GCS_MAVLINK_Plane::send_nav_controller_output() 648 -> 644
  -8 ModeAcro::stabilize() 484 -> 476
  -4 ModeLoiterAltQLand::_enter() 144 -> 140
  -4 ModeQLoiter::run() 928 -> 924
  -4 ModeQRTL::run() 748 -> 744
 -16 ModeThermal::exit_heading_aligned() 164 -> 148
   8 plane 37520 -> 37528
  -4 Plane::ahrs_update() 252 -> 248
  -4 Plane::apply_throttle_limits(float) 336 -> 332
   2 Plane::avoidance_adsb_update() 26 -> 28
   8 Plane::load_parameters() 928 -> 936
   4 Plane::Log_Write_AETR() 156 -> 160
   4 Plane::Log_Write_Attitude() 416 -> 420
   4 Plane::Log_Write_Control_Tuning() 232 -> 236
   4 Plane::Log_Write_Guided() 100 -> 104
   4 Plane::Log_Write_Nav_Tuning() 216 -> 220
   4 Plane::Log_Write_OFG_Guided() 148 -> 152
   2 Plane::Log_Write_RC() 48 -> 50
   4 Plane::Log_Write_Status() 148 -> 152
   4 Plane::Log_Write_Vehicle_Startup_Messages() 100 -> 104
   2 Plane::mode_from_mode_num(Mode::Number) 220 -> 222
  -4 Plane::notify_mode(Mode 48 -> 44
   4 Plane::one_second_loop() 520 -> 524
  -8 Plane::Plane() 1940 -> 1932
   8 Plane::set_takeoff_expected() 140 -> 148
   2 Plane::should_log(unsigned 8 -> 10
   4 Plane::stabilize_stick_mixing_fbw() 508 -> 512
   4 Plane::start_command(AP_Mission::Mission_Command 976 -> 980
  -4 Plane::takeoff_calc_throttle(bool) 148 -> 144
  -4 Plane::trigger_land_abort(float) 224 -> 220
  -4 Plane::update_alt() 444 -> 440
  -2 Plane::update_speed_height() 74 -> 72
   4 QuadPlane::Log_Write_QControl_Tuning() 384 -> 388
  -8 QuadPlane::verify_vtol_land() 688 -> 680
   8 rcoutDriver 672 -> 680
   8 SRV_Channels::calc_pwm() 212 -> 220
  44 SRV_Channels::output_pwm_chan(unsigned 0 -> 44
```

ArduCopter/Attitude.cpp
andyp1per: done

Tools/scripts/extract_features.py
```diff
@@ -196,7 +196,8 @@ def __init__(self, filename, nm="arm-none-eabi-nm", strings="strings"):
             ('AP_GPS_BLENDED_ENABLED', r'AP_GPS::calc_blend_weights\b',),
 
             ('HAL_WITH_DSP', r'AP_HAL::DSP::find_peaks\b',),
-            ('AP_INERTIALSENSOR_HARMONICNOTCH_ENABLED', r'AP_InertialSensor::HarmonicNotch::update_params\b',),
+            ('AP_INERTIALSENSOR_HARMONICNOTCH_ENABLED', r'AP_InertialSensor::get_next_gyro_sample\b',),
```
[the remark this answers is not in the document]
andyp1per: fixed

andyp1per:
Ok done another update to my example. I *think* it shows what I am talking about - and is at least consistent with what I was observing - noticeably that with binary semaphore you get some inconsistency and delay in processing samples:

BSem:
```
tick 6000 3932.320 ops/s missing 0 max_delay 278us
tick 7000 3931.796 ops/s missing -1 max_delay 280us
tick 8000 3932.532 ops/s missing -1 max_delay 277us
tick 9001 3929.536 ops/s missing -1 max_delay 277us
tick 10001 3937.906 ops/s missing 0 max_delay 277us
tick 11001 3931.693 ops/s missing -1 max_delay 277us
tick 12001 3930.709 ops/s missing -1 max_delay 280us
tick 13001 3928.525 ops/s missing -1 max_delay 329us
tick 14001 3929.309 ops/s missing -1 max_delay 277us
tick 15001 3931.921 ops/s missing -1 max_delay 278us
tick 16001 3933.913 ops/s missing -1 max_delay 276us
tick 17001 3930.422 ops/s missing -1 max_delay 278us
tick 18001 3932.618 ops/s missing 0 max_delay 277us
tick 19001 3934.658 ops/s missing 0 max_delay 276us
```
CondMutex:
```
tick 1000 3907.441 ops/s missing 0 max_delay 121us
tick 2000 3918.130 ops/s missing 0 max_delay 123us
tick 3000 3915.374 ops/s missing 0 max_delay 123us
tick 4000 3914.440 ops/s missing 0 max_delay 120us
tick 5000 3913.452 ops/s missing 0 max_delay 125us
tick 6000 3914.499 ops/s missing 0 max_delay 123us
tick 7001 3913.385 ops/s missing 0 max_delay 118us
tick 8001 3913.288 ops/s missing 0 max_delay 120us
tick 9001 3913.362 ops/s missing 0 max_delay 121us
```
This is running on an actual H7 with ChibiOS

andyp1per:
I think a good chunk of the size changes are alignment changes - not sure I can do anything about that

andyp1per:
Re-flown with latest changes - all good

tridge:
why the cost on plane on disables board?
image

libraries/AP_InertialSensor/AP_InertialSensor_Invensense.cpp
```diff
@@ -487,6 +488,16 @@ bool AP_InertialSensor_Invensense::update() /* front end */
     return true;
 }
 
+void AP_InertialSensor_Invensense::set_primary_gyro(uint8_t instance)
+{
+#if AP_INERTIALSENSOR_RATE_LOOP_WINDOW_ENABLED
+    if (!_imu.use_rate_loop_gyro_samples()) {
+        return;
+    }
+    _dev->set_periodic_minimum(instance == gyro_instance ? 0 : 100);
```
tridge, line 497: this seems really dangerous, the SPI thread could completely dominate the CPU

andyp1per:
@tridge I've split out the safe + necessary PR's for separate review to make the unsafe + scary stuff easier to review and control at compile time.

andyp1per:
This is now a 100byte cost on plane, which is mainly due to the small restructuring of INS:
```
antennatracker   104 (+0.0077%)   0 (0.0000%)  0 (0.0000%)    104 (+0.0077%)                                   621084
arducopter       3400 (+0.1877%)  0 (0.0000%)  0 (0.0000%)    3400 (+0.1872%)                                  146872
ardusub          104 (+0.0065%)   0 (0.0000%)  0 (0.0000%)    104 (+0.0065%)                                   359460
blimp            104 (+0.0076%)   0 (0.0000%)  0 (0.0000%)    104 (+0.0076%)                                   599032
arduplane        116 (+0.0064%)   0 (0.0000%)  4 (+0.0015%)   116 (+0.0064%)                                   152832
ardurover        108 (+0.0065%)   0 (0.0000%)  -4 (-0.0015%)  108 (+0.0065%)                                   302264
arducopter-heli  3404 (+0.1880%)  0 (0.0000%)  -4 (-0.0015%)  3404 (+0.1876%)                                  147768
```

libraries/AP_InertialSensor/FastRateBuffer.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+    delete fast_rate_buffer;
+    fast_rate_buffer = nullptr;
+}
+
+uint32_t AP_InertialSensor::get_num_gyro_samples()
+{
+    return fast_rate_buffer->get_num_gyro_samples();
+}
+
```
andyp1per, line 49: add comments

libraries/AP_InertialSensor/FastRateBuffer.cpp
```diff
@@ -0,0 +1,124 @@
...
+void AP_InertialSensor::disable_fast_rate_buffer()
+{
+    delete fast_rate_buffer;
+    fast_rate_buffer = nullptr;
+}
+
+uint32_t AP_InertialSensor::get_num_gyro_samples()
+{
+    return fast_rate_buffer->get_num_gyro_samples();
```
andyp1per, line 47: check for nullptr?

libraries/AP_InertialSensor/FastRateBuffer.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+    if (!use_rate_loop_gyro_samples()) {
+        return false;
+    }
+
+    if (_rate_loop_gyro_window.available() == 0) {
+        _notifier.wait_blocking();
+    }
+
+    WITH_SEMAPHORE(_mutex);
```
rmackay9, line 86: I always thought this had to be within curly braces or else it was equivalent to having it at the top of the method

libraries/AP_HAL_ChibiOS/RCOutput.h
```diff
@@ -579,6 +584,8 @@ class ChibiOS::RCOutput : public AP_HAL::RCOutput
     uint8_t _dshot_cycle;
     // virtual timer for post-push() pulses
     virtual_timer_t _dshot_rate_timer;
+    // force triggering of groups
```
andyp1per, line 587: "This is used by "

ArduCopter/Copter.cpp
```diff
@@ -259,6 +260,9 @@ const AP_Scheduler::Task Copter::scheduler_tasks[] = {
 #if HAL_BUTTON_ENABLED
     SCHED_TASK_CLASS(AP_Button,            &copter.button,              update,           5, 100, 168),
 #endif
+#if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
```
andyp1per, line 263: add a comment - "don't delete"

ArduCopter/Parameters.cpp
```diff
@@ -1232,6 +1233,22 @@ const AP_Param::GroupInfo ParametersG2::var_info2[] = {
...
+#if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
+    // @Param: FSTRATE_ENABLE
+    // @DisplayName: Enable the fast Rate thread
+    // @Description: Enable the fast Rate thread
+    // @User: Advanced
+    // @Values: 0:Disabled,1:Enabled-Dynamic,2:Enabled-FixedWhenArmed,3:Enabled-Fixed
+    AP_GROUPINFO("FSTRATE_ENABLE", 9, ParametersG2, att_enable, 0),
+
+    // @Param: FSTRATE_DIV
+    // @DisplayName: Fast rate thread divisor
```
andyp1per, line 1245: Is there an easier way?

andyp1per: add the value to the used value

ArduCopter/motors.cpp
```diff
@@ -181,7 +181,19 @@ void Copter::motors_output()
...
+        SRV_Channels::push();
+    } else {
+        hal.rcout->push();
+    }
+}
+
+// motors_output from main thread
+void Copter::motors_output_main()
+{
+    if (!using_rate_thread) {
```
andyp1per, line 194: how does this affect servos?

ArduCopter/motors.cpp
```diff
@@ -181,7 +181,19 @@ void Copter::motors_output()
     }
 
     // push all channels
-    SRV_Channels::push();
+    if (full_push) {
```
andyp1per, line 184: Needs a comment

rmackay9, line 184: It's a shame that the copter code needs to worry about the detail of which push() to call.  The copter code essentially knows that the SRV_Channels cannot do what we want it to.  It would be nicer if the SRV_Channel::push() call could figure that out itself somehow

ArduCopter/rate_thread.cpp
```diff
@@ -0,0 +1,501 @@
...
+                was_using_rate_thread = false;
+            }
+            hal.scheduler->delay_microseconds(500);
+            last_run_us = AP_HAL::micros();
+            continue;
+        }
+
+        // set up rate thread requirements
+        if (!using_rate_thread) {
+            enable_fast_rate_loop(rate_decimation, rates);
```
rmackay9, line 229: not a blocker, feel free to resolve immediately but we need to be careful that we don't somehow end up in a loop where we continuously delete and re-create the buffer.

ArduCopter/rate_thread.cpp
```diff
@@ -0,0 +1,501 @@
...
+// @LoggerMessage: RTDT
+// @Description: Attitude controller time deltas
+// @Field: TimeUS: Time since system startup
+// @Field: dt: current time delta
+// @Field: dtAvg: current time delta average
+// @Field: dtMax: Max time delta since last log output
+// @Field: dtMin: Min time delta since last log output
+
+        if (now_ms - last_rtdt_log_ms >= 10) {    // 100 Hz
+            AP::logger().WriteStreaming("RTDT", "TimeUS,dt,dtAvg,dtMax,dtMin", "Qffff",
```
andyp1per, line 264: Do this the fast way

andyp1per: Do we actually need this?

ArduCopter/rate_thread.cpp
```diff
@@ -0,0 +1,501 @@
...
+    }
+    rates.medium_logging_rate = calc_gyro_decimation(rate_decimation, 10);   // 10Hz
+#endif
+    rates.main_loop_rate = calc_gyro_decimation(rate_decimation, AP::scheduler().get_filtered_loop_rate_hz());
+    rates.filter_rate = calc_gyro_decimation(rate_decimation, ins.get_raw_gyro_rate_hz() / 2);
+
+    return 0;
+}
+
+void Copter::enable_fast_rate_loop(uint8_t rate_decimation, RateControllerRates& rates)
```
andyp1per, line 456: add some comments

ArduCopter/rate_thread.cpp
```diff
@@ -0,0 +1,501 @@
...
+#endif
+
+        // immediately output the new motor values
+        if (run_decimated_callback(rates.main_loop_rate, main_loop_count)) {
+            main_loop_count = 0;
+        }
+        motors_output(main_loop_count == 0);
+
+        // process filter updates
+        if (run_decimated_callback(rates.filter_rate, filter_loop_count)) {
```
andyp1per, line 301: Should this be called before the motors output or attitude controller?

libraries/AP_InertialSensor/FastRateBuffer.cpp
andyp1per: Fixed

libraries/AP_InertialSensor/FastRateBuffer.cpp
andyp1per: Fixed

libraries/AP_HAL_ChibiOS/RCOutput.cpp
```diff
@@ -1398,7 +1398,8 @@ void RCOutput::trigger_groups()
     osalSysUnlock();
 #if !defined(HAL_NO_RCOUT_THREAD) || HAL_DSHOT_ENABLED
     // trigger a PWM send
-    if (!in_soft_serial() && hal.scheduler->in_main_thread() && rcout_thread_ctx) {
+    if (!in_soft_serial() &&
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_HAL_ChibiOS/RCOutput.h
andyp1per: Fixed

ArduCopter/Copter.cpp
andyp1per: Fixed

ArduCopter/Parameters.cpp
andyp1per: Fixed docs. Leave config as is

ArduCopter/motors.cpp
andyp1per: Servos not affected

ArduCopter/motors.cpp
andyp1per: Fixed

andyp1per: Left as-is

ArduCopter/rate_thread.cpp
andyp1per: Fixed

ArduCopter/rate_thread.cpp
andyp1per: Reduced rate to 10Hz and done the fast way

ArduCopter/rate_thread.cpp
andyp1per: Fixed

ArduCopter/rate_thread.cpp
andyp1per: called from the right place

peterbarker:
Ping @IamPete1 - good one to review while jet-lagged!

libraries/AP_InertialSensor/FastRateBuffer.h
```diff
@@ -0,0 +1,51 @@
...
+    bool use_rate_loop_gyro_samples() const { return rate_decimation > 0; }
+    bool gyro_samples_available() { return  _rate_loop_gyro_window.available() > 0; }
+    void reset();
+
+private:
+    /*
+      binary semaphore for rate loop to use to start a rate loop when
+      we hav finished filtering the primary IMU
+     */
+    ObjectBuffer<Vector3f> _rate_loop_gyro_window{AP_INERTIAL_SENSOR_RATE_LOOP_BUFFER_SIZE};
```
IamPete1, line 45: Why do we need a buffer? I would have thought you could just keep the latest value and count how many times you update it, then zero the count when you read it.

libraries/AP_InertialSensor/FastRateBuffer.h
```diff
@@ -0,0 +1,51 @@
...
+    uint8_t rate_decimation; // 0 means off
+    uint8_t rate_decimation_count;
+    HAL_BinarySemaphore _notifier;
+    HAL_Semaphore _mutex;
+};
+#endif
```
IamPete1, line 51:
```suggestion
#endif

```

IamPete1 (commented):
Have you tested turtle mode? It twiddles with stuff in motors.

Not a 100% review, still jet-lagged.

IamPete1:
Has any testing been done without DShot? Does PWM still work I guess at it would still be at 400Hz. Do the DroneCAN ESC commands get sent faster with the higher update rate. Do the none motor outputs still work as expected (eg servo gimbal)?

ArduCopter/rate_thread.cpp
```diff
@@ -0,0 +1,493 @@
...
+        rates.fast_logging_rate = calc_gyro_decimation(rate_decimation, 1000);   // 1Khz
+    } else {
+         rates.fast_logging_rate = calc_gyro_decimation(rate_decimation, AP::scheduler().get_filtered_loop_rate_hz());
+    }
+    rates.medium_logging_rate = calc_gyro_decimation(rate_decimation, 10);   // 10Hz
+#endif
+    rates.main_loop_rate = calc_gyro_decimation(rate_decimation, AP::scheduler().get_filtered_loop_rate_hz());
+    rates.filter_rate = calc_gyro_decimation(rate_decimation, ins.get_raw_gyro_rate_hz() / 2);
+
+    return 0;
```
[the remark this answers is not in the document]
andyp1per: Removed

ArduCopter/rate_thread.cpp
```diff
@@ -0,0 +1,493 @@
...
+#endif
+
+        // we must use multiples of the actual sensor rate
+        const float sensor_dt = 1.0f * rate_decimation / ins.get_raw_gyro_rate_hz();
+        const uint32_t now_us = AP_HAL::micros();
+        const uint32_t dt_us = now_us - last_run_us;
+        const float dt = dt_us * 1.0e-6;
+        last_run_us = now_us;
+
+        motors->set_dt(sensor_dt);
```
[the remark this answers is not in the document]
andyp1per: I've moved it all to rate_controller_set_rates

libraries/AP_InertialSensor/AP_InertialSensor_Backend.h
```diff
@@ -53,6 +53,15 @@ class AP_InertialSensor_Backend
      */
     virtual bool update() = 0; /* front end */
 
+    // if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
```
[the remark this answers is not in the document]
andyp1per: Yes, because of the vehicle include protections it can't actually be defined here, the comment is just to indicate that the code is technically controlled

libraries/AP_InertialSensor/AP_InertialSensor.h
```diff
@@ -798,6 +803,29 @@ class AP_InertialSensor : AP_AccelCal_Client
     bool raw_logging_option_set(RAW_LOGGING_OPTION option) const {
         return (raw_logging_options.get() & int32_t(option)) != 0;
     }
+    // if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
```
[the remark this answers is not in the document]
andyp1per: Yes, because of the vehicle include protections it can't actually be defined here, the comment is just to indicate that the code is technically controlled

ArduCopter/Parameters.cpp
```diff
@@ -1232,6 +1233,22 @@ const AP_Param::GroupInfo ParametersG2::var_info2[] = {
...
+    // @Description: Fast rate thread divisor used to control the maximum fast rate update rate. The actual rate is the gyro rate in Hz divided by this value. This value is scaled depending on the configuration of FSTRATE_ENABLE.
+    // @User: Advanced
+    // @Range: 1 10
+    AP_GROUPINFO("FSTRATE_DIV", 10, ParametersG2, att_decimation, 1),
```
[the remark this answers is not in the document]
andyp1per: Yes, it will be downscaled on overload

ArduCopter/rate_thread.cpp
```diff
@@ -0,0 +1,493 @@
...
+
+/*
+  update rate controller rates and return the logging rate
+*/
+uint8_t Copter::rate_controller_set_rates(uint8_t rate_decimation, RateControllerRates& rates, bool warn_cpu_high)
+{
+    const uint32_t attitude_rate = ins.get_raw_gyro_rate_hz() / rate_decimation;
+    attitude_control->set_notch_sample_rate(attitude_rate);
+    hal.rcout->set_dshot_rate(SRV_Channels::get_dshot_rate(), attitude_rate);
+    motors->set_dt(1.0f / attitude_rate);
```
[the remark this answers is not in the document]
andyp1per: Moved it here

libraries/AP_AHRS/AP_AHRS.h
```diff
@@ -620,7 +620,7 @@ class AP_AHRS {
 
     // return primary accels, for lua
     const Vector3f &get_accel(void) const {
-        return AP::ins().get_accel();
+        return AP::ins().get_accel(_get_primary_accel_index());
```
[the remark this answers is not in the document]
andyp1per: I can remove it. It is correct however. What would you prefer?

libraries/AP_HAL_ChibiOS/hwdef/MambaH743v4/hwdef.dat
```diff
@@ -20,6 +20,7 @@ define CH_CFG_ST_RESOLUTION 16
 
 # leave 1 sectors free
 FLASH_RESERVE_START_KB 384
+env OPTIMIZE -O2
```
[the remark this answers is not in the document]
andyp1per: I think so, the PR is quite CPU intensive and I have been testing on this board a lot - this gives more headroom

libraries/AP_InertialSensor/AP_InertialSensor.h
```diff
@@ -9,6 +9,8 @@
 #define AP_INERTIAL_SENSOR_ACCEL_VIBE_FILT_HZ           2.0f    // accel vibration filter hz
 #define AP_INERTIAL_SENSOR_ACCEL_PEAK_DETECT_TIMEOUT_MS 500     // peak-hold detector timeout
 
+#define AP_INERTIAL_SENSOR_RATE_LOOP_BUFFER_SIZE 8     // gyro buffer size for rate loop
```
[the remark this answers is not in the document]
andyp1per: Removed

ArduCopter/Copter.h
```diff
@@ -725,7 +736,25 @@ class Copter : public AP_Vehicle {
     void set_accel_throttle_I_from_pilot_throttle();
     void rotate_body_frame_to_NE(float &x, float &y);
     uint16_t get_pilot_speed_dn() const;
-    void run_rate_controller();
+    void run_rate_controller_main();
+
+    // if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
```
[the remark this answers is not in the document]
andyp1per: Yes, because of the vehicle include protections it can't actually be defined here, the comment is just to indicate that the code is technically controlled

libraries/AP_InertialSensor/FastRateBuffer.h
andyp1per: Still need to process a regular number of samples in the rate controller. I mean you could try and count and interpolate, but its simpler this way and simpler to understand without further changes to the rate controller code. Also in general the buffer only ever contains one item. We could consider reducing the size of the buffer.

libraries/AP_InertialSensor/FastRateBuffer.h
andyp1per: Fixed

andyp1per:
> Has any testing been done without DShot? Does PWM still work I guess at it would still be at 400Hz. Do the DroneCAN ESC commands get sent faster with the higher update rate. Do the none motor outputs still work as expected (eg servo gimbal)?

I will need to do more testing. I believe this is all fine (including turtle mode), but will test further.

libraries/AP_AHRS/AP_AHRS.h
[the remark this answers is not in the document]
IamPete1: A separate PR would be great, thanks.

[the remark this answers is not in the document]
andyp1per: Ok done

ArduCopter/motors.cpp
```diff
@@ -181,7 +182,21 @@ void Copter::motors_output()
...
+        // motor output including servos and other updates that need to run at the main loop rate
+        SRV_Channels::push();
```
tridge, line 187: do we end up with an unpaired cork/push?

tridge (commented):
some change?

tridge:
i'm happy with the system level and the AP_InertialSensor and AP_Vehicle changes, so from the point of view of plane which doesn't use this I'm happy with this PR
I have not properly reviewed the copter specific changes

ArduCopter/motors.cpp
andyp1per: No, all checks out. I have added in the check however.

libraries/AP_InertialSensor/FastRateBuffer.cpp
```diff
@@ -0,0 +1,147 @@
...
+    if (fast_rate_buffer == nullptr) {
+        return;
+    }
+    fast_rate_buffer->set_rate_decimation(rdec);
+}
+
+// whether or not to push the current gyro sample
+bool AP_InertialSensor::is_rate_loop_gyro_enabled(uint8_t instance) const
+{
+    return fast_rate_buffer_enabled && fast_rate_buffer->use_rate_loop_gyro_samples() && instance == AP::ahrs().get_primary_gyro_index();
```
IamPete1, line 76: Should `fast_rate_buffer != nullptr` be in here?

libraries/AP_InertialSensor/FastRateBuffer.cpp
```diff
@@ -0,0 +1,147 @@
...
+// enable the fast rate buffer and start pushing samples to it
+void AP_InertialSensor::enable_fast_rate_buffer()
+{
+    if (fast_rate_buffer_enabled) {
+        return;
+    }
+    if (fast_rate_buffer == nullptr) {
+        fast_rate_buffer = NEW_NOTHROW FastRateBuffer();
+    }
+    fast_rate_buffer_enabled = true;
```
IamPete1, line 43: Should we check the buffer was allocated before enabling ?
```suggestion
    fast_rate_buffer_enabled = fast_rate_buffer != nullptr;
```

libraries/AP_InertialSensor/FastRateBuffer.cpp
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// get the next available gyro sample from the fast rate buffer
+bool AP_InertialSensor::get_next_gyro_sample(Vector3f& gyro)
+{
+    if (!fast_rate_buffer_enabled) {
+        return false;
+    }
+
+    return fast_rate_buffer->get_next_gyro_sample(gyro);
```
IamPete1, line 86: nulprt check?

andyp1per: Fixed

libraries/AP_InertialSensor/FastRateBuffer.cpp
andyp1per: Fixed

libraries/AP_InertialSensor/FastRateBuffer.cpp
andyp1per: fixed

ArduCopter/Copter.cpp
```diff
@@ -741,11 +754,24 @@ void Copter::one_hz_loop()
...
 #if AC_CUSTOMCONTROL_MULTI_ENABLED
     custom_control.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
 #endif
+
+#if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
+    // see if we should have a separate rate thread
+    if (!started_rate_thread && get_fast_rate_type() != FastRateType::FAST_RATE_DISABLED) {
+        if (hal.scheduler->thread_create(FUNCTOR_BIND_MEMBER(&Copter::rate_controller_thread, void),
+                                         "rate",
+                                         1536, AP_HAL::Scheduler::PRIORITY_RCOUT, 1)) {
```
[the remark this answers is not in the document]
andyp1per: Its still memory. This was the value I needed that gave a decent amout of headroom

ArduCopter/Copter.cpp
```diff
@@ -741,11 +754,24 @@ void Copter::one_hz_loop()
...
+    }
     pos_control->get_accel_z_pid().set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
 #if AC_CUSTOMCONTROL_MULTI_ENABLED
     custom_control.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
 #endif
+
+#if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
+    // see if we should have a separate rate thread
+    if (!started_rate_thread && get_fast_rate_type() != FastRateType::FAST_RATE_DISABLED) {
+        if (hal.scheduler->thread_create(FUNCTOR_BIND_MEMBER(&Copter::rate_controller_thread, void),
```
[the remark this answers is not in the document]
andyp1per: Fixed

ArduCopter/Attitude.cpp
```diff
@@ -4,18 +4,21 @@
...
-    attitude_control->set_dt(last_loop_time_s);
     pos_control->set_dt(last_loop_time_s);
+    attitude_control->set_dt(last_loop_time_s);
 
-    // run low level rate controllers that only require IMU data
-    attitude_control->rate_controller_run();
+    if (!using_rate_thread) {
+        motors->set_dt(last_loop_time_s);
+        // only run the rate controller if we are not using the rate thread
+        attitude_control->rate_controller_run();
```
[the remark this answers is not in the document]
andyp1per: I don't understand the comment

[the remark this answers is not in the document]
andyp1per: This was addressed by the rate control threading PR

ArduCopter/Parameters.cpp
```diff
@@ -1232,6 +1233,22 @@ const AP_Param::GroupInfo ParametersG2::var_info2[] = {
     // @User: Advanced
     AP_GROUPINFO("FS_EKF_FILT", 8, ParametersG2, fs_ekf_filt_hz, FS_EKF_FILT_DEFAULT),
 
+#if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
+    // @Param: FSTRATE_ENABLE
+    // @DisplayName: Enable the fast Rate thread
+    // @Description: Enable the fast Rate thread. In the default case the fast rate divisor, which controls the update frequency of the thread, is dynamically scaled from FSTRATE_DIV to avoid overrun in the gyro sample buffer and main loop slow-downs. Other values can be selected to fix the divisor to FSTRATE_DIV on arming or always.
+    // @User: Advanced
+    // @Values: 0:Disabled,1:Enabled-Dynamic,2:Enabled-FixedWhenArmed,3:Enabled-Fixed
+    AP_GROUPINFO("FSTRATE_ENABLE", 9, ParametersG2, att_enable, 0),
```
IamPete1, line 1242: Do you forsee more params here in the future? It would be nice to make this a enable in a little sub tree.

ArduCopter/Copter.h
```diff
@@ -631,6 +631,17 @@ class Copter : public AP_Vehicle {
...
+
+    // type of fast rate attitude controller in operation
+    enum class FastRateType : uint8_t {
+        FAST_RATE_DISABLED            = 0,
+        FAST_RATE_DYNAMIC             = 1,
+        FAST_RATE_FIXED_ARMED         = 2,
+        FAST_RATE_FIXED               = 3,
+    };
+
+    FastRateType get_fast_rate_type() const { return FastRateType(g2.att_enable.get()); }
```
IamPete1, line 643: What happens if I set a out of range value? Lots of places are checking `!= FAST_RATE_DISABLED` what would a values of -1 result in? Unfortunately we have a presdent that -1 is the same as disabled for lots of params.

IamPete1 (approved):
I think this is as good enough to go in, I can't see any more issues from review, just need to start getting wider testing i think.

rafolg:
should HAL_INS_RATE_LOOP be INS_GYRO_RATE instead? I don't see where HAL_INS_RATE_LOOP is used?

rmackay9 (approved):
TBH, I have not reviewed this too thoroughly but I'm happy based on the review of others for this to go in

tridge (approved):

PR #27030, opened 2024-05-09
AP_BattMonitor: ESC: add mask parameter

IamPete1:
This adds a `ESC_MASK` parameter to allow assigning particular ESCs to particular batteries. Default is 0 for use all which is the current behaviour.

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
@@ -20,6 +20,32 @@
 
 #include "AP_BattMonitor_ESC.h"
 
+const AP_Param::GroupInfo AP_BattMonitor_ESC::var_info[] = {
+
+    // Param indexes must be between 36 and 39 to avoid conflict with other battery monitor param tables loaded by pointer
+
+    // @Param: ESC_MASK
+    // @DisplayName: ESC mask
+    // @Description: If 0 all connected ESCs will be used. If non-zero, only those selected in will be used.
+    // @Bitmask: 0: ESC 1, 1: ESC 2, 2: ESC 3, 3: ESC 4, 4: ESC 5, 5: ESC 6, 6: ESC 7, 7: ESC 8, 8: ESC 9, 9: ESC 10, 10: ESC 11, 11: ESC 12, 12: ESC 13, 13: ESC 14, 14: ESC 15, 15: ESC 16, 16: ESC 17, 17: ESC 18, 18: ESC 19, 19: ESC 20, 20: ESC 21, 21: ESC 22, 22: ESC 23, 23: ESC 24, 24: ESC 25, 25: ESC 26, 26: ESC 27, 27: ESC 28, 28: ESC 29, 29: ESC 30, 30: ESC 31, 31: ESC 32
+    // @User: Standard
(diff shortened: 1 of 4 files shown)


libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
```diff
@@ -20,6 +20,33 @@
 
 #include "AP_BattMonitor_ESC.h"
 
+const AP_Param::GroupInfo AP_BattMonitor_ESC::var_info[] = {
+
+    // Param indexes must be between 36 and 39 to avoid conflict with other battery monitor param tables loaded by pointer
+
+    // @Param: ESC_MASK
+    // @DisplayName: ESC mask
+    // @Description: 0: Use all connected ESC's if none zero use those set in mask
```
magicrub, line 29: This 0: at the beginning is weird, and the rest doesn't quite make sense. Maybe add a period or a comma somewhere?

libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
```diff
@@ -20,6 +20,33 @@
...
+    // @Bitmask: 0: ESC 1, 1: ESC 2, 2: ESC 3, 3: ESC 4, 4: ESC 5, 5: ESC 6, 6: ESC 7, 7: ESC 8, 8: ESC 9, 9: ESC 10, 10: ESC 11, 11: ESC 12, 12: ESC 13, 13: ESC 14, 14: ESC 15, 15: ESC 16, 16: ESC 17, 17: ESC 18, 18: ESC 19, 19: ESC 20, 20: ESC 21, 21: ESC 22, 22: ESC 23, 23: ESC 24, 24: ESC 25, 25: ESC 26, 26: ESC 27, 27: ESC 28, 28: ESC 29, 29: ESC 30, 30: ESC 31, 31: ESC 32
```
magicrub, line 30: I know we go up to 32 bits on here and likely elsewhere but in reality we should only go up to 24 due to float-integer roundoff error.

magicrub (changes requested):

libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
IamPete1: I have re-worded .

libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
IamPete1: We use the full 32 bits for all servo stuff.

libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
```diff
@@ -20,6 +20,32 @@
 
 #include "AP_BattMonitor_ESC.h"
 
+const AP_Param::GroupInfo AP_BattMonitor_ESC::var_info[] = {
+
+    // Param indexes must be between 36 and 39 to avoid conflict with other battery monitor param tables loaded by pointer
+
+    // @Param: ESC_MASK
+    // @DisplayName: ESC mask
+    // @Description: If 0 all connected ESCs will be used. If none zero only those selected in will be used.
```
magicrub, line 29:
```suggestion
    // @Description: If 0 all connected ESCs will be used. If non-zero, only those selected in will be used.
```

tridge (approved):

PR #27039, opened 2024-05-11
AP_ExternalAHRS: Fix duplicate condition

Ryanf55:
Fix duplicate logic in the pre-arm check of microstrain7. Filter state was checked above.


IamPete1 (approved):
LGTM

tridge (approved):

PR #27040, opened 2024-05-11
waf: move setting of ROMFS flags to embed step

IamPete1:
Currently on ChibiOS these flags are set before files are added from `ROMFS_custom` so the flags may be incorrect. This moves the setting of the flags to the same step as where the files are imbedded so nothing is missed. This also has the advantage of unifying the approach across all boards. 

I have tested imbedding a script in SITL, I have built a ChibiOS board and grepped the bin to see that the script is embedded correctly, but I have not had a chance to flash to a board to double check.


IamPete1:
Tested on CubeOrangePlus, added script to ROMFS via hwdef folder and ROMFS_custom.

peterbarker (approved):

PR #27042, opened 2024-05-11
AP_Param: Summarize the type definitions settings

muramura:
I don't think it's necessary to separate type definitions and settings.


magicrub (approved):
I'm approving this but would also like to run it by the devcall because there might be a good reason for this that I'm missing

tridge (approved):

PR #27043, opened 2024-05-11
global: Change the IF statement to a SWITCH statement

muramura:
In this case, where the same variable value is evaluated multiple times in IF statements, 
I think it would be better to use a SWITCH statement.

### DIFF
--- a/libraries/AP_Param/AP_Param.cpp
+++ b/libraries/AP_Param/AP_Param.cpp
@@ -2205,23 +2205,33 @@ void AP_Param::set_float(float value, enum ap_var_type var_type)
     float rounding_addition = 0.01f;
         
     // handle variables with standard type IDs
-    if (var_type == AP_PARAM_FLOAT) {
+    switch (var_type) {
+    case AP_PARAM_FLOAT:
         ((AP_Float *)this)->set(value);
-    } else if (var_type == AP_PARAM_INT32) {
+        break;
+    case AP_PARAM_INT32: {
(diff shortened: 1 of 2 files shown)


libraries/AP_Param/AP_Param.cpp
```diff
@@ -2205,23 +2205,33 @@ void AP_Param::set_float(float value, enum ap_var_type var_type)
...
+        break;
+    }
+    case AP_PARAM_INT8: {
         if (value < 0) rounding_addition = -rounding_addition;
         float v = value+rounding_addition;
         v = constrain_float(v, INT8_MIN, INT8_MAX);
         ((AP_Int8 *)this)->set(v);
+        break;
+    }
+    default:
```
magicrub, line 2233: Don't need this default. Also, if we change var_class to be an enum class then we'll NEED to remove the default.

magicrub (changes requested):

peterbarker:
Needs a rebase (Rover changes already merged), and Tom's review comment addressed.

@muramura are you still wanting to chase this?

@magicrub I had a quick play with moving from `ap_var_type` to `enum class AP_Param::`VarType`.  The changes are massive as we use `AP_PARAM_FLOAT` and friends in huge amounts of code.  The conversion tables are possibly also slightly more difficult to read.  I pushed that incomplete branch here: https://github.com/peterbarker/ardupilot/pull/new/archive/pr/vartype-enum-class - but I don't think I'll try to push that forward at this point.

PR #27045, opened 2024-05-12
LD06: Record only 2deg slices

mw46d:
* Some recording changes were taken from the RPLidar driver code. 
* Using the average of the 12 readings/package is not a good idea. Those packages don't describe the same 'angle' for each rotation. So you end up with a lot of different distances per package. I tried to fix things to well-defined 2deg slices and take the minimum reading reported.


peterbarker (approved):
OK, this improves the driver out-of-sight - thanks for this!

I've just tested this on an LD06 locally, and it makes the driver actually usable!

I've taken the liberty of adjusting a few patches in here to conform to our style guidelines, and adjusted the message on your commit to also conform to our requirements.

I've also adjusted the handling of a bad size received to avoid the nasty buffer overflow, rebasing on a previous PR of mine.

Thanks!

@lvale you said you had some of these.  I've no idea how they're at all useful for you before this PR :-)

peterbarker:
I've created a simulator for this device now.  And corrected the sanity check again.

libraries/SITL/SIM_PS_LD06.cpp
```diff
@@ -0,0 +1,162 @@
...
+    // sanity check we're being called often enough:
+    if (time_delta_ms / samples_per_ms > sample_count) {
+        AP_HAL::panic("Not being called often enough");
+    }
+
+    const float degrees_per_s = 2152;  // see example datasheet page 8
+    const float degrees_per_ms = degrees_per_s / 1000.0f;
+    const float degrees_per_sample = degrees_per_ms / samples_per_ms;
+
+    // ::fprintf(stderr, "Packing %u samples in for %ums interval (%f degrees/sample)\n", sample_count, time_delta_ms, degrees_per_sample);
```
rmackay9, line 78: let's get rid of this debug

rmackay9:
@Adithya-Patil, are you OK with this?  Any comments on the orientation?  The wiki docs say the wire should be facing backwards

mw46d:
> @Adithya-Patil, are you OK with this? Any comments on the orientation? The wiki docs say the wire should be facing backwards

That's how my sensor is oriented as well;-)

rmackay9:
Hi @mw46d,

Ah, so you've got one of these sensors?  Do you find that it doesn't actually work with the code in 4.5, 4.6 and/or master?

mw46d:
That's why I started that PR;-)

rmackay9:

rmackay9:
Hi @mw46d, @peterbarker,

We've got some complaints during beta testing regarding this change https://discuss.ardupilot.org/t/looking-to-get-firmware-release-4-6-0-beta2-11-dec-2024/133422/7

rmackay9:
It's a long time ago now but let's remember to add the "Wiki Needed" label when we add new simulators https://github.com/ArduPilot/ardupilot_wiki/pull/6883

PR #27047, opened 2024-05-12
Gps ground course in NMEA and GSOF was incorrect

Ryanf55:
# Purpose

Both the NMEA and GSOF simulators have broken ground course in SITL. 

## Steps to reproduce

Set these params for GSOF against UBX:
```
GPS2_TYPE  11
SIM_GPS2_TYPE    11
```

Or these for NMEA
```
GPS2_TYPE  5
SIM_GPS2_TYPE  5
```

Then, start SITL with plane, arm, and takeoff.

## Before this PR

Here's the GPS course over ground for Ublox (GPS 1) and GSOF (GPS 2).

image

## With the fixed math on ground course, but no pi-wrapping (intermediate):
image

## After this PR

GSOF:
image

NMEA:
image

## Question

Is it bad you can have a flyaway if you don't wrap the GPS ground course from 0-360?

### DIFF
--- a/libraries/AP_GPS/AP_GPS.h
+++ b/libraries/AP_GPS/AP_GPS.h
@@ -197,7 +197,7 @@ public:
         uint16_t time_week;                 ///< GPS week number
         Location location;                  ///< last fix location
         float ground_speed;                 ///< ground speed in m/s
-        float ground_course;                ///< ground course in degrees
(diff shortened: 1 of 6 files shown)


tridge (approved):

rmackay9:

PR #27049, opened 2024-05-12
Rover: Determine the prearm check results

muramura:
The results of the first pre-arm check have not been verified.
It is better to judge the pre-arm check result and check the next pre-arm check.


khancyr (commented):
no, in this case that change in wrong. 
We first test the motor configuration through the arming check. 
Whatever the result is, if we got Torqeedo esc, we test it. 

With your change we won't report Torqeedo start until we fix first the motor configuration. 

So the current way is better as we report as much as we can in a single loop. This is possible as the check independant

khancyr:
Thanks anyway, but I will close this one as that isn't a good change.

PR #27051, opened 2024-05-12
SITL: Move the ENUM definition to the header file

muramura:
ENUM definitions are not processes. There is a header file. 
I think it would be better to move the definitions to the header file.

### DIFF
--- a/libraries/SITL/SIM_GPS_Trimble.cpp
+++ b/libraries/SITL/SIM_GPS_Trimble.cpp
@@ -31,17 +31,6 @@ void GPS_Trimble::publish(const GPS_Data *d)
             // TODO magic number until SITL supports GPS bootcount based on GPSN_ENABLE
             const uint8_t bootcount = 17;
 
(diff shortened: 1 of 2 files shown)


amilcarlucas:
This removes the benefit of encapsulation.

What is the advantage of this change?

muramura:
@amilcarlucas san. 
I have defined the ENUM privately. 
I think there are no benefits to defining the ENUM in the processing section. 
You can obtain the definition through the editor.

khancyr:
@Ryanf55  @peterbarker I think you own this driver. So just need a decision here. I don't think it change anything, so mostly code style

Ryanf55:
> @amilcarlucas san. I have defined the ENUM privately. I think there are no benefits to defining the ENUM in the processing section. You can obtain the definition through the editor.

I agree, the benefit is encapsulation. These enums are an implementation detail, intended to be hidden from the higher level simulation code. I do not agree with moving the code out of an implementation detail. That would violate the interface segregation principal of SOLID design principal.

khancyr:
Then close this PR is !Thanks Muramura-san for bring this.

PR #27052, opened 2024-05-12
AC_AttitudeControl_Heli: passthrough_bf_roll_pitch_rate_yaw use quaterion to update attitude target

IamPete1:
This is a update to remove lots of uses of `_euler_angle_target` this does two things:

- Switching to use quaternions means we don't need to worry about euler wrap handling
- This keeps the quaternion attitude target `_attitude_target` up to date rather than just the euler one `_euler_angle_target`.

As far as I can tell this should make no difference because neither attitude targets are actually used for anything other than logging while in roll/pitch flybar passthrough.


MattKear:
I am fairly sure that this won't change the motor outputs, but it would be nice to confirm with the heli motors comparison tool

lthall (approved):

IamPete1:
Thanks to https://github.com/ArduPilot/ardupilot/pull/32430 this function is now gone so we don't need to fix it.

PR #27059, opened 2024-05-14
AP_Proximity: prevent buffer overflow in LD06 driver

peterbarker:
We're using a value off the wire before it has been validated.  That value is used to limit indexing into a buffer, and that buffer isn't big enough to handle all possible "bad" values that index could take on.  Note that "read" here returns int16_t....

I've tried to simply make minimal changes in this driver to avoid the buffer overrun.

### DIFF
--- a/libraries/AP_Proximity/AP_Proximity_LD06.cpp
+++ b/libraries/AP_Proximity/AP_Proximity_LD06.cpp
@@ -92,21 +92,33 @@ void AP_Proximity_LD06::get_readings()
     // Loops through all bytes that were received
     while (nbytes-- > 0) {
 
-        // Gets and logs the current byte being read
-        const uint8_t c = _uart->read();
+        uint8_t c;
+        if (!_uart->read(c)) {
+            break;
+        }
 
         // Stores the byte in an array if the byte is a start byte or we have already read a start byte
-        if (c == LD_START_CHAR || _response_data) {
-
-            // Sets to true if a start byte has been read, default false otherwise
-            _response_data = true;
-
+        if (c == LD_START_CHAR || _byte_count) {
             // Stores the next byte in an array
             _response[_byte_count] = c;
+            if (_byte_count < START_DATA_LENGTH) {
+                _byte_count++;
+                continue;
+            }
+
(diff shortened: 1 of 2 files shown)


tridge:
@peterbarker if this PR can't be easily tested then we could do a much simpler patch that just checks the array bounds before adding data. That would prefer the buffer overrun while being a minimal change

rishabsingh3003:
Unfortunately, I dont not have the sensor either so I cannot test. Thanks @peterbarker for getting this fixed! 
@lvale I remember you had helped test this driver when we had this merged, would it be possible for you to test it?

burgeruser:
Hello, I have an LD06 laser radar that I purchased early this year. Unfortunately, its performance is subpar; it has a short range of only 12 meters, and the measurement noise is significant. As a result, the SLAM mapping using Cartographer algorithm does not yield satisfactory results.

Unfortunately, the model has become obsolete since it was quite niche. There are now many cheaper, lighter, and more accurate alternatives available in the market. In my testing with the LD06 on the developer version of Ardupilot4.5.0, directly connecting it to the flight controller for obstacle avoidance proved unreliable. The original 'AP_Proximity_LD06.cpp' file, which handles UART communication, heavily utilizes the STM32 microcontroller's RAM and lacks redundancy. In case of a physical UART cable error, it fails to parse data properly. This file tends to cause issues on STM32F4 platforms but runs smoothly on STM32H7 ones.

lvale:
@peterbarker 
sorry for coming late to the discussion. I have a pair of these working fine on Rover. What do you want to get tested ?

rmackay9:
Hi @lvale, thanks very much!  We would just like to confirm that the driver hasn't been broken by these changes so we'd like it if you could just do a quick test to confirm that it still appears to work.  If you ping me the name of the autopilot you're using we can create a firmware to load on the autopilot for the test.

peterbarker:
I've now tested this on a real LD06 and there's no change in behaviour that I can discern.

Unplugged/replugged a few times live, no problems there, seems to resync just fine.

rmackay9 (approved):

rmackay9:

PR #27063, opened 2024-05-14
AC_AutoTune: Convert the IF statement to a SWITCH statement

muramura:
It can be replaced with a SWITCH statement.


peterbarker:
Requires a rebase.

@muramura did you still want to chase this?

peterbarker:
Sorry, I've just had a quick look at this and don't think we should merge this one:
 - this should be enum-class
 - using switch-with-a-default makes this less useful a patch when there are only 2 of the many cases handled in the switch.

Closing this one.

PR #27065, opened 2024-05-14
AP_Scripting: added pullup or pulldown to pinMode()

pkocmoud:
This is a non-breaking change for existing scripts. 
The pin_types are: 
0 - input (floating)
1 - output
2 - input (pullup enabled)
3 - input (pulldown enabled)

### DIFF
--- a/libraries/AP_HAL_ChibiOS/GPIO.cpp
+++ b/libraries/AP_HAL_ChibiOS/GPIO.cpp
@@ -209,17 +209,30 @@ ioline_t GPIO::resolve_alt_config(ioline_t base, PERIPH_TYPE ptype, uint8_t inst
 }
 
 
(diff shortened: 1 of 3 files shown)


magicrub (commented):
This is probably a correct non breaking change but it just screams enum.

pkocmoud:
Thanks @magicrub, I also considered allowing for the user a choice of OUTPUT types: Push Pull and Open Drain. But I suspected some issue occurred in the past which necessitated the Lines 240-248. In the end my goal was to change as little as possible.

peterbarker:
@pkocmoud I think Sid's recent PR https://github.com/ArduPilot/ardupilot/pull/28140/files might make this on obsolete.  It completely ignores any pretense at layering and you just specify the low-level ChibiOS stuff to set the pin up.

Do you think Sid's stuff will suit your needs?

PR #27067, opened 2024-05-15
Add servo 2 and 3 min/max toggle function implementation

democat3457:
Adds missing Servo 2 and 3 min and max toggle functions to the button-handling switch-case.


rmackay9 (changes requested):
Txs for this.

One small thing is the commit description (not to be confused with the PR description) should be prefixed with the directory name.  E.g. "ArduSub".  The reason we do this is because it makes backporting much easier if we can see the subsystem affected by each commit.

democat3457:
@rmackay9 I've renamed the commit and force-pushed the new one.

Williangalvani:
thanks @democat3457 ,
I force-pushed your pr to remove a merge commit. I'll merge it once CI passes.
Thanks for the contribution, I can't believe it took this long for this to be noticed!

PR #26674, opened 2024-04-02
Support per-motor throttle-based harmonic notch

andyp1per:
To enable set the mult-source bit in INS_HNTCx_OPTS and use the throttle notch

PR #27068, opened 2024-05-15
Phase lag pre-arm check

andyp1per:
Made a separate PR for discussion so as to not hold up https://github.com/ArduPilot/ardupilot/pull/26674

@tridge your original suggestion was ```bw > 0.75*freq/nsources```, I assume that was a typo because that will break the configuration of all 14 of my copters that are using this feature very successfully (and for which this feature was originally written) 😄 I have changed that to  ```bw * 0.75 > freq/nsources``` which at least does not break many existing configurations but it does not work on octacopters. It is impossible to get the octacopter test to pass with this setting which was my original concern with having any check like this - it simply does not filter enough noise. My suggestion therefore is either to make this a warning or simply clamp the check at f/4. Either way much discussion and testing required!

### DIFF
--- a/libraries/AP_Vehicle/AP_Vehicle.cpp
+++ b/libraries/AP_Vehicle/AP_Vehicle.cpp
@@ -1111,6 +1111,41 @@ bool AP_Vehicle::block_GCS_mode_change(uint8_t mode_num, const uint8_t *mode_lis
 }
 #endif
 
+#if AP_INERTIALSENSOR_HARMONICNOTCH_ENABLED
+/// pre_arm_check - returns true if all pre-takeoff checks have completed successfully
+bool AP_Vehicle::pre_arm_check_notches() const
+{
+#if APM_BUILD_TYPE(APM_BUILD_ArduPlane) || APM_BUILD_COPTER_OR_HELI || APM_BUILD_TYPE(APM_BUILD_ArduSub)
+    AP_Motors* motors = AP::motors();
+#else  // APM_BUILD_Rover
+    AP_MotorsUGV *motors = AP::motors_ugv();
+#endif
+    const uint8_t nmotors = motors == nullptr ? 1 : __builtin_popcount(motors->get_motor_mask());
+
+    for (auto &notch : ins.harmonic_notches) {
+        if (!notch.params.enabled() || notch.params.hasOption(HarmonicNotchFilterParams::Options::DisableNotchPreArm)) {
+            continue;
+        }
+        float nsources = 1.0f;
+        if (notch.params.hasOption(HarmonicNotchFilterParams::Options::DynamicHarmonic)) {
+            if (notch.params.tracking_mode() == HarmonicNotchDynamicMode::UpdateBLHeli
+                || notch.params.tracking_mode() == HarmonicNotchDynamicMode::UpdateThrottle) {
+                nsources = nmotors;
+            }
+            else if (notch.params.tracking_mode() == HarmonicNotchDynamicMode::UpdateGyroFFT) {
+                nsources = 3;
+            }
+        }
+
+        if (notch.params.bandwidth_hz() > notch.params.center_freq_hz() / sqrtf(nsources)) {
+            return false;
+        }
+    }
+
+    return true;
+}
+#endif
+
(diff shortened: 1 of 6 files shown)


libraries/AP_Vehicle/AP_Vehicle.cpp
```diff
@@ -1109,6 +1109,42 @@ bool AP_Vehicle::block_GCS_mode_change(uint8_t mode_num, const uint8_t *mode_lis
...
+        if (notch.params.hasOption(HarmonicNotchFilterParams::Options::DynamicHarmonic)) {
+            if (notch.params.tracking_mode() == HarmonicNotchDynamicMode::UpdateBLHeli
+                || notch.params.tracking_mode() == HarmonicNotchDynamicMode::UpdateThrottle) {
+                nsources = nmotors;
+            }
+            else if (notch.params.tracking_mode() == HarmonicNotchDynamicMode::UpdateGyroFFT) {
+                nsources = 3;
+            }
+        }
+        if (nsources == 1 && notch.params.bandwidth_hz() > 0.75 * notch.params.center_freq_hz()) {
```
tridge, line 1137: needs an option bit to disable the check
has to be in INS_HNTCH_OPTS

tridge (commented):
maybe go to 1.0 so it passes with quad multi-src with 1:4 ratio?

andyp1per:
After discussion with @lthall we decided f/sqrt(nsources) was a suitable compromise in terms of pre-arm check. Adjusting attenuation unfortunately is unlikely to make enough of a difference.

libraries/AP_Vehicle/AP_Vehicle.cpp
andyp1per: Fixed

tridge:
> Adjusting attenuation unfortunately is unlikely to make enough of a difference.

I'm surprised about this. At zero attenuation there should be no phase lag. I wrote a little test to check this out and got this:
image
this was for a single notch set at 60Hz with bandwidth of 60Hz. Source frequency was 50Hz, and the test frequency varies for the 5 graphs from 10Hz to 50Hz.
The test shows a strong relationship between attenuation and phase lag.
test code here:
https://github.com/ArduPilot/ardupilot/pull/27164

andyp1per:
So Flytrex has pretty conclusively proved for me that 1/nsources is harmful. They flew on an X8 with notch-per-motor and saw *oscillation* at F/8 on their current tune because the noise was too high. F/6 also showed significant noise, whereas F/4 showed really good attenuation and good control on their *existing* tune. So as a pre-arm check 1/sqrt(nsources) is definitely the way to go. This is also proved by my existing autotest.

lthall:
I have a feeling this is a problem because Quicktune is resulting in dangerously small gain margins that are then made unstable by any increase in phase lag. Also true for manual tunes that are too tight but the Quicktune algorithm will make that the "recommended tune".

I think it is silly to have prearms to warn people about sensible parameter settings. All these things eat into flash space, support time and developer time.

lthall:
Another thought. If we are going to be adding a pattern where we put in parameter pre-arm checks that are triggered by sensible parameter settings and then expect the user to turn off each check with an options bit, maybe we should set up a specific structure for that.

This is sort of a new category of "sanity check" where we have changed a parameter so we want to force the pilot to achnolage that they understand the risks of what they have done and have taken appropriate action to mitigate the risk.

In this case it is: - You have turned on additional notches so your tune may become unstable.
But any decrease in filter frequency can also do that. So maybe we should be doing this with any of those things I would normally say "If you change this parameter, be very carful on your next flight and start the flight with a tuning stability check".

lthall:
> So Flytrex has pretty conclusively proved for me that 1/nsources is harmful. They flew on an X8 with notch-per-motor and saw _oscillation_ at F/8 on their current tune because the noise was too high. F/6 also showed significant noise, whereas F/4 showed really good attenuation and good control on their _existing_ tune using a single notch. So as a pre-arm check 1/sqrt(nsources) is definitely the way to go. This is also proved by my existing autotest.

This is great testing to show why we don't want this PR. Do we have any real life testing or examples to show that we need this check?

andyp1per:
Flytrex now have a decent tune with F/4 - so I pretty much conclude that we should close this PR.

PR #27070, opened 2024-05-16
Parameter: CRASH_DETECT Param doc update to add 1 in value set

loki077:
As this was causing a CI check we made locally to compare param values to param_metadata generated xml file.

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -731,7 +731,6 @@ const AP_Param::Info Plane::var_info[] = {
     // @Param: CRASH_DETECT
     // @DisplayName: Crash Detection
     // @Description: Automatically detect a crash during AUTO flight and perform the bitmask selected action(s). Disarm will turn off motor for safety and to help against burning out ESC and motor. Set to 0 to disable crash detection.
-    // @Values: 0:Disabled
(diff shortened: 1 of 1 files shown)


peterbarker (approved):
I've corrected the commit message and marked this for DevCall in case it doesn't get merged sooner.

Note that some of these fields are guidelines only; not necessarily `Values`, but certainly `Ranges`.

magicrub (changes requested):
Why not update the bitmask field too?

peterbarker:
> Why not update the bitmask field too?

Fair call.  In fact, that is the *correct* fix, as both MAVProxy and MissionPlanner ignore the @Values field if @Bitmask is present!

I have created https://github.com/ArduPilot/ardupilot/pull/27082 to enforce the same rule as we have had in libraries for years - you can only have one of @Values or @Bitmask

@loki077 you could simply remove the @Values as part of this PR - or we could merge mine which will close this one as Merged.

loki077:
@peterbarker I have forced pushed the change. You can also make it a part of your PR.

peterbarker:
> @peterbarker I have forced pushed the change. You can also make it a part of your PR.

I've merged that other PR now - so closing this one.

Thanks!

PR #27071, opened 2024-05-16
Copter: Handle DO_MOUNT_CONTROL yaw angle as body frame 

nexton-winjeel:
The demanded yaw angles in the `MOUNT_CONTROL` and `DO_MOUNT_CONTROL` commands are in body frame (and are handled this way in `AP_Mount`). This fixes the Copter handlers where they were being treated as earth frame.

It also rejects these messages if an invalid mode is provided, as the yaw control is only applicable when the mode is `MAV_MOUNT_MODE_MAVLINK_TARGETING`.

FYI: @rmackay9

### DIFF
--- a/libraries/AP_Mission/AP_Mission.cpp
+++ b/libraries/AP_Mission/AP_Mission.cpp
@@ -1251,6 +1251,7 @@ MAV_MISSION_RESULT AP_Mission::mavlink_int_to_mission_cmd(const mavlink_mission_
         break;
 
     case MAV_CMD_DO_MOUNT_CONTROL:                      // MAV ID: 205
+        // TODO: this is only valid if packet.z == MAV_MOUNT_MODE_MAVLINK_TARGETING
         cmd.content.mount_control.pitch = packet.param1;
         cmd.content.mount_control.roll = packet.param2;
         cmd.content.mount_control.yaw = packet.param3;
@@ -1767,6 +1768,7 @@ bool AP_Mission::mission_cmd_to_mavlink_int(const AP_Mission::Mission_Command& c
         packet.param1 = cmd.content.mount_control.pitch;
         packet.param2 = cmd.content.mount_control.roll;
         packet.param3 = cmd.content.mount_control.yaw;
+        packet.z = MAV_MOUNT_MODE_MAVLINK_TARGETING;
         break;
 
     case MAV_CMD_DO_SET_CAM_TRIGG_DIST:                 // MAV ID: 206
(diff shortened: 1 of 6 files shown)


libraries/AP_Mission/AP_Mission.cpp
```diff
@@ -1219,6 +1219,7 @@ MAV_MISSION_RESULT AP_Mission::mavlink_int_to_mission_cmd(const mavlink_mission_
         break;
 
     case MAV_CMD_DO_MOUNT_CONTROL:                      // MAV ID: 205
+        // TODO: this is only valid if packet.z == MAV_MOUNT_MODE_MAVLINK_TARGETING
```
nexton-winjeel, line 1222: Should this be returning `MAV_MISSION_INVALID_PARAM7` if the mode (`packet.z`) is not set to `MAV_MOUNT_MODE_MAVLINK_TARGETING`?

joshanne:
@rmackay9 you might have some interest on this one.

rmackay9:
We talked about this on the dev call and discussed the concern that users of MP may find that if they use the payload control tab that the vehicle may end up doing circles.  The procedure would be to setup a vehicle with a 2-axis gimbal and then:

- move Pan control to 45 deg
- move Tilt up and down

Each time the Tilt is moved the yaw would be re-sent and because after this change it would be in body-frame the vehicle would rotate

On the other hand, the Pan control changing the vehicle's heading to an absolute heading (e.g. North) also doesn't seem very good

I'll test and see what happens in SITL

rmackay9:
Hi @nexton-winjeel,

I've tested this and I agree that it is an improvement.  The existing behaviour results in the vehicle turning to point North if the user clicks on MP's the Payload Control pitch or roll controls which is certainly unexpected.

One small issue I noticed is that this essentially locks in any yaw error so if the user constantly adjusts roll or pitch the vehicle will also yaw slightly.  I think we could fix this by using the current target yaw instead of the actual yaw.

nexton-winjeel:
@rmackay9: 
> I think we could fix this by using the current target yaw instead of the actual yaw.

I've updated to use the current yaw target. I've also added an autotest.

nexton-winjeel:
`test.Plane.DeadreckoningNoAirSpeed` failed when I made Copter changes?

Rebased on `master` to re-trigger the tests.

nexton-winjeel:
And now `test.Copter.AutoTuneYawD` is failing. Can't repro locally though. Flapping?

PR #27074, opened 2024-05-16
AP_Scripting: added lua binding get_last_update for efi and esc

loki077:
We need this data available for Lua scripts

### DIFF
--- a/libraries/AP_Scripting/generator/description/bindings.desc
+++ b/libraries/AP_Scripting/generator/description/bindings.desc
@@ -450,6 +450,7 @@ singleton AP_ESC_Telem method get_usage_seconds boolean uint8_t 0 NUM_SERVO_CHAN
 singleton AP_ESC_Telem method update_rpm void uint8_t 0 NUM_SERVO_CHANNELS uint16_t'skip_check float'skip_check
 singleton AP_ESC_Telem method update_telem_data void uint8_t 0 NUM_SERVO_CHANNELS AP_ESC_Telem_Backend::TelemetryData uint16_t'skip_check
 singleton AP_ESC_Telem method set_rpm_scale void uint8_t 0 NUM_SERVO_CHANNELS float'skip_check
+singleton AP_ESC_Telem method get_last_telem_data_ms uint32_t uint8_t 0 ESC_TELEM_MAX_ESCS
 
 include AP_Param/AP_Param.h
 singleton AP_Param rename param
@@ -851,6 +852,7 @@ singleton AP_EFI depends (AP_EFI_SCRIPTING_ENABLED == 1)
 singleton AP_EFI rename efi
 singleton AP_EFI method get_backend AP_EFI_Backend uint8_t'skip_check
 singleton AP_EFI method get_state void EFI_State'Ref
+singleton AP_EFI method get_last_update_ms uint32_t 
 
 -- ----END EFI Library----
 
--- a/libraries/AP_Scripting/generator/description/bindings.desc
+++ b/libraries/AP_Scripting/generator/description/bindings.desc
@@ -440,16 +440,16 @@ userdata AP_ESC_Telem_Backend::TelemetryData field motor_temp_cdeg int16_t'skip_
 
 singleton AP_ESC_Telem depends HAL_WITH_ESC_TELEM == 1
 singleton AP_ESC_Telem rename esc_telem
-singleton AP_ESC_Telem method get_rpm boolean uint8_t 0 NUM_SERVO_CHANNELS float'Null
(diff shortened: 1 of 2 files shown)


magicrub (approved):

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -441,6 +441,7 @@ singleton AP_ESC_Telem method get_usage_seconds boolean uint8_t 0 NUM_SERVO_CHAN
 singleton AP_ESC_Telem method update_rpm void uint8_t 0 NUM_SERVO_CHANNELS uint16_t'skip_check float'skip_check
 singleton AP_ESC_Telem method update_telem_data void uint8_t 0 NUM_SERVO_CHANNELS AP_ESC_Telem_Backend::TelemetryData uint16_t'skip_check
 singleton AP_ESC_Telem method set_rpm_scale void uint8_t 0 NUM_SERVO_CHANNELS float'skip_check
+singleton AP_ESC_Telem method get_last_telem_data_ms uint32_t uint8_t 0 NUM_SERVO_CHANNELS
```
peterbarker, line 444:
```suggestion
singleton AP_ESC_Telem method get_last_telem_data_ms uint32_t uint8_t 0 ESC_TELEM_MAX_ESCS
```

peterbarker (changes requested):

libraries/AP_Scripting/generator/description/bindings.desc
loki077: In this case shouldn't all the other ESC_telem defines be change from NUM_SERVO_CHANNELS to ESC_TELEM_MAX_ESCS

peterbarker: > In this case shouldn't all the other ESC_telem defines be change from NUM_SERVO_CHANNELS to ESC_TELEM_MAX_ESCS

Possibly?  I based this suggestion on the sanity check in the top of that method, nothing more.

loki077:
@peterbarker @tridge I have made the requested changes.

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -438,16 +438,17 @@ userdata AP_ESC_Telem_Backend::TelemetryData field motor_temp_cdeg int16_t'skip_
...
-singleton AP_ESC_Telem method update_rpm void uint8_t 0 NUM_SERVO_CHANNELS uint16_t'skip_check float'skip_check
-singleton AP_ESC_Telem method update_telem_data void uint8_t 0 NUM_SERVO_CHANNELS AP_ESC_Telem_Backend::TelemetryData uint16_t'skip_check
-singleton AP_ESC_Telem method set_rpm_scale void uint8_t 0 NUM_SERVO_CHANNELS float'skip_check
+singleton AP_ESC_Telem method get_rpm boolean uint8_t 0 ESC_TELEM_MAX_ESCS float'Null
+singleton AP_ESC_Telem method get_temperature boolean uint8_t 0 ESC_TELEM_MAX_ESCS int16_t'Null
+singleton AP_ESC_Telem method get_motor_temperature boolean uint8_t 0 ESC_TELEM_MAX_ESCS int16_t'Null
+singleton AP_ESC_Telem method get_current boolean uint8_t 0 ESC_TELEM_MAX_ESCS float'Null
+singleton AP_ESC_Telem method get_voltage boolean uint8_t 0 ESC_TELEM_MAX_ESCS float'Null
+singleton AP_ESC_Telem method get_consumption_mah boolean uint8_t 0 ESC_TELEM_MAX_ESCS float'Null
+singleton AP_ESC_Telem method get_usage_seconds boolean uint8_t 0 ESC_TELEM_MAX_ESCS uint32_t'Null
```
IamPete1, line 447: I think we can just `'skip_check` for all these that return bool. The result will be null and `AP_ESC_Telem` should deal with it correctly. The problem with using a define is that there is no way for the script to get the value.

Lokesh-Carbonix: Do you want all ESC_TELEM_MAX_ESCS to be replaced with 'skip_check. I am not sure of the define implementation and have been following the previous implementation. Not sure if it wil brake anything else. @IamPete1 if you are good with it I will update the PR.

loki077: @robertlong13 FYI

IamPete1: Yeah, we should be good to remove them all.

tridge (approved):

peterbarker:
Merged it, thanks!

PR #27076, opened 2024-05-16
AP_ICEngine: Add CRANK_DIR for direction control of TCA9554 Engine

loki077:
Added a parameter CRAN_DIR where 0 if forward and 1 or greater is reverse direction for Engine Cranking. This is particular with Hirth Engine where both options are available.

### DIFF
--- a/libraries/AP_ICEngine/AP_ICEngine.cpp
+++ b/libraries/AP_ICEngine/AP_ICEngine.cpp
@@ -145,7 +145,7 @@ const AP_Param::GroupInfo AP_ICEngine::var_info[] = {
     // @Param: OPTIONS
     // @DisplayName: ICE options
     // @Description: Options for ICE control. The Disable ignition in RC failsafe option will cause the ignition to be set off on any R/C failsafe. If Throttle while disarmed is set then throttle control will be allowed while disarmed for planes when in MANUAL mode. If disable while disarmed is set the engine will not start while the vehicle is disarmed unless overriden by the MAVLink DO_ENGINE_CONTROL command.
-    // @Bitmask: 0:Disable ignition in RC failsafe,1:Disable redline governor,2:Throttle while disarmed,3:Disable while disarmed
+    // @Bitmask: 0:Disable ignition in RC failsafe,1:Disable redline governor,2:Throttle while disarmed,3:Disable while disarmed,4:Crank direction Reverse
     AP_GROUPINFO("OPTIONS", 15, AP_ICEngine, options, 0),
 
     // @Param: STARTCHN_MIN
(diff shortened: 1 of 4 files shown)


libraries/AP_ICEngine/AP_ICEngine_TCA9554.cpp
```diff
@@ -82,7 +82,7 @@ void AP_ICEngine_TCA9554::TCA9554_set(TCA9554_state_t value)
     }
 }
 
-void AP_ICEngine_TCA9554::set_starter(bool on)
+void AP_ICEngine_TCA9554::set_starter(bool on, AP_Int8 crank_direction)
```
magicrub, line 85:
```suggestion
void AP_ICEngine_TCA9554::set_starter(bool on, int8_t crank_direction)
```

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -167,6 +167,14 @@ const AP_Param::GroupInfo AP_ICEngine::var_info[] = {
 
     // 18 was IGNITION_RLY
 
+#if AP_ICENGINE_TCA9554_STARTER_ENABLED
+    // @Param: CRANK_DIR
+    // @DisplayName: Engine Cranking Direction for TCA9554
+    // @Description: This will define which direction the engine is cranking for the TCA9554 starter control
+    // @User: Standard
+    // @Values: 0:Forward,1:Reverse
+    AP_GROUPINFO("CRANK_DIR", 19, AP_ICEngine, crank_direction, 0),
```
magicrub, line 176: This belongs inside the driver

magicrub (changes requested):

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -167,6 +167,11 @@ const AP_Param::GroupInfo AP_ICEngine::var_info[] = {
 
     // 18 was IGNITION_RLY
 
+#if AP_ICENGINE_TCA9554_STARTER_ENABLED
+    // @Group: _TCA9554
+    // @Path: AP_ICEngine_TCA9554.cpp
+    AP_SUBGROUPINFO(tca9554_starter, "_TCA", 19, AP_ICEngine, AP_ICEngine_TCA9554),
```
tridge, line 173: maybe just a ICE_OPTIONS bit instead? That would be a lot smaller patch and not clutter up parameters. We only need a single bit for the direction

tridge (changes requested):

libraries/AP_ICEngine/AP_ICEngine.cpp
loki077: yes that would be good, as if there where other implementation of cranking they can use that flag too for direction control.

loki077: @tridge I have made the changes, have a look.

Also, if there is a need to splitting the commit into one for param mods and other for implementation then let me know.

tridge (approved):

PR #27077, opened 2024-05-16
Adding ThM (modified throttle send to engine) to EFIS log

loki077:
- ThM added to help understand the engine logs better
 - Updated the field names to keep the total length 64-characters long.

### DIFF
--- a/libraries/AP_EFI/AP_EFI_Serial_Hirth.cpp
+++ b/libraries/AP_EFI/AP_EFI_Serial_Hirth.cpp
@@ -205,6 +205,7 @@ void AP_EFI_Serial_Hirth::send_request()
 bool AP_EFI_Serial_Hirth::send_target_values(uint16_t thr)
 {
     uint8_t computed_checksum = 0;
+    throttle_to_hirth = 0;
 
     // clear buffer
     memset(raw_data, 0, ARRAY_SIZE(raw_data));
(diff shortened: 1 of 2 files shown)


libraries/AP_EFI/AP_EFI_Serial_Hirth.cpp
```diff
@@ -205,6 +205,7 @@ void AP_EFI_Serial_Hirth::send_request()
 bool AP_EFI_Serial_Hirth::send_target_values(uint16_t thr)
 {
     uint8_t computed_checksum = 0;
+    throttle = 0;
```
peterbarker, line 208: This is a little too generic a name for a member variable IMO.

Perhaps
```suggestion
    throttle_to_hirth = 0;
```

libraries/AP_EFI/AP_EFI_Serial_Hirth.cpp
```diff
@@ -369,23 +370,24 @@ void AP_EFI_Serial_Hirth::log_status(void)
...
-    // @Field: ETSS: Status of Engine temperature sensor
-    // @Field: ATSS: Status of Air temperature sensor
-    // @Field: APSS: Status of Air pressure sensor
+    // @Field: ES1: Status of EGT1 excess temperature error
+    // @Field: ES2: Status of EGT2 excess temperature error
+    // @Field: CS1: Status of CHT1 excess temperature error
+    // @Field: CS2: Status of CHT2 excess temperature error
+    // @Field: ETS: Status of Engine temperature sensor
+    // @Field: ATS: Status of Air temperature sensor
+    // @Field: APS: Status of Air pressure sensor
```
peterbarker, line 379: Many of these are booleans.  You could save columns and logging bandwidth by logging them into a bitmask.

Many analysis tools do have reasonable support for bitmasks now.

peterbarker (changes requested):
Commit message needs fixing.

libraries/AP_EFI/AP_EFI_Serial_Hirth.cpp
loki077: I have change the implementation and grouped all bool into 2 groups EET( Error Excess Temperature) FLAGS ( Sensor Status) I can also group them into one but thought to keep separate @peterbarker let me know your opinion on that. 
Also, will WriteStreaming be able to use FieldBitmaskEnum?

image
image

loki077:
> Commit message needs fixing.

I will fix the commit message once the code is all good.

loki077:
@Pradeep-Carbonix @robertlong13 FYI

libraries/AP_EFI/AP_EFI_Serial_Hirth.h
```diff
@@ -31,6 +31,28 @@ class AP_EFI_Serial_Hirth: public AP_EFI_Backend {
 
     void update() override;
 
+    enum class Error_Excess_Temp_Bitfield : uint16_t {
```
IamPete1, line 34: These ones that are only really used for log documentation you can define right in the `log_status` function in the cpp. That stops it getting used elsewhere my mistake.

loki077: @IamPete1 what i mention above is that ok?

tridge: seems reasonable to me

tridge (approved):

PR #27079, opened 2024-05-16
AP_Terrain: added parameter for terrain cache size

tridge:
for a fast moving aircraft with small TERRAIN_SPACING you need a larger cache to prevent thrashing of the cache


peterbarker (approved):
LGTM

PR #27082, opened 2024-05-17
Remove reduntant @Values from parameter documentation

peterbarker:
we have enforced this one-of-Values-or-Bitmask in libraries for a long time, and should have been enforcing it in vehicles before now.

MissionPlanner and MAVProxy ignore Values if Bitmask is present

Built on top of https://github.com/ArduPilot/ardupilot/pull/27070


ArduPlane/Parameters.cpp
```diff
@@ -1135,8 +1132,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Param: DSPOILER_OPTS
     // @DisplayName: Differential spoiler and crow flaps options
     // @Description: Differential spoiler and crow flaps options
-    // @Values: 0: none, 1: D spoilers have pitch input, 2: use both control surfaces on each wing for roll, 4: Progressive crow flaps only first (0-50% flap in) then crow flaps (50 - 100% flap in)
```
amilcarlucas, line 1138: Some of the details should be added in the description IMHO

amilcarlucas (changes requested):
Nice PR!!!

Rover/Parameters.cpp
```diff
@@ -636,8 +635,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Param: FS_OPTIONS
     // @DisplayName: Failsafe Options
     // @Description: Bitmask to enable failsafe options
-    // @Values: 0:None,1:Failsafe enabled in Hold mode
-    // @Bitmask: 0:Failsafe enabled in Hold mode
+    // @Bitmask: 0:Failsafe enabled in Hold mode,1:Failsafe enabled in Hold mode
```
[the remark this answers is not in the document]
peterbarker: Fixed.

ArduPlane/Parameters.cpp
peterbarker: Done

PR #27083, opened 2024-05-17
SITL: added SIM_WIND_TC

tridge:
wind in the real world does not change instantly. Adding a time constant prevents the EKF from getting confused with instant changes in wind speed

### DIFF
--- a/libraries/SITL/SITL.cpp
+++ b/libraries/SITL/SITL.cpp
@@ -81,12 +81,20 @@ const AP_Param::GroupInfo SIM::var_info[] = {
     // @Units: deg
     // @User: Advanced
     AP_GROUPINFO("WIND_DIR",      10, SIM,  wind_direction,  180),
+
     // @Param: WIND_TURB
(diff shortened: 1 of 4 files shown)


libraries/SITL/SITL.cpp
```diff
@@ -81,13 +81,21 @@ const AP_Param::GroupInfo SIM::var_info[] = {
...
     // @Param: WIND_TURB
     // @DisplayName: Simulated Wind variation
     // @Description: Allows you to emulate random wind variations in sim
     // @Units: m/s
     // @User: Advanced
     AP_GROUPINFO("WIND_TURB",     11, SIM,  wind_turbulance,  0),
 
+    // @Param: WIND_TC
+    // @DisplayName: Wind variation time constant
+    // @Description: this controls the time over which wind changes take effect
```
rmackay9, line 94: normally we don't need the "this controls" at the front because all parameters control something.

peterbarker (approved):

PR #27092, opened 2024-05-18
Plane:correct Q_ASSIST metadata

Hwurzburg:


ArduPlane/quadplane.cpp
```diff
@@ -103,7 +103,7 @@ const AP_Param::GroupInfo QuadPlane::var_info[] = {
...
+    // @Description: This is the speed below which the quad motors will provide stability and lift assistance in fixed wing modes. -1 means no assistance except during transition. Note that if this is set to -1 then other Q_ASSIST features are also disabled. A non-zero,positive high value will lead to more false positives which can waste battery. A lower value will result in less false positive, but will result in assistance taking longer to trigger. If unsure then set to 3 m/s below the minimum airspeed you will fly at. If you don't have an airspeed sensor then use 5 m/s below the minimum airspeed you fly at. If you want to disable the arming check and all non-transition assistance Q_ASSIST_SPEED should be set to -1.
```
IamPete1, line 106:
```suggestion
    // @Description: This is the speed below which the quad motors will provide stability and lift assistance in fixed wing modes. Zero or -1 means no assistance except during transition. Note that if this is set to zero or -1 then other Q_ASSIST features are also disabled. A non-zero,positive high value will lead to more false positives which can waste battery. A lower value will result in less false positive, but will result in assistance taking longer to trigger. If unsure then set to 3 m/s below the minimum airspeed you will fly at. If you don't have an airspeed sensor then use 5 m/s below the minimum airspeed you fly at. If you want to disable the arming check and all non-transition assistance Q_ASSIST_SPEED should be set to -1.
```

Hwurzburg: Pete,as you explained it there is no way now to use 0 for Q_ASSIST it throws a  prearm

IamPete1: Right, but the default is 0. So the description should explain what it does in that case.

tridge (approved):

PR #27094, opened 2024-05-18
AC_Fence: A circular fence with a radius of 0 meters is ignored

muramura:
I am developing an application for vehicles that support MAVLINK messages. I have discovered different behaviors in the setting of circular fences. I want to ignore circles with a radius of 0 meters. By ignoring circles with a radius of 0 meters, I can standardize the processing with vehicles other than ArduPilot.

AFTER
Screenshot from 2024-05-18 23-13-29

Screenshot from 2024-05-18 23-13-36

BEFORE
Screenshot from 2024-05-18 22-11-23

### DIFF
--- a/libraries/AC_Fence/AC_PolyFence_loader.cpp
+++ b/libraries/AC_Fence/AC_PolyFence_loader.cpp
@@ -968,7 +968,7 @@ bool AC_PolyFence_loader::validate_fence(const AC_PolyFenceItem *new_items, uint
                GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "Received incorrect type (want=%u got=%u)", (unsigned)expecting_type, (unsigned)new_items[i].type);
                return false;
             }
-            if (!is_positive(new_items[i].radius)) {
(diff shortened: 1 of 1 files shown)


rmackay9:
I guess the advantage of supporting this is that it allows an external system to essentially disable a fence without removing it from the list.  The downside is that it introduces an edge case that we will need to be careful of in the flight code.  The flight code may or may not be ignoring circular fences with zero radius.

muramura:
@rmackay9 san.
I'm modifying the process for adding circle fences so that only those with a positive radius are registered.

Screenshot from 2024-05-21 06-29-08

peterbarker:
I'm really confused by this - we're enforcing different rules for the fence in different places, which is just really rather odd.

And patching in "write_fence" is very strange.

peterbarker:
Sorry, I can't see this as a good idea.

The current sanity check is there to prevent the sorts of errors which Randy mentions.

You can't have a circle with zero radius - that's called a "point".

Closing this, sorry.

PR #27096, opened 2024-05-19
GCS_MAVLink: Notify access from an invalid system ID

muramura:
I attempted to control the vehicle with a manual control message from the GCS. The vehicle descended in altitude. Upon investigating the code, I discovered that the system ID was unregistered. I know that DRONEKIT and MAVSDK have default system IDs, but I didn't remember the exact values. I plan to notify the GCS of the system ID values to facilitate setting them in the FC.

AFTER
Screenshot from 2024-05-19 12-27-06


libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -6864,6 +6864,7 @@ void GCS_MAVLINK::manual_override(RC_Channel *c, int16_t value_in, const uint16_
 void GCS_MAVLINK::handle_manual_control(const mavlink_message_t &msg)
 {
     if (msg.sysid != sysid_my_gcs()) {
+        GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "Ignore sysid %u", msg.sysid);
```
peterbarker, line 6867: 1. this is only once place a warning might need to be sent (make a `send_sysid_ignored()` method?)
2. this message would need to be rate-limited as these messages can come in *fast*, and you will flood people with messages.

peterbarker (changes requested):

PR #27101, opened 2024-05-20
AP_GPS, PrecLand, EFI, SerialManager: TYPE param docs renamed to MAVLink

rmackay9:
This only affects parameter documentation so should be low-risk.

This makes consistent the Type to be "MAVLink" in those few cases where we call it something else.

There's actually [one more in RC_Channels](https://github.com/ArduPilot/ardupilot/blob/master/libraries/RC_Channel/RC_Channels_VarInfo.h#L104) where we call it "MAVRadio" so I'd like to rename this one as well at some point if people are happy with that but that can be a separate PR.


IamPete1 (approved):

amilcarlucas (approved):
Nice cleanup

peterbarker (approved):
LGTM

PR #27106, opened 2024-05-20
AP_TECS: modify max climb estimate to prevent throttle bouncing on limit

jschall:
During a steady climb where throttle is saturated, throttle will not consistently hold at 100%, but will instead periodically bounce because of the behavior of ```_max_climb_scaler```. The same applies to descents.

Example from log:
bouncing-throttle

This pull request contains a **currently un-tested** proposal for fixing this. ```_max_(climb|sink)_scaler``` will now asymptotically approach the current climb/sink rate (plus a 10% margin, to be discussed) over time, rather than rocketing down toward zero. @magicrub @tridge @samuelctabor @priseborough thoughts?

### DIFF
--- a/libraries/AP_TECS/AP_TECS.cpp
+++ b/libraries/AP_TECS/AP_TECS.cpp
@@ -579,9 +579,13 @@ void AP_TECS::_update_height_demand(void)
             }
             const float hgt_dem_alpha = _DT / MAX(_DT + _hgt_dem_tconst, _DT);
             if (max_climb_condition && _hgt_dem > _hgt_dem_prev) {
-                    _max_climb_scaler *= (1.0f - hgt_dem_alpha);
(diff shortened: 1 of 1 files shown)


tridge:
@priseborough can you take a look?

tridge:
really need the bin log so we can check airspeed and flags

jschall:
tecs_bouncing

jschall:
See also https://github.com/ArduPilot/ardupilot/pull/27114

peterbarker:
> However, the fix is causing these side effects due to the clamping of demanded altitude when saturation is detected.

Sounds like we avoided a regression because you noticed this PR.... can you outline the test you ran to determine the side-effects were a problem, please?  An actual autotest seems to be warranted.

PR #27107, opened 2024-05-20
AP_Compass: magcal keeps soft iron matrix + scale bugfix

jschall:
This patch changes the large vehicle compass calibration so that it keeps the soft-iron calibration matrix and the scale. This way, the user can perform a full compass calibration on a subsystem of the vehicle, without losing it when a large vehicle cal is performed.

It also fixes a bug: large vehicle compass calibration currently does not respect the scale parameter. It needs to either incorporate it into the math or reset it to 1.

Why was the scale param added when the diagonals already exist? Seems like something that can be removed.

### DIFF
--- a/libraries/AP_Compass/AP_Compass_Calibration.cpp
+++ b/libraries/AP_Compass/AP_Compass_Calibration.cpp
@@ -458,34 +458,45 @@ MAV_RESULT Compass::handle_mag_cal_command(const mavlink_command_int_t &packet)
   get mag field with the effects of offsets, diagonals and
   off-diagonals removed
  */
-bool Compass::get_uncorrected_field(uint8_t instance, Vector3f &field) const
+bool Compass::compute_offsets(uint8_t instance, const Vector3f& truth_field, Vector3f &new_offsets) const
 {
+    // uncorrected_field = inv(mat)*corrected_field*1/scale-offsets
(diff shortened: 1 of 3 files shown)


tridge (changes requested):
this would need to be a COMPASS_OPTIONS bit, off by default

peterbarker:
> this would need to be a COMPASS_OPTIONS bit, off by default

Or an option in the mavlink command to retain the diags/off-diags

libraries/AP_Compass/AP_Compass_Calibration.cpp
```diff
@@ -458,34 +458,45 @@ MAV_RESULT Compass::handle_mag_cal_command(const mavlink_command_int_t &packet)
...
+        // remove impact of scale, diagonals, and off-diagonals
+        uncorrected_field = mat_inv * uncorrected_field * 1.0f/scale;
     }
 #endif
 
     // remove impact of offsets
-    field -= get_offsets(instance);
+    uncorrected_field -= get_offsets(instance);
+    
+    new_offsets = mat_inv*truth_field*1.0f/scale - uncorrected_field;
```
magicrub, line 499: scale needs divide by zero check

magicrub:
I agree with @tridge . Needs to be an options bit, default off.

PR #27109, opened 2024-05-20
AP_Mount: Only save converted mount parameter if mount was previously set in the first place

joshanne:
The mount library force configures the mount type on conversion, even if the mount was never configured in the first place.

### DIFF
--- a/libraries/AP_Mount/AP_Mount.cpp
+++ b/libraries/AP_Mount/AP_Mount.cpp
@@ -1014,7 +1014,10 @@ void AP_Mount::convert_params()
     // convert MNT_TYPE to MNT1_TYPE
     int8_t mnt_type = 0;
     IGNORE_RETURN(AP_Param::get_param_by_index(this, 19, AP_PARAM_INT8, &mnt_type));
-    if (mnt_type > 0) {
+    if (mnt_type == 0) {
+        // if the mount was not previously set, no need to perform the upgrade logic
+        return;
+    } else if (mnt_type > 0) {
         int8_t stab_roll = 0;
         int8_t stab_pitch = 0;
         IGNORE_RETURN(AP_Param::get_param_by_index(this, 4, AP_PARAM_INT8, &stab_roll));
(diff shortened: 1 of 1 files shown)


rmackay9:
Hi @joshanne,

Thanks for this.  Have you noticed any issue with this save?  This is actually done on purpose so that we don't run the parameter conversion on each boot.

joshanne:
> Thanks for this. Have you noticed any issue with this save? This is actually done on purpose so that we don't run the parameter conversion on each boot.

If you flash the flight controller then boot, the parameter is immediately considered configured. So if you update to with a configured `defaults.parm` the default mount is not reflected.

The conversion probably should only occur if the old `MNT_TYPE` parameter is non zero and the current mount is not configured && the current mount type is non-zero.

ie. Do not run the conversion if:
* if a defaults file is used then the new parameter is unconfigured but non-zero.
* if a defaults file is not used, and the parameter is non-zero.

What we want to achieve is don't trample the user config if updating from an old version of software.

joshanne:
At the very least the conversion should only occur if the old mount type is set to a non-zero software.

The underlying problem is this conversion runs and writes that mount parameter even if the flight controller ever saw a later version of software.

libraries/AP_Mount/AP_Mount.cpp
```diff
@@ -1024,8 +1027,9 @@ void AP_Mount::convert_params()
             // conversion is still done even if HAL_MOUNT_SERVO_ENABLED is false
             mnt_type = 7;  // (int8_t)Type::BrushlessPWM;
         }
+        // if the mount was previously set, then we need to save the upgraded mount type
+        _params[0].type.set_and_save(mnt_type);
     }
-    _params[0].type.set_and_save(mnt_type);
```
joshanne, line 1028: Moving this inside the else ensures that we only write a value to `MNT1_TYPE` if the previous `MNT_TYPE` was set to a non zero value.

ie. don't set `MNT1_TYPE` to a configured value of zero, as this is no longer "default".

joshanne:
@rmackay9 - updated the changes, comments added to both source and PR.

joshanne:
I still lean toward the initial commit, because upgrading of the other mount parameters could be perfectly valid if the mount type was not set at the time of the upgrade.

rmackay9:
@joshanne,

OK, thanks.  The existing code is a very common pattern though so I think we should discuss with @tridge (and perhaps others) at the EU dev call later today.

tridge (approved):

PR #23658, opened 2023-05-02
Plane: fix rangefinder correction when terrain follow is off

jschall:
Previous behavior was either to flare extremely high or crash into the ground, depending on whether the land waypoint is higher or lower than home.

This allows the user to set a land waypoint altitude other than 0 relative without causing a crash.

PR #27110, opened 2024-05-20
Plane: fix rangefinder correction when terrain follow is off

jschall:
When terrain is disabled and the LAND waypoint altitude is nonzero, the current behavior is to flare at an incorrect height despite the presence of a rangefinder, either flaring high and stalling or not flaring at all and crashing into the ground.

This addresses that bug.


jschall:
closing as duplicate of https://github.com/ArduPilot/ardupilot/pull/23658

PR #27112, opened 2024-05-21
Tools: Scripts: run_lua_language_check: give hint about github-release-downloader

IamPete1:
This gives a hint about how to fix the missing import.

### DIFF
--- a/Tools/scripts/run_lua_language_check.py
+++ b/Tools/scripts/run_lua_language_check.py
@@ -69,7 +69,13 @@ if __name__ == '__main__':
 
     # See if there is a new version (only try on Linux)
     if platform.system() == "Linux":
-        from github_release_downloader import check_and_download_updates, GitHubRepo
+        try:
+            from github_release_downloader import check_and_download_updates, GitHubRepo
+        except ImportError:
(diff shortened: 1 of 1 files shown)


Tools/scripts/run_lua_language_check.py
```diff
@@ -69,7 +69,13 @@ def print_failures(file_name, fails):
 
     # See if there is a new version (only try on Linux)
     if platform.system() == "Linux":
-        from github_release_downloader import check_and_download_updates, GitHubRepo
+        try:
+            from github_release_downloader import check_and_download_updates, GitHubRepo
+        except ImportError:
+            print("Import github-release-downloader failed")
+            print("Install with: python -m pip install github-release-downloader")
```
tridge, line 76: should be: python3 -m pip ....

tridge (changes requested):

Tools/scripts/run_lua_language_check.py
IamPete1: done

peterbarker (approved):

tridge (approved):

PR #27113, opened 2024-05-21
autotest: neaten Copter Loiter test

peterbarker:
... the harness takes care of the cleanup in case of error

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -435,47 +435,38 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
 
         self.context_push()
 
-        ex = None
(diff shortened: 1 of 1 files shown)


rmackay9 (approved):
LGTM.  I'd just like to check that you've tested the failure case to confirm the failure is reported correctly to the user/caller.

peterbarker:
> LGTM. I'd just like to check that you've tested the failure case to confirm the failure is reported correctly to the user/caller.

The code which catches this now is here: https://github.com/ardupilot/ardupilot/blob/master/Tools/autotest/vehicle_test_suite.py#L8458

... that code is there to make the changes I'm making in this PR (and several very similar PRs) possible.

PR #27114, opened 2024-05-21
autotest: add test showing throttle satuation problem

peterbarker:


jschall:
Before #27106 
image
After #27106
image

peterbarker:
I've now modified this to be a regression test, and I intend to merge this after CI has passed. 

The test fails on master but passed on top of #27106 

The test will be disabled - whatever fixes the problem should remove the line which disables the test.

PR #27115, opened 2024-05-21
Plane: check for terrain enable in terrain-rangefinder consistency check

jschall:


tridge:
could you reproduce in SITL?

peterbarker:
Ping @jschall tridge wants to know if whatever this fixes is reproducible in SITL.

PR #27118, opened 2024-05-21
autotest: add test for TECS overspeed

peterbarker:


peterbarker:
Note that this PR is only a placeholder/starting point for a more nuanced PR.  The setting of TECS_SINK_RATE is in complete violation of the parameter description/costraints.

PR #27125, opened 2024-05-21
WIP: autotest: Add catapult test

Georacer:
An autotest for a catapult launch has been added for Plane.
An auxiliary `set_servo` method has been added to `vehicle_test_suite.py`.


Georacer:
The way to call the test is
```bash
./Tools/autotest/autotest.py --no-clean --frame plane-catapult --speedup 10 build.Plane test.Plane.CatapultTakeoff
```

This PR is WIP because it's premature and failing.

There are two issues with it:

1. This test is specific to the `plane-catapult` frame. I'd like some advice on how to prevent it from running automatically for all frames.

2. For some reason the catapult refuses to launch the vehicle. I seem to successfully send the RC trigger to the catapult, but it does not produce forward acceleration.
At some point the throttle picks up a bit (I don't know why) and the plane starts crawling on the ground).
 
image

Even more weirdly, autotest things the airspeed has risen, when the logs shows it hasn't.
```bash
AT-0004.1: airspeed=0.40 (want between 21.0 and 100)
AT-0004.1: AP: Triggered AUTO. GPS speed = 0.8
AT-0004.2: airspeed=13.59 (want between 21.0 and 100)
AT-0004.2: AP: SIM Hit ground at 3.284683 m/s
AT-0004.3: airspeed=27.58 (got between 21.0 and 100) (maintain=0.6/5.0)
AT-0004.3: AP: SIM Hit ground at 4.907945 m/s
AT-0004.3: AP: Holding course 35398 at 28.8m/s (-0.0)
AT-0004.4: airspeed=26.24 (got between 21.0 and 100) (maintain=1.7/5.0)
AT-0004.4: AP: SIM Hit ground at 4.911546 m/s
AT-0004.5: airspeed=25.58 (got between 21.0 and 100) (maintain=2.8/5.0)
AT-0004.6: AP: SIM Hit ground at 3.960280 m/s
AT-0004.6: airspeed=25.28 (got between 21.0 and 100) (maintain=3.8/5.0)
AT-0004.7: AP: SIM Hit ground at 3.963535 m/s
AT-0004.8: airspeed=25.92 (got between 21.0 and 100) (maintain=4.9/5.0)
AT-0004.8: AP: SIM Hit ground at 3.964641 m/s
AT-0004.8: Attained airspeed=25.871821010813992

```

There's a slight chance I'm looking at the wrong .bin. I inspected `buildlogs/*2.bin` and `*3.bin`.

Tools/autotest/vehicle_test_suite.py
```diff
@@ -5406,6 +5406,10 @@ def set_rc(self, chan, pwm, timeout=20):
         """Setup a simulated RC control to a PWM value"""
         self.set_rc_from_map({chan: pwm}, timeout=timeout)
 
+    def set_servo(self, chan, pwm, timeout=0):
```
peterbarker, line 5409:
```suggestion
    def set_servo(self, chan, pwm):
```

Tools/autotest/arduplane.py
```diff
@@ -4160,6 +4160,38 @@ def LandingDrift(self):
...
+        if max_throttle == 0:
+            max_throttle = self.get_parameter("THR_MAX")
+
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
+
+        # Load and start mission.
+        self.load_mission("catapult.txt", strict=True)
+        self.set_current_waypoint(1, check_afterwards=True)
+        self.change_mode('AUTO')
```
peterbarker, line 4182: Usually you'd load your mission before arming your vehicle :-)

I suggest using `wait_groundspeed(..., minimum_duration=5)` to ensure the vehicle doesn't move until the catapult is released.

Tools/autotest/arduplane.py
```diff
@@ -4160,6 +4160,38 @@ def LandingDrift(self):
 
         self.wait_disarmed(timeout=180)
 
+    def CatapultTakeoff(self):
+        '''Test that a catapult takeoff works correctly'''
+
+        self.set_parameters({
```
peterbarker, line 4166:
```suggestion
        self.customise_SITL_commandline(
            [],
            model="plane-catapult",
            defaults_filepath=self.model_defaults_filepath("plane"),
        )

        self.set_parameters({
```

Tools/autotest/arduplane.py
```diff
@@ -4160,6 +4160,38 @@ def LandingDrift(self):
...
+        # Throw the catapult.
+        self.set_servo(7, 2000)
+
+        # Ensure that after the airspeed has been reached, the throttle recedes.
+        max_pwm = max_throttle*20
+        self.wait_airspeed(min_speed-1, 100, minimum_duration=5, timeout=10)
+        self.wait_servo_channel_value(3, max_pwm, timeout=0.1, comparator=operator.lt)
+        # Wait for landing waypoint.
+        self.wait_current_waypoint(11, timeout=1200)
+        self.wait_disarmed()
```
peterbarker, line 4193:
```suggestion
        self.wait_disarmed(timeout=120)
```

Tools/autotest/arduplane.py
```diff
@@ -4160,6 +4160,38 @@ def LandingDrift(self):
...
+        max_throttle = self.get_parameter("TKOFF_THR_MAX")
+        if max_throttle == 0:
+            max_throttle = self.get_parameter("THR_MAX")
+
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
+
+        # Load and start mission.
+        self.load_mission("catapult.txt", strict=True)
+        self.set_current_waypoint(1, check_afterwards=True)
```
peterbarker, line 4181:
```suggestion
```

peterbarker (changes requested):

Tools/autotest/arduplane.py
Georacer: I'll place the mission loading before and just wait for arming checks to pass.
The mission load ended there in an act of frustration ;)

Georacer:
@peterbarker thanks for the suggestions. I've implemented it and it's running now.

Question:
I modified the `basic.txt` mission, that is used in `FlyEachFrame`. It contains a `DO_LAND_START` and more waypoints than I need for this test.
Should I use it regardless, or next time I should code a mission small enough for my needs?

Remark:
I've places this test in the known-failing list, as I believe this behaviour is now bugged.
The throttle never recedes during climb and the aircraft overspeeds:
image

If you agree, we can consider this PR done and I can open an Issue ticket.

Tools/autotest/arduplane.py
```diff
@@ -4160,6 +4160,43 @@ def LandingDrift(self):
...
+            max_throttle = self.get_parameter("THR_MAX")
+
+        # Load and start mission.
+        self.load_mission("catapult.txt", strict=True)
+        self.change_mode('AUTO')
+
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
+
+        # Throw the catapult.
```
peterbarker, line 4189:
```suggestion
        self.progress("Throw the catapult")
```
Many times when I've put a comment like this in I've found it useful to put it in as a progress text instead.  Just a suggestion.

peterbarker:
> Question: I modified the `basic.txt` mission, that is used in `FlyEachFrame`. It contains a `DO_LAND_START` and more waypoints than I need for this test. Should I use it regardless, or next time I should code a mission small enough for my needs?

Whatever is most convenient, really.  Crafting your own mission gives you exactly what you want with no complication, but takes longer.  Note you can also load a "generic mission", a relatively new concept in the suite I'm still playing around with.  e.g.

```
        self.load_generic_mission("CMAC-circuit.txt", strict=False)
```

There's something else you might be able to get away with, depending.  Which is to just use "takeoff mode" to the appropriate altitude (it uses TECS).  This may or may not reproduce, we do a few little interesting things in takeoff mode which might cause the problem to not reproduce.

You could possibly do something like this:
```
        self.takeoff(10)
        loc = self.mav.location()
        self.location_offset_ne(loc, 500, 500)
        loc.alt = 5000
        self.do_reposition(loc)
```

> If you agree, we can consider this PR done and I can open an Issue ticket.

Yes - and you should use that ticket number in the `disabled_tests` method so people can look at the issue/analysis.

Tools/autotest/arduplane.py
Georacer: Thanks! For now it's pretty evident (perhaps even more useful) by plotting RCOU[7]. But I'll keep that in mind.

Georacer:
I have updated the PR to switch from a check for receding throttle to a check for correct airspeed regulation.
This is because currently TECS assumes that the TAKEOFF happens at full throttle.

tridge (approved):

tridge:
closing in favor of the test code in https://github.com/ArduPilot/ardupilot/pull/27174

PR #27128, opened 2024-05-22
add script to generate rst from hwdef

peterbarker:
```
pbarker@fx:~/rc/ardupilot(pr/rst-from-hwdef)$ ./Tools/scripts/generate_board_page.py 
pbarker@fx:~/rc/ardupilot(pr/rst-from-hwdef)$ cat /tmp/some_html/Pixhawk6X.rst 

============
UART Mapping
============

 - SERIAL0 -> OTG1 (USB)
 - SERIAL1 -> UART7 (Telem1) (RTS/CTS pins)
 - SERIAL2 -> UART5 (Telem2) (RTS/CTS pins)
 - SERIAL3 -> USART1 (GPS1)
 - SERIAL4 -> UART8 (GPS2)
 - SERIAL5 -> USART2 (Telem3) (RTS/CTS pins)
 - SERIAL6 -> UART4 (User)
 - SERIAL7 -> USART3 (Debug)
 - SERIAL8 -> OTG2 (USB; MAVLink, can be used for SLCAN with protocol change)
```

I'm not sure what else we can usefully generate.  A lot of things are optional or change over time, so the best you can say from a hwdef is "might have a this, might have a that".  The MatekF405-TE etc is a really good example of that....


Hwurzburg:
I think we need to discuss when you get a chance

Issue #27130, opened 2024-05-22
AP_DDS: Improve instructions to launch plane SITL with colcon


Ryanf55:
## Feature request

**Is your feature request related to a problem? Please describe.**

I can't figure out how to cleanly launch ardupilot SITL with `ros2 launch` for a plane

**Describe the solution you'd like**

Clearly document how to run the plane model with ArduPilot's ROS 2 bringup scripts

**Describe alternatives you've considered**

Hard-cod e the absolute path in a bringup shell script

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[x ] Plane
[  ] Rover
[  ] Submarine

**Additional context**

I can't figure out how to tell the CLI arguments to evaluate the path to the package share directory to change from `copter.parm` to something else.
```python3
            DeclareLaunchArgument(
                "defaults",
                default_value=PathJoinSubstitution(
                    [
                        FindPackageShare("ardupilot_sitl"),
                        "config",
                        "default_params",
                        "copter.parm",
                    ]
                ),
                description="Set path to defaults file.",
            ),
 ```

PR #27131, opened 2024-05-22
Tools: ros2: Improve ability to launch plane in headless mode

Ryanf55:
# Purpose

* Install models for other users such as the `plane.parm` that SITL launches with for plane
* Fix bool parsing in some launch args such as mavproxy and console

# Testing

Create a launch file with the contents;
```xml
<?xml version="1.0"?>
<launch>
  <include file="$(find-pkg-share ardupilot_sitl)/launch/sitl_dds_udp.launch.py">
    <arg name="defaults" value="$(find-pkg-share ardupilot_sitl)/config/models/plane.parm"/>
    <arg name="model" value="plane"/>
    <arg name="command" value="arduplane"/>
  </include>
</launch>
```

```bash
ros2 launch your_package your_launch_file.launch.xml \
   refs:=$(pwd)/src/ardupilot/libraries/AP_DDS/dds_xrce_profile.xml \
   map:=True \
   console:=True
```

Then, change map and console to false and observe them not appear.
The plane should boot up just like sim_vehicle, and be able to use mavproxy.
Mavproxy should show a plane icon.

# Issue

Solves https://github.com/ArduPilot/ardupilot/issues/27130

I found this running `colcon test` and saw a ton of windows popping up.

image

# Follow up

I'd like CMake to install the refs file and default that location so you don't need to specify the refs path when you launch in a directory other than ardupilot.

### DIFF
--- a/Tools/ros2/ardupilot_sitl/CMakeLists.txt
+++ b/Tools/ros2/ardupilot_sitl/CMakeLists.txt
@@ -99,6 +99,12 @@ install(DIRECTORY
   DESTINATION share/${PROJECT_NAME}/config/
 )
 
+# Install additional autotest model params.
+install(DIRECTORY
+  ${ARDUPILOT_ROOT}/Tools/autotest/models
+  DESTINATION share/${PROJECT_NAME}/config/
+)
+
 # Install Python package.
 ament_python_install_package(${PROJECT_NAME}
   PACKAGE_DIR src/${PROJECT_NAME}
--- a/Tools/ros2/ardupilot_sitl/src/ardupilot_sitl/launch.py
+++ b/Tools/ros2/ardupilot_sitl/src/ardupilot_sitl/launch.py
@@ -31,6 +31,9 @@ from launch_ros.substitutions import FindPackageShare
 
 from .actions import ExecuteFunction
 
+TRUE_STRING = "True"
+FALSE_STRING = "False"
+BOOL_STRING_CHOICES = set([TRUE_STRING, FALSE_STRING])
 
 class VirtualPortsLaunch:
     """Launch functions for creating virtual ports using `socat`."""
@@ -307,10 +310,10 @@ class MAVProxyLaunch:
             "--non-interactive ",
         ]
 
-        if console:
+        if console == TRUE_STRING:
             cmd.append("--console ")
 
-        if map:
+        if map == TRUE_STRING:
             cmd.append("--map ")
 
         # Create action.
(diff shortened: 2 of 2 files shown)


Tools/ros2/ardupilot_sitl/CMakeLists.txt
```diff
@@ -99,6 +99,12 @@ install(DIRECTORY
   DESTINATION share/${PROJECT_NAME}/config/
 )
 
+# Install additional autotest model params.
+install(DIRECTORY
+  ${ARDUPILOT_ROOT}/Tools/autotest/models
```
srmainwaring, line 104: Do we want everything from `./Tools/autotest/models` or just the `.parm` files?

srmainwaring (approved):
Looks good. Only question is whether we need the additional JSON files from the directory containing the `plane.parm` file.

srmainwaring:
> I'd like CMake to install the refs file and default that location so you don't need to specify the refs path when you launch in a directory other than ardupilot.

The DDS refs file is already installed to `./install/ardupilot_sitl/share/ardupilot_sitl/config/dds_xrce_profile.xml`. Did you have something else in mind?

Ryanf55:
> Looks good. Only question is whether we need the additional JSON files from the directory containing the `plane.parm` file.
I wasn't sure. If you think I should leave them out, I'll add a matcher to only copy `.parm` and `.param`.

Ryanf55:
> > I'd like CMake to install the refs file and default that location so you don't need to specify the refs path when you launch in a directory other than ardupilot.
> 
> The DDS refs file is already installed to `./install/ardupilot_sitl/share/ardupilot_sitl/config/dds_xrce_profile.xml`. Did you have something else in mind?

No,let me try using that in the launch file.

Tools/ros2/ardupilot_sitl/CMakeLists.txt
Ryanf55: I think it's ok to put everything in just in case people need it.

srmainwaring:
> I wasn't sure. If you think I should leave them out, I'll add a matcher to only copy `.parm` and `.param`

It's good as is.

> Nothing needs to change in this PR, so I think the code is good to go as-is.

LGTM. The launch examples at the end of the  [`/Tools/ros2/README`](https://github.com/ArduPilot/ardupilot/blob/master/Tools/ros2/README.md) use the installed ref files.

Ryanf55:
@srmainwaring  Thoughts on backporting this to 4.5?

PR #27133, opened 2024-05-22
Fix arming check for duplicate scripting RC channels and extra RC checks for rover-quiktune

stephendade:
Fixing two (related) issues in here, from a user support session:
1. The arming check for duplicate RC Aux functions didn't check for Scripting functions, as it only checked for Aux function with an ID <256 (Scripting was 300+)
~~2. Added an extra check in the rover-quiktune script to warn the user if they had a mis-configured (or no) RC channel mapped to scripting~~

### DIFF
--- a/libraries/RC_Channel/RC_Channel.cpp
+++ b/libraries/RC_Channel/RC_Channel.cpp
@@ -1798,7 +1798,7 @@ RC_Channel *RC_Channels::find_channel_for_option(const RC_Channel::AUX_FUNC opti
 // duplicate_options_exist - returns true if any options are duplicated
 bool RC_Channels::duplicate_options_exist()
 {
-    uint8_t auxsw_option_counts[256] = {};
+    uint8_t auxsw_option_counts[512] = {};
     for (uint8_t i=0; i<NUM_RC_CHANNELS; i++) {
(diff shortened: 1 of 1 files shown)


rmackay9:
Hi @stephendade,

Thanks for all the investigation and fix for the issue.

Re the Rover quiktune change, doesn't this mean that if the user has not setup an RC switch to "scripting1" and instead uses Mission Planner's Aux Function tab (this is shown on [the Rover QuikTune wiki page](https://ardupilot.org/rover/docs/quiktune.html#installing-the-script)) that they will see an error until they first push the "Low" button?

stephendade:
> Re the Rover quiktune change, doesn't this mean that if the user has not setup an RC switch to "scripting1" and instead uses Mission Planner's Aux Function tab (this is shown on [the Rover QuikTune wiki page](https://ardupilot.org/rover/docs/quiktune.html#installing-the-script)) that they will see an error until they first push the "Low" button?

Ahh, wasn't aware of that function. I'll remove that check.

peterbarker (approved):
Any bigger and we'll need to change to the N**2 algorithm here.

peterbarker:
Merged, thanks!

PR #27134, opened 2024-05-22
Copter: Enable processing

muramura:
This process will never be executed, so I will enable it. 
I believe that processes that are never executed are unnecessary.


ArduCopter/mode_guided.cpp
```diff
@@ -409,10 +409,9 @@ bool ModeGuided::get_wp(Location& destination) const
         destination = Location(guided_pos_target_cm.tofloat(), guided_pos_terrain_alt ? Location::AltFrame::ABOVE_TERRAIN : Location::AltFrame::ABOVE_ORIGIN);
         return true;
     default:
-        return false;
+        break;
```
peterbarker, line 412: There should be no default here - we should *never* be in an invalid `SubMode`!

muramura: @peterbarker san.
I have described all the definitions.

peterbarker (approved):
LGTM

IamPete1 (approved):

PR #27136, opened 2024-05-22
Use active ESC when only one can be displayed

andyp1per:
Currently the OSD displays ESC 0 only. This attempts to show the most relevant information

### DIFF
--- a/libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
+++ b/libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
@@ -120,6 +120,30 @@ uint32_t AP_ESC_Telem::get_active_esc_mask() const {
     return ret;
 }
 
+// return an active ESC for the purposes of reporting (e.g. in the OSD)
+uint8_t AP_ESC_Telem::get_max_rpm_esc() const
+{
+    uint32_t ret = 0;
+    float max_rpm = 0;
+    const uint32_t now = AP_HAL::millis();
+    const uint32_t now_us = AP_HAL::micros();
+    for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
+        if (_telem_data[i].last_update_ms == 0 && !was_rpm_data_ever_reported(_rpm_data[i])) {
+            // have never seen telem from this ESC
+            continue;
+        }
+        if (_telem_data[i].stale(now)
+            && !rpm_data_within_timeout(_rpm_data[i], now_us, ESC_RPM_DATA_TIMEOUT_US)) {
+            continue;
+        }
+        if (_rpm_data[i].rpm > max_rpm) {
+            max_rpm = _rpm_data[i].rpm;
+            ret = i;
+        }
+    }
+    return ret;
+}
+
 // return number of active ESCs present
 uint8_t AP_ESC_Telem::get_num_active_escs() const {
     uint32_t active = get_active_esc_mask();
--- a/libraries/AP_OSD/AP_OSD_Screen.cpp
+++ b/libraries/AP_OSD/AP_OSD_Screen.cpp
@@ -358,7 +358,7 @@ const AP_Param::GroupInfo AP_OSD_Screen::var_info[] = {
 #if HAL_WITH_ESC_TELEM
     // @Param: ESCTEMP_EN
     // @DisplayName: ESCTEMP_EN
-    // @Description: Displays first esc's temp
+    // @Description: Displays highest temp of all active ESCs, or of a specific ECS if OSDx_ESC_IDX is set
     // @Values: 0:Disabled,1:Enabled
 
     // @Param: ESCTEMP_X
@@ -374,7 +374,7 @@ const AP_Param::GroupInfo AP_OSD_Screen::var_info[] = {
 
     // @Param: ESCRPM_EN
     // @DisplayName: ESCRPM_EN
-    // @Description: Displays first esc's rpm
+    // @Description: Displays highest rpm of all active ESCs, or of a specific ESC if OSDx_ESC_IDX is set
     // @Values: 0:Disabled,1:Enabled
 
     // @Param: ESCRPM_X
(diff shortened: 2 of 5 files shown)


libraries/AP_OSD/AP_OSD_Screen.cpp
```diff
@@ -2004,7 +2004,7 @@ void AP_OSD_Screen::draw_esc_temp(uint8_t x, uint8_t y)
 {
     int16_t etemp;
     // first parameter is index into array of ESC's.  Hardwire to zero (first) for now.
```
IamPete1, line 2006: Comment is now wrong. I also think you should update the X Y params for this to explain its now the highest.

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -120,6 +120,30 @@ uint32_t AP_ESC_Telem::get_active_esc_mask() const {
...
+    uint32_t now_us = AP_HAL::micros();
+    for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
+        if (_telem_data[i].last_update_ms == 0 && !was_rpm_data_ever_reported(_rpm_data[i])) {
+            // have never seen telem from this ESC
+            continue;
+        }
+        if (_telem_data[i].stale(now)
+            && !rpm_data_within_timeout(_rpm_data[i], now_us, ESC_RPM_DATA_TIMEOUT_US)) {
+            continue;
+        }
```
IamPete1, line 138: I think this is confusing. I would rather see that the ESC is not working than have it change to the next one and hide the issue.

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -120,6 +120,30 @@ uint32_t AP_ESC_Telem::get_active_esc_mask() const {
...
+        if (_rpm_data[i].rpm > max_rpm) {
+            max_rpm = _rpm_data[i].rpm;
+            ret = i;
+        }
```
IamPete1, line 142: Its actually picking the one with the highest RPM?

IamPete1 (dismissed):

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
IamPete1: Maybe just two OSD items? "ESC max of all" which is what I think this code does? and add a new index param to the current ESC item with the default being to do whatever it is we do now. 

Another option might be to display the ESC number which the data is for alongside the data in the OSD.

libraries/AP_OSD/AP_OSD_Screen.cpp
andyp1per: Fixed

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
andyp1per: Fixed

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
andyp1per: Fixed

libraries/AP_OSD/AP_OSD_Screen.cpp
```diff
@@ -1166,6 +1166,13 @@ const AP_Param::GroupInfo AP_OSD_Screen::var_info2[] = {
     AP_SUBGROUPINFO(rc_lq, "RC_LQ", 9, AP_OSD_Screen, AP_OSD_Setting),
 #endif
 
+#if HAL_WITH_ESC_TELEM
+    // @Param: ESC_IDX
+    // @DisplayName: ESC_IDX
+    // @Description: Index of the ESC to use for displaying ESC information. 0 means use the most active ESC.
```
IamPete1, line 1172: I'm not a fan of "most active". What does it mean? Maybe just "highest RPM".

andyp1per: Fixed, plus the other bad bug you didn't spot ;)

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
IamPete1: This is still picking the highest?

IamPete1: Ah, I think the plan is now that that is the correct thing to do. Just the function name and comment don't say that.

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -120,6 +120,30 @@ uint32_t AP_ESC_Telem::get_active_esc_mask() const {
     return ret;
 }
 
+// return an active ESC for the purposes of reporting (e.g. in the OSD)
+uint8_t AP_ESC_Telem::get_active_esc() const
```
IamPete1, line 124:
```suggestion
// return the index of the ESC with the highest RPM
uint8_t AP_ESC_Telem::get_max_RPM_index() const
```

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -120,6 +120,30 @@ uint32_t AP_ESC_Telem::get_active_esc_mask() const {
...
+{
+    uint32_t ret = 0;
+    float max_rpm = 0;
+    const uint32_t now = AP_HAL::millis();
+    uint32_t now_us = AP_HAL::micros();
```
peterbarker, line 129:
```suggestion
    const uint32_t now_us = AP_HAL::micros();
```

tridge:
can be merged once name is changed from "active" to something indicating highest RPM

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
andyp1per: fixed

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
andyp1per: fixed

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
andyp1per: fixed

tridge (approved):

PR #27138, opened 2024-05-22
Scale adjusted because QMC5883L modules deliver a multiple value

dg9oaa:


andyp1per (changes requested):
We use many undocumented registers in ArduPIlot - just because they are not in the datasheet does not mean they don't exist. Also the 4x scale cannot generally be correct because these devices are used everywhere and no-one is having problems. What is the problem you are trying to solve?

dg9oaa:
I have several airplanes and drones that I have equipped with GPS modules. Among them are BN-880 and HGLRC M80 PRO. I have used the BN-880s that are also suitable for APM and Pixhawk according to the description. Most of the time, however, a QMC5883 was fitted. The compass could never be calibrated with this chip. In the telemetry of the remote control transmitter I saw that the Gaussian values reached the 4000 mark quite quickly and showed virtually no sine function with further rotation.
I then used the really great MAGFit web tool to calibrate the compass by generating the correction parameters. I noticed that the measured Gaussian values were very large and the error message "PreArm: Check mag field (xy diff:373>100)" appeared in the messages. Of course several times and with different values.

I then started to read out the compass modules with a Raspberry Pi Pico via i2c and took a closer look at the measured Gauss values. I also played around with the parameter settings. I noticed that setting the Gauss Scale Rate does not change the output value, except for a single module and that was an Arduino test board.

So I think it makes sense to leave the Gaussian rate at the default and simply divide the output value by 4. I also got the idea from an INAV, as there is a scale factor that was set to ~4000 for my tested modules.

The other idea I had was to introduce such a scale factor in Adopilot, which would probably be more complex.

Best regards and thank you very much for your great work

andyp1per:
Thanks for your input. So its one particular compass that isn't working?

dg9oaa:
All QMC5883 modules (some BN-880, HGLRC M80 PRO and an Arduino test board) that I have tried do not work with Ardupilot.

PR #27141, opened 2024-05-23
Copter: ModeGuided: add an option_is_enabled method, use it

peterbarker:


peterbarker:
Added a test for one of the options, others done my inspection only

ArduCopter/mode_guided.cpp
```diff
@@ -1009,7 +1014,7 @@ void ModeGuided::set_yaw_state(bool use_yaw, float yaw_cd, bool use_yaw_rate, fl
 // returns true if pilot's yaw input should be used to adjust vehicle's heading
 bool ModeGuided::use_pilot_yaw(void) const
 {
-    return (copter.g2.guided_options.get() & uint32_t(Options::IgnorePilotYaw)) == 0;
+    return option_is_enabled(Option::IgnorePilotYaw);
```
IamPete1, line 1017:
```suggestion
    return !option_is_enabled(Option::IgnorePilotYaw);
```

IamPete1 (changes requested):

ArduCopter/mode.h
```diff
@@ -1138,6 +1138,8 @@ class ModeGuided : public Mode {
         AllowWeatherVaning = (1U << 7)
     };
 
+    bool option_is_enabled(Option option) const;
```
rmackay9, line 1141: let's put a little comment above this one.  sorry, it's perhaps pretty obvious what it does but..

ArduCopter/mode_guided.cpp
```diff
@@ -97,6 +97,11 @@ void ModeGuided::run()
     }
  }
 
+bool ModeGuided::option_is_enabled(Option option) const
```
rmackay9, line 100: put a comment above here too please!

rmackay9:
feel free to merge after those couple of small fixes.

ArduCopter/mode.h
peterbarker: Done

ArduCopter/mode_guided.cpp
peterbarker: Done

peterbarker:
I've confirmed that using the `copter` singleton in the mode header does not work trivially, and my guess is that we'll end up with an include loop if we try to make it work.

PR #27142, opened 2024-05-23
Lua binding added for landing target message

Ullagaddi:


IamPete1 (changes requested):
PR should be against the master branch. 

I think we should either add a scripting backend, or add the ability to send generic MAVLink messages to the vehicle. 

It also needs docs.

PR #27150, opened 2024-05-24
Tools: Install uxrceddsgen in Ubuntu

Ryanf55:
# Purpose

Allow users to more easily install MicroXRCEDDSGen without having to compile it using gradle, for users who use ROS 2 with Ubuntu. This is optional during setup.

### DIFF
--- a/Tools/environment_install/install-prereqs-ubuntu.sh
+++ b/Tools/environment_install/install-prereqs-ubuntu.sh
@@ -482,6 +482,44 @@ grep -Fxq "$exportline4" ~/$SHELL_LOGIN 2>/dev/null || {
         echo "Skipping appending CCache to PATH."
     fi
 }
+
+
+UXRCE_GEN_TOOL_NAME="ArduPilot's MicroXRCEDDSGen tool for ROS 2"
+if [[ -z "${DO_UXRCGEN_INSTALL}" ]] && maybe_prompt_user "Install $UXRCE_GEN_TOOL_NAME [N/y]?" ; then
+    DO_UXRCGEN_INSTALL=1
+fi
+
+if [[ $DO_UXRCGEN_INSTALL -eq 1 ]]; then
+    if ! package_is_installed "default-jre"; then
+        heading "Installing default-jre"
+        $APT_GET install default-jre
+        echo "Done!"
+    fi
+
+    if [[ -z "${UXRCE_GEN_ROOT}" ]]; then
+        UXRCE_GEN_ROOT=$OPT/ardupilot_uxrce
+    fi
+
+    heading "Installing $UXRCE_GEN_TOOL_NAME"
+    echo "Downloading from ArduPilot release"
+    sudo mkdir -p $UXRCE_GEN_ROOT
+    cd $UXRCE_GEN_ROOT
+    sudo wget --progress=dot:giga -P $UXRCE_GEN_ROOT/scripts https://github.com/ArduPilot/Micro-XRCE-DDS-Gen/releases/download/v4.5.0/microxrceddsgen
+    sudo chmod +x scripts/microxrceddsgen
+    sudo wget --progress=dot:giga -P $UXRCE_GEN_ROOT/share/microxrceddsgen/java https://github.com/ArduPilot/Micro-XRCE-DDS-Gen/releases/download/v4.5.0/microxrceddsgen.jar
+
+    exportline="export PATH=$UXRCE_GEN_ROOT/scripts:\$PATH";
+    grep -Fxq "$exportline" ~/$SHELL_LOGIN 2>/dev/null || {
+        if maybe_prompt_user "Add $UXRCE_GEN_ROOT to your PATH [N/y]?" ; then
+            echo $exportline >> ~/$SHELL_LOGIN
+            eval $exportline
+            microxrceddsgen -version
+        else
+            echo "Skipping adding $UXRCE_GEN_ROOT to PATH."
+        fi
+    }
+fi
+
 echo "Done!"
 
 if [[ $SKIP_AP_GIT_CHECK -ne 1 ]]; then


Ryanf55:
Unrelated failure: https://github.com/Ryanf55/ardupilot/actions/runs/9230437664/job/25398673460#step:6:116

Ryanf55:
[0001-Github-Enable-DDS-in-environment-setup-testing.patch](https://github.com/ArduPilot/ardupilot/files/15440681/0001-Github-Enable-DDS-in-environment-setup-testing.patch)

Also, this patch could be helpful to ensure DDS is actually going to work.

Tools/environment_install/install-prereqs-ubuntu.sh
```diff
@@ -482,6 +482,44 @@ grep -Fxq "$exportline4" ~/$SHELL_LOGIN 2>/dev/null || {
...
+    fi
+
+    if [[ -z "${UXRCE_GEN_ROOT}" ]]; then
+        UXRCE_GEN_ROOT=$OPT/ardupilot_uxrce
+    fi
+
+    heading "Installing $UXRCE_GEN_TOOL_NAME"
+    echo "Downloading from ArduPilot release"
+    sudo mkdir -p $UXRCE_GEN_ROOT
+    cd $UXRCE_GEN_ROOT
```
Georacer, line 506: Shouldn't you return to the previous directory after this install step is done?

Tools/environment_install/install-prereqs-ubuntu.sh
```diff
@@ -482,6 +482,44 @@ grep -Fxq "$exportline4" ~/$SHELL_LOGIN 2>/dev/null || {
...
+    sudo wget --progress=dot:giga -P $UXRCE_GEN_ROOT/share/microxrceddsgen/java https://github.com/ArduPilot/Micro-XRCE-DDS-Gen/releases/download/v4.5.0/microxrceddsgen.jar
+
+    exportline="export PATH=$UXRCE_GEN_ROOT/scripts:\$PATH";
+    grep -Fxq "$exportline" ~/$SHELL_LOGIN 2>/dev/null || {
+        if maybe_prompt_user "Add $UXRCE_GEN_ROOT to your PATH [N/y]?" ; then
+            echo $exportline >> ~/$SHELL_LOGIN
+            eval $exportline
+            microxrceddsgen -version
+        else
+            echo "Skipping adding $UXRCE_GEN_ROOT to PATH."
```
Georacer, line 518: Do you indeed want to skip adding to path if the `-y` parameter was passed?

Georacer (commented):
Hi!
Apart from the two comments, a couple more questions:

First, the flags `DO_UXRCGEN_INSTALL` and `UXRCE_GEN_TOOL_NAME` are not visible anywhere. Where do you propose that they will be documented?

If I understand this correctly, you download `microxrceddsgen` and `microxrceddsgen.jar`, but you don't actually use them to produce the DDS libraries. Is this really that useful? The user has to run the generation command anyways.

Also, in your PR descriptions you mention "for users who use ROS2". I don't see a mechanism where these will not be installed with `-y` if a user doesn't plan to user ROS.

Tools/environment_install/install-prereqs-ubuntu.sh
```diff
@@ -482,6 +482,44 @@ grep -Fxq "$exportline4" ~/$SHELL_LOGIN 2>/dev/null || {
...
+        UXRCE_GEN_ROOT=$OPT/ardupilot_uxrce
+    fi
+
+    heading "Installing $UXRCE_GEN_TOOL_NAME"
+    echo "Downloading from ArduPilot release"
+    sudo mkdir -p $UXRCE_GEN_ROOT
+    cd $UXRCE_GEN_ROOT
+    sudo wget --progress=dot:giga -P $UXRCE_GEN_ROOT/scripts https://github.com/ArduPilot/Micro-XRCE-DDS-Gen/releases/download/v4.5.0/microxrceddsgen
+    sudo chmod +x scripts/microxrceddsgen
+    sudo wget --progress=dot:giga -P $UXRCE_GEN_ROOT/share/microxrceddsgen/java https://github.com/ArduPilot/Micro-XRCE-DDS-Gen/releases/download/v4.5.0/microxrceddsgen.jar
```
peterbarker, line 509: I wonder if these artifacts should be committed to the tree, and/or whether we should be cloning the entire repo rather than grabbing artifacts.

Tools/environment_install/install-prereqs-ubuntu.sh
Ryanf55: No need. The others don't, they use absolute paths for directory changes.

Ryanf55:
> Hi! Apart from the two comments, a couple more questions:
> 
> First, the flags `DO_UXRCGEN_INSTALL` and `UXRCE_GEN_TOOL_NAME` are not visible anywhere. Where do you propose that they will be documented?
> 
> If I understand this correctly, you download `microxrceddsgen` and `microxrceddsgen.jar`, but you don't actually use them to produce the DDS libraries. Is this really that useful? The user has to run the generation command anyways.
> 
> Also, in your PR descriptions you mention "for users who use ROS2". I don't see a mechanism where these will not be installed with `-y` if a user doesn't plan to user ROS.

As best I could, I followed existing patterns. That includes the use of undocumented flags, which are only changed in CI. Thus, I propose doing what's already done and adding more flags, that only CI touches. 

No, that is not correct. We download those, which are used to produce the ardupilot binaries. Once the binaries are created, microxrceddsgen is not needed at runtime. 

If a user says "n" to the question, it won't be installed, just like the other optional features in setup. If it doesn't do that, it's a bug.

peterbarker:
```
[8:25 AM]Peter Barker: How often does it change?  How co-dependent is the ArduPilot build on it?  i.e. the install script is run once, so gets the jar once.  If it is out of date/non-functional in 2 months then that's a problem as everyone's DDS setup breaks.
[8:59 AM]RyanF: I hasn't changed since 4.5, but we expect one major version bump before 4.6.
[9:25 AM]Peter Barker: .... breaking everyone's DDS setups.  Thus my question about adding it as an artifact to the ArduPilot repo.  How big is it?
[9:28 AM]RyanF: 406B for the script and 19M for the jar.
```

Ryanf55:
Can we try this workflow out first? We have the 4.5.1 version that fixed some extranneous stderr behavior, but otherwise expect no changes. This PR makes a good improvement significant improvement in user setup time lowering the barrier to entry in using ROS 2.

PR #27152, opened 2024-05-24
GCS_MAVLink: Add support for parity to Serial passthrough

brad112358:
Add code to reflect USB ACM parity setting to the passthrough port alongside existing support for ACM baud rate changes.  Some use cases for serial passthrough require specific parity settings.

For example, even parity is used and required by the USART protocol used in the STM32 system bootloader. This enhancement allows the use of standard flash programming tools such as STM32CubeProgrammer to flash connected STM based peripherals such as Receivers and Telemetry radios via serial passthrough.  Some examples of such peripherals include the FrSky R9 receivers as well as various other STM based LoRa modules used by the mLRS project.

### DIFF
--- a/libraries/AP_HAL/UARTDriver.cpp
+++ b/libraries/AP_HAL/UARTDriver.cpp
@@ -179,6 +179,11 @@ bool AP_HAL::UARTDriver::flow_control_enabled(enum flow_control flow_control_set
     return false;
 }
 
+uint8_t AP_HAL::UARTDriver::get_parity(void)
+{
+    return AP_HAL::UARTDriver::parity;
+}
+
 #if HAL_UART_STATS_ENABLED
 // Take cumulative bytes and return the change since last call
 uint32_t AP_HAL::UARTDriver::StatsTracker::ByteTracker::update(uint32_t bytes)
--- a/libraries/AP_HAL/UARTDriver.h
+++ b/libraries/AP_HAL/UARTDriver.h
@@ -78,6 +78,9 @@ public:
 
     // read buffer from a locked port. If port is locked and key is not correct then -1 is returned
     ssize_t read_locked(uint8_t *buf, size_t count, uint32_t key) WARN_IF_UNUSED;
+
+    // get current parity for passthrough use
+    uint8_t get_parity(void);
     
     // control optional features
     virtual bool set_options(uint16_t options) { _last_options = options; return options==0; }
@@ -189,6 +192,9 @@ public:
     // return true requested baud on USB port
     virtual uint32_t get_usb_baud(void) const { return 0; }
 
+    // return requested parity on USB port
+    virtual uint8_t get_usb_parity(void) const { return parity; }
+
     // disable TX/RX pins for unusued uart
     virtual void disable_rxtx(void) const {}
 
@@ -212,6 +218,8 @@ protected:
     // key for a locked port
     uint32_t lock_write_key;
     uint32_t lock_read_key;
+
+    uint8_t parity;
 
     /*
       backend begin method
--- a/libraries/AP_HAL_ChibiOS/UARTDriver.cpp
+++ b/libraries/AP_HAL_ChibiOS/UARTDriver.cpp
@@ -672,6 +672,19 @@ uint32_t UARTDriver::get_usb_baud() const
     return 0;
 }
 
+/*
+    get the requested usb parity.  Valid if get_usb_baud() returned non-zero.
+*/
+uint8_t UARTDriver::get_usb_parity() const
+{
+#if HAL_USE_SERIAL_USB
+    if (sdef.is_usb) {
+        return ::get_usb_parity(sdef.endpoint_id);
+    }
+#endif
+    return 0;
+}
+
 uint32_t UARTDriver::_available()
 {
     if (!_rx_initialised || _uart_owner_thd != chThdGetSelfX()) {
@@ -1429,6 +1442,7 @@ void UARTDriver::configure_parity(uint8_t v)
         // not possible
         return;
     }
+    UARTDriver::parity = v;
 #if HAL_USE_SERIAL == TRUE
     // stop and start to take effect
     sdStop((SerialDriver*)sdef.serial);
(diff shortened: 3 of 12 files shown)


olliw42:
very very cool !!
👍

olliw42:
tested and found to work great !!
good job, sir
👍

andyp1per (approved):
It looks ok to me, I am not sure about the var recording parity - but seems like it should work. An alternative would be to make the function virtual and pull the parity out of the underlying uart

brad112358:
> It looks ok to me, I am not sure about the var recording parity - but seems like it should work. An alternative would be to make the function virtual and pull the parity out of the underlying uart

I thought about doing it that way, but this is simpler and it's only 1 byte of data.  Also, the very lightest flight stacks (AIO boards) are still only available with 1MB flash the code size should be slightly smaller this way.

brad112358:
I resolved the conflict caused by recent updates to master, rebased and force pushed

tridge (changes requested):
I think we need a SERIAL_PASSOPT parameter with a bitmask of options, with an option bit to enable this feature
the problem with this PR is if your device needs a particular parity and the client you are using doesn't support setting it on USB then you can no longer use passthru as the exiting parity setting in SERIALn_OPTIONS will be cleared. So I think we need an option bit to enable parity mapping
I'd also like someone to test this with u-Center pass-thru on ublox to ensure that still works, as that is the most common use case. Please test with a fw update of a ublox GPS

olliw42:
many thanks for looking at it.

first off, as far as we know there is no "exiting parity setting in SERIALn_OPTIONS". That is actually the whole reason for why Brad did look into this, since we needed it but it wasn't there.

Being able to properly "decouple" the two involved serials is certainly an important point, which we indeed have overlooked a bit. I think this needs some more thinking. Brad and I will do that externally, but We'll be back.

brad112358:
I apologize in advance if everyone has already considered what I'm about to write here:

I could have been more clear in the original PR comment.  The only use case this PR is addressing is where a flight controller is connected via USB with CDC-ACM class to a PC and we want to access a serial peripheral on that FC from the PC via passthrough as a handy alternative to disconnecting the peripheral from the FC and connecting it directly to a serial port on the PC.

In this case, we know the PC software which will be used to interact with the peripheral is capable of accessing the peripheral via a serial connection on the PC.  I am further assuming that the serial connection works with the standard OS capabilities provided by all serial ports so that the software needs no special low level driver installed to interact with the peripheral.

I am far more familiar with Linux internals and drivers than I am with Windows, but I believe the following statements apply to Windows and MacOS as well as Linux.  I did verify this works with the STM bootloader from Windows in addition to Linux.  When a CDC-ACM class USB device is connected, a special serial device instance is created.  On Linux this device is accessed as /dev/ttyACM0, for example.  On windows, it would be something like com9.  When the baud rate or parity is set by an application using this serial device, the OS level CDC-ACM driver sends these settings via the USB connection to the FC and in the case of parity settings, we act on that information as you see in this PR.

Since the baud and parity settings are relayed by the OS level driver and not explicitly by the application and since the application can be assumed to set the correct parity when it interacts directly with the device via a simple serial port, I believe there is no reason to worry that the proper parity might not be set when the application interacts with the serial device associated with the FC via the CDC-ACM class USB connection.  So I don't believe a SERIAL_PASSOPT parameter would ever be needed in this use case.  

@tridge, does this help?  Do you know of a case where an application platform supports CDC-ACM but sends incorrect information in the parity field?  If so, does it send correct info in the baud field or do we need a way to manually configure that too?

A second use case which @olliw42 mentioned might be considered where the connection between the FC and the PC is via a simple serial port instead of via USB.  In this case, we might want something like SERIAL_PASSOPT to correctly configure the serial port on which the peripheral is connected.  In the case of peripheral firmware update via a bootloader, we probably also shouldn't assume the peripheral will communicate at the same baud rate which we have configured for normal use, so if we want to avoid temporary changes to the non-passthrough configuration, we could also add a way to configure the passthrough baud.

This second use case was beyond the scope I intended for this PR.  If we believe a simple serial connection to between the PC and FC is common enough to justify the extra parameters and development effort, I'd suggest this work be undertaken in a separate PR.

brad112358:
With regard to additional testing with GPS, I suspect all of my (low cost) GPS modules are ROM based and don't allow firmware update.  I guess I can check to see if I can find a flash based GPS module somewhere.  @andyp1per, could you find time to test this with ublox GPS firmware update?

brad112358:
> Upgrade worked fine

Thanks for taking the time to test it!

tridge (approved):
discussed dev call, will accept, we may add the option param in the future

olliw42:
cool! THX.

brad112358:
Thanks!

brad112358:
I probably should have asked sooner, but I didn't realize 4.6 wouldn't be out soon.  Now I can't even find anything which shows when the next major Ardupilot release is expected.  If would be nice to have this functionality available to support updating mLRS firmware on the new Mateksys mLRS mR900-22 receiver without unsoldering it.  Backporting this should be safe and easy enough I think.  Is there a better way to request this?

rmackay9:
Hi @brad112358,

We will start 4.6.0 beta testing by the end of October but it normally takes at least 2 months to get through beta testing.  If the PR's "project" is set to "4.5 backports" then it is considered for backporting to the next stable point release.  I've gone ahead and done that although no promises of course

brad112358:
Thanks!

rmackay9:
I could not easily include this in 4.5.7-beta1 due to a merge conflict.  If someone is keen to get this in then a backport PR would be greatly appreciated!

brad112358:
> I could not easily include this in 4.5.7-beta1 due to a merge conflict. If someone is keen to get this in then a backport PR would be greatly appreciated!

I'm not sure how this works and a quick search of the wiki found nothing.  I assume this just requires a PR against one of the 4.5 branches.  Which one?  Is Copter-4.5 OK?  Or do you need PRs for all of the 4.5 branches?

brad112358:
@rmackay9 I went ahead and created a PR #28232 against Copter-4.5.  Please let me know if I should do anything else.

PR #27158, opened 2024-05-25
ADSB: Sagetech MXS: updates

IamPete1:
This includes some driver minor re-work and adds the ability to request a "summary" at a rate given by a new param. This gives a way of limiting the data rate. Currently we use auto reporting which sends a message every time a ADSB transmission is received, if there is a lot of traffic about this can result in very high data rates.  

There are also some changes to remove the test strings set for flight ID and aircraft registration. Both serial ports are also set to 230400 so you can use either. 

Our sim is updated to support the new  functionality.

### DIFF
--- a/libraries/AP_ADSB/AP_ADSB.cpp
+++ b/libraries/AP_ADSB/AP_ADSB.cpp
@@ -62,6 +62,9 @@ extern const AP_HAL::HAL& hal;
 
 AP_ADSB *AP_ADSB::_singleton;
 
+// Per backend params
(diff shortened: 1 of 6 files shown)


libraries/AP_ADSB/AP_ADSB.cpp
```diff
@@ -176,6 +179,10 @@ const AP_Param::GroupInfo AP_ADSB::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("OPTIONS",  15, AP_ADSB, _options, 0),
 
+    // @Group:
```
IamPete1, line 182: This group with a empty name does not get picked up by the param docs. Works fine in AP tho.

PR #27161, opened 2024-05-26
AP_DDS: Use common quaternion initialization function

Ryanf55:
A common initialization function. Follow up to the previous PR.
https://github.com/ArduPilot/ardupilot/pull/27140#discussion_r1614957993

### DIFF
--- a/libraries/AP_DDS/AP_DDS_Client.cpp
+++ b/libraries/AP_DDS/AP_DDS_Client.cpp
@@ -77,6 +77,14 @@ const AP_Param::GroupInfo AP_DDS_Client::var_info[] {
     AP_GROUPEND
 };
 
+static void initialize(geometry_msgs_msg_Quaternion& q)
+{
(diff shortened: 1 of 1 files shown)


libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -77,6 +77,14 @@ const AP_Param::GroupInfo AP_DDS_Client::var_info[] {
     AP_GROUPEND
 };
 
+static inline void initialize(geometry_msgs_msg_Quaternion& q)
```
tridge, line 80: should not be inline

tridge (changes requested):

Ryanf55:
Needs squash  unless you can autosquash on merge.

Ryanf55:
> Needs squash unless you can autosquash on merge.

Squashed. Ready for final check.

PR #27163, opened 2024-05-26
AP_Scripting: added notch_switch example

tridge:
switch between two notch setups using attenuation change

### DIFF
--- a/libraries/AP_Scripting/examples/notch_switch.lua
+++ b/libraries/AP_Scripting/examples/notch_switch.lua
@@ -1,0 +1,48 @@
+--[[
+
+   example script to switch between two notch setups by changing the
+   attenuation to zero on the notch to disable. This allows for easy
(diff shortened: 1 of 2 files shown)


libraries/AP_Scripting/examples/notch_switch.lua
```diff
@@ -0,0 +1,46 @@
...
+local AUX_FN = 300
+
+local attenuation = INS_HNTCH_ATT:get()
+if not attenuation then
+   gcs:send_text(0, string.format("Unable to get attenuation"))
+   return
+end
+
+function update()
+   local sw_current = rc:get_aux_cached(AUX_FN)
```
IamPete1, line 25: I think this works in lua:
```suggestion
   local sw_current = rc:get_aux_cached(AUX_FN) or 0
```
Although I may be wrong....

peterbarker (approved):

PR #27164, opened 2024-05-26
Filter: added test of phase lag vs attenuation

tridge:
image

this was for a single notch set at 60Hz with bandwidth of 60Hz. Source frequency was 50Hz, and the test frequency varies for the 5 graphs from 10Hz to 50Hz.

### DIFF
--- a/libraries/Filter/tests/plot_harmonictest5.gnu
+++ b/libraries/Filter/tests/plot_harmonictest5.gnu
@@ -1,0 +1,12 @@
+#!/usr/bin/gnuplot -persist
+set y2tics 0,10
+set ytics nomirror
+set style data linespoints
+set key autotitle
+set datafile separator ","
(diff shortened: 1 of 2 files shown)


IamPete1:
I guess if you were to interpolate the crossing point it would fix the stepping.

IamPete1:
Fix here: https://github.com/IamPete1/ardupilot/commit/25ccdb3b99c4082b667673754f1d0dfa22df396b

image

We do pick up some funny stuff at 0 attenuation because we don't deal with the phase wrap correctly, moving to floating point math means we get a phase of about 360 and about 0 so the averaging means we get some value in between. 

To elaborate slight more on the issue, because the noise frequency of 50Hz is a exact multiple of the 200Hz sample rate we get the samples happening at the same point in the sine every time. This means the error does not average out.

PR #27167, opened 2024-05-26
AP_Notify: Convert the ENUM definition to a class

muramura:
I think converting the ENUM definition to a class will make it clearer.

### DIFF
--- a/libraries/AP_Notify/RGBLed.h
+++ b/libraries/AP_Notify/RGBLed.h
@@ -105,12 +105,12 @@ private:
     const uint32_t sequence_disarmed_bad_gps = DEFINE_COLOUR_SEQUENCE_SLOW(BLUE);
 
     uint8_t last_step;
-    enum rgb_source_t {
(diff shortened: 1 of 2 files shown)


libraries/AP_Notify/RGBLed.h
```diff
@@ -105,12 +105,12 @@ class RGBLed: public NotifyDevice {
     const uint32_t sequence_disarmed_bad_gps = DEFINE_COLOUR_SEQUENCE_SLOW(BLUE);
 
     uint8_t last_step;
-    enum rgb_source_t {
+    enum class rgbSource {
```
peterbarker, line 108:
```suggestion
    enum class Source {
```

peterbarker (changes requested):

libraries/AP_Notify/RGBLed.h
muramura: @peterbarker san.
I agreed and made the change.

Issue #22615, opened 2023-01-09
clang fix needed for the equivalent gcc fix here: https://github.com/ArduPilot/ardupilot/pull/22614


davidbuzz:
## Bug report

new() expects exceptions, and we compile without them, and so when any compiler optimisation level above -O0 is used, godbolt shows that some did-new-return-nullptr checks are elided by the optimisations that shouldn't be.

see the gcc fix here https://github.com/ArduPilot/ardupilot/pull/22614

_Please describe the problem_

**Version**
_What version was the issue encountered with_

**Platform**
[  X] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
_What type of airframe (flying wing, glider, hex, Y6, octa etc)_

**Hardware type**
_What autopilot hardware was used? (Pixhawk, Cube, Pixracer, Navio2, etc)_

**Logs**
_Please provide a link to any relevant logs that show the issue_

davidbuzz:
clang understands check-new but doesn't use it.

davidbuzz:
the linux boards are the ones usually compiled with clang, so this is mostly gonna afffect them.

davidbuzz:
... above my paygrade to fix clang. :-)

khancyr:
why not use C++ "smart pointer" like unique_ptr, this is the normal replacement for new() and is noexcept ?

davidbuzz:
@khancyr  - thanks , but what this issue needs is an actual fix, not random unimplemented suggestion/s.   if you have a PR that fixes this issue using 'smart pointer' that would be awesome, but there's probably a number of other ways to fix it too.

peterbarker:
> @khancyr - thanks , but what this issue needs is an actual fix, not random unimplemented suggestion/s. if you have a PR that fixes this issue using 'smart pointer' that would be awesome, but there's probably a number of other ways to fix it too.

No, that wouldn't be awesome.

I loathe the smart pointer stuff.

See https://github.com/ArduPilot/ardupilot/pull/16138

kknives:
[...]

PR #22614, opened 2023-01-09
waf: add -fcheck-new to g++ build

tridge:
this ensures the compiler doesn't assume that new always returns a non-NULL value. Without this the compiler may remove the error path in code like this:

```
MyObject *x = new MyObject;
if (x == nullptr) {
  ::printf("Alloc failed\n");
}
```

the reason it can do this is the new operator is marked as throwing an exception on failure, which means the error path is unreachable. As we don't have C++ exceptions in ArduPilot could (and does!) have code that ends up losing protection against allocation failures

@andyp1per I believe this is possible the cause of your PixracerPro errors

an example of the impact of this is here:
```
void AP_RollController::autotune_start(void)
{
    if (autotune == nullptr) {
        autotune = new AP_AutoTune(gains, AP_AutoTune::AUTOTUNE_ROLL, aparm, rate_pid);
        if (autotune == nullptr) {
            if (!failed_autotune_alloc) {
                GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "AutoTune: failed roll allocation");
            }
            failed_autotune_alloc = true;
        }
    }
    if (autotune != nullptr) {
        autotune->start();
    }
}
```
the compiler was removing the nullptr check, and would run autotune->start() on a nullptr if we ran out of memory. The "AutoTune: failed roll allocation" string does not appear anywhere in the binary without this PR. The compiler thought it could remove the nullptr check as it assumed that if new failed then an exception would be raised, so it wouldn't get to that line.
A test on a H7 that has been deliberately run low on memory was done, then fixed wing auto-tune was started. The board did a null ptr dereference

PR #27169, opened 2024-05-27
Moved from using new to new(std::nothrow)

tridge:
Unfortunately clang++ accepts -fcheck-new on the command line but does not implement it, which means our strategy of relying on -fcheck-new to avoid the problem of elided code for nullptr checks does not work on any build using clang++.

This switches us to use a NEW_NOTHROW macro instead for new, and adds link checks to ensure we don't miss any cases

See also https://github.com/ArduPilot/ardupilot/pull/22614 and #22615 

Note that we could avoid the NEW_NOTHROW macro and use new(std::nothrow) directly, it is largely a matter of taste.
One advantage of directly using new(std::nothrow) is it may display better in C++ editors

UPDATE: this now maps NEW_NOTHROW to new on ChibiOS, and maps it to new(std::nothrow) on other HALs. This avoids the extra flash on ChibiOS where -fcheck-new works

### DIFF
--- a/.github/workflows/test_chibios.yml
+++ b/.github/workflows/test_chibios.yml
@@ -159,6 +159,7 @@ jobs:
             CubeOrange-PPP,
             SOHW,
             Pixhawk6X-PPPGW,
+            new-check,
         ]
         toolchain: [
             chibios,
--- a/AntennaTracker/GCS_Tracker.h
+++ b/AntennaTracker/GCS_Tracker.h
@@ -28,7 +28,7 @@ protected:
 
     GCS_MAVLINK_Tracker *new_gcs_mavlink_backend(GCS_MAVLINK_Parameters &params,
                                                  AP_HAL::UARTDriver &uart) override {
-        return new GCS_MAVLINK_Tracker(params, uart);
+        return NEW_NOTHROW GCS_MAVLINK_Tracker(params, uart);
     }
 
 private:
--- a/ArduCopter/GCS_Copter.h
+++ b/ArduCopter/GCS_Copter.h
@@ -43,7 +43,7 @@ protected:
 
     GCS_MAVLINK_Copter *new_gcs_mavlink_backend(GCS_MAVLINK_Parameters &params,
                                                 AP_HAL::UARTDriver &uart) override {
-        return new GCS_MAVLINK_Copter(params, uart);
+        return NEW_NOTHROW GCS_MAVLINK_Copter(params, uart);
     }
 
 };
--- a/ArduCopter/system.cpp
+++ b/ArduCopter/system.cpp
@@ -361,54 +361,54 @@ void Copter::allocate_motors(void)
         case AP_Motors::MOTOR_FRAME_DECA:
         case AP_Motors::MOTOR_FRAME_SCRIPTING_MATRIX:
         default:
-            motors = new AP_MotorsMatrix(copter.scheduler.get_loop_rate_hz());
+            motors = NEW_NOTHROW AP_MotorsMatrix(copter.scheduler.get_loop_rate_hz());
             motors_var_info = AP_MotorsMatrix::var_info;
             break;
         case AP_Motors::MOTOR_FRAME_TRI:
-            motors = new AP_MotorsTri(copter.scheduler.get_loop_rate_hz());
+            motors = NEW_NOTHROW AP_MotorsTri(copter.scheduler.get_loop_rate_hz());
             motors_var_info = AP_MotorsTri::var_info;
             AP_Param::set_frame_type_flags(AP_PARAM_FRAME_TRICOPTER);
             break;
         case AP_Motors::MOTOR_FRAME_SINGLE:
-            motors = new AP_MotorsSingle(copter.scheduler.get_loop_rate_hz());
+            motors = NEW_NOTHROW AP_MotorsSingle(copter.scheduler.get_loop_rate_hz());
             motors_var_info = AP_MotorsSingle::var_info;
             break;
         case AP_Motors::MOTOR_FRAME_COAX:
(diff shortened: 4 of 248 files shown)


amilcarlucas:
```
ardusub          1908 (+0.1210%)  0 (0.0000%)  -4 (-0.0015%)  1908 (+0.1208%)                                  384556
antennatracker   1404 (+0.1057%)  0 (0.0000%)  -4 (-0.0015%)  1404 (+0.1055%)                                  633752
ardurover        2012 (+0.1228%)  0 (0.0000%)  4 (+0.0015%)   2012 (+0.1226%)                                  322784
arduplane        1984 (+0.1113%)  0 (0.0000%)  0 (0.0000%)    1984 (+0.1111%)                                  178004
arducopter-heli  2032 (+0.1134%)  0 (0.0000%)  0 (0.0000%)    2032 (+0.1132%)                                  168884
blimp            1424 (+0.1053%)  0 (0.0000%)  0 (0.0000%)    1424 (+0.1052%)                                  610532
arducopter       2028 (+0.1136%)  0 (0.0000%)  -4 (-0.0015%)  2028 (+0.1134%)                                  175276
```

Is the size change expected?

tridge:
> Is the size change expected?

no, I'll look into it

tridge:
@amilcarlucas it seems the compiler does a better job optimising when we use new(std::nothrow). I don't know why. The result is the same regardless of whether we use -fcheck-new or not

tridge:
@amilcarlucas it is weird, the CI test shows the PR costing flash, but when I run the same builds locally (eg. for Durandal) I get a significant reduction in flash usage.
Here is master (hash a1971bb5b7) on Durandal:
```
Build directory: /data/APM/build/Durandal
----------------------------------------------------------------------------------------------------------
bin/arducopter   1790040      4008   258280               1794048          172020  Not Applicable
```
and here is this PR:
```
Build directory: /data/APM.36/build/Durandal
----------------------------------------------------------------------------------------------------------
bin/arducopter   1786836      4008   258284               1790844          175220  Not Applicable
```

tridge:
in CI, I see this for master: ```
BUILD SUMMARY
Build directory: /__w/ardupilot/ardupilot/base_branch/build/Durandal
Target               Text (B)  Data (B)  BSS (B)  Total Flash Used (B)  Free Flash (B)  External Flash Used (B)
---------------------------------------------------------------------------------------------------------------
bin/antennatracker    1328732      2180   260100               1330912          635152  Not Applicable         
bin/arducopter        1784756      4008   258284               1788764          177300  Not Applicable         
bin/arducopter-heli   1791144      4008   258280               1795152          170916  Not Applicable         
bin/arduplane         1782156      3920   258356               1786076          179988  Not Applicable         
bin/ardusub           1576980      2628   259652               1579608          386460  Not Applicable         
bin/blimp             1351928      2196   260072               1354124          611956  Not Applicable         
bin/ardurover         1638640      2628   259648               1641268          324800  Not Applicable         
```
and this for the PR:
```
BUILD SUMMARY
Build directory: /__w/ardupilot/ardupilot/pr/build/Durandal
Target               Text (B)  Data (B)  BSS (B)  Total Flash Used (B)  Free Flash (B)  External Flash Used (B)
---------------------------------------------------------------------------------------------------------------
bin/antennatracker    1330136      2180   260096               1332316          633752  Not Applicable         
bin/arducopter        1786784      4008   258280               1790792          175276  Not Applicable         
bin/arducopter-heli   1793176      4008   258280               1797184          168884  Not Applicable         
bin/arduplane         1784140      3920   258356               1788060          178004  Not Applicable         
bin/ardusub           1578888      2628   259648               1581516          384556  Not Applicable         
bin/blimp             1353352      2196   260072               1355548          610532  Not Applicable         
bin/ardurover         1640652      2628   259652               1643280          322784  Not Applicable         
```

Looking at copter, the size for the PR matches my local build, but the size for master doesn't. 
ahh, it depends on what directory I build in. Starting in a new directory I match the CI test

**Comment by tpwrules** [comment]
According to the GCC docs it should be sufficient to mark the `new` operator implementation as `noexcept`?

**Comment by tpwrules** [comment]
Is it time to make our own named malloc/new functions if we are having to use a macro here? Placement new could be used to place in allocated storage; delete would need to be changed.

There are a few other problems which could come back to bite us I've observed:
* `malloc`'s contents are defined to be unspecified, not 0. The compiler could do funky things with reads to data returned from malloc. Allegedly clang does this.
* It's undefined to allocate with new and pass to free, or allocate with malloc and pass to delete. But the existing code does this.

**Comment by tpwrules** [comment]
Added at the top of `AP_GPS.cpp` and replaced the functions within appropriately:
```c++
#define CONSTRUCT(T, ...) \
    ({ T* t = (T*)malloc(sizeof(T)); if (t) { new (t) T(__VA_ARGS__); }; t; })

template<class T> void DESTRUCT(T* p) { if(p) { p->~T(); } free(p); }
```

Using just `CONSTRUCT` adds 0 bytes on CubeOrangePlus, using DESTRUCT adds 8 bytes. This compares to 40 additional bytes using `new (std::nothrow)`.

The syntax is even uglier but it reinforces that something is going on and lets us safely do stuff like zero the underlying memory.

tridge:
this has been updated to make NEW_NOTHROW map to new on ChibiOS by default. You can make it map to NEW_NOTHROW by configuring with --enable-new-checking, which allows CI to check for new violations

tridge:
@tpwrules I can see how CONSTRUCT would work, but I decided to go with just leaving NEW_NOTHROW as new on ChibiOS, relying on gcc support for -fcheck-new, and using the C++ standard of new(std::nothrow) elsewhere

peterbarker (approved):
LGTM

PR #27170, opened 2024-05-27
SITL: update to work on newer cygwin versions

tridge:
fixes a build error with current cygwin


peterbarker:
Does it still work on older Cygwin versions?

tridge:
> Does it still work on older Cygwin versions?

still builds certainly (tested in CI), I expect it still works fine

Issue #27147, opened 2024-05-23
Overspeed during mission with takeoff item


Georacer:
## Issue report

**Issue details**

In a mission starting with a takeoff waypoint, while the plane climbs during the initial climb, the throttle is wide open and the aircraft overspeeds.

The expected behaviour would be for the plane to maintain cruise airspeed even while climbing at TKOFF_THR_MAX.

A more detailed issue description and investigation can be found here:
[github_issue.pdf](https://github.com/ArduPilot/ardupilot/files/15422753/github_issue.pdf)

The problem was reported in real life by a user launching with a catapult; the issue happens in a normal takeoff as well.
In a log produced with 4.3.1 the problem didn't exist. I believe this release was before the major TECS changes brought in by https://github.com/ArduPilot/ardupilot/pull/22191.

**Version**
Current master.

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[X] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
Reported by user flying normal plane.
Reproduced in SITL with `plane-catapult` frame.

**Hardware type**
N/A

**Logs**
Attached is a SITL log in a normal plane, flying an Auto mission.
mission file: [mission.txt](https://github.com/ArduPilot/ardupilot/files/15422694/mission.txt)
log: [00000005.zip](https://github.com/ArduPilot/ardupilot/files/15422682/00000005.zip)

Georacer:
We have mostly addressed this with the addition of the `TKOFF_OPTIONS[0]` in 4.6. Users who want airspeed control during takeoff can use that.

PR #27174, opened 2024-05-27
TECS: Fix airspeed control during takeoff

Georacer:
This PR addresses #27147 and #27211.

It fixes a lot of bugs currently lodged in TECS during a takeoff, both in TAKEOFF and AUTO modes. The major one was that TECS was operating in reset mode during all of FlightStage::TAKEOFF.

It refactors the throttle limits system for Plane so that: `servo.cpp` is the primary place where throttle limits (upper and lower) are imposed. `mode_takeoff.cpp` will also impose throttle limits according to its functionality description. `AP_TECS.cpp` will take those limits into account, also for the no-airspeed case, which was missed previously.
This has the benefit that TECS doesn't need to incorporate the logic for the various throttle limits decisions.

Logging of the throttle limits that TECS uses is added, in place of duplicate pitch limit logging.

Finally, it introduces `TKOFF_MODE` and `TKOFF_THR_MIN`. By default `TKOFF_MODE` makes it so that all takeoff climbs are done in max throttle (`THR_MAX` or `TKOFF_THR_MAX`).
When `TKOFF_MODE=1` the throttle range past `TKOFF_LVL_ALT` can very between min and max.
ardupilot-Takeoff_Mode_0
ardupilot-Takeoff_Mode_1

`TKOFF_THR_MIN` is also taken into account in quadplane transitions.

A lot of tests have been added to expose the new functionality.

---
Old PR text:

The major change in TECS behaviour is that it will no longer force full throttle and a speed weight of 2 during the whole of a TAKEOFF waypoint. Instead, it will limit this behaviour for the part while the airspeed is below the setpoint.

The changes make it so the aircraft does not overspeed during the climb.

It depends on #27125 for easier testing of the changes, by invoking `./Tools/autotest/autotest.py build.Plane test.Plane.CatapultTakeoff` or equivalent.

The commits of this PR are purposely split for now, to facilitate comparison and testing of each change.
You can find a detailed explanation of the reasoning behind the changes in the attached .pdf.
[PR_explanation.pdf](https://github.com/ArduPilot/ardupilot/files/15458300/PR_explanation.pdf)


libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1335,7 +1336,7 @@ void AP_TECS::update_pitch_throttle(int32_t hgt_dem_cm,
     }
 
     if (flight_stage == AP_FixedWing::FlightStage::TAKEOFF || flight_stage == AP_FixedWing::FlightStage::ABORT_LANDING) {
-        if (!_flags.reached_speed_takeoff && _TAS_state >= _TAS_dem_adj) {
+        if (!_flags.reached_speed_takeoff && _TAS_state >= _TAS_dem && (_TAS_dem > 3.0f)) {
```
tridge, line 1339: i think we could make this > 0 not > 3, the > 3 we used during testing was paranoia

Tools/autotest/arduplane.py
```diff
@@ -5474,4 +5514,5 @@ def disabled_tests(self):
             "LandingDrift": "Flapping test. See https://github.com/ArduPilot/ardupilot/issues/20054",
             "InteractTest": "requires user interaction",
             "ClimbThrottleSaturation": "requires https://github.com/ArduPilot/ardupilot/pull/27106 to pass",
+            "CatapultTakeoff": "Known bug. See https://github.com/ArduPilot/ardupilot/issues/27147",
```
tridge, line 5517: if it is passing now we can remove this line

tridge (changes requested):
in our testing we put in an additional check that throttle doesn't go below cruise in takeoff, should we add that in?

libraries/AP_TECS/AP_TECS.cpp
Georacer: I did at first, but it didn't work.
In the catapult launch, it seems that TECS is initialized after the aircraft is in the air.
I'll give it another shot, just in case I misrememeber.

Tools/autotest/arduplane.py
Georacer: Ah, indeed. Thanks.

Georacer:
> in our testing we put in an additional check that throttle doesn't go below cruise in takeoff, should we add that in?
Do you mean for the whole climb, or just the first segment?
For the first segment I don't think it's necessary, the throttle is very high anyways.

For the rest of the climb, it will be equivalent to setting a lower TECS throttle limit.
In general, I'd rather not add more `ifs` than we strictly need. But it might become necessary if I can't fix the soaring case.

libraries/AP_TECS/AP_TECS.cpp
tridge: maybe use _TAS_dem > min airspeed then?

Georacer: That actually sounds much better!

tridge (approved):

libraries/AP_TECS/AP_TECS.h
```diff
@@ -419,6 +419,9 @@ class AP_TECS {
     // need to reset on next loop
     bool _need_reset;
 
+    // Checks if we reset at the beginning of takeoff.
+    bool _reset_after_takeoff{false};
```
priseborough, line 423: The variable name doesn't match the function. Suggest this is renamed something like _takeoff_entry_reset_cmd

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1168,22 +1163,28 @@ void AP_TECS::_initialise_states(int32_t ptchMinCO_cd, float hgt_afe)
...
+            _max_climb_scaler = 1.0f;
+            _max_sink_scaler = 1.0f;
+            _pitch_demand_lpf.reset(_ahrs.get_pitch());
+            _pitch_measured_lpf.reset(_ahrs.get_pitch());
+        }
 
-        _pitch_demand_lpf.reset(_ahrs.get_pitch());
-        _pitch_measured_lpf.reset(_ahrs.get_pitch());
+        if (!_reset_after_takeoff) {
+            _flags.reset          = true;
```
priseborough, line 1179: Note that _flags.reset is also used in the update_50hz function to reset the complementary filter states. Because  update_50hz is called at a different rate and from a different part of the plane code to update_pitch_throttle we need to check if _flags.reset  is cleared from within update_pitch_throttle._initialise_states before the next frame of update_50hz runs

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -838,10 +835,8 @@ void AP_TECS::_update_throttle_with_airspeed(void)
 float AP_TECS::_get_i_gain(void)
 {
     float i_gain = _integGain;
-    if (_flight_stage == AP_FixedWing::FlightStage::TAKEOFF || _flight_stage == AP_FixedWing::FlightStage::ABORT_LANDING) {
-        if (!is_zero(_integGain_takeoff)) {
-            i_gain = _integGain_takeoff;
-        }
+    if (_in_takeoff_first_stage() || _flight_stage == AP_FixedWing::FlightStage::ABORT_LANDING) {
+        i_gain = _integGain_takeoff;
```
priseborough, line 839: In effect his change means we are defaulting to an integral gain of 0 during takeoff whereas before it was 0.3. This is inconsistent with the parameter description  "When set to 0 then TECS_INTEG_GAIN is used. " and is also a change to the default tuning.

Most takeoff will spend little time with  _in_takeoff_first_stage true and the speed weighting is 2 so I would not expect it to have much effect for most applications. However there may be some low powered setups with large takeoff heights where the integrator will help the pitch angle to  converge to the value needed to maintain airspeed.

Please update the parameter documentation and change TECS_TKOFF_IGAIN to match the TECS_INTEG_GAIN value of 0.3 unless there was a problem with that value in which case we need to find a suitable default.

libraries/AP_TECS/AP_TECS.h
Georacer: I meant for this to be a flag to be queried and checked.
If it's OK with you, I'll rename it to `_flag_have_reset_after_takeoff`.
If not, I'll take your suggestion and call it `_takeoff_entry_reset_cmd`.

libraries/AP_TECS/AP_TECS.cpp
Georacer: So, if I understand correctly, your concern is that `_flags.reset` could be cleared by `update_pitch_throttle()` before `update_50hz()` has a chance to use it?

In principle, `update_50hz()` is called by the `Plane::update_speed_height()` task, scheduled at 50Hz, while `update_pitch_throttle()` is called by the `Plane::update_alt()` task, scheduled at 10Hz.
So unless the scheduler goes terribly wrong, the execution should go something like that:
1. `Plane::update_alt()` calls `update_pitch_throttle()` and sets the `_flags.reset` bit.
2. `Plane::update_speed_height()` runs between 4 to 5 times, each one calling `update_50hz()` with `_flags.reset == true`.
3. Then `Plane::update_alt()` will run once more, invoking `update_pitch_throttle()` and clearing setting `_flags.reset = false`.

Are you satisfied with this argument?

libraries/AP_TECS/AP_TECS.cpp
Georacer: Parameter description updated.

Hwurzburg:
Can't hurt,but I can test on a real plane this week when weather permits

Georacer:
> Can't hurt,but I can test on a real plane this week when weather permits

Unfortunately I have no real-world flight testing capabilities around here. I'd be much obliged if you did.

Georacer:
> Has this been tested with normal hand launched 1m foamies using defaults (which most users would use) both with and without an airspeed sensor?

@Hwurzburg I tried with hand-launching the Aeroscout with RealFlight and a 2x2 test matrix:
- Mode: AUTO, TAKEOFF
- ARSPD_USE: 0, 1

I used the parameter files found here: https://github.com/ArduPilot/SITL_Models/tree/master/RealFlight/Tridge/Planes/Aeroscout except I deleted all `TECS*` related lines, so that the defaults will remain.

I also used this short mission in the tests: 
[rf_mission.txt](https://github.com/ArduPilot/ardupilot/files/15488189/rf_mission.txt)

Here are the logs: 
[logs.zip](https://github.com/ArduPilot/ardupilot/files/15488737/logs.zip)
It flirts with underspeed a bit, as it's not tuned at all, but it didn't fall out of the sky.

libraries/AP_TECS/AP_TECS.h
tridge: @Georacer it also should not be initialised to false - the init just wastes flash space

Georacer:
Okay, I'll change when I'm back from vacation.
Do our build options guarantee initialization to the default value, which I
guess is false for boolean and zero for numerical?

tridge:
> Do our build options guarantee initialization to the default value

yes

Hwurzburg:
Aeroscout AutoTakeoffs
Master (Airspeed Sensor active)
Master
This PR with Airspeed Sensor
withASPD
This PR without Airspeed being used
withoutARSPD

A bit surprised that no full throttle till leveloff....I had recently fixed MODE TAKEOFF not to switch to TECS until alt reached (full throttle climb) where it has switched to TECS once level alt reached....but I dont see a real issue with this.....but will have to fix MODE TAKEOFF again

Georacer:
> A bit surprised that no full throttle is kept until LVL_ALT is achived....I had recently fixed MODE TAKEOFF not to switch to TECS until alt reached (full throttle climb) where it had beeb switching to TECS once level alt reached....but this seems to switch even before that is reached...

@Hwurzburg, the TAKEOFF flight mode is not within the scope of this PR, only AUTO is. If I negatively affected TAKEOFF, we can examine that separately.

> I think full throttle needs to be active until LVL_ALT is reached to get into navigation ASAP....and we will have to fix MODE TAKEOFF again if this merged to revert its behviour also...

The point of this PR is primarily to prevent the aircraft from overspeeding, by **intentionally** giving TECS back its throttle control during the climb.
As I showed above this is beneficial. Do have any objections with it?

libraries/AP_TECS/AP_TECS.h
Georacer: Renamed and removed initialization.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1168,22 +1163,28 @@ void AP_TECS::_initialise_states(int32_t ptchMinCO_cd, float hgt_afe)
...
+        if (_in_takeoff_first_stage()) {
+            _post_TO_hgt_offset   = _climb_rate_limit * _hgt_dem_tconst; // Replacement prevents oscillating hgt_rate_dem.
+            _TAS_dem_adj = _TAS_dem;
+            _max_climb_scaler = 1.0f;
+            _max_sink_scaler = 1.0f;
+            _pitch_demand_lpf.reset(_ahrs.get_pitch());
+            _pitch_measured_lpf.reset(_ahrs.get_pitch());
+        }
 
-        _pitch_demand_lpf.reset(_ahrs.get_pitch());
```
tridge, line 1180: if we never do a takeoff (eg. done via FBWA) then this would mean no lpf resets

tridge (changes requested):

tridge:
> The point of this PR is primarily to prevent the aircraft from overspeeding, by **intentionally** giving TECS back its throttle control during the climb.

we may need a TECS API called by mode_takeoff.cpp that forces for the next loop full throttle to be used, this would apply below TKOFF_LVL_ALT

Hwurzburg:
I must say that I am opposed to
1. not having the option of using TKOFF_THR_MAX for the entire climb with an airspeed sensor and I think this should be the default. Lest someone have a crash due to poor setup on their first auto takeoff, or want max climb all the time but not want it during navigation
2. not having TKOFF_THR_MAX **always** until at least TKOFF_LVL_ALT is reached

libraries/AP_TECS/AP_TECS.cpp
Georacer: @tridge this is already taken care of in L1174: https://github.com/ArduPilot/ardupilot/pull/27174/files#diff-30aee80148363a53a210bc83701e744855f97514d3ce986ee83b1c0dc8fdf569R1174

libraries/AP_TECS/AP_TECS.cpp
Georacer: @priseborough my explanation above is wrong, because `update_50hz()` always sets the value of `_flags.reset` right at its beginning.

Still, I think what you really meant was that since `update_50hz()` has its own, separate criteria for setting `_flags.reset` and that would affect its use in `update_pitch_throttle()`.
But this is also not the case, because the first thing `update_pitch_throttle() -> _initialize_states()` does is to set `_flags.reset = false`. From then on, it decides whether it needs to set to true for the rest of the loop.

Georacer:
> I must say that I am opposed to
> 
>     1. not having the option of using TKOFF_THR_MAX for the entire climb with an airspeed sensor and I think this should be the default. Lest someone have a crash due to poor setup on their first auto takeoff, or want max climb all the time but not want it during navigation
> 
>     2. not having TKOFF_THR_MAX **always** until at least TKOFF_LVL_ALT is reached

Hey @Hwurzburg !

So for today, my intention was to make it so in TAKEOFF mode,
1. the throttle would be kept at the maximum allowable while below TKOFF_LVL_ALT and then
2. past that, TECS would be allowed to fly as normal.
However, as the min pitch is locked at TKOFF_LVL_PITCH for the whole climb, this interferes with TECS' controls and results in underspeed.
I believe that is the old (read: current) behaviour as well. This is a graph produced (more or less) with the new test `TakeoffMinThrottle` on the existing master.
image

It's quite complex to fix this in TAKEOFF mode, as that mode interferes with the FlightStage quite a bit, making it hard for TECS to understand what's going on.
If you want, we can discuss this live.

In any case, I have now adjusted this PR it so the throttle will be kept at max until TKOFF_LVL_ALT, which is the old behaviour.
Up until TKOFF_ALT, TECS will regulate throttle as it sees fit. This holds for both TAKEOFF and AUTO modes.

A graph of how TAKEOFF mode works with this PR:
image

And a graph of how AUTO mode works with this PR, with TKOFF_LVL_ALT=30:
image

Are you satisfied with these changes?

Georacer:
> > The point of this PR is primarily to prevent the aircraft from overspeeding, by **intentionally** giving TECS back its throttle control during the climb.
> 
> we may need a TECS API called by mode_takeoff.cpp that forces for the next loop full throttle to be used, this would apply below TKOFF_LVL_ALT

@tridge I've avoided doing that for now, with the latest changes. I've chosen another way to let TECS know of TKOFF_LVL_ALT, see `ArduPlane.cpp`.
If that's too bold, let me know.

libraries/AP_TECS/AP_TECS.cpp
Georacer: There are dedicated resets for that case in the previous if-case, L1154.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1436,3 +1446,12 @@ void AP_TECS::update_pitch_throttle(int32_t hgt_dem_cm,
     }
 #endif
 }
+
+bool AP_TECS::_in_takeoff_first_stage(void)
+{
+    return (_flight_stage == AP_FixedWing::FlightStage::TAKEOFF) && (
+        !_flags.reached_speed_takeoff || (
+            _tkoff_max_thr_alt > 0 && _height < _tkoff_max_thr_alt
```
tridge, line 1454: i think this may be a problem if we were doing a takeoff from down in a valley, as _height will be home relative, so this will be true during the whole takeoff
this conditional also needs to be broken apart to be easier to read, and commented

ArduPlane/ArduPlane.cpp
```diff
@@ -594,11 +594,11 @@ void Plane::update_alt()
             tecs_target_alt_cm = MAX(tecs_target_alt_cm, prev_WP_loc.alt - home.alt) + (g2.rtl_climb_min+10)*100;
         }
 
+        TECS_controller.set_tkoff_settings(get_takeoff_pitch_min_cd(), mode_takeoff.level_alt);
```
tridge, line 597: i think it would be better to pass a min throttle, with an API that sets the min throttle in TECS for the next TECS update loop, auto-cleared. That would allow the decision to be made by the core plane code, with common code for mode takeoff and auto takeoff

tridge (changes requested):

libraries/AP_TECS/AP_TECS.cpp
Georacer: I have removed this additional height condition and refactored the method with more comments.

For heights below `TKOFF_LVL_ALT` only the max throttle is now overwritten via the new `TECS::set_throttle_min()` method.

ArduPlane/ArduPlane.cpp
Georacer: As requested, I have implemented `TECS::set_min_throttle()` and TAKEOFF mode uses it to command a max throttle climb during FlightStage::Takeoff.

Georacer:
I've made the following changes after today's Dev Call:
1. I've removed the takeoff-first-stage height check from within TECS. TAKEOFF mode will control it by itself...
2. via the new `TECS::set_min_throttle()` method.

---

As a recap, here are the specs I'm trying to hit:

1 In AUTO takeoff
    1.1 before reaching min airspeed, use max throttle.
    1.2 after reaching min airspeed, do airspeed control as normal.
2 In TAKEOFF mode
    2.1 before reaching min airspeed, use max throttle (fallback to 1.1)
    2.2 before reaching TKOFF_LVL_ALT, use max throttle.

Here are the comparison between `master` and this branch, using the two new autotests I've created.

Before, AUTO takeoff waypoint.
Note how the throttle is at max for the whole climb and the vehicle overspeeds.
image

After, AUTO takeoff waypoint.
TECS now keeps max throttle until minimum airspeed is reached, and then controls throttle as normal.
image

Before, TAKEOFF mode, with
- `TKOFF_LVL_ALT=30`
- `TKOFF_THR_MAX=90`
TECS switches to `FlightStage:Normal` as soon as `TKOFF_LVL_ALT` is hit.
image

After, TAKEOFF mode. The functional behaviour is unchanged.
image

Hwurzburg:
how about the same plots without an airspeed sensor?

Georacer:
> how about the same plots without an airspeed sensor?

Right, thanks for the reminder.

AUTO, with the following **same** additional parameters as the _with airspeed_ test, but also with `ARSPD_USE=0`
"ARSPD_USE": 0.0,
"TKOFF_LVL_ALT": 30.0,
"TKOFF_THR_MINACC": 3.0,
"TECS_PITCH_MAX": 35.0,
"PTCH_LIM_MAX_DEG": 35.0

Before:
image

After:
image

TAKEOFF, same as the previous _with airspeed_-setup, but again, adding `ARSPD_USE=0`.
Before:
image

After:
image

I'll spin-up Windows to do a TAKEOFF hand-launch on RealFlight, for good measure.

libraries/AP_TECS/AP_TECS.h
```diff
@@ -134,6 +134,12 @@ class AP_TECS {
         _pitch_max_limit = pitch_limit;
     }
 
+    // set minimum throttle override, 0-1 range
+    // it is applicable for one control cycle only
+    void set_min_throttle(const float thr) {
```
tridge, line 139: maybe set_throttle_range(), and have min and max, so we can do both min and max throttle in one API
note that min can be -ve if we are allowing for reverse throttle, so need to be careful.
we could pass a NaN for the min to mean no lower limit? or maybe -1 meaning no lower limit?

tridge (changes requested):

Hwurzburg (changes requested):
George, here are my concerns with your stated goals :
1. Many have near hgh takeoff obstacles that require tkoff_thr_max for a longer period  than to min_airspeed to get max climb...must have an  option to do this...either option current behavior or via tkoff_lvl_alt in ALL autotakeoffs and have max thrfotlle until that alt in any case
2. not a coder, and not sure this happens, but if aiirpseed sensor is not used, you cannot use ANY airspeed from AP for decisions during an entire autottakeoff
3. AUTO NAV_TAKEOFF  and MODE TAKEOFF climbs to their target alt should behave the same in terms of climb profile (MODE TAKEEOFF can start loitering during the climb where NAV_TAKEOFF does not, but climb profile should be thee same as determined by TKOF_LVL_ALT, and takeoff pitches programmed, and Takeof thr max

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -889,18 +891,6 @@ void AP_TECS::_update_throttle_without_airspeed(int16_t throttle_nudge, float pi
...
-        if (_takeoff_start_ms == 0) {
-            _takeoff_start_ms = now;
-        }
-        const uint32_t dt = now - _takeoff_start_ms;
-        if (dt*0.001 < aparm.takeoff_throttle_max_t) {
-            _throttle_dem = _THRmaxf;
-        }
-    } else {
-        _takeoff_start_ms = 0;
-    }
```
Georacer, line 903: This has been made common to with- and without-airspeed cases.

ArduPlane/Parameters.cpp
```diff
@@ -142,12 +142,28 @@ const AP_Param::Info Plane::var_info[] = {
 
     // @Param: TKOFF_THR_MAX_T
     // @DisplayName: Takeoff throttle maximum time
-    // @Description: This sets the time that maximum throttle will be forced during a fixed wing takeoff without an airspeed sensor. If an airspeed sensor is being used then the throttle is set to maximum until the takeoff airspeed is reached.
+    // @Description: This sets the time that maximum throttle will be forced during a fixed wing takeoff.
```
Georacer, line 145: This has been made common to both with- and without-airspeed cases. I believe it makes it easier to tell the various takeoff combinations apart, without negatively affecting the with-airspeed takeoffs.
The general idea is that they should be max-throttle by default anyways.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1108,8 +1100,8 @@ void AP_TECS::_update_pitch(void)
                                     (double)integSEBdot_max,
                                     (double)_integSEBdot,
                                     (double)_integKE,
-                                    (double)_PITCHminf,
-                                    (double)_PITCHmaxf);
+                                    (double)_THRminf,
+                                    (double)_THRmaxf);
```
Georacer, line 1104: Pitch limit logging also occurs in the `TECS` topic.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1308,9 +1295,6 @@ void AP_TECS::update_pitch_throttle(int32_t hgt_dem_cm,
             // note that this allows a flare pitch outside the normal TECS auto limits
             _PITCHmaxf = _land_pitch_max;
         }
-
-        // and allow zero throttle
-        _THRminf = 0;
```
Georacer, line 1313: This has been migrated to `_update_throttle_limits()`.

Georacer:
I've updated the original PR text, since there has been a significant rework in the last two weeks.
Please read that first.

To address the review comments that were pending:
**1. Are TAKEOFF and AUTO modes' behaviour identical?**
It wasn't 100% before, should be identical now. The only caveat is that mode AUTO doesn't have direct access to `TKOFF_LVL_ALT` and hence can't do full throttle below it.

**2. Taking into account takeoffs that happen mid-mission, from a deep valley.**
No change in the AUTO or TAKEOFF altitude logic has been made, so it shouldn't make things worse. Any TECS changes only affect throttle logic, not altitude thresholds and targets.

**3. How about reverse-throttle setups?**
No new throttle limits have been added, most notably in `servos.cpp` except for FlightStage::TAKEOFF, where reverse throttle isn't used.

**4. What about when ARSPD_USE=0?**
It is deemed that in lack of airspeed measurement, we can't rely on synthetic airspeed during takeoff. So max throttle will be used throughout the climb.

**5. Do we need to increase TKOFF_LVL_ALT?**
No, since the default behaviour for most users who have TKOFF_MODE=0 will be max throttle for the whole climb. Individual users can tailor to their own needs.

New behaviour:
1. The new parameters TKOFF_MODE and TKOFF_THR_MIN are introduced.

2. In the current master with a TAKEOFF mode, the throttle was allowed to drop when climbing past LVL_ALT.
Now this is only allowed when ARSPD_USE=1 and TKOFF_MODE=1.

3. In the current master with an AUTO takeoff and ARSPD_USE=0, the throttle was allowed to recede past TKOFF_THR_MAX_T.
Now it is always at max.

4. In the current master TKOFF_THR_MAX_T applies only when ARSPD_USE=0.
Now it applies always. This is to simplify the TKOFF* parameters.

5. Increased TKOFF_LVL_ALT from 5m to 10m.
At the request of Henry.

Hwurzburg:
sorry wont be able test fly...did a quick sim with no airspeed sensor and defaults...master drops throttle after tkoff_lvl_alt is reached (5m)...but this keeps it maxed until over 2x that alt....why?

Georacer:
> sorry wont be able test fly...did a quick sim with no airspeed sensor and defaults...master drops throttle after tkoff_lvl_alt is reached (5m)...but this keeps it maxed until over 2x that alt....why?

This is intentional (as written above), in order to guarantee that the plane can climb without stalling up until TKOFF_LVL_ALT while fixed at TKOFF_LVL_PITCH or higher.
Master doesn't do a good job maintaining actual airspeed in this case.

If one installs an airspeed sensor and is confident about their tuning, they can set TKOFF_MODE=1 and it will fly with normal TECS past TKOFF_LVL_ALT.

Hwurzburg:
when does thr drop without aspd sensore if not after TKOFF_LVL_ALT?

Georacer:
> when does thr drop without aspd sensore if not after TKOFF_LVL_ALT?
> 
> edit: just read your comment....will recheck my sim..yep its okay my only concern is that AUTO max perf climb for some period cannot be obtained...there should be some way to access the TKOF_LVL_ALT param from mode AUTO

I'm not sure I understand your concern. Is it that while doing an AUTO takeoff the thruster shouldn't be wide open for the whole climb?
If ARSPD_MODE=0, then one can set a lower altitude in the takeoff waypoint.
If ARSPD_MODE=1, then TKOFF_MODE=1 can be set, where there will be normal throttle control after TKOFF_THR_MAX_T.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1436,3 +1420,82 @@ void AP_TECS::update_pitch_throttle(int32_t hgt_dem_cm,
...
+    // Raise min to force max throttle for TKOFF_THR_MAX_T after a takeoff.
+    if (_flight_stage == AP_FixedWing::FlightStage::TAKEOFF) {
+        const uint32_t now = AP_HAL::millis();
+        if (_takeoff_start_ms == 0) {
+            _takeoff_start_ms = now;
+        }
+        const uint32_t dt = now - _takeoff_start_ms;
+        if (dt*0.001 < aparm.takeoff_throttle_max_t) {
+            _THRminf = _THRmaxf;
+        } else {
```
Hwurzburg, line 1477: what is this????

Georacer: This particular if-statement is the implementation of TKOFF_THR_MAX_T: It ensures throttle is at max for the first TKOFF_THR_MAX_T seconds.

I didn't create it; I moved it from `_update_throttle_without_airspeed` further above to make it common with both with- and without-airspeed measurement cases. Thus simplifying the default takeoff logic.

Hwurzburg: Geroge, the i variable is unused

Hwurzburg (approved):

ArduPlane/Parameters.cpp
```diff
@@ -142,12 +142,28 @@ const AP_Param::Info Plane::var_info[] = {
...
+    // @DisplayName: Takeoff minimum throttle
+    // @Description: The minimum throttle to use in takeoffs in AUTO and TAKEOFF flight modes, when TKOFF_MODE=1. Also, the minimum throttle to use in a quadpane forward transition. This can be useful to ensure faster takeoffs or transitions on aircraft where the normal throttle control leads to a slow takeoff or transition. It is used when it is larger than THR_MIN, otherwise THR_MIN is used instead.
+    // @Units: %
+    // @Range: 0 100
+    // @Increment: 1
+    // @User: Standard
+    ASCALAR(takeoff_throttle_min,       "TKOFF_THR_MIN",    60),
+
+    // @Param: TKOFF_MODE
+    // @DisplayName: Takeoff mode
```
tridge, line 162: add TKOFF_OPTIONS, with first bit being this option

tridge (changes requested):
with the change to use TKOFF_OPTIONS I'm happy for this to go in
many thanks to @Hwurzburg for the careful testing!!

libraries/AP_TECS/AP_TECS.cpp
priseborough: Should be OK

priseborough (approved):

tridge (approved):

PR #27177, opened 2024-05-27
AP_Scripting: added tracker Pelco-D control script

julled:
This script uses the scaled output from the antennatracker servos and map them to corresponding Pelco-D messages to be sent via a RS-485 interface to a motorized base (can be anything from motorized tracker to a PTZ camera). If your FCU doesnt offer a RS-485 interface by default, you can use or TTL-RS485- or USB-RS485-adapters. I tested this script using a USB-RS485 adapter using Linux/Obal board and a Hikvision PTZ camera.

### DIFF
--- a/libraries/AP_Scripting/applets/pelco_d_antennatracker.lua
+++ b/libraries/AP_Scripting/applets/pelco_d_antennatracker.lua
@@ -1,0 +1,110 @@
+--[[
+   Pelco-D control implementation for antennatracker.
+   Implemented by using knowledge from the excellent python implementation in https://gist.github.com/jn0/cc5c78f4a0f447a6fb2e45a5d9efa13d.
+--]]
+
+local SERVO_PAN = 71
+local SERVO_TILT = 72
+local SERIAL_BAUD = 9600
+
+-- Antennattracker modes
+local MODE_STOP = 1
+local MODE_SCAN = 2
+local MODE_SERVOTEST = 3
+local MODE_GUIDED = 4
+
+gcs:send_text(0, "Starting Pelco-D Control")
+
+local port = assert(serial:find_serial(0), "Pelco-D: No Scripting Serial Port")
+port:begin(SERIAL_BAUD)
+port:set_flow_control(0)
+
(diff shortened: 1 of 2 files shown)


libraries/AP_Scripting/applets/pelco_d_antennatracker.lua
```diff
@@ -0,0 +1,114 @@
...
+local baud = 9600
+
+-- Antennattracker modes
+local MODE_STOP = 1
+local MODE_SCAN = 2
+local MODE_SERVOTEST = 3
+local MODE_GUIDED = 4
+
+gcs:send_text(0, "Starting Pelco-D Control")
+local port = serial:find_serial(0)
```
IamPete1, line 17:
```suggestion
local port = assert(serial:find_serial(0), "Pelco-D: No Scripting Serial Port")
```
To fix the lua CI warning. The assert checks that `port` is not `nil`. Before you call begin on it.

You do have the check on line 21, so you could also move the begin and flow control calls after that. 

With the assert you can remove the check on 21.

libraries/AP_Scripting/applets/pelco_d_antennatracker.lua
```diff
@@ -0,0 +1,112 @@
+--[[
+   Pelco-D control implementation for antennatracker.
+   Implemented by using knowledge from the excellent python implementation in https://gist.github.com/jn0/cc5c78f4a0f447a6fb2e45a5d9efa13d.
+--]]
+
+---@diagnostic disable: need-check-nil
```
IamPete1, line 6: This is cheating. The scripting checks are still quite new, so we have a bunch of these in the code base for pre-exiting scripts. I would like to not add any more. 

What is the error it gives?

IamPete1: The assert should have fixed that?

IamPete1: Looks like were good.

PR #27178, opened 2024-05-27
AP_Scripting: added GPS yaw binding

tridge:
useful for seeing if switching to a GPS yaw source set in EK3 will work


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2819,6 +2819,13 @@ gps.GPS_OK_FIX_2D = enum_integer
 gps.NO_FIX = enum_integer
 gps.NO_GPS = enum_integer
 
+-- get yaw from GPS in degrees
+---@param instance -- instance number
```
peterbarker, line 2823:
```suggestion
---@param instance integer -- instance number
```

PR #27179, opened 2024-05-27
autotest.py: remove unused frame parameter

peterbarker:
this is unused

... should we make it used?  Why?  The tests aren't written for different frame dynamics.


andyp1per (approved):

Georacer:
I'm a bit late to the party, but I think it is actually used, at least syntactically.

It seems to be passed to https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/vehicle_test_suite.py#L1680, via https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/autotest.py#L524

Not sure why this PR didn't trigger an issue, though. Perhaps the `frame` attribute is always overwritten by our workflow?

PR #27184, opened 2024-05-28
GCS_MAVLink: Remove any unnecessary processing

muramura:
I discovered that I was performing the same process multiple times. 
I will remove the unnecessary processing.


khancyr (commented):
Logic is the same, probably just codestyle. 
we can use > 1 as start_stop is uint type so no real use of checking negative.

peterbarker:
Makes the code larger; no *real* benefit here.

Sorry, closing this one.

PR #27192, opened 2024-05-29
change iomcu_f103_8MHz to iomcu-f103-8MHz

burgeruser:
change iomcu_f103_8MHz to iomcu-f103-8MHz


burgeruser:
@rmackay9 
I've also revised the title for this.

rmackay9:
Hi @burgeruser,

So, to be clear, I mean the commit's title.  Not the PR title.  So I think you've updated the PR's title, that's temporary and we're not too concerned about that.  The commit title remains forever though and we like to prefix it with the subsystem because it allows us to more easily do backports.

amilcarlucas (approved):

burgeruser:
> Hi @burgeruser,
> 
> So, to be clear, I mean the commit's title. Not the PR title. So I think you've updated the PR's title, that's temporary and we're not too concerned about that. The commit title remains forever though and we like to prefix it with the subsystem because it allows us to more easily do backports.

@rmackay9 
Apologies, I didn't submit promptly; I just reapplied the commit and modified the commit title.

PR #27195, opened 2024-05-29
AP_NavEKF3: ensure EK3 does not read GPS when disabled

tridge:
this allows for a change of source set to disable use of GPS. Otherwise the GPS will still be used even when the source set had no GPS enabled
This came up while testing dead-reckoning, where the vehicle was flying on EKF2 and we wanted EKF2 to retain GPS. We used a lua script to change EKF3 source set and expected that to stop the use of GPS for EKF3

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
@@ -547,6 +547,11 @@ void NavEKF3_core::readGpsData()
     // check for new GPS data
     const auto &gps = dal.gps();
 
+    if (frontend->sources.getPosXYSource() != AP_NavEKF_Source::SourceXY::GPS) {
+        // don't read GPS data when disabled in the current source set
+        return;
+    }
+
     // limit update rate to avoid overflowing the FIFO buffer
     if (gps.last_message_time_ms(selected_gps) - lastTimeGpsReceived_ms <= frontend->sensorIntervalMin_ms) {
         return;
--- a/libraries/AP_NavEKF3/AP_NavEKF3_Outputs.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_Outputs.cpp
@@ -234,7 +234,8 @@ bool NavEKF3_core::getPosNE(Vector2f &posNE) const
         // In constant position mode the EKF position states are at the origin, so we cannot use them as a position estimate
         if(validOrigin) {
             auto &gps = dal.gps();
-            if ((gps.status(selected_gps) >= AP_DAL_GPS::GPS_OK_FIX_2D)) {
+            if (frontend->sources.getPosXYSource() == AP_NavEKF_Source::SourceXY::GPS
+                && gps.status(selected_gps) >= AP_DAL_GPS::GPS_OK_FIX_2D) {
                 // If the origin has been set and we have GPS, then return the GPS position relative to the origin
                 const Location &gpsloc = gps.location(selected_gps);
                 posNE = public_origin.get_distance_NE_ftype(gpsloc).tofloat();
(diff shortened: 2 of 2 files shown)


libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
```diff
@@ -542,6 +542,11 @@ void NavEKF3_core::readGpsData()
     // check for new GPS data
     const auto &gps = dal.gps();
 
+    if (frontend->sources.getPosXYSource() != AP_NavEKF_Source::SourceXY::GPS) {
```
rmackay9, line 545: Although it's unusual a user could enable only altitude or only velocity (xy or z) while not using xy position so I think we need a more fine grained method to disable it.  Also remember that even if the active source isn't GPS, it is still possible if EK3_SRC_OPTIONS = 1 (Fuse all velocities) that GPS velocities from another source set could be used.

libraries/AP_NavEKF3/AP_NavEKF3_Outputs.cpp
```diff
@@ -230,7 +230,8 @@ bool NavEKF3_core::getPosNE(Vector2f &posNE) const
         // In constant position mode the EKF position states are at the origin, so we cannot use them as a position estimate
         if(validOrigin) {
             auto &gps = dal.gps();
-            if ((gps.status(selected_gps) >= AP_DAL_GPS::GPS_OK_FIX_2D)) {
+            if (frontend->sources.getPosXYSource() == AP_NavEKF_Source::SourceXY::GPS
```
rmackay9, line 233: One concern is that the position may jump as the user changes between source sets.  So a source set that includes GPS will show the latest GPS position but if the user then switches back to external nav, it'll jump back to the last good position estimate.

I actually agree with this change or something like it.  I've always thought it odd that the EKF3 sometimes reports the raw GPS position instead of the EKF position.  We already report the raw GPS position to the GCS so I'm not sure we really need to report it again through the EKF.  That's probably too big a change for the GCSs to absorb though.

peterbarker: > I actually agree with this change or something like it. I've always thought it odd that the EKF3 sometimes reports the raw GPS position instead of the EKF position. We already report the raw GPS position to the GCS so I'm not sure we really need to report it again through the EKF. That's probably too big a change for the GCSs to absorb though.

Resolving that problem is complicated by the fact that the EKF can return a position estimate even when it is feeling ill.  It will do this if it has ever managed to get an origin (i.e. be happy with where it is), and when the GPS is not capable of providing a location.

What that means is that never returning a GPS from the EKF would mean changing the interface we use onto the backends to be more complicated.   e.g. "EK3 give me a location".  What, you can't do that?  OK, "GPS, what about you?"  Really?!  OK, "EK3, please give me your best guess".  I don't think it will be all that pretty.

peterbarker: ... you can't really use the data if getLLH returns false; if it has never seen an origin and the GPS isn't giving a location then `false` really means that it hasn't given you anything at all.  This is the sort of case you might end up flying to 0,0

moreba1:
> this allows for a change of source set to disable use of GPS. Otherwise the GPS will still be used even when the source set had no GPS enabled This came up while testing dead-reckoning, where the vehicle was flying on EKF2 and we wanted EKF2 to retain GPS. We used a lua script to change EKF3 source set and expected that to stop the use of GPS for EKF3

Currently EKF3 use GPS for update origin even if EK3_SRC_POSZ is baro .  I will use gps for horizontal position and use baro for altitude . EKF origin must be not changed after arming and during flight. 

related to this issue : https://github.com/ArduPilot/ardupilot/issues/33431

PR #27200, opened 2024-05-30
AP_Scripting: do not send fragmented messages

muramura:
The text status is limited to 50 bytes. 
If it exceeds 50 bytes, it is fragmented and sent. 
The fragmented strings become garbage. 
When I send, I truncate the text at this 50-byte limit. 
With this process, I will no longer send text that has become garbage.

MISSION PLANNER
AFETR
Screenshot from 2024-05-30 22-24-14

BEFORE
Screenshot from 2024-05-30 22-30-59

### DIFF
--- a/libraries/AP_Scripting/AP_Scripting.cpp
+++ b/libraries/AP_Scripting/AP_Scripting.cpp
@@ -380,6 +380,7 @@ bool AP_Scripting::arming_checks(size_t buflen, char *buffer) const
     lua_scripts::get_last_error_semaphore()->take_blocking();
     const char *error_buf = lua_scripts::get_last_error_message();
     if (error_buf != nullptr) {
+        buflen = buflen < 44 ? buflen : 43;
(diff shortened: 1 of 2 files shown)


IamPete1:
I don't think its better, were loosing information. Mission planner should be able to re-assemble the fragments, I'm not sure why its doesn't.

muramura:
@IamPete1 san.
I understand that MAVLINK2 includes definitions for fragmentation. This phenomenon means that the process is not within MAVLINK2. I will check the ArduPilot processes.

Thank you for the comment.

peterbarker:
@IamPete1 has the right of it here.

Both MAVProxy and QGroundControl reassemble these statustexts from the chunks.

Patches are required against MissionPlanner here, not ArduPilot.

Closing, but thanks anyway.  And I'll review patches to add the feature to MissionPlanner (badly).

Issue #27171, opened 2024-05-27
iomcu reset with bdshot on 6X


tridge:
With a CUAV 6X with Pixhawk6X-bdshot firmware, with 6 blheli32 ESCs connected on a hexacopter (4 ESCs on main, 2 on aux) I get a reproducible internal error 0x1000 for IOMCU reset after arming. The copter still flies fine, but we need to find the cause of this.
Logs here:
https://drive.google.com/drive/folders/1fveumuzs7zw7zCqw7jKLHqiaXxx3k5nC?usp=drive_link
ping @andyp1per

burgeruser:
I encountered a similar issue earlier, and my solution was either to exclusively use D-shot outputs from the IOMCU or have the MAINMCU handle all D-shot outputs. I frequently ran into error code 0x1000 back then.

PR #27201, opened 2024-05-30
Fix IOMCU transmission errors resulting from DMA timer interference

andyp1per:
This fixes https://github.com/ArduPilot/ardupilot/issues/27171

We basically needed to reset the PWM peripheral when releasing the IC DMA lock in the same way we do for the main DMA channel.


tridge:
@andyp1per what is the consequences of this bug? I assume it affects 4.5 as well?

andyp1per:
> @andyp1per what is the consequences of this bug? I assume it affects 4.5 as well?

It affects 4.5 as well. Only happens with bdshot on channels 3/4. Consequences are dropped packets on the IOMCU uart because you get missing bytes at the beginning of frames. Its also somewhat timing dependent, if you build with assertions the problem goes away, if you put in a little sleep the problem goes away, it doesn't happen every frame - only now and then - hence why you were able to fly without issue.

tridge (approved):

tridge:
I've confirmed this bug impacts 4.5.3

PR #27203, opened 2024-05-31
autotest: correct Valgrind error detection 

peterbarker:
Somehow we lost this aspect of autotest where a non-zero valgrind log means failiure.

Fix that.

I've tested the failure case here and CI hsould cover the no-failure case.


rmackay9:

PR #27211, opened 2024-06-01
Plane: added TKOFF_THR_MIN

tridge:
this allows for a minimum throttle to be applied during a takeoff or quadplane forward transition in auto throttle modes. This is useful on high drag aircraft where normal TECS throttle control results in a slow transition

### DIFF
--- a/ArduPlane/servos.cpp
+++ b/ArduPlane/servos.cpp
@@ -521,6 +521,9 @@ float Plane::apply_throttle_limits(float throttle_in)
 
     // Query the conditions where TKOFF_THR_MAX applies.
     const bool use_takeoff_throttle =
+#if HAL_QUADPLANE_ENABLED
+        quadplane.in_slt_fwd_transition() ||
+#endif
         (flight_stage == AP_FixedWing::FlightStage::TAKEOFF) ||
         (flight_stage == AP_FixedWing::FlightStage::ABORT_LANDING);
 
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -3973,6 +3973,21 @@ bool QuadPlane::in_transition(void) const
 }
 
 /*
+  return true if we are in a fwd transition for a SLT quadplane
+  and in an auto-throttle mode
+ */
+bool QuadPlane::in_slt_fwd_transition(void) const
+{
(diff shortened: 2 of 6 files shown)


ArduPlane/servos.cpp
```diff
@@ -540,7 +540,21 @@ float Plane::apply_throttle_limits(float throttle_in)
...
+    auto throttle = constrain_float(throttle_in, min_throttle, max_throttle);
+
+    /*
+      apply TKOFF_THR_MIN if enabled and in an auto-throttle forward
+      transition
+     */
+    if (use_takeoff_throttle_max) {
+        const auto tmin_thr = g2.takeoff_throttle_min;
+        if (tmin_thr > throttle) {
+            throttle = MAX(throttle, tmin_thr);
```
IamPete1, line 552: This means I could set my take off min throttle larger than `THR_MAX` (or `TKOFF_THR_MAX`). I think we should still constrain down in that case.

ArduPlane/servos.cpp
```diff
@@ -540,7 +540,21 @@ float Plane::apply_throttle_limits(float throttle_in)
...
 #endif
 
-    return constrain_float(throttle_in, min_throttle, max_throttle);
+    auto throttle = constrain_float(throttle_in, min_throttle, max_throttle);
+
+    /*
+      apply TKOFF_THR_MIN if enabled and in an auto-throttle forward
+      transition
+     */
+    if (use_takeoff_throttle_max) {
```
IamPete1, line 549: Why not apply in the same case as `takeoff_throttle_max`. Presumably because you want to do the TECS reset. But if that is important we should probably be doing the TECS reset for the existing ICE idle, max take off throttle, battery comp and watt limit cases too....

IamPete1: also, maybe the bool should be `use_takeoff_throttle_limits` now.

IamPete1 (commented):
We could pass the actual throttle limits to TECS so it knows what is going on. I guess the reset means that it will actually chop throttle to zero and have under speed before the I term builds up.

tridge:
@Georacer updated to handle SLT transition

peterbarker:
https://github.com/ArduPilot/ardupilot/pull/27174 added this parameter and https://github.com/ArduPilot/ardupilot/pull/27770 fixed its usage.

Closing this one now.

PR #27214, opened 2024-06-01
SITL: delete unnecessary settings

muramura:
Since the elements are being cleared, the subsequent zero settings are unnecessary.


khancyr:
Thanks anyway. I will close this

PR #27215, opened 2024-06-01
SITL: correct valgrind failure in SIM_GPS_SBP2

muramura:
Make it the same process as commit hash b2b8eb93 .


peterbarker:
Thanks @muramura , good catch.

PR #27219, opened 2024-06-02
AP_Scripting: introduce serial device simulation support

tpwrules:
Allows a script to simulate a device attached via any serial protocol. The script can read and write data and have it handled according to the protocol as if exchanged over a serial port. The script can then do protocol translation, data filtering and validation, hardware-in-the-loop simulation, experimentation, etc., especially in combination with the scripting protocol which lets the script itself handle an attached device and so interpose any communication.

Costs about 850B flash on Cube Orange Plus.

Todo:
* [X] Test in SITL
* [X] Test that it properly disables
* [x] Test on real vehicle
* [x] Bikeshed names
* [x] Put parameters into a subgroup?
* [x] Examples
* [x] Binding docs
* [ ] Wiki docs
* [x] Autotest script


tridge:
@tpwrules interesting!
I'd like to see a working example, maybe a script that presents a NMEA GPS, mirroring real GPS data?

tpwrules:
That should be possible, I'll give it a try.

tpwrules:
I've added such an example, it works well in SITL. The extension to having a switch that starts fiddling with the data is obvious, but I'm not sure of a realistic way to do that.

tpwrules:
Some more system questions:
* Are protocols going to get confused that the baud rate is 0? I wasn't sure how to update the parameter, I can't just assign to it like a normal variable and exposing it as a real parameter doesn't really make sense IMO.
* Are protocols going to get confused if the script doesn't drain the buffer and it fills up?

tpwrules:
I tested the example script on a real vehicle and it does work nice. There is a small amount of driftiness/toiletbowling possibly caused by the lag of the script processing the data (not helped by running on an F4 autopilot) or the unavailability of the 3D velocity information and ground course uncertainty, and the not particularly great mags on the tested drone.

But it worked fantastic as a protocol demo.

IamPete1 (commented):
With this we now have two serial port userdata objects in lua. I suspect we could generalize so we only need one. I think you would only have to remove `get_protocol_index`. Maybe the `get_device_port` could become `find_virtual_serial(portoocl)` or something so you don't need the driver level protocol anymore.

tpwrules:
I see your point, but the problem is that for the device the script needs to read from the write buffer and write to the read buffer. There aren't any methods on `AP_HAL::UARTDriver` to accomplish this, and we need the subclass of `AP_SerialManager::RegisteredPort` anyway, so binding to that makes sense.

I suppose we could introduce a scripting-specific object that replaces both serial userdatas with some munging to access `AP_HAL::UARTDriver` or `AP_SerialManager::RegisteredPort` depending on the direction but it's not immediately clear that this would save anything. Is an additional userdata really that expensive?

IamPete1:
`AP_SerialManager::RegisteredPort` is already a subclass of `AP_HAL::UARTDriver` I think it should "just work" so long as we don't need any of the new methods not in the base class.

tpwrules:
Consider the loopback example, both userdatas from `find_port` and `get_device_port` are bindings of the exact same `AP_HAL::UARTDriver` instance, but the former is accessing methods on the superclass and the latter methods on the subclass.

So we would need some sort of multiplex class that the script binds to that decides which to do. I don't think that's impossible, but it doesn't seem like it would save anything. I will investigate though.

I do like the idea of finding by protocol, I will try implementing that as well.

tpwrules:
I was able to implement your suggestions and get good savings. It saves about 300 bytes, but the base branch has changed somehow and the total cost is now only ~850 bytes. Thanks for the input.

tpwrules:
Tested again on a real drone and everything still works nice.

libraries/AP_Scripting/AP_Scripting_SerialDevice.cpp
```diff
@@ -0,0 +1,156 @@
...
+ */
+void AP_Scripting_SerialDevice::Port::init(void)
+{
+    begin(1000000, 0, 0); // assume 1MBaud rate even though it's a bit meaningless
+}
+
+void AP_Scripting_SerialDevice::Port::clear(void)
+{
+    WITH_SEMAPHORE(sem);
+    if (readbuffer) readbuffer->clear();
```
rmackay9, line 56: nitpick: we normally split this across a couple of lines.

libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
```diff
@@ -0,0 +1,75 @@
...
+size_t AP_Scripting_SerialAccess::write(uint8_t c)
+{
+    return write(&c, 1);
+}
+
+size_t AP_Scripting_SerialAccess::write(const uint8_t *buffer, size_t size)
+{
+    if (!check_is_device_port()) {
+        return stream->write(buffer, size);
+    } else {
```
peterbarker, line 36: `else` here is redundant

libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
```diff
@@ -0,0 +1,75 @@
...
+        return ON_DEVICE_PORT(write, buffer, size);
+    }
+}
+
+int16_t AP_Scripting_SerialAccess::read(void)
```
peterbarker, line 41: Do we *really* need this?  It's *really* awful as an interface!

libraries/AP_Scripting/AP_Scripting_SerialDevice.cpp
```diff
@@ -0,0 +1,156 @@
...
+  initialise scripting serial ports
+*/
+void AP_Scripting_SerialDevice::init(void)
+{
+    if (enable == 0) {
+        return;
+    }
+
+    for (uint8_t i=0; i<ARRAY_SIZE(ports); i++) {
+        auto &p = ports[i];
```
peterbarker, line 31:
```suggestion
    for (auto &p : ports) {
```

similarly elsewhere

Hwurzburg:
needs a build_options.py and extract_feature.py

peterbarker (changes requested):

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -391,18 +391,21 @@ singleton RC_Channels method lua_rc_channel RC_Channel uint8_t 1 NUM_RC_CHANNELS
...
+userdata AP_Scripting_SerialAccess method begin void uint32_t 1U UINT32_MAX
+userdata AP_Scripting_SerialAccess method write uint32_t uint8_t'skip_check
+userdata AP_Scripting_SerialAccess manual writestring lua_serial_writestring 1 1
+userdata AP_Scripting_SerialAccess method read int16_t
+userdata AP_Scripting_SerialAccess manual readstring lua_serial_readstring 1 1
+userdata AP_Scripting_SerialAccess method available uint32_t
+userdata AP_Scripting_SerialAccess method set_flow_control void AP_HAL::UARTDriver::flow_control'enum AP_HAL::UARTDriver::FLOW_CONTROL_DISABLE AP_HAL::UARTDriver::FLOW_CONTROL_RTS_DE
+
+singleton TotallyFakeSerialSingletonType rename serial
+singleton TotallyFakeSerialSingletonType depends HAL_GCS_ENABLED
```
rmackay9, line 406: Let's rename "TotallyFake" to something.  Even just "Fake" would be better.

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -391,18 +391,21 @@ singleton RC_Channels method lua_rc_channel RC_Channel uint8_t 1 NUM_RC_CHANNELS
...
-ap_object AP_HAL::UARTDriver method write uint32_t uint8_t'skip_check
-ap_object AP_HAL::UARTDriver method available uint32_t
-ap_object AP_HAL::UARTDriver method set_flow_control void AP_HAL::UARTDriver::flow_control'enum AP_HAL::UARTDriver::FLOW_CONTROL_DISABLE AP_HAL::UARTDriver::FLOW_CONTROL_RTS_DE
-
-singleton AP_SerialManager rename serial
-singleton AP_SerialManager depends HAL_GCS_ENABLED
-singleton AP_SerialManager method find_serial AP_HAL::UARTDriver AP_SerialManager::SerialProtocol_Scripting'literal uint8_t'skip_check
+include AP_Scripting/AP_Scripting_SerialAccess.h
+-- don't let user create access objects
+userdata AP_Scripting_SerialAccess creation null -1
```
rmackay9, line 396: I wonder if we should rename this to "AP_Scripting_Serial".

libraries/AP_Scripting/AP_Scripting.cpp
```diff
@@ -160,6 +160,40 @@ const AP_Param::GroupInfo AP_Scripting::var_info[] = {
...
+    AP_GROUPINFO("SER_D2_PRO", 17,  AP_Scripting, _serialdevice.ports[1].state.protocol, -1),
+#endif
+
+#if AP_SCRIPTING_SERIALDEVICE_NUM_PORTS > 2
+    // @Param: SER_D3_PRO
+    // @DisplayName: Serial protocol of scripting serial device
+    // @Description: Serial protocol of scripting serial device
+    // @CopyFieldsFrom: SCR_SER_D1_PRO
+    AP_GROUPINFO("SER_D3_PRO", 18,  AP_Scripting, _serialdevice.ports[2].state.protocol, -1),
+#endif
```
rmackay9, line 195: add a comment here to warn devs away from adding parameters here because they will be hidden by the enable.

libraries/AP_Scripting/AP_Scripting.cpp
```diff
@@ -160,6 +160,40 @@ const AP_Param::GroupInfo AP_Scripting::var_info[] = {
...
+    // @User: Advanced
+    AP_GROUPINFO_FLAGS("SER_EN", 15,  AP_Scripting, _serialdevice.enable, 0, AP_PARAM_FLAG_ENABLE),
+
+    // @Param: SER_D1_PRO
+    // @DisplayName: Serial protocol of scripting serial device
+    // @Description: Serial protocol of scripting serial device
+    // @CopyFieldsFrom: SERIAL1_PROTOCOL
+    // @RebootRequired: True
+    // @User: Advanced
+    AP_GROUPINFO("SER_D1_PRO", 16,  AP_Scripting, _serialdevice.ports[0].state.protocol, -1),
```
rmackay9, line 179: Maybe we could rename this parameter to SER1_PROTO?

libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
tpwrules: Can you help me understand you are pointing to and what is awful about it?

The ultimate problem is that the objects returned by `serial:find_serial` and `serial:find_simulated_device` must be able to access two different sets of methods on the same C++ object. So either they can each return a different type of userdata (which was done formerly and objected to by @IamPete1) or there needs to be a new class (here `AP_Scripting_SerialAccess`) which chooses dynamically. The latter way does save flash (roughly 300B), so I'd say that one.

Here is the first attempt if you want to compare: https://github.com/tpwrules/ardupilot/tree/ssd-v1

libraries/AP_Scripting/generator/description/bindings.desc
tpwrules: `NonexistentSerialSingletonType` ? It just needs some name, not a real type, as all the functions are `manual`.

IamPete1: why not just `serial` and no rename?

libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
tpwrules: Had a brief discussion on the Discord and the objection here is to the `int16_t` return type. It would be less error prone to have something like `bool read(uint8_t&)`. But this would break existing scripts and representing a reference type like this in Lua is not possible (it would have to be a second return value).

In any case I had a plan to revise some other types in a separate PR so I will fold this into that.

libraries/AP_Scripting/generator/description/bindings.desc
tpwrules: Duh, thank you...

libraries/AP_Scripting/generator/description/bindings.desc
tpwrules: I like shorter names. It's not quite a port though, I will probably still use that word in the docs.

libraries/AP_Scripting/AP_Scripting_SerialDevice.cpp
tpwrules: I changed elsewhere, but in this particular case I need `i` in the loop.

libraries/AP_Scripting/generator/description/bindings.desc
tpwrules: I did some thinking and I don't think `AP_Scripting_Serial` is descriptive enough. The only reason I'd see to shorten it is to save the few bytes of flash.

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -737,6 +713,111 @@ int lua_get_CAN_device2(lua_State *L) {
...
+    // write up to that number of bytes
+    const uint32_t written_bytes = port->write((const uint8_t*)data, req_bytes);
+
+    // return the number of bytes that were actually written
+    lua_pushinteger(L, written_bytes);
+
+    return 1;
+}
+
+int lua_serial_readstring(lua_State *L) {
```
tridge, line 799: why doesn't the existing readstring work? A RegisteredPort inherits from AP_HAL::UARTDriver

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -737,6 +713,111 @@ int lua_get_CAN_device2(lua_State *L) {
...
+
+    return 1;
+}
+#endif // AP_SCRIPTING_SERIALDEVICE_ENABLED
+
+int lua_serial_writestring(lua_State *L)
+{
+    binding_argcheck(L, 2);
+
+    AP_Scripting_SerialAccess * port = check_AP_Scripting_SerialAccess(L, 1);
```
tridge, line 784: we should be able to added this as: ```
ap_object AP_HAL::UARTDriver manual writestring AP_HAL__UARTDriver_readstring 1 1
```
and get the functionality for existing ports too

File: libraries/AP_Scripting/tests/serial_loopback.lua [hunk 11]
```diff
@@ -0,0 +1,36 @@
+local ser_driver = serial:find_serial(0)
+local ser_device = serial:find_simulated_device(28, 0)
```
tridge, line 2: it would be nice to not tie this method to just the SCR_SER devices. I've wanted for a while to be able to attach to serial devices that don't have their protocol set to scripting. For example, when writing a rangefinder driver in lua it would be nice for the user to be able to set the protocol to rangefinder instead of scripting.

tridge (changes requested):

libraries/AP_Scripting/lua_bindings.cpp
tpwrules: This is the existing readstring but refactored. It works for both types of ports.

libraries/AP_Scripting/lua_bindings.cpp
tpwrules: This does work for existing ports.

libraries/AP_Scripting/tests/serial_loopback.lua
tpwrules: I don’t understand how that relates. That could be a future improvement, but we would have to figure out how to disconnect the existing driver I think?

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -737,6 +713,111 @@ int lua_get_CAN_device2(lua_State *L) {
...
+
+int lua_serial_readstring(lua_State *L) {
+    binding_argcheck(L, 2);
+
+    AP_Scripting_SerialAccess * port = check_AP_Scripting_SerialAccess(L, 1);
+
+    // create a buffer sized to hold the number of bytes the user wants to read
+    luaL_Buffer b;
+    const uint16_t req_bytes = get_uint16_t(L, 2);
+    uint8_t *data = (uint8_t *)luaL_buffinitsize(L, &b, req_bytes);
```
tridge, line 807: please test that if too large that we do get a lua exception here and not nullptr

tridge (changes requested):

libraries/AP_Scripting/lua_bindings.cpp
tpwrules: Yes, I tested in SITL with a size of 65535 bytes (max supported) and we get a "not enough memory for buffer allocation" exception.

tpwrules:
Checked that the exceptions are safe and it's good, we should be ready to merge! Thanks!

tridge (approved):

PR #27220, opened 2024-06-02
Copter: Clear the counter when rudder arm operation is disabled

muramura:
Initialize the counter if it is changed during operation.


IamPete1 (commented):
Maybe Toy mode too?

khancyr (commented):
it is ok. I don't think it really matter as it would need you to change the parameter twice in 2s and do rudder arming ...

Without the change, it may arm within the 2s 
With the PR, it will arm 2s after changing the parameter to get back the rudder arming

PR #27221, opened 2024-06-02
Copter: Change the description(NFC)

muramura:
Include the methods for LUA scripts within the range of the preprocessor.

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -448,13 +448,13 @@ bool Copter::has_ekf_failsafed() const
 
 #endif // AP_SCRIPTING_ENABLED
 
-// returns true if vehicle is landing. Only used by Lua scripts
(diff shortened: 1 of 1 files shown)


IamPete1 (changes requested):
I don't think this is correct, these methods are used by DroneCAN.

https://github.com/ArduPilot/ardupilot/blob/eb8eeba8516f5fc3a83b82a7f03235c2ca6578ae/libraries/AP_DroneCAN/AP_DroneCAN.cpp#L1033-L1041

muramura:
@IamPete1 san.
Thank you for the review. I will revert the changes and remove the phrase "script only."

muramura:
@IamPete1 san.
I changed the title.
I changed the description.

peterbarker:
Corrected commit message, marking for merge

PR #25459, opened 2023-11-04
Vehicles: Move landed state handling from GCS class to Vehicle class

olliw42:
This PR moves the GCS' landed_state() method to the AP_Vehicle class.

The primary issue this solves is that the GCS' landed_state() method is private and hence cannot be accessed from other classes. For instance the landed_state() would be useful in the AUTOPILOT_STATE_FOR_GIMBAL_DEVICE MAVLink message, which has a respective field, but mounts may have to not use the common function to account for specific behavior and for disadvantages of the common function (like jitter and broadcast).

The simplest approach would be to simply make landed_state() public, but accessing wouldn't be nice. A nicer approach would be to add a public getter to the GCS class. In this PR the more intrusive approach of moving the code from the GCS class to the Vehicle class is taken, since logically the landed state is really a vehicle property and not a GCS thing. Indeed it nicely ammends the existing is_landing() and is_taking_off().

The PR should be fully nfc, since only code is shuffled around. Compiles for all vehicles, and lightly tetsed for copter.

PR #27222, opened 2024-06-02
LUA: Add a function to retrieve the landed state

muramura:
I want to get the value of LANDED_STATE to determine the landing status in LUA. I will add a method to retrieve the LANDED_STATE. This ticket is related to #27221 .

RESULT
Screenshot from 2024-06-02 15-32-46

LUA SCRIPT
Screenshot from 2024-06-02 17-21-16

### DIFF
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -2388,6 +2388,15 @@ function vehicle:is_taking_off() end
 function vehicle:is_landing() end
 
 -- desc
+---@return integer
+---| '0' # unknown
+---| '1' # landed (on ground) 
+---| '2' # in air
+---| '3' # currently taking off
+---| '4' # currently landing 
+function vehicle:get_landed_state() end
+
+-- desc
 onvif = {}
 
 -- desc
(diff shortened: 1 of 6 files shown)


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2387,6 +2387,10 @@ function vehicle:is_taking_off() end
 ---@return boolean
 function vehicle:is_landing() end
 
+-- desc
+---@return integer
+function vehicle:get_landed_state() end
```
IamPete1, line 2392: Full descriptions please. 

You can enumerate the return states like this:

https://github.com/ArduPilot/ardupilot/blob/eb8eeba8516f5fc3a83b82a7f03235c2ca6578ae/libraries/AP_Scripting/docs/docs.lua#L1416-L1423

IamPete1 (changes requested):

olliw42:
this is a bit strange. There is already a landed_state() in GCS_MAVlink which does exactly the same, so it's a good part code duplication. Also, why should only Copter get this lua. 

Note that there had been this PR https://github.com/ArduPilot/ardupilot/pull/25459, proposing to move em to the vehicle class, which should also adress this here.

olliw42:
you do not need to use either to call it in lua
the function you define newly here is exactly identical to the function I mentioned is already there. Whatever name you give it, it doesn't change that fact :)

muramura:
I have implemented the process to determine the LAND state in the COPTER.
By doing this, multiple methods are no longer needed.
I have added a description of the return value to the LUA function.

PR #27223, opened 2024-06-02
Lua binding to set airspeed for a fixed wing Plane

timtuxworth:
For a method to allow plane airspeed to be set for a fixed wing. Currently this will only work in Guided and Auto mode.

### DIFF
--- a/ArduPlane/ArduPlane.cpp
+++ b/ArduPlane/ArduPlane.cpp
@@ -963,6 +963,18 @@ bool Plane::flight_option_enabled(FlightOptions flight_option) const
     return g2.flight_options & flight_option;
 }
 
+#if AP_SCRIPTING_ENABLED
+// this implements the virtual function defined in AP_Vehicle for Plane to set the desired airspeed in m/s
+bool Plane::set_desired_airspeed(float airspeed_new)
+{
+    if (control_mode->can_override_throttle()) {
+        plane.new_airspeed_cm = constrain_float(airspeed_new, aparm.airspeed_min, aparm.airspeed_max) * 100.0f;
+        return true;
+    }
+    return false;
+}
+#endif
+
 #if AC_PRECLAND_ENABLED
 void Plane::precland_update(void)
 {
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -2427,6 +2427,11 @@ function vehicle:nav_scripting_enable(param1) end
 ---@return boolean
 function vehicle:set_desired_speed(param1) end
 
+-- Set autopilot desired airspeed (Plane)
+---@param airspeed_new number -- new airspeed in m/s
+---@return boolean -- true if successful
+function vehicle:set_desired_airspeed(airspeed_new) end
+
 -- desc
 ---@param param1 number
 ---@param param2 number
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -1045,6 +1045,9 @@ private:
     void update_flight_stage();
     void set_flight_stage(AP_FixedWing::FlightStage fs);
     bool flight_option_enabled(FlightOptions flight_option) const;
+#if AP_SCRIPTING_ENABLED
+    bool set_desired_airspeed(float airspeed_new) override;    // different from set_desired_speed() in rover and copter
+#endif
 
     // navigation.cpp
     void loiter_angle_reset(void);
--- a/ArduPlane/mode.h
+++ b/ArduPlane/mode.h
@@ -118,6 +118,10 @@ public:
     // true if the mode sets the vehicle destination, which controls
     // whether control input is ignored with STICK_MIXING=0
     virtual bool does_auto_throttle() const { return false; }
+
+    // flags if the throttle can be overriden by setting new_airspeed_cm
+    // eventually it should return does_auto_throttle() but for now this is not correct
+    virtual bool can_override_throttle() const { return false; }
     
     // true if the mode supports autotuning (via switch for modes other
     // that AUTOTUNE itself
@@ -225,7 +229,9 @@ public:
     bool does_auto_navigation() const override;
 
     bool does_auto_throttle() const override;
-    
+
+    bool can_override_throttle() const override { return does_auto_throttle(); }
+
     bool mode_allows_autotuning() const override { return true; }
 
     bool is_landing() const override;
@@ -301,6 +307,8 @@ public:
 
     bool does_auto_throttle() const override { return true; }
 
+    bool can_override_throttle() const override { return does_auto_throttle(); }
+    
     // handle a guided target request from GCS
     bool handle_guided_request(Location target_loc) override;
 
--- a/libraries/AP_Scripting/examples/plane-set-airspeed.lua
+++ b/libraries/AP_Scripting/examples/plane-set-airspeed.lua
@@ -1,0 +1,46 @@
+--[[
+   demonstrate setting the airspeed for a plane in fixed wing mode
+--]]
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+
+local AUX_FN = 300
+local last_sw = -1
+
+--[[
+   If the user changes the position of the scripting switch choose either
+   LOW: AIRSPEED_MIN
+   MID: AIRSPEED_CRUISE
+   HIGH: AIRSPEED_MAX
+--]]
+local function set_desired_speed()
+    local sw_current = rc:get_aux_cached(AUX_FN)
+    if not sw_current then
+       -- ignore untill the user flips the switch
+       return
+    end
+    if sw_current ~= last_sw then
+        last_sw = sw_current
+        if sw_current == 0 then
+           new_airspeed = Parameter("AIRSPEED_MIN"):get()
+           gcs:send_text(MAV_SEVERITY.INFO, string.format("Speed set to AIRSPEED_MIN %f", new_airspeed))
+        elseif sw_current == 1 then
+            new_airspeed = Parameter("AIRSPEED_CRUISE"):get()
+            gcs:send_text(MAV_SEVERITY.INFO, string.format("Speed set to AIRSPEED_CRUISE %f", new_airspeed))
+        else
+            new_airspeed = Parameter("AIRSPEED_MAX"):get()
+            gcs:send_text(MAV_SEVERITY.INFO, string.format("Speed set to AIRSPEED_MAX %f", new_airspeed))
+        end
+        if not vehicle:set_desired_airspeed(new_airspeed) then
+            gcs:send_text(MAV_SEVERITY.ERROR, string.format("Failed to set airspeed"))
+        end
+     end
+end
+
+local function update()
+   if arming:is_armed() then
+      set_desired_speed()
+   end
+   return update, 1000
+end
+
+return update, 1000
--- a/libraries/AP_Scripting/generator/description/bindings.desc
+++ b/libraries/AP_Scripting/generator/description/bindings.desc
@@ -339,6 +339,7 @@ singleton AP_Vehicle method set_target_throttle_rate_rpy void float -100 100 flo
 singleton AP_Vehicle method set_rudder_offset void float'skip_check boolean
 singleton AP_Vehicle method set_desired_turn_rate_and_speed boolean float'skip_check float'skip_check
 singleton AP_Vehicle method set_desired_speed boolean float'skip_check
+singleton AP_Vehicle method set_desired_airspeed boolean float'skip_check
 singleton AP_Vehicle method nav_scripting_enable boolean uint8_t'skip_check
 singleton AP_Vehicle method set_velocity_match boolean Vector2f
 singleton AP_Vehicle method set_land_descent_rate boolean float'skip_check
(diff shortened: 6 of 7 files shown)


timtuxworth:
I thought it might be a good idea to split this out of my scripting PR's so the AP/c++ code can get potentially merged without getting caught up in discussions about the Lua code. I'm thinking others might also find this useful. @IamPete1 could I ask you to take a look?

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2349,6 +2349,12 @@ function vehicle:nav_scripting_enable(param1) end
 ---@return boolean
 function vehicle:set_desired_speed(param1) end
 
+-- desc set the guided mode radious and direction used for the final loiter at the target
```
IamPete1, line 2352:
```suggestion
-- Set the guided mode radius and direction used for the final loiter at the target
```

ArduPlane/ArduPlane.cpp
```diff
@@ -963,6 +963,28 @@ bool Plane::flight_option_enabled(FlightOptions flight_option) const
...
+    
+    return true;
+}
+#endif
+
+#if AP_SCRIPTING_ENABLED
+// Helper function to let scripting set the guided mode radius
+bool Plane::set_guided_radius_and_direction(float radius, bool direction_is_ccw)
+{
+    if (control_mode->mode_number() == Mode::GUIDED) {
```
IamPete1, line 980:
```suggestion
    if (control_mode == &plane.mode_guided) {
```

ArduPlane/ArduPlane.cpp
```diff
@@ -963,6 +963,28 @@ bool Plane::flight_option_enabled(FlightOptions flight_option) const
     return g2.flight_options & flight_option;
 }
 
+#if AP_SCRIPTING_ENABLED
+// this implements the virtual function defined in AP_Vehicle and already used in Copter
+bool Plane::set_desired_speed(float speed)
```
IamPete1, line 968: I'm really not sure about using `set_desired_speed` which both copter and rover implement as a ground speed to set airspeed on plane. Firstly because its inconsistent and secondly because it prevents us implementing it as a ground speed in the future (VTOL, taxiing ect)

I would much rather see a new `set_desired_airspeed`.

IamPete1: In any case, you would need to update the docs, because the currently say it only work on copter and rover.

https://github.com/ArduPilot/ardupilot/blob/eb8eeba8516f5fc3a83b82a7f03235c2ca6578ae/libraries/AP_Scripting/docs/docs.lua#L2347

timtuxworth: Glad I asked. I didn't realize that, I'll make the change.

timtuxworth: I'd like to make set_desired_airspeed() take m/s instead of cm/s. Do you think that makes sense @IamPete1 ?

libraries/AP_Scripting/docs/docs.lua
timtuxworth: done

ArduPlane/ArduPlane.cpp
timtuxworth: done

ArduPlane/ArduPlane.cpp
timtuxworth: done

timtuxworth:
Do you have any thoughts @Hwurzburg ?

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2349,6 +2349,17 @@ function vehicle:nav_scripting_enable(param1) end
 ---@return boolean
 function vehicle:set_desired_speed(param1) end
 
+-- desc sets autopilot desired airspeed (Plane)
```
IamPete1, line 2352:
```suggestion
-- Set autopilot desired airspeed (Plane)
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2349,6 +2349,17 @@ function vehicle:nav_scripting_enable(param1) end
...
+---@param airspeed_new number -- new airspeed in m/s
+---@return boolean
+function vehicle:set_desired_airspeed(airspeed_new) end
+
+-- desc set the guided mode radius and direction used for the final loiter at the target
```
IamPete1, line 2357:
```suggestion
-- Set the guided mode radius and direction used for the final loiter at the target (Plane)
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2349,6 +2349,17 @@ function vehicle:nav_scripting_enable(param1) end
 ---@return boolean
 function vehicle:set_desired_speed(param1) end
 
+-- desc sets autopilot desired airspeed (Plane)
+---@param airspeed_new number -- new airspeed in m/s
+---@return boolean
```
IamPete1, line 2354:
```suggestion
---@return boolean -- true if successful
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2349,6 +2349,17 @@ function vehicle:nav_scripting_enable(param1) end
...
+function vehicle:set_desired_airspeed(airspeed_new) end
+
+-- desc set the guided mode radius and direction used for the final loiter at the target
+---@return boolean
```
IamPete1, line 2358:
```suggestion
---@return boolean -- true if successful 
```

ArduPlane/ArduPlane.cpp
```diff
@@ -963,6 +963,28 @@ bool Plane::flight_option_enabled(FlightOptions flight_option) const
     return g2.flight_options & flight_option;
 }
 
+#if AP_SCRIPTING_ENABLED
+// this implements the virtual function defined in AP_Vehicle for Plane to set the desired airspeed in m/s
+bool Plane::set_desired_airspeed(float airspeed_new)
+{
+    plane.new_airspeed_cm = constrain_float(airspeed_new, aparm.airspeed_min, aparm.airspeed_max) * 100.0f;
```
IamPete1, line 970: I wonder if we should return false if outside the valid range and not do anything. I'm not sure...

IamPete1: Looks like this only works in guided and auto, so we should return false in any other mode.

IamPete1 (dismissed):

libraries/AP_Scripting/docs/docs.lua
timtuxworth: Done

libraries/AP_Scripting/docs/docs.lua
timtuxworth: Done

libraries/AP_Scripting/docs/docs.lua
timtuxworth: Done

libraries/AP_Scripting/docs/docs.lua
timtuxworth: Done

ArduPlane/ArduPlane.cpp
timtuxworth: Done

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2389,6 +2389,17 @@ function vehicle:nav_scripting_enable(param1) end
...
 
+-- Set autopilot desired airspeed (Plane)
+---@param airspeed_new number -- new airspeed in m/s
+---@return boolean -- true if successful
+function vehicle:set_desired_airspeed(airspeed_new) end
+
+-- Set the guided mode radius and direction used for the final loiter at the target (Plane)
+---@return boolean -- true if successful
+---@param radius number
+---@param direction_is_ccw boolean
```
IamPete1, line 2400:
```suggestion
---@param radius number
---@param direction_is_ccw boolean
---@return boolean -- true if successful
```
Sorry, just spotted this, the return is typically at the bottom.

timtuxworth: Dammit I broke it.

ArduPlane/ArduPlane.cpp
```diff
@@ -963,6 +963,30 @@ bool Plane::flight_option_enabled(FlightOptions flight_option) const
...
+{
+    if (control_mode == &mode_guided || control_mode == &mode_auto) {
+        plane.new_airspeed_cm = constrain_float(airspeed_new, aparm.airspeed_min, aparm.airspeed_max) * 100.0f;
+        return true;
+    }
+    return false;
+}
+#endif
+
+#if AP_SCRIPTING_ENABLED
```
peterbarker, line 978:
```suggestion

```

ArduPlane/ArduPlane.cpp
```diff
@@ -963,6 +963,30 @@ bool Plane::flight_option_enabled(FlightOptions flight_option) const
...
+// Helper function to let scripting set the guided mode loiter radius and direction
+bool Plane::set_guided_radius_and_direction(float radius, bool direction_is_ccw)
```
rmackay9, line 980: I think we should remove or replace the "guided" part of the method name.  So perhaps change it to just, "set_radius_and_direction" or "set_circle_radius_and_direction" or "set_loiter_radius_and_direction".

timtuxworth: This is implementing a virtual method that is defined in AP_Vehicle.h. I'm thinking we wouldn't want to change that.

timtuxworth: The thing is it really only sets the radius and direction _for guided mode_ - it doesn't set the loiter radius/direction for many other places in plane, only the specific one used for guided mode. 
The way the guided mode radius_direction works is it is an override for the general loiter behavior. If not set then guided mode will use the general loiter values, but if set, they will apply for guided mode only.

ArduPlane/ArduPlane.cpp
timtuxworth: fixed

timtuxworth:
This has been tested on a real quadplane by a partner.

timtuxworth:
Based on the dev call 2024-06-17 could you take a look at this please @tridge ? (no rush)

ArduPlane/ArduPlane.cpp
```diff
@@ -963,6 +963,28 @@ bool Plane::flight_option_enabled(FlightOptions flight_option) const
...
+{
+    if (control_mode->is_guided_mode() || control_mode == &mode_auto) {
+        plane.new_airspeed_cm = constrain_float(airspeed_new, aparm.airspeed_min, aparm.airspeed_max) * 100.0f;
+        return true;
+    }
+    return false;
+}
+
+// Helper function to let scripting set the guided mode loiter radius and direction
+bool Plane::set_guided_radius_and_direction(float radius, bool direction_is_ccw)
```
tridge, line 978: I would prefer to have bindings to the functionality under the OFFBOARD_GUIDED code in plane, which is currently accessed via MAVLink with MAV_CMD_GUIDED_xxx funtions.
That will give a lot more functionality and re-uses existing tested capability.
The set_guided_radius_and_direction() method is really just a workaround for not having the OFFBOARD_GUIDED APIs

tridge (changes requested):
@peterbarker I think this PR is a great example of where we need generic access to MAV_CMD_xxx from lua. As we have discussed I think we should just bind it to the first mavlink GCS instance instead of creating a dedicated instance (at least for now), but open to discussion
If we did that then all the MAV_CMD_GUIDED_xxx functionality would be available at very low cost, along with all the other fns

ArduPlane/ArduPlane.cpp
```diff
@@ -963,6 +963,28 @@ bool Plane::flight_option_enabled(FlightOptions flight_option) const
     return g2.flight_options & flight_option;
 }
 
+#if AP_SCRIPTING_ENABLED
+// this implements the virtual function defined in AP_Vehicle for Plane to set the desired airspeed in m/s
+bool Plane::set_desired_airspeed(float airspeed_new)
+{
+    if (control_mode->is_guided_mode() || control_mode == &mode_auto) {
```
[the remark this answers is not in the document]
IamPete1: As it is currently implemented those are the only two modes that this command works in, hence the check. Totally agree that it should work in more modes in the future. A virtual mode method makes sense. But we don't necessarily have to have that before merging this.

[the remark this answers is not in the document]
timtuxworth: I agree with @tridge  but for a different reason. Regardless what modes can/can't be overridden, it shouldn't be code in this API layer that prevents something from working that might later change at a lower level. The mode knows if it can be overridden or not, so creating a can_override_airspeed() on the mode is a better way to implement this check.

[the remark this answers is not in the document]
timtuxworth: I've changed it to call a "working" version of can_override_airspeed() that returns the correct values based on current functionality.

ArduPlane/ArduPlane.cpp
timtuxworth: I've removed this in favor of Tridge's MAVLink Api suggestion #37265

ArduPlane/ArduPlane.cpp
timtuxworth: Ive removed this. I'll use the MAVLink API.

timtuxworth:
Tridge's MavLink Command_int Lua interface PR is at least as good as this.

PR #27224, opened 2024-06-02
AP_Scripting: checks fixes and improvements 

IamPete1:
A number of fixes:

- luacheck ignores the new lua-language-sever lua code that is downloaded when you first run the new checks (the code does not pass the checks configured for the AP environment)
- lua-language checks can now be run on a single file. The file is copied to a new temporary directory the checks run and the directory is then deleted. 
- lua-language checks stop using gitignore so they will check scripts your working on in the SITL `scripts` dir.
- lua-language checks report the number of files it has tested and errors out if the number is 0 or not known.
- the docs document the logger write function both with and without units and multipliers so the checks will now pass in both cases.

### DIFF
--- a/Tools/scripts/run_lua_language_check.py
+++ b/Tools/scripts/run_lua_language_check.py
@@ -16,11 +16,16 @@ import json
 import shutil
 import platform
 from urllib.parse import unquote
+import subprocess
+import re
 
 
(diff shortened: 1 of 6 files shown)


tridge (approved):

tridge:
thannks Pete!

PR #27225, opened 2024-06-03
AC_PrecLand: Add Aux Switch to enable/disable Copter Prec Land

rishabsingh3003:


libraries/AC_PrecLand/AC_PrecLand.h
```diff
@@ -132,6 +132,11 @@ class AC_PrecLand
     // parameter var table
     static const struct AP_Param::GroupInfo var_info[];
 
+    // Allow Aux Function to enable/disable Copter PrecLand routine.
+    // This is different from the "Enable" parameter since we want the plnd data to be logged and the sensor to be working normally.
+    void plnd_active(bool active) { _plnd_active = active; }
```
peterbarker, line 137: Please don't override like this.
```suggestion
    void set_active(bool active) { _plnd_active = active; }
```

libraries/AC_PrecLand/AC_PrecLand.h
```diff
@@ -132,6 +132,11 @@ class AC_PrecLand
...
+    bool plnd_active() { return _plnd_active; }
```
peterbarker, line 138:
```suggestion
    bool active() { return _plnd_active; }
```

`plnd_` is implied

libraries/AC_PrecLand/AC_PrecLand.h
```diff
@@ -232,6 +237,8 @@ class AC_PrecLand
 
     TargetState                 _current_target_state;  // Current status of the landing target
 
+    bool                       _plnd_active;            // user can use aux function to turn Copter PrecLand on/off
```
peterbarker, line 240:
```suggestion
    bool                       _active;            // user can use aux function to turn Copter PrecLand on/off
```

I'm really not sure about using the word "active" to describe this.  It's a bit ambiguous as it might also mean that it is actively tracking a target.

enabled/disabled would be more typical.

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -1702,6 +1706,27 @@ bool RC_Channel::do_aux_function(const AUX_FUNC ch_option, const AuxSwitchPos ch
...
+        if (_precland == nullptr) {
+            // precland not enabled
+            break;
+        }
+
+        switch (ch_flag) {
+        case AuxSwitchPos::HIGH:
+            _precland->plnd_active(true);
+            break;
+        case AuxSwitchPos::MIDDLE:
```
peterbarker, line 1722:
```suggestion
        case AuxSwitchPos::MIDDLE:
            break;
```

this allows us to use the middle position for something else later.

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -1702,6 +1706,27 @@ bool RC_Channel::do_aux_function(const AUX_FUNC ch_option, const AuxSwitchPos ch
     }
 #endif
 
+
+#if AC_PRECLAND_ENABLED
+    case AUX_FUNC::PLND: {
```
peterbarker, line 1711:
```suggestion
    case AUX_FUNC::PRECLAND: {
```

peterbarker (changes requested):
Have you considered the initialisation of this new aux function?  i.e. does the feature get turned on/off at boot based on switch position?

libraries/AC_PrecLand/AC_PrecLand.h
rishabsingh3003: ENABLED was already used in the parameter to "enable" the library, ie PLND_ENABLED.  So to keep it independent of that, I used "active". Happy to word it differently though.

peterbarker: Perhaps look at where that existing function is used and see if we can untie it from the parameter?

khancyr (approved):
LGTM. I think  active is ok. That is was we use on some other place for plnd anyway

peterbarker:
Ping @rishabsingh3003 - needs a rebase.

PR #27226, opened 2024-06-03
Update MAVLink parsing for wiki inclusion

ES-Alexander:
ArduPilot Wiki uses reStructuredText (rst) source files, so supporting rst output is essential for this script to generate wiki files. Ideally this can be run automatically for every vehicle type and version, like the existing parameter lists.

- [x] Add rst support
- [ ] Add per-vehicle dropdown version selection support, like the parameter lists
    - Not yet sure how this is done
- [ ] Set up automatic generation per vehicle, per version
    - May be done via autotest, [like for parameters](https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/autotest.py#L194)
        - Moving this to autotest could be useful for notifying of changes in MAVLink support on PRs, but is probably too much effort for this PR
    - Will likely get done externally, in what is currently [the parameters repository](https://github.com/ArduPilot/ParameterRepository/blob/main/scripts/run_param_parse.py#L132)
        - Extra attractive because the script requires Python &geq; 3.11, which will be easier to specify there than forcing autotest to update

Example file for ArduSub 4.1, with all options included:
[ArduSub_Sub-4.1_MAVLink_Messages.rst.zip](https://github.com/user-attachments/files/15531781/ArduSub_Sub-4.1_MAVLink_Messages.rst.zip)

### DIFF
--- a/Tools/scripts/mavlink_parse.py
+++ b/Tools/scripts/mavlink_parse.py
@@ -152,7 +152,8 @@ class MAVLinkDetector:
     ARDUPILOT_URL = 'https://github.com/ArduPilot/ardupilot/tree/{branch}/{source}'
     EXPORT_FILETYPES = {
         'csv': 'csv',
-        'markdown': 'md'
+        'markdown': 'md',
+        'rst': 'rst',
     }
 
     MARKDOWN_INTRO = (
@@ -169,6 +170,9 @@ class MAVLinkDetector:
         ' least shows that the autopilot is aware it exists, and will try'
         ' to do something meaningful with it.{unsupported}{stream_groups}'
     )
+
+    # Convert markdown hyperlinks into rst syntax
+    RST_INTRO = MARKDOWN_INTRO.replace('[', '`').replace('](', ' <').replace(')', '>`_')
 
     VEHICLES = ('AntennaTracker', 'ArduCopter', 'ArduPlane', 'ArduSub', 'Rover')
 
@@ -422,6 +426,74 @@ class MAVLinkDetector:
 
                     print(name, source, dialect, sep=' | ', file=file)
 
+    def export_rst(self, file, iterable, branch='master', header=None, 
+                        use_intro=True, **extra_kwargs):
+        if header == 'ardupilot_wiki':
+            header = '\n'.join((
+                '.. _mavlink_support:',
+                '',
+                '===============',
+                'MAVLink Support',
+                '===============',
+                '\n',
+            ))
+
+        if header:
+            print(header, file=file)
+
+        if use_intro:
+            commands = stream_groups = unsupported = ''
+            if extra_kwargs['include_commands']:
+                commands = ' (and commands)'
+            if extra_kwargs['include_unsupported']:
+                unsupported = (
+                    '\n\nKnown :ref:`unsupported messages <mavlink_missing_messages>`'
+                    f'{commands} are shown at the end.'
+                )
+            if extra_kwargs['include_stream_groups']:
+                stream_groups = (
+                    '\n\nThe autopilot includes a set of :ref:`mavlink_stream_groups`'
+                    ' for convenience, which allow configuring the stream rates of'
+                    ' groups of requestable messages by setting parameter values. '
+                    'It is also possible to manually request messages, and request'
+                    ' individual messages be streamed at a specified rate.'
+                )
+            vehicle = self.vehicle.replace('ALL', 'ArduPilot')
+             
+            print(self.RST_INTRO.format(
+                vehicle=vehicle, commands=commands, 
+                stream_groups=stream_groups, unsupported=unsupported
+            ), '\n', file=file)
+            
+        for data in iterable:
+            match data:
+                case str() as type_:
+                    reference = f'mavlink_{type_}'
+                    heading = type_.title().replace('_', ' ')
+                    source_header = (
+                        'Code Source' if type_ != 'stream_groups' else
+                        'Stream Group Parameter'
+                    )
+                    print(f'\n.. _{reference}:\n\n{heading}\n{"="*len(heading)}\n',
+                          self.get_description(type_).replace('`','``'),
+                          '\n.. csv-table::',
+                          f'  :header: MAVLink Message, {source_header}, MAVLink Dialect\n\n',
+                          sep='\n', file=file)
+                case MAVLinkMessage() as message:
+                    name, source, dialect = message.as_tuple()
+                    if dialect != MAVLinkDialect.UNKNOWN:
+                        msg_url = self.MAVLINK_URL.format(dialect=dialect,
+                                                          message_name=name.split(':')[0])
+                        name = f'`{name} <{msg_url}>`_'
+                    if source != 'UNSUPPORTED' and not source.startswith('SRn'):
+                        folder = source.split('/')[0]
+                        base = 'libraries/' if folder not in self.VEHICLES else '' 
+                        code_url = self.ARDUPILOT_URL.format(branch=branch,
+                                                             source=base+source)
+                        source = f'`{source} <{code_url}>`_'
+
+                    print(f'  {name}', source, dialect, sep=', ', file=file)
+
 
 if __name__ == '__main__':
     from inspect import signature
@@ -431,6 +503,7 @@ if __name__ == '__main__':
     default_vehicle = detector_init_params['vehicle'].default
     vehicle_options = [default_vehicle, *MAVLinkDetector.VEHICLES]
     default_exclusions = detector_init_params['exclude_libraries'].default
+    format_options = [*MAVLinkDetector.EXPORT_FILETYPES, 'none']
 
     parser = ArgumentParser(formatter_class=ArgumentDefaultsHelpFormatter)
     parse_opts = parser.add_argument_group('parsing options')
@@ -449,7 +522,7 @@ if __name__ == '__main__':
     export_opts.add_argument('-q', '--quiet', action='store_true',
                              help='Disable printout, only export a file.')
     export_opts.add_argument('-f', '--format', default='markdown',
-                             choices=['csv', 'markdown', 'none'],
+                             choices=format_options,
                              help='Desired format for the exported file.')
     export_opts.add_argument('-b', '--branch',
                              help=('The branch to link to in markdown mode.'


amilcarlucas:
This seams related to this file that I have been working on for the past weeks:

```python
#!/usr/bin/python3

'''
Rsync apm.pdef.xml files for different versions of the ArduPilot firmware
For each version, it checks out the corresponding tag, generates parameter metadata,
and finally rsyncs the updated parameter metadata pdef.xml files.

(C) 2024 Amilcar do Carmo Lucas

SPDX-License-Identifier:    GPL-3
'''

import os
import getpass
import datetime
import shutil
import subprocess

VEHICLE_TYPES = ["Copter", "Plane", "Rover", "ArduSub", "Tracker"]  # Add future vehicle types here
RSYNC_USERNAME = 'amilcar'

# Store the current working directory
old_cwd = os.getcwd()

def get_vehicle_tags(vehicle_type):
    """
    Lists all tags in the ArduPilot repository that start with the given vehicle type followed by '-'.
    Returns a list of tag names.
    """
    try:
        # Change to the ArduPilot directory
        os.chdir('../ardupilot/')
        tags_output = subprocess.check_output(['git', 'tag', '--list', f'{vehicle_type}-[0-9]\\.[0-9]\\.[0-9]'], text=True)
        return tags_output.splitlines()
    except Exception as e: # pylint: disable=broad-exception-caught
        print(f"Error getting {vehicle_type} tags: {e}")
        return []

def generate_vehicle_versions():
    """
    Generates arrays for each vehicle type based on the tags fetched from the ArduPilot repository.
    """
    vehicle_versions = {}

    for vehicle_type in VEHICLE_TYPES:
        tags = get_vehicle_tags(vehicle_type)
        if tags:
            vehicle_versions[vehicle_type] = [tag.split('-')[1] for tag in tags]

    return vehicle_versions

def create_one_pdef_xml_file(vehicle_type: str, dst_dir: str, git_tag: str):
    os.chdir('../ardupilot')
    subprocess.run(['git', 'checkout', git_tag], check=True)
    # subprocess.run(['git', 'pull'], check=True)
    subprocess.run(['Tools/autotest/param_metadata/param_parse.py', '--vehicle', vehicle_type, '--format', 'xml'], check=True)
    # Return to the old working directory
    os.chdir(old_cwd)

    if not os.path.exists(dst_dir):
        os.makedirs(dst_dir)

    # Insert an XML comment on line 3 in the ../ardupilot/apm.pdef.xml file to indicate
    # the tag used to generate the file and the current date
    with open('../ardupilot/apm.pdef.xml', 'r', encoding='utf-8') as f:
        lines = f.readlines()
    lines.insert(2, f'<!-- Generated from git tag {git_tag} on {datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")} -->\n')
    with open('../ardupilot/apm.pdef.xml', 'w', encoding='utf-8') as f:
        f.writelines(lines)
    shutil.copy('../ardupilot/apm.pdef.xml', f'{dst_dir}/apm.pdef.xml')

# Function to sync files using rsync
def sync_to_remote(vehicle_type: str) -> None:
    src_dir = f'{vehicle_type}/'
    dst_user = RSYNC_USERNAME
    dst_host = 'firmware.ardupilot.org'
    dst_path = f'param_versioned/{vehicle_type}/'

    # Construct the rsync command
    rsync_cmd = [
        'rsync',
        '-avz',
        '--progress',
        '--password-file=-',  # Use stdin for password
        src_dir,
        f'{dst_user}@{dst_host}::{dst_path}'
    ]

    print(f'Synchronizing {src_dir} to {dst_path}...')
    print(rsync_cmd)
    subprocess.run(rsync_cmd, check=True)

def main():
    vehicle_versions = generate_vehicle_versions()

    # Prompt for the password securely
    _remote_password = getpass.getpass("Enter the remote server password: ")

    # Iterate over the vehicle_versions list
    for vehicle_type, versions in vehicle_versions.items():
        for version in versions:

            if version[0] == '3' and vehicle_type != 'AP_Periph':
                continue # Skip ArduPilot 3.x versions, as param_parse.py does not support them out of the box
            if version[0] == '4' and version[2] == '0' and vehicle_type != 'ArduSub':
                continue # Skip ArduPilot 4.0.x versions, as param_parse.py does not support them out of the box
            vehicle_dir = vehicle_type

            if vehicle_type == 'ArduSub':
                vehicle_dir = 'Sub'

            create_one_pdef_xml_file(vehicle_type, f'{vehicle_dir}/stable-{version}', f'{vehicle_type}-{version}')
        sync_to_remote(vehicle_type)

if __name__ == '__main__':
    main()
```

Maybe we can combine this together?

ES-Alexander:
@amilcarlucas where/when are you expecting/wanting yours to run?

As far as I'm aware the [parameters repository](https://github.com/ArduPilot/ParameterRepository/tree/main) already generates all the `apm.pdef.xml` files (with CI checks for new tags to process every 5 days), so I expect it could be reasonably straightforward to add your rsync updates [there](https://github.com/ArduPilot/ParameterRepository/blob/main/scripts/run_param_parse.py#L145) as just an extra step?

I'm not sure how much sense it would make to combine efforts on a PR in that repo, given your feature is an additional endpoint for the parameters, whereas mine is a new script to run and store the outputs of, but if you're also wanting to include the MAVLink message support in your rsync transfers then I suppose that ties them together somewhat?

amilcarlucas:
Nevermind, I jumped to conclusions.

ES-Alexander:
Did some more digging and found that adding a version selection dropdown makes more sense to handle as a patch during generation in the specific applications where it's relevant. Since both of the remaining "planned" components of this PR can be handled in separate repositories, I think this is ready for review / can be merged in 🙂

ES-Alexander:
Not really sure who should review this PR, but @Hwurzburg might have some ideas or preferences around the formatting of the generated files 🤷‍♂️

Hwurzburg:
Was all the formatting in the example sub file generated by this script?

ES-Alexander:
> Was all the formatting in the example sub file generated by this script?

Yeah - the `.rst` example file is just directly from the script output :-)

I generated it from the `Sub-4.1` branch with the following invocation:
```
python mavlink_parse.py -v ArduSub --format rst -b Sub-4.1 --header ardupilot_wiki -gucq
```

(noting once again that the script requires Python 3.11 to run)

Hwurzburg:
I've tagged @peterbarker and @IamPete1  to review if the groupings and generated data is correct....with PeterB's help I could add per vehicle pages into the Dev sites wiki under a MAVLink commands topic....

ES-Alexander:
> I've tagged @peterbarker and @IamPete1 to review if the groupings and generated data is correct

I've just rebased over master, and generated some more rst files for different vehicle types, to make that checking a bit more straightforward:

[MultiVehicle_master_MAVLink_Messages.zip](https://github.com/user-attachments/files/15793798/MultiVehicle_master_MAVLink_Messages.zip)

> with PeterB's help I could add per vehicle pages into the Dev sites wiki under a MAVLink commands topic....

Nice! 😃 
I'm happy to make a PR to generate these automatically in what is currently the parameters repository, if that's helpful, but also happy for some other approach to be taken.

I'd note that because the MAVLink protocol updates, and I'm currently including the "unsupported messages" in the output generation, the outputs for each version should ideally be re-generated each time with the latest script version, and the most recent pymavlink release, rather than tying the script to the vehicle branch and only generating once per version. That said, that's not a critical requirement, and generating once per version could be ok if we either accept that newer messages won't show up, or just don't generate the unsupported messages - I just like the idea of being able to search for a message and finding out clearly if it is definitely unsupported 🤷‍♂️

Hwurzburg:
having to manually update the page generation once in a while is fine for now, IMO....a PR to generate the page would be good, then I can add it to the Dev wiki.....very few users, and only devs, will be using it but it will certainly help them, even if occasionally out of date...

Hwurzburg:
sorry its taken me so long to get back to this, but I wanted to add wiki files for Plane,Copter,Rover messages as of current master....the params repo does not even have files for 4.5...could you generate these three and I will ad to dev wiki...thanks

ES-Alexander:
> sorry its taken me so long to get back to this

I get it - unfortunately the list of interesting/useful things to do seems to grow much faster than the available time to do things 😅

> could you generate these three and I will ad to dev wiki

It's been a while since I did those initial generations, so I'll need to figure out the relevant setup/process again. I'm hoping it's mostly covered by the Action I made in the parameters repo, although from memory that's reliant on this PR being merged, so I might need to make some custom branches to do the manual generations for now.

I'll try to sort that out this week :-)

> I wanted to add wiki files for Plane,Copter,Rover messages as of current master....the params repo does not even have files for 4.5...could you generate these three

To confirm, did you want files for just the 4.5 release, or current master as well?
If I get to doing this before you respond I'll just generate both - I don't expect it to require much extra effort once I have the process working :-)

Hwurzburg:
4.6 is about to drop...so master only will be fine

ES-Alexander:
Just got around to this.

> 4.6 is about to drop...so master only will be fine

Here ya go 🙂
[MultiVehicle_master_MAVLink_Messages_Nov-5.zip](https://github.com/user-attachments/files/17621527/MultiVehicle_master_MAVLink_Messages_Nov-5.zip) 

> the params repo does not even have files for 4.5

Following up on this, it _did_ have 4.5 files, [per my initial pull request to that repo](https://github.com/ArduPilot/ParameterRepository/pull/7/files). Those files then got auto-removed as part of a recent metadata update, and weren't regenerated because [the process for doing so](https://github.com/ArduPilot/ParameterRepository/pull/8) is reliant on this PR being merged so the relevant parsing script option (rst format) is available. Until this PR is merged that will continue to happen, unless I make a hacky workaround that references a branch in my ardupilot fork, which would then need to be updated again once this PR is merged.

Given you're willing to put its outputs in the wiki, is there anything holding back this PR from getting merged?

Hwurzburg:
@ES-Alexander PR for wiki inclusion...thanks https://github.com/ArduPilot/ardupilot_wiki/pull/6379

ES-Alexander:
> PR for wiki inclusion...[ArduPilot/ardupilot_wiki#6379](https://github.com/ArduPilot/ardupilot_wiki/pull/6379)

Nice! Thanks for adding those in :-)

> ... had to reformat a few things in the files.

The generated files? If so, do you remember what you had to change, and are they things we should be able to edit the generator script to provide from the start?

Hwurzburg:
the tag must be different in each file....I retitled the headers

ES-Alexander:
I have a vague recollection I made those the same on purpose, with the idea they'd be replaced automatically as part of the version selection support. That said, I may be misremembering, and it should be possible to at least roughly do something similar as part of the initial generator script if we want to, since it does already know the vehicle(s) being generated for, and the branch.

Hwurzburg:
not required at this point, I can manually run every major release to keep up to date.....going to merge as is...thanks

Tools/scripts/mavlink_parse.py
```diff
@@ -7,7 +7,7 @@
 from dataclasses import dataclass, astuple
 
 from pymavlink.dialects.v20 import (
-    common, icarous, cubepilot, uAvionix, ardupilotmega
+    common, icarous, cubepilot, uAvionix, ardupilotmega, development
```
tridge, line 10: should be just be 'all'? either that or needs minimal
does this end up with lots of duplicates via the recursion of ardupilot -> common -> minimal?

tridge (changes requested):
needs to handle recursion

Tools/scripts/mavlink_parse.py
```diff
@@ -193,7 +198,12 @@ def __init__(self, common_files, vehicle='ALL',
             folder = file.parent.stem
             if folder in exclude_libraries:
                 continue
-            text = file.read_text()
+
+            try:
+                text = file.read_text()
+            except FileNotFoundError:  # Broken symlink
+                continue
```
peterbarker, line 205: Why would we ignore this?!

Tools/scripts/mavlink_parse.py
peterbarker, line 10: Missing minimal.xml and/or parsing of include files and use all.xml?

Check Wiki to see if HEARTBEAT is missing?

peterbarker (changes requested):
maybe just try `all.xml`

Tools/scripts/mavlink_parse.py
ES-Alexander: > should be just be 'all'?

It's nice to break them out into dialects, so that the message definitions can be linked to directly within the MAVLink documentation.

> either that or needs minimal

The messages defined in `minimal` [are duplicated in the `common` doc](https://mavlink.io/en/messages/common.html#HEARTBEAT), so I didn't think it was necessary to break it out further (because the links already work fine). 

Happy to make that change if you think it's preferable to link to the minimal docs instead of the common docs for those couple of messages. Thoughts?

> does this end up with lots of duplicates via the recursion of ardupilot -> common -> minimal?

Nope - the dialects are checked in order (see `determine_dialect`), so it just stops when it gets to the first one that matches. The `MAVLinkDetector` class maps relevant messages to their dialects (and corresponding documentation links) as needed, so the only message duplicates are if the firmware handles the same message in different ways that are checked for (e.g. `SYSTEM_TIME` can be incoming, (automatically) outgoing, and also requestable, so appears in all three tables).

ES-Alexander:
To anyone following along, I've just:
- Rebased this over master
- Cleaned up some CLI argument descriptions
- Added ID numbers to the first column of the tables
    - That was previously [added independently in the docs](https://github.com/ArduPilot/ardupilot_wiki/pull/6789), so it seemed to make sense to include in future generations
- Added `minimal` and some other (new) missing message sets that ArduPilot includes
- Added "Blimp" in the listed vehicle options
- Added error logs when searched for files are missing

Tools/scripts/mavlink_parse.py
ES-Alexander: I've refactored the import code to be dynamic, based on the enum, and added in `minimal` (and the other currently missing message sets) so it shows up correctly.

ES-Alexander: I've refactored the import code to be dynamic, based on the enum (so there's a single source of truth), and added in `minimal` (and the other currently missing message sets that ArduPilot includes) so it shows up correctly (instead of HEARTBEAT being in `common`).

Tools/scripts/mavlink_parse.py
ES-Alexander: I've made it print the missing thing to stderr, so it's at least possible to see it in logs and determine whether it's expected or something has gone wrong.

peterbarker:
>     * Added `minimal` and some other (new) missing message sets that ArduPilot includes

Any reason you're not using "all.xml"?

peterbarker (approved):
LGTM.

@Hwurzburg  has to be happy, 'though.

Hwurzburg:
unfortunately I cant run this to check formats, I don't use 3.11 and changing would cause me many issues on my Linux box that Tridge would have to help me fix...so lets merge this and if there are issues I will flag them

Hwurzburg (approved):

ES-Alexander:
Thanks for the reviews :-)

> Any reason you're not using "all.xml"?

@peterbarker As I understand it `all` shouldn't include any unique messages, so it isn't useful for generating links to the MAVLink docs about the definition of a given message.

That said, on reflection it probably makes sense as a more useful / earlier fallback than "unknown" if a dialect gets added to pymavlink and supported in ArduPilot firmware before it gets included in this script, so I've added it (a single extra line in the "add missing message sets" commit). As a side effect, that does also mean any messages from dialects ArduPilot doesn't use will show up in the unsupported table at the end of the generated files, which could be interpreted as either valuable (for someone searching for a specific message), or just extra noise to wade through...

> I cant run this to check formats

@Hwurzburg the outputs from this branch are generated every 5 days for ArduPilot's `master` branch, over in [this repo](https://github.com/ArduPilot/ParameterRepository) if that's helpful for checking? The *-4.7 folders (and AP_Periph-1.9) are all from the latest version of the script (except the "all" dialect inclusion which I just added, which currently only adds some more unsupported messages).

PR #27228, opened 2024-06-03
Heli dual: Constrain cyclic roll for intermeshing

MattKear:
Currently, we do not constrain cyclic output for roll on intermeshing/coax configs.
This PR fixes it.

### DIFF
--- a/libraries/AP_Motors/AP_MotorsHeli_Dual.cpp
+++ b/libraries/AP_Motors/AP_MotorsHeli_Dual.cpp
@@ -404,7 +404,8 @@ void AP_MotorsHeli_Dual::move_actuators(float roll_out, float pitch_out, float c
             pitch_out = _cyclic_max/4500.0f;
             limit.pitch = true;
         }
-    } else {
+    }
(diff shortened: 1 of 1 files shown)


IamPete1:
I don't understand why we would want to leave the cyclic unconstrained on any type.

MattKear:
This is my thinking: 
It's not that we want to leave the cyclic unconstrained, but for some frame types the input maps to collective not cyclic.
Take roll control for transverse dual for example.  We want to control roll from differential collective, so I don't think it is correct to limit the roll control based on a cyclic limitation.
Do you not agree?

bnsgeyer:
I agree with @MattKear. You want cyclic limiting for intermeshing but not for a side by side. We don’t use the cyclic limit for the tandem in this mixer. We want to be consistent but also I don’t think it is correct to limit the differential collective pitch like it was cyclic.

MattKear:
We have now flight tested the addition of the roll cyclic constraint on an interesting heli.

IamPete1 (approved):
LGTM

bnsgeyer (approved):

PR #27230, opened 2024-06-04
SITL: make simulated SBP2 vastly more accurate

peterbarker:
I have no idea if this is actually the sort of number which these devices will return.

However, for the EKF to be happy with the GPS the reported speed accuracy must be much lower than the SBP2 driver reports when we give it these numbers.

It might be that we are interpretting these fields incorrectly in the driver and that the simulator is, in fact, correct.


IamPete1 (approved):
LGTM, its only the sim.

PR #27233, opened 2024-06-04
AP_BattMonitor: prevent memory corruption when manipulating DroneCAN backends

peterbarker:
```
./Tools/autotest/autotest.py --gdb --debug build.SITLPeriphBattMon build.Copter test.BattCAN
```

Casting a backend to the wrong pointer type and then dereferencing  it is unlikely to end well.  We were doing that in at least two places.

No reason this can't be bad in flight, but the operator is probably insane manipulating these parameters in-flight.

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor.cpp
@@ -480,7 +480,8 @@ AP_BattMonitor::init()
         memset(&state[instance].cell_voltages, 0xFF, sizeof(cells));
         state[instance].instance = instance;
 
-        switch (get_type(instance)) {
+        const auto allocation_type = configured_type(instance);
+        switch (allocation_type) {
 #if AP_BATTERY_ANALOG_ENABLED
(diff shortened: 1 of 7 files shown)


libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -699,7 +702,16 @@ void AP_BattMonitor::read()
...
     for (uint8_t i=0; i<_num_instances; i++) {
-        if (drivers[i] != nullptr && get_type(i) != Type::NONE) {
+            if (drivers[i] == nullptr) {
+                continue;
+            }
+            if (allocated_type(i) != configured_type(i)) {
+                continue;
+            }
+            // allow run-time disabling; this is technically redundant
+            if (configured_type(i) == Type::NONE) {
```
IamPete1, line 712: I agree with the comment, why do we need this?

IamPete1 (commented):
I suspect we have the same issues in lots of drivers.....

tridge (approved):

PR #26606, opened 2024-03-24
Heli: Add blade pitch angle logging

MattKear:
This pr adds logging for blade pitch angle contributions from collective and cyclic.  Whilst it is possible to calculate these values from other log fields it requires the user to understand how we do the blade angle conversions and to know how the mixers work.  This is therefore a convenient logging field.

This PR only makes changes to logging.  There are no functional changes to flight dynamics or actuators.
For cyclic angle contributions to be accurate users will need to set the newly added H_CYC_MAX_ANG param.

Example from dual heli logging in SITL:
image

PR #27234, opened 2024-06-04
AP_Motors: Heli: Fix SWSH logging for reversed collectives

MattKear:
I didn't account for reversed collective directions correctly in my original PR that added swash plate angle logging: https://github.com/ArduPilot/ardupilot/pull/26606

Plot below shows the issue.  This aircraft has two reversed collectives and the collective angle range has come out reversed:
image

This fix up resolves the issue by storing the collective value before we apply the collective reversal for the control outputs.

### DIFF
--- a/libraries/AP_Motors/AP_MotorsHeli_Swash.cpp
+++ b/libraries/AP_Motors/AP_MotorsHeli_Swash.cpp
@@ -211,15 +211,16 @@ void AP_MotorsHeli_Swash::add_servo_raw(uint8_t num, float roll, float pitch, fl
 // calculates servo output
 void AP_MotorsHeli_Swash::calculate(float roll, float pitch, float collective)
 {
+
+    // Store inputs for logging, store col before col reversal to ensure logging comes out with the correct sign (+/-)
+    _roll_input = roll;
+    _pitch_input = pitch;
+    _collective_input_scaled = collective;
+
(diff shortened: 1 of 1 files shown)


libraries/AP_Motors/AP_MotorsHeli_Swash.cpp
```diff
@@ -211,16 +211,17 @@ void AP_MotorsHeli_Swash::add_servo_raw(uint8_t num, float roll, float pitch, fl
...
-    }
 
-    // Store inputs for logging
+    // Store inputs for logging, store col before col reversal to ensure logging comes out with the correct sign (+/-)
     _roll_input = roll;
     _pitch_input = pitch;
     _collective_input_scaled = collective;
 
+    // Collective control direction. Swash moves up for negative collective pitch, down for positive collective pitch
+    if (_collective_direction == COLLECTIVE_DIRECTION_REVERSED){
```
IamPete1, line 221:
```suggestion
    if (_collective_direction == COLLECTIVE_DIRECTION_REVERSED) {
```
could take the opportunity to fix the missing space ;)

IamPete1:
We tested this fix, collective now coming out as expected.

image

IamPete1 (approved):

bnsgeyer (approved):

PR #27237, opened 2024-06-04
Sub: update default streamrates

Williangalvani:


rmackay9:
For what it's worth (I'm not the sub maintainer) but I wouldn't have made this change.  The issue was probably that the GCSs or companion computers aren't doing their part to request just the data they need.  https://ardupilot.org/dev/docs/mavlink-requesting-data.html

PR #27239, opened 2024-06-05
Plane: Merge ArduPlane.cpp and Plane.cpp

peterbarker:
- merge Plane.cpp -> ArduPlane.cpp
 - rename ArduPlane.cpp -> Plane.cpp

Makes it look like Blimp, Copter, Tracker and Rover.

image


IamPete1:
Does make this comment wrong:
https://github.com/ArduPilot/ardupilot/blob/7cf96fb0a1e8451ae8942c7079540356c068227b/ArduPlane/Plane.h#L1017

I think the merge is fine, although it would be nice to split things out into new files if we can.

tridge:
I'd like to delay till the remaining NOAA glider changes are in

tridge (approved):

PR #27240, opened 2024-06-05
Copter: tidy failsafe action handling

peterbarker:
Mode changes will not work when mode not compiled in

### DIFF
--- a/ArduCopter/events.cpp
+++ b/ArduCopter/events.cpp
@@ -379,41 +379,46 @@ void Copter::failsafe_deadreckon_check()
 //  this is always called from a failsafe so we trigger notification to pilot
 void Copter::set_mode_RTL_or_land_with_pause(ModeReason reason)
 {
+#if MODE_RTL_ENABLED
     // attempt to switch to RTL, if this fails then switch to Land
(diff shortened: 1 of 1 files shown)


andyp1per (approved):

rmackay9 (approved):
Looks good but I'd like to check that some testing has been done?  I'm 99% sure it's fine but just to be sure..

IamPete1 (approved):
LGTM.

I wonder if we should have a arming check if the user has set the failsafe action to a mode that does not exist. Or SRTL with SRTL disabled.

peterbarker:
> Looks good but I'd like to check that some testing has been done? I'm 99% sure it's fine but just to be sure..

Yes, I've compiled without RTL and it behaves appropriately.  Only user-visible change is they don't see an attempt to change into an unknown mode before the enter LAND.

peterbarker:
> LGTM.
> 
> I wonder if we should have a arming check if the user has set the failsafe action to a mode that does not exist. Or SRTL with SRTL disabled.

I think that's would be a good check.

I think there might some cleanup available - we've got a series of defines in here which are used for the parameter value - but we might be able to use the FailsafeAction enumeration instead, which might be nice.

PR #27243, opened 2024-06-05
Add back the support for generating compile commands

bugobliterator:
* Adds improvement over the older process, the compile_commands.json are generated before compilation rather than after.

### DIFF
--- a/Tools/ardupilotwaf/ap_library.py
+++ b/Tools/ardupilotwaf/ap_library.py
@@ -29,7 +29,7 @@ vehicle-related headers and fails the build if they do.
 import os
 import re
 
-from waflib import Errors, Task, Utils
+from waflib import Errors, Task, Utils, Logs
 from waflib.Configure import conf
 from waflib.TaskGen import after_method, before_method, feature
 from waflib.Tools import c_preproc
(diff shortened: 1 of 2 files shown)


Tools/ardupilotwaf/ardupilotwaf.py
```diff
@@ -680,7 +680,87 @@ def options(opt):
     g.add_option('--ubsan-abort',
         action='store_true',
         help='''Build using the gcc undefined behaviour sanitizer and abort on error''')
-    
+
+
+def write_compilation_database(bld):
+    """
+    Write the clang compilation database as JSON
```
peterbarker, line 687: clang still relevant here?

Doesn't seem like it, so a comment at least would be good.

MichelleRos (changes requested):
Just tried building Copter for SITL on this branch, but I only got this for compile_commands.json:
image

Tools/ardupilotwaf/ardupilotwaf.py
bugobliterator: removed use of clang

bugobliterator:
> Just tried building Copter for SITL on this branch, but I only got this for compile_commands.json: image

Yeah, there was a minor issue. it should be fixed now @MichelleRos

MichelleRos (changes requested):
Thanks. It is now generating the .json, but it's still struggling with include errors:
image
When I run a diff on the .json generated by yours compared to what I get if I take the same point in master that your PR is based on and instead just reverted the original two commits that started this, what I get is it seems like your PR's json is missing the "-I" commands and some of the "-D" ones for the files in the ArduCopter folder:
image
image

I'm starting to wonder whether this compile commands thing is the right way to go about this... Just removing the compileCommands tag from the c_cpp_properties.json file and adding the whole workspace folder to the include path seems to solve most of the issues (for some reason, whenever that compileCommands tag is set, it seems to ignore the fact the include path). We may need to add a bunch of the defines in the defines section, but I'm wondering whether that might be an easier path than this massive json database?

bugobliterator:
@MichelleRos the issue should be fixed now. Just needed to move the generation stage bit more deeper.

MichelleRos:
@bugobliterator I tried again, but this time it seems like nearly all the libraries' files themselves are missing. Opening any file within the vehicle folder worked correctly, but most of the library files I opened would have include errors.

Also the .json file is ~886 KB whereas my json from the previous one is ~4048 KB (this is with just copter, plane and blimp compiled for sitl).

MichelleRos:
Looks like it's working now :)

PR #27244, opened 2024-06-05
AP_Scripting: add an example to use Copter Standby mode in SITL with LUA

khancyr:
Add an example of standby usage on SITL with a LUA scripting to do the switching of motor control

Will make a blog post and wiki entry about this

### DIFF
--- a/libraries/AP_Scripting/examples/sitl_standby_sim.lua
+++ b/libraries/AP_Scripting/examples/sitl_standby_sim.lua
@@ -1,0 +1,87 @@
+-- This script is a allow to switch the standby function by the switch of the remote controller on 2 simulated fcus.
+
+-- variable to limit the number of messages sent to GCS
+local last_rc_input = 0
+
+-- constants
+local switch_high = 2
+local switch_low = 0
+
+-- Standby function number in ArduPilot
+local standby_function = 76
+-- RC pin to use to select the FCU
+local rc_input_pin = 8
+
+
+-- for fast param access it is better to get a param object,
+-- this saves the code searching for the param by name every time
+local JSON_MASTER = Parameter()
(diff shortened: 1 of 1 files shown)


libraries/AP_Scripting/examples/standby-sim.lua
```diff
@@ -0,0 +1,80 @@
...
+local switch_low = 0
+local standby_function = 76
+local rc_input_pin = 8
+
+
+-- for fast param access it is better to get a param object,
+-- this saves the code searching for the param by name every time
+local JSON_MASTER = Parameter()
+if not JSON_MASTER:init('SIM_JSON_MASTER') then
+  gcs:send_text(6, 'get JSON_MASTER failed')
```
rmackay9, line 18: Looking pretty good.  I'd probably use the entire variable name, "SIM_JSON_MASTER" instead of shortening it because it will help users more quickly identify the issue

rmackay9 (approved):

rmackay9:
One small thing is it might be good to prefix "sitl_" to the front of the script name to make it clear that it's only for use on simulated vehicle?

khancyr:
Updated with @rmackay9 comments

rmackay9:
Hi @khancyr,

Looks like the Lua checks have failed so many you can fixup the issue?

```
Diagnosis complete, 1 problems found, see /__w/ardupilot/ardupilot/lualogs/check.json
/__w/ardupilot/ardupilot/libraries/AP_Scripting/examples/sitl_standby_sim.lua:77:7
	Code: cast-local-type
	This variable is defined as type `integer`. Cannot convert its type to `integer|nil`.
	- `nil` cannot match `integer`
	- Type `nil` cannot match `integer`
```

PR #27245, opened 2024-06-05
GCS_MAVLink: remove MOUNT_CONFIGURE/MOUNT_CONTROL handling by default 

peterbarker:
Also add lifetime hint comments so we remove these in 4.7


rmackay9 (approved):
I'm happy with the mount changes.  I think that the GCSs were updated more than a year ago to use the commands instead of the messages.

PR #27250, opened 2024-06-06
AP_Follow: change to using position-from-origin internally

peterbarker:

In preparation for kinematically-consistent target information.

### DIFF
--- a/libraries/AP_Follow/AP_Follow.cpp
+++ b/libraries/AP_Follow/AP_Follow.cpp
@@ -168,6 +168,26 @@ void AP_Follow::clear_offsets_if_required()
 // get target's estimated location
 bool AP_Follow::get_target_location_and_velocity(Location &loc, Vector3f &vel_ned) const
 {
+    Vector3p pos_ned;
+    Vector3f local_vel_ned;  // so we either change both return parameters or neither
+    if (!get_target_position_and_velocity(pos_ned, local_vel_ned)) {
+        return false;
+    }
+    if (!AP::ahrs().get_location_from_origin_offset_NED(loc, pos_ned)) {
+        return false;
+    }
+    vel_ned = local_vel_ned;
+
+    if (_alt_type == AP_FOLLOW_ALTITUDE_TYPE_RELATIVE) {
+        loc.change_alt_frame(Location::AltFrame::ABOVE_HOME);
+    }
+
+    return true;
+}
+
+// get target's estimated location and velocity, both NED SI from origin
+bool AP_Follow::get_target_position_and_velocity(Vector3p &pos_ned, Vector3f &vel_ned) const
+{
     // exit immediately if not enabled
     if (!_enabled) {
         return false;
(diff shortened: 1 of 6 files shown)


amilcarlucas:
Lokks like a relatively small change:
```
ardusub          0 (0.0000%)     0 (0.0000%)  0 (0.0000%)    0 (0.0000%)                                      388464
antennatracker   0 (0.0000%)     0 (0.0000%)  0 (0.0000%)    0 (0.0000%)                                      637308
ardurover        684 (+0.0418%)  0 (0.0000%)  -4 (-0.0015%)  684 (+0.0417%)                                   326604
arduplane        752 (+0.0422%)  0 (0.0000%)  0 (0.0000%)    752 (+0.0421%)                                   180200
arducopter-heli  732 (+0.0409%)  0 (0.0000%)  4 (+0.0015%)   732 (+0.0408%)                                   171880
blimp            0 (0.0000%)     0 (0.0000%)  0 (0.0000%)    0 (0.0000%)                                      614112
arducopter       716 (+0.0401%)  0 (0.0000%)  -4 (-0.0015%)  716 (+0.0400%)                                   177320
```

peterbarker:
Fixed problems caused by rebase on top of function rename in master

rmackay9:
I think we just need to be careful that we cover all situation:

1. handle AHRS failing to provide a position
2. handle situations where the lead vehicle is using different altitude frames (absolute, relative and terrain following)

timtuxworth:
There are Lua bindings for some of these functions so they are not "internal". 
I'm currently using them and expecting them to return lat/lng - is that going to change?

timtuxworth:
> In preparation for kinematically-consistent target information.

What does that mean?

lthall:
> > In preparation for kinematically-consistent target information.
> 
> What does that mean?

Currently the position jumps each time we get a new update and the interpolation between updates is very basic.

I have been using real-time S-Curves to generate a kinematicly consistent interpolation of the follow me data. I am doing this in pos/vel/accel and heading/heading_rate. This lets you do wonderfully smooth follow-me behaviours. It also acts as a filter for position / velocity noise in your follow-me data.

timtuxworth:
I'm relying on these methods returning lat/long via the Lua bindings as I use the results to call back in via the new MAVLink DO_REPOSITION interface to point the vehicle at the calculated target location. If these change to returning position relative to origin, will my Lua scripts break?

```
get_target_location_and_velocity()
get_target_location_and_velocity_ofs()
get_target_dist_and_vel_ned()
```

peterbarker:
> I'm relying on these methods returning lat/long via the Lua bindings as I use the results to call back in via the new MAVLink DO_REPOSITION interface to point the vehicle at the calculated target location. If these change to returning position relative to origin, will my Lua scripts break?

Yes, they absolutely would break.  Really, really badly.

OTOH, you might have missed the word "internally" in the PR title.

We have no intention of breaking everybody's (eg.) Plane ship-landing scripts by changing the existing "location" bindings, but we will almost certainly add some "position" (offset-from-origin) bindings to LUA at some stage.

timtuxworth:
> > I'm relying on these methods returning lat/long via the Lua bindings as I use the results to call back in via the new MAVLink DO_REPOSITION interface to point the vehicle at the calculated target location. If these change to returning position relative to origin, will my Lua scripts break?
> 
> Yes, they absolutely would break. Really, really badly.
> 
> OTOH, you might have missed the word "internally" in the PR title.
> 
> We have no intention of breaking everybody's (eg.) Plane ship-landing scripts by changing the existing "location" bindings, but we will almost certainly add some "position" (offset-from-origin) bindings to LUA at some stage.

Thanks @peterbarker - I'm really looking forward to this feature going in. After talking to @lthall - now I understand what it's doing, I expect kinematic extrapolation to make a significant improvement to the plane follow Lua function too.

lthall:
Is this ready to take to a dev call?

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -129,7 +129,7 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
     // @Param: _ALT_TYPE
     // @DisplayName: Follow altitude type
     // @Description: Follow altitude type
-    // @Values: 0:absolute, 1:relative
+    // @Values: 0:absolute, 1:relative (assume target's home is at the same altitude as this vehicle's)
```
tridge: move comment to the description

tridge (commented):
please check quadplane ship landing lands within a reasonable distance of the ship target

rmackay9 (approved):

lthall (approved):

libraries/AP_Follow/AP_Follow.cpp
peterbarker: Removed this instead

timtuxworth:
This change has caused my plane follow Lua script to fail. Specifically DO_REPOSITION called by gcs:run_command_int() does not fly to the specified location. It seems to fly to the home location rather than the location specified in the call. 

I've nailed it down to the one commit that changes AP_Follow.cpp and AP_Follow.h - I cherry picked this commit into my local working copy and watched a fully working system start to fail.

Any thoughts on where I should look to figure out why this might be happening @peterbarker ?

PR #27253, opened 2024-06-06
AP_Common: rearrange get_vector_from_origin_NEU to be less destructive

peterbarker:
this change means that if the origin call failes we don't half-update the object before returning false

Extracted from https://github.com/ArduPilot/ardupilot/pull/27250


khancyr (approved):
make sense. LGTM. 
The WARN_IF_UNUSED would protect again such wrong use of the return, but we can still see in the code some place where the half updated vector can be use. So definitively solve some issues

tridge (approved):

PR #27254, opened 2024-06-06
Tidy AP_Follow library

peterbarker:
This PR simply splits up the monolithic `handle_message` function into 4 separate functions:
 - `should_handle_message to clarify the logic, remove the modification of state during this stage and make the total file delta in this PR smaller
 - two methods for handling the two different messages we might consume
 - a method that takes care of the logging

Extracted from https://github.com/ArduPilot/ardupilot/pull/27250

### DIFF
--- a/libraries/AP_Follow/AP_Follow.cpp
+++ b/libraries/AP_Follow/AP_Follow.cpp
@@ -272,27 +272,40 @@ bool AP_Follow::get_target_heading_deg(float &heading) const
     return true;
 }
 
-// handle mavlink DISTANCE_SENSOR messages
-void AP_Follow::handle_msg(const mavlink_message_t &msg)
+// returns true if we should extract information from msg
+bool AP_Follow::should_handle_message(const mavlink_message_t &msg) const
 {
     // exit immediately if not enabled
     if (!_enabled) {
-        return;
+        return false;
     }
 
     // skip our own messages
     if (msg.sysid == mavlink_system.sysid) {
-        return;
+        return false;
     }
 
     // skip message if not from our target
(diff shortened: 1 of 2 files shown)


libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -272,27 +272,40 @@ bool AP_Follow::get_target_heading_deg(float &heading) const
...
+    // this method should be called from an "update()" method:
+    if (_automatic_sysid) {
+        // maybe timeout who we were following...
+        if ((_last_location_update_ms == 0) ||
+            (AP_HAL::millis() - _last_location_update_ms > AP_FOLLOW_SYSID_TIMEOUT_MS)) {
+            _sysid.set(0);
         }
+    }
+
+    if (!should_handle_message(msg)) {
```
khancyr, line 308: shouldn't this be before the sysid set ? that would avoid doing it uncessary if we don't enable follow ?

peterbarker: We need to retire that setting regardless of whether we get a mavlink message, really.

The sysid-set is really cheap, or I wouldn't have done it this way :-)

khancyr: millis isn't that cheap I think. But no big deal. Agree that the check is in a wrong place, for another PR , like a bunch of other refactoring in this file ...

IamPete1 (approved):
LGTM

tridge:
need test of auto-sysid code

tridge:
can merge when tested

peterbarker:
Tested by manipulating the ModeFollow autotes to test different sysids and adding breakpoints.

PR #20074, opened 2022-02-13
Plane: use correct "down" rangefinder on tailsitters

IamPete1:
On a tailsitter in VTOL flight `ROTATION_PITCH_270` rangefinder is not down. This changes to use `ROTATION_YAW_180` in that case. It also takes into account `Q_TRIM_PITCH` if set. 

This does mean that were swapping the source for the rangefinder distance calculation depending on mode.  

In the longer term it would be nice to abstract the height above ground stuff to a library, copter has two instances (up and down), plane should probably also have two (FW and VTOL). There is also no reason we should only use one rangfinder, we should use the one that is closest to pointing down at the time, for example a plane in a 70 deg bank would get a better estimate from a `YAW_90`/`YAW_270` rangefinder than the normal `ROTATION_PITCH_270`. We could also use proximity sensors as a source of readings.

PR #27262, opened 2024-06-06
Plane: support inverted rangefinder for landing

tridge:
this allows for planes where the rangefinder is setup so it sees the ground when landing inverted.

Inverted landing can be useful for some types of payload

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1084,6 +1084,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Bitmask: 11: Disable suppression of fixed wing rate gains in ground mode
     // @Bitmask: 12: Enable FBWB style loiter altitude control
     // @Bitmask: 13: Indicate takeoff waiting for neutral rudder with flight control surfaces
+    // @Bitmask: 14: Support inverted landings
     // @User: Advanced
     AP_GROUPINFO("FLIGHT_OPTIONS", 13, ParametersG2, flight_options, 0),
 
--- a/ArduPlane/altitude.cpp
+++ b/ArduPlane/altitude.cpp
@@ -667,6 +667,14 @@ void Plane::rangefinder_height_update(void)
         // correct the range for attitude (multiply by DCM.c.z, which
         // is cos(roll)*cos(pitch))
         rangefinder_state.height_estimate = distance * ahrs.get_rotation_body_to_ned().c.z;
+        if (flight_option_enabled(FlightOptions::INVERTED_LANDING) && fly_inverted()) {
+            /*
+              allow for rangefinders which point upwards for an
+              inverted landing. Some aircraft are setup for inverted
+              landing to protect camera payloads
+             */
+            rangefinder_state.height_estimate *= -1;
+        }
 
         rangefinder_terrain_correction(rangefinder_state.height_estimate);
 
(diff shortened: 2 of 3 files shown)


IamPete1:
I'm not a fan, this requires you set a upwards facing rangefinder as facing downwards. I would like to see support of this via the new surface distance lib. We would need start using it in plane and then allow it to automatically select the best rangefinder based on the vehicle attitude.

magicrub:
I agree with @IamPete1 , we should add support to Orientation_90

zyromskij:
Hi everyone, this functionality is extremely important for us - for now we had to resort to an "emergency flare" lua script which triggers flare action below a certain rangefinder altitude but obviously doesn't allow for precise terrain tracking, etc. making our landings a bit unsafe.
Is there anything we can do to help with this? We can start test flights right away if someone could confirm that it's implemented correctly.

tridge:
> I would like to see support of this via the new surface distance lib. We would need start using it in plane and then allow it to automatically select the best rangefinder based on the vehicle attitude

we only use rangefinders for landing in fixed wing. When you only do inverted landings this PR offers a very simple and robust solution. We don't want to do ceiling tracking like copter can do, it just makes no sense for planes, and would add a whole heap of complexity and entanglement unnecessarily

tridge:
> we should add support to Orientation_90

possible, but we'd need to add a parameter specifying the orientation we want to use for landing. That is much heaver weight as a change than this is.
Or we make this option same option bit from the PR make plane use Orientation_90 for the landing. It will add quite a lot more flash, no more functionality, but maybe you'd prefer the aesthetics?

IamPete1:
> we only use rangefinders for landing in fixed wing. When you only do inverted landings this PR offers a very simple and robust solution. We don't want to do ceiling tracking like copter can do, it just makes no sense for planes, and would add a whole heap of complexity and entanglement unnecessarily

I will explain in more detail.

There should be no reason we need to specify a rangefinder, and no reason we need to use only one. We know the attitude of the vehicle and the orientation of the rangefinder relative to the vehicle. So we can automatically pick the one which is pointing at the ground. That is my eventual goal for the surface tracking library. What it adds is for the rangefinder being used to change during flight (or use more that one at once) so need some method to transition between them smoothly. The same thing could be used on copter to move between the -90 deg rangefinder to a -45 deg one which will be closer to down if the vehicle is pitched over at 30 deg. It also fixes the current issue of rangefinder landing not working on tailsitters because the -90 deg rangefinder is not down in VTOL modes.

tridge:
@IamPete1 ok, but that is a much larger change, and is addressing something that nobody has (to my knowledge) asked for on plane. This costs 36 bytes and solves the exact problem that the partner has

IamPete1:
> @IamPete1 ok, but that is a much larger change, and is addressing something that nobody has (to my knowledge) asked for on plane. This costs 36 bytes and solves the exact problem that the partner has

The tailsitter issue is long standing, I have a PR from 2022 (https://github.com/ArduPilot/ardupilot/pull/20074) although I was never that happy with it.

My main problem with this is that is really takes users down a dead end where that have to set a range finder that is really facing up as facing down. That leaves us no path to update to a more comprehensive method in the future.

ArduPlane/Parameters.cpp
```diff
@@ -1084,6 +1084,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Bitmask: 11: Disable suppression of fixed wing rate gains in ground mode
     // @Bitmask: 12: Enable FBWB style loiter altitude control
     // @Bitmask: 13: Indicate takeoff waiting for neutral rudder with flight control surfaces
+    // @Bitmask: 14: Support inverted landings
```
tridge, line 1087: henry asked for expanding the comment

tridge:
think about a vehicle:set_hagl(height_agl, timeout_ms)

PR #27271, opened 2024-06-07
Add and use new ModeReason::AUX_FUNCTION

IamPete1:
Aux function are no longer only from RC. This changes to use a dedicated mode reason.

We don't check for the old `ModeReason::RC_COMMAND` anywhere in the main code. Nor is it exposed over MAVLink. There is a scripting binding, and it is used in logging.

The scripting binding is `vehicle:get_control_mode_reason()`, it is not used by any of our examples.

### DIFF
--- a/libraries/AP_Vehicle/ModeReason.h
+++ b/libraries/AP_Vehicle/ModeReason.h
@@ -70,4 +70,5 @@ enum class ModeReason : uint8_t {
   DEADRECKON_FAILSAFE = 50,
   MODE_TAKEOFF_FAILSAFE = 51,
   DDS_COMMAND = 52,
+  AUX_FUNCTION = 53,
 };
--- a/ArduCopter/RC_Channel.cpp
+++ b/ArduCopter/RC_Channel.cpp
@@ -144,7 +144,7 @@ void RC_Channel_Copter::do_aux_function_change_mode(const Mode::Number mode,
     switch(ch_flag) {
     case AuxSwitchPos::HIGH: {
         // engage mode (if not possible we remain in current flight mode)
-        copter.set_mode(mode, ModeReason::RC_COMMAND);
+        copter.set_mode(mode, ModeReason::AUX_FUNCTION);
         break;
     }
     default:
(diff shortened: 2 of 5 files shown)


libraries/AP_Vehicle/ModeReason.h
```diff
@@ -70,4 +70,5 @@ enum class ModeReason : uint8_t {
   DEADRECKON_FAILSAFE = 50,
   MODE_TAKEOFF_FAILSAFE = 51,
   DDS_COMMAND = 52,
+  AUX_FUNCTION = 53,
```
peterbarker, line 73:
```suggestion
  AUX_FUNCTION = 1,
```
... and move it up next to the other one.

I think we can get away with making this an alias for `RC_COMMAND`.  And perhaps add a comment indicating what's going on...

tridge: if its an alias we don't get the distinction in logs, which I think we do want

IamPete1: Yeah, the goal was to get it in the log.

peterbarker: > if its an alias we don't get the distinction in logs, which I think we do want

Ah, I'd missed the fact the mode switch continues to use RC_COMMAND.

peterbarker:
@IamPete1 we could, perhaps, have multiple `AUX_FUNCTION` enumeration entries, and choose a different one based on `AuxFuncTriggerSource` - as you say, there are many ways we can trigger aux functions now-adays.

IamPete1:
> @IamPete1 we could, perhaps, have multiple `AUX_FUNCTION` enumeration entries, and choose a different one based on `AuxFuncTriggerSource` - as you say, there are many ways we can trigger aux functions now-adays.

We also have `AUXF` log, so this mode reason should be enough to tell you to go and look at the `AUXF` where the source is logged.

rmackay9 (approved):
I'm fine with this solution or PeterB's suggestion.

PR #27273, opened 2024-06-08
SITL: log number of times sim paused on serial0 buffer

peterbarker:
SITL pauses the simulation if we do not have a minimum amount of space in its out queue.

Log the number of times we do this.

Fora run of the AutoTuneSwitch test:

image

...shows we drop a packet at the start before we bring up the mavlink connection from autotest.  I think.  `SIM_SPEEDUP` was the default 100.

I've also tested ctrl-z on the terminal running autotest, and the sim starts complaining a lot:

image


peterbarker:
Interesting graph zooming in on that ctrl-z test:
image

tridge (approved):

PR #27277, opened 2024-06-08
AP_Scripting: add handle_external_position_estimate binding

tridge:
can be used for offboard navigation systems
we had this in mavlink but not in lua.

### DIFF
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -3251,6 +3251,13 @@ function arming:disarm() end
 -- It provides estimates for the vehicles attitude, and position.
 ahrs = {}
 
+-- supply an external position estimate to the AHRS (supported by EKF3)
+---@param location Location_ud -- estimated location, altitude is ignored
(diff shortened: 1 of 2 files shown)


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3136,6 +3136,13 @@ function arming:disarm() end
 -- It provides estimates for the vehicles attitude, and position.
 ahrs = {}
 
+-- supply an external position estimate to the AHRS (supported by EKF3)
+---@param location Location_ud -- estimated location, altitude is ignored
+---@param accuracy number -- 1-sigma accuracy in meters
+---@param timestamp_ms uint32_t_ud -- timestamp of reading in ms since boot
```
IamPete1, line 3142:
```suggestion
---@param timestamp_ms uint32_t_ud|integer|number -- timestamp of reading in ms since boot
```
Because of the magic casting we do to uint32.

peterbarker: Done

IamPete1 (approved):
LGMT

PR #27278, opened 2024-06-09
AP_Logger: take a blocking semaphore when writing messages using the block logger

andyp1per:
The current block logger can generate logs that have significant holes. It turns out this is due to the very short time it waits to acquire the write buffer lock before giving up. The file logger blocks and the block logger should do the same. Tested on my Marmotte 5

### DIFF
--- a/libraries/AP_Logger/AP_Logger_Block.cpp
+++ b/libraries/AP_Logger/AP_Logger_Block.cpp
@@ -137,10 +137,7 @@ bool AP_Logger_Block::_WritePrioritisedBlock(const void *pBuffer, uint16_t size,
         return false;
     }
 
-    if (!write_sem.take(1)) {
-         _dropped++;
-        return false;
-    }
+    WITH_SEMAPHORE(write_sem);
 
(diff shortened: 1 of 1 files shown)


libraries/AP_Logger/AP_Logger_Block.cpp
```diff
@@ -137,10 +137,7 @@ bool AP_Logger_Block::_WritePrioritisedBlock(const void *pBuffer, uint16_t size,
         return false;
     }
 
-    if (!write_sem.take(1)) {
-         _dropped++;
-        return false;
-    }
+    write_sem.take_blocking();
```
tridge, line 140: just switch to WITH_SEMAPHORE() and remove the write_sem.give() calls

tridge (changes requested):

libraries/AP_Logger/AP_Logger_Block.cpp
andyp1per: Fixed

peterbarker:
Merged, thanks @andyp1per

PR #27283, opened 2024-06-11
Copter: after declaring an EKF failsafe keep checking for a possible lane switch to recover

andyp1per:
I had an incident where I had an EKF failsafe from which the copter never recovered. This was (I think) because the second lane which was switched to was overall much worse than the first lane which had a temporary glitch, but the code was unable to switch lanes back again once the glitch had cleared.

### DIFF
--- a/ArduCopter/ekf_check.cpp
+++ b/ArduCopter/ekf_check.cpp
@@ -20,6 +20,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 static struct {
     uint8_t fail_count;         // number of iterations ekf or dcm have been out of tolerances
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks for this.  I suspect that incident was no fun at all.

I wonder though if perhaps we're trying to fix the problem within the EKF itself instead of in the vehicle code.  It seems like a better outcome would have been if the EKF didn't switch lanes in the first place or if it had more quickly switched back to the original lane.  I worry that we're just making the switching logic more complex and not addressing the underlying cause.

peterbarker (commented):
Andy will get a log together and discuss with tridge and Paul

PR #27284, opened 2024-06-11
AP_OSD:Update HD fonts

Hwurzburg:
Update HD fonts for use with ArduPilot for DJI and Walksnail HD Goggles. Non functional change. Once merged I can update the links in the wiki.


andyp1per:
@Hwurzburg does this include the latest SneakyFPV fonts? They are the best for WS. Think we already have those ...

Hwurzburg:
These are his latest

rmaia3d:
> These are his latest, and no we don't already have these, only the old Europa...these include a unified ini file, and fonts for the optional CRSF extend link panels, if custom built

In the CRSF extended link panels PR, I only included one variation of the SneakyFPV fonts for Walksnail. Here you are including all the variations with support for the extended panels, which is great, more options for users!  👍 

However, I'm just not sure if these WTFOS fonts have support for the panel extensions. Sneaky's latest fonts for WTFOS do have that support, but I don't know the if the fonts included in this directory are from him or the original WTFOS fonts without the extended icons. Just a heads up.

Hwurzburg:
Rodrigo, you are right .....WTFOS just changed last month to png font....I had worked with Shannon on a unified set of all AP fonts for WS but now have included directories for both the old .bin WTFOS and new (which includes the new CRSF font additions) PNG style....only the Europa one though since you cant select fonts via the goggle menu like WS, but this will allow users reading the wiki to avoid the hunt thru Shannons zip files and get running

rmackay9 (approved):
I'm not completely sure that I'm qualified to review this bug it looks ok to me.

Hwurzburg:
thanks...will update wiki by tomorrow...

rmaia3d:
> but this will allow users reading the wiki to avoid the hunt thru Shannons zip files and get running

Yes, this is great! Hunting through those zip files can be a bit dauting at first, so having the files already neatly organized and sorted in the Ardu package will make life a lot easier, specially for new users, or users "new" to HD systems. :+1

Issue #27268, opened 2024-06-07
Relay switch for video transmitter voltage control not working in 4.5 (RADIX2HD)


mluessi:
The RADIX2HD uses the Relay2 functionality to power on/off the voltage regulator for the video transmitter. This was working correctly in 4.4 but no longer works as expected in 4.5.  By default, when no RC channel is mapped to the relay, the video transmitter should be powered on, but it is powered off.  When mapping an RC channel to the relay (e.g. `RC8_OPTION = 34`), the relay can only be toggled once.

The default state of the relay can be fixed by using `RELAY2_DEFAULT = 1`. The target currently has `RELAY_DEFAULT = 1` in its `defaults.param` file and I assume this needs to be changed there. With that changed, the video transmitter voltage turns on when no RC channel is mapped to the relay. The relay can also be controlled from MissionPlanner.

However, controlling the relay using an RC channel still does not work. When using channel 8 ( `RC8_OPTION = 34`) the relay can only toggled once (on to off) using the switch on the RC transmitter. I confirmed that the GPIO pin does not toggle, and this has been confirmed with several flight controllers. So it does not seem to be a hardware problem but a bug or misconfiguration.

The relevant parameters are:

    RELAY2_PIN = 81
    RELAY2_FUNCTION = 1
    RELAY2_DEFAULT = 1

For control with RC channel 8:

    RC8_OPTION = 34

**Version**
not working in 4.5.3, 4.5.2, 4.5.1, working in 4.4.4

**Platform**
[  ] All
[  ] AntennaTracker
[ x] Copter
[ x ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
NA

**Hardware type**
[RADIX2HD](https://ardupilot.org/copter/docs/common-radix2hd.html)

**Logs**
NA

IamPete1:
Can you upload a log of the test please. You can enable disarm logging rather than arming.

mluessi:
@IamPete1 I recorded a log. It booted with RC8 high and then I toggled it a few times every 10 s. I can see RCIN.C8 changing, but not sure what I need to look at to see if the relay should be toggling. 

Log file: [radix2hd_vreg_switch_test_20240608.log](https://drive.google.com/file/d/1F1UZU7_VJyz_WIgOw0S5gvDY4oTiLG2v/view?usp=sharing)

IamPete1:
Your switch is conflicting with the flight mode. ` FLTMODE_CH` is 8.

[...]

PR #25108, opened 2023-09-26
AP_Relay: Refactor to support RELAYx_FUNCTION

WickedShell:
This refactors the AP_Relay library as proposed on the dev call yesterday.

Things that stand out from the refactoring:

- You can now have per pin defaults. This is something I've been desperate for, because a number of our devices need opposite polarity. I've actually handled this in the past with inverters, but it's still frustrating.
- Each relay is hidden as an enable parameter, so you don't see/notice the cost of the extra default/function parameters unless you are using a given relay
- There was a small edge case before where changing a pair of relays one after the other (such as what we have to do in ICE), would result in logging just the first change, this now will catch both of these, but still rate limits logging.
- In order to walk through the parameters, you need to have an update call, but this is part of what made the logging handling above better.
- If a relay has a key feature assigned to it (say in this case ignition), a GCS user can't accidentally send the wrong relay command and turn off their ignition. This is actually a very significant win for my use cases. I have some relay outputs that scripts currently control, and the user has been known to try and toggle them with the relay interface in MP, which has resulted in very erroneous (and semi dangerous output). If we add some scripting functions to the list this entire failure mode disappears.

Downsides:
- If you are using all 6 relays then you will be viewing more parameters then before. As we start to refactor away various libraries to use this interface we should see that reduced.
- Initial setup is a bit more complicated because changing the function is an enable parameter, necessitating a param download.

Things missing as currently presented:
- [x] Parameter conversion
- [x] Need to correct that the params currently show up as `RELAY_X_FUNCTION` and the like. That's straightforward.

I haven't done the work to do parameter conversion yet, I wanted to present a prototype that is working with a CubeOrangePlus on the bench, with a oscilloscope hooked up to the outputs. Pin states were toggled using both the `relay` and `engine` commands in `MAVProxy`.

PR #27286, opened 2024-06-12
AP_HAL_ChibiOS: update relay parameter names

IamPete1:
This updates the param names in `defaults.parm` files. This was forgotten in https://github.com/ArduPilot/ardupilot/pull/25108.

There is still some change in behavior. The old relay default parameter covered all relays as we now have one param per relay I have only converted for those relays which also have there pins defined by default. So a user setup relay does not get the old default. We now support upto 16 relays (6 by default), so the full conversion would need the default param for each one.

Closes https://github.com/ArduPilot/ardupilot/issues/27268


andyp1per (approved):

libraries/AP_HAL_ChibiOS/hwdef/KakuteH7Mini/defaults.parm
```diff
@@ -1,7 +1,7 @@
 # setup for LEDs on chan9
 SERVO9_FUNCTION 120
 # Default VTX power to on
-RELAY_DEFAULT 1
+RELAY2_DEFAULT 1
```
amilcarlucas, line 4: Shouldn't this be RELAY1_DEFAULT ?

IamPete1: It uses relay 2. 

https://github.com/ArduPilot/ardupilot/blob/1d945cc5ac8b6555e0c86500653b74fb41106111/libraries/AP_HAL_ChibiOS/hwdef/KakuteH7Mini/hwdef.dat#L75-L77

It seems to be a pattern for lots of the boards, I guess to leave relay 1 free for user stuff.

mluessi:
@IamPete1 thanks! Looks good to me.

PR #27287, opened 2024-06-12
AP_Scripting: Adds I2C transfer() bindings and an example 

IanBurwell:
Adds LUA bindings for the I2CDevice `transfer()` function. This is often needed over `write_register()` as I2C devices might not have 8 bit registers or desire some other custom transfer. 

An example script is also added that allows shutdown of a BQ40Z bms chip which would not be possible with LUA without the `transfer()` function. This script has functionality similar to a sibling PR that implements such functionality more generally in the AP_BattMonitor domain.

### DIFF
--- a/libraries/AP_Scripting/examples/BQ40Z_bms_shutdown.lua
+++ b/libraries/AP_Scripting/examples/BQ40Z_bms_shutdown.lua
@@ -1,0 +1,108 @@
+-- TI BQ40Z BMS shutdown script
+
+local mavlink_msgs = require("MAVLink/mavlink_msgs")
+
+local COMMAND_ACK_ID = mavlink_msgs.get_msgid("COMMAND_ACK")
+local COMMAND_LONG_ID = mavlink_msgs.get_msgid("COMMAND_LONG")
+local msg_map = {}
+msg_map[COMMAND_ACK_ID] = "COMMAND_ACK"
+msg_map[COMMAND_LONG_ID] = "COMMAND_LONG"
+
+local PARAM_TABLE_KEY = 51
+local PARAM_TABLE_PREFIX = "BATT_BQ40Z_"
+-- add a parameter and bind it to a variable
+local function bind_add_param(name, idx, default_value)
+    assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), string.format('could not add param %s', name))
+    return Parameter(PARAM_TABLE_PREFIX .. name)
+end
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 2), 'could not add param table')
+--[[
+  // @Param: BATT_BQ40Z_BUS
+  // @DisplayName: Bus number for the BQ40Z
+  // @Description: Bus number for the BQ40Z
+  // @Range: 0 3
+  // @User: Standard
+--]]
+local BATT_BQ40Z_BUS = bind_add_param('BUS', 1, 0)
+
+
+local MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN = 246
+local MAV_RESULT_ACCEPTED = 0
+local MAV_RESULT_FAILED = 4
+
+-- Initialize MAVLink rx with number of messages, and buffer depth
+mavlink:init(1, 10)
+-- Register message id to receive
+mavlink:register_rx_msgid(COMMAND_LONG_ID)
+-- Block MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN so we can handle the ACK
+mavlink:block_command(MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN)
+
+-- Init BMS i2c device
+local i2c_addr = BATT_BQ40Z_BUS:get()
+assert(i2c_addr ~= nil, "BATT_BQ40Z_BUS not retrievable")
+local bms = i2c:get_device(math.floor(i2c_addr), 0x0B)
+
+-- Exit emergency shutdown (for BQ40Z60, twice for redundancy)
+bms:transfer("\x00\xA7\x23", 0)
+bms:transfer("\x00\xA7\x23", 0)
+
+-- Function that is returned to the AP scheduler when we want to shutdown
+local function shutdown_loop()
+    local ret = bms:transfer("\x00\x10\x00", 0)
+    if ret == nil then
+        gcs:send_text(0, "BQ40Z shutdown transfer failed")
+    end
+
+    return shutdown_loop, 500
+end
+
+-- Main loop
+local function update()
+    local msg, chan = mavlink:receive_chan()
+    local parsed_msg = nil
+
+    if (msg ~= nil) then
+        parsed_msg = mavlink_msgs.decode(msg, msg_map)
+    end
+
+    if parsed_msg ~= nil and (parsed_msg.msgid == COMMAND_LONG_ID) and (parsed_msg.command == MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN) then
+        -- Prepare ack
+        local ack = {}
+        ack.command = parsed_msg.command
+        ack.result = MAV_RESULT_ACCEPTED
+        ack.progress = 0
+        ack.result_param2 = 0
+        ack.target_system = parsed_msg.sysid
+        ack.target_component = parsed_msg.compid
+
+        -- Don't shutdown if armed
+        if arming:is_armed() and parsed_msg.param1 == 2 then
+            gcs:send_text(1, "Not sutting down BQ40Z as vehicle is armed")
+            ack.result = MAV_RESULT_FAILED
+            mavlink:send_chan(chan, mavlink_msgs.encode("COMMAND_ACK", ack))
+
+        -- Shutdown
+        elseif parsed_msg.param1 == 2 then
+            gcs:send_text(1, "Shutting down BQ40Z!!!")
+            mavlink:send_chan(chan, mavlink_msgs.encode("COMMAND_ACK", ack))
+            return shutdown_loop, 0
+
+        -- Pass through the command if it isn't requesting shutdown
+        else
+            local command_int = {
+                p1 = parsed_msg.param1,
+                p2 = parsed_msg.param2,
+                p3 = parsed_msg.param3,
+                p4 = parsed_msg.param4,
+            }
+            local result = gcs:run_command_int(MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN, command_int)
+            ack.result = result
+            mavlink:send_chan(chan, mavlink_msgs.encode("COMMAND_ACK", ack))
+        end
+    end
+
+    return update, 1000
+end
+
+
+return update, 1000
--- a/libraries/AP_Scripting/examples/RM3100_self_test.lua
+++ b/libraries/AP_Scripting/examples/RM3100_self_test.lua
@@ -1,0 +1,107 @@
+-- Runs the Built-In Self Test on the RM3100 LR circuits
+-- Note COMPASS_DISBLMSK should have the 16th bit set to 1 (RM3100)
+
+-- Init RM3100 on bus 0
+local rm3100 = i2c:get_device(0, 0x20)
+assert(rm3100 ~= nil, "i2c get_device error, cannot run RM3100 self test")
+
+-- Queues a Built-In Self Test
+function queue_test()
+    gcs:send_text(1, "Running RM3100 self test")
+
+    -- Queue a self test by setting BIST register
+    local ret = rm3100:transfer("\x33\x8F", 0)
+    if ret == nil then
+        gcs:send_text(1, "Rm3100 BIST transfer failed")
+        return queue_test, 1000
+    end
+
+    -- Send a POLL request to run a BIST
+    ret = rm3100:transfer("\x00\x70", 0)
+    if ret == nil then
+        gcs:send_text(1, "Rm3100 POLL transfer failed")
+        return queue_test, 1000
+    end
+
+    -- As a measurement takes time, delay a bit by scheduling a different function
+    return read_test, 1000
+end
+
+
+-- Reads back values from a Built-In Self Test
+function read_test()
+    -- Read the BIST results
+    local results_str = rm3100:transfer("\x33", 1)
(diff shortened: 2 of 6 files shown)


IanBurwell:
Sibling PR: #27288

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -578,6 +578,7 @@ ap_object AP_HAL::I2CDevice semaphore-pointer
 ap_object AP_HAL::I2CDevice method set_retries void uint8_t 0 20
 ap_object AP_HAL::I2CDevice method write_register boolean uint8_t'skip_check uint8_t'skip_check
 ap_object AP_HAL::I2CDevice manual read_registers AP_HAL__I2CDevice_read_registers 2 1
+ap_object AP_HAL::I2CDevice manual transfer AP_HAL__I2CDevice_transfer 2
```
IamPete1, line 581:
```suggestion
ap_object AP_HAL::I2CDevice manual transfer AP_HAL__I2CDevice_transfer 2 1
```

IanBurwell: Just added that +1

IamPete1:
How do you feel about a string vs the tables? The pack and unpack string functions are very powerful for extracting values with the correct format. Its a shame we have the table precedent in the existing read function.

IanBurwell:
I hadn't considered taking/returning a string which would be more memory efficient, and it would totally be satisfying to be able to do something like
```lua
local some_uint32 = string.unpack("<I4", dev:transfer(string.pack("b", 0xAB), 4))
```
where the string.unpack in handling the byte assembly into a uint32.

But there is also a nice simplicity with using tables. If I wanted just one byte back after a multi byte write it seems like a bit much to do: 
```lua
local some_byte = string.unpack("b", dev:transfer(string.pack("c3", 0xAB, 0xCD, 0xEF), 1))
```
vs
```lua
local some_byte = dev:transfer({0xAB, 0xCD, 0xEF}, 1)[1]
```

But the more I think about this the more a string seems to help even if it incurs a bit more boilerplate. I am going to try and throw together something that uses strings instead.

IanBurwell:
@IamPete1 I have added an alternative `transfer_str` binding for comparison and an example that I actually used it to debug with. Do you think we should go with one over the other, or keep both (maybe renaming them slightly)?

libraries/AP_Scripting/examples/BQ40Z_bms_shutdown.lua
```diff
@@ -0,0 +1,74 @@
+-- TI BQ40Z BMS shutdown script
+---@diagnostic disable: need-check-nil
```
IamPete1, line 2: This is cheating. if you post the warning I'm sure we can work out how fix it.

libraries/AP_Scripting/examples/RM3100_self_test.lua
```diff
@@ -0,0 +1,107 @@
+-- Runs the Built-In Self Test on the RM3100 LR circuits
+-- Note COMPASS_DISBLMSK should have the 16th bit set to 1 (RM3100)
+---@diagnostic disable: need-check-nil
```
IamPete1, line 3: Cheating again ;)

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -655,6 +655,79 @@ int AP_HAL__I2CDevice_read_registers(lua_State *L) {
...
+        }
+    }
+    return success;
+}
+
+int AP_HAL__I2CDevice_transfer_str(lua_State *L) {
+    const int args = lua_gettop(L);
+    if (args != 3){
+        return luaL_argerror(L, args, "expected 2 arguments");
+    }
```
IamPete1, line 704:
```suggestion
    binding_argcheck(L, 3);
```

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -655,6 +655,79 @@ int AP_HAL__I2CDevice_read_registers(lua_State *L) {
...
+    if (!lua_isstring(L, 2)) {
+        return luaL_argerror(L, 2, "expected a string as the first argument");
+    }
```
IamPete1, line 707:
```suggestion
```
Don't `lua_tolstring` do this for us? I think it will fall over there already.

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -655,6 +655,79 @@ int AP_HAL__I2CDevice_read_registers(lua_State *L) {
...
+
+    AP_HAL::I2CDevice * ud = *check_AP_HAL__I2CDevice(L, 1);
+
+    // Parse string of bytes to send
+    size_t send_len;
+    const uint8_t* send_data = (const uint8_t*)(lua_tolstring(L, 2, &send_len));
+
+    // Parse and setup rx buffer
+    uint32_t rx_len = get_uint32(L, 3, 0, UINT32_MAX);
+    uint8_t rx_data[rx_len];
```
IamPete1, line 717: I know the transfer function takes a uint32, but it looks like we actually use a uint16 internally. Even then we probably don't want to let lua try and allocate 65KB. Maybe just use a uint8?  
```suggestion
    // Parse and setup rx buffer
    uint8_t rx_len = get_uint8_t(L, 3);
    uint8_t rx_data[rx_len];
```

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -655,6 +655,79 @@ int AP_HAL__I2CDevice_read_registers(lua_State *L) {
...
+
+    // Transfer
+    ud->get_semaphore()->take_blocking();
+    const bool success = static_cast<bool>(ud->transfer(send_data, send_len, rx_data, rx_len));
```
IamPete1, line 721:
```suggestion
    const bool success = ud->transfer(send_data, send_len, rx_data, rx_len);
```
I'm not sure why we used a case for the `read_registers` method, we really don't need it.

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -655,6 +655,79 @@ int AP_HAL__I2CDevice_read_registers(lua_State *L) {
...
+    ud->get_semaphore()->give();
+
+    // Return a string
+    if (success) {
+        lua_pushlstring(L, (const char *)rx_data, rx_len);
+    }
+    return success;
```
IamPete1, line 728:
```suggestion
    if (!success) {
        return 0;
    }
    
    lua_pushlstring(L, (const char *)rx_data, rx_len);
    return 1;
```
Bad examples in `read_registers` again ;). I think a early return is easier to follow. But I'll leave it to you to pick.

IamPete1 (commented):
I have only really reviewed the string version, I don't think we need both, it will just confuse people. Were tending to use strings now for serial and mavlink. I think string is the way to go, but then I have not used either. You have used both, which do you prefer?

libraries/AP_Scripting/examples/BQ40Z_bms_shutdown.lua
IanBurwell: I removed this script as it was for the table-based transfer function. I can adapt it but as it duplicates the functionality in #27288 I just removed it.

libraries/AP_Scripting/examples/RM3100_self_test.lua
IanBurwell: Threw in an actual nil check :+1:. But it makes me wonder when `get_device()` actually returns nil. Every return before the `AP_HAL__I2CDevice` is created is associated with an error which would result in the script erroring out rather than really returning nil right?

libraries/AP_Scripting/lua_bindings.cpp
IanBurwell: I wondered about that cast, but wanted to be safe. Removed :+1:

IanBurwell:
Using them both more I do prefer the string version even though it brings some extra boilerplate with it. `string.unpack` is just too powerful.

libraries/AP_Scripting/examples/RM3100_self_test.lua
IamPete1: Yeah, you are correct. In which case you can remove the nil check and update the docs here:
https://github.com/ArduPilot/ardupilot/blob/62d9b14a65f2276534ae17306e531f67fcc0cdd8/libraries/AP_Scripting/docs/docs.lua#L138

To just be:

```
---@return AP_HAL__I2CDevice_ud
```

IanBurwell:
I have added back the BQ40Z battery shutdown example as per discussion in the July 15th 2024 dev call and suggestion by @tridge. It works as desired on our dreamer quadrotors, and could be a useful example. With it this PR should be ready for merge pending review.

libraries/AP_Scripting/examples/BQ40Z_bms_shutdown.lua
```diff
@@ -0,0 +1,93 @@
+-- TI BQ40Z BMS shutdown script
+---@diagnostic disable: param-type-mismatch
```
IamPete1, line 2: This is cheating, we need to fix the issue not ignore. I think you said that we needed to update the docs for `get_device` https://github.com/ArduPilot/ardupilot/pull/27287#discussion_r1674770170

libraries/AP_Scripting/examples/BQ40Z_bms_shutdown.lua
```diff
@@ -0,0 +1,93 @@
...
+
+            -- Send ack
+            local ack = {}
+            ack.command = parsed_msg.command
+            ack.result = MAV_RESULT_ACCEPTED
+            ack.progress = 0
+            ack.result_param2 = 0
+            ack.target_system = parsed_msg.sysid
+            ack.target_component = parsed_msg.compid
+            mavlink:send_chan(chan, mavlink_msgs.encode("COMMAND_ACK", ack))
```
IamPete1, line 82: I think you'll have to block the command from being used by ap. with `mavlink:block_command(MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN)` Otherwise AP will ACK unsupported before you ack accepted. 

We now have a binding for `gcs:run_command_int` so if you don't see the param 2 you could pass the command on to AP.

IanBurwell: Ah I didn't know we had a binding for running a command int. Yes I will get on that

libraries/AP_Scripting/examples/BQ40Z_bms_shutdown.lua
IanBurwell: I need to look back into it, but I did this as I was grabbing a parameter and the get() function can return nil according to the docs. I'll look into the bindings a bit more and find a less-cheaty way.

libraries/AP_Scripting/examples/BQ40Z_bms_shutdown.lua
```diff
@@ -0,0 +1,93 @@
...
+    local parsed_msg = nil
+
+    if (msg ~= nil) then
+        parsed_msg = mavlink_msgs.decode(msg, msg_map)
+    end
+
+    if parsed_msg ~= nil and (parsed_msg.msgid == COMMAND_LONG_ID) and (parsed_msg.command == MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN) then
+        gcs:send_text(0, "Got MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN")
+
+        if parsed_msg.param1 == 2 then
```
IamPete1, line 71: Might want to check disarmed armed too.

IamPete1 (approved):
LGTM, Thanks.

libraries/AP_Scripting/examples/RM3100_self_test.lua
```diff
@@ -0,0 +1,107 @@
+-- Runs the Built-In Self Test on the RM3100 LR circuits
+-- Note COMPASS_DISBLMSK should have the 16th bit set to 1 (RM3100)
+
+-- Init RM3100 on bus 0
+local rm3100 = i2c:get_device(0, 0x20)
+assert(rm3100 ~= nil, "i2c get_device error, cannot run RM3100 self test")
+
+-- Queues a Built-In Self Test
+function queue_test()
+    gcs:send_text('1', "Running RM3100 self test")
```
tpwrules, line 10:
```suggestion
    gcs:send_text(1, "Running RM3100 self test")
```

Same in other places.

libraries/AP_Scripting/examples/RM3100_self_test.lua
```diff
@@ -0,0 +1,107 @@
...
+    return read_test, 1000
+end
+
+
+-- Reads back values from a Built-In Self Test
+function read_test()
+    -- Read the BIST results
+    local results_str = rm3100:transfer(string.pack("B", 0x33), 1)
+    if results_str ~= nil then
+        local results = string.unpack("B", results_str)
```
tpwrules, line 36: Could be simply `results_str:byte()`.

tpwrules (commented):
Mostly some style tweaks but I am concerned about the change to `gcs:run_command_int`, `docs.lua` at least needs to be updated.

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -1178,10 +1205,8 @@ int lua_GCS_command_int(lua_State *L)
 
     auto result = _gcs->lua_command_int_packet(pkt);
 
-    // map MAV_RESULT to a boolean
-    bool ok = result == MAV_RESULT_ACCEPTED;
-
-    lua_pushboolean(L, ok);
+    // Return the resulting MAV_RESULT
+    lua_pushinteger(L, result);
```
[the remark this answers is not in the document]
IanBurwell: Somehow that escaped me, thank you for the catch. I think it is not too out of scope of this PR and is required by it. I can split it into a different commit if you think that is justified.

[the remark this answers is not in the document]
tpwrules: Yes, it definitely should be a separate commit at least.

I recognize it's necessary for the PR but it is technically backwards incompatible. It doesn't look like any examples currently check the return values though and it hasn't made it to a stable release so I'd be okay with it going in.

PR #27288, opened 2024-06-12
AP_BattMonitor: Adds smart battery shutdown framework and TI BQ battery monitor

IanBurwell:
Adds ability to command shutdown of a smart-battery system via `PREFLIGHT_REBOOT_SHUTDOWN`, with an example BattMonitor backend that supports being commanded to poweroff.
We use the TI BQ40Z60 and BQ40Z60 chips at GreenSight in our Dreamer line of drones, and use this shutdown functionality to allow full system shutdown from a companion computer.
Note that the TI BQ BattMonitor backend also includes logic to reset the chip's emergency shutdown flag upon successful boot. This is used to ensure the battery stays on after charging is initiated.

Similar functionality was also implemented in a sibling PR that uses LUA but which lacks the broader framework for battery shutdown.


libraries/AP_BattMonitor/AP_BattMonitor_SMBus_TIBQ.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+    uint8_t cmd[] = {0x00, 0x10, 0x00};
+    if (!_dev->transfer(cmd, 3, nullptr, 0)) {
+        GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "Failed to shutdown TIBQ");
+    } else {
+        _state.is_powering_off = true;
+    }
+}
+
+#endif // AP_BATTERY_SMBUS_TIBQ_ENABLED
```
IamPete1, line 48:
```suggestion
#endif // AP_BATTERY_SMBUS_TIBQ_ENABLED

```

libraries/AP_BattMonitor/AP_BattMonitor_SMBus_TIBQ.h
```diff
@@ -0,0 +1,28 @@
...
+
+    // returns true if the battery can be shutdown with shutdown()
+    bool can_shutdown() override { return true; };
+    // shuts the battery down if supported
+    void shutdown() override;
+
+    bool _exit_emshut;
+};
+
+#endif // AP_BATTERY_SMBUS_TIBQ_ENABLED
```
IamPete1, line 28:
```suggestion
#endif // AP_BATTERY_SMBUS_TIBQ_ENABLED

```

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3318,6 +3318,19 @@ MAV_RESULT GCS_MAVLINK::handle_preflight_reboot(const mavlink_command_int_t &pac
         }
     }
 
+#if AP_BATTERY_ENABLED
+    // Check if shutdown is requested
+    if(is_equal(packet.param1, 2.0f)){
```
IamPete1, line 3323:
```suggestion
    if (is_equal(packet.param1, 2.0f)) {
```

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3318,6 +3318,19 @@ MAV_RESULT GCS_MAVLINK::handle_preflight_reboot(const mavlink_command_int_t &pac
...
+        // If the battery monitor in use doesn't support shutdown
+        if(!AP::battery().can_shutdown())
+            return MAV_RESULT_FAILED;
+
+        AP::battery().shutdown();
```
IamPete1, line 3328: Could these be merged? So the shutdown call returns true/false.

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3318,6 +3318,19 @@ MAV_RESULT GCS_MAVLINK::handle_preflight_reboot(const mavlink_command_int_t &pac
         }
     }
 
+#if AP_BATTERY_ENABLED
+    // Check if shutdown is requested
+    if(is_equal(packet.param1, 2.0f)){
+        // If the battery monitor in use doesn't support shutdown
+        if(!AP::battery().can_shutdown())
```
IamPete1, line 3325:
```suggestion
        if (!AP::battery().can_shutdown()) {
        
```

libraries/AP_BattMonitor/AP_BattMonitor_Backend.h
```diff
@@ -85,13 +85,18 @@ class AP_BattMonitor_Backend
...
     void Log_Write_BAT(const uint8_t instance, const uint64_t time_us) const;
     void Log_Write_BCL(const uint8_t instance, const uint64_t time_us) const;
 
     // set desired MPPT powered state (enabled/disabled)
     virtual void mppt_set_powered_state(bool power_on) {};
 
+    // returns true if the battery can be shutdown with shutdown()
+    virtual bool can_shutdown() { return false; };
+    // shuts the battery down if supported
+    virtual void shutdown() {};
```
IamPete1, line 98: Could these be combined?

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1164,6 +1170,61 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
 }
 #endif
 
+// returns true if the battery has shutdown functionality
+bool AP_BattMonitor::can_shutdown(uint8_t instance) const {
```
IamPete1, line 1174:
```suggestion
bool AP_BattMonitor::can_shutdown(uint8_t instance) const
{
```

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1164,6 +1170,61 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+    if (instance >= _num_instances || drivers[instance] == nullptr) {
+        return false;
+    }
+
+    return drivers[instance]->can_shutdown();
+}
+
+// returns true if all connected batteries have shutdown functionality
+bool AP_BattMonitor::can_shutdown() const {
```
IamPete1, line 1183:
```suggestion
bool AP_BattMonitor::can_shutdown() const
{
```

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1164,6 +1170,61 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+    if(_num_instances == 0){
```
IamPete1, line 1184:
```suggestion
    if (_num_instances == 0) {
```

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1164,6 +1170,61 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+        if (configured_type(i) != Type::NONE && !can_shutdown(i)) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+// attempts to shut down a battery (if supported)
+void AP_BattMonitor::shutdown(uint8_t instance){
```
IamPete1, line 1198:
```suggestion
void AP_BattMonitor::shutdown(uint8_t instance)
{
```

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1164,6 +1170,61 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+    if (instance < _num_instances && drivers[instance] != nullptr) {
+        drivers[instance]->shutdown();
+    }
+}
+
+// attempts to shut down all batteries that support doing so
+void AP_BattMonitor::shutdown(){
```
IamPete1, line 1205:
```suggestion
void AP_BattMonitor::shutdown()
{
```

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1164,6 +1170,61 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+    if(_num_instances == 0){
```
IamPete1, line 1206:
```suggestion
    if (_num_instances == 0) {
```

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1164,6 +1170,61 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+        return;
+    }
+
+    for (uint8_t i=0; i< _num_instances; i++) {
+        // skip the primary battery for now
+        if(i == AP_BATT_PRIMARY_INSTANCE) {
+            continue;
+        }
+
+        if (configured_type(i) != Type::NONE) {
```
IamPete1, line 1216: If the check is needed the `shutdown(i);` function should do it.

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1164,6 +1170,61 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+}
+
+// attempts to shut down all batteries that support doing so
+void AP_BattMonitor::shutdown(){
+    if(_num_instances == 0){
+        return;
+    }
+
+    for (uint8_t i=0; i< _num_instances; i++) {
+        // skip the primary battery for now
```
IamPete1, line 1211:
```suggestion
        // Save primary battery for last
```

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1164,6 +1170,61 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+        if(i == AP_BATT_PRIMARY_INSTANCE) {
```
IamPete1, line 1212:
```suggestion
        if (i == AP_BATT_PRIMARY_INSTANCE) {
```

IamPete1 (commented):
I think this should hook into the existing battery power off stuff somehow.

https://github.com/ArduPilot/ardupilot/blob/e83afcf9c43f729d2a203a60779ef576295fb9c8/libraries/AP_BattMonitor/AP_BattMonitor.cpp#L1013-L1014

That is for the shutdown signal coming in from the battery rather than the other way round. But we probably should send that mavlink `MAV_CMD_POWER_OFF_INITIATED` in this case too.

libraries/AP_BattMonitor/AP_BattMonitor_Backend.h
IanBurwell: The reason I separated those is so that we could check if a battery could be shutdown without actually trying to shut it down. Thus, if we for some reason had one battery monitor that supported shutdown and another that didnt, we wouldn't allow shutting down of the system. What are your thoughts there? Maybe a "shutdown" should attempt to shut down everything that supports it and leave the others online?

libraries/AP_BattMonitor/AP_BattMonitor_SMBus_TIBQ.cpp
```diff
@@ -0,0 +1,51 @@
...
+    uint8_t cmd[] = {0x00, 0x10, 0x00};
+    if (!_dev->transfer(cmd, 3, nullptr, 0)) {
+        GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "Failed to shutdown TIBQ");
+        return false;
+    } else {
+        _state.is_powering_off = true;
+    }
+
+    return true;
+}
```
IanBurwell, line 49: @IamPete1 
> But we probably should send that mavlink MAV_CMD_POWER_OFF_INITIATED in this case too.
I have actually already integrated that in the backend when I overwrote the shutdown function via `_state.is_powering_off = true`. This maybe could be done in the front end on a successful shutdown, but then the front end would need access to the back end's `_state` variable or some extra getter function would be needed.

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1176,6 +1182,68 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+        return false;
+    }
+
+    for (uint8_t i=0; i< _num_instances; i++) {
+        // Save primary battery for last
+        if (i == AP_BATT_PRIMARY_INSTANCE) {
+            continue;
+        }
+
+        if (!shutdown(i)){
```
IamPete1, line 1238:
```suggestion
        if (!shutdown(i)) {
```

libraries/AP_BattMonitor/AP_BattMonitor_SMBus_TIBQ.cpp
```diff
@@ -0,0 +1,51 @@
...
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "BQ40Z bms exited shutdown");
+            _exit_emshut = false;
+        }
+    }
+
+    AP_BattMonitor_SMBus_Generic::timer();
+}
+
+
+bool AP_BattMonitor_SMBus_TIBQ::shutdown(){
```
IamPete1, line 36:
```suggestion
bool AP_BattMonitor_SMBus_TIBQ::shutdown() {
```

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3398,6 +3398,18 @@ MAV_RESULT GCS_MAVLINK::handle_preflight_reboot(const mavlink_command_int_t &pac
         }
     }
 
+#if AP_BATTERY_ENABLED
+    // Check if shutdown is requested
+    if (is_equal(packet.param1, 2.0f)){
```
IamPete1, line 3403:
```suggestion
    if (is_equal(packet.param1, 2.0f)) {
```

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3398,6 +3398,18 @@ MAV_RESULT GCS_MAVLINK::handle_preflight_reboot(const mavlink_command_int_t &pac
...
+        // If the battery monitor in use doesn't support shutdown
+        if (!AP::battery().shutdown()){
```
IamPete1, line 3405:
```suggestion
        if (!AP::battery().shutdown()) {
```

IamPete1 (commented):
Is the shutdown functionality unique to the TI BQ? Or is the new driver really just a flag to say that its SMBus battery that can shutdown?

Hwurzburg (changes requested):
need to update https://github.com/ArduPilot/ardupilot/blob/8734aa854d5b359501d40b7c8c33f0196d8f2c29/libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp#L24

IanBurwell:
I believe the functionality to command shutdown is not unique to the BQ, but that the back-end interface is chipset specific. Shutting down a TI BQ40Z chip is initiated via SMBus ManufactureAccess() which is not consistent between different SMBus batteries. The bootup logic to clear the EMSHUT flag on the other hand is a BQ-specific requirement, which is why I didn't include any sort of abstraction in AP_BatteryMonitor for that sequence.

libraries/AP_BattMonitor/AP_BattMonitor_SMBus_TIBQ.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    AP_BattMonitor_SMBus_Generic::timer();
+}
+
+
+bool AP_BattMonitor_SMBus_TIBQ::shutdown() {
+    // Semaphore is needed in case this is called from another thread
+    WITH_SEMAPHORE(_dev->get_semaphore());
+
+    uint8_t cmd[] = {0x00, 0x10, 0x00};
```
peterbarker, line 39:
```suggestion
    static const uint8_t cmd[] {0x00, 0x10, 0x00};
```

libraries/AP_BattMonitor/AP_BattMonitor_SMBus_TIBQ.cpp
```diff
@@ -0,0 +1,50 @@
...
+    if (!_dev->transfer(cmd, 3, nullptr, 0)) {
```
peterbarker, line 40:
```suggestion
    if (!_dev->transfer(cmd, sizeof(cmd), nullptr, 0)) {
```

libraries/AP_BattMonitor/AP_BattMonitor_SMBus_TIBQ.cpp
```diff
@@ -0,0 +1,50 @@
...
+        GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "Failed to shutdown TIBQ");
+        return false;
+    } else {
+        _state.is_powering_off = true;
+    }
```
peterbarker, line 45:
```suggestion
    }
    _state.is_powering_off = true;
```

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3398,6 +3398,18 @@ MAV_RESULT GCS_MAVLINK::handle_preflight_reboot(const mavlink_command_int_t &pac
         }
     }
 
+#if AP_BATTERY_ENABLED
+    // Check if shutdown is requested
+    if (is_equal(packet.param1, 2.0f)) {
```
tridge, line 3403: I don't think this should be in PREFLIGHT_REBOOT_SHUTDOWN
I think we need MAV_CMD_BATTERY_CONTROL command, which takes a bitmask of batterys to apply this to, and which allows for both shutdown and startup

libraries/AP_BattMonitor/AP_BattMonitor_SMBus_TIBQ.cpp
```diff
@@ -0,0 +1,50 @@
...
+    : AP_BattMonitor_SMBus_Generic(mon, mon_state, params)
+{
+    _exit_emshut = true;
+}
+
+
+void AP_BattMonitor_SMBus_TIBQ::timer() {
+    if (_exit_emshut) {
+        // Exit EMERGENCY SHUTDOWN state in case it was engaged on the last poweroff:
+        uint8_t cmd[] = {0x00, 0xA7, 0x23};
```
peterbarker, line 24:
```suggestion
        static const uint8_t cmd[] {0x00, 0xA7, 0x23};
```

libraries/AP_BattMonitor/AP_BattMonitor_SMBus_TIBQ.cpp
```diff
@@ -0,0 +1,50 @@
...
+        if (_dev->transfer(cmd, 3, nullptr, 0)) {
```
peterbarker, line 25:
```suggestion
        if (_dev->transfer(cmd, sizeof(cmd), nullptr, 0)) {
```

libraries/AP_BattMonitor/AP_BattMonitor.cpp
```diff
@@ -1176,6 +1182,68 @@ bool AP_BattMonitor::handle_scripting(uint8_t idx, const BattMonitorScript_State
...
+        return true;
+    }
+
+    return drivers[instance]->shutdown();
+}
+
+// Attempts to shut down all batteries that support doing so
+bool AP_BattMonitor::shutdown()
+{
+    if (!can_shutdown() || _num_instances == 0) {
```
peterbarker, line 1228: Shouldn't we be returning "true" for "we can shut down all 0 batteries"?

tridge (changes requested):
please look at how MAV_CMD_BATTERY_RESET is done, this should be similar

peterbarker (changes requested):

tridge:
discussed on dev call, decided the lua approach is better
https://github.com/ArduPilot/ardupilot/pull/27287

IanBurwell:
Closing as it was decided that this would better fit as a script, which is implemented in #27287

PR #27289, opened 2024-06-12
ArduCopter: get MAV_STATE_BOOT on reboot

richaravoil:
this change add a step in the booting process to let know a companion computer if the autopilot has rebooted

### DIFF
--- a/ArduCopter/GCS_Mavlink.cpp
+++ b/ArduCopter/GCS_Mavlink.cpp
@@ -89,6 +89,10 @@ MAV_STATE GCS_MAVLINK_Copter::vehicle_system_status() const
         return MAV_STATE_STANDBY;
     }
 
+    if (!copter.ap.initialised) {
+    	return MAV_STATE_BOOT;
+    }
+
     return MAV_STATE_ACTIVE;
 }
 


ArduCopter/GCS_Mavlink.cpp
```diff
@@ -89,6 +89,10 @@ MAV_STATE GCS_MAVLINK_Copter::vehicle_system_status() const
         return MAV_STATE_STANDBY;
     }
 
+    if(!copter.ap.initialised){
```
IamPete1, line 92:
```suggestion
    if (!copter.ap.initialised) {
```

IamPete1:
Do any heartbeats make it out before we have initialised?

khancyr:
it is possible, but no guaranty as it relies on a delay() happening during init to send the heartbeat message.

Otherwise you use the STAT_BOOTCNT parameter, that is what we are using on autotest to detect reboot.

Dronotique:
> it is possible, but no guaranty as it relies on a delay() happening during init to send the heartbeat message.
> 
> Otherwise you use the STAT_BOOTCNT parameter, that is what we are using on autotest to detect reboot.

Hello,

With this code I receive 4 heartbeat Mavlink message with a status of MAV_STATE_BOOT before the status MAV_STATE_STANDBY

Whithout this code, the 4 firsts heartbeats received are with a status of  MAV_STATE_ACTIVE before going back to MAV_STATE_STANDBY

Good idea to use also STAT_BOOTCNT to check for a reboot, but I think this PR code is alos a good thing.

khancyr:
yep, not a bad change but we couldn't trust this to get the reboot state as it rely on the system to have delay to send message during init. If you check on default SITL, you will have 2 heartbeats : on coming from the the long baro calibration from a specific sensor, the second from the gyro.

Staying in MAV_STATE_BOOT until we finishing truely booting is indeed a good thing. May worth upgrading the PR to do the same on all vehicles if you got some time.

Dronotique:
It seems this function is managed in the same way for Copter, Sub and Blimp.
But Rover and Plane are managed differently with an Initialisation mode and sendind heartbeat in calibrating state until this flag goes true.

We can implement this PR for Blimp and Sub, but we have no way to test

khancyr:
The initial mode doesn't matter. Initialisation  mode is just a mode that doesn't need anything to enter and don't permit to do anything movement on rover and plane, we just failed to implement it in copter in the past. We can do another PR for other vehicle

peterbarker:
Good stuff.

This is how you can test for Blimp and Sub: https://github.com/ArduPilot/ardupilot/compare/master...peterbarker:ardupilot:pr/boot-state-test?expand=1

(moving that code into the common vehicle base class and adding an entry in their test list).

Have you checked how MissionPlanner reacts to those 4 heartbeats in `MAV_STATE_BOOT`?  I doubt it matters, but it would be bad if these PRs caused random error messages to pop up.

rmackay9 (commented):
OK, I guess it's ok.

Dronotique:
> Have you checked how MissionPlanner reacts to those 4 heartbeats in `MAV_STATE_BOOT`? I doubt it matters, but it would be bad if these PRs caused random error messages to pop up.

I don't notice any reaction in Mission Planner.

> OK, I guess it's ok. I worry slightly that we are making the vehicles inconsistent.
We implemented this change in Sub and Blimp with those PRs : 
https://github.com/ArduPilot/ardupilot/pull/27304
https://github.com/ArduPilot/ardupilot/pull/27303

khancyr:
@rmackay9 I got this in loop, I will do the PR for other vehicle and autotest

PR #27294, opened 2024-06-13
scripts: do not correct IMU values if TCAL is running

andyp1per:
The tempcal script incorrectly tries to correct IMU values when TCAL is running, but these values are unchanged when tcal is running


amilcarlucas:
I will need to fix this in the methodic configurator as well. It uses a linted copy of this code

amilcarlucas:
@tridge this is the original PR from Andy that we spoke of last week.

Tools/scripts/tempcal_IMU.py
amilcarlucas, line 1: @andyp1per Tridge thinks that the correct fix is to either set false on line 308 or 312.

PR #27298, opened 2024-06-14
Param: added arming check for eeprom full

tridge:
this allows users to know if their params won't persist, which is dangerous for a common
preflight reboot

### DIFF
--- a/libraries/AP_Arming/AP_Arming.cpp
+++ b/libraries/AP_Arming/AP_Arming.cpp
@@ -1046,6 +1046,11 @@ bool AP_Arming::system_checks(bool report)
             return false;
         }
 
+        if (AP_Param::get_eeprom_full()) {
+            check_failed(ARMING_CHECK_PARAMETERS, report, "parameter storage full");
+            return false;
+        }
+        
         // check main loop rate is at least 90% of expected value
         const float actual_loop_rate = AP::scheduler().get_filtered_loop_rate_hz();
         const uint16_t expected_loop_rate = AP::scheduler().get_loop_rate_hz();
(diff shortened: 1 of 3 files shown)


rmackay9:
It's great to see this.  I guess it's been tested on systems with no eeprom?  Don't we have some small number of boards that store parameter values to flash?

peterbarker (approved):

peterbarker:
> It's great to see this. I guess it's been tested on systems with no eeprom? Don't we have some small number of boards that store parameter values to flash?

Most of our supported boards use flash for parameter storage :-)

It's usually the "premium" boards which feature a RAMTRON device.

This PR has only been tested in SITL (simulated RAMTRON) and on CubeOrange, so far.

flash-page vs RAMTRON is hidden behind the StorageManager layer, so if we have bugs around EEPROM being full when using flash pages for storage they will probably manifest differently to what we saw here.

tridge:
@rmackay9 all boards have "eeprom" meaning storage, we can't do params without it, this is not specific to the storage type

tridge:
this has been tested on a CubeOrange that had full storage, parameter sets were failing

IamPete1:
Can we tell how close it is? It might be worth giving some small margin, 10 params or something. I think this currently only reports when you actually try and write the parameter which pushes it over the edge?

What should the user do to fix?

tridge:
@IamPete1 we can tell how close. We log it in every log: ```
Param space used: 5113/5120
```

> I think this currently only reports when you actually try and write the parameter which pushes it over the edge?

right.  With this PR they will then get an arming failure. I don't think we should prevent arming earlier than that - users would be rightly annoyed if we had space and couldn't fly

> What should the user do to fix?

backup params, set FORMAT_VERSION=0 to wipe params then reload saved params

rmackay9:
We should add a new item to this wiki page https://ardupilot.org/copter/docs/common-prearm-safety-checks.html

PR #27300, opened 2024-06-14
AP_Logger: tidy running of LoggerMessageWriter sub-writers

peterbarker:
... use the fact these all have a common base class to reduce effectively duplicate code.

Also un-virtualise and const a function which wasn't overridden and can be const

Tested in SITL by creating fence, mission and rally and ensuring they still get to the log.

```
CubeOrange-periph-heavy  16                                                                    
Durandal                            -64    *           -64     -64               -64    -64    -64
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     -88    *           -88     -88               -96    -88    -96
MatekF405                           -96    *           -96     -88               -96    -96    -88
Pixhawk1-1M-bdshot                  -88                -96     -96               -88    -88    -96
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           -88    *           -96     -88               -88    -96    -88
skyviper-v2450                                         -88                                     
```

Some more saving to be made by making `PARM`, `UNIT`, `MULTI` and `FMTU` all subwriters.

### DIFF
--- a/libraries/AP_Logger/LoggerMessageWriter.cpp
+++ b/libraries/AP_Logger/LoggerMessageWriter.cpp
@@ -177,40 +177,30 @@ void LoggerMessageWriter_DFLogStart::process()
         stage = Stage::RUNNING_SUBWRITERS;
         FALLTHROUGH;
 
-    case Stage::RUNNING_SUBWRITERS:
-        if (!_writesysinfo.finished()) {
-            _writesysinfo.process();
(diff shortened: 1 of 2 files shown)


peterbarker:
.... I'll also note that we can probably declare and persist the list of pointers somewhere else, this was just convenience.

khancyr (approved):
LGTM, no mistkaes there

PR #27305, opened 2024-06-14
AP_Mount: Alexmos: increase precision on parsed angle

MattKear:
Noticed this issue when helping someone out.  We are using integer math when calculating the angle in degrees from the provided value.  This minor rework helps to maintaining a little more accuracy so that we don't round to integer degrees.

I have also done a little tidying by moving the defines to the .cpp.

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Alexmos.cpp
+++ b/libraries/AP_Mount/AP_Mount_Alexmos.cpp
@@ -3,6 +3,57 @@
 #if HAL_MOUNT_ALEXMOS_ENABLED
 #include <AP_SerialManager/AP_SerialManager.h>
 
+//definition of the commands id for the Alexmos Serial Protocol
+#define CMD_READ_PARAMS 'R'
+#define CMD_WRITE_PARAMS 'W'
+#define CMD_REALTIME_DATA 'D'
+#define CMD_BOARD_INFO 'V'
+#define CMD_CALIB_ACC 'A'
+#define CMD_CALIB_GYRO 'g'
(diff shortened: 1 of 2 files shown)


libraries/AP_Mount/AP_Mount_Alexmos.cpp
```diff
@@ -3,6 +3,57 @@
...
+#define AP_MOUNT_ALEXMOS_MODE_NO_CONTROL 0
+#define AP_MOUNT_ALEXMOS_MODE_SPEED 1
+#define AP_MOUNT_ALEXMOS_MODE_ANGLE 2
+#define AP_MOUNT_ALEXMOS_MODE_SPEED_ANGLE 3
+#define AP_MOUNT_ALEXMOS_MODE_RC 4
+
+#define AP_MOUNT_ALEXMOS_SPEED 30 // deg/s
+
+#define INT14_DEGREES = (360.0f / 0x3FFF) // 1 full rotation in degrees over 14 bit range
+#define VALUE_TO_DEGREE(d) (float(d)*INT14_DEGREES)
```
MattKear, line 53: This is the key bit that has changed to improve the angle precision

rmackay9 (approved):

rmackay9:
This looks OK to me and it has been tested of course?

MattKear:
A client is testing this for me.  I will report back once the testing has shown the improvement works.

IamPete1:
Before we get binned value, although not exactly integers:

image

After we get much smoother:
image

PR #27306, opened 2024-06-14
AP_RPM: Demystifying RPM sensors 

MattKear:
Lots of users struggle with setting up interrupt-based RPM sensors.  (see here for a number of users asking for help: https://discuss.ardupilot.org/t/rpm-sensor-wiki-knowledge-share/45091/83).

In an attempt to demystify RPM sensors, this PR improves the logging:
- Move to instance based logging.
- Always log the rpm reading regardless of quality.
- Log the calculated quality value.
- Log the bool so that we can clearly see when AP reports the sensor as healthy.

It is my hope that adding this information will help users to figure out what is going wrong.

**Example from Testing:**  

Tested IRL on Heli.
Thankfully I had a dodgy RPM sensor for testing 😬
So now, by always logging raw RPM I can see that something is not right in the measurement and I can see that the corresponding quality is measure is on the floor.
image

Here is an example of the 2nd instance working as well:
image

### DIFF
--- a/libraries/AP_RPM/AP_RPM.cpp
+++ b/libraries/AP_RPM/AP_RPM.cpp
@@ -214,9 +214,7 @@ void AP_RPM::update(void)
     }
 
 #if HAL_LOGGING_ENABLED
-    if (enabled(0) || enabled(1)) {
-        Log_RPM();
-    }
+    Log_RPM();
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


libraries/AP_RPM/AP_RPM.cpp
```diff
@@ -264,6 +264,11 @@ bool AP_RPM::get_rpm(uint8_t instance, float &rpm_value) const
     return true;
 }
 
+float AP_RPM::get_rpm_no_check(uint8_t instance) const
+{
+    return state[instance].rate_rpm;
+}
+
```
IamPete1, line 271: Do we need the function for this? At least it should be private, this really should not be used anywhere except logging.

libraries/AP_RPM/AP_RPM.cpp
```diff
@@ -295,18 +300,26 @@ bool AP_RPM::arming_checks(size_t buflen, char *buffer) const
...
-    const struct log_RPM pkt{
-        LOG_PACKET_HEADER_INIT(LOG_RPM_MSG),
-        time_us     : AP_HAL::micros64(),
-        rpm1        : rpm1,
-        rpm2        : rpm2
-    };
-    AP::logger().WriteBlock(&pkt, sizeof(pkt));
+    // update logging for each instance
+    for (uint8_t i=0; i<num_instances; i++) {
+        if (drivers[i] == nullptr) {
```
IamPete1, line 305:
```suggestion
        if ((drivers[i] == nullptr) || !enabled(i)) {
```
Then you can remove the check on 217.

libraries/AP_RPM/AP_RPM.cpp
```diff
@@ -295,18 +300,26 @@ bool AP_RPM::arming_checks(size_t buflen, char *buffer) const
...
+            // don't log unused instances
+            continue;
+        }
+        float rpm = get_rpm_no_check(i);
+        float qual = get_signal_quality(i);
+        bool h = healthy(i);
```
IamPete1, line 311: Might as well put these straight into the structure. (or leave them here and make them const)

IamPete1 (commented):
Nice, we have needed this for ages.

MattKear:
This has now been flight tested on heli:
More dodgy RPM sensors 😆 
image 
image

tridge (approved):

libraries/AP_RPM/AP_RPM.cpp
```diff
@@ -295,18 +293,23 @@ bool AP_RPM::arming_checks(size_t buflen, char *buffer) const
 #if HAL_LOGGING_ENABLED
 void AP_RPM::Log_RPM() const
 {
-    float rpm1 = -1, rpm2 = -1;
-
-    get_rpm(0, rpm1);
-    get_rpm(1, rpm2);
+    // update logging for each instance
+    for (uint8_t i=0; i<num_instances; i++) {
+        if (drivers[i] == nullptr || enabled(i)) {
```
IamPete1, line 298:
```suggestion
        if (drivers[i] == nullptr || !enabled(i)) {
```

MattKear: well spotted.  Thanks.

PR #27307, opened 2024-06-14
AP_Scripting: add `uint64_t` userdata and auto generate `uint32_t` operators.

IamPete1:
This adds a new userdata for uint64 much like the existing one for uint32. This makes big numbers much easier. A small section is added to the test script to make sure there doing the correct thing. There is also support for logging the new type.

Currently you could use it is a argument for a binding, but it would be better to do some more work so that it can be "coerced" so you could pass a normal lua number if you wanted to. We don't have many functions that take a uint64 argument, so I have not added that (it will still work, you just have to be sure to pass the correct type).

This adds a binding to get unix epoc from gps which needs the uint64 type. In the future we might want to add support for `micros64` too.

I have also moved the uint32 operators to automatically generated bindings and increased the possible auto generated operator types.

### DIFF
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -44,6 +44,44 @@ function uint32_t_ud:tofloat() end
 ---@return integer
 function uint32_t_ud:toint() end
 
+---@class (exact) uint64_t_ud
+---@operator add(uint64_t_ud|uint32_t_ud|integer|number): uint64_t_ud
+---@operator sub(uint64_t_ud|uint32_t_ud|integer|number): uint64_t_ud
+---@operator mul(uint64_t_ud|uint32_t_ud|integer|number): uint64_t_ud
+---@operator div(uint64_t_ud|uint32_t_ud|integer|number): uint64_t_ud
+---@operator mod(uint64_t_ud|uint32_t_ud|integer|number): uint64_t_ud
+---@operator band(uint64_t_ud|uint32_t_ud|integer|number): uint64_t_ud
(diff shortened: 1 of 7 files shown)


IamPete1:
Does cost quite a bit of flash....
```
Binary Name      Text [B]         Data [B]     BSS (B)      Total Flash Change [B] (%)      Flash Free After PR (B)
---------------  ---------------  -----------  -----------  ----------------------------  -------------------------
ardusub          1728 (+0.1096%)  0 (0.0000%)  0 (0.0000%)  1728 (+0.1094%)                                  384728
antennatracker   1728 (+0.1301%)  0 (0.0000%)  0 (0.0000%)  1728 (+0.1298%)                                  633476
ardurover        1728 (+0.1055%)  0 (0.0000%)  0 (0.0000%)  1728 (+0.1053%)                                  323412
arduplane        1728 (+0.0969%)  0 (0.0000%)  0 (0.0000%)  1728 (+0.0967%)                                  177072
arducopter-heli  1728 (+0.0965%)  0 (0.0000%)  0 (0.0000%)  1728 (+0.0962%)                                  168752
blimp            1728 (+0.1278%)  0 (0.0000%)  0 (0.0000%)  1728 (+0.1276%)                                  610296
arducopter       1728 (+0.0967%)  0 (0.0000%)  0 (0.0000%)  1728 (+0.0965%)                                  174200
```
I have added ALL the operators for uint64, we could add subset to start with. EG: + - * /

rmackay9 (approved):
I guess the flash cost is a little high but I think it looks good.

PR #27310, opened 2024-06-16
Copter: Move to the processing section to be used

muramura:
I believe it's better to perform processing where it's needed. 
I will move the processing to the necessary location. 
This change should reduce unnecessary access.


peterbarker (approved):
LGTM.

PR #27311, opened 2024-06-16
AP_Scripting: add REPL applet

tpwrules:
Implements a reasonably comfortable REPL accessible over serial entirely as a loadable script. Also accessible using QGroundControl's MAVLink Console feature, if so configured.

Removes never-implemented `SERIAL_CONSOLE_DEV_SHELL` handling and saves a few dozen bytes on Cube Orange.

Tested in SITL and on the bench with a Cube Orange. Was able to run a couple examples through it.

### DIFF
--- a/libraries/AP_Scripting/applets/repl.lua
+++ b/libraries/AP_Scripting/applets/repl.lua
@@ -1,0 +1,358 @@
+-- Interactive REPL (read-evaluate-print-loop) for the Lua scripting engine
+-- accessible over serial, with line editing, history, and output formatting.
+
+-- 0-based index of Scripting protocol port to use, or nil to use MAVLink
+local PORT_IDX = 0
+local MAX_HISTORY = 50 -- number of lines of history to keep (must be >= 1)
+local VERSION = "v1.0" -- version is convenience for the user
+
+local port
+if PORT_IDX == nil then
+  port = require("mavport")
+else
+  port = serial:find_serial(PORT_IDX)
+end
+assert(port, "REPL scripting port not configured")
+
+-- scan through parameters to find our port and grab its baud rate
+do
+  local serial_info = ""
+  local baud = 115200
+  if PORT_IDX ~= nil then
+    local port_num = 0
+    while PORT_IDX >= 0 and port_num <= 9 do
+      local protocol = param:get(("SERIAL%d_PROTOCOL"):format(port_num))
+      port_num = port_num + 1
+      if protocol == 28 then PORT_IDX = PORT_IDX - 1 end
+    end
+    if PORT_IDX == -1 then -- correct port index found
+      port_num = port_num - 1
+      baud = param:get(("SERIAL%d_BAUD"):format(port_num)) or 115200
+      serial_info = (" on SERIAL%d, BAUD=%d"):format(port_num, baud)
+    end
+  end
+  -- if we can't find the right port, the baud probably does not matter
+  -- (e.g. CAN or network port)
+  port:begin(baud)
+
+  gcs:send_text(6, "Lua REPL "..VERSION.." starting"..serial_info)
+end
+
+-- grab things we use from the environment in case the user messes them up
+local string = string
+local table = table
+
+-- declaration of main state variable and functions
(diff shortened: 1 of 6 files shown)


IamPete1 (commented):
Nice! We really need a binding to get the baud rate for a serial port.

tpwrules:
Yeah, it would be nice, but it looked a bit troublesome. I also want coroutines and `writestring`, but I filed all that under future work.

tpwrules:
I've been using this a while and am quite happy with it. Seems to work reliably and has been very useful for my recent scripting PRs.

Added a few features and tweaked some things and would like to see it merged for others to use.

tridge:
looks nice!

libraries/AP_Scripting/modules/mavport.lua
```diff
@@ -0,0 +1,182 @@
...
+    local avail = 0
+    for _, b in ipairs(self._rx_buf) do
+        avail = avail + #b
+    end
+
+    return avail
+end
+
+-- for completeness
+function mavport:set_flow_control(flow_control_setting) -- luacheck: ignore 212
```
IamPete1, line 179:
```suggestion
function mavport:set_flow_control(_)
```
If you use a underscore you can tell it that your ignoring the argument. It might be enough to just prefix with a underscore. I don't recall.

libraries/AP_Scripting/modules/mavport.lua
```diff
@@ -0,0 +1,182 @@
...
+
+local mavlink_msgs = require("MAVLink/mavlink_msgs")
+
+local SERIAL_CONTROL_DATA_LEN = 70
+local SERIAL_CONTROL = require("MAVLink/mavlink_msg_SERIAL_CONTROL")
+local msg_map = {
+    [SERIAL_CONTROL.id] = "SERIAL_CONTROL",
+}
+
+function mavport:begin(baud_rate) -- luacheck: ignore 212
```
IamPete1, line 15: underscore again.

libraries/AP_Scripting/applets/repl.lua
```diff
@@ -0,0 +1,365 @@
...
+          tx_buf = { s:sub(ci) }
+          break
+        end
+      end
+    end
+  end
+end
+
+-- substitute print function for within the REPL that prints to the port
+function print(...) -- luacheck: ignore
```
IamPete1, line 84: What is this warning?

libraries/AP_Scripting/applets/repl.lua
```diff
@@ -0,0 +1,365 @@
...
+  local t = table.pack(...)
+  for i = 1, t.n do
+    writestring(tostring(t[i]))
+    writestring((i ~= t.n) and "\t" or "\r\n")
+  end
+end
+local print = print_real -- luacheck: ignore
```
IamPete1, line 91: What is the warning?

IamPete1 (commented):
Per line lua check ignores are much less offensive that for per file, but it would be nice to not have to ignore anything.

libraries/AP_Scripting/applets/repl.lua
tpwrules: Lua check doesn’t like shadowing a built in. I will add the number but I don’t think there’s a good way around the ignore.

IamPete1: Fair enough, if you add the number and a short comment explaining why its needed I think its OK. I just want to minimize the number of things we have to ignore so we can prevent people copy and pasting there way into bad code, although we still need to fixup lots of the pre-existing examples.

tpwrules: Did find a way to avoid it.

libraries/AP_Scripting/modules/mavport.lua
tpwrules: Apparently on the version I have an underscore prefix is enough, but not so with the one in CI.

IamPete1 (approved):
Scripting stuff looks good, it would be nice not to have to ignore the checks on a few lines, but this is quite a advanced script, so maybe its OK.

PR #27318, opened 2024-06-17
Copter: fix MAV_CMD_CONDITION_YAW with relative angle when WP_YAW_BEH…

### LINKED ISSUE #27317: when WP_YAW_BEHAVIOR = 0,  MAV_CMD_CONDITION_YAW with relative angle cause wrong heading
## Bug report

**Issue details**
When WP_YAW_BEHAVIOR is set to 0 (never change yaw), the copter is in guided mode and receives MAV_CMD_CONDITION_YAW with relative angle (param4 = 1). The copter does not rotate to the angle relative to the current yaw, instead it rotates to the angle relative to heading to the next waypoint.

yaw_issue
In the above log, the copter should rotate from 350 degrees to 0 degrees after receiving MAV_CMD_CONDITION_YAW with 10 degrees clockwise relative, but it rotated to 280 degrees instead.

To reproduce this issue in SITL:
1. param set WP_YAW_BEHAVIOR 0
2. switch to guided mode & take-off
3. fly to any destination
4. when the copter is flying to the destination, long 10 0 1 1

**Version**
master

**Platform**
[  ] All
[  ] AntennaTracker
[X] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Logs**
[yaw_problem.zip](https://github.com/user-attachments/files/15860563/yaw_problem.zip)

chobitsfan:
…AVIOR = 0

fix #27317

It is tested in SITL and ph4-mini quadcopter
[yaw_problem_fix.zip](https://github.com/user-attachments/files/15860604/yaw_problem_fix.zip)

### DIFF
--- a/ArduCopter/autoyaw.cpp
+++ b/ArduCopter/autoyaw.cpp
@@ -111,6 +111,9 @@ void Mode::AutoYaw::set_fixed_yaw(float angle_deg, float turn_rate_ds, int8_t di
 
     // calculate final angle as relative to vehicle heading or absolute
     if (relative_angle) {
+        if (_mode == Mode::HOLD) {
+            _yaw_angle_cd = copter.ahrs.yaw_sensor;
+        }
         _fixed_yaw_offset_cd = angle_deg * 100.0 * (direction >= 0 ? 1.0 : -1.0);
     } else {
         // absolute angle


rmackay9:
Thanks very much @chobitsfan for finding and fixing this issue!

lthall (approved):
Looks good!

PR #27319, opened 2024-06-17
Rover: use enum-class for mission-done behaviours

peterbarker:
I can't help but wonder why we don't support `MAV_CMD_DO_CHANGE_MODE` as a mission item.  All of these options could be done with one of those.

Testing is limited to a quick spin in SITL and checking that the `EndMissionBehaviour` autotest output seems to cover the cases.

### DIFF
--- a/Rover/Parameters.h
+++ b/Rover/Parameters.h
@@ -384,8 +384,10 @@ public:
     // windvane
     AP_WindVane windvane;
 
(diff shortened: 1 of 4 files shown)


peterbarker:
```
Board,rover
CubeOrange,-24
```

rmackay9 (approved):

PR #27321, opened 2024-06-17
AP_Scripting: add an example which denies arming if home not set

peterbarker:
Note sure if this shouldn't do additional checks that the AHRS does when gps is required.  But it's a little tricky given the structure of those prearms...

### DIFF
--- a/libraries/AP_Scripting/examples/require-home-set-to-arm.lua
+++ b/libraries/AP_Scripting/examples/require-home-set-to-arm.lua
@@ -1,0 +1,20 @@
+-- example script for ensuring home is set before allowing arming
+-- useful so users don't take off in stabilize before RTL will work.
+
+---@dxiagnostic disable: param-type-mismatch
+
+auth_id = arming:get_aux_auth_id()
+
+function update ()
+
+    if not ahrs:home_is_set() then
+      arming:set_aux_auth_failed(auth_id, "Home is not set")
+    else
+      arming:set_aux_auth_passed(auth_id)
+    end
+
+    return update, 5000
+
+end
+
+return update()


rmackay9 (approved):

libraries/AP_Scripting/examples/require-home-set-to-arm.lua
```diff
@@ -0,0 +1,20 @@
+-- example script for ensuring home is set before allowing arming
+-- useful so users don't take off in stabilize before RTL will work.
+
+---@dxiagnostic disable: param-type-mismatch
```
IamPete1, line 4: you really should not need this on such a small script.

IamPete1: infact, I think there is a typo anyway, so its not working.

libraries/AP_Scripting/examples/require-home-set-to-arm.lua
```diff
@@ -0,0 +1,20 @@
...
+
+auth_id = arming:get_aux_auth_id()
```
IamPete1, line 6:
```suggestion
local auth_id = assert(arming:get_aux_auth_id(), "failed to get arming auth ID")
```

IamPete1: that should fix the warning, just need to check that auth_id is not nil.

libraries/AP_Scripting/examples/require-home-set-to-arm.lua
peterbarker: > infact, I think there is a typo anyway, so its not working.

I introduced the typo to see if the line was needed :-)

peterbarker:
We now have an option bit in the main firmware to enforce this, so the script is redundant.

I don't think it offers anything more than the existing remote-auth and get-home examples offer, so I'm going to close this one.

PR #27322, opened 2024-06-17
AP_Scripting: allow logging lua boolean with uint8 format `B`

IamPete1:
Currently you cant just log a bool, it must be a integer value to use `B` (uint8_t) format. This now allows bools.

EG this currently you have to do this:
```
local bool = true
logger:write('BOOL', 'boolean', 'B', bool and 1 or 0)
```

With this PR it you can just use the bool:
```
local bool = true
logger:write('BOOL', 'boolean', 'B', bool)
```

### DIFF
--- a/libraries/AP_Scripting/lua_bindings.cpp
+++ b/libraries/AP_Scripting/lua_bindings.cpp
@@ -460,11 +460,17 @@ int AP_Logger_Write(lua_State *L) {
             case 'M': // uint8_t (flight mode)
             case 'B': { // uint8_t
                 int isnum;
-                const lua_Integer tmp1 = lua_tointegerx(L, arg_index, &isnum);
+                lua_Integer tmp1 = lua_tointegerx(L, arg_index, &isnum);
                 if (!isnum || (tmp1 < 0) || (tmp1 > UINT8_MAX)) {
-                    luaM_free(L, buffer);
-                    luaL_argerror(L, arg_index, "argument out of range");
-                    // no return
+                    // Also allow boolean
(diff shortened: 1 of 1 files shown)


IamPete1:
```
ardusub          32 (+0.0020%)  0 (0.0000%)  0 (0.0000%)  32 (+0.0020%)                                    385672
antennatracker   32 (+0.0024%)  0 (0.0000%)  0 (0.0000%)  32 (+0.0024%)                                    634428
ardurover        32 (+0.0020%)  0 (0.0000%)  0 (0.0000%)  32 (+0.0019%)                                    324380
arduplane        32 (+0.0018%)  0 (0.0000%)  0 (0.0000%)  32 (+0.0018%)                                    178024
arducopter-heli  32 (+0.0018%)  0 (0.0000%)  0 (0.0000%)  32 (+0.0018%)                                    169696
blimp            32 (+0.0024%)  0 (0.0000%)  0 (0.0000%)  32 (+0.0024%)                                    611240
arducopter       32 (+0.0018%)  0 (0.0000%)  0 (0.0000%)  32 (+0.0018%)                                    175144
```

peterbarker (dismissed):
tridge would really like the logging example to be modified to use this new 'B' specifier

IamPete1:
> tridge would really like the logging example to be modified to use this new 'B' specifier

`B` is not new, its the existing specifier for uint8 its used in lots of places already. The change is that you can pass a lua bool now, before you had to pass a lua number.

tridge (approved):

tridge:
@IamPete1 thanks for clarifying

PR #27324, opened 2024-06-18
board-types.txt: Hakrc initial IDs and reservation 

jpkh:
Adding initial ID's and small reservation block for HAKRC flight controllers.


peterbarker:
I've corrected the commit list and commit message and marked as MergeOnCIPass

peterbarker:
I've corrected the commit list again and force-pushed.  Please don't create merge commits.

PR #27325, opened 2024-06-18
ArduPlane: Add help text for QPOS log message

shancock884:
This PR adds the missing help text for the QPOS log message in quadplane.cpp.

I have checked the generated output files (xml, rst, etc) after running Tools/autotest/logger_metadata/parse.py, and the output looks correctly parsed, including enum values for QPOS_xxx states.


IamPete1 (approved):
Looks good, Thanks

peterbarker (approved):

PR #27329, opened 2024-06-18
AP_Math: added comments and a test for euler ordering

tridge:
our main euler functions did not have a comment on the ordering convention

### DIFF
--- a/libraries/AP_Math/matrix3.h
+++ b/libraries/AP_Math/matrix3.h
@@ -236,7 +236,7 @@ public:
     }
 
     /*
-      create a rotation matrix from Euler angles in 321 euler orderin
+      create a rotation matrix from Euler angles in 321 euler ordering
     */
     void        from_euler(T roll, T pitch, T yaw);
 
--- a/libraries/AP_Math/matrix3.h
+++ b/libraries/AP_Math/matrix3.h
@@ -235,13 +235,17 @@ public:
         return a.is_nan() || b.is_nan() || c.is_nan();
     }
 
-    // create a rotation matrix from Euler angles
+    /*
+      create a rotation matrix from Euler angles in 321 euler orderin
+    */
     void        from_euler(T roll, T pitch, T yaw);
 
-    // create eulers from a rotation matrix.
-    // roll is from -Pi to Pi
-    // pitch is from -Pi/2 to Pi/2
-    // yaw is from -Pi to Pi
+    /* create eulers from a rotation matrix.
+       roll is from -Pi to Pi
+       pitch is from -Pi/2 to Pi/2
+       yaw is from -Pi to Pi
+       euler order is 321
+    */
     void        to_euler(T *roll, T *pitch, T *yaw) const;
 
     // create matrix from rotation enum
--- a/libraries/AP_Math/matrix3.cpp
+++ b/libraries/AP_Math/matrix3.cpp
@@ -21,7 +21,7 @@
 #include "AP_Math.h"
 
 // create a rotation matrix given some euler angles
-// this is based on http://gentlenav.googlecode.com/files/EulerAngles.pdf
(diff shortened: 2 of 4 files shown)


rmackay9 (approved):

libraries/AP_Math/matrix3.h
```diff
@@ -235,13 +235,17 @@ class Matrix3 {
         return a.is_nan() || b.is_nan() || c.is_nan();
     }
 
-    // create a rotation matrix from Euler angles
+    /*
+      create a rotation matrix from Euler angles in 321 euler orderin
```
peterbarker, line 239:
```suggestion
      create a rotation matrix from Euler angles in 321 euler ordering
```

peterbarker (approved):

andyp1per:
Having scratched my head against this for some time it would be great to see the additional clarifying commentary:

- Say whether rotations are intrinsic or extrinsic? (https://github.com/ArduPilot/ardupilot/pull/27335)
- Say what the positive direction is for each axis
- This doc in matrix3.cpp now gives a 404 - http://gentlenav.googlecode.com/files/EulerAngles.pdf

andyp1per (commented):
I think its great as far as it goes, but possible could do with some additional commentary

isso11:
> * positive direction is for each axis
I think so too. In the context of Euler angles and the North-East-Down (NED) coordinates frame:
Yaw (ψ): Positive is clockwise when viewed from above, around the Down (Z) axis.
Pitch (θ): Positive is nose up, around the East (Y) axis.
Roll (𝜙): Positive is right wing down, around the North (X) axis.

andyp1per:
@tridge I manually checked one of these and either euler 321 means something different to what I think it does or these rotations are euler 123. An easy example take ROTATION_ROLL_90_YAW_90, this means roll 90 followed by yaw 90. The code swaps around vector values so that [x, y, z] -> [z, x, y] which indeed is roll 90 followed by yaw 90 it is *not* yaw 90 followed by roll 90. So either the meaning of euler 321 is very weird - which I am prepared to believe - or your documentation is actually wrong.

andyp1per (changes requested):
Need to resolve my very confused understanding of the meaning of euler 321

isso11:
> euler 321

Aren't these rotations basically for transformations between body fixed and local inertial NED frames? If so, then the rotations 321 should, by standard definition, mean  [Yaw (3, around the 3rd axis, Z), Pitch (2, around the 2nd axis, Y), Roll(1, around the 1st axis, X)] in that order.

andyp1per:
> Aren't these rotations basically for transformations between body fixed and local inertial NED frames? If so, then the rotations 321 should, by standard definition, mean [Yaw (3, around the 3rd axis, Z), Pitch (2, around the 2nd axis, Y), Roll(1, around the 1st axis, X)] in that order.

That is my understanding

Georacer:
> That is my understanding

Mine as well.

tridge:
@andyp1per @Georacer see the rotate_ordered() function in this PR

tridge:
> * Say whether rotations are intrinsic or extrinsic?

as demonstrated by the rotate_ordered() function it is intrinsic

tridge:
> * This doc in matrix3.cpp now gives a 404

it is common that URLs like this disappear. If we can find the PDF we could put it in https://github.com/ArduPilot/Datasheets

andyp1per:
@tridge I have pushed a test that fails to prove my point.

libraries/AP_Math/matrix3.h
tridge: changed

tridge:
@andyp1per I've found the original EulerAngles.pdf from Bill Premerlani and added it to out Datasheets repo and updated the link in the code in this PR

peterbarker:
Ping @andyp1per can we merge this one?

andyp1per (approved):

peterbarker:
Merged, thanks!

PR #27330, opened 2024-06-19
Rover: remove redundant setting of optical flow bits

peterbarker:
identical code in the parent class


rmackay9 (approved):
This looks correct to me.  As always, let's do a quick test to confirm that the flags are being set as expected post change.

peterbarker:
> This looks correct to me. As always, let's do a quick test to confirm that the flags are being set as expected post change.

Done

PR #27334, opened 2024-06-19
Copter: add Copter::option_enabled(...) replacing bitops vs copter.g2…

peterbarker:
….flight_options

I've already added a test for an option bit in autotest and played with same in SITL

### DIFF
--- a/ArduCopter/Copter.h
+++ b/ArduCopter/Copter.h
@@ -622,11 +622,15 @@ private:
     };
 
 
-    enum class FlightOptions {
+    enum class FlightOption : uint32_t {
         DISABLE_THRUST_LOSS_CHECK     = (1<<0),   // 1
         DISABLE_YAW_IMBALANCE_WARNING = (1<<1),   // 2
         RELEASE_GRIPPER_ON_THRUST_LOSS = (1<<2),  // 4
     };
+    // returns true if option is enabled for this vehicle
+    bool option_is_enabled(FlightOption option) const {
+        return (g2.flight_options & uint32_t(option)) != 0;
+    }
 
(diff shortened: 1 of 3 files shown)


ArduCopter/Copter.h
```diff
@@ -622,11 +622,13 @@ class Copter : public AP_Vehicle {
     };
 
 
-    enum class FlightOptions {
+    enum class FlightOption {
```
peterbarker, line 625:
```suggestion
    enum class FlightOption : uint32_t {
```

peterbarker:
Now also fixes intermittent new test by fixing the Gripper code to emit messages correctly.

libraries/AP_Gripper/AP_Gripper_Servo.cpp
```diff
@@ -104,8 +104,10 @@ void AP_Gripper_Servo::update_gripper()
     // Check for successful grabbed or released
     if (config.state == AP_Gripper::STATE_GRABBING && has_state_pwm(config.grab_pwm)) {
         config.state = AP_Gripper::STATE_GRABBED;
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Gripper load grabbed");
```
rmackay9, line 107: Does this perhaps result in duplicate messages being output?  I see we're already outputting these messages higher up in the file

peterbarker:
On Wed, 19 Jun 2024, Randy Mackay wrote:

> Does this perhaps result in duplicate messages being output? I see we're already outputting these messages higher up in the file

Only if "release" is called again - and in that case we're supposed to let 
the user know that we've done that, so duplkicates are in the design.

lthall (approved):

PR #27337, opened 2024-06-19
Copter: Clearly show the preprocessor

muramura:
I think the preprocessor is getting mixed in with the processing and is difficult to distinguish. I want to clearly identify the preprocessor. The description of the preprocessor in this definition is different only in this file.


peterbarker (approved):
LGTM.

These have also bugged me.

peterbarker (approved):
LGTM

PR #19890, opened 2022-01-26
Copter: add flight option to release gripper on thrust loss

IamPete1:
Adds a option to release the gripper on thrust loss. Similar to https://github.com/ArduPilot/ardupilot/pull/15069 that relases it on failsafe. 

To match up with the with the option to disable thrust loss check this is in `FLIGHT_OPTIONS`

PR #27339, opened 2024-06-20
Copter: Remove Payload release on thrust loss detection

lthall:
This PR removes a potentially very dangerous "feature" added in this PR #19890.

The previous PR will release the payload any time the aircraft uses more than 90% thrust for a second while descending under control and less than 15 degrees lean angle.

The thrust loss detection is designed to be a sub check of further dependent checks like the motor out test. It is a check that is expected to happen in normal flight.

This feature above looks like a sensible check to enable but is expected to result in random releases of the payload during landings on low battery for example. The aircraft is heavily loaded because it has a payload, the aircraft is descending and it is normal for the industry to specify their aircraft maximum take-off at 80% thrust on a full battery.

Releasing payloads in an unplanned and automated way, over landing zones or on approach is literally a risk of serious or fatal injury to ground crew.

-----------------------------

I have tried to improve the comments on thrust_loss_check() to make it clearer what this check is doing.
We may also be able to improve the output message to the user to be clearer or more appropriate. It is currently "Potential Thrust Loss (%d)" and a motor. However, it does not require a motor lost so it will say motor 1 when no motor is lost I think.

The name is also misleading. Perhaps the name "Thrust Limited Detection" would be better.

We could also reduce the likelihood of this going off by changing the check for decent to make it "less than commanded decent" in alt_hold based modes.

### DIFF
--- a/ArduCopter/crash_check.cpp
+++ b/ArduCopter/crash_check.cpp
@@ -169,12 +169,6 @@ void Copter::thrust_loss_check()
         // enable thrust loss handling
         motors->set_thrust_boost(true);
         // the motors library disables this when it is no longer needed to achieve the commanded output
-
-#if AP_GRIPPER_ENABLED
-        if (copter.option_is_enabled(FlightOption::RELEASE_GRIPPER_ON_THRUST_LOSS)) {
-            gripper.release();
-        }
-#endif
     }
 }
 
--- a/ArduCopter/crash_check.cpp
+++ b/ArduCopter/crash_check.cpp
@@ -96,7 +96,12 @@ void Copter::crash_check()
     }
 }
 
-// check for loss of thrust and trigger thrust boost in motors library
+// thrust_loss_check - returns true when the aircraft detects for 1 second that it is:
+//                      above 90% thrust,
+//                      less than 15 degrees lean angle target,
+//                      less than 30 degrees of attitude error,
+//                      is descending.
+// This check provides a warning that the aircraft is near it's maximum thrust limits and is used trigger thrust boost algorithm in the motors library
 void Copter::thrust_loss_check()
 {
     static uint16_t thrust_loss_counter;  // number of iterations vehicle may have been crashed


geofrancis:
don't you have to enable this option first?

// @Bitmask: 0:Disable thrust loss check, 1:Disable yaw imbalance warning, 2:Release gripper on thrust loss

khancyr (approved):
Agree. "Randomly" dropping payload is a serious concern.

It would need to check if this wasn't included in a recent releases

lthall:
I have talked to Pete about this and the application was for a drone that dragged a line. If the line got snagged the drone would crash.

This RELEASE_GRIPPER_ON_THRUST_LOSS would mean that anytime the aircraft hit maximum thrust and started reducing in altitude it would release.

The interesting thing is that it would only work if the drone was leant over by less than 15 degrees. So that would mean the aircraft would need to be pulled down almost vertically. I wonder how this could be effective at all.

ArduCopter/crash_check.cpp
```diff
@@ -96,7 +96,12 @@ void Copter::crash_check()
     }
 }
 
-// check for loss of thrust and trigger thrust boost in motors library
+// thrust_loss_check - returns true when the aircraft detects for 1 second that it is:
+//                      above 90% thrust,
```
IamPete1, line 100:
```suggestion
//                      above 90% demanded thrust or high thrust saturation,
```

ArduCopter/crash_check.cpp
```diff
@@ -169,12 +174,6 @@ void Copter::thrust_loss_check()
         // enable thrust loss handling
         motors->set_thrust_boost(true);
         // the motors library disables this when it is no longer needed to achieve the commanded output
-
-#if AP_GRIPPER_ENABLED
-        if (copter.option_is_enabled(FlightOption::RELEASE_GRIPPER_ON_THRUST_LOSS)) {
-            gripper.release();
-        }
-#endif
```
IamPete1, line 177: We can't just remove this. The would need to be a arming check if the option is set and the feature is gone.

rmackay9:
I think we could potentially close this PR and instead focus on this updated version https://github.com/ArduPilot/ardupilot/pull/27396

rmackay9:
I'm going to go ahead and close this because @lthall has approved the other PR

PR #27341, opened 2024-06-20
Detune to stop ground oscillation

lthall:
This pr provides the ability to reduce the gains of roll, pitch, and yaw to remove ground oscillation while landed. Each axis has a gain reduction variable that is set when the aircraft is landed. The gain returns to normal in 2 x ATC_INPUT_TC.

This is a common problem on aircraft with flexible landing gear. It is often seen as a yaw oscillation during spool up on more solidly built aircraft also.

### DIFF
--- a/ArduCopter/mode.cpp
+++ b/ArduCopter/mode.cpp
@@ -423,6 +423,7 @@ void Copter::update_flight_mode()
 #if AP_RANGEFINDER_ENABLED
     surface_tracking.invalidate_for_logging();  // invalidate surface tracking alt, flight mode will set to true if used
 #endif
+    attitude_control->landed_gain_reduction(copter.ap.land_complete); // Adjust gains when landed to attenuate ground oscillation
 
     flightmode->run();
 }
--- a/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
+++ b/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
@@ -150,6 +150,27 @@ const AP_Param::GroupInfo AC_AttitudeControl::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("INPUT_TC", 20, AC_AttitudeControl, _input_tc, AC_ATTITUDE_CONTROL_INPUT_TC_DEFAULT),
 
+    // @Param: LAND_R_MULT
+    // @DisplayName: Landed roll gain multiplier
+    // @Description: Roll gain multiplier active when landed. A factor of 1.0 means no reduction in gain while landed. Reduce this factor to reduce ground oscitation in the roll axis. 
+    // @Range: 0.25 1.0
+    // @User: Advanced
+    AP_GROUPINFO("LAND_R_MULT", 21, AC_AttitudeControl, _land_roll_mult, 1.0),
+
+    // @Param: LAND_P_MULT
+    // @DisplayName: Landed pitch gain multiplier
+    // @Description: Pitch gain multiplier active when landed. A factor of 1.0 means no reduction in gain while landed. Reduce this factor to reduce ground oscitation in the pitch axis. 
+    // @Range: 0.25 1.0
+    // @User: Advanced
+    AP_GROUPINFO("LAND_P_MULT", 22, AC_AttitudeControl, _land_pitch_mult, 1.0),
+
+    // @Param: LAND_Y_MULT
+    // @DisplayName: Landed yaw gain multiplier
+    // @Description: Yaw gain multiplier active when landed. A factor of 1.0 means no reduction in gain while landed. Reduce this factor to reduce ground oscitation in the yaw axis. 
+    // @Range: 0.25 1.0
+    // @User: Advanced
+    AP_GROUPINFO("LAND_Y_MULT", 23, AC_AttitudeControl, _land_yaw_mult, 1.0),
+
     AP_GROUPEND
 };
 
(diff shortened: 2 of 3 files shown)


ArduCopter/mode_stabilize.cpp
```diff
@@ -62,6 +62,9 @@ void ModeStabilize::run()
         break;
     }
 
+    // reduce gains when landed;
+    attitude_control->landed_gain_reduction(copter.ap.land_complete);
```
andyp1per, line 66: Only stabilize?

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -150,6 +150,27 @@ const AP_Param::GroupInfo AC_AttitudeControl::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("INPUT_TC", 20, AC_AttitudeControl, _input_tc, AC_ATTITUDE_CONTROL_INPUT_TC_DEFAULT),
 
+    // @Param: LAND_R_MULT
+    // @DisplayName: Roll gain multiplier used when landed
```
andyp1per, line 154: I know a multiplier makes sense mathematically, but no-one is ever going to set it above 1 and so its really a divider. I would prefer a name that conveys to the use a bit more what it does.

MichelleRos: That would be a bit confusing. Unless we also invert the parameter...

lthall: I think people understand that multiplying by a number between 0 and 1 will reduce the gains.

rmackay9:
@lthall I'm sure this will be popular.  Do we need to have both Roll and Pitch parameters?  I wonder if we can imagine a frame where we'd want them to be different.

bnsgeyer:
> @lthall I'm sure this will be popular. Do we need to have both Roll and Pitch parameters? I wonder if we can imagine a frame where we'd want them to be different.

I think for heli there could be significantly different gains between pitch and roll.  But I don't know if the same backoff factor could work for both axes.

ArduCopter/mode_stabilize.cpp
bnsgeyer: @lthall do you intend to have this work in all modes at some point. I'm curious like @andyp1per as to why this is just stabilize

lthall: Yes, this is just an example of how I need to do it right now. I was looking for suggestions on how I could do it better.

lthall:
I considered this but I believe it would be a mistake to not separate them. We don't want to compromise control authority unnecessarily.  There are numerous aircraft that have significantly different ground interaction modes between all three axis.

ArduCopter/mode.cpp
```diff
@@ -420,6 +420,7 @@ bool Copter::set_mode(const uint8_t new_mode, const ModeReason reason)
 void Copter::update_flight_mode()
 {
     surface_tracking.invalidate_for_logging();  // invalidate surface tracking alt, flight mode will set to true if used
+    attitude_control->landed_gain_reduction(copter.ap.land_complete); // Adjust gains when landed to attenuate ground oscillation
```
lthall, line 423: I have called this feature here to capture all flight modes in a single line. It seems like a good fit but I would like to get others feedback.

andyp1per: is land_complete all that you need?

andyp1per (approved):

ArduCopter/mode.cpp
lthall: Yes, this is a reasonably conservative test.

IamPete1 (approved):
Looks good. Just to confirm, it does should up in the `ATSC` log message?

lthall:
@IamPete1 Sorry mate, I didn't understand what you were asking.

IamPete1:
> @IamPete1 Sorry mate, I didn't understand what you were asking.

The `ATSC`  message contains the used values of the angle P scale and the rate PD scale if all ones. The logging should "just work" with your changes, but it would be worth double checking just so we can track what is going on. You should see them slew up and down, Once they get to all ones the log should stop.

lthall:
image

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -204,6 +225,21 @@ void AC_AttitudeControl::reset_rate_controller_I_terms_smoothly()
     get_rate_yaw_pid().relax_integrator(0.0, _dt, AC_ATTITUDE_RATE_RELAX_TC);
 }
 
+// Reduce attitude control gains while landed to stop ground resonance
+void AC_AttitudeControl::landed_gain_reduction(bool landed)
+{
+    // use 2.0 x tc to match the response time to 86% commanded
+    const float spool_step = _dt / (2.0 * _input_tc);
```
rmackay9, line 232: I wonder if we should catch the possible divide by zero here

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -150,6 +150,27 @@ const AP_Param::GroupInfo AC_AttitudeControl::var_info[] = {
...
 
+    // @Param: LAND_R_MULT
+    // @DisplayName: Roll gain multiplier used when landed
+    // @Description: Roll gain multiplier used when landed
+    // @Range: 0 1
+    // @User: Advanced
+    AP_GROUPINFO("LAND_R_MULT", 21, AC_AttitudeControl, _land_roll_mult, 1.0),
+
+    // @Param: LAND_P_MULT
+    // @DisplayName: Pitch gain multiplier used when landed
```
rmackay9, line 161: we should consider putting a Land at the front of the Display name

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -150,6 +150,27 @@ const AP_Param::GroupInfo AC_AttitudeControl::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("INPUT_TC", 20, AC_AttitudeControl, _input_tc, AC_ATTITUDE_CONTROL_INPUT_TC_DEFAULT),
 
+    // @Param: LAND_R_MULT
+    // @DisplayName: Roll gain multiplier used when landed
+    // @Description: Roll gain multiplier used when landed
```
rmackay9, line 155: We should give some instructions on what lower vs higher values do.  E.g. "lower values result in lower gains during landing, higher values result in higher gains during landing"

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -150,6 +150,27 @@ const AP_Param::GroupInfo AC_AttitudeControl::var_info[] = {
...
+    // @Range: 0 1
```
rmackay9, line 156: Is the lower range really 0?

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
lthall: I am not sure if I like what I did here.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
lthall: Maybe 
"Landed pitch gain multiplier"
"Landed pitch gain multiplier, only active when landed" (seems to say it twice though)

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
lthall: Changing to 0.25 to 1.0

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
lthall: Does this work?
    // @Description: Roll gain multiplier used when landed. A factor of 1.0 means no reduction in gain while landed. Reduce this factor to reduce ground oscitation in the roll axis.

rmackay9 (approved):
LGTM, thanks for the changes

PR #27342, opened 2024-06-20
AP_Mount: Siyi gets rate map, PI control, accel limit

rmackay9:
This improves the Siyi gimbal control inspired by the work done in the [MAVProxy Siyi module](https://github.com/ArduPilot/MAVProxy/blob/master/MAVProxy/modules/mavproxy_SIYI):

- Rate mapping table compensates for the non-linear rate response of the gimbal
- Angle controller is enhanced in these ways
  - P controller is extended to a PI controller to reduce long term error
  - Rate targets are acceleration limited to reduce impossibly fast changes and thus reduce overshoot
  - I-term build up is avoided by using limit flags
- Debug logging of angle and rate has been added to somewhat ease support if this doesn't go well.  This debugging is off by default

This has been tested on a Siyi A8 and some screen shots are shown below

Desired vs Actual Pitch Angle
image

Desired vs Actual Yaw Angle
image

Desired vs Actual Pitch Rate
image

Desired vs Actual Yaw Rate
image


rmackay9:
Hi @timtuxworth,

I wonder if you'd be able to help me test this?  I've bench tested with an A8 and I can bench test with a Z10 as well but a flight test is slightly more difficult for me.

No pressure of course but if you can help, just ping me with your autopilot and I'll create a binary.  Discord also works of course.

timtuxworth:
Hi Randy - I currently have the A8 Mini installed on an EDU-650 with a Cube Red. I'd likely be able to test it on the weekend.

rmackay9:
Hi @timtuxworth,

Thanks very much!

Here is a binary for the CubeRedPrimary based on this branch. https://www.dropbox.com/scl/fi/k7f1yxtqa7y50ebtdu256/copter-460-siyiPI-cuberedprimary-27Jun2024.zip?rlkey=ae1mfjr1ybq3n5njpcad43lug&dl=0

I've rebased it on master as well.

timtuxworth:
I tried testing this but found that this firmware seems to break NET_Px_PROTOCOL access to the A8 Mini. I have mine connected to the FC via ethernet. I proved this by flashing back to 4.5.0 beta 2 which worked. 
Unfortunately I just broke my FC by accidentally flashing plane on it while trying to get back to 4.5.4 stable, so now I have some messing around to recover.

rmackay9:
Hi @timtuxworth,

Now that we've resolved the networking issue, I wonder if you'd be willing to test this change again?  I've put a new CubeRedPrimary binary here which is this PR rebased on master
https://www.dropbox.com/scl/fi/l0apmxdhfom4ksen95rx3/copter-460-siyiPI-cuberedprimary-17Jul2024.zip?rlkey=1xajqqvzd0wa6njyajcuki9nk&dl=0

timtuxworth:
Hi @rmackay9  - where does this sit? What would it take to get it into master?

rmackay9:
Hi @timtuxworth,

I think more testing and confirmation that it actually improves performance.  Looking at the logs it does improve performance for the A8 at least but it's difficult to know if users would even notice the improvement.

timtuxworth:
I watched the behavior today and I'd have to say it seems more like a "glitch" than "stuck". Here's a screen recording of QGround Control. It shows an ROI being set about 1:00 into the video. The camera tracks the ROI then at about 1:10 swings back to face forward then swings back to the ROI again. This happens continually through the flight. 

https://www.dropbox.com/scl/fi/6n3rg2k6xqcvfmreqk5mu/QGroundControl-2025-03-11-11-50-36.mp4?rlkey=14oc1ut4enaqco8f3rjqpj9x2&dl=0

rmackay9:
Just for anyone reading this, the issue reported by @timtuxworth above is not related to the changes in this PR and are most likely due to issues in the Siyi camera gimbals themselves

PR #27343, opened 2024-06-20
Add support for docking clamps, including automated Copter release

peterbarker:
This work sponsored by FreeSpace.


ArduCopter/takeoff.cpp
```diff
@@ -218,6 +223,11 @@ void _AutoTakeoff::run()
     if (complete) {
         const Vector3p& _complete_pos = copter.pos_control->get_pos_target_cm();
         complete_pos = Vector3p{_complete_pos.x, _complete_pos.y, pos_z};
+#if AP_TIE_DOWN_CLAMPS_ENABLED
```
peterbarker, line 226: @lthall why do we need this block?  I'm not sure how we get to here without going through the init function which should already have released the clamps.

lthall: This sets the clamps back to the closed position after the take-off has completed. That way the state matches the switch position if it is being used and if it isn't the clamp is ready for reattachment when the aircraft lands again.

peterbarker: I've added a comment, thanks.

peterbarker:
Is the option bit really needed?

libraries/SRV_Channel/SRV_Channel_aux.cpp
```diff
@@ -142,6 +143,9 @@ void SRV_Channel::aux_servo_function_setup(void)
     case k_egg_drop:
         set_range(100);
         break;
+#if AP_TIE_DOWN_CLAMPS_ENABLED
+    case k_tie_down_release:
```
IamPete1, line 147: my preference for new options would be to set the range to 1.0, its a float now so the value doesn't matter. Eventually I would like to move everything to 0 -> 1 or -1 -> 1

IamPete1: you could also call `set_range` from landing gear of course, that would remove the need for any change in this lib.

libraries/SRV_Channel/SRV_Channel.h
```diff
@@ -188,6 +188,7 @@ class SRV_Channel {
         k_rcin14_mapped         = 153,
         k_rcin15_mapped         = 154,
         k_rcin16_mapped         = 155,
+        k_tie_down_release      = 156,
```
IamPete1, line 191: Needs param description.

IamPete1:
This does feel like it might be more of a scripting feature. We do have `vehicle:is_landing()` and `vehicle:is_taking_off()`.

Hwurzburg (changes requested):
needs to update Parameters.cpp also

rmackay9:
I agree with @IamPete1, it's so simple that it seems like it could be handled in scripting.  I suspect the reason that scripting isn't being used because scripts are a bit more difficult to deploy.  If that's the reason then perhaps we should investigate how we could  make deploying scripts easier.

libraries/SRV_Channel/SRV_Channel_aux.cpp
peterbarker: Done and done

libraries/SRV_Channel/SRV_Channel.h
peterbarker: Added

Hwurzburg (changes requested):
still needs FLIGHT_OPTIONS metadata addition

peterbarker:
> still needs FLIGHT_OPTIONS metadata addition

Added metadata.

Hwurzburg (approved):

ArduCopter/Parameters.cpp
```diff
@@ -1000,7 +1000,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Param: FLIGHT_OPTIONS
     // @DisplayName: Flight mode options
     // @Description: Flight mode specific options
-    // @Bitmask: 0:Disable thrust loss check, 1:Disable yaw imbalance warning, 2:Release gripper on thrust loss, 3:Require position for arming
+    // @Bitmask: 0:Disable thrust loss check, 1:Disable yaw imbalance warning, 2:Release gripper on thrust loss, 3:Require position for arming, 4:Enable automated docking clamp release
```
rmackay9, line 1003: I think we could simplify this by removing the FLIGHT_OPTIONS bit and just enable the feature if the user defines the SERVOx_FUNCTION.

ArduCopter/takeoff.cpp
```diff
@@ -54,6 +54,11 @@ void Mode::_TakeOff::start(float alt_cm)
     _running = true;
     take_off_start_alt = copter.pos_control->get_pos_target_z_cm();
     take_off_complete_alt  = take_off_start_alt + alt_cm;
+#if AP_TIE_DOWN_CLAMPS_ENABLED
+    if (copter.option_is_enabled(Copter::FlightOption::TAKEOFF_TIEDOWN_RELEASE)) {
+        copter.landinggear.tie_down_release();
```
rmackay9, line 59: How about moving this outside of landing gear.  While the tie-down is perhaps physically on the landing gear I'm not sure it's too related is it?

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -1283,6 +1289,26 @@ bool RC_Channel::run_aux_function(AUX_FUNC ch_option, AuxSwitchPos pos, AuxFuncT
     return ret;
 }
 
+#if AP_TIE_DOWN_CLAMPS_ENABLED
+void RC_Channel::do_aux_function_TIE_DOWN_RELEASE(const AuxSwitchPos ch_flag)
```
rmackay9, line 1293: How do you feel about putting "TIE_DOWN_RELEASE" in lower case?  It's a little shouty :-)

rmackay9:
Re lua vs C++, I apparently said C++ was fine at some point (I have no memory) and so if it's OK with @IamPete1 and others, I think it's only fair that we allow it to be added to the c++.  I suspect that the flash cost is low and if we did it as a script, I can imagine that adding access to the state takeoff state machine could require as much flash as the C++ equivalent.

ArduCopter/Parameters.cpp
peterbarker: I think we need this option in case the user wants to have docking clamps which they can trigger with an auxiliary switch but don't necessarily want them to *automatically* release.

ie. in this case the servo function defines what it operates, not entirely how it is used.

We could add a second servo function if you want.  We do this in Plane for flaps - we have a channel which controls flaps and a second channel which operates just like the first one except that the flaps also get automatically activated during landing.  I'm not keen on this because if we find another use for the clamps we might end up with a cross-product problem!

ArduCopter/takeoff.cpp
peterbarker: I agree they're only kind-of-sort-of related.

I do want this in libraries so other vehicles can use the functionality - do you have suggestions on where this might go?  Creating a new library just for this seems a little overkill, but I can do that.

lthall (approved):

lthall:
I think this is fundermental capability for multirotors. Given how small this is I think it is worth putting it in C.

libraries/SRV_Channel/SRV_Channel.h
```diff
@@ -210,6 +210,7 @@ class SRV_Channel {
         k_motor31               = 178,
         k_motor32               = 179,
         k_cam_zoom              = 180,
+        k_tie_down_release      = 181,   ///< nominates servos to trigger for releasing tie-down clamps
```
IamPete1, line 213: Needs param docs.

libraries/RC_Channel/RC_Channel.h
```diff
@@ -264,6 +264,8 @@ class RC_Channel {
         AHRS_AUTO_TRIM =     182,  // in-flight AHRS autotrim
         AUTOLAND =           183,  //Fixed Wing AUTOLAND Mode
 
+        TIE_DOWN_RELEASE =   184, // Operate tie down release low=open, middle=auto, high=close
```
IamPete1, line 267: Needs param docs.

IamPete1 (changes requested):

rmackay9:
Looks good to me, let's move the bit to AP_LandingGear's OPTIONS bit.  then I'm happy for it to be merged, thanks!

ArduCopter/Parameters.cpp
```diff
@@ -846,7 +846,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Param: FLIGHT_OPTIONS
     // @DisplayName: Flight mode options
     // @Description: Flight mode specific options
-    // @Bitmask: 0:Disable thrust loss check, 1:Disable yaw imbalance warning, 2:Release gripper on thrust loss
+    // @Bitmask: 0:Disable thrust loss check, 1:Disable yaw imbalance warning, 2:Release gripper on thrust loss, 4:Enable automated docking clamp release
```
IamPete1, line 849: Is there any benefit to the option? Looks like its gated on the servo being assigned in any-case. If we do want the option there is a landing gear option param where this could live.

libraries/AP_LandingGear/AP_LandingGear.cpp
```diff
@@ -339,4 +343,25 @@ void AP_LandingGear::deploy_for_landing()
...
+#if AP_TIE_DOWN_CLAMPS_ENABLED
+// support for tie-down clamps:
+static constexpr auto tie_down_function = SRV_Channel::k_tie_down_release;
+void AP_LandingGear::tie_down_secure()
+{
+    if (!SRV_Channels::function_assigned(tie_down_function)) {
+        return;
+    }
+    GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Secure tie down");
+    SRV_Channels::set_output_scaled(tie_down_function, 1);
```
IamPete1, line 355: My preference would be to always set the servo but only print the message is the function is assigned. This means you could use the scaled value in scripting without having to assign it to a output.

rmackay9:
We discussed this on the dev call and everyone agreed that multiple grippers makes sense.  It's less clear if tie-downs should be part of the regular C++ code or implemented as a lua script.

ArduCopter/Parameters.cpp
LupusTheCanine: Somebody may want servos marked as clamps but controlled by a script or companion computer.

PR #27346, opened 2024-06-20
Rename get_primary_*() to get_first_healthy_*() and record primary instance in inertial sensors

andyp1per:

This PR makes explicit the difference between the "primary gyro" as selected by the EKF and the "first healthy gyro" as selected by the IMU. Both are used in different ways in different places, and sometimes they are the same. The existing naming was far too error prone.

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor.cpp
@@ -1933,13 +1933,25 @@ void AP_InertialSensor::update(void)
         // set primary to first healthy accel and gyro
         for (uint8_t i=0; i<INS_MAX_INSTANCES; i++) {
             if (_gyro_healthy[i] && _use(i)) {
-                _primary_gyro = i;
+                _first_healthy_gyro = i;
+#if AP_AHRS_ENABLED
+                // ask AHRS for the true primary, might just be us though
+                _primary_gyro = AP::ahrs().get_primary_gyro_index();
+#else
+                _primary_gyro = _first_healthy_gyro;
+#endif
                 break;
             }
         }
         for (uint8_t i=0; i<INS_MAX_INSTANCES; i++) {
             if (_accel_healthy[i] && _use(i)) {
-                _primary_accel = i;
+                _first_healthy_accel = i;
+#if AP_AHRS_ENABLED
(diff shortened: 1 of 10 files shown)


peterbarker:
OK, fundamental problem.

The first healthy gyro isn't necessarily in use, so this would be a bad rename.

My apologies, I didn't realise we filtered with the use mask or I would have countenanced against this rename sooner.

So `AP::ins().get_primary()` is actually closer to the truth than `AP::ins().get_first_healthy()`.

BTW, there's a bunch of logic changes involved in here which should be a straight-forward method rename, I'd have thought....

andyp1per:
@peterbarker the whole point is that AP::ins().get_primary() != AP::ahrs().get_primary() that is the source of the problem and I already found one bug where one was being used when it should be the other. I don't care what the names are as long as they are different. AP::ins().get_primary() is a bad name because *inside* the IMU code we are using a "primary" gyro that is the AHRS version - all highly confusing

I think AP::ahrs().get_primary() is the right name since that is the primary gyro for all flight control. So really that just leaves what we call the IMU version.

```get_first_active_gyro()```?
```get_first_healthy_active_gyro()```?

File: libraries/AP_InertialSensor/AP_InertialSensor.cpp [hunk 1]
```diff
@@ -1933,13 +1933,25 @@ void AP_InertialSensor::update(void)
         // set primary to first healthy accel and gyro
         for (uint8_t i=0; i<INS_MAX_INSTANCES; i++) {
             if (_gyro_healthy[i] && _use(i)) {
-                _primary_gyro = i;
+                _first_healthy_gyro = i;
+#if AP_AHRS_ENABLED
+                // ask AHRS for the true primary, might just be us though
+                _primary_gyro = AP::ahrs().get_primary_gyro_index();
```
tridge, line 1939: this could flip-flop between two IMUs, having AP_InertialSensor call into AHRS is dangerous, as AHRS itself depends on AP_InertialSensor, this is a nasty layering violation

tridge (changes requested):

peterbarker:
`get_first_usable`

peterbarker:
Replaced by https://github.com/ArduPilot/ardupilot/pull/27383

PR #27347, opened 2024-06-21
Fix ECC Failure Check in Bootloader and recovery from it 

bugobliterator:
For STM32H7 board the RAM[0] is actually non DMA'ble memory in bootloader. And for doing ECC check we were trying to copy values from Flash to non-DMA'ble memory which will just fail and not raise an error when the ECC fail actually happens.

Edit: moved to adding support for Heap allocation on boards with bootloader flash size > 128K and using malloc_dma to allocate DMA'ble memory


peterbarker:
image

andyp1per (approved):

Tools/AP_Bootloader/support.cpp
```diff
@@ -553,11 +553,7 @@ void check_ecc_errors(void)
             break;
         }
 #endif
-        dmaStartMemCopy(dma,
-                        STM32_DMA_CR_PL(0) | STM32_DMA_CR_PSIZE_BYTE |
-                        STM32_DMA_CR_MSIZE_BYTE,
-                        ofs+(uint8_t*)FLASH_BASE, buf, sizeof(buf));
-        dmaWaitCompletion(dma);
+        memcpy((uint8_t*)buf, (uint8_t*)FLASH_BASE+ofs, sizeof(buf));
```
tridge, line 556: this would completely break the functionality as you'd get a hard-fault. this is absolutely the wrong fix

tridge (changes requested):

tridge:
@bugobliterator you need to discuss with me

Tools/AP_Bootloader/support.cpp
```diff
@@ -538,11 +538,23 @@ void port_setbaud(uint32_t baudrate)
...
  */
+#define ECC_CHECK_CHUNK_SIZE (32*sizeof(uint32_t))
 void check_ecc_errors(void)
 {
     __disable_fault_irq();
+    // stm32_flash_corrupt(0x8043200);
     auto *dma = dmaStreamAlloc(STM32_DMA_STREAM_ID(1, 1), 0, nullptr, nullptr);
+
+#if HAL_DMA_SAFE_MEM_START
+    uint32_t *buf = (uint32_t *)HAL_DMA_SAFE_MEM_START;
```
tridge, line 549: need to check that check_err_errors() is called before we do heap initialisation, or if we do heap init then we should call malloc_dma() (ie. check if CH_CFG_USE_HEAP == TRUE then use malloc_dma())

tridge (changes requested):

Tools/AP_Bootloader/support.cpp
tridge: simpler is to always enable CH_CFG_USE_HEAP on H7, then unconditionally use malloc_dma(), that gives very simple logic

bugobliterator: modified to use CH_CFG_USE_HEAP if Flash size for bootloader >128K

tpwrules:
This is actually boards with >= 128K of bootloader space, right? Do we have any with > 128K?

magicrub:
Cube red is 256kb bootloader.

tpwrules:
Looks like it applies to the 128K boards too so the comment is wrong.

bugobliterator:
> Cube red is 256kb bootloader.

No it is 128K only. And yes @tpwrules is correct the patch applies to boards with bl flash >= 128 . Will fix the commit name

tridge (approved):

tridge:
ignoring the comment error to get this PR in

PR #27352, opened 2024-06-21
MOUNT: Add a mount2 retract

muramura:
Rovers and other vehicles sometimes have cameras mounted on both the front and rear.

### DIFF
--- a/libraries/RC_Channel/RC_Channel.cpp
+++ b/libraries/RC_Channel/RC_Channel.cpp
@@ -209,6 +209,7 @@ const AP_Param::GroupInfo RC_Channel::var_info[] = {
     // @Values{Copter}: 109:use Custom Controller
     // @Values{Copter, Rover, Plane, Blimp}:  110:KillIMU3
     // @Values{Copter,Plane,Rover,Blimp,Sub,Tracker}: 112:SwitchExternalAHRS
+    // @Values{Copter, Rover, Plane}: 113:Retract Mount2
     // @Values{Plane}: 150:CRUISE Mode
     // @Values{Copter}: 151:TURTLE Mode
     // @Values{Copter}: 152:SIMPLE heading reset
@@ -699,6 +700,7 @@ void RC_Channel::init_aux_function(const AUX_FUNC ch_option, const AuxSwitchPos 
     case AUX_FUNC::FFT_NOTCH_TUNE:
 #if HAL_MOUNT_ENABLED
     case AUX_FUNC::RETRACT_MOUNT1:
+    case AUX_FUNC::RETRACT_MOUNT2:
     case AUX_FUNC::MOUNT_LOCK:
 #endif
     case AUX_FUNC::LOG_PAUSE:
@@ -734,7 +736,10 @@ const RC_Channel::LookupTable RC_Channel::lookuptable[] = {
     { AUX_FUNC::PARACHUTE_RELEASE,"ParachuteRelease"},
     { AUX_FUNC::PARACHUTE_3POS,"Parachute3Position"},
     { AUX_FUNC::MISSION_RESET,"MissionReset"},
+#if HAL_MOUNT_ENABLED
     { AUX_FUNC::RETRACT_MOUNT1,"RetractMount1"},
+    { AUX_FUNC::RETRACT_MOUNT2,"RetractMount2"},
+#endif
     { AUX_FUNC::RELAY,"Relay1"},
     { AUX_FUNC::MOTOR_ESTOP,"MotorEStop"},
     { AUX_FUNC::MOTOR_INTERLOCK,"MotorInterlock"},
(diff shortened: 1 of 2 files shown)


khancyr (approved):
LGTM, same as for mount1

rmackay9:
Hi @muramura,

We discussed at this week's EU dev call and to save flash could you please add a new method called, "do_aux_function_retract_mount()" that takes a "ch_flag" and a "uint8_t instance" which is set to 0 or 1 depending upon whether we're handling RETRACT_MOUNT1 or RETRACT_MOUNT2?

muramura:
@rmackay9 san.   I understand. I will make the changes.

rmackay9 (approved):

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -1248,6 +1251,34 @@ void RC_Channel::do_aux_function_fft_notch_tune(const AuxSwitchPos ch_flag)
...
+ * 
+ * @param [in] ch_flag  Position of the switch. HIGH, MIDDLE and LOW.
+ * @param [in] instance 0: RETRACT MOUNT 1 <br>
+ *                      1: RETRACT MOUNT 2
+*/
+void RC_Channel::do_aux_function_retract_mount(const AuxSwitchPos ch_flag, const uint8_t instance)
+{
+#if HAL_MOUNT_ENABLED
+    AP_Mount *mount = AP::mount();
+    if (mount == nullptr || mount->get_num_instances() < (instance+1u)) {
```
IamPete1, line 1265: Mount `set_mode` call already deals with a invalid instance. There should be no need to check again here. That means we can save a little flash because we don't need the new `get_num_instances` method.

IamPete1 (changes requested):

libraries/AP_Mount/AP_Mount.h
```diff
@@ -129,6 +129,8 @@ class AP_Mount
     // return primary instance ID
     uint8_t get_primary_instance() const { return _primary; }
 
+    uint8_t get_num_instances() const { return _num_instances; }
+
```
IamPete1, line 133: Don't need this anymore.

libraries/RC_Channel/RC_Channel.cpp
muramura: I understand. I will delete it.

libraries/AP_Mount/AP_Mount.h
muramura: I understand. I will delete it.

muramura:
@IamPete1 san.   I made changes.

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -735,6 +737,7 @@ const RC_Channel::LookupTable RC_Channel::lookuptable[] = {
     { AUX_FUNC::PARACHUTE_3POS,"Parachute3Position"},
     { AUX_FUNC::MISSION_RESET,"MissionReset"},
     { AUX_FUNC::RETRACT_MOUNT1,"RetractMount1"},
+    { AUX_FUNC::RETRACT_MOUNT2,"RetractMount2"},
```
IamPete1, line 740:
```suggestion
#if HAL_MOUNT_ENABLED
    { AUX_FUNC::RETRACT_MOUNT1,"RetractMount1"},
    { AUX_FUNC::RETRACT_MOUNT2,"RetractMount2"},
#endif
```

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -1248,6 +1251,34 @@ void RC_Channel::do_aux_function_fft_notch_tune(const AuxSwitchPos ch_flag)
...
+/**
+ * Perform the RETRACT_MOUNT 1/2 process.
+ * 
+ * @param [in] ch_flag  Position of the switch. HIGH, MIDDLE and LOW.
+ * @param [in] instance 0: RETRACT MOUNT 1 <br>
+ *                      1: RETRACT MOUNT 2
+*/
+void RC_Channel::do_aux_function_retract_mount(const AuxSwitchPos ch_flag, const uint8_t instance)
+{
+#if HAL_MOUNT_ENABLED
```
IamPete1, line 1263:
```suggestion
#if HAL_MOUNT_ENABLED
void RC_Channel::do_aux_function_retract_mount(const AuxSwitchPos ch_flag, const uint8_t instance)
{
```

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -1248,6 +1251,34 @@ void RC_Channel::do_aux_function_fft_notch_tune(const AuxSwitchPos ch_flag)
...
+        break;
+    case AuxSwitchPos::MIDDLE:
+        // nothing
+        break;
+    case AuxSwitchPos::LOW:
+        mount->set_mode_to_default(instance);
+        break;
+    }
+#endif  // HAL_MOUNT_ENABLED
+}
```
IamPete1, line 1280:
```suggestion
}
#endif  // HAL_MOUNT_ENABLED

```

libraries/RC_Channel/RC_Channel.cpp
IamPete1: By defining out the whole function we can make sure there will be a compiler error is someone trys to use it without mount in the future.

IamPete1 (commented):
Sorry, spotted some more things.

muramura:
@IamPete1 san.   I agreed.

PR #27353, opened 2024-06-21
Sub: use cork() when writing motors,  move servo updates to motors_output()

Williangalvani:
it turns out that sub was missing this, which meant that there were two code paths able to change pwm, which harmed cpu usage and made the fifty_hz_loop() take longer than it should. 

I also moved calc_pwm from the fifty_hz loop down to motors_output, which despite causing us to calculate the servos more often, improves things as we no longer write to rcout on that task

### DIFF
--- a/ArduSub/ArduSub.cpp
+++ b/ArduSub/ArduSub.cpp
@@ -158,10 +158,7 @@ void Sub::fifty_hz_loop()
 
     failsafe_sensors_check();
 
-    // Update rc input/output
     rc().read_input();
-    SRV_Channels::calc_pwm();
-    SRV_Channels::output_ch_all();
 }
 
 // update_batt_compass - read battery and compass
--- a/ArduSub/motors.cpp
+++ b/ArduSub/motors.cpp
@@ -18,7 +18,11 @@ void Sub::motors_output()
         verify_motor_test();
     } else {
         motors.set_interlock(true);
+        SRV_Channels::cork();
+        SRV_Channels::calc_pwm();
+        SRV_Channels::output_ch_all();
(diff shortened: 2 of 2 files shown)


ES-Alexander:
We partly found this by looking at the `@SYS/tasks.txt` file.

Could be cool/useful to include some kind of checks on the CPU usage amounts and proportions of each scheduled task during auto-testing, in a similar way to how the size test determines changes in the flash usage for each PR... :-)

Williangalvani:
and now it looks like this is not quite the right fix, I'm looking further into it

ArduSub/ArduSub.cpp
```diff
@@ -160,8 +160,6 @@ void Sub::fifty_hz_loop()
 
     // Update rc input/output
```
ES-Alexander, line 161:
```suggestion
    // Update rc input (output gets corked and updated in motors.cpp)
```

ES-Alexander:
I tested on a Pixhawk, and the motor and non-motor servo outputs are still working 👍 

---

Pulled the `@SYS/tasks.txt` files for reference:
[Sub-4.5_tasks.txt](https://github.com/user-attachments/files/16022917/Sub-4.5_tasks.txt)
[PR_tasks.txt](https://github.com/user-attachments/files/16022918/PR_tasks.txt)

Main changes are:
| Function | Sub-4.5 AVG | PR (4.6) AVG |
| --- | --- | --- |
 `motors_output` | 71 | 102
`read_AHRS` | 660 | 698
`update_send` | 71 | 105
`fifty_hz_loop` | 215 | 11

PR #27354, opened 2024-06-21
Ardupilot-MAVlink battery monitor

FAIZALR:
This commit introduces functionality to support battery monitors using the MAVLink protocol. The following changes are included:

- Added conditional compilation for MAVLink support (`#if AP_BATTERY_MAVLINK_ENABLED`).
- Defined `AP_BATTERY_MAVLINK_ENABLED` to default to `AP_BATTERY_BACKEND_DEFAULT_ENABLED` if not explicitly set.
- Implemented `handle_msg(const mavlink_message_t &msg)` in `AP_BattMonitor` to iterate through configured instances and pass MAVLink messages to appropriate drivers.
- Updated driver instantiation to include MAVLink type (`Type::MAVLink`) when configured (`BATTx_MONITOR=30`).

These changes enables the battery monitor to interface with MAVLink-enabled battery devices, enabling the flight controller to receive and process battery status updates via MAVLink messages. The implementation ensures compatibility and functionality when MAVLink battery monitor driver  is enabled through configuration (**BATTx_MONITOR=30**) and tested in SITL.

mavproxy
mission_planner


khancyr (approved):
Nicely done ! Thanks that looks good

Hwurzburg (changes requested):
- Needs new monitor type included in parameter metadata
- Needs this option defaulted to not be included (this is an esoteric monitor and most firmware should not pay the flash cost) and the option added to build_options.py and extract_features.py

tridge (changes requested):
this PR has some major fundamental issues. We could be getting BATTERY_STATUS messages from another vehicle, or even from ourselves via a loopback. There is no protection against this. Using a sensor reporting msg as sensor input is inherently difficult, it also doesn't support multiple batteries.
Why do you want to do this? Can you explain the setup?
Note that you could implement exactly this functionality using lua, which can parse mavlink messages and create a battery backend

peterbarker:
Filter based on src system to make sure it's coming from the same vehicle.

peterbarker:
Ping @FAIZALR - you won't be able to reuse that mavlink message, at least.

Did you still wish to pursue this PR?

peterbarker:
Closing this one now.  It's not a terrible concept, just can't use that message!

PR #27355, opened 2024-06-22
AP_Math: split CRCs from checksums

tpwrules:
All CRCs are checksums but not all checksums are CRCs.

Paves the way for deduplication and unification of CRC implementations, including interface and standard of implementation.

No functional change yet. Hopefully this cleanup is appreciated.

### DIFF
--- a/libraries/AP_RCProtocol/AP_RCProtocol_FPort2.cpp
+++ b/libraries/AP_RCProtocol/AP_RCProtocol_FPort2.cpp
@@ -179,7 +179,7 @@ void AP_RCProtocol_FPort2::decode_downlink(const FPort2_Frame &frame)
         // get fresh telem_data in the next call
         telem_data.available = false;
     }
-    buf[9] = crc_sum8_with_carry(&buf[1], 8);
+    buf[9] = cksum_fport(&buf[1], 8);
     
     uart->write(buf, sizeof(buf));
 #endif
@@ -286,7 +286,7 @@ reset:
 // check checksum byte
 bool AP_RCProtocol_FPort2::check_checksum(void)
 {
-    return crc_sum8_with_carry(&byte_input.buf[1], byte_input.control_len-1) == 0;
+    return cksum_fport(&byte_input.buf[1], byte_input.control_len-1) == 0;
 }
 
 // support byte input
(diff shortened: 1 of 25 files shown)


tridge:
> No functional change yet. Hopefully this cleanup is appreciated.

I'm a bit sceptical of the benefit of this change. Can you point at some examples of the ultimate benefit? I'm well aware of the distinction between crc and checksum, but for our purposes that is an implementation detail and really only of academic interest

tpwrules:
The benefit is that by separating out the CRCs, we can switch to a common implementation for them. There is a known taxonomy (see section 15 [here](https://www.zlib.net/crc_v3.txt)) and there are various code generators and template libraries available for the different types.

This lets us identify and remove duplicates (e.g. `crc_xmodem`, `crc16_ccitt`) and intelligently switch between bit-wise or byte-wise computation with the space/speed tradeoff. We can also remove all the warts of the different functions (argument order, width of input size) and reduce flash usage.

peterbarker:
> The benefit is that by separating out the CRCs, we can switch to a common implementation for them. There is a known taxonomy (see section 15 [here](https://www.zlib.net/crc_v3.txt)) and there are various code generators and template libraries available given those parameters.
> 
> This lets us identify and remove duplicates (e.g. `crc_xmodem`, `crc16_ccitt`) and intelligently switch between bit-wise or byte-wise computation with the space/speed tradeoff. We can also remove all the warts of the different functions (argument order, width of input size) and reduce flash usage.

That all sounds good.  Have you seen a project which does this sort of heavy-reuse of CRC calculation?

libraries/AP_RCProtocol/AP_RCProtocol_FPort2.cpp
```diff
@@ -286,7 +286,7 @@ void AP_RCProtocol_FPort2::_process_byte(uint32_t timestamp_us, uint8_t b)
 // check checksum byte
 bool AP_RCProtocol_FPort2::check_checksum(void)
 {
-    return crc_sum8_with_carry(&byte_input.buf[1], byte_input.control_len-1) == 0;
```
peterbarker, line 289: Decided this was a bad rename

tpwrules:
Going to come at this in a less invasive manner some time in the future.

PR #27358, opened 2024-06-22
AP_BatteryMonitor: add missing failsafe with no action

IamPete1:
Currently if your battery monitor goes away the only way to tell is a bit in sys status. This is quite easy to miss. This now results in a full battery failsafe with no action. We might want to consider adding a parameter in the future. This uses a 5 second timeout.

This becomes more likely with CAN battery monitors which are often powered from the battery itself, so if the battery fails the node also disappears rather than telling us that the battery is gone. 

Example:
image

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor.h
+++ b/libraries/AP_BattMonitor/AP_BattMonitor.h
@@ -81,6 +81,7 @@ public:
     // battery failsafes must be defined in levels of severity so that vehicles wont fall backwards
     enum class Failsafe : uint8_t {
         None = 0,
+        Unhealthy,
         Low,
         Critical
     };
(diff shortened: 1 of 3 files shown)


libraries/AP_BattMonitor/AP_BattMonitor.h
```diff
@@ -81,6 +81,7 @@ class AP_BattMonitor
     // battery failsafes must be defined in levels of severity so that vehicles wont fall backwards
     enum class Failsafe : uint8_t {
         None = 0,
+        Unhealthy,
```
tridge, line 84: need to check if this is exposed anywhere (logging? lua scripting?) - this changes the numbering

peterbarker (approved):

libraries/AP_BattMonitor/AP_BattMonitor.h
IamPete1: I have confirmed this is not exposed. I changed the enum name and the state variable name and built to make sure I checked all the users.

tridge (approved):

PR #27359, opened 2024-06-22
AP_BattMonitor: Sum, ESC: all selected items must be found for health.

IamPete1:
Currently both the sum and ESC backends will hide a failure because they take a average. So if one item is lost the voltage stays the same and no fail-safe is tripped. This will now mark the monitor unhealthy if any item is missing. With https://github.com/ArduPilot/ardupilot/pull/27358 this gives a clear warning to the pilot.

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
@@ -99,12 +99,27 @@ void AP_BattMonitor_ESC::read(void)
         }
     }
 
+    // Max number of ESCs ever seen
+    esc_count = MAX(esc_count, voltage_escs);
+
+    if (voltage_escs == 0) {
+        // Need at least one to be healthy
+        _state.healthy = false;
+
+    } else if (all_enabled) {
+        // All ESCs must be found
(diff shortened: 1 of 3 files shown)


tridge (changes requested):
we need to think about a case such as an octo-quad copter, using ESC telem for voltage, with battery failsafes enabled, and an ESC blows up, so we lose telemettry
the ESC based voltage monitor would now be unhealthy, does this change behaviour? do we lose the ability to do failsafes?
looks like unhealthy prevents current accumulation for battery percentage failsafe

tridge:
so what happens when an esc fails on an octo using ESC telem for battery monitoring? single statustext? continual messages?

IamPete1:
> so what happens when an esc fails on an octo using ESC telem for battery monitoring? single statustext? continual messages?

You hit the new missing battery failsafe, no action but same reporting behavior as the existing failsafes. Single message and beeper going. The low voltage/capacity failfafes will still trigger as before.

IamPete1:
This costs about 200B on 1M boards and 100B on periph. We could define out the new method on periph for zero flash cost.

tridge:
i'll do the test for octo-quad esc failure

Issue #27360, opened 2024-06-22
8266 based ELRS or mavesp8266 fails due to DEFAULTGPIO driving their TX lines low in bootloader


Hwurzburg:
Within the last year many boards started using the DEFAULTGPIO.....LOW instruction in their bootloaders. Recently Andy Piper got reports that ELRS modules would not work with ArduPilot and has started adding commands on "likely" RCIN UART RX pins to pull high in the bootloader. Discussing with the ELRS devs, it seem that an ESP8266 will enter a hardware test mode that cannot be exited except by reboot if its UART TX is pulled low on power up. This means that if an ESP8266 is attached to an ArduPilot autopilot, either as a MAVESP telem wifi or ELRS, it will lock in a hardware selftest mode rendering it useable. 
Andy's solution is not really viable unless you pull ALL UART RX pins high or tristate during bootloader. This would be a massive code effort. Alternatively, we could remove all DEFAULTGPIO statements from the bootloader code....or change to tristate?
We need a viable solution...

Why hasn't this come up before? most ELRS are ESP32 based and don't have this issue and the MAVesp is not widely used either, but ELRS has been ported to R9 modules and others now and they are ESP8266 based, as I understand it...so just now starting to emerge

**Platform**
[ x] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

tridge:
as far as I can see there is only one board with a DEFAULTGPIO before the MCU statement in a hwdef-bl.dat, the BlitzF745AIO
@andyp1per I think this should be removed, so it is floating. I think we have to rely on the makers of the devices having a weak pullup so we should float

PR #27361, opened 2024-06-23
hwdef: don't use DEFAULTGPIO in bootloaders

tridge:
setting up a DEFAULTGPIO pulldown in bootloaders is a bad idea as it overrides any hardware pullups that have been put in place as part of a "hold in bootloader" mechanism. See discussion in #27360 for ELRS. Instead we want undefined pins to be floating, which is the default anyway. This does mean higher ESD vulnerability in bootloaders, but we really have no choice as we have no way to predict which pins need what pullup/pulldown to avoid peripherals from not booting

note that this only impacts one board that I can see, the BlitzF745AIO, due to the MCU vs DEFAULTGPIO interaction


andyp1per:
So while this will fix the ELRS issue it will not fix the servo issue which we get reports of - basically people report that they can get random servo movements due to the floating state of the servo pins in the bootloader - particularly affects plane users. So I feel like we need a more nuanced solution here.

tridge:
> So while this will fix the ELRS issue it will not fix the servo issue which we get reports of - basically people report that they can get random servo movements due to the floating state of the servo pins in the bootloader

i don't think the float state is what causes this, I think its a chopped PWM issue. I can reproduce on one of my aircraft. Happy to discuss possible fixes

andyp1per:
FWIW leaving it floating does *not* work at least in the circumstances I saw - see the history here - https://discuss.ardupilot.org/t/iflight-blitz-f7-aio/99619 - you have to pull the pins explicitly high

PR #27364, opened 2024-06-23
deleted

burgeruser:


rmackay9:
Hi @burgeruser,

Thanks for the PR.  I wonder how much thought has been put into the flight controller name?  I'm not a lawyer of course but it seems like it could be problematic.

PR #27365, opened 2024-06-23
Scripting: added command_int API for accessing MAV_CMD_xxx

tridge:
This makes it possible to call most MAV_CMD_xxx calls from lua scripts, gaining us a lot of functionality with a small amount of flash
This was prompted by #27223 and the MAV_CMD_GUIDED_xxx commands, but is really very broadly useful

some questions:
 - should we return the full MAV_RESULT enum? Is it useful?
 - should we create a dedicated GCS_MAVLink channel for lua?

this also adds an example that tests DO_SET_MODE and DO_REPOSITION

### DIFF
--- a/libraries/AP_Scripting/examples/command_int.lua
+++ b/libraries/AP_Scripting/examples/command_int.lua
@@ -1,0 +1,56 @@
+--[[
+   demonstrate using the gcs:command_int() interface to send commands from MAVLink MAV_CMD_xxx set
+--]]
+
+local MAV_FRAME_GLOBAL_RELATIVE_ALT = 3
+
+local MAV_CMD_DO_SET_MODE = 176
+local MAV_CMD_DO_REPOSITION = 192
+
+-- some plane flight modes for testing
+local MODE_LOITER = 12
+local MODE_GUIDED = 15
+
+local MAV_MODE_FLAG_CUSTOM_MODE_ENABLED = 1
+
+--[[
+create a NaN value
+--]]
+local function NaN()
+   return 0/0
+end
+
+--[[
+   test API calls. When in LOITER mode change to GUIDED and force flying to a location NE of home
+--]]
+local function test_command_int()
+   if vehicle:get_mode() ~= MODE_LOITER then
+      return
+   end
+   local home = ahrs:get_home()
+   if not home then
+      return
+   end
+
+   -- force mode GUIDED
+   gcs:run_command_int(MAV_CMD_DO_SET_MODE, { p1 = MAV_MODE_FLAG_CUSTOM_MODE_ENABLED, p2 = MODE_GUIDED })
+
+   -- and fly to 200m NE of home and 100m above home
+   local dest = home:copy()
+   dest:offset_bearing(45, 200)
+
+   gcs:run_command_int(MAV_CMD_DO_REPOSITION, { frame = MAV_FRAME_GLOBAL_RELATIVE_ALT,
+                                                p1 = -1,
+                                                p4 = NaN(),
+                                                x = dest:lat(),
+                                                y = dest:lng(),
+                                                z = 100 })
+end
+
+local function update()
+   test_command_int()
+   return update, 1000
+end
+
+return update, 1000
+
--- a/libraries/GCS_MAVLink/GCS.h
+++ b/libraries/GCS_MAVLink/GCS.h
@@ -1271,6 +1271,11 @@ public:
 
     virtual uint8_t sysid_this_mav() const = 0;
 
+#if AP_SCRIPTING_ENABLED
+    // lua access to command_int
+    MAV_RESULT lua_command_int_packet(const mavlink_command_int_t &packet);
+#endif
+
 protected:
 
     virtual GCS_MAVLINK *new_gcs_mavlink_backend(GCS_MAVLINK_Parameters &params,
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -2605,6 +2605,12 @@ function gcs:send_text(severity, text) end
 ---@return uint32_t_ud -- system time in milliseconds
 function gcs:last_seen() end
 
+-- call a MAVLink MAV_CMD_xxx command via command_int interface
+---@param command integer -- MAV_CMD_xxx
+---@param params table -- parameters of p1, p2, p3, p4, x, y and z and frame. Any not specified taken as zero
+---@return boolean
+function gcs:run_command_int(command, params) end
+
 -- The relay library provides access to controlling relay outputs.
 relay = {}
 
--- a/libraries/AP_Scripting/generator/description/bindings.desc
+++ b/libraries/AP_Scripting/generator/description/bindings.desc
@@ -296,6 +296,7 @@ singleton GCS method get_high_latency_status depends HAL_HIGH_LATENCY2_ENABLED =
 
 singleton GCS method enable_high_latency_connections void boolean
 singleton GCS method enable_high_latency_connections depends HAL_HIGH_LATENCY2_ENABLED == 1
+singleton GCS manual run_command_int lua_GCS_command_int 2 1
 
 include AP_ONVIF/AP_ONVIF.h depends ENABLE_ONVIF == 1
 singleton AP_ONVIF depends ENABLE_ONVIF == 1
--- a/libraries/AP_Scripting/lua_bindings.cpp
+++ b/libraries/AP_Scripting/lua_bindings.cpp
@@ -1112,4 +1112,78 @@ void lua_abort()
 #endif
 }
 
+#if HAL_GCS_ENABLED
+/*
+  implement gcs:command_int() access to MAV_CMD_xxx commands
+ */
+int lua_GCS_command_int(lua_State *L)
+{
+    GCS *_gcs = check_GCS(L);
+    binding_argcheck(L, 3);
+
+    const uint16_t command = get_uint16_t(L, 2);
+    if (!lua_istable(L, 3)) {
+        // must have parameter table
+        return 0;
+    }
+
+    mavlink_command_int_t pkt {};
+
+    pkt.command = command;
+
+    float *params = &pkt.param1;
+    int32_t *xy = &pkt.x;
+
+    // extract the first 4 parameters as floats
+    for (uint8_t i=0; i<4; i++) {
+        char pname[3] { 'p' , char('1' + i), 0 };
+        lua_pushstring(L, pname);
+        lua_gettable(L, 3);
+        if (lua_isnumber(L, -1)) {
+            params[i] = lua_tonumber(L, -1);
+        }
+        lua_pop(L, 1);
+    }
+
+    // extract the xy values
+    for (uint8_t i=0; i<2; i++) {
+        const char *names[] = { "x", "y" };
+        lua_pushstring(L, names[i]);
+        lua_gettable(L, 3);
+        if (lua_isinteger(L, -1)) {
+            xy[i] = lua_tointeger(L, -1);
+        }
+        lua_pop(L, 1);
+    }
+
+    // and z
+    lua_pushstring(L, "z");
+    lua_gettable(L, 3);
+    if (lua_isnumber(L, -1)) {
+        pkt.z = lua_tonumber(L, -1);
+    }
+    lua_pop(L, 1);
+
+    // optional frame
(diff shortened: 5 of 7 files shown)


peterbarker:
>     * should we return the full MAV_RESULT enum? Is it useful?

It'd probably only be useful during script development - but it could be very useful there.  Most of the time the script probably couldn't do anything useful with the more-specific result code.

>     * should we create a dedicated GCS_MAVLink channel for lua?

Probably not for now.... only problem i can think of is commands that have side-effects on the channel e.g. setting stream rates.

> this also adds an example that tests DO_SET_MODE and DO_REPOSITION

The interface would be *much*, *much nicer with optional parameters, in the same way that `run_cmd_int` works in autotest.  Some of those look like this: `self.run_cmd_int(COMMAND_NUM, p1=xyzzy)` which is nice.

I wonder if the lack of target component might cause problems somewhere.  I can't think of anywhere that might matter, however.

tridge:
@peterbarker did you read the example? It does have optional parameters

IamPete1:
This https://github.com/ArduPilot/ardupilot/pull/27219 allows this and lots more, but it will be a little more complex to deal with.

tpwrules:
I think this introduces locking issues? Ordinarily commands are processed starting at the GCS's `update_receive` method from the main scheduler. But as I understand Lua runs in a different thread, so it could enter the command handler while received data is being processed and commands are being handled.

libraries/AP_Scripting/examples/command_int.lua
```diff
@@ -0,0 +1,45 @@
...
+   if vehicle:get_mode() ~= MODE_LOITER then
+      return
+   end
+   local home = ahrs:get_home()
+   if not home then
+      return
+   end
+
+   -- force mode GUIDED
+   gcs:command_int(MAV_FRAME_GLOBAL, MAV_CMD_DO_SET_MODE, { MAV_MODE_FLAG_CUSTOM_MODE_ENABLED, MODE_GUIDED })
```
timtuxworth, line 29: The vehicle:update_target_location(old_location, new_location) method validates the old location hasn't changed before setting new location to "prevent a race condition where the user changes mode or commands a different target in the controlling lua script". Is that not a problem when using MAVLink?

libraries/AP_Scripting/examples/command_int.lua
```diff
@@ -0,0 +1,45 @@
...
+
+   -- and fly to 200m NE of home and 100m above home
+   local dest = home:copy()
+   dest:offset_bearing(45, 200)
+   dest:alt(home:alt()+100*100)
+
+   gcs:command_int(MAV_FRAME_GLOBAL, MAV_CMD_DO_REPOSITION, { -1, 0, 0, 0, dest:lat(), dest:lng(), dest:alt()*0.01 })
```
timtuxworth, line 36: The vehicle:update_target_location(old_location, new_location) method validates the old location hasn't changed before setting new location to "prevent a race condition where the user changes mode or commands a different target in the controlling lua script". Is that not a problem when using MAVLink?

tridge: you can still use that method, this just offers access to a much broader range of functions, with the same semantics you get as over MAVLink from a companion computer.
If you need that race free update then you should use vehicle:update_target_location()

tridge:
> I think this introduces locking issues? Ordinarily commands are processed starting at the GCS's `update_receive` method from the main scheduler.

good catch, I've added a GCS mutex

tridge:
> This #27219 allows this and lots more, but it will be a little more complex to deal with.

a little more complex? It would be horribly complex to configure then construct mavlink messages, then parse the reply to work out if it succeeded. How would you even do it as a synchonous lua call? You'd need to wait for GCS processing by the main thread then wait for the reply, which means adding a state machine in the lua code.

IamPete1:
> a little more complex? It would be horribly complex to configure then construct mavlink messages, then parse the reply to work out if it succeeded. How would you even do it as a synchonous lua call? You'd need to wait for GCS processing by the main thread then wait for the reply, which means adding a state machine in the lua code.

I would hope we would update the examples to do the tricky stuff. The synchronous issue is still a problem for this PR, although in a much smaller way when you get a `MAV_RESULT_IN_PROGRESS`, how do you get the final result in that case?

I guess my point is really this: Are we going to add helpers for every MAVLink message that might be useful for a script to send to the vehicle? I agree that the commands do tick off lots of them, but going the direct MAVLink way gets everything. Not to mention all the messages going the other way that give information we may not have bindings for.

tpwrules:
I think also such a state machine could be nicely encapsulated and appear synchronous to the user if coroutines made a return.

tridge:
> Are we going to add helpers for every MAVLink message that might be useful for a script to send to the vehicle?

with command_int we get access to a huge pile of functions with one API

> I would hope we would update the examples to do the tricky stuff

it is still going to be:
 - slow (waiting for main thread to process commands)
 - much more complex to setup for user

> when you get a MAV_RESULT_IN_PROGRESS, how do you get the final result in that case?

that is indeed tricky, but commands that do that are very rare (only one that comes to mind is airspeed cal), and as far as I know the final result is always a pass?

There certainly are some MAV_CMD_xxx that won't make sense. For example, the one that triggers sending of MAVCAN makes no sense. It will work, but will just cause the pkts to come out on USB

timtuxworth:
I've tested this in SITL as a replacement for update_target_location() for plane follow, and it has the same problem - it tries to loiter around the target aircraft if it gets too close to the target location.

libraries/GCS_MAVLink/GCS.h
```diff
@@ -1288,6 +1293,7 @@ class GCS
 private:
 
     static GCS *_singleton;
+    HAL_Semaphore _sem;
```
peterbarker, line 1296: I'm not sure if this is insufficent protection or too blunt.

It's too blunt if you're protecting against mavlink-command-ints being run at the same time - that should go within `handle_command_int`.

It might not be sufficient protection depending on where the main thread is up to when the script command is executed.  Remembering, of course, that we process mavlink while we're sleeping (at least that's only likely when we're disarmed....)

tridge: I presume you mean in AP_Vehicle::scheduler_delay_callback() ? that calls gcs().update_receive() which is exactly the call the mutex in this PR protects.

peterbarker:
> I think also such a state machine could be nicely encapsulated and appear synchronous to the user if coroutines made a return.

These commands *must* run fast as they're processed on the main thread ATM when executed via mavlink input.

You *may* get handed back an "in progress" result... and that would require some clever handling in LUA...

libraries/GCS_MAVLink/GCS.h
tpwrules: Upon further thought I agree that this is insufficient locking. Consider, for example `AP_Gripper_Servo::release`. It can be called through `GCS_MAVLINK::handle_command_do_gripper` which is called by `update_receive` and is protected by that new lock. But on the same object `AP_Gripper_Servo::grab` can be called through `AP_Gripper_Backend::update()` which is called by the main scheduler on the main thread too and not protected by any lock.

The scheduler won't run two things at once (IIUC) but if Lua is releasing while the main thread decides to grab, who wins? We need another lock in the gripper methods too. Then we have to start being concerned about locking order and deadlocks, and look at locking in everything that can be done by the command int handler...

This problem for Gripper is already triggerable through the aux function handler and bindings in any case (though adding a generic lock there would be a lot more straightforward), but I don't think we want it to be worse.

tpwrules:
> > I think also such a state machine could be nicely encapsulated and appear synchronous to the user if coroutines made a return.
> 
> These commands _must_ run fast as they're processed on the main thread ATM when executed via mavlink input.
> 
> You _may_ get handed back an "in progress" result... and that would require some clever handling in LUA...



> > I think also such a state machine could be nicely encapsulated and appear synchronous to the user if coroutines made a return.
> 
> These commands _must_ run fast as they're processed on the main thread ATM when executed via mavlink input.
> 
> You _may_ get handed back an "in progress" result... and that would require some clever handling in LUA...

Like I said this can be done elegantly with coroutines. I think we should study that possibility. It would require more complex scripting for sure and use up valuable scripting memory, but would need no new bindings and create no new problems.

timtuxworth:
> > I think also such a state machine could be nicely encapsulated and appear synchronous to the user if coroutines made a return.
> 
> These commands _must_ run fast as they're processed on the main thread ATM when executed via mavlink input.
> 
> You _may_ get handed back an "in progress" result... and that would require some clever handling in LUA...
As far as I can see "you" (i.e. Lua) will never get an "in progress" result, it seems to be hidden from Lua as lua_GCS_command_int only returns true/false success/fail.

libraries/AP_Scripting/examples/command_int.lua
```diff
@@ -0,0 +1,48 @@
...
+   if vehicle:get_mode() ~= MODE_LOITER then
+      return
+   end
+   local home = ahrs:get_home()
+   if not home then
+      return
+   end
+
+   -- force mode GUIDED
+   gcs:command_int(MAV_CMD_DO_SET_MODE, { p1 = MAV_MODE_FLAG_CUSTOM_MODE_ENABLED, p2 = MODE_GUIDED })
```
tridge, line 29: run_command_int instead

timtuxworth:
For DO_REPOSITION param4 (Yaw) says "Yaw heading. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.). For planes indicates loiter direction (0: clockwise, 1: counter clockwise)"

So if I want to **not set** this value (yaw) and leave it as whatever it was before, should I not pass "Nil" from Lua? except that lua_GCS_command_int doesn't seem to support that.

tpwrules:
You can generate a NaN in Lua with the expression `0/0`. In SITL this appears to be the NaN encoded as `0xFFC00000`.

timtuxworth:
I'm trying to use DO_REPOSITION to set the vehicle target lat/lng and altitude. The vehicle successfully heads to the new target lat/lng but appears to ignore the altitude. This is the wrapper I use to call it, passing in alt as cm.

```
local function set_vehicle_target_location(target)
   if not gcs:command_int(MAV_FRAME.GLOBAL_RELATIVE_ALT, MAV_CMD_INT.DO_REPOSITION, { target.groundspeed or -1, target.options or 0, target.radius or 2, target.yaw or 0,
                           x = target.lat, y = target.lng, z = target.alt * 0.01 }) then
      gcs:send_text(MAV_SEVERITY.ERROR, SCRIPT_NAME_SHORT .. ": MAVLink DO_REPOSITION returned false")
   end
end

```

timtuxworth:
My plane follow script uses do_reposition to set the target lat/lng/alt to the location returned by AP_Follow, but it it gets too close to the point, it switches to GUIDED_CHANGE_HEADING to maintain heading alongside the target plane.

GUIDED_CHANGE_HEADING sets plane.guided_state.target_heading_type, to either COG or HEADING but there is now ay to change it back to NONE, so once the script has set heading, it can't revert to position mode without switching out of GUIDED and back again. 

I found I could get this to work bu adding one line handle_command_int_do_reposition() right before the plane.set_mode() at line 844

`        plane.guided_state.target_heading_type = GUIDED_HEADING_NONE;
`

tridge:
> You can generate a NaN in Lua with the expression `0/0`

yep, I've added NaN() to the example and used it for yaw

tridge:
> This is the wrapper I use to call it, passing in target.alt as m.

please update to the current PR, this is the old syntax

timtuxworth:
> please update to the current PR, this is the old syntax

I love the new syntax - very clean.

tridge:
> GUIDED_CHANGE_HEADING sets plane.guided_state.target_heading_type, to either COG or HEADING but there is no way to change it back to NONE

as a separate PR we should add GUIDED_HEADING_DEFAULT to the enum in the HEADING_TYPE enum in the mavlink XML and then allow setting of that type in GUIDED_CHANGE_HEADING.

peterbarker (approved):
LGTM

timtuxworth:
> > GUIDED_CHANGE_HEADING sets plane.guided_state.target_heading_type, to either COG or HEADING but there is no way to change it back to NONE
> 
> as a separate PR we should add GUIDED_HEADING_DEFAULT to the enum in the HEADING_TYPE enum in the mavlink XML and then allow setting of that type in GUIDED_CHANGE_HEADING.

That seems to require a PR in the mavlink repo am I right?

PR #27366, opened 2024-06-23
Plane: Change IF statement to SWITCH statement

muramura:
C++ has a SWITCH statement.
I think the SWITCH statement is better than the IF statement when you want to select different processes with the same variable.


peterbarker (approved):
LGTM

peterbarker:
CI failure is an internal github error on a bootloader build.  I think ti's safe to merge anyway

tridge:
not keen on this, the testing to validate it isn't worth the change

PR #27368, opened 2024-06-23
autotest: install multiple test scripts with one call

peterbarker:


Tools/autotest/vehicle_test_suite.py
```diff
@@ -4632,11 +4632,17 @@ def install_example_script_context(self, scriptname):
...
+    def install_test_script_context(self, scriptnames):
         '''installs an test script which will be removed when the context goes
         away'''
-        self.install_test_script(scriptname)
-        self.context_get().installed_scripts.append(scriptname)
+        if getattr(scriptnames, '__iter__', None) is None:
+            scriptname = [scriptnames]
+        for scriptname in scriptnames:
+            self.install_test_script(scriptname)
+            self.context_get().installed_scripts.append(scriptname)
```
tpwrules, line 4642: Strings have an `__iter__` method too, this will install every letter as a separate string. The test should probably be `if isinstance(scriptnames, str)`.

Tools/autotest/vehicle_test_suite.py
```diff
@@ -4632,11 +4632,17 @@ def install_example_script_context(self, scriptname):
...
+
+    def install_test_scripts_context(self, *args, **kwargs):
+        return self.install_test_script_context(*args, **kwargs)
```
tpwrules, line 4645: Why an additional function?

tpwrules (changes requested):

Tools/autotest/vehicle_test_suite.py
tpwrules: The assignment is also incorrect.

Tools/autotest/vehicle_test_suite.py
peterbarker: > Why an additional function?

Yeah, should have added that comment....

It's just pluralisation to avoid annoying someone who wants to use the function.

Tools/autotest/vehicle_test_suite.py
peterbarker: > Strings have an `__iter__` method too, this will install every letter as a separate string. The test should probably be `if isinstance(scriptnames, str)`.

Fixed, thanks!

peterbarker: > The assignment is also incorrect.

Wow, really needed extra coffee.

Probably shouldn't have removed the only caller to the singular-script case....

Tools/autotest/vehicle_test_suite.py
tpwrules: Why not have the plural function just call the single function repeatedly? Then you don’t have to worry about the type.

peterbarker: > Why not have the plural function just call the single function repeatedly? Then you don’t have to worry about the type.

The current implementation allows the user to call either with either parameter.

Trying not to repeat the problem with `set_parameter` vs `set_parameters`, which I'm still mulling up fixing (just to make it more forgiving about calling one instead of the other where getting it exactly right really shouldn't matter).

peterbarker:
Merged this - thanks @tpwrules

PR #27369, opened 2024-06-24
AP_Beacon: use enum class for type

peterbarker:
Relying on autotests to make sure this works.

zero-byte-change on CubeOrange/ArduCopter.


khancyr (approved):
LGTM

Compile time checked now, so that is fine

PR #27370, opened 2024-06-24
AC_Autotune: Clean up Multi Variables and non functional changes

lthall:
I have moved the multi only variable declaration variables into the AC_AutoTune_Multi.h.

I removed the "f" from the floats.

I simplified a couple functions.

All these changes should be non-functional.

I am a little disgusted by some of the structures that I made here. This still has some work to do before it could be considered "clean".

### DIFF
--- a/libraries/AC_AutoTune/AC_AutoTune.h
+++ b/libraries/AC_AutoTune/AC_AutoTune.h
@@ -259,24 +259,15 @@ protected:
     bool     positive_direction;         // false = tuning in negative direction (i.e. left for roll), true = positive direction (i.e. right for roll)
     StepType step;                       // see StepType for what steps are performed
     TuneType tune_type;                  // see TuneType
-    bool     ignore_next;                // true = ignore the next test
     bool     twitch_first_iter;          // true on first iteration of a twitch (used to signal we must step the attitude or rate target)
     uint8_t  axes_completed;             // bitmask of completed axes
-    float    test_rate_min;                         // the minimum angular rate achieved during TESTING_RATE step-multi only
-    float    test_rate_max;                         // the maximum angular rate achieved during TESTING_RATE step-multi only
-    float    test_angle_min;                        // the minimum angle achieved during TESTING_ANGLE step-multi only
-    float    test_angle_max;                        // the maximum angle achieved during TESTING_ANGLE step-multi only
     uint32_t step_start_time_ms;                    // start time of current tuning step (used for timeout checks)
     uint32_t step_time_limit_ms;                    // time limit of current autotune process
     uint32_t level_start_time_ms;                   // start time of waiting for level
     int8_t   counter;                               // counter for tuning gains
-    float    target_rate;                           // target rate-multi only
-    float    target_angle;                          // target angle-multi only
     float    start_angle;                           // start angle
     float    start_rate;                            // start rate - parent and multi
-    float    rate_max;                              // maximum rate variable - parent and multi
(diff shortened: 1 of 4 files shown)


libraries/AC_AutoTune/AC_AutoTune_Multi.cpp
```diff
@@ -653,11 +649,10 @@ void AC_AutoTune_Multi::twitching_test_angle(float angle, float rate, float angl
 }
 
 // twitching_measure_acceleration - measure rate of change of measurement
-void AC_AutoTune_Multi::twitching_measure_acceleration(float &rate_of_change, float rate_measurement, float &rate_measurement_max) const
+void AC_AutoTune_Multi::twitching_measure_acceleration(float &accel_average, float rate, float meas_rate_max) const
 {
-    if (rate_measurement_max < rate_measurement) {
-        rate_measurement_max = rate_measurement;
-        rate_of_change = (1000.0f*rate_measurement_max)/(AP_HAL::millis() - step_start_time_ms);
+    if (is_equal(rate, meas_rate_max)) {
```
peterbarker, line 654: Is this really non-functional?

lthall: Effectivly but I am going to remove it anyway.

libraries/AC_AutoTune/AC_AutoTune.h
```diff
@@ -259,24 +260,15 @@ class AC_AutoTune
...
-    float    test_angle_max;                        // the maximum angle achieved during TESTING_ANGLE step-multi only
     uint32_t step_start_time_ms;                    // start time of current tuning step (used for timeout checks)
     uint32_t step_time_limit_ms;                    // time limit of current autotune process
     uint32_t level_start_time_ms;                   // start time of waiting for level
     int8_t   counter;                               // counter for tuning gains
-    float    target_rate;                           // target rate-multi only
-    float    target_angle;                          // target angle-multi only
     float    start_angle;                           // start angle
     float    start_rate;                            // start rate - parent and multi
-    float    rate_max;                              // maximum rate variable - parent and multi
```
rmackay9, line 277: somewhat surprisingly AC_AutoTune_Heli also defines a "rate_max" variable but it is a parameter.  I'm slightly surprised that the compiler didn't catch this "shadowing" but in any case it's good to get rid of it.

I think @bnsgeyer should confirm that Heli autotune seems to work OK

libraries/AC_AutoTune/AC_AutoTune_Multi.cpp
```diff
@@ -1221,7 +1220,12 @@ void AC_AutoTune_Multi::twitch_test_init()
     } else {
         rotation_rate_filt.reset(0);
     }
-
+    twitch_first_iter = true;
+    test_rate_max = 0.0;
+    test_rate_min = 0.0;
+    test_angle_max = 0.0;
+    test_angle_min = 0.0;
+    rotation_rate_filt.reset(0.0);
```
rmackay9, line 1228: Hi @lthall,

I think there may be an issue here with the rotation_rate_filt.reset(0.0) call because it is called juste a few lines higher up where it may set the rate to the "start_rate".  That line has no effect with this change

lthall: Thanks, fixed.

rmackay9 (approved):

PR #27373, opened 2024-06-24
Constellation choice for septentrio gnss

chiara-septentrio:
# Changes
## General overview
This pull request adds the feature to chose the system used for Septentrio receivers through the GPS_GNSS_MODE parameter. 
Any configuration is possible outside of IMES only (see below) and will take effect on start-up.

## Special case : IMES
The IMES system is not implemented as there is no feature for it on the Septentrio receivers. The same way, NavIc and L-Band that can be configured on the receiver but are not present in the current parameters are not implemented.
If the user choses to use only IMES, it is treated as an absence of choice and the system is left as is.

## SBAS parameter 
The Septentrio driver also uses the GPS_SBAS_MODE parameter, this parameter takes precedence over this new feature. To take only this new feature into account, the parameter can be left as NoChange. Redundant choices should not cause any issue.

## Tests
This new feature has been tested with the CubeOrangePlus connected to one mosaic-go through its com2 port, and the copter build. It has been tested with all possible configurations. It has also been tested to work as intended with the SBAS parameter.
RxControl has been used to monitor the receiver and confirm the drop in satellites used corresponded to the right systems.

  Example of corresponding parameter and receiver configuration:
image
*Parameter on mission planner*

image
*Configuration on RxControl*

### DIFF
--- a/libraries/AP_GPS/AP_GPS_SBF.cpp
+++ b/libraries/AP_GPS/AP_GPS_SBF.cpp
@@ -138,6 +138,19 @@ AP_GPS_SBF::read(void)
                                     config_string = nullptr;
                                 }
                                 break;
+                            case Config_State::Constellation:
+                                if ((params.gnss_mode&0x6F)!=0) {
+                                    //IMES not taken into account by Septentrio receivers
+                                    if (asprintf(&config_string, "sst, %s%s%s%s%s%s\n", (params.gnss_mode&(1U<<0))!=0 ? "GPS" : "",
+                                                            (params.gnss_mode&(1U<<1))!=0 ? ((params.gnss_mode&0x01)==0 ? "SBAS" : "+SBAS") : "",
+                                                            (params.gnss_mode&(1U<<2))!=0 ? ((params.gnss_mode&0x03)==0  ? "GALILEO" : "+GALILEO") : "",
+                                                            (params.gnss_mode&(1U<<3))!=0 ? ((params.gnss_mode&0x07)==0 ? "BEIDOU" : "+BEIDOU") : "",
+                                                            (params.gnss_mode&(1U<<5))!=0 ? ((params.gnss_mode&0x0F)==0 ? "QZSS" : "+QZSS") : "",
+                                                            (params.gnss_mode&(1U<<6))!=0 ? ((params.gnss_mode&0x2F)==0  ? "GLONASS" : "+GLONASS") : "") == -1) {
+                                        config_string=nullptr;
+                                    }
+                                }
+                                break;
                             case Config_State::Blob:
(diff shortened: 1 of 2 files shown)


andyp1per (changes requested):
Commits need to be squashed into a single commit with a commit message that starts with "AP_GPS: "

chiara-septentrio:
The commit issue has been amended

chiara-septentrio:
I updated the pull request to the main, would it be possible to have a new review @andyp1per ?

andyp1per (approved):
Looks ok to me - what testing has been done?

chiara-septentrio:
I've tested it with a CubeOrangePlus and mosaic-go, all configurations worked as expected (if you want to test it yourself, just note GLONASS alone will not provide PVT but it's expected with the Seprentrio receivers)

libraries/AP_GPS/AP_GPS_SBF.cpp
```diff
@@ -138,6 +138,19 @@ AP_GPS_SBF::read(void)
...
+                            case Config_State::Constellation:
+                                if ((params.gnss_mode&0x6F)!=0) {
+                                    //IMES not taken into account by Septentrio receivers
+                                    if (asprintf(&config_string, "sst, %s%s%s%s%s%s\n", (params.gnss_mode&(1U<<0))!=0 ? "GPS" : "",
+                                                            (params.gnss_mode&(1U<<1))!=0 ? ((params.gnss_mode&0x01)==0 ? "SBAS" : "+SBAS") : "",
+                                                            (params.gnss_mode&(1U<<2))!=0 ? ((params.gnss_mode&0x03)==0  ? "GALILEO" : "+GALILEO") : "",
+                                                            (params.gnss_mode&(1U<<3))!=0 ? ((params.gnss_mode&0x07)==0 ? "BEIDOU" : "+BEIDOU") : "",
+                                                            (params.gnss_mode&(1U<<5))!=0 ? ((params.gnss_mode&0x0F)==0 ? "QZSS" : "+QZSS") : "",
+                                                            (params.gnss_mode&(1U<<6))!=0 ? ((params.gnss_mode&0x2F)==0  ? "GLONASS" : "+GLONASS") : "") == -1) {
+                                        config_string=nullptr;
```
tridge, line 150: this looks like a memory leak, we should add a private fn free_config_string() which does this:

 ```
 void AP_GPS_SBF::free_config_string(void)
{
  free(config_string);
  config_string = nullptr;
}
```

tridge (changes requested):

peterbarker (approved):

tridge (approved):

PR #27378, opened 2024-06-24
AP_Generator: Add hook for error_code to LUA

EosBandi:
Expose AP_Generator error_code to LUA script,
- Allows script based management guide
- Dynamic speed change based on Fuel Cell state


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3571,3 +3571,11 @@ function visual_odom:healthy() end
 -- visual odometry quality as a percentage from 1 to 100 or 0 if unknown
 ---@return integer
 function visual_odom:quality() end
+
+--Gnerator object
+--@class generator
+generator = {}
+
+-- returns the generator error code
+--- @return uint32_t_ud
```
IamPete1, line 3580:
```suggestion
---@return uint32_t_ud
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3571,3 +3571,11 @@ function visual_odom:healthy() end
 -- visual odometry quality as a percentage from 1 to 100 or 0 if unknown
 ---@return integer
 function visual_odom:quality() end
+
+--Gnerator object
+--@class generator
```
IamPete1, line 3576:
```suggestion
--Generator object
```
Only need the class thing on userdata.

EosBandi:
@IamPete1 Thanks, I was pulling my hair to find why script test failed on docs check, but it seems it was only a space....

libraries/AP_Generator/AP_Generator_IE_FuelCell.h
```diff
@@ -26,6 +26,9 @@ class AP_Generator_IE_FuelCell : public AP_Generator_Backend
     // Update fuel cell, expected to be called at 20hz
     void update(void) override;
 
+    // Allow generator error code to pass up to frontend and export to LUA
+    uint32_t get_errorcode(void) override {return _err_code;}
```
peterbarker, line 30: I'm *really* not a fan of returning driver-specifc error codes in a `uint32_t`.

Options include
 - creating an enumeration in `AP_Generator` which describes all possible bits which could be set, and which the backends can supply
 - create singletons for the different generator backends individually and add bindings for those singletones.

peterbarker (changes requested):

PR #27379, opened 2024-06-24
Rover: call stats set_flying from 1Hz loop

IamPete1:
Rover had a `stats_update` function, but its not called from anywhere. Rather than keeping the function this moves the one line that was in it to the 1Hz loop.

### DIFF
--- a/Rover/Rover.cpp
+++ b/Rover/Rover.cpp
@@ -293,17 +293,6 @@ void Rover::nav_script_time_done(uint16_t id)
 }
 #endif // AP_SCRIPTING_ENABLED
 
-#if AP_STATS_ENABLED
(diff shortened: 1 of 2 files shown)


rmackay9 (approved):

rmackay9:
@peterbarker, you're happy with this?  It looks like some vehicle's call the set_flying() directly while others (like tracker) have it in its own function.

peterbarker:
Merged, thanks!

peterbarker:
> @peterbarker, you're happy with this? It looks like some vehicle's call the set_flying() directly while others (like tracker) have it in its own function.

Can't complain with the "it now works" aspect of the PR :-)

PR #27380, opened 2024-06-24
AP_Arming: Change the initialization of the string buffer

muramura:
It's not necessary to clear the entire string buffer with zeros.
It's sufficient to set the first byte of the string buffer to NULL.
This method can free up 8 bytes of FLUSH memory.

STM32 has limited FLUSH memory.
I want to prioritize practical benefits over appearance.

AFTER
UILD SUMMARY
Build directory: /home/muramura/work2/ardupilot/build/fmuv3
bin/arducopter   1550208      3988   192768               1554196          526560  Not Applicable   

BEFORE
BUILD SUMMARY
Build directory: /home/muramura/work2/ardupilot/build/fmuv3
bin/arducopter   1550284      3988   192772               1554272          526480  Not Applicable

### DIFF
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -44,7 +44,8 @@ bool AP_Arming_Copter::run_pre_arm_checks(bool display_failure)
     }
 
     // always check motors
-    char failure_msg[50] {};
+    char failure_msg[50];
+    failure_msg[0] = 0;
     if (!copter.motors->arming_checks(ARRAY_SIZE(failure_msg), failure_msg)) {
         check_failed(display_failure, "Motors: %s", failure_msg);
         passed = false;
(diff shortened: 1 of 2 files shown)


peterbarker:
If memory serves this buffer gets memcpy'd into the packet we send on the network, leaking information from the autopilot.

Can you confirm you have checked this doesn't happen, please?

peterbarker (approved):
Looking at this, we use these as arguments to a varargs snprintf-like function.

That means we only require these to be null-terminated, and these changes do actually ensure that's the case.

Saves real numbers of bytes:
```
CubeOrange-periph-heavy  *                                                                     
Durandal                            -64    *           -112    -104              -64    -64    -64
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     -24    *           -64     -64               -32    -32    -32
MatekF405                           -16    *           -32     -32               -16    -24    -24
Pixhawk1-1M-bdshot                  -24                -40     -40               -32    -24    -32
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           -16    *           -40     -32               -16    -16    -24
skyviper-v2450                                         -40                                     
```

I gave it a spin through valgrind and it didn't find anything either.

rmackay9:
I can't say I'm a fan of this pattern really.  What was one line is now two and it's less clear.  Anyway..

peterbarker:
OK, I was convinced by the flash savings.

I was concerned by the change in pattern - leaving the bytes uninitialised on the stack is safe here, but isn't safe most of the time.

tridge was also concerned with that.

Randy doesn't like the change in coding style.

So, I'm sorry, while this *does* achieve your goal of saving some flash space, I think in this case we are going to spend a few bytes because it makes the code slightly safer, doesn't use a pattern we're trying to avoid, and does lead to more code lines (which the reader has to decipher each time to assure themselves it is correct).

Closing this now.

PR #27381, opened 2024-06-24
hwdef: JHEMCU-H743HD correct for rename of W25 dataflash driver

peterbarker:


peterbarker:
Doesn't compile before, does afterwards.

Hwurzburg (approved):

PR #27382, opened 2024-06-25
Tidy GPS pre-arm checks

peterbarker:
Will allow backends to have their own pre-arm checks.

Rename pre-arm checks to be consistent with our libraries.  Note that the pre-arm check doesn't actually check the backends are healthy, despite the previous name...

### DIFF
--- a/libraries/AP_GPS/AP_GPS_DroneCAN.cpp
+++ b/libraries/AP_GPS/AP_GPS_DroneCAN.cpp
@@ -227,24 +227,13 @@ AP_GPS_Backend* AP_GPS_DroneCAN::probe(AP_GPS &_gps, AP_GPS::GPS_State &_state)
 
 bool AP_GPS_DroneCAN::inter_instance_pre_arm_checks(char failure_msg[], uint16_t failure_msg_len)
 {
-    // only do these checks if there is at least one DroneCAN GPS:
-    bool found_uavan_type = false;
-    for (uint8_t i = 0; i < GPS_MAX_RECEIVERS; i++) {
-        const auto type = AP::gps().params[i].type;
-        if (type != AP_GPS::GPS_TYPE_UAVCAN &&
-            type != AP_GPS::GPS_TYPE_UAVCAN_RTK_BASE &&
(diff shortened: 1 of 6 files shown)


libraries/AP_GPS/AP_GPS_DroneCAN.cpp
```diff
@@ -225,8 +225,24 @@ AP_GPS_Backend* AP_GPS_DroneCAN::probe(AP_GPS &_gps, AP_GPS::GPS_State &_state)
     return backend;
 }
 
-bool AP_GPS_DroneCAN::backends_healthy(char failure_msg[], uint16_t failure_msg_len)
+bool AP_GPS_DroneCAN::pre_arm_checks(char failure_msg[], uint16_t failure_msg_len)
 {
+    // only do these checks if there is at least one DroneCAN GPS:
+    bool found_uavan_type = false;
```
tridge, line 231: uavan -> DroneCAN

libraries/AP_GPS/AP_GPS_DroneCAN.cpp
```diff
@@ -225,8 +225,24 @@ AP_GPS_Backend* AP_GPS_DroneCAN::probe(AP_GPS &_gps, AP_GPS::GPS_State &_state)
...
+        }
+        found_uavan_type = true;
+   }
+    if (!found_uavan_type) {
+        return true;
+    }
+
+    // lint parameters and detected node IDs:
     for (uint8_t i = 0; i < GPS_MAX_RECEIVERS; i++) {
         const auto &params_i = AP::gps().params[i];
```
tridge, line 247: maybe check for DroneCAN inside loop, and remove above?

tridge (changes requested):

libraries/AP_GPS/AP_GPS_DroneCAN.cpp
peterbarker: This line's now been removed

libraries/AP_GPS/AP_GPS_DroneCAN.cpp
peterbarker: Done

tridge (approved):

Issue #27320, opened 2024-06-17
ArduPilot should flash green lights when AHRS-is-happy, *not* GPS-is-happy


peterbarker:
## Feature request

**Is your feature request related to a problem? Please describe.**
The most visible sign that a Copter is ready to go is the flashing light.

Which is blue if you're good to arm but don't have a good GPS fix.

Which is flashing-green if you're good to arm and the GPS prearms are passing.

Just because your GPS prearms are passing doesn't mean that your vehicle knows where it is and has set its home location.

The latter would seem to be much more useful information than "have GPS lock".  It should reflect whether the vehicle can do waypoint navigation, not whether the GPS is talking to us and has a reasonable fix.

**Describe the solution you'd like**
Change the AP_Notify library's gps-checks-passing to ahrs-checks-passing-and-home-is-set.

**Describe alternatives you've considered**
None

**Platform**
[  ] All
[  ] AntennaTracker
[ x ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

okok125:
Thank You  Peter BarKer.

rmackay9:
I'm pretty sure that for Copter, it needs to have a good position estimate for the LEDs to go green.  It's not just that it has good GPS signal.

okok125:
This is the vedio that I wanna reflect. 
[https://youtu.be/UFu3tds1VjA](url)

peterbarker:
> I'm pretty sure that for Copter, it needs to have a good position estimate for the LEDs to go green. It's not just that it has good GPS signal.

That depends on which mode you are in.

Try this experiment:
```
./Tools/autotest/sim_vehicle.py -v ArduCopter --rgbled --console
param set SIM_GPS_NUMSATS 3
reboot
```

The EKF will never be happy with less than 6 satellites, even if everything else looks good.  This is a reasonable simulation for someone starting up their vehicle in a marginal GPS environment - 'though the EKF would usually reject for poor GPS error estimates rather than number-of-sats with modern equipment.

Wait for the vehicle to stabilize.

Now flip between `LOITER` and `STABILIZE`.

In `STABILIZE` mode the lights will flash green (rapidly as they will have fix-type 6)
In `LOITER` mode the lights will flash yellow indicating a lack of lock.  Adding an enabled inclusion fence will do the same thing (make the lights flash yellow) as AHRS position is then required.

This also appears in the console:
[...]

PR #27367, opened 2024-06-23
ArduCopter: add Copter flight option for requiring position to arm

peterbarker:

PR #27385, opened 2024-06-25
AP_Notify: flash green lights based off location not GPS

peterbarker:
Your Copter in stabilize mode can flash rapid-green indicating a good DGPS lock or better, and yet your vehicle still doesn't have a good idea of where it is.

That means that your vehicle may end up RTL'ing onto a point somewhere along your flight path, when you were given an all-green indication by teh vehicle before you armed in stabilize.

Past this PR, we require the same GPS as before, but in addition we must have been happy enough with our location to set the navigation origin, *and* currently be able to get a location.

A user will receive slow-flashing blue lights if they can't currently get a location, or the navigation origin isn't set, even if they've got a "good" fix.

We also require a good location to get a solid green light - you will get a solid blue light if you can't get a location and have a navigation origin, even if you have a good GPS lock

Closes https://github.com/ArduPilot/ardupilot/issues/27320

Tested in SITL using `--rgbled`

### DIFF
--- a/libraries/AP_Notify/RGBLed.cpp
+++ b/libraries/AP_Notify/RGBLed.cpp
@@ -20,6 +20,7 @@
 #include <AP_GPS/AP_GPS.h>
 #include "RGBLed.h"
 #include "AP_Notify.h"
+#include <AP_AHRS/AP_AHRS.h>
 
 extern const AP_HAL::HAL& hal;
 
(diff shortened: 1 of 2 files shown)


andyp1per:
I really want this as a pre-arm option as well. I am forever waiting for status in loiter before switching to acro and flying because I want to make sure that RTL will work well.

okok125:
Thank You Peter Barker.

rmackay9 (approved):
LGTM.

Now we can look forward to the support calls, "I upgraded to 4.6 and I can't get a GPS lock any more".  Despite this I still think it's a good change, thanks!

peterbarker:
> I really want this as a pre-arm option as well. I am forever waiting for status in loiter before switching to acro and flying because I want to make sure that RTL will work well.

https://github.com/ArduPilot/ardupilot/pull/27367

You're welcome ;-)

rmackay9:
Hi @peterbarker,

This wiki page should probably be updated https://ardupilot.org/copter/docs/common-leds-pixhawk.html

I've added an item to my 4.6.0 release warnings text document so when it comes time to do the beta testing and final release I'll add a warning to the blog post and release announcements

okok125:
😃 so Happy

Randy Mackay ***@***.***> 於 2024年6月26日 週三 上午10:25 寫道：

Hwurzburg:
@peterbarker this seems incomplete...what about systems using LEDA and LEDB??this only affects RGB notify leds

PR #27387, opened 2024-06-25
Use 64 bit time for PPS interrupt in Ublox driver

bugobliterator:
If using 32 bit time for PPS interrupt the corrected time can be set incorrectly after 2^32 microsecond, and cause ublox redetection code to run again and result in Ublox module to drop out after around 1hour and 12 minutes.


tridge (changes requested):
I'd prefer the muxh simpler fix of: 

```
    _last_pps_time_us = AP_HAL::micros64();
```
in AP_GPS_UBLOX::pps_interrupt, with a comment as to why we can't use the passed in timestamp_us

tridge (approved):

PR #27392, opened 2024-06-26
Fix CubeRed and Dual Core Bootloader issues

bugobliterator:
* Disable second core by default
* Fix bug in dual usb bootloader related to timeout.
* Rebuild bootloaders for dual core Cubes


Tools/ardupilotwaf/boards.py
```diff
@@ -286,15 +286,15 @@ def configure_env(self, cfg, env):
 
         if cfg.env.DEBUG:
             env.CFLAGS += [
-                '-g',
+                '-g3',
```
tridge, line 289: check works on clang? or enable only on gcc

libraries/AP_HAL_ChibiOS/hwdef/common/board.c
```diff
@@ -269,6 +269,30 @@ void __early_init(void) {
                      STM32_NOCACHE_MPU_REGION_2_SIZE |
                      MPU_RASR_ENABLE);
 #endif
+#if defined(DUAL_CORE)
```
tridge, line 272: not critical, but prefer a fn in stm32_util.c

tridge (changes requested):

Tools/ardupilotwaf/boards.py
bugobliterator: --debug build added to CI check.

libraries/AP_HAL_ChibiOS/hwdef/common/board.c
bugobliterator: done

tridge (approved):

PR #27394, opened 2024-06-26
AP_Notify: Perform common checks first

muramura:
The two if statements are making the same judgment.
Refactor this code to consolidate the common judgment and clarify the differing judgments.


peterbarker (approved):
LGTM

PR #27396, opened 2024-06-27
Copter: Remove Payload release on thrust loss detection (attempt2)

rmackay9:
This is a replacement for PR https://github.com/ArduPilot/ardupilot/pull/27339 and attempts to resolve the concerns raised with simply removing the option:

1. Arming check added if the user has set the FLIGHT_OPTIONS bit for "Release gripper on thrust loss"
2. Scripting gripper bindings added
3. Example script that releases the gripper on thrust loss (or something close it)

I've done some simple testing in SITL to confirm the arming check works
image

.. and to confirm that the new script does release the gripper on thrust loss or more accurately when these cases are true for a full second:

- the vehicle is armed
- "thrust boost" is active
- the vehicle is descending
- the gripper is "grabbed" (e.g. not already released)

I tested by setting SIM_ENGINE_MUL = 0.6 which causes the quadcopter to lose altitude and hit the ground.
image


ArduCopter/Parameters.cpp
```diff
@@ -1000,7 +1000,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Param: FLIGHT_OPTIONS
     // @DisplayName: Flight mode options
     // @Description: Flight mode specific options
-    // @Bitmask: 0:Disable thrust loss check, 1:Disable yaw imbalance warning, 2:Release gripper on thrust loss, 3:Require position for arming
+    // @Bitmask: 0:Disable thrust loss check, 1:Disable yaw imbalance warning, 2:Release gripper on thrust loss (deprecated), 3:Require position for arming
```
tpwrules, line 1003:
```suggestion
    // @Bitmask: 0:Disable thrust loss check, 1:Disable yaw imbalance warning, 2:Release gripper on thrust loss (removed), 3:Require position for arming
```

If it were deprecated, it would still work but issue a warning.

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -789,6 +789,13 @@ singleton AP_Winch method release_length void float'skip_check
 singleton AP_Winch method set_desired_rate void float'skip_check
 singleton AP_Winch method get_rate_max float
 
+include AP_Gripper/AP_Gripper.h
+singleton AP_Gripper depends AP_GRIPPER_ENABLED
+singleton AP_Gripper rename gripper
+singleton AP_Gripper method grab void
+singleton AP_Gripper method release void
+singleton AP_Gripper method released boolean
```
tpwrules, line 797: There are locking issues with using the gripper from Lua. Releasing the gripper (with the same issues) is already possible with something like `rc:run_aux_function(19, 2)`. You could do an edge detect on the thrust loss instead of checking if the gripper is already released.

Can we use already existing functionality instead of adding a new problematic binding? We can come back once the issues get addressed.

tpwrules (changes requested):

ArduCopter/Parameters.cpp
rmackay9: ok, changed to (removed in 4.6+).  By the way there are other places (like in the enum and warning to user) which still uses the word "deprecated" but I've left these.  I think it's clear to the user what the issue is despite being possibly slightly incorrect wording.

libraries/AP_Scripting/generator/description/bindings.desc
rmackay9: Hi @tpwrules, thanks for the feedback.

I'm not aware of the locking issues but regarding using edge detection, this also crossed my mind.  The way it is here it will override a pilot which is perhaps not a good thing but I'm not sure it really matters.  I don't think we expect this script to be used much so while I normally strive for perfection, I think it's good enough.

tpwrules: Here is the background: https://github.com/ArduPilot/ardupilot/pull/27365#discussion_r1653164720

If it is a rarely used script then adding a potentially problematic binding just for it seems like an avoidable choice. I mentioned edge detection as a way to avoid needing a binding to check if the gripper is released or not.

ArduCopter/crash_check.cpp
```diff
@@ -96,7 +96,12 @@ void Copter::crash_check()
     }
 }
 
-// check for loss of thrust and trigger thrust boost in motors library
+// thrust_loss_check - returns true when the aircraft detects for 1 second that it is:
+//                      above 90% thrust,
```
IamPete1, line 100:
```suggestion
//                      above 90% demanded thrust or high thrust saturation,,
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -1317,6 +1317,19 @@ function winch:relax() end
 ---@return boolean
 function winch:healthy() end
 
+-- desc gripper object
```
IamPete1, line 1320:
```suggestion
-- Gripper object
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -1317,6 +1317,19 @@ function winch:relax() end
...
+gripper = {}
+
+-- desc move the gripper to the grab position
```
IamPete1, line 1323:
```suggestion
-- Move the gripper to the grab position
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -1317,6 +1317,19 @@ function winch:relax() end
...
+function gripper:grab() end
+
+-- desc move the gripper to the release position
```
IamPete1, line 1326:
```suggestion
-- Move the gripper to the release position
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -1317,6 +1317,19 @@ function winch:relax() end
...
+function gripper:release() end
+
+-- desc returns true if gripper is currently in the released position
```
IamPete1, line 1329:
```suggestion
-- Returns true if gripper is currently in the released position
```

rmackay9:
OK, I'll switch this to use the aux function in order to save flash space.

rmackay9:
I've updated this to use the RC auxiliary function as discussed on the dev call and re-tested and I think it's working OK
image

libraries/AP_Scripting/generator/description/bindings.desc
rmackay9: OK, switched to use the aux function as suggested, thanks!

Tools/autotest/arducopter.py
```diff
@@ -11499,26 +11499,6 @@ def Clamp(self):
 
         self.reboot_sitl()  # because we set home
 
-    def GripperReleaseOnThrustLoss(self):
-        '''tests that gripper is released on thrust loss if option set'''
-
-        self.context_push()
```
peterbarker, line 11505: We can retain this test - all we need to do is `install_example_script_context` here for your new script

peterbarker (changes requested):

rmackay9:
@MattKear is going to review logs from the original customer.

rmackay9:
rebased on master

PR #27398, opened 2024-06-27
AP_Frsky_Telem: avoid use if int16-t read call

peterbarker:
Tested in SITL with gdb

### DIFF
--- a/libraries/AP_Frsky_Telem/AP_Frsky_SPort.cpp
+++ b/libraries/AP_Frsky_Telem/AP_Frsky_SPort.cpp
@@ -43,7 +43,10 @@ void AP_Frsky_SPort::send(void)
     }
 
     for (int16_t i = 0; i < numc; i++) {
-        int16_t readbyte = _port->read();
+        uint8_t readbyte;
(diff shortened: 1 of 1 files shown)


tpwrules:
The number of characters is checked above, is it actually possible for the read to fail here? If so that’s certainly a gotcha that needs to be documented.

peterbarker:
> The number of characters is checked above, is it actually possible for the read to fail here? If so that’s certainly a gotcha that needs to be documented.

Technically yes, but you *really* have to work at it and/or do silly things.  e.g. read from the same port in a higher-priority thread (you might be able to arrange it using mavlink-passthrough, for example).  There's no technical mutex against multiple consumers on a serial port, 'though we do have protections floating around for various scenarios (e.g. that pass-through support)

Still, as I've done in https://github.com/ArduPilot/ardupilot/pull/27397/files , you can close that hole and avoid problems revolving around that int16_t return value at the same time...

tpwrules:
Multiple consumers on a port is definitely a little scary... What do you mean by mavlink passthrough?

tridge (approved):

PR #27399, opened 2024-06-27
Plane: Quadplane: update Q_OPTION bit 22 description

IamPete1:
This updates the description to match what the option actually does.


andyp1per (approved):

Hwurzburg:
@andyp1per this option is still undescribed in the wiki....its on the old 4.5 wiki list as still open....if you could describe WHEN this option should be used, I can add it in the wiki

tridge (approved):

tridge (approved):

PR #27400, opened 2024-06-27
AP_GPS: hdop and vdop are unitless

Williangalvani:
https://en.wikipedia.org/wiki/Dilution_of_precision_(navigation)

### DIFF
--- a/libraries/AP_GPS/AP_GPS.h
+++ b/libraries/AP_GPS/AP_GPS.h
@@ -201,8 +201,8 @@ public:
         float gps_yaw;                      ///< GPS derived yaw information, if available (degrees)
         uint32_t gps_yaw_time_ms;           ///< timestamp of last GPS yaw reading
         bool  gps_yaw_configured;           ///< GPS is configured to provide yaw
-        uint16_t hdop;                      ///< horizontal dilution of precision in cm
-        uint16_t vdop;                      ///< vertical dilution of precision in cm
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

tpwrules:
They don’t have a physical unit, but they are scaled by 100. I.e. if the field has a value of 157, then a GCS would display an hdop of 1.57.

The comment should explain this scaling.

Williangalvani:
> They don’t have a physical unit, but they are scaled by 100. I.e. if the field has a value of 157, then a GCS would display an hdop of 1.57.
> 
> The comment should explain this scaling.

Hah, I had looked at the code for UBX and it sets the hdop directly. but UBX itself has it scaled:
image
I'll update the comment

peterbarker:
> They don’t have a physical unit, but they are scaled by 100. I.e. if the field has a value of 157, then a GCS would display an hdop of 1.57.
> 
> The comment should explain this scaling.

Not sure how "GCS" enters into the equation here.  These variables are internal to ArduPilot; you have to go via the mavlink layer to get data to the GCS (typically).

It looks like the descriptions on the mavlink messages could be improved in a manner similar to this PR, 'though!
```
      <field type="uint16_t" name="eph">GPS HDOP horizontal dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
```

PR #27401, opened 2024-06-27
set the last_updated_ms field of efi_state

berman00:
The last_update_ms variable was never set in the DroneCAN implementation of the Electronic Fuel Injector sensor. This causes the function `bool AP_EFI::is_healthy` to always return false. Solved by adding one line of code to set the variable to the correct value.


peterbarker (approved):
I've fixed the commit message to conform to our standards, and squashed all the commits into one.

Not sure why this was just a draft, I plan on merging it anyway.

berman00:
> I've fixed the commit message to conform to our standards, and squashed all the commits into one.
> 
> Not sure why this was just a draft, I plan on merging it anyway.

Hey! I didn't know you could see it if it was a draft. I wanted to get around to testing is first. Thanks! you saved me some work : )

PR #17461, opened 2021-05-16
AP_Motors: add throttle factor to MotorsMatrix

IamPete1:
This adds throttle factor to motors matrix allowing GC offsets to be accounted for, a better approach than  https://github.com/ArduPilot/ardupilot/pull/17455. The new factors can only be set via scripting and the scripting matrix frame class. They would need to be calculated on a per vehicle basis. 

There is no clever handling of ranges, this should be done prior to setting from the script. If one just wants all out thrust at the expense of the CG offset at high throttle they should be scaled such that the min value is 1. If the offset should be accounted for at full throttle they should be scaled with the max value as 1.

We could potently do a better job of priority, but that would be a extensive re-write, it is expected the throttle factors should be close to 1, any larger deviation should be fixed in hardware.

PR #27402, opened 2024-06-27
Lua Script for copter CoM compensation

Georacer:
This PR is a script that corrects for the CoM discrepancy in a quad-X frame. It is applicable both in Copter and (quad)Plane.
I guess it's what #17461 was envisioned for.

It uses the dynamic mixer matrix to update the front-to-back thrust ratio on the fly.

I've been testing with RealFlight on [this model](https://github.com/ArduPilot/SITL_Models/tree/master/RealFlight/Released_Models/QuadPlanes/Skywalker_X8_QuadPlane), after "bringing it close to reality" by reverting the "cheat" I had made in the motor positions: 
Screenshot from 2024-06-24 19-45-46

The issue with this model is that due to the CoM misplacement, a throttle-up causes a pitch-down. The script corrects the buildup of the pitch I-term and suppresses the pitch-down motion upon throttling up.

The results have been documented in [this blog post](https://discuss.ardupilot.org/t/compensating-for-com-misalignment-with-lua-scripts/120008).

If you think this script is too dangerous to exist in the `applets` directory, I'm happy with it being in the `examples` too.

### DIFF
--- a/libraries/AP_Scripting/applets/x-quad-cg-allocation.lua
+++ b/libraries/AP_Scripting/applets/x-quad-cg-allocation.lua
@@ -1,0 +1,283 @@
+-- x-quad-cg-allocation.lua: Adjust the control allocation matrix for offset CoG.
+--
+-- WARNING: This script is applicable only to X-type quadrotors and quadplanes.
+--
+-- How To Use
+--   1. Place this script in the "scripts" directory.
+--   2. Set FRAME_CLASS or Q_FRAME_CLASS to 17 to enable the dynamic scriptable mixer.
+--   3. Enable Lua scripting via the SCR_ENABLE parameter.
+--   4. Reboot.
+--   5. Fly the vehicle.
+--   6. Adjust the value of the CGA_RATIO parameter.
+--
+-- How It Works
+--   1. The control allocation matrix is adjusted for thrust and pitch based on the ??? parameter value.
+
+--[[
+Global definitions.
+--]]
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+local SCRIPT_NAME = "CoG adjust script"
+local LOOP_RATE_HZ = 10
+local last_warning_time_ms = uint32_t() -- Time we last sent a warning message to the user.
+local WARNING_DEADTIME_MS = 1000 -- How often the user should be warned.
+local is_mixer_matrix_static = false
+local is_mixer_matrix_dynamic = false
+local last_ratio = 1
+
+-- State machine states.
+local FSM_STATE = {
+    INACTIVE = 0,
+    INITIALIZE = 1,
+    ACTIVE = 2,
+    FINISHED = 3
+}
+local current_state = FSM_STATE.INACTIVE
+local next_state = FSM_STATE.INACTIVE
+
+
+--[[
+New parameter declarations
+--]]
+local PARAM_TABLE_KEY = 139
+local PARAM_TABLE_PREFIX = "CGA_"
+
+-- Bind a parameter to a variable.
+function bind_param(name)
+    return Parameter(name)
+end
+
+-- Add a parameter and bind it to a variable.
+function bind_add_param(name, idx, default_value)
+    assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), string.format('Could not add param %s', name))
+    return bind_param(PARAM_TABLE_PREFIX .. name)
+end
+
+-- Add param table.
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 1), SCRIPT_NAME .. ': Could not add param table.')
+
+--[[
+  // @Param: CGA_RATIO
+  // @DisplayName: CoG adjustment ratio
+  // @Description: The ratio between the front and back motor outputs during steady-state hover. Positive when the CoG is in front of the motors midpoint (front motors work harder).
+  // @Range: 0.5 2
+  // @User: Advanced
+--]]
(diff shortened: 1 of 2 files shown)


libraries/AP_Scripting/applets/x-quad-cg-allocation.lua
```diff
@@ -0,0 +1,239 @@
...
+  local success, err = pcall(update)
+  if not success then
+     gcs:send_text(MAV_SEVERITY.EMERGENCY, "Internal Error: " .. err)
+     return protected_wrapper, 1000
+  end
+  return protected_wrapper, 1000.0/LOOP_RATE_HZ
+end
+
+-- Start running update loop
+return protected_wrapper()
```
IamPete1, line 239:
```suggestion
return protected_wrapper()

```

libraries/AP_Scripting/applets/x-quad-cg-allocation.lua
```diff
@@ -0,0 +1,239 @@
...
+    end
+
+    current_state = next_state
+end
+
+--[[
+Main loop function
+--]]
+function update()
+    fsm_step()
```
IamPete1, line 220: You might as well call this directly

IamPete1 (commented):
Dynamic mixer is great for finding the correct value, but it would be nice to transition people onto the normal mixer. Then you can just have a script that runs at boot with no callbacks. I think the `load_factors` call skips the normalization step that we do on a "normal" matrix. So there could be some change in tune.

IamPete1:
> @IamPete1 A static mixer would definitely bring better performance, agreed. But I really wanted to be able to tune on the fly.
> How does the idea that I modify the present script so that it switches between dynamic and static based on a parameter sound?

You could check if you can see the `Motors_dynamic` or `AP_MotorsMatrix` singletons so it would switch based on frame class.

tridge:
this is so common that I wonder if a param in the C++ code is warranted?

Georacer:
> this is so common that I wonder if a param in the C++ code is warranted?

You mean skip the .lua script entirely and implement this in C++? If so, I'd like to have this merged first and then dive into those deep waters 😅.

Georacer:
> Dynamic mixer is great for finding the correct value, but it would be nice to transition people onto the normal mixer. Then you can just have a script that runs at boot with no callbacks. I think the `load_factors` call skips the normalization step that we do on a "normal" matrix. So there could be some change in tune.

How about I add the normalization step to the `AP_MotorsMatrix_Scripting_Dynamic::init()` as well?

IamPete1:
> this is so common that I wonder if a param in the C++ code is warranted?

Using throttle factors does add some inaccuracy in the mixer. We do the full mix and then apply throttle factors afterwards, so there not taken into account for the saturation handling. Unless there is a quite significant issue its probably better not to set them and I would not recommend large values at all. As ever, a hardware fix is better if possible. 

> How about I add the normalization step to the `AP_MotorsMatrix_Scripting_Dynamic::init()` as well?

You would have to do it on every call to `load_factors`. I think its fine that the dynamic mix doesn't do it automatically. You could add it in the script. They thing is that if you do the normalization then you can use the same mix with both scripting mixers.

Georacer:
> > @IamPete1 A static mixer would definitely bring better performance, agreed. But I really wanted to be able to tune on the fly.
> > How does the idea that I modify the present script so that it switches between dynamic and static based on a parameter sound?
> 
> You could check if you can see the `Motors_dynamic` or `AP_MotorsMatrix` singletons so it would switch based on frame class.

@IamPete1 is this what you meant?
```lua
    if MotorsMatrix then
        is_mixer_matrix_static = true
        warn_user("Found static mixer", MAV_SEVERITY.INFO)
    end
```
This check follows the truthy path even if `Q_FRAME_CLASS=1`.
Did you have another check in mind?

Georacer:
@IamPete1 I've pushed another another version.

I've managed to implement dual operation, both dynamic and static scriptable matrices.
The switch is done via the `[Q_]FRAME_CLASS` parameter. I wasn't sure how to achieve it via the singletons.

I've tested it in RealFlight. Some examples:

When `Q_FRAME_CLASS=1`:
Screenshot from 2024-07-02 15-15-32

Operation in `Q_FRAME_CLASS=17`:
Screenshot from 2024-07-02 15-46-10

Operation in `Q_FRAME_CLASS=15`
Screenshot from 2024-07-02 15-49-19

Do you require any more specific testing?

Georacer:
Got another update in, to delete two unused variables.
For some reason `run_lua_language_check.py` didn't catch them.

```bash
❯ ./Tools/scripts/run_lua_language_check.py libraries/AP_Scripting/applets/x-quad-cg-allocation.lua
Initializing ...
>>>>>>>>>>========== 1/2
>>>>>>>>>>>>>>>>>>>> 2/2
Diagnosis completed, no problems found
2 Files checked
```

IamPete1:
> I wasn't sure how to achieve it via the singletons.

I think you can just see if the singleton is nil. Something like:
```
if Motors_dynamic then

else if AP_MotorsMatrix then

end
```

> For some reason `run_lua_language_check.py` didn't catch them.

Yeah, it doesn't get everything. The other one, `run_luacheck` might get them tho.

Georacer:
> I think you can just see if the singleton is nil.
I had tried the following before, but I believe was getting exceptions.

```lua
-- Decide if the UA is using a static mixer matrix.
function inspect_mixer_matrix_static()
    if MotorsMatrix then
        is_mixer_matrix_static = true
        warn_user("Found static mixer", MAV_SEVERITY.INFO)
    end
end

-- Decide if the UA is using a dynamic mixer matrix.
function inspect_mixer_matrix_dynamic()
    if Motors_dynamic then
        is_mixer_matrix_dynamic = true
        warn_user("Found dynamic mixer", MAV_SEVERITY.INFO)
    end
end
```

tridge (approved):

PR #27404, opened 2024-06-27
AP_Scripting: only open the scripts directory on filesystems that support it

andyp1per:


IamPete1:
If we make this change I think we also need to narrow the scope of the error here:

https://github.com/ArduPilot/ardupilot/blob/868a650fba3ab13538b15bbe6c74b49a5224b6c6/libraries/AP_Scripting/AP_Scripting_config.h#L11

It should be `!AP_FILESYSTEM_FILE_WRITING_ENABLED && !defined(HAL_HAVE_AP_ROMFS_EMBEDDED_LUA)`

libraries/AP_Scripting/lua_scripts.cpp
```diff
@@ -504,10 +504,12 @@ void lua_scripts::run(void) {
     // Skip those directores disabled with SCR_DIR_DISABLE param
     uint16_t dir_disable = AP_Scripting::get_singleton()->get_disabled_dir();
     bool loaded = false;
+#if AP_FILESYSTEM_FILE_WRITING_ENABLED
```
IamPete1, line 507: Its not clear to me that his is correct in every case. Obviously loading a script is a read only operation. Also note that `SCRIPTING_DIRECTORY` is a define that can be overridden.

andyp1per: Unfortunately there is no "AP_FILESYSTEM_FILE_OPENDIR_ENABLED", we could define one but it would have the same definition as AP_FILESYSTEM_FILE_WRITING_ENABLED

andyp1per:
> If we make this change I think we also need to narrow the scope of the error here:
> 
> https://github.com/ArduPilot/ardupilot/blob/868a650fba3ab13538b15bbe6c74b49a5224b6c6/libraries/AP_Scripting/AP_Scripting_config.h#L11
> 
> It should be `!AP_FILESYSTEM_FILE_WRITING_ENABLED && !defined(HAL_HAVE_AP_ROMFS_EMBEDDED_LUA)`

I'm not sure HAL_HAVE_AP_ROMFS_EMBEDDED_LUA needs to exist. We already have:
```
#define AP_FILESYSTEM_FILE_READING_ENABLED (AP_FILESYSTEM_FILE_WRITING_ENABLED || AP_FILESYSTEM_ROMFS_ENABLED)
```
So I think we could actually leave this and simply change HAL_HAVE_AP_ROMFS_EMBEDDED_LUA to AP_FILESYSTEM_ROMFS_ENABLED

peterbarker:
> So I think we could actually leave this and simply change HAL_HAVE_AP_ROMFS_EMBEDDED_LUA to AP_FILESYSTEM_ROMFS_ENABLED

Someone may not want to embed LUA scripts but still want to have ROMFS (e.g. defaults.parm files)

libraries/AP_Scripting/lua_scripts.cpp
peterbarker: > Unfortunately there is no "AP_FILESYSTEM_FILE_OPENDIR_ENABLED", we could define one but it would have the same definition as 

I'm in favour of `AP_FILESYSTEM_FILE_OPENDIR_ENABLED` rather than reusing  `AP_FILESYSTEM_FILE_WRITING_ENABLED`

However, I'd really rather handle that string emission more gracefully.  So either don't call the load-all-scripts method if the directory doesn't (or can't) exist, or remove the warning from the load-all-scripts method and put it in the places where it is appropriate to warn.

peterbarker (commented):
You don't actually state what the bug is here.

I'm *guessing* that it's the fact it emits a warning, "Lua: open directory (%s) failed" on boards that can't opendir?  Which boards/filesystems are they, BTW?

andyp1per:
> I'm _guessing_ that it's the fact it emits a warning, "Lua: open directory (%s) failed" on boards that can't opendir? Which boards/filesystems are they, BTW?

Yes that's it - MambaH743v4 was the one I was testing

libraries/AP_Scripting/lua_scripts.cpp
andyp1per: That's what this change does - don't call load-all-scripts. In what way would you do this differently.

nexton-winjeel, line 507: Shouldn't this use `AP_FILESYSTEM_FILE_READING_ENABLED`?

peterbarker:
I've created this alternative PR which takes a different approach: https://github.com/ArduPilot/ardupilot/pull/28012

libraries/AP_Scripting/lua_scripts.cpp
andyp1per: Fixed

andyp1per: Fixed

tridge:
@andyp1per we have merged the alternative from @peterbarker - it's less likely to have unintended side effects

PR #27406, opened 2024-06-28
AP_Periph: add ESC_RATE parameter to RC_OUT peripherals

tpwrules:
Like other vehicles (which use RC_SPEED), ESC_RATE is used to set the PWM output rate for outputs whose functions are set to MotorN so that ESCs can be driven at a fast speed (400Hz, same default as aerial vehicles) while servos still run at normal speed (50Hz, controlled by OUT_RATE).

Tested the PWM outputs with real devices and a scope. Also tested that PWM and DShot600 still fly well.

### DIFF
--- a/Tools/AP_Periph/Parameters.cpp
+++ b/Tools/AP_Periph/Parameters.cpp
@@ -407,6 +407,15 @@ const AP_Param::Info AP_Periph_FW::var_info[] = {
     // @Path: ../libraries/SRV_Channel/SRV_Channels.cpp
     GOBJECT(servo_channels, "OUT",     SRV_Channels),
 
+    // @Param: ESC_RATE
+    // @DisplayName: ESC Update Rate
+    // @Description: Rate in Hz that ESC PWM outputs (function is MotorN) will update at
+    // @Units: Hz
+    // @Range: 50 400
+    // @Increment: 1
+    // @User: Advanced
+    GSCALAR(esc_rate, "ESC_RATE", 400), // effective Copter and QuadPlane default after clamping
+
     // @Param: ESC_PWM_TYPE
     // @DisplayName: Output PWM type
     // @Description: This selects the output PWM type, allowing for normal PWM continuous output, OneShot, brushed or DShot motor output
--- a/Tools/AP_Periph/Parameters.h
+++ b/Tools/AP_Periph/Parameters.h
@@ -95,6 +95,7 @@ public:
         k_param_rangefinder_port1,
         k_param_options,
         k_param_rpm_msg_rate,
+        k_param_esc_rate,
     };
 
     AP_Int16 format_version;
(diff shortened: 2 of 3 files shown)


andyp1per (commented):
This will need careful testing of dshot. set_freq() can easily mess with the DMA setup resulting in dshot issues. In particular the order of calls is important

tpwrules:
How would issues manifest?

magicrub:
I'm not a fan of the param name. Feels like it needs a "PWM" in it instead of RC

tpwrules:
It is slightly wonky but I took the name from the other vehicles. I personally would call it ESC_SPEED to fit with the other parameters on periph.

tpwrules:
Rebased on top of https://github.com/ArduPilot/ardupilot/pull/27444 . It looks like the call to `set_freq` probably belongs before `SRV_Channels::init` as that function makes comments about doing late setup.

In any case, the new version has been tested in flight using both 400Hz Normal PWM and DShot600 with a BLHeli_S ESC and performed great in both cases. I also had a 50Hz servo attached to a different timer on the periph to make sure that functionality still worked. Verified the correct PWM waveforms on the scope too.

Tools/AP_Periph/Parameters.cpp
```diff
@@ -407,6 +407,15 @@ const AP_Param::Info AP_Periph_FW::var_info[] = {
...
     GOBJECT(servo_channels, "OUT",     SRV_Channels),
 
+    // @Param: RC_SPEED
+    // @DisplayName: ESC Update Speed
+    // @Description: Speed in Hz that ESC PWM outputs (function is MotorN) will update at
+    // @Units: Hz
+    // @Range: 50 490
+    // @Increment: 1
+    // @User: Advanced
+    GSCALAR(rc_speed, "RC_SPEED", 490), // Copter and QuadPlane default
```
IamPete1, line 417:
```suggestion
    GSCALAR(rc_speed, "RC_SPEED", 400), // Copter and QuadPlane default
```

Tools/AP_Periph/Parameters.cpp
```diff
@@ -407,6 +407,15 @@ const AP_Param::Info AP_Periph_FW::var_info[] = {
     // @Path: ../libraries/SRV_Channel/SRV_Channels.cpp
     GOBJECT(servo_channels, "OUT",     SRV_Channels),
 
+    // @Param: RC_SPEED
+    // @DisplayName: ESC Update Speed
+    // @Description: Speed in Hz that ESC PWM outputs (function is MotorN) will update at
+    // @Units: Hz
+    // @Range: 50 490
```
IamPete1, line 414:
```suggestion
    // @Range: 50 500
```

Tools/AP_Periph/Parameters.cpp
tridge: maybe ESC_RATE?

tridge (approved):

PR #27408, opened 2024-06-28
AP_GPS: don't consider uBlox PVT time correct unless we have 2D fix

peterbarker:
the time may be set by an offline assistance client and may not be accurate

... this is not my commit, but it's in a WIP branch...

MAVProxy can supply this assistance data, so perhaps this patch is worth looking at?


peterbarker:
```
Author: Andrew Tridgell <andrew@tridgell.net>
Date:   Thu Jun 1 18:12:53 2017 +1000
```

tridge (approved):

PR #27409, opened 2024-06-28
Allow full compass calibration with location sources other than GPS

### LINKED ISSUE #18229: Sub: Yaw drift issue
We can sporadically see a persistent yaw drift despite having a very good magnetic heading.
This happens randomly and seems to away when rebooting.

This has happened both on Navigators and Pixhawk 1

It seems to be related to SUB having INS_GYR_CAL=0 for default, which can lead to significant gyro bias at boot.

We have multiple reports of it happening in 4.0.
Related thread in discuss: https://discuss.bluerobotics.com/t/steady-uncommanded-yaw-in-depth-and-stabilize-modes/10260

As it is not trivial to replicate, I'm not 100% sure if this still happens in master.
We still need a good log with LOG_DISARMED=1 and LOG_REPLAY=1.

Current reports/logs:

[gcelec at discuss.bluerobotics](https://discuss.bluerobotics.com/t/unstable-heading-pitch-and-roll/10384/2)

Williangalvani:
Not having a GPS connected currently causes compass calibration to skip the fix_radius step, which prevents using the WMM in the ekf.

I also added a warning so users can know something is wrong.

this is a step towards fix #18229, and is extracted from #26506

### DIFF
--- a/libraries/AP_Compass/CompassCalibrator.cpp
+++ b/libraries/AP_Compass/CompassCalibrator.cpp
@@ -1067,6 +1067,7 @@ bool CompassCalibrator::fix_radius(void)
 {
     Location loc;
     if (!AP::ahrs().get_location(loc) && !AP::ahrs().get_origin(loc)) {
+        GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "MagCal: No location, fix_radius skipped");
         // we don't have a position, leave scale factor as 0. This
         // will disable use of WMM in the EKF. Users can manually set
         // scale factor after calibration if it is known
--- a/libraries/AP_Compass/CompassCalibrator.cpp
+++ b/libraries/AP_Compass/CompassCalibrator.cpp
@@ -1065,14 +1065,14 @@ bool CompassCalibrator::calculate_orientation(void)
  */
 bool CompassCalibrator::fix_radius(void)
 {
-    if (AP::gps().status() < AP_GPS::GPS_OK_FIX_2D) {
+    Location loc;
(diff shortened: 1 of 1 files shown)


tridge (approved):

libraries/AP_Compass/CompassCalibrator.cpp
```diff
@@ -1065,14 +1065,15 @@ bool CompassCalibrator::calculate_orientation(void)
  */
 bool CompassCalibrator::fix_radius(void)
 {
-    if (AP::gps().status() < AP_GPS::GPS_OK_FIX_2D) {
+    Location loc;
+    if (!AP::ahrs().get_location(loc)) {
```
tridge, line 1069: i think we need to accept either AHRS location or GPS location, so lack of healthy compass doesn't prevent compass calibration to fix the issue

tridge (changes requested):

libraries/AP_Compass/CompassCalibrator.cpp
```diff
@@ -1065,14 +1065,15 @@ bool CompassCalibrator::calculate_orientation(void)
...
+        gcs().send_text(MAV_SEVERITY_WARNING, "MagCal: No location, fix_radius skipped");
```
rmackay9, line 1070: We should probably use GCS_SEND_TEXT and I think we should check the prefix is consistent with other text messages from this library.

libraries/AP_Compass/CompassCalibrator.cpp
tridge: could also do AP::ahrs().get_origin()

tridge (approved):

PR #27411, opened 2024-06-28
AP_Scripting: refresh serial bindings

tpwrules:
Removes some annoyances with unnecessary `uint32_t_ud` and pitfalls with -1 return values. This does technically break compatibility, not sure if there is any policy on that. I tried to edit the scripts in the repo to fix them but user scripts will need to be adjusted in minor ways.

Tested in SITL and on Cube Orange with my REPL script.

### DIFF
--- a/libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
+++ b/libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
@@ -37,13 +37,9 @@ size_t AP_Scripting_SerialAccess::write(const uint8_t *buffer, size_t size)
     return ON_DEVICE_PORT(write, buffer, size);
 }
 
-int16_t AP_Scripting_SerialAccess::read(void)
+bool AP_Scripting_SerialAccess::read(uint8_t &c)
 {
-    uint8_t c;
-    if (read(&c, 1) != 1) {
-        return -1;
-    }
-    return c;
+    return read(&c, 1) > 0;
 }
 
 ssize_t AP_Scripting_SerialAccess::read(uint8_t* buffer, uint16_t count)
--- a/libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
+++ b/libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
@@ -24,9 +24,9 @@ void AP_Scripting_SerialAccess::begin(uint32_t baud)
     }
 }
 
-int32_t AP_Scripting_SerialAccess::write(uint8_t c)
+bool AP_Scripting_SerialAccess::write(uint8_t c)
 {
-    return (int32_t)write(&c, 1); // return value will be 0 or 1
+    return write(&c, 1) > 0;
 }
 
 size_t AP_Scripting_SerialAccess::write(const uint8_t *buffer, size_t size)
--- a/libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
+++ b/libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
@@ -24,9 +24,9 @@ void AP_Scripting_SerialAccess::begin(uint32_t baud)
     }
 }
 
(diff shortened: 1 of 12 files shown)


libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -814,13 +830,13 @@ int lua_serial_readstring(lua_State *L) {
 
     // read up to that number of bytes
     const ssize_t read_bytes = port->read(data, req_bytes);
-    if (read_bytes < 0) {
-        return 0; // error, return nil
+    if (read_bytes >= 0) {
+        // push the buffer as a string, truncated to the number of bytes actually read
+        luaL_pushresultsize(&b, read_bytes);
+    } else {
+        lua_pushnil(L); // error, return nil
```
IamPete1, line 837: You can just return 0 in this case. No need to push nil.

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -802,6 +803,21 @@ int lua_serial_writestring(lua_State *L)
...
+int lua_serial_read(lua_State *L) {
+    binding_argcheck(L, 1);
+
+    AP_Scripting_SerialAccess * port = check_AP_Scripting_SerialAccess(L, 1);
+
+    uint8_t c;
+    if (port->read(c)) {
+        lua_pushinteger(L, c);
+    } else {
+        lua_pushnil(L); // error, return nil
```
IamPete1, line 815: Return 0

libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
```diff
@@ -24,9 +24,9 @@ void AP_Scripting_SerialAccess::begin(uint32_t baud)
     }
 }
 
-size_t AP_Scripting_SerialAccess::write(uint8_t c)
+int32_t AP_Scripting_SerialAccess::write(uint8_t c)
 {
-    return write(&c, 1);
+    return (int32_t)write(&c, 1); // return value will be 0 or 1
```
IamPete1, line 29: Read returns bool, why not here too?

libraries/AP_Scripting/lua_bindings.cpp
tpwrules: We have to push nil specifically so that we return a value. Otherwise e.g. `string.char(port:read())` will return an empty string if there is no data available instead of raising an error that it tried to do something with nil. (Previously, it would raise an error that -1 is out of range).

libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
tpwrules: Would the Lua result also be a bool? Not a bad idea, but Lua bools do not decay to int so it would be harder to e.g. keep a total of bytes sent.

libraries/AP_Scripting/lua_bindings.cpp
tpwrules: Technically this does not actually result in returning nil. It's not a big deal here but is more of a problem for `port:read()`.

libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
tpwrules: On the other hand, in Lua, 0 is truthy so it is probably best to do that so that `if port:write(c) then` works as expected.

IamPete1:
I do think this is a improvement, the key thing we have to decide is if were OK with the breaking change.

tridge (changes requested):
we can't break all the existing scripts that use serial ports

tpwrules:
Yeah, the breakage would be unfortunate.

We could allow ourselves to remove uint32_t_ud return values by adding a metatable to numbers with the same methods. We also could deprecate the current serial `read` and add a new one with a new name.

It's questionable whether these are worth the flash space. They are not a high priority for me right now in any case.

PR #26944, opened 2024-05-01
AP_Mount: update mount_open servo output to work for all backends

rohan-pandeyy:
---
AP_Mount: update mount_open servo output should work for all backends Issue #26892
---

## Changes:
New function in AP_Mount_Backend
- Added a new protected function control_mount_open_servo() to the AP_Mount_Backend class which checks if the mount_open or mount2_open function is assigned and controls the corresponding servo based on the current mount mode.

Update Backend Update functions
- Modified the update() function of each mount backend (e.g., AP_Mount_Servo) to call the control_mount_open_servo() function to ensure that the open/close servo is consistently controlled across all backend implementations.

Removed redundant code
- Removed the code block in AP_Mount_Servo::update() that directly controlled the open servo using move_servo().

PR #27417, opened 2024-06-29
dhruvak001/issue26892

dhruvak001:


rmackay9:
This looks like a mistake so I think I'll close this.  The code change included is already here in this PR I think https://github.com/ArduPilot/ardupilot/pull/26944

PR #27419, opened 2024-06-29
AP_Temperature:expand metadata for analog sensors

Hwurzburg:
NFC....expands metadata which tends to imply only the values listed work


IamPete1 (approved):
It would be great if we could standardize our analog pin descriptions. There are loads of them in various places.

We have "copyfrom" but its not clear what should be the "master copy".

Issue #27418, opened 2024-06-29
Tradheli: After landing in Autotune and motor interlock disables, motor can restart


bnsgeyer:
## Bug report

**Issue details**
User experienced an unexpected motor restart after landing in Autotune (for heli, motor interlock disables once land complete flag is true).  Here is the [User's Post on this event](https://discuss.ardupilot.org/t/copter-wobbling-at-high-head-speed/118575/82)

It was determined that Autotune does not handle the spool state similar to other non-manual throttle flight modes that are not intended for takeoff (like Brake mode).  For heli's, once the aircraft lands, the spool state is set to GROUND_IDLE.  This disables motor interlock in the code while the user's switch still has it enabled.
Here is the lines of code in Mode_Autotune responsible for this behavior
`        if (target_climb_rate < 0.0f) {`
`            copter.motors->set_desired_spool_state(AP_Motors::DesiredSpoolState::GROUND_IDLE);`
`        } else {`
`            copter.motors->set_desired_spool_state(AP_Motors::DesiredSpoolState::THROTTLE_UNLIMITED);`
`        }`

Thus when the user accidentally raised the throttle above center which triggered a positive commanded climb rate, then the motor interlock was enabled again and the motor started.  

I will start a PR with the solution that I think will work for multi and heli.  Just wanted to document this issue.

**Version**
ArduCopter 4.5.1

**Platform**
[  ] All
[  ] AntennaTracker
[X] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
Helicopter - Single

**Hardware type**
CUAV Nora

**Logs**
Link to relevant log is provided in discuss post

bnsgeyer:
fixed in PR #27423
closing

bnsgeyer:
fixed in PR https://github.com/ArduPilot/ardupilot/pull/27423
closing

bnsgeyer:
fixed in PR https://github.com/ArduPilot/ardupilot/pull/27423
closing

PR #27423, opened 2024-06-29
AC_AutoTune: make safe shutdown for tradheli when landing in Autotune

bnsgeyer:
This PR fixes a bug in autotune that was documented in issue #27418.  A tradheli user discovered this bug after landing in autotune and as the aircraft rotor system was spooling down, accidentally raised the collective, commanding positive climb rate.  This caused the motor interlock to engage unintentionally and caused the main gear to be stripped.  This would not usually happen in this case but did because of a bad setting for H_RSC_IDLE.  Regardless, this illustrates the possibility for the rotor to spool up once landed.

The fix to this bug is to change the logic so the interlock can never be enabled after landing in autotune.  In fact this PR causes the aircraft to disarm once ground idle is reached.  So this will affect the behavior for multirotors too.  @lthall I wasn't sure what the original intention of keeping the aircraft from disarming after the user lands in autotune as well as allowing the spool state to go throttle unlimited if the user commands a climb rate.

### DIFF
--- a/ArduCopter/mode.h
+++ b/ArduCopter/mode.h
@@ -184,6 +184,8 @@ public:
     // pause and resume a mode
     virtual bool pause() { return false; };
     virtual bool resume() { return false; };
+
+    // handle situations where the vehicle is on the ground waiting for takeoff
     void make_safe_ground_handling(bool force_throttle_unlimited = false);
 
     // true if weathervaning is allowed in the current mode
--- a/ArduCopter/mode.h
+++ b/ArduCopter/mode.h
@@ -184,6 +184,7 @@ public:
     // pause and resume a mode
     virtual bool pause() { return false; };
     virtual bool resume() { return false; };
+    void make_safe_ground_handling(bool force_throttle_unlimited = false);
 
     // true if weathervaning is allowed in the current mode
 #if WEATHERVANE_ENABLED == ENABLED
@@ -196,7 +197,6 @@ protected:
     bool is_disarmed_or_landed() const;
     void zero_throttle_and_relax_ac(bool spool_up = false);
     void zero_throttle_and_hold_attitude();
-    void make_safe_ground_handling(bool force_throttle_unlimited = false);
 
     // Return stopping point as a location with above origin alt frame
     Location get_stopping_point() const;
--- a/ArduCopter/mode_autotune.cpp
+++ b/ArduCopter/mode_autotune.cpp
@@ -38,30 +38,20 @@ void AutoTune::run()
     // apply SIMPLE mode transform to pilot inputs
     copter.update_simple_mode();
 
-    // reset target lean angles and heading while landed
-    if (copter.ap.land_complete) {
-        // we are landed, shut down
-        float target_climb_rate = get_pilot_desired_climb_rate_cms();
-
-        // set motors to spin-when-armed if throttle below deadzone, otherwise full range (but motors will only spin at min throttle)
-        if (target_climb_rate < 0.0f) {
-            copter.motors->set_desired_spool_state(AP_Motors::DesiredSpoolState::GROUND_IDLE);
-        } else {
-            copter.motors->set_desired_spool_state(AP_Motors::DesiredSpoolState::THROTTLE_UNLIMITED);
-        }
-        copter.attitude_control->reset_rate_controller_I_terms_smoothly();
-        copter.attitude_control->reset_yaw_target_and_rate();
-
-        float target_roll, target_pitch, target_yaw_rate;
-        get_pilot_desired_rp_yrate_cd(target_roll, target_pitch, target_yaw_rate);
-
-        copter.attitude_control->input_euler_angle_roll_pitch_euler_rate_yaw(target_roll, target_pitch, target_yaw_rate);
-        copter.pos_control->relax_z_controller(0.0f);
-        copter.pos_control->update_z_controller();
-    } else {
-        // run autotune mode
-        AC_AutoTune::run();
+    // disarm when the landing detector says we've landed and spool state is ground idle
+    if (copter.ap.land_complete && motors->get_spool_state() == AP_Motors::SpoolState::GROUND_IDLE) {
+        copter.arming.disarm(AP_Arming::Method::LANDED);
     }
+
+    // if not armed set throttle to zero and exit immediately
+    if (copter.ap.land_complete) {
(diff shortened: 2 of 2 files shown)


ArduCopter/mode.h
```diff
@@ -184,6 +184,7 @@ class Mode {
     // pause and resume a mode
     virtual bool pause() { return false; };
     virtual bool resume() { return false; };
+    void make_safe_ground_handling(bool force_throttle_unlimited = false);
```
rmackay9, line 187: Just a small thing but any reason why the declaration is moved higher?  I think it was better in its original place.

ArduCopter/mode_autotune.cpp
```diff
@@ -38,30 +38,20 @@ void AutoTune::run()
     // apply SIMPLE mode transform to pilot inputs
     copter.update_simple_mode();
 
-    // reset target lean angles and heading while landed
+    // disarm when the landing detector says we've landed and spool state is ground idle
```
rmackay9, line 41: disarming when the vehicle becomes landed is available as a flight option but its not the default so I think this would surprise users

ArduCopter/mode.h
bnsgeyer: I couldn't call it from the AC_Autotune library.  So I had to make it public.  Any ideas on how to do it otherwise?

ArduCopter/mode_autotune.cpp
bnsgeyer: Well, we would never allow the user to takeoff in autotune so why would we leave the aircraft in an armed state after landing.  Plus disarming would then save the gains which i think is helpful for the user.

ArduCopter/mode.h
rmackay9: from within mode_autotune.cpp I think it should work if the "copter.flightmode->" prefix is removed.

ArduCopter/mode_autotune.cpp
rmackay9: I think we don't allow users to arm in autotune but they can takeoff in autotune.  It's really more a matter of consistency though.  If we want the vehicle to always disarm while landed we can discuss and perhaps make that change but having different modes behave differently is confusing to the user

bnsgeyer: There isn’t any logic in autotune to support takeoff.

ArduCopter/mode.h
bnsgeyer: I’m pretty sure I tried that first but it doesn’t work because the code is in autotune and not mode_autotune

ArduCopter/mode_autotune.cpp
```diff
@@ -38,30 +38,20 @@ void AutoTune::run()
...
+    }
+
+    // if not armed set throttle to zero and exit immediately
     if (copter.ap.land_complete) {
-        // we are landed, shut down
-        float target_climb_rate = get_pilot_desired_climb_rate_cms();
-
-        // set motors to spin-when-armed if throttle below deadzone, otherwise full range (but motors will only spin at min throttle)
-        if (target_climb_rate < 0.0f) {
-            copter.motors->set_desired_spool_state(AP_Motors::DesiredSpoolState::GROUND_IDLE);
```
bnsgeyer, line 48: @rmackay9 I suppose this would allow users to takeoff.  however this is far from the current code in althold and loiter used for spool state management and takeoff.

 IMHO, I don't think it is a good/safe practice to allow takeoff in autotune. @lthall what do you think?

ArduCopter/mode_autotune.cpp
andyp1per: I tend to agree with @rmackay9 I want disarm to be under my control, not autotune - this would also be a surprise.

lthall:
Autotune does not support take-off because it can twitch at any time. To support take-off we would need to only take-off if autotune has completed.

I don't think shutting down as soon as we land will surprise people given that people have just been watching potentially crash inducing pitch and roll changes. In general I think people would be happy to see the aircraft shut down quickly once on the ground. I can't imagine that people are deliberately doing  touch and go's just after autotune.

Don't Auto modes automatically shut down once landed, even if they are not intended to land? I am thinking of Autotune more of an automated mode than a manual mode. So I am not too concerned that it shuts down automatically like an auto mode.

ArduCopter/mode_autotune.cpp
lthall: The first twitch of autotune can be very large and aggressive. If we let people take-off in autotune and they hover at 1m for a second that first twitch could easily cause a crash.

Given we switch to Autotune from Loiter or Alt_Hold to determine what behaviour we use I don't think we should be starting in Autotune.

ArduCopter/mode_autotune.cpp
lthall: The alternative is to ignore any positive climb rate or switch out of mot_spin_arm after landed but still let the user disarm.

ArduCopter/mode_autotune.cpp
rmackay9: In case it matters, I think this is an existing issue

ArduCopter/mode.h
rmackay9: @bnsgeyer,

OK, you're right that we can't easily remove the prefix without it becoming awkward.

Could you just add this comment above the function?
// handle situations where the vehicle is on the ground waiting for takeoff

bnsgeyer: done.  Thanks!

rmackay9 (approved):

lthall (approved):
Looks good.

rmackay9:
@bnsgeyer, feel free to merge when you like

bnsgeyer:
@rmackay9 is there any desire to back port this to 4.5?

rmackay9:
Hi @bnsgeyer,

Because of the change in behaviour I'd personally prefer it waited until 4.6 (which is 2 or 3 months from starting beta testing) but if we think it is a real hazard then we could backport it.

bnsgeyer:
@rmackay9 No.  I think that is fine.  This latest event was a problem because the user a wrong setting which caused a rapid acceleration of the motor and stripped his main gear.  In most cases, the user would realize that the engine is starting again (which would be a slow acceleration) and disable motor interlock on their transmitter.

PR #27425, opened 2024-06-30
AP_Scripting: ignore hidden Lua files

tpwrules:
On macOS, sometimes ._script.lua is created to store metadata when the user copies script.lua over to their SD card. Previously, the scripting engine would barf since the file is not Lua. Now, these files are ignored.

Also avoids a case where a hidden and valid script might be loaded without the user's knowledge.

Tested on SITL that a hidden file is rejected and non hidden files are still loaded.


IamPete1 (approved):
I have never hit this, but it seems like reasonable precaution.

tpwrules:
I have seen at least one user seeking help on the Discord because they hit the problem with macOS independently.

tridge (approved):

PR #27428, opened 2024-07-01
Plane: support MavLink HEADING_TYPE_DEFAULT in GUIDED_CHANGE_HEADING

timtuxworth:
Once Guided mode has switched to "HEADING" tracking, it can't be switched back to default GUIDED_HEADING_NONE, so using DO_REPOSITION (for example) will not work. This fixes that.

Requires the associated change to ardupilotmega.xml here https://github.com/mavlink/mavlink/pull/2130


peterbarker:
This was done correctly in https://github.com/ArduPilot/ardupilot/pull/28527

PR #27430, opened 2024-07-01
AP_NavEKF3: log mag fusion selection to XKFS

Williangalvani:
This helps us better understand what kind of yaw fusion the ekf is doing.
It was very useful while debugging  #18229

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_MagFusion.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_MagFusion.cpp
@@ -425,12 +425,14 @@ void NavEKF3_core::SelectMagFusion()
     if (dataReady) {
         // use the simple method of declination to maintain heading if we cannot use the magnetic field states
         if(inhibitMagStates || magStateResetRequest || !magStateInitComplete) {
+            magFusionSel = MagFuseSel::FUSE_YAW;
             fuseEulerYaw(yawFusionMethod::MAGNETOMETER);
 
             // zero the test ratio output from the inactive 3-axis magnetometer fusion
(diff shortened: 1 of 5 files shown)


libraries/AP_NavEKF3/AP_NavEKF3_MagFusion.cpp
```diff
@@ -462,6 +475,20 @@ void NavEKF3_core::SelectMagFusion()
     }
 }
 
+#if HAL_LOGGING_ENABLED
+// log changes to magnetometer fusion selection
+void NavEKF3_core::WriteMagFusionSelection()
+{
+    const struct log_XKMF pkt{
```
tridge, line 482: quite an expensive way to log 1 byte, could be a field in XKFS, or could jst document how to tell using XKF3.IYAW vs XKF3.IMX/IMY/IMZ

tridge (changes requested):

libraries/AP_NavEKF3/LogStructure.h
```diff
@@ -15,6 +15,7 @@
     LOG_XKQ_MSG,  \
     LOG_XKT_MSG,  \
     LOG_XKTV_MSG, \
+    LOG_XKMF_MSG, \
```
peterbarker, line 18: No longer required?

peterbarker (changes requested):

libraries/AP_NavEKF3/LogStructure.h
Williangalvani: oops, removed

peterbarker (approved):

PR #25066, opened 2023-09-21
Feature/rover enum class

khancyr:
Add a bit of enum class on Rover. 

Override https://github.com/ArduPilot/ardupilot/pull/20011

gain some space

Add missing TERMINATE handler for fence breach and FS

PR #27431, opened 2024-07-01
Rover: fix ModeReason on fence breaches

Williangalvani:
I am not sure if there's an underlying reason for these to be set to BATTERY_FAILSAFE of it it was a mistake.
If it was on purpose, we should at least add a comment explaining what is going on.


khancyr (approved):
Issue comes from PR : https://github.com/ArduPilot/ardupilot/pull/25066 .
Good catch. We should correct the other too:

rmackay9:
Merged, thanks!

Perhaps as @khancyr says we could do even better by mapping the failsafe_trigger method's failsafe_type argument to a ModeReason::FAILSAFE_RADIO or ModeReason::FAILSAFE_GCS but this PR is still an improvement so I've merged, thanks!

PR #27433, opened 2024-07-02
AP_Scripting: added NMEA 2000 EFI driver

tridge:
this has been tested on a marine engine, and correctly produces key telemetry data

the NMEA_2000.lua module is broken out to make it easier to add other NMEA 2000 based drivers

### DIFF
--- a/libraries/AP_Scripting/drivers/EFI_NMEA2k.lua
+++ b/libraries/AP_Scripting/drivers/EFI_NMEA2k.lua
@@ -1,0 +1,196 @@
+--[[
+   EFI driver using NMEA 2000 marine CAN protocol
+--]]
+
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+
+PARAM_TABLE_KEY = 48
+PARAM_TABLE_PREFIX = "EFI_2K_"
+
+local efi_backend = nil
+local efi_state = EFI_State()
+local cylinder_state = Cylinder_Status()
+if not efi_state or not cylinder_state then
+   return
+end
+
+-- bind a parameter to a variable given
+function bind_param(name)
+    local p = Parameter(name)
+    assert(p, string.format('could not find %s parameter', name))
+    return p
+end
+
+-- add a parameter and bind it to a variable
+function bind_add_param(name, idx, default_value)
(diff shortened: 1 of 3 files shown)


tpwrules:
Is this an 11 or 29 bit protocol? Could it be documented whether it can be mixed with DroneCAN?

tridge:
> Is this an 11 or 29 bit protocol? Could it be documented whether it can be mixed with DroneCAN?

it is 29 bit, and 250kbaud, so incompatible with DroneCAN in multiple ways unfortunately

libraries/AP_Scripting/drivers/EFI_NMEA2k.lua
```diff
@@ -0,0 +1,196 @@
...
+end
+
+-- add a parameter and bind it to a variable
+function bind_add_param(name, idx, default_value)
+    assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), string.format('could not add param %s', name))
+    return bind_param(PARAM_TABLE_PREFIX .. name)
+end
+
+-- Setup EFI Parameters
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 10), 'could not add EFI_2K param table')
```
Georacer, line 31: Is there any drawback in declaring a table for 10 positions and using only 3?

libraries/AP_Scripting/drivers/EFI_NMEA2k.lua
```diff
@@ -0,0 +1,196 @@
...
+         elseif pgn == PGN_ENGINE_PARAM_DYNAMIC then
+            parse_engine_param_dynamic(data)
+         end
+      end
+   end
+
+   return update, 2
+end
+
+return update()
```
Georacer, line 196: Is it worth using the protected wrapper instead?

IamPete1: Not really related to this PR. But I'm not a fan of just putting in the protected wrapper everywhere. With adequate testing there should be no need. We typically also just try the same thing again rather than doing any sort of reset to try and clear the error.

libraries/AP_Scripting/drivers/EFI_NMEA2k.lua
tridge: it consumes a few more bytes, doesn't really matter

rmackay9:
I guess the NMEA200.lua file was intentionally put in the modules directory?

PR #27436, opened 2024-07-03
Replay: add option to stop compass after specified time in ms

bugobliterator:


tridge (changes requested):
lets remove the replay changes as this type of replay hackery doesn't scale (there are many things like this one does with replay)
I think AP_COMPASS_KILL_MAG_ENABLED should be zero by default

peterbarker:
Ping @bugobliterator - you were going to remove the replay stuff from this and just leave the option to kill compasses.

PR #27437, opened 2024-07-03
VN: Add validation to initialization configuration commands

lashVN:
Currently, all configuration commands sent to the VN unit are not validated for a response, leading to difficult-to-diagnose errors when a command doesn't go through or is rejected by the VN. All commands send to the unit (particularly setup commands) should have their acceptance confirmation checked before the driver begins to use the data.


lashVN:
The last commit is purely cosmetic, it looked like styling was inconsistent so I ran clang-format on the files. I couldn't find the file recommended or used project-wide, so I took my best guess based on looking through the code: 

```
BasedOnStyle: Google
IndentWidth: 4
TabWidth: 4
UseTab: Never
ColumnLimit: 100
AllowShortFunctionsOnASingleLine: false
AllowShortIfStatementsOnASingleLine: false
AllowShortLoopsOnASingleLine: false
AlignAfterOpenBracket: Align
AlignConsecutiveAssignments: true
AlignOperands: true
AlignTrailingComments: true
IndentCaseLabels: true
PointerAlignment: Left
SpacesInParentheses: false
SortIncludes: false
IncludeBlocks: Preserve
AccessModifierOffset: -4
```

Let me know if these were incorrect

lashVN:
@IamPete1 

These changes are built upon PR #27285 , so I expect the first few commits which are contained within this PR (and resultant diffs) will be removed when that is merged. If not, I'll remove them after merge.

Otherwise, I've tested these changes for successful arming of a AP connected to a VN, so this should be ready for review.

PR #27441, opened 2024-07-03
AC_AutoTune: (heli-only) Rename and remove some variables

peterbarker:
The first patch just adds _cd to some_cd variables.

The second patch removes some variables which are no longer in use.

Post-the-first-patch, there are no binary changes (below).  The second patch must be correct if everything still compiles.  It saves a few bytes of both flash and (presumably) RAM.

```
Durandal                       *      *           *       *                 *      *      *
Hitec-Airspeed      *                                                            
KakuteH7-bdshot                *      *           *       *                 *      *      *
MatekF405                      *      *           *       *                 *      *      *
Pixhawk1-1M-bdshot             *                  *       *                 *      *      *
f103-QiotekPeriph   *                                                            
f303-Universal      *                                                            
revo-mini                      *      *           *       *                 *      *      *
skyviper-v2450                                    *                              
```


bnsgeyer (approved):
LGTM.

PR #27442, opened 2024-07-03
AP_TECS: Add a separate stall speed parameter used for minimum speed scaling

WickedShell:
When stall prevention is enabled we were scaling from the aircraft's minimum flight speed. However this is normally already picked as being above the stall speed, and for a variety of reasons we may want to pin the aircraft at a higher minimum speed. But if the aircraft was commanded to fly to close to that minimum speed as soon as it banked for a pattern it would command a increase in speed to keep it away from stalling. However if your minimum speed is far from stalling this increase was incorrect. To make it worse what this actually results in happening is an aircraft diving for more speed (over 10 m/s on some aircraft) as well as descending to gain that speed resulting in over 200 foot deviations in altitude control.

This was found on a real vehicle (in a rather alarming flight), I've recreated the behavior in the sim easily, and tested the patch in the sim. The one question I have is do we want to keep this parameter within TECS which is the only user currently, or should it be moved into the vehicle side just incase we want to do more with it in the future?

If the stall speed isn't configured this will result in no change in behavior.

### DIFF
--- a/libraries/AP_TECS/AP_TECS.cpp
+++ b/libraries/AP_TECS/AP_TECS.cpp
@@ -415,7 +415,11 @@ void AP_TECS::_update_speed(float DT)
     if (aparm.stall_prevention) {
         // when stall prevention is active we raise the minimum
         // airspeed based on aerodynamic load factor
-        _TASmin *= _load_factor;
+        if (is_positive(aparm.airspeed_stall)) {
+            _TASmin = MAX(_TASmin, aparm.airspeed_stall*EAS2TAS*_load_factor);
+        } else {
+            _TASmin *= _load_factor;
+        }
     }
 
     if (_TASmax < _TASmin) {
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -296,6 +296,14 @@ const AP_Param::Info Plane::var_info[] = {
     // @User: Standard
     ASCALAR(airspeed_max, "AIRSPEED_MAX",  AIRSPEED_FBW_MAX),
 
+    // @Param: AIRSPEED_STALL
+    // @DisplayName: Stall airspeed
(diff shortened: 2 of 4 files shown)


timtuxworth:
This seems like a very good idea to me.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -283,6 +283,15 @@ const AP_Param::GroupInfo AP_TECS::var_info[] = {
...
     AP_GROUPINFO("HDEM_TCONST", 33, AP_TECS, _hgt_dem_tconst, 3.0f),
 
+    // @Param: STALL_SPEED
+    // @DisplayName: Stall speed used for scaling minimum speed
+    // @Description: If stall prevention is enabled this speed is used to calculate the minimum airspeed while banking. If this is set to 0 then the speed is scaled from the minimum speed. Typically set slightly higher then true stall speed.
+    // @Range: 5.0 75.0
+    // @Units: m/s
+    // @Increment: 1
+    // @User: Advanced
+    AP_GROUPINFO("STALL_SPEED", 34, AP_TECS, _stall_speed, 0),
```
tridge, line 293: nice find!
needs to be an EAS not a TAS to be consistent with other airspeed params, ideally it would be AIRSPEED_STALL and go along with AIRSPEEED_CRUISE etc
also, stall prevention (limiting bank) happens when TECS is not running, so should not be in TECS

tridge (changes requested):

libraries/AP_TECS/AP_TECS.cpp
WickedShell: It is in EAS. When I use it, I multiply by EAS2TAS.

Okay, happy to move where the parameter lives. That is sort of the thought I was having, I only stuck it here as that's what was using it.

WickedShell: Param moved, and retested in SITL.

tridge (approved):

PR #27444, opened 2024-07-03
AP_Periph: ensure the dshot type gets set

tpwrules:
Same sort of difficulty as https://github.com/ArduPilot/ardupilot/pull/27093 and same sort of fix.

Observed problems with a BLHeli_S ESC and DShot600. Applying this patch fixes it.

### DIFF
--- a/Tools/AP_Periph/rc_out.cpp
+++ b/Tools/AP_Periph/rc_out.cpp
@@ -63,6 +63,10 @@ void AP_Periph_FW::rcout_init()
     // run this once and at 1Hz to configure aux and esc ranges
     rcout_init_1Hz();
 
+#if HAL_DSHOT_ENABLED
+    hal.rcout->set_dshot_esc_type(SRV_Channels::get_dshot_esc_type());
+#endif
+
     // setup ESCs with the desired PWM type, allowing for DShot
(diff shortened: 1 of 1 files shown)


andyp1per (approved):

andyp1per:
@tridge this needs to go in the periph stable release

andyp1per:
Thanks @tpwrules ! And thanks for testing!

tpwrules:
Tested also that everything still works fine in Normal mode and at DShot300.

tridge (approved):

Tools/AP_Periph/rc_out.cpp
```diff
@@ -63,6 +63,8 @@ void AP_Periph_FW::rcout_init()
     // run this once and at 1Hz to configure aux and esc ranges
     rcout_init_1Hz();
 
+    hal.rcout->set_dshot_esc_type(SRV_Channels::get_dshot_esc_type());
```
peterbarker, line 66:
```suggestion
#if HAL_DSHOT_ENABLED
    hal.rcout->set_dshot_esc_type(SRV_Channels::get_dshot_esc_type());
#endif
```

this same define guards the code

PR #27445, opened 2024-07-04
AP_Proximity: avoid use of int16_t-read call

peterbarker:
Tested in SITL using gdb and the simulated device


libraries/AP_Proximity/AP_Proximity_TeraRangerTower.cpp
```diff
@@ -62,12 +62,10 @@ bool AP_Proximity_TeraRangerTower::read_sensor_data()
     }
 
     uint16_t message_count = 0;
-    int16_t nbytes = _uart->available();
-
-    while (nbytes-- > 0) {
-        int16_t c = _uart->read();
-        if (c==-1) {
-            return false;
+    for (auto i=0; i<INT16_MAX; i++) {  // process at most n bytes
```
rmackay9, line 65: I guess it might make sense for this to be 1024 as well.  I can imagine being hesitant to include functional changes that would require testing but what we have now is a bit dangerous.

peterbarker: > I guess it might make sense for this to be 1024 as well. I can imagine being hesitant to include functional changes that would require testing but what we have now is a bit dangerous.

Yes, bearing in mind I don't have the hardware to test this on :-)

INT16_MAX isn't disastrous - 32,000 iterations in here won't kill a vehicle.

Now `AP_Proximity_Cygbot_D1` is a much greater concern - it's iterating up to UINT32_MAX, which *will* cause us to fall out of the sky.  It's a good thing out UARTDriver layer is well-behaved :-)

rmackay9 (approved):

libraries/AP_Proximity/AP_Proximity_TeraRangerTower.cpp
tridge: i think 1024 limit would be good

PR #27446, opened 2024-07-04
AP_Logger: remove format sent_mask, use backend formats-sent bitmask

peterbarker:
This stops us tracking formats-written in two places.

The Write() method will now emit formats as required and set the relevant bit in the backend-owned bitmask of formats sent.

Update: evidently this doesn't fix the DataFlash Erase issues...

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -11644,6 +11644,7 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
             self.Clamp,
             self.GripperReleaseOnThrustLoss,
             self.REQUIRE_POSITION_FOR_ARMING,
+            self.LoggingFormat,
         ])
         return ret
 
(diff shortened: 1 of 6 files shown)


peterbarker:
Just to explain what happens now.

When WriteStreaming is called we use the data supplied in the call to allocate a msg_type for the "NAME", and stash away all of the data required to make a LogStructure.

When WritePrioritisedBlock is called we extract the message ID from the block, and ensure that we have written the format out for that ID.  This is true whether we've entered via `WriteBlock` directly (the vast majority of our log calls), or via `WriteV` (e.g. `WriteStreaming`).  

Write_Emit_Fmt calls into the frontend to fill a LogStructure, the frontend will fill it in either from the static list of structures, or from the dynamic list created via the `WriteV`/`WriteStreaming` calls.

peterbarker:
Flight tested on one of my Bixlers and a 5-inch quad.

tridge (approved):

PR #27448, opened 2024-07-04
AP_NavEKF3: remove storedRange member variable if rangefinder measure…

peterbarker:
…ments disabled

```
Board               blimp  bootloader  copter  heli  iofirmware  plane  rover  sub
Durandal            -32    *           -16     -16               -24    -24    -16
KakuteH7-bdshot     -104   *           -112    -104              -104   -112   -112
MatekF405           -72    *           -72     -72               -64    -16    -64
Pixhawk1-1M-bdshot  -96                -96     -88               -88    -24    -88
iomcu                                                *                         
revo-mini           -72    *           -72     -64               -64    -24    -64
```

We do need to come back and clean up these defines.  We should have one define for "do we store rangefinder measurements?" (which we have), but we should also have a "can we directly fuse rangefinder measurements?" and "can we provide rangefinder measurements for optical flow fusion?"

At the moment we essentially only have one define controlling all of these things (`AP_RANGEFINDER_ENABLED`)

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
@@ -394,7 +394,9 @@ void NavEKF3_core::InitialiseVariables()
     storedGPS.reset();
     storedBaro.reset();
     storedTAS.reset();
+#if EK3_FEATURE_RANGEFINDER_MEASUREMENTS
     storedRange.reset();
+#endif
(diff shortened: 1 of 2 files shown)


khancyr (approved):
LGTM

EK3_FEATURE_RANGEFINDER_MEASUREMENTS vs AP_RANGEFINDER_ENABLED is confusing even if they are the same for now

PR #27450, opened 2024-07-04
AC_WPNav: correct calculation of predict-accel when zeroing pilot desired accel

peterbarker:
This is a (currently failing) test which checks what happens if we re-enter LOITER from AUTO while we are in RC failsafe.

This picture shows that we maintain 1.3m/s in LOITER when in RC failsafe following an auto mission.

image

If the set-parameter of `SIM_RC_FAIL` is removed, and the re-setting of the roll input to 1500 is commented back in the test passes (i.e. the vehicle comes to a stop)

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -11552,6 +11552,35 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
         self.context_pop()
         self.reboot_sitl()
 
+    def AutoContinueOnRCFailsafe(self):
+        '''check LOITER when entered after RC failsafe is ignored in auto'''
+        self.set_parameters({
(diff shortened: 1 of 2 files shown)


khancyr:
shouldn't the drone crash from Auto to Loiter in rc failsafe ? there is no throttle so it should go down

peterbarker:
> shouldn't the drone crash from Auto to Loiter in rc failsafe ? there is no throttle so it should go down

Most modes where the user controls throttle that would be true, but loiter mode tries to handle the case where we are in RC failsafe *somehow*.  Some defensive programming, the comments seem to indicate.

I believe it's just the defensive measures had bitrotted over time.

peterbarker:
Merged based on verbal approval from Leonard on DevCall :-)

PR #27452, opened 2024-07-04
AP_TECS: Constrained throttle in airspeedless

Georacer:
This is a throwaway PR to highlight that we don't constrain the output throttle limits when in airspeedless TECS mode.

Is there a reason we don't do that? I have observed in simulation that TECS indeed doesn't respect them when configured for airspeedless flight.

### DIFF
--- a/libraries/AP_TECS/AP_TECS.cpp
+++ b/libraries/AP_TECS/AP_TECS.cpp
@@ -823,7 +823,12 @@ void AP_TECS::_update_throttle_with_airspeed(void)
 #endif
     }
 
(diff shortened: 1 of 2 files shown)


tridge (changes requested):
we need an AP_TECS API for plane to tell TECS the min and max throttle for the next loop, and the plane code should apply that constraint (so applied in all modes except MANUAL mode) and TECS can be aware of the constraint for integrator handling

Georacer:
Closing as it's integrated in #27174.

PR #27456, opened 2024-07-05
SITL: remove duplicated crc code

peterbarker:
same as one we already have

... also the one used in the driver itself


tpwrules:
I'd like to tackle this sort of thing in a bit more cohesive manner: https://github.com/ArduPilot/ardupilot/pull/27355

Don't think this should affect merging of this but if you'd like to weigh in.

peterbarker:
> Don't think this should affect merging of this but if you'd like to weigh in.

I've weighed in.  I suggest you PR a separate branch on top of your existing PR branch which merges the ccitt and xmodem implementations if they are, indeed, duplicates.  Bringing a PR which saves a kilobyte is a good strategy to getting this sort of thing merged IMO.

PR #25736, opened 2023-12-08
AP_Mount: added sending of position and velocity to Siyi

tridge:
This will allow for exif tags in photos with position of aircraft

PR #27457, opened 2024-07-05
AP_Mount: added sending of position data to Siyi gimbal

rmackay9:
This replaces PR https://github.com/ArduPilot/ardupilot/pull/25736

This adds support for sending the vehicle's position and velocity to the siyi gimbal which allows the lat, lon and altitude to be recorded in each image via EXIF.  This has been tested on real hardware and works.  Below is a screen shot from the Siyi manual describing the command structure.

image

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Siyi.cpp
+++ b/libraries/AP_Mount/AP_Mount_Siyi.cpp
@@ -85,7 +85,7 @@ void AP_Mount_Siyi::update()
     // send attitude to gimbal at 10Hz
     if (now_ms - _last_attitude_send_ms > 100) {
         _last_attitude_send_ms = now_ms;
-        send_attitude();
+        send_attitude_position();
     }
 
     // run zoom control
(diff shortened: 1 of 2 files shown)


tridge (approved):

timtuxworth:
I tested this on master. It didn't work for me, all I get is "Mount not healthy". If I roll back to 4.5.4 the Siyi mount works.

rmackay9:
Hi @timtuxworth, thanks very much for the report.  Maybe the older Siyi firmware does not like this new message and fails for some reason.  Could you try upgrading the Siyi A8 to the latest firmware from Siyi?  I tested using an A8 running Siyi gimbal firmware 0.3.7 and camera ver 0.2.6.

timtuxworth:
I was running 0.3.7 and 0.2.6

rmackay9:
Hi @timtuxworth, OK, thanks.  I wonder if perhaps I was using a beta version of the Siyi firmware and not the standard one.  Let me try again.

timtuxworth:
WIth help from @tridge and @rmackay9 - my setup still had SERIAL1_PROTOCOL = 8 (gimbal) even though it wasn't connected. This meant there were two ports (one serial, one net) that used SERIAL1_PROTOCOL = 8. This worked on 4.5.4 but seems to have confused the new master code. I disabled SERIAL1_PROTOCOL (set to 0) and the Siyi A8 mini camera now works.

PR #27458, opened 2024-07-05
4.5 Backport Bootloader Fixes for H7 based devices 

bugobliterator:
* Disable second core to avoid spurious writes to RAM at boot
* Fix CubeRed bootloader dual endpoint update method


tpwrules:
This also has the incorrect description for the required bootloader size (needs to be >= 128K instead of >128K).

rmackay9:
I think we're waiting on this PR to produce a 4.5.5-beta2.  If we don't merge this then I think we have a slight regression on CubeRed.

rmackay9:
I think if we don't have a good reason to update the bootloader then we should leave that change out.

tridge:
we should not change CubeOrange bootloader, its a 743

Tools/AP_Bootloader/support.cpp
```diff
@@ -491,19 +526,27 @@ void port_setbaud(uint32_t baudrate)
   check if flash has any ECC errors and if it does then erase all of
   flash
  */
+#define ECC_CHECK_CHUNK_SIZE (32*sizeof(uint32_t))
 void check_ecc_errors(void)
 {
     __disable_fault_irq();
+    // stm32_flash_corrupt(0x8043200);
```
tridge, line 533: can we add that fn commented out in flash.c?

tridge:
we've decided for the current beta to not do the ECC changes, limit just to the otg2 and the dual-core fixes

bugobliterator:
> we've decided for the current beta to not do the ECC changes, limit just to the otg2 and the dual-core fixes

I have removed the bootloader changes

tridge (approved):

tridge:
needs bootloader updates

bugobliterator:
> needs bootloader updates

updated

rmackay9:
OK, we will do a 4.5.5-beta2!

rmackay9:
This is included in 4.5.5-beta2

PR #27459, opened 2024-07-05
Add validation to initialization configuration commands

lashVN:
Currently, all configuration commands sent to the VN unit are not validated for a response, leading to difficult-to-diagnose errors when a command doesn't go through or is rejected by the VN. All commands send to the unit (particularly setup commands) should have their acceptance confirmation checked before the driver begins to use the data.

### DIFF
--- a/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
+++ b/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
@@ -286,21 +286,18 @@ reset:
     return true;
 }
 
-// Send command to read given register number and wait for response
+// Send command and wait for response
 // Only run from thread! This blocks until a response is received
 #define READ_REQUEST_RETRY_MS 500
-void AP_ExternalAHRS_VectorNav::wait_register_responce(const uint8_t register_num)
+void AP_ExternalAHRS_VectorNav::send_command_blocking()
 {
-    nmea.register_number = register_num;
-
     uint32_t request_sent = 0;
     while (true) {
         hal.scheduler->delay(1);
 
         const uint32_t now = AP_HAL::millis();
         if (now - request_sent > READ_REQUEST_RETRY_MS) {
-            // Send request to read
-            nmea_printf(uart, "$%s%u", "VNRRG,", nmea.register_number);
+            nmea_printf(uart, "$%s", message_to_send);
             request_sent = now;
         }
 
@@ -371,57 +368,58 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
 }
 
 // decode the most recently consumed term
-// returns true if new sentence has just passed checksum test and is validated
+// returns true if new term is valid
 bool AP_ExternalAHRS_VectorNav::decode_latest_term()
 {
+    // Check the first two terms (In most cases header + reg number) that they match the sent
+    // message. If not, the response is invalid.
     switch (nmea.term_number) {
         case 0:
-            if (strcmp(nmea.term, "VNRRG") != 0) {
+            if (strncmp(nmea.term, message_to_send, nmea.term_offset) != 0) {  // ignore sync byte in comparison
+                return false;
+            } else if (strncmp(nmea.term, "VNERR", nmea.term_offset) == 0) {
+                AP_HAL::panic("VectorNav received VN error response");
+            }
+            break;
+        case 1: 
+            if (strncmp(nmea.term, message_to_send + 6,
+                        nmea.term_offset) != 0) {  // Start after "VNXXX,"
                 return false;
             }
             break;
-
-        case 1:
-            if (nmea.register_number != strtoul(nmea.term, nullptr, 10)) {
-                return false;
+        case 2: 
+            if (strncmp(nmea.term, "VN-", 3) == 0) {
+                // This term is the model number
+                strncpy(model_name, nmea.term, sizeof(model_name));
+                break;
             }
             break;
-
-        case 2:
-            strncpy(model_name, nmea.term, sizeof(model_name));
-            break;
-
         default:
-            return false;
+            break;
     }
     return true;
 }
 
-void AP_ExternalAHRS_VectorNav::update_thread()
-{
+void AP_ExternalAHRS_VectorNav::initialize() {
     // Open port in the thread
     uart->begin(baudrate, 1024, 512);
 
-    // Reset and wait for module to reboot
-    // VN_100 takes 1.25 seconds
-    //nmea_printf(uart, "$VNRST");
-    //hal.scheduler->delay(3000);
+    hal.util->snprintf(message_to_send, sizeof(message_to_send), "VNWRG,06,0");
+    send_command_blocking();
 
-    // Stop NMEA Async Outputs (this UART only)
-    nmea_printf(uart, "$VNWRG,6,0");
-
-    // Detect version
     // Read Model Number Register, ID 1
-    wait_register_responce(1);
+    hal.util->snprintf(message_to_send, sizeof(message_to_send), "VNRRG,01");
+    send_command_blocking();
 
     // Setup for messages respective model types (on both UARTs)
     if (strncmp(model_name, "VN-1", 4) == 0) {
-        // VN-100
+        // VN-1X0
         type = TYPE::VN_AHRS;
 
         // This assumes unit is still configured at its default rate of 800hz
-        nmea_printf(uart, "$VNWRG,75,3,%u,14,073E,0004", unsigned(800/get_rate()));
-
+        hal.util->snprintf(message_to_send, sizeof(message_to_send),
+                           "VNWRG,75,3,%u,14,073E,0004", unsigned(800 / get_rate()));
+        send_command_blocking();
     } else {
         // Default to setup for sensors other than VN-100 or VN-110
         // This assumes unit is still configured at its default IMU rate of 400hz for VN-300, 800hz for others
@@ -432,11 +430,19 @@ void AP_ExternalAHRS_VectorNav::update_thread()
         if (strncmp(model_name, "VN-3", 4) == 0) {
             has_dual_gnss = true;
         }
-        nmea_printf(uart, "$VNWRG,75,3,%u,34,072E,0106,0612", unsigned(imu_rate/get_rate()));
-        nmea_printf(uart, "$VNWRG,76,3,%u,4E,0002,0010,20B8,0018", unsigned(imu_rate/5));
+        hal.util->snprintf(message_to_send, sizeof(message_to_send),
+                           "VNWRG,75,3,%u,34,072E,0106,0612", unsigned(imu_rate / get_rate()));
+        send_command_blocking();
+        hal.util->snprintf(message_to_send, sizeof(message_to_send),
+                           "VNWRG,76,3,%u,4E,0002,0010,20B8,0018", unsigned(imu_rate / 5));
+        send_command_blocking();
     }
 
     setup_complete = true;
+}
+
+void AP_ExternalAHRS_VectorNav::update_thread() {
+    initialize();
     while (true) {
         if (!check_uart()) {
             hal.scheduler->delay(1);
--- a/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
+++ b/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
@@ -305,6 +305,10 @@ void AP_ExternalAHRS_VectorNav::send_command_blocking()
         while (nbytes-- > 0) {
             char c = uart->read();
             if (decode(c)) {
+                if (nmea.error_response) {
+                    // Received a valid VNERR. Try to resend after the timeout length
+                    break;
+                }
                 return;
             }
         }
@@ -353,6 +357,7 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
         nmea.checksum = 0;
         nmea.term_is_checksum = false;
         nmea.sentence_done = false;
+        nmea.error_response = false;
         return false;
     }
 
@@ -378,10 +383,14 @@ bool AP_ExternalAHRS_VectorNav::decode_latest_term()
             if (strncmp(nmea.term, message_to_send, nmea.term_offset) != 0) {  // ignore sync byte in comparison
                 return false;
             } else if (strncmp(nmea.term, "VNERR", nmea.term_offset) == 0) {
-                AP_HAL::panic("VectorNav received VN error response");
+                nmea.error_response = true;  // Message will be printed on next term 
             }
             break;
         case 1: 
+            if (nmea.error_response) {
+                GCS_SEND_TEXT(MAV_SEVERITY_CRITICAL, "VectorNav received VNERR code: %s", nmea.term);
+                break;
+            }
             if (strncmp(nmea.term, message_to_send + 6,
                         nmea.term_offset) != 0) {  // Start after "VNXXX,"
                 return false;
--- a/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
+++ b/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
@@ -289,8 +289,14 @@ reset:
 // Send command and wait for response
 // Only run from thread! This blocks and retries until a non-error response is received
 #define READ_REQUEST_RETRY_MS 500
-void AP_ExternalAHRS_VectorNav::send_command_blocking()
+void AP_ExternalAHRS_VectorNav::run_command(const char * fmt, ...)
 {
+    va_list ap;
+
+    va_start(ap, fmt);
+    hal.util->vsnprintf(message_to_send, sizeof(message_to_send), fmt, ap);
+    va_end(ap);
+
     uint32_t request_sent = 0;
     while (true) {
         hal.scheduler->delay(1);
@@ -305,7 +311,7 @@ void AP_ExternalAHRS_VectorNav::send_command_blocking()
         while (nbytes-- > 0) {
             char c = uart->read();
             if (decode(c)) {
-                if (nmea.error_response) {
+                if (nmea.error_response && nmea.sentence_done) {
                     // Received a valid VNERR. Try to resend after the timeout length
                     break;
                 }
@@ -380,18 +386,17 @@ bool AP_ExternalAHRS_VectorNav::decode_latest_term()
     // message. If not, the response is invalid.
     switch (nmea.term_number) {
         case 0:
-            if (strncmp(nmea.term, message_to_send, nmea.term_offset) != 0) {
-                return false;
-            }
             if (strncmp(nmea.term, "VNERR", nmea.term_offset) == 0) {
                 nmea.error_response = true;  // Message will be printed on next term 
+            } else if (strncmp(nmea.term, message_to_send, nmea.term_offset) != 0) {
+                return false;
             }
             return true;
         case 1: 
             if (nmea.error_response) {
                 GCS_SEND_TEXT(MAV_SEVERITY_CRITICAL, "VectorNav received VNERR code: %s", nmea.term);
-            } else if (strncmp(nmea.term, message_to_send + 6,
-                       nmea.term_offset) != 0) {  // Start after "VNXXX,"
+            } else if (strlen(message_to_send) > 6 && 
+                       strncmp(nmea.term, &message_to_send[6], nmea.term_offset != 0)) {  // Start after "VNXXX,"
                 return false;
             }
             return true;
@@ -411,19 +416,15 @@ void AP_ExternalAHRS_VectorNav::initialize() {
     uart->begin(baudrate, 1024, 512);
 
     // Pause asynchronous communications to simplify packet finding
-    hal.util->snprintf(message_to_send, sizeof(message_to_send), "VNASY,0");
-    send_command_blocking();
+    run_command("VNASY,0");
 
     // Stop ASCII async outputs for both UARTs. If only active UART is disabled, we get a baudrate
     // overflow on the other UART when configuring binary outputs (reg 75 and 76) to both UARTs
-    hal.util->snprintf(message_to_send, sizeof(message_to_send), "VNWRG,06,1");
-    send_command_blocking();
-    hal.util->snprintf(message_to_send, sizeof(message_to_send), "VNWRG,06,2");
-    send_command_blocking();
+    run_command("VNWRG,06,0,1");
+    run_command("VNWRG,06,0,2");
 
     // Read Model Number Register, ID 1
-    hal.util->snprintf(message_to_send, sizeof(message_to_send), "VNRRG,01");
-    send_command_blocking();
+    run_command("VNRRG,01");
 
     // Setup for messages respective model types (on both UARTs)
     if (strncmp(model_name, "VN-1", 4) == 0) {
@@ -431,9 +432,7 @@ void AP_ExternalAHRS_VectorNav::initialize() {
         type = TYPE::VN_AHRS;
 
         // This assumes unit is still configured at its default rate of 800hz
-        hal.util->snprintf(message_to_send, sizeof(message_to_send),
-                           "VNWRG,75,3,%u,14,073E,0004", unsigned(800 / get_rate()));
-        send_command_blocking();
+        run_command("VNWRG,75,3,%u,14,073E,0004", unsigned(800 / get_rate()));
     } else {
         // Default to setup for sensors other than VN-100 or VN-110
         // This assumes unit is still configured at its default IMU rate of 400hz for VN-300, 800hz for others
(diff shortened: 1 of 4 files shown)


IamPete1:
I'm not a fan of the formatting changes in the last commit. Its significantly worse in some places. It also hides the core functional changes in this PR, and makes the history more complex if we ever need to look back. If your very keen you could revert the cases where its worse and open it in a new PR with only those changes. Then you can run [size_compare_branches](https://github.com/ArduPilot/ardupilot/blob/master/Tools/scripts/size_compare_branches.py) to prove there is no difference to compiled output.

lashVN:
@IamPete1 

Thanks for taking a look. Which spots are you talking about? I generally tried to come up with a clangformat from what I found in the guide and looking around the codebase. Alternatively, if there is a different clangformat I should use, please let me know.

Most of the changes seem to be 
1. Preceeding underscore for private members
2. Correcting brace newline convention
3. Pointer alignment
4. Assignment operator alignment
5. operator and parenthesis whitespace
6. Statement newline wrapping (from what I could tell the codebase largely has a 100 character column limit? That's smaller than I tend to prefer, but it looked to be used)
7. Case label tabbing (happy to revert, just noticed style guide accepts unindented case labels)

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -316,184 +318,202 @@ void AP_ExternalAHRS_VectorNav::wait_register_responce(const uint8_t register_nu
...
 
     const bool use_uncomp = option_is_set(AP_ExternalAHRS::OPTIONS::VN_UNCOMP_IMU);
 
     {
         WITH_SEMAPHORE(state.sem);
         if (use_uncomp) {
             state.accel = Vector3f{pkt1.uncompAccel[0], pkt1.uncompAccel[1], pkt1.uncompAccel[2]};
-            state.gyro = Vector3f{pkt1.uncompAngRate[0], pkt1.uncompAngRate[1], pkt1.uncompAngRate[2]};
+            state.gyro =
+                Vector3f{pkt1.uncompAngRate[0], pkt1.uncompAngRate[1], pkt1.uncompAngRate[2]};
```
IamPete1, line 492: This is worse.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -316,184 +318,202 @@ void AP_ExternalAHRS_VectorNav::wait_register_responce(const uint8_t register_nu
...
         } else {
             state.accel = Vector3f{pkt1.accel[0], pkt1.accel[1], pkt1.accel[2]};
-            state.gyro = Vector3f{pkt1.gyro[0], pkt1.gyro[1], pkt1.gyro[2]};
+            state.gyro  = Vector3f{pkt1.gyro[0], pkt1.gyro[1], pkt1.gyro[2]};
         }
 
-        state.quat = Quaternion{pkt1.quaternion[3], pkt1.quaternion[0], pkt1.quaternion[1], pkt1.quaternion[2]};
+        state.quat = Quaternion{pkt1.quaternion[3], pkt1.quaternion[0], pkt1.quaternion[1],
+                                pkt1.quaternion[2]};
```
IamPete1, line 499: Again, the wrap here is worse.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -316,184 +318,202 @@ void AP_ExternalAHRS_VectorNav::wait_register_responce(const uint8_t register_nu
...
+        state.velocity      = Vector3f{pkt1.velNED[0], pkt1.velNED[1], pkt1.velNED[2]};
         state.have_velocity = true;
 
-        state.location = Location{int32_t(pkt1.positionLLA[0] * 1.0e7),
-                                  int32_t(pkt1.positionLLA[1] * 1.0e7),
-                                  int32_t(pkt1.positionLLA[2] * 1.0e2),
-                                  Location::AltFrame::ABSOLUTE};
+        state.location =
+            Location{int32_t(pkt1.positionLLA[0] * 1.0e7), int32_t(pkt1.positionLLA[1] * 1.0e7),
+                     int32_t(pkt1.positionLLA[2] * 1.0e2), Location::AltFrame::ABSOLUTE};
```
IamPete1, line 507: This is worse.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -524,43 +544,41 @@ void AP_ExternalAHRS_VectorNav::process_ins_packet1(const uint8_t *b)
...
 
     if (gps.fix_type >= 3 && !state.have_origin) {
         WITH_SEMAPHORE(state.sem);
-        state.origin = Location{int32_t(pkt2.GPS1posLLA[0] * 1.0e7),
-                                int32_t(pkt2.GPS1posLLA[1] * 1.0e7),
-                                int32_t(pkt2.GPS1posLLA[2] * 1.0e2),
-                                Location::AltFrame::ABSOLUTE};
+        state.origin =
+            Location{int32_t(pkt2.GPS1posLLA[0] * 1.0e7), int32_t(pkt2.GPS1posLLA[1] * 1.0e7),
+                     int32_t(pkt2.GPS1posLLA[2] * 1.0e2), Location::AltFrame::ABSOLUTE};
```
IamPete1, line 581: This is worse.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -587,18 +604,19 @@ void AP_ExternalAHRS_VectorNav::process_ahrs_packet(const uint8_t *b)
             state.gyro = Vector3f{pkt.uncompAngRate[0], pkt.uncompAngRate[1], pkt.uncompAngRate[2]};
         } else {
             state.accel = Vector3f{pkt.accel[0], pkt.accel[1], pkt.accel[2]};
-            state.gyro = Vector3f{pkt.gyro[0], pkt.gyro[1], pkt.gyro[2]};
+            state.gyro  = Vector3f{pkt.gyro[0], pkt.gyro[1], pkt.gyro[2]};
         }
 
-        state.quat = Quaternion{pkt.quaternion[3], pkt.quaternion[0], pkt.quaternion[1], pkt.quaternion[2]};
+        state.quat =
+            Quaternion{pkt.quaternion[3], pkt.quaternion[0], pkt.quaternion[1], pkt.quaternion[2]};
```
IamPete1, line 611: This is worse.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -649,54 +667,46 @@ void AP_ExternalAHRS_VectorNav::process_ahrs_packet(const uint8_t *b)
...
-                       state.gyro[0], state.gyro[1], state.gyro[2],
-                       state.quat[0], state.quat[1], state.quat[2], state.quat[3]);
+    AP::logger().WriteStreaming(
+        "EAH3", "TimeUS,Temp,Pres,MX,MY,MZ,AX,AY,AZ,GX,GY,GZ,Q1,Q2,Q3,Q4", "sdPGGGoooEEE----",
+        "F000000000000000", "Qfffffffffffffff", AP_HAL::micros64(), pkt.temp, pkt.pressure * 1e3,
+        use_uncomp ? pkt.uncompMag[0] : pkt.mag[0], use_uncomp ? pkt.uncompMag[1] : pkt.mag[1],
+        use_uncomp ? pkt.uncompMag[2] : pkt.mag[2], state.accel[0], state.accel[1], state.accel[2],
+        state.gyro[0], state.gyro[1], state.gyro[2], state.quat[0], state.quat[1], state.quat[2],
+        state.quat[3]);
+
```
IamPete1, line 677: This is much worse.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -796,14 +804,13 @@ void AP_ExternalAHRS_VectorNav::send_status_report(GCS_MAVLINK &link) const
...
-    const float hgt_gate = 5;
-    const float mag_var = 0;
-    mavlink_msg_ekf_status_report_send(link.get_chan(), flags,
-                                       pkt1.velU/vel_gate, pkt1.posU/pos_gate, pkt1.posU/hgt_gate,
-                                       mag_var, 0, 0);
+    const struct VN_INS_packet1& pkt1 = *(struct VN_INS_packet1*)_last_ins_pkt1;
+    const float vel_gate              = 5;
+    const float pos_gate              = 5;
+    const float hgt_gate              = 5;
+    const float mag_var               = 0;
```
IamPete1, line 811: This is worse.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.h
```diff
@@ -25,80 +25,77 @@
...
-        uint8_t checksum;         // checksum accumulator
-        bool term_is_checksum;    // current term is the checksum
-        bool sentence_valid;      // is current sentence valid so far
-        bool sentence_done;       // true if this sentence has already been decoded
-        uint8_t register_number;  // VectorNAV register number were reading
-    } nmea;
-
+        char term[20];          // buffer for the current term within the current sentence
+        uint8_t term_offset;    // offset within the _term buffer where the next character should be
+                                // placed
```
IamPete1, line 92: This is worse.

IamPete1 (commented):
Highlighting where some the formatting changes have made the code harder to read.

IamPete1:
> Thanks for taking a look. Which spots are you talking about? I generally tried to come up with a clangformat from what I found in the guide and looking around the codebase. Alternatively, if there is a different clangformat I should use, please let me know.

We do have a guide, but we don't enforce all of it strictly on new code. Its generally left up to the person writing the code what they prefer. Unless the formatting is particularity offensive then its better to just leave it rather than having lots of code churn. As such not all libraries use the same pattern, underscores for private members for example. 

I'm not saying that we would not accept a formatting improvement for this library but it should be in a separate PR to any functional changes.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -371,57 +368,67 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
...
+// returns true if new term is valid
 bool AP_ExternalAHRS_VectorNav::decode_latest_term()
 {
+    // Check the first two terms (In most cases header + reg number) that they match the sent
+    // message. If not, the response is invalid.
     switch (nmea.term_number) {
         case 0:
-            if (strcmp(nmea.term, "VNRRG") != 0) {
+            if (strncmp(nmea.term, message_to_send, nmea.term_offset) !=
+                0) {  // ignore sync byte in comparison
```
IamPete1, line 379:
```suggestion
            if (strncmp(nmea.term, message_to_send, nmea.term_offset) != 0) { 
                // ignore sync byte in comparison
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -371,57 +368,67 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
...
                 return false;
+            } else if (strncmp(nmea.term, "VNERR", nmea.term_offset) == 0) {
+                AP_HAL::panic("VectorNav received VN error response");
```
IamPete1, line 382: Panicing is not a great idea, it effectively just gives up. Would you expect to see this in normal operation? In which case it should trigger a pre-arm. Or is it only if there is some edge case and very unlikely? I which case you could throw a internal error.  

Would it be worth trying a reset of the vector nav to see if we could recover?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -371,57 +368,67 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
...
+void AP_ExternalAHRS_VectorNav::initialize() {
     // Open port in the thread
     uart->begin(baudrate, 1024, 512);
 
-    // Reset and wait for module to reboot
-    // VN_100 takes 1.25 seconds
-    //nmea_printf(uart, "$VNRST");
-    //hal.scheduler->delay(3000);
+    // Pause asynchronous communications to simplify packet finding
+    hal.util->snprintf(message_to_send, sizeof(message_to_send), "$VNASY,0");
```
IamPete1, line 409:
```suggestion
    hal.util->snprintf(message_to_send, sizeof(message_to_send), "VNASY,0");
```
`send_command_blocking` adds the `$` for you.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -432,11 +439,22 @@ void AP_ExternalAHRS_VectorNav::update_thread()
...
+        hal.util->snprintf(message_to_send, sizeof(message_to_send),
+                           "VNWRG,75,3,%u,34,072E,0106,0612", unsigned(imu_rate / get_rate()));
+        send_command_blocking();
+        hal.util->snprintf(message_to_send, sizeof(message_to_send),
+                           "VNWRG,76,3,%u,4E,0002,0010,20B8,0018", unsigned(imu_rate / 5));
+        send_command_blocking();
     }
 
+    // Resume asynchronous communications
+    hal.util->snprintf(message_to_send, sizeof(message_to_send), "$VNASY,1");
```
IamPete1, line 451:
```suggestion
    hal.util->snprintf(message_to_send, sizeof(message_to_send), "VNASY,1");
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: complete

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: complete

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: complete

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: @IamPete1 

The only cases I'd expect to see an error in response is 

1. If something is catastrophically wrong with the VN sensor (hardware failure, etc)
2. If the code was incorrectly edited and sent a command that isn't acceptable to the sensor
3. If the AP-configured IMU Rate (via get_rate()) is set too high, we will reply with insufficient baudrate VNERR

I don't think any of those cases would be resolved by resetting the sensor, but I agree especially since (3) is very viable with normal use panicing may not be the not the correct response. Given the failure cases, what would you suggest instead?

IamPete1: OK, in that case, I would print the error specific (`GCS_SEND_TEXT`) so the user will have some feedback. Then you need to make sure that they can't arm. Hopefully the existing health checks are sufficient, if not then update those so they give a "init failed" error. 

Really there are three things that you need to acheave:
- Not prevent the rest of the code running. 
- Give the user a clue so they can track down how to resolve
- Prevent the vehicle flying in a unsafe state

lashVN: Just pushed a change to resolve this issue.

In this change, the system will continue to send the message to the VectorNav every READ_REQUEST_RETRY_MS (500ms). The pre-arm check should continue to show as a failure, because the system will never pass initialization (assuming the VNERR is never resolved) THis is consistent with previous behavior: block initialization and retry until a valid response is received. It will result in the received error message being sent to GCS every 500ms.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -286,28 +286,29 @@ bool AP_ExternalAHRS_VectorNav::check_uart()
     return true;
 }
 
-// Send command to read given register number and wait for response
+// Send command and wait for response
 // Only run from thread! This blocks until a response is received
 #define READ_REQUEST_RETRY_MS 500
-void AP_ExternalAHRS_VectorNav::wait_register_responce(const uint8_t register_num)
+void AP_ExternalAHRS_VectorNav::send_command_blocking()
```
peterbarker, line 292: You're not blocking on sending, you're blocking on the processing of the message.


```suggestion
void AP_ExternalAHRS_VectorNav::process_command()
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -286,28 +286,29 @@ bool AP_ExternalAHRS_VectorNav::check_uart()
     return true;
 }
 
-// Send command to read given register number and wait for response
+// Send command and wait for response
 // Only run from thread! This blocks until a response is received
```
peterbarker, line 290:
```suggestion
// Only run from thread! This blocks and retries until a non-error response is received
```

What if the error is permanent?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -371,57 +373,70 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
...
 bool AP_ExternalAHRS_VectorNav::decode_latest_term()
 {
+    // Check the first two terms (In most cases header + reg number) that they match the sent
+    // message. If not, the response is invalid.
     switch (nmea.term_number) {
         case 0:
-            if (strcmp(nmea.term, "VNRRG") != 0) {
+            if (strncmp(nmea.term, message_to_send, nmea.term_offset) != 0) {  // ignore sync byte in comparison
                 return false;
+            } else if (strncmp(nmea.term, "VNERR", nmea.term_offset) == 0) {
```
peterbarker, line 385: pointless `else`

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -371,57 +373,70 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
...
+                nmea.error_response = true;  // Message will be printed on next term 
             }
             break;
```
peterbarker, line 388: This code would be clearer if you just did `return true;` in these `true` cases instead of `break`ing.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: I am sending the command, and that's the primary action of the method. The blocking on response is secondary functionality to assure that the command I sent was accepted. To me, "process_command" implies I receive a command and I have to manipulate it in some way.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: As discussed above, if the error is permanent then the system will continue to attempt to send the command every 0.5 seconds, receiving a VNERR response and sending that to GCS. In these cases, initialization will not complete and the unit will not arm.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: Complete

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: Compete

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
peterbarker: My objection is that "blocking" when used with "send" is usually waiting for a message to be sent, *not* a response to be received, thus searching for a better term.

I do use `run_command` elsewhere, that might be better?

lashVN: @peterbarker 

I've pushed to address this comment, renaming "send_command_blocking" to "run_command"

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -371,57 +373,66 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
...
             }
-            break;
-
-        case 1:
-            if (nmea.register_number != strtoul(nmea.term, nullptr, 10)) {
+            return true;
+        case 1: 
+            if (nmea.error_response) {
+                GCS_SEND_TEXT(MAV_SEVERITY_CRITICAL, "VectorNav received VNERR code: %s", nmea.term);
+            } else if (strncmp(nmea.term, message_to_send + 6,
```
peterbarker, line 392:
```suggestion
            } else if (strncmp(nmea.term, MIN(message_to_send + 6, strlen(message_to_send))
```

I'm not a fan of the structure of this driver, and I haven't even really understood it yet.  But the code that's currently in here looks fall-out-of-sky dangerous here.  It's probably not possible to get in here with `strlen(message_to_send)` less than 6, but if we *do* get here then bad things may happen.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -371,57 +373,66 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
...
+            return true;
+        case 2: 
+            if (strncmp(nmea.term, "VN-", 3) == 0) {
+                // This term is the model number
+                strncpy(model_name, nmea.term, sizeof(model_name));
+            }
+            return true;
         default:
-            return false;
+            return true;
```
peterbarker, line 404: This deserves a comment - we ignore any terms past the third?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -371,57 +373,66 @@ bool AP_ExternalAHRS_VectorNav::decode(char c)
...
+void AP_ExternalAHRS_VectorNav::initialize() {
     // Open port in the thread
     uart->begin(baudrate, 1024, 512);
 
-    // Reset and wait for module to reboot
-    // VN_100 takes 1.25 seconds
-    //nmea_printf(uart, "$VNRST");
-    //hal.scheduler->delay(3000);
+    // Pause asynchronous communications to simplify packet finding
+    hal.util->snprintf(message_to_send, sizeof(message_to_send), "VNASY,0");
```
peterbarker, line 413: Maybe just `strncpy`?

Similarly elsewhere

peterbarker (changes requested):
Needs to pass CI.  *ALSO* needs to not access an uninitalised string on failure:
image

Please also see https://github.com/peterbarker/ardupilot/pull/new/pr/vectornav-run-command-format on how `run_command` could be done rather more neatly.  The method would need to be marked as taking varargs, however.  I'm still not really happy with the way that command is passed into decode, but I think you'd agree the varargs message_format is rather nicer.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: The prior iteration of this driver only received responses up to 3 terms, meaning it should mark any response it saw above 3 as invalid. This will now receive 5 or 6 term responses, so it shouldn't invalidate the responses longer than 3 terms.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: @peterbarker 

It should never be possible to hit this point with length < 6 (All responses must begin with "VNXXX,", and to hit the term==1, there has to have been something received after the comma)

I'm also not clear that it would be particularly dangerous, would it not just worst case compare to unutilized memory allocated to message_to_send? Of course I'm assuming message_to_send capacity is greater than 6 + nmea.term_offset

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -286,28 +286,29 @@ bool AP_ExternalAHRS_VectorNav::check_uart()
...
-            nmea_printf(uart, "$%s%u", "VNRRG,", nmea.register_number);
+            nmea_printf(uart, "$%s", message_to_send);
             request_sent = now;
         }
 
         int16_t nbytes = uart->available();
         while (nbytes-- > 0) {
             char c = uart->read();
             if (decode(c)) {
+                if (nmea.error_response && nmea.sentence_done) {
```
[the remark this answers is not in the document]
lashVN: @peterbarker 

I don't think I understand this comment. 

I agree that the driver assumes that a blocking call doesn't have an interrupting call. Is that what you mean?

peterbarker (approved):
LGTM

peterbarker:
@IamPete1 are you happy for this to be merged in its current state?

IamPete1:
> @IamPete1 are you happy for this to be merged in its current state?

Yeah, looks OK.

tridge (approved):

PR #24411, opened 2023-07-27
PID notches and D feed-foward

andyp1per:
This is a PR for discussion and experimentation only. We have discussed how notch filtering of the target and error terms might be useful for things like frame resonance and how allowing target D feedforward might bring some of the benefits in control that betaflight see. This change lumps them all together so that they can be tried.

PR #27460, opened 2024-07-06
AC_PID: correct error calculation to use latest target

IamPete1:
This fixes a bug that was introduced in https://github.com/ArduPilot/ardupilot/pull/24411.

Currently the error is calculated before the target filter is applied, this means the error calculation is using the target from the last loop. 

The fix is to do all the target filtering, then calculate the error and do all of the error filtering.

### DIFF
--- a/libraries/AC_PID/AC_PID.cpp
+++ b/libraries/AC_PID/AC_PID.cpp
@@ -201,34 +201,48 @@ float AC_PID::update_all(float target, float measurement, float dt, bool limit, 
     _pid_info.reset = _flags._reset_filter;
     if (_flags._reset_filter) {
         _flags._reset_filter = false;
+
+        // Reset target filter
         _target = target;
-        _error = _target - measurement;
-        _derivative = 0.0f;
-        _target_derivative = 0.0f;
 #if AP_FILTER_ENABLED
         if (_target_notch != nullptr) {
             _target_notch->reset();
             _target = _target_notch->apply(_target);
         }
+#endif
+
+        // Calculate error and reset error filter
+        _error = _target - measurement;
+#if AP_FILTER_ENABLED
         if (_error_notch != nullptr) {
             _error_notch->reset();
             _error = _error_notch->apply(_error);
         }
 #endif
+        // Zero derivatives
+        _derivative = 0.0f;
+        _target_derivative = 0.0f;
+
     } else {
-        float error_last = _error;
-        float target_last = _target;
-        float error = _target - measurement;
+
+        // Apply target filters
+        const float target_last = _target;
 #if AP_FILTER_ENABLED
         // apply notch filters before FTLD/FLTE to avoid shot noise
         if (_target_notch != nullptr) {
             target = _target_notch->apply(target);
         }
+#endif
+        _target += get_filt_T_alpha(dt) * (target - _target);
+
+        // Calculate error and apply error filter
+        const float error_last = _error;
+        float error = _target - measurement;
+#if AP_FILTER_ENABLED
         if (_error_notch != nullptr) {
             error = _error_notch->apply(error);
         }
 #endif
-        _target += get_filt_T_alpha(dt) * (target - _target);
         _error += get_filt_E_alpha(dt) * (error - _error);
 
         // calculate and filter derivative


libraries/AC_PID/AC_PID.cpp
```diff
@@ -201,34 +201,48 @@ float AC_PID::update_all(float target, float measurement, float dt, bool limit,
...
         }
 #endif
+        // Zero derivatives
+        _derivative = 0.0f;
+        _target_derivative = 0.0f;
+
     } else {
-        float error_last = _error;
-        float target_last = _target;
-        float error = _target - measurement;
```
IamPete1, line 221: This is the core of the bug. Here `_target` is the value from the last call. It is only updated using the new target on 231.

IamPete1, line 221: This is the core of the bug. Here `_target` is the value from the last call. It is only updated using the new target on 231.

lthall:
Nice catch!!

lthall (approved):
Well done. I am very happy you noticed this!

The PR looks good!!

libraries/AC_PID/AC_PID.cpp
andyp1per: As far as I can tell what is actually required is a one line change moving the calculation of error after the target filter has been applied? For such a subtle issue I would prefer would try and do the absolute minimum. So basically this:

```
diff --git a/libraries/AC_PID/AC_PID.cpp b/libraries/AC_PID/AC_PID.cpp
index 3f6d22e562..08890a18ad 100644
--- a/libraries/AC_PID/AC_PID.cpp
+++ b/libraries/AC_PID/AC_PID.cpp
@@ -202,7 +202,6 @@ float AC_PID::update_all(float target, float measurement, float dt, bool limit,
     if (_flags._reset_filter) {
         _flags._reset_filter = false;
         _target = target;
-        _error = _target - measurement;
         _derivative = 0.0f;
         _target_derivative = 0.0f;
 #if AP_FILTER_ENABLED
@@ -210,6 +209,9 @@ float AC_PID::update_all(float target, float measurement, float dt, bool limit,
             _target_notch->reset();
             _target = _target_notch->apply(_target);
         }
+#endif
+        _error = _target - measurement;
+#if AP_FILTER_ENABLED
         if (_error_notch != nullptr) {
             _error_notch->reset();
             _error = _error_notch->apply(_error);
@@ -218,12 +220,14 @@ float AC_PID::update_all(float target, float measurement, float dt, bool limit,
     } else {
         float error_last = _error;
         float target_last = _target;
-        float error = _target - measurement;
 #if AP_FILTER_ENABLED
         // apply notch filters before FTLD/FLTE to avoid shot noise
         if (_target_notch != nullptr) {
             target = _target_notch->apply(target);
         }
+#endif
+        float error = _target - measurement;
+#if AP_FILTER_ENABLED
         if (_error_notch != nullptr) {
             error = _error_notch->apply(error);
         }
```
?

IamPete1: You proposed fix is only applying the notch filter to the target, it completely removes the low pass.

The goal is not only to fix the bug but also to comment and structure the code such that were less likely to make a similar mistake again.

andyp1per: I understand the reasoning, but the restructure is not making it clear to me 😛

andyp1per: Or put another way I think your restructure is obfuscating the fix. Note that I am pretty sure Leonard gave me this code, so I don't have an axe to grind here 😉

lthall: While I understand where the principle of minimal changes come from I believe the extra effort in ensuring the code is also clear to the future developers can't be understated.

I think this PR has both fixed the bug and also made it clearer to help prevent problems in the future or simply how easy the code is to read and understand.

andyp1per (approved):

andyp1per:
@IamPete1 explained it to me - it was not what I was thinking

bnsgeyer (approved):
LGTM

rmackay9:

PR #27463, opened 2024-07-06
tools: add name

booska777:


rmackay9:
Hi @booska777,

Can you rebase on master and also fix the commit title to be, "Tools: add name to GitTools" (or something like that.

rmackay9:
Sorry, I tried to resolve the conflict by using the web page but it doesn't work properly it seems (or I don't know how to do it).

PR #27468, opened 2024-07-06
Tools: add name to GIT_Success.txt

sichimur:
Tools: add name to GIT_Success.txt


rmackay9:
Very nicely done, thanks!

PR #27476, opened 2024-07-06
SITL: fix json airspeed

eppravitra:
Airspeed calculation for JSON was not quite working because velocity_air_ef was not updated in SIM_JSON.cpp

### DIFF
--- a/libraries/SITL/SIM_JSON.cpp
+++ b/libraries/SITL/SIM_JSON.cpp
@@ -323,14 +323,15 @@ void JSON::recv_fdm(const struct sitl_input &input)
 
         airspeed_pitot = state.airspeed;
     } else {
+        
+        // wind is not supported yet for JSON sim, assume zero for now        
+        wind_ef.zero(); 
+
+        // velocity relative to airmass in Earth's frame
+        velocity_air_ef = velocity_ef - wind_ef;
+
         // velocity relative to airmass in body frame
-        velocity_air_bf = dcm.transposed() * velocity_ef;
-
-        // airspeed
-        airspeed = velocity_air_bf.length();
-
-        // airspeed as seen by a fwd pitot tube (limited to 120m/s)
-        airspeed_pitot = constrain_float(velocity_air_bf * Vector3f(1.0f, 0.0f, 0.0f), 0.0f, 120.0f);
+        velocity_air_bf = dcm.transposed() * velocity_air_ef;
 
         // airspeed fix for eas2tas
         update_eas_airspeed();


eppravitra:
@tridge, Could you please take a look at this? My Gazebo simulation is not working anymore, and I think I've figured out why.

libraries/SITL/SIM_JSON.cpp
```diff
@@ -323,6 +323,13 @@ void JSON::recv_fdm(const struct sitl_input &input)
 
         airspeed_pitot = state.airspeed;
     } else {
+        
+        //wind is not supported yet JSON sim, assume zero for now        
+        wind_ef = Vector3f(0.0f, 0.0f, 0.0f);
```
IamPete1, line 328:
```suggestion
        // wind is not supported yet JSON sim, assume zero for now        
        wind_ef.zero(); 
```

libraries/SITL/SIM_JSON.cpp
```diff
@@ -323,6 +323,13 @@ void JSON::recv_fdm(const struct sitl_input &input)
...
     } else {
+        
+        //wind is not supported yet JSON sim, assume zero for now        
+        wind_ef = Vector3f(0.0f, 0.0f, 0.0f); 
+
+        // velocity relative to airmass in Earth's frame
+        velocity_air_ef = velocity_ef - wind_ef;
+
         // velocity relative to airmass in body frame
         velocity_air_bf = dcm.transposed() * velocity_ef;
```
IamPete1, line 334:
```suggestion
        velocity_air_bf = dcm.transposed() * velocity_air_ef;
```

eppravitra:
Thanks for the suggestions @IamPete1. I think you are right about the changes. Could you add bug and DevCall labels to this, since Gazebo sim is broken right now?

libraries/SITL/SIM_JSON.cpp
```diff
@@ -323,14 +323,15 @@ void JSON::recv_fdm(const struct sitl_input &input)
...
         airspeed_pitot = state.airspeed;
     } else {
-        // velocity relative to airmass in body frame
-        velocity_air_bf = dcm.transposed() * velocity_ef;
+        
+        // wind is not supported yet for JSON sim, assume zero for now        
+        wind_ef.zero(); 
 
-        // airspeed
-        airspeed = velocity_air_bf.length();
```
peterbarker, line 330: I see other backends still updating these; why do we no longer need to do that here?

eppravitra: The calculation is already done inside update_eas_airspeed(). If left as it is, the variable airspeed will be 
 rewritten anyway. I figure it's better to remove the line to avoid confusion.

peterbarker:
@IamPete1 are you OK with this now?

IamPete1 (approved):
LGTM, thanks.

@peterbarker Yeah, I think its good to merge.

peterbarker:
Merged, thanks!

eppravitra:
Thanks!

Issue #16248, opened 2021-01-06
Plane: reach the target altitude of the next waypoint immediately during AUTO and GUIDED, respecting the limits of the TECS


robustini:
With Ardupilot while performing a waypoint flight, the plane flies along a virtual straight line between the previous and the next waypoint.
In the aeronautical world a new flight level is usually executed immediately.
It would therefore be interesting in Ardupilot there was an option to do the same thing.
Example:

- altitude of the waypoint just reached: 10 meters
- altitude of the waypoint to reach: 110 meters
- distance between the two waypoints: 1000 meters
- cruise nav speed: 20 m/s
- maximum height change (climb rate in this case) set in the TECS: 10 meters per second

Currently the plane would be 500 meters away (halfway) between the two waypoints after 25 seconds at a height of 60 meters.
Instead what it should do with the new feature is to reach an altitude of 110 meters only after 10 seconds, at a distance of 200 meters from the first waypoint.
Same thing will do if the next waypoint is at a lower height.
This is aeronautically more logical, the manned planes behave this way during a flight for example in IFR.
And of course this should be an option that can be activated with a parameter, perhaps that you can choose whether this behavior occurs only during a climb, a descent or both.
This solution could also optimize consumption and it might also be interesting for the Copter.
I'm aware that I can already do it now by adding intermediate waypoints, but it seems absurd to have to waste events for this, and it would get boring on a complex mission.
I ping @tridge and @magicrub because they sure could be involved in this.
I am attaching a simple sketch to better explain what I mean:

![wp](https://user-images.githubusercontent.com/3864712/103776128-4c605a00-502f-11eb-963c-2b297b7cc932.jpg)

dcamatti:
@robustini , I fully agree with your request. I I want an altitude change, especially just after takeoff, I want to get rid of obstacles as soon as possibile using the max climb performance of the UAV. Like it is today I can use the max climb performance only by manually placing an additional wpt close enought to cause the UAV climb at max climb rate. Looking forward to have this implemented. Thanks - Daniele Camatti

peterbarker:
On Wed, 6 Jan 2021, Marco Robustini wrote:

> Instead what it should do with the new feature is to reach an altitude of 110 meters only after 10 seconds, at a distance of 200 meters from the first
> waypoint.

I do like the idea.

I have concerns about changing the way the vehicle executes missions based 
on parameters.

Better to represent this in the mission IMO, in a standardised fashion. 
[...]

PR #27477, opened 2024-07-07
Plane: option to immediately climb in AUTO mode (not doing glide slope)

eppravitra:
"Aviation style" climb profile where the aircraft reaches final altitude way before reaching the desired location. 

This has been suggested before in #16248, and I finally got around to giving it a try.

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1086,6 +1086,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Bitmask: 11: Disable suppression of fixed wing rate gains in ground mode
     // @Bitmask: 12: Enable FBWB style loiter altitude control
     // @Bitmask: 13: Indicate takeoff waiting for neutral rudder with flight control surfaces
+    // @Bitmask: 14: In AUTO - climb to next waypoint altitude immediately instead of linear climb
     // @User: Advanced
     AP_GROUPINFO("FLIGHT_OPTIONS", 13, ParametersG2, flight_options, 0),
 
--- a/ArduPlane/altitude.cpp
+++ b/ArduPlane/altitude.cpp
@@ -81,6 +81,13 @@ void Plane::setup_glide_slope(void)
         break;
 
     case Mode::Number::AUTO:
+
+        //climb without doing glide slope if option is enabled
+        if (!above_location_current(next_WP_loc) && plane.flight_option_enabled(FlightOptions::IMMEDIATE_CLIMB_IN_AUTO)) {
+            reset_offset_altitude();
+            break;
+        }
+
         // we only do glide slide handling in AUTO when above 20m or
         // when descending. The 20 meter threshold is arbitrary, and
         // is basically to prevent situations where we try to slowly
--- a/ArduPlane/defines.h
+++ b/ArduPlane/defines.h
@@ -168,6 +168,7 @@ enum FlightOptions {
     DISABLE_GROUND_PID_SUPPRESSION = (1<<11),
     ENABLE_LOITER_ALT_CONTROL = (1<<12),
     INDICATE_WAITING_FOR_RUDDER_NEUTRAL = (1<<13),
+    IMMEDIATE_CLIMB_IN_AUTO = (1<<14),
 };
 
 enum CrowFlapOptions {


robustini:
Thanks for the implementation attempt but it doesn't work, tested on SITL and it keeps making direct lines between waypoints.

robustini:
A log demonstrating its successful operation.

[00000026.ZIP](https://github.com/user-attachments/files/16118509/00000026.ZIP)

eppravitra:
I forgot to mention that I only did it for the climb.

Descent still works the same way. Descents will always do glide slope. 

The change works regardless of GLIDE_SLOPE_MIN

To enable the behavior, set FLIGHT_OPTIONS bit 14 to 1

shancock884:
> To enable the behavior, set FLIGHT_OPTIONS bit 14 to 1

If you add a bit to FLIGHT_OPTIONS, remember to add some helpful info about it to Parameters.cpp for the wiki!
https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/Parameters.cpp#L1088

eppravitra:
Thanks for the suggestion @shancock884 .

The comment has been added.

peterbarker:
I've marked this for DevCallTopic, and NeedsTesting so it might get tested this weekend.

peterbarker:
@eppravitra why is the parameter `GLIDE_SLOPE_MIN` not sufficient for your purposes?

eppravitra:
Hi @peterbarker ,

Honestly, I totally missed GLIDE_SLOPE_MIN part of the code.

After playing with it, I still stand by my idea that there should be a way to do glide slope during descent and not doing glide slope during climb. 

Setting GLIDE_SLOPE_MIN to 0 effectively disables glide slope for both climb and descent.

I welcome suggestions for better implementation or a better way to organize the parameters though.

robustini:
> Setting GLIDE_SLOPE_MIN to 0 effectively disables glide slope for both climb and descent.

Which is what I had pointed out to you earlier.

peterbarker:
> Honestly, I totally missed GLIDE_SLOPE_MIN part of the code.

Yep, this was also an unfamiliar parameter to me.  tridge knew it existed, 'though!

> After playing with it, I still stand by my idea that there should be a way to do glide slope during descent and not doing glide slope during climb.

I agree.  I think the preference would be to re-use the existing parameter somehow, rather than adding another parameter.

> I welcome suggestions for better implementation or a better way to organize the parameters though.

Two ideas spring to mind.  Negative values mean use for climb only.  Add a `FLIGHT_OPTION` bit to limit it to climb only.

eppravitra:
> Two ideas spring to mind. Negative values mean use for climb only. Add a `FLIGHT_OPTION` bit to limit it to climb only.

Thanks @peterbarker 

I’ll leave it up to you and @tridge to make a call on what is the best way to enable/disable this feature. Just tell me what you guys prefer, and I’m happy to recode.

I’ll admit I do find this a bit confusing. For me, the way I have it (using FLIGHT_OPTIONS) is the least confusing way I can think of.

IamPete1:
We have two bits of mission storage we could use on normal way-points to make this configurable in the mission. Or add a "DO_SET" waypoint for it.

timtuxworth:
Will this work in GUIDED mode? (or if not .. could it?)

tridge:
I'd be happy with a FLIGHT_OPTIONS bit to disable slope climb

ArduPlane/Parameters.cpp
```diff
@@ -1086,6 +1086,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Bitmask: 11: Disable suppression of fixed wing rate gains in ground mode
     // @Bitmask: 12: Enable FBWB style loiter altitude control
     // @Bitmask: 13: Indicate takeoff waiting for neutral rudder with flight control surfaces
+    // @Bitmask: 14: Climb the aircraft without doing slope in AUTO
```
Hwurzburg, line 1089: this description needs work....most dont know "slope" in regards to AUTO

Hwurzburg (changes requested):

ArduPlane/altitude.cpp
```diff
@@ -81,6 +81,13 @@ void Plane::setup_glide_slope(void)
         break;
 
     case Mode::Number::AUTO:
+
+        //climb without doing glide slope if option is enabled
+        if (!above_location_current(next_WP_loc) && plane.flight_option_enabled(FlightOptions::IMMEDIATE_CLIMB_IN_AUTO)){
```
IamPete1, line 86:
```suggestion
        if (!above_location_current(next_WP_loc) && plane.flight_option_enabled(FlightOptions::IMMEDIATE_CLIMB_IN_AUTO)) {
```

tridge:
@eppravitra what testing have you done? do you have a SITL log?

tridge:
@robustini did you test this PR?

ArduPlane/Parameters.cpp
eppravitra: I added more description as you suggested

eppravitra:
> Will this work in GUIDED mode? (or if not .. could it?)

GUIDED already climbs with maximum TECS performance and descends with slope regardless of this PR.

This PR only affects AUTO mode, effectively allowing AUTO mode to behave the same way as what GUIDED is already doing.

eppravitra:
[log_140_2024-7-16-10-59-17.zip](https://github.com/user-attachments/files/16242986/log_140_2024-7-16-10-59-17.zip)


@tridge 
I just ran another SITL. Please see the attachment.

robustini:
> @robustini did you test this PR in its current form?

@tridge yes and it works as it should, it acts in AUTO obviously only on the climb.
Here a recent log based on the last commit of this PR, with various tests enabling and disabling it.
In my opinion it works perfectly, merge can be done and I am closing my related request about this.

[00000029.ZIP](https://github.com/user-attachments/files/16243291/00000029.ZIP)

eppravitra:
>  Obviously a further implementation that also goes to remove the limit imposed by the GLIDE_SLOPE_MIN without having to set it to 0 would be appreciated but still not essential, it is okay as it is.

This code actually does not require you to change GLIDE_SLOPE_MIN to 0. Just enable the flag and it should work

robustini:
> > Obviously a further implementation that also goes to remove the limit imposed by the GLIDE_SLOPE_MIN without having to set it to 0 would be appreciated but still not essential, it is okay as it is.
> 
> This code actually does not require you to change GLIDE_SLOPE_MIN to 0. Just enable the flag and it should work

I don't see this behavior, you can see it from my log.
To speed up the descent I had to set it to 0.

eppravitra:
Oh, I guess we are talking about different things.

My change only concerns about climb.

ArduPlane/Parameters.cpp
```diff
@@ -1086,6 +1086,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Bitmask: 11: Disable suppression of fixed wing rate gains in ground mode
     // @Bitmask: 12: Enable FBWB style loiter altitude control
     // @Bitmask: 13: Indicate takeoff waiting for neutral rudder with flight control surfaces
+    // @Bitmask: 14: Climb the aircraft without doing slope in AUTO. If enabled, aircraft will climb with maximum TECS performance. If disabled, aircraft will gradually increase altitude so that altitude and final location are reached at the same time.
```
IamPete1, line 1089: I don't think you can use commas here. It confuses the metadata parser.

eppravitra: The comment has been changed

Hwurzburg (changes requested):
per my previous review comment...the label metadata is too long. Something closer to my suggestion is needed..I dont approve still

tridge:
@eppravitra the only thing blocking this PR is the param documentation, the description of the option. Can you please change it to the suggestion from @Hwurzburg (Henry is docs lead)
this PR is ok to merge once that param doc change is made

ArduPlane/Parameters.cpp
eppravitra: I made the changes as suggested

Hwurzburg (approved):

ArduPlane/Parameters.cpp
```diff
@@ -1086,6 +1086,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Bitmask: 11: Disable suppression of fixed wing rate gains in ground mode
     // @Bitmask: 12: Enable FBWB style loiter altitude control
     // @Bitmask: 13: Indicate takeoff waiting for neutral rudder with flight control surfaces
+    // @Bitmask: 14: In AUTO, climb to next waypoint altitude immediately instead of linear climb
```
peterbarker, line 1089:
```suggestion
    // @Bitmask: 14: In AUTO - climb to next waypoint altitude immediately instead of linear climb
```
Can't have commas here.

commas separate bitmask entries, so you can have `Bitmask: 5:Bob, 6:Mary` - a syntax we use elsewhere.  Splitting the bitmask across lines kind of confuses things...

peterbarker (changes requested):

eppravitra:
Thanks!

PR #27481, opened 2024-07-07
AP_Common: fix and clean up Bitmask class

tpwrules:
Please see commits for details.

Tested full test suite and that a vehicle seems to work okay. Saves 1.3K flash on Cube Orange.

### DIFF
--- a/libraries/AP_Common/Bitmask.h
+++ b/libraries/AP_Common/Bitmask.h
@@ -23,28 +23,28 @@
 
 #include <AP_InternalError/AP_InternalError.h>
 
-template<uint16_t num_bits>
+template<uint16_t NUMBITS>
 class Bitmask {
-    static_assert(num_bits > 0, "must store something");
+    static constexpr uint16_t NUMWORDS = ((NUMBITS+31)/32);
+
+    static_assert(NUMBITS > 0, "must store something");
     // for first_set()'s return value
-    static_assert(num_bits <= INT16_MAX, "must fit in int16_t");
+    static_assert(NUMBITS <= INT16_MAX, "must fit in int16_t");
     // so that 1U << bits is in range
     static_assert(sizeof(unsigned int) >= sizeof(uint32_t), "int too small");
 
 public:
-    Bitmask() :
-        numbits(num_bits),
-        numwords((num_bits+31)/32) {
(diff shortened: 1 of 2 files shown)


libraries/AP_Common/Bitmask.h
```diff
@@ -23,26 +23,26 @@
 
 #include <AP_InternalError/AP_InternalError.h>
 
-template<uint16_t num_bits>
+#define NUMWORDS ((NUMBITS+31)/32)
```
khancyr, line 26: isn't this require more division calculation when previously, we store do it once and store the result into numwords?

khancyr: can we change the define to a constexpr function return to get rid of the division calculation at compile time, if that isn't done by compiler already ?

peterbarker:
Looks nice, but size-wise it could be better...

OTOH, CubeOrange is somewhere where we are going to have size problems soon...


```
Board               AP_Periph  blimp  bootloader  copter  heli   iofirmware  plane  rover  sub
Durandal                       -968   *           -1336   -1120              -1192  -1352  -1096
Hitec-Airspeed      -16                                                                    
KakuteH7-bdshot                48     *           40      40                 104    88     80
MatekF405                      -8     *           24      -48                -8     48     -8
Pixhawk1-1M-bdshot             40                 72      40                 40     64     40
f103-QiotekPeriph   -16                                                                    
f303-Universal      -16                                                                    
iomcu                                                            *                         
revo-mini                      8      *           112     56                 8      48     8
skyviper-v2450                                    48                                       
```

libraries/AP_Common/Bitmask.h
tpwrules: The compiler will optimize it either way as `NUMBITS` is a constant here but it is a bit cleaner to keep it as a constant variable than a define.

tpwrules:
We could make it a size reduction across the board by removing the new range tests in `clear` and `get`. But there is already one in `set` and having it comprehensive is IMO better for the health of the code.

Plus, having robust checks here means we can remove them elsewhere (in particular in the DroneCAN DNA server) and get some of the flash back.

tridge (approved):

PR #27482, opened 2024-07-08
AP_Mission: Allow Param4 to be uploaded with NaN

joshanne:
By default, QGroundControl will attempt to upload Loiter Unlimited with a NaN in param4.
Given this field could be NaN, we allow it through the parser.

See Param4 in https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_UNLIM


peterbarker:
Presumably we don't actually manage to store NaN, and we always ignore this field?

I guess technically that means we should be bouncing any attempt to upload a waypoint *without* this field?  I'm not suggesting we make that change as it will break vast amounts of stuff...

joshanne:
AP doesn't currently make use of that field - it is not stored.

At the moment, any mission containing LOITER_UNLIM _must_ have a non-nan p4 field for the mission upload to succeed.  MAVLink specification allows for that particular field in the message to be nan

There is possibly other waypoint types that may cause an entire mission to upload if the users selected ground station does not support the AP implementation of those messages.

peterbarker (approved):
Rebased this on top of CI fixes, marking as MergeOnCIPass

PR #27483, opened 2024-07-08
autotest: add a test for switching to auto with low speed and alt

peterbarker:
SITL/autotest reproduction of something which happened in a real flight.

After switching to auto, TECS doesn't seem to get updated, throttle goes to zero, vehicle flies into ground (or close to it, in the real flight).

image

Note the messages includes "holding course", but never "Takeoff level-off starting at".  That's because relative_alt_cm is below 1000cm.  That's unlikely to change if our throttle is zero and we're below that value to start with.

We are suppressing throttle because we are below 5m-rel-alt:
image

### DIFF
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -112,7 +112,7 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
         self.wait_altitude(alt-alt_epsilon, alt+alt_epsilon, relative=True, timeout=timeout)
         self.context_pop()
 
-    def takeoff_in_FBWA(self, alt=150, alt_max=None, relative=True, mode=None, timeout=30):
+    def takeoff_in_FBWA(self, alt=150, alt_max=None, relative=True, mode=None, timeout=30, throttle=2000):
         if alt_max is None:
             alt_max = alt + 30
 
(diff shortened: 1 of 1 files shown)


tridge:
@peterbarker ok, so it is the throttle suppression heuristics?

peterbarker:
> @peterbarker ok, so it is the throttle suppression heuristics?

Yes, it appeared so in my debugging session.

I modified the test:
```
diff --git a/Tools/autotest/arduplane.py b/Tools/autotest/arduplane.py
index d8b22a8383f..8b94c218385 100644
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -5447,6 +5447,8 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
         self.takeoff_in_FBWA(3, throttle=1450)
         self.set_rc(3, 1200)
         self.change_mode('AUTO')
+        self.delay_sim_time(0.5)
+        self.send_debug_trap()
         self.wait_current_waypoint(3)
         self.fly_home_land_and_disarm()
```

and ran the simulation
```
./Tools/autotest/autotest.py --gdb --debug test.Plane.AutoTakeoffWhileFlying  --disable-break -B Plane::suppress_throttle
```

... then in gdb
```
ena
c
```

to start walking through the relevant code.

tridge:
we've decided a FLIGHT_OPTIONS bit to disable the suppression code would be worthwhile

PR #27484, opened 2024-07-08
Tools: fix flaky RngfndQuality test

tpwrules:
The RngfndQuality test does an initial check that the reported status of the rangefinder Lua backend is NotConnected. This should be the case after a reboot as the Lua backend has never sent any data. There is a normally harmless race condition between enabling scripting and installing the test script which might cause it to start before the SITL reboot.

Unfortunately, the previous test uses the Lua rangefinder backend, so once it finishes, the status of the backend will be NoData (after the data times out) instead of NotConnected. If the test script does start before the reboot, it will see the leftover NoData and raise a failure.

Don't collect STATUSTEXTs until after the reboot so that a spurious failure is ignored.

### DIFF
--- a/Tools/autotest/ardusub.py
+++ b/Tools/autotest/ardusub.py
@@ -172,7 +172,6 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
     def RngfndQuality(self):
         """Check lua Range Finder quality information flow"""
         self.context_push()
-        self.context_collect('STATUSTEXT')
 
         ex = None
         try:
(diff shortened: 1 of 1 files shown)


tpwrules:
Upon further reflection this is the wrong solution. If messages are only captured after the reboot the test success message might be missed. But if messages are captured before reboot then a spurious test failure message might be seen. The correct fix is to fix the tests that leave the Lua rangefinder backend in a weird state; this is done in #27485 .

PR #27485, opened 2024-07-08
autotest: reboot after reverting rangefinder settings

peterbarker:
Co-authored-by: Thomas Watson <twatson52@icloud.com>

Thomas noted that tests prior to RngFndQuality were leaving a LUA rangefinder hanging around in https://github.com/ArduPilot/ardupilot/pull/27484

Thomas' patch fixes RngFndQuality to only collect context messages after a reboot, which will fix the problem.

But tests should *not* leave the vehicle in this state, so this patch fixes that.

### DIFF
--- a/Tools/autotest/ardusub.py
+++ b/Tools/autotest/ardusub.py
@@ -382,6 +382,7 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
 
         self.disarm_vehicle()
         self.context_pop()
+        self.reboot_sitl()  # e.g. revert rangefinder configuration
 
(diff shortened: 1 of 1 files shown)


tpwrules (approved):
This looks to cover all the tests that use the Lua backend and so could leave it in a state that breaks RngfndQuality. Rebooting will clear that state up.

I'd encourage to revise the comment to say that it's clearing driver state and maybe put some more context in the commit message.

Thanks for the correct fix!

peterbarker:
Merged this one, thanks @tpwrules

PR #27496, opened 2024-07-09
AP_Proximity: avoid use of int16_t-read call

peterbarker:
Replaces https://github.com/ArduPilot/ardupilot/pull/27059 and is careful not to change any logic.


tridge:
@peterbarker looks good.
btw, after the last discussion I realised that uart->available() is actually constrained to the size of the readbuf on all HALs. That means <= 4k unless someone has done an unusual begin()

tridge (approved):

PR #27500, opened 2024-07-09
Add: parachute timeout parameter

Sztosik:


libraries/AP_Parachute/AP_Parachute.h
```diff
@@ -104,6 +108,7 @@ class AP_Parachute {
     AP_Int16    _servo_off_pwm; // PWM value to move servo to when shutter is deactivated
     AP_Int16    _alt_min;       // min altitude the vehicle should have before parachute is released
     AP_Int16    _delay_ms;      // delay before chute release for motors to stop
+    AP_Int16     _timeout;      // loss of control timeout value in seconds
```
IamPete1, line 111: If its in seconds I think it should be a float.

Sztosik: I have corrected it. What do you think about the range from 0.5 to 5 seconds?

libraries/AP_Parachute/AP_Parachute.h
```diff
@@ -104,6 +108,7 @@ class AP_Parachute {
     AP_Int16    _servo_off_pwm; // PWM value to move servo to when shutter is deactivated
     AP_Int16    _alt_min;       // min altitude the vehicle should have before parachute is released
     AP_Int16    _delay_ms;      // delay before chute release for motors to stop
+    AP_Float     _timeout;      // loss of control timeout value in seconds
```
rmackay9, line 111: nitpick: if possible I like to keep the parameter definitions in the order they appear in the parameter list in the .cpp file or some other logical order.  It looks like someone put the _options parameter in the wrong place and also didn't add a comment .. let's put that aside for the moment and maybe put the new parameter below _critical_sink?

rmackay9:
This looks pretty good to me.  Could you squash the commits together?

rmackay9 (approved):
This looks good to me.  @Sztosik could you confirm that you've tested this somehow?  I'm pretty sure it does but with such an important subsystem we should be sure.  Perhaps set the timeout to a really large number (like 10 seconds) so that it is really obvious that the parameter is having an effect.

Sztosik:
> This looks good to me. @Sztosik could you confirm that you've tested this somehow? I'm pretty sure it does but with such an important subsystem we should be sure. Perhaps set the timeout to a really large number (like 10 seconds) so that it is really obvious that the parameter is having an effect.

I tested it on a Hexsoon EDU450 witch Cube Orange+. I removed the propellers, switched to AltHold mode, armed the drone and raised it, then changed attitudes. Unfortunately, this does not cause the parachute to deploy (I performed an identical test on the stable version and it works fine). Now I'm trying to investigate it.

IamPete1:
> Unfortunately, this does not cause the parachute to deploy (I performed an identical test on the stable version and it works fine). Now I'm trying to investigate it.

I would guess its the baro altitude check preventing deployment if not falling.

Sztosik:
> I would guess its the baro altitude check preventing deployment if not falling.

I'm not sure, `CHUTE_ALT_MIN` was set to 1 m and I've rised it to 2 m. Chute was deployed while testing on stable version (Copter 4.5.4) with same configuration.

Sztosik:
> > I would guess its the baro altitude check preventing deployment if not falling.
> 
> I'm not sure, `CHUTE_ALT_MIN` was set to 1 m and I've rised it to 2 m. Chute was deployed while testing on stable version (Copter 4.5.4) with same configuration.

Never mind, you have right @IamPete1

Sztosik:
> I tested it on a Hexsoon EDU450 witch Cube Orange+. I removed the propellers, switched to AltHold mode, armed the drone and raised it, then changed attitudes. Unfortunately, this does not cause the parachute to deploy (I performed an identical test on the stable version and it works fine). Now I'm trying to investigate it.

I figured out that value of `CRASH_CHECK_TRIGGER_SEC` should be always grater than `CHUTE_TIMEOUT' param. In other case UAV might disarm without deploying parachute.

Sztosik:
@rmackay9, @IamPete1  I have found another issue that also exist in stable version. In case inability to keep attitude, but without altitude decrease parachute  won't be deployed even if vehicle reach `CHUTE_CRT_SINK`. For now I'm marking this PR as draft and I will try to fix it.

IamPete1:
> @rmackay9, @IamPete1 I have found another issue that also exist in stable version. In case inability to keep attitude, but without altitude decrease parachute won't be deployed even if vehicle reach `CHUTE_CRT_SINK`. For now I'm marking this PR as draft and I will try to fix it.

I think this time out fix is worth having as it is, you can always come back with the additional changes in a new PR.

libraries/AP_Parachute/AP_Parachute.cpp
```diff
@@ -83,6 +83,14 @@ const AP_Param::GroupInfo AP_Parachute::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("OPTIONS", 7, AP_Parachute, _options, AP_PARACHUTE_OPTIONS_DEFAULT),
 
+    // @Param: TIMEOUT
+    // @DisplayName: Parachute timeout
+    // @Description: Triggers the parachute if the loss of control lasts for the time specified by this parameter
```
IamPete1, line 88: Need to mention that it's copter only.

rmackay9: maybe we should make the parameters only appear for copter..?

Sztosik: @rmackay9 how to achieve it? I've tried with `APM_BUILD_COPTER_OR_HELI` but automatic checks failed.

ArduCopter/crash_check.cpp
```diff
@@ -1,7 +1,13 @@
 #include "Copter.h"
 
 // Code to detect a crash main ArduCopter code
-#define CRASH_CHECK_TRIGGER_SEC         2       // 2 seconds inverted indicates a crash
+#if PARACHUTE == ENABLED
+    // CRASH_CHECK_TRIGGER_SEC should be grater than CHUTE_TIMEOUT, to prevent disarming without deploying parachute
+    #define CRASH_CHECK_TRIGGER_SEC parachute.get_chute_timeout() + 1.0f
```
tridge, line 6: this define is not doing what you want, as it is used in a multiplication, you need parantheses around the define 
```
    #define CRASH_CHECK_TRIGGER_SEC  (parachute.get_chute_timeout() + 1.0)
```

ArduCopter/crash_check.cpp
```diff
@@ -86,7 +92,7 @@ void Copter::crash_check()
     // we may be crashing
     crash_counter++;
 
-    // check if crashing for 2 seconds
+    // check if crashing for time defined by CRASH_CHECK_TRIGGER_SEC seconds
     if (crash_counter >= (CRASH_CHECK_TRIGGER_SEC * scheduler.get_loop_rate_hz())) {
```
tridge, line 96: the multiply in this line ends up multiplying by 1.0, not the users value

tridge (changes requested):

tridge:
@Sztosik I've pushed fixes for the build issue and multiply to the PR

PR #27501, opened 2024-07-09
AP_Mount: Increment the serial instance

muramura:
The serial instance value does not match the next serial instance. I will adjust it to match the others.


rmackay9 (approved):
Thank you very much!

PR #27503, opened 2024-07-09
Sub: make gcs failsafe timeout a parameter

Williangalvani:


ArduSub/defines.h
```diff
@@ -79,8 +79,8 @@ enum LoggingParameters {
 #ifndef FS_GCS
 # define FS_GCS                        DISABLED
 #endif
-#ifndef FS_GCS_TIMEOUT_MS
-# define FS_GCS_TIMEOUT_MS             2500    // gcs failsafe triggers after this number of milliseconds with no GCS heartbeat
+#ifndef FS_GCS_TIMEOUT_S
+# define FS_GCS_TIMEOUT_S             5.0    // gcs failsafe triggers after this number of milliseconds with no GCS heartbeat
```
peterbarker, line 83:
```suggestion
# define FS_GCS_TIMEOUT_S             5.0    // gcs failsafe triggers after this number of seconds with no GCS heartbeat
```

peterbarker (changes requested):

ArduSub/defines.h
Williangalvani: fixed, thanks!

ArduSub/failsafe.cpp
```diff
@@ -321,7 +321,8 @@ void Sub::failsafe_gcs_check()
     uint32_t tnow = AP_HAL::millis();
 
     // Check if we have gotten a GCS heartbeat recently (GCS sysid must match SYSID_MYGCS parameter)
-    if (tnow - gcs_last_seen_ms < FS_GCS_TIMEOUT_MS) {
+    const float validated_timeout = constrain_float(g.failsafe_gcs_timeout, 2.0, 120);
```
peterbarker, line 324: I really have my reservations about these constraints - particularly the lower one.

This is magic stuff happening in the code, there's nothing the user can know about this from the parameter description.

I don't think it's unreasonable for the user to set outside the bounds you've put in here.

Are you sure you want to bound things here?

Williangalvani: I agree with you.
I though I had seen that somewhere else but I cant find it anywhere...
Anyway. code updated, thanks for the review

peterbarker (approved):
LGTM

PR #27504, opened 2024-07-10
QURT preparation

tridge:
This breaks out some non-QURT patches from #27091 that simplify that PR to make review easier

ping @peterbarker


tridge:
all commits are merged

PR #27507, opened 2024-07-10
Printf warning fixes

peterbarker:
Extracted from https://github.com/ArduPilot/ardupilot/pull/27504


peterbarker:
We probably should come back around and work out why we're using `printf` in these cases and stop doing that.  It's just line noise for most users.

PR #27509, opened 2024-07-10
AP_Baro: avoid i2c errors with ICP101XX

peterbarker:
this sensor doesn't like reading at higher than expected rate


rmackay9:

PR #27511, opened 2024-07-10
AP_BattMonitor: support I2C INA231 battery monitor

peterbarker:
Extracted from https://github.com/ArduPilot/ardupilot/pull/27504

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor_INA2xx.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor_INA2xx.cpp
@@ -46,6 +46,17 @@ extern const AP_HAL::HAL& hal;
 #define REG_238_DIETEMP       0x06
 #define INA_238_TEMP_C_LSB    7.8125e-3 // need to mask bottom 4 bits
 
+// INA231 specific registers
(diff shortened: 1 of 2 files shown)


libraries/AP_BattMonitor/AP_BattMonitor_INA2xx.cpp
```diff
@@ -168,6 +179,16 @@ bool AP_BattMonitor_INA2XX::configure(DevType dtype)
...
         break;
     }
+
+    case DevType::INA231: {
+        // no configuration needed
+        voltage_LSB = 1.25e-3;
+        current_LSB = max_amps / (1U<<15);
+        const uint16_t cal = 0.00512 / (current_LSB * rShunt);
+        if (write_word(REG_231_CALIBRATION, cal)) {
+            return true;
```
IamPete1, line 189: `dev_type`  is never set, so the detection is never done? Was this tested?

IamPete1: Fix here https://github.com/ArduPilot/ardupilot/pull/31324 (I guess anyway I don't have one to test)

PR #27513, opened 2024-07-10
AP_Filesystem: make fgets() much more efficient

peterbarker:
normally fgets is on a buffered FILE handle. For AP_Filesystem we use an unbuffered file descriptor. This means we were reading one byte at a time from the file

this uses lseek to make fgets() much more efficient by reading the max buffer size at a time in the file

Extracted https://github.com/ArduPilot/ardupilot/pull/27504


tridge:
@peterbarker updated to check lseek at end

PR #27515, opened 2024-07-10
Remove use of <vector> in HAL

peterbarker:
Problematic in the QURT HAL


tridge:
passing in peters repo:
image

PR #27524, opened 2024-07-10
Sub: tell ahrs we can fly in multiple directions

Williangalvani:


rmackay9:
This is fine but I wonder if it makes a difference.  I would assume that it defaults to false

Williangalvani:
> This is fine but I wonder if it makes a difference. I would assume that it defaults to false

Yes, I think it does default to false, though I don't see it being initialized  anywhere else.

Even AntennaTracker has this. so I think it makes sense for Sub to have, too.

rmackay9 (approved):

PR #27525, opened 2024-07-10
Sub: allow setting backup origin parameters for a happier gps-less operation

Williangalvani:
This change makes it so that, when arming, if no GPS is present and the origin wasn't set some other way (dvl or ugps), we use set a backup origin so we are able to use the WMM table even without position data

### DIFF
--- a/ArduSub/Parameters.cpp
+++ b/ArduSub/Parameters.cpp
@@ -709,14 +709,14 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Param: ORIGIN_LAT
     // @DisplayName: Backup latitude for EKF origin
     // @Description:  Backup EKF origin latitude used when not using a positioning system.
-    // @Units: degrees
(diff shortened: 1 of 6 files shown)


ArduSub/Parameters.cpp
```diff
@@ -706,7 +706,24 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
 
     // 18 was scripting
 
-    // 19 was airspeed
+    // @Param: ORIGIN_LAT
+    // @DisplayName: Backup latitude for EKF origin
+    // @Description:  Backup EKF origin latitude used when not using a positioning system.
```
tridge, line 711: units degrees? Should have @Units:

tridge (changes requested):
just needs units

ArduSub/Parameters.cpp
Williangalvani: done!

PR #27527, opened 2024-07-10
Tools: autotest: copter: add test for unhealthy pre-arm and failsafe

IamPete1:
This adds a test for the battery unhealthy fail safe added in https://github.com/ArduPilot/ardupilot/pull/27358.

So we can easily make a monitor unhealthy this changes SITL `ADCSource::set_pin` to return false for a pin of value `ANALOG_INPUT_NONE` (-1).

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -1115,6 +1115,43 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
 
         self.progress("All Battery failsafe tests complete")
 
+    def BatteryMissing(self):
+        ''' Test battery health pre-arm and missing failsafe'''
+        self.context_push()
+
+        # Should be good to arm with no changes
+        self.wait_ready_to_arm()
+
+        # Make monitor unhealthy, this should result in unhealthy prearm
+        self.set_parameters({
+            'BATT_VOLT_PIN': -1,
+        })
+
+        self.drain_mav()
+
+        # Battery should go unhealthy immediately
+        self.assert_prearm_failure("Battery 1 unhealthy", other_prearm_failures_fatal=False)
+
+        # Return monitor to health
+        self.context_pop()
+        self.context_push()
+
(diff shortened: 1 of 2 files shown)


Tools/autotest/arducopter.py
```diff
@@ -1115,6 +1115,43 @@ def test_battery_failsafe(self, timeout=300):
...
+
+        # Return monitor to health
+        self.context_pop()
+        self.context_push()
+
+        self.wait_ready_to_arm()
+
+        # take off and then trigger in flight
+        self.takeoff(10, mode="LOITER")
+        self.set_parameters({
```
peterbarker, line 1143:
```suggestion
        self.context_collect('STATUSTEXT')
        self.set_parameters({
```

Tools/autotest/arducopter.py
```diff
@@ -1115,6 +1115,43 @@ def test_battery_failsafe(self, timeout=300):
...
+            'BATT_VOLT_PIN': -1,
+        })
+
+        # Should trigger missing failsafe
+        self.wait_statustext("Battery 1 is missing")
```
peterbarker, line 1148:
```suggestion
        self.wait_statustext("Battery 1 is missing", check_context=True)
```

Tools/autotest/arducopter.py
```diff
@@ -1115,6 +1115,43 @@ def test_battery_failsafe(self, timeout=300):
...
+
+        # Done, reset params and reboot to clear failsafe
+        self.land_and_disarm()
+        self.context_pop()
+        self.reboot_sitl()
```
peterbarker, line 1153:
```suggestion
        self.reboot_sitl()  # clear battery failsafe
```

peterbarker (approved):
I'm going to merge it as-is, but if it flaps we know how to fix it :-)

peterbarker:
Merged, thanks!

PR #27530, opened 2024-07-11
AP_Logger: Fix logged units

Georacer:
I believe the existing `m/s` units are wrong and should be `rad/s` instead, based on:
- https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Logger/LogStructure.h#L881
- https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_OpticalFlow/AP_OpticalFlow.h#L97

@MattKear FYI.


andyp1per (approved):
Nice catch!

PR #27531, opened 2024-07-11
AP_Scripting: add Lua bindings to check VTOL approach and land phases

CallanDaniel:
This PR adds Lua bindings for the follwing functions from quadplane.h :

- in_vtol_land_approach()
- in_vtol_land_final() 
- in_vtol_land_sequence() 
- in_vtol_land_poscontrol() 
- in_vtol_airbrake()

It also makes the functions public in order to expose them to Lua scritping. 

Tested in SITL with the example script running:

Screenshot 2024-07-11 114433

### DIFF
--- a/libraries/AP_Scripting/examples/vtol_land_status.lua
+++ b/libraries/AP_Scripting/examples/vtol_land_status.lua
@@ -1,0 +1,51 @@
+--[[
+Example script to test VTOL approach bindings. Prints a message to the console once in the VTOL land sequence 
+indicating which phase of the approach and landing the aircraft is in.
+]]--
+
+-- Function to retrieve VTOL status
+function get_vtol_status()
+    -- Retrieve the status from the QuadPlane methods
+    local in_vtol_land_approach = quadplane:in_vtol_land_approach()
+    local in_vtol_land_final = quadplane:in_vtol_land_final()
+    local in_vtol_land_poscontrol = quadplane:in_vtol_land_poscontrol()
+    local in_vtol_airbrake = quadplane:in_vtol_airbrake()
+
+    -- Return the status as a table
+    return {
+      approach = in_vtol_land_approach,
+      final = in_vtol_land_final,
+      poscontrol = in_vtol_land_poscontrol,
+      airbrake = in_vtol_airbrake
+    }
+end
+
+-- Function to send VTOL status to GCS
+function send_status_to_gcs()
+    -- Get the VTOL status
+    local status = get_vtol_status()
+
+    -- Create the message to send to the GCS
+    local message = string.format("VTOL Status - Approach: %s, Final: %s, PosControl: %s, Airbrake: %s",
+                                  tostring(status.approach),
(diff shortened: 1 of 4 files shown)


libraries/AP_Scripting/examples/vtol_land_status.lua
```diff
@@ -0,0 +1,51 @@
...
+    if quadplane:in_vtol_land_sequence() then
+     send_status_to_gcs()
+    end
+
+    -- Schedule the next run of this function in 1000 ms (1 Hz)
+    return update, 1000
+end
+
+-- Start the main loop
+return update, 1000
```
IamPete1, line 51:
```suggestion
return update, 1000

```

libraries/AP_Scripting/examples/vtol_land_status.lua
```diff
@@ -0,0 +1,51 @@
...
+    local status = get_vtol_status()
+
+    -- Create the message to send to the GCS
+    local message = string.format("VTOL Status - Approach: %s, Final: %s, PosControl: %s, Airbrake: %s",
+                                  tostring(status.approach),
+                                  tostring(status.final),
+                                  tostring(status.poscontrol),
+                                  tostring(status.airbrake))
+-- Send the message to the GCS if in VTOL land sequence
+gcs:send_text(6, message)
```
IamPete1, line 35:
```suggestion
    -- Send the message to the GCS if in VTOL land sequence
    gcs:send_text(6, message)
```

libraries/AP_Scripting/examples/vtol_land_status.lua
CallanDaniel: @IamPete1 did you just add an empty line at the end here?

Hwurzburg: he did
image
see the minus sign warning of no blank line at the end

CallanDaniel: Copy that @Hwurzburg appologies, read up on that now and makes complete sense. Thanks!

CallanDaniel:
@Hwurzburg @IamPete1 I added the empty line, but it seems to dissapear when I commit. Appologies for the confusion, still learning.

IamPete1:
> @Hwurzburg @IamPete1 I added the empty line, but it seems to dissapear when I commit. Appologies for the confusion, still learning.

It could be VS Code or git itself, sometimes it will not let you commit pure whitespace changes.

tpwrules:
Bindings are kind of expensive, could you have an enum and one public function which checks and returns a specific property?

CallanDaniel:
@tpwrules I don't see why not, this had actually crossed my mind when I was writing the example script. I'll do that. Thanks for the feedback!

CallanDaniel:
@tpwrules something like this? Yet to do the bindings and a new example script - just wanted to check I'm going in the right direction.

PR #27533, opened 2024-07-11
AP_Notify: add support for blinking 1 LED for notify

peterbarker:
... also a simulator for it


peterbarker:
Tested on SDMODELH7V1.

libraries/SITL/SIM_GPIO_LED_1.h
```diff
@@ -0,0 +1,46 @@
...
+    };
+
+    SIM_LED_n<1> leds{"GPIO_LED_1", colours};
+
+    uint8_t LED_A_PIN;
+};
+
+} // namespace SITL
+
+#endif  // AP_SIM_GPIO_LED_2_ENABLED
```
andyp1per, line 46: LED_1

peterbarker: Fixed, thanks

IamPete1:
I know there is at least one board that defines a second LED pin which is not really connected just so they get a blinky light.

peterbarker:
> I know there is at least one board that defines a second LED pin which is not really connected just so they get a blinky light.

Ah, great, thanks for the confirmation.  I thought, "surely we must blink LEDs on these boards that don't define 2+ LEDs".  But I accepted the insanity and added support, half-thinking someone would point out some alternate pathway through which we blinked LEDs!

peterbarker:
Incidentally, suggestions for what patterns to show when are welcome.  I juts put something together which looked like a good launching point.

tridge (approved):

PR #27539, opened 2024-07-13
AP_HAL_SITL: allow disabling CAN interfaces

tpwrules:
Also corrects an issue where setting an invalid type could result in undefined behavior.

Tested by running two peripherals on two different buses.

### DIFF
--- a/libraries/AP_HAL_SITL/CANSocketIface.cpp
+++ b/libraries/AP_HAL_SITL/CANSocketIface.cpp
@@ -224,6 +224,10 @@ bool CANIface::init(const uint32_t bitrate, const OperatingMode mode)
         transport = NEW_NOTHROW CAN_SocketCAN();
 #endif
         break;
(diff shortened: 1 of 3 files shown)


peterbarker:
"We"?

Did you consider just making 0 "None", shifting everything along one and making the default 1?

It's much more in keeping with what we normally do and the number of users of this could be counted on one hand of a clumsy carpenter.

tpwrules:
“We” is a grammar quirk I have, the idea was wholly my own :)

I did consider renumbering but was concerned it would change CAN behavior in people’s existing SITL instances. If updating the default is sufficient to avoid that, and/or the change is not a big problem, then yes I think renumbering is a lot better.

tpwrules:
I checked and updating the default is indeed sufficient, so I went with 0=None.

tridge (approved):

PR #27540, opened 2024-07-13
AP_HAL_ESP32: use correct unformatted system ID length

tpwrules:
Avoids stuffing uninitialized data into the ID.

### DIFF
--- a/libraries/AP_HAL_ESP32/Util.cpp
+++ b/libraries/AP_HAL_ESP32/Util.cpp
@@ -243,15 +243,13 @@ bool Util::get_system_id(char buf[50])
 
 bool Util::get_system_id_unformatted(uint8_t buf[], uint8_t &len)
 {
-    len = MIN(12, len);
-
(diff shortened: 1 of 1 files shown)


libraries/AP_HAL_ESP32/Util.cpp
```diff
@@ -243,8 +243,7 @@ bool Util::get_system_id(char buf[50])
 
 bool Util::get_system_id_unformatted(uint8_t buf[], uint8_t &len)
 {
-    len = MIN(12, len);
-
+    len = MIN(6, len);
```
peterbarker, line 246: This would be better done as `len = MIN(len, ARRAY_SIZE(base_mac_addr)`, and moved down to where we are doing the mempy.

Not a blocker, this is good-enough as-is IMO

peterbarker (approved):

davidbuzz (approved):

PR #27541, opened 2024-07-13
AP_HAL_SITL: improve periph simulation efficiency

tpwrules:
Removes busywait for simulation state packet, dramatically reducing CPU usage. The dominant wait time in AP_Periph is 1024 microseconds as this is the default value of HAL_PERIPH_LOOP_DELAY_US, so a 1ms wait is unlikely to be a problem.

### DIFF
--- a/libraries/AP_HAL_SITL/SITL_Periph_State.cpp
+++ b/libraries/AP_HAL_SITL/SITL_Periph_State.cpp
@@ -124,10 +124,9 @@ void SITL_State::wait_clock(uint64_t wait_time_usec)
 {
     while (AP_HAL::micros64() < wait_time_usec) {
         struct sitl_input input {};
-        sitl_model->update(input);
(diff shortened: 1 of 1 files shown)


tridge:
I'd like to know this is OK with either quad-can or quadplane-can flying with a SIM_SPEEDUP of 5

tpwrules:
Tested quad-can with a SIM_SPEEDUP of 5 on the vehicle. Connected over Ethernet to be sure (the sim actually wouldn't start without a network connection), and the vehicle moves fast as if it were moving at 5x.

Everything seemed to look good. The desired vs. actual attitude control in a log looks beautiful.

tridge (approved):

tridge:
@tpwrules note that pinging me in code_review would have been better than adding MergeOnCiPass, that tag should only be used by people who do have merge permission for the subsystem

PR #27543, opened 2024-07-13
AP_DroneCAN: fix and clean up DNA server

tpwrules:
Please see commit messages for details.

Fixes the following issues:
* Correctly annotates and validates the node ID parameter set by the user
* Fix various cases where invalid node IDs could be used or stored
* Fixes 1/256 chance where a node's unique ID could not be registered
* Cleans up the code

Does not yet address the elephant in the room of multiple servers using the same storage. That will come next; this all makes that easier.

Tested on SITL and a Cube Orange on the bench using various configurations of AP_Periph nodes. Tested that the DNA database is preserved and that the DroneCAN GUI tool still works properly as well.

### DIFF
--- a/libraries/AP_DroneCAN/AP_DroneCAN.cpp
+++ b/libraries/AP_DroneCAN/AP_DroneCAN.cpp
@@ -84,6 +84,10 @@ extern const AP_HAL::HAL& hal;
 #define AP_DRONECAN_VOLZ_FEEDBACK_ENABLED 0
 #endif
 
+#ifndef AP_DRONECAN_DEFAULT_NODE
+#define AP_DRONECAN_DEFAULT_NODE 10
+#endif
+
 #define AP_DRONECAN_GETSET_TIMEOUT_MS 100       // timeout waiting for response from node after 0.1 sec
 
 #define debug_dronecan(level_debug, fmt, args...) do { AP::can().log_text(level_debug, "DroneCAN", fmt, ##args); } while (0)
@@ -96,11 +100,11 @@ extern const AP_HAL::HAL& hal;
 // table of user settable CAN bus parameters
 const AP_Param::GroupInfo AP_DroneCAN::var_info[] = {
     // @Param: NODE
-    // @DisplayName: DroneCAN node that is used for this network
-    // @Description: DroneCAN node should be set implicitly
-    // @Range: 1 250
+    // @DisplayName: Own node ID
+    // @Description: DroneCAN node ID used by the driver itself on this network
+    // @Range: 1 125
     // @User: Advanced
-    AP_GROUPINFO("NODE", 1, AP_DroneCAN, _dronecan_node, 10),
+    AP_GROUPINFO("NODE", 1, AP_DroneCAN, _dronecan_node, AP_DRONECAN_DEFAULT_NODE),
 
     // @Param: SRV_BM
     // @DisplayName: Output channels to be transmitted as servo over DroneCAN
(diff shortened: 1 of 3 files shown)


libraries/AP_DroneCAN/AP_DroneCAN.cpp
```diff
@@ -326,6 +330,11 @@ void AP_DroneCAN::init(uint8_t driver_index, bool enable_filters)
         debug_dronecan(AP_CANManager::LOG_ERROR, "DroneCAN: init called more than once\n\r");
         return;
     }
+    uint8_t node = _dronecan_node;
+    if (node < 1 || node > 125) { // reset to default if invalid
+        _dronecan_node.set_and_save(AP_DRONECAN_DEFAULT_NODE);
```
tridge, line 335: should be a set(), not set_and_save()

libraries/AP_DroneCAN/AP_DroneCAN_DNA_Server.cpp
```diff
@@ -61,53 +65,42 @@ void AP_DroneCAN_DNA_Server::getHash(NodeData &node_data, const uint8_t unique_i
...
-    WITH_SEMAPHORE(storage_sem);
-    if (!storage.read_block(&data, (node_id * sizeof(struct NodeData)) + NODEDATA_MAGIC_LEN, sizeof(struct NodeData))) {
-        //This will fall through to Prearm Check
-        server_state = STORAGE_FAILURE;
-        return false;
-    }
-    return true;
+    uint8_t crc = crc_crc8(node_data.hwid_hash, sizeof(node_data.hwid_hash));
+    // 0 signifies empty data record so we have to return 1 instead
+    return crc != 0 ? crc : 1;
```
tridge, line 73: i think we should consider an entry to be invalid only if all of hwid_hash is zero

tridge (changes requested):

libraries/AP_DroneCAN/AP_DroneCAN_DNA_Server.cpp
tpwrules: Couldn't find a zero checker function, just used memcmp.

tpwrules:
Performed the following tests with a Cube Orange and 3x MatekL431 node boards:
* Set up flight controller with 4.5.4 and default parameters
* Plugged in each node individually to assign IDs and program firmware with MatekL431-Periph 1.7.0. 125 has nothing plugged in, 124 and 123 have UART GPSes
* Rebooted and plugged in each node individually in ID order to make sure the IDs are preserved
* Set `GPS_CAN_NODEID1` to 123 and `GPS_CAN_NODEID2` to 124, and both GPS types to DroneCAN
* Verified that disabling a particular GPS caused the expected GPS to fail in MAVProxy
* Upgraded the autopilot to this branch
* Rebooted and plugged in each node individually in ID order to make sure the IDs are preserved
* Plugged in each GPS individually and verified that the expected GPS worked properly in MAVProxy
* Plugged in both simultaneously and verified that disabling a particular GPS caused the expected GPS to fail in MAVProxy

So I am confident that the GPSes will continue to work and this can be merged.

libraries/AP_DroneCAN/AP_DroneCAN_DNA_Server.cpp
```diff
@@ -117,93 +106,52 @@ bool AP_DroneCAN_DNA_Server::freeNodeID(uint8_t node_id)
...
 
 //Checks if a valid Server Record is present for specified Node ID
 bool AP_DroneCAN_DNA_Server::isValidNodeDataAvailable(uint8_t node_id)
 {
     NodeData node_data;
     readNodeData(node_data, node_id);
+
+    uint8_t empty_hwid[sizeof(NodeData::hwid_hash)] = {0};
     uint8_t crc = crc_crc8(node_data.hwid_hash, sizeof(node_data.hwid_hash));
-    if (crc == node_data.crc && node_data.crc != 0) {
```
tridge, line 206: I think we want to backport just this small fix (in minimal form) for 4.5.x

tridge (commented):
@bugobliterator has offered to review this PR (Sid is the original author of the code)

bugobliterator (changes requested):
What is the reasoning behind removing the checks if storage operations were successful?

bugobliterator:
> What is the reasoning behind removing the checks if storage operations were successful?

Actually, nevermind. Had a look through the our StorageManager, we write to files on separate thread, and don't return error on io failures. So certainly logical to remove the runtime checks.

bugobliterator (approved):

tridge (approved):

PR #27544, opened 2024-07-14
AP_Notify: Fix building BUG, otherwise u will build fail on the real FC!

makermare:
If anyone's hwdef files have these content, will build error.

"
define AP_NOTIFY_GPIO_LED_RGB_ENABLED
define HAL_GPIO_LED_ON 0
define AP_NOTIFY_GPIO_LED_RGB_RED_PIN 90
define AP_NOTIFY_GPIO_LED_RGB_GREEN_PIN 91
define AP_NOTIFY_GPIO_LED_RGB_BLUE_PIN 92
"

So I fixed the problem and can be build  on the real FC successful!

### DIFF
--- a/libraries/AP_Notify/AP_Notify.cpp
+++ b/libraries/AP_Notify/AP_Notify.cpp
@@ -305,7 +305,7 @@ void AP_Notify::add_backends(void)
                 ADD_BACKEND(NEW_NOTHROW ExternalLED()); // despite the name this is a built in set of onboard LED's
 #endif
 
-#if AP_NOTIFY_GPIO_LED_RGB_ENABLED
+#ifdef AP_NOTIFY_GPIO_LED_RGB_ENABLED
                 ADD_BACKEND(NEW_NOTHROW PixRacerLED());
 #elif AP_NOTIFY_VRBOARD_LED_ENABLED
                 ADD_BACKEND(NEW_NOTHROW VRBoard_LED());
--- a/libraries/AP_Notify/PixRacerLED.cpp
+++ b/libraries/AP_Notify/PixRacerLED.cpp
@@ -15,7 +15,7 @@
 
 #include "AP_Notify_config.h"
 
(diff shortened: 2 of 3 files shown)


rmackay9:
Hi @makermare, could you change the commit title to include the "AP_Notify: " prefix?  We always prefix the commits with the subsystem (or file directory) to make backporting easier.

peterbarker:
@makermare please provide instructions on how to reproduce the build failure on the master branch of ArduPilot.

peterbarker:
@makermare I'm guessing that in your branch you probably need to change
```
#define AP_NOTIFY_GPIO_LED_RGB_ENABLED
```

to

```
#define AP_NOTIFY_GPIO_LED_RGB_ENABLED 1
```

... note the "1" on the end there.

peterbarker:
> > @makermare I'm guessing that in your branch you probably need to change
> > ```
> > #define AP_NOTIFY_GPIO_LED_RGB_ENABLED
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to
> > ```
> > #define AP_NOTIFY_GPIO_LED_RGB_ENABLED 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ... note the "1" on the end there.
> > Thank u so much!
> > Why I using "AP_NOTIFY_GPIO_LED_RGB_ENABLED" without  the "1" on the end there.???
> > Because This is my first build fail, and the terminal tell me to change into "AP_NOTIFY_GPIO_LED_RGB_ENABLED" and PINs, So, I change into it just on my habit. I don't know I need to add the "1" on the end in fact!

Yeah, we've had a lot of bugs relating to `defined(X)` rather than checking using `#if`.

So we're moving *away* from `#ifdef` now-adays.

PR #27550, opened 2024-07-15
QURT: update motor order and add safety state

tridge:
Not all ModalAI quads have the same motor ordering, so we are better off doing it with parameters 

Also added safety state to guarantee no outputs during initialisation


rmackay9 (approved):

Tools/Frame_params/ModalAI/AutonomyDevKit.parm
```diff
@@ -6,6 +6,12 @@ FLTMODE1         5
 FLTMODE4         2
 FLTMODE_CH       6
 
+# motor ordering
+SERVO1_FUNCTION 34
+SERVO2_FUNCTION 35
+SERVO3_FUNCTION 33
+SERVO4_FUNCTION 36
```
IamPete1, line 13: I guess this is not equivalent to one of our existing frame types?

tridge: its not

PR #27551, opened 2024-07-15
Rover: add frame type Omni3Mecanum

srmainwaring:
Add a frame type to support 3 wheel rovers that use Mecanum wheels. The frame requires different mixing weights to Omni3 to ensure the linear (v_x, v_y) and angular (omega_z) velocities align with the vehicle axes. 

### Testing

Mecanum wheeled rover examples have been added to the Gazebo models in https://github.com/ArduPilot/SITL_Models/pull/139.

_Figure: `rc 3` controls forward and back movement._
omni3rover_fwd

_Figure: `rc 4` controls lateral movement._
omni3rover_lat

_Figure: `rc 1` controls yaw._
omni3rover_yaw

### References

The weights are determined using the approach to derive the inverse kinematics described in A. Gfrerrer. _"Geometry and kinematics of the Mecanum wheel"_, Computer Aided Geometric Design 25 (2008) 784–791. Retrieved from https://www.geometrie.tugraz.at/gfrerrer/publications/MecanumWheel.pdf.

### DIFF
--- a/Rover/Parameters.cpp
+++ b/Rover/Parameters.cpp
@@ -523,7 +523,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Param: FRAME_TYPE
     // @DisplayName: Frame Type
     // @Description: Frame Type
-    // @Values: 0:Undefined,1:Omni3,2:OmniX,3:OmniPlus
+    // @Values: 0:Undefined,1:Omni3,2:OmniX,3:OmniPlus,4:Omni3Mecanum
     // @User: Standard
     // @RebootRequired: True
     AP_GROUPINFO("FRAME_TYPE", 24, ParametersG2, frame_type, 0),
--- a/libraries/AR_Motors/AP_MotorsUGV.cpp
+++ b/libraries/AR_Motors/AP_MotorsUGV.cpp
@@ -647,6 +647,13 @@ void AP_MotorsUGV::setup_omni()
         add_omni_motor(2, 0.0f, -1.0f, 1.0f);
         add_omni_motor(3, 1.0f, 0.0f, 0.0f);
         break;
+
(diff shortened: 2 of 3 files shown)


IamPete1 (approved):
Code looks good, and after staring at the vids for a little while I have convinced myself that the wheels are moving correctly. 

Is the model independent or written from the same paper as the mixer? It would be nice to confirm with something, but I don't think we need to hold up the PR. 

Is there a standard "roller angle" for macanum wheels? I guess we would need a new mixer for each type.

rmackay9 (approved):

rmackay9:
Great thanks, we can merge once it passes CI.  We will need to update this wiki page at some point as well.  https://ardupilot.org/rover/docs/rover-motor-and-servo-connections.html

srmainwaring:
Thanks for the review @IamPete1. 

> Is the model independent or written from the same paper as the mixer? It would be nice to confirm with something, but I don't think we need to hold up the PR.

The Gazebo model is independent of the calculation in the paper. It doesn't have any knowledge of the inverse kinematics, instead the Mecanum wheels are modelled directly (each wheel has 9 rollers angled at 45 deg). The movement arises from the rigid body and contact dynamics. The main approximation is the collision mesh for each roller is a sphere - as a primitive type it has support in DART for fast collision logic which improves performance.   

> Is there a standard "roller angle" for macanum wheels? I guess we would need a new mixer for each type.

Yes, rollers at 45 deg seems to be the standard. The vehicle uses the 'left' orientation for the wheels and are positioned at the corners of an equilateral triangle. The mixer weights are scale invariant, but changes in orientation or relative distance of each wheel from the CoM will need a different mixer. I originally though to add a scripting interface like the one for the 6DoF motor model in copter, but it was a bigger change. It could be a nice follow up, allowing up to some max number of motors to be configured.

PR #27560, opened 2024-07-16
Add SITL support for skid-steering boat

stephendade:
Adding a SITL model for a skid-steering motorboat, as we don't currently have one in the built-in simulator.

Makes it easier to debug pivot turns, etc for boats.

Use ``sim_vehicle.py --vehicle=Rover --frame=motorboat-skid`` to use it

### DIFF
--- a/libraries/SITL/SIM_Sailboat.cpp
+++ b/libraries/SITL/SIM_Sailboat.cpp
@@ -32,6 +32,8 @@ namespace SITL {
 #define STEERING_SERVO_CH   0   // steering controlled by servo output 1
 #define MAINSAIL_SERVO_CH   3   // main sail controlled by servo output 4
 #define THROTTLE_SERVO_CH   2   // throttle controlled by servo output 3
+#define MOTORLEFT_SERVO_CH   0   // skid-steering left motor controlled by servo output 1
+#define MOTORRIGHT_SERVO_CH   2   // skid-steering right motor controlled by servo output 3
 #define DIRECT_WING_SERVO_CH 4
 
     // very roughly sort of a stability factors for waves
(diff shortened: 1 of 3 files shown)


libraries/SITL/SIM_Sailboat.cpp
```diff
@@ -97,13 +100,19 @@ float Sailboat::get_turn_circle(float steering) const
 // return yaw rate in deg/sec given a steering input (in the range -1 to +1) and speed in m/s
 float Sailboat::get_yaw_rate(float steering, float speed) const
 {
-    if (is_zero(steering) || is_zero(speed)) {
-        return 0;
+    float rate = 0.0f;
+    if (is_zero(steering) || (!skid_steering && is_zero(speed))) {
+        return rate;
+    }
```
rmackay9, line 106: Thanks for this.  Just a nitpick on style, we normally put the else on the same line as the preceding curly bracket

stephendade: Fixed

rmackay9:
This looks OK to me but this is all @IamPete1's work so I'll give him the last word.

libraries/SITL/SIM_Sailboat.cpp
```diff
@@ -179,7 +188,14 @@ void Sailboat::update(const struct sitl_input &input)
     update_wind(input);
 
     // in sailboats the steering controls the rudder, the throttle controls the main sail position
-    float steering = 2*((input.servos[STEERING_SERVO_CH]-1000)/1000.0f - 0.5f);
+    float steering = 0.0f;
+    if(skid_steering) {
```
IamPete1, line 192:
```suggestion
    if (skid_steering) {
```

IamPete1 (approved):
LGTM

PR #27562, opened 2024-07-16
AP_Camera: proper string formatting

landswellsong:
I believe the original code was a typo, because `.32` as a string doesn't make sense (and it did confuse me for a long time). It was probably meant as a safeguard against a full 32 byte string situation where the mavlink library would not add a terminator. Without this fix a camera advertising such a name would cause a buffer overflow.


rmackay9 (approved):
Thanks for this.  This is certainly my mistake

rmackay9:

PR #27565, opened 2024-07-16
AP_HAL_SITL: If HAL_CAN_WITH_SOCKETCAN is undefined, treat it as NONE

muramura:
If HAL_CAN_WITH_SOCKETCAN is undefined, set transport to NULLPTR instead of leaving it undefined.
When undefined, treat it the same as NONE, and set it to NULLPTR.


tpwrules (approved):
Thanks, looks good!

PR #26666, opened 2024-04-01
AP_Mount: added topotek gimbal driver.

laozhou-fujian:
This adds at least two enhancements on top of PR https://github.com/ArduPilot/ardupilot/pull/27053 which adds Topotek gimbal support:

1. retrieve gimbal model name
2. improve tracking status tracking

PR #27568, opened 2024-07-17
AP_Mount: Topotek image tracking fix

rmackay9:
This replaces PR https://github.com/ArduPilot/ardupilot/pull/26666

This PR adds 3 enhancements to the Topotek driver:

1. versions can be returned by the gimbal in "1.2.3" format (previously it was only "123"
2. gimbal's model name is retrieve
3. image tracking status is improved to handle gimbals that do not have this feature

This has been lightly tested on a real gimbal but we need to clear up these issues:

1. the model name appears as "000000000"
2. I don't really understand the new tracking state "waiting" and how the driver should decide what state to put the gimbal into

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Topotek.cpp
+++ b/libraries/AP_Mount/AP_Mount_Topotek.cpp
@@ -770,7 +770,7 @@ void AP_Mount_Topotek::request_track_status()
 // request gimbal version
 void AP_Mount_Topotek::request_gimbal_version()
 {
-    // sample command: #TPPD2rVSN00
+    // sample command: #TPUD2rVSN00
     send_fixedlen_packet(AddressByte::SYSTEM_AND_IMAGE, AP_MOUNT_TOPOTEK_ID3CHAR_GET_VERSION, false, 0);
 }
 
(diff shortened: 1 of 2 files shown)


laozhou-fujian:
> This replaces PR #26666
> 
> This PR adds 3 enhancements to the Topotek driver:
> 
> 1. versions can be returned by the gimbal in "1.2.3" format (previously it was only "123"
> 2. gimbal's model name is retrieve
> 3. image tracking status is improved to handle gimbals that do not have this feature
> 
> This has been lightly tested on a real gimbal but we need to clear up these issues:
> 
> 1. the model name appears as "000000000"
> 2. I don't really understand the new tracking state "waiting" and how the driver should decide what state to put the gimbal into

Hello
    Because the "model_name" parameter was not set previously, "000000000" appeared. Now, our gimbal cameras are all configured with this parameter during production.
    In the new type of gimbal camera that we are preparing to launch, a new tracking state "waiting" has been added. In this tracking state, the gimbal can automatically detect and recognize certain features such as vehicles, people, etc., and dynamically overlay them on the video stream in real-time, providing reference targets for the operator.

tridge (approved):

PR #27569, opened 2024-07-17
Copter: add and use private ModeAuto::option_is_enabled method

peterbarker:
Tested a couple of these in SITL.

### DIFF
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -202,15 +202,23 @@ void ModeAuto::set_submode(SubMode new_submode)
     }
 }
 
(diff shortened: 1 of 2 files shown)


ArduCopter/mode_auto.cpp
```diff
@@ -202,15 +202,23 @@ void ModeAuto::set_submode(SubMode new_submode)
     }
 }
 
+bool ModeAuto::option_is_enabled(Option option) const
```
rmackay9, line 205: nitpick: maybe put a comment above the method?  Sorry, it's pretty obvious I know but ..

rmackay9 (approved):

tridge:
merging with known DDS CI error

Issue #1898, opened 2015-02-21
Keep track of battery resistance and optionally warn if it's too low before arming


marcmerlin:
I had a discussion with Tridge about an expensive (and potentially dangerous) crash with an octocopter where the 6000mAh battery (with about 10 uses) died flat after 4300mAh used.

He mentioned that while the details could be tricky, especially if you swap batteries, the APM could keep track of the battery internal resistance and warn before takeoff if it has increased too much.
This could help detect a battery that is likely to die before its expected capacity.

He suggested I file this bug :)

lthall:
Hi Marc,
We are currently adding code that estimates the battery impedance each time you arm and take off. This is being used for another purpose at the moment but we may be able to include the feature you are suggesting.

I don't know how we could calculate battery impedance before take off and be able to abort take off in a safe manner. If you have any suggestions about how battery impedance could be measured without the need for significant current draws (and therefore copter take off), I would welcome your suggestions!!

Do you have any documentation on the characteristics of battery impedance  during failure?

marcmerlin:
Hi Leonard, thanks for your answer. I'm not sure how to include Tridge on this, he seemed to have an idea on how to do it when we chatted. Do you know how to Cc him on a github ticket?
I can't give you info from the battery that caused a crash since it ran dry in flight and is now damaged way beyond what it was before takeoff.

But if I had to guess, how about this:
When you arm, it does spin the props at low speed, is that good enough to tell?
If not, could arm spin the props faster until the accelerometers detect the quad is about to take off, and then spin down? That should give a better I to calculate R=U/I .

jschall:
I think the solution to this is smart batteries that store battery state
information onboard.

They are already prevalent in commercial and consumer multicopters but have
yet to work their way into the hobbyist market.

lthall:
The only thing I can suggest is sounding an alarm if the battery impedance is greater than a user specified value. That would then leave it up to the pilot to decide to land and check the battery.

marcmerlin:
@lthall you mentioned a year ago there was code being added to help with this.
Do you know if it happened and whether we should resolve this?

OXINARF:
Linked to #6351

lthall:
The only practical implementation of the remaining part of this issue is to initiate a failsafe if the battery impedance is over a user defined amount after a predetermined flight time or mAh usage percentage.

Otherwise this is a smart battery option and could be closed.

Naterater:
[...]

PR #27571, opened 2024-07-17
Add script for monitoring battery resistance

peterbarker:
Closes https://github.com/ArduPilot/ardupilot/issues/1898

### DIFF
--- a/libraries/AP_Scripting/examples/battery_internal_resistance_check.lua
+++ b/libraries/AP_Scripting/examples/battery_internal_resistance_check.lua
@@ -1,0 +1,35 @@
+--[[
+ script implementing pre-arm check that internal resistance is sensible
+--]]
+
+local MAX_RESISTANCE = 0.03  -- Ohms
+
+local auth_id = assert(arming:get_aux_auth_id())
+
+local warning_last_sent_ms = uint32_t() -- time we last sent a warning message to the user
+warning_interval_ms = 10000
+
+function update()
+   local num_batts = battery:num_instances()
+   local ok = true
+   for i=0,num_batts do
+      local resistance = battery:get_resistance(i)
+      failed = resistance > MAX_RESISTANCE
+      if failed then
+         msg = string.format("Batt[%u] high internal resistance %.5f Ohms", i+1, resistance)
+         if millis() - warning_last_sent_ms > warning_interval_ms then
+            gcs:send_text(0, msg)
+            warning_last_sent_ms = millis()
+         end
+         arming:set_aux_auth_failed(auth_id, msg)
+         ok = false
+      end
+   end
+
+   if ok then
+      arming:set_aux_auth_passed(auth_id)
+   end
+   return update, 500
+end
+
+return update()
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -3147,6 +3147,11 @@ function battery:current_amps(instance) end
 ---@return number -- resting voltage
 function battery:voltage_resting_estimate(instance) end
 
+-- Returns the estimated internal battery resistance in Ohms
+---@param instance integer -- battery instance
+---@return number -- estimated internal resistance in Ohms
+function battery:get_resistance(instance) end
+
 -- Returns the voltage of the selected battery instance.
 ---@param instance integer -- battery instance
 ---@return number -- voltage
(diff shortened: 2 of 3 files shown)


tpwrules:
Can you add docs too? I assume these are returned in Ohms.

Does this actually solve the problem? In my experience internal resistance estimates are pretty bad until after takeoff and the expected current is being consumed. I'm not sure how a pre-arm check, especially after a fresh boot, would help. I would think you need to wait a couple seconds into flight or so, then perhaps switch to land mode if it's high.

tridge (approved):

tridge (approved):

libraries/AP_Scripting/examples/battery_internal_resistance_check.lua
```diff
@@ -0,0 +1,35 @@
+--[[
+ script implementing pre-arm check that internal resistance is sensible
+--]]
+
+local MAX_RESISTANCE = 0.03
+
+local auth_id = arming:get_aux_auth_id()
```
IamPete1, line 7: You could assert this to pass ci

peterbarker: Could you explain this in shorter words, please?

I *am* struggling with the cleanliness checks...

peterbarker:
> Can you add docs too? I assume these are returned in Ohms.

Docs have been added.  I've annotated the variable.

> Does this actually solve the problem? In my experience internal resistance estimates are pretty bad (5-10x expected values) until after takeoff and the expected current is being consumed. It's also easy to imagine that if the current sensor is slightly different the estimates could be unrealistically low too.

Not expected to fix things for your bog-standard copter for initial takeoff.


> I'm not sure how a pre-arm check, especially after a fresh boot, would help. I would think you need to wait a couple seconds into flight or so, then perhaps switch to land mode if it's high. Past the first couple seconds it would be better to just warn the user.

You'll get a warning in-air and a pre-arm check once you disarm after landing.  I started off writing a full-blown C++ thing, triggering a low-battery failsafe on a parameter-defined high-resistance value.  Then I decided to punt to a simple script for now.

Fancier batteries might start to report a persistently-stored internal resistance, in which case you'd get protection before first takeoff.

libraries/AP_Scripting/examples/battery_internal_resistance_check.lua
IamPete1: This is the error:
```
scripts/battery_internal_resistance_check.lua:30:34
        Code: param-type-mismatch
        Cannot assign `integer|nil` to parameter `integer`.
        - `nil` cannot match `integer`
        - Type `nil` cannot match `integer`
scripts/battery_internal_resistance_check.lua:24:37
        Code: param-type-mismatch
        Cannot assign `integer|nil` to parameter `integer`.
        - `nil` cannot match `integer`
        - Type `nil` cannot match `integer`
```

The errors are for:
```
arming:set_aux_auth_passed(auth_id)
```
and
```
arming:set_aux_auth_failed(auth_id, msg)
```

The issue is that both functions are expecting `auth_id` to be a integer type. However it is declared as:
```
local auth_id = arming:get_aux_auth_id()
```

The problem is that `arming:get_aux_auth_id()` can fail and return nil if its run out  of ids to give out. So `auth_id` could be a integer or it could be nil. 

If we change the definition of `auth_id` to:
```
local auth_id = assert(arming:get_aux_auth_id(), "No auth IDs available")
```
The assert will error out if the returned `auth_id` is nil, so now the checker knows that `auth_id` must be and integer. The types then match and CI is happy.

tpwrules:
> Docs have been added. I've annotated the variable.
> 

Thank you. Could you specifically note in docs.lua that the return quantity is resistance in Ohms?

> You'll get a warning in-air and a pre-arm check once you disarm after landing. I started off writing a full-blown C++ thing, triggering a low-battery failsafe on a parameter-defined high-resistance value. Then I decided to punt to a simple script for now.

My problem is that, if I put this script on any of my vehicles, they could never take off because the calculated internal resistance is always dangerously high after boot. It only settles to an accurate value a few seconds after takeoff. So I question the value of the pre-arm check; maybe my vehicles are an edge case?

peterbarker:
> > Docs have been added. I've annotated the variable.
> 
> Thank you. Could you specifically note in docs.lua that the return quantity is resistance in Ohms?

Done.

> My problem is that, if I put this script on any of my vehicles, they could never take off because the calculated internal resistance is always dangerously high after boot. It only settles to an accurate value a few seconds after takeoff. So I question the value of the pre-arm check; maybe my vehicles are an edge case?

We're starting to get into serious heuristics now, which is why this is currently and example. not an applet.

libraries/AP_Scripting/examples/battery_internal_resistance_check.lua
peterbarker: > The assert will error out if the returned `auth_id` is nil, so now the checker knows that `auth_id` must be a integer after the assert. The types then match and CI is happy.

Many thanks, I now understand a little more about LUA :-)

I've made the suggested change to this PR.

PR #27573, opened 2024-07-17
Do not Read FIFO faster than requested rate for ICM45686

bugobliterator:
TLDR; this PR allows users to be able to better control how fast we read ICM45686, previously we were only limited to 3.2KHz and 6.4KHz. With this PR 0.8, 1.6, 3.2 and 6.4 KHz will be achievable. And for all boards except CubeRedPrimary, the default Sensor read (FIFO read) rate will be 1.6KHz, which was previously 3.2KHz.

Additional Notes:

* We seem to be always setting ICM45686 backend loop rates atleast at 3.2KHz when fast_sampling was requested. And at 1.6KHz when it isn't.
* This PR modifies the rate at which we read from FIFO depending on requested gyro rates. We still sample at 3.2KHz or 6.4 KHz depending on requested backend loop rates, we just read from FIFO based on what read rate is requested.

When fast sampling is not requested we read FIFO at 0.8KHz

Currently there are 6 flight boards that are using ICM45686, CubeOrangePlus, CubeRed, Here4FC, Pixhawk6X and ZeroOneX6. As it stands in master all 6 boards are polling the sensor 3.2KHz. 

There is a hard limitation in current master which doesn't even allow users to set the backend rate to value lower than 3.2KHz. Basically even if you had fast sampling turned off the backend rate will be stuck at 3.2KHz.

With this PR following configurations are achievable:

On CubeOrangePlus, Here4FC, Pixhawk 6X and ZeroOneX6, this PR will change the polling rate from 3.2KHz to more reasonable value of 1.6KHz as the default (As all of these boards have FAST_SAMPLING turned on for all IMUs by default). The sample rate on the IMU still remains as 3.2KHz though
If user needs to have faster polling, they can set INS_GYRO_RATE parameters to get polling rate of 3.2 and 6.4. 

On CubeRedPrimary, this will put the primary IMU at polling rate of 1.6KHz and other two at 800Hz. This is because default INS_FAST_SAMPLE mask is set to 1.

As a means of safety this PR will ensure that current users setup don't break in case they are using higher SCHED_LOOP_RATE values https://github.com/ArduPilot/ardupilot/pull/28672

The PR also adds support for showing both the Loop Rate and Sample Rate of the IMU for InvensenseV3. This signifies the difference between the two values. We have been displaying these values for Invensensv2 and v1 as well. We did it by the means of two values separated by / like so :
```
        snprintf(banner, banner_len, "IMU%u: fast sampling enabled %.1fkHz/%.1fkHz",
            gyro_instance, _gyro_backend_rate_hz * _gyro_fifo_downsample_rate * 0.001, _gyro_backend_rate_hz * 0.001);
```

Flight Logs:

CubeOrangePlus LR 0.8 / SR 3.2 : https://droneshare.cubepilot.com/#/s/log?b=1721188076568&c=5a75a9ee-43ef-11ef-a740-b360934a095f&d=CubeOrangePlus+LR+0.8+%2F+SR+3.2&e=2&a=cbfcd2fc-3fa9-11ed-b631-b97a8244eade

CubeOrangePlus LR 1.6 / SR 3.2 : https://droneshare.cubepilot.com/#/s/log?b=1721188057353&c=4f01d6a9-43ef-11ef-a740-7fca103c5c87&d=CubeOrangePlus+LR+1.6+%2F+SR+3.2&e=2&a=cbfcd2fc-3fa9-11ed-b631-b97a8244eade

CubeOrangePlus LR 3.2/ SR 3.2 : https://droneshare.cubepilot.com/#/s/log?b=1721187986195&c=2497d6a7-43ef-11ef-a740-11b81cb1ae46&d=CubeOrangePlus+LR+3.2%2F+SR+3.2&e=2&a=cbfcd2fc-3fa9-11ed-b631-b97a8244eade

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
@@ -343,14 +343,10 @@ void AP_InertialSensor_Invensensev3::start()
     // pre-calculate backend period
     backend_period_us = 1000000UL / backend_rate_hz;
 
-    if (!_imu.register_gyro(gyro_instance, backend_rate_hz, dev->get_bus_id_devtype(devtype)) ||
-        !_imu.register_accel(accel_instance, backend_rate_hz, dev->get_bus_id_devtype(devtype))) {
+    if (!_imu.register_gyro(gyro_instance, sampling_rate_hz, dev->get_bus_id_devtype(devtype)) ||
+        !_imu.register_accel(accel_instance, sampling_rate_hz, dev->get_bus_id_devtype(devtype))) {
         return;
     }
-
-    // update backend sample rate
-    _set_accel_raw_sample_rate(accel_instance, backend_rate_hz);
-    _set_gyro_raw_sample_rate(gyro_instance, backend_rate_hz);
 
     // indicate what multiplier is appropriate for the sensors'
     // readings to fit them into an int16_t:
@@ -381,16 +377,15 @@ void AP_InertialSensor_Invensensev3::start()
 
 // get a startup banner to output to the GCS
 bool AP_InertialSensor_Invensensev3::get_output_banner(char* banner, uint8_t banner_len) {
-    if (fast_sampling) {
-        snprintf(banner, banner_len, "IMU%u: fast%s sampling enabled %.1fkHz",
-            gyro_instance,
+    snprintf(banner, banner_len, "IMU%u:%s%s sampling %.1fkHz/%.1fkHz",
+        gyro_instance,
+        fast_sampling ? " fast" : " normal",
 #if HAL_INS_HIGHRES_SAMPLE
-            highres_sampling ? ", high-resolution" :
+        highres_sampling ? " hi-res" :
 #endif
-            "" , backend_rate_hz * 0.001);
-        return true;
-    }
-    return false;
+        "", backend_rate_hz * 0.001,
+        sampling_rate_hz * 0.001);
+    return true;
 }
 
 /*
@@ -669,21 +664,21 @@ void AP_InertialSensor_Invensensev3::register_write_bank(uint8_t bank, uint8_t r
 }
 
 // calculate the fast sampling backend rate
-uint16_t AP_InertialSensor_Invensensev3::calculate_fast_sampling_backend_rate(uint16_t base_odr, uint16_t max_odr) const
+uint16_t AP_InertialSensor_Invensensev3::calculate_fast_sampling_backend_rate(uint16_t base_backend_rate, uint16_t max_backend_rate) const
 {
     // constrain the gyro rate to be at least the loop rate
-    uint8_t loop_limit = 1;
-    if (get_loop_rate_hz() > base_odr) {
-        loop_limit = 2;
+    uint8_t min_base_rate_multiplier = 1;
+    if (get_loop_rate_hz() > base_backend_rate) {
+        min_base_rate_multiplier = 2;
     }
-    if (get_loop_rate_hz() > base_odr * 2) {
-        loop_limit = 4;
+    if (get_loop_rate_hz() > base_backend_rate * 2) {
+        min_base_rate_multiplier = 4;
     }
     // constrain the gyro rate to be a 2^N multiple
-    uint8_t fast_sampling_rate = constrain_int16(get_fast_sampling_rate(), loop_limit, 8);
+    uint8_t fast_sampling_rate_multiplier = constrain_int16(get_fast_sampling_rate(), min_base_rate_multiplier, 8);
 
     // calculate rate we will be giving samples to the backend
-    return constrain_int16(base_odr * fast_sampling_rate, base_odr, max_odr);
+    return constrain_int16(base_backend_rate * fast_sampling_rate_multiplier, base_backend_rate, max_backend_rate);
 }
 
 /*
@@ -806,6 +801,7 @@ void AP_InertialSensor_Invensensev3::set_filter_and_scaling(void)
             }
         }
     }
+    sampling_rate_hz = backend_rate_hz; // sampling rate and backend rate are the same
 
     // disable gyro and accel as per 12.9 in the ICM-42688 docs
     register_write(INV3REG_PWR_MGMT0, 0x00);
@@ -862,6 +858,7 @@ void AP_InertialSensor_Invensensev3::set_filter_and_scaling(void)
 void AP_InertialSensor_Invensensev3::set_filter_and_scaling_icm42670(void)
 {
     backend_rate_hz = 1600;
+    sampling_rate_hz = 1600;
     // use low-noise mode
     register_write(INV3REG_70_PWR_MGMT0, 0x0f);
     hal.scheduler->delay_microseconds(300);
@@ -886,27 +883,24 @@ void AP_InertialSensor_Invensensev3::set_filter_and_scaling_icm42670(void)
  */
 void AP_InertialSensor_Invensensev3::set_filter_and_scaling_icm456xy(void)
 {
-    uint8_t odr_config = 4;
-    backend_rate_hz = 1600;
-    // always fast sampling
-    fast_sampling = dev->bus_type() == AP_HAL::Device::BUS_TYPE_SPI;
+    uint8_t odr_config;
+    backend_rate_hz = 800;
 
-    if (enable_fast_sampling(accel_instance) && get_fast_sampling_rate() > 1) {
-        backend_rate_hz = calculate_fast_sampling_backend_rate(backend_rate_hz, backend_rate_hz * 4);
+    fast_sampling = dev->bus_type() == AP_HAL::Device::BUS_TYPE_SPI;
+    if (enable_fast_sampling(accel_instance) && get_fast_sampling_rate() && fast_sampling) {
+        // For ICM45686 we only set 3200 or 6400Hz as sampling rates
+        // we don't need to read FIFO faster than requested rate
+        backend_rate_hz = calculate_fast_sampling_backend_rate(800, 6400);
+    } else {
+        fast_sampling = false;
     }
 
-    // this sensor actually only supports 2 speeds
-    backend_rate_hz = constrain_int16(backend_rate_hz, 3200, 6400);
-
-    switch (backend_rate_hz) {
-    case 6400: // 6.4Khz
-        odr_config = 3;
-        break;
-    case 3200: // 3.2Khz
-        odr_config = 4;
-        break;
-    default:
-        break;
+    if (backend_rate_hz <= 3200) {
+        odr_config = 0x04; // 3200Hz
+        sampling_rate_hz = 3200;
+    } else {
+        odr_config = 0x03; // 6400Hz
+        sampling_rate_hz = 6400;
     }
 
     // Disable FIFO first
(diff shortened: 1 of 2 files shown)


tridge:
@andyp1per need your review on this one

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.h
```diff
@@ -140,4 +140,5 @@ class AP_InertialSensor_Invensensev3 : public AP_InertialSensor_Backend
 
     float temp_filtered;
     LowPassFilter2pFloat temp_filter;
+    uint32_t sampling_rate_hz;
```
andyp1per, line 143: Where is this used other than in the boot message?

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
```diff
@@ -865,27 +866,24 @@ void AP_InertialSensor_Invensensev3::set_filter_and_scaling_icm42670(void)
  */
 void AP_InertialSensor_Invensensev3::set_filter_and_scaling_icm456xy(void)
 {
-    uint8_t odr_config = 4;
-    backend_rate_hz = 1600;
-    // always fast sampling
-    fast_sampling = dev->bus_type() == AP_HAL::Device::BUS_TYPE_SPI;
+    uint8_t odr_config;
+    backend_rate_hz = 800;
```
andyp1per, line 870: Problem is this changes the backend rate for all boards using 45686, not just the CubeRed. I suspect it will also change people's tunes, although I haven't fully got my head around what this is actually changing other than lowering the backend rate (which is the filter rate and thus changing people's tunes). I think you are going to need some way of keeping current setups as they are while changing the default for new setups.

andyp1per (changes requested):
I think this is probably a reasonable direction to go in, but needs to be done in a way that people's existing tunes and setups are not compromised.

libraries/AP_HAL_ChibiOS/hwdef/CubeRedPrimary/hwdef.dat
```diff
@@ -105,22 +105,25 @@ PE5    SPI4_MISO    SPI4
...
-SPIDEV icm45686         SPI4 DEVID1  ICM_2_CS           MODE3  2*MHZ  8*MHZ
+SPIDEV icm45686         SPI4 DEVID1  ICM_2_CS           MODE3  2*MHZ  24*MHZ
 SPIDEV ms5611_1         SPI4 DEVID2  BARO_1_CS          MODE3 20*MHZ 20*MHZ
 
 IMU Invensensev3 SPI:icm42688_1 ROTATION_YAW_90
 IMU Invensensev3 SPI:icm42688_0 ROTATION_PITCH_180_YAW_270
 IMU Invensensev2 SPI:icm20649 ROTATION_PITCH_180
 IMU Invensensev3 SPI:icm45686 ROTATION_YAW_180
 
+define HAL_INS_HIGHRES_SAMPLE 5
```
andyp1per, line 124: why not 7?

bugobliterator: For variety

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
bugobliterator: This is not changing the sampling rate though, the samples are still getting consumed at the same rate as before i.e. 3.2KHz. We just flush the data from IMU FIFO at rate slower than 3.2KHz.

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.h
bugobliterator: Yes, that's it. Its used to convey that backend rate is different than the actual sampling rate

tridge:
need to defer till @andyp1per can flight test this

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
```diff
@@ -373,16 +373,15 @@ void AP_InertialSensor_Invensensev3::start()
 
 // get a startup banner to output to the GCS
 bool AP_InertialSensor_Invensensev3::get_output_banner(char* banner, uint8_t banner_len) {
-    if (fast_sampling) {
-        snprintf(banner, banner_len, "IMU%u: fast%s sampling enabled %.1fkHz",
-            gyro_instance,
+    snprintf(banner, banner_len, "IMU%u:%s%s sampling LR:%.1fkHz SR:%.1fkHz",
```
andyp1per, line 376: Not sure LR/SR is descriptive enough _ I certainly don't know what they mean. The rate that matters is the one that the filters are run at.

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
andyp1per: The problem is that you have increased the latency, and its the latency that matters for control - there are two benefits to faster rates - filtering at higher rates which benefits aliasing and lower latency is response which benefits control. Your change means we now only get the first.

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
```diff
@@ -865,27 +866,24 @@ void AP_InertialSensor_Invensensev3::set_filter_and_scaling_icm42670(void)
  */
 void AP_InertialSensor_Invensensev3::set_filter_and_scaling_icm456xy(void)
 {
-    uint8_t odr_config = 4;
-    backend_rate_hz = 1600;
-    // always fast sampling
-    fast_sampling = dev->bus_type() == AP_HAL::Device::BUS_TYPE_SPI;
+    uint8_t odr_config;
```
andyp1per, line 869: Where is the 800Hz value set in the ODR?

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
```diff
@@ -649,21 +648,21 @@ void AP_InertialSensor_Invensensev3::register_write_bank(uint8_t bank, uint8_t r
 }
 
 // calculate the fast sampling backend rate
-uint16_t AP_InertialSensor_Invensensev3::calculate_fast_sampling_backend_rate(uint16_t base_odr, uint16_t max_odr) const
+uint16_t AP_InertialSensor_Invensensev3::calculate_fast_sampling_backend_rate(uint16_t base_backend_rate, uint16_t max_backend_rate) const
```
andyp1per, line 651: This is just a rename, I think you should remove this from the PR to make review a little easier

andyp1per (changes requested):
I think what this PR is achieving is reducing the default non fast sample rate from 1600 to 800 and the default fast sampling rate on H7 from 3.2Khz to 1.6Khz. I think the PR should *just* do that - the renaming and message changes are confusing and not helping IMO. I also think there is a bug in the rate setting that is hard to see with the other changes. In general I agree that changing the range from 1.6Khz->6.4Khz to 800Hz->6.4Khz is a sensible one and ties in better with our existing range of 1Khz->8Khz, but it will affect the latency of samples for existing users. That's probably ok as part of the 4.6 cycle, but probably not ok as a 4.5 backport.

tridge:
we really need to know why the CPU cost is so high. @bugobliterator will try to do some profile to see why that is

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
```diff
@@ -865,27 +866,24 @@ void AP_InertialSensor_Invensensev3::set_filter_and_scaling_icm42670(void)
...
+    uint8_t odr_config;
+    backend_rate_hz = 800;
 
-    if (enable_fast_sampling(accel_instance) && get_fast_sampling_rate() > 1) {
-        backend_rate_hz = calculate_fast_sampling_backend_rate(backend_rate_hz, backend_rate_hz * 4);
+    fast_sampling = dev->bus_type() == AP_HAL::Device::BUS_TYPE_SPI;
+    if (enable_fast_sampling(accel_instance) && get_fast_sampling_rate() && fast_sampling) {
+        // For ICM45686 we only set 3200 or 6400Hz as sampling rates
+        // we don't need to read FIFO faster than requested rate
+        backend_rate_hz = calculate_fast_sampling_backend_rate(800, 6400);
```
andyp1per, line 876: Changes fast sampling backend rate

andyp1per (changes requested):
None of my comments have been addressed and it does affect fast sampling

andyp1per:
I really don't think this is the right way to do this. To get what you want on cube's I think you should introduce a new parameter say "INS_GYRO_FO_RATE" which allows you to control the backend rate independently from the sample rate. You could even call it "INS_GYRO_DIV_RTE" with a default of 1 which represents the FIFO rate divider. You can then set this as high as you like on Cube's but also allows you to do some kind of upgrade path so as to not affect current users. The its not one size fits all and both the current behaviour and what you want can be supported. 

Even better you could resurrect https://github.com/ArduPilot/ardupilot/pull/27841 and then for the non-primary set the FIFO rate really low. You would then get the best of all worlds - low latency on the primary gyro where it matters and low CPU on the backups. The EKF won't care, its only the rate and attitude controllers that care about the latency.

andyp1per:
@bugobliterator I have done a demo in https://github.com/ArduPilot/ardupilot/pull/27841 of what I mean. I think this is a much better way and means *everyone* gets a CPU improvement without *anyone* being penalized for lower latency. I think it also means that all Cube's benefit rather than just the ones with only this one sensor.

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
tridge: @andyp1per what do you mean by latency here? The filtered gyro should be identical in both cases, exactly the same samples are going through exactly the same filters, it is just batching them together

tridge:
@andyp1per I really don't understand what you mean by latency in your comments. Can you explain it in terms of what happens on a 400Hz copter? As far as I can see this PR just changes the batching of the FIFO reads, so we do less SPI transactions when the loop rate doesn't need the transactions at the higher rate. I can't see any sort of latency that impacts for a vehicle not using the rate thread

andyp1per (changes requested):

andyp1per:
@bugobliterator I have done a simpler alternative here - https://github.com/ArduPilot/ardupilot/pull/28682
I think this gives you what you want because on CubeRed/CubeOrange the H7 means that the gyro rate is set to 1 which means that you get 2x the backend rate. Since the lower bound is now 800Hz this means that your backend rate is 1600Hz which is the same as in your PR (so you get the CPU benefit you are looking for). It does still keep the backend and ODR rates the same (so ODR is 1600Hz in this example), but I would prefer to keep it that way for the backport to 4.6 and then address the backend/ODR split more generically via either a new config parameter or via dynamic FIFO rates (
I do agree that we should reduce the FIFO rate relative to the ODR rate in a way that makes sense, but we should do it for all sensors rather than ad-hoc using some arbitrary rules).

bugobliterator:
Actually I don't think with your PR I get what I want. My goal is to keep the Sensor Rate above 1000Hz, and use FIFO to buffer sample. In your PR you are reducing the ODR of the sensor, It will give more CPU time sure, but at a severe cost of low sampling rate.

I also don't believe I am applying any arbitrary rule, in fact I am making the driver more consistent with Invensensev1 and v2. Splitting the notion of backend rate and sampling rate is not novel we have been doing it for both of the remaining two versions of Invensense sensors.

andyp1per (approved):
Ok, I've been through this again - I'm still unhappy that there are drive-by changes in here which IMO are unnecessary for a backport and removing them would have certainly made the discussion shorter, but we've discussed this enough and my main concerns have been addressed.

rmackay9:

PR #27574, opened 2024-07-17
Double Data Rate on Cube Red with W25 DDR capable External Flash 

bugobliterator:
* Adds support for w25-dtr external flash in Double Data Rate mode.

Gives us 60% increase from 100MHz to 160MHz transfer rate


tridge:
looks like this change will impact the YJUAV_A6SE and the SPRacingH7
@andyp1per can you test SPRacingH7?
@peterbarker can we test this PR on your plane?

andyp1per:
This needs to be off on the Spracing it is very deliberately not turned on as it yields no benefit and is slightly less reliable.Sid did you actually test that this makes a difference? The raw speed is higher but some of the other timing has to be loosened so that the net result is often a washAndySent from my iPhoneOn 17 Jul 2024, at 10:25, Andrew Tridgell ***@***.***> wrote:﻿
@tridge requested your review on: #27574 Double Data Rate on Cube Red with W25 DDR capable External Flash .

andyp1per (changes requested):
Need to not affect existing boards

bugobliterator:
Made changes so there's a selection to turn the DTR feature on or off.

tridge:
marked a DelayMerge pending some performance testing that shows a real improvement

PR #27576, opened 2024-07-17
Split SoloGimbal physics out into a separate library

peterbarker:
Will allow for re-use of the physics.


tridge (approved):
happy to merge if this is tested, to not break solo gimbal

PR #27577, opened 2024-07-17
Do PPP networking between Primary and Secondary MCU 

bugobliterator:
* also adds support for TCP reconnection, uart error stats and low noise setting on UART


tridge (changes requested):
nice to see this come together!

libraries/AP_HAL_ChibiOS/UARTDriver.h
```diff
@@ -79,6 +79,7 @@ class ChibiOS::UARTDriver : public AP_HAL::UARTDriver {
         uint8_t get_index(void) const {
             return uint8_t(this - &_serial_tab[0]);
         }
+        bool low_noise_line;
```
tridge, line 82: can this be under ifdef?

bugobliterator: fixed

tridge (approved):

PR #20911, opened 2022-06-05
AP_Math: calc_lowpass_alpha_dt: remove unused constrain

IamPete1:
A little bit of a drive by, the constrain will never trigger.

`dt/(dt+rc)` is already constrained to 0 -> 1.

dt = 0 or rc = inf

0/(0+rc) = 0

rc = 0 or dt = inf

dt/(dt + 0) = 1

PR #27582, opened 2024-07-17
Filter: LowPass: use `calc_lowpass_alpha_dt` in `apply` and prefix all class varables with `_`

IamPete1:
A small simplification to our base low pass filter. It now uses `calc_lowpass_alpha_dt` and `apply` so the existing code is reused rather than being re-implemented. 

There are some changes in behavior in the invalid dt/cutoff_freq cases. Previously the local copy of alpha was not updated where as it now is. This should not matter because the two apply methods should not be used together. 

**Future work**
I would like to split the filter in to two classes one for each of the methods outlined here:

https://github.com/ArduPilot/ardupilot/blob/5a54d9a2ecbff9a32d26aa7feee72f0056a01027/libraries/Filter/LowPassFilter.h#L25-L39

The two methods should not be mixed, so separate classes will ensure that. 

Then I would like to add a filter class which includes the cutoff frequency parameter since that is a very common usecase.

### DIFF
--- a/libraries/Filter/LowPassFilter.cpp
+++ b/libraries/Filter/LowPassFilter.cpp
@@ -9,7 +9,6 @@
 #pragma GCC optimize("O2")
 #endif
 #include "LowPassFilter.h"
-#include <AP_InternalError/AP_InternalError.h>
 
 ////////////////////////////////////////////////////////////////////////////////////////////
 // DigitalLPF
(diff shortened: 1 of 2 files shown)


libraries/Filter/LowPassFilter.cpp
```diff
@@ -59,9 +40,9 @@ T DigitalLPF<T>::apply(const T &sample) {
 template <class T>
 void DigitalLPF<T>::compute_alpha(float sample_freq, float cutoff_freq) {
     if (sample_freq <= 0) {
-        alpha = 1;
+        _alpha = 1;
     } else {
-        alpha = calc_lowpass_alpha_dt(1.0/sample_freq, cutoff_freq);
+        _alpha = calc_lowpass_alpha_dt(1.0/sample_freq, cutoff_freq);
```
IamPete1, line 45: note that we are already using the `calc_lowpass_alpha_dt` function in this case.

libraries/Filter/LowPassFilter.cpp
```diff
@@ -24,33 +23,15 @@ DigitalLPF<T>::DigitalLPF() {
...
-    }
-    if (is_zero(cutoff_freq)) {
-        _output = sample;
-        return _output;
-    }
-    if (is_zero(dt)) {
-        return _output;
-    }
-    float rc = 1.0f/(M_2PI*cutoff_freq);
-    alpha = constrain_float(dt/(dt+rc), 0.0f, 1.0f);
```
andyp1per, line 40: should this constrain be added to calc_lowpass_alpha_dt?

IamPete1: See: https://github.com/ArduPilot/ardupilot/pull/20911

The constrain will never trigger because we have already checked both values are >= 0.

libraries/Filter/LowPassFilter.h
```diff
@@ -59,18 +59,18 @@ class DigitalLPF {
...
-        initialised = false;
+        _initialised = false;
     }
 
 private:
     T _output;
-    float alpha = 1.0f;
-    bool initialised;
+    float _alpha = 1.0f;
+    bool _initialised;
```
peterbarker, line 73: I thought we were trying to canonicalise the other way, to remove leading underscores over  time

tridge:
it would be good to be sure the two API approaches are not currently mixed

IamPete1:
replaced by https://github.com/ArduPilot/ardupilot/pull/27932

PR #27584, opened 2024-07-17
AP_InertialSensor: Check the gyro/accel id has not been previously registered

joshanne:
Discovered on CubeOrangePlus BG3 in a production environment.

On the very first boot of the CubeOrangePlus-BG3 after a parameter wipe, 4 IMU instances are registered. These register under the Parameters with the IDs:
* INS_ACC_ID       3867682
* INS_ACC2_ID      3867938
* INS_ACC3_ID      3867658 **<< This appears to be a legitimate ICM45686 discovered with CHECK_ICM45686_EXT in CHECK_IMU2_PRESENT**
* INS4_ACC_ID      3867658 **<< NOTICE DUPLICATE OF INS_ACC3_ID - This is added when AUX:3867658 is processed**

On subsequent boots, it appears the macro `ADD_BACKEND_AUX` filters the AUX IMU so it is never probed or registered in the first place.

The problem arises if you take the BG3 out of the box, flash AP and go straight into an accelcal on the first boot. Since IMU4 is registered, it calibrates for IMU4 as well. The next reboot then fails pre-arm check requiring 3D Accel Calibration (even though it's already been performed, because now IMU4 is not present).

I've tested this on CubeOrangePlus BG _**and**_ BG3 - and all the IMUs I expect to be present are present with this change in place.

Disclosure: I do not fully understand how AUX IMUs are intended to work, I couldn't find a relevant wiki page - as such this may not be the appropriate fix, but it seems to address the experienced issue.

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor.cpp
@@ -736,6 +736,14 @@ bool AP_InertialSensor::register_gyro(uint8_t &instance, uint16_t raw_sample_rat
         return false;
     }
 
+    // Loop over the existing instances and check if the instance already exists
+    for (uint8_t instance_to_check = 0; instance_to_check < _gyro_count; instance_to_check++) {
+        if ((uint32_t)_gyro_id(instance_to_check) == id) {
+            // if it does, then bail
+            return false;
+        }
+    }
+
     _gyro_raw_sample_rates[_gyro_count] = raw_sample_rate_hz;
     _gyro_over_sampling[_gyro_count] = 1;
     _gyro_raw_sampling_multiplier[_gyro_count] = INT16_MAX/radians(2000);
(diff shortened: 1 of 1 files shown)


andyp1per (approved):

libraries/AP_InertialSensor/AP_InertialSensor.cpp
```diff
@@ -736,6 +736,15 @@ bool AP_InertialSensor::register_gyro(uint8_t &instance, uint16_t raw_sample_rat
         return false;
     }
 
+    // Loop over the existing instances and check if the instance already exists
+    for (uint8_t instance_to_check = 0; instance_to_check < _gyro_count; instance_to_check++) {
+        if ((uint32_t)_gyro_id(instance_to_check) == id) {
+            // if it does, inform and bail
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "Failed to register - gyro id %u already registered", unsigned(id));
```
tridge, line 743: we should drop the warnings

tridge (changes requested):

joshanne:
To be very clear - this occurs because on the COP-BG3, the third IMU (3867658) is also the same ID as the Aux IMU (3867658).

On the first boot, there is no filtering of what is added to the probe list, because there are no parameters in storage to filter against. On subsequent boots, the filtering takes place normally.

HOWEVER, if you accelcal on that very first boot after a parameter wipe, you calibrate the "_4_" (3 is real, 1 is false and a duplicate detected) IMUs. Then on subsequent boots, prearm will fail because only the 3 real IMUs are detected (thanks to filtering of the probe list).

@bugobliterator - I'm unavailable to discuss during the devcall tomorrow, but feel free to discord message me privately or in our partner chat

libraries/AP_InertialSensor/AP_InertialSensor.cpp
joshanne: Warnings have been removed.

rmackay9:

joshanne:
Thanks!

PR #27585, opened 2024-07-17
VectorNav: Reorganize message packets

lashVN:
A few key outputs for logging are missing, and the current structure prevents 800Hz IMU data using TYPE::INS. I would split the messages into:

IMU packet, EKF packet, GNSS packet

There should also be additional logging messages for the new packets coming through

This both aids in debugging, and allows use of VectorNav uncertainty and status information for internal publishing and use

### DIFF
--- a/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
+++ b/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
@@ -16,6 +16,7 @@
   support for serial connected AHRS systems
  */
 
+#include <cstdint>
 #define ALLOW_DOUBLE_MATH_FUNCTIONS
 
 #include "AP_ExternalAHRS_config.h"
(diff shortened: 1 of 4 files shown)


lashVN:
@peterbarker @IamPete1 

I've tested that I can arm on a VN-210 and VN-100, and that all new logging messages output correct data. Please let me know if there's something else necessary.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -16,6 +16,7 @@
   support for serial connected AHRS systems
  */
 
+#include <cstdint>
```
IamPete1, line 19: Why do you need this?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -625,11 +518,7 @@ void AP_ExternalAHRS_VectorNav::process_ahrs_packet(const uint8_t *b)
 #if AP_COMPASS_EXTERNALAHRS_ENABLED
     {
         AP_ExternalAHRS::mag_data_message_t mag;
-        if (use_uncomp) {
```
IamPete1, line 628: This existing functionality has been removed?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.h
```diff
@@ -63,20 +63,23 @@ class AP_ExternalAHRS_VectorNav : public AP_ExternalAHRS_backend {
...
 
     uint8_t *pktbuf;
     uint16_t pktoffset;
     uint16_t bufsize;
 
-    struct VN_INS_packet1 *last_ins_pkt1;
-    struct VN_INS_packet2 *last_ins_pkt2;
+    struct VN_imu_packet const *latest_imu_packet = nullptr;
+    struct VN_INS_ekf_packet const *latest_ins_ekf_packet = nullptr;
+    struct VN_INS_gnss_packet const *latest_ins_gnss_packet = nullptr;
```
IamPete1, line 78:
```suggestion
    struct VN_imu_packet const *latest_imu_packet;
    struct VN_INS_ekf_packet const *latest_ins_ekf_packet;
    struct VN_INS_gnss_packet const *latest_ins_gnss_packet;
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.h
```diff
@@ -63,20 +63,23 @@ class AP_ExternalAHRS_VectorNav : public AP_ExternalAHRS_backend {
...
 
-    uint32_t last_pkt1_ms;
-    uint32_t last_pkt2_ms;
+    uint32_t last_pkt1_ms = UINT32_MAX;
+    uint32_t last_pkt2_ms = UINT32_MAX;
+    uint32_t last_pkt3_ms = UINT32_MAX;
```
IamPete1, line 82: Why not 0?

IamPete1 (commented):
You will also need to update the simulators to match. 

Is the total data rate more or less? What rate is it trying to write all the logs?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.h
lashVN: Agreed, will update

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.h
lashVN: Is it improper to initialize to nullptr?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: While the naming of uncompaccel/uncompgyro/uncompmag implies they're handled the same, a distinct difference is the mag (compensated) isn't estimated by the internal EKF; our on-board Hard Iron estimator is a pretty rarely used feature, so it seemed like unnecessary bloat and bytes to me (especially given we're using it in our high-rate message). If we think that's still valuable I can readd it.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: Agreed, will remove

lashVN:
> You will also need to update the simulators to match.
> 
> Is the total data rate more or less? What rate is it trying to write all the logs?

I will update the simulators in my next push.

The total data rate is the same by default (50Hz + 50Hz + 5Hz), but allows for higher IMU rates. I believe because the messages were combined, the INS (used on VN-300) could only achieve 400Hz + 5Hz. It can now output at 800Hz + 50Hz + 5Hz.

It may be reasonable to introduce a config parameter to allow customizing the update rate of the EKF packet (which most imporantly includes attitude) but I decided to defer that to later work.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.h
IamPete1: You get it for free in AP, doing it manually just costs flash.

lashVN: Oh interesting, didn't know that. Thanks! Will update

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
IamPete1: It is not only removed for mag? Its removed for accel/gyro too?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.h
IamPete1: If zero you don't need to initialize manually (you get it for frees link nullptr).

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
lashVN: It's available for accel/gyro, [here](https://github.com/ArduPilot/ardupilot/pull/27585/files#diff-d01800740fba47f0cbbd7c65d64588ac01381d0464bd10075b00e46ab0149cefL602)

IamPete1: Ah, OK. So is the mag compensated or not now? My preference would be uncompensated I think. Maybe it is possible to turn off the compensation on the VN side so the user can still pick if they really care?

My argument for uncompensated is that AP should be able to do a better job of calibration as it can use battery current if it needs to where as the VN does not have that information.

lashVN: It's technically using compensated, but unless the user has explicitly configured the VN to run the HSI estimator, the compensated and uncompensated outputs are both uncompensated.

tridge (approved):

PR #27588, opened 2024-07-18
Added MFT-SEMA100 Board

EternAlmox:


rmackay9:
Hi @EternAlmox,

Thanks for the PR!

A couple of small administrative issues:

- could you rebase on master so that the "merge" commits disappear?
- we always like to prefix our commits with the folder name (or subsystem name) in order to make backporting easier.  I allows us to quickly look at each commit and see which area of the code is affected.  Could you add a prefix to each of the commits?
  - rename "Add files via upload" to something like, "AP_HAL_ChibiOS: add support for MFT-SEMA100"
- I think it might be possible to "squash" some of the commits togethers.  For example, the two commits called "Add files via upload" and the "Create README.md" could be merged together.

Hwurzburg (changes requested):
Started to review this and had some issues in the defs...so putting off README review until these get resolve:
~~1. Biggest is that UART ORDER and configured UARTS do not match....cant tell what is really desired~~
2. The USB config is unfamiliar to me, and apparently CAN2 cant be used if normal HS USB is enabled? needs review by Tridge
~~3. The bootloader does not setup SD card which I think is desireable now~~
4. I would expect BiDir could be setup on some output groups
~~5. Dont understand why SERIAL4 baud is setup...should not be needed for GPS~~
~~6. Need to know if any CAN periphs are bundled with autopilot to decide on enabling CAN1 by default~~

EternAlmox:
Couldnt managed to add BiDir to any of the outputs. I might need professional help but couldnt find any yet.

There isnt any CAN periphs bundled with autopilot so no need to be enabled by default

Hwurzburg:
we dont put .apj in bootloaders dir....just .bin and also .hex...but .bin is the required one

Hwurzburg:
Removed .hex not .apj

tridge:
rebased and squashed commits

tridge (approved):

rmackay9:

PR #27589, opened 2024-07-18
SIH glider improvements

Georacer:
This is a continuation of https://github.com/ArduPilot/ardupilot/pull/27559.

Up to now, in a typical Simulation in Hardware (SIH) you could not enable glider and balloon simulation on SIH, because `SIM_config.h` would have at its end:
```c++
#ifndef AP_SIM_GLIDER_ENABLED
#define AP_SIM_GLIDER_ENABLED (CONFIG_HAL_BOARD == HAL_BOARD_SITL)
#endif
```
which evaluates to false for SIH.

This PR allows the use of `./Tools/scripts/sitl-on-hardware/sitl-on-hw.py --vehicle plane --simclass Glider --frame Plane` to capture the `simclass` and enable the `AP_SIM_GLIDER_ENABLED` flag.

Additionally, it fixes a bug where the balloon channel would initialize in a negative value, resulting in the balloon getting downwards velocity and going underground, dragging the glider with it.

To test it:
1. Run `./Tools/scripts/sitl-on-hardware/sitl-on-hw.py --board CubeOrangePlus-SimOnHardWare --vehicle plane --simclass Glider --frame Plane --upload` and upload to your flight controller (here assuming Cube Orange+).
2. `rc 3 1000`, `arm throttle`.
3. `servo set 6 1200` to launch the balloon and the plane altitude rise after a few seconds.
4. `servo set 6 2000` to have the balloon climb at the full `SIM_GLD_BLN_RATE`.
5. `servo set 10 2000` to cut the glider off the balloon. It will start falling to the ground shortly afterwards.

Unfortunately, its aerodynamic parameters will have it fall flat. RTLing cannot take over control. This is a known, existing issue.

### DIFF
--- a/Tools/scripts/sitl-on-hardware/sitl-on-hw.py
+++ b/Tools/scripts/sitl-on-hardware/sitl-on-hw.py
@@ -66,6 +66,8 @@ if args.defaults:
     defaults_write(open(args.defaults,"r").read() + "\n")
 
 if args.simclass:
+    if args.simclass == 'Glider':
+        hwdef_write("define AP_SIM_GLIDER_ENABLED 1\n")
     hwdef_write("define AP_SIM_FRAME_CLASS %s\n" % args.simclass)
 if args.frame:
     hwdef_write('define AP_SIM_FRAME_STRING "%s"\n' % args.frame)
--- a/libraries/SITL/SIM_Glider.cpp
+++ b/libraries/SITL/SIM_Glider.cpp
@@ -231,7 +231,7 @@ void Glider::calculate_forces(const struct sitl_input &input, Vector3f &rot_acce
     float aileron  = 0.5*(filtered_servo_angle(input, 1) + filtered_servo_angle(input, 4));
     float elevator = filtered_servo_angle(input, 2);
     float rudder   = filtered_servo_angle(input, 3);
-    float balloon  = filtered_servo_range(input, 5);
+    float balloon  = MAX(0.0f, filtered_servo_range(input, 5)); // Don't let the balloon receive downwards commands.
     float balloon_cut = filtered_servo_range(input, 9);
 
     // Move balloon upwards using balloon velocity from channel 6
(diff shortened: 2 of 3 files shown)


Tools/scripts/sitl-on-hardware/sitl-on-hw.py
```diff
@@ -66,6 +66,8 @@ def sohw_path(fname):
     defaults_write(open(args.defaults,"r").read() + "\n")
 
 if args.simclass:
+    if args.simclass == 'Glider':
+        hwdef_write("define AP_SIM_GLIDER_ENABLED 1\n")
```
peterbarker, line 70: Interesting approach.  We'll see if it works.

Alternate approach would be to enable the feature based on `AP_SIM_ENABLED` and always have it in there, even if it isn't actually usable.

libraries/SITL/SIM_Glider.cpp
```diff
@@ -127,7 +127,7 @@ Vector3f Glider::getTorque(float inputAileron, float inputElevator, float inputR
     float Mz = Cn * qPa * m.Sref * m.refSpan;
 
 
-#if 0
```
peterbarker, line 130: This is going to be *extremely* high rate.

Are you sure you want it in?

Have you checked the SoH logs to make sure we're not just dropping vast numbers of messages with this?

libraries/SITL/SIM_Glider.cpp
```diff
@@ -241,7 +241,7 @@ void Glider::calculate_forces(const struct sitl_input &input, Vector3f &rot_acce
         balloon_velocity = Vector3f(-wind_ef.x, -wind_ef.y, -wind_ef.z -balloon_rate * balloon);
         balloon_position += balloon_velocity * (1.0e-6 * (float)frame_time_us);
         const float height_AMSL = 0.01f * (float)home.alt - position.z;
-        // release at burst height or when channel 9 goes high
+        // release at burst height or when baloon cut output goes high
```
peterbarker, line 244:
```suggestion
        // release at burst height or when balloon cut output goes high
```

libraries/SITL/SIM_Glider.cpp
Georacer: Actually, I haven't put that much thought into this line.

I saw that the logging was always disabled and thought I'd enable it, since it looks equivalent to the `AP_Logger().Write` of the `Glider::getForce` just below.

Am I missing something? Are the enclosing `if`s of this one so different?
Should we perhaps delete this inaccessible log write instead?

tridge: I agree we need to leave it as #if 0, 1200Hz logging of AOA is not really needed

Tools/scripts/sitl-on-hardware/sitl-on-hw.py
Georacer: Maybe I should have mentioned, it does work on my local tests.
Unless you had some other testing in mind.

PR #27592, opened 2024-07-18
Add documentation for SIM_ACC?_RND parameters

peterbarker:


peterbarker:
@khancyr example of a PR which isn't undergoing any CI, when e.g. `test.Copter.Parameters` would be an appropriate CI test to run for the files being changed.

PR #27596, opened 2024-07-19
Add support for ZeroOne_V6X autopilot

ZeroOne-Aero:
Add ZeroOneX6_bl.bin and ZeroOneX6_bl.hex.


peterbarker:
I've cleaned up the commit history.

@ZeroOne-Aero I suggest using a topic branch rather than "master": https://ardupilot.org/dev/docs/git-branch.html

I've fixed the title of the PR to reflect it's adding a new autopilot.

I've requested that @Hwurzburg and @andyp1per have a look at your new board.

I have to query why this hwdef does not inherit from the Pixhawk6X hwdef file - it seems to be very, very similar.

peterbarker:
Note a closed duplicate PR - make sure concerns mentioned on that PR are addressed here.

andyp1per (approved):
Looks ok to me now. Henry will surely have comments on the README :)

peterbarker:
https://github.com/ArduPilot/ardupilot/pull/27646 is a more recent version of this

PR #27598, opened 2024-07-20
SITL: added support for up to 32 rotors

tridge:
a user asked if we could support up to 32 rotors, so here is a PR that demonstrates flying in SITL with 32 rotors
not really expecting this to be merged as is :-)


IamPete1:
32 might seem a bit mad, but there is almost no difference in flash cost vs just adding one or two more. We do have a PR to move to 16, https://github.com/ArduPilot/ardupilot/pull/21157.

I do think we will need support for more than 12 eventually.

libraries/SRV_Channel/SRV_Channels.cpp
```diff
@@ -355,7 +355,7 @@ SRV_Channels::SRV_Channels(void)
 #if NUM_SERVO_CHANNELS > 16
         if (i >= 16) {
             // default to GPIO, this disables the pin and stops logging
```
IamPete1, line 357: comment needs a update. Does none still stop logging?

peterbarker:
This code was merged as part of #29020; closing this now.

PR #27599, opened 2024-07-20
tools: firmware_version_decoder.py: fix detection for BETA+n firmware

Williangalvani:
The script is failing when we do `#define FW_TYPE FIRMWARE_VERSION_TYPE_BETA+1` and so on...

Required for https://github.com/bluerobotics/BlueOS/issues/2833


patrickelectric:
If this is for an case where FIRMWARE_VERSION_TYPE_BETA+N reflects in beta.N as release version, it should be fine.

patrickelectric (approved):

PR #27607, opened 2024-07-22
AP_Mount: Conditionally define `serial_instance` to fix unused variable compile error

mduclehcm:
Change:
- Wrapped the definition of `serial_instance` with preprocessor directives to ensure it is only defined when necessary.
- This resolves the compile error caused by the unused variable when no features requiring `serial_instance` are enabled.

### DIFF
--- a/libraries/AP_Mount/AP_Mount.cpp
+++ b/libraries/AP_Mount/AP_Mount.cpp
@@ -46,7 +46,7 @@ AP_Mount::AP_Mount()
     }
     _singleton = this;
 
-	AP_Param::setup_object_defaults(this, var_info);
+    AP_Param::setup_object_defaults(this, var_info);
 }
 
 // init - detect and initialise all mounts
@@ -184,6 +184,8 @@ void AP_Mount::init()
             set_mode_to_default(instance);
         }
     }
+
+    (void)serial_instance;
 }
 
 // update - give mount opportunity to update servos.  should be called at 10hz or higher


rmackay9:
Thanks very much!  I've tagged PeterB on this because he most likely knows best

libraries/AP_Mount/AP_Mount.cpp
```diff
@@ -63,8 +63,13 @@ void AP_Mount::init()
     // primary is reset to the first instantiated mount
     bool primary_set = false;
 
+#if HAL_MOUNT_STORM32SERIAL_ENABLED || \
+    HAL_MOUNT_SIYI_ENABLED || \
+    HAL_MOUNT_VIEWPRO_ENABLED || \
+    HAL_MOUNT_TOPOTEK_ENABLED
     // keep track of number of serial instances for initialisation
     uint8_t serial_instance = 0;
+#endif
```
peterbarker, line 72: There's an alternative to the #ifs here, and that's:
```suggestion
        (void)serial_instance;
#endif
```

It means when someone adds another serial driver they don't have to add their new thing here.

Thoughts?

peterbarker (approved):
LGeTM.

I might have voided it just under where it was declared, but this looks good, thanks!

rmackay9:
Hi @mduclehcm,

Any chance you could rebase this on master?  We don't normally allow "merge" commits.

peterbarker:
I've corrected the commit list

PR #27609, opened 2024-07-22
Tools: install-prereqs-ubuntu.sh: fix script after setuptools>=71.0.0 release

ben-xD:
Explicitly upgrade to use newest `packaging` to avoid `packaging` incompatibility with `setuptools>=71.0`

As https://github.com/pypa/setuptools/issues/4496#issuecomment-2240322375 mentions:
> setuptools>=71.0 does not work with packaging<24.0, but will not upgrade it if packaging is already installed

Also relevant: https://github.com/pypa/setuptools/issues/4496

### DIFF
--- a/Tools/environment_install/install-prereqs-arch.sh
+++ b/Tools/environment_install/install-prereqs-arch.sh
@@ -26,7 +26,7 @@ BASE_PKGS="base-devel ccache git gsfonts tk wget gcc"
 SITL_PKGS="python-pip python-setuptools python-wheel python-wxpython opencv python-numpy python-scipy"
 PX4_PKGS="lib32-glibc zip zlib ncurses"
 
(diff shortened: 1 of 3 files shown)


khancyr (changes requested):
LGTM, just need to fix the install on focal

peterbarker:
I've merged this as it fixes the Jammy environment test.

We should follow up with a fix for Focal as Pierre suggests, but it is a separate problem

Thanks @ben-xD !

ben-xD:
@rmackay9, does that mean testing the changes and opening 1 PR for each of branches:
```
  remotes/upstream/Copter-4.5                     9a755a82d7 Plane: version to 4.5.7-beta1
  remotes/upstream/Plane-4.5                      0523b2c7cf Plane: version to 4.5.7-beta1
  remotes/upstream/Rover-4.5                      7b5ef6df77 Plane: version to 4.5.7-beta1
  remotes/upstream/Sub-4.5                        cede211a7c CI: add navigator64 build
  remotes/upstream/Tracker-4.5                    22e8f1e84e Plane: version to 4.5.7-beta1
```

PR #27610, opened 2024-07-22
DroneCAN: fix dangerous bug for twin-motor Planes

robertlong13:
Without this, twin motor planes with DroneCAN ESCs need to set a
dummy throttle channel for the `scale_esc_to_unity` to work correctly.

Worse yet, if you don't do this, your ESCs will receive a max throttle command while disarmed as soon as the safety switch is disengaged.

The first commit prevents the max-throttle symptom when esc scaling is unconfigured or misconfigured (costs 16 bytes). The second commit allows the left/right throttle servos to be used to configure esc scaling (costs 24 bytes).

Params to reproduce this bug on CubeOrange, disengage the safety switch and look at RawCommand coming out of the autopilot in DroneCAN GUI
```
CAN_D1_UC_ESC_BM 3
CAN_P1_DRIVER 1
CAN_P2_DRIVER 1
SERIAL6_PROTOCOL 2
SERVO1_FUNCTION 74
SERVO1_MAX 2000
SERVO1_MIN 1000
SERVO2_FUNCTION 73
SERVO2_MAX 2000
SERVO2_MIN 1000
SERVO3_FUNCTION 0
SERVO4_FUNCTION 0
```

### DIFF
--- a/libraries/AP_DroneCAN/AP_DroneCAN.cpp
+++ b/libraries/AP_DroneCAN/AP_DroneCAN.cpp
@@ -709,14 +709,17 @@ void AP_DroneCAN::handle_node_info_request(const CanardRxTransfer& transfer, con
 
 int16_t AP_DroneCAN::scale_esc_output(uint8_t idx){
     static const int16_t cmd_max = ((1<<13)-1);
-    float scaled = 0;
-
+    float scaled = hal.rcout->scale_esc_to_unity(_SRV_conf[idx].pulse);
+    // Prevent invalid values (from misconfigured scaling parameters) from sending non-zero commands
+    if (!isfinite(scaled)) {
+        return 0;
+    }
+    scaled = constrain_float(scaled, -1, 1);
     //Check if this channel has a reversible ESC. If it does, we can send negative commands.
     if ((((uint32_t) 1) << idx) & _esc_rv) {
-        scaled = cmd_max * (hal.rcout->scale_esc_to_unity(_SRV_conf[idx].pulse));
+        scaled *= cmd_max;
     } else {
-        scaled = cmd_max * (hal.rcout->scale_esc_to_unity(_SRV_conf[idx].pulse) + 1.0) / 2.0;
-        scaled = constrain_float(scaled, 0, cmd_max);
+        scaled = cmd_max * (scaled + 1.0) / 2.0;
     }
 
     return static_cast<int16_t>(scaled);
--- a/ArduPlane/radio.cpp
+++ b/ArduPlane/radio.cpp
@@ -58,6 +58,8 @@ void Plane::set_control_channels(void)
         // setup correct scaling for ESCs like the UAVCAN ESCs which
         // take a proportion of speed. For quadplanes we use AP_Motors
         // scaling
+        g2.servo_channels.set_esc_scaling_for(SRV_Channel::k_throttleLeft);
+        g2.servo_channels.set_esc_scaling_for(SRV_Channel::k_throttleRight);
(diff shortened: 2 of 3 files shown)


libraries/AP_DroneCAN/AP_DroneCAN.cpp
```diff
@@ -706,11 +706,16 @@ int16_t AP_DroneCAN::scale_esc_output(uint8_t idx){
     static const int16_t cmd_max = ((1<<13)-1);
     float scaled = 0;
 
+    scaled = hal.rcout->scale_esc_to_unity(_SRV_conf[idx].pulse);
+    // Prevent invalid values (from misconfigured scaling parameters) from sending non-zero commands
```
IamPete1, line 710:
```suggestion
    const float scaled = hal.rcout->scale_esc_to_unity(_SRV_conf[idx].pulse);
    // Prevent invalid values (from misconfigured scaling parameters) from sending non-zero commands
```

IamPete1: Strictly I think we also need to constrain to -1 to 1. Maybe that would be better done in the `scale_esc_to_unity` method.

IamPete1 (commented):
The whole scale to unity thing really should be removed at somepoint. It will cause issues if you were to set different ranges on your throttles, or `PWM_MIN`/`PWM_MAX` for quadplane. 

We should calculate the scaled output per channel.

libraries/AP_DroneCAN/AP_DroneCAN.cpp
robertlong13: You're right. Only the non-reversed servo channels got any constraint applied. I have fixed that now by moving the contrain_float up out of the if block.

It would be nice to do it in `scale_esc_to_unity`, but we don't know inside that function whether "off" is 0.0 or -1.0, and no way to signal to the caller that something is wrong and they should not spin anything.

It's not the best solution, and you're absolutely right that we should eliminate this at some point, but I'm restricting myself to the smallest change possible to fix the bug.

tridge (changes requested):
I'd like to AP_HAL/RCOutput.h change to init the min/max to 1000/2000

robertlong13:
Finally got around to the testing to confirm that I haven't inadvertently messed with the math:

Params, all default except:
```
CAN_D1_UC_ESC_BM,4 # Servo 3 (throttle)
CAN_D1_UC_ESC_OF,2 # Offset by 2 (so only 3 comes in)
CAN_D1_UC_ESC_RV,4 # Servo 3 is reversable (turned this on and off to test both)
CAN_P1_DRIVER,1
CAN_P2_DRIVER,1
SERIAL6_PROTOCOL,2 # So I could have Planner and DroneCAN GUI connected simultaneously
```

Armed in manual and throttled up using vjoy through Mission Planner and looked at the RawCommand messages using DroneCAN GUI

Tested in/outs:
RC3 % | CMD (reversable) | CMD (non-reversable)
---|---|---
0 | 0 | -8191
25 | 2047 | -4095
50 | 4095 | 0
75 | 6143 | 4095
100 | 8191 | 8191

Also, I've made the change the Tridge requested.

And of course I also tested that this PR does still in fact fix the bug.

tridge (approved):

tridge:
@robertlong13 I do hope that table has swapped column headers for reversible/non-reversible?

robertlong13:
> @robertlong13 I do hope that table has swapped column headers for reversible/non-reversible?

Yup, swapped headers. Sorry. Editing the comment now

PR #27611, opened 2024-07-22
AP_HAL_SITL: initialize rcout safety state

robertlong13:
This allows BRD_SAFETY_DEFLT=1 to work correctly for SITL.

### DIFF
--- a/Tools/autotest/rover.py
+++ b/Tools/autotest/rover.py
@@ -6042,10 +6042,10 @@ Brakes have negligible effect (with=%0.2fm without=%0.2fm delta=%0.2fm)
     def RangeFinder(self):
         '''Test RangeFinder'''
         # the following magic numbers correspond to the post locations in SITL
-        home_string = "%s,%s,%s,%s" % (51.8752066, 14.6487840, 54.15, 315)
+        home_string = "%s,%s,%s,%s" % (51.8752066, 14.6487840, 54.15, 231)
 
         rangefinder_params = {
-            "SIM_SONAR_ROT": 6,
+            "SIM_SONAR_ROT": 0,
         }
         rangefinder_params.update(self.analog_rangefinder_parameters())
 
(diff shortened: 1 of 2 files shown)


IamPete1:
I would have expected a change in `AP_HAL_SITL` why does it need to be in the vehicle?

khancyr (commented):
looks wrong. 
you can pass BOARD_SAFETY_ENABLE_DEFAULT 1 to force it . I don't think that using safety button on SITL is convenient things

robertlong13:
> BOARD_SAFETY_ENABLE_DEFAULT

robertlong13:
> looks wrong. you can pass BOARD_SAFETY_ENABLE_DEFAULT 1 to force it . I don't think that using safety button on SITL is convenient things

Nope, that doesn't work. That's exactly what this PR is fixing. The entire BRD_SAFETY_DEFLT parameter is broken in SITL.

As for why I wanted this change:
1. I was trying to reproduce a bug in SITL that I was only able to reproduce on Cube. I was trying to see if safety had anything to do with it (it didn't) but there's no way at all for SITL to start with the safety switch on at boot.
2. Simulation training.

robertlong13:
> I would have expected a change in `AP_HAL_SITL` why does it need to be in the vehicle?

Yeah, looking at this in the cold light of day, I have no idea why I thought this was the best place to do this 😆. Fixed.

IamPete1 (approved):
LGTM. `BRD_SAFETY_DEFLT` is zero in SITL, so this should have no change in behavior unless you change it.

tridge (approved):

peterbarker:
I've rebased this on a fix for Rover's RangeFinder test.

I think what's happening is that the vehicle used to move forward a little but the change in safety state stops that happening.

We used to point the rangefinder laterally,so moving forward brought it in-line with the post.

Very strange.

PR #27612, opened 2024-07-22
Actions: fix tests that don't like repos not called "ardupilot"

IamPete1:
A few of the git safe directory calls assume the repo is called `ardupilot`, if your fork happens not to be called that then CI fails. GitHub provides us with a environmental variable to get the correct directory which we already use in most places. The tricky one is cygwin which uses the repo name, but on a different path, so in that case I have to some messing about to extract the repo name from `GITHUB_REPOSITORY` and `GITHUB_REPOSITORY_OWNER` (see https://tom-gallacher.co.uk/articles/how-to-github-actions-repo-name/)

This also ups the timeout for the environment install scripts from 45 mins to 60 mins. These seem to run a bit slower outside of the main repo, I'm not sure why.

### DIFF
--- a/.github/workflows/test_environment.yml
+++ b/.github/workflows/test_environment.yml
@@ -79,7 +79,7 @@ jobs:
         with:
             submodules: 'recursive'
       - name: test install environment ${{matrix.os}}.${{matrix.name}}
-        timeout-minutes: 45
+        timeout-minutes: 60
         env:
           DISABLE_MAVNATIVE: True
           DEBIAN_FRONTEND: noninteractive
--- a/.github/workflows/test_environment.yml
+++ b/.github/workflows/test_environment.yml
@@ -103,7 +103,7 @@ jobs:
             *"archlinux"*)
             cp /etc/skel/.bashrc /root
             cp /etc/skel/.bashrc /github/home
-            git config --global --add safe.directory /__w/ardupilot/ardupilot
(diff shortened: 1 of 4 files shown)


.github/workflows/test_environment.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         with:
             submodules: 'recursive'
       - name: test install environment ${{matrix.os}}.${{matrix.name}}
-        timeout-minutes: 45
+        timeout-minutes: 60
```
khancyr, line 82: why raising to 60 ? it should be a bit less 30min without cache . But not blocking

khancyr (approved):
LTGM

truely a git annoyance ...

khancyr:
for the question why it is faster on ardupilot, I think we got an upgraded plan as FOSS project an have better machine for CI.

khancyr:
We can merge without waiting environement tests to pass as they are failing due to some setuptools issues, so not related

rmackay9:
My repo is one of these actually

tridge (approved):

tridge:
@khancyr can you do a PR against this PR that fixes the tests so env install works again?

peterbarker:
The Jammy environment install was fixed in https://github.com/ArduPilot/ardupilot/pull/27609

Focal is still outstanding

PR #27616, opened 2024-07-23
AP_Mount: Topotek safely parses version packet

rmackay9:
This corrects a potential issue in the Topotek driver (introduced this morning by PR https://github.com/ArduPilot/ardupilot/pull/27568) that could lead to writing past the end of the "version" variable while parsing the version packet received from the gimbal.

This has been lightly tested on real hardware but the exact code path affected is not run due to the hardware I'm using sending the old style version string
image

Thanks @peterbarker for finding this issue

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Topotek.cpp
+++ b/libraries/AP_Mount/AP_Mount_Topotek.cpp
@@ -1060,15 +1060,15 @@ void AP_Mount_Topotek::gimbal_version_analyse()
     const uint8_t data_buf_len = char_to_hex(_msg_buff[5]);
 
     // check for "."
-    bool constains_period = false;
+    bool contains_period = false;
     for (uint8_t i = 0; i < data_buf_len; i++) {
-        constains_period |= _msg_buff[10 + i] == '.';
+        contains_period |= _msg_buff[10 + i] == '.';
     }
 
     // if contains period, extract version number
     uint32_t ver_num = 0;
     uint8_t ver_count = 0;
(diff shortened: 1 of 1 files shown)


libraries/AP_Mount/AP_Mount_Topotek.cpp
```diff
@@ -1060,16 +1060,16 @@ void AP_Mount_Topotek::gimbal_version_analyse()
...
+        contains_period |= _msg_buff[10 + i] == '.';
     }
 
     // if contains period, extract version number
     uint32_t ver_num = 0;
     uint8_t ver_count = 0;
-    if (constains_period) {
-        for (uint8_t i = 0; i < data_buf_len; i++) {
+    if (contains_period) {
+        for (uint8_t i = 0; i < MIN(data_buf_len, ARRAY_SIZE(version)); i++) {
```
peterbarker, line 1072:
```suggestion
        for (uint8_t i = 0; i < data_buf_len; i++) {
            if (ver_count >= ARRAY_SIZE(version)) {
                break;
            }
```

peterbarker (commented):
I don't think your patch will parse enough of the inbound string.

My suggestion is a bit simpler, I think?

libraries/AP_Mount/AP_Mount_Topotek.cpp
rmackay9: yes, of course, thanks!

tridge (approved):

Issue #27517, opened 2024-07-10
Copter: if SmartRTL is interrupted on its path home, it should restore the last waypoint it was heading to


rmackay9:
When SmartRTL is on its way home, removes a point from the return path and gives it to the navigation controller.  The problem is that if the user switches out of SmartRTL to another mode, that point is not restored back onto the return path.  This means that it is lost and when the user switches back into SmartRTL again, it will skip that point

An extreme situation should be reproduce-able in SITL.  Fly a path in a manual-ish mode (like Loiter) with some noticeable corners.  Engage SmartRTL but then engage Brake mode soon after.  Repeat switching between SmartRTL and Brake and with each switch a point should be lost.  Eventually the vehicle should attempt to fly straight home (which is not good)

rmackay9:
I believe I have a fix for this issue now https://github.com/ArduPilot/ardupilot/pull/27619

peterbarker:
We've merged Randy's fix now.

PR #27619, opened 2024-07-23
Copter: SmartRTL restores last waypoint if users switches to another mode

rmackay9:
This resolves issue https://github.com/ArduPilot/ardupilot/issues/27517 by backing up the current destination and then adding it back to the path if the user exits SmartRTL mode while returning.

This has been tested in SITL and resolves the problem.  During testing I created a zig zag mission as shown below and after the vehicle reached the final waypoint on the top-left I repeatedly switched the vehicle back and forth between SmartRTL and Brake mode about 20 times.

In master when finally switched to SmartRTL for the last time, the vehicle simply landed in place.
image

With this PR applied the vehicle correctly flew the zig-zag pattern home (apologies for mixing MP and MAVProxy views).
image

### DIFF
--- a/ArduCopter/mode.h
+++ b/ArduCopter/mode.h
@@ -1541,6 +1541,10 @@ private:
     // point while following our path home.  If we take too long we
     // may choose to land the vehicle.
     uint32_t path_follow_last_pop_fail_ms;
+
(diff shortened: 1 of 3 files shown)


rmackay9:

PR #27620, opened 2024-07-23
AP_RangeFinder: populate last_reading_ms in the backend for some sensors

rishabsingh3003:
This was noticed when trying to use Lightware LW20 I2C with AP_Periph. The Rangefinder support inside AP_Periph checks for the backends last update time so that we don't send duplicates: 

https://github.com/ArduPilot/ardupilot/blob/master/Tools/AP_Periph/rangefinder.cpp#L67

I did a drive-through of all the other backend and it looks like only one other backend suffers from the issue. 

This problem has been reported a few times in the past: https://discuss.ardupilot.org/t/ap-periph-not-working-with-lightwarei2c/95913

### DIFF
--- a/libraries/AP_RangeFinder/AP_RangeFinder_HC_SR04.cpp
+++ b/libraries/AP_RangeFinder/AP_RangeFinder_HC_SR04.cpp
@@ -99,7 +99,7 @@ void AP_RangeFinder_HC_SR04::update(void)
     const uint32_t now = AP_HAL::millis();
     if (value_us == 0) {
         // no reading; check for timeout:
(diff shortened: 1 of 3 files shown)


rmackay9:
@rishabsingh3003 this only affect AP_Periph or are we losing readings on regular vehicles as well?

rishabsingh3003:
@rubenp02 RNGFND1_TYPE not showing up is likely a hwdef/firware build issue unrelated to this PR.

tridge (approved):

rmackay9 (approved):

rmackay9:

PR #27623, opened 2024-07-23
board_types.txt: Reserving board IDs for Easy Aerial

rishabsingh3003:
Reserve board IDs for some new Easy Aerial boards.

### DIFF
--- a/Tools/AP_Bootloader/board_types.txt
+++ b/Tools/AP_Bootloader/board_types.txt
@@ -367,6 +367,8 @@ AP_HW_ZeroOne_GNSS                   5602
 
 # IDs 6600-6699 reserved for Eagle Eye Drones
 
+# IDs 6900-6909 reserved for Easy Aerial
+
 # IDs 7000-7099 reserved for CUAV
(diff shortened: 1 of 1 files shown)


Tools/AP_Bootloader/board_types.txt
```diff
@@ -367,6 +367,8 @@ AP_HW_ZeroOne_GNSS                   5602
 
 # IDs 6600-6699 reserved for Eagle Eye Drones
 
+# IDs 6900-6999 reserved for Easy Aerial
```
peterbarker, line 370: Does Easy-Aerial need 100 IDs?

If we allocate 100 IDs to every block allocation we can only have 90 allocations in here.

We've been too liberal, really.

tridge: ok, lets do it in blocks of 10

tridge (changes requested):

peterbarker:
Per DevCall discussion, I've adjusted the allocation to 10 IDs and added a comment and marked this as MergeOnCIPass

PR #27626, opened 2024-07-23
autotest: test mag fusion

clydemcqueen:
Test several EK3_MAG_CAL values for sub

### DIFF
--- a/Tools/autotest/ardusub.py
+++ b/Tools/autotest/ardusub.py
@@ -36,6 +36,22 @@ class Joystick():
     Lateral = 6
 
 
+# Values for EK3_MAG_CAL
+class MagCal():
+    WHEN_FLYING = 0
+    WHEN_MANOEUVRING = 1
+    NEVER = 2
+    AFTER_FIRST_CLIMB = 3
+    ALWAYS = 4
+
+
+# Values for XKFS.MAG_FUSION
+class MagFuseSel():
+    NOT_FUSING = 0
+    FUSE_YAW = 1
+    FUSE_MAG = 2
+
+
 class AutoTestSub(vehicle_test_suite.TestSuite):
     @staticmethod
     def get_not_armable_mode_list():
@@ -821,7 +837,7 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
         # restart GPS driver
         self.reboot_sitl()
 
-    def backup_home(self):
+    def backup_origin(self):
         """Test ORIGIN_LAT and ORIGIN_LON parameters"""
 
         self.context_push()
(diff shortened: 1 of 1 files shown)


clydemcqueen:
@Williangalvani FYI, not hugely important, but may catch future breaking changes

CMakeLists.txt
```diff
@@ -0,0 +1 @@
+/home/clyde/projects/ex_ardu_ign/ap_cmake/ap_master_CMakeLists.txt
```
peterbarker, line 1: Shouldn't be here?!

Tools/autotest/ardusub.py
```diff
@@ -852,6 +868,58 @@ def backup_home(self):
...
+        self.delay_sim_time(1)
+        self.assert_mag_fusion_selection(MagFuseSel.FUSE_YAW)
+
+        # ALWAYS
+        self.set_parameters({'EK3_MAG_CAL': MagCal.ALWAYS})
+        self.reboot_sitl()
+        self.wait_ready_to_arm()
+        self.assert_mag_fusion_selection(MagFuseSel.FUSE_MAG)
+
+        self.context_pop()
```
peterbarker, line 921:
```suggestion
```

the harness gives you an implicit context

Tools/autotest/ardusub.py
```diff
@@ -852,6 +868,58 @@ def backup_home(self):
...
+            m = mlog.recv_match(type='XKFS')
+            if m is None:
+                break
+            found_sel = m.MAG_FUSION
+        if found_sel != expect_sel:
+            raise NotAchievedException("Expected mag fusion selection %d, found %d" % (expect_sel, found_sel))
+
+    def fuse_mag(self):
+        """Test EK3_MAG_CAL values"""
+        self.context_push()
```
peterbarker, line 886:
```suggestion
```

peterbarker (changes requested):
Nice.

CMakeLists.txt
clydemcqueen: Doh! Thanks for catching this

Tools/autotest/ardusub.py
```diff
@@ -879,7 +944,8 @@ def tests(self):
             self.MAV_CMD_CONDITION_YAW,
             self.TerrainMission,
             self.SetGlobalOrigin,
-            self.backup_home,
+            self.backup_origin,
+            self.fuse_mag,
```
peterbarker, line 948: Just as a future cleanup, these don't match our naming convention for tests, which you can see in the list above.

Concept is that you can visually distinguish helper functions from tests (kind of entry points, really) that way.

peterbarker (approved):
LGTM

Tools/autotest/ardusub.py
clydemcqueen: Ack, I was confused about naming, thanks for clearing this up

PR #27628, opened 2024-07-24
AC_FENCE:Add AUTOENABLE deprecation warning

Hwurzburg:
Can be changed to prearm checkfailed in 4.7

### DIFF
--- a/libraries/AC_Fence/AC_Fence.cpp
+++ b/libraries/AC_Fence/AC_Fence.cpp
@@ -429,6 +429,17 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
         hal.util->snprintf(failure_msg, failure_msg_len, "Fences enabled, but none selected");
         return false;
     }
+    
+    // if AUTOENABLE = 1 or 2 warn now, but fail in a later release
+    // PARAMETER_CONVERSION - Added: Jul-2024 for ArduPilot-4.6
+    if (_auto_enabled == 1 || _auto_enabled == 2) {
+        static uint32_t last_autoenable_warn_ms;
+        const uint32_t now_ms = AP_HAL::millis();
+        if (now_ms - last_autoenable_warn_ms > 60000) {
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "FENCE_AUTOENABLE is %u, will be removed in 4.7, use 3", unsigned(_auto_enabled));
+            last_autoenable_warn_ms = now_ms;
+        }
+    }
 
     // if not enabled or not fence set-up always return true
     if ((!enabled() && !_auto_enabled) || !_configured_fences) {


libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -429,6 +429,16 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
         hal.util->snprintf(failure_msg, failure_msg_len, "Fences enabled, but none selected");
         return false;
     }
+    
+    // if AUTOENABLE = 1 or 2 warn now, but fail in a later release
+    static uint8_t i = 0 ;
+    i++;
+    if (i > 60) {
```
peterbarker, line 436:
```suggestion
    // PARAMETER_CONVERSION - Added: Jul-2024 for ArduPilot-4.6
    const uint32_t now_ms = AP_HAL::millis();
    if (now_ms - last_autoenable_warn_ms > 60000) {
```

... and add last_autoenable_warn_ms in the header

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -429,6 +429,16 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
...
+        if (_auto_enabled == 1 or _auto_enabled == 2) {
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "FENCE_AUTOENABLE set to 1 or 2, will be deprecated next release, use 0 or 3");
+            i = 0;
```
peterbarker, line 439:
```suggestion
```

can't use a counter to decimate as this can be called at variable rates.

peterbarker (changes requested):

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -429,6 +429,17 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
...
         return false;
     }
+    
+    // if AUTOENABLE = 1 or 2 warn now, but fail in a later release
+    // PARAMETER_CONVERSION - Added: Jul-2024 for ArduPilot-4.6
+    const uint32_t now_ms = AP_HAL::millis();
+    static uint32_t _last_autoenable_warn_ms = now_ms - 60000; //needs this static since function is const
+    if (now_ms - _last_autoenable_warn_ms > 60000) {
+        if (_auto_enabled == 1 or _auto_enabled == 2) {
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "FENCE_AUTOENABLE set to 1 or 2, will be deprecated next release, use 0 or 3");
```
peterbarker, line 439:
```suggestion
            GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "FENCE_AUTOENABLE is %u, will be deprecated next release, use 0 or 3", unsigned(_auto_enabled));
```

peterbarker (changes requested):
Commit messageshould be `AC_Fence`., not AC_FENCE

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -429,6 +429,17 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
...
         return false;
     }
+    
+    // if AUTOENABLE = 1 or 2 warn now, but fail in a later release
+    // PARAMETER_CONVERSION - Added: Jul-2024 for ArduPilot-4.6
+    const uint32_t now_ms = AP_HAL::millis();
+    static uint32_t _last_autoenable_warn_ms = now_ms - 60000; //needs this static since function is const
+    if (now_ms - _last_autoenable_warn_ms > 60000) {
+        if (_auto_enabled == 1 or _auto_enabled == 2) {
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "FENCE_AUTOENABLE is %u, will be deprecated in 4.7, use 3", unsigned(_auto_enabled));
```
peterbarker, line 439:
```suggestion
            GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "FENCE_AUTOENABLE is %u, will be removed in 4.7, use 3", unsigned(_auto_enabled));
```

peterbarker (changes requested):

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -429,6 +429,17 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
         hal.util->snprintf(failure_msg, failure_msg_len, "Fences enabled, but none selected");
         return false;
     }
+    
+    // if AUTOENABLE = 1 or 2 warn now, but fail in a later release
+    // PARAMETER_CONVERSION - Added: Jul-2024 for ArduPilot-4.6
+    const uint32_t now_ms = AP_HAL::millis();
+    static uint32_t _last_autoenable_warn_ms = now_ms - 60000; //needs this static since function is const
```
andyp1per, line 436: These timings seem fairly arbitrary and pre-arm already has its own timings for output. I would just make this a one-line message and print every time.

andyp1per (changes requested):
I think this could be simpler, but I would also at least like to have the discussion about these options first before simply deprecating them. They were put in for a reason so let's at least make sure that that reason is no longer valid.

libraries/AC_Fence/AC_Fence.cpp
Hwurzburg: its not a pre-arm yet and we dont want to spam at 30 second rate, nor to make it a one off thats missed

Hwurzburg:
Tridge asked for this....the autoenable was a series of patches to correct issues with fence as far as I can tell....
1 was to allow takeoffs when fences would otherwise interfere if enabled

but then 2 was needed when alt min was added to allow landings...

3 was the only really correct way....

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -429,6 +429,17 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
...
+    if (now_ms - _last_autoenable_warn_ms > 60000) {
+        if (_auto_enabled == 1 or _auto_enabled == 2) {
```
tridge: should be || not or

tridge: if for the setting should be on outside

libraries/AC_Fence/AC_Fence.cpp
tridge: the static var should not be initialised at all

tridge (approved):

andyp1per (approved):

Hwurzburg:
@peterbarker merge on CI pass?

PR #27630, opened 2024-07-24
AP_Scripting: reduce flash usage

tpwrules:
A pile of optimizations to the way bindings are generated and used in order to improve the code and reduce flash usage. Please see the commit messages individually for details.

Saves ~4.3K flash on Cube Orange. There are no user visible script changes except for error message text. Tested in SITL and on the bench through my REPL and some example scripts.

### DIFF
--- a/libraries/AP_Scripting/generator/src/main.c
+++ b/libraries/AP_Scripting/generator/src/main.c
@@ -1902,11 +1902,6 @@ void emit_userdata_method(const struct userdata *data, const struct method *meth
   // emit comments on expected arg/type
   struct argument *arg = method->arguments;
 
-  if ((data->ud_type == UD_SINGLETON) && !(data->flags & UD_FLAG_LITERAL)) {
-      // fetch and check the singleton pointer
-      fprintf(source, "    %s * ud = check_%s(L);\n", data->name, data->sanatized_name);
-  }
-
   // emit warning if configured
   if (method->deprecate != NULL) {
     fprintf(source, "    static bool warned = false;\n");
@@ -1915,7 +1910,6 @@ void emit_userdata_method(const struct userdata *data, const struct method *meth
     fprintf(source, "        warned = true;\n");
     fprintf(source, "    }\n\n");
   }
-
 
   // sanity check number of args called with
   arg_count = 1;
@@ -1933,8 +1927,12 @@ void emit_userdata_method(const struct userdata *data, const struct method *meth
       fprintf(source, "    %s * ud = check_%s(L, 1);\n", data->name, data->sanatized_name);
       break;
     case UD_SINGLETON:
+      if (!(data->flags & UD_FLAG_LITERAL)) {
+        // fetch and check the singleton pointer
+        fprintf(source, "    %s * ud = check_%s(L);\n", data->name, data->sanatized_name);
+      }
+      break;
     case UD_GLOBAL:
-      // this was bound early
       break;
     case UD_AP_OBJECT:
       // extract the userdata, it was a pointer, so we need to grab it
--- a/libraries/AP_Scripting/generator/src/main.c
+++ b/libraries/AP_Scripting/generator/src/main.c
@@ -2459,15 +2459,11 @@ void emit_index(struct userdata *head) {
     }
 
     fprintf(source, "static int %s_index(lua_State *L) {\n", node->sanatized_name);
-    fprintf(source, "    const char * name = luaL_checkstring(L, 2);\n");
-    fprintf(source, "    if (load_function(L,%s_meta,ARRAY_SIZE(%s_meta),name)",node->sanatized_name,node->sanatized_name);
+    fprintf(source, "    return load_function(L,%s_meta,ARRAY_SIZE(%s_meta))",node->sanatized_name,node->sanatized_name);
     if (node->enums != NULL) {
-      fprintf(source, " || load_enum(L,%s_enums,ARRAY_SIZE(%s_enums),name)",node->sanatized_name,node->sanatized_name);
+      fprintf(source, " || load_enum(L,%s_enums,ARRAY_SIZE(%s_enums))",node->sanatized_name,node->sanatized_name);
     }
-    fprintf(source, ") {\n");
-    fprintf(source, "        return 1;\n");
-    fprintf(source, "    }\n");
-    fprintf(source, "    return 0;\n");
+    fprintf(source, ";\n");
     fprintf(source, "}\n");
     end_dependency(source, node->dependency);
     fprintf(source, "\n");
@@ -2992,14 +2988,15 @@ void emit_docs(struct userdata *node, int is_userdata, int emit_creation) {
 
 
 void emit_index_helpers(void) {
-  fprintf(source, "static bool load_function(lua_State *L, const luaL_Reg *list, const uint8_t length, const char* name) {\n");
+  fprintf(source, "static int load_function(lua_State *L, const luaL_Reg *list, const uint8_t length) {\n");
+  fprintf(source, "    const char * name = luaL_checkstring(L, 2);\n");
   fprintf(source, "    for (uint8_t i = 0; i < length; i++) {\n");
   fprintf(source, "        if (strcmp(name,list[i].name) == 0) {\n");
   fprintf(source, "            lua_pushcfunction(L, list[i].func);\n");
-  fprintf(source, "            return true;\n");
+  fprintf(source, "            return 1;\n");
   fprintf(source, "        }\n");
   fprintf(source, "    }\n");
-  fprintf(source, "    return false;\n");
+  fprintf(source, "    return 0;\n");
   fprintf(source, "}\n\n");
 
   // If enough stuff is defined out we can end up with no enums.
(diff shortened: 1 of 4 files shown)


tridge:
nice work!
really needs review from @IamPete1

libraries/AP_Scripting/generator/src/main.c
```diff
@@ -2495,16 +2491,19 @@ void emit_type_index_with_operators(struct userdata * data, char * meta_name) {
...
+  emit_singleton_type_index(parsed_singletons, "singleton");
+  emit_ap_object_type_index(parsed_ap_objects, "ap_object");
 
   fprintf(source, "static int binding_index(lua_State *L) {\n");
   fprintf(source, "    const char * name = luaL_checkstring(L, 2);\n");
   fprintf(source, "\n");
   fprintf(source, "    bool found = false;\n");
   fprintf(source, "    for (uint32_t i = 0; i < ARRAY_SIZE(singleton_fun); i++) {\n");
   fprintf(source, "        if (strcmp(name, singleton_fun[i].name) == 0) {\n");
+  fprintf(source, "            if (singleton_fun[i].getter && (*singleton_fun[i].getter)() == nullptr) {\n");
```
IamPete1, line 2504: Hum, the assumption is that if its available when it is first called then it will be available at any point afterwards. I'm not sure that is valid in all case, and even if it is now it may not be in the future. 

How about we keep the runtime nulprt check but replace the error string with a generic one. That should keep the a lot of the saving and protect us against the edge case.

libraries/AP_Scripting/generator/src/main.c
```diff
@@ -1844,9 +1829,20 @@ void emit_singleton_fields() {
...
+  const struct argument *count_arg = arg;
+  while (count_arg != NULL) {
+    if (count_arg->type.flags & (TYPE_FLAGS_NULLABLE | TYPE_FLAGS_REFERNCE)) {
+      return_count++;
+    }
+    count_arg = count_arg->next;
+  }
+  // add one to have a spare stack slot for userdata creation funcs
+  fprintf(source, "%sif (%d > LUA_MINSTACK) {\n", tab, return_count+1);
+  fprintf(source, "%s  luaL_checkstack(L, %d, nullptr);\n", tab, return_count+1);
```
IamPete1, line 1842:
```suggestion
  fprintf(source, "%s    luaL_checkstack(L, %d, nullptr);\n", tab, return_count+1);
```

libraries/AP_Scripting/generator/src/main.c
```diff
@@ -1844,9 +1829,20 @@ void emit_singleton_fields() {
...
+  // count arguments to return so we know if we need to check the stack
+  const struct argument *count_arg = arg;
+  while (count_arg != NULL) {
+    if (count_arg->type.flags & (TYPE_FLAGS_NULLABLE | TYPE_FLAGS_REFERNCE)) {
+      return_count++;
+    }
+    count_arg = count_arg->next;
+  }
+  // add one to have a spare stack slot for userdata creation funcs
+  fprintf(source, "%sif (%d > LUA_MINSTACK) {\n", tab, return_count+1);
```
IamPete1, line 1841: I guess this could be a `#if` too. Hopefully the compiler is smart enough that the result is the same.

libraries/AP_Scripting/lua/src/lobject.c
```diff
@@ -63,6 +63,13 @@ int luaO_fb2int (int x) {
 ** Computes ceil(log2(x))
 */
 int luaO_ceillog2 (unsigned int x) {
+  // never called with 0
+#if defined(__GNUC__) || defined(__clang__)
+  const int bitwidth = CHAR_BIT*sizeof(x);
+  x--;
+  const int clz = x ? __builtin_clz(x) : bitwidth; // x == 0 is undefined
+  return bitwidth-clz;
+#else
```
IamPete1, line 72: its not clear by inspection that these are the same. I wonder if we could run both in SITL and panic if they don't match. 

We also typically have been putting `#if defined(ARDUPILOT_BUILD)` round things we change in the lua source just so we can find it again. 

I'm not sure this is worth 300 bytes, on the scale of how much scripting costs its a small fraction.

IamPete1 (commented):
Some nice improvements, thanks!

libraries/AP_Scripting/generator/src/main.c
tpwrules: Most of the Ardupilot code base doesn’t check for null pointers. I almost removed the check entirely except for the possibly unique case of Motors dynamic. The function call overhead is also just large compared to the simple pointer read the singleton load usually gets compiled to.

I will fiddle with this more but possibly bring it back as a separate PR to discuss in more detail.

libraries/AP_Scripting/lua/src/lobject.c
tpwrules: It is going to be faster too. I did a bunch of tests to confirm that it’s the same but I can see the apprehension.

I’m probably going to bring this back as a separate PR.

libraries/AP_Scripting/generator/src/main.c
IamPete1: > Most of the Ardupilot code base doesn’t check for null singletons.

If you grep for `get_singleton` I think you will find we do check in the vast majority of cases.

tpwrules: I take back my statement about "most", there are definitely a lot of checks. But the pattern is still confusing to me.

Take e.g. `AP_Logger`. Most (but not all) of the 32 calls to `AP_Logger::get_singleton()`, do check for nullptr. However they also all seem to be guarded by `#if HAL_LOGGING_ENABLED` But the singleton is a member variable of AP_Vehicle and AP_Vehicle's subclass is a static variable so it's unclear how it could ever be nullptr.

Additionally, there are 311 calls to `AP::logger()` which returns the singleton without checking for nullptr. Some of these calls are protected by a previous check for nullptr, but certainly not all. Some singleton getter methods on `AP` return pointers, does that mean those are more likely to be nullptr?

Are there any "rules of Ardupilot singletons" anywhere? I don't really know what the developer intended guarantees are for AP_Logger, much less all of them. If I want to use the logger in a new feature, which of the three ways should I do it?

IamPete1: I'm probably not the person to answer that. I think we need to keep to the existing pattern here in scripting for now. As you say, if we were going to make it consistent across all singleton users then we could also remove the checks here. But that is outside the scope of this PR.

libraries/AP_Scripting/generator/src/main.c
tpwrules: Done.

libraries/AP_Scripting/generator/src/main.c
tpwrules: Makes more sense, done.

libraries/AP_Scripting/generator/src/main.c
```diff
@@ -1742,8 +1730,7 @@ void emit_field(const struct userdata_field *field, const char* object_name, con
         fprintf(source, "%slua_pushinteger(L, static_cast<int32_t>(%s%s%s%s));\n", indent, object_name, object_access, field->name, index_string);
         break;
       case TYPE_UINT32_T:
-        fprintf(source, "%snew_uint32_t(L);\n", indent);
-        fprintf(source, "%s*check_uint32_t(L, -1) = %s%s%s%s;\n", indent, object_name, object_access, field->name, index_string);
+        fprintf(source, "%s*new_uint32_t(L) = %s%s%s%s;\n", indent, object_name, object_access, field->name, index_string);
```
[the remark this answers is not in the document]
tpwrules: Not sure what you mean? `check_uint32_t` isn't reference anywhere in the generated bindings after this PR. Were you looking at an old copy?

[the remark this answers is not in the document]
IamPete1: Hum, yeah your right, I guess I got confused by the diff.

libraries/AP_Scripting/generator/src/main.c
tpwrules: I took your suggestion of simplifying the error message, we can come back to singletons later.

libraries/AP_Scripting/lua/src/lobject.c
tpwrules: This stands well on its own, dropping for now.

libraries/AP_Scripting/generator/src/main.c
IamPete1: Sorry, I what I had in mind was that it was OK to have a simple error message if you were doing the verbose check on first load.

Although I guess this might not be too painful. Let me see what the error look like. 

```
Lua: ./scripts/MotorMatrix_setup.lua:4: calling 'add_motor_raw' on bad self (singleton not available)
```
I think its OK. That should be enough to give a bit of a clue. Although not as good as what we had. I guess wait and see how much flash we get ;)

tpwrules: Removed the changes here entirely, couldn't come up with anything that would preserve the message and save a meaningful amount of flash.

IamPete1 (approved):
LGTM, thanks.

PR #27635, opened 2024-07-24
AC_FENCE:correct prearm message

Hwurzburg:
NFC....can be inside or outside a fence to be in breach

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -1498,7 +1498,7 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
         self.load_fence("fence-in-middle-of-nowhere.txt")
 
         self.delay_sim_time(5) # let fence check run so it loads-from-eeprom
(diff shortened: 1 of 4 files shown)


rmackay9:
Thanks, let's get @andyp1per to approve to be sure it doesn't interfere with his fence-min-alt PR

peterbarker (approved):

peterbarker:
> Thanks, let's get @andyp1per to approve to be sure it doesn't interfere with his fence-min-alt PR

We've already merged that one.  Sorry, perhaps that was ahead of time?!

Merging this one.

peterbarker (changes requested):
Not really NFC - even changing a string like this is functional - as the autotests failing show.

Autotests must pass

peterbarker:
... commit message should also be `AC_Fence` rather than `AC_FENCE`

andyp1per (commented):
Sorry, I did see this Henry but didn't have time to check where it was being changed. I am ok with this if the tests are fixed to match.

Issue #26650, opened 2024-03-29
Copter: No wind estimate when using EK3_SRC1_YAW = GPS


TheRealKaufmann:
## Bug report

**Issue details**
I am using a HERE 4 Dual GPS Configuration for Yaw and it works really well. However, when selecting EK3_SRC1_YAW = (GPS or GPS with Compass Fallback), there are no wind estimates shown in the wind_vel and wind_dir messages. Also, there is no wind estimate shown in the Log under VWN and VWE. The EK3_DRAG_BCOEF_* and EK3_DRAG_MCOEF parameters have been configures correctly. When selecting EK3_SRC1_YAW = Compass, the wind estimates are shown and available. 

**Version**
ArduCopter 4.4.4

**Platform**
[  ] All
[  ] AntennaTracker
[ X ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
QUAD X

**Hardware type**
Cube Orange

**Logs**
-

rmackay9:
Thanks for the report, I've added this to the [Copter-4.5.0 issues list](https://github.com/ArduPilot/ardupilot/issues/26103) although I see you said the same issue appears with 4.4.

In any case, I'll try to reproduce the issue in SITL.

TheRealKaufmann:
@rmackay9 Thank you! Could you please give me a hint where in the code this problem might arise? 
I might be able to work it through, if it is only in the EKF. For us, this is a fairly critical thing since we want to take off on a large ship with strong winds soon.

rmackay9:
This issue can be reproduced in SITL meaning that it is likely caused by a bug in the EKF.  The steps to reproduce are below:

First confirm that wind estimates work with all default parameters

1. cd ardupilot/ArduCopter
2. ../Tools/autotest/sim_vehicle.py --map --console
3. param load [../Tools/autotest/default_params/copter-wind-estimates.parm](https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/default_params/copter-wind-estimates.parm)
4. restart SITL
5. arm throttle
6. GUIDED
7. takeoff 20
8. status WIND (to show the direction and speed, note the non-zero values)

Next load the gps-for-yaw parameters and repeat the flight
10. param load [../Tools/autotest/default_params/copter-gps-for-yaw.parm](https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/default_params/copter-gps-for-yaw.parm)
[...]

PR #24899, opened 2023-09-06
EKF: cope better with GPS jamming

tridge:
This is a revival of https://github.com/ArduPilot/ardupilot/pull/23010
re-opened so I can push to it to get it to pass CI

PR #27636, opened 2024-07-24
AP_NavEKF3: Fix yaw alignment bug preventing copter  wind estimation

priseborough:
This patch fixes the bug that was preventing copters using GPS yaw from using drag fusion to estimate wind. See https://github.com/ArduPilot/ardupilot/issues/26650 for details

Testing in SITL using the https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/default_params/copter-gps-for-yaw.parm parameters results in a wind speed being reported at the ground station with varying readings when EK3_DRAG_MCOEF is also set to a positive value.

Screenshot 2024-07-25 at 6 06 49 AM

The values reported are  large and varying because the default SITL physics model does not model X and Y axis specific forces, however this test does show that that after the patch, wind estimation is active with a GPS yaw source.

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_MagFusion.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_MagFusion.cpp
@@ -294,6 +294,7 @@ void NavEKF3_core::SelectMagFusion()
                 have_fused_gps_yaw = true;
                 lastSynthYawTime_ms = imuSampleTime_ms;
                 last_gps_yaw_fuse_ms = imuSampleTime_ms;
+                recordYawResetsCompleted();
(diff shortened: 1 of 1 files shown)


tridge (approved):

rmackay9:
Great to see this, thanks!

rmackay9:

rmackay9:
Sorry, I attempted to include this in 4.5.6-beta1 but it resulted in a compile error because this PR is built on top of https://github.com/ArduPilot/ardupilot/pull/24899

bobzwik:
Hello! Do you think this fix will be included in the next beta? Can't wait to test it out!

PR #27637, opened 2024-07-24
Sub: Support MAV_CMD_DO_REPOSITION on COMMAND_INT 

### LINKED ISSUE #14783: Sub: add support for DO_REPOSITION
We've recently [added support for MAV_CMD_DO_REPOSITION to Copter](https://github.com/ArduPilot/ardupilot/pull/14720) and Plane but we should also add it to Rover and Sub as well.

clydemcqueen:
This adds support for MAV_CMD_DO_REPOSITION on COMMAND_INT for Sub. The implementation is basically copied from Copter.

I added an autotest that tests this with 2 frames: MAV_FRAME_GLOBAL_RELATIVE_ALT and MAV_FRAME_GLOBAL_TERRAIN_ALT.

The main benefit is to align with the other vehicle types.

Fixes #14783

### DIFF
--- a/ArduSub/GCS_Mavlink.cpp
+++ b/ArduSub/GCS_Mavlink.cpp
@@ -483,6 +483,46 @@ MAV_RESULT GCS_MAVLINK_Sub::handle_command_do_set_roi(const Location &roi_loc)
     return MAV_RESULT_ACCEPTED;
 }
 
+MAV_RESULT GCS_MAVLINK_Sub::handle_command_int_do_reposition(const mavlink_command_int_t &packet)
+{
+    const bool change_modes = ((int32_t)packet.param2 & MAV_DO_REPOSITION_FLAGS_CHANGE_MODE) == MAV_DO_REPOSITION_FLAGS_CHANGE_MODE;
+    if (!sub.flightmode->in_guided_mode() && !change_modes) {
+        return MAV_RESULT_DENIED;
+    }
+
+    // sanity check location
+    if (!check_latlng(packet.x, packet.y)) {
+        return MAV_RESULT_DENIED;
+    }
+
+    Location request_location;
+    if (!location_from_command_t(packet, request_location)) {
+        return MAV_RESULT_DENIED;
+    }
+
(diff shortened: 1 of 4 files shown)


ArduSub/GCS_Mavlink.cpp
```diff
@@ -483,6 +483,46 @@ MAV_RESULT GCS_MAVLINK_Sub::handle_command_do_set_roi(const Location &roi_loc)
...
+        return MAV_RESULT_DENIED;
+    }
+
+    if (request_location.sanitize(sub.current_loc)) {
+        // if the location wasn't already sane don't load it
+        return MAV_RESULT_DENIED; // failed as the location is not valid
+    }
+
+    // we need to do this first, as we don't want to change the flight mode unless we can also set the target
+    if (!sub.mode_guided.guided_set_destination(request_location)) {
```
peterbarker, line 509: This duplicates a bug we still need to work out with this command in both Plane and Copter (and probably Rover)

If the `set_mode` call below fails then you will have changed your wp destination, so if you were in auto mode unfortunate things may happen...

Just noting it for awareness.

ArduSub/GCS_Mavlink.cpp
```diff
@@ -483,6 +483,46 @@ MAV_RESULT GCS_MAVLINK_Sub::handle_command_do_set_roi(const Location &roi_loc)
...
+        return MAV_RESULT_FAILED;
+    }
+
+    if (!sub.flightmode->in_guided_mode()) {
+        if (!sub.set_mode(Mode::Number::GUIDED, ModeReason::GCS_COMMAND)) {
+            return MAV_RESULT_FAILED;
+        }
+        // the position won't have been loaded if we had to change the flight mode, so load it again
```
peterbarker, line 517: This comment isn't actually correct; the position gets loaded just fine, but cleared when the mode is entered.

peterbarker (approved):
LGTM

peterbarker:
Merged, thanks!

PR #27640, opened 2024-07-25
AP_Camera: TYPE param description get Topotek, Viewpro, Xacti

rmackay9:
When users setup a combined camera+gimbal they normally need to set the CAMx_TYPE to 4 (Mount).  This updates the parameter description to name all the possible camera gimbal types it might refer to which now includes Topotek, Viewpro and Xacti.

I've created two commits so we can backport the 1st one (which doesn't include Topotek) into 4.5 if we choose

FYI @laozhou-fujian

### DIFF
--- a/libraries/AP_Camera/AP_Camera_Params.cpp
+++ b/libraries/AP_Camera/AP_Camera_Params.cpp
@@ -8,7 +8,7 @@ const AP_Param::GroupInfo AP_Camera_Params::var_info[] = {
     // @Param: _TYPE
     // @DisplayName: Camera shutter (trigger) type
     // @Description: how to trigger the camera to take a picture
-    // @Values: 0:None, 1:Servo, 2:Relay, 3:GoPro in Solo Gimbal, 4:Mount (Siyi), 5:MAVLink, 6:MAVLinkCamV2, 7:Scripting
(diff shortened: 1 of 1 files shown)


libraries/AP_Camera/AP_Camera_Params.cpp
```diff
@@ -8,7 +8,7 @@ const AP_Param::GroupInfo AP_Camera_Params::var_info[] = {
     // @Param: _TYPE
     // @DisplayName: Camera shutter (trigger) type
     // @Description: how to trigger the camera to take a picture
-    // @Values: 0:None, 1:Servo, 2:Relay, 3:GoPro in Solo Gimbal, 4:Mount (Siyi), 5:MAVLink, 6:MAVLinkCamV2, 7:Scripting
+    // @Values: 0:None, 1:Servo, 2:Relay, 3:GoPro in Solo Gimbal, 4:Mount (Siyi, Topotek, Viewpro, Xacti), 5:MAVLink, 6:MAVLinkCamV2, 7:Scripting
```
peterbarker, line 11:
```suggestion
    // @Values: 0:None, 1:Servo, 2:Relay, 3:GoPro in Solo Gimbal, 4:Mount (Siyi/Topotek/Viewpro/Xacti), 5:MAVLink, 6:MAVLinkCamV2, 7:Scripting
```

peterbarker (changes requested):
commas separate vvalues

libraries/AP_Camera/AP_Camera_Params.cpp
rmackay9: Thanks again @peterbarker!

PR #27641, opened 2024-07-25
AP_Bootloader: check ecc error flags only in Flash Status registers

bugobliterator:
* We were erasing flash for any of the status register values set, which can trigger mass erase simply when using external flash tools like STLink or Jlink, which can leave flags pending and cause bootloader to erase what was written.


tridge:
as long as this has been tested with the code that triggers a deliberate flash error then happy to merge

bugobliterator:
@tridge Tested Working.

PR #20000, opened 2022-02-05
Copter: If flight mode is disabled, it won't be built in

muramura:
ACRO mode and FLIP mode are optional flight modes.
I did not incorporate these two types into the code if they are disabled.

PR #27643, opened 2024-07-25
Copter: add a Mode method to disable crash check

peterbarker:
prevents looking for specific modes in the crash checker

Replaces https://github.com/ArduPilot/ardupilot/pull/20000/files

Tested the crash checker works on the bench in Loiter with a CubeOrange.

Can't easily test in SITL because of attitude reset on ground interaction

### DIFF
--- a/ArduCopter/crash_check.cpp
+++ b/ArduCopter/crash_check.cpp
@@ -42,7 +42,7 @@ void Copter::crash_check()
     }
 
     // return immediately if we are not in an angle stabilize flight mode or we are flipping
-    if (flightmode->mode_number() == Mode::Number::ACRO || flightmode->mode_number() == Mode::Number::FLIP) {
+    if (flightmode->disable_crash_check()) {
         crash_counter = 0;
         return;
     }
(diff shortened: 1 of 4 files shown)


rmackay9:
Hi @peterbarker,

Thanks for this.  Could we make it a positive question?  E.g. crash_check_enabled()?  I think avoiding the double negative makes it easier to understand

peterbarker:
> Hi @peterbarker,
> 
> Thanks for this. Could we make it a positive question? E.g. crash_check_enabled()? I think avoiding the double negative makes it easier to understand

Done - I went backwards and forwards on that a few times!

rmackay9:
I didn't realise that ENGINE_FAIL was a bitmask.  Maybe we should add a bitmask to the parameter description?

lthall:
Opps, saw my name on the review request list and added myself accidentally. PR looks sensible to me though. No behaviour change but provides flexibility for the future.

peterbarker:
> I didn't realise that ENGINE_FAIL was a bitmask. Maybe we should add a bitmask to the parameter description?

Done

lthall (approved):

rmackay9 (approved):

sdo91:
I just made a comment [here](https://github.com/ArduPilot/ardupilot/commit/93704c573c6b63615c68357b3f602344a5f464f5#commitcomment-157055443), but I don't think ENGINE_FAIL is a bitmask. It's just the index of which single engine will fail, correct?

PR #19230, opened 2021-11-11
RC_Channel: Express option numbers in bits

muramura:
I would change the number of arrays to the maximum value +1 to check for duplication of all channel options.

PR #27644, opened 2024-07-25
RC_Channel: Rework duplicate-rc-options sanity check

peterbarker:
Rework of https://github.com/ArduPilot/ardupilot/pull/19230/files

### DIFF
--- a/libraries/RC_Channel/RC_Channel.cpp
+++ b/libraries/RC_Channel/RC_Channel.cpp
@@ -1859,7 +1859,7 @@ RC_Channel *RC_Channels::find_channel_for_option(const RC_Channel::AUX_FUNC opti
 // duplicate_options_exist - returns true if any options are duplicated
 bool RC_Channels::duplicate_options_exist()
 {
-    uint8_t auxsw_option_counts[512] = {};
+    Bitmask<(uint16_t)RC_Channel::AUX_FUNC::AUX_FUNCTION_MAX> used_auxsw_options;
     for (uint8_t i=0; i<NUM_RC_CHANNELS; i++) {
         const RC_Channel *c = channel(i);
         if (c == nullptr) {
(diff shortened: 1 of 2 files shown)


libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -1859,27 +1859,24 @@ RC_Channel *RC_Channels::find_channel_for_option(const RC_Channel::AUX_FUNC opti
 // duplicate_options_exist - returns true if any options are duplicated
 bool RC_Channels::duplicate_options_exist()
 {
-    uint8_t auxsw_option_counts[512] = {};
+    Bitmask<512> used_auxsw_options;
```
tpwrules, line 1862:
```suggestion
    Bitmask<(uint16_t)RC_Channel::AUX_FUNC::AUX_FUNCTION_MAX> used_auxsw_options;
```

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -1859,27 +1859,24 @@ RC_Channel *RC_Channels::find_channel_for_option(const RC_Channel::AUX_FUNC opti
...
     for (uint8_t i=0; i<NUM_RC_CHANNELS; i++) {
         const RC_Channel *c = channel(i);
         if (c == nullptr) {
             // odd?
             continue;
         }
         const uint16_t option = c->option.get();
-        if (option >= sizeof(auxsw_option_counts)) {
+        if (option == 0) { // MAGIC VALUE! This is AUXSW_DO_NOTHING
```
tpwrules, line 1870:
```suggestion
        if (option == (uint16_t)RC_Channel::AUX_FUNC::DO_NOTHING) {
```

tpwrules (commented):
Looks nice, but better to use the enum instead of keeping magic values.

Saves eight bytes of flash on Cube Orange as a bonus!

libraries/RC_Channel/RC_Channel.cpp
peterbarker: Huh.  We have an _MAX entry now.  Interesting, I'd forgotten that....

Done.  I'm not *really* sure it's a win.  Yeah, the code won't get out of date with the number of options, but that won't happen for a while.  And obviously the stack space isn't an issue :-)

With this formulation - pity the poor bugger in his fork who decides to use option number 40,000 for his extra-special-super-secret AUXFUNC... at least he's likely to get a frame-size compilation failure rather than a runtime crash, but still an interesting issue.

libraries/RC_Channel/RC_Channel.cpp
peterbarker: Done.  No idea why I used a literal zero when I first wrote this.

libraries/RC_Channel/RC_Channel.cpp
tpwrules: That is fair, I had considered suggesting a static_assert that the bit mask didn’t end up over 512 bytes as we had decent evidence that much stack space was safe, but forgot about it. We’ll see what the dev call thinks.

tridge (approved):

PR #27651, opened 2024-07-25
Enable logging on iFlight 2RAW H743 and rename board

andyp1per:
iFlight requested the name change.

The flash chip is actually accessible via the microSD interface


peterbarker:
@andyp1per I suggest creating a PR against the 4.5 branch, rather than just labelling this one.  You will need to adjust the LED defines at the very least for the 4.5 branch

Hwurzburg:
this impacts the wiki page already pending and needs readme changes....hard to catch these without review or wiki label! a little help would be appreciated!

rmackay9:

rmackay9:
I've removed the 4.5 backport label because this PR is included in https://github.com/ArduPilot/ardupilot/pull/27662

PR #27652, opened 2024-07-25
AP_Scripting: fix failures with require()

tpwrules:
The `require()` function needs the environment of the currently executing script to get the table of loaded modules. It currently does this by getting the first upvalue of the function currently used to run the script and hoping that is the environment table it needs. However, that function (and hence the upvalues) can change every time the script reschedules, so there is no guarantee that the environment table will be where `require()` needs it, and indeed it is possible to accidentally create a script where `require()` fails with a mysterious Lua error.

Correct the issue by splitting the script reference originally used by `require()` into a reference to the function used to run the script, and a separate reference specifically for the environment table, which `require()` now uses. Also add a (n admittedly fragile) test that the issue is resolved. Tested in SITL and on Cube Orange with my REPL that `require()` works properly and the issue is fixed.

Bonus: saves 184 bytes of flash on Cube Orange because `lua_getupvalue` is no longer used anywhere

### DIFF
--- a/Tools/autotest/rover.py
+++ b/Tools/autotest/rover.py
@@ -5308,6 +5308,7 @@ Brakes have negligible effect (with=%0.2fm without=%0.2fm delta=%0.2fm)
 
         self.install_test_scripts_context([
             "scripting_test.lua",
+            "scripting_require_test_2.lua",
             "math.lua",
             "strings.lua",
             "mavlink_test.lua",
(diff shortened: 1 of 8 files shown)


bugobliterator (approved):

IamPete1 (approved):
This seems reasonable, I'm not really up to speed with lua enough to know if this is correct.

We do already have a few auto tests that use require, and this adds another. So I think we should merge and see how it goes in the wild.

PR #27654, opened 2024-07-26
AP_Mission: add and use an option_is_set method

peterbarker:


timtuxworth:
This looks like a very sensible idea. I'd like to use it in my pr #27639

IamPete1 (approved):
LGTM

PR #17419, opened 2021-05-12
Add test for SmartRTL/ALT_HOLD switching

peterbarker:
This test is expected to fail.

It's based on another PR which adds a simple, working, test for SmartRTL.

PR #27656, opened 2024-07-26
autotest: add SmartRTL test for rapid switch between smartrtl and alt…

peterbarker:
…hold

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -9868,6 +9868,23 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
         if ex is not None:
             raise ex
 
+    def SMART_RTL_EnterLeave(self):
+        '''check SmartRTL behaviour when entering/leaving'''
+        # we had a bug where we would consume points when re-entering smartrtl
+
+        self.upload_simple_relhome_mission([
+            #                                      N   E  U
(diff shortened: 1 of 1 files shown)


Tools/autotest/arducopter.py
```diff
@@ -9868,6 +9868,23 @@ def GPSForYaw(self):
         if ex is not None:
             raise ex
 
+    def SMART_RTL_EnterLeave(self):
+        '''check SmartRTL behaviour when entering/leaving'''
+        # we had a bug where we would consume points when re-entering smartrtl
```
rmackay9, line 9873: Thanks for this.  Not that it matters too much because it's all in the past but the bug was that points would be lost if SmartRTL mode was exited before reaching home

Would it really "Cause a panic"?  A panicking pilot perhaps but not a code execution panic I don't think.

peterbarker: This bug isn't in the past.  This test is being merged *disabled*.  It still causes SITL to panic.

I've pulled it out of https://github.com/ArduPilot/ardupilot/pull/17419 so the test doesn't bitrot (which it had).  I'll be rebasing the fix PR on top of master soon (it's a non-trivial rebase as it conflicts badly with the recent consumption-of-points fix we merged.

rmackay9: Ah right, sorry.  I assumed this was related to my recent fix but no, this is for that long standing bug.  OK thanks!

PR #27658, opened 2024-07-26
Update FlyingMoon hwdefs to remove MAX7456

peterbarker:
No longer in use on these boards


peterbarker:
These boards aren't generally for sale, and these are the vendor's files, so I think they can go straight in

PR #27659, opened 2024-07-26
hwdef: remove MatekL431-Periph

peterbarker:
this was a kitchen-sink AP_Periph build.  We're moving away from these.

This board does not currently build.  We could make it build, but it will die again and again.


tpwrules:
I will miss it dearly (and had gotten it to build) 😭

Why is a speedybee change mixed in?

IamPete1:
That is for this board:
https://www.mateksys.com/?portfolio=can-l431#tab-id-1

This build is recommended on there page?

I have used these a lot. Typically as a base with extra stuff added on, relay temp sensors ect. 

I think we should fix it rather than removing. Maybe remove GPS, I think remote GPS via a external node is not common.

andyp1per:
> Maybe remove GPS, I think remote GPS via a external node is not common.

Yes, agreed

tpwrules:
I looked and it does seem GPS changes are what pushed it over the edge. Can we just remove some of the rarer GPS types? It does have a GPS plug and a GPS is a suggested application. But maybe we could push users to the GPS-specific build for that.

tpwrules:
https://github.com/ArduPilot/ardupilot/pull/27685 will bring this board back from the dead without any hwdef changes.

joshanne:
+1 for keeping this build around. As Pete mentioned, that board is a kitchen sink board. I tend to rely on the base hwdef to customise this board.

Ryanf55:
You can drop GSOF from periph on this board as long as we still check GSOF on Periph.

peterbarker:
OK, obviously a lot of love for the MatekL431-Periph build.  Lesson learnt :-)

PR #27660, opened 2024-07-26
hwdef: remove features from SpeedyBeeF405WING

peterbarker:
Doesn't compile, out of flash space, this seems an obvious candidate


Hwurzburg (changes requested):
that would leave users with no way to aututune...these are being used in small vtols...I would rather see the mount stuff lost...its less common..and why arent a lot of 405s in trouble? (I havent looked recently)

peterbarker:
> that would leave users with no way to aututune...these are being used in small vtols...I would rather see the mount stuff lost...its less common..and why arent a lot of 405s in trouble? (I havent looked recently)

It really does come down to the extra bits at the bottom.  14kB for QAutoTune

Hwurzburg (approved):

PR #27661, opened 2024-07-26
Replace HAL_NO_MONITOR_THREAD with HAL_MONITOR_THREAD_ENABLED

peterbarker:
I've run this through size-compare-branches.

Everything that compiles is no-compiler-output change.

I've created PRs to fix compilation on the boards that didn't compile.


peterbarker:
... I think it goes without saying that we should probably work out why the boards that include the monitor thread do so...

tridge (approved):

PR #27669, opened 2024-07-27
Update GCS_Mavlink.cpp

haydenbanz:


khancyr:
hello,

You seems to have miss to commit something and we would need some explanation on the change you are doing. 
I will be closing this, please open another PR with your change. Thx

PR #27670, opened 2024-07-27
Filter: LowPassFilter: split into two classes for constant and varable dt

IamPete1:
This splits our existing `LowPassFilter` class into a class which uses variable time step and passes the dt in the apply call and a class which uses constant dt. The variable dt class retains the `LowPassFilter` name, the constant dt class is `LowPassFilter_const_dt`.

The key thing is to only provide methods that will work together. 

This found two bugs, copters flow hold `set_cutoff_frequency` call would not work while in the mode, only on init, so a cuttoff frequency param change would not take effect in teal time. 

Secondly the SITL battery voltage filter is currently none functional because it was using the constant dt methods without setting the sample rate. 

Where needed the class has been changed to the constant dt version.


libraries/Filter/LowPassFilter.cpp
```diff
@@ -12,134 +10,134 @@
...
+    return cutoff_freq;
 }
 
-// return the cutoff frequency
+// add a new raw value to the filter, retrieve the filtered result
 template <class T>
-float LowPassFilter<T>::get_cutoff_freq(void) const {
-    return _cutoff_freq;
+T LowPassFilter_const_dt<T>::apply(const T &sample) {
+    return this->_apply(sample, alpha);
```
IamPete1, line 88: using `this->` is a little odd, its due to class template inheritance. see: https://kernhanda.github.io/templates-inheritance-methods/

clydemcqueen:
test.Sub.RngfndQuality passes locally for me on this PR. Might be flaky. I'll take a look.

IamPete1:
> test.Sub.RngfndQuality passes locally for me on this PR. Might be flaky. I'll take a look.

I have restarted it for now, Thanks for taking a look!

libraries/Filter/LowPassFilter.h
```diff
@@ -15,22 +15,20 @@
...
+   LowPassFilter: providing dt on every sample, and calling apply like this:
 
       // call once
       filter.set_cutoff_frequency(frequency_hz);
 
       // then on each sample
       output = filter.apply(sample, dt);
 
-   2) providing a sample freq and cutoff_freq once at start
+   LowPassFilter_const_dt: providing a sample freq and cutoff_freq once at start
```
andyp1per, line 31: I'm really not enamoured with the name!

IamPete1: I'm open to better suggestions, I just wanted to make it clear which should be used in each case.

andyp1per: FixedLowPassFilter? StaticLowPassFilter?

IamPete1: Its not fixed in cutoff frequency, its fixed in sample rate. So i'm not sure fixed really works. Or at least it does not line up with what we would call a fixed notch.

libraries/Filter/LowPassFilter.cpp
```diff
@@ -12,134 +10,134 @@
...
+template class DigitalLPF<Vector3f>;
+
+////////////////////////////////////////////////////////////////////////////////////////////
+// Low pass filter with constant time step
+////////////////////////////////////////////////////////////////////////////////////////////
 
+// constructor
 template <class T>
-LowPassFilter<T>::LowPassFilter(float sample_freq, float cutoff_freq)
+LowPassFilter_const_dt<T>::LowPassFilter_const_dt(const float &sample_freq, const float &new_cutoff_freq)
```
rmackay9, line 62: I think we should consider naming it LowPassFilterConstDt

libraries/Filter/LowPassFilter.h
rmackay9: LowPassFilterConstDt would be my vote

tridge:
structure good, just needs testing and/or unit tests

IamPete1:
I have updated the transfer function test to run both the new classes, these are with 1000Hz sample rate and 50Hz cuttoff: 
`LowPassFilterFloat`:
image

`LowPassFilterConstDtFloat`:
image

Both still exactly match the analytical solution. So there is no change in behavior.

tridge:
looks good! marked for flight testing on the weekend

tridge:
tested on a 5" quad and a plane on the weekend with no noticeable issues

tridge (approved):

PR #27676, opened 2024-07-27
Ardutracker verbosity updates

landswellsong:
I'm doing some tracker research and I've realized that `AUTO` can do at least 3 different things at once without the operator being aware of what the device is actually trying to accomplish. It sure isn't a problem for a well-set tracker, but during development being aware of that information saves a lot of time from dumb mistakes (e.g. "are my servos dead? hold on I forgot I'm withing minimum range"). I've also reduced the message spam in `GUIDED` so that only the updates are posted and fixed a seemingly small oversight when `vehicle.location_valid` would be automatically set to `true` at startup even without the target being present in the network.


AntennaTracker/mode.h
```diff
@@ -36,13 +36,29 @@ class Mode {
...
+        IDLE=0,             // Not doing anything in particular
+        TRACKING=1,         // Actively tracking a valid target
+        SUPPRESSED=2,       // Target within minimum tracking range
+        SCANNING=3,         // Target lost, scanning for target
+    };
+
+    void switch_state(State st) { _state = st; }
+
+private:
+    State _state{State::IDLE};
```
IamPete1, line 60:
```suggestion
    State _state;
```
You don't need to init to 0 in AP.

AntennaTracker/system.cpp
```diff
@@ -191,6 +192,11 @@ void Tracker::set_mode(Mode &newmode, const ModeReason reason)
     gcs().send_message(MSG_HEARTBEAT);
 
     nav_status.bearing = ahrs.yaw_sensor * 0.01f;
+
+    // reset the auto mode state
+    if (&mode_auto == &newmode) {
+        mode_auto.switch_state(ModeAuto::State::IDLE);
+    }
```
IamPete1, line 199: We don't have it in tracker. On other vehicles we typically have a mode enter function where you can do mode specific initialisations.

AntennaTracker/mode_guided.cpp
```diff
@@ -3,15 +3,23 @@
...
     float target_roll, target_pitch, target_yaw;
     _target_att.to_euler(target_roll, target_pitch, target_yaw);
-    if (now - last_debug > 5000) {
+    bool yaw_changed = isnan(last_target_yaw) || fabsf(last_target_yaw - target_yaw) > eps;
+    bool pitch_changed = isnan(last_target_pitch) || fabsf(last_target_pitch - target_pitch) > eps;
+    if (now - last_debug > DEBUG_TIMEOUT && (yaw_changed || pitch_changed)) {
         last_debug = now;
         gcs().send_text(MAV_SEVERITY_INFO, "target_yaw=%f target_pitch=%f", degrees(target_yaw), degrees(target_pitch));
+        last_target_yaw = target_yaw;
+        last_target_pitch = target_pitch;
```
IamPete1, line 22: Not really a fan off this. We try not to use statics and nans.

How about adding a "last_set_angle_ms" timestamp for the "set_angle" call which updates the `_target_att`. Then you could check if the timestamp is larger than the last debug print.

AntennaTracker/mode_auto.cpp
```diff
@@ -4,9 +4,48 @@
...
+            default:
+                break;
+        }
     }
 }
+
+bool ModeAuto::scan_for_any_target() const
+{
+    return (tracker.g.auto_opts.get() & (1 << 0)) != 0;
+}
```
IamPete1, line 51: If your feeling keen we typically do this sort of thing with a `option_is_set` helper and a enum for the options. There are lots of examples.

AntennaTracker/mode_auto.cpp
```diff
@@ -4,9 +4,48 @@
...
+                gcs().send_text(MAV_SEVERITY_INFO, "Tracker suppressed while the target is within %d m", (short)tracker.g.distance_min);
+                break;
+
+            case State::IDLE:
+                gcs().send_text(MAV_SEVERITY_INFO, "Tracker idle");
+                break;
+
+            default:
+                break;
+        }
```
IamPete1, line 44: This could be in the new `switch_state` helper.

AntennaTracker/mode_auto.cpp
```diff
@@ -4,9 +4,48 @@
...
+        switch (_state) {
+            case State::TRACKING:
+                gcs().send_text(MAV_SEVERITY_INFO, "Tracker locked on the vehicle %d", (short)tracker.g.sysid_target);
+                break;
+
+            case State::SCANNING:
+                if (scan_for_any_target()) {
+                    gcs().send_text(MAV_SEVERITY_INFO, "Tracker scanning for a vehicle");
+                } else {
+                    gcs().send_text(MAV_SEVERITY_INFO, "Tracker scanning for the vehicle %d", (short)tracker.g.sysid_target);
```
IamPete1, line 30:
```suggestion
                    gcs().send_text(MAV_SEVERITY_INFO, "Tracker scanning for the vehicle %i", tracker.g.sysid_target);
```

AntennaTracker/mode_auto.cpp
```diff
@@ -4,9 +4,48 @@
...
+                }
+                break;
+
+            case State::SUPPRESSED:
+                gcs().send_text(MAV_SEVERITY_INFO, "Tracker suppressed while the target is within %d m", (short)tracker.g.distance_min);
```
IamPete1, line 35:
```suggestion
                gcs().send_text(MAV_SEVERITY_INFO, "Tracker suppressed while the target is within %i m", tracker.g.distance_min);
```

IamPete1 (changes requested):
Some nice clean ups, thanks.

AntennaTracker/system.cpp
landswellsong: @IamPete1 what's the best way to ensure that the state is reset upon entering the mode then? I was looking for some enter method to be honest and this part felt bad to write.

My rationale for the whole thing is so that if you do e.g. `AUTO` ⇒ `MANUAL` ⇒ `AUTO` transition, you get a state change announcement. But maybe I'm overthinking and this whole thing can be deleted, after all 99% of the time you want to sit in `AUTO` and never change out of it.

IamPete1: The `mode_auto` enter function would call `switch_state(ModeAuto::State::IDLE)`

AntennaTracker/mode_auto.cpp
landswellsong: I would have expected this to work, but:
```
../../AntennaTracker/mode_auto.cpp: In member function ‘virtual void ModeAuto::update()’:
../../AntennaTracker/mode_auto.cpp:23:84: error: format ‘%i’ expects argument of type ‘int’, but argument 4 has type ‘AP_Int16’ {aka ‘AP_ParamT<short int, AP_PARAM_INT16>’} [-Werror=format=]
   23 |                 gcs().send_text(MAV_SEVERITY_INFO, "Tracker locked on the vehicle %i", tracker.g.sysid_target);
      |                                                                                   ~^   ~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                    |             |
      |                                                                                    int           AP_Int16 {aka AP_ParamT<short int, AP_PARAM_INT16>}
```

Compiler: `gcc (Gentoo 13.2.1_p20230826 p7) 13.2.1 20230826`
Clang (LSP client) warns about the same. I tried `%hi` as well, but for some reason it's also treated as `int`?

IamPete1: Ah, you could try adding the `get()`. That makes the AP_Param type return the underlying type.
```suggestion
                gcs().send_text(MAV_SEVERITY_INFO, "Tracker suppressed while the target is within %i m", tracker.g.distance_min.get());
```

landswellsong:
Resolved the bits that were addressed, removed the mode switch part in `system.cpp` altogether, but kept the conversation in case there's a better way.

landswellsong:
@IamPete1 I've transplanted the mode enter/exit from Plane, the exits are currently completely futureproofing as well as the boolean return of the enter, but I think it can be useful at some point and doesn't really cost too much to have.

landswellsong:
Also added a small commit introducing a new bit to `AUTO_OPTIONS` that allows disabling the automatic scan functionality altogether (off by default of course).

landswellsong:
Should I rebase it over the recent master?

IamPete1:
I think the changes look good. We are failing the tracker CI test tho. So there must be some change in behavior, we need to understand what that is and either fix it or decide its better and update the test. You can try it locally with:
```
./Tools/autotest/autotest.py build.Tracker test.Tracker 
```

landswellsong:
@IamPete1 thanks I'll do that and see what can be done, plus rebase

landswellsong:
@IamPete1 I did a few attempts at the test, but still figuring things out. I renamed the PR to better capture the real meaning and rebased over 4.7. 

I've been using this code for a while now and it seems to work alright (on 4.5.7 at least). That said I ran into similar issues with minimal distance and CR servos as the lack of mode switch clarity inside `AUTO` caused the servo to basically keep repeating the last rotation speed when the target is within minimal radius (which is very obviously a bug) and the "stop updating servos" logic wouldn't hold here because it's based on positional servo logic. I was meant to do that as a PR itself, but I honestly don't like how much of a mess `AUTO` code is and will be after even more changes.

I have another suggestion while I'm still on the test. What if we make a special mode for it than can only be entered from `AUTO` itself? E.g. call it `IDLE` and use it for the cases when the tracker is both active (i.e. not `STOP` where the user expects servos to never move), but not tracking at the moment. This would cover both cases:
- Waiting for the radio (subject of this PR)
- Within minimal tracking distance

This way we get a benefit of just using mode enter/exit instead of a very clunky layered sub-state machine as well as the user gets a clear indication that the servos aren't moving because they are suppressed (on the field it was really a headscratcher quite a few times). Currently the user would need to rely on GCS texts to see these transitions (including ones I've added here), but with another mode the texts would rather be an extra touch. This also helps with automation a lot. Simple state diagram of what I mean:

Diagram1

(If it's a bad idea I'll continue on with the test and maybe add the CR servo fix too)

landswellsong:
The changes in [38103a1](https://github.com/ArduPilot/ardupilot/pull/27676/commits/38103a1ccf0625672a14e9404302d183a1d59e0d) should actually apply to an issue in master branch as well, although the commit isn't cherry-pickable as is. When we set `DISTANCE_MIN` and the target vehicle enters the suppression radius, the servos stop updating, but this means that CR servos (and I assume OnOff too) instead of being frozen continue to play the last PWM commanded, i.e. move uncontrollably. Seen this happen to a Pelco-D setup in the wild. The fix forces servos into a state similar to being disarmed without actually marking the vehicle as disarmed (which makes sense from the operator POV as the servos might just start moving again).

The only thing I'm not super sure about is the `SAFE_DISARM_PWM` parameter. I assume this is meant to give the positional servos a predictable disarmed orientation, but the way I implemented it that would also concern the suppressed states (target too close, scan disabled). I'm not sure if I should handle it differently because I honestly can't grasp the usecase for this parameter. In my understanding if the user wants to stow the disarmed antenna then they will also want it stowed when it's suppressed, but I may be wrong from UX perspective.

If needed I can make a separate PR to address this issue without the rest of this branch code. It would go into `update_auto` instead.

PR #27677, opened 2024-07-27
CI: only run qurt build on ArduPilot main repo

tridge:
this prevents user repos getting failures

thanks to Pete for the suggestion


IamPete1:
Looks like the test is correctly skipped on user repos:
image

And it has already passed here:
image

IamPete1 (approved):
LGTM

PR #27678, opened 2024-07-27
Rover: improve circle / loiter turns drive to edge phase

IamPete1:
Currently rover can get stuck driving to the circle edge, this is particularity true when using a loiter turns waypoint where the rover may start outside the circle. 

This adds a loiter turns waypoint to the drive mission autotest, it fails on master and passes on this PR.

**Master**

image

The auto test mission:

image

**PR**

image

The new auto test mission
image


Rover/mode_circle.cpp
```diff
@@ -144,19 +144,50 @@ void ModeCircle::update()
...
+{
+    // check if vehicle has reached edge of circle
+    const float dist_thresh_m = MAX(g2.turn_radius, AR_CIRCLE_REACHED_EDGE_DIST);
+    reached_edge |= dist_to_edge_m <= dist_thresh_m;
+
+    // calculate target point's position, velocity and acceleration
+    target.pos = config.center_pos.topostype();
+    target.pos.offset_bearing(degrees(target.yaw_rad), config.radius);
+
+    g2.pos_control.input_pos_target(target.pos, rover.G_Dt);
```
IamPete1, line 181: The key change is to use just a pos target, the target is the same but we don't add velocity/accel.

IamPete1:
Circle mode where we start inside the circle does still work:
image

Rover/mode_circle.cpp
```diff
@@ -144,19 +144,50 @@ void ModeCircle::update()
...
+
+    } else {
+        update_circling();
     }
 
+    g2.pos_control.update(rover.G_Dt);
+
+    // get desired speed and turn rate from pos_control
+    const float desired_speed = g2.pos_control.get_desired_speed();
+    const float desired_turn_rate = g2.pos_control.get_desired_turn_rate_rads();
```
MattKear, line 164: nitpick: I know you have only copy-pasted these from else where but we could just pass direct to the function argument

MattKear (approved):
Nice.  LGTM.  Just a nitpick comment, but it is personal preference.

Rover/mode_circle.cpp
IamPete1: It should result in the same compiler output. Keeping them separate means they get a nice comment.

tridge:
@rmackay9 needs your review, thanks

rmackay9 (approved):

rmackay9:
Thanks!

PR #27679, opened 2024-07-27
hwdef: remove EKF opinion from FlywooF745

tpwrules:
No effect as lines have always been commented out, but not useful and EKF2 is not something to encourage.

Verified no compiler output change.


IamPete1 (approved):

peterbarker:
Merged, thanks!

Issue #27648, opened 2024-07-25
RELAY set to MAIN (IOMCU) pin doesn't turn off on versions 4.5.0 and later


LEGIONBOSS:
## Bug report

**Issue details**
When I set up a MAIN pin as a RELAY on an Orange Cube running ArduCopter 4.5.4, it turns ON when instructed, then gets stuck ON when I try to turn it OFF. This issue is not present on AUX (FMU) pins.

Steps to re-create the issue

On Copter 4.4.4 and older:
1. Set SERVO1_FUNCTION to -1 (GPIO)
2. Set RELAY_PIN to 101 (MainOut1)
3. Reboot the board

On Copter 4.5.0 and later:
1. Set SERVO1_FUNCTION to -1 (GPIO)
2. Set RELAY1_FUNCTION to 1 (Relay)
4. Refresh the parameter list to reveal the one below
5. Set RELAY1_PIN to 101 (MainOut1)
6. Reboot the board

On both, after setting the parameters:
1. Monitor the DC voltage between MAIN OUT 1 Signal (S) and GND (-) lines
2. Toggle the relay (Relay0 in Mission Planner, if RELAY1 parameters were set) using the High and Low buttons on the Servo/Relay tab on the DATA screen in Mission Planner (the Toggle button is not a toggle really, but a Pulse)

- On Copter 4.4.4 and older, the Relay works as intended, toggles between ~0 and ~3.3 volts
- On Copter 4.5.0 and later, the Relay only turns on, but stays on when instructed to turn off

**Version**
4.5.4

**Platform**
[  ] All
[  ] AntennaTracker
[X] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
Nothing, just 

**Hardware type**
Cube Orange

**Logs**

andyp1per:
I tried this on a cube orange plus, does not work for me on either 4.4 or 4.5 @IamPete1

IamPete1:
> I tried this on a cube orange plus, does not work for me on either 4.4 or 4.5 @IamPete1

I have tested here and it works on 4.4.4 it works on CubeOrange but not a CubeOrangePlus or Pixhawk1. I'm not sure why that would be, as far as I know the IOMCU config should be the same. In anycase, it seems a little pointless to get into that too much on 4.4 since 4.5 will be different. Do you have a CubeOrange (not plus to test?)

IamPete1:
[...]

PR #27683, opened 2024-07-28
IOMCU: add virtual GPIO read so relays work again

IamPete1:
https://github.com/ArduPilot/ardupilot/pull/25108 Added logging of relay where it only called set and wrote the log if there would be a state change. However, because pin read was never implemented on IOMCU read would always return 0 and the pin never turn off. 

This fixes that by adding a "virtual read" method to get the commanded state of the IOMCU pin. Much like the DroneCAN pins this is NOT a real read, just a return of the last commanded state. However, this is enough to get relay to work again. 

This also adds a getter for the IOMCU GPIO mask so its state is correctly represented in the `RCOut` send text. EG With `SERVO1_FUNCTION` -1: 
```
 RCOut: PWM:2-14
```

This fixes https://github.com/ArduPilot/ardupilot/issues/27648

This has been tested on CubeOrange, however in my testing and in that of @andyp1per this functionality does not work on CubeOrangePlus. It seems to be a long standing issue, it is present in ever since the OrangePlus was added. I'm not sure why this would be. We do see the pin respond momentarily and then go low again.

### DIFF
--- a/libraries/AP_HAL_ChibiOS/GPIO.cpp
+++ b/libraries/AP_HAL_ChibiOS/GPIO.cpp
@@ -241,6 +241,11 @@ uint8_t GPIO::read(uint8_t pin)
     if (g) {
         return palReadLine(g->pal_line);
     }
+#if HAL_WITH_IO_MCU
+    if (AP_BoardConfig::io_enabled() && iomcu.valid_GPIO_pin(pin)) {
+        return iomcu.read_virtual_GPIO(pin);
+    }
+#endif
     return 0;
 }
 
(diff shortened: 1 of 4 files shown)


andyp1per (approved):
LGTM - thanks!

rmackay9:

rmackay9:
I've removed the 4.5 backports project from this PR because the above linked PR has been merged or more accurately the commits from the above linked PR have been included in this release PR https://github.com/ArduPilot/ardupilot/pull/27868

PR #27684, opened 2024-07-28
autotest: large outliers in sub terrain test have low sq

clydemcqueen:
The current test.Sub.SimTerrainSurftrak noise configuration is causing ~more~ test failures ~than we'd like~.

This PR adds `outlier_good_sq_limit` to the configuration. If the error exceeds `outlier_good_sq_limit` then the signal quality drops from 100 (best) to 50 (bad). `Sub::read_rangefinder()` tests for this, so large outliers are rejected. This should ~minimize~ eliminate test failures, and it also exercises the sq test in `Sub::read_rangefinder()`.

Edit: I also fixed some test names and a few typos

### DIFF
--- a/Tools/autotest/ardusub.py
+++ b/Tools/autotest/ardusub.py
@@ -287,7 +287,7 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
         self.disarm_vehicle()
         self.context_pop()
 
-    def prepare_synthetic_seafloor_test(self, sea_floor_depth):
+    def prepare_synthetic_seafloor_test(self, sea_floor_depth, rf_target):
         self.set_parameters({
             "SCR_ENABLE": 1,
             "RNGFND1_TYPE": 36,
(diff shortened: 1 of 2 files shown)


clydemcqueen:
@ptrmu Please take a look

peterbarker:
Just a note.  No test failures are acceptable - the test must be 100% (as-designed ;-) ), even in the face of adding random noise.

People seem to have this thing against flapping tests :-)

clydemcqueen:
@peterbarker got it!

I pushed another commit (I'll squash later), this should eliminate all test failures:
* all outliers > 1m have low signal quality, so they will be dropped
* the rf target is explicitly set

ptrmu:
Looks great!

peterbarker (approved):
Anything which reduces false positives is good in my book :-)

Williangalvani (approved):
Looks good!

peterbarker:
Merged, thanks!

PR #27686, opened 2024-07-28
Add MPU6000 to hwdef for ReaperF745 to Support V4 FC

ericjohnson97:
ReaperF745 V4 is now coming with the MPU6000. 
https://www.foxeer.com/foxeer-reaper-aio-v4-f745-mpu6000-35a-bluejay-bls-2-4s-g-542


Hwurzburg (approved):

libraries/AP_HAL_ChibiOS/hwdef/ReaperF745/hwdef.dat
```diff
@@ -121,6 +121,7 @@ define HAL_LOGGING_DATAFLASH_ENABLED 1
 # spi devices bmi270(v2) icm42688 (v3)
 SPIDEV bmi270    SPI3 DEVID1 GYRO_CS  MODE3  1*MHZ  4*MHZ
 SPIDEV icm42688  SPI3 DEVID1 GYRO_CS  MODE3  2*MHZ  8*MHZ
+SPIDEV mpu6000   SPI3 DEVID1 GYRO_CS  MODE3  1*MHZ  4*MHZ
```
andyp1per, line 124: 4Mhz is pretty slow. I would stick with 8Mhz if that works ok

andyp1per (changes requested):

libraries/AP_HAL_ChibiOS/hwdef/ReaperF745/hwdef.dat
ericjohnson97: sorry to go dark. I copied that line from another board. I'll try the IMU at 8k. I haven't gotten my build air born yet, but I will report back after I am able to flight test.

peterbarker: ping @ericjohnson97 how'd the flight test go?

ericjohnson97: I've run into a number of issues that I haven't worked through yet. First, I took my new drone out to test at the beginning of august in Texas and after a bit of time in the sun and on the hot concrete the 3d printed frame started melting. 
😂

I also was unable to get a GPS lock even though the board reported detecting a ublox gps.  I reprinted my drone in ASA and went to a GPS module i've had success with in the past.  I took it on the roof of my apartment to see if I could get a GPS lock this weekend and unfortunately it still was not getting a GPS lock even though once again it detected a ublox GPS. 

This time I did notice that after powering on the drone the estimated velocity was growing bigger and bigger over time. I re-did all of the calibration, but no luck. I was going to reflash it and start the setup from scratch again, but I've been a little busy. 

Let me know if you have any advice for those issues. I'll report beck once I hopefully get my aircraft in the air. 

here are some pics of my setup for context





I'm using OpenHD for video and telemetry.

andyp1per:
Without GPS and in heat you will likely get an accel bias which will cause your velocity to increase, we make some attempt to compensate for this but a GPS is your best answer. No idea why you would not get GPS lock.

peterbarker:
For your GPS, you can:
 - for testing purposes unclip it from its holder and stretch it out as far from the other electronics as you can - does that get lock?  If so, you've got interference problems.
 - check the voltage to the GPS unit.  They *really* do not like to be undervolted!
 - add a chunky copper groundplane underneath the GPS (unetched circuit board works well for this)
 - check the accel biases in the dataflash logs (https://ardupilot.org/copter/docs/logmessages.html#xkf2)
 - see if the velocity problems occur when the vehicle is kept at a constant temperature
 - consider doing an IMU temperature calibration (https://ardupilot.org/copter/docs/common-imutempcal.html)

@andyp1per is there any way that we could become happy with this PR without flight?  Any amount of ground manipulation we could use instead?

ericjohnson97:
@peterbarker those all good pieces of advice. I'll try those steps

andyp1per:
I don't think these issues can be anything to do with the IMU, I think we should just merge this @peterbarker

peterbarker:
I've fixed the commit message and marked for MergeOnCIPass

Thanks!

PR #27578, opened 2024-07-17
Quicktune C++ conversion (WIP)

MichelleRos:
Add Quicktune to C++ code (conversion from the VTOL-quicktune lua script).

Tested in SITL, multiple multcopters and quadplanes

Edit: Tridge has added support for quadplanes and an autotest for it and made it not run for helis. I plan to add an autotest for the copter version shortly.
I have also updated the PR to limit quicktune to only alt_hold and loiter in copter, and Tridge did something similar for quadplanes.

Edit 2: Tridge updated it to make it always allocate, to be more in line with the pattern in copter, and moved it to its own thread. He also added a max attitude error, so it auto-aborts if the vehicle starts oscillating badly during the tune and made it abort the tune if the pilot changes to a mode that doesn't support quicktune. Note that it does not abort if the pilot changes between modes that do support quicktune, allowing people to, for example, do the tuning in loiter mode, then switch to alt_hold to test the tune before saving the gains.

Now tested on a quadplane as well.

As for the flash concerns, yes, it uses around 5k of flash, but autotune uses around 15k of flash, so for a tuner, quicktune is actually rather light.

PR #27688, opened 2024-07-29
Copter: QuickTune as part of Autotune

IamPete1:
This is a very WIP version of what QuickTune might look like as part of autotune. This builds on https://github.com/ArduPilot/ardupilot/pull/27578. Still a fair bit of tidy up needed (in the heli case for example).  

This adds a new `ATUN_TYPE` param. 1 for the existing autotune, 3 for quick tune. I had to shorten `AUTOTUNE` to `ATUN` for the existing quick tune params to fit. After that is set appropriately the existing autotune works exactly as before. QuickTune also works exactly as it does now, except you have to switch into autotune mode first. The autotests should confirm this.

One big advantage of this method is that you can set `ATUN_TYPE` 0 (which is the default) and you get no additional params and neither methods of autotuning are allocated, so you save RAM.

I have not done param conversion (I don't think we really need it in anycase), and I have not implemented on plane. The plane changes would be very similar to copter.

The removes some of the weirdness in copter, getting rid of the pointer used for params and passing references (attitude control, pos control, interal nav and motors.) to autotune one time rather than on each mode intit.

This allows quicktune in more "modes" than the current PR, it adds pos hold and stabilize.

This makes it a little more painless to compile out one or other tuneing methods since the vehicle dosn't need to know. It also makes it easyer to add a new method in the future.


lthall:
I was not expecting that. You are a MACHINE!

andyp1per:
That's amazing @IamPete1 !

MichelleRos:
I really think this is a _very_ bad idea which will leave us with no end of issues in the future.

These are two totally different tuning algorithms, and I don't really like my work being just taken and mangled into this.

andyp1per:
This is a great step forward! It will allow us to produce a future that results in better usability for users as well as superior tunes. I am 100% behind this approach.

IamPete1:
> Looks like it also _destroys_ one of the main things that makes quicktune so much easier to use: Having a position controller that actually holds the copter in one position.

Have you tried it? ;) . It behaves exactly as your PR if you enter from loiter it behaves like loiter, enter from alt hold and it behaves like alt hold ect. See https://github.com/ArduPilot/ardupilot/pull/27688/files#diff-4412323d997c97af2a4b5d097b50c81a23dc38a2e724e61cf39193fa75d860efR127-R132

lthall:
We have an alternative that can be made to work in our existing Copter framework. Tridge can put what ever he wants into Quadplane and remove it again if we do better here. But in the meantime we still have the LUA script for Copter for those that want to use it. That gives us time to do a really good job here. (Although with all the Control Dev's being willing to contribute to this I suspect it will be faster than expected)

In this form we can:
- replace and test the oscillation detection method with my "pre-oscillation detection" idea if it works as I hope in MATLAB,
- develop a switch pattern that can work consistently across all three autotune variants,
- make this consistent with existing tuning practises (not setting yaw D being the big one),
- start with a simplified parameter set that is potentially shared with the other Autotune modes where appropriate to improve the user experience of both without duplication,
- extend this method to cover some of the simpler control loops where appropriate.

Given we are currently reworking the existing Autotune it is a great time to design a human interface that covers everything effectively with minimal compromises. I don't see why we can't achieve almost the same user experience as the current QuickTune LUA script, consistently over all Autotune variants, while reducing or mitigating some of the performance/safety issues.

MichelleRos:
> Have you tried it? ;)

No. I'm not really interested in trying code that's been *stolen* from me.

MichelleRos:
> Have you tried it? ;)

No. I'm not really interested in trying code that's been taken from my work and shoved into another library.

tridge:
@IamPete1 I'm glad to see you involved in helping bring quicktune into C++, but I don't think the approach in this PR  is better for users. Apart from the rather odd trick of calling last_mode->run() (I'm not sure if I should admire that hack or not!) it still doesn't achieve the workflow on quicktune.
Take for example the common pattern of starting the tune in LOITER, then once done but before saving switching to ALT_HOLD, testing the tune with aggressive stick inputs before deciding if you will save or not. Testing on this PR that aborts when I change to ALT_HOLD. You need to then switch back to AUTOTUNE then re-activate the tune.
This additional complexity in the workflow stems from running the quicktune inside AUTOTUNE mode, but using the mode doesn't buy us anything. I think it is a "tail wagging the dog" situation, where we are using the mode not because it makes for a good user interface, but because that is the structure of AC_AutoTune. So code structure is driving interface, where good design starts with the user experience we want then makes the code fit that user experience. 
The awkward last_mode->run() is a symptom of this, but isn't the root cause. The root cause is the assumption that all tuning has to be inside AC_AutoTune, and I think we are much better off keeping them separate and letting them evolve separately.

tridge:
@IamPete1 I also think there is a bug in here somewhere. The first run I did of your PR I got up to slew rates of 50 on roll with the builtin copter. Unfortunately I hadn't set LOG_BITMASK to 65535, so the slew rates were not recorded. I fixed that for later tests but it didn't happen again. Something weird happened with the tune - the roll slew rate got very high after the end of the roll tune.

lthall:
> it still doesn't achieve the workflow of quicktune.

I also don't agree that the workflow of quicktune is necessarily what we want to achieve, a good starting point but there is room for significant improvement. I also don't agree with the repeated premise that the workflow of quicktune is a wonderful user experience and that the autotune work flow can't be improved to compete. I think this is why the advocates of quicktune are refusing to engage in the conversation on how we could improve the workflow of Autotune to learn from the quicktune experience.

In fact, Autotune initially had a very similar user interface. There is no reason we can't make this consistent between the different approaches. (with some safety improvements)

I think one of the key benefits of this PR is it lets us improve the workflow of quicktune to improve safety, reliability and overall outcome.

> So code structure is driving interface, where good design starts with the user experience we want then makes the code fit that user experience.

I think we can all agree on this. However, I think the control dev's emphasize the tuning experience as a whole, not just quick tune in isolation. We also place a high priority on the positive user experience derived from reliability, consistency, safety of the tuning algorithms.

Best I can tell the Copter developers, the control focused devs in particular, overwhelmingly preferer the approach taken in this PR.

tridge:
> I also don't agree that the workflow of quicktune is necessarily what we want to achieve, a good starting point but there is room for significant improvement.

what workflow would you like to see? Have you tried the quicktune workflow? have you tried the workflow in this PR?
Do you realise that with this PR it is not possible to do quicktune followed by autotune in one flight (you have to reboot between), but with the quicktune PR proposed by Michelle you can do quicktune followed by autotune in one flight.

> I also don't agree with the repeated premise that the workflow of quicktune is a wonderful user experience and that the autotune work flow can't be improved to compete

It isn't a competition!
I'm trying to get a good user experience for users starting out with ArduPilot. I've watched lots of users (via our support proxy system) struggling to get an initial tune on their aircraft. I've learned from that what users find hard and what they don't find hard.


> However, I think the control dev's emphasize the tuning experience as a whole, not just quick tune in isolation.

and yet it makes that experience as a whole worse. You can't test a tune in ALT_HOLD before accepting it. You can't do quicktune followed by autotune in one flight. You have to use both an aux switch and a mode to trigger quicktune, which really makes no sense.
The comments I've seen from @andyp1per @IamPete1 and you all seem focussed on code structure, ie "it should be part of autotune". It is code structure driving user experience, and it is common in the ArduPilot world, and is one of the reasons ArduPilot has a reputation for being hard to use.
Can you tell me any way in which this PR gives a better user experience for a new user than the PR proposed by Michelle? The maths is the same (modulo possible bugs), so the only change in this PR is user experience, and having tried both (yes, I've run this PR several times in SITL while looking at what you have to do to make it work) it is very obvious this is a worse user experience. An experienced user will cope with that, but that isn't who we are aiming for. You Andy and Pete are not the target users.

lthall:
> what workflow would you like to see?

I am still trying to get feedback and consensus but my leading option is identical to Quicktune with gains saved on disarm if in the tuned gains position.

Low - Load original gains / do nothing
Mid - Tune or do nothing if complete (we can make this change to autotune mode too potentially but only if the tune is not completed)
High - Load tuned gains if complete and save them on disarm

In theory the user could:

- Wait for a calm day and go to an open area with good GPS reception
- Connect with a ground station (e.g Mission Planner or QGC) and ensure that the Messages tab can be seen. This is where output from the tune will appear
- Move the RC switch to the low position OR push MP’s Aux Function’s “Low” button
- Arm and takeoff in Loiter mode and climb to a height of about 3m
- Begin the tune by moving the RC switch to the middle position OR push MP’s Aux Function’s “Mid” button
- Monitor the progress of the tune using the GCS’s Messages tab
- If necessary reposition the vehicle using the RC transmitter. This will temporarily pause tuning and restore the original gains. Tuning will resume a few seconds after the RC sticks are returned to their center position
- _If the vehicle begins oscillating violently cancel the tune by moving the RC switch to the low position OR push MP’s Aux Function’s “Low” button_ **(I think we can remove this risk and therefor this line)**
- Once the tune has completed accept the new gains by moving the RC aux switch to the high position OR push MP’s Aux Function’s “High” button
- **The original and tuned gains can be compared by switching between the high and low position while in any mode.**
- Land and disarm the vehicle. **At this time the gains are stored if the switch is in the high position.**

Yes, that is directly copied from the QuickTune Lua script instructions with some small changes to improve safety.

> You can't test a tune in ALT_HOLD before accepting it.

We once could do exactly that and can make it do that again.

> You can't do quicktune followed by autotune in one flight.

I don't see any reason why we can't make that possible if we consider it safe to do so.

> You have to use both an aux switch and a mode to trigger quicktune, which really makes no sense.

We could remove this requirement as we originally did with Autotune. This capability was removed accidentally despite it still passing the autotest.

> Can you tell me any way in which this PR gives a better user experience for a new user than the PR proposed by Michelle?

1. There does not appear to be any reason why we can't provide an identical user experience while using this structure. (However I think we can improve it a little)
2. We can share parameters with the other AutoTune methods to reduce the complexity.
3. We can share the same process with other AutoTune methods to reduce complexity.
4. We have more freedom to add stimulus in a safe way. This potentially improves the quality and reliability of the tune.

> The comments I've seen from @andyp1per @IamPete1 and you all seem focussed on code structure, ie "it should be part of autotune".

This is because we think we can do a better job in every way, especially user experience. The structure enables some of these capabilities.

tridge:
> I am still trying to get feedback and consensus but my leading option is identical to Quicktune with gains saved on disarm if in the tuned gains position.

I think the concept of saving gains on disarm is a bad idea. It violates one of the principles of UX design of not overloading functions. Disarming is not naturally associated with saving gains, and it leads to situations where bad gains can be saved. For example,a tune goes really badly wrong and the vehicle either goes out of control or crashes. The first instinct of the pilot is to disarm to stop the motors spinning. That action should not then save the gains, especially as those gains may be a one of the reasons the vehicle lost control. Expecting pilots to remember to change modes first, or move the tuning switch before disarming isn't great as they may be highly stressed at seeing the vehicle lose control - in that situation they are less likely to remember the required sequence.

> In theory the user could:

I note that the steps you outline don't involve the use of a special AUTOTUNE mode. That is good, but it is not anything like what this PR does. It is much closer to what is achieved by the PR from Michelle that you dislike so much.

> There does not appear to be any reason why we can't provide an identical user experience while using this structure. (However I think we can improve it a little)

again, this PR doesn't achieve that, or in fact anything like it. 



> We can share parameters with the other AutoTune methods to reduce the complexity.

except you can't share parameters with the structure Pete has chosen in this PR, because only one backends parameters are instantiated at a time (with the structure in this PR you have to reboot to change which parameter set is available). It could be achieved with the PR from Michelle as AP_QuickTune could be accessed via our usual singleton method between libraries.

I am however rather dubious as to the value of being able to share parameters. What parameter would we share? I suppose we could share the PI ratio, but the win from sharing the one parameter doesn't seem worth the additional complexity.



> 3\. We can share the same process with other AutoTune methods to reduce complexity.

again, this PR doesn't achieve this at all. This PR makes running multicopter autotune harder (you have to set an additional parameter, then reboot and refresh parameters), and makes running qucktune harder, loses key features (see above comments for feature loss) and adds complexity to the process of tuning, bringing in an unnecessary mode.

> 4\. We have more freedom to add stimulus in a safe way. This potentially improves the quality and reliability of the tune.

why can't we build in stimulus in AP_QuickTune? I'd prefer to get the current no-stimulus method in to start with as it has worked so well, but there is no reason we couldn't add a stimulus parameter for how much additional stimulus to provide, once we work out the methodology and test it sufficiently.

lthall:
> I think the concept of saving gains on disarm is a bad idea.

You are saying that saving gains is dangerous because you could:
1. crash with bad gains selected,
2. save them on disarm because you were able to land right way up (crash check shouldn't save gains)
3. fix your aircraft,
4. not check your parameters,
5. take off again and crash due to the initial bad gains again.

Vs Quicktune's current save on a switch that can easily get knocked to save or moved in the wrong direction under the pressure of bad tune wobbles. This is indicative of the ridiculous double standards you need to apply to argue in favour of Quicktune.

> I note that the steps you outline don't involve the use of a special AUTOTUNE mode.

No, but it can and will use AUTOTUNE mode. But it clearly shows that your "structure is driving interface" argument doesn't hold water.

> again, this PR doesn't achieve that, or in fact anything like it.
x3
No but it clearly shows that your "it is too much work to incorporate Quicktune into the existing Auotutne structure" doesn't hold water. The switch interface will be done in Autotune in some form even if we keep Quicktune a LUA script for Copter.

> except you can't share parameters with the structure Pete has chosen in this PR

You are not going to make the code retard of the team point out how we can share parameters are you? I know better than this and you know better than this.

> I suppose we could share the ....

Thank you for conceding a reasonable point.

> why can't we build in stimulus in AP_QuickTune?

Because you are not in control of the mode you are flying in. You will add stimulus to make quicktune work reliably. Your next step will be to ask to make changes to other modes to support your need to make QuickTune safe with that stimulous. Then we Copter is left with Quicktune code cancer as it spreads tendrils through everything in the name of "safety".

tridge:
> Vs Quicktune's current practice of saving on a switch that can easily get knocked to save or moved in the wrong direction under the pressure of bad tune wobbles. This is indicative of the ridiculous double standards you need to apply to argue in favour of Quicktune.

It isn't a double standard, it is applying UX principles. The arm/disarm switch is for arming and disarming. The aux switch for quicktune is for tuning, aborting and saving gains. You are used to "save on disarm" so it doesn't seem strange to you, but try to view it from a new users perspective. Disarming having side effects of saving tuning gains is an oddity that new users will not expect.
yes, you can knock a switch and save gains when you don't mean to, but ArduPilot is being told to save the gains. Disarming is not a command to save gains, it was just done that way because it made it easier for developers, not because it was good UX design.


> Because you are not in control of the mode you are flying in. You will add stimulus to make quicktune work reliably. Your next step will be to ask to make changes to other modes to support your need to make QuickTune safe with that stimulous. Then we Copter is left with Quicktune code cancer as it spreads tendrils through everything in the name of "safety".

have you noticed that this PR from Pete does "last_mode->run()" ? So the same issues arise. You could be running ALT_HOLD or LOITER or POS_HOLD under the hood when running quicktune on this PR, it is just hidden from the user.

lthall:
> It isn't a double standard

We are comparing:
Proposed Autotune: Switch selects between original and tuned gains where the tuned gains are made permanent if selected during disarming.
Is more dangerous than
Quicktune: Switch selects between original and tuned gains where the tuned gains are saved immediately if the switch is bumped to the wrong position. With no way to fix the problem.

And the example you give argues that the proposed Autotune behaviour is more dangerous than the QuickTune approach because the gains would get saved after landing. Yet in the current QuickTune methodology they would have been saved already in the air.

> especially as those gains may be a one of the reasons the vehicle lost control.

With the QuickTune approach the problem of the tuned gains can't be fixed once the switch goes to the saved position, even if the pilot recognises their mistake. With the approach I am proposing it can be fixed by simply dropping the switch the original gain position.

It is a very clear double standard to make the claim the proposed switch behaviour is more dangerous when the exact same situation is clearly more likely to happen with the current Quicktune behaviour.

> Disarming having side effects of saving tuning gains is an oddity that new users will not expect.

The current selected tune being made permeant on disarming being an oddity is a stretch given it will be transparent to the user. How can the user even tell the difference between that and us loading and saving the new tune / original tune based on switch position. Now I say that, the only down side of loading and saving the original/new rather than just loading the original/new is the ability to not save the new tune if we have a crash detection. That removes the need to add code to the disarm process and keeps it all nicely contained to the flight mode.

Nice, I really like that. That is actually very helpful as I didn't like having to add code to our disarm process.

> Disarming is not a command to save gains, it was just done that way because it made it easier for developers, not because it was good UX design.

We just fixed this problem, thanks! Just comes down to if others agree with you. I am happy with each approach as they are both identical from the users perspective. But I am leaning heavily towards the approach of saving them immediately limits code tendrils where we don't need them.

> The aux switch for quicktune is for tuning, aborting and saving gains.

This is a good summary of on of my main safety concerns.  The switch is used tuning, aborting and saving gains. However it can't abort saved gains.

Now that is something that "is an oddity that new users will not expect".... as they crash.

> Disarming is not a command to save gains, 

True, but our options are limited and it was safe because it gave the user the ability to check the results before flying with them.

> it was just done that way because it made it easier for developers, not because it was good UX design.

This is just an unproductive insult regarding the effort and thought Randy, myself and others into the process and safety of Autotune. Lets keep this discussion clean mate.

> have you noticed that this PR from Pete does "last_mode->run()" ? So the same issues arise.
The difference is we don't have to this way. This PR was done with the intention of clearly demonstrating that we could to it this way in reaction to your "it is too much work to incorporate quicktune into the existing Auotutne structure" argument. However it provides much more scope to fix safety problems with further changes. In short, Pete wrote this explicitly to have the same problems the current quicktune LUA script has.

As uncomfortable as arguing with you is, it has been very helpful refining the next set of changes I plan to make to Autotune. In particular regarding reviving the switch control capability and the form it should take.

timtuxworth:
From where I sit - both approaches seem to be sub-optimal and potentially error prone. I'd take a step back. If we want a good, safe, easy to understand and explain UX (given all the limitations blah, blah), what would we want?

1. A simple way to start tuning
2. A simple way to test new vs old gains from the tuning before committing the change
3. An obvious way to discard the new gains if they are not working
4. A very clean and simple way to "really save" the new gains once really, really sure they are good

The essence of the problem is that the "really save" shouldn't happen while flying. I agree with Leonard on this, but I agree with Tridge that disarm = save = bad. So to keep it simple what about this?

While flying the 3pos (or aux) switch is

- low: old gains
- mid: tune
- high new gains

AFTER disarm the 3pos switch is:
If disarmed after tuning has happened - GCS message "Tuning - switch to mid to finish"
User must then switch to the mid position if not already there, then they can either flip the switch:

- low: discard tune and use the old gains - GCS message "tune discarded"
- high: "really save" the new gains - GCS message "tune saved"

power off or reboot before doing this, will revert to old gains

PS: I'd love to see this applied to plane fixed wing tune too - it has the same problems.

lthall:
Thanks @timtuxworth 

I appreciate the clear, well thought out and justified suggestion!

mbuzdalov:
> If disarmed after tuning has happened - GCS message "Tuning - switch to mid to finish"

From a point of view of a person who typically tunes small craft without any GCS connection and, recently, even without an FPV feed, I would appreciate additional visual indication at this stage via LEDs: so that a user sees something requires their attention actively, and they remember to perform the final selection sequence. Studying logs for parameters to save is nice, but realistically may not be feasible on smaller craft with only 8Mb or so of flash.

(For similar reasons I would love to have a visual LED-based indication of a tuning stage finished successfully or failed, but that's a slightly different story)

PR #27689, opened 2024-07-29
Tools: environment_install: fix focal install

peterbarker:
this was only intended to narrow an error being seen with Focal, but appears to fix it instead


peterbarker:
If this issue is resolved we might be able to unpin this package: https://github.com/pypa/setuptools/issues/4511

PR #27693, opened 2024-07-29
Fix for unhandled TKOFF_THR_MAX

Georacer:
This patches a bug that was introduced via https://github.com/ArduPilot/ardupilot/pull/27174.

It fixes an unhandled case of `TKOFF_THR_MAX=0`, where `THR_MAX` would not be sourced as a value for the minimum throttle limit during takeoff.
The relevant conditions are:
- TKOFF_OPTIONS=0 (default)
- TKOFF_THR_MAX=0 (default)

This would result in incorrect backing-off of the minimum throttle during the climb, which @Hwurzburg spotted:
image

Now it behaves as it should:
image
Log: https://www.dropbox.com/scl/fi/akdqr9o5niv3qleexh3yd/122221_TakeoffTakeoff4.bin?rlkey=o9uykk6cm7785oxctkj17txxq&dl=0

---
Side-note:

I notice that there was significant pitch and throttle oscillation in Henry's log.
I don't think this is related to #27174, as the oscillation persists even after TAKEOFF mode has finished climbing, where TECS operates in a regime which I haven't edited.
I believe that a sudden pitch-down and throttle-cut was triggered (because of this bug) during the climb and the default tuning and model physics would yield bad control behaviour.

### DIFF
--- a/ArduPlane/servos.cpp
+++ b/ArduPlane/servos.cpp
@@ -534,12 +534,17 @@ float Plane::apply_throttle_limits(float throttle_in)
             // Or (in contrast) to give some extra throttle during the initial climb.
             max_throttle = aparm.takeoff_throttle_max.get();
         }
+
         // Do not allow min throttle to go below a lower threshold.
         // This is typically done to protect against premature stalls close to the ground.
         const bool use_throttle_range = (aparm.takeoff_options & (uint32_t)AP_FixedWing::TakeoffOption::THROTTLE_RANGE);
         if (!use_throttle_range || !ahrs.using_airspeed_sensor()) {
             // Use a constant max throttle throughout the takeoff or when airspeed readings are not available.
-            min_throttle = MAX(min_throttle, aparm.takeoff_throttle_max.get());
+            if (aparm.takeoff_throttle_max.get() == 0) {
+                min_throttle = MAX(min_throttle, aparm.throttle_max.get());
+            } else {
+                min_throttle = MAX(min_throttle, aparm.takeoff_throttle_max.get());
+            }
         } else if (use_throttle_range) { // Use a throttle range through the takeoff.
             if (aparm.takeoff_throttle_min.get() != 0) { // This is enabled by TKOFF_MODE==1.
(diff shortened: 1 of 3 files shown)


Hwurzburg:
This makes it safe from my sitl testing...
in another PR we need to get the model fixed so as not to go unstable
and , more importantly, get the takeoff leveling working again to prevent the overshoots in altitude on auto takeoffs

Hwurzburg (approved):

peterbarker:
Pretty sure this is breaking several Plane autotests.

Georacer:
@peterbarker I believe the latest commit fixed it.
It was a case of the minimum throttle during a transition being set to TAKEOFF_THR_MAX. This is because in servos.cpp the transition throttle limits were historically being treated the same as the takeoff throttle limits. And this resulted in a back-transition getting a **minimum** throttle allowance of TAKEOFF_THR_MAX.

For now I have reverted to the behaviour before #27174.
@tridge I'll have to find another way to introduce the minimum throttle during a transition (and find a way to tell between a forward and backward transition).

The failing autotests passed locally, but let's see what the CI server says about it.

ArduPlane/servos.cpp
```diff
@@ -551,6 +553,15 @@ float Plane::apply_throttle_limits(float throttle_in)
         min_throttle = 0;
     }
 
+    // Handle throttle limits for transition conditions.
+#if HAL_QUADPLANE_ENABLED
+    if (quadplane.in_transition()) {
```
tridge, line 558: why not this?
image

tridge (changes requested):
i think we can make min takeoff throttle work for fwd transition

ArduPlane/servos.cpp
tridge: note that in_transition() is only true for fwd transitions

tridge: turns out in_transiton() is true for tailsitters all the time, which doesn't seem right
I think we need a new in_forward_transition() which is only true for the fwd transition

tridge (approved):

PR #27694, opened 2024-07-29
Plane: PitchController: add rate and actuator output Sys ID inject

IamPete1:
This adds a path for injecting outputs for the purposes of system identification of fixed wing vehicles. This follows the pattern of how its done for copter in AC_AttitudeControl.

Just a draft for now, needs testing and a similar patch for roll.


IamPete1:
This has been overtaken by https://github.com/ArduPilot/ardupilot/pull/29608

Issue #9419, opened 2018-09-13
Enhancement Request: iBus/MavLink interoperation & compatiblity


Saijin-Naib:
#### Issue details
Desired Feature:
iBus Support in ArduPilot to enable direct usage of FlySky/Turnigy receivers and transmitters, as well as telemetry modules.
Reasoning:
Low-cost, high quality/reliability, bi-directional capable, iBus apparently has lower latency than even sbus (via [Oscar Liang](https://oscarliang.com/pwm-ppm-sbus-dsm2-dsmx-sumd-difference/))
Expectation:
ArduPilot/MavLink integration with iBus receivers and telemetry modules (input of things like voltage, optical/magnetic RPM, altitude/variometer, voltage, etc)
MavLink to iBus sensor integration:
MavLink data to be transformed/transmitted via iBus to show up as a valid telemetry sensors on an iBus Tx (altitude, voltage, speed, etc)

#### Version
ArduPilot Master

#### Platform
[X] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

#### Airframe type
3DR Solo (quad (X))

#### Hardware type
Pixhawk 2.0 Cube

#### Logs
N/A
Related post:
https://discuss.ardupilot.org/t/mavlink-telemetry-on-flysky-i6/32848

Related GitHub code for MavLink to iBus telemetry (running on separate ATMega) :
https://github.com/qba667/MAVLinkToIbus

b14ckyy:
Agree! 
As I'm very happy with my Flysky i6 with 10ch FW, I'd like to use all channels instead only 8 with PPM. Please add iBUS support too.

b14ckyy:
@PraiseSatan where do you have an iBUS port on your pixhawk? I have a Pixhak 2.4.8 here and there is only one RC input and this one only supports PPM and S-BUS. When connecting any iBus reciever (X6B), Arducopter does not detect any signals. Only PPM works. If you know any option to use iBus, I'd be happy to get some details :)

PraiseSatan:
@b14ckyy I meant the ibus port on my FS-ia6b, the RX for the remote. I connect from there to the RC in on the Pixhawk.

With the 10ch firmware I have, in the setup menu, there is a RX setup menu. In that there is a PPM Output selection, which allows you to change how the RX sends its data.
I have mine set to SBUS/PWM.

I'm not sure if that will work with all RXs, or just the FS-ia6B.
[...]

PR #16545, opened 2021-02-07
Adding ibus telemetry support via serial to arducopter

jverhuls:
This is my first pull request. I tried to follow the guidelines as much as possible. 
Sorry if i made newbie mistakes.

PR #27695, opened 2024-07-30
Add support for i-BUS telemetry

nicole-ashley:
Implements: [#9419: Enhancement Request: iBus/MavLink interoperation & compatiblity](https://github.com/ArduPilot/ardupilot/issues/9419)

Wiki PR: https://github.com/ArduPilot/ardupilot_wiki/pull/6154

Originally based on [#16545: Adding ibus telemetry support via serial to arducopter](https://github.com/ArduPilot/ardupilot/pull/16545) but heavily rewritten and expanded. The old PR has not been updated in two years and I offered to take over the implementation with no objections.

I'm an experienced software engineer but C++ is not my wheelhouse and this is my first contribution to ArduPilot, so I welcome maintainer feedback especially around memory and CPU management, and potential for causing crashes. This code is tested on an fmuv3 board and has so far proven to be stable. Builds based on Stable for SITL and other boards I've seen requested are available here: https://1drv.ms/f/s!AoxGPryyVyPN1pNwsxI6dRJVblKOug?e=YVYKls

### DIFF
--- a/libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
+++ b/libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
@@ -1,0 +1,792 @@
+/*
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+   i-BUS telemetry for FlySky/Turnigy receivers and other peripherals
+   (eg iA6B, iA10) by Nicole Ashley <nicole@ashley.kiwi>.
+
+   Originally based on work by Jan Verhulst:
+     https://github.com/ArduPilot/ardupilot/pull/16545
+
+   Libraries used for reference and inspiration:
+
+   * iBUStelemetry
+     https://github.com/Hrastovc/iBUStelemetry
+
+   * IBusBM
+     https://github.com/bmellink/IBusBM
+
+   * BetaFlight
+     https://github.com/betaflight/betaflight/blob/master/src/main/telemetry/ibus_shared.c
+ */
+
+#include <AP_IBus_Telem/AP_IBus_Telem.h>
+
+#if AP_IBUS_TELEM_ENABLED
+
+#include <AP_HAL/AP_HAL.h>
+#include <AP_Vehicle/AP_Vehicle.h>
+#include <AP_BattMonitor/AP_BattMonitor.h>
+#include <AP_Arming/AP_Arming.h>
+#include <AP_GPS/AP_GPS.h>
+#include <AP_RPM/AP_RPM.h>
+
+// 2-byte values
+#define IBUS_SENSOR_TYPE_TEMPERATURE          0x01 // Temperature (in 0.1 degrees, where 0=-40'C)
+#define IBUS_SENSOR_TYPE_RPM_FLYSKY           0x02 // FlySky-specific throttle value
+#define IBUS_SENSOR_TYPE_EXTERNAL_VOLTAGE     0x03 // External voltage (in centivolts, so 1450 is 14.50V)
+#define IBUS_SENSOR_TYPE_AVERAGE_CELL_VOLTAGE 0x04 // Avg cell voltage (in centivolts, so 1450 is 14.50V)
+#define IBUS_SENSOR_TYPE_BATTERY_CURRENT      0x05 // Battery current (centi-amps)
+#define IBUS_SENSOR_TYPE_FUEL                 0x06 // Remaining battery percentage
+#define IBUS_SENSOR_TYPE_RPM                  0x07 // Throttle value (in 0.01, so 1200 is 12.00%)
+#define IBUS_SENSOR_TYPE_COMPASS_HEADING      0x08 // Heading (0-360 degrees)
+#define IBUS_SENSOR_TYPE_CLIMB_RATE           0x09 // Climb rate (cm/s)
+#define IBUS_SENSOR_TYPE_COG                  0x0a // Course over ground (centidegrees, so 27015 is 270.15 degrees)
+#define IBUS_SENSOR_TYPE_GPS_STATUS           0x0b // GPS status (2 values: fix type, and number of satellites)
+#define IBUS_SENSOR_TYPE_ACC_X                0x0c // Acc X (cm/s)
+#define IBUS_SENSOR_TYPE_ACC_Y                0x0d // Acc Y (cm/s)
+#define IBUS_SENSOR_TYPE_ACC_Z                0x0e // Acc Z (cm/s)
+#define IBUS_SENSOR_TYPE_ROLL                 0x0f // Roll (centidegrees)
+#define IBUS_SENSOR_TYPE_PITCH                0x10 // Pitch (centidegrees)
+#define IBUS_SENSOR_TYPE_YAW                  0x11 // Yaw (centidegrees)
+#define IBUS_SENSOR_TYPE_VERTICAL_SPEED       0x12 // Vertical speed (cm/s)
+#define IBUS_SENSOR_TYPE_GROUND_SPEED         0x13 // Speed (cm/s)
+#define IBUS_SENSOR_TYPE_GPS_DIST             0x14 // Distance from home (m)
+#define IBUS_SENSOR_TYPE_ARMED                0x15 // Armed / unarmed (1 = armed, 0 = unarmed)
+#define IBUS_SENSOR_TYPE_FLIGHT_MODE          0x16 // Flight mode
+#define IBUS_SENSOR_TYPE_ODO1                 0x7c // Odometer1
+#define IBUS_SENSOR_TYPE_ODO2                 0x7d // Odometer2
+#define IBUS_SENSOR_TYPE_SPEED                0x7e // Speed km/h
+#define IBUS_SENSOR_TYPE_ALT_FLYSKY           0xf9 // FlySky-specific altitude (metres)
+
+// 4-byte values
+#define IBUS_SENSOR_TYPE_TEMPERATURE_PRESSURE 0x41 // Combined temperature & pressure value
+#define IBUS_SENSOR_TYPE_GPS_LAT              0x80 // WGS84 in degrees * 1E7
+#define IBUS_SENSOR_TYPE_GPS_LNG              0x81 // WGS84 in degrees * 1E7
+#define IBUS_SENSOR_TYPE_GPS_ALT              0x82 // GPS (cm)
+#define IBUS_SENSOR_TYPE_ALT                  0x83 // Alt (cm)
+#define IBUS_SENSOR_TYPE_ALT_MAX              0x84 // MaxAlt (cm)
+
+// i-BUS vehicle modes
+#define IBUS_VEHICLE_MODE_STAB    0
+#define IBUS_VEHICLE_MODE_ACRO    1
+#define IBUS_VEHICLE_MODE_AHOLD   2
+#define IBUS_VEHICLE_MODE_AUTO    3
+#define IBUS_VEHICLE_MODE_GUIDED  4
+#define IBUS_VEHICLE_MODE_LOITER  5
+#define IBUS_VEHICLE_MODE_RTL     6
+#define IBUS_VEHICLE_MODE_CIRCLE  7
+#define IBUS_VEHICLE_MODE_PHOLD   8
+#define IBUS_VEHICLE_MODE_LAND    9
+#define IBUS_VEHICLE_MODE_UNKNOWN 255 // Must be positive and 0 is already used; out of range blanks the value
+
+// All the sensors we can accurately provide are listed here.
+// i-BUS will generally only query up to 15 sensors, so subjectively
+// higher-value sensors are sorted to the top to make the most of a
+// small telemetry window. In the future these could be configurable.
+static const AP_IBus_Telem::SensorDefinition sensors[] {
+#if AP_ARMING_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_ARMED, .sensor_length = 2},
+#endif
+    {.sensor_type = IBUS_SENSOR_TYPE_FLIGHT_MODE, .sensor_length = 2},
+#if AP_GPS_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_GPS_STATUS, .sensor_length = 2},
+#endif
+#if AP_BATTERY_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_FUEL, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_EXTERNAL_VOLTAGE, .sensor_length = 2},
+#endif
+#if AP_BARO_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_ALT, .sensor_length = 4},
+#endif
+#if AP_AHRS_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_GPS_DIST, .sensor_length = 2},
+#endif
+#if AP_BARO_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_CLIMB_RATE, .sensor_length = 2},
+#endif
+#if AP_GPS_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_GROUND_SPEED, .sensor_length = 2},
+#endif
+#if AP_AHRS_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_ROLL, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_PITCH, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_YAW, .sensor_length = 2},
+#endif
+#if AP_AIRSPEED_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_SPEED, .sensor_length = 2},
+#endif
+#if AP_BARO_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_TEMPERATURE_PRESSURE, .sensor_length = 4},
+#endif
+#if AP_RPM_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_RPM, .sensor_length = 2},
+#endif
+#if AP_BATTERY_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_BATTERY_CURRENT, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_AVERAGE_CELL_VOLTAGE, .sensor_length = 2},
+#endif
+#if AP_AHRS_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_COMPASS_HEADING, .sensor_length = 2},
+#endif
+#if AP_GPS_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_COG, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_GPS_LAT, .sensor_length = 4},
+    {.sensor_type = IBUS_SENSOR_TYPE_GPS_LNG, .sensor_length = 4},
+    {.sensor_type = IBUS_SENSOR_TYPE_GPS_ALT, .sensor_length = 4},
+#endif
+#if AP_AHRS_ENABLED
+    {.sensor_type = IBUS_SENSOR_TYPE_ACC_X, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_ACC_Y, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_ACC_Z, .sensor_length = 2},
+#endif
+};
+
+#if APM_BUILD_TYPE(APM_BUILD_Rover)
+
+/* Rover modes:
+    MANUAL       = 0
+    ACRO         = 1
+    STEERING     = 3
+    HOLD         = 4
+    LOITER       = 5
+    FOLLOW       = 6
+    SIMPLE       = 7
+    DOCK         = 8
+    CIRCLE       = 9
+    AUTO         = 10
+    RTL          = 11
+    SMART_RTL    = 12
+    GUIDED       = 15
+    INITIALISING = 16
+*/
+static const AP_IBus_Telem::ModeMap mode_map[] {
+    {.ap_mode = 1, .ibus_mode = IBUS_VEHICLE_MODE_ACRO},
+    {.ap_mode = 4, .ibus_mode = IBUS_VEHICLE_MODE_PHOLD},
+    {.ap_mode = 5, .ibus_mode = IBUS_VEHICLE_MODE_LOITER},
+    {.ap_mode = 9, .ibus_mode = IBUS_VEHICLE_MODE_CIRCLE},
+    {.ap_mode = 10, .ibus_mode = IBUS_VEHICLE_MODE_AUTO},
+    {.ap_mode = 11, .ibus_mode = IBUS_VEHICLE_MODE_RTL},
+    {.ap_mode = 12, .ibus_mode = IBUS_VEHICLE_MODE_RTL},
+    {.ap_mode = 15, .ibus_mode = IBUS_VEHICLE_MODE_GUIDED},
+};
+
+#elif APM_BUILD_COPTER_OR_HELI
+
+/* Copter modes:
+    STABILIZE    =  0
+    ACRO         =  1
+    ALT_HOLD     =  2
+    AUTO         =  3
+    GUIDED       =  4
+    LOITER       =  5
+    RTL          =  6
+    CIRCLE       =  7
+    LAND         =  9
+    DRIFT        = 11
+    SPORT        = 13
+    FLIP         = 14
+    AUTOTUNE     = 15
+    POSHOLD      = 16
+    BRAKE        = 17
+    THROW        = 18
+    AVOID_ADSB   = 19
+    GUIDED_NOGPS = 20
+    SMART_RTL    = 21
+    FLOWHOLD     = 22
+    FOLLOW       = 23
+    ZIGZAG       = 24
+    SYSTEMID     = 25
+    AUTOROTATE   = 26
+    AUTO_RTL     = 27
+    TURTLE       = 28
+*/
+static const AP_IBus_Telem::ModeMap mode_map[] {
+    {.ap_mode = 0, .ibus_mode = IBUS_VEHICLE_MODE_STAB},
+    {.ap_mode = 1, .ibus_mode = IBUS_VEHICLE_MODE_ACRO},
+    {.ap_mode = 2, .ibus_mode = IBUS_VEHICLE_MODE_AHOLD},
+    {.ap_mode = 3, .ibus_mode = IBUS_VEHICLE_MODE_AUTO},
+    {.ap_mode = 4, .ibus_mode = IBUS_VEHICLE_MODE_GUIDED},
+    {.ap_mode = 5, .ibus_mode = IBUS_VEHICLE_MODE_LOITER},
+    {.ap_mode = 6, .ibus_mode = IBUS_VEHICLE_MODE_RTL},
+    {.ap_mode = 7, .ibus_mode = IBUS_VEHICLE_MODE_CIRCLE},
+    {.ap_mode = 9, .ibus_mode = IBUS_VEHICLE_MODE_LAND},
+    {.ap_mode = 16, .ibus_mode = IBUS_VEHICLE_MODE_PHOLD},
+    {.ap_mode = 21, .ibus_mode = IBUS_VEHICLE_MODE_RTL},
+    {.ap_mode = 22, .ibus_mode = IBUS_VEHICLE_MODE_PHOLD},
+    {.ap_mode = 27, .ibus_mode = IBUS_VEHICLE_MODE_RTL},
+};
+
+#elif APM_BUILD_TYPE(APM_BUILD_ArduPlane)
+
+/* Plane modes:
+    MANUAL        = 0
+    CIRCLE        = 1
+    STABILIZE     = 2
+    TRAINING      = 3
+    ACRO          = 4
+    FLY_BY_WIRE_A = 5
+    FLY_BY_WIRE_B = 6
+    CRUISE        = 7
+    AUTOTUNE      = 8
+    AUTO          = 10
+    RTL           = 11
+    LOITER        = 12
+    TAKEOFF       = 13
+    AVOID_ADSB    = 14
+    GUIDED        = 15
+    INITIALISING  = 16
+    QSTABILIZE    = 17
+    QHOVER        = 18
+    QLOITER       = 19
+    QLAND         = 20
+    QRTL          = 21
+    QAUTOTUNE     = 22
+    QACRO         = 23
+    THERMAL       = 24
+    LOITER_ALT_QLAND = 25
+*/
+static const AP_IBus_Telem::ModeMap mode_map[] {
+    {.ap_mode = 1, .ibus_mode = IBUS_VEHICLE_MODE_CIRCLE},
+    {.ap_mode = 2, .ibus_mode = IBUS_VEHICLE_MODE_STAB},
+    {.ap_mode = 4, .ibus_mode = IBUS_VEHICLE_MODE_ACRO},
+    {.ap_mode = 5, .ibus_mode = IBUS_VEHICLE_MODE_STAB},
+    {.ap_mode = 6, .ibus_mode = IBUS_VEHICLE_MODE_STAB},
+    {.ap_mode = 7, .ibus_mode = IBUS_VEHICLE_MODE_STAB},
+    {.ap_mode = 10, .ibus_mode = IBUS_VEHICLE_MODE_AUTO},
+    {.ap_mode = 11, .ibus_mode = IBUS_VEHICLE_MODE_RTL},
+    {.ap_mode = 12, .ibus_mode = IBUS_VEHICLE_MODE_LOITER},
+    {.ap_mode = 13, .ibus_mode = IBUS_VEHICLE_MODE_AUTO},
+    {.ap_mode = 15, .ibus_mode = IBUS_VEHICLE_MODE_GUIDED},
+    {.ap_mode = 17, .ibus_mode = IBUS_VEHICLE_MODE_STAB},
+    {.ap_mode = 18, .ibus_mode = IBUS_VEHICLE_MODE_PHOLD},
+    {.ap_mode = 19, .ibus_mode = IBUS_VEHICLE_MODE_LOITER},
+    {.ap_mode = 20, .ibus_mode = IBUS_VEHICLE_MODE_LAND},
+    {.ap_mode = 21, .ibus_mode = IBUS_VEHICLE_MODE_RTL},
+    {.ap_mode = 23, .ibus_mode = IBUS_VEHICLE_MODE_ACRO},
+    {.ap_mode = 25, .ibus_mode = IBUS_VEHICLE_MODE_LOITER},
+};
+
+#elif APM_BUILD_TYPE(APM_BUILD_ArduSub)
+
+/* Submarine modes:
+    STABILIZE    = 0
+    ACRO         = 1
+    ALT_HOLD     = 2
+    AUTO         = 3
+    GUIDED       = 4
+    CIRCLE       = 7
+    SURFACE      = 9
+    POSHOLD      = 16
+    MANUAL       = 19
+    MOTOR_DETECT = 20
+*/
+static const AP_IBus_Telem::ModeMap mode_map[] {
+    {.ap_mode = 0, .ibus_mode = IBUS_VEHICLE_MODE_STAB},
+    {.ap_mode = 1, .ibus_mode = IBUS_VEHICLE_MODE_ACRO},
+    {.ap_mode = 2, .ibus_mode = IBUS_VEHICLE_MODE_AHOLD},
+    {.ap_mode = 3, .ibus_mode = IBUS_VEHICLE_MODE_AUTO},
+    {.ap_mode = 4, .ibus_mode = IBUS_VEHICLE_MODE_GUIDED},
+    {.ap_mode = 5, .ibus_mode = IBUS_VEHICLE_MODE_LOITER},
+    {.ap_mode = 6, .ibus_mode = IBUS_VEHICLE_MODE_RTL},
+    {.ap_mode = 7, .ibus_mode = IBUS_VEHICLE_MODE_CIRCLE},
+    {.ap_mode = 8, .ibus_mode = IBUS_VEHICLE_MODE_PHOLD},
+    {.ap_mode = 9, .ibus_mode = IBUS_VEHICLE_MODE_LAND},
+    {.ap_mode = 16, .ibus_mode = IBUS_VEHICLE_MODE_PHOLD},
+};
+
+#elif APM_BUILD_TYPE(APM_BUILD_Blimp)
+
+/* Blimp modes:
+    LAND     = 0
+    MANUAL   = 1
+    VELOCITY = 2
+    LOITER   = 3
+    RTL      = 4
+*/
+static const AP_IBus_Telem::ModeMap mode_map[] {
+    {.ap_mode = 0, .ibus_mode = IBUS_VEHICLE_MODE_LAND},
+    {.ap_mode = 3, .ibus_mode = IBUS_VEHICLE_MODE_LOITER},
+    {.ap_mode = 4, .ibus_mode = IBUS_VEHICLE_MODE_RTL},
+};
+
+#else
+
+static const AP_IBus_Telem::ModeMap mode_map[] {};
+
+#endif
+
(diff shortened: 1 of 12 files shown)


libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+}
+
+void AP_IBus_Telem::tick(void)
+{
+    ensure_listening();
+
+    uint16_t available = MIN(_port->available(), 1024U);
+    if (available == 0) {
+        return;
+    } else {
```
peterbarker, line 61:
```suggestion
    }
```
No `else` clauses after a return statement...

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+        reset_read_state_after_timegap();
```
peterbarker, line 62: Please just in-line this.

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    for (uint16_t i = 0; i < available; i++) {
```
peterbarker, line 65:
```suggestion
    for (auto i = 0; i < available; i++) {
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+        const int16_t c = _port->read();
+        if (c < 0) {
+            return;
+        }
```
peterbarker, line 69:
```suggestion
        uint8_t c;
        if (!_port->read(c)) {
            return;
        }
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+{
+    // Sensor 0 is reserved for the transmitter, so if for some reason it's requested we should ignore it
+    if (_incoming_sensor_id == 0) {
+        return;
+    }
+
+    // If we've reached the end of our sensor list, we shouldn't respond; this tells the receiver that there
+    // are no more sensors to discover.
+    if (
+        _incoming_sensor_id > sizeof(_sensors) / sizeof(_sensors[0])
```
peterbarker, line 133:
```suggestion
    if (_incoming_sensor_id > ARRAY_SIZE(_sensors) ||
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+        || !get_sensor_definition(_incoming_sensor_id)->sensor_type
```
peterbarker, line 134:
```suggestion
        || get_sensor_definition(_incoming_sensor_id)->sensor_type == 0
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
peterbarker: ... actually, change this out for:
```
    const auto *sensor_defintiion = get_sensor_definition(_incoming_sensor_id);
    if (sensor_definition == nullptr || sensor_definition->sensor_type == 0) {
        return;
    }
```

... and change the calls to dependent functions to:
```
 handle_discover_command(*sensor_definition);
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+        (uint8_t)(PROTOCOL_COMMAND_VALUE | _incoming_sensor_id),
+        {value.byte[0], value.byte[1], value.byte[2], value.byte[3]}
+    };
+    populate_checksum((uint8_t*)&packet, sizeof(packet));
+    _port->write((uint8_t*)&packet, sizeof(packet));
+}
+
+AP_IBus_Telem::_sensor_definition* AP_IBus_Telem::get_sensor_definition(uint8_t sensor_id)
+{
+    return &_sensors[sensor_id - 1];
```
peterbarker, line 273: There will probably only be one caller to this with above-requested changes, and this could probably then be in-lined.

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+    default:
+        value.int32 = 0;
+    }
+
+    return value;
+}
+
+uint16_t AP_IBus_Telem::get_average_cell_voltage_cV()
+{
+    if (AP::battery().has_cell_voltages()) {
```
peterbarker, line 395:
```suggestion
    if (!AP::battery().has_cell_voltages()) {
        return 0;
    }
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+        const auto &cell_voltages = AP::battery().get_cell_voltages();
+        const uint8_t number_of_cells = sizeof(cell_voltages.cells) / sizeof(cell_voltages.cells[0]);
```
peterbarker, line 397: use ARRAY_SIZE

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,507 @@
...
+    if (AP::ahrs().get_relative_position_NE_home(home)) {
+        return home.length();
+    }
+    return 0;
+}
+
+uint16_t AP_IBus_Telem::get_vehicle_mode()
+{
+    const AP_AHRS::VehicleClass vehicle_class = AP::ahrs().get_vehicle_class();
+    if (vehicle_class == AP_AHRS::VehicleClass::GROUND) {
```
peterbarker, line 470: `switch`

libraries/AP_IBus_Telem/AP_IBus_Telem.h
```diff
@@ -0,0 +1,355 @@
...
+#define IBUS_VEHICLE_MODE_UNKNOWN 255 // Must be positive and 0 is already used; out of range blanks the value
+
+class AP_IBus_Telem
+{
+public:
+    AP_IBus_Telem() {}
+
+    /* Do not allow copies */
+    AP_IBus_Telem(const AP_IBus_Telem &other) = delete;
+    AP_IBus_Telem &operator=(const AP_IBus_Telem &) = delete;
```
peterbarker, line 102:
```suggestion
               CLASS_NO_COPY(AP_IBus_Telem);
```

libraries/AP_IBus_Telem/AP_IBus_Telem.h
```diff
@@ -0,0 +1,355 @@
...
+    uint8_t _incoming_command;               // The command coming from the receiver
+    uint8_t _incoming_sensor_id;             // The sensor the receiver is talking to us about
+    uint16_t _expected_checksum;             // Checksum calculation
+    uint8_t _incoming_checksum_high_byte;    // Checksum high byte storage
+
+    // All the sensors we can accurately provide are listed here.
+    // i-BUS will generally only query up to 15 sensors, so subjectively
+    // higher-value sensors are sorted to the top to make the most of a
+    // small telemetry window. In the future these could be configurable.
+    _sensor_definition _sensors[25] = {
```
peterbarker, line 179:
```suggestion
    static const SensorDefinition _sensors[25] = {
```

similarly for the other definitions.

These currently consume RAM, which is bad.  `static const` moves it to just flash.

libraries/AP_IBus_Telem/AP_IBus_Telem.h
```diff
@@ -0,0 +1,355 @@
...
+
+#ifndef HAL_IBUS_TELEM_ENABLED
+#define HAL_IBUS_TELEM_ENABLED 1
+#endif
+
+#if HAL_IBUS_TELEM_ENABLED
+
+#include <AP_HAL/AP_HAL.h>
+#include <AP_Math/vector3.h>
+#include <AP_RPM/AP_RPM.h>
```
peterbarker, line 44:
```suggestion
#include <AP_RPM/AP_RPM_config.h>
```

peterbarker (changes requested):
Partial review only, sorry.  This is looking quite reasonable.

Fantastic to get this work!

I have a test suite for IBus which I wrote in Feb 2021 for this which I will get merged (and run this through).

libraries/AP_IBus_Telem/AP_IBus_Telem.h
nicole-ashley: I'm being told I can't initialise these here without also adding `constexpr`; that seems to be working but let me know if there's a preferred way.

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
nicole-ashley: With the change to `static const` that also removes the size ambiguity, meaning this check is now simply `_incoming_sensor_id > ARRAY_SIZE(_sensors)` with no parameter value checks.

However I've retained the `*sensor_definition` variable and passing it to the dependent functions.

peterbarker:
I've fixed up the suite I wrote years ago, and I've PR'd it here: https://github.com/ArduPilot/ardupilot/pull/27695

Needs this patch to run the suite for Copter:
```
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -11770,6 +11770,7 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
             self.InvalidJumpTags,
             self.IMUConsistency,
             self.AHRSTrimLand,
+            self.IBus,
             self.GuidedYawRate,
             self.NoArmWithoutMissionItems,
             self.DO_CHANGE_SPEED_in_guided,
```

And then it's `./Tools/autotest/autotest.py --gdb --debug build.Copter test.Copter.IBus` to run the suite against your code.  Your branch passed that suite when I rebased it on top of my PR branch there.

peterbarker:
We'll want an entry in `Tools/scripts/test_build_options.py` and `extract_features.py` to make it available on our custom build server (https://custom.ardupilot.org).  Swipe the lines used for the LTM telemetry backend)

We will probably need to put this into `minimize_common.py` to save the flash on lower-end boards.  Many of these are right on the edge of not compiling - but the custom build server allows users to remove other features to fit things like this in.

libraries/AP_IBus_Telem/AP_IBus_Telem.h
nicole-ashley: I notice you changed the case for this struct type, so I've applied that to the other struct types as well. Easy to change it back to snake case if that was accidental.

nicole-ashley:
Thanks @peterbarker for the initial review! I've made all the changes and left you a couple of comments as well. Please let me know if I can contribute to the tests, I realise that a huge amount has changed since that initial PR.

peterbarker:
This page links to our standards around how we like our commit history to look: https://ardupilot.org/dev/docs/git-interactive-rebase.html

Basically a commit-per-directory, prefixed with the logical subsystem name.  So you might have `AP_RCProtocol_IBUS: add support for iBus protocol` and `autotest: add tests for iBus porotocol`

nicole-ashley:
I've added this to `extract_features.py` and `build_options.py`. Happy to add it to `minimize_common.inc` if you're sure.

nicole-ashley:
I've split out the commits. Tools is all in one but I can easily split that further if you'd like.

nicole-ashley:
@peterbarker Would you like me to include your tests in this PR as their own commit, or will you keep them separate?

peterbarker:
> @peterbarker Would you like me to include your tests in this PR as their own commit, or will you keep them separate?

I intend to merge those tests in that other PR - this PR should include a patch to run them as part of the Copter autotest suite.

Please consider this alternate parser: https://github.com/ArduPilot/ardupilot/commit/391ee28f6714118c2c70f0b80ba6aee97e6ef7ae

We didn't do much frame-gapped parsing way back when.

Advantages of that approach are the lack of a state machine, and not keeping state around in the object which is then assumed by the "handle" methods.

Please note I haven't actually tested that patch I reference as there were link errors on the branch I started from.  I've probably screwed up some endian-ness thing in there somewhere (e.g. the checksumming).  Quite happy to get it fully working if you want to change to that style of parsing rather than the state-machine which is currently in this PR.

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,674 @@
...
+#include <AP_BattMonitor/AP_BattMonitor.h>
+#include <AP_Arming/AP_Arming.h>
+#include <AP_GPS/AP_GPS.h>
+#include <AP_RPM/AP_RPM.h>
+
+// All the sensors we can accurately provide are listed here.
+// i-BUS will generally only query up to 15 sensors, so subjectively
+// higher-value sensors are sorted to the top to make the most of a
+// small telemetry window. In the future these could be configurable.
+static const AP_IBus_Telem::SensorDefinition _sensors[] = {
```
peterbarker, line 49:
```suggestion
static const AP_IBus_Telem::SensorDefinition sensors[] {
```

It doesn't matter here where it's a static-const, but omitting the equals can have benefits - not constructing objects on the stack and increased type-checking.  So just as a pattern we're preferring it.

... also removed the leading underscore on the variable name.  We're moving away from prepending underscores to things - so if we could remove them from the object's variables that would be good...

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,674 @@
...
+   To respond to a type query, we send:
+   * 0x06: Message length
+   * 0x91: 0x90 for type + 0x01 for sensor ID 1
+   * 0x03: Sensor type, eg 0x03 for external voltage
+   * 0x02: Sensor length (2 or 4 bytes)
+   * 0x63: Checksum low byte
+   * 0xFF: Checksum high byte
+   Checksums are 0xFFFF minus the sum of the previous bytes
+ */
+void AP_IBus_Telem::handle_type_command(SensorDefinition sensor)
```
peterbarker, line 361:
```suggestion
void AP_IBus_Telem::handle_type_command(const SensorDefinition &sensor)
```

& avoids putting data onto the stack, and `const` is just generally useful for keeping logic correct.

similarly elsewhere

nicole-ashley:
I much prefer that parser! And it's usually how I would write this kind of thing (I don't like instance-level state). I'll rewrite and test that now (will be tomorrow before I come back with it though).

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,674 @@
...
+}
+
+uint16_t AP_IBus_Telem::get_average_cell_voltage_cV()
+{
+    if (!AP::battery().has_cell_voltages()) {
+        return 0;
+    }
+
+    const auto &cell_voltages = AP::battery().get_cell_voltages();
+    const uint8_t number_of_cells = ARRAY_SIZE(cell_voltages.cells);
```
peterbarker, line 561: defensive programming says we should check for zero to avoid the divide-by-zero case below.

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,674 @@
...
+    } else {
+        return 0;
+    }
+}
+
+Vector3f AP_IBus_Telem::get_unbiased_acceleration()
+{
+    return (AP::ahrs().get_accel() - AP::ahrs().get_accel_bias());
+}
+
```
peterbarker, line 618: Maybe just inline this?  Should save a few bytes

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,674 @@
...
+    }
+    return 0;
+}
+
+uint16_t AP_IBus_Telem::get_vehicle_mode()
+{
+    const AP_AHRS::VehicleClass vehicle_class = AP::ahrs().get_vehicle_class();
+
+    switch (vehicle_class) {
+    case AP_AHRS::VehicleClass::GROUND:
```
peterbarker, line 633: I *think* we can make this a compile-time decision instead.

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,674 @@
...
+    SIMPLE       = 7
+    DOCK         = 8
+    CIRCLE       = 9
+    AUTO         = 10
+    RTL          = 11
+    SMART_RTL    = 12
+    GUIDED       = 15
+    INITIALISING = 16
+*/
+static const AP_IBus_Telem::ModeMap _ground_vehicle_mode_map[] = {
```
peterbarker, line 95:
```suggestion
#if APM_BUILD_TYPE(APM_BUILD_ArduPlane)
static const AP_IBus_Telem::ModeMap mode_map[] = {
.
.
};
#elif APM_BUILD_TYPE(APM_BUILD_ArduCopter)
static const AP_IBus_Telem::ModeMap mode_map[] = {
.
.
```

That way only ever have one `mode_map` to look through (avoiding the switch on the AHRS concept of the vehicle)

libraries/AP_IBus_Telem/AP_IBus_Telem.h
```diff
@@ -0,0 +1,171 @@
...
+
+class AP_IBus_Telem
+{
+public:
+    AP_IBus_Telem() {}
+
+    CLASS_NO_COPY(AP_IBus_Telem);
+
+    void init();
+    void listen();
```
peterbarker, line 103: Most of these can be `private`

peterbarker (changes requested):

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
nicole-ashley: I did originally build it this way but I was getting `@Invalid_use_of_APM_BUILD_TYPE` errors. Maybe that'll have changed now it's in the .cpp file? I'll give it a go.

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
nicole-ashley: I originally had no underscores but when I did a style guide check I noticed it said to put them in, so I did 😆 

I'll update!

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
peterbarker: Yep - the defines I mentioned aren't allowed to be used in headers which aren't on a difficult-to-find whitelist.  It's done this way to try to keep the compilation the same for different vehicles, leading to faster build-times.

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
peterbarker: Oh, wow.  Wiki PR coming up for DevCall to fix that, my apologies.

peterbarker: https://github.com/ArduPilot/ardupilot_wiki/pull/6150

nicole-ashley: Wonderful, thank you :)

nicole-ashley:
I've made all the changes and they're compiling locally, but I haven't bench-tested them yet. I'll do that tomorrow morning and may have some more tweaks based on my findings.

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
nicole-ashley: I ended up reverting this (well not this one because it's gone but any remaining `for auto` expressions) because it was breaking SITL builds:
```
error: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Werror=sign-compare]
  311 |     for (auto i = 0; i < sizeof(u.buffer) - 2; i++) {

error: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Werror=sign-compare]
  660 |     for (auto i = 0; i < ARRAY_SIZE(mode_map); i++) {
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,681 @@
...
+        return;
+    }
+
+    // If we've reached the end of our sensor list, we shouldn't respond; this tells the receiver that there
+    // are no more sensors to discover.
+    if (incoming.sensor_id > ARRAY_SIZE(sensors)) {
+        return;
+    }
+
+    const auto *sensor_definition = &sensors[incoming.sensor_id - 1];
```
peterbarker, line 335:
```suggestion
    const auto &sensor_definition = sensors[incoming.sensor_id - 1];
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,681 @@
...
+
+    switch (incoming.command << 4) {
+    case PROTOCOL_COMMAND_DISCOVER:
+        handle_discover_command(incoming);
+        break;
+
+    case PROTOCOL_COMMAND_TYPE:
+        handle_type_command(incoming, *sensor_definition);
```
peterbarker, line 343:
```suggestion
        handle_type_command(incoming, sensor_definition);
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,681 @@
...
+        break;
+
+    case PROTOCOL_COMMAND_VALUE:
+        handle_value_command(incoming, *sensor_definition);
```
peterbarker, line 347:
```suggestion
        handle_value_command(incoming, sensor_definition);
```

libraries/AP_IBus_Telem/AP_IBus_Telem.h
```diff
@@ -0,0 +1,162 @@
...
+#endif
+
+#if AP_IBUS_TELEM_ENABLED
+
+#include <AP_HAL/AP_HAL.h>
+#include <AP_Math/vector3.h>
+#include <AP_RPM/AP_RPM_config.h>
+
+// 2-byte values
+#define IBUS_SENSOR_TYPE_TEMPERATURE          0x01 // Temperature (in 0.1 degrees, where 0=-40'C)
```
peterbarker, line 47: Could move this stuff into the .cpp file - nothing else in ArduPilot needs to know about this.

libraries/AP_IBus_Telem/AP_IBus_Telem.h
```diff
@@ -0,0 +1,162 @@
...
+
+    typedef struct {
+        uint8_t sensor_type;   // Sensor type (IBUS_SENSOR_TYPE_* above)
+        uint8_t sensor_length; // Data length for defined sensor (can be 2 or 4 bytes)
+    } SensorDefinition;
+
+    typedef struct {
+        uint8_t ap_mode;
+        uint8_t ibus_mode;
+    } ModeMap;
```
peterbarker, line 112: make these private

libraries/AP_IBus_Telem/AP_IBus_Telem.h
```diff
@@ -0,0 +1,162 @@
...
+        uint32_t uint32;
+        int16_t int16;
+        int32_t int32;
+        uint8_t byte[4];
+    } SensorValue;
+
+    struct PACKED CommandPacket {
+        uint8_t message_length;
+        uint8_t command : 4;
+        uint8_t sensor_id : 4;
```
peterbarker, line 127:
```suggestion
        uint8_t sensor_id : 4;
        uint8_t command : 4;
```

... it was close!

peterbarker (changes requested):

peterbarker:
Instead of fighting with clang (the cause of the MacOSX compilation failure), move the mode map entirely into the `.cpp` file - don't need namespacing or whatever provided by the header file.

libraries/AP_OSD/AP_OSD_ParamSetting.cpp
```diff
@@ -139,7 +139,7 @@ static const char* SERIAL_PROTOCOL_VALUES[] = {
     "FSKY_TX", "LID360", "", "BEACN", "VOLZ", "SBUS", "ESC_TLM", "DEV_TLM", "OPTFLW", "RBTSRV",
     "NMEA", "WNDVNE", "SLCAN", "RCIN", "MGSQRT", "LTM", "RUNCAM", "HOT_TLM", "SCRIPT", "CRSF",
     "GEN", "WNCH", "MSP", "DJI", "AIRSPD", "ADSB", "AHRS", "AUDIO", "FETTEC", "TORQ",
-    "AIS", "CD_ESC", "MSP_DP", "MAV_HL", "TRAMP", "DDS", "IMUOUT", "IQ", "PPP",
+    "AIS", "CD_ESC", "MSP_DP", "MAV_HL", "TRAMP", "DDS", "IMUOUT", "IQ", "PPP", "IBUS_TLM"
```
andyp1per, line 142: We try and make these 6 chars so that they fit in the various display contexts

andyp1per (changes requested):
One small comment

libraries/AP_OSD/AP_OSD_ParamSetting.cpp
nicole-ashley: Would `IBUSTLM` work? There are a few 7char ones. I feel like the "TLM" is important because not all serial protocols are telemetry.

libraries/AP_IBus_Telem/AP_IBus_Telem.h
nicole-ashley: This could be me lacking C++ knowledge here, but if I make them private I'm then told they're inaccessible when using them as types in the .cpp file (as they're being used outside class context, I expect). I could move them out of the class entirely? Or is there a better way?

nicole-ashley: Possibly that's what you mean here? https://github.com/ArduPilot/ardupilot/pull/27695#issuecomment-2258260241

libraries/AP_OSD/AP_OSD_ParamSetting.cpp
andyp1per: If there are 7 char ones already then ignore me

nicole-ashley: 7 is the longest so far. `IBUS_TLM` is 8 which would be longer. I can make it `IBUSTLM` but that breaks convention with the other telemetry ones which all use an underscore. Happy to do whichever you prefer!

nicole-ashley:
I've made all the changes other than two I've asked questions on. Happy to make them once I understand them. Thank you for all the C++ and embedded help!

Bench testing on the new parser has passed. It's how I probably would have written it if I started from scratch, I think that was just the last holdover from the original PR that I hadn't rewritten.

I've updated the builds folder with the latest code based on stable-4.5.4.

peterbarker (changes requested):

libraries/AP_IBus_Telem/AP_IBus_Telem.h
peterbarker: Never mind - it would require additional patches in the cpp file to change

```
   98 | static const AP_IBus_Telem::SensorDefinition sensors[] {
```
to
```
   98 | static const AP_IBus_Telem::SensorDefinition AP_IBus_Telem::sensors[] {
```

and the addition in the header of `sensors`.

The definitions could also be moved to the `cpp` file, of course, but we're starting to get lost in the weeds a bit.

We'll just forget I mentioned this possibility :-)

peterbarker:
> I've made all the changes other than two I've asked questions on. Happy to make them once I understand them. Thank you for all the C++ and embedded help!

All good :-)  Really good to get the rapid turn-arounds - that was a lot of churn in a short amount of time, and I know that multiple round-trips with large amounts of requested changes can be frustrating.

> Bench testing on the new parser has passed. It's how I probably would have written it if I started from scratch, I think that was just the last holdover from the original PR that I hadn't rewritten.

:-)

I think @andyp1per had worked on some infrastructure to make writing frame-gapped packet parsing even easier; some sort of an API on the UARTDriver.  I *believe* that one's still a WIP.

> I've updated the builds folder with the latest code based on stable-4.5.4.

OK, this is a bit of a strange comment.  AFAICS this is all based on master.  We only merge features into master, never a stable branch, 'though some get backported after they're in master.  Usually the "builds" folder is a transient directory used by the build system rather than something you'd actively update, too.  So I'm a touch confused :-)

nicole-ashley:
And now I can report that about 30 minutes of real-world flight testing has also worked perfectly with no noticable issues or side-effects.

@peterbarker Possibly the build comment is misleading, and it's not important for the purposes of the code review. All the code in this PR is based purely on `master`.

In the opening description I provided a folder with some builds for bench testing. They're this feature based on stable so that any issues noticed are definitely to do with only this feature and not other unreleased changes. This is a holdover from when I was resurrecting the code from the old PR, in case anyone who didn't have a full build environment could help me test it. Possibly there's a more official way to do this but as I didn't have push access to the old PR that was a way to offer up testable binaries. Now that there's been a large functional rewrite (the parser) I updated the builds so that the newest way could easily be tested.

libraries/AP_IBus_Telem/AP_IBus_Telem.h
nicole-ashley: This'll be burning in my mind now so I'm sure I'll come back to it for a future iteration 🙂

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,730 @@
...
+        value.uint16 = AP::arming().is_armed();
+        break;
+
+    case IBUS_SENSOR_TYPE_FLIGHT_MODE:
+        value.uint16 = get_vehicle_mode();
+        break;
+
+    case IBUS_SENSOR_TYPE_SPEED:
+        value.uint16 = AP::airspeed()->get_airspeed();
+        break;
```
peterbarker, line 601:
```suggestion
#if AP_AIRSPEED_ENABLE
        case IBUS_SENSOR_TYPE_SPEED:
        value.uint16 = AP::airspeed()->get_airspeed();
        break;
#endif
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,730 @@
...
+        value.uint16 = AP::gps().ground_speed_cm();
+        break;
+
+    case IBUS_SENSOR_TYPE_GPS_DIST:
+        value.uint16 = get_distance_from_home_m();
+        break;
+
+    case IBUS_SENSOR_TYPE_ARMED:
+        value.uint16 = AP::arming().is_armed();
+        break;
```
peterbarker, line 593:
```suggestion
#if AP_ARMING_ENABLED
    case IBUS_SENSOR_TYPE_ARMED:
        value.uint16 = AP::arming().is_armed();
        break;
#endif
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,730 @@
...
+        value.int16 = AP::ahrs().roll_sensor;
+        break;
+
+    case IBUS_SENSOR_TYPE_PITCH:
+        value.int16 = AP::ahrs().pitch_sensor;
+        break;
+
+    case IBUS_SENSOR_TYPE_YAW:
+        value.int16 = AP::ahrs().yaw_sensor;
+        break;
```
peterbarker, line 581:
```suggestion
        break;
#endif  // AP_AHRS_ENABLED
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,730 @@
...
+    case IBUS_SENSOR_TYPE_COG:
+        value.uint16 = AP::gps().ground_course() * 100;
+        break;
+
+    case IBUS_SENSOR_TYPE_GPS_STATUS:
+        value.byte[0] = get_gps_status();
+        value.byte[1] = AP::gps().num_sats();
+        break;
+
+    case IBUS_SENSOR_TYPE_ACC_X:
```
peterbarker, line 559:
```suggestion
#if AP_AHRS_ENABLED
    case IBUS_SENSOR_TYPE_ACC_X:
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,730 @@
...
+    case IBUS_SENSOR_TYPE_COMPASS_HEADING:
+        value.uint16 = AP::ahrs().yaw_sensor * 0.01;
+        break;
+
+    case IBUS_SENSOR_TYPE_CLIMB_RATE:
+    case IBUS_SENSOR_TYPE_VERTICAL_SPEED:
+        value.int16 = AP::baro().get_climb_rate() * 100;
+        break;
+
+    case IBUS_SENSOR_TYPE_COG:
```
peterbarker, line 550:
```suggestion
#if AP_GPS_ENABLED
    case IBUS_SENSOR_TYPE_COG:
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,730 @@
...
+    case IBUS_SENSOR_TYPE_RPM:
+    case IBUS_SENSOR_TYPE_RPM_FLYSKY:
+        value.uint16 = get_rpm();
+        break;
+
+    case IBUS_SENSOR_TYPE_COMPASS_HEADING:
+        value.uint16 = AP::ahrs().yaw_sensor * 0.01;
+        break;
+
+    case IBUS_SENSOR_TYPE_CLIMB_RATE:
```
peterbarker, line 545:
```suggestion
#if AP_BARO_ENABLED
    case IBUS_SENSOR_TYPE_CLIMB_RATE:
```

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,730 @@
...
+    {.sensor_type = IBUS_SENSOR_TYPE_GROUND_SPEED, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_ROLL, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_PITCH, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_YAW, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_SPEED, .sensor_length = 2},
+    {.sensor_type = IBUS_SENSOR_TYPE_TEMPERATURE_PRESSURE, .sensor_length = 4},
+#if AP_RPM_ENABLED == 1
+    {.sensor_type = IBUS_SENSOR_TYPE_RPM, .sensor_length = 2},
+#endif
+    {.sensor_type = IBUS_SENSOR_TYPE_BATTERY_CURRENT, .sensor_length = 2},
```
peterbarker, line 116:
```suggestion
#if AP_BATTERY_ENABLED
    {.sensor_type = IBUS_SENSOR_TYPE_BATTERY_CURRENT, .sensor_length = 2},
```

etc

libraries/AP_IBus_Telem/AP_IBus_Telem.h
```diff
@@ -0,0 +1,113 @@
...
+   * IBusBM
+     https://github.com/bmellink/IBusBM
+
+   * BetaFlight
+     https://github.com/betaflight/betaflight/blob/master/src/main/telemetry/ibus_shared.c
+ */
+
+#pragma once
+
+#ifndef AP_IBUS_TELEM_ENABLED
```
peterbarker, line 36:
```suggestion
#include <AP_HAL/AP_HAL_Boards.h>

#ifndef AP_IBUS_TELEM_ENABLED
```

will resolve the compile failure with redefinition of this define

peterbarker (changes requested):
Need to be pretty liberal with cutting out code with the `#if`s.  We have a script to test the build options (`Tools/autotest/test_build_options.py`) which will catch them all, but it takes a long time to run.  I can do that either on the next iteration of this PR or after the thing is merged (i.e. fix them with a follow-up PR.  The concept is that code should compile with every option in `build_options.py` turned off.

A run with `./Tools/autotest/test_build_options.py --no-enable-in-turn --no-disable-in-turn` might save a few round-trips to github trying to get CI to pass.

nicole-ashley:
OK! I've made all the changes and also `test_build_options.py --no-enable-in-turn --no-disable-in-turn` has passed locally. So fingers crossed! 🤞

nicole-ashley:
The last unaddressed comments are:

@peterbarker I haven't added this to `minimize_common.inc` yet. It'd be sad not to have it in builds automatically but I totally understand why and knowing the build server exists is good. Presumably it will automatically be in builds for boards that aren't so tight on memory? Let me know if you want me to add it to `minimize_common.inc`.

@andyp1per I've left a comment regarding the OSD name; totally happy to take your call.

peterbarker:
Closes https://github.com/ArduPilot/ardupilot/issues/9419

tridge:
@nicole-ashley very nicely done PR!
My initial reaction is that this should be enabled only via the custom build server, as we are already very low on flash. See https://firmware.ardupilot.org/Tools/BuildSizes/builds.html
What is the typical flight controller used by the community that uses i-Bus? We could enable it on some of the 2M flash boards, but that only makes sense if that overlaps with the likely user community that use i-Bus.
we should also enable it always on SITL so it gets built and doesn't bit rot

libraries/AP_IBus_Telem/AP_IBus_Telem.cpp
```diff
@@ -0,0 +1,796 @@
...
+    default:
+        value.int32 = 0;
+    }
+
+    return value;
+}
+
+uint16_t AP_IBus_Telem::get_average_cell_voltage_cV()
+{
+#if AP_BATTERY_ENABLED
```
peterbarker, line 685:
```suggestion
#if AP_BATTERY_ENABLED
uint16_t AP_IBus_Telem::get_average_cell_voltage_cV()
{
```

... and remove the call

Similarly elsewhere

peterbarker (changes requested):

peterbarker:
This is probably the way to limit the boards this is built on: https://github.com/ArduPilot/ardupilot/pull/27717

nicole-ashley:
@peterbarker The 2048 suggestion sounds good to me - we get it automatically in the bigger boards and it can still be selected for custom builds.

Thanks @tridge! I'm not actually sure what a common FC for i-BUS users is. I can only speak from my own experience: i-BUS is more commonly in the entry-level TX/RX range, and as I was entry-level myself and didn't have A) lots of money for a FC but also B) enough knowledge to make a call on a very specific or limited design. Therefore I went for something cost effective while still being generic, plug-and-play, and enough memory to know I wouldn't regret it: that lead me to one of the classic Pixhawks running hde fmuv3 build.

The old PR also had interest from users of Pixracer, MatekF405, and SpeedyBeeF405Wing boards.

peterbarker:
> @peterbarker I suspect the >=2048 limit will probably rule out most of the boards that are commonly used by owners of i-BUS hardware, meaning they'll need to do a custom build (which is also not an entry-level step). It would be great to get it under 2048 by default. Completely your call though.

Definitely not my call on the 2048 thing - that's tridge as technical lead (the poor bugger).

We did agonize over it a chunk on DevCallEU tonight, musing that it would be great to have a way of working out what to put in the builds which was based on numbers not just gut feel.  We've also mused on integration of the custom build server functionality with MissionPlanner to make the process of adding support for these things more straight-forward.

As tridge mentioned  If we put this on 1MB boards today then we'd instantly break several boards.  If this goes in there by default then something has to go out - and because those things have been in we run the risk of breaking people's working setups, which we don't like doing.

nicole-ashley:
@peterbarker @tridge Totally understand with the 1MB boards. It would be great however to see it by default on the 2MB boards (which if I understand correctly would also be excluded by the `>2048` rule). 

However I'm just pleading my case and will absolutely take your (well, @tridge's) call. The important thing is that it's in the core codebase, available for people to use, and even more importantly that it's in the documentation with a reference on how to get it on your board if you want it. Discovery and availability is key even if it's not there for a lot of people by default. I totally get that it is, after all, an edge use case.

nicole-ashley:
Of the failing tests:

* sitltest-plane has missing logs but that build is working for me locally (maybe a runner issue?)
* sitltest-copter-tests1d seems to be a timeout possibly from an overloaded runner
* sitltest-copter-tests2b is failing due to the missing i-BUS tests

peterbarker:
Please cherry-pick https://github.com/ArduPilot/ardupilot/pull/27717/commits/5bcfd941125470be986b2b28c4d21d49b0634732 to reinstate the feature in SITL / linux / STM32-boards-with-external-flash.

peterbarker (approved):
LGTM

peterbarker:
Merged, thanks!

rmackay9:
Hi @nicole-ashley @peterbarker, @andyp1per,

We've got a user [here in the 4.6 beta testing forums](https://discuss.ardupilot.org/t/copter-4-6-0-beta2-is-available-for-beta-testing/127445/11) strugglingto use this feature.  I wonder if you could help?

Also it looks like we missed adding the new item to the SERIALx_PROTOCOL parameter description

nicole-ashley:
@rmackay9 I'm travelling at the moment with intermittent internet access. Hard to tell what's wrong on the face of it but I've suggested some things to check.

I do have my computer so I can make the missing change. A quick pointer about where to do it would be helpful though, or I can spend more time on it when I get home in a few weeks.

rmackay9:
Hi @nicole-ashley, Thanks very much for that!  Hopefully the user will figure it out or maybe we can find someone else to test it and confirm it works or not

nicole-ashley:
When I'm back home I can test the new beta as well to make sure everything is still working as expected.

PR #26832, opened 2024-04-17
ESC Telem Pre ARM check added

loki077:
The goal is to have ESC telem Pre Arm check.

PR #21499, opened 2022-08-20
Copter: takeoff check that motors are spinning at TKOFF_RPM_MIN 

andyp1per:
Alternative implementation of https://github.com/ArduPilot/ardupilot/pull/21223

PR #27696, opened 2024-07-30
Implement ESC Minimum RPM Check Before Takeoff

loki077:
This PR introduces a code implementation to perform an ESC minimum RPM check before takeoff. The purpose of this check is to ensure that motors are properly spooling before takeoff, leveraging ESC telemetry when available. This aims to prevent flights with undetected motor failures.

Initially, I considered integrating this check with a Pre-ARM check (#26832), but it was highlighted that a similar method already exists in Copter (#21499).

With assistance from @peterbarker found a different method to generalise the check into AP_motors. 

This code is still a work in progress and needs further discussion during dev call if this is the right approach.


loki077:
@Pradeep-Carbonix @robertlong13 FYI

peterbarker:
@WickedShell and @magicrub have had ideas on this sort of thing before.

... and it's @andyp1per 's code we're co-opting

rmackay9:
thanks for this, I'm not sure that architecturally we should be putting this down in the motors library.  This check depends on other subsystems so I think it naturally belongs higher

ArduPlane/quadplane.cpp
```diff
@@ -1926,6 +1941,10 @@ void QuadPlane::update_throttle_hover()
  */
 void QuadPlane::motors_output(bool run_rate_controller)
 {
+#if HAL_WITH_ESC_TELEM
+    motors->takeoff_check(takeoff_rpm_min, takeoff_rpm_max, false);
```
tridge, line 1945: the land_complete false is a problem, and means this doesn't work at all.
We need this to only impact automatic takeoff, not manual takeoff or a back transition. If we lose RPM monitoring on a motor we must not fail to spool up on back transition

tridge (changes requested):
maybe this common code should go in AP_Vehicle?

ArduPlane/quadplane.cpp
tridge: can use in_vtol_takeoff()

ArduPlane/quadplane.h
```diff
@@ -612,6 +612,11 @@ class QuadPlane
     // AHRS alt for land abort and package place, meters
     float land_descend_start_alt;
 
+    // support for blocking takeoff until RPMs reach parameter-defined
+    // minimum levels:
+    AP_Float takeoff_rpm_min;
```
tridge, line 617: these are AP_Int16 in copter, we should make them the same type in plane

tridge, line 617: needs #if HAL_WITH_ESC_TELEM around these

ArduPlane/quadplane.cpp
```diff
@@ -1926,6 +1941,11 @@ void QuadPlane::update_throttle_hover()
  */
 void QuadPlane::motors_output(bool run_rate_controller)
 {
+#if HAL_WITH_ESC_TELEM
```
tridge, line 1944: I'd much rather have this in QuadPlane::takeoff_controller(), and operate just like the existing check for tiltrotor.fully_up()
I don't think it should apply to manually piloted takeoffs
I don't think we should be calling set_spoolup_block() at all in quadplanes, we can just delay takeoff, and print a message as to why it is delayed

tridge (changes requested):

PR #27698, opened 2024-07-30
AP_DroneCAN: DNA_Server: fix handling of empty entry

tpwrules:
Replaces the check for a CRC of 0 with a check that the hwid is 0. Substantially reduces 1/256 chance that a particular node ID couldn't be stored.

The core locking fix will not be back-port-able so I have decided to get this done now. The storage area is already protected by semaphores in the StorageManager and HAL, the mis-use of the lock in the DNA server is not a few-line fix, and nobody has complained so far.

Tested on plane on CubeOrange using a selection of nodes and GPSes that the DNA server is preserved on upgrade and correctly registers and remembers nodes after the upgrade.

### DIFF
--- a/libraries/AP_DroneCAN/AP_DroneCAN_DNA_Server.cpp
+++ b/libraries/AP_DroneCAN/AP_DroneCAN_DNA_Server.cpp
@@ -202,8 +202,10 @@ bool AP_DroneCAN_DNA_Server::isValidNodeDataAvailable(uint8_t node_id)
 {
     NodeData node_data;
     readNodeData(node_data, node_id);
+
+    uint8_t empty_hwid[sizeof(NodeData::hwid_hash)] {};
     uint8_t crc = crc_crc8(node_data.hwid_hash, sizeof(node_data.hwid_hash));
-    if (crc == node_data.crc && node_data.crc != 0) {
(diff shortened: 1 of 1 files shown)


libraries/AP_DroneCAN/AP_DroneCAN_DNA_Server.cpp
```diff
@@ -202,8 +202,10 @@ bool AP_DroneCAN_DNA_Server::isValidNodeDataAvailable(uint8_t node_id)
 {
     NodeData node_data;
     readNodeData(node_data, node_id);
+
+    uint8_t empty_hwid[sizeof(NodeData::hwid_hash)] = {0};
```
tridge, line 206: i'm happy with this, but I suspect @peterbarker will insist on removing the = and the 0 in the init, you can do this:

```
  uint8_t xxx[10] {};
```

it generates slightly smaller code

tridge (changes requested):
thanks for the backport!

libraries/AP_DroneCAN/AP_DroneCAN_DNA_Server.cpp
tpwrules: It generated identical code for me, but it's probably good to keep the habit so I changed it. I think it matters more for static variables?

tridge (approved):

rmackay9:

PR #27699, opened 2024-07-30
AP_Scripting: Change serial depends HAL_GCS to AP_SERIALMANAGER

haydendonald:
Serial bindings were dependant on HAL_GCS_ENABLED but this is not ideal, it should be dependant on AP_SERIALMANAGER_ENABLED so that serial can function without GCS.

Tested on: CubePilot-PPPGW running AP_Periph, CubeOrangePlus running Plane


peterbarker (approved):
LGTM

IamPete1:
Can you double check scripting still builds on periph please.

PR #10899, opened 2019-03-26
GCS_MAVLink: Pass external Statustext messages to FrSky passthrough

juichi:
These changes will allow externally generated Statustext MAVLink messages to get passed to FrSky passthrough telemetry. 

In this revision to the code, the messages have the sysid and compid appended.  I think this is okay but it is something that may be changed.  Also, this is a very simple change which keeps the messages at the info level.  I am still working to get the skills up to be able to decode the message's severity from the message itself.  

Also, first pull request ever.  Excited.

PR #27706, opened 2024-07-30
add send statustext to osd for sysid 10

Elaugaste:
I propose implementing a feature that allows statustext messages to be sent to the OSD. This would be beneficial for those utilizing a Companion Computer.

https://github.com/ArduPilot/ardupilot/pull/10899

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -3596,7 +3596,14 @@ void GCS_MAVLINK::handle_statustext(const mavlink_message_t &msg) const
     char text[text_len] = { 'G','C','S',':'};
     uint8_t offset = strlen(text);
 
-    if (msg.sysid != sysid_my_gcs()) {
+    if (msg.sysid == 10) {
+                    offset = hal.util->snprintf(text,
+                                    max_prefix_len,
+                                    "CC:");
+        offset = MIN(offset, max_prefix_len);
+    }
+
+    if (msg.sysid != sysid_my_gcs() && msg.sysid != 10) {
         offset = hal.util->snprintf(text,
                                     max_prefix_len,
                                     "SRC=%u/%u:",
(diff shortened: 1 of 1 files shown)


libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3606,7 +3613,7 @@ void GCS_MAVLINK::handle_statustext(const mavlink_message_t &msg) const
     }
 
     memcpy(&text[offset], packet.text, MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN);
-
+    send_text(MAV_SEVERITY_INFO, "%s", text);
```
peterbarker, line 3616: Instant traffic reflection attack - if you have two ArduPilot instances that can see each other messages will never leave the system except via dropped messages or resource starvation.  The fun really starts if you have thee instances.

peterbarker (changes requested):
I understand the desire for this functionality, but we are *not* going to special-case the number 10 here :-)

Elaugaste:
> I understand the desire for this functionality, but we are _not_ going to special-case the number 10 here :-)

i use sysid 10 because its "default" sysid  for project mavp2p. :)
I was afraid that the changes might break something, so I made a restriction.

I think it's worth making a small range for the CC prefix.

libraries/GCS_MAVLink/GCS_Common.cpp
Elaugaste: > Instant traffic reflection attack - if you have two ArduPilot instances that can see each other messages will never leave the system except via dropped messages or resource starvation. The fun really starts if you have thee instances.

Is this true even if different ardupilot instances have different SYSID_THISMAV?

Elaugaste:
* changed CC sysid to 254
* add send_text2sysid() for sending msg only for CC (this must resolve "Instant traffic reflection attack")

PR #27709, opened 2024-07-30
Rename PID setter methods to include set_

peterbarker:
There are very few places in ArduPilot where we don't have "set_" for setters.

PIDs are one of them - make them uniform.

This is a no-compiler-output change on the smaller set of builds I run.  I will be running the full set of builds.

### DIFF
--- a/ArduCopter/tuning.cpp
+++ b/ArduCopter/tuning.cpp
@@ -37,70 +37,70 @@ void Copter::tuning()
 
     // Roll, Pitch tuning
     case TUNING_STABILIZE_ROLL_PITCH_KP:
-        attitude_control->get_angle_roll_p().kP(tuning_value);
-        attitude_control->get_angle_pitch_p().kP(tuning_value);
+        attitude_control->get_angle_roll_p().set_kP(tuning_value);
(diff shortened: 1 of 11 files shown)


andyp1per (approved):
I prefer this, but really @lthall and @bnsgeyer get the final say

lthall:
I think those names have been a part of the project longer than I have!!

Does anybody know why we used the "k" in front of the P, I and D?

IamPete1 (approved):

rmackay9:
These are even before my time, we'd need to ask Jason Short.

peterbarker:
> Does anybody know why we used the "k" in front of the P, I and D?

`k` is for `konstant` AFAIK :-)  So to help distinguish between the output component and the gain, I think?

We can remove them too as part of this PR, if you like - what do you think?

peterbarker:
... removing the `k` would make it consistent with the "newer" entries in the API - `ff` and the filter rates.

rmackay9:
@peterbarker, thanks.  Making it clear whether the caller is getting the gain or the output seems important :-).  So either in the name or the comments above the function.  I guess we could make it set_P_gain() or set_gain_P().

lthall:
This all sounds good to me!

PR #27711, opened 2024-07-30
AP_NavEKF3: Corrected logged test ratios

Georacer:
I noticed that the description of the logged EKF3 test ratios `XKF4/[SV,SP,SH,SM,SVT]` is described to have a nominal threshold of 0-1.
But in practice the values had a nominal range of 0-100, as they are premultiplied by 100.

Additionally, they are guaranteed to be positive, as they are the result of a `sqrtF()`, returned by `NavEKF3_core::getVariances()`.

Sidenote: I believe most graphing tools take the unit multipliers into account, so perhaps we can omit the description change.

Before (somehow MAVExplorer knows the correct scaling):
image
image

After:
image
image


libraries/AP_NavEKF3/LogStructure.h
```diff
@@ -437,7 +437,7 @@ struct PACKED log_XKV {
     { LOG_XKF3_MSG, sizeof(log_XKF3), \
       "XKF3","QBcccccchhhccff","TimeUS,C,IVN,IVE,IVD,IPN,IPE,IPD,IMX,IMY,IMZ,IYAW,IVT,RErr,ErSc", "s#nnnmmmGGGd?--", "F-BBBBBBCCCBB00" , true }, \
     { LOG_XKF4_MSG, sizeof(log_XKF4), \
-      "XKF4","QBcccccfffHBIHb","TimeUS,C,SV,SP,SH,SM,SVT,errRP,OFN,OFE,FS,TS,SS,GPS,PI", "s#------mm-----", "F-------??-----" , true }, \
+      "XKF4","QBcccccfffHBIHb","TimeUS,C,SV,SP,SH,SM,SVT,errRP,OFN,OFE,FS,TS,SS,GPS,PI", "s#------mm-----", "F-BBBBB-??-----" , true }, \
```
tridge, line 440: need to remove the c, should be H

tridge (changes requested):

libraries/AP_NavEKF3/LogStructure.h
```diff
@@ -173,7 +173,7 @@ struct PACKED log_XKF3 {
...
+// @Description: EKF3 variances.  SV, SP, SH and SM are probably best described as 'Squared Innovation Test Ratios' where values <100 tells us the measurement was accepted and >100 tells us it was rejected. They represent the square of the (innovation / maximum allowed innovation) where the innovation is the difference between predicted and measured value and the maximum allowed innovation is determined from the uncertainty of the measurement, uncertainty of the prediction and scaled using the number of standard deviations set by the innovation gate parameter for that measurement, eg EK3_MAG_I_GATE, EK3_HGT_I_GATE, etc
```
tridge, line 176: should be < 1, revert this change

tridge (changes requested):

tridge (approved):

PR #27719, opened 2024-07-31
Feature/vehicle compile

khancyr:
fix compile


peterbarker (approved):

peterbarker:
Merged, thanks!

This would have been caught if AP_EXTERNAL_CONTROL_ENABLED was in `build_options.py` - I wonder why it isn't...

khancyr:
Probably missed as that is a new things use only with ros I think

Issue #6378, opened 2017-06-05
Picture trigger should be decoupled from GPS updates


rmackay9:
This is an edge case but at the moment "do_take_picture" is triggered from the "Copter::update_GPS" function but it's theoretically possible that we are using a different position estimate source beside the GPS (i.e. visual odometry).

OXINARF:
@rmackay9 I've updated the title as this affects all vehicles. Feel free to change it again if you don't find it clear enough.

7princekumar:
I would like to work on it but have no clue . I am new to open source contribution. I need your help to solve this.

WickedShell:
Done in #14804

rmackay9:
Except that this line still exists meaning the camera trigger still won't work without a GPS. https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Camera/AP_Camera.cpp#L344

WickedShell:
Hmm, yeah okay, I yield there, that should probably be addressed as well :)

PR #27720, opened 2024-07-31
AP_Camera: rely on AHRS for position, not GPS

peterbarker:
our location may not be coming from a GPS

Closes https://github.com/ArduPilot/ardupilot/issues/6378

### DIFF
--- a/libraries/AP_Camera/AP_Camera_Backend.cpp
+++ b/libraries/AP_Camera/AP_Camera_Backend.cpp
@@ -1,9 +1,13 @@
-#include "AP_Camera_Backend.h"
+#include "AP_Camera_config.h"
 
 #if AP_CAMERA_ENABLED
+
+#include "AP_Camera_Backend.h"
+
 #include <GCS_MAVLink/GCS.h>
 #include <AP_GPS/AP_GPS.h>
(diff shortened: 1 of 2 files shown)


libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -57,8 +61,11 @@ void AP_Camera_Backend::update()
         return;
     }
 
-    // check GPS status
-    if (AP::gps().status() < AP_GPS::GPS_OK_FIX_3D) {
+    Location current_loc;
+    const AP_AHRS &ahrs = AP::ahrs();
+    if (!ahrs.get_location(current_loc)) {
+        last_location.lat = 0;
```
rmackay9, line 67: I love to see this change but I wonder why we need to set last_location here

peterbarker: I copied that pattern from a clause above.

Which was a mistake; I've removed it.

GPS is still used in here for a timestamp.  The date should probably come from `AP::rtc()`, but we probably also need to have an option bit or something...

rmackay9 (approved):
Super happy to see this, thanks!

PR #27724, opened 2024-08-01
BMP390 driver update, add to Navigator

Williangalvani:
I had to do these changes in order to get the BMP390 working on a new revision of the navigator.
needs testing on other platforms.

Changes are based on [Adafruit's library](https://github.com/adafruit/Adafruit_BMP3XX)

could it be an i2c issue? that would explain why no one else has seem the issue...


libraries/AP_Baro/AP_Baro_BMP388.cpp
```diff
@@ -107,6 +109,9 @@ bool AP_Baro_BMP388::init()
         return false;
     }
 
+    dev->write_register(BMP388_REG_CMD, BMP388_SOFT_RESET, true);
```
tridge, line 112: the true is wrong, this is not a checked register, this will reset your baro every 20 loops

tridge (dismissed):

libraries/AP_Baro/AP_Baro_BMP388.cpp
Williangalvani: I removed this line, so now the PR only adds the new sensor to the navigator.
I tested it here and it worked fine on my board without the soft reset. If the issue comes up again, I'll open a new PR for it.

PR #25139, opened 2023-09-28
AP_ICEngine: Fixup the docs on ICE_STARTCHN_MIN to match the behaviour

WickedShell:
The docs on this were incorrect to the current behavior. This documents the behavior as it is currently implemented.

PR #27730, opened 2024-08-01
AP_ICEngine: stop engine in E-Stop

IamPete1:
We already force throttle to 0 in E-Stop, so the engine is unlikely to be idling. The key thing is to also stop the ignition and prevent the starter from turning over the engine. Just running the starter can be quite dangerous even if the engine is not going to start it gets the motor going fast and gives no warning. 

I have also added to the `STARTCHN_MIN` description to better describe what it does.


IamPete1:
The `STARTCHN_MIN` description fixup is also in https://github.com/ArduPilot/ardupilot/pull/25139

tridge (approved):

PR #21865, opened 2022-10-03
Add ARKV6X

AlexKlimaj:
This PR adds support for the ARKV6X flight controller. It is based on the Pixhawk Autopilot Bus form factor and FMUV6X pinout. There are differences in sensors with the FMUV6X, enough to warrant its own hwdef. There is also a timer connected to all the IMU's to provide an external clock.

https://arkelectron.com/product/arkv6x/

Current Testing Status
- [X] Builds
- [X] Communicates with QGC daily
- [X] Flashes through QGC
- [X] All sensors present
- [X] IMU orientations correct
- [x] PWM outputs
- [x] RC input
- [x] IO MCU communication
- [x] Digital power module communication - Need to set BATT_I2C_BUS to 1 or 2. Are both supported at the same time?
- [x] Telem ports
- [x] GPS ports
- [x] Buzzer
- [x] CAN ports
- [x] I2C ports
- [x] Flight tested

[00000012.zip](https://github.com/ArduPilot/ardupilot/files/9803837/00000012.zip)
[00000013.zip](https://github.com/ArduPilot/ardupilot/files/9803899/00000013.zip)
[00000015.zip](https://github.com/ArduPilot/ardupilot/files/9804075/00000015.zip)

PR #27736, opened 2024-08-02
ARKV6X RC IN USART6 4.5 Backport

AlexKlimaj:
Backport of @dakejahl work in master to fix SBUS input on the RC IN port.


libraries/AP_HAL_ChibiOS/hwdef/ARKV6X/hwdef.dat
```diff
@@ -81,7 +85,8 @@ PD9 USART3_RX USART3
 
 # USART6 is for IOMCU
 PC6 USART6_TX USART6
-PC7 USART6_RX USART6
+# used for RC SBUS
+#PC7 USART6_RX USART6
```
tridge, line 89: happy to bring this in, but I'm curious why SBUS didn't work on this UART? Did the user set SERIAL8_PROTOCOL to RCIN?

peterbarker: Ping @dakejahl - unanswered question here.

dakejahl: The conversation is in a now deleted slack workspace but I think ultimately the issue ended up being the user had bad hardware. I can test again to be sure. Are there concerns with this change?

peterbarker: > The conversation is in a now deleted slack workspace but I think ultimately the issue ended up being the user had bad hardware. I can test again to be sure. Are there concerns with this change?

I think tridge was just curious.  He'll see your response as part of the backport process.

dakejahl: I'm not as familiar with ardupilot so if something doesn't look right let me know. Thanks for the review

tridge:
this does seem a strange change, what testing has been done?

dakejahl:
I tested a few months ago and it was working. I can't remember everything I did and I can't seem to figure it out from the documentation. SBUS is not working at all for me using 4.5.5 or master (with or without this PR)

SERIAL8_PROTOCOL set to RCIN
RC_PROTOCOL set to 1 (all)

Am I missing something?


I even went back and tried 4.3.3 since this was when the board support was first introduced and SBUS worked back then (I checked the logs alex shared) but it's still not working.
https://github.com/ArduPilot/ardupilot/pull/21865

I am testing on a variant of this board with an IIM42653 IMU which does not yet have support in ardupilot, so I am getting a constant error about "failed to initialize INS", could this be interfering with the SBUS setup somehow?

I know the hardware is working because when I flash to PX4 it works.

peterbarker:
Closing per directions in #28015

rmackay9:
I've removed the 4.5 backport project because it doesn't look like we want to do that

PR #27740, opened 2024-08-03
SRV_Channel: added AM32 to the SERVO_DSHOT_ESC param docs

tridge:
setting 3 works well with AM32


tridge:
@andyp1per what value should users use for BlueJay?

andyp1per:
@tridge BlueJay should be 2 and 4. I think AM32 should be 1 and 3 - the timing is the same for both, its just that 1 does not request EDT information.

tridge:
@andyp1per thanks, updated PR. I had a user ask me about BlueJay setting today

andyp1per (approved):
Note that there is an argument that says if there are no versions of AM32 that don't support EDT then we should always recommend 3, but what you have here is undoubtedly correct.

Issue #24245, opened 2023-07-07
SITL: dumpstack/dumpcore do not run if PWD is not in the repository and sh != bash


durka:
## Bug report

**Issue details**

There is a comment here: https://github.com/ArduPilot/ardupilot/blob/a480c0a74a3edc7e9905bfd9821dfa973d416077/libraries/AP_HAL_SITL/system.cpp#L58

Unfortunately, trusting in `$PATH` is a hazardous course of action! This function uses `sh` to invoke the program. On my system (Ubuntu 18.04), `sh` is symlinked to `dash`, which does NOT search `$PATH` when given a command that doesn't include a slash. 

The man page makes it sound like it does, but it doesn't.

>     Path Search
>      When locating a command, the shell first looks to see if it has a shell function by that name.  Then it looks for a builtin command by that name.  If a builtin command is not found, one of two things happen:
>      1.   Command names containing a slash are simply executed without performing any searches.
>      2.   The shell searches each entry in PATH in turn for the command.  The value of the PATH variable should be a series of entries separated by colons.  Each entry consists of a directory name.  The current directory may be indicated implicitly by an empty directory name, or explicitly by a single period.

```
$ mkdir test
$ echo -e '#!/bin/sh\necho $0 "$@"\n' >test/test.sh
$ PATH=$PWD/test:$PATH bash test.sh 1 2 3
test.sh 1 2 3
$ PATH=$PWD/test:$PATH dash test.sh 1 2 3
dash: 0: cannot open test.sh: No such file
```

Here are a couple of workarounds:
1. Use `bash` instead of `sh`
1. Use `sh -c`, but note that in that case the full command needs to be in quotes, i.e. `sh -c "dumpstack.sh PID"` or `sh -c "dumpstack.sh $0" PID`
2. If the command isn't found in one of the canned search directories, find it from `$PATH` using `which`, then pass *that* to `sh`

**Version**
4.2.3

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine
[x] SITL

**Airframe type**
N/A

**Hardware type**
N/A

**Logs**
N/A

peterbarker:
We can remove any hack required for this when this sequence of commands results in something sensible:

```
pbarker@fx:~/rc/ardupilot(master)$ ls ~/bin/bin2elf.sh 
/home/pbarker/bin/bin2elf.sh
[...]

PR #27742, opened 2024-08-03
AP_HAL_SITL: work around bug in dash

peterbarker:
Closes https://github.com/ArduPilot/ardupilot/issues/24245

### DIFF
--- a/libraries/AP_HAL_SITL/system.cpp
+++ b/libraries/AP_HAL_SITL/system.cpp
@@ -65,13 +65,24 @@ static void run_command_on_ownpid(const char *commandname)
     // find dumpstack command:
     const char *command_filepath = commandname; // if we can't find it trust in PATH
     struct stat statbuf;
+    const char *custom_scripts_dir_path = getenv("AP_SCRIPTS_DIR_PATH");
+    char *custom_scripts_dir_path_pattern = nullptr;
+    if (custom_scripts_dir_path != nullptr) {
+        if (asprintf(&custom_scripts_dir_path_pattern, "%s/%%s", custom_scripts_dir_path) == -1) {
+            custom_scripts_dir_path_pattern = nullptr;
+        }
+    }
     const char *paths[] {
+        custom_scripts_dir_path_pattern,
         "Tools/scripts/%s",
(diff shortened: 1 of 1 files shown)


libraries/AP_HAL_SITL/system.cpp
```diff
@@ -96,6 +96,13 @@ static void run_command_on_ownpid(const char *commandname)
         p = progname;
     }
 
+    // work around a bug in dash; its manpage says it searches $PATH
+    // but it does't.  https://github.com/ArduPilot/ardupilot/issues/24245
+    const char *cmd_fmt = "sh %s %d >%s 2>&1";
+    if (getenv("AP_DASH_HACKERY")) {
```
tridge, line 102: how does this env variable get set?

peterbarker: By the user.

They already need to be running SITL (or autotest) with their cwd set to a place we're not expecting, and to have modified their $PATH to include `scripts`, so I figured an additional environment variable would be OK.

This patch was designed as least-risk-to-existing-users.

durka: I'm thinking it might be easier to understand (for the users like me w/ weird setups) if the env var was something more like AP_OVERRIDE_PATH_TO_SCRIPTS, to just specify the full path to where dumpstack/dumpcore are to be found, rather than the two-step process of adding that dir to PATH and also setting this other var.

peterbarker: > I'm thinking it might be easier to understand (for the users like me w/ weird setups) if the env var was something more like AP_OVERRIDE_PATH_TO_SCRIPTS=/path/to/scripts, to just specify the full path to where dumpstack/dumpcore are to be found, rather than the two-step process of adding that dir to PATH and also setting this other var.

Yeah, I considered doing this approach but went with your suggestion instead as a tested solution :-)

Decided at DevCall to go with this, looking for an env variable speciying the path.  The reason I prefer this one is that you don't need to modify your path to find the tool....

peterbarker:
Fixed to use a path from the environment to the ArduPilot scripts.

tridge (approved):

PR #27743, opened 2024-08-03
Tools: remove redundant DroneCAN packet buffer initialization

tpwrules:
The `<msg>_encode` method always zeros the buffer up to `<msg>_MAX_SIZE` bytes so there is no need to do it before calling the function.

Saves at least 8 bytes per instance, e.g. 136 bytes on MatekL431-GPS. Tested in SITL and on a couple nodes.


tridge (approved):

tridge:
thanks!

PR #27745, opened 2024-08-03
AP_ICEngine: move to using servo endpoints and RC option

IamPete1:
This removes the separate set of servo endpoints that ICE was using. `ICE_PWM_IGN_ON`, `ICE_PWM_IGN_OFF`, `ICE_PWM_STRT_ON`, `ICE_PWM_STRT_OFF`. Instead the normal servo function min/max/reversed params are used. This also removes `ICE_START_CHAN` in favor of a new aux function. 

The param conversion has been tested in SITL. 

In order to allow the user to go back and forth between versions a new hidden param has been added to act as a format version for the ICE lib. This ensures the conversion is only done one time. 

In order to maintain the existing behavior a new `set_output_min_max_defaults` was added to the `SRV_Channel` lib.

There should be no change in behavior. The goal is just to clear up conflicting parameters to make it easier to setup.


MattKear (approved):
This is a very nice clean up.
This will need an update to the wiki page:
https://ardupilot.org/plane/docs/common-ice.html

tridge (dismissed):
looks like a reasonable approach, thanks!
missed an ICE_START_CHAN in plane-ice.parm

IamPete1:
> missed an ICE_START_CHAN in plane-ice.parm

This actually broke CI because it was set to 7 in the defaults and then quadplane set 11, so the option was setup twice. I have removed the manual setting of it in the autotest.

IamPete1:
CI should be fixed now, turned out that the defaults from file does not get applied for the RC option in CI because other tests had previously used that channel. The fix is to move to channel 11 rather than 7 so none of the other tools have set that option so we can still use the default. We could also keep it on 7 and have it in the default AND also set it manually in the test.

tridge (approved):

PR #22023, opened 2022-10-20
APM_Control: factor out an AP_PitchRollController base class

peterbarker:
Saves ~100 bytes on MatekF405

PR #27747, opened 2024-08-04
APM_Control: add AP_FW_Controller as common base class to roll and pitch controlers

IamPete1:
This moves to a shared base class between the plane roll and pitch controllers, there is lots of shared code which was duplicated. This helps with future tidyups as they now only need to happen in one place.

### DIFF
--- a/libraries/APM_Control/examples/AP_FW_Controller_test/wscript
+++ b/libraries/APM_Control/examples/AP_FW_Controller_test/wscript
@@ -1,0 +1,19 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+def build(bld):
+
+    if bld.env.BOARD != 'linux':
+        return
+
+    # Turn off the normal AHRS so we can define a custom one
+    bld.env.DEFINES += ['AP_AHRS_ENABLED=0']
+
+    # Turn off other things that are expecting a complete AHRS
+    bld.env.DEFINES += ['AP_FRSKY_SPORT_PASSTHROUGH_ENABLED=0', 'AP_AIRSPEED_ENABLED=0']
+
+    bld.ap_program(
+        use='ap',
+        program_groups=['examples'],
+    )
+
--- a/libraries/APM_Control/examples/AP_FW_Controller_test/wscript
+++ b/libraries/APM_Control/examples/AP_FW_Controller_test/wscript
@@ -3,17 +3,19 @@
 
 def build(bld):
 
-    if bld.env.BOARD != 'linux':
+    if bld.env.BOARD != 'sitl':
         return
 
-    # Turn off the normal AHRS so we can define a custom one
-    bld.env.DEFINES += ['AP_AHRS_ENABLED=0']
-
-    # Turn off other things that are expecting a complete AHRS
-    bld.env.DEFINES += ['AP_FRSKY_SPORT_PASSTHROUGH_ENABLED=0', 'AP_AIRSPEED_ENABLED=0']
+    bld.ap_stlib(
+        name='AP_FW_Controller_test_libs',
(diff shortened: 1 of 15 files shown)


IamPete1:
This is a example of something it would simplify. https://github.com/ArduPilot/ardupilot/pull/27694

peterbarker:
I tried this on for size and got bounced: https://github.com/ArduPilot/ardupilot/pull/22023

"Needs a lot of testing" was not the only comment made.

From the meeting notes:
```
UTC0016 - [APM_Control: factor out an AP_PitchRollController base class by peterbarker · Pull Request #22023 · ArduPilot/ardupilot · GitHub](https://github.com/ArduPilot/ardupilot/pull/22023/files)

    Parameter conversion
    Check scale factor
    Pitch tracking in TECS
```

... and it was *also* stated that increasing complexity in this area was not wanted and burning those extra bytes was fine.

timtuxworth:
This is awesome. It's a very nice structural change that should make some other things much easier in future.

IamPete1:
Rebased and add enum for autotune type so the `autotune_start` function can be in the base class. Still need to come-up with a test to prove correctness.

robustini:
Thanks Peter, this would be great for a smoother management evolution of automatic controls in the future.

IamPete1:
I have added a test that runs a 60 second chirp through the roll and pitch controllers. This is then run in a test matrix with a range of roll/pitch/airspeed/flags. Total runs of 882 (441 for roll and 441 for pitch). The sweep is dumped to a csv with time, angle error, output, rate target, rate actual, rate error, P, I D, FF, DFF, DMod, slew rate, limit flag, PD limit flag, reset flag and I term set flag. The floating point values are printed to four decimal places. There is no change in these output files with the rework. 

A plot of one run using the included python:
image

image

image

image

image

tridge (commented):
I'd like to see some manual testing on:
 - inverted flight
 - auto-tune

libraries/APM_Control/examples/AP_FW_Controller_test/wscript
```diff
@@ -0,0 +1,19 @@
...
+def build(bld):
+
+    if bld.env.BOARD != 'linux':
+        return
+
+    # Turn off the normal AHRS so we can define a custom one
+    bld.env.DEFINES += ['AP_AHRS_ENABLED=0']
+
+    # Turn off other things that are expecting a complete AHRS
+    bld.env.DEFINES += ['AP_FRSKY_SPORT_PASSTHROUGH_ENABLED=0', 'AP_AIRSPEED_ENABLED=0']
```
tridge, line 13: better to have:
```
bld.env.AP_LIB_REMOVE_LIBRARY['AP_FW_Controller_test'] = ['AP_AHRS']
```

IamPete1: I swapped to HAL SITL in the end and used the sim backend. Slightly less hacky.

IamPete1:
> * inverted flight

Wider range of roll and pitch added to the test sweep. Roll from -170 to +180 in 10 get steps, pitch from -80 to + 80 in 10 deg steps. This brings the total number of test runs to 11016, still no change in output.

IamPete1:
> * auto-tune

Added results checking to the existing autotest. Roll is very consistent, values within 2% every time. Pitch is not, there seems to be three solutions for P and two for D. There is no change as a result of this PR, but we might want to look into it...

IamPete1:
I have added some more cases to the pitch result checker, I ran it 25 times localy and got this:
image

No guarantees that it won't fail occasionally, but it does seem to be a genuine problem that we should try and understand, roll is the same every time.

tridge (approved):

PR #27748, opened 2024-08-05
ArduPlane: add build_option.py entry for Plane offboard-guided control

peterbarker:
This is already off in minimize_features.inc - so this will allow someone to get these back on the custom build server on small boards (or take it out on the bigger boards)

Saves 16 bytes on the smaller boards because the symbol is going away entirely.

Ran it through test_build_options.py (Feature is ~2.2kB)

### DIFF
--- a/Tools/scripts/extract_features.py
+++ b/Tools/scripts/extract_features.py
@@ -263,6 +263,8 @@ class ExtractFeatures(object):
             ('AP_RSSI_ENABLED', r'AP_RSSI::init'),
 
             ('AP_ROVER_ADVANCED_FAILSAFE_ENABLED', r'Rover::afs_fs_check'),
+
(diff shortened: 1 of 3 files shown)


Tools/scripts/extract_features.py
```diff
@@ -261,6 +261,7 @@ def __init__(self, filename, nm="arm-none-eabi-nm", strings="strings"):
             ('AP_MAVLINK_MSG_HIL_GPS_ENABLED', r'mavlink_msg_hil_gps_decode'),
             ('AP_BARO_PROBE_EXTERNAL_I2C_BUSES', r'AP_Compass::_probe_external_i2c_compasses'),
             ('AP_RSSI_ENABLED', r'AP_RSSI::init'),
+            ('AP_PLANE_OFFBOARD_GUIDED_ENABLED', r'GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands'),
```
tridge, line 264: missing _SLEW

tridge (changes requested):

PR #27750, opened 2024-08-05
AP_DroneCAN: DNA_Server: lock all high level database operations

tpwrules:
This adds a database class which contains the lock over all high level database operations (as most are read-modify-write) and does the database operations on the storage, and a static instance of it which is shared among all server instances. Please see commits for details.

There should be no behavior or logic changes at all to users using only one driver, for better or worse. Users using multiple won't have duplicate IDs assigned when attaching new nodes to different drivers without rebooting in between. They also won't risk corrupting the database due to race conditions, however unlikely that probably was in practice.

It doesn't look immediately feasible in Ardupilot right now to have the same ID on two different drivers anyway as the parameters which store the GPS node IDs can't handle that. That work will have to be done too (not sure if there are other places which don't store the driver index), in addition to expanding the storage. I do not plan to do this work; hoping that 120 IDs should be enough for anybody. Fortunately, once that is done, the DNA server will not need any modification except for the instantiation of one database per server.

Tested in SITL using a node on each bus. Testing on hardware to come soon.

### DIFF
--- a/libraries/AP_DroneCAN/AP_DroneCAN_DNA_Server.cpp
+++ b/libraries/AP_DroneCAN/AP_DroneCAN_DNA_Server.cpp
@@ -53,7 +53,7 @@ AP_DroneCAN_DNA_Server::AP_DroneCAN_DNA_Server(AP_DroneCAN &ap_dronecan, CanardI
 
 /* Method to generate 6byte hash from the Unique ID.
 We return it packed inside the referenced NodeData structure */
-void AP_DroneCAN_DNA_Server::getHash(NodeData &node_data, const uint8_t unique_id[], uint8_t size) const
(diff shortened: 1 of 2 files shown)


tridge:
just like to consider just a static struct db instead of the class, but happy to go with your judgement

tpwrules:
Basically the short answer is that I'd like it to look like this eventually for type safety and additional features I want to add and cleanups I want to do. We could do a separate PR with a static struct and migrate to a class later but that seems like unnecessary overhead.

tpwrules:
Bench tested on real hardware with some nodes, paying particular attention to preservation of the database across upgrades and correct ordering of dual GPSes. It all looks good.

bugobliterator (approved):
LGTM

PR #27752, opened 2024-08-05
AP_Scripting: add SITL RealFlight gimbal driver

robertlong13:
Useful for testing this PR: https://github.com/ArduPilot/MissionPlanner/pull/3386

Goes along with this SITL Models PR as well: []


Georacer:
Perhaps it's a silly question, but: What makes this driver targeted to RealFlight? Isn't it applicable to any PWM gimbal?

robertlong13:
> Perhaps it's a silly question, but: What makes this driver targeted to RealFlight? Isn't it applicable to any PWM gimbal?

No, you're right, it would be applicable to any PWM gimbal that stabilizes roll and pitch internally, takes PWM commands for body-frame roll and pitch angle commands, is set to yaw-follow, and has a servo channel for zooming the camera (though that last part is optional). If there are any of that kind of gimbal still on the market (I feel like they all dried up in 2017) then this would be a nice backend for that.

The only reason this has RealFlight written all over it is that I made it for RealFlight and it works well with it.

Georacer:
Then if it's not too much trouble, I'd prefer seeing this be turned into such a generic PWM driver. Especially since it's placed in the `drivers` folder.
I realize it might need a bit more scrutiny in that case.

Of course, staying a RealFlight script (and perhaps moving it to `examples`) is also fine by me.

robertlong13:
> Then if it's not too much trouble, I'd prefer seeing this be turned into such a generic PWM driver. Especially since it's placed in the `drivers` folder. I realize it might need a bit more scrutiny in that case.

If someone has one they want to flight test for me, then I'd happily make it generic.

isso11:
> > Then if it's not too much trouble, I'd prefer seeing this be turned into such a generic PWM driver. Especially since it's placed in the `drivers` folder. I realize it might need a bit more scrutiny in that case.
> 
> If someone has one they want to flight test for me, then I'd happily make it generic.
> 
> I can't even find one to buy that isn't STorM32 based, which wouldn't need this driver.

Hi @robertlong13 
 I can test this on a real hardware. I have some gimbals at my university and can test if you want.

PR #27753, opened 2024-08-05
CAMERA_FOV_STATUS altitude unit fix

robertlong13:
Should report mm and not cm


robertlong13:
I've reduced the scope of this to just the unit fix

rmackay9 (approved):

tridge (approved):

PR #27754, opened 2024-08-05
Autotune: Add switch to test gains after tune is complete

IamPete1:
This adds a switch allowing the user to test gains in any mode once the autotune has been completed. If the user has used the switch they can then disarm in any mode and the gains will be saved if they are active. If the user does not use the switch then they must disarm in autotune to save gains per the current behavior. 

This allows easy testing of the tune in any mode and makes the procedure to save a little simpler. 

The load test gains functions have been updated to ensure they use the same criteria as the saving of gains for the load. Currently they were not checking if a axis had completed, only that the axis was enabled.

### DIFF
--- a/ArduCopter/RC_Channel.cpp
+++ b/ArduCopter/RC_Channel.cpp
@@ -76,7 +76,8 @@ void RC_Channel_Copter::init_aux_function(const AUX_FUNC ch_option, const AuxSwi
     // the following functions do not need to be initialised:
     case AUX_FUNC::ALTHOLD:
     case AUX_FUNC::AUTO:
-    case AUX_FUNC::AUTOTUNE:
+    case AUX_FUNC::AUTOTUNE_MODE:
+    case AUX_FUNC::AUTOTUNE_TEST_GAINS:
     case AUX_FUNC::BRAKE:
     case AUX_FUNC::CIRCLE:
     case AUX_FUNC::DRIFT:
@@ -291,8 +292,11 @@ bool RC_Channel_Copter::do_aux_function(const AUX_FUNC ch_option, const AuxSwitc
 #endif
 
 #if AUTOTUNE_ENABLED == ENABLED
-        case AUX_FUNC::AUTOTUNE:
+        case AUX_FUNC::AUTOTUNE_MODE:
             do_aux_function_change_mode(Mode::Number::AUTOTUNE, ch_flag);
+            break;
+        case AUX_FUNC::AUTOTUNE_TEST_GAINS:
+            copter.mode_autotune.autotune.do_aux_function(ch_flag);
             break;
 #endif
 
--- a/libraries/AC_AutoTune/AC_AutoTune.cpp
+++ b/libraries/AC_AutoTune/AC_AutoTune.cpp
@@ -109,6 +109,51 @@ void AC_AutoTune::stop()
     // we expect the caller will change the flight mode back to the flight mode indicated by the flight mode switch
 }
 
+// Autotune aux function trigger
+void AC_AutoTune::do_aux_function(const RC_Channel::AuxSwitchPos ch_flag)
+{
+    if (mode != TuneMode::SUCCESS) {
+        if (ch_flag == RC_Channel::AuxSwitchPos::HIGH) {
+            gcs().send_text(MAV_SEVERITY_NOTICE,"AutoTune: must be complete to test gains");
+        }
+        return;
+    }
+
+    switch(ch_flag) {
+        case RC_Channel::AuxSwitchPos::LOW:
+            // load original gains
+            load_gains(GainType::GAIN_ORIGINAL);
+            update_gcs(AUTOTUNE_MESSAGE_TESTING_END);
+            break;
+        case RC_Channel::AuxSwitchPos::MIDDLE:
+            // Middle position is unused for now
+            break;
+        case RC_Channel::AuxSwitchPos::HIGH:
+            // Load tuned gains
+            load_gains(GainType::GAIN_TUNED);
+            update_gcs(AUTOTUNE_MESSAGE_TESTING);
+            break;
+    }
+
+    have_pilot_testing_command = true;
+}
+
+// Possibly save gains, called on disarm
+void AC_AutoTune::disarmed(const bool in_autotune_mode)
+{
+    // True if pilot is testing tuned gains
+    const bool testing_tuned = have_pilot_testing_command && (loaded_gains == GainType::GAIN_TUNED);
+
+    // True if in autotune mode and no pilot testing commands have been received
+    const bool tune_complete_no_testing = !have_pilot_testing_command && in_autotune_mode;
+
+    if (tune_complete_no_testing || testing_tuned) {
+        save_tuning_gains();
+    } else {
+        reset();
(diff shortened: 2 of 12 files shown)


ArduCopter/RC_Channel.cpp
```diff
@@ -291,9 +292,12 @@ bool RC_Channel_Copter::do_aux_function(const AUX_FUNC ch_option, const AuxSwitc
 #endif
 
 #if AUTOTUNE_ENABLED == ENABLED
-        case AUX_FUNC::AUTOTUNE:
+        case AUX_FUNC::AUTOTUNE_MODE:
             do_aux_function_change_mode(Mode::Number::AUTOTUNE, ch_flag);
             break;
+        case AUX_FUNC::AUTOTUNE_SWITCH:
```
amilcarlucas, line 298: How about renaming to AUTOTUNE_SAVE_SWITCH ?

IamPete1: It don't save, it just allows testing of new or old gains. Save is still triggered on disarm. We could call it something more descriptive of course, the key thing was just to differentiate from the autotune mode function.

andyp1per (approved):
LGTM

libraries/AC_AutoTune/AC_AutoTune.cpp
```diff
@@ -109,6 +109,51 @@ void AC_AutoTune::stop()
...
+    }
+
+    switch(ch_flag) {
+        case RC_Channel::AuxSwitchPos::LOW:
+            // load original gains
+            load_gains(GainType::GAIN_ORIGINAL);
+            update_gcs(AUTOTUNE_MESSAGE_TESTING_END);
+            break;
+        case RC_Channel::AuxSwitchPos::MIDDLE:
+            // Middle position is unused for now
```
lthall, line 129: This is where we could change to mode AutoTune if the tune has not been completed.

********* For Discussion *********

lthall (commented):
This looks great. However I think one of my previous PR's may have introduced a bug that keeps ddt zero. I will find this problem and report back.

lthall:
We do have a bug here somewhere:
Master:
image

PR:
image

So this is a problem in the PR somewhere. Still working out why....

lthall:
Confirmed that this showed a problem in master :)

libraries/RC_Channel/RC_Channel.h
```diff
@@ -252,6 +252,7 @@ class RC_Channel {
         VFWD_THR_OVERRIDE =  176, // force enabled VTOL forward throttle method
         MOUNT_LRF_ENABLE =   177,  // mount LRF enable/disable
         FLIGHTMODE_PAUSE =   178,  // e.g. pause movement towards waypoint
+        AUTOTUNE_SWITCH =    180, // auto tune tuning switch
```
rmackay9, line 255: maybe this should be AUTOTUNE_TEST_GAINS?

rmackay9: or maybe AUTOTUNE_TEST

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -239,6 +239,7 @@ const AP_Param::GroupInfo RC_Channel::var_info[] = {
     // @Values{Plane}: 176:Quadplane Fwd Throttle Override enable
     // @Values{Copter, Rover, Plane, Blimp}: 177:Mount LRF enable
     // @Values{Copter}: 178:FlightMode Pause/Resume
+    // @Values{Copter}: 180:Autotune test tuned gains
```
IamPete1, line 242:
```suggestion
    // @Values{Copter, Plane}: 180:Autotune test tuned gains
```

rmackay9 (approved):
LGTM besides the slight switch name question

rmackay9:
We decided to wait for @bnsgeyer to test this

ArduCopter/RC_Channel.cpp
IamPete1: changed to "AUTOTUNE_TEST_GAINS"

libraries/RC_Channel/RC_Channel.h
IamPete1: changed to "AUTOTUNE_TEST_GAINS"

libraries/RC_Channel/RC_Channel.cpp
IamPete1: fixed.

bnsgeyer (approved):
Tested in SITL on heli.   LGTM.

tridge (commented):
there is a bug in the ordering of the hal.util->set_soft_armed(false); and the copter.mode_autotune.save_tuning_gains(); as otherwise AP_Param will only save the first 30 gains
randy has asked the fix to be in a separate PR, does need fixing

libraries/AC_AutoTune/AC_AutoTune.cpp
```diff
@@ -109,6 +109,51 @@ void AC_AutoTune::stop()
...
+void AC_AutoTune::disarmed(const bool in_autotune_mode)
+{
+    // True if pilot is testing tuned gains
+    const bool testing_tuned = have_pilot_testing_command && (loaded_gains == GainType::GAIN_TUNED);
+
+    // True if in autotune mode and no pilot testing commands have been received
+    const bool tune_complete_no_testing = !have_pilot_testing_command && in_autotune_mode;
+
+    if (tune_complete_no_testing || testing_tuned) {
+        save_tuning_gains();
```
tridge, line 151: do you intend to save_tuning_gains() if switch is in lower position so we currently have the old gains loaded?

tridge (changes requested):

libraries/AC_AutoTune/AC_AutoTune.cpp
lthall: If you have used the switch to test the gains and you have the switch in the "original gains" position then it should not save the tuned gains.

If you have not used the switch to test the gains and are in AutoTune mode then it will save the gains. The is should be the same as the current behaviour.

timtuxworth: Sounds confusing.

IamPete1: I'm open to better suggestions but I could not think of anything else while preserving the current behaviour, we can't make it a requirement that everyone must setup the new switch to run autotune.

Now we can send aux functions from the gcs we can't really check if the user has setup a switch until they first trigger it.

timtuxworth: How about reverse the switch low = save, high = test?

IamPete1: I'm not sure how that is less confusing, the pattern (there are one or two exceptions) for functions is that high is to take some action. In this case the action is the thing we have added namely the ability to test gains in any mode. Original gains and discarding in other modes is the existing behavior, it would be abit odd to require the user to flip a switch to high for that.

Don't forget that aux functions are edge triggered, it does not matter where the switch is until autotune has completed, only changes after that (and before you disarm) will do anything.

PR #27248, opened 2024-06-06
Apd f120 new packet modification

Pradeep-Carbonix:
Added support for ESC extended status telemetry data for CubeOrange and AP_Periph : 
 - INPUT_DUTY, 
 - OUTPUT_DUTY,
 - STATUS_FLAGS

PR #26252, opened 2024-02-18
AP_ESC_Telem: log NaNs on no measurements, various cleanups

mbuzdalov:
This PR is, in a sense, continuation of #25978 as, among other things, it implements a suggestion from that conversation.

What it does is:

- initializes RPM values, to be subsequently logged, with NaN instead of 0, so that an absence of a measurement will be reported as such;
- in `update()`, deduplicates calls to `micros64()` when actually logging ESC data;
- replaces repeated constructions such as `_telem_data[esc_index]` with an appropriate reference;
- cleans up some whitespace and redundant initialization.

PR #27755, opened 2024-08-05
AP_ESC_Telem / AP_DroneCAN: add support for extended ESC status message.

IamPete1:
This is a rework of https://github.com/ArduPilot/ardupilot/pull/27248. This adds support for the DroneCAN extended ESC status message. The new fields input duty, output duty and flags are logged in a new message. The extended message also contains motor temperature which is logged in the existing ESC message. This functionality is enabled with `AP_EXTENDED_ESC_TELEM_ENABLED` which is true if DroneCAN is enabled. 

A number of ESCs from APD and Currawong support this message.


peterbarker (approved):

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -492,6 +492,18 @@ void AP_ESC_Telem::update_telem_data(const uint8_t esc_index, const AP_ESC_Telem
         telemdata.usage_s = new_data.usage_s;
     }
 
+#if AP_EXTENDED_ESC_TELEM_ENABLED
+    if (data_mask & AP_ESC_Telem_Backend::TelemetryType::INPUT_DUTY) {
+        _telem_data[esc_index].input_duty = new_data.input_duty;
```
robertlong13, line 497: Should `_telem_data[esc_index]` be replaced with `telemdata`? When we opened the original PR you based this, we did not notice the changes made in #26252 

I can open the PR to fix that if you think we should.

IamPete1: @robertlong13 yeah your right. A PR would be great, thanks!

PR #27757, opened 2024-08-05
Copter: Change the definitions of the direct values 0 and 1 to the definition names DISABLED and ENABLED

muramura:
Copter defines DISABLED as 0 and ENABLED as 1. Using these defined names makes the settings clearer.


peterbarker:
We agreed at DevCall to go entirely the other direction, and we've merged PRs into Rover, Tracker, Blimp and Plane to do that.

Did you want to rework this to go the other way?

peterbarker:
We're going the other way with this.  The raw `ENABLED` and `DISABLED` defines make using these words in the libraries problematic.  It's why we sometimes use "NONE" in enumerations rather than "DISABLED"!

https://github.com/ArduPilot/ardupilot/pull/28004

Thanks anyway!

PR #27758, opened 2024-08-05
Plane: Takeoff improvements

Georacer:
This PR fixes TECS oscillations that would occur once mode TAKEOFF would reach `TKOFF_ALT`.
Additionally, it harmonizes the takeoff behaviour between modes AUTO and TAKEOFF.
Finally, it fixes a few bugs.

Many thanks to @Hwurzburg for sharing code for fixing the pitch setpoint.

## Notable behaviour changes

1. Mode TAKEOFF and AUTO now behave identically during a takeoff.
2. `TKOFF_LVL_ALT` now also affects AUTO mode.
3. TAKEOFF mode now respects the level-off angles, as it approaches altitude. There should now be significantly less overshoot.
4. The behaviour of TKOFF_ROTATION_SPD now works as advertized. It was inactive before.

## Notable code changes

1. Mode TAKEOFF now spends all of the climb in FlightStage::TAKEOFF, just like AUTO mode did.
2. `TECS::set_pitch_max_limit()` is now split into `TECS::set_pitch_max()` and `TECS::set_pitch_min()`. Its usage in `quadplane.cpp` has been modified accordingly.
3. TECS pitch limits are now applied at the end of the calculation. Before, vertical acceleration limitations were applied last.
5. TECS: `_post_TO_hgt_offset` will now never pull height demand above the target altitude.
6. `mode_takeoff.cpp` is now prevented from switching behaviours from past `TKOFF_ALT`-2m back to climb behaviour. This would cause behaviour switching if the plane would drop slightly lower than 2m from `TKOFF_ALT`.
7. There are many occasions were TECS is overridden both in pitch and throttle. When that happens, TECS is now effectively inactive and will reset itself.
8. Changed behaviour of `TECS_PITCH_MIN` to match description.
9. `servos.cpp` now doesn't try to enforce all of the takeoff throttle logic combinations. Only `TKOFF_THR_MAX` and `TKOFF_THR_MIN`.

## Tests

The following were tested in autotests:

1. Altitude and pitch no longer oscillation upon reaching the takeoff altitude. ✅
2. TAKEOFF and AUTO takeoffs now behave exactly the same in all combinations of `ARSPD_USE` and `TKOFF_OPTIONS`. ✅
3. The level-off of minimum pitch demand now works in TAKEOFF. ✅
4. When `TKOFF_ROTATE_SPD` is nonzero, pitch demand will be 5deg before the rotation speed and will climb up to the minimum takeoff pitch for groundspeed equal to cruise airspeed. ✅

The following were tested in manual SITL:

1. Altitude and pitch no longer oscillation upon reaching the takeoff altitude. ✅
2. The level-off of minimum pitch demand now works in TAKEOFF. ✅
3. When `TKOFF_ROTATE_SPD` is nonzero, pitch demand will be 5deg before the rotation speed and will climb up to the minimum takeoff pitch for groundspeed equal to cruise airspeed. ✅

## Known issues

1. Ground speed is used in some airspeed comparisons.
5. If AUTO is engaged while in flight and has a NAV_TAKEOFF point, it will force the pitch to positive while it gets processed. This lasts for a few cycles.
6. The SITL plane doesn't regulate pitch well when it's flying at full throttle and fast. Nothing to do with this PR, just an observation.

## Future work

1. I noticed that if you load a mission before you get a GNSS fix, the NAV_TAKEOFF waypoint (i.e. next_WP_loc) will be set at coordinates at the home coordinates and altitude. This can break some takeoff logic in AUTO mode.
2. Render `TKOFF_LVL_ALT` a higher-level parameter, as it is now used also by AUTO mode, not just TAKEOFF.

### DIFF
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -446,6 +446,10 @@ private:
         uint32_t accel_event_ms;
         uint32_t start_time_ms;
         bool waiting_for_rudder_neutral;
+        float throttle_lim_max;
+        float throttle_lim_min;
+        uint32_t throttle_max_timer_ms;
+        // Good candidate for keeping the initial time for TKOFF_THR_MAX_T.
     } takeoff_state;
 
     // ground steering controller state
@@ -1131,7 +1135,7 @@ private:
     bool auto_takeoff_check(void);
     void takeoff_calc_roll(void);
     void takeoff_calc_pitch(void);
-    void takeoff_calc_throttle(const bool use_max_throttle=false);
+    void takeoff_calc_throttle();
     int8_t takeoff_tail_hold(void);
     int16_t get_takeoff_pitch_min_cd(void);
     void landing_gear_update(void);
--- a/ArduPlane/mode_takeoff.cpp
+++ b/ArduPlane/mode_takeoff.cpp
@@ -81,6 +81,7 @@ void ModeTakeoff::update()
     const float alt = target_alt;
     const float dist = target_dist;
     if (!takeoff_mode_setup) {
+        plane.auto_state.takeoff_altitude_rel_cm = alt * 100;
         const uint16_t altitude = plane.relative_ground_altitude(false,true);
         const float direction = degrees(ahrs.get_yaw());
         // see if we will skip takeoff as already flying
@@ -92,11 +93,12 @@ void ModeTakeoff::update()
                 plane.set_flight_stage(AP_FixedWing::FlightStage::NORMAL);
             } else {
                 gcs().send_text(MAV_SEVERITY_INFO, "Climbing to TKOFF alt then loitering");
+                start_loc = plane.current_loc;
                 plane.next_WP_loc = plane.current_loc;
                 plane.next_WP_loc.alt += ((alt - altitude) *100);
                 plane.next_WP_loc.offset_bearing(direction, dist);
                 takeoff_mode_setup = true;
-                plane.set_flight_stage(AP_FixedWing::FlightStage::NORMAL);
+                plane.set_flight_stage(AP_FixedWing::FlightStage::TAKEOFF);
             }
             // not flying so do a full takeoff sequence
         } else {
@@ -117,8 +119,9 @@ void ModeTakeoff::update()
                 gcs().send_text(MAV_SEVERITY_INFO, "Takeoff to %.0fm for %.1fm heading %.1f deg",
                                 alt, dist, direction);
                 plane.takeoff_state.start_time_ms = millis();
+                plane.takeoff_state.throttle_max_timer_ms = millis();
                 takeoff_mode_setup = true;
-
+                plane.steer_state.hold_course_cd = wrap_360_cd(direction*100); // Necessary to allow Plane::takeoff_calc_roll() to function.
             }
         }
     }
@@ -126,48 +129,49 @@ void ModeTakeoff::update()
     if (plane.check_takeoff_timeout()) {
         plane.set_flight_stage(AP_FixedWing::FlightStage::NORMAL);
         takeoff_mode_setup = false;
-
     }
-        
-    // we finish the initial level takeoff if we climb past
-    // TKOFF_LVL_ALT or we pass the target location. The check for
-    // target location prevents us flying forever if we can't climb
-    // reset the loiter waypoint target to be correct bearing and dist
-    // from starting location in case original yaw used to set it was off due to EKF
-    // reset or compass interference from max throttle
+
+    // We update the waypoint to follow once we're past TKOFF_LVL_ALT or we
+    // pass the target location. The check for target location prevents us
+    // flying towards a wrong location if we can't climb.
+    // This will correct any yaw estimation errors (caused by EKF reset
+    // or compass interference from max throttle), since it's using position bearing.
     const float altitude_cm = plane.current_loc.alt - start_loc.alt;
-    if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF &&
-        (altitude_cm >= level_alt*100 ||
-         start_loc.get_distance(plane.current_loc) >= dist)) {
-        // reset the target loiter waypoint using current yaw which should be close to correct starting heading
+    if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF
+        && plane.steer_state.hold_course_cd == -1 // This is the enter-once flag.
+        && (altitude_cm >= (level_alt * 100.0f) || start_loc.get_distance(plane.current_loc) >= dist)
+    ) {
         const float direction = start_loc.get_bearing_to(plane.current_loc) * 0.01;
         plane.next_WP_loc = start_loc;
         plane.next_WP_loc.offset_bearing(direction, dist);
         plane.next_WP_loc.alt += alt*100.0;
+        plane.steer_state.hold_course_cd = wrap_360_cd(direction*100); // Necessary to allow Plane::takeoff_calc_roll() to function.
+    }
+        
+    // We finish the initial level takeoff if we climb past
+    // TKOFF_ALT or we pass the target location. The check for
+    // target location prevents us flying forever if we can't climb.
+    if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF &&
+        (altitude_cm >= (alt*100 - 200) ||
+        start_loc.get_distance(plane.current_loc) >= dist)) {
         plane.set_flight_stage(AP_FixedWing::FlightStage::NORMAL);
     }
 
     if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF) {
-        //below TAKOFF_LVL_ALT
+        //below TKOFF_ALT
         plane.takeoff_calc_roll();
         plane.takeoff_calc_pitch();
-        plane.takeoff_calc_throttle(true);
+        plane.takeoff_calc_throttle();
     } else {
-        if ((altitude_cm >= alt * 100 - 200)) { //within 2m of TKOFF_ALT, or above and loitering
 #if AP_FENCE_ENABLED
-            if (!have_autoenabled_fences) {
-                plane.fence.auto_enable_fence_after_takeoff();
-                have_autoenabled_fences = true;
-            }
-#endif
-            plane.calc_nav_roll();
-            plane.calc_nav_pitch();
-            plane.calc_throttle();
-        } else { // still climbing to TAKEOFF_ALT; may be loitering
-            plane.takeoff_calc_throttle();
-            plane.takeoff_calc_roll();
-            plane.takeoff_calc_pitch();
+        if (!have_autoenabled_fences) {
+            plane.fence.auto_enable_fence_after_takeoff();
+            have_autoenabled_fences = true;
         }
+#endif
+        plane.calc_nav_roll();
+        plane.calc_nav_pitch();
+        plane.calc_throttle();
         
         //check if in long failsafe due to being in initial TAKEOFF stage; if it is, recall long failsafe now to get fs action via events call
         if (plane.long_failsafe_pending) {
--- a/ArduPlane/takeoff.cpp
+++ b/ArduPlane/takeoff.cpp
@@ -121,6 +121,7 @@ bool Plane::auto_takeoff_check(void)
         takeoff_state.launchTimerStarted = false;
         takeoff_state.last_tkoff_arm_time = 0;
         takeoff_state.start_time_ms = now;
+        takeoff_state.throttle_max_timer_ms = now;
         steer_state.locked_course_err = 0; // use current heading without any error offset
         return true;
     }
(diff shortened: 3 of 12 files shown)


Georacer:
This is the new behaviour of arming in mode TAKEOFF in stock SITL:
image

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1156,7 +1158,16 @@ void AP_TECS::_initialise_states(int32_t ptchMinCO_cd, float hgt_afe)
...
+        
+        if (!_flag_throttle_forced) {
+            // Calculate the takeoff target height offset before _hgt_dem_in_raw gets reset below.
+            // Prevent the offset from becoming negative.
+            _post_TO_hgt_offset = MAX(MIN(_climb_rate_limit * _hgt_dem_tconst, _hgt_dem_in_raw - hgt_afe), 0);
+        } else {
+            // If throttle is externally forced, this mechanism of adding energy is unnecessary.
+            _post_TO_hgt_offset = 0;
+        }
+
```
Georacer, line 1170: The offset is now a little smarter:
If TECS is externally overridden, there's little point in having it.

In normal operation, we don't want to have the offset bring the demanded height above the raw target, as it leads to altitude overshoots in general. And we also don't want it to become negative.

Here;s a small opening to a potential bug. This offset is the only mechanism for adding throttle (energy) during `FlightStage::TAKOFF`. As it becomes zero as the altitude error becomes zero, throttle would also become zero and could lead to the altitude error never actually closing.
`TKOFF_THR_MIN` (default=0.6) is necessary to ensure that we will reach the altitude.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1460,10 +1394,9 @@ void AP_TECS::_update_throttle_limits() {
 
     // Configure min throttle.
 
-    // If less min throttle is allowed during takeoff, use it.
+    // Use takeoff min throttle, if applicable.
     bool use_takeoff_throttle = _flight_stage == AP_FixedWing::FlightStage::TAKEOFF || _flight_stage == AP_FixedWing::FlightStage::ABORT_LANDING;
-    const bool use_throttle_range = (aparm.takeoff_options & (uint32_t)AP_FixedWing::TakeoffOption::THROTTLE_RANGE);
-    use_takeoff_throttle = use_takeoff_throttle && (use_throttle_range == 1) && (aparm.takeoff_throttle_min != 0);
```
Georacer, line 1466: `use_throttle_range` was a bug here on my part. When `TKOFF_OPTIONS=1` is irrelevant to using `TKOFF_THR_MIN`.

Granted, this section would be better if it was identical to how `TKOFF_THR_MAX` is used above.

Georacer:
Selected responses:

`ARSPD_USE=1`, `TKOFF_OPTIONS=0`:
image

`ARSPD_USE=1`, `TKOFF_OPTIONS=1`:
image

`ARPSD_USE=0`
image

Hwurzburg (approved):

tridge (changes requested):
I think there should be only one call into TECS to give the limits for the next loop, not two calls giving different limits

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1067,6 +1064,9 @@ void AP_TECS::_update_pitch(void)
         _pitch_dem = _last_pitch_dem - ptchRateIncr;
     }
 
+    // Constrain pitch demand
+    _pitch_dem = constrain_float(_pitch_dem_unc, _PITCHminf, _PITCHmaxf);
+
```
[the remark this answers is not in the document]
tridge: @priseborough are you ok with this re-ordering?

Georacer:
This is me thinking out loud and writing down my thought process:

In the [last dev call](https://discuss.ardupilot.org/t/ardupilot-eu-dev-call-2024-08-14/122304) one of the review points was that TECS throttle limits should be set (through the `TECS.set_throttle_min/max()` in one place in the codebase.
The current situation where multiple places invoke this and affect the limits is confusing.

Currently there are 3 places in the codebase that make decisions on the throttle limits and 2 places where the throttle is actually written.
- `Plane::update_control_mode()` runs as a FAST_TASK, on every loop, calling the `update()` of each throttle-controlling flight mode.
	- This invokes eventually `Plane::takeoff_calc_throttle()`, which writes throttle limits to TECS,
	- then asks TECS for the throttle output in `Plane::calc_throttle(),
	- then writes on
```c++
SRV_Channels::set_output_scaled(SRV_Channel::k_throttle, commanded_throttle);
```
- `servos.cpp:set_servos()` runs as a FAST_TASK, on every loop.
	- These will overwrite the `SRV_Channel::k_throttle` values (assuming it runs after `update_control_mode()`, as it has lower priority).
	- They will also set the new, final limits to TECS.
- TECS control updates occur under the `update_alt()` task, at 10Hz.

Regarding which piece of code enacts which piece of logic:

TECS logic takes into account:
- `THR_MAX/MIN`
- `TKOFF_THR_MAX/MIN`
- `TKOFF_THR_MAX_T`
- `FlightStage`, as a dependency of the above.
- `landing.is_flaring()` to set min to 0.

`Plane::takeoff_calc_throttle()` logic takes into account:
- `TKOFF_THR_MAX/MIN`
- `TKOFF_OPTIONS`

`Plane::apply_throttle_limits()` logic takes into account:
- `THR_MAX/MIN`
- `TKOFF_THR_MAX/MIN`
- `landing.is_flaring()` to set min to 0.

What I think I'll do is:

1. Move all usage of `THR_MAX/MIN` to `Plane::apply_throttle_limits()`
1. Move all usage of `landing.is_flaring()` to `apply_throttle_limits()`.
1. Move all usage of `TKOFF_THR_MAX/MIN`, `TKOFF_THR_MAX_T`,`TKOFF_OPTIONS` to `takeoff_calc_throttle()`. After all the takeoff part of the codebase is the one that should logically apply the takeoff-related parameters.
1. `takeoff_calc_throttle()` will calculate throttle limits and store them in `_tkoff_thr_max/min`. These will also be retrievable via `get_tkoff_thr_max/min`. Care is needed that they will get restored when `takeoff_calc_throttle()` is no longer called.
1. `apply_throttle_limits()` will bring in `get_tkoff_thr_max/min()` and take it into account towards calculating the `limited_throttle`. It will then set those limits to TECS. This will be the only place to do so, because it knows the final limits that are used.

Georacer:
I've pushed another pair of commits to address the review comments in the last Dev Call:
The throttle limits logic has been refactored such that:
- TECS is now completely agnostic of the throttle limits. On every control iteration `servos.cpp:apply_throttle_limits()` is expected to set them.
- `servos.cpp:apply_throttle_limits()` does not take into account takeoff-related throttle parameters. They have all been moved to `takeoff.cpp`. `apply_throttle_limits()` is expected to query `takeoff.cpp` for the limts, where applicable.
- `takeoff_calc_throttle()` now writes the takeoff related throttle limits into `takeoff_state` new struct members.

Georacer:
I went over @peterbarker 's flight test logs.
All went as expected: he had set up `ARSPD_USE=0` and `TKOFF_OPTIONS=0`, so the aircraft climbed at full throttle until `TKOFF_ALT` or until `TKOFF_DIST`
It followed the takeoff profile desired pitch exactly, except for the intervals where the pilot mixed pitch input, as expected.

Thanks Peter!

Georacer:
> this is failing a flapping copter fence test that I thin @andypiper has a PR to disable

Ah, thanks for mentioning it. It did pass on my PC locally, but failing twice in a row in CI server got me a bit worried.

ArduPlane/mode_takeoff.cpp
```diff
@@ -117,57 +119,57 @@ void ModeTakeoff::update()
...
+    // This will correct any yaw estimation errors (caused by EKF reset
+    // or compass interference from max throttle), since it's using position bearing.
     const float altitude_cm = plane.current_loc.alt - start_loc.alt;
-    if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF &&
-        (altitude_cm >= level_alt*100 ||
-         start_loc.get_distance(plane.current_loc) >= dist)) {
-        // reset the target loiter waypoint using current yaw which should be close to correct starting heading
+    if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF
+        && plane.steer_state.hold_course_cd == -1 // This is the enter-once flag.
+        && altitude_cm >= level_alt
```
tridge, line 140: this seems to be comparing a cm value with a meter value

tridge (changes requested):

libraries/AP_TECS/AP_TECS.h
```diff
@@ -429,6 +431,10 @@ class AP_TECS {
 
     // need to reset on next loop
     bool _need_reset;
+    // Flag if someone else drives pitch externally.
+    bool _flag_pitch_forced;
```
tridge, line 435: i'm not keen on the reset code as a special case for under 1 degree pitch range, it implies we have some bad behaviour at narrow pitch ranges, eg. 1.5 degrees. If we don't have any bad behaviour then why do we need the special case?

ArduPlane/takeoff.cpp
```diff
@@ -179,66 +180,111 @@ void Plane::takeoff_calc_roll(void)
...
+            TECS_controller.set_pitch_min(0.01f*nav_pitch_cd);
+            TECS_controller.set_pitch_max(0.01f*nav_pitch_cd);
+            return;
+        } else if (gps.ground_speed() <= (float)aparm.airspeed_cruise) {
+            // If rotate speed applied, gradually transition from TKOFF_GND_PITCH to the climb angle.
+            // This is recommended for ground takeoffs, so delay rotation until ground speed indicates adequate airspeed.
+            const uint16_t min_pitch_cd = 500; // Set a minimum of 5 deg climb angle.
+            nav_pitch_cd = (gps.ground_speed() / (float)aparm.airspeed_cruise) * auto_state.takeoff_pitch_cd;
+            nav_pitch_cd = constrain_int32(nav_pitch_cd, min_pitch_cd, auto_state.takeoff_pitch_cd); 
+            TECS_controller.set_pitch_min(0.01f*nav_pitch_cd);
```
tridge, line 199: this doesn't look right, this is using fixed pitch after rotate and before cruise speed. That doesn't sound right at all, we won't be able to rotate

tridge: we've lost the case of with an airspeed sensor and doing rotating rolling takeoff

tridge (changes requested):

ArduPlane/Plane.h
```diff
@@ -439,6 +439,10 @@ class Plane : public AP_Vehicle {
         uint32_t accel_event_ms;
         uint32_t start_time_ms;
         bool waiting_for_rudder_neutral;
+        float throttle_lim_max;
```
tridge, line 442: my concern is these variables may persist longer than we want them to, we could reset them in Plane::update_control_mode() to ensure we can't use stale values

ArduPlane/mode_takeoff.cpp
```diff
@@ -117,57 +119,57 @@ void ModeTakeoff::update()
...
-    // from starting location in case original yaw used to set it was off due to EKF
-    // reset or compass interference from max throttle
+
+    // We update the waypoint to follow once we're past TKOFF_LVL_ALT.
+    // This will correct any yaw estimation errors (caused by EKF reset
+    // or compass interference from max throttle), since it's using position bearing.
     const float altitude_cm = plane.current_loc.alt - start_loc.alt;
-    if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF &&
-        (altitude_cm >= level_alt*100 ||
-         start_loc.get_distance(plane.current_loc) >= dist)) {
```
tridge, line 141: i think we need to keep this safety mechanism, and the comment that explains why it is needed (eg. user doesn't fully charge battery, or sets a silly level alt)

ArduPlane/takeoff.cpp
```diff
@@ -179,66 +180,111 @@ void Plane::takeoff_calc_roll(void)
...
-        if (nav_pitch_cd < takeoff_pitch_min_cd) {
-            nav_pitch_cd = takeoff_pitch_min_cd;
+        // At any rate, we don't want to go lower than the minimum pitch bound.
+        if (nav_pitch_cd < pitch_min_cd) {
+            nav_pitch_cd = pitch_min_cd;
         }
     } else {
-        if (g.takeoff_rotate_speed > 0) {
-            // Rise off ground takeoff so delay rotation until ground speed indicates adequate airspeed
-            nav_pitch_cd = (gps.ground_speed() / (float)aparm.airspeed_cruise) * auto_state.takeoff_pitch_cd;
```
tridge, line 197: why is this being removed?

tridge (changes requested):

tridge:
@Hwurzburg If you have the time, a set of SITL tests on this PR would be very worthwhile!

Hwurzburg:
Everything looks good to me....alt overshoot is about 4m...within expectations I think for TECS...
Auto with and without AS
Auto_noAS
Auto
Mode Takeoff with and without AS
modetkoff_50_noAS
modetkoff_50

also verified that rotation speed and level alt works in both auto and mode takeoff

I did not check George's option, just default operation which 99% of users will use

Hwurzburg:
@tridge per your request, I ran the SITL tests above in RF using your AeroScout (am PRing a move to Released models also)....everything seemed fine with those physics and model

Hwurzburg (approved):

tridge (approved):

tridge:
many thanks to @Hwurzburg and @Georacer for all your great work and testing on this PR!

peterbarker:
I'm concerned that the glider-pull-up stuff is breaking in here...

Georacer:
> I'm concerned that the glider-pull-up stuff is breaking in here...

Henry told me a couple of days ago that might be an issue with the recent pullup PR itself; so I didn't pay much attention.
But I can take another look, of course, in the next few days.

tridge:
@Hwurzburg we do need to work out why GliderPullup is failing

Georacer:
The `GliderPullup` test should now pass on my latest commit.
What changed is that now pitch target is (as in master) first capped by min/max and then rate-limited. **(CASE M)**
Instead of being rate-limited first and then capped **(CASE P)**.

Case M can lead to the TECS pitch target being outside of the externally passed min/max.

---

In my opinion that test is only accidentally passing, thanks to this TECS "quirk".
In Case M, the TECS pitch reference only gradually rises to compensate for altitude, even though the mission Takeoff item force a positive minimum pitch
This gives enough time for the vehicle to gain forwards velocity:
image

In Case P, when the Takeoff mission item is processed (t~=271) the TECS minimum pitch is set to 15deg and the demanded pitch is immediately set there.
Thus the aircraft pulls the nose up earlier and never gains forward velocity.
It falls flat all the way to the ground.

---

In any case... I'm happy to forfeit this change for now to make the test content.

peterbarker:
Would be nice to update the documentation to note that the pitch-max should be negative on a glider like this, @Hwurzburg

Hwurzburg:
> Would be nice to update the documentation to note that the pitch-max should be negative on a glider like this, @Hwurzburg (`TECS_PITCH_MAX`)

@Georacer has a  wiki PR for this and it should be incorporated into his PR

Georacer:
> > Would be nice to update the documentation to note that the pitch-max should be negative on a glider like this, @Hwurzburg (`TECS_PITCH_MAX`)
> 
> @Georacer has a wiki PR for this and it should be incorporated into his PR

@Hwurzburg what Peter is talking about is something different:

We were discussing in the dev call today that pure gliders (no motor) should probably be set with a maximum pitch angle that is close to zero (or negative), such that they can't pull their nose up and stall.

Hwurzburg:
okay will add to TECS page where that param is discussed

timtuxworth:
I tested this today. Definitely much improved "overshoot" behavior on auto takeoff. It no longer hits the top of the climb and then dives back down like it was doing before this.

Georacer:
@timtuxworth that's great to hear!

PR #27759, opened 2024-08-05
Plane: Allow a Lua script to return to a previous mission track

timtuxworth:
As discussed in #27639 on the dev call on 29 Aug 2024, added a way to get the previous WP location prior to switching out of AUTO mode and then restore it again after (in my case) doing some terrain object avoidance. This allows a plane to resume cross tracking mid mission which is required for magnetic field measurements in wilderness areas.

I changed the names proposed by @tridge in the call because:
1. getting the previous WP could conceivably be used for other purposes not just cross tracking, so I'm proposing vehicle:get_previous_location() - REMOVED based on the dev call on 12 Aug 2024
2. Restoring the previous location for crosstracking also requires setting the internal crosstrack boolean to true, so proposing to call this vehicle:set_crosstrack_start() which both sets the prev_WP_loc and also sets auto_state.crosstrack = true to resume cross tracking for the current mission item.

The sample script includes a LocationTracker() class which can be used to save previous target locations that can be passed to set_crosstrack_start()

Includes a sample lua script which requires you to be running a mission. A switch will save the previous location (high) and switch to loiter, then resume the mission and restore crosstracking (low). This screen shot shows a loiter being triggered, then switching back to crosstracking and resuming the path by crosstracking to the mission track rather than just heading direct for the next WP (WP 10).

Screenshot 2024-08-05 100641

### DIFF
--- a/ArduPlane/ArduPlane.cpp
+++ b/ArduPlane/ArduPlane.cpp
@@ -908,6 +908,16 @@ bool Plane::set_land_descent_rate(float descent_rate)
 #endif
     return false;
 }
+
+// Allow for scripting to have control over the crosstracking when exiting and resuming missions or guided flight
+// It's up to the Lua script to ensure the provided location makes sense
+bool Plane::set_crosstrack_start(const Location &new_start_location)
+{        
+    prev_WP_loc = new_start_location;
+    auto_state.crosstrack = true;
+    return true;
+}
+
 #endif // AP_SCRIPTING_ENABLED
 
 // returns true if vehicle is landing.
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -2506,6 +2506,12 @@ function vehicle:is_taking_off() end
 ---@return boolean
 function vehicle:is_landing() end
 
+-- Set the previous target location for crosstrack and crosstrack if available in the current mode
+-- It's up to the Lua code to ensure the new_start_location makes sense
+---@param new_start_location Location_ud
+---@return boolean -- true on success
+function vehicle:set_crosstrack_start(new_start_location) end
+
 -- desc
 onvif = {}
 
--- a/libraries/AP_Scripting/examples/crosstrack_restore.lua
+++ b/libraries/AP_Scripting/examples/crosstrack_restore.lua
@@ -1,0 +1,99 @@
+--[[
+
+   example script to show interrupting a mission and then 
+   reseting the crosstracking to the correct line when
+   returning to the mission after the interruption
+
+   this functionality is only available in Plane
+--]]
+
+SCRIPT_NAME = "Crosstrack Restore"
+SCRIPT_NAME_SHORT = "XTrack"
+SCRIPT_VERSION = "4.6.0-001"
+
+
+MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+FLIGHT_MODE = {AUTO=10, RTL=11, LOITER=12, GUIDED=15, QHOVER=18, QLOITER=19, QRTL=21}
+
+local last_sw = -1
+local AUX_FN = 300
+
+-- Attempts to duplicate the code that updates the prev_WP_loc variable in the c++ code
+local function LocationTracker()
+
+   local self = {}
+
+   -- to get this to work, need to keep 2 prior generations of "target_location"
+   local previous_target_location            -- the target prior to the current one
+   local previous_previous_target_location   -- the target prior to that - this is the one we want
+
+   function self.same_loc_as(A, B)
+      if A == nil or B == nil then
+         return false
+      end
+      if (A:lat() ~= B:lat()) or (A:lng() ~= B:lng()) then
+         return false
+      end
+      return (A:alt() == B:alt()) and (A:get_alt_frame() == B:get_alt_frame())
+   end
+
+   function self.save_previous_target_location()
+      local target_location = vehicle:get_target_location()
+      if target_location ~= nil then
+         if not self.same_loc_as(previous_target_location, target_location) then
+            -- maintain three generations of location
+            previous_previous_target_location = previous_target_location
+            previous_target_location = target_location
+         end
+      else
+         previous_target_location = ahrs:get_location()
+         previous_previous_target_location = previous_target_location
+      end
+   end
+
+   function self.get_saved_location()
+      return previous_previous_target_location
+   end
+
+   return self
+end
+
+local location_tracker = LocationTracker()
+
+local function update()
+
+   -- save the previous target location only if in auto mode, if restoring it in AUTO mode
+   if vehicle:get_mode() == FLIGHT_MODE.AUTO and location_tracker ~= nil then
+      location_tracker.save_previous_target_location()
+   end
+
+   local sw_current = rc:get_aux_cached(AUX_FN)
+   if not sw_current then
+      sw_current = 0
+   end
+   if sw_current ~= last_sw then
+      last_sw = sw_current
+      if sw_current == 0 then
+        vehicle:set_mode(FLIGHT_MODE.AUTO)
+        if location_tracker ~= nil then
+            local previous_location = location_tracker.get_saved_location()
+            if previous_location ~= nil then
+               vehicle:set_crosstrack_start(previous_location)
+            end
+         end
+         gcs:send_text(MAV_SEVERITY.INFO, string.format("%s: Switched to AUTO", SCRIPT_NAME_SHORT))
+      else
+         vehicle:set_mode(FLIGHT_MODE.LOITER)
+         gcs:send_text(MAV_SEVERITY.INFO, string.format("%s: Switched to LOITER", SCRIPT_NAME_SHORT))
+      end
+   end
+
+   return update,100
+end
+
+if FWVersion:type() == 3 then
+   gcs:send_text(MAV_SEVERITY.NOTICE, string.format("%s %s script loaded", SCRIPT_NAME, SCRIPT_VERSION) )
+   return update()
+else
+   gcs:send_text(MAV_SEVERITY.NOTICE,string.format("%s: Must run on Plane", SCRIPT_NAME_SHORT))
+end
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -1286,6 +1286,11 @@ public:
 
     // allow for landing descent rate to be overridden by a script, may be -ve to climb
     bool set_land_descent_rate(float descent_rate) override;
+
+    // allow scripts to override mission/guided crosstrack behaviour
+    // It's up to the Lua script to ensure the provided location makes sense
+    bool set_crosstrack_start(const Location &new_start_location) override;
+
 #endif // AP_SCRIPTING_ENABLED
 
 };
--- a/libraries/AP_Scripting/generator/description/bindings.desc
+++ b/libraries/AP_Scripting/generator/description/bindings.desc
@@ -349,6 +349,7 @@ singleton AP_Vehicle method has_ekf_failsafed boolean
 singleton AP_Vehicle method reboot void boolean
 singleton AP_Vehicle method is_landing boolean
 singleton AP_Vehicle method is_taking_off boolean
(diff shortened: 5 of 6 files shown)


ArduPlane/ArduPlane.cpp
```diff
@@ -908,6 +908,55 @@ bool Plane::set_land_descent_rate(float descent_rate)
...
+    case Mode::Number::RTL:
+    case Mode::Number::AVOID_ADSB:
+    case Mode::Number::GUIDED:
+    case Mode::Number::AUTO:
+    case Mode::Number::LOITER:
+    case Mode::Number::TAKEOFF:
+#if HAL_QUADPLANE_ENABLED
+    case Mode::Number::QLOITER:
+    case Mode::Number::QLAND:
+    case Mode::Number::QRTL:
```
IamPete1, line 948: This certainly wont do anything in some of these modes. Maybe just auto and the two RTLs?

timtuxworth: I'm going to say I'd want it in Guided.

timtuxworth: Fixed

timtuxworth:
Tested on SITL and by Overhead Intelligence on a MFE Fighter VTOL

ArduPlane/ArduPlane.cpp
```diff
@@ -908,6 +908,50 @@ bool Plane::set_land_descent_rate(float descent_rate)
...
+#endif
```
IamPete1, line 926: I think this should be the same modes as allow setting it. The problem is that if the mode does not use that location it really could be anywhere so you might end up navigating to somewhere new rather than somewhere you have already been.

ArduPlane/ArduPlane.cpp
```diff
@@ -908,6 +908,50 @@ bool Plane::set_land_descent_rate(float descent_rate)
 #endif
     return false;
 }
+
+// allow scripte to find access prev_WP_loc
```
IamPete1, line 912:
```suggestion
// allow script to find access prev_WP_loc
```

libraries/AP_Scripting/examples/crosstrack_restore.lua
```diff
@@ -0,0 +1,47 @@
...
+      last_sw = sw_current
+      if sw_current == 0 then
+        vehicle:set_mode(FLIGHT_MODE.AUTO)
+        if previous_location ~= nil then
+            vehicle:set_crosstrack_start(previous_location)
+            gcs:send_text(MAV_SEVERITY.INFO, string.format("%s: Restored previous location", SCRIPT_NAME_SHORT))
+        end
+        gcs:send_text(MAV_SEVERITY.INFO, string.format("%s: Switched to AUTO", SCRIPT_NAME_SHORT))
+      else
+        previous_location = vehicle:get_previous_location()
```
IamPete1, line 36: If nil you could use the current location.

ArduPlane/ArduPlane.cpp
IamPete1: TBH, I'm not really a fan of this method at all. We have a binding for current destination. You could easly use that in lua for the same result. 

If we were to add a binding for the `same_loc_as` location method:
```
local previous_location

local function update()

    local target = vehicle:get_target_location()
    if (target and (not previous_location or not previous_location:same_loc_as(target)) then
        previous_location = target
    end

   ....
end
```
If we dont add a binding for `same_loc_as` then you can add a little helper:
```
local function same_loc_as(A, B)
    if (A:lat() ~= B:lat()) or (A:lon() ~= B:lon()) then
        return false
    end
    return (A:alt() == B:alt()) and (A:get_alt_frame() == B:get_alt_frame())
end
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2497,6 +2497,15 @@ function vehicle:is_taking_off() end
 ---@return boolean
 function vehicle:is_landing() end
 
+-- Get the previous target location if available in current mode
+---@return Location_ud|nil -- previous target ocation
+function vehicle:get_previous_location() end
+
+-- Set the previous target location for crosstrack and enable crosstrack if available in the current mode
+---@param new_start_location Location_ud -- target location
```
IamPete1, line 2505:
```suggestion
---@param new_start_location Location_ud
```

ArduPlane/ArduPlane.cpp
timtuxworth: > I think this should be the same modes as allow setting it. The problem is that if the mode does not use that location it really could be anywhere so you might end up navigating to somewhere new rather than somewhere you have already been.

prev_WP_loc is set in Plane::update_GPS_10Hz as well as a number of other places like do_RTL(), do_takeoff(), verify_nav_wp() set_next_wp(), MAV_CMD_GUIDED_CHANGE_HEADING, ModeCruise::navigate()

My suggestion is to allow it to be used "as required" in Lua regardless of where its set. Why restrict it?

libraries/AP_Scripting/examples/crosstrack_restore.lua
timtuxworth: Pretty sure that happens anyway, so I'm not sure why I would need/want to do this here.

ArduPlane/ArduPlane.cpp
timtuxworth: > TBH, I'm not really a fan of this method at all. We have a binding for current destination. You could easly use that in lua for the same result.
> 
> If we were to add a binding for the `same_loc_as` location method:
> 
> ```
> local previous_location
> 
> local function update()
> 
>     local target = vehicle:get_target_location()
>     if (target and (not previous_location or not previous_location:same_loc_as(target)) then
>         previous_location = target
>     end
> 
>    ....
> end
> ```
> 
> If we dont add a binding for `same_loc_as` then you can add a little helper:
> 
> ```
> local function same_loc_as(A, B)
>     if (A:lat() ~= B:lat()) or (A:lon() ~= B:lon()) then
>         return false
>     end
>     return (A:alt() == B:alt()) and (A:get_alt_frame() == B:get_alt_frame())
> end
> ```

I don't understand this

timtuxworth: > TBH, I'm not really a fan of this method at all. We have a binding for current destination. You could easly use that in lua for the same result.

As far as I can see prev_WP_loc may never have been a previous "target_location" based on all the places it is set in the code.

prev_WP_loc is set in Plane::update_GPS_10Hz as well as a number of other places like do_RTL(), do_takeoff(), verify_nav_wp() set_next_wp(), MAV_CMD_GUIDED_CHANGE_HEADING, ModeCruise::navigate() ...

libraries/AP_Scripting/examples/crosstrack_restore.lua
```diff
@@ -0,0 +1,47 @@
...
+
+local last_sw = -1
+local AUX_FN = 300
+
+local previous_location
+
+local function update()
+   local sw_current = rc:get_aux_cached(AUX_FN)
+   if not sw_current then
+      -- treat unset as notch1
```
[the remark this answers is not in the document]
timtuxworth: Fixed

libraries/AP_Scripting/examples/crosstrack_restore.lua
```diff
@@ -0,0 +1,47 @@
+--[[
+
```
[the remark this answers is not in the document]
timtuxworth: Fixed

libraries/AP_Scripting/examples/crosstrack_restore.lua
IamPete1: In any of the modes form `get_previous_location` returns false then `previous_location` will be nil here.

timtuxworth: Which is why line 31 tests for nil before using it. What would you do?

libraries/AP_Scripting/examples/crosstrack_restore.lua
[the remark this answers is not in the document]
IamPete1: Sorry, I should have been more clear. 

I meant more that the script should not be used except on plane. It will run fine on other vehicles. As your comment points warns the functionality won't work. But it will change modes just fine. Except it won't be loiter and auto on other vehicles. It will be whatever modes happen have the same number.

You could check in the script and error out if not plane. I think this should work:
```
assert(FWVersion:type() == 3, "Must be plane")
```

libraries/AP_Scripting/examples/crosstrack_restore.lua
IamPete1: It would just make the cross track from where if you flip the switch if your in a mode that does not support `get_previous_location`. 

I guess otherwise you could only accept the interrupt if in a mode that supports it. Here you always switch to loiter with the switch no matter if you have a valid cross track or not.

libraries/AP_Scripting/examples/crosstrack_restore.lua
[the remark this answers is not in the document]
timtuxworth: This will throw and error but then the scheduler will try to run the script again. I used an if statement instead which only runs update() if the test passes.

ArduPlane/ArduPlane.cpp
```diff
@@ -908,6 +908,50 @@ bool Plane::set_land_descent_rate(float descent_rate)
...
+        break;
+    default:
+        break;
+    }
+    return false;
+}
+
+bool Plane::set_crosstrack_start(const Location &new_start_location)
+{
+    switch (control_mode->mode_number()) {
```
tridge, line 938: i think we should remove the switch and unconditionally set prev_WP_loc and auto_state.crosstrack, and leave it up to script to determine when it is valid

ArduPlane/ArduPlane.cpp
```diff
@@ -908,6 +908,50 @@ bool Plane::set_land_descent_rate(float descent_rate)
 #endif
     return false;
 }
+
+// allow scripts to find access prev_WP_loc
+bool Plane::get_previous_location(Location &previous_location)
```
tridge, line 913: I think Pete is right that this functionality can be done in the script without a binding

tridge (changes requested):

ArduPlane/ArduPlane.cpp
timtuxworth: > TBH, I'm not really a fan of this method at all. We have a binding for current destination. You could easly use that in lua for the same result.
> 
> If we were to add a binding for the `same_loc_as` location method:
> 
> ```
> local previous_location
> 
> local function update()
> 
>     local target = vehicle:get_target_location()
>     if (target and (not previous_location or not previous_location:same_loc_as(target)) then
>         previous_location = target
>     end
> 
>    ....
> end
> ```
> 
> If we dont add a binding for `same_loc_as` then you can add a little helper:
> 
> ```
> local function same_loc_as(A, B)
>     if (A:lat() ~= B:lat()) or (A:lon() ~= B:lon()) then
>         return false
>     end
>     return (A:alt() == B:alt()) and (A:get_alt_frame() == B:get_alt_frame())
> end
> ```

I've thought about this a lot and I think the idea of duplicating code that already exists in c++ in Lua is a bad idea. In principle - if the functionality exists in c++ the correct solution is to create a binding.

timtuxworth: I got this to work, example script updated.

ArduPlane/ArduPlane.cpp
timtuxworth: Done

timtuxworth:
I got a CI fail on sitltest-plane. I downloaded the log but can't see the problem. Can you help @IamPete1 ?

IamPete1:
> I got a CI fail on sitltest-plane. I downloaded the log but can't see the problem. Can you help @IamPete1 ?

I have restarted it, that test has been failing occasionally.

timtuxworth:
> > I got a CI fail on sitltest-plane. I downloaded the log but can't see the problem. Can you help @IamPete1 ?
> 
> I have restarted it, that test has been failing occasionally.

Is that something I can do myself?

IamPete1:
> Is that something I can do myself?

Not directly, maintainers only. However, whenever you push it will restart all of CI, so you can always rebase to re-trigger everything.

timtuxworth:
Rebased on master

tridge (approved):

PR #27762, opened 2024-08-06
AC_Autotune: Fix bug introduced in #27370

lthall:
During the clean up to try to make things easer to understand due to poorly named variables, one of those variables was deleted and another similarly named variable used in it's place.

This fixes this problem.


rmackay9 (approved):

rmackay9:
I've marked this as MergeOnCIPass because this bug results in the vehicle's max accelerations being set to 4000 which is dangerously low.

@lthall is confident this change is correct and it looks OK to me as well

PR #27764, opened 2024-08-06
Copter: use mavlink enumeration to clarify do_change_speed implementa…

peterbarker:
…tion

I think this also really clarifies the hole we are in with groundspeed vs airspeed

### DIFF
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -1918,15 +1918,20 @@ void ModeAuto::do_yaw(const AP_Mission::Mission_Command& cmd)
 void ModeAuto::do_change_speed(const AP_Mission::Mission_Command& cmd)
 {
     if (cmd.content.speed.target_ms > 0) {
-        if (cmd.content.speed.speed_type == 2)  {
+        switch (cmd.content.speed.speed_type) {
+        case SPEED_TYPE_CLIMB_SPEED:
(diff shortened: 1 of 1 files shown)


khancyr (approved):
Match mavlink spec. LGTM

ArduCopter/mode_auto.cpp
```diff
@@ -1918,15 +1918,20 @@ void ModeAuto::do_yaw(const AP_Mission::Mission_Command& cmd)
...
             desired_speed_override.up = cmd.content.speed.target_ms;
-        } else if (cmd.content.speed.speed_type == 3)  {
+            break;
+        case SPEED_TYPE_DESCENT_SPEED:
             copter.wp_nav->set_speed_down(cmd.content.speed.target_ms * 100.0f);
             desired_speed_override.down = cmd.content.speed.target_ms;
-        } else {
+            break;
+        case SPEED_TYPE_AIRSPEED:
+        case SPEED_TYPE_GROUNDSPEED:
```
IamPete1, line 1931: do we need a default here to maintain behavior?

peterbarker: Technically yes.  I could add a default with existing behaviour but with warnings that you're using out-of-spec behaviour?

I'm also tempted to use the same code to warn that you're specifying an airspeed on a Copter which doesn't understand that.  That way when we *do* offer the ability to fly an airspeed to Copter we can use the correct enumeration entry....

rmackay9 (approved):

PR #27765, opened 2024-08-06
Move HELI Logging Down into Motors

MattKear:
Currently we only log variables from AP_MotorsHeli in the HELI log message.  This PR moves the log write function down into the motors library.  This is being proposed for the following reasons:
- We can delete accessor functions in motors that are only used by the logging function.  Making the code a little less difficult to read.
- It makes it easier to add more motors data to the log (I have another PR associated with autorotations coming shortly that will make use of this)
- ~~We get a small saving in flash cost.~~

There is no functional change.

Tested in SITL.  here is an example log output:
image

### DIFF
--- a/libraries/AP_Motors/AP_MotorsHeli_RSC.h
+++ b/libraries/AP_Motors/AP_MotorsHeli_RSC.h
@@ -154,8 +154,8 @@ private:
     const uint8_t   _instance;
 
     // channel setup for aux function
-    SRV_Channel::Aux_servo_function_t _aux_fn;
-    uint8_t         _default_channel;
+    const SRV_Channel::Aux_servo_function_t _aux_fn;
+    const uint8_t _default_channel;
 
     // internal variables
     RotorControlMode _control_mode = ROTOR_CONTROL_MODE_DISABLED;   // motor control mode, Passthrough or Setpoint
--- a/libraries/AP_Motors/AP_MotorsHeli_RSC.h
+++ b/libraries/AP_Motors/AP_MotorsHeli_RSC.h
@@ -4,6 +4,7 @@
 #include <AP_Math/AP_Math.h>            // ArduPilot Mega Vector/Matrix math Library
 #include <RC_Channel/RC_Channel.h>
 #include <SRV_Channel/SRV_Channel.h>
+#include <AP_Logger/AP_Logger.h>
 
 // default main rotor speed (ch8 out) as a number from 0 ~ 100
 #define AP_MOTORS_HELI_RSC_SETPOINT             70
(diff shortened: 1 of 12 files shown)


amilcarlucas:
```
Binary Name      Text [B]       Data [B]     BSS (B)       Total Flash Change [B] (%)      Flash Free After PR (B)
---------------  -------------  -----------  ------------  ----------------------------  -------------------------
arducopter       0 (0.0000%)    0 (0.0000%)  0 (0.0000%)   0 (0.0000%)                                      164892
ardusub          0 (0.0000%)    0 (0.0000%)  0 (0.0000%)   0 (0.0000%)                                      373828
ardurover        0 (0.0000%)    0 (0.0000%)  0 (0.0000%)   0 (0.0000%)                                      315656
antennatracker   0 (0.0000%)    0 (0.0000%)  0 (0.0000%)   0 (0.0000%)                                      633244
arducopter-heli  12 (+0.0007%)  0 (0.0000%)  4 (+0.0015%)  12 (+0.0007%)                                    165372
arduplane        0 (0.0000%)    0 (0.0000%)  0 (0.0000%)   0 (0.0000%)                                      168580
blimp            0 (0.0000%)    0 (0.0000%)  0 (0.0000%)   0 (0.0000%)                                      611032
```

libraries/AP_Motors/AP_MotorsHeli_RSC.h
```diff
@@ -147,6 +151,7 @@ class AP_MotorsHeli_RSC {
 
 private:
     uint64_t        _last_update_us;
+    uint8_t         _instance;
```
IamPete1, line 154:
```suggestion
    const uint8_t         _instance;
```
The other stuff set in the constructor should probably be const too....

MattKear:
Having chatted it through with @IamPete1 I have done a rework to the original PR.  As all variables being logged are from heli RSC the logging has been moved down a level further to the RSC.

There was already a logging function in motors, so the RSC log is now called from that.

I have renamed the log field to HRSC (Heli RSC) so "HELI" remains available should we need vehicle level logging for heli in the future.

As we have multiple RSCs in single heli (main and tail rotor) we have per instance logging.  0 is main and 1 is tail.  Tail logging will become more useful in the future, because we will need to run the tail rotor RSC without the main for the autorotation with DDFP configs.

Testing in SITL.  Single:
image

Dual:
image

Noting that only one instance is present for the Dual case (only main rotor RSC)

MattKear:
looks like we didn't get that flash saving I was hoping for....

bnsgeyer (approved):
LGTM

bnsgeyer:
@MattKear @IamPete1 Should this be taken to devcall just to have other eyes on it or do we feel comfortable amoungst us 3 to call it good and have me merge it?

peterbarker:
> @MattKear @IamPete1 Should this be taken to devcall just to have other eyes on it or do we feel comfortable amoungst us 3 to call it good and have me merge it?

I think you should merge it.

bnsgeyer:
Thanks!  Merged

PR #25814, opened 2023-12-21
Enhanced attitude control via scripting

andyp1per:

PR #27767, opened 2024-08-06
Sub: support ABOVE_TERRAIN frame in GUIDED mode posvel case

clydemcqueen:
Sub Guided_PosVel mode currently assumes `Location::AltFrame::ABOVE_ORIGIN`. This PR adds support for `Location::AltFrame::ABOVE_TERRAIN`.

This new functionality is exposed two ways:
* Using MAVLink: `SET_TARGET_POSITION_GLOBAL_INT.frame = MAV_FRAME_GLOBAL_TERRAIN_ALT_INT`
* Using Lua: `set_target_posvel_terrain(Vector3f, Vector3f)`

The motivation is to allow the development of scripts that drive the ROV at a constant xy velocity and at a constant height above the seafloor, while giving the pilot control over heading. There's a new autotest that demonstrates this.

Possible open issues / future PRs:
* Is it better to add an optional frame parameter to the `set_target_posvel_NED(Vector3f, Vector3f)` Lua binding?
* Sub Guided_WP mode could also support `Location::AltFrame::ABOVE_TERRAIN`, though this mode does not support rapid target updates

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -320,10 +320,15 @@ bool Copter::set_target_pos_NED(const Vector3f& target_pos_ned_m, bool use_yaw, 
 }
 
 // set target position and velocity (for use by scripting)
-bool Copter::set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms)
+bool Copter::set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms, Location::AltFrame alt_frame)
 {
     // exit if vehicle is not in Guided mode or Auto-Guided mode
     if (!flightmode->in_guided_mode()) {
+        return false;
+    }
+
+    // only support above origin frame for now
+    if (alt_frame != Location::AltFrame::ABOVE_ORIGIN) {
         return false;
     }
 
--- a/ArduCopter/Copter.h
+++ b/ArduCopter/Copter.h
@@ -680,7 +680,7 @@ private:
     bool get_target_location(Location& target_loc) override;
     bool update_target_location(const Location &old_loc, const Location &new_loc) override;
     bool set_target_pos_NED(const Vector3f& target_pos_ned_m, bool use_yaw, float yaw_deg, bool use_yaw_rate, float yaw_rate_degs, bool yaw_relative, bool is_terrain_alt) override;
-    bool set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms) override;
+    bool set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms, Location::AltFrame alt_frame) override;
     bool set_target_posvelaccel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms, const Vector3f& target_accel_ned_mss, bool use_yaw, float yaw_deg, bool use_yaw_rate, float yaw_rate_degs, bool yaw_relative) override;
     bool set_target_velocity_NED(const Vector3f& vel_ned_ms, bool align_yaw_to_target) override;
     bool set_target_velaccel_NED(const Vector3f& target_vel_ned_ms, const Vector3f& target_accel_ned_mss, bool use_yaw, float yaw_deg, bool use_yaw_rate, float yaw_rate_degs, bool relative_yaw) override;
--- a/ArduSub/GCS_MAVLink_Sub.cpp
+++ b/ArduSub/GCS_MAVLink_Sub.cpp
@@ -166,8 +166,10 @@ bool GCS_MAVLINK_Sub::send_info()
     CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
     send_named_float("RollPitch", sub.roll_pitch_flag);
 
+#if AP_RANGEFINDER_ENABLED
     CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
-    send_named_float("RFTarget", sub.mode_surftrak.get_rangefinder_target_cm() * 0.01f);
+    send_named_float("RFTarget", sub.flightmode->get_rangefinder_target());
+#endif
 
     return true;
 }
@@ -706,9 +708,16 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
             break;
         }
 
-        Vector3f pos_neu_cm;  // position (North, East, Up coordinates) in centimeters
+        // Guided submodes do not support acceleration control
+        if (!acc_ignore) {
+            break;
+        }
 
-        if (!pos_ignore) {
+        if (pos_ignore) {
+            if (!vel_ignore) {
+                sub.mode_guided.guided_set_velocity(Vector3f(packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f));
+            }
+        } else {
             // sanity check location
             if (!check_latlng(packet.lat_int, packet.lon_int)) {
                 break;
@@ -718,23 +727,45 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
                 // unknown coordinate frame
                 break;
             }
-            const Location loc{
-                packet.lat_int,
-                packet.lon_int,
-                int32_t(packet.alt*100),
-                frame,
-            };
-            if (!loc.get_vector_from_origin_NEU_cm(pos_neu_cm)) {
-                break;
-            }
-        }
 
-        if (!pos_ignore && !vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_destination_posvel(pos_neu_cm, Vector3f(packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f));
-        } else if (pos_ignore && !vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_velocity(Vector3f(packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f));
-        } else if (!pos_ignore && vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_destination(pos_neu_cm);
+            const Vector3f vel_neu_cm{packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f};
+
+            switch (frame) {
+                case Location::AltFrame::ABOVE_TERRAIN: {
+                    // transform global horizontal (lat, lon) input to a local (x, y) position
+                    const Location loc{packet.lat_int, packet.lon_int, 0, frame};
+                    Vector2f pos_ne_cm;
+                    if (!loc.get_vector_xy_from_origin_NE_cm(pos_ne_cm)) {
+                        break;
+                    }
+
+                    // velocity is required
+                    if (!vel_ignore) {
+                        // packet.alt is the target rangefinder range, not the target altitude
+                        sub.mode_guided.guided_set_destination_posvel(Vector3f(pos_ne_cm.x, pos_ne_cm.y, packet.alt * 100.0f), vel_neu_cm, frame);
+                    }
+                    break;
+                }
+                case Location::AltFrame::ABOVE_HOME: {
+                    // transform point from ABOVE_HOME to ABOVE_ORIGIN
+                    const Location loc{packet.lat_int, packet.lon_int, int32_t(packet.alt * 100.0f), frame};
+                    Vector3f pos_neu_cm;
+                    if (!loc.get_vector_from_origin_NEU_cm(pos_neu_cm)) {
+                        break;
+                    }
+
+                    if (vel_ignore) {
+                        sub.mode_guided.guided_set_destination(pos_neu_cm);
+                    } else {
+                        sub.mode_guided.guided_set_destination_posvel(pos_neu_cm, vel_neu_cm, Location::AltFrame::ABOVE_ORIGIN);
+                    }
+                    break;
+                }
+                case Location::AltFrame::ABSOLUTE:
+                case Location::AltFrame::ABOVE_ORIGIN:
+                    // not supported
+                    break;
+            }
         }
 
         break;
--- a/ArduSub/Log.cpp
+++ b/ArduSub/Log.cpp
@@ -47,7 +47,7 @@ void Sub::Log_Write_Control_Tuning()
         desired_alt         : pos_control.get_pos_target_U_cm() * 0.01f,
         inav_alt            : pos_control.get_pos_estimate_U_m(),
         baro_alt            : barometer.get_altitude(),
-        desired_rangefinder_alt   : mode_surftrak.get_rangefinder_target_cm() * 0.01,
+        desired_rangefinder_alt   : flightmode->get_rangefinder_target(),
         rangefinder_alt           : rangefinder_state.alt,
         terr_alt            : terr_alt,
         target_climb_rate   : (int16_t)pos_control.get_vel_target_U_cms(),
@@ -176,6 +176,7 @@ struct PACKED log_GuidedTarget {
     LOG_PACKET_HEADER;
     uint64_t time_us;
     uint8_t type;
+    uint8_t alt_frame;
     float pos_target_x;
     float pos_target_y;
     float pos_target_z;
@@ -185,12 +186,14 @@ struct PACKED log_GuidedTarget {
 };
 
 // Write a Guided mode target
-void Sub::Log_Write_GuidedTarget(uint8_t target_type, const Vector3f& pos_target, const Vector3f& vel_target)
+void Sub::Log_Write_GuidedTarget(uint8_t target_type, const Vector3f& pos_target, const Vector3f& vel_target,
+                                 Location::AltFrame alt_frame)
 {
     struct log_GuidedTarget pkt = {
         LOG_PACKET_HEADER_INIT(LOG_GUIDEDTARGET_MSG),
         time_us         : AP_HAL::micros64(),
         type            : target_type,
+        alt_frame       : static_cast<uint8_t>(alt_frame),
         pos_target_x    : pos_target.x,
         pos_target_y    : pos_target.y,
         pos_target_z    : pos_target.z,
@@ -251,6 +254,7 @@ void Sub::Log_Write_GuidedTarget(uint8_t target_type, const Vector3f& pos_target
 // @Description: Guided mode target information
 // @Field: TimeUS: Time since system startup
 // @Field: Type: Type of guided mode
+// @Field: Frame: Location.AltFrame
 // @Field: pX: Target position, X-Axis
 // @Field: pY: Target position, Y-Axis
 // @Field: pZ: Target position, Z-Axis
@@ -276,7 +280,7 @@ const struct LogStructure Sub::log_structure[] = {
     { LOG_DATA_FLOAT_MSG, sizeof(log_Data_Float),         
       "DFLT",  "QBf",         "TimeUS,Id,Value", "s--", "F--" },
     { LOG_GUIDEDTARGET_MSG, sizeof(log_GuidedTarget),
-      "GUIP",  "QBffffff",    "TimeUS,Type,pX,pY,pZ,vX,vY,vZ", "s-mmmnnn", "F-000000" },
+      "GUIP",  "QBBffffff",    "TimeUS,Type,Frame,pX,pY,pZ,vX,vY,vZ", "s--mmmnnn", "F--000000" },
 };
 
 uint8_t Sub::get_num_log_structures() const
--- a/ArduSub/Sub.cpp
+++ b/ArduSub/Sub.cpp
@@ -465,6 +465,37 @@ void Sub::rc_loop()
 }
 #endif
 
+#if AP_SCRIPTING_ENABLED
+// set target position and velocity in supplied AltFrame
+bool Sub::set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms, Location::AltFrame alt_frame)
+{
+    // exit if the vehicle is not in Guided mode or Auto-Guided mode
+    if (!flightmode->in_guided_mode()) {
+        return false;
+    }
+
+    float dest_u_m = 0.0f;
+    switch (alt_frame) {
+        case Location::AltFrame::ABOVE_HOME:
+            // distance above home (usually at the water surface)
+            dest_u_m = -target_pos_ned_m.z;
+            break;
+        case Location::AltFrame::ABOVE_TERRAIN:
+            // distance above terrain
+            dest_u_m = target_pos_ned_m.z;
+            break;
+        case Location::AltFrame::ABSOLUTE:
+        case Location::AltFrame::ABOVE_ORIGIN:
+            // not supported
+            return false;
+    }
+
+    const Vector3f pos_neu_cm(target_pos_ned_m.x * 100.0f, target_pos_ned_m.y * 100.0f, dest_u_m * 100.0f);
+    const Vector3f vel_neu_cms(target_vel_ned_ms.x * 100.0f, target_vel_ned_ms.y * 100.0f, -target_vel_ned_ms.z * 100.0f);
+    return mode_guided.guided_set_destination_posvel(pos_neu_cm, vel_neu_cms, alt_frame);
+}
+#endif  // AP_SCRIPTING_ENABLED
+
 Sub *Sub::_singleton = nullptr;
 
 Sub sub;
--- a/ArduSub/Sub.h
+++ b/ArduSub/Sub.h
@@ -646,6 +646,9 @@ public:
     float get_rangefinder_target_cm() const WARN_IF_UNUSED { return flightmode->get_rangefinder_target() * 100.0f; }
     bool set_rangefinder_target_cm(float new_target_cm) { return mode_surftrak.set_rangefinder_target_cm(new_target_cm); }
 #endif // AP_RANGEFINDER_ENABLED
+
+    // set target position and velocity
+    bool set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms, Location::AltFrame alt_frame) override;
 #endif // AP_SCRIPTING_ENABLED
 };
 
--- a/ArduSub/Sub.h
+++ b/ArduSub/Sub.h
@@ -417,7 +417,7 @@ private:
     void Log_Write_Data(LogDataID id, int16_t value);
     void Log_Write_Data(LogDataID id, uint16_t value);
     void Log_Write_Data(LogDataID id, float value);
-    void Log_Write_GuidedTarget(uint8_t target_type, const Vector3f& pos_target, const Vector3f& vel_target);
+    void Log_Write_GuidedTarget(uint8_t target_type, const Vector3f& pos_target, const Vector3f& vel_target, Location::AltFrame alt_frame);
     void Log_Write_Vehicle_Startup_Messages();
 #endif
     void load_parameters(void) override;
@@ -643,7 +643,7 @@ public:
     uint8_t get_and_clear_button_count(uint8_t index);
 
 #if AP_RANGEFINDER_ENABLED
-    float get_rangefinder_target_cm() const WARN_IF_UNUSED { return mode_surftrak.get_rangefinder_target_cm(); }
+    float get_rangefinder_target_cm() const WARN_IF_UNUSED { return flightmode->get_rangefinder_target() * 100.0f; }
     bool set_rangefinder_target_cm(float new_target_cm) { return mode_surftrak.set_rangefinder_target_cm(new_target_cm); }
 #endif // AP_RANGEFINDER_ENABLED
 #endif // AP_SCRIPTING_ENABLED
--- a/ArduSub/mode.h
+++ b/ArduSub/mode.h
@@ -68,6 +68,7 @@ public:
     virtual bool allows_arming(bool from_gcs) const = 0;
     virtual bool is_autopilot() const { return false; }
     virtual bool in_guided_mode() const { return false; }
+    virtual float get_rangefinder_target() const { return 0; }
 
     // return a string for this flightmode
     virtual const char *name() const = 0;
@@ -225,7 +226,7 @@ public:
 
     bool init(bool ignore_checks) override;
 
-    float get_rangefinder_target_cm() const WARN_IF_UNUSED { return rangefinder_target_cm; }
+    float get_rangefinder_target() const override WARN_IF_UNUSED { return rangefinder_target_cm * 0.01f; }
     bool set_rangefinder_target_cm(float target_cm);
 
 protected:
@@ -261,12 +262,13 @@ public:
     bool allows_arming(bool from_gcs) const override { return true; }
     bool is_autopilot() const override { return true; }
     bool in_guided_mode() const override { return true; }
+    float get_rangefinder_target() const override WARN_IF_UNUSED;
     bool guided_limit_check();
     void guided_limit_init_time_and_pos();
     void guided_set_angle(const Quaternion &q, float climb_rate_cms, bool use_yaw_rate, float yaw_rate_rads);
     void guided_set_angle(const Quaternion&, float);
     void guided_limit_set(uint32_t timeout_ms, float alt_min_cm, float alt_max_cm, float horiz_max_cm);
-    bool guided_set_destination_posvel(const Vector3f& destination, const Vector3f& velocity);
+    bool guided_set_destination_posvel(const Vector3f& destination, const Vector3f& velocity, Location::AltFrame alt_frame);
     bool guided_set_destination_posvel(const Vector3f& destination, const Vector3f& velocity, bool use_yaw, float yaw_cd, bool use_yaw_rate, float yaw_rate_cds, bool relative_yaw);
     bool guided_set_destination(const Vector3f& destination);
     bool guided_set_destination(const Location&);
@@ -294,7 +296,7 @@ private:
     void guided_takeoff_run();
     void guided_pos_control_start();
     void guided_vel_control_start();
-    void guided_posvel_control_start();
+    void guided_posvel_control_start(Location::AltFrame alt_frame);
     void guided_angle_control_start();
 };
 
--- a/ArduSub/mode_guided.cpp
+++ b/ArduSub/mode_guided.cpp
@@ -9,6 +9,7 @@
 
 static Vector3p posvel_pos_target_cm;
 static Vector3f posvel_vel_target_cms;
+static Location::AltFrame posvel_frame;
 static uint32_t update_time_ms;
 
 struct {
@@ -118,10 +119,11 @@ void ModeGuided::guided_vel_control_start()
 }
 
 // initialise guided mode's posvel controller
-void ModeGuided::guided_posvel_control_start()
+void ModeGuided::guided_posvel_control_start(Location::AltFrame alt_frame)
 {
-    // set guided_mode to velocity controller
+    // set guided_mode to posvel controller
     sub.guided_mode = Guided_PosVel;
+    posvel_frame = alt_frame;
 
     // set vertical speed and acceleration
     // All limits must be positive
@@ -188,7 +190,7 @@ bool ModeGuided::guided_set_destination(const Vector3f& destination)
 
 #if HAL_LOGGING_ENABLED
     // log target
-    sub.Log_Write_GuidedTarget(sub.guided_mode, destination, Vector3f());
+    sub.Log_Write_GuidedTarget(sub.guided_mode, destination, Vector3f(), Location::AltFrame::ABOVE_ORIGIN);
 #endif
 
     return true;
@@ -223,7 +225,7 @@ bool ModeGuided::guided_set_destination(const Location& dest_loc)
 
 #if HAL_LOGGING_ENABLED
     // log target
-    sub.Log_Write_GuidedTarget(sub.guided_mode, Vector3f(dest_loc.lat, dest_loc.lng, dest_loc.alt),Vector3f());
+    sub.Log_Write_GuidedTarget(sub.guided_mode, Vector3f(dest_loc.lat, dest_loc.lng, dest_loc.alt),Vector3f(), Location::AltFrame::ABOVE_ORIGIN);
 #endif
 
     return true;
@@ -259,7 +261,7 @@ bool ModeGuided::guided_set_destination(const Vector3f& destination, bool use_ya
 
 #if HAL_LOGGING_ENABLED
     // log target
-    sub.Log_Write_GuidedTarget(sub.guided_mode, destination, Vector3f());
+    sub.Log_Write_GuidedTarget(sub.guided_mode, destination, Vector3f(), Location::AltFrame::ABOVE_ORIGIN);
 #endif
 
     return true;
@@ -298,11 +300,11 @@ void ModeGuided::guided_set_velocity(const Vector3f& velocity, bool use_yaw, flo
 }
 
 // set guided mode posvel target
-bool ModeGuided::guided_set_destination_posvel(const Vector3f& destination, const Vector3f& velocity)
+bool ModeGuided::guided_set_destination_posvel(const Vector3f& destination, const Vector3f& velocity, Location::AltFrame alt_frame)
 {
 #if AP_FENCE_ENABLED
     // reject destination if outside the fence
-    const Location dest_loc(destination, Location::AltFrame::ABOVE_ORIGIN);
+    const Location dest_loc(destination, alt_frame);
     if (!sub.fence.check_location_within_fence(dest_loc)) {
         LOGGER_WRITE_ERROR(LogErrorSubsystem::NAVIGATION, LogErrorCode::DEST_OUTSIDE_FENCE);
         // failure is propagated to GCS with NAK
@@ -310,23 +312,54 @@ bool ModeGuided::guided_set_destination_posvel(const Vector3f& destination, cons
     }
 #endif
 
-    // check we are in posvel control mode
-    if (sub.guided_mode != Guided_PosVel) {
-        guided_posvel_control_start();
+    // move into Guided_PosVel sub-mode with the specified frame (as necessary)
+    if (sub.guided_mode != Guided_PosVel || posvel_frame != alt_frame) {
+        switch (alt_frame) {
+            case Location::AltFrame::ABOVE_ORIGIN:
+                guided_posvel_control_start(alt_frame);
+                break;
+            case Location::AltFrame::ABOVE_TERRAIN:
+                if (!sub.rangefinder_alt_ok()) {
+                    gcs().send_text(MAV_SEVERITY_WARNING, "Terrain data (rangefinder) not available");
+                    return false;
+                }
+                guided_posvel_control_start(alt_frame);
+
+                // bumpless transfer
+                position_control->init_pos_terrain_D_m((destination.z - sub.rangefinder_state.inertial_alt_cm) * 0.01f);
+                position_control->set_pos_terrain_target_D_m(-sub.rangefinder_state.rangefinder_terrain_offset_cm * 0.01f);
+                break;
+            case Location::AltFrame::ABSOLUTE:
+            case Location::AltFrame::ABOVE_HOME:
+                // not supported
+                return false;
+        }
+    } else {  // already in the intended sub-mode + frame
+        if (alt_frame == Location::AltFrame::ABOVE_TERRAIN) {
+            // rangefinder target may have changed
+            position_control->init_pos_terrain_D_m(position_control->get_pos_terrain_D_m() + (destination.z - posvel_pos_target_cm.z) * 0.01f);
+        }
     }
 
     update_time_ms = AP_HAL::millis();
     posvel_pos_target_cm = destination.topostype();
     posvel_vel_target_cms = velocity;
 
-    position_control->input_pos_vel_accel_NE_cm(posvel_pos_target_cm.xy(), posvel_vel_target_cms.xy(), Vector2f());
-    float dz = posvel_pos_target_cm.z;
-    position_control->input_pos_vel_accel_U_cm(dz, posvel_vel_target_cms.z, 0);
-    posvel_pos_target_cm.z = dz;
+    Vector2p pos_ne_m = posvel_pos_target_cm.xy() * 0.01f;
+    Vector2f vel_ne_ms = posvel_vel_target_cms.xy() * 0.01f;
+    position_control->input_pos_vel_accel_NE_m(pos_ne_m, vel_ne_ms, Vector2f());
+    posvel_pos_target_cm.xy() = pos_ne_m * 100.0f;
+    posvel_vel_target_cms.xy() = vel_ne_ms * 100.0f;
+
+    float pz_d_m = -posvel_pos_target_cm.z * 0.01f;
+    float vz_d_ms = -posvel_vel_target_cms.z * 0.01f;
+    position_control->input_pos_vel_accel_D_m(pz_d_m, vz_d_ms, 0);
+    posvel_pos_target_cm.z = -pz_d_m * 100.0f;
+    posvel_vel_target_cms.z = -vz_d_ms * 100.0f;
 
 #if HAL_LOGGING_ENABLED
     // log target
-    sub.Log_Write_GuidedTarget(sub.guided_mode, destination, velocity);
+    sub.Log_Write_GuidedTarget(sub.guided_mode, destination, velocity, alt_frame);
 #endif
 
     return true;
@@ -347,7 +380,7 @@ bool ModeGuided::guided_set_destination_posvel(const Vector3f& destination, cons
 
     // check we are in posvel control mode
     if (sub.guided_mode != Guided_PosVel) {
-        guided_posvel_control_start();
+        guided_posvel_control_start(Location::AltFrame::ABOVE_ORIGIN);
     }
 
     // set yaw state
@@ -365,7 +398,7 @@ bool ModeGuided::guided_set_destination_posvel(const Vector3f& destination, cons
 
 #if HAL_LOGGING_ENABLED
     // log target
-    sub.Log_Write_GuidedTarget(sub.guided_mode, destination, velocity);
+    sub.Log_Write_GuidedTarget(sub.guided_mode, destination, velocity, Location::AltFrame::ABOVE_ORIGIN);
 #endif
 
     return true;
@@ -641,6 +674,14 @@ void ModeGuided::guided_posvel_control_run()
 
     // send position and velocity targets to position controller
     position_control->input_pos_vel_accel_NE_cm(posvel_pos_target_cm.xy(), posvel_vel_target_cms.xy(), Vector2f());
+
+#if AP_RANGEFINDER_ENABLED
+    if (posvel_frame == Location::AltFrame::ABOVE_TERRAIN && sub.rangefinder_alt_ok()) {
+        // surftrak: set the offset target to the current terrain altitude estimate
+        position_control->set_pos_terrain_target_U_cm(sub.rangefinder_state.rangefinder_terrain_offset_cm);
+    }
+#endif
+
     float pz = posvel_pos_target_cm.z;
     position_control->input_pos_vel_accel_U_cm(pz, posvel_vel_target_cms.z, 0);
     posvel_pos_target_cm.z = pz;
@@ -888,4 +929,13 @@ bool ModeGuided::guided_limit_check()
 
     // if we got this far we must be within limits
     return false;
+}
+
+// returns target for rangefinder, or 0 if there is no rangefinder target
+float ModeGuided::get_rangefinder_target() const
+{
+    if (sub.guided_mode == Guided_PosVel && posvel_frame == Location::AltFrame::ABOVE_TERRAIN) {
+        return posvel_pos_target_cm.z * 0.01f;
+    }
+    return 0.0f;
 }
--- a/Tools/autotest/ardusub.py
+++ b/Tools/autotest/ardusub.py
@@ -7,7 +7,9 @@ Parameters are in-code defaults plus default_params/sub.parm
 AP_FLAKE8_CLEAN
 '''
 
+import math
 import os
+import re
 
 from math import degrees
 from math import radians
@@ -17,12 +19,16 @@ from pymavlink import mavutil
 
 import vehicle_test_suite
 
+from pysim import util
 from vehicle_test_suite import NotAchievedException
 
 # get location of scripts
 testdir = os.path.dirname(os.path.realpath(__file__))
 
 SITL_START_LOCATION = mavutil.location(33.810313, -118.393867, 0, 185)
+
+# Extract true range from STATUSTEXT messages sent by sub_test_synthetic_seafloor.lua
+RE_TR_SEARCH = re.compile(r'#TR#\s*([-+]?\d*\.?\d+)')
 
 
 class Joystick():
@@ -373,13 +379,11 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
 
         def get_true_distance():
             """Return the True distance from the simulated range finder"""
-            m_true = self.mav.recv_match(type='STATUSTEXT', blocking=True, timeout=3.0)
-            if m_true is None:
-                return m_true
-            idx_tr = m_true.text.find('#TR#')
-            if idx_tr < 0:
+            m_true = self.assert_receive_message('STATUSTEXT', timeout=3.0)
+            match = RE_TR_SEARCH.search(m_true.text)
+            if not match:
                 return None
-            return float(m_true.text[(idx_tr+4):(idx_tr+12)])
+            return float(match.group(1))
 
         tstart = self.get_sim_time_cached()
         self.progress('Distance to be watched: %.2f (+/- %.2f)' % (match_distance, delta))
@@ -420,7 +424,7 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
         end_altitude = start_altitude - 10
         validation_delta = 1.5  # Largest allowed distance between sub height and desired height
 
-        self.context_push()
+        # Load the synthetic seafloor, this will push a context and reboot
         self.prepare_synthetic_seafloor_test(sea_floor_depth, match_distance)
 
         # Dive to match_distance off the bottom in preparation for the mission
@@ -452,9 +456,8 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
 
         self.set_rc(Joystick.Forward, 1500)
 
+        # Disarm before context pop
         self.disarm_vehicle()
-        self.context_pop()
-        self.reboot_sitl()  # e.g. revert rangefinder configuration
 
     def SimTerrainMission(self):
         """Mission at a constant height above synthetic sea floor"""
@@ -465,7 +468,7 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
         end_altitude = start_altitude - 10
         validation_delta = 1.5  # Largest allowed distance between sub height and desired height
 
-        self.context_push()
+        # Load the synthetic seafloor, this will push a context and reboot
         self.prepare_synthetic_seafloor_test(sea_floor_depth, match_distance)
 
         # The synthetic seafloor has an east-west ridge south of the sub.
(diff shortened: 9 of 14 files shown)


timtuxworth:
Very interesting I'm working on something similar for Plane.

clydemcqueen:
@timtuxworth Thanks for reaching out! I will read your PR and try to stay in sync where I can.

My use case doesn't depend on a GCS, instead I am using a Lua script to send targets at 10Hz. This allows the sub pilot to control heading while GUIDED mode handles the xyz motion. Here is the script that I am testing: https://github.com/clydemcqueen/ardusub_surftrak/blob/guided_above_terrain/lua/transect.lua

ArduSub/Log.cpp
```diff
@@ -178,6 +178,7 @@ struct PACKED log_GuidedTarget {
     LOG_PACKET_HEADER;
     uint64_t time_us;
     uint8_t type;
+    uint8_t alt_frame;
```
timtuxworth, line 181: Have you thought about making this a Location::AltFrame instead of uint8_t?

clydemcqueen: Hmmm, will all compilers map Location::AltFrame to uint8_t? I don't see this usage anywhere else in the code.

clydemcqueen:
We tested this w/ an A50 DVL off Seattle's Pier 62 on 21-Aug-2024 with good success. Learnings:
* We forgot to set WPNAV_ACCEL_Z to to match PILOT_ACCEL_Z, so we had some strong up/down oscillation due to sonar delay. (SURFTRAK performed as expected since it uses PILOT_ACCEL_Z).
* We learned that the lower limit for WPNAV_SPEED is 20 cm/s, vs the desired 10 cm/s. This has now been changed in master.
* The script tried to set the target position even when the rangefinder was unhealthy, resulting in a bunch of warnings in the GCS. (It looks like mavlink-router drops some messages, so the rangefinder is unhealthy more often than I expected, see https://github.com/bluerobotics/BlueOS-Water-Linked-DVL/issues/44.) I updated the script to avoid generating warnings.

We will test again in a few weeks.

ArduSub/Log.cpp
timtuxworth: I've used a static cast

clydemcqueen:
We tested this on a different ROV with an A50 DVL off a vessel in Elliott Bay. I still saw vertical oscillations in guided mode compared to surftrak. I tracked this down how `ModeGuided::guided_set_destination_posvel` was calling `position_control->set_pos_offset_z_cm`. I changed the code to distinguish between the first call and subsequent calls.

We will test this again in the coming weeks.

clydemcqueen:
I am finally returning to this project after a long break!

The code has been substantially re-written and there is a much better autotest.

We just had a successful test off Pier 59 in Seattle. This includes the code in this PR and a Lua script that implements a virtual target following algorithm that we're calling "constant velocity SURFTRAK". I uploaded the Lua script and a partial BIN log [here](https://github.com/clydemcqueen/guided_above_terrain).

I'm marking this as "ready for review" though I know Willian is on vacation for another week. Thanks!

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -157,8 +157,10 @@ bool GCS_MAVLINK_Sub::send_info()
     CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
     send_named_float("RollPitch", sub.roll_pitch_flag);
 
-    CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
-    send_named_float("RFTarget", sub.mode_surftrak.get_rangefinder_target_cm() * 0.01f);
+    if (sub.control_mode == Mode::Number::SURFTRAK) {
+        CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
+        send_named_float("RFTarget", sub.mode_surftrak.get_rangefinder_target_cm() * 0.01f);
```
clydemcqueen: The value `sub.mode_surftrak.get_rangefinder_target_cm()` is only valid if we're in SURFTRAK mode.

It probably makes sense to also publish the rangefinder target in GUIDED mode iff submode == Guided_PosVel and posvel_frame == ABOVE_TERRAIN. This will require a bit more refactoring of mode_guided to expose these values. Thoughts from reviewers?

ES-Alexander: Seems reasonable to me - if the value is used in multiple situations but only exposed in a subset of them then that seems likely to lead to user confusion.

Longer term it should perhaps be transferred to some kind of dedicated MAVLink message (rather than NAMED_VALUEs), but that's not a priority at this stage.

ES-Alexander:
From a brief look, the code seems already to me.

Not sure how best this can be tested, although I suppose a SITL test should at least be possible

clydemcqueen:
Thanks @ES-Alexander !

These are the tests I have done:
* `test.Sub.GuidedPosVel` uses a Lua script to simulate a rising and falling seafloor. The test alternates between ABOVE_ORIGIN and ABOVE_TERRAIN frames. See the graph below.
* I did a bunch of SITL tests, this has a flat seafloor at -50m so not super interesting
* I did a bunch SITL + Gazebo tests using these models: https://github.com/clydemcqueen/bluerov2_gz/
* We recently did a live test with a custom Lua script: https://github.com/clydemcqueen/guided_above_terrain

Let me know if there is anything else!

image

ArduSub/Sub.cpp
```diff
@@ -526,6 +526,37 @@ void Sub::rc_loop()
...
+        return false;
+    }
+
+    // pos.z is rf target, do not flip sign
+    const Vector3f pos_neu_cm(target_pos.x * 100.0f, target_pos.y * 100.0f, target_pos.z * 100.0f);
+    const Vector3f vel_neu_cms(target_vel.x * 100.0f, target_vel.y * 100.0f, -target_vel.z * 100.0f);
+
+    return mode_guided.guided_set_destination_posvel(pos_neu_cm, vel_neu_cms, Location::AltFrame::ABOVE_TERRAIN);
+}
+#endif
```
ES-Alexander:
```suggestion
#endif  // AP_SCRIPTING_ENABLED
```
I think there's enough code between the tags that it's useful to specify what is being ended, especially since it's not trivially obvious these functions are only relevant to / used in Lua scripts.

Tools/autotest/ardusub.py
```diff
@@ -468,6 +469,133 @@ def SimTerrainMission(self):
...
+                       mavutil.mavlink.POSITION_TARGET_TYPEMASK_AY_IGNORE |
+                       mavutil.mavlink.POSITION_TARGET_TYPEMASK_AZ_IGNORE |
+                       mavutil.mavlink.POSITION_TARGET_TYPEMASK_YAW_IGNORE |
+                       mavutil.mavlink.POSITION_TARGET_TYPEMASK_YAW_RATE_IGNORE)
+
+        # Everything in meters and meters/second unless noted
+        seafloor_depth = 50
+        speed = 0.5
+
+        # Generate a synthetic seafloor at -50m
```
ES-Alexander: Comments shouldn't include variable values (to avoid extra maintenance load).

More generally this comment (if it should exist at all) is likely better placed as a section descriptor, above the seafloor_depth variable definition.

Tools/autotest/ardusub.py
```diff
@@ -468,6 +469,133 @@ def SimTerrainMission(self):
...
+        self.context_push()
+        self.prepare_synthetic_seafloor_test(seafloor_depth, 10)  # rf_target is not used
+
+        # Guided_PosVel uses WPNAV_SPEED
```
ES-Alexander:
```suggestion
        # Guided_PosVel uses WPNAV_SPEED (cm/s)
```
Useful to explain that the value adjustment is a unit conversion (e.g. instead of compensation for the simulation rate, or some other situation).

Tools/autotest/ardusub.py
```diff
@@ -468,6 +469,133 @@ def SimTerrainMission(self):
...
+
+        # Run between 2 locations, 30m apart
+        distance = 30
+        timeout = distance / speed + 5  # Add a little time to accelerate, etc.
+
+        runs = [{
+            # Hold depth at -15m as the terrain rises 10m
+            'frame': mavutil.mavlink.MAV_FRAME_GLOBAL_RELATIVE_ALT,
+            'bearing': 180,
+            'target_alt': -15,  # Altitude
```
ES-Alexander: Could be worth explaining [which altitude this is](https://ardupilot.org/copter/docs/common-understanding-altitude.html), although admittedly I'm having a hard time coming up with a concise way of saying that it effectively resolves to ASL, though is technically a value for altitude "above" HOME (if I'm understanding correctly).

Tools/autotest/ardusub.py
```diff
@@ -468,6 +469,133 @@ def SimTerrainMission(self):
...
+            'max_error_allowed': 1.0,
+        }, {
+            # Hold range at 25m as the terrain falls by 10m
+            'frame': mavutil.mavlink.MAV_FRAME_GLOBAL_TERRAIN_ALT,
+            'bearing': 0,
+            'target_alt': 25,  # Distance above seafloor
```
ES-Alexander:
```suggestion
            'target_alt': 25,  # Distance above seafloor (AGL)
```

Tools/autotest/ardusub.py
```diff
@@ -468,6 +469,133 @@ def SimTerrainMission(self):
...
+                    self.progress('Frame %u reached destination at time %f, max_error %f' %
+                                  (run['frame'], self.get_sim_time_cached() - start_time, max_error))
+                    break
+                elif self.get_sim_time_cached() - start_time > timeout:
+                    raise NotAchievedException('Frame %u took too long to reach the destination' % run['frame'])
+
+                if current_alt is None:
+                    continue
+
+                alt_error = abs(current_alt - run['target_alt'])
```
ES-Alexander, line 592: This is within the frame being considered. Could be valuable to also check the difference between the rangefinder and EKF (baro?) estimates to better ensure the terrain is actually being followed.

ES-Alexander (approved):
From a more thorough look, the approach and code both look ok to me, with just some minor suggestions for improvement/clarity/maintainability - nice! :D

Noting that I'm not familiar with these parts of the codebase, so I've primarily reviewed the code that is shown, and can't really comment on how well this new code fits into the existing functionality, or whether there are regions that should have been touched but have been overlooked.

ArduSub/Sub.cpp
clydemcqueen: Fixed

Tools/autotest/ardusub.py
clydemcqueen: Removed comment

I also found other comments where I included numbers and cleaned them up

Tools/autotest/ardusub.py
clydemcqueen: Fixed

ArduSub/mode.h
```diff
@@ -267,7 +267,9 @@ class ModeGuided : public Mode
     void guided_set_angle(const Quaternion &q, float climb_rate_cms, bool use_yaw_rate, float yaw_rate_rads);
     void guided_set_angle(const Quaternion&, float);
     void guided_limit_set(uint32_t timeout_ms, float alt_min_cm, float alt_max_cm, float horiz_max_cm);
-    bool guided_set_destination_posvel(const Vector3f& destination, const Vector3f& velocity);
+    void set_posvel_frame_above_home();
+    void set_posvel_frame_above_terrain(float destination_z);
```
clydemcqueen: Whoops, `set_posvel_frame_above_home` and `set_posvel_frame_above_terrain` are from a previous iteration. Removed.

Tools/autotest/ardusub.py
clydemcqueen: Yeah, AFAIK sub basically forces origin.z to be 0 at the "surface", which is ASL most places but could easily be the surface of a lake in the mountains. This doesn't really jive with AP nomenclature, nor does it match MAVLink naming. E.g., `MAV_FRAME_GLOBAL_RELATIVE_ALT` is defined as "above home" but the sub mode_guided code uses `Location::AltFrame::ABOVE_ORIGIN`.

I'll label these "# Altitude above origin" and "# Distance to seafloor".

Tools/autotest/ardusub.py
clydemcqueen: See above

Tools/autotest/ardusub.py
clydemcqueen: Hmmm, IIUC, we are comparing the range target to the ground truth `true_range_m` value from the Lua script, so we know the terrain is being followed.

I could also do this by looking at DISTANCE_SENSOR messages, but the terrain generator adds some noise and dropouts, and so some DISTANCE_SENSOR messages might have weird values. I just tried this and the test fails due to noise. It passes with configuration 1 (no noise) in `prepare_synthetic_seafloor_test`. LMK what is preferred, I can go either way.

Using a Lua script to simulate terrain and a noisy rangefinder is cumbersome. It would be nice to have a terrain generator in SimSubmarine.cpp.

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: Fixed

clydemcqueen:
Thanks for the review, @ES-Alexander!

ArduSub/mode_guided.cpp
```diff
@@ -73,7 +74,6 @@ autopilot_yaw_mode ModeGuided::get_default_auto_yaw_mode(bool rtl) const
     }
 }
 
-
```
ES-Alexander, line 76: Unexpected whitespace change. IIRC ArduPilot prefers them to be in dedicated non-functionality commits, and only exist if there's a reason for the change.

Tools/autotest/ardusub.py
ES-Alexander: > ... we are comparing the range target to the ground truth true_range_m value from the Lua script, so we know the terrain is being followed.
> 
> I could also do this by looking at DISTANCE_SENSOR messages ...

I was more thinking of using the `GLOBAL_POSITION_INT` altitude to confirm the vehicle did actually move vertically when it was expected to follow terrain, rather than maintaining its ASL altitude.

> It would be nice to have a terrain generator in SimSubmarine.cpp.

Fair enough. Perhaps out of scope for this PR, but a simple approach could be to have a sinusoid with a 1m period, with a globally linear slope in one direction? That way moving by integer meters horizontally should keep the altitude the same, except for that linear slope, but there's also some simple variation in between. Nice and predictable, with some spice.

Something like
```
depth = 50 + sin(x/(2*pi))*sin(y/(2*pi)) + 0.1*x
```

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -157,8 +157,13 @@ bool GCS_MAVLINK_Sub::send_info()
...
+    } else if (sub.control_mode == Mode::Number::GUIDED) {
+        CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
+        send_named_float("RFTarget", sub.mode_guided.get_rangefinder_target_cm() * 0.01f);
```
ES-Alexander: Maybe worth a comment like "sends 0 in unsupported sub-modes"? Not super fussed either way.

ES-Alexander (approved):

ArduSub/mode_guided.cpp
clydemcqueen: Fixed

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: Fixed

vshie:
This is exciting! My ROV with Cerulean DVL and ROVL system is currently being re-worked, but I'd love to test this in a week or two!

clydemcqueen:
> This is exciting! My ROV with Cerulean DVL and ROVL system is currently being re-worked, but I'd love to test this in a week or two!

@vshie  Hey, Tony -- here are the parameters that we are using:

```
WPNAV_SPEED     20      # move forward at 20 cms
WPNAV_SPEED_UP  30      # ascend at 30 cms
WPNAV_SPEED_DN  30      # descend at 30 cms

PSC_JERK_Z 8.0          # force KPa = 1.6, KPv = 0.8
PILOT_ACCEL_Z 500
WPNAV_ACCEL_Z 500
```

IMHO the WPNAV defaults are too high; you don't want the sub to dive at 1 m/s.

The changes to jerk and accel params are to deal with sonar delay. You'll need to use PILOT params for SURFTRAK, but WPNAV params for GUIDED.

Also, there was a recent AP_Math PR that merged recently, you might want to set jerk to 4.0 if you still see vertical oscillations.

I am eager to see the results!

Update: WPNAV_SPEED* is now WP_SPD*, and are in m/s, not cm/s

Williangalvani (approved):

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2768,6 +2768,12 @@ function vehicle:set_target_posvelaccel_NED(target_pos, target_vel, target_accel
 ---@return boolean
 function vehicle:set_target_posvel_NED(target_pos, target_vel) end
 
+-- Sets the target position and velocity in a guided mode, with ABOVE_TERRAIN frame.
```
ES-Alexander: Peter Barker has suggested "Why not a binding which takes a frame?"

clydemcqueen: I agree, adding a frame parameter is a bit more future-proof. There is a nice pattern in https://github.com/ArduPilot/ardupilot/pull/25814 that I can follow.

clydemcqueen: Fixed

clydemcqueen:
Changes:
* removed `set_target_posvel_NED_terrain` scripting method
* added an optional `alt_frame` param to the `set_target_posvel_NED` scripting method, the default is ABOVE_ORIGIN
* rebased to pick up changes to WPNAV parameter names and units

ArduCopter/Copter.cpp
```diff
@@ -320,7 +320,7 @@ bool Copter::set_target_pos_NED(const Vector3f& target_pos_ned_m, bool use_yaw,
 }
 
 // set target position and velocity (for use by scripting)
-bool Copter::set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms)
+bool Copter::set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms, Location::AltFrame alt_frame)
```
peterbarker, line 323: I don't think this is valid - we can't take an alt_frame parameter and ignore it on Copter, thinks will start to fly into the ground.

You can just reject the command if the `alt_frame` isn't above-origin, 'though....

clydemcqueen: Fixed

clydemcqueen:
Changes:
* reject call to `Copter::set_target_posvel_NED` if frame is not ABOVE_ORIGIN
* rebased

ES-Alexander:
To me this feels close to going in :-)
Possibly just waiting for @lthall, since he self-requested a review?

Have you done any testing with the updated Lua function?

clydemcqueen:
> To me this feels close to going in :-)

Hi @ES-Alexander -- that is good news!

> Have you done any testing with the updated Lua function?

Do you mean the test we ran on 7-Jan-2026? This is the script we tested: https://github.com/clydemcqueen/guided_above_terrain/blob/main/lua/transect3.lua

I am happy to schedule a test of 4.7-BETA once that is live.

ES-Alexander:
Was just asking whether the updated `set_target_posvel_NED` Lua function had been tested to work as expected, after the transition from the initial `set_target_posvel_NED_terrain` approach (since I know I haven't tested it 😅).

clydemcqueen:
@ES-Alexander Good catch. I updated my scripts and re-ran some tests. Everything works well! See https://github.com/clydemcqueen/guided_above_terrain/pull/1

clydemcqueen:
Rebased

lthall (commented):
I am not a blocker for this.

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -166,8 +166,14 @@ bool GCS_MAVLINK_Sub::send_info()
     CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
     send_named_float("RollPitch", sub.roll_pitch_flag);
 
-    CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
-    send_named_float("RFTarget", sub.mode_surftrak.get_rangefinder_target_cm() * 0.01f);
+    if (sub.control_mode == Mode::Number::SURFTRAK) {
```
peterbarker: So 3 things on this line.

In Copter we don't check the control mode number, we check the pointer to the current flight mode object instead.

The other thing we do is not check the mode or mode number but *ask* the current flightmode if we should do something.  So in this case it might be 

```suggestion
    if (sub.flightmode->should_send_RFTarget()) {
```

The third is that we should absolutely not be introducing new "standards" for GCSs to consume out of `NAMED_VALUE_FLOAT` messages.  Good for diagnostics, fantastic for local use - but quite rude as a mechanism put forward to get information to e.g. QGroundControl.  We have the mavlink message set for a reason - so people can discuss and come to a consensus on things.  I'm not blocking, I'm just saying this isn't the right way in the least to get data to a GCS on a permanent basis.

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -166,8 +166,14 @@ bool GCS_MAVLINK_Sub::send_info()
...
+        CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
+        send_named_float("RFTarget", sub.mode_surftrak.get_rangefinder_target_cm() * 0.01f);
+    } else if (sub.control_mode == Mode::Number::GUIDED) {
+        CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
+        // Sends 0 in unsupported sub-modes
```
peterbarker: So this is wrong :-)

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -166,8 +166,14 @@ bool GCS_MAVLINK_Sub::send_info()
     CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
     send_named_float("RollPitch", sub.roll_pitch_flag);
 
-    CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
-    send_named_float("RFTarget", sub.mode_surftrak.get_rangefinder_target_cm() * 0.01f);
+    if (sub.control_mode == Mode::Number::SURFTRAK) {
+        CHECK_PAYLOAD_SIZE(NAMED_VALUE_FLOAT);
+        send_named_float("RFTarget", sub.mode_surftrak.get_rangefinder_target_cm() * 0.01f);
```
peterbarker: why not just a call onto `sub.flightmode`?

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -727,6 +733,17 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
                 int32_t(packet.alt*100),
                 frame,
             };
+
+            // posvel case supports ABOVE_TERRAIN frame
+            if (!vel_ignore && acc_ignore && frame == Location::AltFrame::ABOVE_TERRAIN) {
+                if (!loc.get_vector_xy_from_origin_NE_cm(pos_neu_cm.xy())) {
+                    break;
+                }
+                pos_neu_cm.z = loc.alt;
```
peterbarker, line 738: Why not `get_vector_from_origin_NED_cm`?!

Generally avoid taking `.alt` from a location object.  We have accessor for it now which we should be using.

ArduSub/mode.h
```diff
@@ -278,6 +278,7 @@ class ModeGuided : public Mode
     float get_auto_heading();
     void guided_limit_clear();
     void set_auto_yaw_mode(autopilot_yaw_mode yaw_mode);
+    float get_rangefinder_target_cm() const WARN_IF_UNUSED;
```
peterbarker: Consider
```suggestion
    float get_rangefinder_target() const WARN_IF_UNUSED;
```

... we're moving to metres generally in AP.  Not a blocker, just a suggestion.

ArduSub/mode_guided.cpp
```diff
@@ -298,21 +299,40 @@ void ModeGuided::guided_set_velocity(const Vector3f& velocity, bool use_yaw, flo
...
         return false;
     }
 #endif
 
-    // check we are in posvel control mode
-    if (sub.guided_mode != Guided_PosVel) {
-        guided_posvel_control_start();
+    // ensure we are in Guided_PosVel sub-mode and the frames match
+    if (sub.guided_mode != Guided_PosVel || posvel_frame != alt_frame)
+    {
```
peterbarker:
```suggestion
    if (sub.guided_mode != Guided_PosVel || posvel_frame != alt_frame) {
```

ArduSub/mode_guided.cpp
```diff
@@ -298,21 +299,40 @@ void ModeGuided::guided_set_velocity(const Vector3f& velocity, bool use_yaw, flo
...
+            // bumpless transfer: _pos_desired_neu_m.z = destination.z
+            position_control->init_pos_terrain_U_cm(sub.rangefinder_state.inertial_alt_cm - destination.z);
+            position_control->set_pos_terrain_target_U_cm(sub.rangefinder_state.rangefinder_terrain_offset_cm);
+        } else {
+            guided_posvel_control_start(alt_frame);
+        }
+    } else {
+        if (alt_frame == Location::AltFrame::ABOVE_TERRAIN) {
+            // rangefinder target may have changed: _pos_desired_neu_m.z -= delta
+            position_control->init_pos_terrain_U_cm(-position_control->get_pos_terrain_D_m() * 100.0 + posvel_pos_target_cm.z - destination.z);
```
peterbarker: Consider making all new calls into the `_m` rather than `_cm` methods so you gradually move to the new methods

ArduSub/Sub.cpp
```diff
@@ -468,6 +468,24 @@ void Sub::rc_loop()
...
+// set target position and velocity, alt frame is Location::AltFrame::ABOVE_ORIGIN
+bool Sub::set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms, Location::AltFrame alt_frame)
+{
+    // exit if the vehicle is not in Guided mode or Auto-Guided mode
+    if (!flightmode->in_guided_mode()) {
+        return false;
+    }
+
+    // dest_z_m is either distance above the origin (the water surface) or the distance to the seafloor, depending on the altitude frame
+    auto dest_z_m = alt_frame == Location::AltFrame::ABOVE_ORIGIN ? -target_pos_ned_m.z : target_pos_ned_m.z;
```
peterbarker:
```suggestion
    const auto dest_z_m = alt_frame == Location::AltFrame::ABOVE_ORIGIN ? -target_pos_ned_m.z : target_pos_ned_m.z;
```

Tools/autotest/ardusub.py
```diff
@@ -510,6 +511,132 @@ def SimTerrainMission(self):
...
+
+        # Dive to starting altitude
+        start_altitude = -15
+        pwm = 1300 if self.get_altitude(relative=True) > start_altitude else 1700
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
+        self.set_rc(Joystick.Throttle, pwm)
+        self.wait_altitude(altitude_min=start_altitude-1, altitude_max=start_altitude, relative=False, timeout=120)
+        self.set_rc(Joystick.Throttle, 1500)
+        self.delay_sim_time(1)
```
peterbarker, line 546: Consider creating the equivalent of `self.takeoff(...)` which we use in Plane and Copter to good effect.

Tools/autotest/ardusub.py
```diff
@@ -510,6 +511,132 @@ def SimTerrainMission(self):
...
+            'bearing': 0,
+            'target_alt': 15,  # Distance to seafloor
+            'max_error_allowed': 1.0,
+        }]
+
+        # Stay in GUIDED mode for the duration
+        self.change_mode('GUIDED')
+
+        for run in runs:
+            msg = self.mav.recv_match(type='GLOBAL_POSITION_INT', blocking=True)
```
peterbarker:
```suggestion
            msg = self.assert_receive_message('GLOBAL_POSITION_INT')
```
(and elsewhere)

Tools/autotest/ardusub.py
```diff
@@ -510,6 +511,132 @@ def SimTerrainMission(self):
...
+
+            # Go!
+            start_time = self.get_sim_time()
+            while msg := self.mav.recv_match(type=['GLOBAL_POSITION_INT', 'STATUSTEXT'], blocking=True):
+                # Get ground truth (sans noise) from the terrain generator
+                if msg.get_type() == 'STATUSTEXT':
+                    if run['frame'] == mavutil.mavlink.MAV_FRAME_GLOBAL_TERRAIN_ALT:
+                        idx_tr = msg.text.find('#TR#')
+                        if idx_tr > 0:
+                            current_alt = float(msg.text[(idx_tr + 4):(idx_tr + 12)])
```
peterbarker: Maybe use a regex?

ArduSub/GCS_MAVLink_Sub.cpp
peterbarker: ...  you need these for the test - consider wrapping them in an #ifdef for SITL-only.

peterbarker (changes requested):

clydemcqueen:
@peterbarker Huge thanks for the detailed and thoughtful review. I'll hit each one of the comments.

ArduSub/mode_guided.cpp
clydemcqueen: I'll update to use the latest functions.

Tools/autotest/ardusub.py
clydemcqueen: This is a very good idea, but it will touch a dozen spots. I'll draft a separate PR.

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: I am overloading "pos_neu_cm.z" to mean "target rangefinder" when frame == ABOVE_TERRAIN.

I could add some comments, etc. Thoughts?

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: The comment is wonky; it means "in Sub mode GUIDED, there are submodes, and some of the submodes send 0."

I will switch to `sub.flightmode->get_rangefinder_target()` across all modes, so this is moot anyway.

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: Points 1 & 2: I will switch to a `sub.flightmode->get_rangefinder_target()` so this pushes the per-mode logic to the modes.

Point 3: Ack. These have been around for a bunch of years, but I agree that this is not ideal. I thought that there was an issue on this topic, but I can't find it.

clydemcqueen:
Friendly ping @peterbarker -- anything else?

Tools/autotest/ardusub.py
clydemcqueen: Done

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: I rewrote this to avoid accessing Location members directly, to avoid overloading variables like `pos_neu_cm`, and to make it more obvious how the various options map to GUIDED sub-modes. There should be no change to existing functionality. Please take another look.

ArduCopter/Copter.h
```diff
@@ -150,6 +150,9 @@
 
 #if AP_SCRIPTING_ENABLED
 #include <AP_Scripting/AP_Scripting.h>
+#if MODE_GUIDED_ENABLED
+#include <AP_Common/Location.h>
+#endif
```
peterbarker:
```suggestion
#include <AP_Scripting/AP_Scripting.h>
```

already included

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -721,23 +729,33 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
...
 
-        if (!pos_ignore && !vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_destination_posvel(pos_neu_cm, Vector3f(packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f));
-        } else if (pos_ignore && !vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_velocity(Vector3f(packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f));
-        } else if (!pos_ignore && vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_destination(pos_neu_cm);
+            const Vector3f vel_neu_cm{packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f};
+
+            if (frame == Location::AltFrame::ABOVE_TERRAIN) {
```
peterbarker:
```suggestion
            switch (frame) {
            case Location::AltFrame::ABOVE_TERRAIN:
```

... you shouldn't be treating anything which isn't nav-terrain as above origin - just reject :-)

ArduSub/mode_guided.cpp
```diff
@@ -298,35 +299,60 @@ void ModeGuided::guided_set_velocity(const Vector3f& velocity, bool use_yaw, flo
...
 
     update_time_ms = AP_HAL::millis();
     posvel_pos_target_cm = destination.topostype();
     posvel_vel_target_cms = velocity;
 
-    position_control->input_pos_vel_accel_NE_cm(posvel_pos_target_cm.xy(), posvel_vel_target_cms.xy(), Vector2f());
-    float dz = posvel_pos_target_cm.z;
-    position_control->input_pos_vel_accel_U_cm(dz, posvel_vel_target_cms.z, 0);
-    posvel_pos_target_cm.z = dz;
+    Vector2p pos_ne_m = posvel_pos_target_cm.xy() * 0.01f;
```
peterbarker, line 348:
```suggestion
    const Vector2p pos_ne_m = posvel_pos_target_cm.xy() * 0.01f;
```

ArduSub/mode_guided.cpp
```diff
@@ -298,35 +299,60 @@ void ModeGuided::guided_set_velocity(const Vector3f& velocity, bool use_yaw, flo
...
+    Vector2f vel_ne_ms = posvel_vel_target_cms.xy() * 0.01f;
```
peterbarker, line 349:
```suggestion
    const Vector2f vel_ne_ms = posvel_vel_target_cms.xy() * 0.01f;
```

Note that if you were to do this properly like this:


```suggestion
    const Vector2f vel_ne_ms { posvel_vel_target_cms.xy() * 0.01f };
```

then you would probably get a compilation failure as you're losing precision through here!


... similarly elsewhere here on both the `const` thing *and* the losing precision thing

ArduSub/Sub.cpp
```diff
@@ -466,6 +466,24 @@ void Sub::rc_loop()
...
+// set target position and velocity, alt frame is Location::AltFrame::ABOVE_ORIGIN
+bool Sub::set_target_posvel_NED(const Vector3f& target_pos_ned_m, const Vector3f& target_vel_ned_ms, Location::AltFrame alt_frame)
+{
+    // exit if the vehicle is not in Guided mode or Auto-Guided mode
+    if (!flightmode->in_guided_mode()) {
+        return false;
+    }
+
+    // dest_z_m is either distance above the origin (the water surface) or the distance to the seafloor, depending on the altitude frame
+    const auto dest_z_m = alt_frame == Location::AltFrame::ABOVE_ORIGIN ? -target_pos_ned_m.z : target_pos_ned_m.z;
```
peterbarker: Don't assume alt_frame is one of two values!  Assume that (a) callers will screw up and (b) at some stage in the future someone will want to navigate AMSL for some reason and you want a path towards doing that without screwing over the people from (a)

ArduSub/Sub.h
```diff
@@ -425,7 +425,8 @@ class Sub : public AP_Vehicle {
     void Log_Write_Data(LogDataID id, int16_t value);
     void Log_Write_Data(LogDataID id, uint16_t value);
     void Log_Write_Data(LogDataID id, float value);
-    void Log_Write_GuidedTarget(uint8_t target_type, const Vector3f& pos_target, const Vector3f& vel_target);
+    void Log_Write_GuidedTarget(uint8_t target_type, const Vector3f& pos_target, const Vector3f& vel_target,
```
peterbarker: I suggest not defaulting here and make the callers supply the value.

ArduSub/Sub.h
```diff
@@ -651,9 +652,12 @@ class Sub : public AP_Vehicle {
     uint8_t get_and_clear_button_count(uint8_t index);
 
 #if AP_RANGEFINDER_ENABLED
-    float get_rangefinder_target_cm() const WARN_IF_UNUSED { return mode_surftrak.get_rangefinder_target_cm(); }
+    float get_rangefinder_target_cm() const WARN_IF_UNUSED { return flightmode->get_rangefinder_target() * 100.0f; }
```
peterbarker, line 646: This is a nice gradual improvement.

Tools/autotest/ardusub.py
```diff
@@ -372,13 +378,11 @@ def watch_true_distance_maintained(self, match_distance, delta=0.3, timeout=5.0,
 
         def get_true_distance():
             """Return the True distance from the simulated range finder"""
-            m_true = self.mav.recv_match(type='STATUSTEXT', blocking=True, timeout=3.0)
-            if m_true is None:
-                return m_true
-            idx_tr = m_true.text.find('#TR#')
-            if idx_tr < 0:
+            m_true = self.assert_receive_message('STATUSTEXT', timeout=3.0)
```
peterbarker, line 382: Not really sure what "true" means here - but nice improvement here.

Tools/autotest/ardusub.py
```diff
@@ -487,6 +491,126 @@ def SimTerrainMission(self):
...
+                target_vel = [speed * math.cos(math.radians(run['bearing'])), speed * math.sin(math.radians(run['bearing']))]
+
+                self.mav.mav.set_position_target_global_int_send(
+                    0, 1, 1, run['frame'], posvel_mode,
+                    int(target_loc[0] * 1e7), int(target_loc[1] * 1e7), run['target_alt'],
+                    target_vel[0], target_vel[1], 0, 0, 0, 0, 0, 0)
+
+        self.disarm_vehicle()
+        self.context_pop()
+        self.reboot_sitl()  # e.g. revert rangefinder configuration
```
peterbarker: I don't think you need this or the context; you get an implicit context as part of your test, and if you reboot within that context then we reboot the vehicle after popping the context.

Basically to remove this sort of boilerplate.

I'm not even seeing you reboot the vehicle to set the rangefinder config up, so not sure why you'd need to reboot to turn it off...

peterbarker (changes requested):

ArduSub/mode_guided.cpp
clydemcqueen: These can't be const: `input_pos_vel_accel_NE_m` and `input_pos_vel_accel_D_m` need non-const references.

This will get cleaned up: https://github.com/ArduPilot/ardupilot/issues/32876

ArduSub/mode_guided.cpp
clydemcqueen: Ditto

Tools/autotest/ardusub.py
clydemcqueen: There is a context push and reboot in `prepare_synthetic_seafloor_test`.

Good to know on the automatic reboot and context pop! I updated all of the tests that use `prepare_synthetic_seafloor_test`.

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: Done. Noting that this could cause existing [buggy] scripts to fail.

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -562,12 +564,12 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
...
         // a RC override message is considered to be a 'heartbeat'
         // from the ground station for failsafe purposes
-        
+
         handle_rc_channels_override(msg);
         break;
     }
 
-    
+
```
ES-Alexander: Here

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -818,9 +843,9 @@ uint8_t GCS_MAVLINK_Sub::high_latency_tgt_heading() const
         // need to convert -18000->18000 to 0->360/2
         return wrap_360_cd(sub.wp_nav.get_wp_bearing_to_destination_cd()) / 200;
     }
-    return 0;      
+    return 0;
 }
-    
+
```
ES-Alexander: Here too.

ArduSub/mode_guided.cpp
```diff
@@ -415,7 +441,7 @@ void ModeGuided::guided_set_yaw_state(bool use_yaw, float yaw_cd, bool use_yaw_r
     if (use_yaw && !use_yaw_rate) {
         sub.yaw_rate_only = false;
         sub.mode_auto.set_auto_yaw_look_at_heading(yaw_cd * 0.01f, 0.0f, direction, relative_angle);
-    } else if (use_yaw && use_yaw_rate) { 
+    } else if (use_yaw && use_yaw_rate) {
```
ES-Alexander: And here

ArduSub/mode_guided.cpp
```diff
@@ -775,7 +809,7 @@ void ModeGuided::set_auto_yaw_mode(autopilot_yaw_mode yaw_mode)
     case AUTO_YAW_RESETTOARMEDYAW:
         // initial_armed_bearing will be set during arming so no init required
         break;
-    
+
```
ES-Alexander: And here

ES-Alexander (changes requested):
As mentioned previously, it's preferred not to have whitespace changes for parts of the code you're not already touching, because it can make the git history confusing when blaming.

This is Sub code so we have some extra discretion, but if you want this to be included in the PR then I'd recommend pulling it out into a dedicated (whitespace cleanup) commit that happens before the functional ones. Otherwise you can just remove those changes - up to you :-)

clydemcqueen:
Arghh, I am still tweaking my new IDE setup, and these changes didn't show up in my IDE diff. I'll remove them. Sorry!

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -721,23 +729,40 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
...
+                    if (vel_ignore) {
+                        sub.mode_guided.guided_set_destination(pos_neu_cm);
+                    } else {
+                        sub.mode_guided.guided_set_destination_posvel(pos_neu_cm, vel_neu_cm, Location::AltFrame::ABOVE_ORIGIN);
+                    }
+                    break;
+                }
+                default:
+                    break;
```
peterbarker: Put all of the enumeration values in don't `default`.

Suggestion only here, but it's good practice.

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -721,23 +729,40 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
...
+                    if (!loc.get_vector_xy_from_origin_NE_cm(pos_ne_cm)) {
+                        break;
+                    }
+
+                    if (!vel_ignore) {
+                        sub.mode_guided.guided_set_destination_posvel(Vector3f(pos_ne_cm.x, pos_ne_cm.y, packet.alt * 100.0f), vel_neu_cm, frame);
+                    }
+                    break;
+                }
+                case Location::AltFrame::ABOVE_HOME: {
```
peterbarker, line 749: This should be ABOVE_ORIGIN, surely?!

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -721,23 +729,40 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
...
+            switch (frame) {
+                case Location::AltFrame::ABOVE_TERRAIN: {
+                    // packet.alt is the target rangefinder range, not the target altitude
+                    const Location loc{packet.lat_int, packet.lon_int, 0, frame};
+                    Vector2f pos_ne_cm;
+                    if (!loc.get_vector_xy_from_origin_NE_cm(pos_ne_cm)) {
+                        break;
+                    }
+
+                    if (!vel_ignore) {
```
peterbarker, line 743: Are you missing an `else` clause here?!

If intentional, you need a comment :-)

ArduSub/Sub.cpp
```diff
@@ -466,6 +467,36 @@ void Sub::rc_loop()
 }
 #endif
 
+#if AP_SCRIPTING_ENABLED
+// set target position and velocity, alt frame is Location::AltFrame::ABOVE_ORIGIN
```
peterbarker:
```suggestion
// set target position and velocity in supplied AltFrame
```

peterbarker (dismissed):

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: This is intentional, at least for now. Supporting the WP cases would be a nice addition, I created an issue to track these.

I added a comment.

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: Actually, I do mean ABOVE_HOME:
* mavlink_coordinate_frame_to_location_alt_frame maps RELATIVE_ALT to ABOVE_HOME
* get_vector_from_origin_NEU_cm transforms the point from ABOVE_HOME to ABOVE_ORIGIN
* mode_guided expects ABOVE_ORIGIN (or ABOVE_TERRAIN in the posvel case)

In fact, in my first attempt at using a switch statement, I wrote ABOVE_ORIGIN and the tests promptly failed. So... thanks for the feedback!

I added some comments.

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -721,23 +729,45 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
...
-            sub.mode_guided.guided_set_destination_posvel(pos_neu_cm, Vector3f(packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f));
-        } else if (pos_ignore && !vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_velocity(Vector3f(packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f));
-        } else if (!pos_ignore && vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_destination(pos_neu_cm);
+            const Vector3f vel_neu_cm{packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f};
+
+            switch (frame) {
+                case Location::AltFrame::ABOVE_TERRAIN: {
+                    // transform (lat, lon) from ABOVE_HOME to ABOVE_ORIGIN
```
peterbarker:
```suggestion
                    // transform (lat, lon) from TERRAIN to ABOVE_ORIGIN
```

ArduSub/GCS_MAVLink_Sub.cpp
peterbarker: Ah!  I just really was not expecting people to be navigating relhome.  This looks fine.

Note that you could factor out this "transform point from" to be outside the `switch` statement.  You will do the work even if the frame isn't supported, but *shrug*.

There is teh question as to why the velocity handling is different depending on the frame here.  That's weird.  This is related to me "else" clause above.

peterbarker (commented):
I don't think your routine working out how to shove stuff into guided mode is quite right yet.  But if you think this is a real improvement I don't think that should stop you merging this and coming back to it.

clydemcqueen:
This failed in test.Plane.AUTOTUNE. It passes locally.

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -711,9 +713,15 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
             break;
         }
 
-        Vector3f pos_neu_cm;  // position (North, East, Up coordinates) in centimeters
+        if (!acc_ignore) {
```
ES-Alexander, line 712: Am I understanding correctly that this guard is here because acceleration targets are not (yet) supported as part of position targeting? I feel like that shouldn't be a silent failure, but I suppose only commands have result responses (maybe this _should_ be a command, but that's way out of scope here), so the only option would be a `gcs().send_text` call. That said, other similar parts of the message handling don't do that, and I'm unsure if there's an explicit reason for it - fine to leave out for this PR.

Either way, a comment here saying why there's an early exit is a good idea :-)

ArduSub/mode_guided.cpp
```diff
@@ -298,35 +300,60 @@ void ModeGuided::guided_set_velocity(const Vector3f& velocity, bool use_yaw, flo
...
         LOGGER_WRITE_ERROR(LogErrorSubsystem::NAVIGATION, LogErrorCode::DEST_OUTSIDE_FENCE);
         // failure is propagated to GCS with NAK
         return false;
     }
 #endif
 
-    // check we are in posvel control mode
-    if (sub.guided_mode != Guided_PosVel) {
-        guided_posvel_control_start();
+    // ensure we are in Guided_PosVel sub-mode and the frames match
```
ES-Alexander:
```suggestion
    // move into Guided_PosVel sub-mode with the specified frame (as necessary)
```
The original comment made me think it was supposed to be a validation check, so I got confused by the implementation (particularly because of the `else` clause, which doesn't help ensure either of those things). Definitely "my reading" of it, but I think this alternative might be slightly harder to misinterpret :-)

Alternatively we could leave the original comment and switch the code blocks around, with the conditional checking `sub.guided_mode == Guided_PosVel && posvel_frame == alt_frame` instead, but then the code is slightly harder to follow because initialisation comes later than updating.

ArduSub/mode_guided.cpp
```diff
@@ -298,35 +300,60 @@ void ModeGuided::guided_set_velocity(const Vector3f& velocity, bool use_yaw, flo
...
+
+            guided_posvel_control_start(alt_frame);
+
+            // bumpless transfer: _pos_desired_neu_m.z = destination.z
+            position_control->init_pos_terrain_D_m((destination.z - sub.rangefinder_state.inertial_alt_cm) * 0.01f);
+            position_control->set_pos_terrain_target_D_m(-sub.rangefinder_state.rangefinder_terrain_offset_cm * 0.01f);
+        } else {
+            guided_posvel_control_start(alt_frame);
+        }
+    } else {
```
ES-Alexander:
```suggestion
    } else {  // already in the intended sub-mode + frame
```

ES-Alexander (approved):
Looks ok to me - just a few minor suggestions on making the code easier to follow :-)

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: > Am I understanding correctly that this guard is here because acceleration targets are not (yet) supported as part of position targeting?

Correct. FYI that acceleration targets would be straightforward to add. I could imagine an external obstacle avoidance system that would provide precise, smooth {pos, vel, accel} response to obstacles.

> other similar parts of the message handling don't do that, and I'm unsure if there's an explicit reason for it

Vehicles should send POSITION_TARGET_GLOBAL_INT (if requested) so that a controller can check for problems. But... I just tested this and discovered that this only works for the Guided_WP sub-mode, not for the other Guided submodes. I'll raise this in an issue.

For now I just added a comment.

clydemcqueen: I noted the lack of POSITION_TARGET_GLOBAL_INT messages here: https://github.com/ArduPilot/ardupilot/issues/33098

ES-Alexander (approved):
Looks good - thanks! :D

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -1255,6 +1255,39 @@ int lua_AP_Vehicle_set_target_velocity_NED(lua_State *L)
...
+    }
+
+    AP_Vehicle *ud = check_AP_Vehicle(L);
+    Vector3f &pos = *check_Vector3f(L, 2);
+    Vector3f &vel = *check_Vector3f(L, 3);
+
+    Location::AltFrame alt_frame = Location::AltFrame::ABOVE_ORIGIN;
+
+    if (args == 4) {
+        alt_frame = static_cast<Location::AltFrame>(get_integer(L, 4, 0, UINT8_MAX));
```
IamPete1:
```suggestion
        alt_frame = static_cast<Location::AltFrame>(get_uint8_t(L, 4));
```

IamPete1:
Is this the first case of allowing a alt frame without at lat/lon directly with it? Somehow the altframe is only really valid at a location. For example flying 50 m absolute altitude might be fine in one location, but in another you will crash. But here I guess we still have a location associated with the alt frame, just it is not a lat/lon. I think it is fine, I guess going forward we might want to make a new type for Vector3 with alt frame like we do with location such that the alt frame is always associated with its position.

In anycase, I think the scripting and copter stuff looks good.

libraries/AP_Scripting/lua_bindings.cpp
clydemcqueen: Done.

clydemcqueen:
> I guess going forward we might want to make a new type for Vector3 with alt frame like we do with location such that the alt frame is always associated with its position.

Yes, that would be nifty.

Thanks for the review @IamPete1

clydemcqueen:
Rebase to force a CI re-run

clydemcqueen:
Looks like just `colcon build/test` failed due to 403 error. But everything else passes!

peterbarker:
Sorry, I've popped MergeOnCIPass off as it touches the Copter directory and @rmackay9 has not reviewed this.

Williangalvani:
> Sorry, I've popped MergeOnCIPass off as it touches the Copter directory and @rmackay9 has not reviewed this.

My bad! I pinged @iampete for the lua sections but missed @rmackay9 !
sorry!

ArduSub/mode_guided.cpp
```diff
@@ -298,35 +300,60 @@ void ModeGuided::guided_set_velocity(const Vector3f& velocity, bool use_yaw, flo
...
+                gcs().send_text(MAV_SEVERITY_WARNING, "Terrain data (rangefinder) not available");
+                return false;
+            }
+
+            guided_posvel_control_start(alt_frame);
+
+            // bumpless transfer: _pos_desired_neu_m.z = destination.z
+            position_control->init_pos_terrain_D_m((destination.z - sub.rangefinder_state.inertial_alt_cm) * 0.01f);
+            position_control->set_pos_terrain_target_D_m(-sub.rangefinder_state.rangefinder_terrain_offset_cm * 0.01f);
+        } else {
```
peterbarker: You should not be assuming that if things are not above-terrain then they are above-origin. 

`switch` instead.

rmackay9:
So this looks like we're simply translating an alt-above-terrain into an alt-above-ekf-origin and then using that as the target.  The way we should do this is to do terrain following on the way to the target.

clydemcqueen:
> So this looks like we're simply translating an alt-above-terrain into an alt-above-ekf-origin and then using that as the target. The way we should do this is to do terrain following on the way to the target.

Hi @rmackay9 -- thanks for the review!

Are you referring terrain following in AC_WPNav? [Indeed this is a great idea](https://github.com/ArduPilot/ardupilot/issues/32961).

While WP navigation is desired, it isn't always possible. We run 100 m transects at a constant speed (10 cm/s) and at a constant height (80 cm) above the seafloor. The terrain is typically sloping, sometimes rocky, and vegetated in the summer, with kelp stipes running from the seafloor to the surface blocking the path. The pilot's job is move from A to B in a reasonably straight line in these conditions. SURFTRAK has been a huge help in maintaining constant height, but the pilot still has to carefully maintain a slow, constant forward velocity for ~20 minutes, which is difficult.

This PR allows external controllers to use `Guided_PosVel` mode and rapid target updates to maintain both constant height and constant velocity, while letting the pilot maintain control over yaw. The `GuidedPosVel` autotest demonstrates this using MAVLink. We've tried this in real conditions and is works great -- in fact we have been running this since December.

I am happy to discuss further!

ArduSub/mode_guided.cpp
clydemcqueen: Great catch! Sorry I missed that.

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -718,23 +727,45 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
...
-            sub.mode_guided.guided_set_destination_posvel(pos_neu_cm, Vector3f(packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f));
-        } else if (pos_ignore && !vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_velocity(Vector3f(packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f));
-        } else if (!pos_ignore && vel_ignore && acc_ignore) {
-            sub.mode_guided.guided_set_destination(pos_neu_cm);
+            const Vector3f vel_neu_cm{packet.vx * 100.0f, packet.vy * 100.0f, -packet.vz * 100.0f};
+
+            switch (frame) {
+                case Location::AltFrame::ABOVE_TERRAIN: {
+                    // transform (lat, lon) from ABOVE_TERRAIN to ABOVE_ORIGIN
```
ES-Alexander: This feels misleading, and may be what's confusing Randy.

The functionality is about transforming the horizontal positions, whereas the "ABOVE" nomenclature makes it seem like it's changing a terrain-focused target distance into an origin-focused altitude target (which would be incorrect behaviour).

```suggestion
                    // transform global horizontal (lat, lon) input to a local (x, y) position
```

ArduSub/GCS_MAVLink_Sub.cpp
```diff
@@ -718,23 +727,45 @@ void GCS_MAVLINK_Sub::handle_message(const mavlink_message_t &msg)
...
+
+                    // velocity is required
+                    if (!vel_ignore) {
+                        // packet.alt is the target rangefinder range, not the target altitude
+                        sub.mode_guided.guided_set_destination_posvel(Vector3f(pos_ne_cm.x, pos_ne_cm.y, packet.alt * 100.0f), vel_neu_cm, frame);
+                    }
+                    break;
+                }
+                case Location::AltFrame::ABOVE_HOME: {
+                    // transform point from ABOVE_HOME to ABOVE_ORIGIN
```
ES-Alexander, line 750: If this comment is correct, it seems like supporting the `ABOVE_ORIGIN` frame would require almost no additional code (just a different transformation), because the actual functionality is apparently already rel-origin.

Am I misunderstanding what's happening here? Not sure if there are nuances elsewhere that stop that from being applicable, or if that just determined as "more to review and test, so better to handle in a later PR".

ArduSub/Sub.cpp
```diff
@@ -465,6 +465,37 @@ void Sub::rc_loop()
...
+{
+    // exit if the vehicle is not in Guided mode or Auto-Guided mode
+    if (!flightmode->in_guided_mode()) {
+        return false;
+    }
+
+    float dest_u_m = 0.0f;
+    switch (alt_frame) {
+        case Location::AltFrame::ABOVE_HOME:
+            // distance above water surface
```
ES-Alexander:
```suggestion
            // distance above home (usually at the water surface)
```
This feels more correct, though I still don't love the wording.

I'm a little unsure whether the functionality works as intended if home is somewhere below (or above) the local current water surface, but I can't see a reason why it wouldn't, and it's quite possible it's not this PR's fault if there do turn out to be issues with how ArduSub handles that.

ArduSub/GCS_MAVLink_Sub.cpp
clydemcqueen: See `GCS_MAVLINK::mavlink_coordinate_frame_to_location_alt_frame`. IIUC, MAVLink doesn't have the notion of `ABOVE_ORIGIN`.

It would be easy to support `ABOVE_ORIGIN` via Lua, and  `ABSOLUTE` via MAVLink and Lua.

rmackay9:
@clydemcqueen,

I think you'll find though that this change means that if there's an underwater hill between where the vehicle is now and where it wants to go that it will hit the side of the hill.  I'm really not sure it's a good idea to have two different implementations of this.  I think it would be better to pass the output from the rangefinder into AC_WPNav and then let it do the terrain following for the user

peterbarker:
> let it do the terrain following for the user

This is actually a problem in the way ArduPilot treats terrain.

You have a destination.  It's a point in space.  You specify its altitude using a frame, and we convert it from that frame into either an AMSL altitude (Planes in fixed-wing), or into a relative-to-origin offset (down) (most other things!).

We *also* have a path-planning requirement on how to get there.  Do you fly diagonally (shortest path)?  Do you fly laterally and then vertically?  Do you fly vertically then laterally?  Do you fly vertically, then laterally, then vertically again?  We do all of those across vehicles - and that's not even considering obstacle avoidance!

ArduPilot absolute conflates these two concepts in various places!

Randy is suggesting that Sub do what we do on other vehicles.  If you specify a destination in terrain frame that absolutely infers that you come to the desired destination altitude straight away and maintain that altitude terrain-relative altitude en-route to the destination.  Certainly useful for avoiding running into intermediate terrain.

The sub guys want to be able to take shortest-path to a point in space - where the altitude is defined in above-terrain as they're trying to survey this way.

It feels like if you're trying to survey at a constant height above terrain then terrain-following is appropriate.  If the operator is just trying to put the vehicle somewhere they can then start their survey from... perhaps it is better to not terrain-follow so it's a cleaner story?!

In any case, I've never really like the fact we conflate the way we are specifying the destination with how we get there.  We should discuss that and get that fixed on some really long time-scale.

rmackay9:
Hi @peterbarker,

One correction though, the line below is not quite how Copter/Heli interpret terrain altitudes.  Instead the vehicle slowly increases/decreases its alt-above-terrain so that it perfectly reaches the target altitude-above-terrain when it arrives at the destination.  We have [this section on the terrain following wiki page that tries to explain this](https://ardupilot.org/copter/docs/terrain-following.html#warning).

`you come to the desired destination altitude straight away and maintain that altitude terrain-relative altitude `

clydemcqueen:
Hi @rmackay9  and @peterbarker thanks for the thoughtful comments.

> I think you'll find though that this change means that if there's an underwater hill between where the vehicle is now and where it wants to go that it will hit the side of the hill.

For sure flying 80 cm above the seafloor is tricky, which is why we typically run the transect at 10 cm/s and do manual obstacle avoidance.

> I think it would be better to pass the output from the rangefinder into AC_WPNav and then let it do the terrain following for the user

Our transects are roughly straight, but they may curve to follow a contour, or the pilot may have to move around boulders or kelp stipes. We do this by sending targets at 10 Hz based on the yaw that the pilot controls. Maybe I missing something here, but my understanding is that AC_WPNav doesn't support this sort of navigation?

> I'm really not sure it's a good idea to have two different implementations of this.

Got it. I modified Sub GuidedSubMode::Guided_PosVel because it was a fairly small change. The PosVel xy and z controllers were independent; I left them that way. This independent behavior is useful, but it isn't essential for our use case.

There is another option:

* Add a new sub-mode to Sub: GuidedSubMode::**Guided_Pos**. This would basically be the same as SubMode::**Pos** in Copter. It would handle both the WP case and the non-WP case, and it would handle both RELATIVE and TERRAIN frames. The non-WP case would support rapid target updates.
* To disambiguate the WP and non-WP cases in MAVLink, add a GUIDED_OPTIONS parameter to Sub with a single option (for now): ModeGuided::Options::**WPNavUsedForPosControl**. This would default to true to preserve existing Sub behavior.

This a bigger change, and requires more testing. Thoughts?

ES-Alexander:
Ok, from a bunch of discussion and some sleuthing, my latest understanding is:

1. Guided_PosVel allows for external trajectory control, with rapid updates of targets to move _towards_ (e.g. from a companion computer, or user inputs augmented with a Lua script)
    - The general expectation is that new updates are provided before reaching the target
    - Copter has PosVelAccel, which is even better for this
        - We may want to add Accel support to Sub too, but that's out of scope here
    - _This_ PR is trying to add `ABOVE_TERRAIN` frame as an option for Sub's Guided_PosVel targets, then following the terrain to get to them
        - Copter's PosVelAccel may not support that at the moment (?), but could perhaps implement it in future
1. Guided_WP is for high-level specification of locations to move _to_/_through_
    - `ABOVE_TERRAIN` is already supported there (via `MAV_CMD_DO_REPOSITION`), but there's also ongoing work to [make that more accessible](https://github.com/ArduPilot/ardupilot/pull/33412)
    - This matches Copter's WP mode - it's not suited to full/rapid external trajectory control, because when a target update is received the autopilot relaxes its controllers and starts planning a new trajectory to stop at that new target
1. Copter also has Pos mode, with an option bit in [`GUID_OPTIONS`](https://ardupilot.org/copter/docs/parameters.html#guid-options-guided-mode-options) to select between the direct / trajectory-focused position controller (but with zero velocity and acceleration at each target position), or using the full WPNav controller
    - The position controller option couples together the horizontal and vertical controllers
    - Replicating that for Sub _may_ be desirable, but is also out of scope for this PR
1. Peter B's point stands that it would be nice to decouple ways of specifying "where to go" from "how to get there"
    - That's a broader ArduPilot issue/consideration that's definitely out of scope here

Tools/autotest/ardusub.py
```diff
@@ -484,9 +487,127 @@ def SimTerrainMission(self):
         self.wait_altitude(altitude_min=end_altitude-validation_delta/2, altitude_max=end_altitude+validation_delta/2,
                            relative=False, timeout=1)
 
+        # Disarm before context pop
+        self.disarm_vehicle()
+
+    def GuidedPosVel(self):
```
ES-Alexander, line 493: Ideally this will make use of some shared functionality with the GuidedWP autotest.

ES-Alexander (commented):
Seemingly needs a rebase, I suspect post the merging of #33412.

clydemcqueen:
I met with @Williangalvani and @lthall and we came up with a revised plan. I am closing this PR.

PR #27769, opened 2024-08-06
OSD should display ESC temp rather than motor temp

andyp1per:
Motor temp is rarely implemented and was used before.


robustini:
Work fine (esc temp at bottom right), thanks @andyp1per!

DJI_0001 MOV_snapshot_00 17 040

tridge (approved):

rmackay9:

PR #27770, opened 2024-08-06
Plane: TKOFF_THR_MIN is applied to SLT transitions

Georacer:
This PR applies `TKOFF_THR_MIN` throttle to forward transitions of SLT-type quadplanes.

Screenshot from 2024-08-12 21-55-42

## Changes

- Apply `TKOFF_THR_MIN` throttle to transitions of SLT-type quadplanes.
- Split `quadplane.in_transition()` into forward and backward cases. SLT and tiltrotor never go into a back-transition. Tailsitter can return `in_back_transition()==true`.

## Tests

In SITL the following has been tested:
- In SLT, the `TKOFF_THR_MIN` is applied during the forward transition.
- In SLT, `in_frwd_transition` is never true during a back-transition.

## Notes

- By default `TKOFF_THR_MIN` is 60%. We had put it thus for classic FW takeoffs. Is it a problem that all SLT-quadplanes will now transition at 60% min throttle?

### DIFF
--- a/ArduPlane/Attitude.cpp
+++ b/ArduPlane/Attitude.cpp
@@ -204,9 +204,9 @@ float Plane::stabilize_pitch_get_pitch_out()
 #endif
     // if LANDING_FLARE RCx_OPTION switch is set and in FW mode, manual throttle,throttle idle then set pitch to LAND_PITCH_DEG if flight option FORCE_FLARE_ATTITUDE is set
 #if HAL_QUADPLANE_ENABLED
-    const bool quadplane_in_transition = quadplane.in_transition();
+    const bool quadplane_in_frwd_transition = quadplane.in_frwd_transition();
 #else
-    const bool quadplane_in_transition = false;
+    const bool quadplane_in_frwd_transition = false;
 #endif
 
     int32_t demanded_pitch = nav_pitch_cd + int32_t(g.pitch_trim * 100.0) + SRV_Channels::get_output_scaled(SRV_Channel::k_throttle) * g.kff_throttle_to_pitch;
@@ -219,7 +219,7 @@ float Plane::stabilize_pitch_get_pitch_out()
        - throttle stick at zero thrust
        - in fixed wing non auto-throttle mode
     */
-    if (!quadplane_in_transition &&
+    if (!quadplane_in_frwd_transition &&
         !control_mode->is_vtol_mode() &&
         !control_mode->does_auto_throttle() &&
         flare_mode == FlareMode::ENABLED_PITCH_TARGET &&
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -3971,28 +3971,11 @@ bool QuadPlane::is_vtol_land(uint16_t id) const
 }
 
 /*
-  return true if we are in any kind of transition
- */
-bool QuadPlane::in_transition(void) const
-{
-    return in_frwd_transition() || in_back_transition();
-}
-
-/*
   return true if we are in a transition to fwd flight from hover
  */
 bool QuadPlane::in_frwd_transition(void) const
 {
     return available() && transition->active_frwd();
-}
-
-/*
-  return true if we are in a transition to hover from fwd flight
- */
-bool QuadPlane::in_back_transition(void) const
-{
-    // By default the 
-    return available() && transition->active_back();
 }
 
 /*
@@ -4374,11 +4357,6 @@ bool SLT_Transition::allow_update_throttle_mix() const
 bool SLT_Transition::active_frwd() const
 {
     return quadplane.assisted_flight && ((transition_state == TRANSITION_AIRSPEED_WAIT) || (transition_state == TRANSITION_TIMER));
-}
-
-bool SLT_Transition::active_back() const
-{
-    return false;
 }
 
 /*
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -810,6 +810,10 @@ bool QuadPlane::setup(void)
     pilot_speed_z_max_dn.convert_centi_parameter(AP_PARAM_INT16);
     pilot_accel_z.convert_centi_parameter(AP_PARAM_INT16);
 
+    // Provisionally assign the SLT thrust type.
+    // It will be overwritten by tailsitter or tiltorotor setups.
+    thrust_type = ThrustType::SLT;
+
     tailsitter.setup();
 
     tiltrotor.setup();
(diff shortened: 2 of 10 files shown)


Georacer:
Case where QAssist will employ TKOFF_THR_MIN:
Screenshot from 2024-08-01 17-26-33

ArduPlane/servos.cpp
```diff
@@ -559,6 +559,14 @@ float Plane::apply_throttle_limits(float throttle_in)
         if (aparm.takeoff_throttle_max != 0) {
             max_throttle = aparm.takeoff_throttle_max.get();
         }
+
+        // Apply minimum throttle limits only for SLT thrust types.
+        // The other types don't support it well.
+        if (quadplane.thrust_type == QuadPlane::ThrustType::THRUST_TYPE_SLT) {
```
tridge, line 565: need a && control_mode->does_auto_throttle()
normally could be a quadplane.get_thrust_type() then thrust_type would be private

tridge (changes requested):

ArduPlane/quadplane.h
```diff
@@ -76,6 +76,14 @@ class QuadPlane
     void control_auto(void);
     bool setup(void);
 
+    enum class ThrustType : uint8_t {
+        THRUST_TYPE_UNDEFINED=0, // Undefined. Means that no backend has configured it yet.
```
tridge, line 80: UNDEFINED not needed

peterbarker, line 80:
```suggestion
        UNDEFINED=0, // Undefined. Means that no backend has configured it yet.
```

superfluous.  similarly on other entries

peterbarker (changes requested):

ArduPlane/quadplane.h
Georacer: Done.

ArduPlane/servos.cpp
Georacer: Done.

ArduPlane/quadplane.h
Georacer: Amended.

ArduPlane/quadplane.cpp
```diff
@@ -4350,11 +4371,16 @@ bool SLT_Transition::allow_update_throttle_mix() const
...
 
-bool SLT_Transition::active() const
+bool SLT_Transition::active_frwd() const
 {
     return quadplane.assisted_flight && ((transition_state == TRANSITION_AIRSPEED_WAIT) || (transition_state == TRANSITION_TIMER));
 }
 
+bool SLT_Transition::active_back() const
+{
+    return false;
```
tridge, line 4381: there is a back transition :-) 
we can remove active_back() for now to avoid having to define it

ArduPlane/Attitude.cpp
```diff
@@ -219,7 +219,7 @@ float Plane::stabilize_pitch_get_pitch_out()
        - throttle stick at zero thrust
        - in fixed wing non auto-throttle mode
     */
-    if (!quadplane_in_transition &&
+    if (!quadplane_in_frwd_transition &&
```
tridge, line 222: this is specifically aiming for TVBS landing, so not fwd transition

tridge: we will need @Hwurzburg to have a look at how this is defined, and for what classes of aircraft

tridge (changes requested):
@Hwurzburg can you explain how the flare code works, what type of aircraft is it used on?

Hwurzburg:
> @Hwurzburg can you explain how the flare code works, what type of aircraft is it used on?

tilt rotor VTOLs landing conventionally as fixed wing....with switch active in pilot throttle controlled modes it will force tilts up at low throttle to clear props...at highest switch position it also sets pitch to flare pitch .....use it on all my tvbs for emergency fw landing when batts are low

ArduPlane/Attitude.cpp
Georacer: > tilt rotor VTOLs landing conventionally as fixed wing....with switch active in pilot throttle controlled modes it will force tilts up at low throttle to clear props...at highest switch position it also sets pitch to flare pitch .....use it on all my tvbs for emergency fw landing when batts are low

@Hwurzburg the reason we ask is that we want to decide if the `quadplane.in_transition()` test needs to be replaced with a check for forward transition only.
Based on your description, I understand now that the `!quadplane_in_transition` check is meant to protect against the case where the pilot is currently commanding a forward transition while on low throttle. In this case we don't want to force the landing pitch.

@tridge, on second inspection, I see that there is an additional check here for `!control_mode->is_vtol_mode()`. When `quadplane.is_back_transition()==true` then tautologically `is_vtol_mode()==true` as well.
So adding a definition of and using `is_back_transition()` is superfluous.

I think we can simply replace `in_transition()` with `in_frwd_transition()` and spare us the definition of `in_back_transition()`.

ArduPlane/quadplane.cpp
Georacer: I propose removing it and I have a commit ready for it.
See my comment on the other conversation.

Georacer:
I just tested the feature. I loaded Convergence in RealFligth, and set `long DO_AUX_FUNCTION 89 1`. This is the flare switch.
Took off in QSTAB, then I transitioned to FBWA.
Every time I closed the throttle to zero, the front tilt servos would raise the motors vertically.

ArduPlane/Attitude.cpp
Georacer: I just tested the feature. I loaded Convergence in RealFligth, and set `long DO_AUX_FUNCTION 89 1`. This is the flare switch.
Took off in QSTAB, then I transitioned to FBWA.
Every time I closed the throttle to zero, the front tilt servos would raise the motors vertically.

Hwurzburg: great!

tridge:
@Georacer good catch on the default for TKOFF_THR_MIN, I think change to zero, and where the variable is used, if it is zero then use TRIM_THROTTLE, I think it is reasonable that a takeoff always uses at least trim (cruise) throttle
on some aircraft 60% throttle may lead to overspeed on takeoff

Georacer:
> @Georacer good catch on the default for TKOFF_THR_MIN, I think change to zero, and where the variable is used, if it is zero then use TRIM_THROTTLE, I think it is reasonable that a takeoff always uses at least trim (cruise) throttle on some aircraft 60% throttle may lead to overspeed on takeoff

I've set `TKOFF_THR_MIN` to 0. When it is set to 0, the transition will now use `TRIM_THROTTLE`.

On the other hand, for a normal plane takeoff (not transition), when `TKOFF_THR_MIN==0`, `TRIM_THROTTLE` will be used instead. as we have been discussing in https://github.com/ArduPilot/ardupilot/pull/27758

tridge (approved):

PR #27772, opened 2024-08-07
AP_Periph: add support for extended ESC DroneCAN message. 

IamPete1:
This builds on https://github.com/ArduPilot/ardupilot/pull/27771 to allow testing of both it and https://github.com/ArduPilot/ardupilot/pull/27755.

They do work:
image

On Periph this is not enabled by default. If it is enabled it sends each ESC in turn such that they are each sent at the set `ESC_EXT_TLM_RATE`, by default this is 10 times lower than the default ESC status message rate.  

I had to add getters into AP_ESC_Telem for periph to read the values. I also added some token values to the SITL esc backend. 

I have also added a helper to ESC telem for the get functions to check the type bitmask and if the data is stale in one call.

### DIFF
--- a/Tools/AP_Periph/AP_Periph.h
+++ b/Tools/AP_Periph/AP_Periph.h
@@ -334,9 +334,15 @@ public:
 #ifdef HAL_PERIPH_ENABLE_RC_OUT
 #if HAL_WITH_ESC_TELEM
     AP_ESC_Telem esc_telem;
+    uint8_t get_motor_number(const uint8_t esc_number) const;
     uint32_t last_esc_telem_update_ms;
     void esc_telem_update();
     uint32_t esc_telem_update_period_ms;
+#if AP_EXTENDED_ESC_TELEM_ENABLED
+    void esc_telem_extended_update(const uint32_t &now_ms);
+    uint32_t last_esc_telem_extended_update;
+    uint8_t last_esc_telem_extended_sent_id;
+#endif
 #endif
 
     SRV_Channels servo_channels;
--- a/Tools/AP_Periph/Parameters.cpp
+++ b/Tools/AP_Periph/Parameters.cpp
@@ -712,6 +712,17 @@ const AP_Param::Info AP_Periph_FW::var_info[] = {
     GSCALAR(rpm_msg_rate, "RPM_MSG_RATE", 0),
 #endif
 
+#if AP_EXTENDED_ESC_TELEM_ENABLED && HAL_WITH_ESC_TELEM
+    // @Param: ESC_EXT_TLM_RATE
+    // @DisplayName: ESC Extended telemetry message rate
+    // @Description: This is the rate at which extended ESC Telemetry will be sent across the CAN bus for each ESC
+    // @Units: Hz
+    // @Range: 0 50
+    // @Increment: 1
+    // @User: Advanced
+    GSCALAR(esc_extended_telem_rate, "ESC_EXT_TLM_RATE", AP_PERIPH_ESC_TELEM_RATE_DEFAULT / 10),
+#endif
+
     AP_VAREND
 };
 
--- a/Tools/AP_Periph/can.cpp
+++ b/Tools/AP_Periph/can.cpp
@@ -1670,6 +1670,20 @@ void AP_Periph_FW::can_start()
 
 #ifdef HAL_PERIPH_ENABLE_RC_OUT
 #if HAL_WITH_ESC_TELEM
+// try to map the ESC number to a motor number. This is needed
+// for when we have multiple CAN nodes, one for each ESC
+uint8_t AP_Periph_FW::get_motor_number(const uint8_t esc_number) const
+{
+    const auto *channel = SRV_Channels::srv_channel(esc_number);
+    // try to map the ESC number to a motor number. This is needed
+    // for when we have multiple CAN nodes, one for each ESC
+    if (channel == nullptr) {
+        return esc_number;
+    }
+    const int8_t motor_num = channel->get_motor_num();
+    return (motor_num == -1) ? esc_number : motor_num;
+}
+
 /*
   send ESC status packets based on AP_ESC_Telem
  */
@@ -1681,15 +1695,8 @@ void AP_Periph_FW::esc_telem_update()
         mask &= ~(1U<<i);
         const float nan = nanf("");
         uavcan_equipment_esc_Status pkt {};
-        const auto *channel = SRV_Channels::srv_channel(i);
-        // try to map the ESC number to a motor number. This is needed
-        // for when we have multiple CAN nodes, one for each ESC
-        if (channel == nullptr) {
-            pkt.esc_index = i;
-        } else {
-            const int8_t motor_num = channel->get_motor_num();
-            pkt.esc_index = motor_num==-1? i:motor_num;
-        }
+        pkt.esc_index = get_motor_number(i);
+
         if (!esc_telem.get_voltage(i, pkt.voltage)) {
             pkt.voltage = nan;
         }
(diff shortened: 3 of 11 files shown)


libraries/AP_ESC_Telem/AP_ESC_Telem_Backend.h
```diff
@@ -27,10 +27,14 @@ class AP_ESC_Telem_Backend {
...
         uint32_t flags;             // Status flags
+        uint8_t power_percentage;   // Percentage of output power
 #endif // AP_EXTENDED_ESC_TELEM_ENABLED
 
         // return true if the data is stale
-        bool stale(uint32_t now_ms=0) const volatile;
+        bool stale(uint32_t now_ms) const volatile;
+
+        //  return true if the requested type of data is available and not stale
+        bool valid(const uint16_t type_mask) const volatile;
```
IamPete1, line 37: I'm not sure if this really needs to be volatile....

Tools/AP_Periph/can.cpp
```diff
@@ -1722,6 +1736,73 @@ void AP_Periph_FW::esc_telem_update()
...
+        return;
+    }
+
+    uint32_t mask = esc_telem.get_active_esc_mask();
+    if (mask == 0) {
+        // No ESCs to report
+        return;
+    }
+
+    // ESCs are sent in turn to minimise used bandwidth, to make the rate param matchs the status message we multiplty
```
IamPete1, line 1753:
```suggestion
    // ESCs are sent in turn to minimise used bandwidth, to make the rate param match the status message we multiply
```

Tools/AP_Periph/can.cpp
```diff
@@ -1722,6 +1736,73 @@ void AP_Periph_FW::esc_telem_update()
...
+        // Too soon!
+        return;
+    }
+    last_esc_telem_extended_update = now_ms;
+
+    for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
+        // Send each ESC in turn
+        const uint8_t index = (last_esc_telem_extended_sent_id + 1 + i) % ESC_TELEM_MAX_ESCS;
+
+        if ((mask & (1U << i)) == 0) {
```
IamPete1, line 1766:
```suggestion
        if ((mask & (1U << index)) == 0) {
```

Tools/AP_Periph/Parameters.cpp
```diff
@@ -712,6 +712,17 @@ const AP_Param::Info AP_Periph_FW::var_info[] = {
     GSCALAR(rpm_msg_rate, "RPM_MSG_RATE", 0),
 #endif
 
+#if AP_EXTENDED_ESC_TELEM_ENABLED && HAL_WITH_ESC_TELEM
+    // @Param: ESC_EXT_TLM_RATE
+    // @DisplayName: ESC Extended telemetry message rate
+    // @Description: This is the rate at which ESC Telemetry will be sent across the CAN bus for each ESC
```
tpwrules, line 718:
```suggestion
    // @Description: This is the rate at which extended ESC Telemetry will be sent across the CAN bus for each ESC
```

tridge:
I think we can remove the volatiles, then merge on CI pass

Tools/AP_Periph/Parameters.cpp
IamPete1: Done thanks.

IamPete1:
> I think we can remove the volatiles, then merge on CI pass

I have done some more reading, i think its correct here. It enforces that it must be called on a object which is marked volatile. So because the objects are declared volatile here:
https://github.com/ArduPilot/ardupilot/blob/d225683d69493ba3e5f3c78ea376482b7e9017f2/libraries/AP_ESC_Telem/AP_ESC_Telem.cpp#L241

The volatile on the method enforces that, so it should be a compiler error if you were to try calling it on a none volatile instance of the object.

Tools/AP_Periph/can.cpp
```diff
@@ -1708,7 +1715,14 @@ void AP_Periph_FW::esc_telem_update()
         if (esc_telem.get_raw_rpm(i, rpm)) {
             pkt.rpm = rpm;
         }
-        pkt.power_rating_pct = 0;
+
+#if AP_EXTENDED_ESC_TELEM_ENABLED
```
tpwrules, line 1719: Technically this is not extended, just new to this PR? Should it really be classified as such?

Tools/AP_Periph/can.cpp
```diff
@@ -1722,6 +1736,73 @@ void AP_Periph_FW::esc_telem_update()
...
+            pkt.motor_temperature_degC = motor_temp_cdeg * 0.01;
+            have_data = true;
+        }
+
+        have_data |= esc_telem.get_input_duty(index, pkt.input_pct);
+        have_data |= esc_telem.get_output_duty(index, pkt.output_pct);
+        have_data |= esc_telem.get_flags(index, pkt.status_flags);
+
+        if (have_data) {
+            pkt.esc_index = get_motor_number(i);
```
tpwrules, line 1788:
```suggestion
            pkt.esc_index = get_motor_number(index);
```

tpwrules (changes requested):
Found a bug and a few opportunities for cleanup, but once fixed I think this is ready to go in. The data looks good in the DroneCAN GUI tool in SITL (after the fix).

Tools/AP_Periph/AP_Periph.h
```diff
@@ -334,9 +334,15 @@ class AP_Periph_FW {
 #ifdef HAL_PERIPH_ENABLE_RC_OUT
 #if HAL_WITH_ESC_TELEM
     AP_ESC_Telem esc_telem;
+    uint8_t get_motor_number(const uint8_t esc_number) const;
     uint32_t last_esc_telem_update_ms;
     void esc_telem_update();
     uint32_t esc_telem_update_period_ms;
+#if AP_EXTENDED_ESC_TELEM_ENABLED
+    void esc_telem_extended_update(const uint32_t &now_ms);
```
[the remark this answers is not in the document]
IamPete1: I guess it really comes out the same in this case, but passing the reference is often less data than passing the object.

Tools/AP_Periph/can.cpp
```diff
@@ -1708,7 +1715,14 @@ void AP_Periph_FW::esc_telem_update()
         if (esc_telem.get_raw_rpm(i, rpm)) {
             pkt.rpm = rpm;
         }
-        pkt.power_rating_pct = 0;
+
+#if AP_EXTENDED_ESC_TELEM_ENABLED
+        uint8_t power_rating_pct;
+        if (esc_telem.get_power_percentage(i, power_rating_pct)) {
+            pkt.power_rating_pct = power_rating_pct;
+        }
```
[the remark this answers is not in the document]
IamPete1: Yeah, but then you need the whole `IGNORE_RETURN` thing for the bool. I think this is clearer.

Tools/AP_Periph/can.cpp
IamPete1: Not new in this PR. Its not in the DroneCAN extended message, but it is a thing we only get over DroneCAN none of the other ESC protocols report it. The extended telem define and log seemed like a better fit than trying to squeeze it into the existing log where all the other protocol would have to pay the cost too.

Tools/AP_Periph/can.cpp
IamPete1: Yep, Fixed, Thanks.

Tools/AP_Periph/AP_Periph.h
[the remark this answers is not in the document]
tpwrules: It will save a bit of flash though.

Tools/AP_Periph/can.cpp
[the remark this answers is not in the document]
IamPete1: It really should be marked to warn..... but I didn't want  to get into that in this PR.

tpwrules (approved):
Looks good now, tested it with the DroneCAN GUI tool and SITL.

Decided that the volatile was probably necessary due to potential cross-thread access, and that the check for an out of range ESC number needed to be moved to avoid UB.

tridge (approved):

PR #27773, opened 2024-08-07
AP_Camera: Add Camera Tracking functionality (WIP)

khanasif786:
This adds support for a companion computer to perform image tracking using any of the backends

### DIFF
--- a/ArduCopter/mode_follow.cpp
+++ b/ArduCopter/mode_follow.cpp
@@ -2,6 +2,7 @@
 
 #if MODE_FOLLOW_ENABLED == ENABLED
 
+#include<AP_Camera/AP_Camera_Tracking.h>
 /*
  * mode_follow.cpp - follow another mavlink-enabled vehicle by system id
  *
@@ -128,7 +129,49 @@ void ModeFollow::run()
                 }
                 break;
             }
-
+#if AP_CAMERA_OFFBOARD_TRACKING_ENABLED
+            case AP_Follow::YAW_BEHAVE_OBJECT_FOLLOW: {
+                AP_Mount *mount = AP_Mount::get_singleton();
+                float yaw_max, pitch_max;
+                float yaw_min, pitch_min;
+                mount->get_mount_yaw_limits(yaw_min,yaw_max);
+                mount->get_mount_pitch_limits(pitch_min,pitch_max);
+                // Get gimbal attitude
+                float roll, pitch, yaw;
+                if (!mount->get_attitude_euler(mount->get_primary_instance(), roll, pitch, yaw)) {
+                    break; 
+                }
+                float last_yaw_tobe_pointed_to;
+                bool last_yaw_tobe_pointed_to_isvalid = AP_Follow::get_singleton()->get_yaw_under_gimbal_limit_to_point_to(last_yaw_tobe_pointed_to);
+                
+                bool cond1 = yaw > (yaw_max - g2.follow.get_object_follow_yaw_reset());
+                bool cond2 = yaw < (yaw_min + g2.follow.get_object_follow_yaw_reset());
+                bool cond3 = !last_yaw_tobe_pointed_to_isvalid;
+                bool cond4 = pitch < (pitch_min + g2.follow.get_object_follow_pitch_reset());
+                
+                if (cond1 || cond2 || cond3 || cond4) {
+                    // gcs().send_text(MAV_SEVERITY_INFO, "Reset: y=%.1f p=%.1f c1=%d c2=%d c3=%d c4=%d", yaw, pitch, cond1, cond2, cond3, cond4);
+                    
+                    if (dist_vec.xy().length_squared() > 1.0) {
+                        yaw_cd = get_bearing_cd(Vector2f{}, dist_vec.xy());
+                        AP_Follow::get_singleton()->set_yaw_under_gimbal_limit_to_point_to(yaw_cd);
+                        use_yaw = true;
+                    } else {
+                        // code this
+                        // Its when your object is going (slowly or by some way) such that you can't get the distance vector less than 1 and
+                        // You have achieved your yaw and pitch limits
+                    }
+                }
+                last_yaw_tobe_pointed_to_isvalid = AP_Follow::get_singleton()->get_yaw_under_gimbal_limit_to_point_to(last_yaw_tobe_pointed_to);
+                if (last_yaw_tobe_pointed_to_isvalid) {
+                    yaw_cd = last_yaw_tobe_pointed_to;
+                    use_yaw = true;
+                } else {
+                    // code this
+                }
+                break;
+            }
+#endif
             case AP_Follow::YAW_BEHAVE_NONE:
             default:
                 // do nothing


libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -14,6 +14,16 @@ AP_Camera_Backend::AP_Camera_Backend(AP_Camera &frontend, AP_Camera_Params &para
     _instance(instance)
 {}
 
+// init - performs any required initialisation
+void AP_Camera_Backend::init() {
```
peterbarker, line 18:
```suggestion
void AP_Camera_Backend::init()
{
```

similarly elsewere throughout this file

libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -415,4 +425,61 @@ void AP_Camera_Backend::log_picture()
 }
 #endif
 
+bool AP_Camera_Backend::set_tracking_external(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2) {
```
rmackay9, line 428: let's put comments above every function (both in the .h and .cpp)

libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -14,6 +14,16 @@ AP_Camera_Backend::AP_Camera_Backend(AP_Camera &frontend, AP_Camera_Params &para
     _instance(instance)
 {}
 
+// init - performs any required initialisation
+void AP_Camera_Backend::init() {
+    // initialise the tracking object
+#if AP_CAMERA_TRACKING_ENABLED
```
peterbarker, line 20:
```suggestion
#if AP_CAMERA_TRACKING_ENABLED
    // initialise the tracking object
```

libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -415,4 +425,61 @@ void AP_Camera_Backend::log_picture()
 }
 #endif
 
+bool AP_Camera_Backend::set_tracking_external(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2) {
+#if AP_CAMERA_TRACKING_ENABLED
```
peterbarker, line 429:
```suggestion
#if AP_CAMERA_TRACKING_ENABLED
bool AP_Camera_Backend::set_tracking_external(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2) {
```

Don't call the method in the first place

libraries/AP_Camera/AP_Camera_Backend.cpp
peterbarker, line 428: Needs better names instead of p1 and p2

libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -415,4 +425,61 @@ void AP_Camera_Backend::log_picture()
...
+
+bool AP_Camera_Backend::set_tracking(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2) {
+    if (_params.track_enable == 1) {
+        return set_tracking_external(tracking_type, p1, p2);
+    } else {
+        return set_tracking_internal(tracking_type, p1, p2);
+    }
+}
+
+void AP_Camera_Backend::handle_message_camera_information(mavlink_channel_t chan, const mavlink_message_t &msg)  {
```
rmackay9, line 447: I think we will face conflicts when using a mavlink enabled camera (e.g. CAMx_TYPE = MavlinkCamV2) because we will also receive a camera_information message from these cameras.

libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -415,4 +425,61 @@ void AP_Camera_Backend::log_picture()
...
+    if (tracker == nullptr) {
+        return false;
+    }
+    return tracker->set_tracking(tracking_type, p1, p2, _params.track_sysid, _params.track_compid, _cam_info);
+#else
+    return false;
+#endif
+}
+
+bool AP_Camera_Backend::set_tracking(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2) {
```
peterbarker, line 439: This should be gated on `AP_CAMERA_TRACKING_ENABLED`

libraries/AP_Camera/AP_Camera_Backend.h
```diff
@@ -183,6 +196,10 @@ class AP_Camera_Backend
     Location last_location;         // Location that last picture was taken at (used for trigg_dist calculation)
     uint16_t image_index;           // number of pictures taken since boot
     bool last_is_armed;             // stores last arm/disarm state. true if it was armed lastly
+    bool _got_camera_info;      // true once camera has provided CAMERA_INFORMATION
```
rmackay9, line 199: If these are only used for the image tracking then let's put these into a structure

libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -415,4 +425,61 @@ void AP_Camera_Backend::log_picture()
...
+    if (_params.track_enable == 1) {
+        return set_tracking_external(tracking_type, p1, p2);
+    } else {
+        return set_tracking_internal(tracking_type, p1, p2);
```
peterbarker, line 443:
```suggestion
    }
    return set_tracking_internal(tracking_type, p1, p2);
```

libraries/AP_Camera/AP_Camera_Backend.h
```diff
@@ -80,7 +81,13 @@ class AP_Camera_Backend
     // set tracking to none, point or rectangle (see TrackingType enum)
     // if POINT only p1 is used, if RECTANGLE then p1 is top-left, p2 is bottom-right
     // p1,p2 are in range 0 to 1.  0 is left or top, 1 is right or bottom
-    virtual bool set_tracking(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2) { return false; }
+    bool set_tracking(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2);
+
+    // default tracking supported by camera
+    virtual bool set_tracking_internal(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2) { return false; }
```
peterbarker, line 87: None of this should exist unless tracking is enabled

libraries/AP_Camera/AP_Camera_MAVLinkCamV2.cpp
```diff
@@ -123,36 +123,6 @@ SetFocusResult AP_Camera_MAVLinkCamV2::set_focus(FocusType focus_type, float foc
     return SetFocusResult::ACCEPTED;
 }
 
-// handle incoming mavlink message including CAMERA_INFORMATION
```
rmackay9, line 126: I think we should not remove this method.  I think the backend should catch it first and then forward it on

libraries/AP_Camera/AP_Camera_Params.cpp
```diff
@@ -103,6 +103,25 @@ const AP_Param::GroupInfo AP_Camera_Params::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("_VFOV", 13, AP_Camera_Params, vfov, 0),
 
+    // @Param: _TRK_ENABLE
```
peterbarker, line 106: Needs define guards

libraries/AP_Camera/AP_Camera_Backend.h
```diff
@@ -95,7 +102,10 @@ class AP_Camera_Backend
     float vertical_fov() const { return MAX(0, _params.vfov); }
 
     // handle MAVLink messages from the camera
-    virtual void handle_message(mavlink_channel_t chan, const mavlink_message_t &msg) {}
```
rmackay9, line 98: Let's leave this handle_message() as virtual but then the MAVlinkCamV2 backend should call the parent handle_message() as either the first thing or last thing it does.  e.g call AP_Camera_Backend::handle_message()

rmackay9: maybe add a comment to make it clear that all child classes, if they implement this method, must still call the parent's handle_message() method.

libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -68,4 +68,13 @@ void AP_Camera_Servo::configure(float shooting_mode, float shutter_speed, float
     }
 }
 
+// // handle MAVLink messages for the camera
```
peterbarker, line 71: dead code?

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,70 @@
...
+               return false;
+            }
+            break;
+        case TrackingType::TRK_RECTANGLE:
+            if (!(_cam_info.flags & CAMERA_CAP_FLAGS_HAS_TRACKING_RECTANGLE)) {
+               return false;
+            }
+            break;
+        default:
+            return false;
```
peterbarker, line 27:
```suggestion
```

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    gcs().send_text(MAV_SEVERITY_INFO,"Tracking: New Tracking request");
+    uint8_t proxy_device_compid=0, proxy_device_sysid=0;
```
peterbarker, line 31:
```suggestion
    uint8_t proxy_device_compid;
    uint8_t proxy_device_sysid;
```

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+// set tracking to none, point or rectangle (see TrackingType enum)
+// if POINT only p1 is used, if RECTANGLE then p1 is top-left, p2 is bottom-right
+// p1,p2 are in range 0 to 1.  0 is left or top, 1 is right or bottom
+bool AP_Camera_Tracking::set_tracking(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2, uint8_t tracking_device_sysid, uint8_t tracking_device_compid, mavlink_camera_information_t _cam_info)
+{
+    // if we don't support the required tracking then return
+    switch (tracking_type) {
+        case TrackingType::TRK_NONE:
+            break;
```
rmackay9, line 15: I think this needs to return

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,70 @@
...
+        default:
+            return false;
+    }
+
+    gcs().send_text(MAV_SEVERITY_INFO,"Tracking: New Tracking request");
+    uint8_t proxy_device_compid=0, proxy_device_sysid=0;
+    auto _link = GCS_MAVLINK::find_by_mavtype_and_compid(MAV_TYPE_ONBOARD_CONTROLLER, proxy_device_compid, proxy_device_sysid);
+    if (_link == nullptr) {
+        gcs().send_text(MAV_SEVERITY_WARNING,"AP_Camera: Could Not find any onboard controller registered");
+    }
```
peterbarker, line 35:
```suggestion
        return false;
    }
```

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,70 @@
...
+               return false;
+            }
+            break;
+        default:
+            return false;
+    }
+
+    gcs().send_text(MAV_SEVERITY_INFO,"Tracking: New Tracking request");
+    uint8_t proxy_device_compid=0, proxy_device_sysid=0;
+    auto _link = GCS_MAVLINK::find_by_mavtype_and_compid(MAV_TYPE_ONBOARD_CONTROLLER, proxy_device_compid, proxy_device_sysid);
```
rmackay9, line 32: I think we shouldn't look this up each time.  Let's store this as a variable/member

peterbarker, line 32: Embed _link in the backend instance rather than looking it up each time

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if (_link == nullptr) {
+        gcs().send_text(MAV_SEVERITY_WARNING,"AP_Camera: Could Not find any onboard controller registered");
+    }
+    // prepare and send message
+    mavlink_command_long_t pkt {};
+
+    pkt.confirmation=0;
+    pkt.target_component=tracking_device_compid;
+    pkt.target_system = tracking_device_sysid;
+    pkt.param1 = 0;
```
rmackay9, line 42: I think we don't we need to set these values to zero because of the {} above

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    gcs().send_text(MAV_SEVERITY_INFO,"Tracking: New Tracking request");
+    uint8_t proxy_device_compid=0, proxy_device_sysid=0;
+    auto _link = GCS_MAVLINK::find_by_mavtype_and_compid(MAV_TYPE_ONBOARD_CONTROLLER, proxy_device_compid, proxy_device_sysid);
+    if (_link == nullptr) {
+        gcs().send_text(MAV_SEVERITY_WARNING,"AP_Camera: Could Not find any onboard controller registered");
+    }
+    // prepare and send message
+    mavlink_command_long_t pkt {};
```
rmackay9, line 37: you could use named fields for this

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,70 @@
...
+    pkt.confirmation=0;
+    pkt.target_component=tracking_device_compid;
+    pkt.target_system = tracking_device_sysid;
+    pkt.param1 = 0;
+    pkt.param2 = 0;
+    pkt.param3 = 0;
+    pkt.param4 = 0;
+    pkt.param5 = 0;
+    pkt.param6 = 0;
+    pkt.param7 = 0;
```
peterbarker, line 48: use named-field initialisation here

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    if (tracking_type == TrackingType::TRK_POINT) {
```
peterbarker, line 50: switch

libraries/AP_Camera/AP_Camera_Tracking.h
```diff
@@ -0,0 +1,61 @@
...
+
+    // set tracking to non      nj klbhnjk nm,e, point or rectangle (see TrackingType enum)
+    // if POINT only p1 is used, if RECTANGLE then p1 is top-left, p2 is bottom-right
+    // p1,p2 are in range 0 to 1.  0 is left or top, 1 is right or bottom
+    bool set_tracking(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2, uint8_t tracking_device_sysid, uint8_t tracking_device_compid, mavlink_camera_information_t _cam_info);
+
+    // handle MAVLink messages from the camera
+    void handle_message(mavlink_channel_t chan, const mavlink_message_t &msg);
+
+//     // configure camera
```
rmackay9, line 51: let's just delete this commented out code

libraries/AP_Camera/AP_Camera_Tracking.h
```diff
@@ -0,0 +1,61 @@
...
+  Camera Tracking Library
+ */
+#pragma once
+
+#include <GCS_MAVLink/GCS.h>
+#include "AP_Camera_shareddefs.h"
+// #include "AP_Camera_Backend.h"
+#include "AP_Camera_config.h"
+
+#if AP_CAMERA_TRACKING_ENABLED
```
peterbarker, line 24:
```suggestion
#include "AP_Camera_config.h"

#if AP_CAMERA_TRACKING_ENABLED

#include <GCS_MAVLink/GCS.h>
#include "AP_Camera_shareddefs.h"
// #include "AP_Camera_Backend.h"
#include "AP_Camera_config.h"

```

libraries/AP_Camera/AP_Camera_Tracking.h
```diff
@@ -0,0 +1,61 @@
...
+// #include "AP_Camera.h"
+
+class AP_Camera_Tracking
+{
+public:
+
+    // Constructor
+    AP_Camera_Tracking() {
+        
+    }
```
peterbarker, line 35:
```suggestion
    AP_Camera_Tracking() {}
```

rmackay9:
Could you rebase this and also clarify what testing has been done?

libraries/AP_Camera/AP_Camera_Tracking.h
```diff
@@ -0,0 +1,61 @@
...
+    /* Do not allow copies */
+    CLASS_NO_COPY(AP_Camera_Tracking);
+
+    // initialize the camera tracking library
+    void init();
+
+    // set tracking to non      nj klbhnjk nm,e, point or rectangle (see TrackingType enum)
+    // if POINT only p1 is used, if RECTANGLE then p1 is top-left, p2 is bottom-right
+    // p1,p2 are in range 0 to 1.  0 is left or top, 1 is right or bottom
+    bool set_tracking(TrackingType tracking_type, const Vector2f& p1, const Vector2f& p2, uint8_t tracking_device_sysid, uint8_t tracking_device_compid, mavlink_camera_information_t _cam_info);
```
peterbarker, line 46: top-left / bottom-right naming
```suggestion
    bool set_tracking(TrackingType tracking_type,
                                 const Vector2f& p1,
                                 const Vector2f& p2,
                                 uint8_t tracking_device_sysid,
                                 uint8_t tracking_device_compid,
                                 mavlink_camera_information_t _cam_info);
```

libraries/AP_Camera/AP_Camera_Tracking.h
```diff
@@ -0,0 +1,61 @@
...
+    void handle_message(mavlink_channel_t chan, const mavlink_message_t &msg);
+
+//     // configure camera
+//     void configure(float shooting_mode, float shutter_speed, float aperture, float ISO, int32_t exposure_type, int32_t cmd_id, float engine_cutoff_time) override;
+
+//     // send camera settings message to GCS
+//     void send_camera_settings(mavlink_channel_t chan) const override;
+
+//     // send camera information message to GCS
+//     void send_camera_information(mavlink_channel_t chan) const override;
```
peterbarker, line 58: Dead code?

peterbarker (changes requested):

rmackay9:
Could you create a separate PR with the python script in a new directory AP_Camera/examples?

rmackay9:
We should also consume the [CAMERA_TRACKING_IMAGE_STATUS]( https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/common.xml#L6079C1-L6079C35) message and update the user

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
+import threading
+import time
+from pymavlink import mavutil
+
+ONBOARD_CONTROLLER_SYSID = 245
+ONBOARD_CONTROLLER_COMPID = 0
+
+# Function to connect to MAVLink
+def connect_to_mavlink(ip, port):
```
peterbarker, line 9: Create a class which encapsulates everything, then create an instance of the class and invokes a run method on it

See `size_compare_branches.py` as an example of that structure

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
+import threading
+import time
+from pymavlink import mavutil
+
+ONBOARD_CONTROLLER_SYSID = 245
```
rmackay9, line 5: I think we should accept a command line argument for the system id and component id.  If those arguments aren't specified then I think we should consume HEARTBEAT messages and use the first system id that we receive as long as the MAV_TYPE (of the HEARTBEAT message) is not a GCS.  For component id we should probably default to using MAV_COMP_ID_ONBOARD_COMPUTER (191)

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
...
+ONBOARD_CONTROLLER_COMPID = 0
+
+# Function to connect to MAVLink
+def connect_to_mavlink(ip, port):
+    connection = mavutil.mavlink_connection(f'udp:{ip}:{port}', source_system=ONBOARD_CONTROLLER_SYSID)
```
peterbarker, line 10: Wait for a heartbeat, lock onto the first system ID you see that is probably a vehicle heartbeat (see mavutil's `probably_vehicle_heartbeat` method)

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
...
+    connection.wait_heartbeat()
+    print("Heartbeat received from system (system %u component %u)" % (connection.target_system, connection.target_component))
+    # Send a heartbeat back to register the device
+    connection.mav.heartbeat_send(
```
peterbarker, line 14: heartbeats must be sent regularly - at, say, 1Hz

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,161 @@
+import cv2
+import threading
+from pymavlink import mavutil
+import math 
+from ultralytics import YOLO
+
+master = mavutil.mavlink_connection('127.0.0.1:14560')
```
rmackay9, line 7: we should allow the caller to specify the video source as an argument to the script.

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
...
+        system_status=mavutil.mavlink.MAV_STATE_UNINIT,
+        mavlink_version=3
+    )
+    return connection
+
+ip = "127.0.0.1"
+port = 14570
+
+# Connect to MAVLink
+connection = connect_to_mavlink(ip, port)
```
peterbarker, line 28: This will end up something like `self.connection = self.connect_to_mavlink(ip, port)`

libraries/AP_Camera/examples/tracking.py
rmackay9: instructions on how to install would be good.  E.g. what packages are required and how can they be installed? (e.g. pip)

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
...
+    focal_length = float('nan')  # Not known
+    sensor_size_h = float('nan')  # Not known
+    sensor_size_v = float('nan')  # Not known
+    resolution_h = 640  # Example resolution
+    resolution_v = 480  # Example resolution
+    lens_id = 0  # Not known
+    flags = 4095  # All capabilities
+    cam_definition_version = 0  # Not known
+    cam_definition_uri = b""  # Not known
+    gimbal_device_id = 0  # Not known
```
peterbarker, line 47: Put these all into the send call directly?

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
...
+    print(f"Tracking point parameters: param1={param1}, param2={param2}")
+
+def send_heartbeat(connection):
+    while True:
+        connection.mav.heartbeat_send(
+            type=mavutil.mavlink.MAV_TYPE_ONBOARD_CONTROLLER,
+            autopilot=mavutil.mavlink.MAV_AUTOPILOT_INVALID,
+            base_mode=0,
+            custom_mode=0,
+            system_status=mavutil.mavlink.MAV_STATE_UNINIT,
```
peterbarker, line 83: Should be uninit

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,161 @@
...
+                    
+                    #put box in cam
+                    
+                    final_list = (x1, y1, x2, y2)
+                    cv2.rectangle(frame, (x1, y1), (x2, y2), (255, 0, 255), 3)
+
+            # faces = face_cascade.detectMultiScale(gray, scaleFactor=1.3, minNeighbors=5, minSize=(30, 30))
+            faces= [final_list]
+            if len(faces) == 0:
+                # Do something if no faces were detected
```
rmackay9, line 123: We should be sending the CAMERA_TRACKING_IMAGE_STATUS message at regular intervals (1hz?) so that AP knows if we can't track the object any more

libraries/AP_Camera/examples/tracking.py
peterbarker, line 7: Again, structure by creating an object then calling a method on it

rmackay9:
We should implement [MAV_CMD_CAMERA_STOP_TRACKING](https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/common.xml#L1806)

libraries/AP_Camera/examples/tracking.py
rmackay9: We should support [MAV_CMD_CAMERA_STOP_TRACKING](https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/common.xml#L1806)

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,161 @@
...
+import threading
+from pymavlink import mavutil
+import math 
+from ultralytics import YOLO
+
+master = mavutil.mavlink_connection('127.0.0.1:14560')
+master.wait_heartbeat()
+print("Heartbeat from the system (system %u component %u)" % 
+     (master.target_system, master.target_component))
+def send_gimbal_manager_pitch_yaw_angles(target_system, target_component, flags, gimbal_device_id, pitch, yaw, pitch_rate, yaw_rate):
```
peterbarker, line 11: Many of these parameters will become object state

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,161 @@
...
+                center_x = 640/2
+                center_y = 480/2
+                continue
+            try:
+                for (x, y, w, h) in faces:
+                    cv2.rectangle(frame, (x, y), (w, h), (255, 0, 0), 2)
+                    center_x = (x + w) // 2
+                    center_y = (y + h) // 2
+                    cv2.circle(frame, (center_x, center_y), 3, (0, 255, 0), -1)
+                    #print("Center of the face: ({}, {})".format(center_x, center_y))
```
rmackay9, line 134: instead of commented out lines, perhaps put a debug definition at the top to allow turning on/off debug

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,161 @@
...
+        pitch,
+        yaw,
+        pitch_rate,
+        yaw_rate
+    )
+    master.mav.send(msg)
+    #print("Gimbal manager pitch and yaw angles message sent.")
+send_gimbal_manager_pitch_yaw_angles(0, 0, 0, 0, 0, 0, float("NaN"),float("NaN"))
+def send_command():
+    global centre_x, center_y
```
peterbarker, line 38: this looks odd

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
+import threading
```
peterbarker, line 1:
```
'''
your comment about what this script does goes here

AP_FLAKE8_CLEAN
'''
```

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,161 @@
...
+        #print(diff_x,diff_y)
+
+        send_gimbal_manager_pitch_yaw_angles(0, 0, 0, 0, float("NaN"), float("NaN"), math.radians(diff_y), math.radians(diff_x))
+
+#########################################################
+# Initialize the VideoCapture object
+
+# gst-launch-1.0 -v udpsrc port=5600 caps='application/x-rtp, media=(string)video, clock-rate=(int)90000, encoding-name=(string)H264' ! rtph264depay ! avdec_h264 ! videoconvert ! autovideosink sync=false
+
+gst_pipeline = (
```
rmackay9, line 57: let's get a pipeline for at least the siyi camera gimbals using an ethernet connection to the companion copmuter

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,161 @@
...
+                continue
+            try:
+                for (x, y, w, h) in faces:
+                    cv2.rectangle(frame, (x, y), (w, h), (255, 0, 0), 2)
+                    center_x = (x + w) // 2
+                    center_y = (y + h) // 2
+                    cv2.circle(frame, (center_x, center_y), 3, (0, 255, 0), -1)
+                    #print("Center of the face: ({}, {})".format(center_x, center_y))
+            except:
+                pass
```
peterbarker, line 136: This isn't acceptable in any Python script.  You'd be surprised at how painful this can be.

Catch the exceptions you know about.

libraries/AP_Camera/examples/tracking.py
rmackay9: I think we should move this definition to the top to make it easier to find and possibly we should try and create this pipe line based on arguments input by the caller

peterbarker (changes requested):

Hwurzburg (changes requested):
needs exclusion in minimize include file and build options/extract features files added

libraries/AP_Camera/AP_Camera_Backend.cpp
khanasif786: top_left and bottom_right would be good

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
...
+
+# Main function
+def main():
+    # Send camera information message
+    send_camera_information(connection)
+    print('sent CAMERA INFO')
+    
+    # Start listening for MAVLink messages
+    while True:
+        msg = connection.recv_match(type='COMMAND_LONG', blocking=True)
```
srmainwaring, line 96: Does the call need to be blocking here?

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,161 @@
...
+# Initialize the VideoCapture object
+
+# gst-launch-1.0 -v udpsrc port=5600 caps='application/x-rtp, media=(string)video, clock-rate=(int)90000, encoding-name=(string)H264' ! rtph264depay ! avdec_h264 ! videoconvert ! autovideosink sync=false
+
+gst_pipeline = (
+    "udpsrc port=5600 ! application/x-rtp,encoding-name=H264,payload=96 ! "
+    "rtph264depay ! h264parse ! queue ! avdec_h264 ! videoconvert ! appsink"
+)
+
+cap = cv2.VideoCapture(gst_pipeline, cv2.CAP_GSTREAMER)
```
srmainwaring, line 62: The default versions of `opencv-python` and `opencv-contrib-python` do not have gstreamer support enabled. A custom build of the Python packages from source is required for this (at least this is the case on macOS). To reduce the complexity it may be easier to use the gstreamer Python bindings to capture the video stream and then make this available to opencv. The BlueROV docs have a guide for this: https://www.ardusub.com/developers/opencv.html.

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,106 @@
...
+            custom_mode=0,
+            system_status=mavutil.mavlink.MAV_STATE_UNINIT,
+            mavlink_version=3
+        )
+        time.sleep(1)
+
+# Main function
+def main():
+    # Send camera information message
+    send_camera_information(connection)
```
srmainwaring, line 91: There seems to be two ways to ensure that the `MAV_CMD_CAMERA_TRACK_POINT` response expected below is triggered when sending the camera info. One is in `AP_Camera::handle_command(const mavlink_command_int_t &packet)` and requires a `MAV_CMD_CAMERA_TRACK_POINT` message be sent. The other is to use an aux switch to enable camera image tracking. This is the setup I used for testing. 

```bash
RC13_OPTION      174.0       # Camera Image Tracking
```

Along with

```bash
param set CAM1_TRK_ENABLE 1
param set CAM1_TRK_SYSID 245
param set CAM1_TRK_COMPID 0
```

then run the script, enable tracking with

```bash
MANUAL> rc 13 1900
```

and check the output.

Either way all of this needs to be documented.

khanasif786 (commented):
@rmackay9 @peterbarker @srmainwaring  Have a look

rmackay9:
@khanasif786, thanks for all your efforts on this.  Could you rebase this on master too?

libraries/AP_Camera/AP_Camera_Tracking.cpp
khanasif786: Done

libraries/AP_Camera/AP_Camera_Tracking.h
```diff
@@ -0,0 +1,53 @@
...
+/*
+  Camera Tracking Library
+ */
+#pragma once
+
+#include "AP_Camera_config.h"
+#if AP_CAMERA_TRACKING_ENABLED
+#include <GCS_MAVLink/GCS.h>
+#include "AP_Camera_shareddefs.h"
+// #include "AP_Camera_Backend.h"
```
rmackay9, line 23: let's remove this commented out line

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,87 @@
...
+        case TrackingType::TRK_RECTANGLE:
+            if (!(_cam_info.flags & CAMERA_CAP_FLAGS_HAS_TRACKING_RECTANGLE)) {
+               return false;
+            }
+            break;
+    }
+
+    GCS_SEND_TEXT(MAV_SEVERITY_INFO,"Tracking: New Tracking request");
+    uint8_t proxy_device_compid = tracking_device_compid; 
+    uint8_t proxy_device_sysid;
```
srmainwaring, line 30: Needs initialising to avoid compile error on macOS.

```c++
= tracking_device_compid; 
    uint8_t proxy_device_sysid {};
```

khanasif786: Thanks

amilcarlucas:
```
antennatracker   916 (+0.0688%)   0 (0.0000%)  -4 (-0.0015%)  916 (+0.0687%)                                   631372
arduplane        1048 (+0.0583%)  0 (0.0000%)  0 (0.0000%)    1048 (+0.0582%)                                  164728
blimp            916 (+0.0677%)   0 (0.0000%)  4 (+0.0015%)   916 (+0.0675%)                                   609112
arducopter-heli  1076 (+0.0598%)  0 (0.0000%)  4 (+0.0015%)   1076 (+0.0597%)                                  162656
arducopter       1072 (+0.0596%)  0 (0.0000%)  0 (0.0000%)    1072 (+0.0594%)                                  161680
ardusub          1088 (+0.0684%)  0 (0.0000%)  0 (0.0000%)    1088 (+0.0683%)                                  371252
ardurover        1060 (+0.0643%)  0 (0.0000%)  4 (+0.0015%)   1060 (+0.0642%)                                  313360
```

Hwurzburg (changes requested):
still needs the changes I requested

libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -416,4 +427,66 @@ void AP_Camera_Backend::log_picture()
...
+
+    mavlink_msg_camera_information_decode(&msg, &camera_settings._cam_info);
+
+    const uint8_t fw_ver_major = camera_settings._cam_info.firmware_version & 0x000000FF;
+    const uint8_t fw_ver_minor = (camera_settings._cam_info.firmware_version & 0x0000FF00) >> 8;
+    const uint8_t fw_ver_revision = (camera_settings._cam_info.firmware_version & 0x00FF0000) >> 16;
+    const uint8_t fw_ver_build = (camera_settings._cam_info.firmware_version & 0xFF000000) >> 24;
+
+    // display camera info to user
+    gcs().send_text(MAV_SEVERITY_INFO, "Camera: %s.32 %s.32 fw:%u.%u.%u.%u",
```
landswellsong, line 468: This should be `%.32s` instead, probably carried over from the code before #27562

libraries/AP_Camera/AP_Camera_Tracking.cpp
```diff
@@ -0,0 +1,87 @@
...
+        case TrackingType::TRK_NONE:
+            pkt.command = MAV_CMD_CAMERA_STOP_TRACKING;
+            break;
+        default:
+            // Unknown
+            return false;
+    }
+
+    _link->send_message(MAVLINK_MSG_ID_COMMAND_LONG, (const char*)&pkt);
+    GCS_SEND_TEXT(MAV_SEVERITY_WARNING,"sent message to device sysid %d and comp %d",tracking_device_sysid,tracking_device_compid);
```
srmainwaring, line 83: This could be removed, and the sysid, compid info added to the `Tracking: New Tracking request` message on L28

srmainwaring:
@khanasif786, I think perhaps the follow-functionality should be split into a separate PR, with this one focussing on the core camera tracking changes.

The addition of `MSG_CAMERA_TRACKING_IMAGE_STATUS` is needed though if we are to run the tracker on a companion computer and want to send the GCS information about the current bounding box so it can be displayed in the GUI. I'm running into just this issue in testing.

I've written a MAVProxy module (`camtrack`) to test this PR in simulation and on hardware. There's a mavproxy GUI module to engage / disengage the tracker, and a separate headless module that may run on either the GCS computer, or on the companion computer.

It's working in simulation, still having a few issues setting up correct sys_id and component_id  for all components on hardware (CubeOrangePlus with RPi4 companion). AFAICT the RPi4 copes with the `cv2.legacy.TrackerCSTR` tracker.

The MAVProxy code is here: https://github.com/srmainwaring/MAVProxy/tree/wips/wip-camera-tracking. It is a WIP with a lot of debug message code included that will eventually be pruned. Let me know if you'd like a walk through.

libraries/AP_Camera/AP_Camera_Backend.h
```diff
@@ -186,6 +216,21 @@ class AP_Camera_Backend
...
+        mavlink_camera_tracking_image_status_t _cam_tracking_status {};
+        uint8_t _sysid_camera;      // sysid of camera
+        uint8_t _compid_camera;     // component id of gimbal
+    } camera_settings;
+
+    enum CAMERA_TRACKING_STATUS_FLAGS : uint8_t {
+        CAMERA_TRACKING_STATUS_FLAGS_IDLE,
+        CAMERA_TRACKING_STATUS_FLAGS_ACTIVE,
+        CAMERA_TRACKING_STATUS_FLAGS_ERROR
+    };
```
srmainwaring, line 233: Shadows 

```c++
#define HAVE_ENUM_CAMERA_TRACKING_STATUS_FLAGS
typedef enum CAMERA_TRACKING_STATUS_FLAGS
{
   CAMERA_TRACKING_STATUS_FLAGS_IDLE=0, /* Camera is not tracking | */
   CAMERA_TRACKING_STATUS_FLAGS_ACTIVE=1, /* Camera is tracking | */
   CAMERA_TRACKING_STATUS_FLAGS_ERROR=2, /* Camera tracking in error state | */
   CAMERA_TRACKING_STATUS_FLAGS_ENUM_END=3, /*  | */
} CAMERA_TRACKING_STATUS_FLAGS;
#endif
```
in `mavlink/v2.0/common/common.h`

libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -416,4 +445,73 @@ void AP_Camera_Backend::log_picture()
...
+// handle MAVLink messages from the camera
+void AP_Camera_Backend::handle_message(mavlink_channel_t chan, const mavlink_message_t &msg)
+{
+    // handle CAMERA_INFORMATION
+    switch (msg.msgid) {
+        case MAVLINK_MSG_ID_CAMERA_INFORMATION:
+            handle_message_camera_information(chan,msg);
+            break;
+        case MAVLINK_MSG_ID_CAMERA_TRACKING_IMAGE_STATUS:
+            handle_message_camera_tracking_image_status(chan,msg);
```
srmainwaring, line 511: Missing `break`. Compilation with `clang` fails.

srmainwaring (changes requested):

libraries/AP_Camera/AP_Camera_Backend.h
```diff
@@ -86,6 +86,14 @@ class AP_Camera_Backend
     // p1,p2 are in range 0 to 1.  0 is left or top, 1 is right or bottom
     bool set_tracking(TrackingType tracking_type, const Vector2f& top_left, const Vector2f& bottom_right);
 
+    // returns true if the objkect to be tracked is visible in the frame
```
srmainwaring, line 89: sp. objkect -> object

vicradon:
@khanasif786 
are you still working on this?

khanasif786:
@vicradon , Yes! Testing phase. do you have any suggestions or something let me know.

libraries/AP_Camera/AP_Camera_Tracking.cpp
khanasif786: Fixed Thanks

libraries/AP_Camera/AP_Camera_Backend.h
khanasif786: Thanks Fixed

libraries/AP_Camera/AP_Camera_Backend.cpp
khanasif786: Fixed Thanks

libraries/AP_Camera/AP_Camera_Backend.h
khanasif786: Thanks, Fixed

khanasif786:
The architecture of the Follow Library has been changed so much, which is why the merge conflict, So eventually this PULL will require so much rework. Let me come back here with a new and compatible patch

khanasif786:
The new working PR is here - #30941

PR #27774, opened 2024-08-07
Plane: use enum class for VTOL approach stage

peterbarker:
also namespace it with the state variable which uses this type

Without this you end up with a valid `RTL` token floating around, which is kind of nasty.

```
CubeOrange-periph-heavy  *                                                       
Durandal                            *      *           *       *                 0      *      *
Hitec-Airspeed           *                                                       
KakuteH7-bdshot                     *      *           *       *                 0      *      *
MatekF405                           *      *           *       *                 0      *      *
Pixhawk1-1M-bdshot                  *                  *       *                 0      *      *
f103-QiotekPeriph        *                                                       
f303-Universal           *                                                       
revo-mini                           *      *           *       *                 0      *      *
```

### DIFF
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -366,19 +366,20 @@ private:
         uint32_t AFS_last_valid_rc_ms;
     } failsafe;
 
-    enum Landing_ApproachStage {
-        RTL,
-        LOITER_TO_ALT,
-        ENSURE_RADIUS,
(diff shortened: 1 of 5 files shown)


peterbarker:
Added a breakpoint on `Plane::verify_landing_vtol_approach` and made sure it was still crossed (`VTOLLandSpiral`)

tridge (approved):

PR #27775, opened 2024-08-07
Copter: use HAL_PARACHUTE_ENABLED in place of PARACHUTE_ENABLED

peterbarker:
Last vehicle which uses `PARACHUTE` like this.

Tested for no compiler output change.

Tested with test_build_features.py specifying HAL_PARACHUTE_ENABLED


tridge:
randy approved

PR #27778, opened 2024-08-07
Add and use Battery backend internal-use-only flag 

peterbarker:
sometimes the information from a backend is very useful for logging, but not via telemetry.


MattKear (approved):
Nice feature!

tridge (approved):

tridge:
can be merged when testing complete, thanks!

lthall (approved):
Tested and it does what is says on the box!

PR #27782, opened 2024-08-07
Add arming check that all defaults in defaults files are used 

peterbarker:
Instant problem on CubeOrange:
```
AP: PreArm: unknown param BATT2_VOLT_MULT in parameter defaults fil
```


peterbarker:
That CubeOrange problem shows we can't really do this.

Georacer:
What was the problem here exactly?
That the 2nd battery monitor wasn't configured and the thing needed a power cycle for the parameter to appear? Sure I'm missing something.

PR #27783, opened 2024-08-07
Copter/AC_PosControl/Scripting: slung payload oscillation suppression

rmackay9:
This PR has expanded from its original scope to now add support for 3D position, velocity and acceleration offsets usable in all autonomous (e.g. Auto, Guide, RTL, etc) and semi-autonomous (e.g Loiter, PosHold, ZigZag, etc) modes.

This has been rigorously tested in SITL but it is a big change so any and all feedback is very welcome.

---------------------------

This adds slung payload oscillation suppression using a Lua script together with a new AC_PosControl feature that allows external callers (e.g. scripts) to add horizontal position or velocity offsets to the vehicle while flying in Auto mode.

The payload oscillation suppression only operates while the vehicle is in Auto mode executing a SCRIPT_TIME or PAYLOAD_PLACE command.  E.g. it does not trying to reduce oscillation while the vehicle if flying towards a waypoint nor in any other flight mode.  [A blog post including video is here](https://discuss.ardupilot.org/t/slung-payload-oscillation-suppression/123466).

Changes include:

1. SITL's slung payload simulator is enhance so that wind (e.g. SIM_WIND_SPD, SIM_WIND_DIR) affects the payload.  This allows us to create an offsets between the slung payload and vehicle
2. AC_PosControl gets new get/set methods for position and velocity offsets.  External callers (including scripting) can use these to add offsets to the vehicle's target position or velocity
3. AP_Vehicle/AP_Copter/AC_WPNav/AP_Scripting get equivalent new methods/bindings to set the position and velocity offsets (e.g. set_pos_offset, set_vel_offset).  Perhaps we should rename/replace these if we want to support more modes besides Auto
4. AP_Scripting's modules directory gets message definitions for GLOBAL_POSITION_INT and HEARTBEAT.  We haven't normally added these but without them it makes it very difficult for normal users to use scripts that consume or sends these mavlink messages
5. slung-payload.lua applet is added to AP_Scripting.  This script only operates while the vehicle is in Auto mode executing a SCRIPT_TIME or PAYLOAD_PLACE command.  It moves the vehicle towards the payload to reduce its oscillation with the speed of movement controlled by the SLUP_POS_P parameter.  It also tries to estimate the payload's long-term position offset from the vehicle and slowly moves the vehicle in the opposite direction so that the package lands right on the waypoint.  The speed of this movement can be adjusted using the SLUP_WP_POS_P parameter

This has been well tested in SITL and moderately well tested on a real vehicle:

- checked that AC_PosControl position and velocity targets result in the expected behaviour.  See screenshots below
auto-offset-testing-screenshot
- checked that the simulated payload lands precisely on the waypoint
- checked that simulated wind affects the payload as expected (e.g. payload is pushed by the wind)
- checked every navigation mission command with an offset of x=20m,y=20m to confirm they worked as expected

Video of SITL test: https://www.youtube.com/watch?v=jL2YUByYv2Y

Known issues we should fix before merging:

- [x] when switching between SCurve navigation and pure position control navigation the offsets are lost leading to a jump in the vehicle's position target.  This happens in Auto mode when switching from a WAYPOINT command and PAYLOAD_PLACE
- [x] when switching from takeoff to SCurve navigation the SCurves seem to use twice the expected position offset

### DIFF
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -498,7 +498,7 @@ void ModeAuto::land_start()
     set_submode(SubMode::LAND);
 }
 
-// auto_circle_movetoedge_start - initialise waypoint controller to move to edge of a circle with it's center at the specified location
+// circle_movetoedge_start - initialise waypoint controller to move to edge of a circle with it's center at the specified location
 //  we assume the caller has performed all required GPS_ok checks
 void ModeAuto::circle_movetoedge_start(const Location &circle_center, float radius_m, bool ccw_turn)
 {
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -1419,7 +1419,7 @@ void PayloadPlace::run()
         // vel_threshold_fraction * max velocity
         const float vel_threshold_fraction = 0.1;
         const float stop_distance = 0.5 * sq(vel_threshold_fraction * copter.pos_control->get_max_speed_up_cms()) / copter.pos_control->get_max_accel_z_cmss();
-        bool reached_altitude = copter.pos_control->get_pos_target_z_cm() >= descent_start_altitude_cm - stop_distance;
+        bool reached_altitude = copter.pos_control->get_pos_desired_z_cm() >= descent_start_altitude_cm - stop_distance;
         if (reached_altitude) {
             state = State::Done;
         }
@@ -1472,6 +1472,8 @@ bool ModeAuto::shift_alt_to_current_alt(Location& target_loc) const
         (wp_nav->get_terrain_source() == AC_WPNav::TerrainSource::TERRAIN_FROM_RANGEFINDER)) {
         int32_t curr_rngfnd_alt_cm;
         if (copter.get_rangefinder_height_interpolated_cm(curr_rngfnd_alt_cm)) {
+            // subtract position offset (if any)
+            curr_rngfnd_alt_cm -= pos_control->get_pos_offset_z_cm();
             // wp_nav is using rangefinder so use current rangefinder alt
             target_loc.set_alt_cm(MAX(curr_rngfnd_alt_cm, 200), Location::AltFrame::ABOVE_TERRAIN);
             return true;
@@ -1486,9 +1488,19 @@ bool ModeAuto::shift_alt_to_current_alt(Location& target_loc) const
         return false;
     }
 
-    // set target_loc's alt
-    target_loc.set_alt_cm(currloc.alt, currloc.get_alt_frame());
+    // set target_loc's alt minus position offset (if any)
+    target_loc.set_alt_cm(currloc.alt - pos_control->get_pos_offset_z_cm(), currloc.get_alt_frame());
     return true;
+}
+
+// subtract position controller offsets from target location
+// should be used when the location will be used as a target for the position controller
+void ModeAuto::subtract_pos_offsets(Location& target_loc) const
+{
+    // subtract position controller offsets from target location
+    const Vector3p& pos_ofs_neu_cm = pos_control->get_pos_offset_cm();
+    Vector3p pos_ofs_ned_m = Vector3p{pos_ofs_neu_cm.x * 0.01, pos_ofs_neu_cm.y * 0.01, -pos_ofs_neu_cm.z * 0.01};
+    target_loc.offset(-pos_ofs_ned_m);
 }
 
 /********************************************************************************/
@@ -1531,6 +1543,10 @@ void ModeAuto::do_nav_wp(const AP_Mission::Mission_Command& cmd)
 {
     // calculate default location used when lat, lon or alt is zero
     Location default_loc = copter.current_loc;
+
+    // subtract position offsets
+    subtract_pos_offsets(default_loc);
+
     if (wp_nav->is_active() && wp_nav->reached_wp_destination()) {
         if (!wp_nav->get_wp_destination_loc(default_loc)) {
             // this should never happen
@@ -1651,32 +1667,22 @@ void ModeAuto::do_land(const AP_Mission::Mission_Command& cmd)
 // note: caller should set yaw_mode
 void ModeAuto::do_loiter_unlimited(const AP_Mission::Mission_Command& cmd)
 {
-    // convert back to location
-    Location target_loc(cmd.content.location);
+    // calculate default location used when lat, lon or alt is zero
+    Location default_loc = copter.current_loc;
 
-    // use current location if not provided
-    if (target_loc.lat == 0 && target_loc.lng == 0) {
-        // To-Do: make this simpler
-        Vector3f temp_pos;
-        copter.wp_nav->get_wp_stopping_point_xy(temp_pos.xy());
-        const Location temp_loc(temp_pos, Location::AltFrame::ABOVE_ORIGIN);
-        target_loc.lat = temp_loc.lat;
-        target_loc.lng = temp_loc.lng;
-    }
+    // subtract position offsets
+    subtract_pos_offsets(default_loc);
 
-    // use current altitude if not provided
-    // To-Do: use z-axis stopping point instead of current alt
-    if (target_loc.alt == 0) {
-        // set to current altitude but in command's alt frame
-        int32_t curr_alt;
-        if (copter.current_loc.get_alt_cm(target_loc.get_alt_frame(),curr_alt)) {
-            target_loc.set_alt_cm(curr_alt, target_loc.get_alt_frame());
-        } else {
-            // default to current altitude as alt-above-home
-            target_loc.set_alt_cm(copter.current_loc.alt,
-                                  copter.current_loc.get_alt_frame());
+    // use previous waypoint destination as default if available
+    if (wp_nav->is_active() && wp_nav->reached_wp_destination()) {
+        if (!wp_nav->get_wp_destination_loc(default_loc)) {
+            // this should never happen
+            INTERNAL_ERROR(AP_InternalError::error_t::flow_of_control);
         }
     }
+
+    // get waypoint's location from command and send to wp_nav
+    const Location target_loc = loc_from_cmd(cmd, default_loc);
 
     // start way point navigator and provide it the desired location
     if (!wp_start(target_loc)) {
@@ -1689,7 +1695,13 @@ void ModeAuto::do_loiter_unlimited(const AP_Mission::Mission_Command& cmd)
 // do_circle - initiate moving in a circle
 void ModeAuto::do_circle(const AP_Mission::Mission_Command& cmd)
 {
-    const Location circle_center = loc_from_cmd(cmd, copter.current_loc);
+    // calculate default location used when lat, lon or alt is zero
+    Location default_loc = copter.current_loc;
+
+    // subtract position offsets
+    subtract_pos_offsets(default_loc);
+
+    const Location circle_center = loc_from_cmd(cmd, default_loc);
 
     // calculate radius
     uint16_t circle_radius_m = HIGHBYTE(cmd.p1); // circle radius held in high byte of p1
@@ -1757,6 +1769,10 @@ void ModeAuto::do_spline_wp(const AP_Mission::Mission_Command& cmd)
 {
     // calculate default location used when lat, lon or alt is zero
     Location default_loc = copter.current_loc;
+
+    // subtract position offsets
+    subtract_pos_offsets(default_loc);
+
     if (wp_nav->is_active() && wp_nav->reached_wp_destination()) {
         if (!wp_nav->get_wp_destination_loc(default_loc)) {
             // this should never happen
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -517,8 +517,8 @@ void ModeAuto::circle_movetoedge_start(const Location &circle_center, float radi
 
     // check our distance from edge of circle
     Vector3f circle_edge_neu;
-    copter.circle_nav->get_closest_point_on_circle(circle_edge_neu);
-    float dist_to_edge = (inertial_nav.get_position_neu_cm() - circle_edge_neu).length();
+    float dist_to_edge;
+    copter.circle_nav->get_closest_point_on_circle(circle_edge_neu, dist_to_edge);
 
     // if more than 3m then fly to edge
     if (dist_to_edge > 300.0f) {
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -1251,7 +1251,6 @@ void PayloadPlace::run()
 
     auto &g2 = copter.g2;
     const auto &g = copter.g;
-    auto &inertial_nav = copter.inertial_nav;
     auto *attitude_control = copter.attitude_control;
     const auto &pos_control = copter.pos_control;
     const auto &wp_nav = copter.wp_nav;
@@ -1293,7 +1292,7 @@ void PayloadPlace::run()
         case State::Descent_Start:
             gcs().send_text(MAV_SEVERITY_INFO, "%s Abort: Gripper Open", prefix_str);
             // Descent_Start has not run so we must also initalise descent_start_altitude_cm
-            descent_start_altitude_cm = inertial_nav.get_position_z_up_cm();
+            descent_start_altitude_cm = pos_control->get_pos_desired_z_cm();
             state = State::Done;
             break;
         case State::Descent:
@@ -1320,7 +1319,7 @@ void PayloadPlace::run()
 
     case State::Descent_Start:
         descent_established_time_ms = now_ms;
-        descent_start_altitude_cm = inertial_nav.get_position_z_up_cm();
+        descent_start_altitude_cm = pos_control->get_pos_desired_z_cm();
         // limiting the decent rate to the limit set in wp_nav is not necessary but done for safety
         descent_speed_cms = MIN((is_positive(g2.pldp_descent_speed_ms)) ? g2.pldp_descent_speed_ms * 100.0 : abs(g.land_speed), wp_nav->get_default_speed_down());
         descent_thrust_level = 1.0;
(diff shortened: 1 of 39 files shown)


libraries/AC_WPNav/AC_WPNav.h
```diff
@@ -220,6 +220,17 @@ class AC_WPNav
     /// return the crosstrack_error - horizontal error of the actual position vs the desired position
     float crosstrack_error() const { return _pos_control.crosstrack_error();}
 
+    ///
+    /// offset methods
+    ///
+    /// add an offset to vehicle's target position in NED meters
+    /// only xy offsets are supported
+    void set_pos_offset(const Vector3p& pos_offset_NED);
```
lthall, line 228: I wonder if these could be just in the position controller and work in any mode then.

ArduCopter/Copter.cpp
```diff
@@ -413,6 +413,28 @@ bool Copter::set_desired_speed(float speed)
...
 
 #if MODE_AUTO_ENABLED == ENABLED
+// add an additional offset to vehicle's target position
+// units are m or m/s in NED frame
+bool Copter::set_auto_pos_offset(const Vector3f &pos_offset_NED)
+{
+    if (flightmode != &mode_auto) {
+        return false;
+    }
+    wp_nav->set_pos_offset(pos_offset_NED.topostype());
```
lthall, line 423: I was thinking about this after our discussion. There isn't any reason we can't do this like guided where we use:
1. Pos, Vel, Accel
2. Vel, Accel

We just leave the later ones zero if we don't what the increased reaction time they provide.

libraries/AC_WPNav/AC_WPNav.h
rmackay9: I've moved the offset accessors to the position controller

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -406,13 +409,14 @@ void AC_PosControl::input_pos_xyz(const Vector3p& pos, float pos_offset_z, float
                         -constrain_float(accel_max_z_cmss, 0.0f, 750.0f), accel_max_z_cmss,
                         jerk_max_z_cmsss, _dt, false);
 
-    // update the vertical position, velocity and acceleration offsets
+    // update the position, velocity and acceleration offsets
+    update_xy_offsets();
     update_pos_offset_z(pos_offset_z);
```
lthall, line 414: I am thinking we should not pass pos_offset_z into update_pos_offset_z and instead just use _pos_offset_target.z in that function like we do update_pos_offset_xy.

We would then set _pos_offset_target.z at the top of this function.

This would be more consistent between xy and z.

libraries/AC_AttitudeControl/AC_PosControl.h
```diff
@@ -481,10 +495,18 @@ class AC_PosControl
...
     bool        _fwd_pitch_is_limited;     // true when the forward pitch demand is being limited to meet acceleration limits
 
-    float       _pos_offset_target_z;   // vertical position offset target, frame NEU in cm relative to the EKF origin
-    float       _pos_offset_z;          // vertical position offset, frame NEU in cm relative to the EKF origin
-    float       _vel_offset_z;          // vertical velocity offset in NEU cm/s calculated by pos_to_rate step
-    float       _accel_offset_z;        // vertical acceleration offset in NEU cm/s/s
+    // offset handling variables
+    enum class XYOffsetType {
+        NONE = 0,
+        POSITION,
```
lthall, line 501: This should be POS_VEL_ACCEL

libraries/AC_AttitudeControl/AC_PosControl.h
```diff
@@ -481,10 +495,18 @@ class AC_PosControl
...
+        VELOCITY
```
lthall, line 502: This should be VEL_ACCEL

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -582,12 +612,67 @@ void AC_PosControl::input_vel_accel_xy(Vector2f& vel, const Vector2f& accel, boo
...
+    case XYOffsetType::NONE:
+        break;
+
+    case XYOffsetType::POSITION: {
+        // update position, velocity, accel offsets for this iteration
+        update_pos_vel_accel_xy(_pos_offset.xy(), _vel_offset.xy(), _accel_offset.xy(), _dt, _limit_vector.xy(), _p_pos_xy.get_error(), _pid_vel_xy.get_error());
+
+        // input shape horizontal position offsets
+        Vector2f vel_zero, accel_zero;
+        shape_pos_vel_accel_xy(_pos_offset_target.xy(), vel_zero, accel_zero,
```
lthall, line 651: To make this generic this would update:
_pos_offset_target
_vel_offset_target
_accel_offset_target

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -582,12 +612,67 @@ void AC_PosControl::input_vel_accel_xy(Vector2f& vel, const Vector2f& accel, boo
...
+
+        // check for velocity offset target timeout
+        uint32_t now_ms = AP_HAL::millis();
+        if (now_ms - _vel_offset_target_ms > POSCONTROL_VEL_OFFSET_TARGET_TIMEOUT_MS) {
+            _vel_offset_target.zero();
+        }
+
+        // input shape horizontal velocity offsets
+        Vector2f accel_zero;
+        shape_vel_accel_xy(_vel_offset_target, accel_zero,
```
lthall, line 669: To make this generic this would update:
_vel_offset_target
_accel_offset_target

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1079,6 +1164,34 @@ Vector3f AC_PosControl::lean_angles_to_accel(const Vector3f& att_target_euler) c
     };
 }
 
+/// set the horizontal position offset target in cm from EKF origin in NE frame
+void AC_PosControl::set_pos_offset_target_xy_cm(const Vector2p& pos_offset_target_xy_cm)
```
lthall, line 1168: Here we would ideally set:
_pos_offset_target
_vel_offset_target
_accel_offset_target

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1079,6 +1164,34 @@ Vector3f AC_PosControl::lean_angles_to_accel(const Vector3f& att_target_euler) c
...
+    // set position offset target
+    _xy_offset_type = XYOffsetType::POSITION;
+    _pos_offset_target.xy() = pos_offset_target_xy_cm;
+
+    // clear velocity offset target
+    _vel_offset_target.zero();
+}
+
+/// set the horizontal velocity offset target in cm/s in NE frame
+void AC_PosControl::set_vel_offset_target_xy_cms(const Vector2f& vel_offset_target_xy_cms)
```
lthall, line 1179: Here we would ideally set:
_vel_offset_target
_accel_offset_target

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -582,12 +612,67 @@ void AC_PosControl::input_vel_accel_xy(Vector2f& vel, const Vector2f& accel, boo
...
+/// update the horizontal position and velocity offsets
+/// this moves the offsets (e.g _pos_offset, _vel_offset, _accel_offset) towards the targets (e.g. _pos_offset_target or _vel_offset_target)
+void AC_PosControl::update_xy_offsets()
+{
+    switch (_xy_offset_type) {
+
+    case XYOffsetType::NONE:
+        break;
+
+    case XYOffsetType::POSITION: {
```
lthall, line 645: Do we need a check to zero the position offset if we don't get an update for an extended time?

lthall (commented):
I have done my first pass and my initial focus is on just the position controller work. I think I have two general concerns:
1. We don't handle xy and z in the same way. I think this can be handled with a small change.
2. We have limited the input to just position or velocity when both these can be provided with higher order inputs ie. We can specify a moving position offset by specifying position, velocity and acceleration to achieve a more responsive offset target movement.

libraries/AC_AttitudeControl/AC_PosControl.cpp
rmackay9: ok thanks.  My concern about changing those lines is it would mean I would need to re-test surface tracking and/or terrain following

libraries/AC_AttitudeControl/AC_PosControl.h
rmackay9: ok changed

libraries/AC_AttitudeControl/AC_PosControl.h
rmackay9: ok changed to VEL_ACCEL

libraries/AC_AttitudeControl/AC_PosControl.cpp
rmackay9: ok, added _vel_offset_target and _accel_offset_target.  I wonder if we need to do this in other places too

libraries/AC_AttitudeControl/AC_PosControl.cpp
rmackay9: done, thanks.

libraries/AC_AttitudeControl/AC_PosControl.cpp
rmackay9: So if the user only wanted to set pos they could just leave vel and accel as zero right?

libraries/AC_AttitudeControl/AC_PosControl.cpp
rmackay9: I'm not sure if we need that or not.  For guided mode control (which I know is a little different) we don't timeout the position, we only time out velocity commands.

rmackay9: If we timeout the velocity offset target (and zero it), we should also zero the accel offset target at the same time right?

rmackay9: I've added a single timeout of 3 seconds for pos, vel and accel

ArduCopter/Copter.cpp
rmackay9: we've changed the interface now to allow callers to provide pos, vel and accel target offsets

IamPete1:
Do we want to log the offsets all the time? For the PD scale stuff we added a new message that was only logged if they were set (not one in that case, not zero in your case).

libraries/AC_AttitudeControl/AC_PosControl.h
```diff
@@ -253,8 +253,8 @@ class AC_PosControl
     /// set_pos_target_xy_cm - sets the position target, frame NEU in cm relative to the EKF origin
     void set_pos_target_xy_cm(float pos_x, float pos_y) { _pos_target.x = pos_x; _pos_target.y = pos_y; }
 
-    /// get_pos_target_cm - returns the position target, frame NEU in cm relative to the EKF origin
-    const Vector3p& get_pos_target_cm() const { return _pos_target; }
+    /// get_pos_target_cm - returns the position target (not including offsets), frame NEU in cm relative to the EKF origin
+    Vector3p get_pos_target_cm() const { return _pos_target - _pos_offset; }
```
tridge, line 257: this changes from a reference to a returned value, we need to change the use cases in places like mode_guided, need to change this:
```
        guided_pos_target_cm.x = pos_control->get_pos_target_cm().x;
        guided_pos_target_cm.y = pos_control->get_pos_target_cm().y;
```
to 
```
guided_pos_target_cm.xy() = pos_control->get_pos_target_cm().xy();
```

tridge (changes requested):

libraries/AC_AttitudeControl/AC_PosControl.h
tridge: should also consider moving to the cpp from the .h

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -2354,6 +2354,25 @@ function vehicle:set_circle_rate(rate_dps) end
 ---@return number|nil
 function vehicle:get_circle_radius() end
 
+-- add an offset to vehicle's target position, velocity and acceleration in autonomous modes
+---@param pos_offset_NED Vector3f_ud
+---@param vel_offset_NED Vector3f_ud
+---@param accel_offset_NED Vector3f_ud
+---@return boolean
+function vehicle:set_posvelaccel_offset(pos_offset_NED, vel_offset_NED, accel_offset_NED) end
```
IamPete1, line 2362: If you add a `AC_PosControl` singleton you can go direct. You would have to depends copter on the bindings. That means 0 cost on other vehicles and we can enable it on plane easily if we wanted to.

IamPete1 (dismissed):

rmackay9:
This was discussed on the dev call and I'll make the following changes:

- [ ] PSCx offsets will be logged to a new message PSCO (?)
- [ ] Scripting bindings will use a new position controller object (instead of going through AP_Vehicle)
- [ ] Check / change places where a reference has been replaced with an object to ensure the code is not doubling up on calculations.  also consider moving from .h to .cpp

libraries/AP_Scripting/docs/docs.lua
rmackay9: done, thanks

libraries/AC_AttitudeControl/AC_PosControl.h
rmackay9: I've done both these suggestions now, thanks

rmackay9:
I've made the suggested changes, rebased on master and re-tested and all seems well.

I think we're ready to merge this now

rmackay9:
Here is the flash cost of this feature
```

arducopter       1420 (+0.1463%)  0 (0.0000%)  4 (+0.0020%)   1420 (+0.1461%)                                   58872
antennatracker   132 (+0.0185%)   0 (0.0000%)  4 (+0.0020%)   132 (+0.0184%)                                   315792
arducopter-heli  1424 (+0.1460%)  0 (0.0000%)  0 (0.0000%)    1424 (+0.1458%)                                   53936
ardusub          1260 (+0.1410%)  0 (0.0000%)  4 (+0.0020%)   1260 (+0.1408%)                                  135928
ardurover        132 (+0.0149%)   0 (0.0000%)  -4 (-0.0020%)  132 (+0.0148%)                                   142440
blimp            120 (+0.0163%)   0 (0.0000%)  0 (0.0000%)    120 (+0.0163%)                                   296720
arduplane        1148 (+0.1133%)  0 (0.0000%)  -4 (-0.0020%)  1148 (+0.1132%)                                   16752
```

lthall (approved):
I am very happy with this one.

ArduCopter/mode_auto.cpp
```diff
@@ -1648,33 +1664,23 @@ void ModeAuto::do_land(const AP_Mission::Mission_Command& cmd)
...
+
+    // use previous waypoint destination as default if available
+    if (wp_nav->is_active() && wp_nav->reached_wp_destination()) {
+        if (!wp_nav->get_wp_destination_loc(default_loc)) {
+            // this should never happen
+            INTERNAL_ERROR(AP_InternalError::error_t::flow_of_control);
         }
     }
 
+    // get waypoint's location from command and send to wp_nav
```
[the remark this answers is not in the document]
rmackay9: I'm not sure why these old comment have only now appeared but in any case I will add a comment above loc_from_cmd in a follow-up PR (it's an existing issue)

libraries/AC_AttitudeControl/AC_PosControl_Logging.cpp
```diff
@@ -40,4 +40,26 @@ void AC_PosControl::Write_PSCD(float pos_target, float pos, float vel_desired, f
...
+                               const Vector2f& accel_offset_ne)
+{
+    const struct log_PSCO pkt{
+        LOG_PACKET_HEADER_INIT(LOG_PSCO_MSG),
+            time_us                 : AP_HAL::micros64(),
+            pos_target_offset_north : pos_target_offset_ne.tofloat().x * 0.01f,
+            pos_offset_north        : pos_offset_ne.tofloat().x * 0.01f,
+            vel_target_offset_north : vel_target_offset_ne.x * 0.01f,
+            vel_offset_north        : vel_offset_ne.x * 0.01f,
+            accel_offset_north      : accel_offset_ne.x * 0.01f,
```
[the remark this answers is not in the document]
rmackay9: logging has been updated to use a separate message for each axis

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1396,7 +1396,7 @@ void AC_PosControl::write_log()
 
         // log offsets if they have ever been used
         if (_xy_offset_type != XYOffsetType::NONE) {
-            Write_PSCO(_pos_offset_target.xy(), _pos_offset.xy(), _vel_offset_target, _vel_offset.xy(), _accel_offset.xy());
```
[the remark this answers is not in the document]
rmackay9: I think this has been resolved now

libraries/AC_WPNav/AC_WPNav.cpp
```diff
@@ -194,7 +194,7 @@ void AC_WPNav::wp_and_spline_init(float speed_cms, Vector3f stopping_point)
     if (stopping_point.is_zero()) {
         get_wp_stopping_point(stopping_point);
     }
-    _origin = _destination = stopping_point;
+    _origin = _destination = stopping_point - _pos_control.get_pos_offset_cm().tofloat();
```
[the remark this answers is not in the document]
rmackay9: this has been all fixed up now, txs!

libraries/AC_WPNav/AC_Circle.cpp
```diff
@@ -200,7 +200,7 @@ bool AC_Circle::update(float climb_rate_cms)
         target.y += - _radius * sinf(-_angle);
 
         // heading is from vehicle to center of circle
-        _yaw = get_bearing_cd(_inav.get_position_xy_cm(), _center.tofloat().xy());
+        _yaw = get_bearing_cd(_inav.get_position_xy_cm() - _pos_control.get_pos_offset_cm().xy().tofloat(), _center.tofloat().xy());
```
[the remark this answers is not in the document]
rmackay9: Leonard's fixed up how we handle offsets

PR #27786, opened 2024-08-07
Simplify Autorotation Mode Switching and Consolidate Autorotation State in RSC

MattKear:
This PR does the following:
- Removes the linking of autorotation flight mode and the interlock.
- Consolidates the autorotation state into its own class within the RSC.
- Moves the bailout logic out of the autorotation mode and into RSC making it common to both the mode and manual autorotations.

Reasons for this:
- The linking of the flight mode switching and interlock was creating lots of edge cases that were proving to difficult to prevent.  This is simpler and just like any other flight mode **(this is still SITL only)**.
- Removing this link between interlock and flight mode means, in the future, we can add partial power reduction in assisted autorotations via the flight mode which will make initial setup and testing more gradual and less intimidating.
- All bailout out handling for the manual autorotation has been tidied up and consolidated into the RSC.  The autorotation flight mode will be changed to use this bailout logic in the RSC to remove duplication.

This PR is WIP.  I still have the following to address:

- [x] Get bailouts working in autorotation flight mode.
- [x] Fixup the autotests to accept this new way of entering the autorotation flight mode.
- [x] Add autorotation flight mode exit handling e.g. don't allow exiting the flight mode to non-manual throttle modes unless the interlock is engaged and the RSC has flagged the bailout as complete.

### DIFF
--- a/ArduCopter/heli.cpp
+++ b/ArduCopter/heli.cpp
@@ -189,29 +189,25 @@ void Copter::heli_update_rotor_speed_targets()
 // to autorotation flight mode if manual collective is not being used.
 void Copter::heli_update_autorotation()
 {
-    // check if flying and interlock disengaged
-    if (!ap.land_complete && !motors->get_interlock()) {
+    bool in_autorotation_mode = false;
 #if MODE_AUTOROTATE_ENABLED
-        if (g2.arot.is_enable()) {
-            if (!flightmode->has_manual_throttle()) {
-                // set autonomous autorotation flight mode
-                set_mode(Mode::Number::AUTOROTATE, ModeReason::AUTOROTATION_START);
-            }
-            // set flag to facilitate both auto and manual autorotations
-            motors->set_in_autorotation(true);
-            motors->set_enable_bailout(true);
-        }
+    in_autorotation_mode = flightmode == &mode_autorotate;
 #endif
-        if (flightmode->has_manual_throttle() && motors->arot_man_enabled()) {
-            // set flag to facilitate both auto and manual autorotations
-            motors->set_in_autorotation(true);
-            motors->set_enable_bailout(true);
-        }
-    } else {
-        motors->set_in_autorotation(false);
-        motors->set_enable_bailout(false);
+
+    // If we have landed then we do not want to be in autorotation and we do not want to via the bailout state
+    if (ap.land_complete || ap.land_complete_maybe) {
+        motors->force_deactivate_autorotation();
+        return;
     }
 
+    // if we got this far we are flying, check for conditions to set autorotation state
+    if (!motors->get_interlock() && (flightmode->has_manual_throttle() || in_autorotation_mode)) {
+        // set state in motors to facilitate manual and assisted autorotations
+        motors->set_autorotation_active(true);
+    } else {
+        // deactivate the autorotation state via the bailout case
+        motors->set_autorotation_active(false);
+    }
 }
 
 // update collective low flag.  Use a debounce time of 400 milliseconds.
--- a/Tools/autotest/helicopter.py
+++ b/Tools/autotest/helicopter.py
@@ -700,25 +700,27 @@ class AutoTestHelicopter(AutoTestCopter):
 
     def TurbineCoolDown(self, timeout=200):
         """Check Turbine Cool Down Feature"""
+        self.context_push()
         # set option for Turbine
         RAMP_TIME = 4
         SETPOINT = 66
         IDLE = 15
         COOLDOWN_TIME = 5
-        self.set_parameter("RC6_OPTION", 161)
-        self.set_parameter("H_RSC_RAMP_TIME", RAMP_TIME)
-        self.set_parameter("H_RSC_SETPOINT", SETPOINT)
-        self.set_parameter("H_RSC_IDLE", IDLE)
-        self.set_parameter("H_RSC_CLDWN_TIME", COOLDOWN_TIME)
+        self.set_parameters({"RC6_OPTION": 161,
+                             "H_RSC_RAMP_TIME": RAMP_TIME,
+                             "H_RSC_SETPOINT": SETPOINT,
+                             "H_RSC_IDLE": IDLE,
+                             "H_RSC_CLDWN_TIME": COOLDOWN_TIME})
         self.set_rc(3, 1000)
         self.set_rc(8, 1000)
 
         self.progress("Starting turbine")
         self.wait_ready_to_arm()
+        self.context_collect("STATUSTEXT")
         self.arm_vehicle()
 
         self.set_rc(6, 2000)
-        self.wait_statustext('Turbine startup')
+        self.wait_statustext('Turbine startup', check_context=True)
 
         # Engage interlock to run up to head speed
         self.set_rc(8, 2000)
@@ -743,6 +745,7 @@ class AutoTestHelicopter(AutoTestCopter):
 
         self.set_rc(6, 1000)
         self.wait_disarmed(timeout=20)
+        self.context_pop()
 
     def TurbineStart(self, timeout=200):
         """Check Turbine Start Feature"""
--- a/Tools/autotest/helicopter.py
+++ b/Tools/autotest/helicopter.py
@@ -302,54 +302,15 @@ class AutoTestHelicopter(AutoTestCopter):
         self.progress("Lowering rotor speed")
         self.set_rc(8, 1000)
 
-    def AutoRotation(self, timeout=600):
+    def Autorotation(self, timeout=600):
         """Check engine-out behaviour"""
-        self.set_parameter("AROT_ENABLE", 1)
-        start_alt = 100 # metres
-        self.set_parameter("PILOT_TKOFF_ALT", start_alt * 100)
-        self.change_mode('POSHOLD')
-        self.set_rc(3, 1000)
-        self.set_rc(8, 1000)
-        self.wait_ready_to_arm()
-        self.arm_vehicle()
-        self.set_rc(8, 2000)
-        self.progress("wait for rotor runup to complete")
-        self.wait_servo_channel_value(8, 1659, timeout=10)
-        self.delay_sim_time(20)
-        self.set_rc(3, 2000)
-        self.wait_altitude(start_alt - 1,
-                           (start_alt + 5),
-                           relative=True,
-                           timeout=timeout)
-        self.context_collect('STATUSTEXT')
-        self.progress("Triggering autorotate by raising interlock")
-        self.set_rc(3, 1000)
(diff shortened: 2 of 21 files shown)


libraries/AC_Autorotation/AC_Autorotation.cpp
```diff
@@ -81,39 +79,30 @@ const AP_Param::GroupInfo AC_Autorotation::var_info[] = {
...
-
     // @Param: HS_SENSOR
     // @DisplayName: Main Rotor RPM Sensor 
     // @Description: Allocate the RPM sensor instance to use for measuring head speed.  RPM1 = 0.  RPM2 = 1.
     // @Units: s
     // @Range: 0.5 3
     // @Increment: 0.1
     // @User: Advanced
-    AP_GROUPINFO("HS_SENSOR", 9, AC_Autorotation, _param_rpm_instance, 0),
+    AP_GROUPINFO("HS_SENSOR", 8, AC_Autorotation, _param_rpm_instance, 0),
```
IamPete1, line 89: Changing the indexes here will result in params being loaded from eeprom wrong. Its fine to leave a gap, just leave a comment what the index 8 was used for so noone trys to use it again in the future.

libraries/AP_Motors/AP_MotorsHeli_RSC.cpp
```diff
@@ -389,38 +402,19 @@ void AP_MotorsHeli_RSC::output(RotorControlState state)
...
 {
-    int8_t ramp_time;
-    int8_t bailout_time;
-    // sanity check ramp time and enable bailout if set
-    if (_ramp_time <= 0) {
-        ramp_time = 1;
-    } else {
-        ramp_time = _ramp_time;
-    }
+    float ramp_time = MAX(float(_ramp_time.get()), 1.0);
```
IamPete1, line 405: const everything you can.

IamPete1 (commented):
just a quick look.

libraries/AC_Autorotation/AC_Autorotation.cpp
MattKear: This is SITL only at the moment though, so I figured it was ok as it is not actually on anyone's eeprom yet?

IamPete1: Ah, OK, as you say, breaking SITL for people is not such a big deal.

MattKear:
Added to DevCallEU for a quick look over/general feedback.  Not looking to get this merged today.

ArduCopter/heli.cpp
```diff
@@ -189,29 +189,25 @@ void Copter::heli_update_rotor_speed_targets()
 // to autorotation flight mode if manual collective is not being used.
 void Copter::heli_update_autorotation()
 {
-    // check if flying and interlock disengaged
-    if (!ap.land_complete && !motors->get_interlock()) {
+    bool in_autorotaion_mode = false;
```
peterbarker, line 192:
```suggestion
    bool in_autorotation_mode = false;
```

MattKear:
I am pretty happy with this PR now.

It has been tested IRL now.  For complete transparency, this code was actually tested with the code in #28209 on top.

The key bit that we need to ensure is working well, is the autorotation state inside the RSC as that code will be available to users to download on 4.6-dev once this is merged.  This plot illustrates the RSC autorotation state transitioning from "active" to "bailing out" to "stopped"  During test the vehicle bailed out from the glide smoothly. 
image

The setup that I tested in real life is using a governor in the ESC, and the autorotation signal window was used to invoke a rapid spool up.
image

I have tested this code with Heli's head speed governor, a lot in Real Flight.  It does work, however, I will say that Heli's governor does not play that nice with autorotations.  However, this is an issue with way that the governor is currently written.  The over-speed/under-speed faults are significantly more likely to be tripped when recovering from an autorotation and the governor just gives up at that point.  This is an existing issue, that is not a problem with this code being added.  I am simply raising this for awareness now, and I will address the issue with the governor in a future PR.

Also for complete transparency, since testing this IRL I have done some restructuring to move the H_RSC_AROT params to the RSC_Autorotation object.  I did not change the logic of the mode, and have re-tested in the sim.  It was just something I noticed in testing, that we should have the RSC_AROT params hidden if not enabled.

bnsgeyer:
@MattKear I agree with the premise of this PR.  It seems a little odd having to switch over to autorotate mode and then disable interlock to initiate the autorotation.  I have tested this in Real Flight sim.  I plan to look at the follow on PR this week.

bnsgeyer (approved):
@MattKear I looked through this a little more thoroughly.  It is a pretty extensive change, for sure.  A little surprised to see the new RSC_autorotation library but it looks like  it simplified a bunch in RSC and the motors library.

Did you check the cooldown feature to ensure it still works as designed.  I know there is a test for the turbine start but I'm not sure if there is one for the cooldown timer.

I am happy with this and approved but want to ensure the cooldown timer still works.

MattKear:
Thanks for taking a look at this.

Yeah, the rework was for a few reasons:
- To simplify the autorotation state within the RSC
- To handle the bailout for more RSC configs (we previously didn't handle external governors with bailout throttle very well) which is why.
- To rename the params so that they appear to users to be more broadly applicable to the autorotation in general not just manual autorotation. This has meant that the duplication of bailout handling in the flight mode and the RSC has been now been removed.

MattKear:
No, I have not tested the turbine cool down functionality.  What should I look for to see that it still works?

bnsgeyer:
@MattKear checking it is pretty easy. It is used for ICE or turbine engines.  You will need to set a non zero H_RSC_IDLE. It will raise that by 50% for the cooldown time specified in H_RSC_CLDWN in seconds after motor interlock is disabled (RSC State is Idle). You should also verify that it doesn’t use the cooldown timer feature if the aircraft is in autorotation.

MattKear:
@bnsgeyer ok, thanks Bill. I will test and report back.

libraries/AP_Motors/AP_MotorsHeli_RSC.cpp
```diff
@@ -305,46 +312,32 @@ void AP_MotorsHeli_RSC::output(RotorControlState state)
...
         _autothrottle = false;
         _governor_fault = false;
-        if (_in_autorotation) {
-            // if in autorotation, set the output to idle for autorotation. This will tell an external governor to use fast ramp for spool up.
-            // if autorotation idle is set to zero then default to the RSC idle value.
-            if (_arot_idle_output == 0) {
+
+        // check if we need to use autorotation idle throttle
+        if (autorotation.get_idle_throttle(_idle_throttle)) {
+            // if we are here then we are autorotating
```
bnsgeyer, line 318: @MattKear this guarantees the aircraft is in autorotation even if the throttle idle value is zero. I guess if you’re in our autorotation, it returns a true value for this call?

MattKear: If you are in the autorotation mode the function autorotation.get_idle_throttle() populates the _idle_throttle variable with the value set by H_RSC_AROT_IDLE and returns true.  This is then assigned to the control output and returns early.  if we are not in the autorotation then _idle_throttle is not modified and the autorotation.get_idle_throttle() returns false giving access to the subsequent cases.
In the case of both turbine and piston, they would have to set H_RSC_AROT_IDLE to match H_RSC_IDLE to maintain the engine at idle during the autorotation.

bnsgeyer: @MattKear I fixed the cooldown feature in this PR.  it required that I change the functionality of get_idle_throttle(). One because when AROT_IDLE was zero then it would use the cooldown idle.  and two, for some reason it would always use the cooldown throttle even when I used the autorotation state.  Look at what I did and let me know what  you think.
@Ferruccio1984 checkout and test the changes.  I think that works as it was meant to work.

libraries/AP_Motors/AP_MotorsHeli.cpp
```diff
@@ -147,6 +148,10 @@ const AP_Param::GroupInfo AP_MotorsHeli::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("COL_LAND_MIN", 32, AP_MotorsHeli, _collective_land_min_deg, AP_MOTORS_HELI_COLLECTIVE_LAND_MIN),
 
+    // @Group: RSC_AROT_
+    // @Path: ../AP_Motors/AP_MotorsHeli_RSC.cpp
```
bnsgeyer, line 152:
```suggestion
    // @Path: ../AC_Autorotation/RSC_Autorotation.cpp
```
This should fix the test that isn't passing CI

MattKear:
@bnsgeyer I have rebased this now that PR #28310 has gone in.

I have added more cases to autotest ManAutorotation() to check all of the conditions that you presented on Discord.

This is what the ManAutorotation test looks like:
image

**Test 1**
H_RSC_IDLE = 5, H_RSC_AROT_RAMP = 2, H_RSC_AROT_IDLE = 0, H_RSC_CLDWN_TIME = 0
image

**Test 2**
H_RSC_IDLE = 0, H_RSC_AROT_RAMP = 0, H_RSC_AROT_IDLE = 20, H_RSC_CLDWN_TIME = 0
image

**Test 3**
H_RSC_IDLE = 5, H_RSC_AROT_RAMP = 2, H_RSC_AROT_IDLE = 0, H_RSC_CLDWN_TIME = 5
image

**Test 4**
H_RSC_IDLE = 5, H_RSC_AROT_RAMP = 2, H_RSC_AROT_IDLE = 10, H_RSC_CLDWN_TIME = 5
image

MattKear:
@peterbarker I have also done the modifications that you recommended in PR #28310 to the turbine cooldown autotest.  This is kept as a separate commit.

bnsgeyer (approved):
LGTM

Tools/autotest/helicopter.py
```diff
@@ -322,85 +326,169 @@ def AutoRotation(self, timeout=600):
...
+
+        # Now we test a config that would be used with an ESC with internal governor and an autorotation window
+        self.progress("testing autorotation with ESC autorotation window config")
+        TestAutorotationConfig(self, rsc_idle=0.0, arot_ramp_time=0.0, arot_idle=20.0, cool_down=0)
+
+        # Check rsc output behavior when using the cool down feature
+        self.progress("testing autorotation with cool down enabled and zero autorotation idle")
+        TestAutorotationConfig(self, rsc_idle=5.0, arot_ramp_time=2.0, arot_idle=0, cool_down=5.0)
+
+        self.progress("testing that H_RSC_AROT_IDLE is used over RSC_IDLE when cool down is enabled")
```
peterbarker, line 488: There's also `start_subtest` which adds a bit more delineation in the logs

PR #27788, opened 2024-08-08
AP_Mission: do not use float functions on integers

peterbarker:
pitch is int8_t, yaw is int16_t

Completely untested.  Well, CI will make sure it compiles.

### DIFF
--- a/libraries/AP_Mission/AP_Mission_Commands.cpp
+++ b/libraries/AP_Mission/AP_Mission_Commands.cpp
@@ -330,8 +330,8 @@ bool AP_Mission::start_command_do_gimbal_manager_pitchyaw(const AP_Mission::Miss
     }
 
     // handle angle target
-    const bool pitch_angle_valid = !isnan(cmd.content.gimbal_manager_pitchyaw.pitch_angle_deg) && (fabsF(cmd.content.gimbal_manager_pitchyaw.pitch_angle_deg) <= 90);
-    const bool yaw_angle_valid = !isnan(cmd.content.gimbal_manager_pitchyaw.yaw_angle_deg) && (fabsF(cmd.content.gimbal_manager_pitchyaw.yaw_angle_deg) <= 360);
+    const bool pitch_angle_valid = abs(cmd.content.gimbal_manager_pitchyaw.pitch_angle_deg) <= 90;
+    const bool yaw_angle_valid = abs(cmd.content.gimbal_manager_pitchyaw.yaw_angle_deg) <= 360;
     if (pitch_angle_valid && yaw_angle_valid) {
         mount->set_angle_target(gimbal_instance, 0, cmd.content.gimbal_manager_pitchyaw.pitch_angle_deg, cmd.content.gimbal_manager_pitchyaw.yaw_angle_deg, cmd.content.gimbal_manager_pitchyaw.flags & GIMBAL_MANAGER_FLAGS_YAW_LOCK);
         return true;


IamPete1 (commented):
This must be correct. But I do wonder if this means were missing the NaN check on the command write. The incomming MAVLink could be NaN here:

https://github.com/ArduPilot/ardupilot/blob/748fe538eac62046b7cb65ec8c70c0570136db84/libraries/AP_Mission/AP_Mission.cpp#L1360-L1367

Presumably if NaN the intended behavior would be that it is not used. So we could set to a out of range value so it is ignored via the abs. I guess we would also have to return a NaN in the MAVLink read....

edit: It looks like NaNs are not in the spec, so we might be OK. https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW

peterbarker:
This is probably the wrong way to fix this.  We should probably add a `uint8_t field_invalid` field to `struct PACKED gimbal_manager_pitchyaw_Command`.

We have the same problem with the handle-rate-commands block below the one I've modified here.

rmackay9:
I'm pretty sure that the caller is meant to set the unused fields to NaN

libraries/AP_Mission/AP_Mission_Commands.cpp
```diff
@@ -330,8 +330,8 @@ bool AP_Mission::start_command_do_gimbal_manager_pitchyaw(const AP_Mission::Miss
     }
 
     // handle angle target
-    const bool pitch_angle_valid = !isnan(cmd.content.gimbal_manager_pitchyaw.pitch_angle_deg) && (fabsF(cmd.content.gimbal_manager_pitchyaw.pitch_angle_deg) <= 90);
-    const bool yaw_angle_valid = !isnan(cmd.content.gimbal_manager_pitchyaw.yaw_angle_deg) && (fabsF(cmd.content.gimbal_manager_pitchyaw.yaw_angle_deg) <= 360);
+    const bool pitch_angle_valid = abs(cmd.content.gimbal_manager_pitchyaw.pitch_angle_deg) <= 90;
```
tridge, line 333: a simple change would be to use -128 as a magic value meaning invalid for int8 rates, and set to -128 for NaN in the mavlink msg. If someone asks for -128 then instead set -127

tridge (changes requested):

nexton-winjeel:
@peterbarker 
> edit: It looks like NaNs are not in the spec, so we might be OK.

It's in the message description, not the field descriptions.

> ### MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW (1000)
> Set gimbal manager pitch/yaw setpoints (low rate command). It is possible to set combinations of the values below. E.g. an angle as well as a desired angular rate can be used to get to this angle at a certain angular rate, or an angular rate only will result in continuous turning. **NaN is to be used to signal unset.** Note: only the gimbal manager will react to this command - it will be ignored by a gimbal device. Use [GIMBAL_MANAGER_SET_PITCHYAW](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_SET_PITCHYAW) if you need to stream pitch/yaw setpoints at higher rate.

peterbarker:
We currently bounce NaN values in mission upload.  The Mount backend doesn't cope if either value is NaN.

I've created a branch which preserves NaN values in storage using INT8_MIN and INT16_MIN, but I reckon that's probably just a waste of flash space at this point.

tridge (approved):

PR #27789, opened 2024-08-08
AP_ExternalAHRS: correct VectorNAV handling of error term

peterbarker:


lashVN:
This isn't an error path, it's the nominal path.

As I read it, this is causing our response integrity checking (validating the second term of the string matches the sent string) not to happen, so a non-confirmation could be interpreted as a confirmation. In practice, this would be a strange case to happen, but theoretically could mean we've not configured the VN when we thought we did. Not critical, but certainly is a bug

IamPete1 (approved):
LGTM

PR #27793, opened 2024-08-08
Copter: Fix payload place bug

lthall:
There is a missing pos_control->update_z_controller();

Before:
image

After:
image


rmackay9 (approved):
LGTM, thanks!

rmackay9:
I've marked this for next week's dev call but I think we should merge before then because it is a relatively straight forward fix to a relatively serious bug.  If any other devs puts an approval on this then I'll likely go ahead and merge.

rmackay9:

PR #27777, opened 2024-08-07
Use ifdefs to remove more code around RC option code

peterbarker:
Moves us closer to being able to remove `default:` clauses.

```
Board               AP_Periph  blimp  bootloader  copter  heli  iofirmware  plane  rover  sub
Durandal                       *      *           *       *                 *      *      *
KakuteH7-bdshot                *      *           *       *                 *      *      *
MatekF405                      -16    *           -56     -48               -40    -16    -8
Pixhawk1-1M-bdshot             -16                -16     -16               -16    -16    -16
f303-Universal      *
iomcu                                                           *
revo-mini                      -8     *           -8      -8                -16    -8     -16

real    5m21.096s
user    133m52.460s
sys     18m32.352s
```

PR #27796, opened 2024-08-09
scripts: Fixed extract_features.py not extracting ExternalAHRS or INS Temp Cal properly

joshanne:
`extract_features.py` is not detecting/correctly extracting:
* IMU Temperature Calibration
* ExternalAHRS

Tested against a local build of CubeOrangePlus with hwdef modifications disabling the feature.

### DIFF
--- a/Tools/scripts/extract_features.py
+++ b/Tools/scripts/extract_features.py
@@ -73,8 +73,8 @@ class ExtractFeatures(object):
             ('HAL_NAVEKF3_AVAILABLE', 'NavEKF3::NavEKF3',),
             ('HAL_NAVEKF2_AVAILABLE', 'NavEKF2::NavEKF2',),
             ('HAL_EXTERNAL_AHRS_ENABLED', r'AP_ExternalAHRS::init\b',),
-            ('AP_EXTERNAL_AHRS_{type}_ENABLED', r'AP_ExternalAHRS_{type}::healthy\b',),
-            ('HAL_INS_TEMPERATURE_CAL_ENABLE', 'AP_InertialSensor::TCal::Learn::save_calibration',),
+            ('AP_EXTERNAL_AHRS_{type}_ENABLED', r'AP_ExternalAHRS_(?P<type>.*)::healthy\b',),
+            ('HAL_INS_TEMPERATURE_CAL_ENABLE', 'AP_InertialSensor_TCal::Learn::save_calibration',),
             ('HAL_VISUALODOM_ENABLED', 'AP_VisualOdom::init',),
 
             ('AP_RANGEFINDER_ENABLED', 'RangeFinder::RangeFinder',),


Tools/scripts/extract_features.py
```diff
@@ -73,14 +73,15 @@ def __init__(self, filename, nm="arm-none-eabi-nm", strings="strings"):
             ('HAL_NAVEKF3_AVAILABLE', 'NavEKF3::NavEKF3',),
             ('HAL_NAVEKF2_AVAILABLE', 'NavEKF2::NavEKF2',),
             ('HAL_EXTERNAL_AHRS_ENABLED', r'AP_ExternalAHRS::init\b',),
-            ('AP_EXTERNAL_AHRS_{type}_ENABLED', r'AP_ExternalAHRS_{type}::healthy\b',),
-            ('HAL_INS_TEMPERATURE_CAL_ENABLE', 'AP_InertialSensor::TCal::Learn::save_calibration',),
+            ('AP_EXTERNAL_AHRS_{type}_ENABLED', r'AP_ExternalAHRS_(?P<type>.*)::healthy\b',),
```
joshanne, line 76: This was not correctly detecting presence of ExternalAHRS_VectorNav

Tools/scripts/extract_features.py
```diff
@@ -73,14 +73,15 @@ def __init__(self, filename, nm="arm-none-eabi-nm", strings="strings"):
...
+            ('HAL_INS_TEMPERATURE_CAL_ENABLE', 'AP_InertialSensor_TCal::Learn::save_calibration',),
```
joshanne, line 77: This was not detecting presence of Inertial Sensor TCal

Tools/scripts/extract_features.py
```diff
@@ -142,7 +143,7 @@ def __init__(self, filename, nm="arm-none-eabi-nm", strings="strings"):
 
             ('HAL_PARACHUTE_ENABLED', 'AP_Parachute::update',),
             ('AP_FENCE_ENABLED', r'AC_Fence::check\b',),
-            ('HAL_RALLY_ENABLED', r'AP_Rally::get_rally_max\b',),
+            ('HAL_RALLY_ENABLED', 'AP_Rally::find_nearest_rally_point',),
```
joshanne, line 146: I noticed Rally was not being detected by default? `get_rally_max` is not showing when running `arm-none-eabi-nm --demangle --print-size` - so I've moved this to a function that is visible and used.

Tools/scripts/extract_features.py
```diff
@@ -257,7 +258,7 @@ def __init__(self, filename, nm="arm-none-eabi-nm", strings="strings"):
             ('FORCE_APJ_DEFAULT_PARAMETERS', 'AP_Param::param_defaults_data'),
             ('HAL_BUTTON_ENABLED', 'AP_Button::update'),
             ('HAL_LOGGING_ENABLED', 'AP_Logger::init'),
-            ('AP_COMPASS_CALIBRATION_FIXED_YAW_ENABLED', 'AP_Compass::mag_cal_fixed_yaw'),
+            ('AP_COMPASS_CALIBRATION_FIXED_YAW_ENABLED', 'Compass::mag_cal_fixed_yaw'),
```
joshanne, line 261: This was not detecting properly. The class is `Compass`, not `AP_Compass` - perhaps `Compass` should be moved to `AP_Compass` at some point?

Tools/scripts/build_options.py
```diff
@@ -34,6 +34,7 @@ def __init__(self,
     Feature('AHRS', 'MicroStrain5', 'AP_EXTERNAL_AHRS_MICROSTRAIN5_ENABLED', 'Enable MICROSTRAIN 5-series External AHRS', 0, "AHRS_EXT"),  # noqa: E501
     Feature('AHRS', 'MicroStrain7', 'AP_EXTERNAL_AHRS_MICROSTRAIN7_ENABLED', 'Enable MICROSTRAIN 7-series External AHRS', 0, "AHRS_EXT"),  # noqa: E501
     Feature('AHRS', 'AHRS_EXT_VECTORNAV', 'AP_EXTERNAL_AHRS_VECTORNAV_ENABLED', 'Enable VectorNav External AHRS', 0, "AHRS_EXT"),  # noqa
+    #Feature('AHRS', 'InertialLabs', 'AP_EXTERNAL_AHRS_INERTIAL_LABS_ENABLED', 'Enable InertialLabs External AHRS', 0, "AHRS_EXT"), # disabled because INERTIAL_LABS is the enable flag
```
joshanne, line 37: There are 9 instances of `AP_EXTERNAL_AHRS_INERTIAL_LABS_ENABLED` in the code base - I'd like to change them to `AP_EXTERNAL_AHRS_INERTIALLABS_ENABLED` so this line can be uncommented and changed correctly.

We can either detect the InertialLabs class correctly, but not use this `build_option` feature externally, OR, incorrectly detect the feature, but correctly use the `build_options` to generate a file.

What should it be?

joshanne: Of course, this would potentially break peoples custom build servers...

Could rename the class. 

Is there a way to map extracted feature classes to a different define name? I couldn't see one

peterbarker:
Thanks for this.  I've just written patches here that allow `./Tools/autotest/test_build_features.py` to detect these.  Those are lurking in https://github.com/ArduPilot/ardupilot/pull/27777

Tools/scripts/build_options.py
peterbarker: I'm fine with this rename

Tools/scripts/extract_features.py
peterbarker: Yes, it should.  I made that change for OpticalFlow, I think.

Tools/scripts/extract_features.py
peterbarker: Compiler was probably inlining; you could try recompiling with a different optimisation level to see if that's true.  But it can't possibly be being elided given where it is used!

Tools/scripts/extract_features.py
peterbarker: Whoops.

peterbarker (approved):

peterbarker:
Merged, thanks!

Tools/scripts/build_options.py
joshanne: Excellent. I'll push a patch for that shortly.

PR #27802, opened 2024-08-10
AUAV Airspeed Sensor Driver

Tdogb:
Drivers for the Amphenol AUAV airspeed sensor. This is a relatively new UAV specific airspeed sensor. I don't know why Amphenol didn't write their own Ardupilot drivers, but I did it for them since this looks like a really awesome sensor.

https://www.mouser.com/new/amphenol/amphenol-auav-air-speed-altitude-sensors/


MallikarjunSE:
These actually provide absolute pressure measurement too.

Tdogb:
> These actually provide absolute pressure measurement too.

Nice! Although I'm going to leave that to someone else to implement if they want :)

libraries/AP_HAL_ChibiOS/hwdef/include/minimize_common.inc
```diff
@@ -127,6 +127,7 @@ define AP_AIRSPEED_ANALOG_ENABLED 1
 define AP_AIRSPEED_MS5525_ENABLED 1
 define AP_AIRSPEED_SDP3X_ENABLED 1
 define AP_AIRSPEED_NMEA_ENABLED 1  # additional checks for vehicle type in .cpp
+define AP_AIRSPEED_AUAV_ENABLED 1
```
Hwurzburg, line 130: uncommon airspeed drivers should not be included in minimize

Hwurzburg (changes requested):
needs extract_features.py also

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
```diff
@@ -0,0 +1,231 @@
...
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+  backend driver for airspeed from a I2C AUAV sensor
+ */
+#include "AP_Airspeed_AUAV.h"
+
+#if AP_AIRSPEED_AUAV_ENABLED
```
peterbarker, line 21:
```suggestion
#include "AP_Airspeed_config.h"

#if AP_AIRSPEED_AUAV_ENABLED

#include "AP_Airspeed_AUAV.h"
```

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include <utility>
+
+extern const AP_HAL::HAL &hal;
+
+
+AP_Airspeed_AUAV::AP_Airspeed_AUAV(AP_Airspeed &_frontend, uint8_t _instance) :
+    AP_Airspeed_Backend(_frontend, _instance)
+{
+}
+
```
peterbarker, line 40:
```suggestion
```

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
```diff
@@ -0,0 +1,231 @@
...
+// probe for a sensor
+bool AP_Airspeed_AUAV::probe(uint8_t bus, uint8_t address)
+{
+    _dev = hal.i2c_mgr->get_device(bus, address);
+    if (!_dev) {
```
peterbarker, line 45:
```suggestion
    if (_dev == nullptr) {
```

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
```diff
@@ -0,0 +1,231 @@
...
+    _pressure = 250+1.25*(pressure-(0.1f*pow(2,23))/pow(2,24))*1000;
+    return true;
+}
+
+// return the current temperature in degrees C, if available
+bool AP_Airspeed_AUAV::get_temperature(float &_temperature)
+{
+    WITH_SEMAPHORE(sem);
+    _temperature = 0;
+    return true;
```
peterbarker, line 228:
```suggestion
    return false;
```

peterbarker (changes requested):

peterbarker:
Ping @Tdogb - also needs to pass CI :-)

Tdogb:
Hey guys, I think it would be super cool to implement the absolute pressure sensing capability. Does anyone have strong opinions on how this is implemented? I don't want someone to unknowingly have it override their baro

tridge:
just as an airspeed sensor initially is the way to go

peterbarker:
@Tdogb did you need assistance on this one, or still plugging away at it?

MallikarjunSE:
@peterbarker i tried this pr. Was able to get differential pressure output. But temperature didn't work for me.

Tdogb:
> @peterbarker i tried this pr. Was able to get differential pressure output. But temperature didn't work for me.

Temperature is compensated for but the real temperature value isn't read because there isn't documentation on it.

I had accidentally pushed some other dev code I have to this branch so I removed it. Although if anyone has interest in using multiple of the same i2c sensors, I can revive the code, it works well.

libraries/AP_Airspeed/AP_Airspeed.cpp
```diff
@@ -441,6 +442,12 @@ void AP_Airspeed::allocate()
             sensor[i] = NEW_NOTHROW AP_Airspeed_External(*this, i);
 #endif
             break;
+        case TYPE_AUAV:
+#if AP_AIRSPEED_AUAV_ENABLED
+            sensor[i] = NEW_NOTHROW AP_Airspeed_AUAV(*this, i);
+#endif
+            break;
+            break;
```
peterbarker, line 450:
```suggestion
            break;
```

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
```diff
@@ -0,0 +1,329 @@
...
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+  backend driver for airspeed from a I2C AUAV sensor
+ */
+#include "AP_Airspeed_AUAV.h"
+#include "AP_Logger/AP_Logger.h"
+#if AP_AIRSPEED_AUAV_ENABLED
```
peterbarker, line 21:
```suggestion
#include "AP_Airspeed_config.h"

#if AP_AIRSPEED_AUAV_ENABLED

#include "AP_Airspeed_AUAV.h"
#include "AP_Logger/AP_Logger.h"
```

peterbarker (changes requested):
There are still changes requested from my previous review

libraries/AP_HAL_ChibiOS/hwdef/include/minimize_common.inc
tridge: fixed

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
tridge: needs constructor for the pressure range in fixed driver

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
tridge: fixed

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
tridge: can't do that with OwnPtr

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
tridge: fixed to return real temperature

libraries/AP_Airspeed/AP_Airspeed.cpp
tridge: fixed

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
tridge: fixed

tridge:
@MallikarjunSE @Tdogb I've force pushed some fixes:
 - temperature now works
 - supports 3 pressure ranges
 - removed the absolute code
 - use 8x sampling for lower noise

I've bench tested buy not flight tested. Please test!

MallikarjunSE:
> @MallikarjunSE @Tdogb I've force pushed some fixes:
> 
> * temperature now works
> * supports 3 pressure ranges
> * removed the absolute code
> * use 8x sampling for lower noise
> 
> I've bench tested buy not flight tested. Please test!

Hey @tridge 
Tested on DroneCAN periph device. Works! 
Thank you! 
image

MallikarjunSE:
@tridge  tested on L30D. Works! 
image

MallikarjunSE:
Hey @tridge 
below is temperature comparison between DroneCAN periph vs I2C directly plugged in. Temperature update rate seems to be too low on DroneCAN? Or it's sufficient?

image

tridge:
> Temperature update rate seems to be too low on DroneCAN? Or it's sufficient?

the reason the graph looks weird is the DroneCAN RawAirData message uses a float16 for temperature, which means it doesn't have much resolution. It is sufficient for how it is used though

Tdogb:
Thanks for helping push this across the finish line guys! 

I have some code on my personal branch for absolute pressure, however, I've been running into issues with the sensor not acking some of the calibration coefficients. Looking at it on the scope makes it seem like an issue with the sensor itself. I'll send photos when I get home

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
```diff
@@ -0,0 +1,246 @@
...
+bool AP_Airspeed_AUAV::init()
+{
+    if (bus_is_configured()) {
+        // the user has configured a specific bus
+        if (probe(get_bus(), AUAV_AIRSPEED_I2C_ADDR)) {
+            goto found_sensor;
+        }
+    } else {
+        // if bus is not configured then fall back to the old
+        // behaviour of probing all buses, external first
```
peterbarker, line 78: "same behaviour as other dirvers"?

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
```diff
@@ -0,0 +1,246 @@
...
+        FOREACH_I2C_EXTERNAL(bus) {
+            if (probe(bus, AUAV_AIRSPEED_I2C_ADDR)) {
+                goto found_sensor;
+            }
+        }
+        FOREACH_I2C_INTERNAL(bus) {
```
peterbarker, line 84: I wonder if we should actually have a force-probe-internal-buses we can enable on those boards which screw this up?  Might save flash on boards which are not stuffed...

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
```diff
@@ -0,0 +1,246 @@
...
+    measurement_started_ms = 0; // It should always get reset by _measure. This is a safety to handle failures of i2c bus
+    uint8_t inbuf[7];
+    if (!_dev->read((uint8_t *)&inbuf, sizeof(inbuf))) {
+        return;
+    }
+    const int32_t Tref_Counts = 7576807; // temperature counts at 25C
+    const float TC50Scale = 100.0f * 100.0f * 167772.2f; // scale TC50 to 1.0% FS0
+    float AP3, BP2, CP, Corr, Pcorr, Pdiff, TC50, Pnfso, Tcorr, PCorrt;
+    int32_t iPraw, Tdiff, iTemp;
+    uint32_t PComp;
```
peterbarker, line 130: Would be nice to initialise these as we declare them.

Hwurzburg (approved):

tridge:
@Tdogb @MallikarjunSE I've changed it to require ARSPD_BUS to be set correctly, which is how other drivers work except for the MS4525 (which was a legacy compatibility hack)
I've also added status checking so the probe is more reliable, and ensures that we don't use invalid data

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
tridge: done

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
tridge: i've changed it to require the bus to be set

libraries/AP_Airspeed/AP_Airspeed_AUAV.cpp
tridge: fixed

MallikarjunSE:
> @Tdogb @MallikarjunSE I've changed it to require ARSPD_BUS to be set correctly, which is how other drivers work except for the MS4525 (which was a legacy compatibility hack) I've also added status checking so the probe is more reliable, and ensures that we don't use invalid data

Tested with CubeOrange and Periph device and works. I2C bus selection works as expected. Previously it used to probe without setting bus number. Now works with bus number set correctly. 
image

criegit:
tested it with L05D on the bench with Pix32 v6 and Matek L431-Cannode. Seems to work very good.
Comparisons with DLVR sensor and with a reference pressure system planned in the near future.

I did my own driver integration of this airspeed sensor to learn how to do that. I was not that far off, so thanks a lot for the code to compare, correct and learn!

I would love to participate in the barometer integration as much as i can. (limited)
As far as i understand, the barometer option of the board are as of now hardcoded in the hwdef file.
So there would be a easy way to implement this sensor for an AP_Periph device that ships with this sensors, but this does notwork for flexible selection of sensors in the Autopilot Parameter list.

I hope to be more present in the upcoming devtalks.

IamPete1:
I have done a PR to add a baro driver. https://github.com/ArduPilot/ardupilot/pull/29535

However, my sensor always fails to read the extended calibration coefficients for that channel despite working fine on the differential channel. I can do builds if anyone else wants to give it a try.

Tdogb:
> I have done a PR to add a baro driver. https://github.com/ArduPilot/ardupilot/pull/29535
> 
> 
> 
> However, my sensor always fails to read the extended calibration coefficients for that channel despite working fine on the differential channel. I can do builds if anyone else wants to give it a try.

I have independently written a driver for abs pressure and have ran into the exact issue. Putting it on the scope the fc is outputting the correct command but the sensor just doesn't respond, it seems like an issue with the sensor itself.

IamPete1:
> I have independently written a driver for abs pressure and have ran into the exact issue. Putting it on the scope the fc is outputting the correct command but the sensor just doesn't respond, it seems like an issue with the sensor itself.

I'm not going crazy then! I did find that it works fine if we just don't bother with the extended calibration. Maybe that is a solution,  but it is certainly strange. 

I did notice the other team has a driver, I wonder if it works there. https://github.com/PX4/PX4-Autopilot/blob/main/src/drivers/differential_pressure/auav/AUAV.cpp

IamPete1:
@Tdogb I now have the baro driver working. It turns out its very sensitive to timings because its clocked much slower than the differential channel for some reason. https://github.com/ArduPilot/ardupilot/pull/29535

PR #27803, opened 2024-08-10
Copter: Suppress multiple anomaly alerts

muramura:
The telemetry transmitter uses UART at 57600 BPS or 38400 BPS.
The signal lines are only TX and RX for the telemetry transmitter.
Sending many messages in a short time is prone to overflow.
If an error is detected in the order, I don't think other processing is necessary.
If it gets better, execute the next process.


peterbarker:
Emitting multiple failures is actually a deliberate - requested! - feature.

These tests are independent, too - one failure doesn't mean the others will.

Sorry, but this isn't how things are desired.

muramura:
@peterbarker san.

It means that you are not assuming multiple errors will occur.

The pre-arm check will not proceed to the next check if even one abnormality is detected.

I think it's not good to check only this part.

PR #27804, opened 2024-08-10
AP_Math: Remove heap allocations from mat_inverse

Horolsky:
This change removes redundant allocations in mat_inverse while keeping compatible API.
The main overload args are extended with the buffer ptr managed by the caller.  If ptr is null, it will be allocated on heap.
Additional template overload deduces the static array capacity, keeping the default buffer on stack for all use cases.


amilcarlucas:
```
antennatracker   -156 (-0.0117%)  0 (0.0000%)  4 (+0.0015%)   -156 (-0.0117%)                                  633372
arduplane        -156 (-0.0087%)  0 (0.0000%)  4 (+0.0015%)   -156 (-0.0087%)                                  168232
blimp            -156 (-0.0115%)  0 (0.0000%)  -4 (-0.0015%)  -156 (-0.0115%)                                  611160
arducopter-heli  -156 (-0.0087%)  0 (0.0000%)  4 (+0.0015%)   -156 (-0.0087%)                                  165800
arducopter       -156 (-0.0087%)  0 (0.0000%)  -4 (-0.0015%)  -156 (-0.0087%)                                  164936
ardusub          -156 (-0.0098%)  0 (0.0000%)  -4 (-0.0015%)  -156 (-0.0098%)                                  373684
ardurover        -156 (-0.0095%)  0 (0.0000%)  -4 (-0.0015%)  -156 (-0.0095%)                                  315880
```

libraries/AP_Common/BufferHandler.h
```diff
@@ -0,0 +1,73 @@
...
+/// Works as a non-owning view on a buffer ptr provided by the caller,
+/// or manages the buffer lifetime on the heap when the nullptr is given.
+/// @tparam T data type
+/// @tparam Card cardinality, num of logical blocks in the buffer
+template <class T, size_t Card = 1>
+class BufferHandler
+{
+    bool const kAllocateHeap;
+    size_t const kBlockSize;
+    T *ptr_;
```
peterbarker, line 30: Why the trailing underscore?

libraries/AP_Common/BufferHandler.h
```diff
@@ -0,0 +1,73 @@
...
+
+public:
+
+    BufferHandler(T* buf, size_t const blocksize)
+        : kAllocateHeap{nullptr == buf}
```
peterbarker, line 35: We don't usually reverse these comparisons

libraries/AP_Common/BufferHandler.h
```diff
@@ -0,0 +1,73 @@
...
+template <class T, size_t Card = 1>
+class BufferHandler
+{
+    bool const kAllocateHeap;
+    size_t const kBlockSize;
+    T *ptr_;
+
+public:
+
+    BufferHandler(T* buf, size_t const blocksize)
```
peterbarker, line 34:
```suggestion
    BufferHandler(T* buf, const size_t blocksize)
```

libraries/AP_Common/BufferHandler.h
```diff
@@ -0,0 +1,73 @@
...
+    BufferHandler(size_t const blocksize) : BufferHandler(nullptr, blocksize) {}
+
+    ~BufferHandler()
+    {
+        if (kAllocateHeap) {
+            delete[] ptr_;
+        }
+    }
+
+    BufferHandler(BufferHandler const& o) : BufferHandler(o.ptr_, o.kBlockSize)
```
peterbarker, line 50: Do we really want to allow copying?

libraries/AP_AccelCal/AccelCalibrator.cpp
```diff
@@ -333,6 +333,7 @@ void AccelCalibrator::run_fit(uint8_t max_iterations, float& fitness)
     union param_u fit_param = _param;
     uint8_t num_iterations = 0;
 
+    MatInvBuffer<float> buffer {ACCEL_CAL_MAX_NUM_PARAMS*ACCEL_CAL_MAX_NUM_PARAMS};
```
peterbarker, line 336: So if allocation fails here then we just rumble on?

I haven't spotted handling within the new class for that - but not handling the allocation failure here seems bad.

libraries/AP_Common/BufferHandler.h
Horolsky: deleted

libraries/AP_Common/BufferHandler.h
Horolsky: updated

libraries/AP_Common/BufferHandler.h
Horolsky: done

libraries/AP_AccelCal/AccelCalibrator.cpp
Horolsky: Hi, that is a good question that I wanted to bring to the review. The original mat_inverse implementation has the same issue,  there are 7 allocations with unhandled failures: ([link](https://github.com/ArduPilot/ardupilot/blob/60c2e4278cc38591187b8a188677bd66ac8f210a/libraries/AP_Math/matrix_alg.cpp#L202)).

Fixing this would require more changes in the dependent modules, so I guess it could be done in a separate pr. I can take that task too, but I'm new to ArduPilot, and I need some support in clarifying the recovery strategy for allocation failures.

Horolsky: I guess the solution is to reserve memory on client class initialization and handle failures accordingly on that step. Another option is to place it on stack, with the price of increasing frame size.

libraries/AP_Math/matrix_alg.cpp
```diff
@@ -149,71 +126,71 @@ static void mat_back_sub(const T *U, T *out, uint16_t n)
...
-static bool mat_inverseN(const T* A, T* inv, uint16_t n)
+static bool mat_inverseN(const T* A, T* inv, const uint16_t n, MatInvBuffer<T> &buf)
 {
-    T *L, *U, *P;
-    bool ret = true;
-    L = NEW_NOTHROW T[n*n];
-    U = NEW_NOTHROW T[n*n];
-    P = NEW_NOTHROW T[n*n];
-    mat_LU_decompose(A,L,U,P,n);
+    T *L      = buf.L();
```
tridge, line 173: i think it would be a lot simpler to declare a structure with the elements it needs, then allocation that with a single NEW_NOTHROW, and return false on failure
I'm just concerned this PR spreads the logic of the mat inverse across 3 files, with entanglement between them, with the U(), P() etc methods in AP_Math.h, which are of specific sizes needed by this code

tridge (changes requested):

PR #27805, opened 2024-08-11
AP_GPS: fixed unconfigured uBlox pre-arm

tridge:
when GPS_GNSS_MODE is zero (the default) we could get stuck trying to configure the GNSS for F9P


andyp1per (approved):
Thanks!

PR #27812, opened 2024-08-12
AP_Scripting: correct compilation when AP_SCRIPTING_SERIALDEVICE_ENAB…

peterbarker:
…LED false

### DIFF
--- a/libraries/AP_Scripting/AP_Scripting.h
+++ b/libraries/AP_Scripting/AP_Scripting.h
@@ -14,6 +14,8 @@
  */
 #pragma once
 
+#include "AP_Scripting/AP_Scripting_config.h"
+
 #if AP_SCRIPTING_ENABLED
 
 #include <GCS_MAVLink/GCS_config.h>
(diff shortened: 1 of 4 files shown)


peterbarker:
> Can you clarify in what cases this fails? I thought I tested that setting it to disabled didn’t break anything but I guess not across all possible configurations.

image

Trying to compile this when the serial manager doesn't allow port registration...

> Also, should we add `AP_SCRIPTING_ENABLED` as a condition to enable `AP_SCRIPTING_SERIALDEVICE_ENABLED` and remove the tests for `AP_SCRIPTING_ENABLED`?

Where?  It is odd that AP_SCRIPTING_SERIALDEVICE_ENABLED isn't dependent on scripting, and I did see that.  But I was after a minimal-change here.

tpwrules:
You are right, I probably did not test with `AP_SERIALMANAGER_REGISTER_ENABLED` set to 0. I was confused by the commit title.

As for where, just search for `AP_SCRIPTING_SERIALDEVICE_ENABLED` across the codebase, various places it's used also check `AP_SCRIPTING_ENABLED`.

tridge (approved):

PR #27815, opened 2024-08-12
AP_Logger: correctly read file header for block logger

andyp1per:
This fixes the issue where log files do not have timestamps when using flash logging


rmackay9:

PR #27818, opened 2024-08-12
Feature/alpine: fix compiliation and add test

khancyr:
add environement setup and test for alpine linux
fix build on alpine due to missing include for timeval struct (same issue as in https://github.com/ArduPilot/ardupilot/pull/27518)


peterbarker:
Your new test needs to pass :-)

peterbarker:
Merged, thanks!

PR #27819, opened 2024-08-12
AP_ExternalAHRS: Fix Build Option for InertialLabs

joshanne:
Opened after fixing some other build_options/extract_features in https://github.com/ArduPilot/ardupilot/pull/27796

Extraction of the feature generates `INERTIALLABS` but the define uses `INERTIAL_LABS`.

This fixes it so the feature matches the build option.


joshanne:
@peterbarker following up from last weeks discussions.

joshanne:
Looks like FenceFloorEnabledLanding is flapping.

peterbarker (approved):
LGTM.

Note that I have https://github.com/ArduPilot/ardupilot/pull/27777 on the way which, while a bit of a monster, will catch more of these issues in CI.

peterbarker:
Merged, thanks!

Issue #27632, opened 2024-07-24
ublox GPS reports fix type of 4


Baudouin0:
**IF YOU DON'T REMOVE THESE FOUR LINES, THEN YOUR REPORT WILL BE CLOSED AUTOMATICALLY**
Questions and user problems should be directed at the forum (http://discuss.ardupilot.org)

ublox GPS reports fix type of 4 - this is GNSS+dead reckoning as the puck contains acceleromters.
If there was a previous discussion in the forum, link to it

## Bug report

**Issue details**

Debug terminal reports 'Unexpected state 33 or 35'

**Version**
4.5.4

**Platform**
[  ] All
[  ] AntennaTracker
[ x ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
F450

**Hardware type**
PixHawk 2.4.8

**Logs**
Unexpected state 33.

Baudouin0:
AP_GPS_UBLOX.cpp:
Make case 3 and 4 the same.

        switch (_buffer.pvt.fix_type)
        {
            case 0:
                state.status = AP_GPS::NO_FIX;
                break;
            case 1:
                state.status = AP_GPS::NO_FIX;
                break;
            case 2:
                state.status = AP_GPS::GPS_OK_FIX_2D;
                break;
            case 3:
            case 4:
                state.status = AP_GPS::GPS_OK_FIX_3D;
                if (_buffer.pvt.flags & 0b00000010)  // diffsoln
                    state.status = AP_GPS::GPS_OK_FIX_3D_DGPS;
                if (_buffer.pvt.flags & 0b01000000)  // carrsoln - float
                    state.status = AP_GPS::GPS_OK_FIX_3D_RTK_FLOAT;
                if (_buffer.pvt.flags & 0b10000000)  // carrsoln - fixed
                    state.status = AP_GPS::GPS_OK_FIX_3D_RTK_FIXED;
                break;
            case 5:
                state.status = AP_GPS::NO_FIX;
                break;
            default:
                state.status = AP_GPS::NO_FIX;
                break;
        }

peterbarker:
Which physical device is this?

The patch you propose here looks like it's probably correct.  Did you want to create a PR for it?

Baudouin0:
It’s a Ublox M8N with built in compass.
NEO-M8 series | u-blox<https://www.u-blox.com/en/product/neo-m8-series>

[...]

PR #27825, opened 2024-08-13
AP_GPS: understand uBlox fix_type of 4 as being GNSS and dead reckoning

peterbarker:
Closes https://github.com/ArduPilot/ardupilot/issues/27632

This L series seems to be able to suck sensor data in to provide an inertial solution.

I'm actually a little concerned about feeding this estimate into our EKF.  Wouldn't the EKF's solution end up with measured acceleration applied twice, once from ArduPilot's sensors, once by whatever is feeding the GPS unit?


tridge:
let's try and get a flight log and look at the innovations, preferably with some movement, so CIRCLE mode, or some significant flight under pilot control

MallikarjunSE:
It could be using this. 
https://www.u-blox.com/en/product/ubx-m9140-ka-dr-chip
Recent DJI drones use this which has built in IMU.

PR #27827, opened 2024-08-13
Plane: Reset TECS along with the rest of the controllers

Georacer:
For the balloon drop use-case, I've noticed that TECS would initialize on the ground. That meant that the desired altitude (which is under an LPF) would not start close to the actual altitude, rendering itself useless for an RTL:
image

With this commit TECS starts at the release altitude, resulting in reasonable altitude error:
image

:warning: Requires https://github.com/ArduPilot/ardupilot/pull/27758 .

### DIFF
--- a/ArduPlane/mode.cpp
+++ b/ArduPlane/mode.cpp
@@ -255,6 +255,9 @@ void Mode::reset_controllers()
     // reset steering controls
     plane.steer_state.locked_course = false;
     plane.steer_state.locked_course_err = 0;
+
+    // reset TECS
+    plane.TECS_controller.reset();
 }
 
 bool Mode::is_taking_off() const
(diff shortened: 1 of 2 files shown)


tridge:
@Georacer this seems good, will you take it out of draft?

Georacer:
> @Georacer this seems good, will you take it out of draft?

If you're convinced by a draft PR and in my absence, maybe I should attend the Dev call less, haha.

Yes, I'll convert it to a normal PR.

MattKear (approved):
This looks to me.  You mention a dependency on #27758.  Is that the case?

Georacer:
> This looks to me. You mention a dependency on #27758. Is that the case?

There is a "shared change" in TECS.cpp that is necessary to get the results that I pasted in the images.

tridge (approved):

PR #27830, opened 2024-08-13
SITL: reboot at different location

magicrub:
This allows including a Lat/long/alt and heading into the PREFLIGHT_REBOOT_SHUTDOWN message to then override the home position that is used when launching the next SITL instance. The helper function is flexible to allow not just editing existing, but adding new entries for future use.

Requirements:

- p1 = 1
- p2 = 0
- p3 = 0
- p4 = valid heading: 0 - 360
- p5/x = valid latitude: +/- 90
- p6/y = valid longitude: +/- 180
- p7/z = valid altitude in meters: +/- (2^23cm / 100)
- non-zero position
- longitude isn't the magic force reboot value

### DIFF
--- a/libraries/AP_HAL_SITL/HAL_SITL_Class.cpp
+++ b/libraries/AP_HAL_SITL/HAL_SITL_Class.cpp
@@ -305,6 +305,22 @@ void HAL_SITL::run(int argc, char * const argv[], Callbacks* callbacks) const
     actually_reboot();
 }
 
+void HAL_SITL::update_new_args(const char* opt, const char* val)
+{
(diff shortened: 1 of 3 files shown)


magicrub:
for a custom GCS I'm working with, they reboot the sim all the time in the cloud. Lots of sims running. Would be nice to not have to kill and restart the app when a simple reboot command solves the problem

peterbarker:
> for a custom GCS I'm working with, they reboot the sim all the time in the cloud. Lots of sims running. Would be nice to not have to kill and restart the app when a simple reboot command solves the problem

You know that you can set the SITL starting position persistently in parameters?

peterbarker:
Ping @magicrub - you can set the startup location persistently...

PR #27831, opened 2024-08-13
Sub: remove ENABLE/DISABLE defines, use 0/1 instead

peterbarker:
same as Plane, Rover, Sub, Tracker

No compiler output change across any target (except the SITL ones...)


Williangalvani (approved):
looks good, thanks!

PR #27832, opened 2024-08-13
Tools: LogAnalyzer: remove

peterbarker:
the web-based tools are supplanting this

Tested this on a couple of modern (and not-so-modern onboard logs and it's broken in several ways.


khancyr:
Agreed!

rmackay9 (approved):
I think this is the "Auto Analysis" in MP and it's badly broken there too

peterbarker:
> I think this is the "Auto Analysis" in MP and it's badly broken there too

This would infer the removal of "Auto Analysis" in MissionPlanner.

PR #27839, opened 2024-08-14
Refactor attitude control to be thread safe

andyp1per:
Broken out from https://github.com/ArduPilot/ardupilot/pull/27029

The essence of these changes is to make sure that the interaction between the attitude controller and the rate controller are thread safe. This involves avoiding setting object state as intermediate values but instead publishing at the end of a calculation. It also means that changes should be accepted as vectors rather than discrete axes. Experiments were also done using locking but this proved to be moderately expensive at high update rates and did not improve the overall effect of this change.

### DIFF
--- a/ArduCopter/Attitude.cpp
+++ b/ArduCopter/Attitude.cpp
@@ -15,7 +15,9 @@ void Copter::run_rate_controller()
     pos_control->set_dt(last_loop_time_s);
 
     // run low level rate controllers that only require IMU data
-    attitude_control->rate_controller_run(); 
+    attitude_control->rate_controller_run();
+     // reset sysid and other temporary inputs
+    attitude_control->rate_controller_target_reset();
 }
 
 /*************************************************************
--- a/ArduCopter/Attitude.cpp
+++ b/ArduCopter/Attitude.cpp
@@ -16,7 +16,7 @@ void Copter::run_rate_controller()
 
     // run low level rate controllers that only require IMU data
     attitude_control->rate_controller_run();
-     // reset sysid and other temporary inputs
+    // reset sysid and other temporary inputs
     attitude_control->rate_controller_target_reset();
 }
 
--- a/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
+++ b/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
@@ -188,15 +188,16 @@ float AC_AttitudeControl::get_slew_yaw_max_degs() const
 // Ensure attitude controller have zero errors to relax rate controller output
 void AC_AttitudeControl::relax_attitude_controllers()
 {
+    // take a copy of the last gyro used by the rate controller before using it
+    Vector3f gyro = _rate_gyro;
     // Initialize the attitude variables to the current attitude
     _ahrs.get_quat_body_to_ned(_attitude_target);
     _attitude_target.to_euler(_euler_angle_target);
     _attitude_ang_error.initialise();
 
     // Initialize the angular rate variables to the current rate
-    _ang_vel_target = _ahrs.get_gyro();
+    _ang_vel_target = gyro;
     ang_vel_to_euler_rate(_attitude_target, _ang_vel_target, _euler_rate_target);
-    _ang_vel_body = _ahrs.get_gyro();
 
     // Initialize remaining variables
     _thrust_error_angle = 0.0f;
@@ -208,6 +209,8 @@ void AC_AttitudeControl::relax_attitude_controllers()
 
     // Reset the I terms
     reset_rate_controller_I_terms();
+    // finally update the attitude target
+    _ang_vel_body = gyro;
 }
 
 void AC_AttitudeControl::reset_rate_controller_I_terms()
@@ -531,6 +534,7 @@ void AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_2(float roll_rate_bf_cds, 
     _attitude_target.to_euler(_euler_angle_target);
     // Convert body-frame angular velocity into euler angle derivative of desired attitude
     ang_vel_to_euler_rate(_attitude_target, _ang_vel_target, _euler_rate_target);
+    // finally update the attitude target
     _ang_vel_body = _ang_vel_target;
 }
 
@@ -554,7 +558,7 @@ void AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_3(float roll_rate_bf_cds, 
         _attitude_ang_error.from_axis_angle(attitude_error);
     }
 
-    Vector3f gyro_latest = _ahrs.get_gyro_latest();
+    Vector3f gyro_latest = _rate_gyro;
     attitude_ang_error_update_quat.from_axis_angle((_ang_vel_target - gyro_latest) * _dt);
     _attitude_ang_error = attitude_ang_error_update_quat * _attitude_ang_error;
     _attitude_ang_error.normalize();
@@ -582,8 +586,11 @@ void AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_3(float roll_rate_bf_cds, 
 
     // Compute the angular velocity target from the integrated rate error
     _attitude_ang_error.to_axis_angle(attitude_error);
-    _ang_vel_body = update_ang_vel_target_from_att_error(attitude_error);
-    _ang_vel_body += _ang_vel_target;
+    Vector3f ang_vel_body = update_ang_vel_target_from_att_error(attitude_error);
+    ang_vel_body += _ang_vel_target;
+
+    // finally update the attitude target
+    _ang_vel_body = ang_vel_body;
 }
 
 // Command an angular step (i.e change) in body frame angle
@@ -788,33 +795,35 @@ void AC_AttitudeControl::attitude_controller_run_quat()
     thrust_heading_rotation_angles(_attitude_target, attitude_body, attitude_error, _thrust_angle, _thrust_error_angle);
 
     // Compute the angular velocity corrections in the body frame from the attitude error
-    _ang_vel_body = update_ang_vel_target_from_att_error(attitude_error);
+    Vector3f ang_vel_body = update_ang_vel_target_from_att_error(attitude_error);
 
     // ensure angular velocity does not go over configured limits
-    ang_vel_limit(_ang_vel_body, radians(_ang_vel_roll_max), radians(_ang_vel_pitch_max), radians(_ang_vel_yaw_max));
+    ang_vel_limit(ang_vel_body, radians(_ang_vel_roll_max), radians(_ang_vel_pitch_max), radians(_ang_vel_yaw_max));
 
     // rotation from the target frame to the body frame
     Quaternion rotation_target_to_body = attitude_body.inverse() * _attitude_target;
 
     // target angle velocity vector in the body frame
     Vector3f ang_vel_body_feedforward = rotation_target_to_body * _ang_vel_target;
-
+    Vector3f gyro = _rate_gyro;
     // Correct the thrust vector and smoothly add feedforward and yaw input
     _feedforward_scalar = 1.0f;
     if (_thrust_error_angle > AC_ATTITUDE_THRUST_ERROR_ANGLE * 2.0f) {
-        _ang_vel_body.z = _ahrs.get_gyro().z;
+        ang_vel_body.z = gyro.z;
     } else if (_thrust_error_angle > AC_ATTITUDE_THRUST_ERROR_ANGLE) {
         _feedforward_scalar = (1.0f - (_thrust_error_angle - AC_ATTITUDE_THRUST_ERROR_ANGLE) / AC_ATTITUDE_THRUST_ERROR_ANGLE);
-        _ang_vel_body.x += ang_vel_body_feedforward.x * _feedforward_scalar;
-        _ang_vel_body.y += ang_vel_body_feedforward.y * _feedforward_scalar;
-        _ang_vel_body.z += ang_vel_body_feedforward.z;
-        _ang_vel_body.z = _ahrs.get_gyro().z * (1.0 - _feedforward_scalar) + _ang_vel_body.z * _feedforward_scalar;
+        ang_vel_body.x += ang_vel_body_feedforward.x * _feedforward_scalar;
+        ang_vel_body.y += ang_vel_body_feedforward.y * _feedforward_scalar;
+        ang_vel_body.z += ang_vel_body_feedforward.z;
+        ang_vel_body.z = gyro.z * (1.0 - _feedforward_scalar) + ang_vel_body.z * _feedforward_scalar;
     } else {
-        _ang_vel_body += ang_vel_body_feedforward;
+        ang_vel_body += ang_vel_body_feedforward;
     }
 
     // Record error to handle EKF resets
     _attitude_ang_error = attitude_body.inverse() * _attitude_target;
+    // finally update the attitude target
+    _ang_vel_body = ang_vel_body;
 }
 
 // thrust_heading_rotation_angles - calculates two ordered rotations to move the attitude_body quaternion to the attitude_target quaternion.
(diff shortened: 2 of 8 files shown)


tpwrules:
Does this just assume that a Vector3f set is atomic?

andyp1per:
> Does this just assume that a Vector3f set is atomic?

It assumes that it doesn't matter that its not. The individual axes are and because all of the output is sequential, even though you might get some object tearing in practice it doesn't affect things. What does matter is setting things to 0 or other values in intermediate steps - that does make a big difference.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -188,15 +188,16 @@ float AC_AttitudeControl::get_slew_yaw_max_degs() const
 // Ensure attitude controller have zero errors to relax rate controller output
 void AC_AttitudeControl::relax_attitude_controllers()
 {
+    // take a copy of the last gyro used by the rate controller before using it
+    Vector3f gyro = _rate_gyro;
```
IamPete1, line 192: Doesn't this mean we relax to the last loops gyro not this loops? Plane at least calls run last.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -497,6 +500,12 @@ void AC_AttitudeControl::input_rate_bf_roll_pitch_yaw(float roll_rate_bf_cds, fl
 
 // Command an angular velocity with angular velocity smoothing using rate loops only with no attitude loop stabilization
 void AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_2(float roll_rate_bf_cds, float pitch_rate_bf_cds, float yaw_rate_bf_cds)
+{
+    _ang_vel_body = input_rate_bf_roll_pitch_yaw_4(roll_rate_bf_cds, pitch_rate_bf_cds, yaw_rate_bf_cds);
+}
+
+// Command an angular velocity with angular velocity smoothing using rate loops only with no attitude loop stabilization
+Vector3f AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_4(float roll_rate_bf_cds, float pitch_rate_bf_cds, float yaw_rate_bf_cds)
```
IamPete1, line 508: I don't understand what making this a wrapper is doing for us?

IamPete1 (commented):
Plane does not call these functions in the same order as copter in a loop. I suspect plane would now be using the gyro from one loop behind?

It also makes me slightly nervous that this is all a bit subtle so we could easily break it in the future.

I wonder if a little structure "rate controller inputs" or something like that would make it a bit clearer. You could also have a sempahore on changing those values, it may be a bit of performance cost but it does at least act as a signal that the developer needs to be careful.

andyp1per:
> I wonder if a little structure "rate controller inputs" or something like that would make it a bit clearer. You could also have a sempahore on changing those values, it may be a bit of performance cost but it does at least act as a signal that the developer needs to be careful.

Perhaps simply using a naming prefix would be sufficient here - I have always found it confusing that we have both attitude control and rate control mixed up in the same class

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
andyp1per: With the rate thread these things are going on concurrently so there is no first/last - so in fact the opposite is true - a much more up-to-date value is being used. In he non-rate case you might be right - I'll have to have a think about how this might be fixed. Obvious answer is to have two functions. Where is this called from?

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
andyp1per: Its also called from input_rate_step_bf_roll_pitch_yaw() which is now required for autotune

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
IamPete1: I meant all of these functions that use `_rate_gyro`. Copter runs the rate controller the first thing in the new loop, so on copter everything that comes later is using the correct value. Plane does all the inputs and then runs the rate controller.

We almost need a `set gyro` call, so both plane and copter do that first with the AHRS update and then whatever comes after will always be using a consistent value.

andyp1per: In copter this behaviour is unchanged - the AHRS is updated after the rate loops have run, so in this case it is a more accurate view. Note that _rate_gyro is the current gyro from the sensor here, not the delayed AHRS version - so again a more accurate version.

tpwrules:
> > Does this just assume that a Vector3f set is atomic?
> 
> It assumes that it doesn't matter that its not. The individual axes are and because all of the output is sequential, even though you might get some object tearing in practice it doesn't affect things. What does matter is setting things to 0 or other values in intermediate steps - that does make a big difference.

I'm concerned about this assumption on non-microcontrollers, the tear could be pretty big. In any case, wouldn't it break replay?

andyp1per:
> I'm concerned about this assumption on non-microcontrollers, the tear could be pretty big. In any case, wouldn't it break replay?

Tearing will only happen when the rate loop is in a separate thread. With replay all bets are off with a separate thread - I don't see how this can be made to work well regardless of the locking strategy.

libraries/AC_AttitudeControl/AC_AttitudeControl.h
```diff
@@ -177,6 +178,24 @@ class AC_AttitudeControl {
     // handle reset of attitude from EKF since the last iteration
     void inertial_frame_reset();
 
+    // Command euler yaw rate and pitch angle with roll angle specified in body frame
+    // (implemented only in AC_AttitudeControl_TS for tailsitter quadplanes)
+    virtual void input_euler_rate_yaw_euler_angle_pitch_bf_roll(bool plane_controls, float euler_roll_angle_cd, 
+        float euler_pitch_angle_cd, float euler_yaw_rate_cds) {}
+
+    ////// begin rate update functions //////
```
lthall, line 186: Thanks for this!

libraries/AP_AHRS/AP_AHRS_Logging.cpp
```diff
@@ -144,23 +144,24 @@ void AP_AHRS_View::Write_AttitudeView(const Vector3f &targets) const
 }
 
 // Write a rate packet
-void AP_AHRS_View::Write_Rate(const AP_Motors &motors, const AC_AttitudeControl &attitude_control,
+void AP_AHRS_View::Write_Rate(const AP_Motors &motors, AC_AttitudeControl &attitude_control,
```
lthall, line 147: Logging should be already fixed in a previous PR and moved to the Attitude Controller. WIP

lthall (approved):
This is looking good. A few PR's to get in first will simplify it further.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
IamPete1: > In copter this behaviour is unchanged

We need to check plane too.

libraries/AC_AutoTune/AC_AutoTune_Multi.cpp
```diff
@@ -1254,23 +1254,22 @@ void AC_AutoTune_Multi::twitch_test_run(AxisType test_axis, const float dir_sign
...
         }
     } else {
-        attitude_control->input_rate_bf_roll_pitch_yaw_2(0.0, 0.0, 0.0);
         // Testing rate P and D gains so will set body-frame rate targets.
         // Rate controller will use existing body-frame rates and convert to motor outputs
         // for all axes except the one we override here.
         switch (test_axis) {
         case AxisType::ROLL:
             // override body-frame roll rate
-            attitude_control->rate_bf_roll_target(dir_sign * target_rate + start_rate);
```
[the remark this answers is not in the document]
andyp1per: Removed

libraries/AC_AutoTune/AC_AutoTune_Multi.cpp
```diff
@@ -1254,23 +1254,22 @@ void AC_AutoTune_Multi::twitch_test_run(AxisType test_axis, const float dir_sign
...
+            attitude_control->input_rate_step_bf_roll_pitch_yaw(dir_sign * target_rate + start_rate, 0.0f, 0.0f);
             break;
         case AxisType::PITCH:
             // override body-frame pitch rate
-            attitude_control->rate_bf_pitch_target(dir_sign * target_rate + start_rate);
```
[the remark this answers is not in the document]
andyp1per: Removed

libraries/AC_AutoTune/AC_AutoTune_Multi.cpp
```diff
@@ -1254,23 +1254,22 @@ void AC_AutoTune_Multi::twitch_test_run(AxisType test_axis, const float dir_sign
...
+            attitude_control->input_rate_step_bf_roll_pitch_yaw(0.0f, dir_sign * target_rate + start_rate, 0.0f);
             break;
         case AxisType::YAW:
         case AxisType::YAW_D:
             // override body-frame yaw rate
-            attitude_control->rate_bf_yaw_target(dir_sign * target_rate + start_rate);
```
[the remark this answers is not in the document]
andyp1per: Left

libraries/AP_AHRS/AP_AHRS_Logging.cpp
andyp1per: Fixed

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -497,6 +500,12 @@ void AC_AttitudeControl::input_rate_bf_roll_pitch_yaw(float roll_rate_bf_cds, fl
 
 // Command an angular velocity with angular velocity smoothing using rate loops only with no attitude loop stabilization
 void AC_AttitudeControl::input_rate_bf_roll_pitch_yaw_2(float roll_rate_bf_cds, float pitch_rate_bf_cds, float yaw_rate_bf_cds)
+{
+    _ang_vel_body = input_rate_bf_roll_pitch_yaw_4(roll_rate_bf_cds, pitch_rate_bf_cds, yaw_rate_bf_cds);
+}
+
+// Command an angular velocity with angular velocity smoothing using rate loops only with no attitude loop stabilization
```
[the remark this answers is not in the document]
lthall: Yes, this is a little off putting.

input_rate_bf_roll_pitch_yaw

This is a fully stabilized acro. This uses the same attitude control methodology as pretty much everything else.

input_rate_bf_roll_pitch_yaw_2

This is like the BetaFlight acro with no attitude feedback. This is rate loops only.

input_rate_bf_roll_pitch_yaw_3

This is an acro with attitude feedback that does not rely on attitude. This is done by integrating rate error. I would remove this but it is still used by Plane. It is an elegant solution that Robert came up with so I will be sad to see it go.

input_rate_bf_roll_pitch_yaw_4

I am thinking we can probably remove this and do the step function directly. I will work with Andy to see if I can do this elegantly.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
andyp1per: I have added a change which means plane now gets the latest AHRS value for all of these, so NFC. Longer term plane should be doing it the way copter does to minimize latency

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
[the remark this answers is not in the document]
andyp1per: Added and removed

andyp1per:
Thanks for your comments @IamPete1 ! They provoked a good debate / change

IamPete1 (commented):
I think this all looks good for copter and plane. Just needs some some fixups for sub and heli. 

Sub needs the new `rate_controller_target_reset` call after its `rate_controller_run`.

The sub classes `AC_AttitudeControl_Heli`, `AC_AttitudeControl_Multi_6DoF` and `AC_AttitudeControl_Sub` all override `rate_controller_run` I think they need to be changed to override the new `rate_controller_run_dt`.

The heli method `passthrough_bf_roll_pitch_rate_yaw` is quite tangled, it has not been made safe, I think that is OK, we just need to add a big comment so we can do it at some future point. Likewise for the functions in `AC_AttitudeControl_TS`

andyp1per:
> I think this all looks good for copter and plane. Just needs some some fixups for sub and heli.
> 
> Sub needs the new `rate_controller_target_reset` call after its `rate_controller_run`.
> 
> The sub classes `AC_AttitudeControl_Heli`, `AC_AttitudeControl_Multi_6DoF` and `AC_AttitudeControl_Sub` all override `rate_controller_run` I think they need to be changed to override the new `rate_controller_run_dt`.
> 
> The heli method `passthrough_bf_roll_pitch_rate_yaw` is quite tangled, it has not been made safe, I think that is OK, we just need to add a big comment so we can do it at some future point. Likewise for the functions in `AC_AttitudeControl_TS`

I believe that all the subclasses that override rate_controller_run() are correct. None of them support sysid or pd_scale from the base class (which is what rate_controller_target_reset does) and so can be left as-is for now.

I agree that Heli could be refactored to use that function, but was planning to leave it alone for now since @bnsgeyer doesn't want fast rate support in Heli for now. 

rate_controller_run_dt is currently the equivalent of what was rate_controller_run() which was only used by copter and plane and only exists in AC_AttitudeControl_Multi so cannot be overridden.

So I agree but that some further refactoring could be done but propose that we leave it as-is in this PR to avoid changing the behaviour of these other vehicles.

andyp1per:
@IamPete1 I have added a config error to the default implementation of rate_controller_run_dt() to avoid it getting called accidentally

ArduCopter/Attitude.cpp
```diff
@@ -15,7 +15,9 @@ void Copter::run_rate_controller()
     pos_control->set_dt(last_loop_time_s);
 
     // run low level rate controllers that only require IMU data
-    attitude_control->rate_controller_run(); 
+    attitude_control->rate_controller_run();
+     // reset sysid and other temporary inputs
```
rmackay9, line 19: nitpick: tiny little white space issue here, just an extra space

tridge:
needs testing on quadplane, either real aircraft or realflight

libraries/AC_AttitudeControl/AC_AttitudeControl_Multi.cpp
```diff
@@ -439,35 +439,49 @@ void AC_AttitudeControl_Multi::update_throttle_rpy_mix()
     _throttle_rpy_mix = constrain_float(_throttle_rpy_mix, 0.1f, AC_ATTITUDE_CONTROL_MAX);
 }
 
-void AC_AttitudeControl_Multi::rate_controller_run()
+void AC_AttitudeControl_Multi::rate_controller_run_dt(float dt, const Vector3f& gyro)
```
rmackay9, line 442: it doesn't really matter but normally we put "dt" as the last argument

rmackay9:
This looks OK to me

I agree that we should be sure to test Heli and Sub to be sure they're unaffected and of course, we should test with the mindset that there is a problem and we just haven't found it yet rather than just to prove that it works as expected.

tridge:
I've tested quadplane in RealFlight and it does look good to me

ArduCopter/Attitude.cpp
andyp1per: Thanks fixed

libraries/AC_AttitudeControl/AC_AttitudeControl_Multi.cpp
andyp1per: Thanks, fixed

andyp1per:
We have done manual testing of the gain backoff on landing and takeoff successfully and I have also written an autotest to verify which passes - https://github.com/ArduPilot/ardupilot/pull/28220

Williangalvani:
couldn't find any issues running this on Sub

tridge (approved):

peterbarker:
Just waiting on @bnsgeyer to be happy with this and then we can merge it!

MattKear:
I will have some time tomorrow to test this on heli in Real Flight, if @bnsgeyer doesn't have chance before then.

andyp1per:
Thanks @MattKear !

bnsgeyer:
@andyp1per @rmackay9  I tested heli in real flight.  It looks fine.  No issues.

andyp1per:
Thanks all!

PR #27841, opened 2024-08-15
Dynamic FIFO Buffering

andyp1per:

This normalizes the usage of the "primary" gyro in the IMU drivers and gives an opportunity to take action when it changes. We were doing this before with random calls into the AHRS, but really the IMU needs to know this information. This makes it possible to take specific required action in future PRs.

I've turned this into a demo of what this can be used for, most notably to slow down FIFO read rates on non-primary gyros.
On a CubeOrange+ with ```INS_GYRO_RATE=2``` (4KHz):
Before:
```
idle          PRI=  1 sp=0x30021CF8 STACK= 296/ 504 LOAD=22.0%
SPI4          PRI=181 sp=0x3002AF58 STACK= 896/1464 LOAD=23.2%
SPI1          PRI=181 sp=0x3002B668 STACK= 824/1464 LOAD=25.2%
```
After:
```
idle          PRI=  1 sp=0x30021D00 STACK= 296/ 504 LOAD=34.7%
SPI4          PRI=181 sp=0x3002AF58 STACK= 816/1464 LOAD=22.0%
SPI1          PRI=181 sp=0x3002B668 STACK= 816/1464 LOAD=14.1%
```

The saving is not as great as you might expect, probably down to SPI overhead which can't be avoided but still substantial.

On my fast rate copter with 2x ICM42688 running at 16Mhz or so I get:
Before:
```
idle          PRI=  1 sp=0x3001F668 STACK= 296/ 504 LOAD= 4.7%
SPI1          PRI=181 sp=0x20003538 STACK= 848/1464 LOAD=17.7%
SPI4          PRI=181 sp=0x20003B50 STACK= 848/1464 LOAD=16.3%
```
After:
```
idle          PRI=  1 sp=0x3001F670 STACK= 296/ 504 LOAD=12.8%
SPI1          PRI=181 sp=0x30021838 STACK= 816/1464 LOAD=19.1%
SPI4          PRI=181 sp=0x20002720 STACK= 824/1464 LOAD= 8.3%*
```
Which is the difference between on the edge and comfortable

With @bugobliterator 's SPI PR https://github.com/ArduPilot/ardupilot/pull/28093 I get further benefit:
```
idle          PRI=  1 sp=0x3001F670 STACK= 296/ 504 LOAD=15.4%
SPI1          PRI=181 sp=0x30022188 STACK= 816/1464 LOAD=15.9%
SPI4          PRI=181 sp=0x30021B70 STACK= 824/1464 LOAD= 7.6%*
```


libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
```diff
@@ -799,6 +796,8 @@ void AP_InertialSensor_Backend::update_gyro(uint8_t instance) /* front end */
         }
     }
 #endif
+
+    set_primary_gyro(_imu._primary_gyro);
```
IamPete1, line 800: This function dons't do anything? If its a backed should it pass a bool "is_primary" rather than the instance?

libraries/AP_InertialSensor/AP_InertialSensor.cpp
```diff
@@ -1950,12 +1950,24 @@ void AP_InertialSensor::update(void)
         for (uint8_t i=0; i<INS_MAX_INSTANCES; i++) {
             if (_gyro_healthy[i] && _use(i)) {
                 _first_usable_gyro = i;
+#if AP_AHRS_ENABLED
+                // ask AHRS for the true primary, might just be us though
+                _primary_gyro = AP::ahrs().get_primary_gyro_index();
+#else
+                _primary_gyro = _first_usable_gyro;
+#endif
```
IamPete1, line 1958: Shouldn't this be outside the for loop?

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
andyp1per: Right, this is to give the backends an opportunity to do something - that something is the more controversial bit, so just putting the framework in place for now.

bool is_primary would also work if you think that's better

libraries/AP_InertialSensor/AP_InertialSensor.cpp
andyp1per: its inside because I need _first_usable_gyro to be definitely set. Could split this up though with a #define if you think that's better?

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
andyp1per: Fixed

libraries/AP_InertialSensor/AP_InertialSensor.cpp
andyp1per: Fixed

andyp1per:
Closing because I can live without changing the SPI delay and needs restructuring to fix the layering

peterbarker:
"Not a hard no"... need to be careful

precursor for poking the primary sensor into the library, rather than fetching from AP_AHRS would be good

libraries/AP_AHRS/AP_AHRS.cpp
```diff
@@ -340,9 +340,21 @@ void AP_AHRS::reset_gyro_drift(void)
  */
 void AP_AHRS::update_state(void)
 {
+    const uint8_t primary_gyro = _get_primary_gyro_index();
+    const uint8_t primary_accel = _get_primary_accel_index();
+
+    // tell the IMUS about primary changes
```
peterbarker, line 346:
```
#if AP_INERTIALSENSOR_ENABLED
```

andyp1per: Fixed

andyp1per:
Flown on my 5". This really works very well indeed! Giving the fast rate loop extra CPU cycles results in even smoother flying - it was amazing!

andyp1per:
@tridge I have changed this so that the IMU changes are only active on copter with fast rate enabled. Hopefully this goes some way to alleviate your concerns!

tridge (commented):
i'm nervous about the changes on invensense v1 and v2, can we restrict the change to v3?

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
```diff
@@ -170,7 +171,7 @@ struct PACKED FIFODataHighRes {
 static_assert(sizeof(FIFOData) == 16, "FIFOData must be 16 bytes");
 static_assert(sizeof(FIFODataHighRes) == 20, "FIFODataHighRes must be 20 bytes");
 
-#define INV3_FIFO_BUFFER_LEN 8
+#define INV3_FIFO_BUFFER_LEN 16
```
tridge, line 174: is this needed?

libraries/AP_InertialSensor/AP_InertialSensor_Backend.h
```diff
@@ -292,6 +296,12 @@ class AP_InertialSensor_Backend
     void update_accel(uint8_t instance) __RAMFUNC__; /* front end */
     void update_accel_filters(uint8_t instance) __RAMFUNC__; /* front end */
 
+    // catch updates to the primary gyro and accel
+    void update_primary_gyro() __RAMFUNC__; /* backend */
+    void update_primary_accel() __RAMFUNC__; /* backend */
+    virtual void set_primary_gyro(bool is_primary) {}
+    virtual void set_primary_accel(bool is_primary) {}
```
tridge, line 303: can remove all the accel based switching

libraries/AP_InertialSensor/FastRateBuffer.cpp
```diff
@@ -79,6 +79,17 @@ bool AP_InertialSensor::is_rate_loop_gyro_enabled(uint8_t instance) const
     return fast_rate_buffer->use_rate_loop_gyro_samples() && instance == AP::ahrs().get_primary_gyro_index();
 }
 
+// whether or not to use the dynamic fifo
+bool AP_InertialSensor::is_dynamic_fifo_enabled(uint8_t instance) const
+{
+    if (!fast_rate_buffer_enabled || fast_rate_buffer == nullptr) {
+        return false;
+    }
+    return ((1U<<instance) & AP_INERTIALSENSOR_DYNAMIC_FIFO_MASK)
```
tridge, line 88: not sure this mask is useful? I think it can be removed

tridge (changes requested):

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
```diff
@@ -798,6 +801,20 @@ void AP_InertialSensor_Backend::update_gyro(uint8_t instance) /* front end */
     update_gyro_filters(instance);
 }
 
+void AP_InertialSensor_Backend::update_primary_gyro()
+{
+    // timing changes need to be made in the bus thread in order to take effect which is
+    // why they are actioned here
+    const bool is_new_primary = gyro_instance == _imu._primary_gyro;
+    uint32_t now_us = AP_HAL::micros();
+    if (is_primary_gyro != is_new_primary
```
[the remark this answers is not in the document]
andyp1per: Don't think so, has to be checked every time. You could save the micros call if the gyro is in fact wrong, but not otherwise

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
andyp1per: Yes I believe it is

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
```diff
@@ -841,6 +858,19 @@ void AP_InertialSensor_Backend::update_accel(uint8_t instance) /* front end */
     update_accel_filters(instance);
 }
 
+void AP_InertialSensor_Backend::update_primary_accel()
+{
```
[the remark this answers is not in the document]
andyp1per: removed

libraries/AP_InertialSensor/AP_InertialSensor_Backend.h
andyp1per: Need to keep the set* for drivers that have different FIFOs and callbacks for accel vs gyro

libraries/AP_InertialSensor/FastRateBuffer.cpp
andyp1per: removed

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
```diff
@@ -841,6 +858,19 @@ void AP_InertialSensor_Backend::update_accel(uint8_t instance) /* front end */
...
+    // timing changes need to be made in the bus thread in order to take effect which is
+    // why they are actioned here
+    const bool is_new_primary = accel_instance == _imu._primary_accel;
+    uint32_t now_us = AP_HAL::micros();
```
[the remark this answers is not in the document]
andyp1per: Nop, see above.

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev2.cpp
```diff
@@ -23,6 +23,7 @@
 
 #include <AP_HAL/AP_HAL.h>
 
+#include "AP_InertialSensor_rate_config.h"
```
peterbarker, line 26: No longer needed?

libraries/AP_InertialSensor/AP_InertialSensor_Invensense.cpp
```diff
@@ -26,6 +26,7 @@
 #include <AP_InternalError/AP_InternalError.h>
 #include <AP_Logger/AP_Logger.h>
 
+#include "AP_InertialSensor_rate_config.h"
```
peterbarker, line 29: No longer required?

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
peterbarker: .... needs explaining because this is definitely the sort of thing we end up needing a point release in stable for when we discover some random v3 sensors doesn't like the increased length...

peterbarker (changes requested):

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev2.cpp
andyp1per: Removed

libraries/AP_InertialSensor/AP_InertialSensor_Invensense.cpp
andyp1per: Removed

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
andyp1per: Added a comment

andyp1per:
Test flown, EKF innovations look the same. CPU on second IMU way down.

andyp1per:
Full fight test on two vehicles. EKF happy, CPU load great. Good to go.

libraries/AP_InertialSensor/AP_InertialSensor.cpp
```diff
@@ -1953,12 +1963,18 @@ void AP_InertialSensor::update(void)
         for (uint8_t i=0; i<INS_MAX_INSTANCES; i++) {
             if (_gyro_healthy[i] && _use(i)) {
                 _first_usable_gyro = i;
+#if !AP_AHRS_ENABLED
+                _primary_gyro = _first_usable_gyro;
+#endif
```
peterbarker, line 1968: It might be a little clearer to drop this and the accel equivalent outside these two loops, but right below them.

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
```diff
@@ -807,6 +808,21 @@ void AP_InertialSensor_Backend::update_gyro(uint8_t instance) /* front end */
...
 
+void AP_InertialSensor_Backend::update_primary()
+{
+    // timing changes need to be made in the bus thread in order to take effect which is
+    // why they are actioned here. Currently the _primary_gyro and _primary_accel can never
+    // be different for a particular IMU
+    const bool is_new_primary = gyro_instance == _imu._primary_gyro;
+    uint32_t now_us = AP_HAL::micros();
+    if (is_primary != is_new_primary
+        || AP_HAL::timeout_expired(last_primary_update_us, now_us, PRIMARY_UPDATE_TIMEOUT_US)) {
```
peterbarker, line 819: Why do we need a timeout here?  A call to `update_primary` during init should work?

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
```diff
@@ -807,6 +808,21 @@ void AP_InertialSensor_Backend::update_gyro(uint8_t instance) /* front end */
     update_gyro_filters(instance);
 }
 
+void AP_InertialSensor_Backend::update_primary()
+{
+    // timing changes need to be made in the bus thread in order to take effect which is
+    // why they are actioned here. Currently the _primary_gyro and _primary_accel can never
+    // be different for a particular IMU
+    const bool is_new_primary = gyro_instance == _imu._primary_gyro;
```
peterbarker, line 816:
```suggestion
    const bool is_new_primary = (gyro_instance == _imu._primary_gyro);
```

peterbarker (changes requested):
Looks pretty good to me.

libraries/AP_InertialSensor/AP_InertialSensor.cpp
andyp1per: I don't think that's logically correct, because we only want to make this change if there is a new _first_usable_gyro. I think its clearer leaving it where it is

libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
andyp1per: Fixed, thanks

libraries/AP_AHRS/AP_AHRS.cpp
```diff
@@ -340,9 +340,21 @@ void AP_AHRS::reset_gyro_drift(void)
...
 {
+    const uint8_t primary_gyro = _get_primary_gyro_index();
+    const uint8_t primary_accel = _get_primary_accel_index();
+#if AP_INERTIALSENSOR_ENABLED
+    // tell the IMUS about primary changes
+    if (primary_gyro != state.primary_gyro) {
+        AP::ins().set_primary_gyro(primary_gyro);
+    }
+
+    if (primary_accel != state.primary_accel) {
```
tridge, line 351: i think we don't need to do anything with accel in this PR, we tie the accel and gyro index in AP_AHRS::_get_primary_IMU_index() now anyway

tridge (changes requested):
i think with the accel changes removed we should merge this, I don't love it, but we need to get to the finish line :-)

libraries/AP_AHRS/AP_AHRS.cpp
andyp1per: Removed

tridge (approved):

PR #27843, opened 2024-08-15
Servo output optimization

andyp1per:

This is an optimization that avoids a repeated N^2 lookup in motor output that proves to be quite expensive at high rates (i.e. involving dshot).

### DIFF
--- a/libraries/AP_Motors/AP_MotorsMulticopter.cpp
+++ b/libraries/AP_Motors/AP_MotorsMulticopter.cpp
@@ -248,6 +248,9 @@ void AP_MotorsMulticopter::output()
     // update throttle filter
     update_throttle_filter();
 
+    // update servo map
+    update_servo_map();
+
     // calc filtered battery voltage and lift_max
     thr_lin.update_lift_max_from_batt_voltage();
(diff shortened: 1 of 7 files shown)


IamPete1 (commented):
I think we could re-work `find_channel` to pull the index of the first set bit in the existing `channel_mask`, the result should be the same except it would work for all outputs.

andyp1per:
> I think we could re-work `find_channel` to pull the index of the first set bit in the existing `channel_mask`, the result should be the same except it would work for all outputs.

That's a neat idea - I will try it!

andyp1per:
> I think we could re-work `find_channel` to pull the index of the first set bit in the existing `channel_mask`, the result should be the same except it would work for all outputs.

Actually I don't think this works as there is no map from function to channel - you first have to find a channel with the right function before you can find the output. Avoid the long iteration though the function map is the main thing being avoided here. Maybe I have misunderstood your idea - if so could you please elaborate?

andyp1per:
Closing is favour of https://github.com/ArduPilot/ardupilot/pull/27849

PR #27844, opened 2024-08-15
Allow configuration of device periodic delay

andyp1per:

Currently the value is hardcoded and fairly arbitrary. This allows devices to decide.

Here is the motor output using a 0 minimum on the MPU6000. You can see that the output rate is the desired 4Khz (this is the gyro sample rate) and is pretty regular:

image

Here is with the default 100us. You can see that the output is not able to keep up with the inbound rate because a fixed 100us delay is being added after the SPI thread callback has run and we only get 2Khz output and worse we will be dropping samples.

image

Interestingly the ICM42688 *is* able to cope:

image

I think this is to do with the SPI bus speed. On the Matek H743 we have:

```
SPIDEV icm42688  SPI1 DEVID1 IMU1_CS      MODE3  2*MHZ 16*MHZ 
SPIDEV mpu6000   SPI1 DEVID1 IMU1_CS      MODE3  1*MHZ  4*MHZ
SPIDEV icm20602  SPI4 DEVID1 IMU2_CS      MODE3  1*MHZ  4*MHZ
SPIDEV icm42605  SPI4 DEVID1 IMU3_CS      MODE3  2*MHZ 16*MHZ
```

So MPU6000 running really pretty slowly. If I put some simple timing capture in the INS thread the MPU6000 takes about 240us to process a FIFO read and the ICM24688 takes about 120us - so quite a substantial difference and of course easily explains why one can keep up and the other cannot. I think perhaps on that basis the answer is not to modify the periodic delay but simply not allow the fast rate code to be run at higher rates on these slow IMUs. I guess we might be able to put something in the chibios_hwdef.py script that figures out whether the INS read is fast enough,

### DIFF
--- a/libraries/AP_HAL/Device.h
+++ b/libraries/AP_HAL/Device.h
@@ -260,6 +260,12 @@ public:
     virtual PeriodicHandle register_periodic_callback(uint32_t period_usec, PeriodicCb) = 0;
 
     /*
+     * Adjust the minimum time allowed between cycles of the periodic callback registered with
+     * #register_periodic_callback.
+     */
+    virtual void set_periodic_minimum(uint32_t min_usec) = 0;
+
+    /*
      * Adjust the time for the periodic callback registered with
      * #register_periodic_callback. Note that the time will be re-calculated
      * from the moment this call is made and expire after @period_usec.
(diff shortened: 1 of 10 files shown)


andyp1per:
Closing, because I think here is a better way - although I still think that 100us is a bit arbitrary

PR #27849, opened 2024-08-15
SRV_Channels: find_channel: use channel mask remove need for search over all channels

IamPete1:
This should speed up the find channel call so its less painful at higher rates. https://github.com/ArduPilot/ardupilot/pull/27843

~~I have not tested this, or had a in depth look to see if its a good idea.~~ 

I think this is OK, there will be slightly more lag when a function is changed, we should not be changing functions at runtime so I think its OK.


andyp1per (approved):
Nice work Pete! I flew this successfully

IamPete1:
Fixed a typo in a comment.

tridge (approved):

PR #27851, opened 2024-08-15
AHRS: added common origin between backends

tridge:
This ensures that EKF2, EKF3 and ExternalAHRS all have a common origin

Note that the origin will come from the first running backend that gets an origin, even if it is not the active backend. This is needed because EKF2 and EKF3 refuse to change origin once it is set.
The placement of this code in the update function for each backend is done as two backends may get an origin on the same loop, so we must check for an origin as soon as we call the backends update

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -11889,6 +11889,58 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
         self.set_rc(10, 2000)
         self.wait_statustext('Using EKF Source Set 3', check_context=True)
 
+    def CommonOrigin(self):
+        """Test common origin between EKF2 and EKF3"""
+        self.context_push()
+
+        # start on EKF2
+        self.set_parameters({
+            'AHRS_EKF_TYPE': 2,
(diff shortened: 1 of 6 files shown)


peterbarker (commented):
I like it.  The copied block will go away when I manage to get the common base class stuff sorted....

Need to test it on real vehicles a bit.

peterbarker:
.. also should have an autotest

tridge:
this was flight tested on a 5" quad, switching between EK2 and EK3 did not change position

tridge:
> .. also should have an autotest

not really sure how to do that ... the obvious method of not setting EK2_ENABLE=1 until after EK3 has set origin won't work, as this PR only does the origin set once

peterbarker (approved):

rmackay9 (approved):
an autotest would be nice but if it's too difficult then merging is OK

tridge:
i'll try to do an autotest

tridge:
back for dev call this week. I think we should merge this now, but separately see if we can stop the EKFs from changing the origin height on baro reset, but as a separate PR

rmackay9 (approved):

PR #27853, opened 2024-08-15
QURT: allow for external compasses

tridge:
this allows for additional compasses on I2C


tridge:
flight tested on voxl-2

PR #15669, opened 2020-10-28
GCS_MAVlink: Allow clearing during an active mission upload

rywunder:
If an upload link is active Ardupilot already allows a different upload from the same GCS to take its place.  This code allows a clear command from the same GCS to overwrite and close an active upload link. Without this change a GCS must wait for a lost link upload to timeout before being able to clear a mission.

Tested this using mavproxy and and a simple pymavlink script with a timing out upload.

PR #27854, opened 2024-08-15
GCS_MAVLink: allow uploading GCS to cancel upload using mission_clear_all

peterbarker:
~these would appear to be identical operations.~  clear differs from a 0-item-upload in the "clear" is only allowed disarmed, and resets a whole bunch of mission state (e.g. the loaded item).

Replaces https://github.com/ArduPilot/ardupilot/pull/15669 .  Like the PR it replaces, this allows a GCS to cancel an upload with a clear-all command as if they'd issues a mission-upload with zero items.

### DIFF
--- a/libraries/GCS_MAVLink/MissionItemProtocol.cpp
+++ b/libraries/GCS_MAVLink/MissionItemProtocol.cpp
@@ -33,7 +33,7 @@ void MissionItemProtocol::handle_mission_clear_all(const GCS_MAVLINK &_link,
                                                    const mavlink_message_t &msg)
 {
     bool success = true;
-    success = success && !receiving;
+    success = success && cancel_upload(_link, msg);
     success = success && clear_all_items();
     send_mission_ack(_link, msg, success ? MAV_MISSION_ACCEPTED : MAV_MISSION_ERROR);
 }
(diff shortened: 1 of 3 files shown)


libraries/GCS_MAVLink/MissionItemProtocol.cpp
```diff
@@ -52,23 +52,16 @@ bool MissionItemProtocol::mavlink2_requirement_met(const GCS_MAVLINK &_link, con
     return false;
 }
 
-void MissionItemProtocol::handle_mission_count(
-    GCS_MAVLINK &_link,
-    const mavlink_mission_count_t &packet,
-    const mavlink_message_t &msg)
+bool MissionItemProtocol::check_receiving_cancel(const GCS_MAVLINK &_link, const mavlink_message_t &msg)
```
tridge, line 55: cancel_upload ?

tridge (changes requested):
just the name as cancel_upload, thanks

libraries/GCS_MAVLink/MissionItemProtocol.cpp
peterbarker: Done

PR #27857, opened 2024-08-16
Ma copter sur 4.3.0.3 esc logging

AbhiMalloy:
This is to log the RCOUT at 25hz instead of 10 Hz


IamPete1:
I don't think you meant to open a PR against this branch.

PR #27858, opened 2024-08-16
autotest: fix FenceFloorEnabledLanding test

andyp1per:
This test had erroneous behaviour, now fixed.

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -1789,7 +1789,6 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
         self.user_takeoff(alt_min=25)
 
         # Check fence is enabled
-        self.do_fence_enable()
         self.assert_fence_enabled()
 
         # Change to RC controlled mode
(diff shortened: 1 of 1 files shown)


peterbarker:
Could you explain what this test is supposed to be testing?

The test description is the same as another test in there, so I can't really trust it - the description does look correct for the *other* test.

The test doesn't seem to make sense as-written; it enters recovery mode, then waits for it to timeout, which should re-trigger the fence action - RTL - but then it fiddles with RC as if that will make a difference.

If I understand what's happening correctly this is actually a pretty serious bug in the fence logic.

When passing, the vehicle enters RTL and the disarm-on-land code *within the mode* triggers.  When failing the vehicle stays in LOITER for what seems an unlimited period below the fence.  The vehicle will *not* disarm in loiter when the landing detector fires as the relevant option bit it not set.

So... what *is* supposed to happen in this test?  An `assert_mode` or two in here would not go astray!

peterbarker:
image

peterbarker (approved):
LGTM, thanks!

PR #27859, opened 2024-08-16
SITL : added a basic simulated SBF GPS receiver

chiara-septentrio:
A basic simulator that returns PVT with the PVTGeodetic v1 SBF block


libraries/SITL/SIM_GPS.h
```diff
@@ -120,6 +120,9 @@ class GPS : public SerialDevice {
 #if AP_SIM_GPS_SBP2_ENABLED
         SBP2  =  9,
 #endif
+#if AP_SIM_GPS_SBF_ENABLED
+        SBF = 10, //matches GPS_TYPE
```
Ryanf55, line 124:
```suggestion
        SBF = 10, //matches GPS1_TYPE 
```

Ryanf55 (approved):
LGTM, nice and simple. Always great to see more simulators added!

Ryanf55:
Did you want this added to the autotest suite?

libraries/SITL/SIM_GPS.h
peterbarker: Fixed

peterbarker:
@chiara-septentrio I have made this simulator work.

Next time you open a PR, please state what testing it has had.  @Ryanf55 assumed a minimal level of testing, probably thinking you had at least verified this was working in SITL.  Given the several problems I encountered, surely the only level of testing this could have had was "it compiles"?

I have added it to the autotest suite, so now we have some level of protection against regressions in the Septentrio driver.

I have also made a patch to the Septentrio driver to not raise a numeric exception when the undulation comes in as DNU; the simulator was supplying DNU there.  Please verify that that patch is correct.

chiara-septentrio:
It's my bad, I misunderstood how to test for it correctly. I launched a simulation without the console or map and since I didn't see any error, I thought it was fine. I had never used SITL before and I obviously didn't correctly understand to documentation, I'm sorry.
I tested the changes with the console and map this time, using mode guided and circle, and it seemed to work fine. 
I just changed the function to check the configuration is done to take simulation into account otherwise the function was waiting for the GPS to respond to commands infinitely. From what I understand it should still be able to use an actual GPS with the simulation.

peterbarker:
OK, LGTM.  Thanks for re-testing it!

I assume you're OK with the change in the actual driver, and I'm marking it MergeOnCIPass.

Thanks for coming back and creating this for us!

peterbarker:
Merged, thanks @chiara-septentrio !

PR #27863, opened 2024-08-17
AP_DroneCAN: DNA_Server: polish db comments and organize

tpwrules:
Followup to https://github.com/ArduPilot/ardupilot/pull/27750 .

No compiler output change until the last two commits, last two have no size change (at least on CubeOrange).


tpwrules:
Tested in SITL for good measure.

tridge (approved):

PR #27873, opened 2024-08-20
AP_Scripting: message_interval example handles 0hz

rmackay9:
This is a small enhancement to the message_interval lua script that is useful for forcibly setting the rate at which messages are sent to the ground station.

This comes out of the [slung payload project](https://github.com/ArduPilot/ardupilot/pull/27783) where I want to force the payload to send its position to the vehicle at 10hz.

The changes are:

- add more definitions for MAV_SEVERITY, commonly used message ids and serial port used
- add ability to disable a message completely
- added a welcome message

This has been lightly tested in SITL and on a real hardware (a Here4FC on the payload)

message-interval-testing

### DIFF
--- a/libraries/AP_Scripting/examples/message_interval.lua
+++ b/libraries/AP_Scripting/examples/message_interval.lua
@@ -5,17 +5,50 @@
 -- during the initial connection, so by resetting them periodically we ensure we get
 -- the expected telemetry rate
 
--- intervals is a table which contains a table for each entry we want to adjust
--- each entry is arranged as {the serial link to adjust, the mavlink message ID, and the broadcast interval in Hz}
-local intervals = {{0, uint32_t(30), 2.0},  -- First serial, ATTITUDE, 2Hz
-                   {0, uint32_t(163), 5.0}} -- First serial, AHRS, 5Hz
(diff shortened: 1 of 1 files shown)


libraries/AP_Scripting/examples/message_interval.lua
```diff
@@ -5,17 +5,50 @@
...
+
+-- serial port
+local serial_port = 0
+
 -- intervals is a table which contains a table for each entry we want to adjust
 -- each entry is arranged as {the serial link to adjust, the mavlink message ID, and the broadcast interval in Hz}
-local intervals = {{0, uint32_t(30), 2.0},  -- First serial, ATTITUDE, 2Hz
-                   {0, uint32_t(163), 5.0}} -- First serial, AHRS, 5Hz
+local intervals = {{serial_port, ATTITUDE_ID, 2.0}, -- ATTITUDE, 2Hz
+                   {serial_port, AHRS_ID, 1.0},     -- AHRS, 5Hz
```
peterbarker, line 37: Comment is out of date

peterbarker (approved):
LGTM - might want to fix the comment?

libraries/AP_Scripting/examples/message_interval.lua
rmackay9: Thanks for noticing that.  I've updated the rate to match the comment, they're just examples in any case, for the payload the table is filled in differently (everything is off except global-position-int)

PR #27875, opened 2024-08-20
Plane: Updated QTUN and NTUN units to modern multiplier

Georacer:
Did not change the format character, as pymavlink seems to need it, due to a bug.

Before:
Screenshot from 2024-08-20 16-37-12
Screenshot from 2024-08-20 16-39-23

After:
Screenshot from 2024-08-20 16-37-37
Screenshot from 2024-08-20 16-44-27

### DIFF
--- a/ArduPlane/Log.cpp
+++ b/ArduPlane/Log.cpp
@@ -341,7 +341,7 @@ const struct LogStructure Plane::log_structure[] = {
 // @Field: TAT: target altitude TECS
 // @Field: TAsp: target airspeed
     { LOG_NTUN_MSG, sizeof(log_Nav_Tuning),         
-      "NTUN", "QfcccfffLLeee",  "TimeUS,Dist,TBrg,NavBrg,AltE,XT,XTi,AsE,TLat,TLng,TAW,TAT,TAsp", "smddmmmnDUmmn", "F0BBB0B0GG000" , true },
+      "NTUN", "QfcccfffLLeee",  "TimeUS,Dist,TBrg,NavBrg,AltE,XT,XTi,AsE,TLat,TLng,TAW,TAT,TAsp", "smddmmmnDUmmn", "F0BBB0B0GGBB0" , true },
 
 // @LoggerMessage: ATRP
 // @Description: Plane AutoTune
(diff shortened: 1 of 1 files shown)


ArduPlane/Log.cpp
```diff
@@ -341,7 +341,7 @@ const struct LogStructure Plane::log_structure[] = {
 // @Field: TAT: target altitude TECS
 // @Field: TAsp: target airspeed
     { LOG_NTUN_MSG, sizeof(log_Nav_Tuning),         
-      "NTUN", "QfcccfffLLeee",  "TimeUS,Dist,TBrg,NavBrg,AltE,XT,XTi,AsE,TLat,TLng,TAW,TAT,TAsp", "smddmmmnDUmmn", "F0BBB0B0GG000" , true },
+      "NTUN", "QfcccfffLLeee",  "TimeUS,Dist,TBrg,NavBrg,AltE,XT,XTi,AsE,TLat,TLng,TAW,TAT,TAsp", "smddmmmnDUmmn", "F0BBB0B0GGBB0" , true },
```
tridge, line 344: the 'e' already has the * 100 built in, so if we added the B it would apply twice, and get the wrong answer

tridge (changes requested):

ArduPlane/Log.cpp
tridge: we should probably change the '0' to a '-' meaning no multiplier (thanks @peterbarker )

Georacer: I don't think it applies twice. At least not in my testing. I vaguely remember that the MAVProxy logic is that if a format with multiplier is found, then the separate multiplier is not taken into account.

I would prefer changing the format `e` into `i`, but seeing this triggered some unwanted behaviour last time, I thought I'd double-dip and do scaling via both the format and the individual multiplier. Mostly because doing it right doesn't seem to work correctly: https://github.com/ArduPilot/pymavlink/pull/965.

It's not a big deal, let's discuss live.

tridge:
need to get this fixed first:
https://github.com/ArduPilot/pymavlink/pull/965

PR #27877, opened 2024-08-20
Add osd switching via stick inputs in disarm state

prosolvo4th:
Sticks inputs: roll mid && pitch high && yaw low && throttle mid

### DIFF
--- a/libraries/AP_OSD/AP_OSD.cpp
+++ b/libraries/AP_OSD/AP_OSD.cpp
@@ -30,6 +30,7 @@
 #include <AP_HAL/AP_HAL.h>
 #include <AP_HAL/Util.h>
 #include <RC_Channel/RC_Channel.h>
+#include <AP_RCMapper/AP_RCMapper.h>
 #include <AP_AHRS/AP_AHRS.h>
 #include <AP_BattMonitor/AP_BattMonitor.h>
 #include <utility>
@@ -597,6 +598,32 @@ void AP_OSD::update_current_screen()
             }
         } else {
             last_switch_ms = 0;
+        }
+        break;
+    case STICKS_INPUT:
+        if (!AP_Notify::flags.armed) {
+            RC_Channels *rc_channels = RC_Channels::get_singleton();
+            RC_Channel *roll = rc_channels->channel(AP::rcmap()->roll() - 1);
+            RC_Channel *pitch = rc_channels->channel(AP::rcmap()->pitch() - 1);
+            RC_Channel *yaw = rc_channels->channel(AP::rcmap()->yaw() - 1);
+            RC_Channel *throttle = rc_channels->channel(AP::rcmap()->throttle() - 1);
+            if (roll->get_aux_switch_pos() == RC_Channel::AuxSwitchPos::MIDDLE &&
+                pitch->get_aux_switch_pos() == RC_Channel::AuxSwitchPos::HIGH &&
+                yaw->get_aux_switch_pos() == RC_Channel::AuxSwitchPos::LOW &&
+                throttle->get_aux_switch_pos() == RC_Channel::AuxSwitchPos::MIDDLE) {
+                if (switch_debouncer) {
+                    uint32_t now = AP_HAL::millis();
+                    if (now - last_switch_ms > 1000) {
+                        next_screen();
+                        last_switch_ms = now;
+                    }
+                } else {
+                    switch_debouncer = true;
+                    return;
+                }
+            } else {
+                last_switch_ms = 0;
+            }
         }
         break;
     }
--- a/libraries/AP_OSD/AP_OSD.h
+++ b/libraries/AP_OSD/AP_OSD.h
@@ -562,6 +562,7 @@ public:
         TOGGLE=0,
         PWM_RANGE=1,
         AUTO_SWITCH=2,
+        STICKS_INPUT=3,
     };
 
     AP_Int8 osd_type;


libraries/AP_OSD/AP_OSD.cpp
```diff
@@ -599,6 +600,32 @@ void AP_OSD::update_current_screen()
             last_switch_ms = 0;
         }
         break;
+    case STICKS_INPUT:
+        if (!AP_Notify::flags.armed) {
+            RC_Channels *rc_channels = RC_Channels::get_singleton();
+            RC_Channel *roll = rc_channels->channel(AP::rcmap()->roll() - 1);
```
peterbarker, line 606:
```suggestion
            auto &rc_channels = AP::rc();
            auto &roll = rc_channels.get_roll_channel();
```

libraries/AP_OSD/AP_OSD.cpp
```diff
@@ -30,6 +30,7 @@
 #include <AP_HAL/AP_HAL.h>
 #include <AP_HAL/Util.h>
 #include <RC_Channel/RC_Channel.h>
+#include <AP_RCMapper/AP_RCMapper.h>
```
peterbarker, line 33:
```suggestion
```

peterbarker (changes requested):
The new option would need to be documented in the relevant parameter description in AP_OSD.cpp

Not sure why we haven't supported this before.  Seems odd.

Hwurzburg:
needs Andy's approval that this does not interfere with his param change code

andyp1per:
Pretty sure this is going to interact very badly with the param stuff and runcam menus. I think probably at a minimum you need to check that you are not in the runcam osd menus and disable this behaviour if you are. The param stuff is the same I think - you need to check you are not in these menus when enabling this stuff.

andyp1per (changes requested):
Need to check whether you are in runcam or osd parameter menus

libraries/AP_OSD/AP_OSD.cpp
prosolvo4th: Done
Thank you for your time

libraries/AP_OSD/AP_OSD.cpp
prosolvo4th: done

prosolvo4th:
> Need to check whether you are in runcam or OSD parameter menus

Hi!
I added checking for runcam menu state (kind of).
But I don't want to add a check on Screen5 or Screen6. I use this functionality to go to Screen5 to adjust the VTX.

andyp1per:
> > Need to check whether you are in runcam or OSD parameter menus
> 
> Hi! I added checking for runcam menu state (kind of). But I don't want to add a check on Screen5 or Screen6. I use this functionality to go to Screen5 to adjust the VTX.

Does it work ok? I mean if its working well with the parameter menus then that's great.

prosolvo4th:
> > > Need to check whether you are in runcam or OSD parameter menus
> > 
> > 
> > Hi! I added checking for runcam menu state (kind of). But I don't want to add a check on Screen5 or Screen6. I use this functionality to go to Screen5 to adjust the VTX.
> 
> Does it work ok? I mean if its working well with the parameter menus then that's great.

yep, so far so good

libraries/AP_OSD/AP_OSD.cpp
```diff
@@ -599,6 +600,36 @@ void AP_OSD::update_current_screen()
             last_switch_ms = 0;
         }
         break;
+    case STICKS_INPUT:
+        auto *runcam = AP::runcam();
+        const uint8_t msg_size = 20;
+        char ret_msg[msg_size] = {0};
+        bool runcam_menu = runcam->pre_arm_check(ret_msg, msg_size);
+        if (!AP_Notify::flags.armed && (!runcam_menu || strcmp(ret_msg, "In OSD menu") != 0)) {
```
andyp1per, line 608: Please just make is_arming_prevented() public and use that

prosolvo4th: sure

prosolvo4th:
@Hwurzburg 
Here are changes in the wiki:
https://github.com/ArduPilot/ardupilot_wiki/pull/6278

libraries/AP_OSD/AP_OSD.cpp
prosolvo4th: done

libraries/AP_OSD/AP_OSD.cpp
```diff
@@ -78,7 +79,7 @@ const AP_Param::GroupInfo AP_OSD::var_info[] = {
     // @Param: _SW_METHOD
     // @DisplayName: Screen switch method
     // @Description: This sets the method used to switch different OSD screens.
-    // @Values: 0: switch to next screen if channel value was changed, 1: select screen based on pwm ranges specified for each screen, 2: switch to next screen after low to high transition and every 1s while channel value is high
+    // @Values: 0: switch to next screen if channel value was changed, 1: select screen based on pwm ranges specified for each screen, 2: switch to next screen after low to high transition and every 1s while channel value is high, 3: switches to next screen if the sticks in the next position: roll - middle, pitch - high, throttle - middle, yaw - low. Keeps toggling to next screen every 1s while sticks in mentioned positions
```
Hwurzburg, line 82: Yaw-low.  Yaw stick can't be low..only left or right

Hwurzburg (changes requested):

libraries/AP_OSD/AP_OSD.cpp
prosolvo4th: Done. Changed to left

libraries/AP_Camera/AP_RunCam.h
```diff
@@ -98,6 +98,8 @@ class AP_RunCam
     void update();
     // Check whether arming is allowed
     bool pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) const;
+    // whether or not we can arm
+    bool is_arming_prevented() const { return _in_menu > _menu_enter_level; }
```
peterbarker, line 102: Probably want an 
```suggestion
    bool in_menus() const { return _in_menu > _menu_enter_level; }
```

and have is_arming_prevented call it.

peterbarker (changes requested):
We don't care if RunCam is preventing arming.

We care that we're in the RunCam menus.

So use an appropriately-named method.

peterbarker:
Byte-cost for this feature:

```
CubeOrange-periph-heavy  *                                                                     
Durandal                            *      *           272     272               272    272    *
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     *      *           120     120               120    112    *
MatekF405                           *      *           112     112               120    112    *
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           *      *           120     112               120    112    *
```

peterbarker:
... also has to pass CI

prosolvo4th:
> ... also has to pass CI

This can be quite difficult.
For some reason sometimes tests can fail for no reason.
I created a simple change and PR with this change in my fork, and CI failed in 2 checks.
I am afraid, I am not able to get a CI 100% pass...

libraries/AP_Camera/AP_RunCam.h
prosolvo4th: Thank you for your time.
Done

libraries/AP_OSD/AP_OSD.cpp
```diff
@@ -599,6 +606,32 @@ void AP_OSD::update_current_screen()
             last_switch_ms = 0;
         }
         break;
+    case STICKS_INPUT:
+        if (!AP_Notify::flags.armed && !AP::runcam()->in_menu()) {
```
andyp1per, line 610: You don't need to check for armed here. You can't be in menu and armed

libraries/AP_Camera/AP_RunCam.cpp
```diff
@@ -244,7 +244,7 @@ bool AP_RunCam::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len)
     }
 
     // currently in the OSD menu, do not allow arming
-    if (is_arming_prevented()) {
```
andyp1per, line 247: please don't change these - it makes it less clear. I don't think this is what @peterbarker was proposing, he suggested having two functions and I think that would be fine.

andyp1per (changes requested):
You will also need to squash the commits

andyp1per:
> We don't care if RunCam is preventing arming.
> 
> We care that we're in the RunCam menus.
> 
> So use an appropriately-named method.

We do internally. I object to this rename internally, but fine for it to be used externally.

libraries/AP_OSD/AP_OSD.cpp
peterbarker: ... also the wrong way to check if you're armed; you ask the arming library directly if you're armed.

peterbarker: > You don't need to check for armed here. You can't be in menu and armed

Disagree.  We should not pre-suppose the inner-workings of AP_RunCam here.

We should both check we are not armed *and* that we aren't going to screw over AP_RunCam somehow.
```suggestion
        if (!AP::arming().is_armed() && !AP::runcam()->in_menu()) {
```

peterbarker: ... actually, you will need to add `#if`s around the call to the runcam singleton here, or we'll end up with compilations failures when runcam isn't compiled in

libraries/AP_OSD/AP_OSD.cpp
```diff
@@ -78,7 +79,7 @@ const AP_Param::GroupInfo AP_OSD::var_info[] = {
     // @Param: _SW_METHOD
     // @DisplayName: Screen switch method
     // @Description: This sets the method used to switch different OSD screens.
-    // @Values: 0: switch to next screen if channel value was changed, 1: select screen based on pwm ranges specified for each screen, 2: switch to next screen after low to high transition and every 1s while channel value is high
+    // @Values: 0: switch to next screen if channel value was changed, 1: select screen based on pwm ranges specified for each screen, 2: switch to next screen after low to high transition and every 1s while channel value is high, 3: switches to next screen if the sticks in the next position: roll - middle, pitch - high, throttle - middle, yaw - left. Keeps toggling to next screen every 1s while sticks in mentioned positions
```
Hwurzburg, line 82: this causes a CI metadata failure
```suggestion
    // @Values: 0: switch to next screen if channel value was changed, 1: select screen based on pwm ranges specified for each screen, 2: switch to next screen after low to high transition and every 1s while channel value is high, 3: switches to next screen if the sticks are held roll-middle pitch-high throttle-middle and yaw-left. Keeps toggling to the next screen every 1s while sticks are in the mentioned positions
```

Hwurzburg (changes requested):

PR #27883, opened 2024-08-21
Plane: Enable quadplane dead reckoning navigation to start earlier

peterbarker:
This reduces EKF errors during transition fromVTOL to FW flight when operating without a GPS fix.

### DIFF
--- a/ArduPlane/ArduPlane.cpp
+++ b/ArduPlane/ArduPlane.cpp
@@ -505,6 +505,11 @@ void Plane::update_fly_forward(void)
         return;
     }
 
+    if (quadplane.in_transition()) {
+        ahrs.set_fly_forward(true);
+        return;
+    }
+
     if (quadplane.in_vtol_mode() ||
         quadplane.in_assisted_flight()) {
         ahrs.set_fly_forward(false);


ArduPlane/ArduPlane.cpp
```diff
@@ -505,6 +505,11 @@ void Plane::update_fly_forward(void)
         return;
     }
 
+    if (quadplane.in_transition()) {
```
tridge, line 508: I think this does need an option, and off by default, likely need Q_OPTION2

tridge (changes requested):

peterbarker:
Logs reflect the new bit being set, 'though does highlights oddities in setting and unsetting this bit as we go through transitions.

This image shows that the bit has the intended effect of setting fly-forward earlier in the transition from qhover to cruise:
image

this image (with the bit set,) shows we glitch fly-forward flag to zero in some stage of the forward transition:
image

ArduPlane/ArduPlane.cpp
```diff
@@ -505,6 +505,11 @@ void Plane::update_fly_forward(void)
...
+        ahrs.set_fly_forward(true);
+        return;
+    }
+
     if (quadplane.in_vtol_mode() ||
         quadplane.in_assisted_flight()) {
```
IamPete1, line 514: Couldn't removing this line amount to the same thing?

IamPete1 (changes requested):
Needs a rebase on https://github.com/ArduPilot/ardupilot/pull/27770

IamPete1:
I think its a good idea, but I think just in transition is is a bit too coarse. Maybe a new method for the the transition class? I think we would be OK to turn on fly forward in the timer stage but maybe not airspeed wait.

libraries/AP_DAL/AP_DAL.h
```diff
@@ -87,14 +87,21 @@ class AP_DAL {
...
+        WHEELENCODER_ENABLED = (1U<<5),
+        TAKEOFF_EXPECTED = (1U<<6),
+        TOUCHDOWN_EXPECTED = (1U<<7),
     };
 
     // EKF ID for timing checks
     enum class EKFType : uint8_t {
-        EKF2 = 0,
-        EKF3 = 1,
+        EKF2 = 2,
```
peterbarker, line 103: Why is this OK?

libraries/AP_DAL/AP_DAL.h
```diff
@@ -87,14 +87,21 @@ class AP_DAL {
     void log_writeDefaultAirSpeed3(const float aspeed, const float uncertainty);
     void log_writeEulerYawAngle(float yawAngle, float yawAngleErr, uint32_t timeStamp_ms, uint8_t type);
 
-    enum class StateMask {
+    enum class RFRNFlags {
```
tridge, line 90: should have a bitwidth?

libraries/AP_DAL/AP_DAL.h
tridge, line 103: can you explain this?

tridge (changes requested):
please check replaying an old log, and double check my 23 bits junkcode test

tridge:
test here for bits, please check!
https://github.com/tridge/junkcode/blob/master/float_int.c

ArduPlane/ArduPlane.cpp
peterbarker: No longer relevant, I think - the option bit makes it different

libraries/AP_DAL/AP_DAL.h
peterbarker: ... because we're just shoving the AHRS value in here

PR #27058, opened 2024-05-13
Create common GSOF library

Ryanf55:
# Purpose

Factor out the parser into its own library so we can use it in EARHS for the PX1. Do it with as little change to the code as possible. 

# Testing Done

SITL test flights in a few different modes on copter and plane. I've also tested on the bench hardware connected to SITL. and a flight test.

# What changed

95% of the code is the same as before, the only difference is that the number of expected packets (the `valid` count), is now exposed up to higher levels. The parser used to hard-code 5 different packet types, which breaks down in EARHS. 
Otherwise, the parser no longer depends directly on types in AP_GPS, so there are now a function in AP_GPS_GSOF called `pack_state_data` to go from the packed GSOF data into the AP_GPS struct.

Longer term, I'd like to switch to a bitfield on which packets were received per DCOL frame.

# Next up

Add the EARHS. To see the WIP branch, see here: https://github.com/Ryanf55/ardupilot/tree/27033-gsof-add-eahrs-driver

# Issue

* Relates to #27033

PR #27887, opened 2024-08-21
AP_GSOF: Use sparse endian instead of custom code

Ryanf55:
# Purpose

Follow up from https://github.com/ArduPilot/ardupilot/pull/27058#discussion_r1691530795

This uses the sparse-endian bit manipulation code rather than the custom stuff in AP_GSOF to save flash.

# Demo

Start plane, arm, takeoff, and observe it fly around in SITL. 
image

CI also has some checks that there are no regressions.


peterbarker (approved):
LGTM

PR #27890, opened 2024-08-21
AP_Mount: Siyi ZT6 hardware id and set-lens fix

rmackay9:
This fixes two problems with the AP_Mount Siyi driver to improve support for the ZT6

1. corrects the Hardware ID.  The manual says it should be '8', '2' but when testing with real hardware it appears to actually be '8','3'    
    image
2. fixes the set-lens method (used when the pilot selects the active lens using an Aux switch or the MP AuxFunctions tab).  The order is now:
    - low: RGB
    - mid: Thermal
    - high: Picture-in-picture RGB / Thermal

This has been tested on real hardware (a ZT6 only, I do not have a ZT30).  Below is a screen shot of MP's messages tab.  Before this change the "Mount: Siyi ZT6" message did not appear.

image

I think we should backport this to 4.5 because we claimed that we had ZT6 support in an earlier point release.  The ZT6 actually does partially work without this fix in that the gimbal can be moved around.  It's just that set-len doesn't work properly.

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Siyi.cpp
+++ b/libraries/AP_Mount/AP_Mount_Siyi.cpp
@@ -28,7 +28,7 @@ const AP_Mount_Siyi::HWInfo AP_Mount_Siyi::hardware_lookup_table[] {
         {{'7','3'}, "A8"},
         {{'6','B'}, "ZR10"},
         {{'7','8'}, "ZR30"},
-        {{'8','2'}, "ZT6"},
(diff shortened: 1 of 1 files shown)


rmackay9:
Siyi has confirmed that there was a mistake in the documentation so I think this is safe to merge

tridge (approved):

rmackay9:
this could not be included in 4.5.7-beta1 due to a merge conflict

PR #27893, opened 2024-08-21
AP_InertialSensor: support fast rate primary on MPU6000

andyp1per:
### Summary

Give the Invensense v1 driver the same set_primary() handling as v3 so the fast rate loop's dynamic FIFO works on an MPU6000: a non-primary v1 IMU is read at a reduced rate and the primary at the full backend rate. Split out from #27029.

### Classification & Testing

- [x] Tested on hardware
- [x] Logs available on request

Tested on a Matek H743 Slim (MPU6000 primary) and on a client board with an MPU6000 in the second slot, fast sampling on and off. The current head differs from what was flown only by removing the resync described below, which review showed never fired in steady state.

### Description

Testing with an MPU6000 in the second slot found that the v3 bound of 2x loop rate is too slow for v1: the v1 FIFO fills at the 8kHz sensor rate whatever the backend rate, and at 400Hz beats it reached the depth where the driver already discards corrupt samples ("stop at 8 of 48" and temperature resets). A fast-sampling non-primary is now held at 1kHz, one read buffer per beat; the 2x loop rate scaling applies only without fast sampling, where the FIFO fills at the backend rate as on v3.

An earlier version also resynced the primary's beat to the FIFO when a read came up short. The check compared the raw gyro count with the downsample rate, and since the count is only reset by the accel block it never fired in steady state. Fixing it is the wrong answer: adjust_periodic_callback() moves the next beat to now + period, and now is after the wake latency and the count read, so each correction overshoots the drift it is fixing and is undone within a few beats. A model of the DeviceBus timer gives roughly 50 empty beats and 50 double emits per second at 1-4kHz with an IMU clock 0.1% slower than the MCU, against a handful with no resync, and with the rate thread on the backend's 200ms re-anchor in update_primary() dominates anyway. The resync has been dropped.

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor_Invensense.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor_Invensense.cpp
@@ -26,6 +26,7 @@
 #include <AP_InternalError/AP_InternalError.h>
 #include <AP_Logger/AP_Logger.h>
 
+#include "AP_InertialSensor_rate_config.h"
 #include "AP_InertialSensor_Invensense.h"
 #include <GCS_MAVLink/GCS.h>
 
@@ -436,7 +437,7 @@ void AP_InertialSensor_Invensense::start()
     }
 
     // start the timer process to read samples, using the fastest rate avilable
-    _dev->register_periodic_callback(1000000UL / _gyro_backend_rate_hz, FUNCTOR_BIND_MEMBER(&AP_InertialSensor_Invensense::_poll_data, void));
+    periodic_handle = _dev->register_periodic_callback(1000000UL / _gyro_backend_rate_hz, FUNCTOR_BIND_MEMBER(&AP_InertialSensor_Invensense::_poll_data, void));
 }
 
 // get a startup banner to output to the GCS
@@ -481,6 +482,21 @@ bool AP_InertialSensor_Invensense::update() /* front end */
     }
 
     return true;
+}
+
+void AP_InertialSensor_Invensense::set_primary(bool _is_primary)
+{
+#if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
+    if (_imu.is_dynamic_fifo_enabled(gyro_instance)) {
+        if (_is_primary) {
+            _dev->adjust_periodic_callback(periodic_handle, 1000000UL / _gyro_backend_rate_hz);
+        } else {
+            // scale down non-primary to 2x loop rate, but no greater than the default sampling rate
+            _dev->adjust_periodic_callback(periodic_handle,
+                                           1000000UL / constrain_int16(get_loop_rate_hz() * 2, 400, 1000));
+        }
+    }
+#endif
 }
 
 /*
--- a/libraries/AP_InertialSensor/AP_InertialSensor_Invensense.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor_Invensense.cpp
@@ -74,6 +74,9 @@ extern const AP_HAL::HAL& hal;
 #define MPU_SAMPLE_SIZE 14
 #define MPU_FIFO_BUFFER_LEN 8
 
+// rate the FIFO is filled at when fast sampling, independent of the backend rate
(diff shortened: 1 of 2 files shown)


tridge:
@andyp1per what testing have you done on boards with MPU6k or MPU9250, with and without fast sampling?

andyp1per:
> @andyp1per what testing have you done on boards with MPU6k or MPU9250, with and without fast sampling?

I tested on a Matek H743 Slim v1 which has the MPU 6000. The fast rate code will not work without this change. I switched it on and off successfully. I have not tried without fast sampling but can do that as well.

andyp1per:
Worked with a client with this IMU, tried both on and off successfully - found another bug in the process.

tridge:

<details><summary>Previous review</summary>

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_02/devcall_pr_reviews.html#pr27893

Reviewed at head `3329994a30`. **REQUEST CHANGES** — one functional gap in the mechanism this PR adds.

### The resync never fires at the maximum fast rate

The new re-adjust in `_read_fifo()` is gated on `is_primary && n_samples + _accum.gyro_count < _gyro_fifo_downsample_rate`. At `fast_sampling_rate == 8` the downsample rate is **1**, and `_accum.gyro_count` is zeroed *only* inside the accel block (line 706) and then unconditionally incremented at 710 — the gyro emit block at 721 zeroes the accumulator vector but never the count. So the count is essentially always ≥1, and `n_samples + (≥1) < 1` is false even with an empty FIFO. The adjustment therefore runs at most once, before the first processed sample, which defeats the primary-rate synchronisation the PR exists to add.

A dedicated pending-gyro-phase counter (or carefully validated modulo logic) rather than raw `_accum.gyro_count` would fix it.

### Checked and clear

`periodic_handle` being left uninitialised is correct under house style — both probe paths use `NEW_NOTHROW`, whose global implementation calls `calloc`. Re-adjusting the callback does not thrash the timer either: on ChibiOS `adjust_periodic_callback()` only rewrites `period_usec`/`next_usec` with no reallocation or re-registration; Linux takes one `timerfd_settime()` and only on a short read.

Minor: in `set_primary()`, when `_fast_sampling` is true `min_rate_hz` is 8000/8 = 1000 and the constrain upper bound is also 1000, so the loop-rate term is dead in that branch. Both review passes read that as intentional (the comment requires draining a whole FIFO buffer per beat); noting it only because it reads as a live computation. CI 99/99 green.

</details>

tridge:

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_02/devcall_pr_reviews.html#pr27893

Re-reviewed at head `b4c8c0e9be` (was `3329994a30`); my earlier comment above is superseded. **Both findings are resolved — APPROVE.**

**The resync that could never fire is gone.** Rather than repairing the guard, the whole block gated on `is_primary && n_samples + _accum.gyro_count < _gyro_fifo_downsample_rate` has been removed from `_read_fifo()` — there is no longer any `adjust_periodic_callback` inside the FIFO read at all. That is the better outcome: `_accum.gyro_count` survives only in its original accel/gyro downsampling roles, so nothing else depended on the semantics that made the guard dead.

**And the feature is not lost with it.** `set_primary()` now branches explicitly — the primary beats at `_gyro_backend_rate_hz`, and a non-primary with `_fast_sampling` beats at `MPU_FAST_SAMPLE_RATE_HZ / MPU_FIFO_BUFFER_LEN` = 8000/8 = **1 kHz**, i.e. one full 8-sample buffer per beat against an 8 kHz fill. That is balanced and well clear of the depth at which `_read_fifo()` finds corrupt samples, which is what the removed dynamic adjustment was reaching for.

The minor note from last round is resolved by the same change: the `_fast_sampling` branch previously had `min_rate_hz` = 1000 against a constrain upper bound of 1000, making the loop-rate term dead. It is now its own branch with no constrain, and the non-fast branch keeps `constrain_int16(loop_rate*2, 400, 1000)` unchanged.

CI still running (1 passing, 83 pending).

tridge (approved):

PR #27895, opened 2024-08-21
Support LTO Builds

andyp1per:
Examples work, but not sure much else


tpwrules:
I had investigated this in the past and done some work to fix related compiler errors but I wasn't sure if it was possible. In particular I'm not sure wrapped functions are safe to use, even if they are marked as used? IIUC there is a [GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88643) where it can resolve functions before they can get wrapped, although it looks like maybe there's a fix there?

I think we should kill the wrapped functions ultimately though.

There's also the problem that `-O2` boards now optimize too well and overflow the flash.

andyp1per:
@tpwrules great! Do you want to collaborate? The examples work for me, so clearly not all is broken - I am hoping its something simple. Apparently all the ChibiOS examples are built with LTO so there should be good underlying support.

PR #27896, opened 2024-08-21
Copter: 4.5.6-beta1 fixup

rmackay9:
This resolves a small difference between Copter/Rover and Plane/Tracker caused by slightly different versions of PR https://github.com/ArduPilot/ardupilot/pull/27698 being merged to Plane-4.5 vs master


tpwrules (approved):
LGTM

PR #27901, opened 2024-08-22
AP_Mount_Siyi: Bugfix for GPS position in EXIF from Siyi camera

igarashi87:
The longitude in the EXIF data was not being saved correctly when shooting with the Siyi A8 mini*, so I made a slight modification to the code.
I will show the properties of the captured images before and after the code modification below(Please forgive that this is a Japanese-language environment).

- Before(Longitude is an abnormal value)
before

- After(Longitude was corrected to the appropriate value)
after

The shooting location is Tokyo, Japan. Normally, the latitude should be around 35.6, and the longitude should be around 139.6.

Please confirm this modification.

*Camera Firmware: v0.2.6, Gimbal Firmware: v0.3.7

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Siyi.cpp
+++ b/libraries/AP_Mount/AP_Mount_Siyi.cpp
@@ -1180,7 +1180,7 @@ void AP_Mount_Siyi::send_attitude_position(void)
         uint32_t time_boot_ms;
         int32_t lat, lon;
         int32_t alt_msl, alt_ellipsoid;
-        Vector3l velocity_ned_int32;
+        Vector3f velocity_ned_int32;
     } position;
     Location loc;
     Vector3f velocity_ned;


rmackay9:
Hi @igarashi87,

There are some known issues is the Siyi gimbal firmware that cause incorrect EXIF location data to be written to the images.

The Siyi developers specifically told me that the velocities have been changed to int32 in mm which is why we changed the format in this PR https://github.com/ArduPilot/ardupilot/pull/27457

I think we need to wait until the next version of the Siyi firmware is released

igarashi87:
Hi @rmackay9

Thanks for your reply. I now know why it was set to int32t.
I too hope that the next version of the Siyi firmware will be released.

rmackay9:
@igarashi87,

Txs, I ping'd Siyi yesterday about a few issues including the EXIF issue so I hope they will respond in the coming days.  By the way, it's great to see other developers working on this driver!

peterbarker:
Ping @rmackay9 - there's been movement on this, I believe.

rmackay9:
Hi @peterbarker, @igarashi87,

I have not received an updated A8 firmware yet but they sent me this picture for the ZT6 a few days ago and it again shows the speeds being sent using int32_t.  So I'm pretty sure we should not merge this PR.

image

I'm going to go ahead and close this PR actually but we can re-open if we find out that we need the change

PR #27903, opened 2024-08-22
AP_camera: Camera tracking example scripts

khanasif786:
This PR realtes to #27773
There are currently two scripts more scripts may be added.
1.  send_camera_information send the camera information from companion computer to the FC running Ardupilot over network. 2. 2. the tracking.py scripts runs in parallel thread detecting humans and then sending PITCH YAW commands to FC. You need to specify the gstreamer pipeline command in string format to the example script. By default the example script uses the udp stream.  ehich the gstreamer plugin from gazebo uses


timtuxworth:
Could you use an RTSP stream direct without needing to use gstreamer?

khanasif786:
@timtuxworth , yeah there are workarounds, just need to catch the rtsp stream and give it to cv2 capture

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,98 @@
...
+    def send_camera_information(self):
+        self.connection.mav.camera_information_send(
+            int(time.time() * 1000) & 0xFFFFFFFF,          # time_boot_ms
+            b"CameraVendor123" + b'\0' * (32 - len("CameraVendor123")),  # vendor_name
+            b"CameraModel123" + b'\0' * (32 - len("CameraModel123")),    # model_name
+            (1 << 24) | (0 << 16) | (0 << 8) | 1,          # firmware_version
+            float('nan'),                                  # focal_length
+            float('nan'),                                  # sensor_size_h
+            float('nan'),                                  # sensor_size_v
+            640,                                           # resolution_h
```
rmackay9, line 40: ideally we should pull these resolutions from the rtsp stream that we're using.  Not a blocker though

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,98 @@
...
+            480,                                           # resolution_v
+            0,                                             # lens_id
+            4095,                                          # flags
```
peterbarker, line 43: SHould use enumeration entries rather than just a number.  Maybe create a `flags` local variable

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,98 @@
...
+        self.send_camera_information()
+
+        # Start the heartbeat thread
+        heartbeat_thread = threading.Thread(target=self.send_heartbeat)
+        heartbeat_thread.daemon = True
+        heartbeat_thread.start()
+
+        while True:
+            msg = self.connection.recv_match(type='COMMAND_LONG', blocking=True)
+            if msg and msg.get_type() == 'COMMAND_LONG' and msg.command == mavutil.mavlink.MAV_CMD_CAMERA_TRACK_POINT:
```
rmackay9, line 79: we should eventually implement TRACK_RECTANGLE.  again not a blocker though

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,131 @@
...
+from pymavlink import mavutil
+import math 
+from ultralytics import YOLO
+import argparse
+
+class DroneGimbalControl:
+    def __init__(self, video_source=None):
+        # Set default video source if not provided
+        if video_source is None:
+            video_source = 'udpsrc port=5600 ! application/x-rtp,encoding-name=H264,payload=96 ! rtph264depay ! h264parse ! queue ! avdec_h264 ! videoconvert ! appsink'
```
rmackay9, line 18: I thought that with the siyi cameras, only H265 worked.  I don't think H264 streaming works with them

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,131 @@
...
+            centre_x_copy = int(self.center_x)
+            centre_y_copy = int(self.center_y)
+            if (centre_x_copy == 0 and centre_y_copy == 0):
+                diff_x = 0
+                diff_y = 0
+            else:
+                diff_x = (centre_x_copy - (640 / 2)) / 2
+                diff_y = -(centre_y_copy - (480 / 2)) / 2
+
+            self.send_gimbal_manager_pitch_yaw_angles(float("NaN"), float("NaN"), math.radians(diff_y), math.radians(diff_x))
```
peterbarker, line 62: Need to rate-limit-this

rmackay9, line 62: As PeterB mentioned, we should ensure that this runs no faster than 50hz or maybe even 20hz

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def update_frame(self):
+        while True:
+            ret, frame = self.cap.read()
```
peterbarker, line 66: See if you find a blocking API, or in some other way stop this burning an entire core.

`time.sleep(0.01)` if nothing else

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,131 @@
...
+            if ret:
+                with self.frame_lock:
+                    self.latest_frame = frame
+
+    def detect_faces(self):
+        final_list = []
+        while True:
+            if self.latest_frame is not None:
```
peterbarker, line 74: early return and `continue` patterns would make this indenting less extreme.

peterbarker (changes requested):
Review data integrity

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,131 @@
...
+        self.target_system = self.master.target_system
+        self.target_component = self.master.target_component
+        self.flags = 0
+        self.gimbal_device_id = 0
+
+        self.cap = cv2.VideoCapture(video_source, cv2.CAP_GSTREAMER)
+        self.model = YOLO("yolov8n.pt")
+        self.frame_lock = threading.Lock()
+        self.latest_frame = None
+        self.center_y = 0
```
peterbarker, line 35: Need a lock for these as they are shared state.

Consider a little "Centre" object to contain x, y, and the lock

peterbarker (changes requested):

libraries/AP_Camera/examples/tracking.py
khanasif786: Thanks, No longer applicable actually changed this.

libraries/AP_Camera/examples/tracking.py
khanasif786: Done

libraries/AP_Camera/examples/tracking.py
khanasif786: done

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,148 @@
...
+                if msg.target_system == self.sysid:
+                    if msg.command == mavutil.mavlink.MAV_CMD_CAMERA_TRACK_POINT:
+                        self.handle_camera_track_point(msg)
+                    elif msg.command == mavutil.mavlink.MAV_CMD_CAMERA_TRACK_RECTANGLE:
+                        self.handle_camera_track_rectangle(msg)
+                else:
+                    print("Received but not for us")
+
+def main():
+    parser = argparse.ArgumentParser(description="A script to demonstrate command-line arguments for sysid and compid.")
```
amilcarlucas, line 136: "A script to send camera information" would be a better description

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,172 @@
...
+            data, addr = self.sock.recvfrom(16)
+            x, y, w, h = struct.unpack('!ffff', data)
+            print(f"Received new tracking coordinates: x={x}, y={y}, w={w}, h={h}")
+            # Convert normalized coordinates to pixel values
+            x, y, w, h = int(x * self.video_streamer.frame_width), int(y * self.video_streamer.frame_height), int(w * self.video_streamer.frame_width), int(h * self.video_streamer.frame_height)
+            self.video_streamer.update_tracker((x, y, w-x, h-y))
+
+
+def main():
+    gimbal_control = GimbalControl('127.0.0.1:14560')
```
amilcarlucas, line 165: some argparse arguments and documentation would be nice

amilcarlucas (changes requested):

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, input_port, output_port):
+        self.input_port = input_port
+        self.output_port = output_port
+        self.cap = self.initialize_video_capture()
+        self.frame_width = int(self.cap.get(cv2.CAP_PROP_FRAME_WIDTH))
+        self.frame_height = int(self.cap.get(cv2.CAP_PROP_FRAME_HEIGHT))
+        self.out = self.initialize_video_writer()
+        self.frame = None
+
+        self.tracker = cv2.legacy.TrackerCSRT_create()  # Change to legacy namespace
```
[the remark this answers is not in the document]
khanasif786: Thanks, i was thinking similar. 😃

libraries/AP_Camera/examples/ReadMe.md
```diff
@@ -0,0 +1,79 @@
...
+sudo apt install -y build-essential cmake git pkg-config libgtk-3-dev libavcodec-dev libavformat-dev libswscale-dev
+
+sudo apt install -y python3-dev python3-numpy python3-pip
+
+sudo apt install -y libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev \
+                    gstreamer1.0-plugins-good gstreamer1.0-plugins-bad \
+                    gstreamer1.0-plugins-ugly gstreamer1.0-libav gstreamer1.0-tools
+```
+
+3. default opencv from pip should be uninstalled and again built with gstreamer support (If already did this go to next step)
```
srmainwaring, line 26: I think we should avoid requiring a custom build of opencv to include gstreamer support, and instead use standard packages. We can add a class to convert from the gstreamer frames to ones compatible with opencv, such as the example here: https://www.ardusub.com/developers/opencv.html

libraries/AP_Camera/examples/tracking.py
```diff
@@ -0,0 +1,183 @@
...
+                centre_y_copy = int(self.center_y)
+
+            if (centre_x_copy == 0 and centre_y_copy == 0):
+                diff_x = 0
+                diff_y = 0
+            else:
+                diff_x = (centre_x_copy - (640 / 2)) / 2
+                diff_y = -(centre_y_copy - (480 / 2)) / 2
+
+            self.send_gimbal_manager_pitch_yaw_angles(float("NaN"), float("NaN"), math.radians(diff_y), math.radians(diff_x))
```
srmainwaring, line 49: I think we're going to need PI(D) controllers to set gains on the conversion from `(diff_x, diff_y)` to `(pitch_rate_rads, yaw_rate_rads)`. From initial testing with a SIYI A8, P gains of around 0.1 are a reasonable start. Unadjusted, the gimbal overshoots and tracking is lost immediately.

libraries/AP_Camera/examples/send_camera_information.py
```diff
@@ -0,0 +1,152 @@
...
+        param2 = msg.param2
+        print(f"Tracking point parameters: param1={param1}, param2={param2}")
+
+    def handle_camera_track_rectangle(self, msg):
+        print("Received MAV_CMD_CAMERA_TRACK_RECTANGLE command.")
+        # These should remain as floats (normalized coordinates)
+        norm_x = msg.param1
+        norm_y = msg.param2
+        norm_w = msg.param3
+        norm_h = msg.param4
```
srmainwaring, line 98: https://mavlink.io/en/messages/common.html#MAV_CMD_CAMERA_TRACK_RECTANGLE

- param3 is bottom right corner x not width
- param4 is bottom right corner y not height

srmainwaring:
@khanasif786 - I've rebased this PR on master here: https://github.com/srmainwaring/ardupilot/tree/mav_cmd_camera_example_scripts. 

`ReadMe.md` in this PR need to be renamed before rebasing `README.md`. This has to be done in a few steps to avoid issues on case insensitive filesystems such as the one used by macOS. The steps are `ReadMe.md` -> `ReadMe_.md` -> squash into 2d753992 -> `ReadMe_.md` -> `README.md` -> squash, then rebase on master. I tried to force push my branch but don't have permission.

rmackay9:
A couple of small things:

1. we should squash the commits
2. let's change the README file name to match conventional case

PR #27906, opened 2024-08-23
HAL_ChibiOS: fixed the imu orientation of RadiolinkPIX6

radiolinkW:
HAL_ChibiOS: fixed the imu orientation of RadiolinkPIX6
In order to obtain better baro data, the orientation of IMU has been adjusted.


peterbarker:
Please tell me that none of these boards are have been sent to anyone?

ie. this sort of change makes using boards with the old orientation bad,

radiolinkW:
> Please tell me that none of these boards are have been sent to anyone?
> 
> ie. this sort of change makes using boards with the old orientation bad,

We have alreay thought about what you are worried about, so don't worry. 
The previous boards were only used for testing, and the ones we sent out were the latest adjusted ones.

tridge (approved):

PR #27909, opened 2024-08-23
Plane: Allow for Guided mode to support Terrain following

timtuxworth:
Starting with support for MAVLink CMD_GUIDED_CHANGE_ALTITUDE and CMD_DO_REPOSITION to support MAV_FRAME_GLOBAL_ALT, this adds translation from the MAVlink frame to Location::AltFrame:ABOVE_TERRAIN.

To make this easier to understand and maintain, I also changed the plane.guided_state.target_alt_frame variable directly support Location::AltFrame directly (it was a uint8_t)

Then lower level changes required to support the functionality were required in mode_guided.cpp.

The example script "works around" the current lack of the ability of GCS such as Mission Planner or QGroundControl to send altitude request messages with terrain. The script will convert non-terrain guided terrain targets into terrain targets. There is a parameter ZGT_MODE which allows choosing how the script will do this:
1: reset the requested terrain altitude to the current plane terrain altitude
2: reset to the requested altitude above/below home as a terrain altitude to target
3: reset to the current altitude + the above/below home altitude as the terrain altitude to target

### DIFF
--- a/libraries/AP_Scripting/examples/plane_guided_terrain.lua
+++ b/libraries/AP_Scripting/examples/plane_guided_terrain.lua
@@ -1,0 +1,185 @@
+--[[
+
+   example script to show interrupting a mission and then 
+   reseting the crosstracking to the correct line when
+   returning to the mission after the interruption
+
+   this functionality is only available in Plane
+--]]
+
+SCRIPT_NAME = "OverheadIntel Guided Terrain"
+SCRIPT_NAME_SHORT = "TerrGuided"
+SCRIPT_VERSION = "4.6.0-002"
+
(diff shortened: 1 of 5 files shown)


timtuxworth:
tagging @Ryanf55 I think it's the right Ryan

libraries/AP_Scripting/examples/plane_guided_terrain.lua
```diff
@@ -0,0 +1,185 @@
+--[[
+
+   example script to show interrupting a mission and then 
+   reseting the crosstracking to the correct line when
```
clydemcqueen, line 4: Typo: resetting

timtuxworth:
Closed in favour of #27921

PR #27911, opened 2024-08-24
AP_Camera/AP_Mount/AP_Scripting: add change settings method

rmackay9:
This adds a new general purpose "change_settings" method to AP_Camera (and AP_Mount) which allows us to add support for various camera settings without creating methods for each.  The vast majority of camera settings are simply a number.

The new method is used to implement three commonly used settings for the Siyi ZT6 and ZT30 thermal cameras:

- Thermal Palette (e.g. WhiteHot, BlackHot, Rainbow, etc)
- Thermal Gain (aka temperature range)
- Thermal Raw Data (stores raw temp data along with thermal images when a picture is taken)

A new change_camera_settings.lua script is also added which creates user visible parameters for these settings.

This has been tested on real hardware and below is a picture of a test where I changed the thermal palette on a ZT6 between "4" (Rainbow) and "10" (BlackHot)

image
image
image

We should eventually implement the change for the ViewPro and Topotek drivers as well which also support these features.  Topotek in particular has specifically mentioned they would like the ability to easily change the Palette.

### DIFF
--- a/libraries/AP_Camera/AP_Camera.cpp
+++ b/libraries/AP_Camera/AP_Camera.cpp
@@ -780,6 +780,19 @@ bool AP_Camera::get_state(uint8_t instance, camera_state_t& cam_state)
     }
     return backend->get_state(cam_state);
 }
+
+// change camera settings not normally used by autopilot
+bool AP_Camera::change_setting(uint8_t instance, CameraSetting setting, float value)
+{
+    WITH_SEMAPHORE(_rsem);
+
+    auto *backend = get_instance(instance);
+    if (backend == nullptr) {
+        return false;
+    }
+    return backend->change_setting(setting, value);
+}
+
 #endif // #if AP_CAMERA_SCRIPTING_ENABLED
 
 // return backend for instance number
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -1409,6 +1409,16 @@ function AP_Camera__camera_state_t_ud:take_pic_incr() end
 ---@return AP_Camera__camera_state_t_ud|nil
 function camera:get_state(instance) end
 
+-- Change a camera setting to a given value
+---@param instance integer
+---@param setting integer
+---| '0' # THERMAL_PALETTE
+---| '1' # THERMAL_GAIN
+---| '2' # THERMAL_RAW_DATA
+---@param value number
+---@return boolean
+function camera:change_setting(instance, setting, value) end
+
 -- desc
 mount = {}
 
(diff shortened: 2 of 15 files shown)


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -1409,6 +1409,17 @@ function AP_Camera__camera_state_t_ud:take_pic_incr() end
 ---@return AP_Camera__camera_state_t_ud|nil
 function camera:get_state(instance) end
 
+---
+-- desc
```
IamPete1, line 1413:
```suggestion
-- Change a numbered camera setting to a given value
```

libraries/AP_Camera/AP_Camera.cpp
```diff
@@ -780,6 +780,19 @@ bool AP_Camera::get_state(uint8_t instance, camera_state_t& cam_state)
     }
     return backend->get_state(cam_state);
 }
+
+// change camera settings not normally used by autopilot
```
peterbarker, line 784:
```suggestion
#if AP_CAMERA_CHANGE_SETTING_API_ENABLED
// change camera settings not normally used by autopilot
```

peterbarker (changes requested):
... and something in `build_options.py`

libraries/AP_Camera/AP_Camera.cpp
Hwurzburg: I edited @peterbarker note above

libraries/AP_Scripting/docs/docs.lua
rmackay9: ok thanks, I've changed the wording to, "Change a camera setting to a given value"

libraries/AP_Camera/AP_Camera.cpp
rmackay9: I really think we are trying to be too granular.  This is a tiny feature already hidden behind AP_SCRIPTING.  I'll will go ahead and add it because that's the easiest path to getting this merged.

rmackay9:
Here's the sizes

```
antennatracker   212 (+0.0159%)  0 (0.0000%)  -4 (-0.0015%)  212 (+0.0159%)                                   632076
arduplane        484 (+0.0269%)  0 (0.0000%)  4 (+0.0015%)   484 (+0.0269%)                                   165288
blimp            212 (+0.0157%)  0 (0.0000%)  4 (+0.0015%)   212 (+0.0156%)                                   609816
arducopter-heli  484 (+0.0269%)  0 (0.0000%)  4 (+0.0015%)   484 (+0.0269%)                                   163248
arducopter       484 (+0.0269%)  0 (0.0000%)  4 (+0.0015%)   484 (+0.0268%)                                   162264
ardusub          484 (+0.0304%)  0 (0.0000%)  -4 (-0.0015%)  484 (+0.0304%)                                   371860
ardurover        484 (+0.0293%)  0 (0.0000%)  4 (+0.0015%)   484 (+0.0293%)                                   313936
```

PR #27913, opened 2024-08-24
AP_RangeFinder: RDS02UF: avoid nuking partial messages after good parse

peterbarker:
there may be a fragment of another packet in the buffer


rishabsingh3003 (approved):
Looks good! Thanks.
I have tested the RDS02UF I have on bench with this branch and it seems to be working normally.

peterbarker:
> Looks good! Thanks. I have tested the RDS02UF I have on bench with this branch and it seems to be working normally.

Thanks for testing it!

PR #27915, opened 2024-08-24
AP_Camera/AP_Mount/GCS_MAVLink: Siyi implements camera-thermal-range

rmackay9:
This PR adds support for sending the min and max temperatures and their XY positions in the live video stream to the ground station when using Siyi ZT6 and ZT30 cameras.  We can add support for ViewPro and Topotek cameras in a follow-up PRs.

This is useful for search & rescue and firefighting applications because the ground station can overlay a rectangle over the hotspot along with its temperature.

This PR relies on this mavlink PR https://github.com/ArduPilot/mavlink/pull/369 (and the equivalent upstream PR https://github.com/mavlink/mavlink/pull/2145).

It is also built on PR https://github.com/ArduPilot/ardupilot/pull/27890 and PR https://github.com/ArduPilot/ardupilot/pull/27911 so those should be merged first.

This has been lightly tested on real-hardware (a CubeOrange + Siyi ZT6) and below is a screen shot showing the temperature jump when I moved in front of the camera although please note that the test used named value pairs to send the temperatures and positions instead of using the new mavlink message.

temp-minmax-test


rmackay9:
This has been updated inline with discussion on the upstream mavlink PR https://github.com/mavlink/mavlink/pull/2145

rmackay9:
I've updated this to use the final CAMERA_THERMAL_RANGE mavlink message definition and also hidden the feature behind a define

Currently the new message is only sent if requested by the GCS (e.g. not sent by default)

libraries/AP_Camera/AP_Camera_Mount.cpp
```diff
@@ -109,6 +109,19 @@ void AP_Camera_Mount::send_camera_capture_status(mavlink_channel_t chan) const
...
+void AP_Camera_Mount::send_camera_thermal_range(mavlink_channel_t chan) const
+{
+#if AP_MOUNT_SEND_THERMAL_RANGE_ENABLED
+    AP_Mount* mount = AP::mount();
+    if (mount != nullptr) {
+        mount->send_camera_thermal_range(get_mount_instance(), chan);
+    }
+#endif
+}
+#endif
```
peterbarker, line 123:
```suggestion
#endif  // AP_CAMERA_SEND_THERMAL_RANGE_ENABLED
```

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -1060,6 +1060,7 @@ ap_message GCS_MAVLINK::mavlink_id_to_ap_message_id(const uint32_t mavlink_id) c
         { MAVLINK_MSG_ID_CAMERA_SETTINGS,       MSG_CAMERA_SETTINGS},
         { MAVLINK_MSG_ID_CAMERA_FOV_STATUS,     MSG_CAMERA_FOV_STATUS},
         { MAVLINK_MSG_ID_CAMERA_CAPTURE_STATUS, MSG_CAMERA_CAPTURE_STATUS},
+        { MAVLINK_MSG_ID_CAMERA_THERMAL_RANGE,  MSG_CAMERA_THERMAL_RANGE},
```
peterbarker, line 1063:
```suggestion
#if AP_CAMERA_SEND_THERMAL_RANGE_ENABLED
        { MAVLINK_MSG_ID_CAMERA_THERMAL_RANGE,  MSG_CAMERA_THERMAL_RANGE},
#endif
```

peterbarker (changes requested):
LGTM past these

libraries/AP_Camera/AP_Camera_Mount.cpp
rmackay9: comment added, thanks!

libraries/GCS_MAVLink/GCS_Common.cpp
rmackay9: I might leave this one out if that's OK.  If you look up a few lines you'll see that the CAMERA_FOV_STATUS is also not behind a definition.

rmackay9:
feature size info
```
arducopter       692 (+0.0384%)  0 (0.0000%)  4 (+0.0015%)   692 (+0.0384%)                                   161192
antennatracker   108 (+0.0081%)  0 (0.0000%)  4 (+0.0015%)   108 (+0.0081%)                                   631524
arducopter-heli  692 (+0.0385%)  0 (0.0000%)  -4 (-0.0015%)  692 (+0.0384%)                                   162160
ardusub          692 (+0.0435%)  0 (0.0000%)  -4 (-0.0015%)  692 (+0.0434%)                                   370788
ardurover        688 (+0.0417%)  0 (0.0000%)  0 (0.0000%)    688 (+0.0416%)                                   312848
blimp            112 (+0.0083%)  0 (0.0000%)  0 (0.0000%)    112 (+0.0083%)                                   609312
arduplane        696 (+0.0387%)  0 (0.0000%)  0 (0.0000%)    696 (+0.0386%)                                   164248
```

libraries/GCS_MAVLink/GCS_Common.cpp
peterbarker: This one really should be in.  Feel free to fix the other one - and group them together would be even nicer!

rmackay9: ok, added, thanks!

PR #27916, opened 2024-08-24
AP_Math: allow rand_float() on STM32

peterbarker:


peterbarker:
```
CubeOrange-periph-heavy  *                                                                     
Hitec-Airspeed           *                                                                     
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
```

tpwrules:
The non-SITL implementation is wrong? It can't ever generate a negative number.

16/32 bits is a bit coarse too, but not sure if that matters for intended uses.

peterbarker:
> The non-SITL implementation is wrong? It can't ever generate a negative number.

Wow, whoops.  That was my bad.  Should be fixed now.

> The SITL comment probably needs updating too. 16/32 bits is a bit coarse, but not sure if that matters for intended uses.

This affects simulation-on-hardware - just for adding noise, really.

Thanks!

libraries/AP_Math/AP_Math.cpp
```diff
@@ -339,14 +339,13 @@ uint16_t get_random16(void)
...
-#if AP_SIM_ENABLED
-// generate a random float between -1 and 1, for use in SITL
+// generate a random float between -1 and 1
 float rand_float(void)
 {
 #if CONFIG_HAL_BOARD == HAL_BOARD_SITL
     return ((((unsigned)random()) % 2000000) - 1.0e6) / 1.0e6;
 #else
-    return get_random16() / 65535.0;
+    return ((get_random16() / 65535.0) -1) * 2;
```
tpwrules, line 348:
```suggestion
    return ((get_random16() / 65535.0) - 0.5) * 2;
```

You fixed it too well, now it can’t return positive numbers.

peterbarker: Man.  Order of operations failure.

I've patched this.  And added a test....

tpwrules (approved):
LGTM

PR #27917, opened 2024-08-24
Plane: Allow wind relative dead reckoning when lift motors are assisting

peterbarker:
When lift motors are assisting the wing during forward flight, the airspeed and zero sideslip assumption are still sufficiently valid to keep navigating. Without this patch a VTOL operating in FW mode without GPS and reliant on air data for dead reckoning will stop navigating and lose its position estimate if the lift motors start assisting.

I'm extracting this from a larger piece of work.

This reasoning for this patch (abnove, as part of the patch comment) is in direct contravention of the reasoning in https://github.com/ArduPilot/ardupilot/pull/6153 , the patch which added this.

### DIFF
--- a/ArduPlane/Plane.cpp
+++ b/ArduPlane/Plane.cpp
@@ -533,11 +533,6 @@ void Plane::update_fly_forward(void)
             ahrs.set_fly_forward(false);
             return;
         }
-
-        if (quadplane.in_assisted_flight()) {
-            ahrs.set_fly_forward(false);
-            return;
-        }
     }
 #endif
 


priseborough:
The original concern about quadplanes violating the fly forward assumption when assisting turned out to be invalid because they weathervane as airspeed rises.

IamPete1:
Seem to be the same idea as https://github.com/ArduPilot/ardupilot/pull/27917?

tridge:
> The original concern about quadplanes violating the fly forward assumption when assisting and corrupting wind estimates turned out to be invalid because they weathervane as airspeed rises and the effect on dead reckoning navigation with assisting activating during forward flight was not foreseen.

I've seen tilt vector quadplanes like skywalker X8 slew around really badly in fwd transition, could this cause the yaw state to become corrupted?

tridge:
this also means if we stall in fixed wing mode then during recovery we will be setting fly forwad, is it OK to be lying to the EKF about our fly fwd state?

tridge:
same in FBWA mode at zero throttle, we would be saying we are in fly fwd, we are in assisted flight in that case, possibly with zero airspeed

peterbarker:
> Seem to be the same idea as #27917?

Ah yes, that's this PR :-)

Can you recall which one you meant to reference?

PR #23173, opened 2023-03-10
GPS: simple jamming simulator

tridge:
this is meant to complement #23010 to make it easier to test making ArduPilot more resilient to jamming

PR #27918, opened 2024-08-24
Correct Jamming simulator use of accuracies

peterbarker:
Replaces https://github.com/ArduPilot/ardupilot/pull/23173/files , which got forgotten and https://github.com/ArduPilot/ardupilot/pull/23173 was merged instead.

The merged PR didn't change the backends to use the information out of the supplied simulated GPS state object, so while position might move around due to jamming the accuracies did not.

Add a test that everything goes wonky.


tridge (approved):

rmackay9:
I've added the "WikiNeeded" label because we should really have an entry in our [SITL wiki here](https://ardupilot.org/dev/docs/using-sitl-for-ardupilot-testing.html) telling users that the feature exists and the basics of how to use it

PR #27921, opened 2024-08-24
ArduPlane: use Location::AltFrame for guided_state target frame references enabling ABOVE_TERRAIN

timtuxworth:
This PR changes the c++ code in Plane Guided mode to use Location:AltFrame for internal references to target Frame. 

This only applies for AP_PLANE_OFFBOARD_GUIDED_SLEW_ENABLED - effectively guided mode commands received via MAVLink, which could also be from Lua using the command_int Lua interface.

A very welcome side effect of this change is that it enables plane MAVLink CMD_GUIDED_CHANGE_ALTITUDE to work for  terrain altitudes in addition to the existing GLOBAL/Absolute and RELATIVE/Home altitudes.

Extracted from #27909 as requested by @IamPete1 to fix guided_state.target_alt_frame being a uint8_t whereas it actually stores AltFrame values from Location::AltFrame.

Similar to changes proposed for Sub in #27767 

Mavlink alt frame values (probably) needed for altitude navigation:

```
   MAV_FRAME_GLOBAL=0, /* Global (WGS84) coordinate frame + altitude relative to mean sea level (MSL). | */
   MAV_FRAME_GLOBAL_RELATIVE_ALT=3, /* 
          Global (WGS84) coordinate frame + altitude relative to the home position.
         | */
   MAV_FRAME_GLOBAL_TERRAIN_ALT=10, /* Global (WGS84) coordinate frame with AGL altitude (altitude at ground level). | */
```

whereas the internal Location:AltFrame values are 

```
    enum class AltFrame {
        ABSOLUTE = 0,
        ABOVE_HOME = 1,
        ABOVE_ORIGIN = 2,
        ABOVE_TERRAIN = 3
    };
```

So as you can see ABSOLUTE is co-incidentally the same (zero) but none of the other values match. Even worse MAV_FRAME_GLOBAL_RELATIVE_ALT=3 == AltFrame::ABOVE_TERRAIN=3.

So the MAVLink values need to be translated by GCS_Mavlink.cpp to values that can be used by Location and all of the other internal code that uses Location::ALtFrame.

### DIFF
--- a/ArduPlane/GCS_Mavlink.cpp
+++ b/ArduPlane/GCS_Mavlink.cpp
@@ -932,48 +932,25 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands(const mavl
             return MAV_RESULT_DENIED;
         }
 
-         // the requested alt data might be relative or absolute
-        float new_target_alt = packet.z * 100;
-        float new_target_alt_rel = packet.z * 100 + plane.home.alt;
-
-         // only global/relative/terrain frames are supported
-        switch(packet.frame) {
-            case MAV_FRAME_GLOBAL_RELATIVE_ALT: {
-                if   (is_equal(plane.guided_state.target_alt,new_target_alt_rel) ) { // compare two floats as near-enough
-                    // no need to process any new packet/s with the same ALT any further, if we are already doing it.
-                    return MAV_RESULT_ACCEPTED;
-                }
-                plane.guided_state.target_alt = new_target_alt_rel;
-                break;
-            }
-            case MAV_FRAME_GLOBAL: {
-                if   (is_equal(plane.guided_state.target_alt,new_target_alt) ) {  // compare two floats as near-enough
-                    // no need to process any new packet/s with the same ALT any further, if we are already doing it.
-                    return MAV_RESULT_ACCEPTED;
-                }
-                plane.guided_state.target_alt = new_target_alt;
-                break;
-            }
-            default:
-                //  MAV_RESULT_DENIED  means Command is invalid (is supported but has invalid parameters).
-                return MAV_RESULT_DENIED;
+        Location::AltFrame new_target_alt_frame;
+        if (!mavlink_coordinate_frame_to_location_alt_frame((MAV_FRAME)packet.frame, new_target_alt_frame)) {
+            return MAV_RESULT_DENIED;
         }
+        // keep a copy of what came in via MAVLink - this is needed for logging, but not for anything else
+        plane.guided_state.target_mav_frame = packet.frame;
 
-        plane.guided_state.target_alt_frame = packet.frame;
-        plane.guided_state.last_target_alt = plane.current_loc.alt; // FIXME: Reference frame is not corrected for here
+        const int32_t new_target_alt_cm = packet.z * 100;
+        plane.guided_state.target_location.set_alt_cm(new_target_alt_cm, new_target_alt_frame); 
         plane.guided_state.target_alt_time_ms = AP_HAL::millis();
 
+        // param3 contains the desired vertical velocity (not acceleration)
         if (is_zero(packet.param3)) {
-            // the user wanted /maximum acceleration, pick a large value as close enough
-            plane.guided_state.target_alt_accel = 1000.0;
+            // the user wanted /maximum altitude change rate, pick a large value as close enough
+            plane.guided_state.target_alt_rate = 1000.0;
         } else {
-            plane.guided_state.target_alt_accel = fabsf(packet.param3);
+            plane.guided_state.target_alt_rate = fabsf(packet.param3);
         }
 
-         // assign an acceleration direction
-        if (plane.guided_state.target_alt < plane.current_loc.alt) {
-            plane.guided_state.target_alt_accel *= -1.0f;
-        }
         return MAV_RESULT_ACCEPTED;
     }
 
--- a/ArduPlane/Log.cpp
+++ b/ArduPlane/Log.cpp
@@ -128,10 +128,11 @@ struct PACKED log_OFG_Guided {
     float target_airspeed_cm;
     float target_airspeed_accel;
     float target_alt;
-    float target_alt_accel;
-    uint8_t target_alt_frame;
+    float target_alt_rate;
+    uint8_t target_mav_frame;   // received MavLink frame
     float target_heading;
     float target_heading_limit;
+    uint8_t target_alt_frame;   // internal AltFrame
 };
 
 // Write a OFG Guided packet.
@@ -142,11 +143,12 @@ void Plane::Log_Write_OFG_Guided()
         time_us                : AP_HAL::micros64(),
         target_airspeed_cm     : (float)guided_state.target_airspeed_cm*(float)0.01,
         target_airspeed_accel  : guided_state.target_airspeed_accel,
-        target_alt             : guided_state.target_alt,
-        target_alt_accel       : guided_state.target_alt_accel,
-        target_alt_frame       : guided_state.target_alt_frame,
+        target_alt             : guided_state.target_location.alt * 0.01,
+        target_alt_rate        : guided_state.target_alt_rate,
+        target_mav_frame       : guided_state.target_mav_frame,
         target_heading         : guided_state.target_heading,
-        target_heading_limit   : guided_state.target_heading_accel_limit
+        target_heading_limit   : guided_state.target_heading_accel_limit,
+        target_alt_frame       : static_cast<uint8_t>(guided_state.target_location.get_alt_frame()),
     };
     logger.WriteBlock(&pkt, sizeof(pkt));
 }
@@ -274,7 +276,7 @@ void Plane::Log_Write_Guided(void)
         logger.Write_PID(LOG_PIDG_MSG, g2.guidedHeading.get_pid_info());
     }
 
-    if ( is_positive(guided_state.target_alt) || is_positive(guided_state.target_airspeed_cm) ) {
+    if ( guided_state.target_location.alt != -1 || is_positive(guided_state.target_airspeed_cm) ) {
         Log_Write_OFG_Guided();
     }
 #endif // AP_PLANE_OFFBOARD_GUIDED_SLEW_ENABLED
@@ -490,12 +492,13 @@ const struct LogStructure Plane::log_structure[] = {
 // @Field: Arsp:  target airspeed cm
 // @Field: ArspA:  target airspeed accel
 // @Field: Alt:  target alt
-// @Field: AltA: target alt accel
-// @Field: AltF: target alt frame
+// @Field: AltA: target alt velocity (rate of change)
+// @Field: AltF: target alt frame (MAVLink)
 // @Field: Hdg:  target heading
 // @Field: HdgA: target heading lim
+// @Field: AltL: target alt frame (Location)
     { LOG_OFG_MSG, sizeof(log_OFG_Guided),     
-      "OFG", "QffffBff",    "TimeUS,Arsp,ArspA,Alt,AltA,AltF,Hdg,HdgA", "s-------", "F-------" , true }, 
+      "OFG", "QffffBffB",    "TimeUS,Arsp,ArspA,Alt,AltA,AltF,Hdg,HdgA,AltL", "snnmo-d--", "F--------" , true }, 
 #endif
 };
 
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -577,11 +577,10 @@ private:
         uint32_t target_airspeed_time_ms;
 
         // altitude adjustments
-        float target_alt = -1;   // don't default to zero here, as zero is a valid alt.
-        uint32_t last_target_alt = 0;
-        float target_alt_accel;
+        Location target_location;
+        float target_alt_rate;
         uint32_t target_alt_time_ms = 0;
-        uint8_t target_alt_frame = 0;
+        uint8_t target_mav_frame = -1;
 
         // heading track
         float target_heading = -4; // don't default to zero or -1 here, as both are valid headings in radians
--- a/ArduPlane/mode.cpp
+++ b/ArduPlane/mode.cpp
@@ -58,9 +58,8 @@ bool Mode::enter()
     plane.guided_state.target_heading = -4; // radians here are in range -3.14 to 3.14, so a default value needs to be outside that range
     plane.guided_state.target_heading_type = GUIDED_HEADING_NONE;
     plane.guided_state.target_airspeed_cm = -1; // same as above, although an airspeed of -1 is rare on plane.
-    plane.guided_state.target_alt = -1; // same as above, although a target alt of -1 is rare on plane.
     plane.guided_state.target_alt_time_ms = 0;
-    plane.guided_state.last_target_alt = 0;
+    plane.guided_state.target_location.set_alt_cm(-1, Location::AltFrame::ABSOLUTE); 
 #endif
 
 #if AP_CAMERA_ENABLED
--- a/ArduPlane/mode_guided.cpp
+++ b/ArduPlane/mode_guided.cpp
@@ -129,24 +129,29 @@ void ModeGuided::set_radius_and_direction(const float radius, const bool directi
 void ModeGuided::update_target_altitude()
 {
 #if AP_PLANE_OFFBOARD_GUIDED_SLEW_ENABLED
-    if (((plane.guided_state.target_alt_time_ms != 0) || plane.guided_state.target_alt > -0.001 )) { // target_alt now defaults to -1, and _time_ms defaults to zero.
+    // target altitude can be negative (e.g. flying below home altitude from the top of a mountain)
+    if (((plane.guided_state.target_alt_time_ms != 0) || plane.guided_state.target_location.alt != -1 )) { // target_alt now defaults to -1, and _time_ms defaults to zero.
         // offboard altitude demanded
         uint32_t now = AP_HAL::millis();
         float delta = 1e-3f * (now - plane.guided_state.target_alt_time_ms);
         plane.guided_state.target_alt_time_ms = now;
         // determine delta accurately as a float
-        float delta_amt_f = delta * plane.guided_state.target_alt_accel;
+        float delta_amt_f = delta * plane.guided_state.target_alt_rate;
         // then scale x100 to match last_target_alt and convert to a signed int32_t as it may be negative
         int32_t delta_amt_i = (int32_t)(100.0 * delta_amt_f); 
-        Location temp {};
-        temp.alt = plane.guided_state.last_target_alt + delta_amt_i; // ...to avoid floats here, 
-        if (is_positive(plane.guided_state.target_alt_accel)) {
-            temp.alt = MIN(plane.guided_state.target_alt, temp.alt);
-        } else {
-            temp.alt = MAX(plane.guided_state.target_alt, temp.alt);
+        // To calculate the required velocity (up or down), we need to target and current altitudes in the target frame
+        const Location::AltFrame target_frame = plane.guided_state.target_location.get_alt_frame();
+        int32_t target_alt_previous_cm;
+        if (plane.current_loc.initialised() && plane.guided_state.target_location.initialised() && 
+            plane.current_loc.get_alt_cm(target_frame, target_alt_previous_cm)) {
+            // create a new interim target location that that takes current_location and moves delta_amt_i in the right direction
+            int32_t temp_alt_cm = constrain_int32(plane.guided_state.target_location.alt, target_alt_previous_cm - delta_amt_i,  target_alt_previous_cm + delta_amt_i);
+            Location temp_location = plane.guided_state.target_location;
+            temp_location.set_alt_cm(temp_alt_cm, target_frame);
+
+            // incrementally step the altitude towards the target            
+            plane.set_target_altitude_location(temp_location);
         }
-        plane.guided_state.last_target_alt = temp.alt;
-        plane.set_target_altitude_location(temp);
     } else 
 #endif // AP_PLANE_OFFBOARD_GUIDED_SLEW_ENABLED
         {
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -5533,6 +5533,7 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
 
     def MAV_CMD_GUIDED_CHANGE_ALTITUDE(self):
         '''test handling of MAV_CMD_GUIDED_CHANGE_ALTITUDE'''
+        self.start_subtest("set home relative altitude")
         self.takeoff(30, relative=True)
         self.change_mode('GUIDED')
         for alt in 50, 70:
@@ -5544,6 +5545,7 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
             self.wait_altitude(alt-1, alt+1, timeout=30, relative=True)
 
         # test for #24535
+        self.start_subtest("switch to loiter and resume guided maintains home relative altitude")
         self.change_mode('LOITER')
         self.delay_sim_time(5)
         self.change_mode('GUIDED')
@@ -5554,6 +5556,55 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
             timeout=30,
             relative=True,
         )
+        # test that this works if switching between RELATIVE (HOME) and GLOBAL(AMSL)
+        self.start_subtest("set global/AMSL altitude, switch to loiter and resume guided")
+        alt = 625
+        self.run_cmd_int(
+            mavutil.mavlink.MAV_CMD_GUIDED_CHANGE_ALTITUDE,
+            p7=alt,
+            frame=mavutil.mavlink.MAV_FRAME_GLOBAL,
+        )
+        self.wait_altitude(alt-3, alt+3, timeout=30, relative=False)
+        self.change_mode('LOITER')
+        self.delay_sim_time(5)
+        self.change_mode('GUIDED')
+        self.wait_altitude(
+            alt-5,  # NOTE: reuse of alt from above CHANGE_ALTITUDE
+            alt+5,
+            minimum_duration=10,
+            timeout=30,
+            relative=False,
+        )
+        # test that this works if switching between RELATIVE (HOME) and terrain
+        self.start_subtest("set terrain altitude, switch to loiter and resume guided")
+        self.change_mode('GUIDED')
+        alt = 100
+        self.run_cmd_int(
+            mavutil.mavlink.MAV_CMD_GUIDED_CHANGE_ALTITUDE,
+            p7=alt,
+            frame=mavutil.mavlink.MAV_FRAME_GLOBAL_TERRAIN_ALT,
+        )
+        self.wait_altitude(
+            alt-5,  # NOTE: reuse of alt from abovE
+            alt+5,  # use a 5m buffer as the plane needs to go up and down a bit to maintain terrain distance
+            minimum_duration=10,
+            timeout=30,
+            relative=False,
+            altitude_source="TERRAIN_REPORT.current_height"
+        )
+        self.change_mode('LOITER')
+        self.delay_sim_time(5)
+        self.change_mode('GUIDED')
+        self.wait_altitude(
+            alt-5,  # NOTE: reuse of alt from abovE
+            alt+5,  # use a 5m buffer as the plane needs to go up and down a bit to maintain terrain distance
+            minimum_duration=10,
+            timeout=30,
+            relative=False,
+            altitude_source="TERRAIN_REPORT.current_height"
+        )
+
+        self.delay_sim_time(5)
         self.fly_home_land_and_disarm()
 
     def _MAV_CMD_PREFLIGHT_CALIBRATION(self, command):
--- a/Tools/autotest/vehicle_test_suite.py
+++ b/Tools/autotest/vehicle_test_suite.py
@@ -7176,6 +7176,10 @@ class TestSuite(ABC):
                 altitude_source = "GLOBAL_POSITION_INT.relative_alt"
             else:
                 altitude_source = "GLOBAL_POSITION_INT.alt"
+        if altitude_source == "TERRAIN_REPORT.current_height":
+            terrain = self.assert_receive_message('TERRAIN_REPORT')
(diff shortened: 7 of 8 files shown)


ArduPlane/Plane.h
```diff
@@ -563,9 +563,10 @@ class Plane : public AP_Vehicle {
         // altitude adjustments
         float target_alt = -1;   // don't default to zero here, as zero is a valid alt.
         uint32_t last_target_alt = 0;
+        Location::AltFrame last_target_alt_frame = Location::AltFrame::ABSOLUTE;
```
IamPete1, line 566: This is unused.

IamPete1 (commented):
This has broken CI. It also changes the logged value from being the MAVLink enum to the AP one.

timtuxworth:
> This has broken CI. It also changes the logged value from being the MAVLink enum to the AP one.

I think that is a bug. Why would plane.guided_state internally use the MAVLink enum for the target_alt_frame associated with the internal target_alt value it sends to TECS?

ArduPlane/Plane.h
timtuxworth: In this PR yes, but when it's used in my other PR this was extracted from.

ArduPlane/mode_guided.cpp
```diff
@@ -140,6 +140,9 @@ void ModeGuided::update_target_altitude()
         int32_t delta_amt_i = (int32_t)(100.0 * delta_amt_f); 
         Location temp {};
         temp.alt = plane.guided_state.last_target_alt + delta_amt_i; // ...to avoid floats here, 
+        if(plane.guided_state.target_alt_frame == Location::AltFrame::ABOVE_HOME) {
+            temp.relative_alt = true;
+        }
```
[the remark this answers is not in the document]
timtuxworth: Which is why it never would have worked. MAVLink was passing in the (MAVLink enum) Frame value which was never being used except for logging and was the wrong value to drive AP internal altitude frames. This fixes that.

timtuxworth:
> This has broken CI. It also changes the logged value from being the MAVLink enum to the AP one.

Is there an easy way to find out which test failed?

timtuxworth:
> This has broken CI. It also changes the logged value from being the MAVLink enum to the AP one.

I'm wondering if the test is valid? I think that switching out of GUIDED to LOITER and back to GUIDED should reset the plane.guided_state.target_alt back to -1, so I'm not sure why the test is expecting the altitude to be 70 +/-3 - can you explain that to me?

IamPete1:
> I'm wondering if the test is valid? I think that switching out of GUIDED to LOITER and back to GUIDED should reset the plane.guided_state.target_alt back to -1, so I'm not sure why the test is expecting the altitude to be 70 +/-3 - can you explain that to me?

It was circling at 70m in guided. Changing to loiter should maintain that 70m altitude. Then changing back to guided again should still maintain 70m altitude. Until a command is received it should stay at the altitude at which the mode was entered.

ArduPlane/Plane.h
IamPete1: > Should I move this back over there?

Yes please.

ArduPlane/mode_guided.cpp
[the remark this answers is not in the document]
IamPete1: I think you just need to delete this new if and you will pass CI.

ArduPlane/Plane.h
timtuxworth: Turns out I do need to use it to fix the CI fail

ArduPlane/mode_guided.cpp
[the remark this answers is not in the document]
timtuxworth: True - but that is not actually correct. I think the real fix is that ModeGuiided::update_target_location() which is using last_target_alt should also use last_target_alt_frame.

timtuxworth:
> > I'm wondering if the test is valid? I think that switching out of GUIDED to LOITER and back to GUIDED should reset the plane.guided_state.target_alt back to -1, so I'm not sure why the test is expecting the altitude to be 70 +/-3 - can you explain that to me?
> 
> It was circling at 70m in guided. Changing to loiter should maintain that 70m altitude. Then changing back to guided again should still maintain 70m altitude. Until a command is received it should stay at the altitude at which the mode was entered.

Fixed - see my change to mode_guided.cpp

IamPete1:
> Fixed - see my new change to mode_guided.cpp

I don't understand why we need to fix something that wasn't broken? Maybe I'm missing some case where we get it wrong currently?

timtuxworth:
> > Fixed - see my new change to mode_guided.cpp
> 
> I don't understand why we need to fix something that wasn't broken? Maybe I'm missing some case where we get it wrong currently?

I think it was broken, which caused the CI fail. It was using last_target_alt without a frame. The way I see it, it was purely coincidence that the sequence of events in the CI test would pass this test.

timtuxworth:
> It also changes the logged value from being the MAVLink enum to the AP one.

How do you think I should deal with this? I would argue it's a bug that it had been logging the MAVLink enum since that's not what the code is using, so it was probably not very useful.

ArduPlane/mode_guided.cpp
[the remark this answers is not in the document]
timtuxworth: Now implemented - and passes CI

amilcarlucas:
Are there static asserts that the AP internal frame defenition enums have the same values as the mavlink external frame defenition enums?

timtuxworth:
> Are there static asserts that the AP internal frame defenition enums have the same values as the mavlink external frame defenition enums?

They don't have the same values. That's the fundamental problem. See the first post on this PR, I lay out the mismatch.

clydemcqueen:
`GCS_MAVLINK::mavlink_coordinate_frame_to_location_alt_frame` is handy, and used by Copter and Sub.

https://github.com/ArduPilot/ardupilot/blob/1439aebf94ae388a82f28199aa7dd333c009b98c/libraries/GCS_MAVLink/GCS_Common.cpp#L6880

timtuxworth:
> `GCS_MAVLINK::mavlink_coordinate_frame_to_location_alt_frame` is handy, and used by Copter and Sub.
> 
> https://github.com/ArduPilot/ardupilot/blob/1439aebf94ae388a82f28199aa7dd333c009b98c/libraries/GCS_MAVLink/GCS_Common.cpp#L6880

Oh that's awesome, I didn't know that existed!

Tools/autotest/arduplane.py
```diff
@@ -5529,6 +5531,26 @@ def MAV_CMD_GUIDED_CHANGE_ALTITUDE(self):
...
+        self.start_subtest("set global/AMSL altitude switch to loiter and resume")
+        alt = 600
+        self.run_cmd_int(
+            mavutil.mavlink.MAV_CMD_GUIDED_CHANGE_ALTITUDE,
+            p7=alt,
+            frame=mavutil.mavlink.MAV_FRAME_GLOBAL,
+        )
+        self.wait_altitude(alt-3, alt+3, timeout=30, relative=False)
+        self.change_mode('LOITER')
+        self.delay_sim_time(5)
```
peterbarker, line 5544: Shouldn't there be a `wait_altitude` in here?

Tools/autotest/arduplane.py
```diff
@@ -5529,6 +5531,26 @@ def MAV_CMD_GUIDED_CHANGE_ALTITUDE(self):
...
+        self.change_mode('GUIDED')
+        self.wait_altitude(
+            alt-5,  # NOTE: reuse of alt from above CHANGE_ALTITUDE
+            alt+5,
+            minimum_duration=10,
+            timeout=30,
+            relative=False,
+        )
+        self.delay_sim_time(5)
```
peterbarker, line 5553: Why's this one in here?  Seems to be redundant after the `wait_altitude`
```suggestion
```

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -929,35 +929,33 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands(const mavl
...
-                if   (is_equal(plane.guided_state.target_alt,new_target_alt_rel) ) { // compare two floats as near-enough
-                    // no need to process any new packet/s with the same ALT any further, if we are already doing it.
-                    return MAV_RESULT_ACCEPTED;
-                }
-                plane.guided_state.target_alt = new_target_alt_rel;
-                break;
+        if (mavlink_coordinate_frame_to_location_alt_frame((MAV_FRAME)packet.frame, new_target_alt_frame)) {
+
+             // the requested alt data might be relative (to home)
+            if (new_target_alt_frame == Location::AltFrame::ABOVE_HOME) {
```
peterbarker, line 938: This should definitely still be a `switch` statement.

We should be rejecting packets if the new frame isn't one of ABOVE_HOME or ABSOLUTE

ArduPlane/Plane.h
```diff
@@ -563,9 +563,11 @@ class Plane : public AP_Vehicle {
...
         float target_alt_accel;
         uint32_t target_alt_time_ms = 0;
-        uint8_t target_alt_frame = 0;
+        Location::AltFrame target_alt_frame = Location::AltFrame::ABSOLUTE;
+        uint8_t target_mav_frame = -1;
```
peterbarker, line 570: Just leave at zero?

peterbarker (changes requested):

ArduPlane/GCS_Mavlink.cpp
timtuxworth: Interestingly, using mavlink_coordinate_frame_to_location_alt_frame() probably makes it work with ABOVE_TERRAIN "for free". Hence the changes I'm trying to make to the autotest to test that.

ArduPlane/Plane.h
timtuxworth: because the type has changed.

Tools/autotest/arduplane.py
timtuxworth: I've added an altitude_source = "TERRAIN_REPORT.current_height" to wait_altitude. It's a big clunky but it does work.

timtuxworth:
Thanks for looking at this @peterbarker - I've just make a whole bunch of changes which cleans it up a lot. Thanks for your questions/suggestions they lead to a lot of improvements. I'm pretty sure I fixed a couple of bugs in the original code.
@IamPete1 it would be great if you could take another look. I've now added autotests for ABOVE_TERRAIN which works quite nicely.

IamPete1:
This seems to now add support for a new alt frame? Wasn't that going to be in https://github.com/ArduPilot/ardupilot/pull/27909?

timtuxworth:
> This seems to now add support for a new alt frame? Wasn't that going to be in #27909?

It was, but after adding mavlink_coordinate_frame_to_location_alt_frame() there doesn't seem much point. This gives us ABOVE_TERRAIN and I'd actually have to add code to remove it again.

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -967,10 +959,14 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands(const mavl
...
-         // assign an acceleration direction
-        if (plane.guided_state.target_alt < plane.current_loc.alt) {
+         // assign an acceleration direction - this only works if the current and target frame are the same
+         // this should not be happening here - it should be in ModeGuided::update_target_altitude()
+        if (plane.guided_state.last_target_alt_frame == plane.guided_state.target_alt_frame && plane.guided_state.target_alt < plane.current_loc.alt) {
             plane.guided_state.target_alt_accel *= -1.0f;
         }
+        //if ((MAV_FRAME)packet.frame == MAV_FRAME_GLOBAL_TERRAIN_ALT) {
+        //    return MAV_RESULT_ACCEPTED;
+        // }
```
IamPete1, line 969: Remove commented code.

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -929,35 +929,27 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands(const mavl
...
+
+        if (!mavlink_coordinate_frame_to_location_alt_frame((MAV_FRAME)packet.frame, new_target_alt_frame)) {
+            return MAV_RESULT_DENIED;
+        }
+
+        // compare two floats as near-enough
+        const float new_target_alt = packet.z * 100;
+        if (is_equal(plane.guided_state.target_alt,new_target_alt) && plane.guided_state.target_alt_frame == new_target_alt_frame) {
+            // no need to process any new packet/s with the same ALT any further, if we are already doing it.
+            return MAV_RESULT_ACCEPTED;
```
IamPete1, line 942: I know this stuff is there already, but I would be tempted to just remove the early return. I'm not sure what its really getting us.

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -967,10 +959,14 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands(const mavl
             plane.guided_state.target_alt_accel = fabsf(packet.param3);
         }
 
-         // assign an acceleration direction
-        if (plane.guided_state.target_alt < plane.current_loc.alt) {
+         // assign an acceleration direction - this only works if the current and target frame are the same
+         // this should not be happening here - it should be in ModeGuided::update_target_altitude()
```
IamPete1, line 963: I agree, this is such a big rework now you might as well just fix it there. It also means it works with any alt frame and you don't need `last_target_alt_frame` anymore.

ArduPlane/mode_guided.cpp
```diff
@@ -139,13 +139,17 @@ void ModeGuided::update_target_altitude()
         // then scale x100 to match last_target_alt and convert to a signed int32_t as it may be negative
         int32_t delta_amt_i = (int32_t)(100.0 * delta_amt_f); 
         Location temp {};
-        temp.alt = plane.guided_state.last_target_alt + delta_amt_i; // ...to avoid floats here, 
+        int32_t temp_alt = plane.guided_state.last_target_alt + delta_amt_i;
```
IamPete1, line 142:
```suggestion
        int32_t temp_alt = constrain_int32(plane.guided_state.target_alt, plane.guided_state.last_target_alt - delta_amt_i,  plane.guided_state.last_target_alt + delta_amt_i);
```
You can use the constrain trick so we don't need to know the sign of the accel.

IamPete1: That removes the if and the else.

ArduPlane/mode_guided.cpp
```diff
@@ -139,13 +139,17 @@ void ModeGuided::update_target_altitude()
...
+        // Restore the last altitude that the plane had been flying
         if (is_positive(plane.guided_state.target_alt_accel)) {
-            temp.alt = MIN(plane.guided_state.target_alt, temp.alt);
+            temp_alt = MIN(plane.guided_state.target_alt, temp_alt);
         } else {
-            temp.alt = MAX(plane.guided_state.target_alt, temp.alt);
+            temp_alt = MAX(plane.guided_state.target_alt, temp_alt);
         }
+        temp.set_alt_cm(temp_alt, plane.guided_state.target_alt_frame);
```
IamPete1, line 149: This does not work correctly because the location is not initialized so it is not valid in terrain alt frame.

ArduPlane/mode_guided.cpp
IamPete1: Infact, this whole thing is velocity limiting not accel limiting. The command calls for a accel limit. There is either a error on the MAVLink spec or in the code.

ArduPlane/mode_guided.cpp
IamPete1: you need to store the last target location, not just its alt.

ArduPlane/Log.cpp
```diff
@@ -491,11 +493,12 @@ const struct LogStructure Plane::log_structure[] = {
 // @Field: ArspA:  target airspeed accel
 // @Field: Alt:  target alt
 // @Field: AltA: target alt accel
-// @Field: AltF: target alt frame
+// @Field: AltF: target alt frame (MAVLink)
 // @Field: Hdg:  target heading
 // @Field: HdgA: target heading lim
+// @Field: AltL: target alt frame (Location)
```
IamPete1, line 499: I don't see the benefit here, were logging the same information twice?

IamPete1 (commented):
Quite a bit of the existing code looks suspect to me.....

ArduPlane/GCS_Mavlink.cpp
timtuxworth: Done

ArduPlane/GCS_Mavlink.cpp
timtuxworth: I don't think that's a good idea. Removing it would set last_target_alt and last_target_alt_frame even if they are the same values as the incoming values. Leaving this in means those values are tell us the last actually different target values.

ArduPlane/mode_guided.cpp
timtuxworth: Would it not be better to compare the new target location to the **current location** rather than the last target to determine the velocity required

ArduPlane/Log.cpp
timtuxworth: I want to know what came in via MAVLink _and_ what the result was (AltFrame)

ArduPlane/Log.cpp
```diff
@@ -142,11 +143,12 @@ void Plane::Log_Write_OFG_Guided()
...
         target_airspeed_accel  : guided_state.target_airspeed_accel,
-        target_alt             : guided_state.target_alt,
+        target_alt             : guided_state.target_location.alt * 0.01,
         target_alt_accel       : guided_state.target_alt_accel,
-        target_alt_frame       : guided_state.target_alt_frame,
+        target_mav_frame       : guided_state.target_mav_frame,
         target_heading         : guided_state.target_heading,
-        target_heading_limit   : guided_state.target_heading_accel_limit
+        target_heading_limit   : guided_state.target_heading_accel_limit,
+        target_alt_frame       : static_cast<uint8_t>(guided_state.target_location.get_alt_frame()),
```
peterbarker, line 151: rearrangment of field order is a little unfortunate here

Tools/autotest/vehicle_test_suite.py
```diff
@@ -7282,11 +7282,18 @@ def get_altitude(self, relative=False, timeout=30, altitude_source=None):
                 altitude_source = "GLOBAL_POSITION_INT.relative_alt"
             else:
                 altitude_source = "GLOBAL_POSITION_INT.alt"
+        if altitude_source == "TERRAIN_REPORT.current_height":
+            terrain = self.assert_receive_message('TERRAIN_REPORT')
+            if terrain is None:
+                raise NotAchievedException("Did not get TERRAIN_REPORT message")
```
peterbarker, line 7288:
```suggestion
```

You've asserted you've received it :-)

Tools/autotest/vehicle_test_suite.py
```diff
@@ -7282,11 +7282,18 @@ def get_altitude(self, relative=False, timeout=30, altitude_source=None):
...
+            return terrain.current_height
+
         (msg, field) = altitude_source.split('.')
         msg = self.poll_message(msg, quiet=True)
         divisor = 1000.0  # mm is pretty common in mavlink
         if altitude_source == "SIM_STATE.alt":
             divisor = 1.0
+
```
peterbarker, line 7296: Bad whitespace change

Tools/autotest/arduplane.py
```diff
@@ -2942,8 +2942,6 @@ def TerrainLoiter(self):
             gpi = self.assert_receive_message('GLOBAL_POSITION_INT')
             terrain = self.assert_receive_message('TERRAIN_REPORT')
             rel_alt = terrain.current_height
-            self.progress("%um above terrain (%um bove home)" %
-                          (rel_alt, gpi.relative_alt/1000.0))
```
peterbarker, line 2946: Why is this being removed?

Too noisy?

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -929,48 +929,32 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands(const mavl
...
         }
 
-        plane.guided_state.target_alt_frame = packet.frame;
-        plane.guided_state.last_target_alt = plane.current_loc.alt; // FIXME: Reference frame is not corrected for here
+        const int32_t new_target_alt = packet.z * 100;
+        if (is_equal(plane.guided_state.target_location.alt,new_target_alt) && plane.guided_state.target_location.get_alt_frame() == new_target_alt_frame) {
+            // no need to process any new packet/s with the same ALT any further, if we are already doing it.
+            // This is quite important as last_target_alt and last_target_alt_frame should only be set if the new one is different
+            return MAV_RESULT_ACCEPTED;
+        }
```
peterbarker, line 942:
```suggestion
```

peterbarker (changes requested):

libraries/AP_Scripting/examples/plane_guided_terrain.lua
```diff
@@ -0,0 +1,229 @@
...
+
+gcs:send_text(MAV_SEVERITY.NOTICE, string.format("%s %s script loaded", SCRIPT_NAME, SCRIPT_VERSION) )
+
+-- start running update loop
+if FWVersion:type() == 3 then
+    gcs:send_text(MAV_SEVERITY.NOTICE, string.format("%s %s script loaded", SCRIPT_NAME, SCRIPT_VERSION) )
+    return protected_wrapper()
+else
+    gcs:send_text(MAV_SEVERITY.NOTICE,string.format("%s: Must run on Plane", SCRIPT_NAME_SHORT))
+end
```
Ryanf55, line 229:
```suggestion
end

```

tridge (changes requested):

Tools/autotest/arduplane.py
```diff
@@ -5533,6 +5531,7 @@ def RunMissionScript(self):
 
     def MAV_CMD_GUIDED_CHANGE_ALTITUDE(self):
         '''test handling of MAV_CMD_GUIDED_CHANGE_ALTITUDE'''
+        self.start_subtest("set home relative altitude")
         self.takeoff(30, relative=True)
         self.change_mode('GUIDED')
         for alt in 50, 70:
```
Ryanf55, line 5537: Can we also test negative MSL altitudes in death valley? You mentioned it didn't work before, so now we can prove it does work.

ArduPlane/mode.cpp
```diff
@@ -58,9 +58,11 @@ bool Mode::enter()
     plane.guided_state.target_heading = -4; // radians here are in range -3.14 to 3.14, so a default value needs to be outside that range
     plane.guided_state.target_heading_type = GUIDED_HEADING_NONE;
     plane.guided_state.target_airspeed_cm = -1; // same as above, although an airspeed of -1 is rare on plane.
-    plane.guided_state.target_alt = -1; // same as above, although a target alt of -1 is rare on plane.
+    //plane.guided_state.target_alt = -1; // same as above, although a target alt of -1 is rare on plane.
+    //plane.guided_state.target_alt_frame = Location::AltFrame::ABSOLUTE;
```
IamPete1, line 62: Remove commented code.

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -929,48 +929,32 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands(const mavl
...
+        // keep a copy of what came in via MAVLink - this is needed for logging, but not for anything else
+        plane.guided_state.target_mav_frame = packet.frame;
+        plane.guided_state.target_location.set_alt_cm(new_target_alt, new_target_alt_frame); 
+
         plane.guided_state.target_alt_time_ms = AP_HAL::millis();
 
+        // param3 contains the desired vertical velocity (not acceleration)
         if (is_zero(packet.param3)) {
             // the user wanted /maximum acceleration, pick a large value as close enough
             plane.guided_state.target_alt_accel = 1000.0;
```
IamPete1, line 953: rename to `target_alt_rate`

ArduPlane/Log.cpp
timtuxworth: It's not a reordering in terms of the values that end up in the log. The 6th field used to be called target_alt_frame but it contained the MAVLINK frame, so now the 6th field still contains the mavlink frame, the same value at the same position in the log and still called AltF - but it's source is a different value target_mav_frame.

Tools/autotest/vehicle_test_suite.py
timtuxworth: Thanks

Tools/autotest/arduplane.py
timtuxworth: I thought I had put it in myself. I was trying to remove debug code I had added. I'll put it back.

ArduPlane/GCS_Mavlink.cpp
timtuxworth: gone

ArduPlane/Log.cpp
```diff
@@ -129,9 +129,10 @@ struct PACKED log_OFG_Guided {
     float target_airspeed_accel;
     float target_alt;
     float target_alt_accel;
-    uint8_t target_alt_frame;
+    uint8_t target_mav_frame;   // received MavLink frame
```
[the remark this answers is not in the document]
timtuxworth: Yes but it's a MAV_FRAME value as defined in the mavlink common.xml

ArduPlane/GCS_Mavlink.cpp
timtuxworth: Removed

ArduPlane/mode_guided.cpp
timtuxworth: Thanks -done.

Tools/autotest/arduplane.py
timtuxworth: Reverted

libraries/AP_Scripting/examples/plane_guided_terrain.lua
timtuxworth: Thanks - fixed.

ArduPlane/mode.cpp
timtuxworth: Done

Tools/autotest/arduplane.py
peterbarker: Note quite; we print the final value just before dying so you know what the bad values were

ArduPlane/mode_guided.cpp
```diff
@@ -138,15 +139,18 @@ void ModeGuided::update_target_altitude()
...
-        if (is_positive(plane.guided_state.target_alt_accel)) {
-            temp.alt = MIN(plane.guided_state.target_alt, temp.alt);
-        } else {
-            temp.alt = MAX(plane.guided_state.target_alt, temp.alt);
+        // To calculate the required velocity (up or down), we need to target and previous altitudes in the target frame
+        int32_t target_alt_previous;
+        if (plane.current_loc.initialised() && plane.guided_state.target_location.initialised() && 
+            plane.current_loc.get_alt_cm(plane.guided_state.target_location.get_alt_frame(), target_alt_previous)) {
+            // create a new interim target location that that takes current_location and moves delta_amt_i in the right direction
+            int32_t temp_alt = constrain_int32(plane.guided_state.target_location.alt, target_alt_previous - delta_amt_i,  target_alt_previous + delta_amt_i);
```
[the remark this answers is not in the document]
timtuxworth: done

ArduPlane/mode_guided.cpp
timtuxworth: Done

Tools/autotest/arduplane.py
timtuxworth: Got it this time I think.

ArduPlane/mode_guided.cpp
```diff
@@ -129,24 +129,29 @@ void ModeGuided::set_radius_and_direction(const float radius, const bool directi
...
-            temp.alt = MAX(plane.guided_state.target_alt, temp.alt);
+        // To calculate the required velocity (up or down), we need to target and current altitudes in the target frame
+        const Location::AltFrame target_frame = plane.guided_state.target_location.get_alt_frame();
+        int32_t target_alt_previous_cm;
+        if (plane.current_loc.initialised() && plane.guided_state.target_location.initialised() && 
+            plane.current_loc.get_alt_cm(target_frame, target_alt_previous_cm)) {
+            // create a new interim target location that that takes current_location and moves delta_amt_i in the right direction
+            int32_t temp_alt_cm = constrain_int32(plane.guided_state.target_location.alt, target_alt_previous_cm - delta_amt_i,  target_alt_previous_cm + delta_amt_i);
+            Location temp_location = plane.guided_state.target_location;
+            temp_location.set_alt_cm(temp_alt_cm, target_frame);
```
timtuxworth, line 150: Needed to be set_alt_cm() to take the target_frame.

timtuxworth:
Tested on a real plane and found one bug in the c++ code and the lua example needed some major rework.

timtuxworth:
I think this is "ready" - what do you think @IamPete1 ?

Tools/autotest/arduplane.py
timtuxworth: I haven't tested this, follow up?

timtuxworth:
Rebased

timtuxworth:
Only 56 bytes on plane!

```
antennatracker   0 (0.0000%)    0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      621548
arducopter       0 (0.0000%)    0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      151208
ardusub          0 (0.0000%)    0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      360236
blimp            0 (0.0000%)    0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      599496
arduplane        56 (+0.0031%)  0 (0.0000%)  0 (0.0000%)  56 (+0.0031%)                                    153736
ardurover        0 (0.0000%)    0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      302576
arducopter-heli  0 (0.0000%)    0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      152168
```

tridge (approved):

PR #27923, opened 2024-08-25
AP_Beacon: allow hwdefs to override number of beacons

peterbarker:
... also a few other variables.

Also move these out of the main header file as one is internal and the others should go in the ocnfig. in case the values are used elsewhere


rmackay9 (approved):
Looks OK, it's been tested of course?  Slightly hard to imagine how it could go wrong but..

peterbarker:
> Looks OK, it's been tested of course? Slightly hard to imagine how it could go wrong but..

I've tested that the defines take effect.

I'm *not* saying that numbers other that 4 actually work :-)

I'm extracting this from a larger body of work to make that merge more manageable.

PR #27926, opened 2024-08-26
Param for Max Retrial of Engine Cranking.

loki077:
As we use a common battery for Avionics and Engine Cranking we would like to have an option to limit the number of retrials so that it doesn't drain that battery completely. Obvious choice is to separate the batteries but I still think there could be some use of this feature.

My Implementation is by adding Param MAX_RETRY which If set 0 then there is no limit to retrials. If set to a value greater than 0 then the engine will retry starting the engine this many times before giving up. Also the counter to track retrials resets to zero on disarm (could be done only on power cycle rather than disarm). If needed to be override can be done by setting the param to 0.

I would like to get more ideas on the method of its implementation. I have not yest tested this code.

Other ways to implement this is by just keep increasing the time between retrials.

### DIFF
--- a/Tools/autotest/quadplane.py
+++ b/Tools/autotest/quadplane.py
@@ -1116,6 +1116,53 @@ class AutoTestQuadPlane(vehicle_test_suite.TestSuite):
         self.change_mode('FBWA')
         self.wait_servo_channel_value(3, 2000)
 
+        self.start_subtest("Testing automatic restart")
+        # Limit start attempts to 4
+        max_tries = 4
+        self.set_parameter("ICE_STRT_MX_RTRY", max_tries)
+        # Make the engine unable to run (by messing up the RPM sensor)
+        rpm_chan = self.get_parameter("ICE_RPM_CHAN")
+        self.set_parameter("ICE_RPM_CHAN", 120) # Set to a non-existent sensor
+        self.set_rc(rc_engine_start_chan, 2000)
+        self.wait_statustext("Uncommanded engine stop")
+        self.wait_statustext("Starting engine")
+        # Restore the engine
+        self.set_parameter("ICE_RPM_CHAN", rpm_chan)
+        # Make sure the engine continues to run for the next 30 seconds
+        try:
+            self.wait_statustext("Uncommanded engine stop", timeout=30)
+            # The desired result is for the wait_statustext raise AutoTestTimeoutException
+            raise NotAchievedException("Engine stopped unexpectedly")
+        except AutoTestTimeoutException:
+            pass
+        self.context_stop_collecting("STATUSTEXT")
+
+        self.start_subtest("Testing automatic starter attempt limit")
+        # Try this test twice.
+        # For the first run, since the engine has been running successfully in
+        # the previous test for 30 seconds, the limit should reset. For the
+        # second run, after commanding an engine stop, the limit should reset.
+        for i in range(2):
+            self.context_collect("STATUSTEXT")
+            self.set_parameter("ICE_RPM_CHAN", 120) # Set to a non-existent sensor
+            self.set_rc(rc_engine_start_chan, 2000)
+            self.wait_statustext("Engine max crank attempts reached", check_context=True, timeout=30)
+            self.delay_sim_time(30) # wait for another 30 seconds to make sure the engine doesn't restart
+            messages = self.context_get().collections["STATUSTEXT"]
+            self.context_stop_collecting("STATUSTEXT")
+            # check for the exact number of starter attempts
+            attempts = 0
+            for m in messages:
+                if "Starting engine" == m.text:
+                    attempts += 1
+            if attempts != max_tries:
+                raise NotAchievedException(f"Run {i+1}: Expected {max_tries} attempts, got {attempts}")
+            # Command an engine stop
+            self.context_collect("STATUSTEXT")
+            self.set_rc(rc_engine_start_chan, 1000)
+            self.wait_statustext("ignition:0", check_context=True)
+            self.context_stop_collecting("STATUSTEXT")
+
     def ICEngineMission(self):
         '''Test ICE Engine Mission support'''
         rc_engine_start_chan = 11
--- a/libraries/AP_ICEngine/AP_ICEngine.cpp
+++ b/libraries/AP_ICEngine/AP_ICEngine.cpp
@@ -181,6 +181,13 @@ const AP_Param::GroupInfo AP_ICEngine::var_info[] = {
     // This allows one time conversion while allowing user to flash between versions with and without converted params
     AP_GROUPINFO_FLAGS("FMT_VER", 19, AP_ICEngine, param_format_version, 0, AP_PARAM_FLAG_HIDDEN),
 
+    // @Param: STRT_MX_RTRY
+    // @DisplayName: Maximum number of retries
+    // @Description: If set 0 then there is no limit to retrials. If set to a value greater than 0 then the engine will retry starting the engine this many times before giving up.
+    // @User: Standard
+    // @Range: 0 127
+    AP_GROUPINFO("STRT_MX_RTRY", 20, AP_ICEngine, max_crank_retry, 0),
+
     AP_GROUPEND
 };
 
@@ -379,6 +386,10 @@ void AP_ICEngine::update(void)
         if (should_run) {
             state = ICE_START_DELAY;
         }
+        crank_retry_ct = 0;
+        // clear the last uncommanded stop time, we only care about tracking
+        // the last one since the engine was started
+        last_uncommanded_stop_ms = 0;
         break;
 
     case ICE_START_HEIGHT_DELAY: {
@@ -402,8 +413,16 @@ void AP_ICEngine::update(void)
         if (!should_run) {
             state = ICE_OFF;
         } else if (now - starter_last_run_ms >= starter_delay*1000) {
-            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Starting engine");
-            state = ICE_STARTING;
+            // check if we should retry starting the engine
+            if (max_crank_retry <= 0 || crank_retry_ct < max_crank_retry) {
+                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Starting engine");
+                state = ICE_STARTING;
+                crank_retry_ct++;
+            } else {
+                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Engine max crank attempts reached");
+                // Mark the last run now so we don't send this message every loop
+                starter_last_run_ms = now;
+            }
         }
         break;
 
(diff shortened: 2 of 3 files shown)


loki077:
@peterbarker @tridge FYI

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -167,6 +167,12 @@ const AP_Param::GroupInfo AP_ICEngine::var_info[] = {
 
     // 18 was IGNITION_RLY
 
+    // @Param: MAX_RETRY
+    // @DisplayName: Maximum number of retries
+    // @Description: If set 0 then there is no limit to retrials. If set to a value greater than 0 then the engine will retry starting the engine this many times before giving up.
+    // @User: Standard
+    // @Range: 0 254
```
robertlong13, line 174: Underlying type as Int8, not UInt8. Should change this to a max of 127 or should change to uint8

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -360,6 +372,8 @@ void AP_ICEngine::update(void)
             // force ignition off when disarmed
             state = ICE_OFF;
         }
+        // reset the crank retry counter when disarmed
+        crank_retry_ct = 0;
```
robertlong13, line 376: Should probably also reset the counter when we shut the ignition off. This way, someone can shut the engine off/on to reset if they want to force it to retry again.

libraries/AP_ICEngine/AP_ICEngine.cpp
loki077: will fix it.

libraries/AP_ICEngine/AP_ICEngine.cpp
loki077: I am yet not sure of the approach of counter being reset is correct way so waiting for few more feedback. But rather then resetting at ignition off maybe we can reset it at Mavlink command of DO_ENGINE_CONTROL.
I will have a look at it.

peterbarker: Agreed with a reset if the user moves away from the auto-restart position and back into it

peterbarker (commented):
Needs an autotest

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -397,8 +405,14 @@ void AP_ICEngine::update(void)
         if (!should_run) {
             state = ICE_OFF;
         } else if (now - starter_last_run_ms >= starter_delay*1000) {
-            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Starting engine");
-            state = ICE_STARTING;
+            // check if we should retry starting the engine
+            if (max_crank_retry >= 0 && (max_crank_retry == 0 || crank_retry_ct < max_crank_retry)) {
```
robertlong13, line 409: Anything less than zero should be considered infinite retry. As written, -1 means you can't start the engine ever.

Should be
`if (max_crank_retry > 0 && crank_retry_ct < max_crank_retry) {`

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -422,9 +436,9 @@ void AP_ICEngine::update(void)
             float rpm_value;
             if (!AP::rpm()->get_rpm(rpm_instance-1, rpm_value) ||
                 rpm_value < rpm_threshold) {
-                // engine has stopped when it should be running
-                state = ICE_START_DELAY;
-                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Uncommanded engine stop");
+                    // engine has stopped when it should be running
+                    state = ICE_START_DELAY;
+                    GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Uncommanded engine stop");
             }
```
robertlong13, line 442: Whitespace change outside of scope of this PR, but also fixed incorrectly. For our astyle rules, it should actually be:

```
            if (!AP::rpm()->get_rpm(rpm_instance-1, rpm_value) ||
                    rpm_value < rpm_threshold) {
                // engine has stopped when it should be running
                state = ICE_START_DELAY;
                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Uncommanded engine stop");
            }
  ```
  
  i.e., adding an extra indent to line 438, not 439-441

**Review by robertlong13 — CHANGES_REQUESTED** [review]

File: libraries/AP_ICEngine/AP_ICEngine.cpp (hunk 3 above)
Comment (loki077) [reply libraries/AP_ICEngine/AP_ICEngine.cpp:409]: if (max_crank_retry <= 0 || crank_retry_ct < max_crank_retry) 
   
   this will be a fix as we want it to try infinitely when it is 0 or less, or less then maxcount.

File: libraries/AP_ICEngine/AP_ICEngine.cpp (hunk 4 above)
Comment (loki077) [reply libraries/AP_ICEngine/AP_ICEngine.cpp:442]: gitkraken view messed this up for me.....  thanks for pointing it,,,

File: libraries/AP_ICEngine/AP_ICEngine.cpp [hunk 5]
```diff
@@ -402,8 +410,15 @@ void AP_ICEngine::update(void)
...
         } else if (now - starter_last_run_ms >= starter_delay*1000) {
-            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Starting engine");
-            state = ICE_STARTING;
+            // check if we should retry starting the engine
+            if (max_crank_retry <= 0 || crank_retry_ct < max_crank_retry) {
+                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Starting engine");
+                state = ICE_STARTING;
+                crank_retry_ct++;
+            } else {                    
+                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Engine max crank attempts reached");
```
IamPete1, line 419: This probably warrant more than a "info" severity level. Probably critical or error. The definitions are here: https://mavlink.io/en/messages/common.html#MAV_SEVERITY

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -402,8 +410,15 @@ void AP_ICEngine::update(void)
...
+                starter_last_run_ms = now;
```
IamPete1, line 420: I think this deserves a comment.

libraries/AP_ICEngine/AP_ICEngine.h
```diff
@@ -107,6 +107,10 @@ class AP_ICEngine {
     // delay between start attempts (seconds)
     AP_Float starter_delay;
 
+    // max crank retry
+    AP_Int8 max_crank_retry;
+    int8_t crank_retry_ct;
```
IamPete1, line 112: Why not a uint8_t?

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -402,8 +410,15 @@ void AP_ICEngine::update(void)
         if (!should_run) {
             state = ICE_OFF;
         } else if (now - starter_last_run_ms >= starter_delay*1000) {
-            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Starting engine");
-            state = ICE_STARTING;
+            // check if we should retry starting the engine
+            if (max_crank_retry <= 0 || crank_retry_ct < max_crank_retry) {
+                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Starting engine");
+                state = ICE_STARTING;
+                crank_retry_ct++;
```
IamPete1, line 417: I think we should reset this on a un-commanded stop. You could build up some retrys on initial start then run for some long period of time. Then the engine stops without being commanded and you only have one retry left. Here:
https://github.com/ArduPilot/ardupilot/blob/04ee66bec4f6c5aa8e2f0d21a3facd34f1f8f29b/libraries/AP_ICEngine/AP_ICEngine.cpp#L430

IamPete1 (changes requested):

libraries/AP_ICEngine/AP_ICEngine.cpp
robertlong13: This would take a larger refactor of ICE than we wanted to push right now. There's currently no difference in the code between an unsuccessful start attempt and an uncommanded engine stop. Both go through the "engine running" state.

libraries/AP_ICEngine/AP_ICEngine.cpp
robertlong13: Tell that to "Uncommanded engine stop" 😆 

I also think this should be critical or error, but I thought I'd get pushback given the severity of the engine stop message

libraries/AP_ICEngine/AP_ICEngine.cpp
robertlong13: Great point

libraries/AP_ICEngine/AP_ICEngine.h
robertlong13: Basically just felt bad mixing signed and unsigned for the two, and there's no AP_Uint8 for the param.

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -402,8 +410,16 @@ void AP_ICEngine::update(void)
...
```
tridge, line 417: need to make this reset if we have been running for 5x crank time

tridge (changes requested):

tridge (approved):

libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -430,6 +449,13 @@ void AP_ICEngine::update(void)
                 // engine has stopped when it should be running
                 state = ICE_START_DELAY;
                 GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Uncommanded engine stop");
```
peterbarker, line 451: WARN

libraries/AP_ICEngine/AP_ICEngine.cpp
peterbarker: WARN

Tools/autotest/quadplane.py
```diff
@@ -1116,6 +1116,53 @@ def ICEngine(self):
...
+        self.start_subtest("Testing automatic restart")
+        # Limit start attempts to 4
+        max_tries = 4
+        self.set_parameter("ICE_STRT_MX_RTRY", max_tries)
+        # Make the engine unable to run (by messing up the RPM sensor)
+        rpm_chan = self.get_parameter("ICE_RPM_CHAN")
+        self.set_parameter("ICE_RPM_CHAN", 120) # Set to a non-existent sensor
+        self.set_rc(rc_engine_start_chan, 2000)
+        self.wait_statustext("Uncommanded engine stop")
+        self.wait_statustext("Starting engine")
```
peterbarker, line 1128: `check_context`, similarly elsewhere in here

peterbarker (changes requested):

PR #27931, opened 2024-08-26
AP_Notify:correct and expand 1 led flash sequences

Hwurzburg:
This aligns the single led notification flashes to 2 led for LED-A and adds a Failsafe indication


peterbarker (approved):
LGTM, but that might be because I don't care about changing the existing signalling!

PR #27941, opened 2024-08-27
AP_DDS: remove instance check for DDS Client

peterbarker:
this is probably a flow of control problem.

But the code block below this resets some state variables before returning, and will also return false in the same case this removed block does.

Resetting that state might be very important to the caller.


Ryanf55:
By returning false in `update_topic`, it prevents data from being published. If we don't have a fix, which is the statement below what you changed, then we want to publish there is no fix so the consuming application can handle the GPS outage. This PR, although indicating there is a logic problem, doesn't fix the underlying issue. 

I'll propose an alternative in the next few days to discuss. Many thanks for pointing this out!

peterbarker:
> By returning false in `update_topic`, it prevents data from being published. If we don't have a fix, which is the statement below what you changed, then we want to publish there is no fix so the consuming application can handle the GPS outage. This PR, although indicating there is a logic problem, doesn't fix the underlying issue.
> 
> I'll propose an alternative in the next few days to discuss. Many thanks for pointing this out!

The statement below is *not healthy*, *not* "no fix".  If an out-of-bound GPS instance is bassed to is_healthy it will return false, so you will go through the code path which resets the state, which is *probably* desirable.

Ryanf55 (approved):
Thanks for the explanation. Approved!

PR #27943, opened 2024-08-27
AP_GPS:fix metadata typo

Hwurzburg:


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -2021,7 +2021,7 @@ bool AP_GPS::gps_yaw_deg(uint8_t instance, float &yaw_deg, float &accuracy_deg,
     // @User: Advanced
 
     // @Param: _RATE_MS
-    // @DisplayName: GPS update rate in milliseconds. Renamed in 4.6 to GPS1_RAATE_MS
```
peterbarker, line 2024: Aren't we just losing the information about being renamed now?

Hwurzburg: no should only be in description not the displayed name

peterbarker:
Merged, thanks!

PR #27944, opened 2024-08-27
Drone build adding new board

hagerashraf:
I added two new boards to the file which are 5700 and 5701, and 8 other IDs for DroneBuild drones' boards
which means IDs reserved are from 5700 to 5710


peterbarker:
I've fixed up the commit message and squashed erverything into a single commit; see https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://ardupilot.org/dev/docs/git-interactive-rebase.html&ved=2ahUKEwjr78iA6ZaIAxWQr1YBHdv6BCcQFnoECBYQAQ&usg=AOvVaw1UtOQF4vjsu2PygliQN9Qv

Marking as MergeOnCIPass

PR #27945, opened 2024-08-28
hwdef: backport CUAV-7-Nano hwdef defines

cuav-chen2:
Update CUAV-7-Nano hwdef file. Correct LED defines and changes default battery voltage and current scale value on 4.5.
@peterbarker Could you review for it?


rmackay9:
Hi @cuav-chen2,

I had a quick peek at this and it seems like it's different than master?  Normally we make changes first to master and then backport them to the stable release (aka 4.5).  The voltage changes are fine I'm sure, they match master but the LED changes look different.

rmackay9:
After discussing with @peterbarker we think this is safe to merge to the stable branch.  I'll also pull it into the Plane, Rover, Tracker-4.5 branches.  It won't go out to users though until we do another point release but we don't have a specific plan to do a point release just yet.  I'm sure we will eventually but we're just about to start the 4.6 beta and we don't normally run two betas at once.. we can but it's rare.

cuav-chen2:
Hi @rmackay9 ,
Thanks for merge! The LED changes is due to the difference in LED configuration macros between versions 4.5 and 4.6 dev, so there will be differences. Now the LED works perfectly in 4.5.

rmackay9:
This is included in Copter-4.5.7-beta1

PR #27948, opened 2024-08-28
AP_Notify: Fix compilation error in GPIO_LED_1.cpp

mbuzdalov:
There was a space instead of an underscore.
I have no idea how this passed continuous integration.


andyp1per (approved):

peterbarker:
> There was a space instead of an underscore. I have no idea how this passed continuous integration.

We obviously don't build a board that only has a single LED on it in CI.

I'm working towards getting that sorted out so we can build all of the different LED types in SITL, all at once.

Issue #27949, opened 2024-08-28
DDVP tail rotor pitch actuator not working correctly


Ferruccio1984:
## Bug report

**Issue details**

_During setup of a customer helicopter with DDVP tail rotor type, it has been found out that tail rotor pitch actuator is not working.
Log image attached below shows output 4(tail pitch servo) non correct behavior. _
![image](https://github.com/user-attachments/assets/ba663206-25ff-4c28-9691-ce66f441fe49)

**Version**
_4.5.6_

**Platform**
[  ] All
[  ] AntennaTracker
[ x ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
_Helicopter_

**Hardware type**
_tested on a CubeOrange and on SITL_

**Logs**
_Please provide a link to any relevant logs that show the issue_

PR #27950, opened 2024-08-28
AP_Motors: fix DDVP tail servo initialization

Ferruccio1984:
Proposal to fix issue reported here: https://github.com/ArduPilot/ardupilot/issues/27949


MattKear (changes requested):
The fix looks correct to me, but it is failing CI.  I do not think that the heli test that it is failing is a flapping test.  I have not had chance to look into why it is failing though.

This is a PR to copter 4.5.  Can you open a PR to master please (problem is there too). We can then mark it for backport.

Ferruccio1984:
Thanks @MattKear , I have followed your request and opened a new PR against master. Closing this one

PR #27952, opened 2024-08-29
AC_WPNav: reduce WPNAV_SPEED min to 10 cm/s

clydemcqueen:
We are using ArduSub to run transects close to the seafloor (as low as 75 cm) while timed cameras capture high resolution stills. We get the best results when we move at 10 cm/s.

We are now starting to automate transects using GUIDED mode, but we've run into the lower speed limit of 20 cm/s. This limit seems arbitrary, so I'd like to reduce it to 10 cm/s.


rmackay9:
Sure, OK.  The reason we have the lower limit is because we get support calls from users saying, "why isn't my vehicle moving in Auto mode?" and the reason (sometimes) turns out to be that they've set the ACWPNAV_SPEED = 0 or something very low because they've mixed up cm with meters.  Once we introduced the lower limit those support calls all went away.  Hopefully they won't come back.

PR #17751, opened 2021-06-12
Plane: fixed issue with overspeed in QPOS_POSITIION1

tridge:
this fixes the backflip issue on the convergence that Henry found

PR #27953, opened 2024-08-29
Plane: add flag to disable wp speed scaling on heading and groundtrack

magate:
When performing a moving landing it is possible to have the prevailing headwind not in the direction of travel of the moving target. In this case is it undesirable to have the groundspeed scaled.

This adds a flag to optionally disable: https://github.com/ArduPilot/ardupilot/pull/17751

image
Moving platform example.


peterbarker (changes requested):
Needs to compile.

You're trying to use a 32-bit value to store a single bit of information; consider using a `Q_OPTION` bit instead

magate:
Q_OPTIONS 0
image

Q_OPTIONS 8388608 which is 2^23
image

Conditional waypoint hit instead of early exit showing the bit flag gated the code.

magate:
Was there an additional request on this prior to merge? I see the `WkiNeeded` label. Is that after merge or before? Thanks.

PR #27955, opened 2024-08-30
AP_HAL_ChibiOS scripts: Fixed chibios_hwdef checking for defaults presence in one folder and using another

joshanne:
`defaults_path` is relative to the folder the hwdef is in.
`self.default_params_filepath` is relative to the directory the script is run from.

The current implementation could result in attempting to load a defaults file from a location other than the expected one, or not actually loading a file when one is available to be loaded.

### DIFF
--- a/libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
+++ b/libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
@@ -2889,12 +2889,14 @@ Please run: Tools/scripts/build_bootloaders.py %s
     def write_processed_defaults_file(self, filepath):
         # see if board has a defaults.parm file or a --default-parameters file was specified
         defaults_filename = os.path.join(os.path.dirname(self.hwdef[0]), 'defaults.parm')
-        defaults_path = os.path.join(os.path.dirname(self.hwdef[0]), args.params)
-
         defaults_abspath = None
-        if os.path.exists(defaults_path):
-            defaults_abspath = os.path.abspath(self.default_params_filepath)
-            self.progress("Default parameters path from command line: %s" % self.default_params_filepath)
+        if isinstance(self.default_params_filepath, str):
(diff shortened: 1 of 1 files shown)


peterbarker (changes requested):
Needs a patch to `Tools/completion/zsh/_waf`

Needs to remove --params from parser object

Currently breaks the `waf` `--default-parameters` option, which may not be acceptable.  Will just out-and break `waf` building when `--params` is removed.

joshanne:
> Needs to remove --params from parser object

This isn't modifying waf, this is modifying `chibios_hwdef.py` - waf can still be called with `--default-parameters`. Calling with `--default-parameters` ends up calling down through to `generate_hwdef_h` in `chibios.py`.

That function calls into `chibios_hwdef.py` through `subprocess.call("python chibios_hwdef.py ..... --params ...)` as is the case here.

This _should not_ break the `--default-parameters` option, since we're just cleaning up what is _actually_ being used.

If you've got an example board that breaks, shoot me the command. I've tested with say, CubeOrange with the defaults, cube orange with overridden defaults using this parameter.

peterbarker:
Ah, darn.

Sorry, I had missed the fact that args.params (a global) was being used from within the body of the script class.

I contend that the correct fix here is not to remove it from the constructor arguments for ChibiOSHWdef, but to stop using `args.params` within that method body.

As it currently stands, `ChibiOHWDef is not going to fare well when instatiated from places ither than `chibios_hwdef.py` - but this PR is going to make things worse rather than better.

peterbarker:
Past https://github.com/ArduPilot/ardupilot/pull/27988 the use of `args.params` is the only remaining use of `args` within `ChibiOSHWDef` - and it would be good to kill its use in this PR....

joshanne:
This is breaking because the subprocess call turns an empty list into a string when constructing the command line.

I'm going to close this given I have a follow up PR that will effectively resolve this by removing the subprocess calls.

PR #27958, opened 2024-08-30
AP_NavEKF3: use reference for dal in frontend in place of method calls

peterbarker:
Saves some bytes and a little bit of time.

Passes the DAL down into the backend for it to use, too.  This will potentially be useful in a future PR - and makes sense as it should be using the same DAL as the frontend.

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3.cpp
@@ -744,7 +744,8 @@ const AP_Param::GroupInfo NavEKF3::var_info2[] = {
     AP_GROUPEND
 };
 
-NavEKF3::NavEKF3()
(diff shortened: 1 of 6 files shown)


peterbarker:
```
CubeOrange-periph-heavy  *                                                                     
Durandal                            -64    *           -88     -80               -112   -80    -80
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     0      *           -8      -24               120    -88    -48
MatekF405                           -16    *           0       80                80     0      -40
Pixhawk1-1M-bdshot                  32                 0       0                 -72    -40    40
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           -8     *           -8      -96               -72    -16    -16
skyviper-v2450                                         -40                                     
```

tridge (approved):

PR #27960, opened 2024-08-30
AP_MSP: make aux channels 1-4 available via msp rc command

catkira:
Motiviation for this patch: I have a device that needs to read aux channels via MSP. Currently it's only possible to read the 4 standard control channels via MSP.

### DIFF
--- a/libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
+++ b/libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
@@ -1079,18 +1079,30 @@ MSPCommandResult AP_MSP_Telem_Backend::msp_process_out_rc(sbuf_t *dst)
     float pitch = -rc().rc_channel(rcmap->pitch()-1)->norm_input_dz();
     float yaw = rc().rc_channel(rcmap->yaw()-1)->norm_input_dz();
     float throttle = rc().rc_channel(rcmap->throttle()-1)->norm_input_dz();
+    float aux1 = rc().rc_channel(4)->norm_input_dz();
+    float aux2 = rc().rc_channel(5)->norm_input_dz();
+    float aux3 = rc().rc_channel(6)->norm_input_dz();
+    float aux4 = rc().rc_channel(7)->norm_input_dz();
 
     const struct PACKED {
         uint16_t a;
         uint16_t e;
         uint16_t r;
         uint16_t t;
+        uint16_t aux1;
+        uint16_t aux2;
+        uint16_t aux3;
+        uint16_t aux4;
     } rc {
         // send only 4 channels, MSP order is AERT
         a : uint16_t(roll*500+1500),       // A
         e : uint16_t(pitch*500+1500),      // E
         r : uint16_t(yaw*500+1500),        // R
-        t : uint16_t(throttle*1000+1000)    // T
+        t : uint16_t(throttle*1000+1000),  // T
+        aux1 : uint16_t(aux1*500+1500),
+        aux2 : uint16_t(aux2*500+1500),
+        aux3 : uint16_t(aux3*500+1500),
+        aux4 : uint16_t(aux4*500+1500)
     };
 
     sbuf_write_data(dst, &rc, sizeof(rc));


catkira:
This patch needs some improvement, checking if aux 1-4 are available and maybe also output more aux channels than 1-4 if available.

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
```diff
@@ -1079,18 +1079,30 @@ MSPCommandResult AP_MSP_Telem_Backend::msp_process_out_rc(sbuf_t *dst)
     float pitch = -rc().rc_channel(rcmap->pitch()-1)->norm_input_dz();
     float yaw = rc().rc_channel(rcmap->yaw()-1)->norm_input_dz();
     float throttle = rc().rc_channel(rcmap->throttle()-1)->norm_input_dz();
+    float aux1 = rc().rc_channel(4)->norm_input_dz();
+    float aux2 = rc().rc_channel(5)->norm_input_dz();
+    float aux3 = rc().rc_channel(6)->norm_input_dz();
+    float aux4 = rc().rc_channel(7)->norm_input_dz();
```
peterbarker, line 1085: What makes channels 4 through 7 "aux" channels?

We may already be sending some of these channels through because they're in the rcmap as one of rpyt.

We could potentially nominate channels by their `RCn_OPTION` parameter to be MSP aux functions - but that would preclude their use by the autopilot to trigger an aux function.  However, that may also solve the which-channels-to-send question; I very much doubt you want to send 18 channels in a default configuration.

catkira: yes, assuming that aux channels start at ch4 is just a heuristic that is probably generally wrong.
Betaflight just sends all available channels with the MSP_RC command: https://github.com/betaflight/betaflight/blob/3e130de49ce5ceffc0e38e85d0fdb5d06c07a6bd/src/main/msp/msp.c#L1321C1-L1325C15
It's probably not a problem there, because this command is usually only sent over a network connection and not a wireless data link.
@peterbarker If you think using the `Rcn_OPTION` is a good idea, then I will look deeper into it

catkira: Another question is whether the MSP_RC ccommand should output processed channel values or raw PWM values like betaflight and inav does. Is there somewhere a hint about this in the MSP protocol documentation?

catkira: http://www.multiwii.com/wiki/index.php?title=Multiwii_Serial_Protocol here it says MSP_RC command should return raw pwm values
image

peterbarker: ... 1000-2000 isn't raw ;-)

peterbarker: @catkira @MichelleRos disagrees with you - those 1000-2000 do look rather prescriptive, and those were recently modified to normalized 1000-2000 values.

catkira: I tried to use the same conversion for aux as for roll, but that resulted in values from 1500 - 2000.
Is there a better way to get values from 1000 - 2000 without doing conversions with magic values?

MichelleRos: "Aux channels" is Betaflight terminology. In ArduPilot we just call it channel 5, 6 etc, so probably better to use that convention, eg "ch5" etc as the names.

Also, it's not magic values. It's just maths.
You'd need to use the norm_input_dz function to get a -1 to 1 (or a 0 to 1?) for each channel that's normalised based on the min, max, trim and deadzone parameters for that channel and then convert it to 1000 to 2000 form. I'm not sure whether ch5 etc comes out as -1 to 1 or 0 to 1. Depends on whether it's classified as an angle or a range type in the code. Sounds like it's the latter, in which case you need to use the method for throttle, not roll.

@peterbarker I wonder whether I should add a function to the RC library that gives a normalised 1000 to 2000 value? I've needed this kind of conversion more than once...

MichelleRos: @catkira "Aux channels" is Betaflight terminology. In ArduPilot we just call it channel 5, 6 etc, so probably better to use that convention, eg "ch5" etc as the names.

Also, it's not magic values. It's just maths.
You'd need to use the norm_input_dz function to get a -1 to 1 (or a 0 to 1?) for each channel that's normalised based on the min, max, trim and deadzone parameters for that channel and then convert it to 1000 to 2000 form. I'm not sure whether ch5 etc comes out as -1 to 1 or 0 to 1. Depends on whether it's classified as an angle or a range type in the code. Sounds like it's the latter, in which case you need to use the method for throttle, not roll.

@peterbarker I wonder whether I should add a function to the RC library that gives a normalised 1000 to 2000 value? I've needed this kind of conversion more than once...

MichelleRos: @catkira And it was changed from raw to this conversion because the main use of this code is for stick control for changing HDZero etc's settings using the sticks. When it was raw, it meant that if you had any reversals in your parameters (or, if you didn't have the right ones, IIRC), the commands would go the wrong way which was really frustrating and the reason why I did the original PR to make it normalised. (So, it's not that it has to be 1000 to 2000 exactly, but it does have to not be reversed, and some remotes do output channels reversed relative to what AP/MSP expects it to be.)

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
```diff
@@ -1079,18 +1079,30 @@ MSPCommandResult AP_MSP_Telem_Backend::msp_process_out_rc(sbuf_t *dst)
     float pitch = -rc().rc_channel(rcmap->pitch()-1)->norm_input_dz();
     float yaw = rc().rc_channel(rcmap->yaw()-1)->norm_input_dz();
     float throttle = rc().rc_channel(rcmap->throttle()-1)->norm_input_dz();
+    uint16_t aux1 = rc().rc_channel(4)->get_radio_in();
```
peterbarker, line 1082: I'm unhappy about just using channels 4 through 7 here.

We're already using a mapping to get the first four, I think we should use a mapping for the aux channels too.

Could you experiment with adding an `RC_Channel::AUX_FUNC::MSP_AUX_1` option and using that to find the nominated channel, please?  Note that the user may not have nominated one of them, so you do need to check for nullptr and use some default value on that aux channel.

peterbarker (changes requested):

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
MichelleRos: Or preferably, just not send the channel if no channel is set to that aux func. Eg if only msp_aux_1 is set, send only 5 channels.

peterbarker: ... but possibly filling gaps as required with dummy data.

catkira: How about sending all the channels all the time? The result from this command usually does not go over a data link, so a bit more bandwidth is not a big problem.

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
catkira: Thanks for the explaination, now I understand a bit better where it comes from. The aux channels that I tested have a range from 0 to 1. Is it possible to get the information whether its -1 to 1 or 0 to 1 from somewhere with a function call? This could avoid having the conversion to 1000 to 2000 range use magic values (and it would break if the 0 to 1 of aux channel is changed in the future).

MichelleRos: It is not "magic values". Please stop repeating that.
It is literal simple plain maths.

catkira: Yes, but the decision which conversion formula to use looks magic.
`uint16_t(yaw*500+1500)` or `uint16_t(throttle*1000+1000)`
It's not obvious for me, which one to use for aux channels. By trial and error I found out that I have to use the 'throttle-formula'.

MichelleRos: It's pretty easy to figure out that when you have a 0 to 1 number, like throttle, you need to multiply it by 1000 to get it to 0 to 1000 range and add 1000 to get it to 1000 to 2000 range.

Similarly, if you have a -1 to 1 value, you need to multiply it by 500 to get a -500 to 500 number and add 1500 to get a 1000 to 2000 number. 

If that seems too hard then I'm really concerned...

catkira: Yes, but how do I know if I have a 0 to 1 or -1 to 1 number. For throttle it's intuitive, but for aux channels not.

PS: thanks for speculating that I am a moron. It's never good to assume to be the smartest person in the room.

MichelleRos: You said nothing about wanting to know how to see whether it's a -1 to 1 or a 0 to 1. You just talked about "magic numbers."

But in any case, I'm done with this PR.

catkira: image

MichelleRos: "Depends on whether it's classified as an angle or a range type in the code. Sounds like it's the latter, in which case you need to use the method for throttle, not roll."

MichelleRos: image

MichelleRos: > Depends on whether it's classified as an angle or a range type in the code. Sounds like it's the latter, in which case you need to use the method for throttle, not roll.

Answered that already.

catkira: yes, but this classification should not be manually repeated at multiple locations in the code. Therefore I want to get this information without repeating it.
Ie is there a flag or something like that in the rc class that stores the information whether a channel is -1 to 1 or 0 to 1?

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
peterbarker: Doesn't bother me too much.  We can change it later if we feel like it.

But want ot see that mapping :-)

PR #27962, opened 2024-08-30
Tidy AP_DAL memory allocation and freeing

peterbarker:
```
CubeOrange-periph-heavy  *                                                                     
Durandal                            -8     *           -16     -16               -16    -8     -8
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     -16    *           -8      -8                -16    -16    -8
MatekF405                           -8     *           -16     -8                -16    -8     -8
Pixhawk1-1M-bdshot                  -8                 -8      -8                -8     -16    -8
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           -16    *           -16     -16               -16    -8     -16
skyviper-v2450                                         -8                                      
```

Be consistent in the way we use the enumeration (and which enumeration we use...)

Make sure we're symmetric for allocation/free calls by adding a free_type method to the DAL.

### DIFF
--- a/libraries/AP_DAL/AP_DAL.cpp
+++ b/libraries/AP_DAL/AP_DAL.cpp
@@ -272,9 +272,13 @@ int AP_DAL::snprintf(char* str, size_t size, const char *format, ...) const
     return res;
 }
 
-void *AP_DAL::malloc_type(size_t size, Memory_Type mem_type) const
+void *AP_DAL::malloc_type(size_t size, MemoryType mem_type) const
 {
     return hal.util->malloc_type(size, AP_HAL::Util::Memory_Type(mem_type));
+}
+void AP_DAL::free_type(void *ptr, size_t size, MemoryType mem_type) const
+{
+    return hal.util->free_type(ptr, size, AP_HAL::Util::Memory_Type(mem_type));
 }
 
 // map core number for replay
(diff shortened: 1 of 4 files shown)


tpwrules:
Why the redundant alloc/free and MemoryType? Is the point that the EKFs only use the DAL and not the HAL?

Might be able to save more code size by putting those in the .h so they can be inlined.

tridge (approved):

peterbarker:
> Why the redundant alloc/free and MemoryType? Is the point that the EKFs only use the DAL and not the HAL?

As discussed on the call, I think this was an attempt to aid in the "dal-standalone-build" testing.  I don't think it's actually required now!

> Might be able to save more code size by putting those in the .h so they can be inlined.

Also looked at this on the call - the header doesn't know about the HAL ATM.

I *think* my preference would be to remove the wrapping if we can - but that would require making sure the stand-alone build is definitely still doing its job, I think.

peterbarker:
This is now a zero-compiler-output change:
```
CubeOrange-periph-heavy  *                                                                     
Hitec-Airspeed           *                                                                     
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
```

That's because we merged the change which makes the dal in NavEKF3 a reference, so we're not longer making a method call to get the value of the enumeration.

The compiler must be optimising away the new wrapping function, so that's all happy.

PR #27965, opened 2024-08-31
Avoid use of AP_Logger.h in library headers

peterbarker:
this pulls in many more headers, we should avoid using it whereever we can

```
CubeOrange-periph-heavy  *                                                                     
Hitec-Airspeed           *                                                                     
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
```

### DIFF
--- a/libraries/APM_Control/AP_AutoTune.cpp
+++ b/libraries/APM_Control/AP_AutoTune.cpp
@@ -250,7 +250,7 @@ void AP_AutoTune::update(AP_PIDInfo &pinfo, float scaler, float angle_err_deg)
 #if HAL_LOGGING_ENABLED
     if (now - last_log_ms >= 40) {
         // log at 25Hz
-        struct log_ATRP pkt = {
+        const struct log_ATRP pkt {
             LOG_PACKET_HEADER_INIT(LOG_ATRP_MSG),
             time_us : AP_HAL::micros64(),
             type : uint8_t(type),
--- a/libraries/APM_Control/AP_AutoTune.h
+++ b/libraries/APM_Control/AP_AutoTune.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <AP_Logger/AP_Logger.h>
 #include <AP_Logger/LogStructure.h>
 #include <AP_Param/AP_Param.h>
 #include <AP_Vehicle/AP_FixedWing.h>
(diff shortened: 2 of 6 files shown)


libraries/APM_Control/AP_AutoTune.cpp
```diff
@@ -250,7 +250,23 @@ void AP_AutoTune::update(AP_PIDInfo &pinfo, float scaler, float angle_err_deg)
 #if HAL_LOGGING_ENABLED
     if (now - last_log_ms >= 40) {
         // log at 25Hz
-        struct log_ATRP pkt = {
+        const struct PACKED {
```
IamPete1, line 253: I would have expected this to be deleted from where ever it is now.

libraries/APM_Control/AP_AutoTune.h
```diff
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <AP_Logger/AP_Logger.h>
```
tridge, line 3: can remove logstructure.h here too, if you remove log_ATRP

libraries/APM_Control/AP_AutoTune.cpp
peterbarker: Ah, I remember now.  We can't move the definition of this into the .cpp file.  log_ATRP is included into ArduPlane's log structure:

```
SCB-WAF: [1166/1206] Compiling ArduPlane/mode.cpp../../ArduPlane/Log.cpp:363:41: error: 'log_ATRP' is not a member of 'AP_AutoTune'
SCB-WAF: ../../ArduPlane/Log.cpp:363:41: error: 'log_ATRP' is not a member of 'AP_AutoTune'  363 |     { LOG_ATRP_MSG, sizeof(AP_AutoTune::log_ATRP),
SCB-WAF:   363 |     { LOG_ATRP_MSG, sizeof(AP_AutoTune::log_ATRP),      |                                         ^~~~~~~~
SCB-WAF:       |                                         ^~~~~~~~compilation terminated due to -Wfatal-errors.
```

... I discovered that but only half-undid the change :-(

I really think we should look at changing this to a WriteStreaming or similar...

libraries/APM_Control/AP_AutoTune.h
peterbarker: Yeah, can't do that as that message is included in ArduPlane's LogStructure!

Should probably be a WriteStreaming call instead, but for now I've put it back in the header, where we only need `AP_Logger/LogStructure.h`, not `AP_Logger.h`, which is the important change.

peterbarker:
Still no compiler output changes with this.

PR #27970, opened 2024-09-01
Dependency and extraction fixes for buld_options.py

peterbarker:


Hwurzburg:
@peterbarker does this the issues on the wiki features extraction pages now? if not, I need to remove that page from the wiki

PR #27977, opened 2024-09-02
Restart logging after transfer complete when LOG_DISARMED=1

peterbarker:
_read_fd was never unset from -1 after the log transfer was complete.

### DIFF
--- a/libraries/AP_Logger/AP_Logger_MAVLinkLogTransfer.cpp
+++ b/libraries/AP_Logger/AP_Logger_MAVLinkLogTransfer.cpp
@@ -128,7 +128,7 @@ void AP_Logger::handle_log_request_data(GCS_MAVLINK &link, const mavlink_message
         uint16_t num_logs = get_num_logs();
         if (packet.id > num_logs || packet.id < 1) {
             // request for an invalid log; cancel any current download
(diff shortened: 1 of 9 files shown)


libraries/AP_Logger/AP_Logger_MAVLinkLogTransfer.cpp
```diff
@@ -174,11 +174,16 @@ void AP_Logger::handle_log_request_erase(GCS_MAVLINK &link, const mavlink_messag
...
+    // mavlink_log_request_end_t packet;
+    // mavlink_msg_log_request_end_decode(&msg, &packet);
+    end_log_transfer();
+}
 
+void AP_Logger::end_log_transfer()
+{
     transfer_activity = TransferActivity::IDLE;
     _log_sending_link = nullptr;
+    backends[0]->end_log_transfer();
```
andyp1per, line 186: Shouldn't you call this on all of the backends?

peterbarker: > Shouldn't you call this on all of the backends?

We only do transfers from the first backend - there's lots of this [0] stuff around.

tridge (approved):

PR #27978, opened 2024-09-02
AP_HAL_ESP32: improve the reported LOOP_RATE on esp32s3 to similar lo…

davidbuzz:
…op rate to classic esp32

getting a loop rate of around 130 on both now

source modules/esp_idf/export.sh

./waf configure --board=esp32s3empty --debug
./waf copter
cd build/esp32buzz/esp-idf_build/
ninja menuconfig

./waf configure --board=esp32buzz --debug
./waf copter
cd build/esp32buzz/esp-idf_build/
ninja menuconfig

and try to get both the resulting modified sdkconfig as smiilar as possible - this causes s3 to use qio, which is faster, and also puts the wifi stuff onto CORE1 on both configs.

s3:
loop_rate: actual: 148.097137Hz, expected: 400Hz
loop_rate: actual: 147.908691Hz, expected: 400Hz

classic:
loop_rate: actual: 188.718842Hz, expected: 400Hz
loop_rate: actual: 191.819748Hz, expected: 400Hz


tpwrules (approved):
Checked and got similar numbers on `esp32s3empty`. DIO -> QIO is the big change here but the others should help a bit too.

WiFi latency even seems a little better too.

tridge (approved):

PR #27981, opened 2024-09-02
AP_Baro: remove unnecessary "#ifdef HAL_BUILD_AP_PERIPH"

junan76:
Since there are already two implementations of the "healthy" method in file "AP_Baro.cpp", and they are both with clear comments, I think that "#ifdef" thing should be removed from the header file :P

### DIFF
--- a/libraries/AP_Baro/AP_Baro.h
+++ b/libraries/AP_Baro/AP_Baro.h
@@ -56,12 +56,8 @@ public:
 
     // healthy - returns true if sensor and derived altitude are good
     bool healthy(void) const { return healthy(_primary); }
-#ifdef HAL_BUILD_AP_PERIPH
-    // calibration and alt check not valid for AP_Periph
+
     bool healthy(uint8_t instance) const;
-#else
-    bool healthy(uint8_t instance) const;
-#endif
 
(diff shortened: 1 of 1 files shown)


libraries/AP_Baro/AP_Baro.h
```diff
@@ -56,12 +56,9 @@ class AP_Baro
 
     // healthy - returns true if sensor and derived altitude are good
     bool healthy(void) const { return healthy(_primary); }
-#ifdef HAL_BUILD_AP_PERIPH
+
     // calibration and alt check not valid for AP_Periph
```
peterbarker, line 60:
```suggestion
```

peterbarker (changes requested):
Wow.  Nice cleanup!

peterbarker (approved):
LGTM

peterbarker:
```
Board,AP_Periph,blimp,bootloader,copter,heli,plane,rover,sub
Durandal,,*,*,*,*,*,*,*
Hitec-Airspeed,*,,,,,,,
KakuteH7-bdshot,,*,*,*,*,*,*,*
MatekF405,,*,*,*,*,*,*,*
f103-QiotekPeriph,*,,,,,,,
f303-Universal,*,,,,,,,
```

Issue #25825, opened 2023-12-23
Rover: Circle mode instabilities


stephendade:
Whilst attempting to tune my rover, I experienced some serious instabilities and "wandering off" in circle mode:

5m circle at 2m/s speed:
![xmaxx-2ms](https://github.com/ArduPilot/ardupilot/assets/1731976/75b908e8-f6a7-4d9a-a58f-9d1c64e4c18b)

A 5m circle at 1m/s speed was fine.

My current theory is that the default steering PID values are too low for this particular vehicle. I've confirmed in manual mode that the vehicle can definitely turn at the rates needed for a 5m circle at 2m/s (23deg/sec).

I could replicate it in SITL via restricting the ``ATC_STR_ACC_MAX`` and ``ATC_STR_RAT_MAX`` parameters to below the minimum degrees/sec required for the circle. So:
```
param set CIRC_RADIUS 5
param set CIRC_SPEED 4
param set ATC_STR_ACC_MAX 90
param set ATC_STR_RAT_MAX 30
```
SITL result:
![sitl](https://github.com/ArduPilot/ardupilot/assets/1731976/9ace43ee-a505-4bb2-a86c-00f379f4ab4f)

I usually wouldn't be concerned about an untuned vehicle performing badly in these higher level flight modes, but given the Quiktune feature requires running in circle mode, it should be expected that untuned vehicles will be running in this mode.

For a fix, I'd suggest that circle mode needs to stop the vehicle if it gets more than X metres outside the circle (where X is +20% on radius?) and suggest to the user that they either a) Increase the radius b) decrease the speed c) increase the Steering PID FF(?).

Alternatively, maybe the way circle mode calculates it's trajectory needs to be improved? I feel that it does wander more than necessary ... it's like it gives up after a while!

rmackay9:
yes, I've seen these problems as well actually but haven't found a way to protect against the problem.  I think the issue occurs because we pass a target position, velocity and acceleration to the position controller but even a large position error does not overpower the velocity and acceleration inputs.  So maybe the solution should be that the desired velocity and acceleration only get to use whatever control bandwidth is left after the velocity and acceleration required to shrink the position error has been calculated.

stephendade:
As an update to anyone with a similar problem: Increasing ``ATC_STR_RAT_FF`` can "fix" the issue until a quiktune can be run.

PR #27983, opened 2024-09-02
Rover: Circle mode enhancements

stephendade:
Two fixes for circle mode in Rover:
- Correct reporting of target position and distance/bearing to target
- Stop circling (stop the vehicle) if it's unable to properly track the circle. This is usually due to the vehicle not being able to turn fast enough for a given speed/radius circle.

### DIFF
--- a/Rover/mode_circle.cpp
+++ b/Rover/mode_circle.cpp
@@ -70,7 +70,7 @@ bool ModeCircle::set_center(const Location& center_loc, float radius_m, bool dir
     return true;
 }
 
-// initialize dock mode
+// initialize circle mode from current position
 bool ModeCircle::_enter()
 {
     // capture starting point and yaw
@@ -84,6 +84,9 @@ bool ModeCircle::_enter()
     config.speed = is_positive(speed) ? speed : g2.wp_nav.get_default_speed();
     target.yaw_rad = AP::ahrs().get_yaw();
     target.speed = 0;
+
+    // record center as location (only used for reporting)
+    config.center_loc = rover.current_loc;
 
     // check speed around circle does not lead to excessive lateral acceleration
     check_config_speed();
@@ -146,8 +149,8 @@ void ModeCircle::update()
 
     // Update distance to destination and distance to edge
     const Vector2f center_to_veh = curr_pos - config.center_pos;
-    _distance_to_destination = center_to_veh.length();
-    dist_to_edge_m = fabsf(_distance_to_destination - config.radius);
+    _distance_to_destination = (target.pos.tofloat() - curr_pos).length();
+    dist_to_edge_m = fabsf(center_to_veh.length() - config.radius);
 
     // Update depending on stage
     if (!reached_edge) {
@@ -221,7 +224,7 @@ float ModeCircle::wp_bearing() const
         return 0;
     }
     // calc vector from circle center to vehicle
-    Vector2f veh_to_center = (config.center_pos - curr_pos_NE);
+    Vector2f veh_to_center = (target.pos.tofloat() - curr_pos_NE);
     if (veh_to_center.is_zero()) {
         return 0;
     }
@@ -266,6 +269,7 @@ bool ModeCircle::set_desired_speed(float speed_ms)
 bool ModeCircle::get_desired_location(Location& destination) const
 {
     destination = config.center_loc;
+    destination.offset_bearing(degrees(target.yaw_rad), config.radius);
     return true;
 }
 
--- a/Rover/mode_circle.cpp
+++ b/Rover/mode_circle.cpp
@@ -91,6 +91,9 @@ bool ModeCircle::_enter()
     // check speed around circle does not lead to excessive lateral acceleration
     check_config_speed();
 
+    // reset tracking_back 
+    tracking_back = false;
+
     // calculate speed, accel and jerk limits
     // otherwise the vehicle uses wp_nav default speed limit
     float atc_accel_max = MIN(g2.attitude_control.get_accel_max(), g2.attitude_control.get_decel_max());
@@ -155,7 +158,14 @@ void ModeCircle::update()
     // Update depending on stage
     if (!reached_edge) {
         update_drive_to_radius();
-
+    } else if (dist_to_edge_m > 2 * MAX(g2.turn_radius, AR_CIRCLE_REACHED_EDGE_DIST) && !tracking_back) {
+        // if more than 2* turn_radius outside circle radius, slow vehicle by 20%
+        config.speed = 0.8 * config.speed;
+        GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "Circle: cannot keep up, slowing to %.1fm/s", config.speed);
+        tracking_back = true;
+    } else if (dist_to_edge_m < MAX(g2.turn_radius, AR_CIRCLE_REACHED_EDGE_DIST) && tracking_back) {
+        // if within turn_radius, call the vehicle back on track
+        tracking_back = false;
     } else {
         update_circling();
     }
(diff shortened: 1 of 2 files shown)


stephendade:
Fix for #25825

The issue can be replicated in SITL by setting ``ATC_STR_RAT_MAX to 10``

Rover/mode_circle.cpp
```diff
@@ -146,13 +149,17 @@ void ModeCircle::update()
...
 
     // Update depending on stage
     if (!reached_edge) {
         update_drive_to_radius();
-
+    } else if (dist_to_edge_m > config.radius * 0.2) {
+        // if more than 20% inside or outside circle radius, stop vehicle
+        GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "Circle: vehicle unable to turn fast enough to track circle. Stopping vehicle.");
+        stop_vehicle();
+        return;
```
IamPete1, line 162: We also come through here in loiter turns mission items. This is not such a great solution in that case. Also it looks like it will call send text at loop rate?

IamPete1 (commented):
Not sure about the stopping, rest looks OK.

We do already limit the speed on circle enter for the given accel limits, maybe we just need some safety factor there, or to include some missing limiting factors.

stephendade:
> We do already limit the speed on circle enter for the given accel limits, maybe we just need some safety factor there, or to include some missing limiting factors.

I assume you mean the throttle accel limits?

This is more the case when the vehicle turn rate (yaw) is much less than the desired turn rate. The vehicle (currently) does not slow down if it can't reach the desired turn rate. It'd be nice if it did, though - I couldn't quite work out the maths to do that though.

Rover/mode_circle.cpp
stephendade: Yeah, I should add in a flag to only send this message once...

stephendade:
Ok, I've changed my approach here:
If the vehicle can't track the circle, it will slow the vehicle speed by 20% and then attempt to track back onto the circle

Worked well in SITL

stephendade:
Tested in a real vehicle and it worked just fine

@IamPete1 and/or @rmackay9 - are you happy to approve?

Rover/mode_circle.cpp
```diff
@@ -70,7 +70,7 @@ bool ModeCircle::set_center(const Location& center_loc, float radius_m, bool dir
     return true;
 }
 
-// initialize dock mode
+// initialize circle mode from current position
 bool ModeCircle::_enter()
```
IamPete1, line 74: I would have expected a `tracking_back = false;` in here somewhere.

Rover/mode_circle.cpp
```diff
@@ -146,13 +149,21 @@ void ModeCircle::update()
...
     // Update depending on stage
     if (!reached_edge) {
         update_drive_to_radius();
-
+    } else if (dist_to_edge_m > config.radius * 0.2 && !tracking_back) {
+        // if more than 20% inside or outside circle radius, slow vehicle
+        config.speed = 0.8 * config.speed;
+        GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "Circle: Vehicle unable to turn fast enough to track circle. Slowing vehicle to %.1fm/s", config.speed);
+        tracking_back = true;
+    } else if (dist_to_edge_m < config.radius * 0.05 && tracking_back) {
```
IamPete1, line 163: Would it make scene to use the same threshold as we do for starting the circle in the first place?

https://github.com/ArduPilot/ardupilot/blob/1e8e250459e9a68571f1c5265066b89fde7c5f4a/Rover/mode_circle.cpp#L174

Rover/mode_circle.cpp
```diff
@@ -190,12 +201,16 @@ void ModeCircle::update_circling()
...
-    angle_total_rad += angle_dt;
+    // calculate angular rate and update target angle, if the vehicle is not trying to track back
+    if (!tracking_back) {
+        const float circumference = 2.0 * M_PI * config.radius;
+        const float angular_rate_rad = (target.speed / circumference) * M_2PI * (config.dir == Direction::CW ? 1.0 : -1.0);
+        const float angle_dt = angular_rate_rad * rover.G_Dt;
+        target.yaw_rad = wrap_PI(target.yaw_rad + angle_dt);
+        angle_total_rad += angle_dt;
+    } else {
+        init_target_yaw_rad();
```
IamPete1, line 212: This means that we effectively reset the target position to the closest position on the circle. So the position controller will always be pushing radially. But there is still some target velocity and acceleration. Because only the target position is changed were not giving kinematically consistent targets. You could swap to just giving a position target like we do on the `drive_to_radius` stage. You could just kick back into that stage and reuse the `reached_edge` flag.

stephendade: My understanding is that the target velocity/accel are calculated further down in ``update_circling()`` based on the target yaw, keeping it consistent.

A position target won't work because we need to vehicle at the correct heading (velocity) to have a smooth entry into the circle:
Screenshot from 2024-09-09 20-41-01

Rover/mode_circle.cpp
stephendade: Fixed

Rover/mode_circle.cpp
stephendade: Sure, that'll work. Changed to use ``MAX(g2.turn_radius, AR_CIRCLE_REACHED_EDGE_DIST)`` instead

Rover/mode_circle.cpp
stephendade: This is what the PR currently does. I think it's a smoother approach back into the circle:
Screenshot from 2024-09-09 20-52-42

IamPete1: OK, i'm convinced.

IamPete1 (approved):
Looks good.

Rover/mode_circle.cpp
```diff
@@ -146,13 +152,21 @@ void ModeCircle::update()
...
+    dist_to_edge_m = fabsf(center_to_veh.length() - config.radius);
 
     // Update depending on stage
     if (!reached_edge) {
         update_drive_to_radius();
-
+    } else if (dist_to_edge_m > 2 * MAX(g2.turn_radius, AR_CIRCLE_REACHED_EDGE_DIST) && !tracking_back) {
+        // if more than 2* turn_radius outside circle radius, slow vehicle by 20%
+        config.speed = 0.8 * config.speed;
+        GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "Circle: Vehicle unable to turn fast enough to track circle. Slowing vehicle to %.1fm/s", config.speed);
```
rmackay9, line 164: I wonder if we could make this comment shorter.  Perhaps, "Circle: cannot keep up, slowing to %.1fm/s"

Rover/mode_circle.cpp
```diff
@@ -146,13 +152,21 @@ void ModeCircle::update()
...
+        tracking_back = true;
+    } else if (dist_to_edge_m < MAX(g2.turn_radius, AR_CIRCLE_REACHED_EDGE_DIST) && tracking_back) {
+        // if within turn_radius, call the vehicle back on track
+        tracking_back = false;
+        GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "Circle: Vehicle back on circle");
```
rmackay9, line 169: I wonder if we need this message.  Normally we don't send out messages when things are going as they should.  If the vehicle speed is increased back to the default or something we could warn about that.

rmackay9:
This looks pretty good to me, thanks!  I've made some comments on text messages to the user but after that's let's merge

Rover/mode_circle.cpp
stephendade: Done

Rover/mode_circle.cpp
stephendade: Removed message

rmackay9 (approved):
LGTM, thanks!

PR #27985, opened 2024-09-02
Plane: added glider pullup support

tridge:
This brings in a pullup system for gliders developed by @priseborough and myself. It is used for the NOAA hight altitude glider drops
This feature is not compiled in by default except on SITL to minimise flash costs for most users. It is expected that users doing high altitude glider drops will use the custom build server or a custom hwdef.dat. Note that doing this with a lua script was looked at and it would require quite a few new binding extensions which I didn't think was worth doing, and would cost more flash for most users.
The feature is enabled with the PUP_ENABLE parameter. When enabled it adds a new stage to the NAV_ALTITUDE_WAIT mission item, controlled by the PUP_* parameters. This stage controls the pullup manoeuvrer where we do a g-force limited pullup out of a dive, bring the aircraft to a normal flight regime where normal mission commands can take over


ArduPlane/Attitude.cpp
```diff
@@ -392,6 +392,13 @@ void Plane::stabilize()
     }
 #endif
 
+#if AP_PLANE_GLIDER_PULLUP_ENABLED
+    if (pullup.in_pullup()) {
+        pullup.stabilize_pullup(get_speed_scaler());
+        return;
```
IamPete1, line 398: really not a fan of this early return. I was trying to get everything to go through its respective `plane.control_mode->run()` function. Could this be in auto's run function?

Also note that because of the check on the next line we will reset the attitude controllers after the pull-up.

tridge: yep, I've moved it to mode_auto in run()

ArduPlane/navigation.cpp
```diff
@@ -321,7 +321,7 @@ void Plane::update_loiter_update_nav(uint16_t radius)
     if ((loiter.start_time_ms == 0 &&
          (control_mode == &mode_auto || control_mode == &mode_guided) &&
          auto_state.crosstrack &&
-         current_loc.get_distance(next_WP_loc) > radius*3) ||
+         current_loc.get_distance(next_WP_loc) > 3 * nav_controller->loiter_radius(fabsf(radius))) ||
```
IamPete1, line 324:
```suggestion
         current_loc.get_distance(next_WP_loc) > 3 * nav_controller->loiter_radius(radius)) ||
```
Don't need the abs, radius is uint.

tridge: fixed

ArduPlane/pullup.cpp
```diff
@@ -0,0 +1,244 @@
...
+    // @Units: m/s
+    // @Range: 0 100
+    // @User: Advanced
+    AP_GROUPINFO("ARSPD_START", 6, GliderPullup,  airspeed_start, 30),
+
+    // @Param: PITCH_START
+    // @DisplayName: Pullup target pitch
+    // @Description: Target pitch for initial pullup
+    // @Units: deg
+    // @Range: -90 0
```
IamPete1, line 60: -90 is probably a bad idea, looks like its using eulers.

ArduPlane/pullup.cpp
```diff
@@ -0,0 +1,244 @@
...
+  altitude or exceeded descent rate
+*/
+bool GliderPullup::pullup_start(void)
+{
+    if (enable != 1) {
+        return false;
+    }
+
+    // release balloon
+    SRV_Channels::set_output_scaled(SRV_Channel::k_lift_release, 100);
```
IamPete1, line 94: I would have expected this to be end of altitude wait rather than the start of pullup.

IamPete1: Also I have been tending to add new things with a range of 1.

ArduPlane/pullup.cpp
```diff
@@ -0,0 +1,244 @@
...
+        plane.nav_pitch_cd = 0;
+        plane.nav_roll_cd = 0;
+        uint32_t now = AP_HAL::millis();
+        if (now - plane.last_stabilize_ms > 1000) {
+            plane.pitchController.reset_I();
+            plane.yawController.reset_I();
+        }
+        plane.last_stabilize_ms = now;
+        SRV_Channels::set_output_scaled(SRV_Channel::k_aileron, plane.rollController.get_rate_out(0, speed_scaler));
+        plane.logger.Write_PID(LOG_PIDR_MSG, plane.rollController.get_pid_info());
```
IamPete1, line 193: I don't think we should just start logging at loop rate. Even if we should this is a odd way of doing it because the existing PID logging is still going on, so there will be lots of duplication.

ArduPlane/pullup.cpp
```diff
@@ -0,0 +1,244 @@
...
+        ng_demand = 0.0;
+        break;
+    }
+    case Stage::WAIT_PITCH: {
+        plane.yawController.reset_I();
```
IamPete1, line 198: Yaw controller I is not reset continually in `Stage::WAIT_AIRSPEED`, I would have expected it to be the same, looks like its also zeroing the rudder.

ArduPlane/pullup.cpp
```diff
@@ -0,0 +1,244 @@
...
+            // apply a rate of change limit to the ng pullup demand
+            ng_demand += MAX(ng_jerk_limit / ahrs.get_EAS2TAS(), 0.1f) * plane.scheduler.get_loop_period_s();
+            ng_demand = MIN(ng_demand, ng_limit);
+            const float VTAS_ref = ahrs.get_EAS2TAS() * aspeed;
+            const float pullup_accel = ng_demand * GRAVITY_MSS;
+            const float demanded_rate_dps = degrees(pullup_accel / VTAS_ref);
+            const uint32_t elev_trim_offset_cd = 4500.0f * elev_offset * (1.0f - ng_demand / ng_limit);
+            SRV_Channels::set_output_scaled(SRV_Channel::k_elevator, elev_trim_offset_cd + plane.pitchController.get_rate_out(demanded_rate_dps, speed_scaler));
+            plane.logger.Write_PID(LOG_PIDP_MSG, plane.pitchController.get_pid_info());
+            plane.logger.Write_PID(LOG_PIDR_MSG, plane.rollController.get_pid_info());
```
IamPete1, line 216: Not a fan of the bonus logging.

ArduPlane/pullup.cpp
```diff
@@ -0,0 +1,244 @@
...
+        const auto &ahrs = plane.ahrs;
+        if (ahrs.airspeed_estimate(aspeed)) {
+            // apply a rate of change limit to the ng pullup demand
+            ng_demand += MAX(ng_jerk_limit / ahrs.get_EAS2TAS(), 0.1f) * plane.scheduler.get_loop_period_s();
+            ng_demand = MIN(ng_demand, ng_limit);
+            const float VTAS_ref = ahrs.get_EAS2TAS() * aspeed;
+            const float pullup_accel = ng_demand * GRAVITY_MSS;
+            const float demanded_rate_dps = degrees(pullup_accel / VTAS_ref);
+            const uint32_t elev_trim_offset_cd = 4500.0f * elev_offset * (1.0f - ng_demand / ng_limit);
+            SRV_Channels::set_output_scaled(SRV_Channel::k_elevator, elev_trim_offset_cd + plane.pitchController.get_rate_out(demanded_rate_dps, speed_scaler));
```
IamPete1, line 214: Should we set `SRV_Channels::set_output_scaled(SRV_Channel::k_elevator, elev_trim_offset_cd)` if there is no airspeed? Currently it does not set in that case. That should mean its still at the value from the last loop, but its possible we set it somewhere else so that is not the case.

ArduPlane/pullup.cpp
```diff
@@ -0,0 +1,244 @@
...
+        SRV_Channels::set_output_scaled(SRV_Channel::k_aileron, plane.rollController.get_rate_out(0, speed_scaler));
+        plane.logger.Write_PID(LOG_PIDR_MSG, plane.rollController.get_pid_info());
+        ng_demand = 0.0;
+        break;
+    }
+    case Stage::WAIT_PITCH: {
+        plane.yawController.reset_I();
+        plane.nav_roll_cd = 0;
+        plane.nav_pitch_cd = 0;
+        plane.last_stabilize_ms = AP_HAL::millis();
```
IamPete1, line 201: This is done in all the cases.

libraries/SRV_Channel/SRV_Channel_aux.cpp
```diff
@@ -140,6 +140,7 @@ void SRV_Channel::aux_servo_function_setup(void)
     case k_flap:
     case k_flap_auto:
     case k_egg_drop:
+    case k_lift_release:
         set_range(100);
```
IamPete1, line 144: I would like to start setting range/angle of 1 for new things.
```suggestion
    case k_lift_release:
        set_range(1);
```
Eventually we can get rid of all the `* 4500`

ArduPlane/pullup.cpp
tridge: a user could chain together multiple altitude waits, with for example a servo command in between to setup some instrument they need at a particular height. They only definately want to release the lift if they do a pullup
on the range, I don't see any set_range(1) in master? it is still a uint16_t, not a float btw.

ArduPlane/commands_logic.cpp
```diff
@@ -831,17 +849,38 @@ bool Plane::verify_continue_and_change_alt()
...
-        gcs().send_text(MAV_SEVERITY_INFO,"Reached altitude");
-        return true;
+#if AP_PLANE_GLIDER_PULLUP_ENABLED
+    if (plane.pullup.in_pullup()) {
+        return plane.pullup.verify_pullup();
     }
-    if (plane.auto_state.sink_rate > cmd.content.altitude_wait.descent_rate) {
+#endif
+
+    const float alt_diff = plane.current_loc.alt - cmd.content.altitude_wait.altitude*100.0f;
```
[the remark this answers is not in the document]
tridge: I added a comment that it is always AMSL, it actually had to be due to the way the message is, the frame is not available (hasLocation=false in the xml). We can't fit the location and frame and the three parameters in the space in our mission storage

ArduPlane/commands_logic.cpp
```diff
@@ -831,17 +849,38 @@ bool Plane::verify_continue_and_change_alt()
...
+#endif
+        return true;
+    }
+
+    const float time_to_alt = alt_diff / MIN(plane.auto_state.sink_rate, -0.01);
+
     // if requested, wiggle servos
-    if (cmd.content.altitude_wait.wiggle_time != 0) {
+    if (cmd.content.altitude_wait.wiggle_time != 0 &&
+        (plane.auto_state.sink_rate > 0 || time_to_alt > cmd.content.altitude_wait.wiggle_time*5)) {
```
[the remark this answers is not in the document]
tridge: the command is specifically for high alt balloon launches, and only supports going up. I've added a comment

ArduPlane/mode.cpp
```diff
@@ -101,6 +101,9 @@ bool Mode::enter()
     plane.target_altitude.terrain_following_pending = false;
 #endif
 
+    // start with normal servo control
```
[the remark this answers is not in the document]
tridge: done

ArduPlane/pullup.cpp
tridge: changed to -80

ArduPlane/pullup.cpp
tridge: removed, this was primarily for debugging

ArduPlane/pullup.cpp
tridge: done

ArduPlane/pullup.cpp
tridge: removed

ArduPlane/pullup.cpp
tridge: i've added it, but in reality airspeed_estimate() always returns true if we have ever had a GPS lock, which will mean it's always true for this command

ArduPlane/pullup.cpp
tridge: moved to one plane

libraries/SRV_Channel/SRV_Channel_aux.cpp
tridge: there are no set_range(1) calls in master right now, and while the range is a uint16_t I don't think 1 is a good choice as a default as it means it can only be binary. That is OK for this case, but not a good pattern

ArduPlane/commands_logic.cpp
```diff
@@ -766,6 +779,11 @@ bool Plane::verify_loiter_to_alt(const AP_Mission::Mission_Command &cmd)
         result = verify_loiter_heading(false);
     }
 
+    // additional check for altitude target, even if blown off course
+    if (current_loc.alt < cmd.content.location.alt) {
+        result = true;
+    }
```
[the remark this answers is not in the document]
tridge: that was some leftover debugging, thanks for noticing!
removed

ArduPlane/Plane.h
```diff
@@ -1272,6 +1282,10 @@ class Plane : public AP_Vehicle {
     // last target alt we passed to tecs
     int32_t tecs_target_alt_cm;
 
+#if AP_PLANE_GLIDER_PULLUP_ENABLED
+    GliderPullup pullup;
```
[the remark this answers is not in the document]
tridge: moved

ArduPlane/ArduPlane.cpp
```diff
@@ -578,6 +591,16 @@ void Plane::update_alt()
         should_run_tecs = false;
     }
 #endif
+
+    if (auto_state.idle_mode) {
+        should_run_tecs = false;
```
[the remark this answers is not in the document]
tridge: another PR I think

ArduPlane/pullup.cpp
IamPete1: > I don't see any set_range(1) in master? it is still a uint16_t, not a float btw.

There are a few examples (admittedly all added by me):

https://github.com/ArduPilot/ardupilot/blob/d74050c5e328909dba2f3fe3a52825c8fbce7c1d/libraries/AP_ICEngine/AP_ICEngine.cpp#L206-L208

https://github.com/ArduPilot/ardupilot/blob/d74050c5e328909dba2f3fe3a52825c8fbce7c1d/libraries/AP_Motors/AP_MotorsHeli_Single.cpp#L217

The range value is not a float but the scaled value is. So you can have any integer endpoints with no loss of resolution. So we might as well standardize on 1.

ArduPlane/ArduPlane.cpp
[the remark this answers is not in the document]
IamPete1: The refactor could be another PR, but the thing I wanted to highlight was the miss match between `should_run_tecs` thresholds in the two cases.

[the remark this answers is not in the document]
tridge: yep, you're right, fixed to match

ArduPlane/Plane.h
```diff
@@ -515,6 +519,11 @@ class Plane : public AP_Vehicle {
         // have we checked for an auto-land?
         bool checked_for_autoland;
 
+        // Altitude threshold to complete a takeoff command in autonomous modes.  Centimeters
```
peterbarker, line 522: Incorrect comment?

ArduPlane/commands_logic.cpp
```diff
@@ -27,6 +27,9 @@ bool Plane::start_command(const AP_Mission::Mission_Command& cmd)
 
         nav_controller->set_data_is_stale();
 
+        // start non-idle
```
peterbarker, line 30: Bad comment?

IamPete1 (approved):
Looks good, nice new functionality.

henrywarhurst:
@tridge should this also turn the motor on once the MAV_CMD_NAV_ALTITUDE_WAIT is complete? I had a balloon drop flight test last week, but with a powered plane (not a glider) and the motor didn't seem to switch on despite the RCOU channel showing throttle going to 100%. Not a frequent ardupilot dev so I'm still figuring this out, but my test seems to indicate the motor indeed isn't spooling up:

```
    def NAV_ALTITUDE_WAIT_Throttle(self):
        '''test throttle engages after NAV_ALTITUDE_WAIT completes'''

        # Mission: NAV_ALTITUDE_WAIT (with threshold below ground so it triggers
        # immediately) followed by NAV_LOITER_UNLIM. This tests that throttle
        # properly engages after NAV_ALTITUDE_WAIT clears without a takeoff item.
        #
        # Starting location is ~585m AMSL, so setting threshold to 100m means
        # alt_diff = 585 - 100 = 485 > 0, triggering immediate completion.

        self.upload_simple_relhome_mission([
            self.create_MISSION_ITEM_INT(
                mavutil.mavlink.MAV_CMD_NAV_ALTITUDE_WAIT,
                p1=100,  # 100m AMSL - well below start altitude, triggers immediately
                p2=0,    # no descent rate trigger
                p3=0     # no wiggle
            ),
            (mavutil.mavlink.MAV_CMD_NAV_LOITER_UNLIM, 0, 0, 50),  # loiter at 50m AGL
        ])

        throttle_engaged = False
        last_print_time = [0]  # use list to allow modification in nested function

        def check_throttle_engaged(mav, m):
            nonlocal throttle_engaged
            if m.get_type() == 'SERVO_OUTPUT_RAW':
                # Print throttle value periodically (every 0.5 seconds)
                now = self.get_sim_time_cached()
                if now - last_print_time[0] >= 0.5:
                    self.progress("Throttle (servo3): %u" % m.servo3_raw)
                    last_print_time[0] = now
                # Check if throttle is above idle (1000 = 0%, anything above = throttle active)
                if m.servo3_raw > 1300:  # some margin above idle
                    throttle_engaged = True

        self.change_mode('AUTO')
        self.wait_ready_to_arm()
        self.arm_vehicle()

        # Install hook to monitor throttle
        self.context_push()
        self.install_message_hook_context(check_throttle_engaged)

        # Wait for mission to progress - NAV_ALTITUDE_WAIT should complete immediately
        # and takeoff should begin with throttle
        self.wait_current_waypoint(2, timeout=10)

        # Give some time for throttle to engage
        self.delay_sim_time(5)

        self.context_pop()

        if not throttle_engaged:
            raise NotAchievedException("Throttle did not engage after NAV_ALTITUDE_WAIT completed")

        self.progress("Throttle correctly engaged after NAV_ALTITUDE_WAIT")

        self.disarm_vehicle(force=True)
```

EDIT: Explicitly unsupressing throttle when `verify_altitude_wait` is about to clear seems to "fix" this issue but this probably isn't the right implementation. I think the throttle suppression code in servos.cpp should have handled this based on GPS speed, so I can't explain why my vehicle indicating throttle on RCOU

```
    if (completed) {
#if AP_PLANE_GLIDER_PULLUP_ENABLED
        if (pullup.pullup_start()) {
            // we are doing a pullup, ALTITUDE_WAIT not complete until pullup is done
            return false;
        }
#endif
        plane.throttle_suppressed = false;
        return true;
    }
  ```

Issue #27867, opened 2024-08-18
Compiling with cygwin broken!?


olliw42:
I'm trying to compile current master, and get a link error, the culprit seems to be `__wrap__malloc_r`:

![grafik](https://github.com/user-attachments/assets/7dfcd9a7-af89-49c4-a1e7-7d5fb9b305ed)

When I did last time, like 2-3 months back it was working all fine. 
This stuff really is above my paygrade, but it seems to me it is related to the changes in https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Common/c%2B%2B.cpp, which happened in these last months.
When I ensure that the code section in `#ifdef CYGWIN_BUILD` is not included, when it compiles and links fine ... but I sure have no idea if the resulting firmware would be working fine.

I'm on Win10 and for compiling I use something which calls itself "Cygwin64 Terminal". I must admit I don't recall really what this is (well, google leads to a page, but ...), just that I use it since long and it always worked great for me so far.

Any hint on what I should do in the code would be much appreciated :)

olliw42:
maybe these parts of the printout are of interest too

![grafik](https://github.com/user-attachments/assets/63694a10-becd-4268-a416-ea4bb3b7aefe)
![grafik](https://github.com/user-attachments/assets/f8d1b71e-a320-4278-88e7-43684c90240e)

yuri-rage:
IMHO, we should deprecate use of Cygwin entirely. WSL2 is a far better solution. Recommend using it.

I know that's not a "fix" for your current environment, but it is a much cleaner way of doing things overall.

rmackay9:
I agree that using WSL2 is a better solution than using cygwin.  The only downside is it is more difficult to access USB devices from WSL2 but that is also do-able using USB WSL and/or [USB WSL Gui](https://gitlab.com/alelec/wsl-usb-gui)

olliw42:
why then not a code change of the kind

```
#ifdef CYGWIN_BUILD
#error Cygwin is no longer supported.
#endif
```

instead of actual changes ... it's a bit of an obscure way for that purpose I would say ;)

my feeling though is that it only can be a minor code change which would correct the situation ...

robustini:
WSL2 is a far better solution but does not allow compilation of the SITL executable for Windows.

olliw42:
MANY THX
that's weired ... it's really just the changes I've mentioned

is this the same? I note that my count goes to [1200] while yours go to [1335]

[...]

PR #27987, opened 2024-09-02
AP_Common: fixed cygwin for non-SITL builds

tridge:
@olliw42 this may fix the issue in #27867


olliw42:
many thx
compiles fine, I also gave the compiled firmware a light test on the bench and looks good

PR #27988, opened 2024-09-03
Avoid using global "args" object within ChibiOSHWdef object

peterbarker:
... problematic for re-use of this object, which we do in various places


joshanne (approved):
LGTM

PR #27991, opened 2024-09-03
waf: move chibios hwdef generation to explicit call

joshanne:
Draft.

Not sure if this is wanted explicitly - I find it odd that we use the push this off to a subprocess call. Explicitly call the run itself.

This is built upon the work from #27955  which (now) builds upon #27988

Feel free to close if it's not wanted.


Tools/ardupilotwaf/chibios.py
```diff
@@ -624,17 +628,27 @@ def generate_hwdef_h(env):
...
+
+    # Build a list of the hardware definitions
+    hwdef_arr = [env.HWDEF]
     if env.HWDEF_EXTRA:
-        cmd += " '{0}'".format(env.HWDEF_EXTRA)
-    if env.BOOTLOADER_OPTION:
-        cmd += " " + env.BOOTLOADER_OPTION
-    return subprocess.call(cmd, shell=True)
+        # Add any extra hwdefs that may be defined
+        hwdef_arr.append(env.HWDEF_EXTRA)
```
joshanne, line 639: I think these should probably be `extend`, right?

Tools/ardupilotwaf/chibios.py
```diff
@@ -654,21 +668,18 @@ def pre_build(bld):
...
-            bld.env.AP_HAL_ROOT,
-            bld.env.BUILDROOT,
-            bld.env.default_parameters,
-            bld.env.HWDEF)
+    hwdef_arr = [bld.env.HWDEF]
     if bld.env.HWDEF_EXTRA:
-        hwdef_rule += " " + bld.env.HWDEF_EXTRA
-    if bld.env.BOOTLOADER_OPTION:
-        hwdef_rule += " " + bld.env.BOOTLOADER_OPTION
+        hwdef_arr.append(bld.env.HWDEF_EXTRA)
```
joshanne, line 673: I think this should also be `extend` right?

Tools/ardupilotwaf/chibios.py
peterbarker: Same-same; one takes a list, the other  a single item. 

So yes - iff `env.HWDEF_EXTRA` is a list...

joshanne: I believe present behaviour in `chibios.py` is a single extra hwdef is provided to `chibios_hwdef.py`.

However, if I called `chibios_hwdef.py` directly, I could provide multiple hwdefs - they are received as a list and it is capable of setting multiple extra hwdefs.

That said, I'm not attempting to change behaviours here. This implementation is current behaviour.

joshanne:
> I do like this.

Excellent, then i'll move this from draft to PR.

Tools/ardupilotwaf/chibios.py
```diff
@@ -624,17 +628,27 @@ def generate_hwdef_h(env):
...
     if env.HWDEF_EXTRA:
-        cmd += " '{0}'".format(env.HWDEF_EXTRA)
-    if env.BOOTLOADER_OPTION:
-        cmd += " " + env.BOOTLOADER_OPTION
-    return subprocess.call(cmd, shell=True)
+        # Add any extra hwdefs that may be defined
+        hwdef_arr.append(env.HWDEF_EXTRA)
+    ch = chibios_hwdef.ChibiOSHWDef(outdir=hwdef_out,
+                                    bootloader=env.BOOTLOADER,
+                                    signed_fw=env.AP_SIGNED_FIRMWARE,
```
joshanne, line 642: `ChibiOSHWDef` defines `signed_fw` as bool.

If waf is called with `--signed-fw` then `env.AP_SIGNED_FIRMWARE` is set to `True`.
If waf is called without `signed-fw` then `env.AP_SIGNED_FIRMWARE` is set to `[ ]` (empty list)

Given python considers empty list as False, this might be safe. The last commit changes `wscript` treat `AP_SIGNED_FIRMWARE` similarly to `BOOTLOADER`, setting the env flag based on what is passed to it at build time.

Tools/ardupilotwaf/chibios.py
joshanne: Scratch that.

Tools/ardupilotwaf/chibios.py
```diff
@@ -18,6 +18,10 @@
 import base64
 import subprocess
 
+# modify our search path:
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../libraries/AP_HAL_ChibiOS/hwdef/scripts'))
+import chibios_hwdef
```
peterbarker, line 23: Since this was written we sprouted hwdefs for the Linux HAL.  It does what you're doing in here, and it does work well!

Tools/ardupilotwaf/chibios.py
```diff
@@ -640,17 +644,29 @@ def generate_hwdef_h(env):
...
+        bootloader=env.BOOTLOADER,
+        signed_fw=env.AP_SIGNED_FIRMWARE,
+        default_params_filepath=env.DEFAULT_PARAMETERS,
+        hwdef=hwdef_arr,
+    )
+    try:
+        ch.run()
+        # Return success for subprocess call
+        return 0
+    except SystemExit:
```
peterbarker: I don't think that tool should ever SystemExit?!

Tools/ardupilotwaf/chibios.py
```diff
@@ -640,17 +644,29 @@ def generate_hwdef_h(env):
...
+        # The run failed, so return the error code
+        return 1
```
peterbarker: ... and if it did, is the comment wrong, subtly or otherwise?

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -2725,12 +2725,14 @@ def get_processed_defaults_file(self, defaults_filepath, depth=0):
     def write_processed_defaults_file(self, filepath):
         # see if board has a defaults.parm file or a --default-parameters file was specified
         defaults_filename = os.path.join(os.path.dirname(self.hwdef[0]), 'defaults.parm')
-        defaults_path = os.path.join(os.path.dirname(self.hwdef[0]), args.params)
-
         defaults_abspath = None
-        if os.path.exists(defaults_path):
-            defaults_abspath = os.path.abspath(self.default_params_filepath)
-            self.progress("Default parameters path from command line: %s" % self.default_params_filepath)
+        if isinstance(self.default_params_filepath, str):
```
peterbarker, line 2729: This change wasn't in the PR title or called out as a change.  What's it for?

peterbarker (commented):
This was conflicting, so I rebased it on master.

I reformatted the construction of the chibios object a little (whitespace changes)

Tools/ardupilotwaf/chibios.py
joshanne: I cannot for the life of me remember what the reasoning behind this was - possibly attempting to keep some existing behaviour from when it was subprocess call?

Nevertheless, removed and cleaned up.

Tools/ardupilotwaf/chibios.py
joshanne: as above.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
joshanne: The reason for this change is about what happens when `write_processed_defaults_file` runs. When this was a subprocess call, this would result in being passed an empty string.

Now this is is either of the two options, depending if `--default-parameters` is passed in to `waf`:
* Configuring a hwdef with no additional default parameters specified results in `default_params_filepath` being set as empty array (`[]`).
  * eg: `./waf configure --board CubeOrangePlus`
* Configuring a hwdef with a default parameter results in `default_params_filepath` being set as a string.
  * eg: `./waf configure --board CubeOrangePlus --default-parameters=./defaults.txt`

Happy to take advice if there's a better way to handle it. If all it takes is an update to a commit message, I can do that too.

The linux hwdef generator doesn't appear to make use of default parameters.

peterbarker:
Gah!  My apologies, I honestly thought we had merged this one!

We now have this functionality in master via ea8f817affdd9d0b48af53185580ebe604b60b8d

Closing this one now - thanks!

PR #27996, opened 2024-09-03
Pr rate log

andyp1per:


lthall (approved):
Great change but I would like to do the same thing with the ATT message.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -1240,3 +1242,52 @@ void AC_AttitudeControl::get_rpy_srate(float &roll_srate, float &pitch_srate, fl
     pitch_srate = get_rate_pitch_pid().get_pid_info().slew_rate;
     yaw_srate = get_rate_yaw_pid().get_pid_info().slew_rate;
 }
+
+#if HAL_LOGGING_ENABLED
+// Write a rate packet
+void AC_AttitudeControl::Write_Rate(const AP_Motors &motors, const AC_PosControl &pos_control) const
```
IamPete1, line 1248: `AC_AttitudeControl` already has a local `_motors` reference. No need to pass it in.

libraries/AC_AttitudeControl/AC_AttitudeControl.h
```diff
@@ -431,6 +431,16 @@ class AC_AttitudeControl {
     // get the value of the PD scale that was used in the last loop, for logging
     const Vector3f &get_PD_scale_logging(void) const { return _pd_scale_used; }
 
+    // get the latest gyro value that was used by the rate controller
+    const Vector3f &get_gyro_latest(void) const { return _rate_gyro; }
+
+    // get the timestamp of the latest gyro value that was used by the rate controller
+    // primarily used by logging
+    uint64_t get_gyro_latest_timestamp_us(void) const { return _rate_gyro_time_us; }
```
IamPete1, line 439: Either don't need these anymore, or make them private.

libraries/AC_AttitudeControl/AC_AttitudeControl.h
```diff
@@ -431,6 +431,16 @@ class AC_AttitudeControl {
     // get the value of the PD scale that was used in the last loop, for logging
     const Vector3f &get_PD_scale_logging(void) const { return _pd_scale_used; }
```
IamPete1, line 432: Like wise this can be removed or private now.

IamPete1 (commented):
Should be a nice clean up.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
andyp1per: nice catch, fixed

libraries/AC_AttitudeControl/AC_AttitudeControl.h
andyp1per: removed

libraries/AC_AttitudeControl/AC_AttitudeControl.h
andyp1per: removed

IamPete1 (approved):
I think this looks good. The only question is if any log tools have a problem with the log timestamp not matching the time the log was written. I don't think they should. Each message should still see a increasing time. Only between messages would time go backwards. The only other thing would be if we were to write the log twice in one loop for some reason then we would get to timestamps the same. 

I guess we could get round this by adding a new timestamp rather than replacing the existing one, but that would make plotting in log review tools hard as there mostly hard coded to use `TimeUS` on the x axis.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -1240,3 +1242,51 @@ void AC_AttitudeControl::get_rpy_srate(float &roll_srate, float &pitch_srate, fl
...
     yaw_srate = get_rate_yaw_pid().get_pid_info().slew_rate;
 }
+
+#if HAL_LOGGING_ENABLED
+// Write a rate packet
+void AC_AttitudeControl::Write_Rate(const AC_PosControl &pos_control) const
+{
+    Vector3f rate_targets = rate_bf_targets();
+    const Vector3f &accel_target = pos_control.get_accel_target_cmss();
+    const Vector3f gyro_rate = _rate_gyro;
```
peterbarker, line 1252: This is taking a copy for no good reason?

libraries/AC_AttitudeControl/LogStructure.h
```diff
@@ -68,4 +102,6 @@ struct PACKED log_PSCx {
     { LOG_PSCE_MSG, sizeof(log_PSCx), \
       "PSCE", PSCx_FMT, "TimeUS,TPE,PE,DVE,TVE,VE,DAE,TAE,AE", PSCx_UNITS, PSCx_MULTS }, \
     { LOG_PSCD_MSG, sizeof(log_PSCx), \
-      "PSCD", PSCx_FMT, "TimeUS,TPD,PD,DVD,TVD,VD,DAD,TAD,AD", PSCx_UNITS, PSCx_MULTS }
+      "PSCD", PSCx_FMT, "TimeUS,TPD,PD,DVD,TVD,VD,DAD,TAD,AD", PSCx_UNITS, PSCx_MULTS }, \
+    { LOG_RATE_MSG, sizeof(log_Rate), \
+        "RATE", "Qfffffffffffff",  "TimeUS,RDes,R,ROut,PDes,P,POut,YDes,Y,YOut,ADes,A,AOut,AOutSlew", "skk-kk-kk-oo--", "F?????????BB--" , true }
```
peterbarker, line 107: These question-marks should be fixed before merge

rmackay9 (commented):
LGTM, thanks

As discussed on the call, a few of us are worried about the TimeUS potentially going backwards which could confuse some log analysis tools but I think it's worth the risk and it seems important that we give the devs the data they need

peterbarker (changes requested):

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
andyp1per: Fixed

peterbarker (approved):

PR #28000, opened 2024-09-04
hwdef: remove SMBus battery support from all minimize-fpv boards

peterbarker:
I doubt there's a single one of these boards on the planet using stock firmware and smbus batteries.  If they exist, the custom build server can re-add this


MattKear (approved):
Yeah, seems sensible to me.

Hwurzburg (approved):
I'm in

peterbarker:
3.5kB, BTW

Hwurzburg:
yikes!   bet there are still one or two more like this if we look

PR #28002, opened 2024-09-04
AP_Mission: stop populating command with -1 on failure

peterbarker:
Just indicate in the comments that the contents are undefined after this call.

Nothing appears to be relying on this.  The method's nullable, so i don't think scripting cares, and the mavlink ftp transfer stuff doesn't look.


peterbarker:
We went with the alternative

PR #28005, opened 2024-09-04
Use sample time in Attitude logging

andyp1per:
When interpreting rate and attitude control logs it is important to be able to correlate the log message with when the data was captured rather than when the log was written. This is especially important as we increase the frequency of the rate loop and make it asynchronous. This small change adds the ability to capture the sample time from the scheduler and then uses that in the attitude log message.

Split from https://github.com/ArduPilot/ardupilot/pull/27029

### DIFF
--- a/libraries/AP_AHRS/AP_AHRS_Logging.cpp
+++ b/libraries/AP_AHRS/AP_AHRS_Logging.cpp
@@ -4,6 +4,7 @@
 
 #include "AP_AHRS.h"
 #include <AP_Logger/AP_Logger.h>
+#include <AP_Scheduler/AP_Scheduler.h>
 
 #include <AC_AttitudeControl/AC_AttitudeControl.h>
 #include <AC_AttitudeControl/AC_PosControl.h>
@@ -53,7 +54,7 @@ void AP_AHRS::Write_Attitude(const Vector3f &targets) const
 {
     const struct log_Attitude pkt{
         LOG_PACKET_HEADER_INIT(LOG_ATTITUDE_MSG),
-        time_us         : AP_HAL::micros64(),
+        time_us         : AP::scheduler().get_loop_start_time_us(),
         control_roll    : (int16_t)targets.x,
         roll            : (int16_t)roll_sensor,
         control_pitch   : (int16_t)targets.y,
@@ -61,6 +62,7 @@ void AP_AHRS::Write_Attitude(const Vector3f &targets) const
         control_yaw     : (uint16_t)wrap_360_cd(targets.z),
         yaw             : (uint16_t)wrap_360_cd(yaw_sensor),
         active          : uint8_t(active_EKF_type()),
+        sensor_dt       : AP::scheduler().get_last_loop_time_s()
     };
     AP::logger().WriteBlock(&pkt, sizeof(pkt));
 }
@@ -127,7 +129,7 @@ void AP_AHRS_View::Write_AttitudeView(const Vector3f &targets) const
 {
     const struct log_Attitude pkt{
         LOG_PACKET_HEADER_INIT(LOG_ATTITUDE_MSG),
-        time_us         : AP_HAL::micros64(),
+        time_us         : AP::scheduler().get_loop_start_time_us(),
         control_roll    : (int16_t)targets.x,
         roll            : (int16_t)roll_sensor,
         control_pitch   : (int16_t)targets.y,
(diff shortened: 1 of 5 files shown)


peterbarker:
Pretty sure we used to do this, and it allowed time to go backwards in logs, and *that* was painful when you were trying to compare messages...

Maybe an extra field?

andyp1per:
> Pretty sure we used to do this, and it allowed time to go backwards in logs, and _that_ was painful when you were trying to compare messages...
> 
> Maybe an extra field?

Do we guarantee that logs are monotonically increasing? I guess if we do then an extra field would be the way, but that will make log analysis of these messages harder

lthall:
> Pretty sure we used to do this, and it allowed time to go backwards in logs, and _that_ was painful when you were trying to compare messages...
> 
> Maybe an extra field?

I don't recall this but we have line numbers and a time stamp. I don't see why we would have a time stamp associated with the time the log message was communicated to the logging thread rather than the time the data in the message was relevant.

libraries/AP_AHRS/LogStructure.h
```diff
@@ -72,6 +73,7 @@ struct PACKED log_Attitude {
     uint16_t error_rp;
     uint16_t error_yaw;
     uint8_t  active;
+    float sensor_dt;
```
IamPete1, line 76: Would this be better in RATE?

IamPete1:
Will need a rework after https://github.com/ArduPilot/ardupilot/pull/28026 has been merged.

libraries/AC_AttitudeControl/AC_AttitudeControl.h
```diff
@@ -489,6 +494,8 @@ class AC_AttitudeControl {
 
     // Intersampling period in seconds
     float               _dt;
+    // scheduler sample time in microseconds (could be different to rate loop)
+    uint64_t            _sample_time_us;
```
IamPete1, line 498: I don't understand. This is just set and the pulled from here with no changes. The comment say this is because it would be changed in the future in the rate loop. But https://github.com/ArduPilot/ardupilot/pull/28005#discussion_r1749660960 you say that the dt is not associated with the rate loop at all? So why do we need to set this if the rate loop will not be changing it? The log method could get it direct from the scheduler?

andyp1per: Yes, I was thinking this would be a better simplification - would avoid changing the API

andyp1per: There are two dt's going on - one for the rate control and one for the attitude control. Currently they are the same, but with a separate rate loop they will be different. The atttitude logging should log the attitude dt, the rate logging should log the rate dt (if necessary). There are also two samples / times going on - one for the atttitude controller and one for the rate controller. Also currently the same, but soon to be different. Again the attitude logging should log the attitude one / time. But I do agree that this could simplified by the logging method just directly calling into the scheduler - I will change it to do that.

libraries/AP_AHRS/AP_AHRS_View.h
```diff
@@ -173,7 +173,7 @@ class AP_AHRS_View
     }
 
     // Logging Functions
-    void Write_AttitudeView(const Vector3f &targets) const;
```
peterbarker, line 176: random whitespace change

rmackay9 (approved):

libraries/AP_AHRS/AP_AHRS_Logging.cpp
```diff
@@ -53,14 +54,15 @@ void AP_AHRS::Write_Attitude(const Vector3f &targets) const
 {
     const struct log_Attitude pkt{
         LOG_PACKET_HEADER_INIT(LOG_ATTITUDE_MSG),
-        time_us         : AP_HAL::micros64(),
+        time_us         : AP::scheduler().get_loop_start_time_us(),
```
peterbarker, line 57: This introduces a coupling between AHRS and the scheduler.

We merged a change to use the sample time in the rate message - this uses a different pattern where we don't pass that timestamp in...

peterbarker:
Resolution at DevCall is that we'll create an `ANG` message!

Which I think means we want to close this!

tridge:
we decided to create an ANG message in AC_AttitudeControl instead

peterbarker:
Closing this one as we're going with `ANG`.  Reopen if we were wrong, please.

PR #28006, opened 2024-09-04
Tools: param_metadata: do not emit Legacy fields to rst/Wiki

peterbarker:
Allows a marker to be added to parameter markup so that it isn't emitted to the RST, but is to everywhere else.

In lieu of doing proper parameter versioning on the build server...


peterbarker:
Tested with `diff -ur` on master / new generation.

Hwurzburg (approved):
Super! will add the rest of GPS and ICE params in a PR today if the "creek dont rise"

PR #28008, opened 2024-09-04
Log attitude information at full resolution

andyp1per:
We are artificially quantizing the attitude information in the attitude log message. This makes it very hard to relate to the rate logging which is logged at full resolution. There is really no reason to quantize like this other than to save a little bit of log bandwidth.

Split from https://github.com/ArduPilot/ardupilot/pull/27029

### DIFF
--- a/AntennaTracker/Log.cpp
+++ b/AntennaTracker/Log.cpp
@@ -7,9 +7,7 @@
 // Write an attitude packet
 void Tracker::Log_Write_Attitude()
 {
-    Vector3f targets;
-    targets.y = nav_status.pitch * 100.0f;
-    targets.z = wrap_360_cd(nav_status.bearing * 100.0f);
+    const Vector3f targets{0.0f, nav_status.pitch, nav_status.bearing};
     ahrs.Write_Attitude(targets);
     AP::ahrs().Log_Write();
 }
--- a/ArduCopter/Log.cpp
+++ b/ArduCopter/Log.cpp
@@ -75,9 +75,7 @@ void Copter::Log_Write_Control_Tuning()
 // Write an attitude packet
 void Copter::Log_Write_Attitude()
 {
-    Vector3f targets = attitude_control->get_att_target_euler_cd();
-    targets.z = wrap_360_cd(targets.z);
-    ahrs.Write_Attitude(targets);
+    ahrs.Write_Attitude(attitude_control->get_att_target_euler_rad() * RAD_TO_DEG);
     ahrs_view->Write_Rate(*motors, *attitude_control, *pos_control);
  }
 
--- a/ArduPlane/Log.cpp
+++ b/ArduPlane/Log.cpp
@@ -5,10 +5,11 @@
 // Write an attitude packet
 void Plane::Log_Write_Attitude(void)
 {
-    Vector3f targets;       // Package up the targets into a vector for commonality with Copter usage of Log_Wrote_Attitude
-    targets.x = nav_roll_cd;
-    targets.y = nav_pitch_cd;
-    targets.z = 0; //Plane does not have the concept of navyaw. This is a placeholder.
+    Vector3f targets {       // Package up the targets into a vector for commonality with Copter usage of Log_Wrote_Attitude
+        nav_roll_cd * 0.01f,
+        nav_pitch_cd * 0.01f,
+        0 //Plane does not have the concept of navyaw. This is a placeholder.
+    };
 
 #if HAL_QUADPLANE_ENABLED
     if (quadplane.show_vtol_view()) {
@@ -18,8 +19,7 @@ void Plane::Log_Write_Attitude(void)
         // since Euler angles are not used and it is a waste of cpu to compute them at the loop rate.
         // Get them from the quaternion instead:
         quadplane.attitude_control->get_attitude_target_quat().to_euler(targets.x, targets.y, targets.z);
-        targets *= degrees(100.0f);
-        quadplane.ahrs_view->Write_AttitudeView(targets);
+        quadplane.ahrs_view->Write_AttitudeView(targets * RAD_TO_DEG);
     } else
 #endif
             {
--- a/Rover/Log.cpp
+++ b/Rover/Log.cpp
@@ -7,8 +7,8 @@
 // Write an attitude packet
 void Rover::Log_Write_Attitude()
 {
-    float desired_pitch_cd = degrees(g2.attitude_control.get_desired_pitch()) * 100.0f;
-    const Vector3f targets(0.0f, desired_pitch_cd, 0.0f);
+    float desired_pitch = degrees(g2.attitude_control.get_desired_pitch());
+    const Vector3f targets(0.0f, desired_pitch, 0.0f);
 
     ahrs.Write_Attitude(targets);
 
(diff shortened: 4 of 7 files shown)


libraries/AP_AHRS/LogStructure.h
```diff
@@ -199,7 +199,7 @@ struct PACKED log_ATSC {
     { LOG_AOA_SSA_MSG, sizeof(log_AOA_SSA), \
         "AOA", "Qff", "TimeUS,AOA,SSA", "sdd", "F00" , true }, \
     { LOG_ATTITUDE_MSG, sizeof(log_Attitude),\
-        "ATT", "QccccCCCCB", "TimeUS,DesRoll,Roll,DesPitch,Pitch,DesYaw,Yaw,ErrRP,ErrYaw,AEKF", "sddddhhdh-", "FBBBBBBBB-" , true }, \
+        "ATT", "QffffffffB", "TimeUS,DesRoll,Roll,DesPitch,Pitch,DesYaw,Yaw,ErrRP,ErrYaw,AEKF", "sddddhhdh-", "FBBBBBBBB-" , true }, \
```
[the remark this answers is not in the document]
andyp1per: good point, we could probably just change the scaling to degrees if we make this change - the source is not in centi-degrees.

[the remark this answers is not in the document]
andyp1per: Ok fixed. There is a question as to whether you want to log the targets as degrees as well with this. Its more invasive as the callers have to be changed.

peterbarker:
Need to divide the targets:
image

peterbarker:
> We are artificially quantizing the attitude information in the attitude log message. This makes it very hard to relate to the rate logging which is logged at full resolution. There is really no reason to quantize like this other than to save a little bit of log bandwidth.

"a little bit of log bandwidth" sounds a little handy-wavey.  Roughly speaking this will double the size of the message.

Which will move its proportion of a log from maybe 0.75% to 1.5%, based on the Plane.AUTOTEST log I'm looking at here.

peterbarker (changes requested):
Need to scale targets

libraries/AP_AHRS/AP_AHRS_Logging.cpp
```diff
@@ -54,14 +54,14 @@ void AP_AHRS::Write_Attitude(const Vector3f &targets) const
...
-        yaw             : (uint16_t)wrap_360_cd(yaw_sensor),
-        error_rp        : (uint16_t)(get_error_rp() * 100),
-        error_yaw       : (uint16_t)(get_error_yaw() * 100),
+        control_roll    : targets.x,
+        roll            : degrees(roll),
+        control_pitch   : targets.y,
+        pitch           : degrees(pitch),
+        control_yaw     : wrap_360_cd(targets.z),
+        yaw             : wrap_360(degrees(yaw)),
+        error_rp        : get_error_rp(),
```
lthall, line 63: These last three lines are really not appropriate for this message as this is supposed to be the logging of the attitude controller.   I would like to remove the last three lines and move this logging to the AC_AttitudeControl library. This would let us optimise this log entry for what it is supposed to be for, monitoring the Attitude Control.

```
 error_rp        : get_error_rp(),
 error_yaw       : get_error_yaw(),
active          : uint8_t(active_EKF_type()),
```

lthall (changes requested):

andyp1per:
> Need to divide the targets:

Fixed

AntennaTracker/Log.cpp
```diff
@@ -7,9 +7,7 @@
 // Write an attitude packet
 void Tracker::Log_Write_Attitude()
 {
-    Vector3f targets;
-    targets.y = nav_status.pitch * 100.0f;
-    targets.z = wrap_360_cd(nav_status.bearing * 100.0f);
+    Vector3f targets{0.0f, nav_status.pitch, nav_status.bearing};
```
peterbarker, line 10:
```suggestion
    const Vector3f targets{0.0f, nav_status.pitch, nav_status.bearing};
```

ArduCopter/Log.cpp
```diff
@@ -75,9 +75,7 @@ void Copter::Log_Write_Control_Tuning()
 // Write an attitude packet
 void Copter::Log_Write_Attitude()
 {
-    Vector3f targets = attitude_control->get_att_target_euler_cd();
-    targets.z = wrap_360_cd(targets.z);
-    ahrs.Write_Attitude(targets);
+    ahrs.Write_Attitude(attitude_control->get_att_target_euler_rad());
```
peterbarker, line 78: You appear to be passing the targets in in radians in several places here now.

Pretty sure you need to convert to degrees here somewhere...

You can multiply the vector by `RAD_TO_DEG`...

peterbarker (changes requested):
Fewer cd  is good!

ArduCopter/Log.cpp
andyp1per: Thanks, fixed

ArduPlane/Log.cpp
```diff
@@ -6,8 +6,8 @@
 void Plane::Log_Write_Attitude(void)
 {
     Vector3f targets;       // Package up the targets into a vector for commonality with Copter usage of Log_Wrote_Attitude
```
peterbarker, line 8: Tidy this up by 
```
const Vector3f targets {
    nav_roll_cd * 0.01,
    nav_pitch_cd * 0.01,
    0
};
```

?

ArduCopter/Log.cpp
peterbarker: OK, so you've fixed this in `Write_Attitude`.

Two issues there.

tridge probably won't like radians in the interface.

You haven't changed the parameter names to add `_rad` on the end

Rover/Log.cpp
```diff
@@ -7,8 +7,8 @@
 // Write an attitude packet
 void Rover::Log_Write_Attitude()
 {
-    float desired_pitch_cd = degrees(g2.attitude_control.get_desired_pitch()) * 100.0f;
-    const Vector3f targets(0.0f, desired_pitch_cd, 0.0f);
+    float desired_pitch = degrees(g2.attitude_control.get_desired_pitch());
```
peterbarker, line 10: This is now a double-conversion to degrees.

libraries/AP_AHRS/AP_AHRS_Logging.cpp
peterbarker: https://github.com/ArduPilot/ardupilot/pull/28026

peterbarker (changes requested):
I do like the change to log in floats.

But I keep finding these trivial cases where we are breaking `ATT`, one of the most important logging messages we have.  *Please* check the log message is correct for all vehicles after your changes.

ArduCopter/Log.cpp
andyp1per: Fixed

Rover/Log.cpp
andyp1per: Fixed

peterbarker (approved):
LGTM.

Please state what testing you've given the various vehicles.

ArduPlane/Log.cpp
andyp1per: Fixed

andyp1per:
I have eyeballed logs from vehicles to check that desired and actual line up.

tridge:
@andyp1per with the decision to add an ANG message, do you still want this change?

rmackay9:
I like this change

rmackay9 (approved):

Issue #20995, opened 2022-06-17
Copter: unable to arm when throttle is maximized in GUIDED mode


muramura:
## Bug report

**Issue details**

I did an ARM from a QGC in GUIDED mode with the throttle of the radio on top.
I could not arm with "PreArm: Throttle (RC3) is not neutral".
I expect it to ignore the radio input except for the flight mode channel.
I can transition from GUIDED mode to other modes by changing the flight mode channel.

**Version**

ArduCopter V4.3.0-dev (e9942f9a)

**Platform**
[  ] All
[  ] AntennaTracker
[X] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**

QUAD COPTER

**Hardware type**

SITL]

**Logs**

![Screenshot from 2022-06-18 01-29-16](https://user-images.githubusercontent.com/646194/174342930-bba81c51-7c79-4dc4-998e-9e5d424f3018.png)

muramura:
I would add a decision not to check the radio input if the arm is by MAVLINK.
With this addition I could arm ignoring the propo stick.

![Screenshot from 2022-06-18 08-28-23](https://user-images.githubusercontent.com/646194/174412876-0ba5ccce-afdc-4203-9676-add49c5e0b62.png)

rmackay9:
We have a related (perhaps duplicate?) issue here https://github.com/ArduPilot/ardupilot/issues/21415

peterbarker:
We have a bit in RC_OPTIONS that handles this now.

PR #20999, opened 2022-06-18
AP_Arming: MAVLINK arm ignores radio stick operation

muramura:
I resolve #20995 .

I checked with SITL.
I maxed out the throttle and issued ARM instructions from the QGC.
I was able to confirm that it would arm.

PR #28009, opened 2024-09-04
Copter: remove throttle check for arming in guided

peterbarker:
Co-authored-by: muramura <ma2maru@gmail.com>

.... but add a guided option to put it back.

Replaces https://github.com/ArduPilot/ardupilot/pull/20999
Closes https://github.com/ArduPilot/ardupilot/issues/20995

We could ignore throttle unconditionally instead of adding the option bit?

New autotest fails before / passes after.

### DIFF
--- a/ArduCopter/mode.h
+++ b/ArduCopter/mode.h
@@ -121,6 +121,8 @@ public:
     virtual bool requires_GPS() const = 0;
     virtual bool has_manual_throttle() const = 0;
     virtual bool allows_arming(AP_Arming::Method method) const = 0;
+    virtual bool arming_check_throttle() const { return true; }
+
     virtual bool is_autopilot() const { return false; }
     virtual bool has_user_takeoff(bool must_navigate) const { return false; }
     virtual bool in_guided_mode() const { return false; }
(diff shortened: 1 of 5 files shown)


ArduCopter/mode.h
```diff
@@ -1140,7 +1144,8 @@ class ModeGuided : public Mode {
         DoNotStabilizePositionXY = (1U << 4),
         DoNotStabilizeVelocityXY = (1U << 5),
         WPNavUsedForPosControl = (1U << 6),
-        AllowWeatherVaning = (1U << 7)
+        AllowWeatherVaning = (1U << 7),
+        ArmingCheckThrottle = (1U << 8),
```
IamPete1, line 1148: Needs param description.

IamPete1:
Is there a case where we could be waiting for the pilot to raise throttle in a takeoff in guided mode?

ArduCopter/mode.h
peterbarker: Fixed, thanks.

We really should add a sanity check for that.

rmackay9:
We discussed this on the dev call and we think there is already an RC_OPTIONS that provides the same safety check

PR #28010, opened 2024-09-04
Remove roll_sensor/pitch_sensor/yaw_sensor from AP_AHRS

peterbarker:
Replace with `get_roll_rad` and get_roll_deg`, both floats.


peterbarker:
I've simplified this PR to simply add the new methods to the interface, not touching the existing creation of the yaw_sensor variables etc.

I'm also not renaming (eg.) `get_roll` to `get_roll_rad` in this PR now.  This rename costs us 408 bytes on CubeOrange because scripting needs the old binding to exist or we instantly break way too many scripts.  We also can't deprecate an alias, it seems.  We might still choose to do this in the future (I have the patches :-), but not in this PR.

peterbarker:
```
CubeRedPrimary                      -48    *           -136    -152              -16    -56    -48
Durandal                            -24    *           -40     -64               -32    -88    48
KakuteH7-bdshot                     -24    *           -120    -120              80     -80    -184
MatekF405                           -8     *           -32     -24               -96    -8     -24
Pixhawk1-1M-bdshot                  32                 -16     -24               -72    -8     32
revo-mini                           -16    *           -80     -64               -80    -96    -32
skyviper-journey                                       -80                                     
skyviper-v2450                                         -8                                      
```

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -7390,7 +7388,7 @@ void GCS_MAVLINK::send_high_latency2() const
         global_position_current.lng, // [degE7] Longitude
         global_position_current.alt * 0.01f, // [m] Altitude above mean sea level
         high_latency_target_altitude(), // [m] Altitude setpoint
-        (((uint16_t)ahrs.yaw_sensor / 100) % 360) / 2, // [deg/2] Heading
+        uint16_t(ahrs.get_yaw_deg()) / 2, // [deg/2] Heading
```
peterbarker, line 7391: This should probably be changed to do floating-point division and rounding?  Pre-existing problem.

peterbarker:
.... there's a bug in toy_mode.cpp line 294 which we should fix once this is in.

libraries/AP_AHRS/AP_AHRS.h
```diff
@@ -563,6 +563,11 @@ class AP_AHRS {
     float get_pitch() const { return pitch; }
     float get_yaw() const { return yaw; }
 
+    // roll/pitch/yaw euler angles, all in radians
```
clydemcqueen, line 566: Radians?

peterbarker: Fixed, thanks :-)

IamPete1 (approved):
LGTM

tridge (approved):

PR #28011, opened 2024-09-04
Legacy tags

Hwurzburg:
not sure the Movingbaseline ones are correct

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -2008,22 +2008,26 @@ bool AP_GPS::gps_yaw_deg(uint8_t instance, float &yaw_deg, float &accuracy_deg, 
     // @CopyFieldsFrom: GPS_TYPE
     // @DisplayName: 2nd GPS type.Renamed in 4.6 to GPS2_TYPE
     // @Description: GPS type of 2nd GPS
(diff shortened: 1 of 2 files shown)


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -2007,22 +2007,26 @@ bool AP_GPS::gps_yaw_deg(uint8_t instance, float &yaw_deg, float &accuracy_deg,
     // @CopyFieldsFrom: GPS_TYPE
     // @DisplayName: 2nd GPS type.Renamed in 4.6 to GPS2_TYPE
     // @Description: GPS type of 2nd GPS
+    // @Legacy 4.5 param
```
joshanne, line 2010: Assuming this is based on the work from #28006 ...

https://github.com/ArduPilot/ardupilot/blob/e1fe73e731805cbab357be01680f7910c3610e0f/libraries/AP_GPS/AP_GPS.cpp#L2005

All the fields have a colon after the name.

Should these not be:

```suggestion
    // @Legacy: 4.5 param
```

and in other places also.

Hwurzburg: right...

peterbarker (approved):
LGTM

Hwurzburg:
thanks

PR #28014, opened 2024-09-05
Plane: allow for different orientations for landing rangefinder

tridge:
this is principally for tailsitters where rangefinders would be orientation with RNGFND1_ORIENT=12 (PITCH_180), but also allows for custom orientations which will be useful if the rangefinder is tilted forward

Tested in SITL using the quadplane-copter_tailsitter frame

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1274,6 +1274,15 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     AP_SUBGROUPINFO(precland, "PLND_", 35, ParametersG2, AC_PrecLand),
 #endif
 
+#if AP_RANGEFINDER_ENABLED
+    // @Param: RNGFND_LND_ORNT
+    // @DisplayName: rangefinder landing orientation
+    // @Description: The orientation of rangefinder to use for landing detection. Should be set to rangefinder_orientation() for normal downward facing rangefinder and ROTATION_PITCH_180 for rearward facing rangefinder for quadplane tailsitters. Custom orientation can be used with CUST_ROT_ENABLE. The orientation must match at least one of the available rangefinders.
+    // @Values: 12:Back, 25:Down, 101:Custom1, 102:Custom2
+    // @User: Standard
+    AP_GROUPINFO("RNGFND_LND_ORNT", 36, ParametersG2, rangefinder_land_orient, ROTATION_PITCH_270),
+#endif
+    
     AP_GROUPEND
 };
 
--- a/ArduPlane/altitude.cpp
+++ b/ArduPlane/altitude.cpp
@@ -138,7 +138,7 @@ float Plane::relative_ground_altitude(bool use_rangefinder_if_available, bool us
 
 #if HAL_QUADPLANE_ENABLED && AP_RANGEFINDER_ENABLED
    if (use_rangefinder_if_available && quadplane.in_vtol_land_final() &&
-       rangefinder.status_orient(ROTATION_PITCH_270) == RangeFinder::Status::OutOfRangeLow) {
+       rangefinder.status_orient(rangefinder_orientation()) == RangeFinder::Status::OutOfRangeLow) {
        // a special case for quadplane landing when rangefinder goes
        // below minimum. Consider our height above ground to be zero
        return 0;
@@ -679,16 +679,36 @@ void Plane::rangefinder_terrain_correction(float &height)
  */
 void Plane::rangefinder_height_update(void)
 {
-    float distance = rangefinder.distance_orient(ROTATION_PITCH_270);
-    
-    if ((rangefinder.status_orient(ROTATION_PITCH_270) == RangeFinder::Status::Good) && ahrs.home_is_set()) {
+    const auto orientation = rangefinder_orientation();
+    bool range_ok = rangefinder.status_orient(orientation) == RangeFinder::Status::Good;
+    float distance = rangefinder.distance_orient(orientation);
+    float corrected_distance = distance;
+
+    /*
+      correct distance for attitude
+     */
+    if (range_ok) {
+        // correct the range for attitude
+        const auto &dcm = ahrs.get_rotation_body_to_ned();
+
+        Vector3f v{1, 0, 0};
+        v.rotate(orientation);
+        v = dcm * v;
+
+        if (!is_positive(v.z)) {
+            // not pointing at the ground
+            range_ok = false;
+        } else {
+            corrected_distance *= v.z;
+        }
+    }
+
+    if (range_ok && ahrs.home_is_set()) {
         if (!rangefinder_state.have_initial_reading) {
             rangefinder_state.have_initial_reading = true;
             rangefinder_state.initial_range = distance;
         }
-        // correct the range for attitude (multiply by DCM.c.z, which
-        // is cos(roll)*cos(pitch))
-        rangefinder_state.height_estimate = distance * ahrs.get_rotation_body_to_ned().c.z;
+        rangefinder_state.height_estimate = corrected_distance;
 
         rangefinder_terrain_correction(rangefinder_state.height_estimate);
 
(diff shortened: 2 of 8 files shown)


IamPete1:
I do have a PR for this https://github.com/ArduPilot/ardupilot/pull/20074 (it does need a rebase). That automatically selects the rangefinder closest to down for the current attitude. The potential problem there is that it will switch in flight.

Would the new set HAGL command work here too? It just seems a bit mad that we need a parameter to tell us which way down is. In the longer term I would like to use all rangefinders all the time using a average with weighting bases on how close to down they are.

ArduPlane/Parameters.cpp
```diff
@@ -1274,6 +1274,15 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     AP_SUBGROUPINFO(precland, "PLND_", 35, ParametersG2, AC_PrecLand),
 #endif
 
+#if AP_RANGEFINDER_ENABLED
+    // @Param: RNGFND_LND_ORNT
+    // @DisplayName: rangefinder landing orientation
+    // @Description: The orientation of rangefinder to use for landing detection. Should be set to rangefinder_orientation() for normal downward facing rangefinder and ROTATION_PITCH_180 for rearward facing rangefinder for quadplane tailsitters. Custom orientation can be used with CUST_ROT_ENABLE. The orientation must match at least one of the available rangefinders.
```
shancock884, line 1280: I think the description will need to use the same vocabulary as the @Values tag, e.g.:
```suggestion
    // @Description: The orientation of rangefinder to use for landing detection. Should be set to Down for normal downward facing rangefinder and Back for rearward facing rangefinder for quadplane tailsitters. Custom orientation can be used with Custom1 or Custom2. The orientation must match at least one of the available rangefinders.
```

tridge: applied, thanks!

tridge:
> I do have a PR for this #20074 (it does need a rebase). That automatically selects the rangefinder closest to down for the current attitude. The potential problem there is that it will switch in flight.

it also doesn't support custom orientations which I think will be useful on some vehicles. The switching in flight issue I think is a real problem, especially with the code we have for starting the use of a rangefinder.
We could use the HAGL method, but I think the parameter is warranted as tailsitters become more common. In the future we could support value -1 for "automatic", but that just complicates things for now

ArduPlane/altitude.cpp
```diff
@@ -679,16 +679,36 @@ void Plane::rangefinder_terrain_correction(float &height)
...
+    float corrected_distance = distance;
+
+    /*
+      correct distance for attitude
+     */
+    if (range_ok) {
+        // correct the range for attitude
+        const auto &dcm = ahrs.get_rotation_body_to_ned();
+
+        Vector3f v{1, 0, 0};
```
IamPete1, line 694: I think if you put the corrected distance in here rather than multiply at the end it is a bit more intuitive. 
```suggestion
        Vector3f v{corrected_distance, 0, 0};
```
and/or a comment explaining the maths.

ArduPlane/system.cpp
```diff
@@ -39,7 +39,7 @@ void Plane::init_ardupilot()
 #if AP_RANGEFINDER_ENABLED
     // initialise rangefinder
     rangefinder.set_log_rfnd_bit(MASK_LOG_SONAR);
-    rangefinder.init(ROTATION_PITCH_270);
+    rangefinder.init(rangefinder_orientation());
```
IamPete1, line 42: I don't think this is a good idea. This value is used to set the param defaults. For example if I had a existing rangefinder and then added a second one. I would set that second one to a custom orientation and then set the new `RNGFND_LND_ORNT` to that custom rotation. However on reboot I would find my original rangefinder has now also changed to that orientation and I would have to go and manually set it back to down.

ArduPlane/Parameters.cpp
```diff
@@ -1274,6 +1274,15 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     AP_SUBGROUPINFO(precland, "PLND_", 35, ParametersG2, AC_PrecLand),
 #endif
 
+#if AP_RANGEFINDER_ENABLED
+    // @Param: RNGFND_LND_ORNT
+    // @DisplayName: rangefinder landing orientation
+    // @Description: The orientation of rangefinder to use for landing detection. Should be set to Down for normal downward facing rangefinder and Back for rearward facing rangefinder for quadplane tailsitters. Custom orientation can be used with Custom1 or Custom2. The orientation must match at least one of the available rangefinders.
+    // @Values: 12:Back, 25:Down, 101:Custom1, 102:Custom2
```
IamPete1, line 1281: Note that the rangefinder orient param does not list custom 1 and custom 2 as valid values.

IamPete1: In fact it also gives `4: Back`. Here you have `12: Back`. One is `ROTATION_YAW_180` and the other `ROTATION_PITCH_180`. For a rangefinder it amounts to the same, but this should match the rangefinder params so selecting "Back" for both works as expected.

ArduPlane/Parameters.cpp
```diff
@@ -1274,6 +1274,15 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     AP_SUBGROUPINFO(precland, "PLND_", 35, ParametersG2, AC_PrecLand),
 #endif
 
+#if AP_RANGEFINDER_ENABLED
+    // @Param: RNGFND_LND_ORNT
```
IamPete1, line 1278: Do we need this on normal planes or could it be a quadplane param? I guess the custom orientations are really the only use case? We could gate on `AP_CUSTOMROTATIONS_ENABLED` although its enabled on most everything.

ArduPlane/Parameters.cpp
tridge: the problem with using YAW_180 is the special handling of the yaw values in Aircraft::perpendicular_distance_to_rangefinder_surface() in SITL - YAW_180 is assumed to be a proximity sensor, so if you use YAW_180 you can't test in SITL
Using YAW_180 will work on a real aircraft, just not SITL

ArduPlane/Parameters.cpp
tridge: we only disable custom rotations on periphs, so it doesn't win us anything
I think we want the orientation to be settable on normal planes too, you may want one which is pointing fwd 45 degrees

ArduPlane/altitude.cpp
tridge: i'm happy either way, I've changed it to do it the way you prefer

ArduPlane/system.cpp
tridge: yep, good catch! fixed

ArduPlane/Parameters.cpp
tridge: @peterbarker looks like you added the special handling for the YAW values in SIM_Aircraft::perpendicular_distance_to_rangefinder_surface() ? does it really need to be that way or can we have some other way to hack in the proximity sensors?

IamPete1: The problem is that the GCS may just give a drop down not the value, so if the user selects "Back" both this and the rangefinder orient it won't work. Maybe you could add "SITL back", or just don't have that case in the values, its not in the values for rangefinder either.

tridge: unless we change that hack in SITL then we can't use YAW180 and make it testable in SITL
we could make the AP_RangeFinder library use pitch180 instead?

IamPete1: I'm only talking about the `@values`, the two "Backs" must be the same (4). You can still use a value that is not listed for in SITL and it should work fine. You were already having to set a value not in the `@Values` for the rangefinder, you will just have to do that too for the new param.

tridge: we can change the meaning of "Back" in the rangefinder library to be 12, but I'd like to discuss that on the call
or we could rename "4: Back" to "4: Back(Yaw180)" in the rangefinder library and add "12: Back" there

IamPete1 (dismissed):
Just need to change the default, looks good apart from that.

IamPete1 (approved):
Looks good.

peterbarker:
Whoops.  Killed the Rover RangeFinder test - we also use those posts in Europe to test Rover's horizontalk rangefinder, and that seems to be busted.

tridge:
> Killed the Rover RangeFinder test - we also use those posts in Europe to test Rover's horizontalk rangefinder, and that seems to be busted.

I've fixed it up

PR #28015, opened 2024-09-05
hwdef: ARKV6x: fix RC input

dakejahl:
This reverts https://github.com/ArduPilot/ardupilot/commit/9f5e837646778c126a4330606f5bc9919e242d94. I believe when I tested that PR I didn't run waf configure and therefore the hwdef changes never got applied. So I thought I fixed something that was never broken to begin with :facepalm: 

I tested this on an ARK Jetson Carrier with an ARKV6X

**requirements for SBUS**
SERIAL8_PROTOCOL = 23 (RCIN)
RC_PROTOCOL = 1 (all)

@tridge I think this should make more sense, let me know if there's anything I missed!

We can close https://github.com/ArduPilot/ardupilot/pull/27736


peterbarker:
Added WikiNeeded on a hunch.

dakejahl:
> Added WikiNeeded on a hunch.

This wiki here? https://github.com/ArduPilot/ardupilot_wiki

What kind of information typically goes in the wiki? At the very least it would be good to document which serial ports correspond to which SERIAL* parameters. And any other basic getting started information specific to the hardware would be good to document. PX4 docs have some vendor specific documentation, should we add the same for Ardupilot?

eg
ARKV6X
https://docs.px4.io/main/en/flight_controller/arkv6x.html
ARK Flow
https://docs.px4.io/main/en/dronecan/ark_flow.html

PR #28017, opened 2024-09-05
AP_HAL_ChibiOS: CubeRedPrimary: probe external compasses

peterbarker:


tridge (approved):

tpwrules (commented):
Tested with IST8308 connected over GPS2 port. Works properly and didn't work before.

PR #28019, opened 2024-09-05
GCS_MAVLink: reject SET_MESSAGE_INTERVAL commands with p3 set

peterbarker:
this is going to be used for changing the rate of a specific instance of a message at some stage

we have to reject it for now so that when the index is used the GCS is told that their message is invalid in this older version of the autopilot


peterbarker:
```
STABILIZE> long SET_MESSAGE_INTERVAL 100 10 3
STABILIZE> Got COMMAND_ACK: SET_MESSAGE_INTERVAL: DENIED
long SET_MESSAGE_INTERVAL 100 10 0
STABILIZE> Got COMMAND_ACK: SET_MESSAGE_INTERVAL: ACCEPTED
```

tridge (approved):

PR #28027, opened 2024-09-06
AP_Scripting: tweak conditions for scripting-requires-filesystem

peterbarker:
AP_FILESYSTEM_READING_ENABLED isn't quite right as we can't read scripts from @SYS.

### DIFF
--- a/libraries/AP_Scripting/AP_Scripting_config.h
+++ b/libraries/AP_Scripting/AP_Scripting_config.h
@@ -9,7 +9,8 @@
 
 #if AP_SCRIPTING_ENABLED
     #include <AP_Filesystem/AP_Filesystem_config.h>
-    #if !AP_FILESYSTEM_FILE_READING_ENABLED
+    // enumerate all of the possible places we can read a script from.
+    #if !AP_FILESYSTEM_POSIX_ENABLED && !AP_FILESYSTEM_FATFS_ENABLED && !AP_FILESYSTEM_ESP32_ENABLED && !AP_FILESYSTEM_ROMFS_ENABLED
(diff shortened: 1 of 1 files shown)


libraries/AP_Scripting/AP_Scripting_config.h
```diff
@@ -9,7 +9,8 @@
 
 #if AP_SCRIPTING_ENABLED
     #include <AP_Filesystem/AP_Filesystem_config.h>
-    #if !AP_FILESYSTEM_FILE_READING_ENABLED
+    // enumerate all of the possible places we can read a script from.
+    #if !AP_FILESYSTEM_POSIX_ENABLED && !AP_FILESYSTEM_FATFS_ENABLED && !AP_FILESYSTEM_ESP32_ENABLED && !AP_FILESYSTEM_ROMFS_ENABLED
```
IamPete1, line 13:
```suggestion
    #if !AP_FILESYSTEM_POSIX_ENABLED && !AP_FILESYSTEM_FATFS_ENABLED && !AP_FILESYSTEM_ESP32_ENABLED && !defined(HAL_HAVE_AP_ROMFS_EMBEDDED_LUA)
```
I think you can swap the romps one for the romfs with scripts define.

IamPete1: Although the existing error string might be a bit critic in that case.

peterbarker: ... also, I'm allergic to *more* `defined(x)` crap :-)

IamPete1 (approved):
LGTM

PR #28033, opened 2024-09-07
Avoid nullptr dereference on bad rcmap value entry

peterbarker:
Setting a bad rcmap entry doesn't just hard-fault the autopilot but ensures you have to reset all of your parameters before you have a working autopilot again.

This has been the case for a very long time.  But perhaps we should fix it?

Checking the library return values for nullptr doesn't help if we have already persisted the parameters - next time you boot you'll not populate e.g. `channel_roll` with a valid pointer and will then segfault.  So we use a "dummy channel" instead.

A new record for shortest autotest.

### DIFF
--- a/ArduPlane/radio.cpp
+++ b/ArduPlane/radio.cpp
@@ -8,16 +8,17 @@
  */
 void Plane::set_control_channels(void)
 {
+    // the library gaurantees that these are non-nullptr:
     if (g.rudder_only) {
         // in rudder only mode the roll and rudder channels are the
         // same.
-        channel_roll = RC_Channels::rc_channel(rcmap.yaw()-1);
+        channel_roll = &rc().get_yaw_channel();
     } else {
-        channel_roll = RC_Channels::rc_channel(rcmap.roll()-1);
+        channel_roll = &rc().get_roll_channel();
     }
-    channel_pitch    = RC_Channels::rc_channel(rcmap.pitch()-1);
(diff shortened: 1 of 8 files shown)


peterbarker:
I've added an additional patch to MSP to avoid a similar problem.

There are almost certainly more in the codebase, but I want to know this is an acceptable solution before patching them all.

peterbarker:
Problems around optimisation again on Pixhawk1-1M-bdshot; elf-diff shows many small increases in random methods

```
CubeOrange-periph-heavy  *                                                                     
Durandal                            -16    *           8       -56               -256   64     72
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     40     *           64      40                56     208    56
MatekF405                           16     *           8       32                176    0      80
Pixhawk1-1M-bdshot                  56                 344     384               280    48     32
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           32     *           16      136               112    48     56
skyviper-v2450                                         40                                      
```

ArduPlane/radio.cpp
```diff
@@ -8,16 +8,17 @@
  */
 void Plane::set_control_channels(void)
 {
+    // the library gaurantees that these are non-nullptr:
```
Ryanf55, line 11:
```suggestion
    // the library guarantees that these are non-nullptr:
```

tridge (approved):

Issue #26137, opened 2024-02-04
Allow using virtual environment on Ubuntu Jammy when running install-prereqs-ubuntu.sh


Ryanf55:
## Feature request

**Is your feature request related to a problem? Please describe.**

ArduPilot python packages that are installed in user-space can interfere with other toolchains. Python developers recommend virtual environments whenever possible. I find myself having to wipe user python packages because they have version incompatibilities, and other development toolchains assume python packages available outside a virtual environment are all managed properly by the distro. 

**Describe the solution you'd like**

Allow `install-prereqs-ubuntu.sh` to have the option to install everything in a virtual environment on Ubuntu Jammy.

**Describe alternatives you've considered**

Manually installing packages in a virtual environment.

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Additional context**

N/A

Ryanf55:
Dev call said this would be accepted if it did not impact the existing flow. An additional param, with a wiki change, would be accepted.

P81000:
I'm using a Docker container to build the code using an Ubuntu image. I was able to run `install-prereqs` inside the container and everything worked well. Can you provide more context? I'm a new ardupilot developer.

Ryanf55:
Prototype branch: https://github.com/Ryanf55/ardupilot/tree/venv-option-on-jammy

Ryanf55:
@peterbarker Would you have any time to consider adding this? The recent PR you introduced to install a different version of setuptools breaks Ubuntu jammy ROS users; now `colcon test` won't find any python tests.
 See here for context.
https://github.com/ArduPilot/ardupilot/issues/27925#issuecomment-2336227563

peterbarker:
I think I saw a PR for jammy but I can't find it now.  I think we'd allow it as an option you could turn on on the command-line or as an environment variable or something?

@Ryanf55 am I mis-remembering such a PR?

Ryanf55:
Yea, I had a branch, but couldn't finish it because the script was more complicated than I expected and I ran out of time. 
See this comment for the WIP branch: https://github.com/ArduPilot/ardupilot/issues/26137#issuecomment-1932619322

Issue #23664, opened 2023-05-02
install-prereqs-ubuntu.sh fails for ubuntu 23.04


AndKe:
## Bug report

**Issue details**

install-prereqs-ubuntu.sh needs changes for python 

**Version**
master

**Platform**
[ x] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
NA

**Hardware type**
NA

**Logs**

```
---------- ./install-prereqs-ubuntu.sh start ----------
+ '[' 1000 == 0 ']'
+ OPT=/opt
+ ARDUPILOT_TOOLS=Tools/autotest
+ ASSUME_YES=false
+ QUIET=false
+ sep='##############################################'
+ OPTIND=1
+ getopts yq opt
+ APT_GET='sudo apt-get'
+ false
+ false
+ sudo apt-get update
[sudo] password for andre: 
Ign:1 https://repo.vivaldi.com/stable/deb stable InRelease
Hit:2 https://ocean.surfshark.com/debian stretch InRelease                                         
Hit:3 https://repo.vivaldi.com/stable/deb stable Release                                           
Get:4 http://no.archive.ubuntu.com/ubuntu lunar InRelease [267 kB]                                 
Get:6 http://security.ubuntu.com/ubuntu lunar-security InRelease [90,7 kB]
Get:7 http://no.archive.ubuntu.com/ubuntu lunar-updates InRelease [90,7 kB]
Hit:8 http://no.archive.ubuntu.com/ubuntu lunar-backports InRelease   
Fetched 448 kB in 1s (542 kB/s)                           
Reading package lists... Done
+ package_is_installed lsb-release
+ dpkg-query -W '-f=${Status}' lsb-release
+ grep -c 'ok installed'
1
++ lsb_release -c -s
+ RELEASE_CODENAME=lunar
+ case ${RELEASE_CODENAME} in
+ PYTHON_V=python3
+ PIP=pip3
+ '[' lunar == bionic ']'
+ '[' lunar == buster ']'
+ '[' lunar == focal ']'
+ '[' lunar == jammy ']'
+ '[' lunar == groovy ']'
+ '[' lunar == bullseye ']'
++ apt-cache search -n '^libcsfml-audio'
++ cut '-d ' -f1
++ head -1
++ grep -Eo '[+-]?[0-9]+([.][0-9]+)?'
+ SITLCFML_VERSION=2.5
[...]

PR #28035, opened 2024-09-07
Tools: Use apt for pexpect and allow system packages in the virtual environment

Ryanf55:
# Purpose

If an apt package is available for a python dependency, I think we should try to use that.

Doing `python3 -m pip install pexpect` opens us to completely breaking as soon as the next major version comes out.
Instead, we can rely on our packaging team at canonical to ensure our packages are compatible.

Let's see how CI does. 

If we like this approach, it's an alternative to virtual environments for Ubuntu jammy: https://github.com/ArduPilot/ardupilot/issues/26137

Same for 
* python3-future
* python3-pexpect
* And I propose all the rest of the python packages that we can, we switch to apt, especially setuptools

### DIFF
--- a/Tools/environment_install/install-prereqs-ubuntu.sh
+++ b/Tools/environment_install/install-prereqs-ubuntu.sh
@@ -164,8 +164,8 @@ else
 fi
 
 # Lists of packages to install
-BASE_PKGS="build-essential ccache g++ gawk git make wget valgrind screen"
-PYTHON_PKGS="future lxml pymavlink pyserial MAVProxy pexpect geocoder empy==3.3.4 ptyprocess dronecan"
+BASE_PKGS="build-essential ccache g++ gawk git make wget valgrind screen python3-pexpect"
+PYTHON_PKGS="future lxml pymavlink pyserial MAVProxy geocoder empy==3.3.4 ptyprocess dronecan"
 PYTHON_PKGS="$PYTHON_PKGS flake8 junitparser"
 
 # add some Python packages required for commonly-used MAVProxy modules and hex file generation:
--- a/Tools/environment_install/install-prereqs-ubuntu.sh
+++ b/Tools/environment_install/install-prereqs-ubuntu.sh
@@ -386,7 +386,7 @@ fi
 
 if [ -n "$PYTHON_VENV_PACKAGE" ]; then
     $APT_GET install $PYTHON_VENV_PACKAGE
-    python3 -m venv $HOME/venv-ardupilot
+    python3 -m venv --system-site-packages $HOME/venv-ardupilot
 
     # activate it:
     SOURCE_LINE="source $HOME/venv-ardupilot/bin/activate"
--- a/Tools/environment_install/install-prereqs-arch.sh
+++ b/Tools/environment_install/install-prereqs-arch.sh
@@ -54,7 +54,7 @@ sudo usermod -a -G uucp "$USER"
 
 sudo pacman -Syu --noconfirm --needed $BASE_PKGS $SITL_PKGS $PX4_PKGS
 
-python3 -m venv "$HOME"/venv-ardupilot
+python3 -m venv --system-site-packages "$HOME"/venv-ardupilot
 
 # activate it:
 SOURCE_LINE="source $HOME/venv-ardupilot/bin/activate"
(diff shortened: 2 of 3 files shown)


Ryanf55:
This is also recommended by another user here who struggled with empy:
https://github.com/ArduPilot/ardupilot/issues/23664#issuecomment-1533722170

Ryanf55:
Looks like the `manic` build failed.

For pexecpt, even though ubuntu:manic has that package, and it installs, python can't find it.
https://github.com/ArduPilot/ardupilot/actions/runs/10753345775/job/29822682675?pr=28035#step:5:1778

If we are ok, we can add the argument `--system-site-pacakges` to give the virtual environment access to the system packages installed through apt.

khancyr:
The system package things is already needs for Archlinux so we can totally use it for Ubuntu too.

What we should looks after is a requirement.txt ( or newer equivalent) so we can track better the python package needs for release

Ryanf55:
The ROS way of doing things is use system packages whenever possible. If we need to install custom versions not distributed with the system, that is when we we should stick them a virtual environment. 

Yes, a requirements file is great, but there's so much logic in this script I don't know if you'd be off any better.

peterbarker:
Yes, we've had things break when python package maintainers do silly things.  We fix those problems when they occur - often by pinning the packages until we can work out why and how we need to update the way we use a package.

Things generally tend *not* to break.

As to using the system packages - that would seem to remove one of the chief advantages of pyenv - separation of your environment from the system environment.  Now you've got two places that can break things!

If we were keen enough we could freeze every package version in `requirements.txt` - but I don't think we're that keen.

The `empy` case you're referencing there was a particularly egregious case.  There was no reason for upstream to break the API in the way it was - backwards compatability was possible.  The author of that package was quite helpful in trying to help us resolve it, mind you.

Are you trying to solve  *specific* problem here?

Ryanf55:
> Yes, we've had things break when python package maintainers do silly things. We fix those problems when they occur - often by pinning the packages until we can work out why and how we need to update the way we use a package.
> 
> Things generally tend _not_ to break.
Until the next major version comes out. We should at least be pinning that.
> As to using the system packages - that would seem to remove one of the chief advantages of pyenv - separation of your environment from the system environment. Now you've got two places that can break things!
Respectfully, I disagree - package maintainers do not allow major version bumps after a distribution is released, and they do their best to avoid breaking things. It's on the onus of us developers to write cross-platform code that works on the main LTS distros with the versions supplied. And I don't see why penv is related, it's not used in the script; [only mac uses it](https://github.com/search?q=repo%3AArduPilot%2Fardupilot%20pyenv&type=code).
> If we were keen enough we could freeze every version in `requirements.txt` - but I don't think we're that keen.
Agree.
> The `empy` case you're referencing there was a particularly egregious case. There was no reason for upstream to break the API in the way it was - backwards compatability was possible. The author of that package was quite helpful in trying to help us resolve it, mind you.
> 
> Are you trying to solve _specific_ problem here?

Yes, the problem is two-fold:
* For the past two months, the DDS interface on `master` has zero functional test coverage aside from the unit tests and linters, thus it can let in regressions. Approving PR's now takes a lot more time because I have to run all the tests individually through pytest.
* Anyone developer on Ubuntu 22 who runs the setup script will ruin their ability for `setup.py`-based python packages to have tests; this will affect most ROS developers who try, and sadly the problem only manifests as a warning until you notices when you run tests that pytest collects 0 tests.

tridge (approved):

Tools/environment_install/install-prereqs-ubuntu.sh
```diff
@@ -386,7 +386,7 @@ fi
 
 if [ -n "$PYTHON_VENV_PACKAGE" ]; then
     $APT_GET install $PYTHON_VENV_PACKAGE
-    python3 -m venv $HOME/venv-ardupilot
+    python3 -m venv --system-site-packages $HOME/venv-ardupilot
```
tridge, line 389: can you also change the others (eg. arch) so we are as consistent as we can be between OSes

tridge (changes requested):

peterbarker (approved):
LGTM

Tools/environment_install/install-prereqs-ubuntu.sh
Ryanf55: Yep, done!

PR #28036, opened 2024-09-08
AP_HAL_Linux: keep "dfd" open util fsync is done.

junan76:
Since we want to do a fsync on "dfd", it can not be closed before that.


peterbarker (approved):
We are also leaking a file descriptor if that `fgtruncate` fails.

PR #28038, opened 2024-09-08
Networking: added CAN multicast UDP network gateway

tridge:
This allows the PPPGW peripherals to act as CAN multicast gateways, which allows any device on the ethernet network to interact with CAN devices on the flight controllers CAN networks
It also means we can configure and update the PPPGW peripherals without knowing the IP address it is configured for, as we can use DroneCAN GUI tool with interface mcast:0 or mcast:1 to get at CAN1 or CAN2 and configure parameters

This gateway is enabled in both AP_Periph with PPPGW and in the network enabled AP_Periph bootloader, which means this can also be used to update the firmware on the AP_Periph PPPGW device

This costs about 450 bytes, which comes from the change to the hal.can API to allow registration of more than one callback. We could have a new API that only appears when this option is enabled, but it would be a duplicate of the current register function, just with a new cb targets, maybe cb_mcast.


magicrub:
You had me at update FW....

timtuxworth:
If I run this "behind" a Raspberry PI on the plane - so not exposing the IP address of the device directly (e.g. I use a proxy to forward the web server and mavp2p to forward the mavlink traffic over a different interface), how would I forward the CAN traffic?

tridge:
> how would I forward the CAN traffic?

the multicast traffic can be fwded over the link if your IP radio allows that. I'm thinking we may need to have an option to set the TTL to 1 so it doesn't get fwded

note you can use the existing network port fwding to fwd mavlink and then use mavcan

Ryanf55:
Setting up CAN on companion computers is hard. This is a nice alternative.

libraries/AP_HAL/CANIface.cpp
```diff
@@ -70,28 +77,59 @@ int16_t AP_HAL::CANIface::receive(CANFrame& out_frame, uint64_t& out_ts_monotoni
...
             add_to_rx_queue(rx_item);
+            added_to_rx_queue = true;
         }
     }
     return 1;
 }
 
 /*
   register a callback for for sending CAN_FRAME messages
+  on success the returned callback_id can be used to unregister the callback
```
Ryanf55, line 104:
```suggestion
  Register a callback for for sending CAN_FRAME messages.
  On success, the returned callback_id can be used to unregister the callback.
```
nit; These two comments bleed together

libraries/AP_HAL/CANIface.h
```diff
@@ -262,13 +262,21 @@ class AP_HAL::CANIface
...
     virtual int8_t get_iface_num() const = 0;
     virtual bool add_to_rx_queue(const CanRxItem &rx_item) = 0;
 
-    FrameCb frame_callback;
+    struct {
+#ifndef HAL_BOOTLOADER_BUILD
+        HAL_Semaphore sem;
+#endif
+        // allow up to 2 callbacks per interface
+        FrameCb cb[2];
```
Ryanf55, line 277: Could/should this be configurable?

libraries/AP_Networking/AP_Networking_CAN.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+    for (uint8_t bus=0; bus<HAL_NUM_CAN_IFACES; bus++) {
+        auto *cbus = get_caniface(bus);
+        if (cbus == nullptr) {
+            continue;
+        }
+        if (bus_mask & (1U<<bus)) {
+            mcast_sockets[bus] = NEW_NOTHROW SocketAPM(true);
+            if (mcast_sockets[bus] == nullptr) {
+                GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "CAN_MCAST[%u]: failed to create socket", unsigned(bus));
```
Ryanf55, line 105: Would  factoring out `CAN_MCAST[%u]:` into a single place save flash? I've seen this done in other places.

libraries/AP_Networking/AP_Networking_CAN.cpp
```diff
@@ -0,0 +1,226 @@
...
+            struct mcast_pkt pkt;
+            const ssize_t ret = mcast_sockets[bus]->recv((void*)&pkt, sizeof(pkt), 0);
+            if (ret > MCAST_HDR_LENGTH && ret <= sizeof(pkt)) {
+                const uint8_t data_len = ret - MCAST_HDR_LENGTH;
+                bool is_canfd = false;
+#if HAL_CANFD_SUPPORTED
+                is_canfd = (pkt.flags & MCAST_FLAG_CANFD) != 0;
+#endif
+                if (pkt.magic != MCAST_MAGIC) {
+                    continue;
```
Ryanf55, line 167: Is magic intended as a compatibilty flag? If so, perhaps it's worth a one-time warning to the user that an incompatible CAN mcast packet was received. Otherwise, wouldn't this silently drop the data? It could be hard to figure out why two devices with different magic values aren't going to communicate.

libraries/AP_Networking/AP_Networking_CAN.cpp
```diff
@@ -0,0 +1,226 @@
...
+    if (frame.isCanFDFrame()) {
+        pkt.flags |= MCAST_FLAG_CANFD;
+    }
+#endif
+    pkt.message_id = frame.id;
+
+    const uint8_t data_length = AP_HAL::CANFrame::dlcToDataLength(frame.dlc);
+
+    memcpy(pkt.data, frame.data, data_length);
+    pkt.crc = crc16_ccitt((uint8_t*)&pkt.flags, data_length+6, 0xFFFFU);
```
Ryanf55, line 221: Where'd the 6 come from?

libraries/AP_Networking/AP_Networking_CAN.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "AP_Networking.h"
+#include <AP_HAL/utility/Socket.h>
+
+class AP_Networking_CAN {
+public:
+    void start(uint8_t bus_mask);
```
Ryanf55, line 8:
```suggestion
    void start(const uint8_t bus_mask);
```
Could add `const` on the public part of the class.

Ryanf55 (commented):
I read through the PR, really neat Tridge! Just some ideas, largely NFC. 
I tried to find spots for nullptr dereference and it looks good.

peterbarker:
@bugobliterator is reviewing this - hopefully done soon!

libraries/AP_HAL/CANIface.h
tridge: no point in making it configurable yet, but if we allow scripting access in the future it would make sense

libraries/AP_Networking/AP_Networking_CAN.cpp
tridge: i doubt it will save flash, and it would make it harder to read

libraries/AP_HAL/CANIface.cpp
tridge: changed

libraries/AP_Networking/AP_Networking_CAN.cpp
tridge: added a comment

libraries/AP_Networking/AP_Networking_CAN.h
tridge: changed, although it really doesn't matter

libraries/AP_Networking/AP_Networking_CAN.cpp
tridge: no, it is to ensure that some other multicast user that happens to choose the same port doesn't inject bad data. I tried to reserve a port with IANA but they rejected my application

libraries/AP_Networking/AP_Networking_CAN.cpp
Ryanf55: Ok. I'll go check on the other places I did this. It was harder to read, and never checked it actually saved flash.

libraries/AP_Networking/AP_Networking_CAN.cpp
timtuxworth: Maybe that deserves a comment?

Tools/AP_Bootloader/support.cpp
```diff
@@ -364,6 +364,17 @@ void thread_sleep_ms(uint32_t ms)
     }
 }
 
+void thread_sleep_us(uint32_t us)
+{
+    while (us > 0) {
+        // don't sleep more than 65 at a time, to cope with 16 bit
+        // timer
+        const uint32_t dt = us > 6500? 6500: us;
+        chThdSleepMicroseconds(dt);
```
bugobliterator, line 373: might be safer to use chTimeUS2I(1)*us and compare that to (UINT16_MAX-1)? and call chThdSleep on that?

bugobliterator (changes requested):
looks good, except for minor change request to thread_sleep_us. That could be a future trap if someone tries to run an odd conversion frequency between ticks and seconds.

Tools/AP_Bootloader/support.cpp
tridge: notice it is 6.5ms max, and chThdSleepMicroseconds() is what we use elsewhere in the bootloader. I think the call is right

bugobliterator (approved):

robertlong13:
This is the PR that caused the issues with DroneCAN over MAVLink described in #28175

robertlong13:
Specifically, it was either 4a102e2f2be3f977840dd85050c74a1c4ea03ba2 or cc930bd49f1af9a1228573a87e92c84c5c8e8333

PR #28039, opened 2024-09-08
waf: allow any custom build option to be specified in waf configure

tridge:
this makes it easy to configure with any option from build_options.py, with tab completion

Note that the build_options.py table takes an optional enable-XXX label to allow us to replace some of the existing options (eg. --enable-ppp and --disable-ekf3)

we end up with options like this: 
```
    --enable-ppp        Enable PPP networking
    --disable-ppp       Disable PPP networking
    --enable-DISPLAY    Enable I2C Displays
    --disable-DISPLAY   Disable I2C Displays
    --enable-LED_CONTROL
                        Enable MAVLink LED Control
    --disable-LED_CONTROL
                        Disable MAVLink LED Control
    --enable-NOTIFY_NCP5623
                        Enable NCP5623 LED
    --disable-NOTIFY_NCP5623
                        Disable NCP5623 LED
    --enable-NOTIFY_NEOPIXEL
                        Enable NeoPixel
    --disable-NOTIFY_NEOPIXEL
                        Disable NeoPixel
    --enable-NOTIFY_PROFILED
                        Enable ProfiLED
    --disable-NOTIFY_PROFILED
                        Disable ProfiLED
    --enable-NOTIFY_PROFILED_SPI
                        Enable ProfiLED (SPI)
    --disable-NOTIFY_PROFILED_SPI
                        Disable ProfiLED (SPI)
    --enable-PLAY_TUNE  Enable MAVLink Play Tune
    --disable-PLAY_TUNE
                        Disable MAVLink Play Tune
    --enable-TONEALARM  Enable ToneAlarm on PWM
    --disable-TONEALARM
                        Disable ToneAlarm on PWM
    --enable-OSD        Enable OSD
    --disable-OSD       Disable OSD
    --enable-OSD_EXTENDED_LINK_STATS
                        Enable OSD panels with extended link stats data
    --disable-OSD_EXTENDED_LINK_STATS
                        Disable OSD panels with extended link stats data
    --enable-OSD_PARAM  Enable OSD param
    --disable-OSD_PARAM
                        Disable OSD param
```

### DIFF
--- a/Tools/scripts/build_options.py
+++ b/Tools/scripts/build_options.py
@@ -417,7 +417,7 @@ BUILD_OPTIONS = [
     Feature('Filesystem', 'FILESYSTEM_SYS', 'AP_FILESYSTEM_SYS_ENABLED', 'Enable @SYS/ filesystem', 0, None),
     Feature('Filesystem', 'APJ_TOOL_PARAMETERS', 'FORCE_APJ_DEFAULT_PARAMETERS', 'Enable apj_tool parameter area', 0, None),
 
-    Feature('Networking', 'PPP Support', 'AP_NETWORKING_BACKEND_PPP', 'Enable PPP networking', 0, None),
+    Feature('Networking', 'PPP', 'AP_NETWORKING_BACKEND_PPP', 'Enable PPP networking', 0, None),
 
     Feature('DroneCAN', 'DroneCAN', 'HAL_ENABLE_DRONECAN_DRIVERS', 'Enable DroneCAN support', 0, None),
 ]
--- a/Tools/scripts/build_options.py
+++ b/Tools/scripts/build_options.py
@@ -23,6 +23,10 @@ class Feature:
         self.default = default
         self.dependency = dependency
 
+    def config_option(self):
+        '''the name of the configure option to be used by waf'''
+        return "enable-" + self.label.replace(" ", "-")
+
 
 # list of build options to offer NOTE: the dependencies must be
 # written as a single string with commas and no spaces,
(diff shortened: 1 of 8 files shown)


Tools/scripts/build_options.py
```diff
@@ -413,7 +421,7 @@ def __init__(self,
     Feature('Filesystem', 'FILESYSTEM_SYS', 'AP_FILESYSTEM_SYS_ENABLED', 'Enable @SYS/ filesystem', 0, None),
     Feature('Filesystem', 'APJ_TOOL_PARAMETERS', 'FORCE_APJ_DEFAULT_PARAMETERS', 'Enable apj_tool parameter area', 0, None),
 
-    Feature('Networking', 'PPP Support', 'AP_NETWORKING_BACKEND_PPP', 'Enable PPP networking', 0, None),
+    Feature('Networking', 'PPP Support', 'AP_NETWORKING_BACKEND_PPP', 'Enable PPP networking', 0, None, "enable-ppp"),
```
peterbarker, line 424: Could we put these hacks back in `waf`, please?

Should just be a dictionary mapping from A to B?

peterbarker (changes requested):

Tools/scripts/build_options.py
tridge: it needs to be here to keep backward compatibility with existing documented options like --enable-ppp. We won't need it for new options. It would be an awful lot uglier to have this in wscript

tridge: if we're willing to break existing documented configure options we could remove these

peterbarker: Or throw a warning in 4.6 that the option is deprecated and remove in 4.7?

timtuxworth: > if we're willing to break existing documented configure options we could remove these

I'm thinking that breaking them would be ok, although it will be a small inconvenience initially for a very small minority who can probably deal with it, because in the long term, part of the power of this feature comes from it's consistency - everything will work the same way.

timtuxworth:
Just a thought about how the options work, could it be 

`--enable FEATURE`

rather than

`--enable-FEATURE`

Doesn't that make more sense?

Tools/scripts/build_options.py
timtuxworth: Surely 
`Feature('Networking', 'PPP Support', 'AP_NETWORKING_BACKEND_PPP', 'Enable PPP networking', 0, None, "enable-ppp"),`
should be "PPP" not "PPP Support" - it seems to be the only one with a space in the label field.

timtuxworth: If we want to support --enable-ppp, could you not _also_ support --enable-PPP, ditto for ekf2/EKF2, ekf3/EKF3.

I'm thinking that someone might want to build a script that would generate a ./waf configure command line. These special cases would make that difficult, but supporting both (deprecating the lowercase option) might be a good interim solution.

tridge:
> `--enable FEATURE` or even `--enable FEATURE1,FEATURE2,FEATURE3`

we won't be able to tab complete that, or at least not without a lot of work in the completions logic for each shell. Being able to tab complete options is very nice

tridge:
after discussion on the dev call I've removed the backwards compatibility logic

peterbarker:
> after discussion on the dev call I've removed the backwards compatibility logic

We had the forethought to create autotests for the options you've removed :-)

Still, fixing those means we can't easily break this new functionality

peterbarker:
@Hwurzburg for the WikiNeeded, we need to remove reference to the removed options --enable-ekf3 and friends (or make it clear they're pre-4.6), and perhaps add a pointer to using the new capabilities for listing configuration options with `--help` and using them like any other command-line option

PR #28040, opened 2024-09-08
AC_AttitudeControl: Fix dt update order

lthall:
This PR moves the movement of target position to the first step of the attitude controller where it knows the dt.

Currently we assume the next dt will be the same as the last dt. This causes noise in the rate controllers when dt changes and high rates are being requested.

Limitations in ATT logging has made this difficult to observe previously.

Before:
image

After:
image

### DIFF
--- a/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
+++ b/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
@@ -254,14 +254,15 @@ void AC_AttitudeControl::landed_gain_reduction(bool landed)
 // remain very close together.
 //
 // All input functions below follow the same procedure
-// 1. define the desired attitude the aircraft should attempt to achieve using the input variables
-// 2. using the desired attitude and input variables, define the target angular velocity so that it should
+// 1. define the desired attitude or attitude change based on the input variables
+// 2. update the target attitude based on the angular velocity target and the time since the last loop
+// 3. using the desired attitude and input variables, define the target angular velocity so that it should
 //    move the target attitude towards the desired attitude
-// 3. if _rate_bf_ff_enabled is not being used then make the target attitude
+// 4. if _rate_bf_ff_enabled is not being used then make the target attitude
 //    and target angular velocities equal to the desired attitude and desired angular velocities.
-// 4. ensure _attitude_target, _euler_angle_target, _euler_rate_target and
+// 5. ensure _attitude_target, _euler_angle_target, _euler_rate_target and
 //    _ang_vel_target have been defined. This ensures input modes can be changed without discontinuity.
-// 5. attitude_controller_run_quat is then run to pass the target angular velocities to the rate controllers and
+// 6. attitude_controller_run_quat is then run to pass the target angular velocities to the rate controllers and
 //    integrate them into the target attitude. Any errors between the target attitude and the measured attitude are
 //    corrected by first correcting the thrust vector until the angle between the target thrust vector measured
 //    trust vector drops below 2*AC_ATTITUDE_THRUST_ERROR_ANGLE. At this point the heading is also corrected.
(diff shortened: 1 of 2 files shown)


libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -745,6 +765,18 @@ Quaternion AC_AttitudeControl::attitude_from_thrust_vector(Vector3f thrust_vecto
     return thrust_vec_quat*yaw_quat;
 }
 
+// Calculates the body frame angular velocities to follow the target attitude
+void AC_AttitudeControl::update_attitude_target()
+{
+    // rotate target and normalize
+    Quaternion attitude_target_update;
+    attitude_target_update.from_axis_angle(Vector3f{_ang_vel_target.x * _dt, _ang_vel_target.y * _dt, _ang_vel_target.z * _dt});
```
andyp1per, line 773: Given other changes coming I think it would be better to take a copy of the attitude target, do the update and then finally assign back from the local variable. Might also give the optimizer better opportunites.

andyp1per (changes requested):

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
lthall: `attitude_target = _attitude_target;
attitude_target = attitude_target * attitude_target_update;
attitude_target.normalize();
_attitude_target = attitude_target;`

You mean this right?

andyp1per: yeah

andyp1per (approved):

IamPete1:
Just to make sure I have understood this correctly:

Currently we do the integration of `_attitude_target` at the end of the loop. This meant we used the angular velocity's from this loop. However the `_dt` is from the last loop at that point.

By moving the integration to the start of the loop were also using the angular velocity's from the last loop. The result is that the angular velocity's and the `_dt` are both for the same period of time.

IamPete1:
There is a change in behavior in the `input_angle_step_bf_roll_pitch_yaw` case. It does not get the new `update_attitude_target` call but the old method in `attitude_controller_run_quat` would have run, Its only used in multi rotor autotune and as far as I can tell multi rotor autotune does not disable `_rate_bf_ff_enabled`.

IamPete1 (commented):
I suspect this makes the same problem worse on plane. The problem is that copter sets all the dt's in the `run_rate_controller` call which is the first thing that happens in the loop. On plane we also update the dt's in the `rate_controller_run` call. However we do that at the end of the loop. So when the attitude controller uses dt it will be the loop before last not the last loop. 

I think the fix for plane is to move the setting of the dt's away from  the `run_rate_controller` call and do them at the start of the loop the same as copter. I think we could probably do that without too much testing, however moving the actual `run_rate_controller` so we completely match copter would be a big job. 

Anyway, I don't think we need to worry too much about plane in this PR.

lthall:
> There is a change in behavior in the `input_angle_step_bf_roll_pitch_yaw` case. It does not get the new `update_attitude_target` call but the old method in `attitude_controller_run_quat` would have run, Its only used in multi rotor autotune and as far as I can tell multi rotor autotune does not disable `_rate_bf_ff_enabled`.

This function sets:
```
    // Set rate feedforward requests to zero
    _euler_rate_target.zero();
    _ang_vel_target.zero();
```
So the attitude target does not change.

> Anyway, I don't think we need to worry too much about plane in this PR.

I agree. These timing issues become less and less important as aircraft get bigger and the filter frequencies get lower. It would be good for plane to optimise the loop structure but I doubt any of these changes will impact the performance of plane.

lthall:
> I think the changes look good. The only question is that this now updates the attitude target in every case not just if `_rate_bf_ff_enabled` is true?

Yes that is correct. When `_rate_bf_ff_enabled` is false we are effectively only normalising the attitude target. The main reason for this is I realised I was calculating the euler targets before the update of the attitude target.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
tridge: i think the _dt multiply could be outside the Vector3f{}

Ryanf55:
Line 501 can be improved by factoring dt.
`attitude_target_update.from_axis_angle(Vector3f{roll_rate_rads * _dt, pitch_rate_rads * _dt, yaw_rate_rads * _dt});`

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
andyp1per: if you do it like this:
```
    Quaternion attitude_target_update, att = _attitude_target;
    attitude_target_update.from_axis_angle(_ang_vel_target * _dt);
    att *= attitude_target_update;
    
    // ensure Quaternion stay normalised
    att.normalize();
```
it saves 12 bytes

andyp1per:
Tested on my 5" - looks good.

rmackay9 (approved):

IamPete1 (approved):

PR #28041, opened 2024-09-08
AP_NavEKF3: Align CORE determination with other methods

muramura:
The CORE determination is different! CORE, CORE != NULLPTR. 
I think it would be better to align them.

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3.cpp
@@ -787,7 +787,7 @@ bool NavEKF3::InitialiseFilter(void)
     }
 #endif
 
-    if (core == nullptr) {
+    if (!core) {
 
         // don't run multiple filters for 1 IMU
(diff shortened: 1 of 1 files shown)


libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -787,7 +787,7 @@ bool NavEKF3::InitialiseFilter(void)
     }
 #endif
 
-    if (core == nullptr) {
+    if (!core) {
```
peterbarker, line 790: `core == nullptr` is our current standard.

This patch moves us away from that standard.

peterbarker (commented):
Our standard is to compare against nullptr as it adds a very small amount of type-safety.

Note that we won't typically move from `!pointer` to `pointer == nullptr` just because, but if you're modifying code in the area we will.

peterbarker:
Closing this one as it moves us further from our standard

PR #28042, opened 2024-09-08
Migrating AP_HAL_ESP32 to the current ESP-IDF v5.3

arg7:
Main focus of this PR:

- Migration to the current ESP-IDF build environment (v5.3).

Additional updates:

- Installation scripts: Improvements added to simplify setup for new users.
- Scheduler fix: Resolved WDT (Watchdog Timer) issues, which were broken with v5.3.
- ADC (Analog-to-Digital Converter): Ported to the new driver, but still non-functional. Battery voltage and current pin numbers aren't being passed correctly to the low-level ADC functions. Further investigation is ongoing.
- RMT and PWM drivers: Not yet ported to the new drivers. You may see warnings about deprecated stuff.

Best regards,
Ing. A. Romanov


davidbuzz (commented):
- thanks for the idf-update PR, I've started looking at it, and its quite large, so bear with me.     General high-level comments follow:
I like the general "where your going with this"  direction, but its going to need some git and assorted cleanup/s..
 1 -   squashing into a single commit, then running ./Tools/gittools/git-subsystems-split on the resulting commit and seeing how it adds better prefixes and breaks it up by folder/s etc.    This helps Ardrew and Randy review it etc.
 2 -  you've added a new ESP_PLATFORM define where we'd traditionally use something like '#if CONFIG_HAL_BOARD == HAL_BOARD_ESP32' to wrap esp32-specific stuff thats outside its HAL, 
 3 - you've wrapped-up in this PR a lot of non-esp32 stuff like the python virtual-env changes, these should probably be a separate PR, unless there's something in this PR that absolutely doesn't work without them, 
 4 - testing - can u put some information in the discussion of this PR saying what board/s you've tested, and what feature/s youve tested.?  eg:
  "I tested mavlink-over-tcp-over-wifi on a dev board, using the esp32s3devkit' target", and it worked better than before" - or whatever.
  5 -    extra lines,  moving functions around and the dozen or so   of DBG_PRINTF changes make it harder to  review the core/important changes, I'm not saying *have* to leave all of them out, but trying to keep to "just the essentials" makes it easier to review and more likely to be merged.

libraries/AP_BattMonitor/AP_BattMonitor_Analog.cpp
```diff
@@ -86,7 +86,7 @@ AP_BattMonitor_Analog::AP_BattMonitor_Analog(AP_BattMonitor &mon,
     }
 #endif
     _state.var_info = var_info;
-    
+    //AR
```
davidbuzz, line 89: unneeded whitespace change

libraries/AP_HAL_ESP32/AnalogIn.h
```diff
@@ -19,18 +19,25 @@
...
 #include "AP_HAL_ESP32.h"
 
-#if HAL_USE_ADC == TRUE && !defined(HAL_DISABLE_ADC_DRIVER)
+#include "esp_adc/adc_oneshot.h"
+#include "esp_adc/adc_cali.h"
+#include "esp_adc/adc_cali_scheme.h"
+
+#if HAL_USE_ADC
+
+//&& !defined(HAL_DISABLE_ADC_DRIVER)
```
davidbuzz, line 28: unneeded  commented-out code

libraries/AP_HAL_ESP32/HAL_ESP32_Class.cpp
```diff
@@ -26,19 +26,22 @@
 #include "RCInput.h"
 #include "RCOutput.h"
 #include "Storage.h"
+
```
davidbuzz, line 29: unneeded whitespace change/s

libraries/AP_HAL_ESP32/RmtSigReader.cpp
```diff
@@ -7,19 +7,22 @@ using namespace ESP32;
...
+    config.mem_block_num = 4; //each block could store 64 pulses
+
     config.flags = 0;
     config.rx_config.filter_en = true;
     config.rx_config.filter_ticks_thresh = 8;
     config.rx_config.idle_threshold = idle_threshold;
 
     rmt_config(&config);
-    rmt_driver_install(config.channel, max_pulses * 8, 0);
+    rmt_driver_install(config.channel, max_pulses*16, 0 );
```
davidbuzz, line 25: you've deliberately changed this multiplier , and config.mem_block_num above , but there doesn't seem to be any explanation of why?  perhaps an extra 1-line commend here would clarify its purpose.?

libraries/AP_HAL_ESP32/Scheduler.cpp
```diff
@@ -571,8 +559,11 @@ void IRAM_ATTR Scheduler::_main_thread(void *arg)
 
     sched->set_system_initialized();
 
+    //initialize WTD for current thread on FASTCPU, all cores will be (1 << CONFIG_FREERTOS_NUMBER_OF_CORES) - 1
```
davidbuzz, line 562: the code used to be explicit about the fact that it was disabling WDT on both core/s with disableCore0WDT/disableCore1WDT, and now that you've simplified this, it appears wdt is now enabled on one core, and disabled on the other, which is fine, if it works, but perhaps a bit more of a warning for those coming after you in the comments might help here.

libraries/AP_HAL_ESP32/SdCard.cpp
```diff
@@ -254,8 +254,16 @@ void mount_sdcard()
 {
     mount_sdcard_spi();
 }
+
 #endif // end spi
 
+void unmount_sdcard()
```
davidbuzz, line 260: I cant see any reason why this function was moved around

libraries/AP_HAL_ESP32/boards/esp32s3devkit.h
```diff
@@ -14,166 +14,138 @@
...
 // make sensor selection clearer
-#define PROBE_IMU_I2C(driver, bus, addr, args ...) ADD_BACKEND(AP_InertialSensor_ ## driver::probe(*this,GET_I2C_DEVICE(bus, addr),##args))
-#define PROBE_IMU_SPI(driver, devname, args ...) ADD_BACKEND(AP_InertialSensor_ ## driver::probe(*this,hal.spi->get_device(devname),##args))
-#define PROBE_IMU_SPI2(driver, devname1, devname2, args ...) ADD_BACKEND(AP_InertialSensor_ ## driver::probe(*this,hal.spi->get_device(devname1),hal.spi->get_device(devname2),##args))
 
-#define PROBE_BARO_I2C(driver, bus, addr, args ...) ADD_BACKEND(AP_Baro_ ## driver::probe(*this,std::move(GET_I2C_DEVICE(bus, addr)),##args))
-#define PROBE_BARO_SPI(driver, devname, args ...) ADD_BACKEND(AP_Baro_ ## driver::probe(*this,std::move(hal.spi->get_device(devname)),##args))
+#define AP_RCPROTOCOL_ENABLED 1
+
+#define AP_RCPROTOCOL_IBUS_ENABLED 1
```
davidbuzz, line 23: This particular board was created by @nkruzan , and he presumably has hardware that's wired a particular way. 
I dont believe its pinout is documented in any sort of readme at this stage ( unfortunately), so its a bit of guesswork wether he's going to be concerned by having sbus and ppmsum "turned off" on his hardware.  
 I suggest that documenting a public pinout for your board/hardware and using your own specific build-target that matches it is better than re-using this one in this case.  eg README.esp32s3arg and using --board=esp32s3arg   [ or whatever you call it ].
It would be great to have a 's3' target that has its pinout/s documented better.  
conversely, we could ask @nkruzan  how he feels about you re-using his board-type, and he might be fine with it. :-)

libraries/AP_HAL_ESP32/boards/esp32s3devkit.h
```diff
@@ -14,166 +14,138 @@
...
-
+#define HAL_ESP32_WIFI 2
 #define WIFI_SSID "ardupilot123"
 #define WIFI_PWD "ardupilot123"
+// */
 
 //RCOUT which pins are used?
 
-#define HAL_ESP32_RCOUT { GPIO_NUM_11,GPIO_NUM_10, GPIO_NUM_9, GPIO_NUM_8, GPIO_NUM_7, GPIO_NUM_6 }
+#define HAL_ESP32_RCOUT { GPIO_NUM_6, GPIO_NUM_7, GPIO_NUM_10, GPIO_NUM_11, GPIO_NUM_12, GPIO_NUM_13, GPIO_NUM_16 }
```
davidbuzz, line 111: see my above note/s about this sort of thing breaking Nick's hardware. :-)

libraries/AP_HAL_ESP32/boards/esp32s3empty.h
```diff
@@ -26,70 +74,111 @@
...
+
+#define HAL_INS_DEFAULT HAL_INS_MPU9250_I2C
+#define HAL_INS_MPU9250_NAME "MPU9250"
+
+// BARO choices:
+#define HAL_BARO_DEFAULT HAL_BARO_BMP280_I2C
+#define HAL_BARO_BMP280_NAME "BMP280"
+
+// IMU probing:
+#define HAL_INS_PROBE_LIST PROBE_IMU_I2C(Invensense, 0, 0x68, ROTATION_YAW_180)
```
davidbuzz, line 92: the 'xxxempty' board/s are a bit special, and are designed to work with no sensor/s attached, and this looks like its changing that.?

libraries/AP_HAL_ESP32/targets/esp32/esp-idf/sdkconfig.orig
```diff
@@ -0,0 +1,1728 @@
+#
+# Automatically generated file. DO NOT EDIT.
```
davidbuzz, line 2: the .orig file should not be committed

libraries/AP_RSSI/AP_RSSI.cpp
```diff
@@ -127,7 +127,8 @@ AP_RSSI *AP_RSSI::get_singleton()
 void AP_RSSI::init()
 {
     // a pin for reading the receiver RSSI voltage. The scaling by 0.25 
-    // is to take the 0 to 1024 range down to an 8 bit range for MAVLink    
+    // is to take the 0 to 1024 range down to an 8 bit range for MAVLink
+    //AR
```
davidbuzz, line 131: unnecessary change

davidbuzz:
Please let me know when you've done some of this, and I'll take another look.

davidbuzz:
given this size of this PR, it might also be easier/better to break it into a few smaller one/s ( eg your 4 dot points at the top ), but that's up-to you at this point.

davidbuzz:
one more note... your current commit/s shows that you've committed two "submodules" in the "modules" folder.  Please exclude them from your PR/s unless you've specifically/deliverabely made changes to them. 
Often all it needs is running './Tools/gittools/submodule-sync.sh' in your local repo, and it re-sync your submodule/s.

libraries/AP_BattMonitor/AP_BattMonitor_Analog.cpp
arg7: //AR is sort of bookmark indicating problematic block of code (was debugging ADC ), but you are right, it souldn't be commited to the main repo.

libraries/AP_HAL_ESP32/AnalogIn.h
arg7: agree

libraries/AP_HAL_ESP32/HAL_ESP32_Class.cpp
arg7: ok

libraries/AP_HAL_ESP32/RmtSigReader.cpp
arg7: RMT stuff on esps32s3 wasn't working due to SOC_RMT_SUPPORT_RX_PINGPONG bug, before crashing it flooded the console with "RMT buffer is full" messages.
So I tried to add some more buffers with no avail. This change should be ignored.

libraries/AP_HAL_ESP32/Scheduler.cpp
arg7: The issue was that the new Watchdog Timer (WDT) driver did not handle the disabling of the WDT on specific cores well, resulting in the console being flooded with errors about missing WDT initialization. Additionally, I was unable to disable the IDLE task, which made things more complicated. As a result, I decided that explicitly initializing the WDT at the end was a simpler solution, and it worked.

libraries/AP_HAL_ESP32/SdCard.cpp
arg7: ok

libraries/AP_HAL_ESP32/boards/esp32s3devkit.h
arg7: I think it would be better to enhance existing boards (adding pin and sensor description), than proliferating new one for each developer.

If nobody objects to it.

libraries/AP_HAL_ESP32/boards/esp32s3devkit.h
arg7: same as above.

libraries/AP_HAL_ESP32/targets/esp32/esp-idf/sdkconfig.orig
arg7: files esp32s3empty.h and esp32buzz.h ended up in commit by error. I ws testing build system and forgot to cancel modifications.

libraries/AP_RSSI/AP_RSSI.cpp
arg7: agree

arg7:
I will study "Submitting Patches Back to Master" manual and try to fix issues ASAP and fix issues pointed out by you.

libraries/AP_BattMonitor/AP_BattMonitor_Analog.cpp
arg7: My remarks for the points:

1. I agree, I will fix it.

2. I've added the ESP_PLATFORM define in the code branch for ESP32 targets. This works around a bug in ESP-IDF v5.3, which doesn’t define ESP_PLATFORM automatically and breaks some FreeRTOS code.

3. It’s a minor bug. I just noticed it on WSL and fixed it.

4. I tested the build system and install scripts on a clean Ubuntu 24 installation (for both ESP32 and ESP32S3 targets) and tested the built firmware on the ESP32S3 target using the esp32s3devkit.h board definition.
Then the list of fixes:
- Fixed all GCC 13 compatibility issues everywhere.
- Fixed the RCIn issue (it was broken on S3).
- Added enhancements to the ADC (ported to the new driver).
- Fixed WDT issues with ESP-IDF v5.3.
- Added support for the IBUS hack for the FlySky IA6 receiver.

5. I noticed some printf statements that I believe should only be present in debug builds, so I fixed them during debugging. It's not a big deal.

arg7:
I'm closing this PR in order to fix git issues and divide it in the few smaller parts.

PR #28043, opened 2024-09-08
hwdef: remove non-ublox GPSes from MatekL431-Periph

tpwrules:
Saves enough flash to get it building again (11K now free). These backends are default for other boards and so are built in CI.

Tested that it builds and still works with a ublox GPS. Solution as discussed on dev call and https://github.com/ArduPilot/ardupilot/pull/27659 .


peterbarker (approved):
LGTM

IamPete1 (approved):

tridge (approved):

PR #20499, opened 2022-04-10
AP_Scripting: Add bindings for AP_Airspeed

Gone4Dirt:
Adds lua bindings to access airspeed sensors directly in lua.

PR #28044, opened 2024-09-08
Add get_airspeed methode

AlexisMGL:
Add get_airspeed(instance) method, useful for planes with multiple sensors (currently, only the primary sensor is available).
- Added AP_Airspeed chapter in docs.lua and bindings.desc.
- Implemented airspeed:get_airspeed method.
- Tested in SITL with the following_script:

-- This script gives airspeed value if available

function state_init()
    gcs:send_text(6, 'script initiated')
    return state_update, 500
end

function state_update()
    
    local message = string.format("Airspeed: %.2f m/s", airspeed:get_airspeed(0))
    gcs:send_text(6,message)
    return state_update, 6000
end

return state_init,500

### DIFF
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -3209,6 +3209,13 @@ function battery:num_instances() end
 ---@return number|nil
 function battery:get_cell_voltage(instance, cell) end
 
+-- The Airspeed library provides access to airspeed sensors information.
+airspeed = {}
+
+-- desc
+---@param instance integer
+---@return airspeed float
+function airspeed:get_airspeed(instance) end
 
 -- The Arming library provides access to arming status and commands.
 arming = {}
--- a/libraries/AP_Scripting/generator/description/bindings.desc
+++ b/libraries/AP_Scripting/generator/description/bindings.desc
@@ -62,6 +62,10 @@ singleton AP_AHRS method initialised boolean
 singleton AP_AHRS method get_posvelyaw_source_set uint8_t
 singleton AP_AHRS method get_quaternion boolean Quaternion'Null
 
+include AP_Airspeed/AP_Airspeed.h
+singleton AP_Airspeed rename airspeed
+singleton AP_Airspeed method get_airspeed float uint8_t'skip_check
+
 include AP_Arming/AP_Arming.h
 
 singleton AP_Arming rename arming


IamPete1:
We do have a existing PR for this. https://github.com/ArduPilot/ardupilot/pull/20499

I do think you at least also need a binding for `healthy`. 

The further problem is that if you call the `healthy` method with a invalid index you will read past the end the array. Is the unsafe methods that have prevented us from adding bindings so far.....

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3209,6 +3209,13 @@ function battery:num_instances() end
 ---@return number|nil
 function battery:get_cell_voltage(instance, cell) end
 
+-- The Airspeed library provides access to airspeed sensors information.
+airspeed = {}
+
+-- desc
+---@param instance integer
+---@return airspeed float
```
IamPete1, line 3217:
```suggestion
---@return number -- airspeed in m/s -1 if invalid.
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3209,6 +3209,13 @@ function battery:num_instances() end
 ---@return number|nil
 function battery:get_cell_voltage(instance, cell) end
 
+-- The Airspeed library provides access to airspeed sensors information.
+airspeed = {}
+
+-- desc
```
IamPete1, line 3215:
```suggestion
-- Get airspeed for the given sensor index
```

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -62,6 +62,10 @@ singleton AP_AHRS method initialised boolean
 singleton AP_AHRS method get_posvelyaw_source_set uint8_t
 singleton AP_AHRS method get_quaternion boolean Quaternion'Null
 
+include AP_Airspeed/AP_Airspeed.h
```
IamPete1, line 65:
```suggestion
include AP_Airspeed/AP_Airspeed.h
singleton AP_Airspeed depends AP_AIRSPEED_ENABLED
```

IamPete1 (changes requested):

peterbarker:
@AlexisMGL please compare/contrast to the PR from @MattKear which @IamPete1 pointed out

AlexisMGL:
I've added other airspeed methods from #20499. It's unclear to me if I should keep the range check, for instance, or not.

peterbarker:
@AlexisMGL this needs to pass CI...

AlexisMGL:
I can rebase if needed

PR #28051, opened 2024-09-09
AP_Airspeed: healthy: check enabled first for instance range check

IamPete1:
By checking enabled first we can be sure that `i` is a valid instance and don't try and read past the end of the arrays. This is needed if we add a scripting binding without a hard coded range check. https://github.com/ArduPilot/ardupilot/pull/28044


peterbarker (approved):
LGTM

peterbarker:
What testing has this had?

IamPete1:
> What testing has this had?

Only here in CI.

peterbarker:
Merged, thanks!

PR #28052, opened 2024-09-09
AP_HAL_Linux: file descriptor leaking and other issues in "Storage" of Linux

junan76:
- Fixed the fd leaking issue in "_storage_create" found by @peterbarker
- Remove the unnecessary call of "unlinkat" in "_storage_create"
- Simplify the implementation of "init"

### DIFF
--- a/libraries/AP_HAL_Linux/Storage.cpp
+++ b/libraries/AP_HAL_Linux/Storage.cpp
@@ -101,7 +101,6 @@ int Storage::_storage_create(const char *dpath)
         return -1;
     }
 
-    unlinkat(dfd, dpath, 0);
     int fd = openat(dfd, STORAGE_FILE, O_RDWR|O_CREAT|O_CLOEXEC, 0666);
 
     if (fd == -1) {
@@ -114,6 +113,7 @@ int Storage::_storage_create(const char *dpath)
     if (ftruncate(fd, sizeof(_buffer)) == -1) {
         fprintf(stderr, "Failed to set file size to %u kB (%m)\n",
                 unsigned(sizeof(_buffer) / 1024));
+        close(fd);
         goto fail;
     }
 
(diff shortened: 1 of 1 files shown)


peterbarker:
This looks like a nice change.

How has it been tested?

junan76:
> This looks like a nice change.
> 
> How has it been tested?

I tested it on beaglebone black board using gdb, will post the screenshots this evening, you can check the result then.

libraries/AP_HAL_Linux/Storage.cpp
```diff
@@ -145,26 +145,16 @@ void Storage::init()
         dpath = HAL_BOARD_STORAGE_DIRECTORY;
     }
 
-    int fd = open(dpath, O_RDWR|O_CLOEXEC);
+    int fd = _storage_create(dpath);
```
peterbarker, line 148: I wonder if we should rename this to`_storage_open_or_create`.  Just wondering.

junan76:
1. start gdb server on beaglebone board
Screenshot from 2024-09-11 19-09-08
2. during the call of `_storage_create`
  - "dpath" and storage file is opened successfuly
  - `ftruncate` and `fsync` is ok
  - `dfd` is closed and `fd` is returned
Screenshot from 2024-09-11 19-12-43
3. back to `init`
  - we read exactly 16384 bytes data from `fd`
  - `_fd` and `_initialized` are both set with correct value
Screenshot from 2024-09-11 19-14-21
4. connecting the board with qgroundcontrol and set `SCHED_LOOP_RATE` to 100Hz
Screenshot from 2024-09-11 19-17-55

junan76:
Hi peter, What should I do with the `DevCallTopic` label?

Williangalvani:
Just tried it on a Navigator and and everything worked  as expectd

PR #28054, opened 2024-09-10
AP_ESC_Telem: tidy old calls to _telem_data

robertlong13:
When the #27755 (well, the original PR that #27755 was based on anyway) was first opened, #26252 had not been merged yet. #26252 refactored a bit, but the change was not applied to #27755. This PR fixes that.

This should not change the behavior of the code, but it might not be binary identical.


IamPete1 (approved):
Thanks!

amilcarlucas (approved):

andyp1per (approved):

Issue #27925, opened 2024-08-26
AP_DDS - no tests


Ryanf55:
## Bug report

**Issue details**

Something appears to have broken in the colcon tests. None of them are executed.

_Please describe the problem_

**Version**
`master`, `4.5`.

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Logs**

```
$ colcon build --packages-up-to ardupilot_dds_tests
Starting >>> ardupilot_msgs
Finished <<< ardupilot_msgs [0.36s]                     
Starting >>> ardupilot_sitl
Finished <<< ardupilot_sitl [5.18s]                    
Starting >>> ardupilot_dds_tests
/home/ryan/.local/lib/python3.10/site-packages/setuptools/_distutils/dist.py:268: UserWarning: Unknown distribution option: 'tests_require'
  warnings.warn(msg)
--- stderr: ardupilot_dds_tests                   
/home/ryan/.local/lib/python3.10/site-packages/setuptools/_distutils/dist.py:268: UserWarning: Unknown distribution option: 'tests_require'
  warnings.warn(msg)
---
Finished <<< ardupilot_dds_tests [1.09s]

Summary: 3 packages finished [7.10s]
  1 package had stderr output: ardupilot_dds_tests
$ . install/setup.bash
$ colcon test --packages-select ardupilot_dds_tests --event-handlers=console_cohesion+
Starting >>> ardupilot_dds_tests
/home/ryan/.local/lib/python3.10/site-packages/setuptools/_distutils/dist.py:268: UserWarning: Unknown distribution option: 'tests_require'
  warnings.warn(msg)
--- output: ardupilot_dds_tests                   

----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
---
--- stderr: ardupilot_dds_tests

----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
---
Finished <<< ardupilot_dds_tests [0.33s]

Summary: 1 package finished [0.79s]
  1 package had stderr output: ardupilot_dds_tests
$ colcon test-result --all --verbose
Summary: 0 tests, 0 errors, 0 failures, 0 skipped
```

Ryanf55:
ArduPilot is hosing the test environment.

Here, we install the apt version of setuptools
[...]

PR #28056, opened 2024-09-10
Tools: Install setuptools, packaging and wheel through apt

Ryanf55:
# Scope

* Installing these on host OS outside of a VENV can break many other tools
* If in VENV, we upgrade them to the latest just in case
* Remove focal workaround now that upstream fixed it

# Detials

Reverts parts of ArduPilot/ardupilot#27689 because upstream https://github.com/pypa/setuptools/issues/4511 is now fixed, and this change is the root cause of https://github.com/ArduPilot/ardupilot/issues/27925

Instead of a plain revert, I think it's safer and more stable to use the OS supplied version by default.
The maintainers approved this approach in https://github.com/ArduPilot/ardupilot/pull/28035


Tools/environment_install/install-prereqs-ubuntu.sh
```diff
@@ -402,13 +406,6 @@ if [ -n "$PYTHON_VENV_PACKAGE" ]; then
     fi
 fi
 
-# try update packaging, setuptools and wheel before installing pip package that may need compilation
-SETUPTOOLS="setuptools"
-if [ ${RELEASE_CODENAME} == 'focal' ]; then
-    SETUPTOOLS=setuptools==70.3.0
-fi
-$PIP install $PIP_USER_ARGUMENT -U pip packaging $SETUPTOOLS wheel
```
peterbarker, line 410: This changes our setup for *everything*, even if it isn't using venv.

Doesn't this mean we're no longer updating `wheel` and `packaging` outside of those things doing venv?

Ryanf55: Correct. I bundled them together because if you update wheel without updating setuptools, then that is another way to hose the environment. We want to ensure they all stay ABI compatible. Either they all go in as APT and are kept compatible by the OS, or they all stay at latest from pip.

Ryanf55:
Debian buster has an issue: https://github.com/ArduPilot/ardupilot/actions/runs/10786348034/job/29913069056?pr=28056#step:5:2878

Because I didn't change it, not sure what to do.

Ryanf55:
I had to move `pygame` too:
```
134.5 + for PACKAGE in $PYTHON_PKGS
134.5 + pip3 install --user -U pygame
135.0 Collecting pygame
138.5   Downloading https://files.pythonhosted.org/packages/72/49/bd2fcbadb6a55bb24284bad4f530189401c99ffc234d51ba54756a776eb2/pygame-2.6.0.tar.gz (15.8MB)
142.2     Complete output from command python setup.py egg_info:
142.2     Skipping Cython compilation
142.2     
142.2     
142.2     WARNING, No "Setup" File Exists, Running "buildconfig/config.py"
142.2     Using UNIX configuration...
142.2     
142.2     /bin/sh: 1: sdl2-config: not found
142.2     /bin/sh: 1: sdl2-config: not found
142.2     /bin/sh: 1: sdl2-config: not found
142.2     Package freetype2 was not found in the pkg-config search path.
142.2     Perhaps you should add the directory containing `freetype2.pc'
142.2     to the PKG_CONFIG_PATH environment variable
142.2     No package 'freetype2' found
142.2     Package freetype2 was not found in the pkg-config search path.
142.2     Perhaps you should add the directory containing `freetype2.pc'
142.2     to the PKG_CONFIG_PATH environment variable
142.2     No package 'freetype2' found
142.2     Package freetype2 was not found in the pkg-config search path.
142.2     Perhaps you should add the directory containing `freetype2.pc'
142.2     to the PKG_CONFIG_PATH environment variable
142.2     No package 'freetype2' found
142.2     /bin/sh: 1: freetype-config: not found
142.2     /bin/sh: 1: freetype-config: not found
142.2     /bin/sh: 1: freetype-config: not found
142.2     Traceback (most recent call last):
142.2       File "<string>", line 1, in <module>
142.2       File "/tmp/pip-install-ba7a5qs8/pygame/setup.py", line 426, in <module>
142.2         buildconfig.config.main(AUTO_CONFIG)
142.2       File "/tmp/pip-install-ba7a5qs8/pygame/buildconfig/config.py", line 234, in main
142.2         deps = CFG.main(**kwds, auto_config=auto)
142.2       File "/tmp/pip-install-ba7a5qs8/pygame/buildconfig/config_unix.py", line 245, in main
142.2         raise RuntimeError('Unable to run "sdl-config". Please make sure a development version of SDL is installed.')
142.2     RuntimeError: Unable to run "sdl-config". Please make sure a development version of SDL is installed.
142.2     
142.2     Hunting dependencies...
142.2     WARNING: "sdl2-config" failed!
142.2     WARNING: "pkg-config freetype2" failed!
142.2     WARNING: "freetype-config" failed!
142.2     
142.2     ---
142.2     For help with compilation see:
142.2         https://www.pygame.org/wiki/CompileDebian
142.2     To contribute to pygame development see:
142.2         https://www.pygame.org/contribute.html
142.2     ---
142.2     
142.2     
142.2     ----------------------------------------
142.2 Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-ba7a5qs8/pygame/
```

peterbarker:
So it sounds like we need to test every distro we support in there *or* explicitly decide to remove support.

`buster` can *probably* go; nobody should be creating a new environment on it.

Ryanf55:
> So it sounds like we need to test every distro we support in there _or_ explicitly decide to remove support.
> 
> `buster` can _probably_ go; nobody should be creating a new environment on it.



> So it sounds like we need to test every distro we support in there _or_ explicitly decide to remove support.
> 
> `buster` can _probably_ go; nobody should be creating a new environment on it.

As part of this PR, or another? I'd really like to fix the breaking of ROS environments ASAP.

peterbarker:
> As part of this PR, or another? I'd really like to fix the breaking of ROS environments ASAP.



> > So it sounds like we need to test every distro we support in there _or_ explicitly decide to remove support.
> > `buster` can _probably_ go; nobody should be creating a new environment on it.
> 
> > So it sounds like we need to test every distro we support in there _or_ explicitly decide to remove support.
> > `buster` can _probably_ go; nobody should be creating a new environment on it.
> 
> As part of this PR, or another? I'd really like to fix the breaking of ROS environments ASAP.

Anything which can be affected by this PR needs to be tested.  This is why I commented on your change to affect things which don't use venv - it seemed to require a *lot* more testing than the alternative.

Seems like you worked out the problems with buster.

Ryanf55:
> > As part of this PR, or another? I'd really like to fix the breaking of ROS environments ASAP.
> 
> > > So it sounds like we need to test every distro we support in there _or_ explicitly decide to remove support.
> > > `buster` can _probably_ go; nobody should be creating a new environment on it.
> > 
> > 
> > > So it sounds like we need to test every distro we support in there _or_ explicitly decide to remove support.
> > > `buster` can _probably_ go; nobody should be creating a new environment on it.
> > 
> > 
> > As part of this PR, or another? I'd really like to fix the breaking of ROS environments ASAP.
> 
> Anything which can be affected by this PR needs to be tested. This is why I commented on your change to affect things which don't use venv - it seemed to require a _lot_ more testing than the alternative.

Yep, this workflow of this combination of apt deps is working in an isolated Ubuntu 22 environment with ROS 2. 
I've tested mavproxy, sim_vehicle.py with gdb, console, map in both Ubuntu 20, Ubuntu 22, and debian buster.

Is there a better alternative you had in mind that avoids installing the latest version through pip outside a venv? 

> Seems like you worked out the problems with buster.

Yep!

peterbarker:
Have you tested a Bionic install?

Yes, it's EOL.  But we don't want to break things without knowing we have.

Which is why I was trying to suggest structuring this change such that it is clearly only affecting e.g. Noble, and *nothing else*.

That makes it easy to test.  And we avoid the "ah, crap, sorry" moments when someone needs to install on something which we accidentally broke.

The older code should naturally age-out - and that's why I asked about taking buster out.

Ryanf55:
I'm putting this up for dev call because 

> ```shell
> docker build . -t ardupilot --build-arg USER_UID=$(id -u) --build-arg USER_GID=$(id -g) --build-arg BASE_IMAGE=debian --build-arg TAG=buster
> ```

More test results:
* Bionic: python3-pygame is not available. As it stands, this would break bionic users.


If we want to make conditional logic to only do this on `focal` and `jammy`, I can do that.

Ryanf55:
Make a separate PR to remove bionic. 

For tests, check the vagrant file for the lists of manual tests. It's extensive.

Ryanf55:
Bionic is removed. Needs testing on the debian builds better.

peterbarker:
OK, so I've now re-read this and thought about it some more.

One thing about these env install scripts - and particularly the `install-prereqs-ubuntu.sh` script, as it is the most complicated - is that we do not make changes that affect older distributions - not only so we don't break things, but so that people get the same sort of environment over time when they run the install scripts.

If you want to make the argument that we install apt, setuptools, wheel and whatever in the system Python environment, or via apt, or whatever, please do - but we should do that in, for example, `plucky`, not `lunar`!

I *do* have some painful experience with `wxpython` in the system environment vs in venv, so I do know there's a lot of pain to be had in this area :-)

Ryanf55:
Previous behavior was to install apt with setuptools. See here for Plane4.4.
https://github.com/ArduPilot/ardupilot/blob/3e039eb372b196dc169de25e138af6dcab28269b/Tools/environment_install/install-prereqs-ubuntu.sh#L162

I'd prefer to just do that. This needs fixed on Jammy and Noble for ROS users of LTS distros. I don't find your argument of "make changes that affect older distributions" valid - well what I am proposing is going back to the way we did it before on 4.4.

An alternative approach I proposed is to allow ardupilot on Jammy install in a venv. Then it can avoid breaking system setuptools. 
https://github.com/ArduPilot/ardupilot/issues/26137#issuecomment-2869637459

PR #28059, opened 2024-09-10
Attitude Logging (ANG)

andyp1per:
Introduces a new high rate attitude message - ANG - to replace the use of ATT for the attitude controller.

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -566,7 +566,9 @@ void Copter::loop_rate_logging()
 // should be run at 10hz
 void Copter::ten_hz_logging_loop()
 {
-    // log attitude data if we're not already logging at the higher rate
+    // always write AHRS attitude at 10Hz
+    ahrs.Write_Attitude(attitude_control->get_att_target_euler_rad() * RAD_TO_DEG);
+    // log attitude controller data if we're not already logging at the higher rate
     if (should_log(MASK_LOG_ATTITUDE_MED) && !should_log(MASK_LOG_ATTITUDE_FAST) && !copter.flightmode->logs_attitude()) {
         Log_Write_Attitude();
     }
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -1772,6 +1772,10 @@ void QuadPlane::update(void)
     if (motors->armed()) {
         const bool motors_active = in_vtol_mode() || assisted_flight;
         if (motors_active && (motors->get_spool_state() != AP_Motors::SpoolState::SHUT_DOWN)) {
+            // log ANG at main loop rate
+            if (show_vtol_view()) {
+                attitude_control->Write_ANG();
+            }
             // log RATE at main loop rate
             attitude_control->Write_Rate(*pos_control);
 
--- a/ArduSub/ArduSub.cpp
+++ b/ArduSub/ArduSub.cpp
@@ -183,6 +183,7 @@ void Sub::ten_hz_logging_loop()
     // log attitude data if we're not already logging at the higher rate
     if (should_log(MASK_LOG_ATTITUDE_MED) && !should_log(MASK_LOG_ATTITUDE_FAST)) {
         Log_Write_Attitude();
+        attitude_control.Write_ANG();
         attitude_control.Write_Rate(pos_control);
         if (should_log(MASK_LOG_PID)) {
             logger.Write_PID(LOG_PIDR_MSG, attitude_control.get_rate_roll_pid().get_pid_info());
@@ -222,6 +223,7 @@ void Sub::twentyfive_hz_logging()
 {
     if (should_log(MASK_LOG_ATTITUDE_FAST)) {
         Log_Write_Attitude();
+        attitude_control.Write_ANG();
         attitude_control.Write_Rate(pos_control);
         if (should_log(MASK_LOG_PID)) {
             logger.Write_PID(LOG_PIDR_MSG, attitude_control.get_rate_roll_pid().get_pid_info());
--- a/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
+++ b/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
@@ -1,9 +1,7 @@
 #include "AC_AttitudeControl.h"
-#include "AC_PosControl.h"
 #include <AP_HAL/AP_HAL.h>
 #include <AP_Vehicle/AP_Vehicle_Type.h>
 #include <AP_Scheduler/AP_Scheduler.h>
-#include <AP_Logger/AP_Logger.h>
 
 extern const AP_HAL::HAL& hal;
 
@@ -1242,51 +1240,3 @@ void AC_AttitudeControl::get_rpy_srate(float &roll_srate, float &pitch_srate, fl
     pitch_srate = get_rate_pitch_pid().get_pid_info().slew_rate;
     yaw_srate = get_rate_yaw_pid().get_pid_info().slew_rate;
 }
-
-#if HAL_LOGGING_ENABLED
-// Write a rate packet
-void AC_AttitudeControl::Write_Rate(const AC_PosControl &pos_control) const
-{
-    const Vector3f &rate_targets = rate_bf_targets();
-    const Vector3f &accel_target = pos_control.get_accel_target_cmss();
-    const Vector3f &gyro_rate = _rate_gyro;
-    const struct log_Rate pkt_rate{
-        LOG_PACKET_HEADER_INIT(LOG_RATE_MSG),
-        time_us         : _rate_gyro_time_us,
-        control_roll    : degrees(rate_targets.x),
-        roll            : degrees(gyro_rate.x),
-        roll_out        : _motors.get_roll()+_motors.get_roll_ff(),
-        control_pitch   : degrees(rate_targets.y),
-        pitch           : degrees(gyro_rate.y),
-        pitch_out       : _motors.get_pitch()+_motors.get_pitch_ff(),
-        control_yaw     : degrees(rate_targets.z),
-        yaw             : degrees(gyro_rate.z),
-        yaw_out         : _motors.get_yaw()+_motors.get_yaw_ff(),
-        control_accel   : (float)accel_target.z,
-        accel           : (float)(-(_ahrs.get_accel_ef().z + GRAVITY_MSS) * 100.0f),
-        accel_out       : _motors.get_throttle(),
-        throttle_slew   : _motors.get_throttle_slew_rate()
-    };
-    AP::logger().WriteBlock(&pkt_rate, sizeof(pkt_rate));
-
-    /*
-      log P/PD gain scale if not == 1.0
-     */
-    const Vector3f &scale = get_last_angle_P_scale();
-    const Vector3f &pd_scale = _pd_scale_used;
-    if (scale != AC_AttitudeControl::VECTORF_111 || pd_scale != AC_AttitudeControl::VECTORF_111) {
-        const struct log_ATSC pkt_ATSC {
-            LOG_PACKET_HEADER_INIT(LOG_ATSC_MSG),
-            time_us  : _rate_gyro_time_us,
-            scaleP_x : scale.x,
-            scaleP_y : scale.y,
-            scaleP_z : scale.z,
-            scalePD_x : pd_scale.x,
-            scalePD_y : pd_scale.y,
-            scalePD_z : pd_scale.z,
-        };
-        AP::logger().WriteBlock(&pkt_ATSC, sizeof(pkt_ATSC));
-    }
-}
-
-#endif  // HAL_LOGGING_ENABLED
--- a/libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
+++ b/libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
@@ -1,0 +1,75 @@
+#include <AP_Logger/AP_Logger_config.h>
+
+#if HAL_LOGGING_ENABLED
+
+#include "AC_AttitudeControl.h"
+#include "AC_PosControl.h"
+#include <AP_Logger/AP_Logger.h>
+#include <AP_Scheduler/AP_Scheduler.h>
+#include "LogStructure.h"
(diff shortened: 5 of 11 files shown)


libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -1,7 +1,9 @@
 #include "AC_AttitudeControl.h"
+#include "AC_PosControl.h"
 #include <AP_HAL/AP_HAL.h>
 #include <AP_Vehicle/AP_Vehicle_Type.h>
 #include <AP_Scheduler/AP_Scheduler.h>
+#include <AP_Logger/AP_Logger.h>
```
peterbarker, line 6: Not needed

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -1,7 +1,9 @@
 #include "AC_AttitudeControl.h"
+#include "AC_PosControl.h"
```
peterbarker, line 2: Not needed?

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
```diff
@@ -0,0 +1,82 @@
...
+        pitch           : attitude_body_angles.y,
+        control_yaw     : wrap_360(targets.z),
+        yaw             : wrap_360(attitude_body_angles.z),
+        sensor_dt       : AP::scheduler().get_last_loop_time_s()
+    };
+    AP::logger().WriteBlock(&pkt, sizeof(pkt));
+}
+
+// Write a rate packet
+void AC_AttitudeControl::Write_Rate(const AC_PosControl &pos_control) const
```
peterbarker, line 38: Taking the position controller here is a little odd.  Perhaps we should pass in the targets instead like we do for `ATT`?

... it's a bit off that we don't persist the targets within here somehow already...

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
```diff
@@ -0,0 +1,82 @@
...
+    Vector3f targets = get_att_target_euler_rad() * RAD_TO_DEG;
+    // Retrieve quaternion body attitude
+    Quaternion attitude_body;
+    _ahrs.get_quat_body_to_ned(attitude_body);
+    // calculate the attitude body euler angles
+    Vector3f attitude_body_angles;
+    attitude_body.to_euler(attitude_body_angles);
+    attitude_body_angles *= RAD_TO_DEG;
+
+    const struct log_Ang pkt{
```
peterbarker, line 23:
```suggestion
    const struct log_ANG pkt{
```

ArduSub/ArduSub.cpp
```diff
@@ -183,7 +183,8 @@ void Sub::ten_hz_logging_loop()
     // log attitude data if we're not already logging at the higher rate
     if (should_log(MASK_LOG_ATTITUDE_MED) && !should_log(MASK_LOG_ATTITUDE_FAST)) {
         Log_Write_Attitude();
-        ahrs_view.Write_Rate(motors, attitude_control, pos_control);
+        attitude_control.Write_Attitude();
+        attitude_control.Write_Rate(pos_control);
```
peterbarker, line 187: Write_attitude already logs multiple messages, perhaps we should have a generic, "attitude control, do your own logging now" call.

I'd actually prefer the attitude controller to have its own logging control parameters, but this might be divisive.

peterbarker (changes requested):

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
andyp1per: Fixed

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
andyp1per: Its better than it was - took 3 arguments. Regardless this is code that has already been approved and merged so I don't propose to dork with it here.

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
andyp1per: Fixed

ArduSub/ArduSub.cpp
andyp1per: Actually no - they will be called from different threads at different rates so need to be kept different. Sub logging is already messed up - really someone needs to go in and decide whether they really need all this stuff at high rates.

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
```diff
@@ -0,0 +1,82 @@
...
+#if HAL_LOGGING_ENABLED
+
+#include "AC_AttitudeControl.h"
+#include "AC_PosControl.h"
+#include <AP_Logger/AP_Logger.h>
+#include <AP_Scheduler/AP_Scheduler.h>
+#include "LogStructure.h"
+
+// Write an ANG packet
+void AC_AttitudeControl::Write_Attitude() const
```
peterbarker, line 12:
```suggestion
void AC_AttitudeControl::Write_ANG() const
```

peterbarker:
Perhaps a test to make sure `ATT.DesRoll` and `ANG.DesRoll` match in normal operation

Could use
```
        dfreader = self.dfreader_for_current_onboard_log()
```

ArduPlane/quadplane.cpp
```diff
@@ -1763,6 +1763,8 @@ void QuadPlane::update(void)
     if (motors->armed()) {
         const bool motors_active = in_vtol_mode() || assisted_flight;
         if (motors_active && (motors->get_spool_state() != AP_Motors::SpoolState::SHUT_DOWN)) {
+            // log ANG at main loop rate
+            attitude_control->Write_Attitude();
```
tridge, line 1767: should not call then when use_fw_attitude_controllers() is true

tridge (changes requested):

andyp1per:
[00000002.zip](https://github.com/user-attachments/files/16959923/00000002.zip)

ArduPlane/quadplane.cpp
andyp1per: Fixed

libraries/AC_AttitudeControl/LogStructure.h
```diff
@@ -104,4 +127,6 @@ struct PACKED log_Rate {
     { LOG_PSCD_MSG, sizeof(log_PSCx), \
       "PSCD", PSCx_FMT, "TimeUS,TPD,PD,DVD,TVD,VD,DAD,TAD,AD", PSCx_UNITS, PSCx_MULTS }, \
     { LOG_RATE_MSG, sizeof(log_Rate), \
-        "RATE", "Qfffffffffffff",  "TimeUS,RDes,R,ROut,PDes,P,POut,YDes,Y,YOut,ADes,A,AOut,AOutSlew", "skk-kk-kk-oo--", "F?????????BB--" , true }
+        "RATE", "Qfffffffffffff",  "TimeUS,RDes,R,ROut,PDes,P,POut,YDes,Y,YOut,ADes,A,AOut,AOutSlew", "skk-kk-kk-oo--", "F?????????BB--" , true }, \
```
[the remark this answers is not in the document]
andyp1per: Will be done in separate PR

peterbarker (approved):
LGTM

libraries/AC_AttitudeControl/LogStructure.h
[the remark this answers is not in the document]
lthall: This should probably be a separate PR.

ArduPlane/quadplane.cpp
```diff
@@ -1763,6 +1763,10 @@ void QuadPlane::update(void)
     if (motors->armed()) {
         const bool motors_active = in_vtol_mode() || assisted_flight;
         if (motors_active && (motors->get_spool_state() != AP_Motors::SpoolState::SHUT_DOWN)) {
+            // log ANG at main loop rate
+            if (!use_fw_attitude_controllers()) {
```
IamPete1, line 1767:
```suggestion
            if (show_vtol_view()) {
```
That is the method we use to swap the existing attitude log from FW to VTOL.

ArduCopter/Copter.cpp
```diff
@@ -566,7 +566,9 @@ void Copter::loop_rate_logging()
 // should be run at 10hz
 void Copter::ten_hz_logging_loop()
 {
-    // log attitude data if we're not already logging at the higher rate
+    // always write AHRS attitude at 10Hz
+    ahrs.Write_Attitude(attitude_control->get_att_target_euler_rad() * RAD_TO_DEG);
```
IamPete1, line 570: I think I would vote to still log `ATT` at 25hz if medium logging is selected, but it make sense to drop it from 400Hz logging.

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
```diff
@@ -0,0 +1,82 @@
...
+            scaleP_z : scale.z,
+            scalePD_x : pd_scale.x,
+            scalePD_y : pd_scale.y,
+            scalePD_z : pd_scale.z,
+        };
+        AP::logger().WriteBlock(&pkt_ATSC, sizeof(pkt_ATSC));
+    }
+}
+
+#endif // HAL_LOGGING_ENABLED
```
IamPete1, line 82:
```suggestion
#endif // HAL_LOGGING_ENABLED

```

libraries/AC_AttitudeControl/LogStructure.h
```diff
@@ -92,6 +93,28 @@ struct PACKED log_Rate {
     float   throttle_slew;
 };
 
+// @LoggerMessage: ANG
+// @Description: Canonical vehicle attitude
```
IamPete1, line 97: Not any more ;) that was `ATT`.

libraries/AC_AttitudeControl/LogStructure.h
```diff
@@ -92,6 +93,28 @@ struct PACKED log_Rate {
...
+// @Field: TimeUS: Time of gyro sample
```
IamPete1, line 98: Maybe "attitude sample" its a bit odd mentioning gyro here on a angle message.

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
```diff
@@ -0,0 +1,82 @@
...
+void AC_AttitudeControl::Write_ANG() const
+{
+    Vector3f targets = get_att_target_euler_rad() * RAD_TO_DEG;
+    // Retrieve quaternion body attitude
+    Quaternion attitude_body;
+    _ahrs.get_quat_body_to_ned(attitude_body);
+    // calculate the attitude body euler angles
+    Vector3f attitude_body_angles;
+    attitude_body.to_euler(attitude_body_angles);
+    attitude_body_angles *= RAD_TO_DEG;
```
IamPete1, line 21: Why not use `degrees(_ahrs.roll)`, `degrees(_ahrs.pitch)`, `degrees(_ahrs.yaw)`? AHRS has done the conversion for us already.

ArduCopter/Copter.cpp
andyp1per: My understanding from the dev call was to log at 10Hz only. I'm not going to open that can of worms again.

libraries/AC_AttitudeControl/LogStructure.h
andyp1per: Fixed

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
andyp1per: Leonard wants this as separate from AHRS as possible. Other changes incoming in future PR's

libraries/AC_AttitudeControl/LogStructure.h
lthall: Maybe: "Time stamp of the current Attitude loop"

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
lthall: This is attitude referencing used by the Attitude Controller, independent of its source.

lthall: For now you are right. We should just use _ahrs.ect versions.

ArduCopter/Copter.cpp
lthall: Yes, the consensus was that ATT should be logged at a fixed rate of 10 Hz.

The idea is to move away from the ATT message. Why do you want to log ATT faster?

lthall: The current behaviour is to log ATT at 10 Hz if MASK_LOG_ATTITUDE_MED is on and not logging it at all if MASK_LOG_ATTITUDE_MED is off.

It looks like MASK_LOG_ATTITUDE_FAST is being used to log EKF_POS at 10 Hz.

These are a bit of a mess.

andyp1per: Note that ATT is not currently logged at 25Hz. Other things are but not ATT, at least not in copter.

libraries/AC_AttitudeControl/LogStructure.h
andyp1per: Fixed

libraries/AC_AttitudeControl/AC_AttitudeControl_Logging.cpp
andyp1per: Fixed

tridge (approved):

rmackay9:
I guess I wasn't involved in the review of this (probably my fault for not paying attention) but I think it's a bit odd that we now have both ATT and ANG messages with nearly exactly the same information in them.  It's hard to imagine what argument would not allow increasing the rate of the ATT message but would allow a duplicate message

Issue #28060, opened 2024-09-10
Copter: _AutoTakeoff's complete_pos is never consumed


rmackay9:
A small typo in PR https://github.com/ArduPilot/ardupilot/pull/25171 means that _AutoTakeoff::get_position() never returns the completion position of a takeoff command.

The issue has been there for some time and no users have noticed so I think it is fairly innocuous.  It results in the ModeAuto::wp_start() using an all zero stopping point which, when passed in AC_WPNav::wp_and_spline_init() results in the vehicle's stopping point being used instead of the takeoff's target position.  The two are normally quite close so this just means that the SCurve's starting point will be slightly off from the takeoff's final target.

rmackay9:
this is fixed now by the above linked PR

PR #28061, opened 2024-09-10
Copter: auto mode takeoff complete pos fix

rmackay9:
This resolves issue https://github.com/ArduPilot/ardupilot/issues/28060 by resolving the mixup of variable names in _AutoTakeoff::get_position() method.

I've also renamed the method and added some comments in the hope that this makes its purpose a little more clear.

This has been lightly tested in SITL using a simple mission consisting of just a takeoff and a waypoint command.
test-mission

Below is a screenshot of the target position (in the Easterly direction) and "before" we see a slight jump while "after" it is gone.
before
after

Thanks to @peterbarker for helping with the investigation and fix

### DIFF
--- a/ArduCopter/mode.h
+++ b/ArduCopter/mode.h
@@ -15,7 +15,7 @@ class _AutoTakeoff {
 public:
     void run();
     void start(float complete_alt_cm, bool terrain_alt);
-    bool get_position(Vector3p& completion_pos);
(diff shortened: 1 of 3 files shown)


peterbarker:
Closes https://github.com/ArduPilot/ardupilot/issues/28060

PR #28064, opened 2024-09-10
Create hwdef-bl.dat

Nuts214:
SpeedyBeeF405WING add bdshot support


andyp1per:
Please create a single PR with these 3 commits in

PR #28067, opened 2024-09-10
SpeedyBeeF405WING with bdshot support

setupfpv:


setupfpv:
now you can get the data of the esc temperature and rmp using bidirectional-dshot on Bluejay firmware.

PR #28069, opened 2024-09-10
VectorNav: rename dataflash log message names

weavVN:
Change the DataFlash log message names in the VectorNav external AHRS driver to make the log messages better indicate they are from the VectorNav driver.


peterbarker (approved):
LGTM.

I've fixed the commit message up and will mark for MergeOnCIPass.

Note our usage of "dataflash" is generally incorrect and a bad habit we're all trying to break.   You're not helping :-)

PR #28071, opened 2024-09-10
AP_Scripting: allow setting throttle factor from add_motor_raw

magate:
`add_motor_raw` defaults throttle factor to 1: https://github.com/ArduPilot/ardupilot/blob/bb2249f7667bfbf8dee724c2b0591d9a67f1b86d/libraries/AP_Motors/AP_MotorsMatrix.h#L79

There is a scripting `set_throttle_factor`. This does not seem to be working with frame class (https://ardupilot.org/plane/docs/parameters-Plane-stable-V4.5.6.html#q-frame-class-frame-class) 15 though.


IamPete1:
You can't add fields without breaking existing scripts.

It certainly should work, they key thing is that you have to set the factors after adding the motor but prior to calling the init. We have a partial example with the static matrix:

https://github.com/ArduPilot/ardupilot/blob/66a2788663cd345a8624c889b18152d637364f2d/libraries/AP_Scripting/applets/x-quad-cg-allocation.lua#L166-L187

What behavior are you seeing?

magate:
> You can't add fields without breaking existing scripts.
> 
> It certainly should work, they key thing is that you have to set the factors after adding the motor but prior to calling the init. We have a partial example with the static matrix:
> 
> https://github.com/ArduPilot/ardupilot/blob/66a2788663cd345a8624c889b18152d637364f2d/libraries/AP_Scripting/applets/x-quad-cg-allocation.lua#L166-L187
> 
> What behavior are you seeing?

This looks like what I want. I'll play with it a bit more. Thanks.

PR #28072, opened 2024-09-11
Plane: added Q_APPROACH_DIST

tridge:
this sets a minimum distance to use the fixed wing approach logic. It is an alternative to just disabling the approach with Q_OPTIONS which some users do to avoid some short distance problems. This allows the approach to still be used for longer distances in QRTL but have it disabled for shorter distances
The approach code is really needed for longer distances to allow the full fixed wing terrain following logic to be used

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -538,6 +538,15 @@ const AP_Param::GroupInfo QuadPlane::var_info2[] = {
     // @User: Standard
     AP_GROUPINFO("BCK_PIT_LIM", 38, QuadPlane, q_bck_pitch_lim, 10.0f),
 
+    // @Param: APPROACH_DIST
+    // @DisplayName: Q mode approach distance
+    // @Description: The minimum distance from the destination to use the fixed wing airbrake and approach code for landing approach. This is useful if you don't want the fixed wing approach logic to be used when you are close to the destination. Set to zero to always use fixed wing approach.
+    // @Units: m
+    // @Range: 0.0 1000
+    // @Increment: 1
+    // @User: Standard
+    AP_GROUPINFO("APPROACH_DIST", 39, QuadPlane, approach_distance, 0),
+    
     AP_GROUPEND
 };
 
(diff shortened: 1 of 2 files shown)


ArduPlane/quadplane.cpp
```diff
@@ -538,6 +538,15 @@ const AP_Param::GroupInfo QuadPlane::var_info2[] = {
     // @User: Standard
     AP_GROUPINFO("BCK_PIT_LIM", 38, QuadPlane, q_bck_pitch_lim, 10.0f),
 
+    // @Param: APPROACH_DIST
+    // @DisplayName: Q mode approach distance
+    // @Description: The minimum distance from the destination to use the fixed wing airbrake and approach code for landing approach. This is useful if you don't want the fixed wing approach logic to be used when you are close to the destination. Set to zero to use the default value of 1.5 times the calculated stopping_distance based on cruise airspeed and Q_TRANS_DECEL.
```
IamPete1, line 543:
```suggestion
    // @Description: The minimum distance from the destination to use the fixed wing airbrake and approach code for landing approach. This is useful if you don't want the fixed wing approach logic to be used when you are close to the destination. Set to zero to always use fixed wing approach. 
```
I don't see the behavior the descriptions described for the zero case. There is `transition_threshold` function, but that switches straight to air brake rather than pos 1.

timtuxworth:
This seems like a good idea, but the interaction with the Q_OPTIONS is very confusing. If I understand this correctly, DISABLE_APPROACH actually means disable fixed wing approach, is that right? and if so the new behavior seems to be:

16: Disable Approach - always disable use of **fixed wing** approach and airbrake stages in VTOL landing, overriding Q_APPROACH_DIST if set

Do I have that right?

ArduPlane/quadplane.cpp
```diff
@@ -538,6 +538,15 @@ const AP_Param::GroupInfo QuadPlane::var_info2[] = {
...
+    // @Units: m
+    // @Range: 0.0 1000
+    // @Increment: 1
+    // @User: Standard
+    AP_GROUPINFO("APPROACH_DIST", 39, QuadPlane, approach_distance, 0),
```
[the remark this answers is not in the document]
IamPete1: Q_FW_LND_APR_RAD is related to the circle before transition into the wind stuff. Its not the same thing.

ArduPlane/quadplane.cpp
tridge: it does do a 1s airbrake, then goes to pos1. It skips the approach code

tridge: I've changed it anyway

Hwurzburg:
@tridge does this impact the code that prevents switch to fw approach if in vtol closer than 1.5x maxrad? I assume not..

PR #28075, opened 2024-09-11
AP_EFI: Hirth: fix sensor health flags

robertlong13:
There is no crankshaft sensor status reported by this EFI. This line is misleading and should be removed. The sensor health bitmask is already logged elsewhere.

### DIFF
--- a/libraries/AP_EFI/AP_EFI_Serial_Hirth.cpp
+++ b/libraries/AP_EFI/AP_EFI_Serial_Hirth.cpp
@@ -33,7 +33,6 @@
 
 #define ENGINE_RUNNING 4
 #define THROTTLE_POSITION_FACTOR 10
-#define CRANK_SHAFT_SENSOR_OK 0x0F
 #define INJECTION_TIME_RESOLUTION 0.8
 #define THROTTLE_POSITION_RESOLUTION 0.1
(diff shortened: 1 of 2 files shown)


peterbarker:
How did we end up with this in here, then?

Is it in the datasheet?

robertlong13:
> How did we end up with this in here, then?
> 

I really don't know how this got there. I asked around but didn't get a confident answer.

> Is it in the datasheet?
> 

No, that `sensor_ok` bitmask holds 4 bits for 4 different sensor healths, none of which are crankshaft, and all 4 of which are logged later correctly. From the bitwise `&` here, you'd think that `CRANK_SHAFT_SENSOR_OK` held a single bit, but it doesn't: its 0x0F (4 LSB set). So this line is effectively marking crankshaft sensor healthy if any of these 4 sensors are healthy, and unhealthy if all are unhealthy.

peterbarker (approved):

robertlong13:
While I was here, I noticed that our other sensor health bitmask is actually wrong. The datasheet shows bits 1-4, not bits 0-3, and I have confirmed with my Hirth EFI that bit 0 always reports 0.

I have pushed another commit to this PR to address that too.

peterbarker (approved):

peterbarker:
Merged, thanks!

PR #26955, opened 2024-05-02
Rover: Determine common conditions first

muramura:
Common decisions are made first.
This reduces the number of unnecessary decisions.

PR #23430, opened 2023-04-07
AP_DDS: Implement GCS_SEND_TEXT over DDS

Ryanf55:
This is just a draft for now, but it's an idea on how the call to GCS_SEND_TEXT will end up using rosout to send the information to the companion computer, or a DDS-enabled ground station. Due to the difficulties of using AP_GCS, I'll leave this out as a draft for now, until someone can suggest an approach to implement it.

PR #28076, opened 2024-09-11
Rover: reject position targets if acceleration supplied

peterbarker:
Replaces https://github.com/ArduPilot/ardupilot/pull/26955/files which simply factored the check out.

This both factors the check out and stops us accepting the position target if acceleration is set.

We should probably also reject if other bits and pieces are set.

### DIFF
--- a/Rover/GCS_MAVLink_Rover.cpp
+++ b/Rover/GCS_MAVLink_Rover.cpp
@@ -675,6 +675,14 @@ void GCS_MAVLINK_Rover::handle_set_attitude_target(const mavlink_message_t &msg)
     }
 }
 
+// if we receive a message where the user has not masked out
+// acceleration from the input packet we send a curt message
+// informing them:
+void GCS_MAVLINK_Rover::send_acc_ignore_must_be_set_message(const char *msgname)
+{
+    GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Ignoring %s; set ACC_IGNORE in mask", msgname);
+}
+
 void GCS_MAVLINK_Rover::handle_set_position_target_local_ned(const mavlink_message_t &msg)
 {
     // decode packet
@@ -782,6 +790,7 @@ void GCS_MAVLINK_Rover::handle_set_position_target_local_ned(const mavlink_messa
 
     if (!acc_ignore) {
         // ignore any command where acceleration is not ignored
+        send_acc_ignore_must_be_set_message("SET_POSITION_TARGET_LOCAL_NED");
         return;
     }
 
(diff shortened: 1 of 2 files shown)


rmackay9:
Thanks @peterbarker for this.  I guess this has been tested in some way?

peterbarker:
```
CubeOrange-periph-heavy  *                                                                     
Durandal                            *      *           *       *                 *      -24    *
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     *      *           *       *                 *      8      *
MatekF405                           *      *           *       *                 *      0      *
Pixhawk1-1M-bdshot                  *                  *       *                 *      0      *
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           *      *           *       *                 *      0      *
```

peterbarker:
Tested in SITL with these commands:
```
# this one should move the vehicle:
# message set_position_target_local_ned 0 1 1 1 0b1111111111111000 10 10 0 0 0 0 0 0 0 0 0

# this one should be ignored as acceleration is supplied along with position:
# message set_position_target_local_ned 0 1 1 1 0b1111111000111000 10 10 0 0 0 0 0 0 0 0 0

# should work:
# message set_position_target_global_int 0 1 1 0 0b1111111111111000 -353630399 1491652946 0 0 0 0 0 0 0 0 0 

# acceleration not ignored:
# message set_position_target_global_int 0 1 1 0 0b1111111000111000 -353630399 1491652946 0 0 0 0 0 0 0 0 0 
```

(and gdb and breakpoints)

rmackay9 (approved):

rmackay9 (changes requested):
We discussed on the dev call and decided that we should add some text feedback to the user if their message is rejected because it's malformed

Ryanf55:
Relates to #23430 which has the same inability to share errors back to users for silently dropping invalid commands

peterbarker:
As requested:

```
GUIDED> message set_position_target_global_int 0 1 1 0 0b1111111000111000 -353630399 1491652946 0 0 0 0 0 0 0 0 0 
GUIDED> AP: Ignoring SET_POSITION_TARGET_GLOBAL_INT; set ACC_IGNORE in mask
```

Rover/GCS_Mavlink.cpp
```diff
@@ -845,6 +845,11 @@ void GCS_MAVLINK_Rover::handle_set_attitude_target(const mavlink_message_t &msg)
     }
 }
 
+void GCS_MAVLINK_Rover::send_acc_ignore_must_be_set_message(const char *msgname)
```
rmackay9, line 848: sorry to be annoying but can you add a comment above the method?  Just copying the method from the .h is fine.

rmackay9 (approved):
after the comment update, feel free to merge, thanks!

Rover/GCS_Mavlink.cpp
peterbarker: Done

PR #26987, opened 2024-05-06
AP_Compass: Add in BMM350 Driver

cuav-chen2:

PR #28079, opened 2024-09-11
AP_Compass: Add in BMM350 Driver 

peterbarker:
Replaceas https://github.com/ArduPilot/ardupilot/pull/26987 as I could not force-push the branch


cuav-chen2:
I have just modified my pr base on your suggestions and still making a test... Any way, thank you for that!

peterbarker:
Has to wait until https://github.com/ArduPilot/ardupilot/pull/28088 goes in

Fixed a problem with `new` vs `NEW_NOTHROW`

libraries/AP_Compass/AP_Compass_BMM350.cpp
```diff
@@ -0,0 +1,492 @@
...
+    float magx = (float)magx_raw * BMM350_XY_SCALE;
+    float magy = (float)magy_raw * BMM350_XY_SCALE;
+    float magz = (float)magz_raw * BMM350_Z_SCALE;
+    float temp = (float)temp_raw * BMM350_TEMP_SCALE;
+
+    if (temp > 0.0f) {
+        temp -= 25.49f;
+    } else if (temp < 0.0f) {
+        temp += 25.49f;
+    }
```
tpwrules, line 462: This seems quite suspect

libraries/AP_Compass/AP_Compass_BMM350.cpp
```diff
@@ -0,0 +1,492 @@
...
+
+    const float cr_ax_comp_x = (magx - _mag_comp.cross_axis.cross_x_y * magy) / (1 - _mag_comp.cross_axis.cross_y_x * _mag_comp.cross_axis.cross_x_y);
+    const float cr_ax_comp_y = (magy - _mag_comp.cross_axis.cross_y_x * magx) / (1 - _mag_comp.cross_axis.cross_y_x * _mag_comp.cross_axis.cross_x_y);
+    const float cr_ax_comp_z = (magz + (magx * (_mag_comp.cross_axis.cross_y_x * _mag_comp.cross_axis.cross_z_y - _mag_comp.cross_axis.cross_z_x) - 
+                                        magy * (_mag_comp.cross_axis.cross_z_y - _mag_comp.cross_axis.cross_x_y * _mag_comp.cross_axis.cross_z_x)) /
+                                       (1 - _mag_comp.cross_axis.cross_y_x * _mag_comp.cross_axis.cross_x_y));
+
+    // Store in field vector and convert uT to milligauss
+    Vector3f field { cr_ax_comp_x * 10.0f, cr_ax_comp_y * 10.0f, cr_ax_comp_z * 10.0f };
+    accumulate_sample(field, _compass_instance);
```
tpwrules, line 484:
```suggestion
    // Store in field vector and convert uT to milligauss
    Vector3f field { cr_ax_comp_x, cr_ax_comp_y, cr_ax_comp_z };
    field *= 10.0f;
    accumulate_sample(field, _compass_instance);
```

libraries/AP_Compass/AP_Compass_BMM350.cpp
peterbarker: Ping @cuav-chen2 - do you see a problem with these lines?

libraries/AP_Compass/AP_Compass_BMM350.cpp
peterbarker: I did consider suggesting this change but decided against in this instance.  For no particular reason - perhaps just for re-testing reasons.

libraries/AP_Compass/AP_Compass_BMM350.cpp
cuav-chen2: oh...my fault. We should compare with `FLT_EPSILON` here.

tpwrules: I think there is something more serious. Does the offset really change depending on the temperature sign? Right now if the input temperature value changes from e.g. 0.001 to -0.001 the reported temperature will change by 50°C (in the opposite direction too!)

cuav-chen2: Maybe you're right. I got this handle logic from the official routine of the sensor. I could do some testing on it if possible, but I'm not sure if I could control the temperature to this critical value.

peterbarker: Have you checked the data sheet?  This should definitely be in there.

Speaking of which, dropping a link to the datasheet in the comment at the top of the header is always a nice touch.

tpwrules: The datasheet is particularly unhelpful, despite being 63 pages it doesn't have any description of the values and just refers to a broken code link...

I did do some searching and found the [code](https://github.com/boschsensortec/BMM350_SensorAPI/blob/4127534c6262f903683ee210b36b85ab2bb504c8/bmm350.c#L1424-L1435) and it does seem to be equivalent to what is proposed here, so I guess I can't argue any further.

Ardupilot does not log magnetometer temperature and Bosch's official API does not document the temperature units anyhow.

peterbarker: OK, I'm marking this for merge - if it doesn't work we can fix it, nothing will crash because of this :-)

peterbarker:
@cuav-chen2 merged this one

PR #28081, opened 2024-09-11
Fence: Change AC_FENCE_ACTION to an enum

muramura:
Change AC_FENCE_ACTION to an ENUM. 
Using an ENUM allows for grouping actions. 
I believe using an ENUM is better than indicating groups with define names.


libraries/AC_Fence/AC_Fence.h
```diff
@@ -20,14 +20,16 @@
...
 // valid actions should a fence be breached
-#define AC_FENCE_ACTION_REPORT_ONLY                 0       // report to GCS that boundary has been breached but take no further action
-#define AC_FENCE_ACTION_RTL_AND_LAND                1       // return to launch and, if that fails, land
-#define AC_FENCE_ACTION_ALWAYS_LAND                 2       // always land
-#define AC_FENCE_ACTION_SMART_RTL                   3       // smartRTL, if that fails, RTL, it that still fails, land
-#define AC_FENCE_ACTION_BRAKE                       4       // brake, if that fails, land
-#define AC_FENCE_ACTION_SMART_RTL_OR_LAND           5       // SmartRTL, if that fails, Land
-#define AC_FENCE_ACTION_GUIDED                      6       // guided mode, with target waypoint as fence return point
-#define AC_FENCE_ACTION_GUIDED_THROTTLE_PASS        7       // guided mode, but pilot retains manual throttle control
+enum class AC_FENCE_ACTION : uint8_t {
```
peterbarker, line 23: Move this into the fence class, and rename it to `Action`.

peterbarker (changes requested):

Ryanf55:
I like this kind of PR. Any motive for mixing of c-style cast and static cast?

muramura:
@Ryanf55 san.   I changed it to static_cast.

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -878,7 +878,7 @@ bool AC_Fence::sys_status_enabled() const
     if (!sys_status_present()) {
         return false;
     }
-    if (_action == AC_FENCE_ACTION_REPORT_ONLY) {
+    if (_action == uint8_t(AC_FENCE_ACTION::REPORT_ONLY)) {
```
Ryanf55, line 881:
```suggestion
    if (_action == static_cast<uint8_t>(AC_FENCE_ACTION::REPORT_ONLY)) {
```

muramura: @Ryanf55 san, @peterbarker san.
I changed the cast to static_cast<AC_Fence::Action>.

Ryanf55: Excellent.

ArduCopter/fence.cpp
```diff
@@ -30,8 +30,8 @@ void Copter::fence_check()
         }
 
         // if the user wants some kind of response and motors are armed
-        uint8_t fence_act = fence.get_action();
-        if (fence_act != AC_FENCE_ACTION_REPORT_ONLY ) {
+        AC_Fence::Action fence_act = static_cast<AC_Fence::Action>(fence.get_action());
```
peterbarker, line 33:
```suggestion
        const AC_Fence::Action fence_act = static_cast<AC_Fence::Action>(fence.get_action());
```

ArduPlane/fence.cpp
```diff
@@ -76,14 +76,18 @@ void Plane::fence_check()
         fence.print_fence_message("breached", new_breaches);
 
         // if the user wants some kind of response and motors are armed
-        const uint8_t fence_act = fence.get_action();
+        const AC_Fence::Action fence_act = static_cast<AC_Fence::Action>(fence.get_action());
```
peterbarker, line 79: Why does this need a static cast?  Similar question elsewhere.

Surely `get_action` should return an `AC_Fence::Action` now?

peterbarker (changes requested):
I like this a lot.

Please state what testing it has had.

peterbarker:
Ping @muramura don't forget this one

PR #28090, opened 2024-09-12
AP_MSP: MSP_RAW_GPS cog should be decidegrees not centidegrees

catkira:
image

source: http://www.multiwii.com/wiki/index.php?title=Multiwii_Serial_Protocol

### DIFF
--- a/libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
+++ b/libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
@@ -217,7 +217,7 @@ void AP_MSP_Telem_Backend::update_gps_state(gps_state_t &gps_state)
         gps_state.lon = loc.lng;
         gps_state.alt_m = loc.alt/100; // 1m resolution
         gps_state.speed_cms = gps.ground_speed() * 100;
-        gps_state.ground_course_cd = gps.ground_course_cd();
+        gps_state.ground_course_dd = gps.ground_course_cd() / 10;
     }
 }
 #endif
--- a/libraries/AP_MSP/AP_MSP_Telem_Backend.h
+++ b/libraries/AP_MSP/AP_MSP_Telem_Backend.h
@@ -57,7 +57,7 @@ public:
         int32_t lon;
         uint16_t alt_m;
         uint16_t speed_cms;
-        int16_t ground_course_cd;
+        uint16_t ground_course_dd;
     } gps_state_t;
 
     typedef struct airspeed_state_s {


catkira:
I think we should also make ground_course_dd uint16_t

peterbarker:
Looks likely. 

Commit message should be `AP_MSP:`, not `fix:`.  Want to fix that or shall I?

catkira:
I can fix. Should I also change it to uint16_t?

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
```diff
@@ -217,7 +217,7 @@ void AP_MSP_Telem_Backend::update_gps_state(gps_state_t &gps_state)
         gps_state.lon = loc.lng;
         gps_state.alt_m = loc.alt/100; // 1m resolution
         gps_state.speed_cms = gps.ground_speed() * 100;
-        gps_state.ground_course_cd = gps.ground_course_cd();
+        gps_state.ground_course_dd = gps.ground_course_cd() * 10;
```
amilcarlucas, line 220: To convert from centi-degrees to deci-degrees you need to divide by 10

amilcarlucas (changes requested):

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
catkira: yes, good catch :)

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
```diff
@@ -217,7 +217,7 @@ void AP_MSP_Telem_Backend::update_gps_state(gps_state_t &gps_state)
         gps_state.lon = loc.lng;
         gps_state.alt_m = loc.alt/100; // 1m resolution
         gps_state.speed_cms = gps.ground_speed() * 100;
-        gps_state.ground_course_cd = gps.ground_course_cd();
+        gps_state.ground_course_dd = gps.ground_course_cd() / 10;
```
peterbarker, line 220:
```suggestion
        gps_state.ground_course_dd = gps.ground_course_cd() * 0.1;
```

sorry, its an efficiency thing; multiplication is much faster than division.

peterbarker (changes requested):

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
nexton-winjeel: Two things:
1. This is integer division. Adding a `float` conversion is *not* efficient.
2. If this level of micro-optimisation matters so much, shouldn't we just enable the `-freciprocal-math` compiler optimisation? (Compiler Explorer example: https://godbolt.org/z/5zn7ejnxK)

peterbarker: >     1. This is integer division. Adding a `float` conversion is _not_ efficient.

Yep, my bad.  Reviewing too much floating point stuff moving from integer to floating point recently.

>     2. If this level of micro-optimisation matters so much, shouldn't we just enable the `-freciprocal-math` compiler optimisation? (Compiler Explorer example: https://godbolt.org/z/5zn7ejnxK)

I know someone looked at that at some stage but can't remember the resolution.

catkira:
@rmackay9 can you backport this to plane 4.5.6 ?

PR #28091, opened 2024-09-13
Add test for geopose heading

Ryanf55:
# Purpose

Add a test to prove the reported heading on the geopose interface is REP-103 compliant.

This matches the proposed clarifications here, and can be used as an example in the REP.
https://github.com/ros-infrastructure/rep/pull/407

### DIFF
--- a/Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_geopose_msg_received.py
+++ b/Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_geopose_msg_received.py
@@ -13,7 +13,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-"""Bring up ArduPilot SITL and check the GeoPose message is being published."""
+"""
+Bring up ArduPilot SITL and check the GeoPoseStamped message is being published.
+
+colcon test --packages-select ardupilot_dds_tests \
+--event-handlers=console_cohesion+ --pytest-args -k test_geopose_msg_received
+
+"""
 
 import launch_pytest
 import math
@@ -41,7 +47,7 @@ CMAC_ABS_ALT = 584
 CMAC_HEADING = 353
 
 
-def ros_quat_to_heading(quat):
+def ros_quat_to_heading_deg(quat):
     # By default, scipy follows scalar-last order – (x, y, z, w)
     rot = R.from_quat([quat.x, quat.y, quat.z, quat.w])
     r, p, y = rot.as_euler(seq="xyz", degrees=True)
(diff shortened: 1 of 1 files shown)


Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_geopose_msg_received.py
```diff
@@ -41,7 +47,7 @@
 CMAC_HEADING = 353
 
 
-def ros_quat_to_heading(quat):
+def ros_quat_to_heading_deg(quat):
     # By default, scipy follows scalar-last order – (x, y, z, w)
```
srmainwaring, line 51: `scipy` is a large library to pull in just for quaternions and rotations. Have you thought about `transforms3d` (https://matthew-brett.github.io/transforms3d/)?

Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_geopose_msg_received.py
```diff
@@ -57,10 +63,28 @@ def validate_position_cmac(position):
...
 
 def validate_heading_cmac(orientation):
-    """Return true if the vehicle is facing the right way for CMAC."""
-    return math.isclose(ros_quat_to_heading(orientation), CMAC_HEADING)
+    """
+    Return true if the vehicle is facing the right way for CMAC.
+
+    Per ROS REP-103, the quaternion should report 0 when the vehicle faces east,
+    and 90 degrees when facing north.
+    Because CMAC is NNW, we expect ~97 degees.
```
srmainwaring, line 81: `./Tools/autotest/locations.txt` has this for CMAC. Heading is 353 => rotation = 97 as described.

```bash
CMAC=-35.363261,149.165230,584,353
```

srmainwaring (approved):
LGTM. Added notes with suggestions, but no blockers.

Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_geopose_msg_received.py
Ryanf55: Yea, I just had to add a tolerance because the data is through the EAHRS with noise added. Peter mentioned to set the EKF type to match the SIM so it's exact, but we don't depend on mavproxy or support params yet, so I left that out of scope and used a tolerance instead.

Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_geopose_msg_received.py
```diff
@@ -57,10 +63,28 @@ def validate_position_cmac(position):
         and math.isclose(position.altitude, CMAC_ABS_ALT, abs_tol=1.0)
     )
 
+def wrap_360(angle):
```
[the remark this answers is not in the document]
Ryanf55: I pulled this from pymavlink but added recursion so you can handle multi-wrap. If the initial heading was directly north, the comparison would get tricky, but this util is only used in this hard-coded cmac test.

Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_geopose_msg_received.py
Ryanf55: Yea, I just saw scipy was already used a bunch of places, including in the setup script, and thought the team would prefer a library they are familiar with.
https://github.com/ArduPilot/ardupilot/blob/72a013952c62cf645c8f668620ba52a41debe042/Tools/environment_install/install-prereqs-ubuntu.sh#L195

Your link looks like a much more focused library, but it's not available on `apt` yet. 
https://github.com/matthew-brett/transforms3d/issues/65

tridge (approved):

PR #28093, opened 2024-09-13
Reduce overhead when doing SPI transactions

bugobliterator:
* Tested on CubeOrangePlus and CubeRedPrimary with Invensensev3 driver

* I used debug trace to analyse timeline for read_fifo to check if we have any substantial overhead while doing SPI transactions. It turned out that calling spiStart and spiStop which in turn call dmaStreamAlloc and and dealloc to take substantial amount of CPU call almost 3-4us. Also we were calling Shared DMA mutex lock and unlock when the DMA is not shared between peripherals.

* Following changes made SPI transactions much less CPU intensive:
- * Do not reset SPI peripheral using spistop and spistart calls if SPI config hasn't changed.
- * Also make changes to SPI frequencies for CubeOrangePlus and CubeRedPrimary to match on the bus where possible (minor CPU gain through this, just first part was enough as most transactions are IMU anyways).
- * Do not call mutex lock unlock for unshared DMAs.

Results:

CubeRedPrimary:

Before:
SPI1          PRI=181 sp=0x20001AA8 STACK= 792/1464 LOAD=13.4%
SPI4          PRI=181 sp=0x200021C0 STACK= 872/1464 LOAD=13.4%
SPI2          PRI=181 sp=0x20005380 STACK= 896/1464 LOAD=13.3%

After:
SPI1          PRI=181 sp=0x20001B08 STACK= 792/1464 LOAD= 9.7%
SPI4          PRI=181 sp=0x20002220 STACK= 792/1464 LOAD=11.2%
SPI2          PRI=181 sp=0x200053E0 STACK= 840/1464 LOAD= 8.3%

CubeOrangePlus:

Before:
SPI4          PRI=181 sp=0x3002AF58 STACK= 668/1464 LOAD=25.5%
SPI1          PRI=181 sp=0x3002B668 STACK= 668/1464 LOAD=14.5%

After:
SPI4          PRI=181 sp=0x3002AF78 STACK= 668/1464 LOAD=19.2%
SPI1          PRI=181 sp=0x3002B688 STACK= 668/1464 LOAD= 9.5%

>10% total CPU time recovered for both CubeOrangePlus and CubeRedPrimary


libraries/AP_HAL_ChibiOS/shared_dma.cpp
```diff
@@ -87,6 +87,9 @@ void Shared_DMA::unregister()
 // lock one stream
 bool Shared_DMA::lock_stream(uint8_t stream_id)
 {
+    if (!is_shared(stream_id)) {
```
andyp1per, line 90: I'm not sure this is safe. I'm pretty sure the dshot code for instance uses the DMA lock to mediate access, so you would need to also check whether another thread has taken the lock.

libraries/AP_HAL_ChibiOS/hwdef/CubeRedPrimary/hwdef.dat
```diff
@@ -105,15 +105,15 @@ PE5    SPI4_MISO    SPI4
...
 SPIDEV ms5611_0         SPI1 DEVID2  BARO_0_CS          MODE3 20*MHZ 20*MHZ
 
-SPIDEV icm42688_1       SPI2 DEVID1  ICM42688_1_CS      MODE3  2*MHZ  8*MHZ
+SPIDEV icm42688_1       SPI2 DEVID1  ICM42688_1_CS      MODE3  8*MHZ  8*MHZ
 SPIDEV rm3100           SPI2 DEVID2  RM3100_CS          MODE3  1*MHZ  1*MHZ
 
 SPIDEV icm20649         SPI4 DEVID1  ICM_2_CS           MODE3  4*MHZ  8*MHZ
 # alternative to icm20649
-SPIDEV icm45686         SPI4 DEVID1  ICM_2_CS           MODE3  2*MHZ  8*MHZ
+SPIDEV icm45686         SPI4 DEVID1  ICM_2_CS           MODE3 20*MHZ 20*MHZ
```
andyp1per, line 116: I thought the slower rate was slower for a reason, i.e. in case of recovery?

libraries/AP_HAL_ChibiOS/SPIDevice.cpp
```diff
@@ -383,6 +392,25 @@ void SPIBus::start_peripheral(void)
...
+*/
+bool SPIBus::config_changed(void)
+{
+    if (
+#if defined(STM32H7)
+        last_spicfg.cfg1 == spicfg.cfg1 &&
+        last_spicfg.cfg2 == spicfg.cfg2
+#else
+        last_spicfg.cr1 == spicfg.cr1 &&
+        last_spicfg.cr2 == spicfg.cr2
```
andyp1per, line 406: I feel like you might need to check the DMA lock state here as well. If the lock is shared (and therefore the channel shared), the periperal may need to be shutdown to allow another to use the DMA channel.

bugobliterator: @andyp1per we only get to this line near the end of acquire bus which must have been successful.

libraries/AP_HAL_ChibiOS/shared_dma.cpp
bugobliterator: @andyp1per can you point to the code where this is done, I don't think Shared DMA is used that way in RCOutput, its only role is to allocate dma for use for dshot and bdshot, as far as I am aware dshot stuff happens entirely from the single thread. What is DMA lock used to mediate access between?

libraries/AP_HAL_ChibiOS/hwdef/CubeRedPrimary/hwdef.dat
bugobliterator: @andyp1per the reason is simply legacy, we have been using that frequency in older Sensors.

andyp1per: Yeah, but for instance we needed that to solve a problem on ICM42688 (changed it on Matek for instance), so not 100% convinced this is an older sensor problem.

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
```diff
@@ -540,44 +540,50 @@ void AP_InertialSensor_Invensensev3::read_fifo()
...
 
-    // adjust the periodic callback to be synchronous with the incoming data
-    // this means that we rarely run read_fifo() without updating the sensor data
-    dev->adjust_periodic_callback(periodic_handle, backend_period_us);
-
-    while (n_samples > 0) {
-        uint8_t n = MIN(n_samples, INV3_FIFO_BUFFER_LEN);
-        if (!block_read(reg_data, (uint8_t*)fifo_buffer, n * fifo_sample_size)) {
-            goto check_registers;
+    do {
```
andyp1per, line 544: Do you know how much difference this makes? I ask because there is a bug in the v3 driver to do with DMA usage which we should fix - https://github.com/ArduPilot/ardupilot/pull/27841/commits/a56660e634eb7b8efe8b31ed4a5eca3c1eb7a0d7 - that said this also has the cost of two SPI transactions, so I am wondering if there is a way of doing both?

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
```diff
@@ -540,44 +540,50 @@ void AP_InertialSensor_Invensensev3::read_fifo()
...
+        uint8_t n;
+        void *fifo_result = nullptr;
+        if (n_samples == 0) {
+            if (!block_read(reg_counth, (uint8_t*)fifo_buffer, 2 + fifo_sample_size)) {
```
andyp1per, line 548: This needs a comment. I *think* I understand what you are trying to achieve here, but its not at all clear.

andyp1per (changes requested):
Need a comment for the IMU changes

andyp1per:
I did a short flight with just the SPI changes - had no issues and this gives a nice bump of about 5% CPU gain, which with my Dynamic FIFO change (https://github.com/ArduPilot/ardupilot/pull/27841) is giving me about 15% CPU gain

tridge:
@bugobliterator I'd like to discuss this a bit more with you. There are risks to not restarting the SPI peripheral. If (due to eg. a voltage spike) the SPI peripheral gets into a bad state, then with a single sensor bus that state could persist forever and take down the vehicle. We switched to always re-init some time ago when we saw this sort of bad persistent state on I2C peripherals.
I'd really rather fix the real problem which is that we are doing transfers at such a high rate for InvensenseV3

libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp
tridge, line 548: should be using _dev->transfer() like in InvensenseV2 and V1, to save the stack space and copy, and preventing a watchdog

tridge (changes requested):

andyp1per (changes requested):
This needs to be split into SPI changes and IMU changes

PR #28097, opened 2024-09-13
AP_Airspeed: Don't fail on REG_WHOAMI_RECHECK_ID

landswellsong:
This prevents a false negative ID check for the cases when the AP is rebooted independently of the sensor, with the latter keeping a temporary ID that was set during initialization

### DIFF
--- a/libraries/AP_Airspeed/AP_Airspeed_ASP5033.cpp
+++ b/libraries/AP_Airspeed/AP_Airspeed_ASP5033.cpp
@@ -78,7 +78,7 @@ bool AP_Airspeed_ASP5033::confirm_sensor_id(void)
 {
     uint8_t part_id;
     if (!dev->read_registers(REG_PART_ID_SET, &part_id, 1) ||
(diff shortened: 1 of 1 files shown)


peterbarker:
Looks reasonable.

Whitespace is a bit off - we wouldn't usually format our code like that.

Please state what testing you've done on this.

landswellsong:
@peterbarker fixed the formatting though the line looks longuish to me now.

The main concern is that if I restart the ardupilot service, the sensor itself is never reset, which I think is also true for rebooting. It would keep the "testing" device id and thus the check would fail until power down. Alternatively we could reset it to `0x00` after the check, but my version just tries to assume the process might have ended in an unpredicted state. The downside is that we get a false positive for a sensor on those specific 2 IDs that does respond with `0x66` to the exact same register and doesn't complain about the id-set command, but I think the possibility is slim.

Tested on Linux HAL with Qiotek ASP5033 I²C sensors. Cold start naturally does work as it did before. Restarting the process no longer makes the ID check fail.

tridge (approved):

PR #28098, opened 2024-09-13
Add defines for experimental topics such as IMU to allow saving flash for unused topics

Ryanf55:
#  Purpose

* Experimental topics, such as IMU, should have an easy way to be disabled at compile time
* This demonstrates a pattern to add in new topics at compile time
* By updating to use the C++ topic table instead of the XML refs, AP now has full control at compile time over what's enabled in the interface
* Note - IMU is a special case - it's super high rate, and largely useless unless you are doing offboard fusion, which reportedly has not performed as well). We don't want to clog our network with unused data.

# Long term

* Every topic could get a define. It might make the code hard to read - are we actually that flash conscious if using DDS? 
image

I wouldn't think so because a lot of flash-heavy features can be moved offboard like path planning, terrain, avoidance, fences, advanced failsafe, AP_follow, AP_MIssion, Proximity, SmartRTL, Soaring, WPNav.
# Tests performed

Compile with IMU disabled and then check it's not published by running sim_vehicle.
```bash
./Tools/autotest/sim_vehicle.py -v ArduPlane --waf-configure-arg "--define=AP_DDS_IMU_PUB_ENABLED=0" --map --console --enable-dds
```

Terminal 2: Run the MicroROS agent 
```bash
ros2 run micro_ros_agent micro_ros_agent udp4 -p 2019
```

Terminal 3: Check topics exposed
```bash
$ ros2 topic list
/ap/battery/battery0
/ap/clock
/ap/cmd_gps_pose
/ap/cmd_vel
/ap/geopose/filtered
/ap/gps_global_origin/filtered
/ap/joy
/ap/navsat/navsat0
/ap/pose/filtered
/ap/tf
/ap/tf_static
/ap/time
/ap/twist/filtered
/parameter_events
/rosout
````


srmainwaring (approved):
LGTM.

srmainwaring:
@Ryanf55 - thanks for adding. I started a similar effort here: https://github.com/srmainwaring/ardupilot/tree/prs/pr-dds-sensor-sub when looking to add extra subscribers for external sensors, but hadn't factored out the change.

As a follow up to this PR there are a couple of outstanding items for the IMU topic.

- Change the topic name - I don't think `experimental` should be part of the name, it's either there or not.
- Fix the orientation so it's compliant with ROS conventions. This will break the existing use, but I think that is limited and we should correct it before it is more widely adopted.

tridge:
will be nice to be in build_options.py

tridge (approved):

PR #28099, opened 2024-09-14
AP_DroneCAN: DNAServer: optimize allocation strategy

tpwrules:
Please see code for details. There are sure to be a lot of variables but this change demonstrably makes the allocation message count (and therefore real allocation time, though I did not measure this precisely) lower and more consistent.

Number of allocation messages with 3x MatekL431-Periph nodes and Cube Orange (6 trials):

before: 44, 36, 35, 92, 107, 41

after: 28, 28, 28, 28, 26, 28


bugobliterator (approved):
LGTM

PR #28100, opened 2024-09-14
AP_DroneCAN: DNAServer: remove preferred allocation support

tpwrules:
Nothing is known to support it so it can't be tested. Removing it saves flash and reduces complexity.

I would prefer to add support for it to AP_Periph so it can be tested but tridge said he would prefer to just remove support. Saves 168 bytes of flash on Cube Orange. Would consume 32 bytes to add support to AP_Periph.

Tested on the bench that allocation of new and existing nodes works properly still.


bugobliterator (approved):
Agreed. Currently if user wants to force node ids, its best to statically assign on the device, and let the dnaserver register it statically. We should probably verify that our prearm works if it sees multiple devices with same node ids.

tpwrules:
I just checked that on the bench and it does still work with one ID manually assigned and the other automatically, or both manually. I think there are some quirks though and I plan to go over that functionality with a fine toothed comb soon.

PR #28103, opened 2024-09-14
Copter: Initialize stack variables

muramura:
Stack variables are not initialized.
If a value cannot be retrieved, it returns an invalid value.
Make it return 0.

### DIFF
--- a/ArduCopter/mode.cpp
+++ b/ArduCopter/mode.cpp
@@ -516,7 +516,7 @@ void Mode::get_pilot_desired_lean_angles(float &roll_out_cd, float &pitch_out_cd
 // transform pilot's roll or pitch input into a desired velocity
 Vector2f Mode::get_pilot_desired_velocity(float vel_max) const
 {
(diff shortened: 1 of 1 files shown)


ArduCopter/mode.cpp
```diff
@@ -516,7 +516,7 @@ void Mode::get_pilot_desired_lean_angles(float &roll_out_cd, float &pitch_out_cd
 // transform pilot's roll or pitch input into a desired velocity
 Vector2f Mode::get_pilot_desired_velocity(float vel_max) const
 {
-    Vector2f vel;
+    Vector2f vel{0.0f, 0.0f};
```
peterbarker, line 519: These variables are implicitly zero.

https://github.com/ardupilot/ardupilot/blob/master/libraries/AP_Math/vector2.h#L51

Simiarly for Location.

It's just a bit of Magic you have to know about.

Closing - thanks anyway!

PR #28107, opened 2024-09-14
Sub: small guided mode fixes

clydemcqueen:
This PR slightly improves sub guided mode:
* Test the destination against the fence before changing the submode. With this change, the submode is not changed if we return false. This matches the behavior in copter.
* Initialize `sub.yaw_rate_only` in all `guided_set_*` calls, so that any submode change starts from the same default
* Fix a few comments


peterbarker:
Commit messages need fixing.

clydemcqueen:
> Commit messages need fixing.

Whoops, fixed

Williangalvani (approved):
lgtm, thanks!

PR #28110, opened 2024-09-15
Copter: allow adding custom modes from scripting

IamPete1:
This allows a script to register a new mode with a given number and name. This is just a wrapper on guided mode, but it allows clearer interaction with the user. 

The example re-implements flip mode in a script. It uses mode 100, in MAVProxy you can enter with `mode 100`. 

In order to allow the flip I added a new guided binding `set_target_rate_and_throttle`. I also found and fixed a issue that the Z controller would throw a flow of control error when moving from guided angle with throttle to guided angle with climb rate. 

The internal variables of guided mode are changed to static so they apply to to the new modes too. 

There are some more changes needed:

- anywhere that doing a compare on guided mode by reference or number needs to deal with the new modes.

- Come up with a way to communicate to the GCS which modes are available so it could populate its mode list with the new modes (this would also be useful if we want to start removing modes and/or for the custom build server).

- Currently scripting is the only way to register a mode, but there is no reason a companion computer could not do the same with a new MAVLink message. 

- In the future we may want to add more state on the custom mode object, for example a "allow_enter" bool which the script could keep updated. This would allow the script to reject entry if the correct conditions are not met. Currently the flip example will enter but then immediately put the mode back again if it does not like it.
- Allow a custom version of guided no-gps as well as full guided.
- Currently the mode will fail to add if the script is restated because it is already there, it might be nice to allow a re-add of the same mode, although this is only a issue for script dev.


amilcarlucas:
```
arducopter       1092 (+0.0604%)  0 (0.0000%)  -4 (-0.0015%)  1092 (+0.0603%)                                  153864
antennatracker   356 (+0.0266%)   0 (0.0000%)  4 (+0.0015%)   356 (+0.0266%)                                   624868
arducopter-heli  1124 (+0.0622%)  0 (0.0000%)  4 (+0.0015%)   1124 (+0.0621%)                                  154832
ardusub          356 (+0.0223%)   0 (0.0000%)  -4 (-0.0015%)  356 (+0.0222%)                                   364140
ardurover        356 (+0.0215%)   0 (0.0000%)  -4 (-0.0015%)  356 (+0.0214%)                                   305848
blimp            356 (+0.0262%)   0 (0.0000%)  -4 (-0.0015%)  356 (+0.0261%)                                   602752
arduplane        356 (+0.0197%)   0 (0.0000%)  -4 (-0.0015%)  356 (+0.0197%)                                   157544
```

ArduCopter/Copter.cpp
```diff
@@ -389,7 +389,59 @@ bool Copter::set_target_angle_and_climbrate(float roll_deg, float pitch_deg, flo
     mode_guided.set_angle(q, Vector3f{}, climb_rate_ms*100, false);
     return true;
 }
-#endif
+
+bool Copter::set_target_rate_and_throttle(float roll_rate_dps, float pitch_rate_dps, float yaw_rate_dps, float throttle)
```
rmackay9, line 393: let's put a comment above this method.  It's pretty obvious what it does but all other methods have comments so let's keep it consistent

ArduCopter/mode.h
```diff
@@ -1164,14 +1194,31 @@ class ModeGuided : public Mode {
     void set_yaw_state(bool use_yaw, float yaw_cd, bool use_yaw_rate, float yaw_rate_cds, bool relative_angle);
 
     // controls which controller is run (pos or vel):
-    SubMode guided_mode = SubMode::TakeOff;
-    bool send_notification;     // used to send one time notification to ground station
-    bool takeoff_complete;      // true once takeoff has completed (used to trigger retracting of landing gear)
+    static SubMode guided_mode;
```
tridge, line 1197: could these cause problems? it makes it shared between all modes

rmackay9:
This will be a great feature!

IamPete1:
Investigate `AVAILABLE_MODES` MAVLink message: https://github.com/mavlink/mavlink/blob/7b35ec95a1995ba7253f1af47ecb48abf93d8433/message_definitions/v1.0/development.xml#L666

ArduCopter/Copter.cpp
```diff
@@ -389,7 +389,59 @@ bool Copter::set_target_angle_and_climbrate(float roll_deg, float pitch_deg, flo
...
+    // Convert from degrees per second to radians per second
+    Vector3f ang_vel_body { roll_rate_dps, pitch_rate_dps, yaw_rate_dps };
+    ang_vel_body *= DEG_TO_RAD;
+
+    // Pass to guided mode
+    mode_guided.set_angle(q, ang_vel_body, throttle, true);
+    return true;
+}
+
+// Register a custom mode with given number and names
```
andyp1per, line 413: #if AP_SCRIPTING_ENABLED somewhere here?

ArduCopter/Copter.cpp
```diff
@@ -389,7 +389,59 @@ bool Copter::set_target_angle_and_climbrate(float roll_deg, float pitch_deg, flo
...
+bool Copter::register_custom_mode(const uint8_t num, const char* full_name, const char* short_name)
+{
+    // Numbers over 100 are reserved for custom modes
+    if (num < 100) {
```
andyp1per, line 417: #define somewhere for this magic number

ArduCopter/mode.h
andyp1per: Feels like a bear trap!

ArduCopter/mode.h
```diff
@@ -1164,14 +1194,31 @@ class ModeGuided : public Mode {
...
+    static bool send_notification;     // used to send one time notification to ground station
+    static bool takeoff_complete;      // true once takeoff has completed (used to trigger retracting of landing gear)
 
     // guided mode is paused or not
-    bool _paused;
+    static bool _paused;
 };
 
+// Mode which behaves as guided with custom mode number and name
```
andyp1per, line 1205: #if AP_SCRIPTING_ENABLED somewhere?

ArduCopter/Copter.cpp
IamPete1: This section is of the file is in a big `#if AP_SCRIPTING_ENABLED`.

ArduCopter/Copter.cpp
IamPete1: I don't like defining stuff that is only used once like this. Two reasons, firstly it makes the code harder to read because you have to go and find the value. Secondly the define has a much larger scope.

In any case i'm still not 100% on the numbers over 100 thing. I was just thinking to remove future conflicts but it also prevents moving existing out of tree modes into lua with the same number.

ArduCopter/mode.h
IamPete1: Done.

IamPete1:
I have rebased, now lots of the supporting stuff has been merged in the other PRs this is a much smaller change.

ArduCopter/Copter.cpp
IamPete1: Done in https://github.com/ArduPilot/ardupilot/pull/28134

ArduCopter/mode.h
IamPete1: The problem the vast majority of the guided state is static already. This stuff: https://github.com/ArduPilot/ardupilot/blob/25e77a5d9fe360e178c86b386a4d2def90976dc4/ArduCopter/mode_guided.cpp#L9-L33

It either has to be all static or all not. With half and half as it is now you can endup being in the wrong sub mode for the targets you have. Making these static is a smaller change than making the existing stuff not.

Currently there should be no issue because there is nothing in the mode exit function so there is no overlap between one version of guided and the next. So long as everything is reset correctly in the mode init then we start clean (which I think it is already).

As you say, this is a bit nasty for the future, but anything else would require a much larger re-work.

IamPete1:
This now uses the `AVAILABLE_MODES` and `AVAILABLE_MODES_MONITOR` to notify the GCS of the new modes. There is a "allow enter" state which the script can change.

ArduCopter/Copter.cpp
IamPete1: Magic number is gone.

timtuxworth:
I just noticed this PR is prefixed "Copter:" - is it going to work in Plane?

IamPete1:
> I just noticed this PR is prefixed "Copter:" - is it going to work in Plane?

Not in this pr, but eventually yes

timtuxworth:
> > I just noticed this PR is prefixed "Copter:" - is it going to work in Plane?
> 
> Not in this pr, but eventually yes

Looking forward to to it. As soon as it's available I'll use it to implement "FOLLOW" mode in Plane :-)

libraries/AP_Scripting/examples/Flip_Mode.lua
```diff
@@ -0,0 +1,246 @@
...
+-- Used to check for mode changes
+local last_mode_number
+
+local MODES = {
+    STABILIZE = 0,
+    ALT_HOLD = 2,
+}
+
+-- Check if we should be allowed to enter flip mode
+local function allow_enter(previous_mode)
```
IamPete1, line 60: This should be rolled into the new allow enter bool in the state object.

rmackay9 (approved):
awesome work!

tridge (commented):
it would be nice to add an auto-test using the flip2 mode

PR #28111, opened 2024-09-15
AP_ExternalAHRS: Correct the judgment

muramura:
The ILABS_UNIT_STATUS2_GNSS_FUSION_OFF flag does not exist in UNITSTATUS.
I think the decision should be made based on the GNSS-related flags in UNITSTATUS and UNITSTATUS2.


peterbarker:
Ping @valentinb - looks like it could be a problem?

valbv:
Thanks guys!
Yes, this fix is correct

peterbarker:
Merged, thanks!

PR #28114, opened 2024-09-16
AP_Logger: log the Scheduler tasks list at log startup

peterbarker:


peterbarker:
We can't log `tasks.txt` like this.  Once we have requested the file once we start to gather stats, which is a little bit of overhead we may not want

PR #22709, opened 2023-01-21
AP_Airspeed: Add support for new message

IamPete1:
This adds support for the new airspeed message. It does not remove the old named value float, maybe we should have them both until we get GCS support.

We need a similar message for baro and then we can remove the three scaled pressure messages.

PR #28121, opened 2024-09-17
GCS_MAVLink: Fix Airspeed without AHRS

IamPete1:
Fixes builds that includes MAVLink and Airspeed without AHRS, for example periph heavy. This bug was introduced in https://github.com/ArduPilot/ardupilot/pull/22709.


peterbarker (approved):
LGTM

PR #28123, opened 2024-09-17
disable feature due to flash space constraints

peterbarker:
ArduPilot doesn't have room for every feature it can have, and hasn't for some time.


Ryanf55 (approved):
Can we update the wiki entries to tell users the feature is off by default since it was enabled by default earlier?
* https://ardupilot.org/copter/docs/common-external-ahrs.html
* https://ardupilot.org/dev/docs/ros-cartographer-slam.html#configure-ardupilot

I agree on disabling some of these little used features by default.

Hwurzburg:
> Can we update the wiki entries to tell users the feature is off by default since it was enabled by default earlier?
> 
>     * https://ardupilot.org/copter/docs/common-external-ahrs.html
> 
>     * https://ardupilot.org/dev/docs/ros-cartographer-slam.html#configure-ardupilot
> 
> 
> I agree on disabling some of these little used features by default.

1. should be in the release notes
2. features are constantly being removed from boards...the wiki has a features section....supposed to auto-update but has bugs/broken for some time so wiki currently directs users to examine the features.txt files..https://ardupilot.org/plane/docs/common-limited-firmware.html

andyp1per (approved):

peterbarker:
I've adjusted this so that these features remain enabled in SITL.

@Ryanf55 there seems to be an undeclared dependency between AP_DDS's extenal odometry support and the visual odometry library; compiling without odometry seems to cause DDS to fail.

Ryanf55:
> I've adjusted this so that these features remain enabled in SITL.
> 
> @Ryanf55 there seems to be an undeclared dependency between AP_DDS's extenal odometry support and the visual odometry library; compiling without odometry seems to cause DDS to fail.

This config was supposed to make that automatic. https://github.com/ArduPilot/ardupilot/blob/86f51ee5a4d12626881167b76152917eaa4d395a/libraries/AP_DDS/AP_DDS_config.h#L17
Thanks for reporting, I'll take a look!

Ryanf55:
> I've adjusted this so that these features remain enabled in SITL.
> 
> @Ryanf55 there seems to be an undeclared dependency between AP_DDS's extenal odometry support and the visual odometry library; compiling without odometry seems to cause DDS to fail.

@peterbarker Can you let me know how you reproduce this failure? I tried this and it compiles fine.
```bash
./waf configure --enable-dds --disable-VISUALODOM
./waf plane
```

peterbarker:
> > I've adjusted this so that these features remain enabled in SITL.
> > @Ryanf55 there seems to be an undeclared dependency between AP_DDS's extenal odometry support and the visual odometry library; compiling without odometry seems to cause DDS to fail.
> 
> @peterbarker Can you let me know how you reproduce this failure? I tried this and it compiles fine.
> 
> ```shell
> ./waf configure --enable-dds --disable-VISUALODOM
> ./waf plane
> ```

```
[1402/1524] Compiling libraries/AP_GPS/tests/test_gps.cpp
In file included from ../../modules/gtest/googletest/include/gtest/gtest.h:62,
                 from ../../tests/AP_gtest.h:7,
                 from ../../libraries/AP_DDS/tests/test_ap_dds_external_odom.cpp:1:
../../libraries/AP_DDS/tests/test_ap_dds_external_odom.cpp: In member function ‘virtual void AP_DDS_EXTERNAL_ODOM_test_is_odometry_success_Test::TestBody()’:
../../libraries/AP_DDS/tests/test_ap_dds_external_odom.cpp:15:17: error: ‘AP_DDS_External_Odom’ was not declared in this scope
   15 |     ASSERT_TRUE(AP_DDS_External_Odom::is_odometry_frame(msg));
```

From https://github.com/peterbarker/ardupilot/actions/runs/10912857981/job/30288286476

peterbarker:
We chickened out on the AP_ExternalAHRS thing.  At this point

rmackay9:
Visual Odometry is quite commonly used.  Does this remove it for nearly all boards?

peterbarker:
> Visual Odometry is quite commonly used. Does this remove it for nearly all boards?

Yes - all ChibiOS boards.

rmackay9:
Hi @peterbarker,

OK, thanks.  I think taking visual odometry out but leaving External AHRS is an odd decision if it is based upon the number of users.  I think we should look for other areas to save before removing visual odometry

rmackay9 (changes requested):
I don't think we should disable visual odometry from the base builds at this point.  I think we should look for other areas first with lower impact on users

peterbarker:
> I don't think we should disable visual odometry from the base builds at this point. I think we should look for other areas first with lower impact on users

I am absolutely not wedded to removing anything here - but I think the maths does say that we need to do these things.

I did look through the libraries (briefly!) and put forward a set.  `AP_EFI` and `AP_ICEngine` were knocked back.  That left us with these ones :-)

IOW - suggestions welcome as to what to remove.

A smaller set of features in 4.6 and then removing more for 4.7 is distinctly possible.  We could potentially 
even skip this until 4.7

Ryanf55:
> AP_ExternalAHRS

I'd vote for EAHRS as optional because only a few advanced users are using it anyways and they need to read the wiki to set it up. 
Have you run the numbers on how much flash that would buy us vs visual odom?

tridge (approved):

peterbarker:
See Randy's issue here: https://github.com/ArduPilot/ardupilot/issues/28210

peterbarker:
Closing this one as there was only one candidate library in it, beacon.  I've opened a fresh PR for that here: https://github.com/ArduPilot/ardupilot/pull/30299

PR #28128, opened 2024-09-17
CI: build CubeRed-EKF2 not CubeOrange-EKF2 in CI

peterbarker:
we really just care that EKF2 builds here, the board really isn't that important

### DIFF
--- a/.github/workflows/test_chibios.yml
+++ b/.github/workflows/test_chibios.yml
@@ -156,7 +156,7 @@ jobs:
             build-options-defaults-test,
             signing,
             CubeOrange-PPP,
-            CubeOrange-EKF2,
+            CubeRed-EKF2,
             SOHW,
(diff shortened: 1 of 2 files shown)


Ryanf55 (approved):

.github/workflows/test_chibios.yml
```diff
@@ -156,7 +156,7 @@ jobs:
             build-options-defaults-test,
             signing,
             CubeOrange-PPP,
-            CubeOrange-EKF2,
+            CubeBlack-EKF2,
```
tridge, line 159: let's make this CubeRed

tridge (changes requested):

.github/workflows/test_chibios.yml
peterbarker: Done

PR #28129, opened 2024-09-17
Copter: guided: `set_angle` call `init_z_controller` when changing from thrust to climb rate control to avoid flow of control error

IamPete1:
We init the Z controller when we fist enter the angle sub mode of guided. However if you send throttle commands the Z controller is not used, if you then change to sending climb rate commands while remaining in angle sub mode you trigger this flow of control error in the pos controller:

https://github.com/ArduPilot/ardupilot/blob/a9ea760cadfddf39434cb3b7dbcf8533f6c4fb1a/libraries/AC_AttitudeControl/AC_PosControl.cpp#L961-L969

extract from https://github.com/ArduPilot/ardupilot/pull/28110

### DIFF
--- a/ArduCopter/mode_guided.cpp
+++ b/ArduCopter/mode_guided.cpp
@@ -652,6 +652,9 @@ void ModeGuided::set_angle(const Quaternion &attitude_quat, const Vector3f &ang_
     // check we are in velocity control mode
     if (guided_mode != SubMode::Angle) {
         angle_control_start();
+    } else if (!use_thrust && guided_angle_state.use_thrust) {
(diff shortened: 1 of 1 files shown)


rmackay9:
I think this is very likely correct

lthall (approved):
Nice Catch

rmackay9 (approved):
Thanks for this.  The extra spaces above and below could be removed but in general it's OK to merge, thanks!

IamPete1:
> Thanks for this. The extra spaces above and below could be removed but in general it's OK to merge, thanks!

Spaces removed.

PR #28131, opened 2024-09-17
Copter: GCS_MAVLink: populate flags using position controller active flags

IamPete1:
This means we don't need to switch on the mode number, extracted from https://github.com/ArduPilot/ardupilot/pull/28110 

This is the first time we will be using these flags for reporting, until now it is used internally only. There are some changes in behaviour, for example you can be landed in auto mode in which case the pos controller may not report as active (i need to check).

### DIFF
--- a/ArduCopter/GCS_Mavlink.cpp
+++ b/ArduCopter/GCS_Mavlink.cpp
@@ -37,25 +37,11 @@ MAV_MODE GCS_MAVLINK_Copter::base_mode() const
     // only get useful information from the custom_mode, which maps to
     // the APM flight mode and has a well defined meaning in the
     // ArduPlane documentation
-    switch (copter.flightmode->mode_number()) {
-    case Mode::Number::AUTO:
-    case Mode::Number::AUTO_RTL:
-    case Mode::Number::RTL:
-    case Mode::Number::LOITER:
-    case Mode::Number::AVOID_ADSB:
-    case Mode::Number::FOLLOW:
-    case Mode::Number::GUIDED:
-    case Mode::Number::CIRCLE:
-    case Mode::Number::POSHOLD:
-    case Mode::Number::BRAKE:
-    case Mode::Number::SMART_RTL:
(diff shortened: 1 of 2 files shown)


ArduCopter/GCS_Mavlink.cpp
```diff
@@ -37,25 +37,11 @@ MAV_MODE GCS_MAVLINK_Copter::base_mode() const
...
-    case Mode::Number::LOITER:
-    case Mode::Number::AVOID_ADSB:
-    case Mode::Number::FOLLOW:
-    case Mode::Number::GUIDED:
-    case Mode::Number::CIRCLE:
-    case Mode::Number::POSHOLD:
-    case Mode::Number::BRAKE:
-    case Mode::Number::SMART_RTL:
+    if ((copter.pos_control != nullptr) && copter.pos_control->is_active_xy()) {
         _base_mode |= MAV_MODE_FLAG_GUIDED_ENABLED;
```
peterbarker, line 41: This makes our answer here marginally worse.

My reading of that attribute is we should only set it actually chasing a waypoints "flies waypoints / mission items".

Not a blocker, but we really should cut this down to AUTO, AUTO_RTL, RTL and SMART_RTL at some stage.

peterbarker (commented):
I like this change.

IamPete1:
I have rebased, I tested the auto mode case when disarmed or armed but waiting for high throttle in both cases both XY and Z control are active.

However, I can't guarantee that this is the same in very combination of mode and submode. If we wanted identical behavior then this could be a per mode function call.

rmackay9 (approved):
This is probably OK but we decided that a bit of testing with MP and QGC is warranted.  At least a grep through MP/QGC to see if they use the flag at all

robertlong13:
> This is probably OK but we decided that a bit of testing with MP and QGC is warranted. At least a grep through MP/QGC to see if they use the flag at all

MP doesn't look at them.
image

IamPete1:
I have grepped QGC, there are no users of these bits there either.

PR #28134, opened 2024-09-17
Scripting: add `set_target_rate_and_throttle` binding for copter

IamPete1:
This adds a rate control binding with throttle for scripting, extracted from https://github.com/ArduPilot/ardupilot/pull/28110.

Note that we do already have `set_target_throttle_rate_rpy` which is impended on plane.

https://github.com/ArduPilot/ardupilot/blob/a9ea760cadfddf39434cb3b7dbcf8533f6c4fb1a/libraries/AP_Scripting/docs/docs.lua#L2487-L2492

The core idea is the same. However, the existing method does not return false if the current mode is not valid. We could add a return to the exiting plane method with out a breaking change. However it is nice to keep the forward flight bindings separate to the VTOL ones so in the future we could support this method on quadplane as well as the existing one for forward flight.

### DIFF
--- a/libraries/AP_Scripting/generator/description/bindings.desc
+++ b/libraries/AP_Scripting/generator/description/bindings.desc
@@ -329,6 +329,7 @@ singleton AP_Vehicle method set_target_posvelaccel_NED boolean Vector3f Vector3f
 singleton AP_Vehicle method set_target_velaccel_NED boolean Vector3f Vector3f boolean float -360 +360 boolean float'skip_check boolean
 singleton AP_Vehicle method set_target_velocity_NED boolean Vector3f
 singleton AP_Vehicle method set_target_angle_and_climbrate boolean float -180 180 float -90 90 float -360 360 float'skip_check boolean float'skip_check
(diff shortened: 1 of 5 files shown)


lthall (approved):
Looks fine to me.

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -351,6 +351,8 @@ singleton AP_Vehicle method reboot void boolean
 singleton AP_Vehicle method is_landing boolean
 singleton AP_Vehicle method is_taking_off boolean
 singleton AP_Vehicle method set_crosstrack_start boolean Location
+singleton AP_Vehicle method set_target_rate_and_throttle boolean float'skip_check float'skip_check float'skip_check float'skip_check
```
rmackay9, line 354: I'd move this higher in the .desc to be grouped with the others like set_target_angle_and_climbrate

rmackay9 (approved):
LGTM, thanks!

@IamPete1 I've made a small suggestion but it's not a blocker so feel free to merge if you're happy.

libraries/AP_Scripting/generator/description/bindings.desc
IamPete1: Done

PR #28135, opened 2024-09-17
AntennaTracker: refuse to track (0;0)

landswellsong:
I'm playing with Olimex g-Link GPS/LTE (Qualcomm EC25) unit for my setup and I've noticed that it seem to be throwing zero coordinates at some stage during the module startup. I can't seem to pinpoint the exact source of the message in my telemetry log, but the dataflash log from the tracker pretty convincingly shows that there were a couple of zero datapoints incoming so it's definitely that can somehow happen in the wild:
VPOS Lat

Consequently, the tracker wanted to execute a sharp turn since its target jumped about 130° and since it's in the middle of the ocean very far away from me, the minimum tracking distance condition did not help either.

This change prevents such behavior from ever taking place.

I wanted to add this to my [bigger WIP PR on the tracker](https://github.com/ArduPilot/ardupilot/pull/27676) but it's probably a benign change that can be applied as is. Will fly-test on the weekend if there are doubts that it might break anything.

I also changed an outdated comment since both the code, the log and the documentation seem to agree that the vehicle position is logged as `VPOS` and not as `GPS2`.

### DIFF
--- a/AntennaTracker/tracking.cpp
+++ b/AntennaTracker/tracking.cpp
@@ -131,6 +131,11 @@ void Tracker::update_tracking(void)
  */
 void Tracker::tracking_update_position(const mavlink_global_position_int_t &msg)
 {
+    // reject (0;0) coordinates
+    if (!msg.lat && !msg.lon) {
+        return;
+    }
+
     vehicle.location.lat = msg.lat;
     vehicle.location.lng = msg.lon;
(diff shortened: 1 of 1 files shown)


tpwrules:
Why does this check the current position as opposed to the incoming position?

peterbarker (approved):
LGTM

amilcarlucas (approved):

PR #28140, opened 2024-09-18
AP_Scripting: add bindings get/set full mode for GPIO

bugobliterator:
allows setting alternate pin modes pin drive strengths etc.


peterbarker:
Not much of an abstraction at this point - when your API is essentially register-writes....

tridge (approved):

PR #26886, opened 2024-04-25
added driver for MLX90614 Temperature sensor

rahulOCR:
Hey everyone,
I am resubmitting this PR as per code of conduct of Ardupilot after closing #26644
I am submitting this PR to merge into master. I have added MLX90614 temperature sensor into AP_TemperatureSensor library.
Using data-sheet provided [Link of data-sheet]](https://www.melexis.com/en/documents/documentation/datasheets/datasheet-mlx90614). I have used this to get ambient temperature of cabin of battery.

**WIKI:**

Hi I am attaching datasheet, wiki and photos for interfacing. MLX90614 sensor is an infrared thermometer for non-contact temperature measurements. Both the IR sensitive thermopile detector chip and the signal conditioning ASIC are integrated in the same TO-39 can. Integrated into the MLX90614 are a low noise amplifier, 17-bit ADC and powerful DSP unit thus achieving high accuracy and resolution of the thermometer.

set (examples shown for first sensor):

[TEMP1_TYPE](https://ardupilot.org/copter/docs/parameters.html#temp1-type) = 0:Disabled, 1:TSYS01, 2:MCP9600, 3:MAX31865, 4:TSYS03, 5:Analog, 6:DroneCAN, 7:MLX90614 Sensor type

[TEMP1_BUS](https://ardupilot.org/copter/docs/parameters.html#temp1-bus) = 0-3 Board’s I2C port on which the sensor is attached

[TEMP1_ADDR](https://ardupilot.org/copter/docs/parameters.html#temp1-addr) = the sensor’s I2C address (0-127). For this sensor, 90

[TEMP1_SRC](https://ardupilot.org/copter/docs/parameters.html#temp1-src) = which other temperature reporting device type should have its temperature report replaced with this sensor’s.

[TEMP1_SRC_ID](https://ardupilot.org/copter/docs/parameters.html#temp1-src-id) = this identifies which individual device, of the above type, should have its report replaced.

318392511-4a5c6cde-a86b-47e0-a5c3-8bec4d2a5e01

[MLX90614-Datasheet-Melexis.pdf](https://github.com/ArduPilot/ardupilot/files/14823016/MLX90614-Datasheet-Melexis.pdf)

PR #28143, opened 2024-09-18
Add LUA AP_TemperatureSensor bindings

peterbarker:
... and an example

image

Testing https://github.com/ArduPilot/ardupilot/pull/26886


peterbarker:
@rahulOCR Something odd going on here.  `get_temperature` should be returning degrees-celsius, so the conversion I'm putting in here is incorrect.

Any ideas?

rahulOCR:
`get_temperature` will return in kelvin. can you let me know what odd is going on ?

libraries/AP_Scripting/examples/temperature_sensor.lua
```diff
@@ -0,0 +1,19 @@
...
+   simple example of reading a temperature sensor
+--]]
+
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+
+--[[
+   main update function, called at 1Hz
+--]]
+function update()
+   local temperature_K = temperature_sensor:get_temperature(0)
```
rahulOCR, line 11: this returns in degrees calvin

peterbarker:
Rebased on top of @rahulOCR 's fix (thanks!).

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -973,3 +973,9 @@ singleton AP_VisualOdom depends HAL_VISUALODOM_ENABLED
 singleton AP_VisualOdom rename visual_odom
 singleton AP_VisualOdom method healthy boolean
 singleton AP_VisualOdom method quality int8_t
+
+-- ----AP_TemperatureSensor Library----
+include AP_TemperatureSensor/AP_TemperatureSensor.h
+singleton AP_TemperatureSensor depends AP_TEMPERATURE_SENSOR_ENABLED
+singleton AP_TemperatureSensor rename temperature_sensor
+singleton AP_TemperatureSensor method get_temperature boolean float'Null uint8_t 0 ud->num_instances()
```
IamPete1, line 981: We might need to add some checks in `AP_TemperatureSensor` but we have been trying to not limit instances by `ud->num_instances()` like this. The problem is that a script has no way to know `ud->num_instances()` at runtime. A too large value will cause the script to error out (unless you pcall).

IamPete1: Looks like its safe to call with any value. 
```suggestion
singleton AP_TemperatureSensor method get_temperature boolean float'Null uint8_t'skip_check
```

peterbarker: Done and re-tested, thanks!

IamPete1 (approved):
LGTM

peterbarker:
Thanks @IamPete1 , merging.

Issue #28101, opened 2024-09-14
Can't set 100Hz for MAVLINK_MSG_ID_HIGHRES_IMU on Rover


lida2003:
**Setup:**

- latest 4.6-dev
- 921600 (UAVT, MAVLinkv2)

same code from setting MAV_CMD_SET_MESSAGE_INTERVAL,MAVLINK_MSG_ID_HIGHRES_IMU to 100Hz, which is 10000 us.

**Result:**
- F435 Copter 100Hz OK
-  H743 Copter 100Hz/200Hz OK
- H743 Rover NG about 50Hz

PS: Previous discussion here[Can't set 100Hz for MAVLINK_MSG_ID_HIGHRES_IMU?
#4](https://github.com/chobitsfan/mavlink-udp-proxy/issues/4).

stephendade:
Just a guess, but the loop rate in Rover ``SCHED_LOOP_RATE`` is 50 by default. Trying increasing this to 100 or 200.

lida2003:
I successfully set 10Hz(actually 9Hz), 25Hz(actually 24Hz), 50Hz(actually 49~50Hz), but I can't set 40Hz, 100Hz, 200Hz.

But it works for copter code 4.6-dev(same git commit code). It might be copter handles this command different from rover. I'm NOT digging into the code right now. 

Just post this question here, maybe someone can take care of it when have time.

stephendade:
Ok, I've done some digging and created a patch in #28144.

I've confirmed by initial guess that Ardupilot is indeed capping the message rate at ``SCHED_LOOP_RATE``. Try setting this parameter to 250Hz and try your message rates again.

lida2003:
I'll do the test tomorrow and get back to you soon.

stephendade:
> Is there any api for changing "SCHEDULER_LOOP_RATE"?

That's just an Ardupilot parameter that you can set via your GCS.

lida2003:
Thanks.
![图片](https://github.com/user-attachments/assets/b2ac5b33-270e-4f61-b803-f1f41d6dc841)

PR #28144, opened 2024-09-18
GCS_MAVLink: Check if set_message_interval is too fast

stephendade:
From #28101.

Ardupilot caps the maximum message rate at 80% of ``SCHED_LOOP_RATE``.

If a user requests a rate faster than this via ``MAV_CMD_SET_MESSAGE_INTERVAL``, the command now fails and advises the user to increase ``SCHED_LOOP_RATE``.

This will prevent user confusion as per the linked issue.

Tested in SITL.

### DIFF
--- a/Tools/autotest/vehicle_test_suite.py
+++ b/Tools/autotest/vehicle_test_suite.py
@@ -6413,8 +6413,8 @@ class TestSuite(ABC):
                 self.remove_message_hook(hook)
         for script in dead.installed_scripts:
             self.remove_installed_script(script)
-        for (message_id, interval_us) in dead.overridden_message_rates.items():
-            self.set_message_interval(message_id, interval_us)
+        for (message_id, rate_hz) in dead.overridden_message_rates.items():
+            self.set_message_rate_hz(message_id, rate_hz)
         for module in dead.installed_modules:
             print("Removing module (%s)" % module)
             self.remove_installed_modules(module)
--- a/Tools/autotest/vehicle_test_suite.py
+++ b/Tools/autotest/vehicle_test_suite.py
@@ -10866,6 +10866,14 @@ Also, ignores heartbeats not from our target system'''
         if abs(original_rate*2 - round(self.get_message_rate_hz("VFR_HUD", run_cmd=self.run_cmd_int))) > 1:
             raise NotAchievedException("Did not set rate")
 
+        # Try setting a rate well beyond SCHED_LOOP_RATE
+        self.run_cmd(
+            mavutil.mavlink.MAV_CMD_SET_MESSAGE_INTERVAL,
+            p1=mavutil.mavlink.MAVLINK_MSG_ID_VFR_HUD,
+            p2=self.rate_to_interval_us(800),
+            want_result=mavutil.mavlink.MAV_RESULT_DENIED,
+        )
+
         self.start_subtest("Use REQUEST_MESSAGE via COMMAND_INT")
         # 148 is AUTOPILOT_VERSION:
         self.context_collect('AUTOPILOT_VERSION')
--- a/libraries/GCS_MAVLink/GCS.h
+++ b/libraries/GCS_MAVLink/GCS.h
@@ -267,6 +267,9 @@ public:
     // accessor for uart
     AP_HAL::UARTDriver *get_uart() { return _port; }
 
+    // cap the MAVLink message rate. It can't be greater than 0.8 * SCHED_LOOP_RATE
+    uint16_t cap_message_interval(uint16_t interval_ms) const;
+
     virtual uint8_t sysid_my_gcs() const = 0;
     virtual bool sysid_enforce() const { return false; }
 
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -225,6 +225,20 @@ void GCS_MAVLINK::send_power_status(void)
                                   hal.analogin->power_status_flags());
 }
 
+#if AP_SCHEDULER_ENABLED
+// cap the MAVLink message rate. It can't be greater than 0.8 * SCHED_LOOP_RATE
+uint16_t GCS_MAVLINK::cap_message_interval(uint16_t interval_ms) const
+{
+    if (interval_ms == 0) {
+        return 0;
+    }
+    if (interval_ms*800 < AP::scheduler().get_loop_period_us()) {
+        return AP::scheduler().get_loop_period_us()/800.0f;
+    }
+    return interval_ms;
+}
+#endif
+
 #if HAL_WITH_MCU_MONITORING
 // report MCU voltage/temperature status
 void GCS_MAVLINK::send_mcu_status(void)
@@ -1657,11 +1671,7 @@ bool GCS_MAVLINK::set_ap_message_interval(enum ap_message id, uint16_t interval_
     }
 
 #if AP_SCHEDULER_ENABLED
-    // send messages out at most 80% of main loop rate
-    if (interval_ms != 0 &&
-        interval_ms*800 < AP::scheduler().get_loop_period_us()) {
-        interval_ms = AP::scheduler().get_loop_period_us()/800.0f;
-    }
+    interval_ms = cap_message_interval(interval_ms);
 #endif
 
     // check if it's a specially-handled message:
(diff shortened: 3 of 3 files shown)


libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3070,6 +3070,13 @@ MAV_RESULT GCS_MAVLINK::set_message_interval(uint32_t msg_id, int32_t interval_u
     } else {
         interval_ms = interval_us / 1000;
     }
+#if AP_SCHEDULER_ENABLED
+    // message rate can't be greate than 0.8 * SCHED_LOOP_RATE
```
amilcarlucas, line 3074:
```suggestion
    // message rate can't be greater than 0.8 * SCHED_LOOP_RATE
```

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3070,6 +3070,13 @@ MAV_RESULT GCS_MAVLINK::set_message_interval(uint32_t msg_id, int32_t interval_u
...
+    if (interval_ms*800 < AP::scheduler().get_loop_period_us()) {
```
nexton-winjeel, line 3075: Might be nice to have a constant or a function here so it doesn't get out of sync with the logic in `set_ap_message_interval`.

stephendade: Sure. I've now pulled it out to ``libraries/GCS_MAVLink/GCS.h``

libraries/GCS_MAVLink/GCS_Common.cpp
stephendade: Fixed

stephendade:
Appears to be some issue with ``COMMAND_INT`` in the rover autotests. ``COMMAND_LONG`` works fine though.

Works OK with MAVProxy:
```
long SET_MESSAGE_INTERVAL 105 10000
command_int GLOBAL SET_MESSAGE_INTERVAL 0 0 105 10000 0 0 0 0 0
```
are equivalent.

Maybe some issue with the test ... will investigate further.

stephendade:
Appears to be an unrelated test failure in ``DeadreckoningNoAirSpeed`` in Plane. but otherwise the tests are now happy.

Tools/autotest/vehicle_test_suite.py
```diff
@@ -10855,10 +10855,19 @@ def SET_MESSAGE_INTERVAL(self):
     def MESSAGE_INTERVAL_COMMAND_INT(self):
         '''Test MAV_CMD_SET_MESSAGE_INTERVAL works as COMMAND_INT'''
         original_rate = round(self.measure_message_rate("VFR_HUD", 20))
-        self.context_set_message_rate_hz('VFR_HUD', original_rate*2, run_cmd=self.run_cmd_int)
+        self.context_set_message_rate_hz(mavutil.mavlink.MAVLINK_MSG_ID_VFR_HUD, original_rate*2,
```
peterbarker, line 10858: Why change this?

libraries/GCS_MAVLink/GCS.h
```diff
@@ -28,6 +28,7 @@
 #include <AP_AHRS/AP_AHRS_config.h>
 #include <AP_Arming/AP_Arming_config.h>
 #include <AP_Airspeed/AP_Airspeed_config.h>
+#include <AP_Scheduler/AP_Scheduler.h>
```
peterbarker, line 31: Move the function implementation to the .cpp file.

It's too big to go in the header.

libraries/GCS_MAVLink/GCS.h
```diff
@@ -267,6 +268,16 @@ class GCS_MAVLINK
     // accessor for uart
     AP_HAL::UARTDriver *get_uart() { return _port; }
 
+#if AP_SCHEDULER_ENABLED
+    // cap the MAVLink message rate. It can't be greater than 0.8 * SCHED_LOOP_RATE
+    uint16_t cap_message_interval(uint16_t interval_ms) {
```
peterbarker, line 273:
```suggestion
    uint16_t cap_message_interval(uint16_t interval_ms) const {
```

libraries/GCS_MAVLink/GCS.h
```diff
@@ -267,6 +268,16 @@ class GCS_MAVLINK
...
+        if (interval_ms != 0 && interval_ms*800 < AP::scheduler().get_loop_period_us()) {
```
peterbarker, line 274:
```suggestion
        if (interval_ms*800 < AP::scheduler().get_loop_period_us()) {
```

kind of a given.  Unless your loop period is zero, in which case you have bigger problems.  And you'll still return the right result

peterbarker (changes requested):

libraries/GCS_MAVLink/GCS.h
stephendade: Moved to cpp

libraries/GCS_MAVLink/GCS.h
stephendade: removed the zero check

libraries/GCS_MAVLink/GCS.h
stephendade: Fixed

Tools/autotest/vehicle_test_suite.py
stephendade: reverted

stephendade:
For the autotests, there appears to be an issue with ``context_set_message_rate_hz`` where it incorrectly sets the message interval:

```
AT-0002.6: Sending COMMAND_LONG to (1,1) (MAV_CMD_SET_MESSAGE_INTERVAL=511) (p1=74.000000 p2=10.100010 p3=0.000000 p4=0.000000 p5=0.000000 p6=0.000000  p7=0.000000)
AT-0002.6: ACK received: COMMAND_ACK {command : 511, result : 2, progress : 0, result_param2 : 0, target_system : 250, target_component : 250} (0.000000s)
```

So it tries to set an interval of 10.100010us (which is _way_ above the SCHED_LOOP_RATE). Best guess there's a missing Hz->usec conversion somewhere ... can't seem to find it though :(

libraries/GCS_MAVLink/GCS.h
peterbarker, line 31: This include needs to be removed.

It introduces a lot of dependencies, and since a *lot* of places use GCS.h ths will slow builds down.

peterbarker (changes requested):

stephendade:
@peterbarker, I've fixed up the autotest issues. Appears to be 2x unrelated tests failing

peterbarker (approved):
LGTM

stephendade:
For the Dev Call, since I likely won't be able to attend:
This was in response to #28101, where a user tried to get a message at a faster rate than 80% of ``SCHED_LOOP_RATE``. This is more apparent in Rover, where ``SCHED_LOOP_RATE`` defaults to 50Hz.

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -225,6 +225,19 @@ void GCS_MAVLINK::send_power_status(void)
                                   hal.analogin->power_status_flags());
 }
 
+#if AP_SCHEDULER_ENABLED
+// cap the MAVLink message rate. It can't be greater than 0.8 * SCHED_LOOP_RATE
+uint16_t GCS_MAVLINK::cap_message_interval(uint16_t interval_ms) const {
```
rmackay9, line 230: nitpick: normally we put the curly brace on its own line

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3127,6 +3136,12 @@ MAV_RESULT GCS_MAVLINK::set_message_interval(uint32_t msg_id, int32_t interval_u
     } else {
         interval_ms = interval_us / 1000;
     }
+#if AP_SCHEDULER_ENABLED
+    if (interval_ms != 0 && cap_message_interval(interval_ms) > interval_ms) {
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Requested message rate too fast. Increase SCHED_LOOP_RATE");
```
peterbarker, line 3141: Include message number here

peterbarker (changes requested):

libraries/GCS_MAVLink/GCS_Common.cpp
stephendade: Fixed

libraries/GCS_MAVLink/GCS_Common.cpp
stephendade: Added message ID to GCS_SEND_TEXT

peterbarker:
Merged, thanks!

PR #28147, opened 2024-09-18
Fix sdkconfig filename with removed tab character

chuanmin0317:


peterbarker:
@chuanmin0317 somehow this is conflicting?

tpwrules:
The commit contents don't seem to match the message, why are you deleting these portions?

davidbuzz:
I dont think this PR is valid .  Peter its an auto-regenerated file... the esp IDF reads that file, parses it and then re-writes it, and depending on the IDF code/version being used, it evolves over time.

PR #28148, opened 2024-09-19
AP_BattMonitor: add option minimum volt option

loki077:
Allows aggregate monitors like Sum and ESC to report the minimum voltage instead of the average voltage. This is useful for systems with multiple isolated batteries where the lowest voltage is the limiting factor.

@robertlong13

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
@@ -57,6 +57,7 @@ void AP_BattMonitor_ESC::read(void)
     uint8_t voltage_escs = 0;     // number of ESCs with valid voltage
     uint8_t temperature_escs = 0; // number of ESCs with valid temperature
     float voltage_sum = 0;
+    float voltage_min = 0;
     float current_sum = 0;
     float temperature_sum = 0;
     float consumed_mah_sum = 0.0;
@@ -82,6 +83,9 @@ void AP_BattMonitor_ESC::read(void)
         }
 
         if (telem.get_voltage(i, voltage)) {
+            if (voltage_escs == 0 || voltage_min > voltage) {
+                voltage_min = voltage;
+            }
             voltage_sum += voltage;
             voltage_escs++;
         }
@@ -132,6 +136,12 @@ void AP_BattMonitor_ESC::read(void)
         // ESCs provide current but not consumed mah, integrate manually
         update_consumed(_state, dt_us);
 
+    }
+
+    // Check if we want to report the minimum voltage instead of the average
+    // (we do this after calculating consumed Wh so this doesn't affect that calculation)
+    if (option_is_set(AP_BattMonitor_Params::Options::Minimum_Voltage)) {
+        _state.voltage = voltage_min;
     }
 }
 
--- a/libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
@@ -154,7 +154,7 @@ const AP_Param::GroupInfo AP_BattMonitor_Params::var_info[] = {
     // @Param: OPTIONS
     // @DisplayName: Battery monitor options
     // @Description: This sets options to change the behaviour of the battery monitor
-    // @Bitmask: 0:Ignore DroneCAN SoC, 1:MPPT reports input voltage and current, 2:MPPT Powered off when disarmed, 3:MPPT Powered on when armed, 4:MPPT Powered off at boot, 5:MPPT Powered on at boot, 6:Send resistance compensated voltage to GCS, 7:Allow DroneCAN InfoAux to be from a different CAN node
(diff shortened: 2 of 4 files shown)


IamPete1:
You can setup your low voltage failsafes on the individual battery monitors before combining them in the sum. That is not a option for ESC of course. 

I think the problem here is that the consumed energy calculation will be thrown off. Maybe you could apply the min voltage after the `update_consumed` call?

robertlong13:
That works for failsafes, but we want to display the number easily to the pilot. That could be done with GCS customizations instead I suppose, but this felt nicer.

Good point about the consumed energy calculation. We can revise so that this option doesn't impact that calc.

Hwurzburg (changes requested):
this should not be included on minimize boards..

robertlong13:
> this should not be included on minimize boards..

If we need to go this route to prevent this 96 byte change (for Plane on Pixhawk1-1M), then I'd recommend we go further and save 200 bytes over the current master by disabling `BATT#_OPTIONS` entirely on minimized boards. That would certainly be a lot clearer from a documentation standpoint, since compiling out one bit of an options parameter can really lead to confusion. It also makes it much easier to explain in `build_options.py`

I can make that a part of this PR if we want.

Hwurzburg:
> > this should not be included on minimize boards..
> 
> If we need to go this route to prevent this 96 byte change (for Plane on Pixhawk1-1M), then I'd recommend we go further and save 200 bytes over the current master by disabling `BATT#_OPTIONS` entirely on minimized boards. That would certainly be a lot clearer from a documentation standpoint, since compiling out one bit of an options parameter can really lead to confusion. It also makes it much easier to explain in `build_options.py`
> 
> I can make that a part of this PR if we want.

I agree, the options are not used in the minimized boards since its difficult( if not impossible )on some boards to use more than one monitor or a non-analog one(usually built-in) ....thats why everything except analog is omitted from minimize builds....we should make the SUM backend a build option while we are at it like the others are

peterbarker:
Wait - getting rid of `BATTn_OPTIONS` would get rid of value 64, use-resting-voltage.  That would be a very bad change IMO.

I suggest we accept this PR without fussing too much about the 100 bytes.

libraries/AP_BattMonitor/AP_BattMonitor_ESC.cpp
```diff
@@ -133,6 +137,12 @@ void AP_BattMonitor_ESC::read(void)
         update_consumed(_state, dt_us);
 
     }
+
+    // Check if we want to report the minimum voltage instead of the average
+    // (we do this after calculating consumed Wh so this doesn't affect that calculation
```
IamPete1, line 142:
```suggestion
    // (we do this after calculating consumed Wh so this doesn't affect that calculation)
```

libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
```diff
@@ -154,7 +154,7 @@ const AP_Param::GroupInfo AP_BattMonitor_Params::var_info[] = {
     // @Param: OPTIONS
     // @DisplayName: Battery monitor options
     // @Description: This sets options to change the behaviour of the battery monitor
-    // @Bitmask: 0:Ignore DroneCAN SoC, 1:MPPT reports input voltage and current, 2:MPPT Powered off when disarmed, 3:MPPT Powered on when armed, 4:MPPT Powered off at boot, 5:MPPT Powered on at boot, 6:Send resistance compensated voltage to GCS, 7:Allow DroneCAN InfoAux to be from a different CAN node
+    // @Bitmask: 0:Ignore DroneCAN SoC, 1:MPPT reports input voltage and current, 2:MPPT Powered off when disarmed, 3:MPPT Powered on when armed, 4:MPPT Powered off at boot, 5:MPPT Powered on at boot, 6:Send resistance compensated voltage to GCS, 7:Allow DroneCAN InfoAux to be from a different CAN node, 9:Aggregate monitor reports minimum voltage instead of average.
```
IamPete1, line 157: I wonder if it would be better to call out sum and ESC specifically.

libraries/AP_BattMonitor/AP_BattMonitor_Sum.cpp
```diff
@@ -73,6 +74,9 @@ AP_BattMonitor_Sum::read()
         if (!_mon.healthy(i)) {
             continue;
         }
+        if (voltage_count == 0 || _mon.voltage(i) < voltage_min) {
+            voltage_min = _mon.voltage(i);
+        }
         voltage_sum += _mon.voltage(i);
```
IamPete1, line 80: Add the fetched voltage as a const float just to prevent getting it three times.

IamPete1 (commented):
Basically looks good.

libraries/AP_BattMonitor/AP_BattMonitor_Sum.cpp
```diff
@@ -103,6 +108,12 @@ AP_BattMonitor_Sum::read()
 
     update_consumed(_state, dt_us);
 
+    // Check if we want to report the minimum voltage instead of the average
```
tridge, line 111: race condition for scripting or DDS, I think we need to create a local float state_voltage and then set that once

tridge (changes requested):

robertlong13:
Closed in favor of #28227

PR #28149, opened 2024-09-19
Use get_channel_yaw() and friends in AP_Arming

peterbarker:
We have methods which return you the relevant channels (or a dummy if your vehicle is mis-configured).

Use them.

```
CubeOrange-periph-heavy  *                                                       
Durandal                            -48    *           -48     -48               -48    -48    -48
Hitec-Airspeed           *                                                       
KakuteH7-bdshot                     -64    *           -56     -64               -64    -64    -64
MatekF405                           -56    *           -56     -64               -56    -56    -56
Pixhawk1-1M-bdshot                  -64                -56     -64               -64    -40    40
f103-QiotekPeriph        *                                                       
f303-Universal           *                                                       
revo-mini                           -56    *           -64     -64               -64    -56    -56
skyviper-v2450                                         -64                       
```

We appear to faithfully copy what master does here:
```
MANUAL> arm throttle
MANUAL> Got COMMAND_ACK: COMPONENT_ARM_DISARM: ACCEPTED
AP: Throttle armed
ARMED

MANUAL> 
MANUAL> param show RCMAP*
MANUAL> RCMAP_PITCH      2
RCMAP_ROLL       45
RCMAP_THROTTLE   3
RCMAP_YAW        56
Flight battery 100 percent
```

We'll need to get a lot more complicated than a single-bit "enforce RPY are neutral" to do better here.

### DIFF
--- a/libraries/AP_Arming/AP_Arming.cpp
+++ b/libraries/AP_Arming/AP_Arming.cpp
@@ -44,7 +44,6 @@
 #include <AP_Scripting/AP_Scripting.h>
 #include <AP_Camera/AP_RunCam.h>
 #include <AP_GyroFFT/AP_GyroFFT.h>
-#include <AP_RCMapper/AP_RCMapper.h>
 #include <AP_VisualOdom/AP_VisualOdom.h>
 #include <AP_Parachute/AP_Parachute.h>
 #include <AP_OSD/AP_OSD.h>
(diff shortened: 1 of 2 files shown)


peterbarker:
In addition to the autotests I've checked this code in SITL with much use of `rc` and `arm throttle`.

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -736,19 +735,22 @@ bool AP_Arming::rc_arm_checks(AP_Arming::Method method)
         check_failed(ARMING_CHECK_PARAMETERS, true, "Mode channel and RC%d_OPTION conflict", rc().flight_mode_channel_number());
         check_passed = false;
     }
-    const RCMapper * rcmap = AP::rcmap();
-    if (rcmap != nullptr) {
+    {
+        auto &xrc = rc();
         if (!rc().option_is_enabled(RC_Channels::Option::ARMING_SKIP_CHECK_RPY)) {
```
IamPete1, line 740:
```suggestion
        if (!xrc.option_is_enabled(RC_Channels::Option::ARMING_SKIP_CHECK_RPY)) {
```
There are a few places still using `rc()`

IamPete1: Or use `rc()` everywhere I guess.

peterbarker: I did only want to change the lines I needed to.  It would be nice to be consistent, I agree.

peterbarker: ... I've removed the `xrc` intermediate variable, costing 0 bytes on MatekF405 and 8 on CubeOrange.

Worth it to keep things simple, and I was finding `xrc` ugly, too :-)

tridge (approved):

PR #28155, opened 2024-09-19
Copter: don't pass `channel_yaw->norm_input_dz()` into `get_pilot_desired_yaw_rate`

IamPete1:
We always pass the same value into the function so we can move the getting of the value down.


rmackay9 (approved):
looks good, thanks!

probably goes without saying but of course, give it a quick test at least to be sure it's OK

IamPete1:
> probably goes without saying but of course, give it a quick test at least to be sure it's OK

Yeah, yaw still works in SITL.

rmackay9:
Thanks, let's get this out of the way before our dev call

PR #28157, opened 2024-09-20
AP_DroneCAN: Fix constructor initialization order

bugobliterator:
bad ordering was leading to dronecan dna server getting initialised before canard interface leading to bad index setup

This should have been caught by compiler and by the looks of it, wasn't. Have we got contructor initialise ordering check turned off?

### DIFF
--- a/libraries/AP_DroneCAN/AP_DroneCAN.h
+++ b/libraries/AP_DroneCAN/AP_DroneCAN.h
@@ -235,9 +235,11 @@ private:
 
     uint32_t *mem_pool;
 
-    AP_DroneCAN_DNA_Server _dna_server;
-
     uint8_t _driver_index;
+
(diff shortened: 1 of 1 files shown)


magicrub:
Is this a case of the constructor doing too much?

tridge:
@bugobliterator what were the symptoms of this bug? do we need a 4.5 or even 4.4 backport?

robertlong13:
> @bugobliterator what were the symptoms of this bug? do we need a 4.5 or even 4.4 backport?

This only needs a 4.5 backport. Issue was not present in 4.4.

Symptoms are that can driver 2 does not log CAND messages and the autopilot constantly requests NodeInfo from every node on CAN2, leading to a lot of spam on the bus.

tridge (approved):

PR #28159, opened 2024-09-20
Tools: Expose DDS build options in build_options.py

Ryanf55:
As requested by @peterbarker  - should we create the networking options, and if so, how should we deconflict the existing CLI options?


tridge (changes requested):
needs extract_features.py update

Ryanf55:
Remove the original `--enable-dds` from `waf`.

Ryanf55:
For testing: 
* ./Tools/autotest/test_build_options.py --define-match-glob "*_DDS_*"

PR #28168, opened 2024-09-20
Replay: Declare the type of the loop variable and consolidate the processing

muramura:
Type declarations and processing can be done together.


peterbarker (approved):
Looks like someone rearranged code in here and did not do a great job.  Probably me.

PR #28169, opened 2024-09-20
Replay: The number of elements in a char array can be obtained using SIZEOF

muramura:
I think it's better to get the size using SIZEOF rather than a literal value.

### DIFF
--- a/Tools/Replay/MsgHandler.cpp
+++ b/Tools/Replay/MsgHandler.cpp
@@ -230,7 +230,7 @@ void MsgHandler::field_not_found(uint8_t *msg, const char *label)
 {
     char all_labels[256];
     uint8_t type = msg[2];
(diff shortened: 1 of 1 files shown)


magicrub:
even though this is correct in this case, the better solution is to use ARRAY_SIZE(). Anyone else have any preference of sizeof() vs ARRAY_SIZE()?

peterbarker:
> even though this is correct in this case, the better solution is to use ARRAY_SIZE(). Anyone else have any preference of sizeof() vs ARRAY_SIZE()?

Using `ARRAY_SIZE()` consistently is probably a good idea.  `ARRAY_SIZE()` will fail if you use it on something which isn't an array, `sizeof()` won't.

peterbarker:
Changed this to ARRAY_SIZE and marked for MergeOnCIPass

PR #28174, opened 2024-09-20
AP_ExternalAHRS: Change GPS to GNSS

muramura:
The element name in message_ofs is set to GNSS.
I will change the definition name to match the element name.


peterbarker:
@muramura does this change it to match the datasheet?

@valbv your thoughts?

valbv:
It looks better
Thanks for these cosmetic changes

valbv:
I think naming in changes is more correct
But in device documentation used "GPS" in title

image
image

peterbarker:
OK, so sadly I think that while this PR does rename the variables to be more accurate, we should stick with the documentation's nomenclature.  If anything I guess the entry in the union could be changed to reflect the datasheet.

Thanks anyway, @muramura , closing this one.

PR #28182, opened 2024-09-21
AP_CANManager: fixed de-registering CAN forwarding

### LINKED ISSUE #28175: Copter-4.6.0-dev: DroneCAN over MAVLink broken?
Copter-4.6.0-dev seems to have broken something that causes MP's DroneCAN/UAV screen to be unable to communicate DroneCAN enabled devices (e.g. a Here4 GPS)

This is what MP's screen shows when Copter-4.6.0-dev (on a CubeOrange with Here4 GPS:
image

This what it looks like using Copter-4.5.6.
image

tridge:
this prevented MissionPlanner MAVCAN UI from working reliably as the CANManager could think we are registered when we are not

This particularly impacted MissionPlanner as it does a sequence of registration of CAN forwarding followed by de-registering (ie. sending bus -1) before finally registering fully

Fixes #28175

### DIFF
--- a/libraries/AP_CANManager/AP_CANManager.cpp
+++ b/libraries/AP_CANManager/AP_CANManager.cpp
@@ -426,38 +426,42 @@ bool AP_CANManager::handle_can_forward(mavlink_channel_t chan, const mavlink_com
 {
     WITH_SEMAPHORE(can_forward.sem);
     const int8_t bus = int8_t(packet.param1)-1;
+
     if (bus == -1) {
         /*
           a request to stop forwarding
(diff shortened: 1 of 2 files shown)


andyp1per (approved):

tridge:
@andyp1per I found a 2nd issue that impacts using this with multicast, fixed in a 2nd commit

rmackay9:
I've tested this fix and it does resolve the issue with Mission Planner's DroneCAN/UAVCAN screen not working, thanks!
image

PR #28183, opened 2024-09-21
AP_Camera: add Python examples for GStreamer UDP and RTSP video streams

srmainwaring:
Python examples demonstrating how to manipulate UDP and RTSP video streams using GStreamer. 

The UDP to RTSP converter addresses the problem of consuming a simulated video stream from Gazebo in more than one application using standard installations of GStreamer and opencv (no need for a custom build and install of opencv).

### Related:

- https://github.com/ArduPilot/ardupilot/pull/27773
- https://github.com/ArduPilot/ardupilot/pull/27903
- https://github.com/ArduPilot/ardupilot_gazebo?tab=readme-ov-file#3-streaming-camera-video

### DIFF
--- a/libraries/AP_Camera/examples/README.md
+++ b/libraries/AP_Camera/examples/README.md
@@ -1,0 +1,106 @@
+# AP_Camera examples
+
+
+## GStreamer examples
+
+The following examples demonstrate how to manipulate GStreamer pipelines to
+capture and display video streams from a UDP or RTSP source. They are useful
+for providing test images to camera algorithms, or forwarding streams from
+simulators such as Gazebo to ground control stations.  
+
+For further details see:
+
+- [GStreamer tutorials](https://gstreamer.freedesktop.org/documentation/tutorials/index.html?gi-language=c)
+
+
+### Install dependencies
+
+#### Ubuntu
+
+```bash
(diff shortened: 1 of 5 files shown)


peterbarker (approved):

rmackay9:
Another possible place to put this would be in the Tools/camera_gimbals directory where I've put some download scripts https://github.com/ArduPilot/ardupilot/tree/master/Tools/cameras_gimbals

srmainwaring:
@rmackay9 - the idea was to keep the scripts in `AP_Camera/examples` to complement the GSoC camera tracking examples in #27903. 

The other thought was to move them into a MAVProxy module/example. I've started looking through the `mavproxy_SIYI` module, as it has many features that would be nice to have in a general camera / gimbal screen, but also is heavily customised to the SIYI and I'm not sure if the general purpose features can be factored out easily.

Regarding the opencv / gstreamer integration: I've found that a custom build of opencv-python is possible on macOS using `brew` if the formula for opencv is edited and the package is installed using a `--build-from-source` flag. I guess the questions is whether requiring a custom build of opencv on Ubuntu and macOS is something we want users to do? This is also the case  for the `mavproxy_SIYI` module as the video streaming support in `MPImage` also requires opencv with GStreamer support, so perhaps it's already assumed?

rmackay9:
Hi @srmainwaring,

I've also looked at the MAVProxy Siyi module and I think we will need to create a separate, more general purpose camera gimbal control screen.  Hopefully we can re-use some things that the Siyi module has added to (like the MPImage class).

rmackay9 (approved):

PR #28188, opened 2024-09-21
AP_Volz_Protocol: add support for telem

IamPete1:
This adds support for logging desired and actual position, primary and secondary voltage, primary and secondary current, motor and PCB temperature. 

The telem data is at quite a low rate because requesting telem takes the place of a position command. So telem does lower the position update rates. A single servo would send commands at 320 Hz and telem requests at 320Hz, there are three messages we request so a full telem update would be at (320/3) 100Hz. For four servos each servo would be updated at 128Hz and telem also at 128Hz. However there are more servos to cycle through so a full telem update for each servo would happen at (128/(3*4)) 10Hz.

Telem is only used for logging, each servo is logged at 5Hz using a new `VOLZ` log with instances for each servo.

Note that we do get position feedback at a much higher rate, so this could potentially be logged faster. 

A example plot:

image

### DIFF
--- a/libraries/AP_Volz_Protocol/AP_Volz_Protocol.cpp
+++ b/libraries/AP_Volz_Protocol/AP_Volz_Protocol.cpp
@@ -14,7 +14,7 @@
 #include <SRV_Channel/SRV_Channel.h>
 #include <AP_BoardConfig/AP_BoardConfig.h>
 
-#define SET_EXTENDED_POSITION_CMD      0xDC
+#include <AP_Logger/AP_Logger.h>
 
 // Extended Position Data Format defines -100 as 0x0080 decimal 128, we map this to a PWM of 1000 (if range is default)
 #define PWM_POSITION_MIN               1000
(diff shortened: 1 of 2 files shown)


libraries/AP_Volz_Protocol/AP_Volz_Protocol.cpp
```diff
@@ -171,10 +375,55 @@ void AP_Volz_Protocol::update()
...
+            );
+        }
+    }
+#endif // HAL_LOGGING_ENABLED
 }
 
-// calculate CRC for volz serial protocol and send the data.
-void AP_Volz_Protocol::send_command(CMD &cmd)
+// Return the crc for a given command packet
+uint16_t AP_Volz_Protocol::calculate_crc(const CMD &cmd) const
```
tridge, line 426: not a blocker, but check if the same CRC as crc_crc16_ibm()

tridge (approved):

PR #26963, opened 2024-05-03
Rover: Converging message output

muramura:
By making it a ternary operator, one can have one message output.

PR #28190, opened 2024-09-22
Rover: remove handling of prearm empty-string case

peterbarker:
it is clear by inspection that this string can never be empty when the called function returns false.  AP_OAPathPlanner::pre_arm_check is not that complicated!

Lightly tested in SITL by setting `OA_TYPE` to an invalid value

Replaces https://github.com/ArduPilot/ardupilot/pull/26963/files

### DIFF
--- a/Rover/AP_Arming.cpp
+++ b/Rover/AP_Arming.cpp
@@ -168,12 +168,7 @@ bool AP_Arming_Rover::oa_check(bool report)
         return true;
     }
 
-    // display failure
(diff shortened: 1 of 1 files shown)


tpwrules:
Errant change in `AP_HAL_ChibiOS`?

peterbarker:
> Errant change in `AP_HAL_ChibiOS`?

Whoops, fixed.

rmackay9 (approved):

PR #24832, opened 2023-08-31
AP_Compass: Compare with string

muramura:
If you get a string from a device, determine it by the string.

PR #28191, opened 2024-09-22
AP_Compass: use memcmp to check whoami result code from HMC5843 device

peterbarker:
Replaces https://github.com/ArduPilot/ardupilot/pull/24832/files

Tested on a revo-mini (broke it / fixed it to make sure this codepath was crossed)


peterbarker:
.... still saves 8 bytes.

tridge (approved):

PR #13842, opened 2020-03-19
Copter: Match cyclic rate to scheduler rate

muramura:
I'm testing flight with a Copter scheduler rate other than 400.
I think it's better to adjust from 400 fixed to scheduler rate.

PR #28192, opened 2024-09-22
correct initialisation of PrecLand buffers

peterbarker:
this value determines the size of buffers allocated.

Too high and you waste RAM.  Too low and you probably won't store data correctly.

Replaces https://github.com/ArduPilot/ardupilot/pull/13842/files

### DIFF
--- a/ArduCopter/precision_landing.cpp
+++ b/ArduCopter/precision_landing.cpp
@@ -8,7 +8,9 @@
 
 void Copter::init_precland()
 {
-    copter.precland.init(400);
+    // scheduler table specifies 400Hz, but we can call it no faster
+    // than the scheduler loop rate:
+    copter.precland.init(MIN(400, scheduler.get_loop_rate_hz()));
 }
 
 void Copter::update_precland()
--- a/ArduPlane/system.cpp
+++ b/ArduPlane/system.cpp
@@ -162,7 +162,9 @@ void Plane::init_ardupilot()
 #endif
 
 #if AC_PRECLAND_ENABLED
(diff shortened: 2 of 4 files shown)


rishabsingh3003:
Accept for the changes made in Plane, I don't think this will functionally change anything since the init function already takes the min of loop rate and demanded update rate internally: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AC_PrecLand/AC_PrecLand.cpp#L229

peterbarker:
> Except for the changes made in Plane, I don't think this will functionally change anything since the init function already takes the min of loop rate and demanded update rate internally: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AC_PrecLand/AC_PrecLand.cpp#L229

I've pushed up a patch to remove that code - in the basis that PrecLand shouldn't be making any assumptions about how the method is called, jsut trust that the caller knows what it is doing.

OK?

rishabsingh3003 (approved):
LGTM

amilcarlucas:
I do not understand what gets changed with this. Apart the AP_Scheduler.h no longer needs to be included.

peterbarker:
> I do not understand what gets changed with this. Apart that the AP_Scheduler.h no longer needs to be included.

Sooo... what happens when you're not calling this from the scheduler?  The point is you can't assume the rate at this method is being called is limited by the scheduler loop rate.  In all likelihood it is, of course.

rmackay9 (approved):
I'm not quite sure why we're fixing things aren't broken

PR #28194, opened 2024-09-22
vehicle: only send airspeed when enabled

tpwrules:
Avoids debug message spam about sending an unknown message on ESP32 where this is disabled.

Issue not caught with https://github.com/ArduPilot/ardupilot/pull/22709


IamPete1 (approved):
LGTM

PR #20949, opened 2022-06-10
GCS_MAVLink: implement sys_status.errors_comm

amilcarlucas:

PR #28195, opened 2024-09-22
AntennaTracker: logging option for telemetry radio status

landswellsong:
I've noticed that Tracker did not log `RADIO_STATUS` packets from the telemetry radios, which is odd since its whole point revolves around telemetry radio handling. I saw that some unification work has been done on `master` to move all the relevant code outside of vehicle specific classes, but still the Tracker did not implement the flag it wouldn't auto-affect it. The tracker didn't have a `MAS_LOG_PM` define either so I took a bit of creative liberty with the define. I don't think anyone  would really need to debug `PM` on the Tracker, but in case it's also centrally handled now I can unify the define with other vehicles.

Not flight tested yet.

servo.cpp changes were needed since G_Dt is a variable in AP_Vehicle.h and to make it clear that it's a constant

### DIFF
--- a/AntennaTracker/GCS_Mavlink.h
+++ b/AntennaTracker/GCS_Mavlink.h
@@ -2,6 +2,8 @@
 
 #include <GCS_MAVLink/GCS.h>
 
+#include "defines.h"
+
 class GCS_MAVLINK_Tracker : public GCS_MAVLINK
 {
 
@@ -29,6 +31,10 @@ protected:
 
     void send_nav_controller_output() const override;
     void send_pid_tuning() override;
+
+#if HAL_LOGGING_ENABLED
+    uint32_t log_radio_bit() const override { return MASK_LOG_RADIO; }
+#endif
 
 private:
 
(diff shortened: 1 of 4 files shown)


amilcarlucas:
Also take a look at https://github.com/ArduPilot/ardupilot/pull/20949 it might be helpful here

landswellsong:
@amilcarlucas I don't seem to draw any connection here :( Can you explain please?

amilcarlucas:
That other PR makes dropped packets visible. And that is also useful information in an antenna tracker. And could also be added to logging

landswellsong:
I don't think those are the same values. The [radio status](https://mavlink.io/en/messages/common.html#RADIO_STATUS) message already does track the lost packets and moreover it also shows the FEC-corrected count and generally the link can be used for other purposes e.g. video streaming so it would make sense to trust the radio stats generator instead of second guessing the state of the whole link by just a part of it (because shapers can be in question).

That said it's definitely something that can be useful for tracker, do you want me to help with testing?

amilcarlucas:
Yes, it would be great if could test my PR and report the result in the comments.

landswellsong:
Rebased over 4.7. Have been using this for a while, will attach testing data in a few hours.

PR #28197, opened 2024-09-22
Tools/environment_install: print how to activate venv if not default

tpwrules:
Nicer for the users. Split out from https://github.com/ArduPilot/ardupilot/pull/28080/ .


peterbarker:
Merged, thanks!

PR #28198, opened 2024-09-22
Tools/environment_install: ensure venv is installed on all Ubuntus

tpwrules:
Corrects missing 22.04 Jammy case and avoids need for hardcoded table. Split out from #28080 .

Totally untested but seems right...

### DIFF
--- a/Tools/environment_install/install-prereqs-ubuntu.sh
+++ b/Tools/environment_install/install-prereqs-ubuntu.sh
@@ -373,16 +373,10 @@ fi
 PIP_USER_ARGUMENT="--user"
 
 # create a Python venv on more recent releases:
-PYTHON_VENV_PACKAGE=""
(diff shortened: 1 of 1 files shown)


khancyr (changes requested):
doesn't work. It enable venv on focal: see https://github.com/ArduPilot/ardupilot/actions/runs/10983477074/job/30492890484 

We got some autotest for that use the install scripts . If you are used to Docker that is quite simple to test those changes

khancyr:
Not sure why we want venv now on jammy ?

tpwrules:
Why would a venv only be desired on some versions? CI passes on all of them?

I think the motivation is that ESP-IDF needs venv.

tpwrules:
Discussed briefly on the devcall and the consensus was that changing this is not desired. We will come up with an ESP32 specific solution if necessary.

PR #28199, opened 2024-09-22
Plane: added Q_LAND_DIS_DELAY

tridge:
this adds an optional disarm delay after an auto landing, allowing for recovery from false landing detection

This option is in response to a false landing on one quadplane in high wind, where the high activity of the rate controllers led to a false landing detection. The aircraft had Q_LAND_ALTCHG = 0.8, making it more vulnerable to false landings, but that is needed on some aircraft to make landing detection reliable

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -546,6 +546,15 @@ const AP_Param::GroupInfo QuadPlane::var_info2[] = {
     // @Increment: 1
     // @User: Standard
     AP_GROUPINFO("APPROACH_DIST", 39, QuadPlane, approach_distance, 0),
+
+    // @Param: LAND_DIS_DELAY
+    // @DisplayName: auto landing disarm delay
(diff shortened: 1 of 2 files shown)


davidbuzz (commented):
i like it

ArduPlane/quadplane.cpp
```diff
@@ -3461,31 +3472,73 @@ bool QuadPlane::land_detector(uint32_t timeout_ms)
...
+    bool land_detected = land_detector(4000);
 
-        if (plane.in_auto_mission_id(MAV_CMD_NAV_PAYLOAD_PLACE)) {
-            // for payload place with full landing we shutdown motors
-            // and wait for the lua script to trigger a climb (using
-            // landing abort) or disarm
-            set_desired_spool_state(AP_Motors::DesiredSpoolState::SHUT_DOWN);
+    if (poscontrol.get_state() == QPOS_LAND_DISARM_PENDING) {
+        if (!land_detected) {
+            gcs().send_text(MAV_SEVERITY_INFO,"Land false completion, recovering");
```
tridge, line 3484: change to MAV_SEVERITY_ERROR

PR #25068, opened 2023-09-21
SRV_Channel: Consider igntion something that should be e-stopped

WickedShell:
This is a bit of a drive by, but the comment at the top of the function is:
```
// return true if function is for anything that should be stopped in a e-stop situation, ie is dangerous
```

and ignition control sounds exactly like that to me.

PR #28200, opened 2024-09-23
AP_ICEngine: don't run engine with safety engaged

robertlong13:
If you set the option bit for "allow throttle while disarmed", you can crank the starter even while the safety switch is engaged, and unless you use a servo pwm as your starter motor controller, you can't even configure it to not do that.

Some places use the `allow_throttle_while_disarmed` method, which additionally checks the safety switch. Other places directly check the options bit. So some stuff works with the safety engaged, and other parts don't. This PR just unifies everything to use the same check.


peterbarker:
Related: https://github.com/ArduPilot/ardupilot/pull/25068

tridge (approved):

PR #28201, opened 2024-09-23
Direct DMA transfer on Invensense v3

andyp1per:
This is done on the other devices and avoids stack allocation and buffer copying in the SPI driver


peterbarker:
@andyp1per how is this a bug?

How has this been tested?

andyp1per:
> @andyp1per how is this a bug?
> 
> How has this been tested?

In my testing I was getting stack overflow without this. It essentially means we are using more stack in the SPI thread than we should.

tridge (approved):

PR #28202, opened 2024-09-23
Tools:remove BATTERY_SUM from <1K builds

Hwurzburg:
option was recently added, but not used...a second monitor on these boards is difficult if not impossible in most cases


IamPete1:
How about removing the sum backend if there are less than 3 battery monitors and then changing the default to two monitors for 1K builds.

Hwurzburg:
not sure I see the point

IamPete1:
> not sure I see the point

Currently 1M boards support 9 battery monitor instances. The sum backend only makes sense if you have three or more instances, you need at least two instances to sum together and one instance for the result.

peterbarker:
@Hwurzburg I think the point of PH's suggestion is that you will save *even more* resources by cutting down the number of battery backend instances allowed.

Hwurzburg:
the problem I see is there is no way to increase the number of monitors with the custom build server IF somebody wanted to have three sensors on a 1K board...ie analog builtin, I2C external, and sum of the two.....without modifying the config files....

tridge (approved):

tridge:
what symptom do users see if they use the feature and it isn't compiled in?

peterbarker:
I have pushed several patches into this PR.

We now emit a clear message if you're trying to use a SUM backend and it is not compiled in - since we are removing support for this and people may be depending on it on 1MB boards (doubtful-face).

image

We now also emit an "unhealthy" message if backend-type is not `Type::NONE` but the driver is nullptr, or if the allocated type doesn't match the configured type.

I've also added a compile-time error if the user tries to enable the USM backend but has specified an insufficient max-backend-count to use it.  If we reduce the number of allowed backends on small boards to less than three then this check may become problematic.

peterbarker:
I've removed the sanity check for the bacnekd  count; it killed half-a-dozen peripherals who include battery but limit the backend count

PR #28203, opened 2024-09-23
AP_DDS: add params for ping timeout and max retries

srmainwaring:
Add parameters to set the time out and number of retries when the DDS client attempts to ping the XRCE agent. Allows the user to increase the timeout before the client exits.

Currently with a RPi4 companion using a PPP connection to a FC, the DDS client times out and exits before the network, PPP daemon and micro-ROS agent are up after a power cycle. 

## Testing

Tested in SITL / Gazebo:

```bash
sim_vehicle.py --debug -v ArduCopter -f gazebo-iris --model json --enable-dds --console
```

```bash
# DDS params with increased number of retries
STABILIZE>  param show *DDS*
DDS_DOMAIN_ID    0
DDS_ENABLE       1           # Enabled
DDS_IP0          127
DDS_IP1          0
DDS_IP2          0
DDS_IP3          1
DDS_MAX_RETRY    60
DDS_TIMEOUT_MS   1000
DDS_UDP_PORT     2019
```

### DIFF
--- a/libraries/AP_DDS/AP_DDS_Client.cpp
+++ b/libraries/AP_DDS/AP_DDS_Client.cpp
@@ -84,6 +84,25 @@ const AP_Param::GroupInfo AP_DDS_Client::var_info[] {
     // @User: Standard
     AP_GROUPINFO("_DOMAIN_ID", 4, AP_DDS_Client, domain_id, 0),
 
+    // @Param: _TIMEOUT_MS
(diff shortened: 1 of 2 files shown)


Ryanf55:
We can get a maximimum of 10,000 * 100 ms = 1000000 ms = 1000 s  =  16.7 minutes. Seems plenty. The docs look good.

Ryanf55 (approved):

PR #28205, opened 2024-09-23
Sub: limit poshold xy velocity to PILOT_SPEED to avoid bounceback

### LINKED ISSUE #28120: Sub: pilot override in POSHOLD mode results in "bounce back"
## Bug report

**Issue details**

Pilot override in POSHOLD mode results in a ~1m "bounce back".

I saw this on real hardware (BlueROV2 heavy + DVL) running a DEV build. I moved the ROV away from the dock while in POSHOLD mode, when I let go of the stick the ROV seemed to jump back towards the dock.

I can reproduce this in SITL on master and Sub-4.5. I was not able to reproduce on Sub-4.1.

To reproduce:
* run sim_vehicle.py with -w, etc.
* run QGC w/ joystick
* wait for _EKF3 IMU0 is using GPS_
* arm, poshold
* use the joystick to move forward aggressively a few meters
* let go of the stick, the sub will move back a meter or so

**Version**

master, Sub-4.5

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[ x ] Submarine

**Airframe type**

All

**Hardware type**

All

**Logs**

I can provide the BlueROV2 logs, but they are a bit big (replay on, long-ish guided mode test). LMK.

clydemcqueen:
This PR limits horizontal velocity in poshold mode, even when the pilot is using the forward and lateral sticks to reposition the sub. This eliminates the possibility of a PSC-caused bounceback. This is verified by an autotest.

The PR also adds a PILOT_SPEED parameter, so there are now 3 WPNAV_SPEED* parameters and 3 PILOT_SPEED* parameters. Previously, poshold was using WPNAV_SPEED for xy and PILOT_SPEED_UP/DN for z, this PR changes it to use PILOT_SPEED* for both xy and z.

(Circle mode also uses WPNAV_SPEED for xy and PILOT_SPEED_UP/DN for z, but I would argue that it should be using WPNAV_SPEED* for both xy and z. This can be a separate PR.)

There are 2 additional small fixes:
* removes unused variable `Sub::desired_climb_rate`
* `ModePoshold::requires_GPS` returns true, so `Sub::set_mode` will call `sub.position_ok` before accepting a mode switch. (`ModePoshold::init()` function also calls `sub.position_ok`, so this doesn't change behavior.)

This should probably be tested on real hardware.

Fixes #28120

### DIFF
--- a/ArduSub/Sub.h
+++ b/ArduSub/Sub.h
@@ -263,9 +263,6 @@ private:
     // Stores initial bearing when armed
     int32_t initial_armed_bearing;
 
-    // Throttle variables
-    int16_t desired_climb_rate;          // pilot desired climb rate - for logging purposes only
-
     // Loiter control
     uint16_t loiter_time_max;                // How long we should stay in Loiter Mode for mission scripting (time in seconds)
     uint32_t loiter_time;                    // How long have we been loitering - The start time in millis
--- a/ArduSub/Sub.h
+++ b/ArduSub/Sub.h
@@ -403,6 +403,7 @@ private:
     float get_roi_yaw();
     float get_look_ahead_yaw();
     float get_pilot_desired_climb_rate(float throttle_control);
+    float get_pilot_desired_horizontal_rate(RC_Channel *channel) const;
     void rotate_body_frame_to_NE(float &x, float &y);
 #if HAL_LOGGING_ENABLED
     // methods for AP_Vehicle:
--- a/ArduSub/mode_poshold.cpp
+++ b/ArduSub/mode_poshold.cpp
@@ -15,8 +15,8 @@ bool ModePoshold::init(bool ignore_checks)
     }
 
     // initialize vertical speeds and acceleration
-    position_control->set_max_speed_accel_xy(sub.wp_nav.get_default_speed_xy(), sub.wp_nav.get_wp_acceleration());
-    position_control->set_correction_speed_accel_xy(sub.wp_nav.get_default_speed_xy(), sub.wp_nav.get_wp_acceleration());
+    position_control->set_max_speed_accel_xy(g.pilot_speed, g.pilot_accel_z);
+    position_control->set_correction_speed_accel_xy(g.pilot_speed, g.pilot_accel_z);
     position_control->set_max_speed_accel_z(-sub.get_pilot_speed_dn(), g.pilot_speed_up, g.pilot_accel_z);
     position_control->set_correction_speed_accel_z(-sub.get_pilot_speed_dn(), g.pilot_speed_up, g.pilot_accel_z);
 
@@ -54,26 +54,6 @@ void ModePoshold::run()
     // set motors to full range
     motors.set_desired_spool_state(AP_Motors::DesiredSpoolState::THROTTLE_UNLIMITED);
 
-    ///////////////////////
-    // update xy outputs //
-    float pilot_lateral = channel_lateral->norm_input();
-    float pilot_forward = channel_forward->norm_input();
-
-    float lateral_out = 0;
-    float forward_out = 0;
-
-    if (sub.position_ok()) {
-        // Allow pilot to reposition the sub
-        if (fabsf(pilot_lateral) > 0.1 || fabsf(pilot_forward) > 0.1) {
-            position_control->init_xy_controller_stopping_point();
-        }
-        sub.translate_pos_control_rp(lateral_out, forward_out);
-        position_control->update_xy_controller();
-    } else {
-        position_control->init_xy_controller_stopping_point();
-    }
-    motors.set_forward(forward_out + pilot_forward);
-    motors.set_lateral(lateral_out + pilot_lateral);
     /////////////////////
     // Update attitude //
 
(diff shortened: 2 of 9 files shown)


ArduSub/mode_poshold.cpp
```diff
@@ -103,12 +83,69 @@ void ModePoshold::run()
...
+    // update z axis
     control_depth();
+
+    // update xy axis
+    // call this after Sub::get_pilot_desired_climb_rate is called so that THR_DZ is reasonable
+    control_horizontal();
+}
+
+// behavior is similar to Sub::get_pilot_desired_climb_rate
+float Sub::get_pilot_desired_horizontal_rate(RC_Channel *channel) const
```
Williangalvani, line 100: wouldn't it make sense to keep it alongside `get_pilot_desired_climbrate()` in attitude.cpp ?

Williangalvani (commented):
looking good! I'll try to do some real-life testing tomorrow. thanks for the fix!

ArduSub/mode_poshold.cpp
clydemcqueen: Good point, I'll move it.

ArduSub/Sub.h
```diff
@@ -406,6 +403,7 @@ class Sub : public AP_Vehicle {
     float get_roi_yaw();
     float get_look_ahead_yaw();
     float get_pilot_desired_climb_rate(float throttle_control);
+    float get_pilot_desired_horizontal_rate(RC_Channel *channel) const;
```
rmackay9, line 406: not a blocker of course but personally I'd stick with the word "speed"

ArduSub/mode_poshold.cpp
```diff
@@ -103,12 +83,42 @@ void ModePoshold::run()
...
+        position_control->input_vel_accel_xy(earth_rates_cm_s, {0, 0});
+
+        // convert pos control roll and pitch angles back to lateral and forward efforts
+        sub.translate_pos_control_rp(lateral_out, forward_out);
+
+        // udpate the xy controller
+        position_control->update_xy_controller();
+    }
+
+    motors.set_forward(forward_out);
```
rmackay9, line 121: This looks like a good change to me because previously the pilot input was essentially added on top of the position controller output.  That would lead to the position controller fighting the user which is terrible and a pattern we try to avoid.

The change in behaviour for users is that they won't be able to control the vehicle laterally in poshold mode anymore unless the EKF has a good position estimate.

Still, looks like an improvement!

Williangalvani: > The change in behaviour for users is that they won't be able to control the vehicle laterally in poshold mode anymore unless the EKF has a good position estimate.

That's an interesting point. what if we move `body_rates_cm_s` up one level and apply that to thrusters if `posititon_ok()` is false? 

Sensors are not perfect and I think users should keep control when they fail.

Williangalvani:
tested and working well on my pool. but I'm concerned about more challenging conditions for DVLs

ArduSub/mode_poshold.cpp
clydemcqueen: > That's an interesting point. what if we move `body_rates_cm_s` up one level and apply that to thrusters if `posititon_ok()` is false?

Ah, good catch, and good idea. Let me play around with this a bit.

clydemcqueen:
> tested and working well on my pool. but I'm concerned about more challenging conditions for DVLs

Thanks for testing! It is good to know that the pilot experience feels good when the sensor is working. I'll work on the sensor failure case a bit.

rmackay9:
In other vehicles the way we handle loss of position estimate is to change the flight mode.  So when the EKF failsafe triggers, the vehicle can be switched to Land, AltHold, etc.  We don't expect a mode to be controlled both with and without a position sensor.

clydemcqueen:
The code now falls back to manual control (limited by PILOT_SPEED) if positioning fails. I tested this in SITL `param set SIM_GPS_DISABLE 1` and it works as expected.

The position estimates can get a little wonky just before the EKF failsafe triggers, but this is the current behavior.

All-in-all I think that this is an improvement over the current situation. LMK what you think.

ArduSub/mode_poshold.cpp
```diff
@@ -103,12 +83,51 @@ void ModePoshold::run()
...
+        position_control->input_vel_accel_xy(earth_rates_cm_s, {0, 0});
+
+        // convert pos control roll and pitch angles back to lateral and forward efforts
+        sub.translate_pos_control_rp(lateral_out, forward_out);
+
+        // udpate the xy controller
+        position_control->update_xy_controller();
+    } else {
+        // allow the pilot to reposition manually
+        forward_out = body_rates_cm_s.x / (float)g.pilot_speed;
```
rmackay9, line 126: might want to put a check in here that the pilot_speed is non-zero

rmackay9:
It's up to you guys of course.  Rover's Acro mode for skid steering vehicle also does this kind of thing where the throttle is completely manual if the vehicle doesn't have a good position estimate, otherwise it runs the forward-back speed controller.

ArduSub/mode_poshold.cpp
clydemcqueen: Done

clydemcqueen:
I tested this on a DVL-rigged ROV last week and it worked well.

The A50 is rock solid in all of the situations we dive in, so I didn't see the failsafe kick in.

LMK if you'd like to see additional changes. Thanks.

Williangalvani (approved):
lgtm!

just tested and behaves as expected. I also disabled the dvl driver while in poshold, and retained control.

PR #28211, opened 2024-09-24
Copter: Set the message buffer size to twice the message size

muramura:
The number of characters in the message exceeds 50 bytes. 
The message array is currently set to 50 bytes. I will submit a PR to change the array to 100 bytes.
ArduPilot has string splitting functionality.

AFTER
Screenshot from 2024-09-24 11-26-39

BEFORE
Screenshot from 2024-09-24 09-58-46

THIS SOURCE:
Screenshot from 2024-09-24 09-59-20

### DIFF
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -44,7 +44,7 @@ bool AP_Arming_Copter::run_pre_arm_checks(bool display_failure)
     }
 
     // always check motors
-    char failure_msg[50] {};
+    char failure_msg[100] {};
     if (!copter.motors->arming_checks(ARRAY_SIZE(failure_msg), failure_msg)) {
         check_failed(display_failure, "Motors: %s", failure_msg);
         passed = false;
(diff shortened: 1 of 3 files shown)


peterbarker (approved):
Thanks!

The message you point out has bugged me for a while!

rmackay9:
@peterbarker I guess we don't care about an extra 50 bytes of RAM being used?  The case where the message is too long, it's only long by a few bytes I think so we could just make it 60 instead of 100.

take care of the pennies and the pounds will take care of themselves

muramura:
@rmackay9 san.  
The variable that stores the text is an automatic variable. 
Therefore, this area is temporary. The stack and flash memory are different.

rmackay9:
ok, thanks.  I think there must be a cost to having the array larger than necessary

@peterbarker if you want to go ahead and merge this then I'm OK with that.

peterbarker:
> ok, thanks. I think there must be a cost to having the array larger than necessary

Not really - these are stack allocations, so they're only temporary buffers.  Our stack from the main thread is fine for this sort of thing.

The main concern is people going overboard with longer messages now!  I've stared at this particular message a few times and failed to find a way to make it shorter and as understandable.

Merging this one.

PR #28215, opened 2024-09-24
Plane: avoid calling quadplane methods if quadplane not available

peterbarker:
one block was guarded, the second not


davidbuzz (approved):
Looks sensible, passes ci.

tridge (approved):

PR #28217, opened 2024-09-24
AP_BattMonitor: Set the number of battery monitors to the supported number

muramura:
The number of battery monitors is supported up to 16. 
The default maximum value is set to 9. For many devices, the maximum value is set to the supported number. 
I will adjust to match the others.

AFTER
Screenshot from 2024-09-24 18-40-10


peterbarker:
9 is the default as it balances utility vs resources consumed.  Those extra parameters are not free!

A user can specify a larger value for this in their custom hwdef (and some people do!), but for most people 9 is already way too many!

Closing this one.

PR #28220, opened 2024-09-24
autotest: GainBackoffTakeoff test

andyp1per:


Tools/autotest/arducopter.py
```diff
@@ -10769,6 +10770,128 @@ def AHRSTrimLand(self):
...
+                    raise ValueError("I sum is zero")
+                print(f"ValidatePDZero: PID_TUNING count: {self.pid_tuning_count}")
+
+            def process(self, mav, m):
+                if m.get_type() != 'PID_TUNING' or m.axis != self.axis:
+                    return
+                self.pid_tuning_count += 1
+                self.p_sum += m.P
+                self.d_sum += m.D
+                self.i_sum += m.I
```
peterbarker, line 10816: I'd suggest dying here if things are non-zero, rather than in the `hook_removed` call.

It means the error occurs sooner, closer to what went wrong eg. when you set rc2 or when you landed, that sort of thing

Tools/autotest/arducopter.py
```diff
@@ -10769,6 +10770,128 @@ def AHRSTrimLand(self):
...
+        self.context_push()
+        self.set_parameters({
+            "ATC_LAND_R_MULT": 0.0,
+            "ATC_LAND_P_MULT": 0.0,
+            "ATC_LAND_Y_MULT": 0.0,
+            "GCS_PID_MASK" : 7,
+            "LOG_BITMASK": 180222,
+        })
+        self.reboot_sitl()
+        self.progress("Test gains are fully backed off")
```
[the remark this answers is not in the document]
andyp1per: Fixed

Tools/autotest/arducopter.py
andyp1per: This can't be done everywhere as 0 is a legitimate value some of the time, but I have done it where I can

peterbarker (approved):
LGTM

PR #28221, opened 2024-09-24
AP_GPS: Suppress unnecessary telemetry transmission

muramura:
If data cannot be obtained from the GPS device, telemetry transmission should be suppressed. 
The GCS can detect a device malfunction if it is unable to receive telemetry from the GPS. 
Since the GCS on the transmitter has a narrow bandwidth, refraining from sending meaningless telemetry data allows the bandwidth to be allocated to other telemetry.

AFTER
Screenshot from 2024-09-24 20-35-20

BEFORE
Screenshot from 2024-09-24 20-17-17


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -1369,6 +1369,10 @@ uint16_t AP_GPS::gps_yaw_cdeg(uint8_t instance) const
 
 void AP_GPS::send_mavlink_gps_raw(mavlink_channel_t chan)
 {
+    if (params[0].type == GPS_TYPE_NONE || drivers[0] == nullptr) {
```
peterbarker, line 1372: We continue to send this message to "clear" the GCS's concept of what the GPS data is.  So it might stop displaying a position on a map or something.

That's my recollection, anyway.

OK, I looked it up: https://github.com/ArduPilot/ardupilot/commit/e995a198bd71f7b5c03f1a6784be749525cc4428

This behaviour is on purpose.


I think this might be reasonable:
```suggestion
    if (params[0].type == GPS_TYPE_NONE) {
```

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -1405,7 +1409,7 @@ void AP_GPS::send_mavlink_gps_raw(mavlink_channel_t chan)
 void AP_GPS::send_mavlink_gps2_raw(mavlink_channel_t chan)
 {
     // always send the message if 2nd GPS is configured
-    if (params[1].type == GPS_TYPE_NONE) {
+    if (params[1].type == GPS_TYPE_NONE || drivers[1] == nullptr) {
```
peterbarker, line 1412:
```suggestion
    if (params[1].type == GPS_TYPE_NONE) {
```

per referenced PR, this is on purpose

muramura: I will undo it.

libraries/AP_GPS/AP_GPS.cpp
muramura: I agree with the comment.

muramura:
Screenshot from 2024-09-26 23-39-44

Screenshot from 2024-09-26 23-40-44

Screenshot from 2024-09-26 23-42-43

peterbarker:
This PR now makes sending of gps-0 and gps-1 symmetric, which is nice.

Taking this to DevCall.

peterbarker:
... oh, implications are that if you set you GPS type to zero then you will no longer get messages at all, so your position will freeze at the last message, rather than starting to receive empty messages.

tridge:
we need to check the impact of this change on MissionPlanner, mavproxy, plot.ardupilot.org and MAVExplorer

peterbarker:
Will the tools fall over if the GPS_RAW message is not present?
MissionPlanner
QGC
MAVProxy?
Mavlogdump?
plot.ardupilot.org

tridge:
what happens if GPS1_TYPE=0 and GPS2_TYPE != 0 with MissionPlanner?

peterbarker:
Ping @muramura can you test the tooling, please?

EosBandi:
Mission Planner uses the data from GLOBAL_POSITION_INT messages for displaying vehicle location, if no GLOBAL_POSITION_INT messages are received at all or it contains zero lat/lng then it uses GPS_RAW_INT for location data. GPS2_RAW messages are not used to display location.

muramura:
@peterbarker san

MISSION PLANNER recognizes GPS2.

gps2
gps2-1

EosBandi:
@muramura Indeed, however it wont use it for displaying position on the map. As described above.

muramura:
@EosBandi san

The position of GPS2 is reflected in the location data of GLOBAL_POSITION_INT.

gps2-112
gps2-113

EosBandi:
@muramura san, you are right and this is a feat of ArduPilot.
The important thing is that your proposed change does not impact Mission Planner.

muramura:
APM PLANNER2 is OK.

Screenshot from 2024-10-05 15-21-24

PR #28226, opened 2024-09-25
hwdef: correct compilation for MFT-SEMA100

peterbarker:
this board uses 2 128kB sectors for storage, meaning it is very short on space

Remove features we are likely to remove for everyone in 4.7

image

@EternAlmox - those 128kB sectors really do hurt the available flash on these boards. 

We're looking at putting a feature in to use a single page of storage for parameters, but that's problematic from both a migration POV and from a data sanctity POV (there's a race condition when "compacting" storage when it fills up).  We'll see how we go.


peterbarker:
```
Build directory: /home/pbarker/rc/ardupilot/build/MFT-SEMA100
----------------------------------------------------------------------------------------------------------
bin/arducopter   1671196      3924   258372               1675120           28804  Not Applicable         
```

andyp1per (approved):

PR #28227, opened 2024-09-25
AP_BattMonitor: add option minimum volt option

robertlong13:
Alternative approach to #28148

This solves the race condition issue and the Wh consumption calculation. Instead of calling `update_consumed` again, it sums the consumption calculation from the other monitors. This is actually much better than integrating it again manually, especially if the underlying monitors are better at calculating the consumption.

I also removed the implementation on the ESC battmon backend (and updated the param docs and comments to match). I don't think it was nearly as important to have this option for the ESC backend and I couldn't use the same trick with that one.

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor_Sum.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor_Sum.cpp
@@ -50,12 +50,16 @@ void
 AP_BattMonitor_Sum::read()
 {
     float voltage_sum = 0;
+    float voltage_min = 0;
     uint8_t voltage_count = 0;
     float current_sum = 0;
     uint8_t current_count = 0;
 
     float temperature_sum = 0.0;
     uint8_t temperature_count = 0;
+
+    float consumed_mah_sum = 0;
+    float consumed_wh_sum = 0;
 
     for (uint8_t i=0; i<_mon.num_instances(); i++) {
         if (i == _instance) {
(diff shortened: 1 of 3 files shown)


peterbarker (approved):

IamPete1:
> it sums the consumption calculation from the other monitors

Your right, that does sound better, the only possible  issue is the battery reset stuff. Just need to double check it does something sensible.

robertlong13:
@IamPete1 I have double checked that battery reset causes the consumptions to reset to zero. That's all you needed me to check, right?

libraries/AP_BattMonitor/AP_BattMonitor_Sum.cpp
```diff
@@ -86,23 +94,34 @@ AP_BattMonitor_Sum::read()
...
-        _state.current_amps = current_sum;
+        if (option_is_set(AP_BattMonitor_Params::Options::Minimum_Voltage)) {
+            _state.voltage = voltage_min;
+        } else {
+            _state.voltage = voltage_sum / voltage_count;
+        }
     }
+    _state.current_amps = current_sum;
+    _state.consumed_mah = consumed_mah_sum;
+    _state.consumed_wh = consumed_wh_sum;
```
IamPete1, line 119: This is a change in behavior if there is no current readings. Before it would hold at the old value, now it will go to zero. I think it is better to hold at the old value, particularly for the consumed readings. `current_count` could become a bool `has_current`. I think its safe to assume you have the consumed readings if you have current.

robertlong13: Good point. Changed.

IamPete1 (approved):
LGTM

PR #23838, opened 2023-05-20
Copter: Notify the reason for not switching to BRAKE mode

muramura:
There is a PR to add and try to add BRAKE mode option in Radio Fail and GCS Fail.
Operators who set this option expect to switch to BRAKE mode.
I will notify why it did not switch to the expected mode.
The operator can see and understand this notification.

PR #28231, opened 2024-09-25
Copter: Change gcs().send_text to GCS_SEND_TEXT

muramura:
This is a PR affected by #23838.
I will change it to GCS_SEND_TEXT.

### DIFF
--- a/ArduCopter/events.cpp
+++ b/ArduCopter/events.cpp
@@ -400,8 +400,9 @@ void Copter::set_mode_SmartRTL_or_land_with_pause(ModeReason reason)
         AP_Notify::events.failsafe_mode_change = 1;
         return;
     }
+    GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "SmartRTL Unavailable, Using Land Mode");
 #endif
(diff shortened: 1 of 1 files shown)


ArduCopter/events.cpp
```diff
@@ -430,9 +432,9 @@ void Copter::set_mode_auto_do_land_start_or_RTL(ModeReason reason)
         AP_Notify::events.failsafe_mode_change = 1;
         return;
     }
+    GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "AutoRTL Unavailable, Trying RTL Mode");
```
peterbarker, line 435: Can't move these inside the #endif - the idea is that the message is emitted whenever it can't enter the mode (AutoRTL here), regardless of why.  So if it fails to enter AutoRTL because there's no GPS available then we should still emit the message.

peterbarker (changes requested):

peterbarker:
@muramura still going to chase this one?

peterbarker:
No response from author; closing this one.

Issue #26749, opened 2024-04-10
GPS/Arming: dual GPS difference check may need to ignore Z-axis difference


rmackay9:
This report comes from [this Copter-4.5 discussion](https://discuss.ardupilot.org/t/incorrect-calculation-of-gps-positions-diffs-during-arming-checks/116033) where tomashmyh points out that when GPSx_TYPE = 14 (MAVLink) the driver ignores the altitude if GPS_INPUT_IGNORE_FLAG_ALT flag is set ([see code here](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_GPS/AP_GPS_MAV.cpp#L50)).

If the GPS can't provide altitude then we also shouldn't be using it in the Arming GPS consistency check.

PR #28236, opened 2024-09-26
AP_GPS: Ignore Z-axis difference while dual GPS difference check

tomashmyh:
#26749

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -1484,13 +1484,18 @@ bool AP_GPS::all_consistent(float &distance) const
 {
     // return true immediately if only one valid receiver
     if (num_instances <= 1 ||
-        drivers[0] == nullptr || params[0].type == GPS_TYPE_NONE) {
+        drivers[0] == nullptr || params[0].type == GPS_TYPE_NONE || GPS_MAX_INSTANCES <= 1) {
         distance = 0;
         return true;
     }
 
     // calculate distance
-    distance = state[0].location.get_distance_NED(state[1].location).length();
+    if (state[0].have_altitude && state[1].have_altitude) {
+        distance = state[0].location.get_distance_NED(state[1].location).length();
+    } else {
+        distance = state[0].location.get_distance_NE(state[1].location).length();
+    }
+
     // success if distance is within 50m
     return (distance < 50);
 }
--- a/libraries/AP_GPS/AP_GPS_MAV.cpp
+++ b/libraries/AP_GPS/AP_GPS_MAV.cpp
@@ -72,6 +72,7 @@ void AP_GPS_MAV::handle_msg(const mavlink_message_t &msg)
             if (have_alt) {
                 loc.alt = packet.alt * 100; // convert to centimeters
             }
+            state.have_altitude = have_alt;
             state.location = loc;
 
             if (have_hdop) {
--- a/libraries/AP_GPS/AP_GPS.h
+++ b/libraries/AP_GPS/AP_GPS.h
@@ -216,6 +216,7 @@ public:
         bool have_vertical_accuracy;      ///< does GPS give vertical position accuracy? Set to true only once available.
         bool have_gps_yaw;                ///< does GPS give yaw? Set to true only once available.
         bool have_gps_yaw_accuracy;       ///< does the GPS give a heading accuracy estimate? Set to true only once available
+        bool have_altitude;               ///< does GPS give altitude? Set to false only if GPS instance does not provide altitude.
         float undulation;                   //<height that WGS84 is above AMSL at the current location
         bool have_undulation;               ///<do we have a value for the undulation
         uint32_t last_gps_time_ms;          ///< the system time we got the last GPS timestamp, milliseconds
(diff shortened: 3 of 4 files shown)


libraries/AP_GPS/AP_GPS_MAV.cpp
```diff
@@ -71,6 +71,8 @@ void AP_GPS_MAV::handle_msg(const mavlink_message_t &msg)
             loc.lng = packet.lon;
             if (have_alt) {
                 loc.alt = packet.alt * 100; // convert to centimeters
+            } else {
+                state.have_altitude = false;
```
peterbarker, line 75: This could change over time, so we might need to set just `state.have_altitude = have_alt;`?

There might be a reason we shouldn't do that, in which case a comment would be good :-)

peterbarker (commented):
@rmackay9 why is this marked "safety"?  AFAICS this can only make it *easier* to arm, meaning it makes things less safe?

libraries/AP_GPS/AP_GPS_MAV.cpp
tomashmyh: you are right, there is no need to set it here only to false.
changed that

peterbarker:
This definitely looks like it would fix the problem mentioned in the issue.

Please outline what testing this has had.

I've marked this as DevCall for potential merge, just need to know how it's been tested.  Have you tested this on non-mavlink GPSs, for example?

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -1484,13 +1484,18 @@ bool AP_GPS::all_consistent(float &distance) const
 {
     // return true immediately if only one valid receiver
     if (num_instances <= 1 ||
-        drivers[0] == nullptr || params[0].type == GPS_TYPE_NONE) {
+        drivers[0] == nullptr || params[0].type == GPS_TYPE_NONE || GPS_MAX_INSTANCES <= 1) {
```
IamPete1, line 1487: Is this needed? We already check `num_instances`, can `num_instances` be larger than `GPS_MAX_INSTANCES`?

peterbarker: The new code being a compiler-time check, the compiler could choose to remove the code block which will never be crossed.

tomashmyh:
> This definitely looks like it would fix the problem mentioned in the issue.
> 
> Please outline what testing this has had.
> 
> I've marked this as DevCall for potential merge, just need to know how it's been tested. Have you tested this on non-mavlink GPSs, for example?

I have tested this change only with mavlink GPS.
So the test looked like this:
1. First GPS - 'physical' GPS DroneCAN type; second GPS - mavlink GPS type(this GPS data was sent via companion computer)
2. Mavlink GPS type sent data without altiutude included using GPS_INPUT (232) mavlink message
3. ARMING_CHECK=670206 - so GPS lock check was enabled
4. Verification that fix is working as expected was done by approving that with that set up copter was able to arm
I will try to find some screenshots from tests, but unfortunately cannot guarantee that will find some, because I did those tests that around a month ago)

tridge (changes requested):
what is the use case?
accepting this message could result in the EKF consuming an invalid altitude. We can't have 3D fix and no altitude

rmackay9:
Hi @tomashmyh,

Could you clarify the use-case for this?  There are a lot of places in the AP code where we check that the GPS provides 3D fix (which implies the altitude is available as well).

No precisely we're wondering if perhaps you're trying to provide a location from an external system (like a vision based system) in which case we have better ways to provide the location/position rather than using the GPS_INPUT message.  VISION_POSITION_ESTIMATE and VISION_SPEED_ESTIMATE or ODOMETRY messages might be better.

tomashmyh:
> Hi @tomashmyh,
> 
> Could you clarify the use-case for this? There are a lot of places in the AP code where we check that the GPS provides 3D fix (which implies the altitude is available as well).
> 
> No precisely we're wondering if perhaps you're trying to provide a location from an external system (like a vision based system) in which case we have better ways to provide the location/position rather than using the GPS_INPUT message. VISION_POSITION_ESTIMATE and VISION_SPEED_ESTIMATE or ODOMETRY messages might be better.

Hi @rmackay9 
I have no a special use-case for this scenario. I know that there are a much better mavlink messages for providing a location from an external system but this fix is not for those cases. The reason why I thought this fix should be there and why I did it is because I think that for now processing GPS_INPUT is not properly done in AP_GPS_MAV.cpp. User can send this message with fix_type=3(3D fix) and with set GPS_INPUT_IGNORE_FLAG_ALT in ignore_flags bitmap. In this case AP will not allow to arm a vehicle even if the main GPS source and all system in general  is healthy and vehicle is ready for use.
Also there is a possibility that user send fix_type=2(2D fix) and with GPS_INPUT_IGNORE_FLAG_ALT but the system does not allow a vehicle to be armed because for the same reason.

tomashmyh:
> what is the use case? accepting this message could result in the EKF consuming an invalid altitude. We can't have 3D fix and no altitude

but if mavlink GPS source is not the main one so EKF is not consuming its data. So in this case vehicle can not be armed and used. Also there is a possibility that user send fix_type=2(2D fix) and with GPS_INPUT_IGNORE_FLAG_ALT

PR #28219, opened 2024-09-24
DroneCAN: backport of lockup fix

tridge:
partial backport of #28218 
this backports all the changes except the mutex in UARTDriver

PR #28241, opened 2024-09-26
DroneCAN: partial application of deadlock fix

tridge:
this matches https://github.com/ArduPilot/ardupilot/pull/28219 which is what is in 4.5.7-beta1


peterbarker (approved):
LGTM

PR #28244, opened 2024-09-27
Split sending terrain report from terrain request

peterbarker:
Someone tried to set the streamrate of the reporting with set-message-interval.

That doesn't work in master, but you can get that effect by setting the itnerval on the terrain-request.

That's kind of garbage, so split the sending of reports away from requests.


peterbarker:
Tested in SITL.

At the moment the streamrate doesn't really work well for the terrain request - there are internal timing limitations in TerrainGCS which stop that working well.  We should probably remove the internal timing limits within TerrainGCS and ensure that the streamrate for requesting data from the GCS is appropriate (0.5Hz?!).  There's also the fact we'll request at arbitrary rate when we've lost our location to look at.

In any case, this moves us towards tidying the terrain-request timing stuff as the user might want to get `TERRAIN_REPORT.current_height` at a high rate.

Georacer:
I was recently playing with this and in stock SITL I was getting a `TERRAIN_REPORT` regularly at 4Hz.
I traced it and I think it is always streamed with `_EXTRA3`, contrary to the description of that parameter.

tridge (approved):

PR #28246, opened 2024-09-27
Fix mission item logging

peterbarker:
In master we use the "CMD" log message to log both items as they are uploaded *and* when one of the vehicles actually start to execute the commands.  This is weird.

This PR creates a separate log message, `MISE` which is emitted by the AP_Mission library as it starts the execution of an item.  This is a behavioural change; we now log all items, rather than just the ones that the vehicles handle.

I've tested Plane creates the messages appropriately, but need to look at the other affected vehicles.

### DIFF
--- a/Rover/mode_auto.cpp
+++ b/Rover/mode_auto.cpp
@@ -512,13 +512,6 @@ void ModeAuto::send_guided_position_target()
 /********************************************************************************/
 bool ModeAuto::start_command(const AP_Mission::Mission_Command& cmd)
 {
-#if HAL_LOGGING_ENABLED
-    // log when new commands start
-    if (rover.should_log(MASK_LOG_CMD)) {
-        rover.logger.Write_Mission_Cmd(mission, cmd);
-    }
-#endif
-
     switch (cmd.id) {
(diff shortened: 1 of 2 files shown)


rmackay9:
This seems like a good idea to me

Re "Needs Testing", I think we should check how MP handles this in the Log review tool paying particular attention to whether it makes changes on what is shown on the map.

We should also test "Create KML + gpx" which pulls out all the versions of the mission into files
image

peterbarker:
> Re "Needs Testing", I think we should check how MP handles this in the Log review tool paying particular attention to whether it changes what is shown on the map.

That view looks entirely reasonable.  I can't see any difference between/after this PR.

> We should also test "Create KML + gpx" which pulls out all the versions of the mission into files ![image](https://private-user-

That also works fine - but I think that only shows a position tack, doesn't have waypoint information in it?  I can't see any difference before/after.

I made the deliberate choice to make the mission upload retain the "CMD" message name, and make the execution message the new MISE.

tridge (approved):

Georacer:
Yes! The log review gods are please with this one!

PR #28248, opened 2024-09-27
GCS_MAVLink: Common: Allow scripting to receive messages which fail CRC check

IamPete1:
This allows scripting to receive messages which the flight controller does not understand. Note that more work is needed to allow sending. This uses https://github.com/ArduPilot/pymavlink/pull/983.

Tested new CRC check with MAVLink 1 and 2. Tested unknown messages by changing the CRC extra of command long in ArduPilot so it would fail its CRC check, it still works as normal in the script.

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -1860,14 +1860,25 @@ GCS_MAVLINK::update_receive(uint32_t max_time_us)
         bool parsed_packet = false;
 
         // Try to get a new message
-        if (mavlink_frame_char_buffer(channel_buffer(), channel_status(), c, &msg, &status) == MAVLINK_FRAMING_OK) {
+        const uint8_t framing = mavlink_frame_char_buffer(channel_buffer(), channel_status(), c, &msg, &status);
+        if (framing == MAVLINK_FRAMING_OK) {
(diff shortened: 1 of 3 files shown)


libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -1860,14 +1860,25 @@ GCS_MAVLINK::update_receive(uint32_t max_time_us)
...
             alternative.last_mavlink_ms = now_ms;
             hal.util->persistent_data.last_mavlink_msgid = 0;
+
+        }
+#if AP_SCRIPTING_ENABLED
+        else if (framing == MAVLINK_FRAMING_BAD_CRC) {
+            // This may be a valid message that we don't know the crc extra for, pass it to scripting which might
+            AP_Scripting *scripting = AP_Scripting::get_singleton();
+            if (scripting != nullptr) {
+                scripting->handle_message(msg, chan);
```
IamPete1, line 1878: Note that this ignores our usual routing rules from the `check_and_forward` function. We can't call it because we don't want to learn routes based on messages that failed the crc.

LupusTheCanine: Maybe allow the subscribe function to  pass CRC Extra value in mavlink:register_rx_msgid(msg_id)

tridge (approved):

PR #28257, opened 2024-09-28
AHRS: add option to disable EKF innovation check for airspeed

tridge:
This adds an AHRS_OPTION to disable the EKF check for innovation that disables airspeed
This is useful where the user wants to trust the airspeed sensor because the EKF innovation check may be unreliable


Georacer:
I'm happy with this change in order to tackle the specific need for the balloon drop.
It would be even better if we could orchestrate the check disabling to happen only during the drop and shortly after.

Regarding gliders, I imagine it's always better to have it enabled, but I'm not that familiar with glider logic to be sure enough.

Georacer (approved):

PR #26855, opened 2024-04-20
Global: Change division to multiplication

muramura:
The STM32 has fewer clocks for multiplication than for division.

PR #28258, opened 2024-09-28
Global: Change the division of float values to multiplication

muramura:
The STM32 is better at multiplication than division.


libraries/AP_PiccoloCAN/piccolo_protocol/ECUPackets.c
```diff
@@ -603,7 +603,7 @@ int decodeECU_TelemetrySlow1PacketStructure(const void* _pg_pkt, ECU_TelemetrySl
 
     // Input voltage in Volts
     // Range of voltage is 0.0f to 25.5f.
-    _pg_user->voltage = float32ScaledFrom1UnsignedBytes(_pg_data, &_pg_byteindex, 0.0f, 1.0f/10.0f);
+    _pg_user->voltage = float32ScaledFrom1UnsignedBytes(_pg_data, &_pg_byteindex, 0.0f, 0.1f);
```
tpwrules, line 606: Files in this directory are generated, I don't think we should change them. I think constant elimination will take care of this anyway.

tpwrules (changes requested):
This overall looks good and matches the official coding guidelines. I believe all the exponents line up too.

peterbarker:
We wil want this rebased on top of the *other* PR which does very similar things

peterbarker:
The other PR this will be rebased on: https://github.com/ArduPilot/ardupilot/pull/26855

PR #28259, opened 2024-09-28
Plane: fixed bug in pullup and loiter to alt code

tridge:
This fixes a bug in the L1 controller where reached_loiter_target() goes false after capture due to an unachievable loiter radius or wind. This can lead waypoint types that rely on a combination of reached_loiter_target() and another condition (such as being lined up with the next waypoint) to never complete

This also fixes a bug in the pullup code where poor pitch trim can mean we never hit the target airspeed in the pullup

Note that the loiter to alt fix is not really a complete fix. If the initial trajectory prevents capture ever happening then we can still have the same issue. We could add a check for yaw going through more than 360 degrees to make this more robust

Note: This builds on #28257 

@peterbarker this includes a fix for the wait min/max/accuracy code in vehicle test suite that will impact a lot of tests. We were getting things like this: 
```
2024-09-29T03:43:34.6080462Z AT-1349.0: Altitude=65.41 (want 67.000000 +- 3.000000)
2024-09-29T03:43:34.6081493Z AT-1349.1: Altitude=67.74 (got 67.000000 +- 3.000000) (maintain=0.3/10.0)
2024-09-29T03:43:34.6082665Z AT-1349.1: Altitude=69.52 (got 67.000000 +- 3.000000) (maintain=1.3/10.0)
2024-09-29T03:43:34.6083815Z AT-1349.2: Altitude=72.00 (got 67.000000 +- 3.000000) (maintain=2.4/10.0)
2024-09-29T03:43:34.6084763Z AT-1349.2: Altitude=73.28 (want 67.000000 +- 3.000000)
2024-09-29T03:43:34.6085559Z AT-1349.2: Altitude=73.33 (want 67.000000 +- 3.000000)
2024-09-29T03:43:34.6086494Z AT-1349.3: Altitude=72.86 (got 67.000000 +- 3.000000) (maintain=0.3/10.0)
2024-09-29T03:43:34.6087555Z AT-1349.3: Altitude=72.35 (got 67.000000 +- 3.000000) (maintain=1.4/10.0)
2024-09-29T03:43:34.6088620Z AT-1349.4: Altitude=72.02 (got 67.000000 +- 3.000000) (maintain=2.4/10.0)
2024-09-29T03:43:34.6089673Z AT-1349.4: Altitude=71.77 (got 67.000000 +- 3.000000) (maintain=3.5/10.0)
2024-09-29T03:43:34.6090905Z AT-1349.4: Altitude=71.69 (got 67.000000 +- 3.000000) (maintain=4.5/10.0)
2024-09-29T03:43:34.6091944Z AT-1349.5: Altitude=71.69 (got 67.000000 +- 3.000000) (maintain=5.6/10.0)
2024-09-29T03:43:34.6092998Z AT-1349.5: Altitude=71.72 (got 67.000000 +- 3.000000) (maintain=6.6/10.0)
2024-09-29T03:43:34.6094091Z AT-1349.5: Altitude=71.78 (got 67.000000 +- 3.000000) (maintain=7.7/10.0)
2024-09-29T03:43:34.6095129Z AT-1349.6: Altitude=71.81 (got 67.000000 +- 3.000000) (maintain=8.7/10.0)
2024-09-29T03:43:34.6096177Z AT-1349.6: Altitude=71.84 (got 67.000000 +- 3.000000) (maintain=9.8/10.0)
2024-09-29T03:43:34.6097082Z AT-1349.6: Attained Altitude=71.93927722772277
```
which shows it accepting an out of range value


ArduPlane/pullup.cpp
```diff
@@ -115,7 +115,7 @@ bool GliderPullup::verify_pullup(void)
     switch (stage) {
     case Stage::WAIT_AIRSPEED: {
         float aspeed;
-        if (ahrs.airspeed_estimate(aspeed) && aspeed > airspeed_start) {
+        if (ahrs.airspeed_estimate(aspeed) && (aspeed > airspeed_start || ahrs.pitch_sensor*0.01 > pitch_start)) {
```
Georacer, line 118: What will happen in the case of a release, being followed by a flat fall?
The airspeed will be low, but the pitch will pass this new test.

Will this mean that the stage will transition to WAIT_PITCH, which will then go into WAIT_LEVEL and return true with a low airspeed (much lower than expected)?

libraries/AP_L1_Control/AP_L1_Control.h
```diff
@@ -124,6 +124,10 @@ class AP_L1_Control : public AP_Navigation {
 
     AP_Float _loiter_bank_limit;
 
+    // remember reached_loiter_target decision
+    uint32_t last_reached_loiter_target_ms;
+    Location last_center_WP;
```
IamPete1, line 129: All the other class variables in L1 have the `_` prefix.

libraries/AP_L1_Control/AP_L1_Control.cpp
```diff
@@ -442,18 +443,39 @@ void AP_L1_Control::update_loiter(const Location &center_WP, float radius, int8_
...
+        /*
+          if we were previously on the circle and the target has not
+          changed then keep _WPcircle true. This prevents
+          reached_loiter_target() from going false due to a gust of
+          wind or an unachievable loiter radius
+         */
+        if (_WPcircle &&
+            last_reached_loiter_target_ms != 0 &&
+            now_ms - last_reached_loiter_target_ms < 200U &&
+            center_WP.same_loc_as(last_center_WP)) {
```
IamPete1, line 459: Should we also check the radius is the same? Going from a large radius to a smaller one would previously result in `_WPcircle` false. Now it will not. Small to large is OK because we would be inside the radius already.

tridge: I didn't add radius as it varies with height, but I've added it now using the unscaled radius from the entry to the function. Also added loiter direction

libraries/AP_L1_Control/AP_L1_Control.h
tridge: changed

ArduPlane/pullup.cpp
tridge: indeed, but this is much better than staying in WAIT_AIRSPEED forever
note that glider pullup is designed to be used with the glider being held by its tail

IamPete1 (approved):
L1 changes looks good. Pull up change looks good too and is contained to glider pull up so not a high risk. Not looked at the test changes.

peterbarker (approved):

peterbarker:
> L1 changes looks good. Pull up change looks good too and is contained to glider pull up so not a high risk. Not looked at the test changes.

Well, I've *only* looked at the test changes and they LGTM.

PR #28262, opened 2024-09-29
AP_ExternalAHRS: Correct VN100 pre-arm handling

rishabsingh3003:
The VN100 does not have a GPS, is an IMU only device. 

Users running 4.5 were able to get the sensor working without any pre-arm errors, but upgrading to master gave two pre-arm checks, one related to setting GPS properly, and other related to setting origin.  This PR makes it possible for VN100 users to go back to the 4.5 behaviour.

### DIFF
--- a/libraries/AP_ExternalAHRS/AP_ExternalAHRS.cpp
+++ b/libraries/AP_ExternalAHRS/AP_ExternalAHRS.cpp
@@ -271,7 +271,16 @@ bool AP_ExternalAHRS::pre_arm_check(char *failure_msg, uint8_t failure_msg_len) 
         }
     }
 
-    if (!state.have_origin) {
+    bool check_origin = true;
+    // don't run this check for VN100 Series since its IMU only
+    if (DevType(devtype) == DevType::VecNav) {
+        const char* name = get_name();
+        if (name && !strncmp(name, "VN-1", 4)) {
+            check_origin = false;
+        }
+    }
+
+    if (check_origin && !state.have_origin) {
         hal.util->snprintf(failure_msg, failure_msg_len, "ExternalAHRS: No origin");
 	    return false;
     }
--- a/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
+++ b/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
@@ -446,6 +446,11 @@ void AP_ExternalAHRS_VectorNav::initialize() {
         // These assumes unit is still configured at its default rate of 800hz
         run_command("VNWRG,75,3,%u,01,0721", unsigned(800 / get_rate()));
         run_command("VNWRG,76,3,16,11,0001,0106");
+
+        WITH_SEMAPHORE(state.sem);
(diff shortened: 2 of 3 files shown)


libraries/AP_ExternalAHRS/AP_ExternalAHRS.cpp
```diff
@@ -271,7 +271,16 @@ bool AP_ExternalAHRS::pre_arm_check(char *failure_msg, uint8_t failure_msg_len)
         }
     }
 
-    if (!state.have_origin) {
+    bool check_origin = true;
+    // don't run this check for VN100 Series since its IMU only
+    if (DevType(devtype) == DevType::VecNav) {
```
lashVN, line 276: Instead of this VN-specific code, would it be reasonable to have something like 

```
if (num_gps_sensors() == 0) {
check_origin = false;
}
```

I'd generally prefer to avoid VN-specific code, and avoid parsing the string. This also leaves open the possibility later of adding a param to force you into "IMU only" mode, which I'd like to leave open

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -760,6 +771,11 @@ bool AP_ExternalAHRS_VectorNav::pre_arm_check(char *failure_msg, uint8_t failure
     return true;
 }
 
+uint8_t AP_ExternalAHRS_VectorNav::num_gps_sensors(void) const
+{
+    return type == TYPE::VN_INS ? 1 + has_dual_gnss : 0;
```
lashVN, line 776: Diving into this I'm a little confused on the issue. I don't think we are publishing GPS2 data to AP inside our driver anywhere, but I have it documented [here](https://ardupilot.org/plane/docs/common-external-ahrs-vectornav.html#use-as-an-external-sensor-set) to set that param for the VN-300. Should we be publishing GPS2 data somewhere? Or should we remove that param as one to be set, and make it so num_gps_sensors is either 0 or 1?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp
```diff
@@ -459,6 +464,12 @@ void AP_ExternalAHRS_VectorNav::initialize() {
         run_command("VNWRG,75,3,%u,01,0721", unsigned(imu_rate / get_rate()));
         run_command("VNWRG,76,3,%u,31,0001,0106,0613", unsigned(imu_rate / 50));
         run_command("VNWRG,77,3,%u,49,0003,26B8,0018", unsigned(imu_rate / 5));
+
+        WITH_SEMAPHORE(state.sem);
+        set_default_sensors(uint16_t(AP_ExternalAHRS::AvailableSensor::IMU) |
```
lashVN, line 469: Apologies for my ignorance how how this method works

Will this override someone's set param? Let's say they have a VN-300 connected, but don't want to use the IMU in EKF3. Do they have to set this param on every startup, or will it notice their param isn't default and leave it untouched?

peterbarker:
Ping @rishabsingh3003 - you have feedback here

PR #28263, opened 2024-09-29
AP_Scripting: added bindings for ipv4 address

tridge:
This makes a UDP server that responds back to individual clients possible by allowing recv to know the source address of the sender. It also allows for conversion between the string and uint32_t form of IPv4 addresses

### DIFF
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -517,9 +517,19 @@ function motor_factor_table_ud:roll(index, value) end
 local SocketAPM_ud = {}
 
 -- Get a new socket
----@param datagram boolean
+---@param datagram integer -- set to 1 for UDP, 0 for TCP
 ---@return SocketAPM_ud
 function Socket(datagram) end
+
+-- return an IPv4 address given a string
+---@param str_address string -- ipv4 address as string
+---@return uint32_t_ud -- ipv4 address
+function string_to_ipv4_addr(str_address) end
+
+-- return a string representation of ipv4 address
+---@param addr uint32_t_ud|integer|number -- ipv4 address
+---@return string -- string representation of address
+function ipv4_addr_to_string(addr) end
 
 -- return true if a socket is connected
 ---@return boolean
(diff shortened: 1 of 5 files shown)


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -541,6 +551,14 @@ function SocketAPM_ud:listen(backlog) end
 ---@return integer
 function SocketAPM_ud:send(str, len) end
 
+-- send a lua string to a specified address. May contain binary data
+---@param str string
+---@param len uint32_t_ud|integer|number
+---@param ipaddr uint32_t_ud -- ipv4 address
```
IamPete1, line 557:
```suggestion
---@param ipaddr uint32_t_ud|integer|number -- ipv4 address
```
All uint32 params can do this magic casting.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -517,10 +517,20 @@ function motor_factor_table_ud:roll(index, value) end
...
 ---@return SocketAPM_ud
 function Socket(datagram) end
 
+-- return an IPv4 address given a string
+---@param str_address string -- ipv4 address as string
+---@return uint32_t_ud -- ipv4 address
+function string_to_ipv4_addr(str_address) end
+
+-- return a string representation of ipv4 address
+---@param uint32_t_ud addr -- ipv4 address
```
IamPete1, line 530:
```suggestion
---@param addr uint32_t_ud|integer|number -- ipv4 address
```
name and type were swapped. And the magic casting thing.

libraries/AP_Scripting/docs/docs.lua

libraries/AP_Scripting/docs/docs.lua

davidbuzz (approved):

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -1037,6 +1049,31 @@ int SocketAPM_accept(lua_State *L) {
     return 0;
 }
 
+/*
+  convert a uint32_t ipv4 address to a string
+ */
+int SocketAPM_ipv4_addr_to_string(lua_State *L) {
+    binding_argcheck(L, 1);
+    const uint32_t ip_addr = get_uint32(L, 1, 0, UINT32_MAX);
+    static char buf[IP4_STR_LEN];
```
tpwrules, line 1058:
```suggestion
   char buf[IP4_STR_LEN];
```
`IP4_STR_LEN` is only 16 and `static` is a footgun here IMO.

tridge: you're right, it was cut and paste from another file, fixed

PR #27779, opened 2024-08-07
Copter, AP_Mission: Send progress for individual mission items (take 2)

nexton-winjeel:
Replaces #25249

This PR:
- adds the framework in `AP_Mission` for populating and transmitting the progress fields of a new `NAV_CONTROLLER_PROGRESS` message; and
- updates Copter to send this information for a variety of mission items.

Requires: https://github.com/ArduPilot/mavlink/pull/366

PR #28264, opened 2024-09-29
Mission: Do not send notifications about messages for completed missions

muramura:
The GCS can download a MISSION from the FC before executing it. 
The MISSION contains sequence numbers. Each time the FC completes a MISSION item, it sends a MISSION_ITEM_REACHED notification. 
The GCS can obtain commands and other information from the notified sequence numbers. 
Therefore, I want to prevent the FC from sending messages.

AFTER
Screenshot from 2024-09-29 12-59-35

BEFORE
Screenshot from 2024-09-29 11-52-56

MISSION_ITEM_REACHED
Screenshot from 2024-09-29 13-01-18

### DIFF
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -2230,7 +2230,9 @@ bool ModeAuto::verify_nav_wp(const AP_Mission::Mission_Command& cmd)
             // play a tone
             AP_Notify::events.waypoint_complete = 1;
         }
+#if 0  // enable to send reached commands
         gcs().send_text(MAV_SEVERITY_INFO, "Reached command #%i",cmd.index);
+#endif
         return true;
     }
     return false;
--- a/libraries/AP_Mission/AP_Mission.cpp
+++ b/libraries/AP_Mission/AP_Mission.cpp
@@ -422,7 +422,9 @@ bool AP_Mission::start_command(const Mission_Command& cmd)
         break;
 
     default:
+#if 0  // enable to send mission commands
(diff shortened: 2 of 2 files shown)


tpwrules:
This will break RC telemetry systems capable of conveying messages such as ELRS.

andyp1per:
> This will break RC telemetry systems capable of conveying messages such as ELRS.

Yes agreed. Closing on that basis.

muramura:
@andyp1per san.
ELRS is less common than SBUS. Should we configure it to send when using ELRS?

peterbarker:
@muramura We could move to sending the same data via an actual protocol rather than text - but it seems some people are relying on the text currently.

Note also https://github.com/ArduPilot/ardupilot/pull/27779 which would be a nice replacement for the text messages, but GCS support will be years in the making!

andyp1per:
There are lots of systems that rely on the text that the flight controller sends (not just this text but lots of messages) - this PR appears to break that assumption - I can't see how that is useful. I mean maybe a config option to turn down the spam?

timtuxworth:
> @muramura We could move to sending the same data via an actual protocol rather than text - but it seems some people are relying on the text currently.
> 
> Note also #27779 which would be a nice replacement for the text messages, but GCS support will be years in the making!

Having an interface depend the structure and information contained in a text message is a very, very bad idea. These 2 PR's combined seem like a much better idea.

This is a much better idea. If we make it an option we can start now and in a few years when the GCS support is there, we will have a much better system.

timtuxworth:
> Having an interface depend the structure and information contained in a text message is a very, very bad idea. These 2 PR's combined seem like a much better idea.

Every time I read this I want to add another "very". :-)

PR #28267, opened 2024-09-30
AP_HAL_ESP32: Add GPIO driver for ESP32

sblaksono:
Add GPIO driver for ESP32, tested with onboard LED on Wemos board.


davidbuzz:
This PR currently has 6 commits and includes a new board type and a disallowed merge commit as well.
If U just wanted the GPIO driver in the PR, then it should only include that 1 commit?

PR #28271, opened 2024-09-30
Copter: add turtle mode to flight modes metadata

Hwurzburg:
mode is missing from Mission Planner since its not in the metadata

### DIFF
--- a/ArduCopter/Parameters.cpp
+++ b/ArduCopter/Parameters.cpp
@@ -249,7 +249,7 @@ const AP_Param::Info Copter::var_info[] = {
     // @Param: FLTMODE1
     // @DisplayName: Flight Mode 1
     // @Description: Flight mode when pwm of Flightmode channel(FLTMODE_CH) is <= 1230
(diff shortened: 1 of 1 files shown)


peterbarker (changes requested):
Should not be removing README.md!

peterbarker (changes requested):
LGTM

Hwurzburg:
definitely....my git fu is failing me....gone now

peterbarker (approved):

rmackay9 (approved):
That's a surprise and makes me wonder how many users have used Turtle mode

Hwurzburg:
I got back to my multis recently and tried BF turtle and liked not having to walk out to a crashed quad so much that when I converted the Kopis Holybro gave me I discovered this.....really pleased how well a default tune of this works....been flying all around the house with it....always used BF on quads due to its super non-tune default handling....was surprised how well  AP's defaults work on small quads

PR #28273, opened 2024-10-01
Tools: Do not use system packages on openSUSE

Ryanf55:
* We only tested on apt, don't blindly change dev env stuff that's not tested in CI
* The original PR was only for APT, but the dev team thought we should try it on all the scripts for consistency

Motivation: https://github.com/ArduPilot/ardupilot/commit/d18a2b22f922bf9358f39a201010167853bb756e#commitcomment-147393965


peterbarker (approved):
Thanks @Ryanf55 

@Jaaaky for visibility

tridge (changes requested):
the arch use of system packages is needed. I recently tried to help someone with arch install and without this they could not install. The build of wx from pip fails

Ryanf55:
@tridge  OK, arch will keep using the flag. Any users of OpenSUSE and arch - please help by adding CI for these. It makes maintenance much easier.

peterbarker:
Merged, thanks @Ryanf55

PR #28279, opened 2024-10-02
Allow setting of EKF3 wind state from MAV_CMD_EXTERNAL_WIND_ESTIMATE

peterbarker:
This is mostly @priseborough 's work, I've just tidied things up a bit.

 - I've ensured the wind estimates replay correctly
 - I've ensured the `WIND` mavlink message contains the wind estimate while EKF3 is in charge
 - I've checked the XKF2[0].VW* messages to contain the supplied data
 - the supplied wind estimates don't last long during a bog-standard Plane takeoff in SITL, the Plane replaces them quickly


libraries/AP_NavEKF3/AP_NavEKF3_AirDataFusion.cpp
```diff
@@ -760,3 +760,60 @@ void NavEKF3_core::FuseDragForces()
...
 
+#if EK3_FEATURE_SETWIND
+bool NavEKF3_core::setWind(float speed, float speed_accuracy, float direction, float direction_accuracy)
+{
+    // reset the corresponding covariances
+    zeroRows(P,22,23);
+    zeroCols(P,22,23);
+
+    // set the wind state variances
+    if (is_positive(speed_accuracy)) {
```
tridge, line 771: what does is_positive() do for a NaN?

libraries/AP_NavEKF3/AP_NavEKF3_AirDataFusion.cpp
```diff
@@ -760,3 +760,60 @@ void NavEKF3_core::FuseDragForces()
...
+        } else {
+            P[22][22] = P[23][23] = sq(speed_accuracy);
+        }
+    } else {
+        P[22][22] = P[23][23] = sq(WIND_SPD_UNCERTAINTY);
+    }
+
+    // reset the NE wind velocity states to the measurement
+    Vector2F wind_vel_prev = stateStruct.wind_vel;
+    stateStruct.wind_vel.x = -speed * cosF(radians(direction));
```
tridge, line 796: need to check sign convention for ekf3 wind state vs external wind state

peterbarker:
- Try passing a nan in in the autotest suite for the accuracies and see what happens
 - Check the sign is correct; set a wind estimate and a simulated speed to match, take-off and make sure estimates remain consistent

peterbarker:
image

peterbarker:
Before my most timestamp-fix patch:
image


After my most timestamp-fix patch:
image

libraries/AP_NavEKF3/AP_NavEKF3_AirDataFusion.cpp
peterbarker: > what does is_positive() do for a NaN?

```
AP: nanf is not positive
AP: nanf is not negative
AP: nanf is not zero
```

```
--- a/ArduPlane/ArduPlane.cpp
+++ b/ArduPlane/ArduPlane.cpp
@@ -325,6 +325,22 @@ extern AP_IOMCU iomcu;
 
 void Plane::one_second_loop()
 {
+    if (is_positive(NaNf)) {
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "nanf is positive");
+    } else {
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "nanf is not positive");
+    }
+    if (is_negative(NaNf)) {
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "nanf is negative");
+    } else {
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "nanf is not negative");
+    }
+    if (is_zero(NaNf)) {
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "nanf is zero");
+    } else {
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "nanf is not zero");
+    }
+
     // make it possible to change control channel ordering at runtime
     set_control_channels();
 
```

landswellsong:
Just in case, flew several times with this PR plucked over 4.5.7 (with DCM changes too) and it seems to work as intended. Didn't see any scary behaviour, can sieve through logs if there is anything in particular that can help testing.

suraqis:
Just found that this implementation is unsafe in case of using scripting binding that allows to set infinite values to wind speed. Need some basic sanity check.

landswellsong:
I've plucked this over 4.7 with some minimal changes because we sometimes use this functionality. If this PR would benefit from some real world testing data, please ping.

PR #28281, opened 2024-10-02
Tools: use /usr/bin/env to find python3

peterbarker:
going directly to /usr/bin/python3 means we don't use the venv python


Ryanf55:
Great!

PR #28282, opened 2024-10-02
SITL: FlightAxis Improvements

robertlong13:
Adds a new options parameter for the FlightAxis SITL frame:
- Option to automatically reset the aircraft position in RealFlight!!
- Move the existing framestring-passed options to option bits (but keep the existing framestring parsing for now)
- Option to silence the FPS reporting printf


amilcarlucas (changes requested):

libraries/SITL/SIM_FlightAxis.h
```diff
@@ -175,12 +174,24 @@ class FlightAxis : public Aircraft {
 
     struct sitl_input last_input;
 
+    AP_Int8 _options;
+
+    enum class Option {
+        ResetPosition = (1U<<0),
+        Rev4Servos    = (1U<<1),
+        HeliDemix     = (1U<<2),
+        SilenceFPS  = (1U<<3),
```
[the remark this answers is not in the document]
robertlong13: Yeah, accidentally messed that up after renaming. Fixed

libraries/SITL/SITL.cpp
```diff
@@ -1487,6 +1488,12 @@ const AP_Param::GroupInfo SIM::ModelParm::var_info[] = {
     AP_SUBGROUPINFO(slung_payload_sim, "SLUP_", 4, SIM::ModelParm, SlungPayloadSim),
 #endif
 
+#if AP_SIM_FLIGHTAXIS_ENABLED
+    // @Group: FTAX_
+    // @Path: ./SIM_FlightAxis.cpp
+    AP_SUBGROUPPTR(flightaxis_ptr, "FTAX_", 5, SIM::ModelParm, FlightAxis),
```
tridge, line 1494: maybe RFL_ ?

libraries/SITL/SIM_FlightAxis.h
```diff
@@ -175,12 +174,24 @@ class FlightAxis : public Aircraft {
 
     struct sitl_input last_input;
 
+    AP_Int16 _options;
```
tridge, line 177: make it an AP_Int32

tridge (changes requested):
good to merge after these changes

timtuxworth:
It makes sense to not call it FlightAxis - RealFlight stopped calling it that in (I think) Evolution.

robertlong13:
> It makes sense to not call it FlightAxis - RealFlight stopped calling it that in (I think) Evolution.

Agreed. It's an unfortunate historical holdover that we should fix in a future PR. That said, the param name we should go ahead and set correctly now.

tridge (approved):

robustini:
@robertlong13 sorry to comment here that a merge has already been done, but since even before the 4.5.7 release I have been noticing problems with Flightaxis and Realflight.
Never seen this behavior before: moving RC4 with the joystick I also have a shift of about 100 uS up and down of CH8.
Obviously on the controller panel of Realflight this thing by moving that channel does not happen, so I think it is attributable to flight axis.

robertlong13:
@robustini, it's extremely unlikely that this has anything to do with this PR, but I can probably help. We should take this conversation to Discord though.

PR #28284, opened 2024-10-02
Added functionality to prevent activating smartrtl even for disarmed …

harsh839:
…copter

added the code functionality to prevent drone for getting on smartrtl even before getting armed

### DIFF
--- a/ArduCopter/mode_smart_rtl.cpp
+++ b/ArduCopter/mode_smart_rtl.cpp
@@ -11,6 +11,13 @@
 
 bool ModeSmartRTL::init(bool ignore_checks)
 {
+    // Add check to prevent SmartRTL activation when disarmed
+    if (!motors->armed()) {
+        // Send a warning message to the GCS
+        GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "Cannot engage SmartRTL while disarmed");
+        return false;  // Prevent initialization
+    }
+
     if (g2.smart_rtl.is_active()) {
         // initialise waypoint and spline controller
         wp_nav->wp_and_spline_init();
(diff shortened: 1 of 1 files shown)


ArduCopter/mode_smart_rtl.cpp
```diff
@@ -11,6 +11,13 @@
 
 bool ModeSmartRTL::init(bool ignore_checks)
 {
+    // Add check to prevent SmartRTL activation when disarmed
+    if (!motors->armed()) {
```
rmackay9, line 15: In copter we always allow the user to switch to the mode if the vehicle is disarmed.  We do not allow users to arm in SmartRTL though so users are forced to be in a different mode when it comes time to arm.

ArduCopter/mode_smart_rtl.cpp
```diff
@@ -174,8 +181,12 @@ void ModeSmartRTL::pre_land_position_run()
 // save current position for use by the smart_rtl flight mode
 void ModeSmartRTL::save_position()
 {
-    const bool should_save_position = motors->armed() && (copter.flightmode->mode_number() != Mode::Number::SMART_RTL);
+    // Prevent saving positions when the drone is disarmed
+    if (!motors->armed()) {
```
rmackay9, line 185: We've already got a "should_save_position" below so I don't think we need this extra armed check.

I suspect the issue is not that a point is being saved but that home is being saved while the vehicle is disarmed.  This means that if the user moves the vehicle after arming SmartRTL may move to the original position.  Could you confirm this is the issue you're seeing?

rmackay9:
BTW Thanks very much for this!

ArduCopter/mode_smart_rtl.cpp
harsh839: Yes this is the issue i was seeing so that it should not go to smartrtl mode before arming the copter that's why i added that thanks for reviewing!

rmackay9: Hi @harsh839,

OK thanks for that.  I think a better solution then would be to modify the AP_SmartRTL::update() function so that it takes a "save_home" argument that is set to true if the vehicle is disarmed.

rmackay9: Actually after looking at the code a bit more, I see that when the vehicle arms it calls copter.g2.smart_rtl.set_home(copter.position_ok()) so there shouldn't be any need to do what I've mentioned above.

I'm sorry @harsh839, I don't see the problem.  Are you sure that you've reproduced a problem and shown that this PR fixes it?  Could you please share some logs or screen shots showing the before and after?

harsh839: ohk i will be posting here just check my another PR which fixes #17419

PR #28305, opened 2024-10-04
AP_HAL_ESP32: ADC ported to new idf 5.x driver and debugged

arg7:
AP_HAL_ESP32: ADC ported to new idf 5.x driver and debugged


davidbuzz (approved):
Reported in discord as being tested on both classic esp32 and s3

PR #28306, opened 2024-10-04
AP_RCProtocol: IBUS hack for FlySky IA6 receiver

arg7:
This PR adds support for slightly modified IBUS protocol  for cheap FlySky IA6 receiver.
See https://www.youtube.com/watch?v=qEUB6GJQmRc

### DIFF
--- a/libraries/AP_RCProtocol/AP_RCProtocol_IBUS.cpp
+++ b/libraries/AP_RCProtocol/AP_RCProtocol_IBUS.cpp
@@ -20,10 +20,12 @@
 
 #include "AP_RCProtocol_IBUS.h"
 
+static int ia6_hack = 0;
+
 // decode a full IBUS frame
-bool AP_RCProtocol_IBUS::ibus_decode(const uint8_t frame[IBUS_FRAME_SIZE], uint16_t *values, bool *ibus_failsafe)
+bool AP_RCProtocol_IBUS::ibus_decode_std(const uint8_t frame[IBUS_FRAME_SIZE], uint16_t *values, bool *ibus_failsafe)
 {
-    uint32_t chksum = 96;
+    uint16_t chksum = 96;
 
     /* check frame boundary markers to avoid out-of-sync cases */
     if ((frame[0] != 0x20) || (frame[1] != 0x40)) {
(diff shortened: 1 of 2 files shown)


davidbuzz (approved):

libraries/AP_RCProtocol/AP_RCProtocol_IBUS.cpp
```diff
@@ -69,23 +110,32 @@ void AP_RCProtocol_IBUS::_process_byte(uint32_t timestamp_us, uint8_t b)
     const bool have_frame_gap = (timestamp_us - byte_input.last_byte_us >= 2000U);
     byte_input.last_byte_us = timestamp_us;
 
+
```
peterbarker, line 113: Bad whitespace change

libraries/AP_RCProtocol/AP_RCProtocol_IBUS.cpp
```diff
@@ -69,23 +110,32 @@ void AP_RCProtocol_IBUS::_process_byte(uint32_t timestamp_us, uint8_t b)
...
         return;
     }
+
     if (byte_input.ofs == 0 && !have_frame_gap) {
         // must have a frame gap before the start of a new IBUS frame
         return;
     }
 
+    if (byte_input.ofs == 0)
+    {
```
peterbarker, line 131:
```suggestion
    if (byte_input.ofs == 0) {
```

libraries/AP_RCProtocol/AP_RCProtocol_IBUS.cpp
```diff
@@ -20,10 +20,12 @@
 
 #include "AP_RCProtocol_IBUS.h"
 
+static int ia6_hack = 0;
```
peterbarker, line 23: Better to have an `AP_RCProtocol_IBUS_IA6` subclass so that other users can use this modified decoder by using the custom build server.

libraries/AP_RCProtocol/AP_RCProtocol_IBUS.cpp
```diff
@@ -20,10 +20,12 @@
...
+
 // decode a full IBUS frame
-bool AP_RCProtocol_IBUS::ibus_decode(const uint8_t frame[IBUS_FRAME_SIZE], uint16_t *values, bool *ibus_failsafe)
+bool AP_RCProtocol_IBUS::ibus_decode_std(const uint8_t frame[IBUS_FRAME_SIZE], uint16_t *values, bool *ibus_failsafe)
 {
-    uint32_t chksum = 96;
+    uint16_t chksum = 96;
```
peterbarker, line 28: Are you sure this is correct?

ie. the checksum definition *must* state that your result is modulo UINT16_MAX.  If it does not then you may pass the checksum when you shouldn't.

libraries/AP_RCProtocol/AP_RCProtocol_IBUS.cpp
```diff
@@ -69,23 +110,32 @@ void AP_RCProtocol_IBUS::_process_byte(uint32_t timestamp_us, uint8_t b)
...
+
     if (have_frame_gap) {
         // if we have a frame gap then this must be the start of a new
         // frame
         byte_input.ofs = 0;
     }
-    if (b != 0x20 && byte_input.ofs == 0) {
-        // definately not IBUS, missing header byte
+
+    if ( !( (b == 0x20) || (b == 0x55) ) && byte_input.ofs == 0) {
```
peterbarker, line 120: Will need a virtual method to return expected header byte

libraries/AP_RCProtocol/AP_RCProtocol_IBUS.cpp
```diff
@@ -69,23 +110,32 @@ void AP_RCProtocol_IBUS::_process_byte(uint32_t timestamp_us, uint8_t b)
...
 
+    if (byte_input.ofs == 0)
+    {
+        ia6_hack = (b == 0x55)?1:0;
+    }
+  
     byte_input.buf[byte_input.ofs++] = b;
 
-    if (byte_input.ofs == sizeof(byte_input.buf)) {
+    if (byte_input.ofs == (sizeof(byte_input.buf)-ia6_hack)) // IA6 has one byte less
```
peterbarker, line 137: Will need a callback for expected number of bytes in a packet

peterbarker (changes requested):
Let me know if you want help structuring this as a subclass.

We also need to augment the test suite in RCProtocolTest to add a regression test for this protocol - have a look for `ibus_bytes` in there to see what's required.

PR #28308, opened 2024-10-04
AP_DroneCAN: properly convert timeout to deadline for aux frames

tpwrules:
The timeout specified for auxiliary driver frames was passed to the driver where a deadline was expected. The transmission was then started after its "deadline", thereby causing it to be canceled and the data lost if the frame could not be sent immediately.

Fix by converting the timeout to a deadline before passing to the driver. The conversion is done in the Canard interface code as it already does other conversions from timeouts to deadlines.

Tested that my setup used to get timeouts registered in `@SYS/can0_stats.txt` and doesn't anymore.


magicrub (commented):
This looks like a correct change to me but lets make sure @bugobliterator or @tridge takes a peek

tridge (approved):

PR #28309, opened 2024-10-04
Cope with negative ESC rpms in notch filter

andyp1per:


RickReeser:
@andyp1per The test flight that caused me to open the issue was done on Copter 4.5.6. Master has a different behavior: the logged filter frequency is negative for notches that are tracking negative RPM. I haven't read through the code enough yet to know what the filter is actually doing in this case. This is what it looks like on master:
image

With your changes:
image

andyp1per:
It doesn't look quite right, does it. We have made some effort to properly log what values are actually being used.

andyp1per:
Actually, maybe it does look right. I'm just a little thrown by the weird symmetry of the ESC values - I wonder if there is something else going on there. Unless this is just split yaw.

RickReeser:
Yeah I think it is fine. The symmetry is because the drone was yawing at the time (uncommanded; bad tune because this was default tuning params).
image

Here are the logs from the above screenshots.
[negative rpm test flights.zip](https://github.com/user-attachments/files/17267598/negative.rpm.test.flights.zip)

andyp1per:
great!

IamPete1:
My preference would be to just apply the abs in the ESC RPM method. Just so that -1 for a static notch still results in off rather than 1hz. -1 makes no sense for many of the sources.

andyp1per:
IDK, this feels safer preventing negative numbers ever reaching the notches

tridge (approved):

tridge:
@IamPete1 i'm OK with this change, but will delay merge for you to expand on your comment

tridge:
discussed on the EU dev call, decided this is good for now

PR #28310, opened 2024-10-04
Fix heli ice/turbine cooldown and add autotest 

MattKear:
This PR fixes the ice/turbine cool down feature in Heli RSC.

This also adds an autotest to prevent future regression.

### DIFF
--- a/Tools/autotest/helicopter.py
+++ b/Tools/autotest/helicopter.py
@@ -15,6 +15,7 @@ from pymavlink import mavutil
 from pysim import vehicleinfo
 
 import copy
+import operator
 
 
 class AutoTestHelicopter(AutoTestCopter):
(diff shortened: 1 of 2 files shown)


bnsgeyer (approved):
LGTM

Tools/autotest/helicopter.py
```diff
@@ -697,6 +698,52 @@ def check_airspeeds(mav, m):
...
+        # set option for Turbine
+        RAMP_TIME = 4
+        SETPOINT = 66
+        IDLE = 15
+        COOLDOWN_TIME = 5
+        self.set_parameter("RC6_OPTION", 161)
+        self.set_parameter("H_RSC_RAMP_TIME", RAMP_TIME)
+        self.set_parameter("H_RSC_SETPOINT", SETPOINT)
+        self.set_parameter("H_RSC_IDLE", IDLE)
+        self.set_parameter("H_RSC_CLDWN_TIME", COOLDOWN_TIME)
```
peterbarker, line 712: `set_parameters` can be faster.

Tools/autotest/helicopter.py
```diff
@@ -697,6 +698,52 @@ def check_airspeeds(mav, m):
...
+        self.set_rc(3, 1000)
+        self.set_rc(8, 1000)
+
+        self.progress("Starting turbine")
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
+
+        self.set_rc(6, 2000)
+        self.wait_statustext('Turbine startup')
```
peterbarker, line 721: There's a race condition here, should collect statustext in the context and `check_context=True`

peterbarker (approved):
I don't think this will flap, despite the races.

PR #28316, opened 2024-10-05
AC_AttitudeControl: Add function to limit target attitude thrust angle

lthall:
This PR provides a feature for use in plane to limit the maximum target angle.

When the target angle is above a given value the current attitude is used to determine the minimum rotation vector back to within the allowable lean angle using only roll and pitch.

### DIFF
--- a/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
+++ b/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
@@ -275,6 +275,79 @@ void AC_AttitudeControl::landed_gain_reduction(bool landed)
     set_angle_P_scale_mult(scale_mult);
 }
 
+// limit_target_thrust_angle - 
+bool AC_AttitudeControl::limit_target_thrust_angle(float thrust_angle_max_cd)
+{
+    float thrust_angle_max_rad = radians(thrust_angle_max_cd * 0.01f);
+
+    // attitude_body represents a quaternion rotation in NED frame to the body
+    Quaternion attitude_body;
+    _ahrs.get_quat_body_to_ned(attitude_body);
+    
+    Vector3f gyro = _ahrs.get_gyro_latest();
+
+    // angle_delta is the estimated rotation that the aircraft will experience during the correction
+    Vector3f angle_delta = Vector3f{gyro.x / get_angle_roll_p().kP(), gyro.y / get_angle_pitch_p().kP(), gyro.z / get_angle_yaw_p().kP() };
+
+    // attitude_update represents a quaternion rotation representing the expected rotation in the next time_constant
+    Quaternion attitude_update;
+    attitude_update.from_axis_angle(angle_delta);
+    attitude_body *= attitude_update;
+
+    // The direction of thrust is [0,0,-1] is any body-fixed frame, inc. body frame and target frame.
+    const Vector3f thrust_vector_up{0.0f, 0.0f, -1.0f};
+    
+    // Rotating [0,0,-1] by attitude_body expresses (gets a view of) the thrust vector in the inertial frame
+    Vector3f thrust_vec = attitude_body * thrust_vector_up;
+
+    // the dot product is used to calculate the current lean angle
+    float thrust_correction_angle = acosf(constrain_float(thrust_vec * thrust_vector_up, -1.0f, 1.0f));
+
+    if (abs(thrust_correction_angle) <= thrust_angle_max_rad) {
+        return false;
+    }
+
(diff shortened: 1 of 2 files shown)


libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -275,6 +275,71 @@ void AC_AttitudeControl::landed_gain_reduction(bool landed)
...
+
+    // attitude_body represents a quaternion rotation in NED frame to the body
+    Quaternion attitude_body;
+    _ahrs.get_quat_body_to_ned(attitude_body);
+    
+    Vector3f gyro = _ahrs.get_gyro_latest();
+
+    // attitude_update represents a quaternion rotation representing the expected rotation in the next time_constant
+    Quaternion attitude_update;
+    float time_constant = MAX(1.0 / get_angle_roll_p().kP(), 1.0 / get_angle_pitch_p().kP() );
```
IamPete1, line 291: You could use roll and pitch time constant for there respective axis.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -275,6 +275,71 @@ void AC_AttitudeControl::landed_gain_reduction(bool landed)
...
+// limit_target_thrust_angle - 
+void AC_AttitudeControl::limit_target_thrust_angle(float thrust_angle_max_cd)
+{
+    float thrust_angle_max_rad = radians(thrust_angle_max_cd * 0.01f);
+
+    // attitude_body represents a quaternion rotation in NED frame to the body
+    Quaternion attitude_body;
+    _ahrs.get_quat_body_to_ned(attitude_body);
+    
+    Vector3f gyro = _ahrs.get_gyro_latest();
```
IamPete1, line 287: Should we be using the new `_rate_gyro` variable that was recently added?

lthall: I believe so.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
lthall: I agree, I made this change.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -275,6 +275,78 @@ void AC_AttitudeControl::landed_gain_reduction(bool landed)
...
+
+    // the dot product is used to calculate the current lean angle
+    float thrust_correction_angle = acosf(constrain_float(thrust_vec * thrust_vector_up, -1.0f, 1.0f));
+
+    if (abs(thrust_correction_angle) <= thrust_angle_max_rad) {
+        return false;
+    }
+
+    // the cross product of the current thrust vector and vertical thrust vector defines the rotation vector
+    Vector3f thrust_vec_cross = thrust_vec % thrust_vector_up;
```
priseborough, line 311: Can you conform that this vector is the target rotation vector to get back to upright?

priseborough: OK, I figured it out. The subsequent usage is consistent with that definition.

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -275,6 +275,78 @@ void AC_AttitudeControl::landed_gain_reduction(bool landed)
...
+
+    // subtract the maximum lean angle from the current thrust angle
+    thrust_correction_angle -= constrain_float(thrust_correction_angle, -thrust_angle_max_rad, thrust_angle_max_rad);
+
+    // calculate the closest _attitude_target using roll and pitch only, with a thrust angle of thrust_angle_max_rad
+    Quaternion attitude_target_offset;
+    attitude_target_offset.from_axis_angle(thrust_vec_cross, thrust_correction_angle);
+    _attitude_target = attitude_target_offset * attitude_body;
+
+    // Set angular velocity targets when further than 1.25 * thrust_angle_max_cd
```
priseborough, line 337: Based on my reading of this code this would be more accurately described as:

    // Angular velocity targets must be zeroed at large thrust angles to prevent them from preventing efficient recovery
    // so zero angular velocity targets when tilted more than 1.25 * thrust_angle_max_cd
    // Interpolate between an unmodified ang_vel_target at thrust_angle_max and zero at 1.25 * thrust_angle_max
    // This ensures small angles past thrust_angle_max_cd do not zero angular velocity targets and prevent full recovery

lthall:
I have pushed the comment fix.

PR #28317, opened 2024-10-05
Copter: Move the class variables to the class definition

muramura:
The class variables for the mode are defined in a separate file. 
I will move them to align with the other modes

### DIFF
--- a/ArduCopter/mode_guided.cpp
+++ b/ArduCopter/mode_guided.cpp
@@ -2,36 +2,6 @@
 
 #if MODE_GUIDED_ENABLED
 
-/*
- * Init and run calls for guided flight mode
- */
-
-static Vector3p guided_pos_target_cm;       // position target (used by posvel controller only)
-bool guided_pos_terrain_alt;                // true if guided_pos_target_cm.z is an alt above terrain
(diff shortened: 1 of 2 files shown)


peterbarker (commented):
This is a good change.

What testing has this had?

muramura:
@peterbarker san

In SITL, by specifying a destination in GUIDED mode, I was able to confirm that the vehicle moves correctly.

Screenshot from 2024-10-06 15-22-12

Screenshot from 2024-10-06 15-22-56

peterbarker (approved):
LGTM

ArduCopter/mode_guided.cpp
```diff
@@ -2,36 +2,6 @@
 
 #if MODE_GUIDED_ENABLED
 
-/*
- * Init and run calls for guided flight mode
- */
-
-static Vector3p guided_pos_target_cm;       // position target (used by posvel controller only)
```
IamPete1, line 9: Lots of these things need to be static. for guided to work as part of auto for example.

peterbarker: Could you explain your reasoning, please?

AFAICS these are only used in `ModeGuided` methods, and we will only ever have one of those instantiated and there are no static methods bared by ModeGuided.

PR #28318, opened 2024-10-05
AP_BLHeli:correct RVMASK metadata

Hwurzburg:
NFC....obviously a typo...assume it should refer to the only other BLH mask types


andyp1per (changes requested):
Actually the sentence should be removed. It's just 3DMASK this applies to

Hwurzburg:
@andyp1per done

andyp1per (approved):

tridge (approved):

PR #28320, opened 2024-10-05
Plane: added fast attitude recovery for inverted flight in Q modes

tridge:
This makes recovery to level a lot faster from an upset attitude when using a Q mode as a recovery mode. It resets the target thrust angle to be within the limits set for the vehicle, which prevents the input shaping from trying to keep the vehicle inverted

builds on https://github.com/ArduPilot/ardupilot/pull/28316 
thanks to @lthall 

With GriffinPro in RealFlight
image
without this change:
image

### DIFF
--- a/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
+++ b/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
@@ -275,6 +275,78 @@ void AC_AttitudeControl::landed_gain_reduction(bool landed)
     set_angle_P_scale_mult(scale_mult);
 }
 
+// limit_target_thrust_angle - 
+bool AC_AttitudeControl::limit_target_thrust_angle(float thrust_angle_max_cd)
+{
+    float thrust_angle_max_rad = radians(thrust_angle_max_cd * 0.01f);
+
+    // attitude_body represents a quaternion rotation in NED frame to the body
+    Quaternion attitude_body;
+    _ahrs.get_quat_body_to_ned(attitude_body);
+    
+    Vector3f gyro = _ahrs.get_gyro_latest();
+
+    // angle_delta is the estimated rotation that the aircraft will experience during the correction
+    Vector3f angle_delta = Vector3f{gyro.x / get_angle_roll_p().kP(), gyro.y / get_angle_pitch_p().kP(), gyro.z / get_angle_yaw_p().kP() };
+
+    // attitude_update represents a quaternion rotation representing the expected rotation in the next time_constant
+    Quaternion attitude_update;
+    attitude_update.from_axis_angle(angle_delta);
+    attitude_body *= attitude_update;
+
+    // The direction of thrust is [0,0,-1] is any body-fixed frame, inc. body frame and target frame.
+    const Vector3f thrust_vector_up{0.0f, 0.0f, -1.0f};
+    
+    // Rotating [0,0,-1] by attitude_body expresses (gets a view of) the thrust vector in the inertial frame
+    Vector3f thrust_vec = attitude_body * thrust_vector_up;
+
+    // the dot product is used to calculate the current lean angle
+    float thrust_correction_angle = acosf(constrain_float(thrust_vec * thrust_vector_up, -1.0f, 1.0f));
+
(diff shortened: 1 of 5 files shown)


tridge:
@IamPete1 @priseborough @lthall I have a chat to Leonard about this, and he suggested we run this function all the time, not just when starting VTOL control. 
I also hit an interesting issue where requesting a smaller angle actually led to a lot more altitude loss in recovery, which seemed counter intuitive to me. The reason seems to be that with a small angle passed to this function we end up pulling the nose down more during the recovery due to the demanded positive pitch rate in body frame when inverted. That leads to downward lift from the wings and a sharp drop in altitude. Example here:
http://uav.tridgell.net/tmp/att_recovery/
rec1.BIN has a small target angle, rec2.BIN a large target angle. rec1.BIN loses 29m, rec2.bin loses 14m

priseborough:
The algorithm appears to be working as expected looking at the initialisation of the set-point in the log.

peterbarker (approved):
Nice autotest.

tridge:
this needs more discussion

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -275,6 +275,78 @@ void AC_AttitudeControl::landed_gain_reduction(bool landed)
...
+    float thrust_angle_max_rad = radians(thrust_angle_max_cd * 0.01f);
+
+    // attitude_body represents a quaternion rotation in NED frame to the body
+    Quaternion attitude_body;
+    _ahrs.get_quat_body_to_ned(attitude_body);
+    
+    Vector3f gyro = _ahrs.get_gyro_latest();
+
+    // angle_delta is the estimated rotation that the aircraft will experience during the correction
+    Vector3f angle_delta = Vector3f{gyro.x / get_angle_roll_p().kP(), gyro.y / get_angle_pitch_p().kP(), gyro.z / get_angle_yaw_p().kP() };
```
peterbarker, line 290: Dont' really want to assign from the object created on the stack

peterbarker: ... and could be const?

peterbarker, line 290:
```suggestion
    const Vector3f angle_delta{
        gyro.x / get_angle_roll_p().kP(),
        gyro.y / get_angle_pitch_p().kP(),
        gyro.z / get_angle_yaw_p().kP()
    };
```

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -275,6 +275,78 @@ void AC_AttitudeControl::landed_gain_reduction(bool landed)
...
+    // attitude_update represents a quaternion rotation representing the expected rotation in the next time_constant
+    Quaternion attitude_update;
+    attitude_update.from_axis_angle(angle_delta);
+    attitude_body *= attitude_update;
+
+    // The direction of thrust is [0,0,-1] is any body-fixed frame, inc. body frame and target frame.
+    const Vector3f thrust_vector_up{0.0f, 0.0f, -1.0f};
+    
+    // Rotating [0,0,-1] by attitude_body expresses (gets a view of) the thrust vector in the inertial frame
+    Vector3f thrust_vec = attitude_body * thrust_vector_up;
```
peterbarker, line 301:
```suggestion
    const Vector3f thrust_vec {attitude_body * thrust_vector_up};
```

peterbarker: ... similarly elsewhere

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
priseborough: @peterbarker I want to fix this but don't understand your comment about assignment from an object created on the stack.

priseborough:
I've rebased, made changes to review comments submitted a new PR here https://github.com/ArduPilot/ardupilot/pull/29480

tridge:
attempting to re-open

tridge:
attempting to re-open

PR #28323, opened 2024-10-06
Copter: Initialize the stack variable

muramura:
This value is used in the input_vel_accel_xy method. 
VEL is a stack variable. 
Initialize the value.


peterbarker:
Vectors are implicitly zeroed to save flash.

https://github.com/ardupilot/ardupilot/blob/master/libraries/AP_Math/vector2.h#L50

PR #28326, opened 2024-10-06
hwdef: GEPRCF745BTHD: remove parachute and bl-flashing support (flash…

peterbarker:
… overflow)

This board has only been in 3 months, I think it's OK to remove parachute here.


andyp1per (changes requested):
I really hate removing bl support, isn't there anything else we can take out? This is really a copter board - remove some plane features as I am sure that is what is overflowing.

peterbarker:
> I really hate removing bl support, isn't there anything else we can take out? This is really a copter board - remove some plane features as I am sure that is what is overflowing.

This is not removing bootloader support.  We can't do that in ArduPilot yet.

This is removing the ability to flash the bootloader - you have to use boot0 and the button to update the bootloader using the combined hex thing.

The bootloader itself works normally.

andyp1per:
Ok not so bad I suppose, although it feels like there should be other options that could be removed instead

PR #27799, opened 2024-08-09
AP_GPS: Backport correct satellite count for SBF DNU NrSv value

chiara-septentrio:
Backport to Copter 4.5 of  [PR 26628 ](https://github.com/ArduPilot/ardupilot/pull/26628): it sets the satellite count to 0 rather than ignoring it

PR #28328, opened 2024-10-07
Copter, Rover, Tracker, Plane-4.5.7 release

rmackay9:
This is the Copter, Rover, Tracker, Plane-4.5.7 release.  The contents of that release can be found in the [4.5.7-beta1 column of this project](https://github.com/orgs/ArduPilot/projects/8/views/8?layout=board).

This is the same as the 4.5.7-beta1 release but with PR https://github.com/ArduPilot/ardupilot/pull/27799 reverted.  I have not actually determined if [the user report](https://discuss.ardupilot.org/t/coper-4-5-7-beta1-available-for-testing/124355) is correct but I think at this stage it is best to remove it and proceed with the release.

### DIFF
--- a/AntennaTracker/ReleaseNotes.txt
+++ b/AntennaTracker/ReleaseNotes.txt
@@ -1,5 +1,11 @@
 Antenna Tracker Release Notes:
 ------------------------------------------------------------------
+Release 4.5.7 08 Oct 2024
+
+Changes from 4.5.7-beta1
+
+1) Reverted Septentrio GPS sat count correctly drops to zero when 255 received
+------------------------------------------------------------------
 Release 4.5.7-beta1 26 Sep 2024
 
 Changes from 4.5.6
--- a/AntennaTracker/version.h
+++ b/AntennaTracker/version.h
@@ -6,14 +6,14 @@
 
 #include "ap_version.h"
 
-#define THISFIRMWARE "AntennaTracker V4.5.7-beta1"
+#define THISFIRMWARE "AntennaTracker V4.5.7"
 
 // the following line is parsed by the autotest scripts
-#define FIRMWARE_VERSION 4,5,7,FIRMWARE_VERSION_TYPE_BETA
+#define FIRMWARE_VERSION 4,5,7,FIRMWARE_VERSION_TYPE_OFFICIAL
 
 #define FW_MAJOR 4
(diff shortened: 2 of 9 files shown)


tridge (approved):

SeptentrioGNSS:
@rmackay9 could you explain why the DNU value of SVnum commit into AP_GPS was reverted?

SeptentrioGNSS:
Hi @tridge, I am still waiting on an explanation on why this has been reverted back from #27799, could you help on this please? 

The case this was trying to solve was when (under specific circumstance) the reported value turns from a numeric (e.g 25) to the 255 value from one message to the next. With the != approach, the previous value will stay in play and users will not have the info that the Septentrio GNSS receiver is not tracking any satellites. With the approach from @flyingthingsintothings and @chiara-septentrio they solved this as the NrSV value will be fixed to 0. 

With the users report on the forum (mentioned by @rmackay9 in the description here) the discussion goes towards the [HDOP](https://gssc.esa.int/navipedia/index.php?title=Positioning_Error#:~:text=Horizontal%20Dilution%20of%20Precision) value (which is inherently different from NrSV and comes from a completely different [SBF message](https://www.septentrio.com/en/support/product-resources)) it is set to 'zero' when it is not available. Otherwise you have to divide by 100 to get the actual value (also this division I do not see in the code upon parsing the message). 

To give an example, here are consecutive PVTGeodetic blocks with the NrSV values dropping to 255 while the antenna is disconnected (rare use case but still plausible and should be indicated to user other than keeping the latest 'normal value') from the Septentrio GNSS receiver.

nrsv_septentrio

rmackay9:
Hi @SeptentrioGNSS,

We reverted the change because we couldn't determine quickly whether the issue was a real problem or not.  We didn't want to hold up other fixes so we reverted the one questionable change and proceeded with the release.

The change is included in [4.6 which is undergoing beta testing now](https://discuss.ardupilot.org/t/copter-4-6-0-beta2-is-available-for-beta-testing/127445).  It's hard to exactly predict when 4.6 will be released as the stable version but I think within a month is possible.

We could try to include the change again if we do another 4.5.x release as well.  There's not definitive plans to do that at the moment though.

As AP has become more stable and mature, we struggle to find enough beta testers.  Any and all feedback during beta testing is greatly appreciated!

PR #28333, opened 2024-10-07
board_types.txt: reservie ID for F4 2-3S 20A AIO FC V1

jonasniesner:
I want to add full support for this board:
https://betafpv.com/products/f4-2-3s-20a-aio-fc-v1 
I have all other files ready but want to reserve a board id first


peterbarker:
Mreged it!

Issue #28301, opened 2024-10-03
AP_DDS: Provide vehicle type and mode


tizianofiorenzani:
## DDS Interface to obtain vehicle's type and mode.

The DDS interface should provide information about the vehicle type and current mode.

**Solution**
- Create a new topic `/ap/flight_mode` of custom type `FlightMode`.
- `FlightMode `should contain:
  - `vehicle_type`: enum of the vehicle, plane, copter,....
  - `flight_mode`: enum of the current flight mode
  - `armed`: true if armed.
  - `header`
    - `frame_id`: string identifying the vehicle like SYSID_THISMAV.

**Platform**
[ x ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

Ryanf55:
Had you considered the reporting designed in REP-147?
https://ros.org/reps/rep-0147.html#flight-modes

tizianofiorenzani:
Thanks for the link. I find the proposed `FlightMode` message a a little too reductive, as it limits interaction with only Guided/Offline and does not provide any actual detail of the underlying mode. For example the DDS interface provides input for joystick, which could be used to control a airplane in FBWB (copter guided is way more advanced).

We could use the `VehicleStatus` type for what I called the `vehicle_type`, use the proposed `FlightMode` as a simplified mode, but I'd also find it useful to provide the actual flight mode, the same as we provide to a Mavlink interface.

timtuxworth:
frame_id seems confusing - in AP this might mean the frame_type/frame_class (e.g. if a VTOL is a X/H/Y frame and 4/6/8 motors etc) - whereas SYSID_THISMAV is not that.

Ryanf55:
> frame_id seems confusing - in AP this might mean the frame_type/frame_class (e.g. if a VTOL is a X/H/Y frame and 4/6/8 motors etc) - whereas SYSID_THISMAV is not that. Also SYSID_THISMAV is a number, not a string.

frame_id is supposed to correspond to a coordinate frame, not a vehicle number. Each vehicle has it's own `base_link` of the "body" frame in AP terminology, and then attached sensors such as GPS, camera, rangefinder also have their own frames relative to that. I wouldn't worry about populating the frame ID with a value to differentiate which vehicle. Differentiating vehicles for swarming type operations is a separate issue here: https://github.com/ArduPilot/ardupilot/issues/23567

[...]

PR #28337, opened 2024-10-07
AP_DDS: Status Message

tizianofiorenzani:
- Issue: https://github.com/ArduPilot/ardupilot/issues/28301
- Creates a new `/ap/status` message
- Publishes the topic only when the status changes.
- The status reports a flag `external_control`, which will be filled by or after https://github.com/ArduPilot/ardupilot/pull/28429.

**Test**
- Launch the SITL.
- Open a new terminal and echo the new topic.
- Verify that only one message is published.
- Now change the vehicle's mode and verify that a new message has been published.
- Arm and takeoff, verify that a new message has been published only when `armed` flag changes.

Test a failsafe:
- Change the parameter `BATT_LOW_VOLT` to 11.0
- Set the parameter `SIM_BATT_VOLTAGE` to 10.6.
- Verify that the status topics reports a failsafe of type 22
- Disable the RC setting `SIM_RC_FAIL` to 1.
- Verify that the failsafe array now reports 22 and 21

image


libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -42,6 +43,7 @@ static constexpr uint16_t DELAY_GEO_POSE_TOPIC_MS = 33;
 static constexpr uint16_t DELAY_CLOCK_TOPIC_MS = 10;
 static constexpr uint16_t DELAY_GPS_GLOBAL_ORIGIN_TOPIC_MS = 1000;
 static constexpr uint16_t DELAY_PING_MS = 500;
+static constexpr uint16_t DELAY_STATUS_TOPIC_MS = 500;
```
Ryanf55, line 46: Can we make this event driven instead rather than poll?

Ryanf55:
Clean commits before merge.

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -525,6 +527,50 @@ void AP_DDS_Client::update_topic(geographic_msgs_msg_GeoPointStamped& msg)
...
+bool AP_DDS_Client::update_topic(ardupilot_msgs_msg_Status& msg)
+{
+    // Fill the new message.
+    msg.vehicle_type = static_cast<uint8_t>(AP::fwversion().vehicle_type);
+    msg.armed = AP_Notify::flags.armed;
+    msg.mode = AP::vehicle()->get_mode();
+    msg.flying = AP_Notify::flags.flying;
+    uint8_t fs_iter = 0;
+    msg.failsafe_size = 0;
+    if ( AP_Notify::flags.failsafe_radio ) {
```
rmackay9, line 539: nitpick: little whitespace within the brackets

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -525,6 +527,50 @@ void AP_DDS_Client::update_topic(geographic_msgs_msg_GeoPointStamped& msg)
     }
 }
 
+bool AP_DDS_Client::update_topic(ardupilot_msgs_msg_Status& msg)
+{
+    // Fill the new message.
+    msg.vehicle_type = static_cast<uint8_t>(AP::fwversion().vehicle_type);
+    msg.armed = AP_Notify::flags.armed;
```
tridge, line 534: should be hal.util->get_soft_armed()

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -525,6 +527,50 @@ void AP_DDS_Client::update_topic(geographic_msgs_msg_GeoPointStamped& msg)
...
+    msg.mode = AP::vehicle()->get_mode();
+    msg.flying = AP_Notify::flags.flying;
```
tridge, line 536: should be vehicle->get_likely_flying();

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -525,6 +527,50 @@ void AP_DDS_Client::update_topic(geographic_msgs_msg_GeoPointStamped& msg)
     }
 }
 
+bool AP_DDS_Client::update_topic(ardupilot_msgs_msg_Status& msg)
+{
+    // Fill the new message.
+    msg.vehicle_type = static_cast<uint8_t>(AP::fwversion().vehicle_type);
+    msg.armed = AP_Notify::flags.armed;
+    msg.mode = AP::vehicle()->get_mode();
```
tridge, line 535: should do const auto &vehicle = AP::vehicle();

tridge (changes requested):

libraries/AP_DDS/Idl/ardupilot_msgs/msg/Status.idl
```diff
@@ -0,0 +1,52 @@
...
+      const uint8 APM_UNKNOWN = 5;
+      const uint8 APM_REPLAY = 6;
+      const uint8 APM_ARDUSUB = 7;
+      const uint8 APM_IOFIRMWARE = 8;
+      const uint8 APM_AP_PERIPH = 9;
+      const uint8 APM_AP_DAL_STANDALONE = 10;
+      const uint8 APM_AP_BOOTLOADER = 11;
+      const uint8 APM_BLIMP = 12;
+      const uint8 APM_HELI = 13;
+      const uint8 FS_RADIO = 1;
```
tridge, line 23: should this be a separate enum? seems odd to fix FS and APM_xxx

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -525,6 +527,50 @@ void AP_DDS_Client::update_topic(geographic_msgs_msg_GeoPointStamped& msg)
...
+    msg.flying = AP_Notify::flags.flying;
+    uint8_t fs_iter = 0;
+    msg.failsafe_size = 0;
+    if ( AP_Notify::flags.failsafe_radio ) {
+        msg.failsafe[fs_iter++] = FS_RADIO;
+    }
+    if ( AP_Notify::flags.failsafe_battery ) {
```
Ryanf55, line 542: Use this from AP_BattMon

rmackay9, line 542: I wonder if we could get the battery failsafe state from the battery library.

For RC, EKF and GCS failsafe we probably need to do what you've done here

libraries/AP_DDS/Idl/ardupilot_msgs/msg/Status.idl
tizianofiorenzani: this is auto generated, I could set the value to start from 14.

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: Are there events I can use to drive this?

Ryanf55: There are not. ArduPilot does not have any such architecture. The best approach is to poll the state quickly and only publish on change.

tizianofiorenzani:
I applied all the suggestions and tested in SITL again. Here is when a Low battery has been triggered:

image

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: set to 100ms

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: Whitespaces removed

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: set as suggested

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: set as suggested

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: set as suggested

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: set as suggested

libraries/AP_DDS/Idl/ardupilot_msgs/msg/Status.idl
tizianofiorenzani: I set the FS enums starting from 21

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: I don't have an opinion about it, maybe we could port that limitation to a new issue?

Tools/ros2/ardupilot_msgs/msg/Status.msg
```diff
@@ -0,0 +1,27 @@
...
+uint8 APM_BLIMP = 12
+uint8 APM_HELI = 13
+uint8 vehicle_type # From AP_Vehicle_Type.h
+
+bool armed # true if vehicle is armed.
+uint8 mode # Vehicle mode, enum depending on vehicle type.
+bool flying # True if flying/driving/diving/tracking.
+bool external_control # True is external control is enabled.
+
+uint8 FS_RADIO = 21
```
Ryanf55, line 23: If it's supposed to match something, mind adding a comment?

tridge (approved):

rmackay9 (approved):

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -604,6 +608,56 @@ void AP_DDS_Client::update_topic(geographic_msgs_msg_GeoPointStamped& msg)
...
+    const auto &vehicle = AP::vehicle();
+    const auto &battery = AP::battery();
+    msg.vehicle_type = static_cast<uint8_t>(AP::fwversion().vehicle_type);
+    msg.armed = hal.util->get_soft_armed();
+    msg.mode = vehicle->get_mode();
+    msg.flying = vehicle->get_likely_flying();
+    msg.external_control = true; // Always true for now. To be filled after PR#28429.
+    uint8_t fs_iter = 0;
+    msg.failsafe_size = 0;
+    if (AP_Notify::flags.failsafe_radio) {
```
Ryanf55, line 624: Can we add another comment that "TODO move from AP_Notify to vehicle functions"

Ryanf55: `virtual bool in_rc_failsafe() const { return true; }`

Ryanf55: Check the radio failsafe using this call rather than AP_Notify:
https://github.com/ArduPilot/ardupilot/blob/9f29606d1c915f0dd64a06babf628c623311783a/libraries/RC_Channel/RC_Channel.h#L517

The devs do not want us using notify unless there is no other option.

PR #28177, opened 2024-09-21
AP_DDS: Wrap all topics in ifdefs

Ryanf55:
# purpose

* Give ability to enable/disable any topic in DDS through compile options
* Follow up of https://github.com/ArduPilot/ardupilot/pull/28098 but for the remaining topics

# Details

In the header file, I grouped together related data, which made some functions private that were public. 
This reduces the amount of ifdefs. For the topics that share the same datatype, I added an `AP_DDS_NEEDS*` to signify whether we need to include that file.

PR #28341, opened 2024-10-08
AP_DDS: Make GPS DDS features depend on GPS

Ryanf55:
# Purpose

Make the DDS features that depend on GPS rely on `AP_GPS_ENABLED`. 

# Details

Follow up to https://github.com/ArduPilot/ardupilot/pull/28177 where we hard-coded navsatfix enable flag to `1`. 
If you disabled GPS on `master` now and enabled DDS, you would get a compile error,

### DIFF
--- a/libraries/AP_DDS/AP_DDS_Client.cpp
+++ b/libraries/AP_DDS/AP_DDS_Client.cpp
@@ -4,7 +4,9 @@
 #if AP_DDS_ENABLED
 #include <uxr/client/util/ping.h>
 
+#if AP_DDS_NEEDS_GPS
 #include <AP_GPS/AP_GPS.h>
+#endif // AP_DDS_NEEDS_GPS
 #include <AP_HAL/AP_HAL.h>
 #include <RC_Channel/RC_Channel.h>
 #include <AP_RTC/AP_RTC.h>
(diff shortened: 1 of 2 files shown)


peterbarker:
... this one definitely needs to compile :-)

Not really sure what you're doing with that "NEEDS" variable, either...

Ryanf55:
> ... this one definitely needs to compile :-)
> 
> Not really sure what you're doing with that "NEEDS" variable, either...

`NEEDS` are a convenience to only have a single define to decide whether a header is needed. 

Before:

```c++
#if AP_DDS_NAVSATFIX_PUB_ENABLED || AP_DDS_STATIC_TF_PUB_ENABLED // And potentially more!
#include <AP_GPS/AP_GPS.h>
#endif
```

After:
```c++
#if AP_DDS_NEEDS_GPS
#include <AP_GPS/AP_GPS.h>
#endif
```

tridge (approved):

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -3,8 +3,11 @@
 #include "AP_DDS_config.h"
 #if AP_DDS_ENABLED
 #include <uxr/client/util/ping.h>
+#include <AP_GPS/AP_GPS_config.h>
```
peterbarker, line 6:
```suggestion
```

tridge, line 6: not needed

Ryanf55, line 6:
```suggestion
```

peterbarker (changes requested):

peterbarker (approved):

PR #28342, opened 2024-10-08
AP_DroneCAN: fixed using 2 serial ports with one node

tridge:
we need to match node ID and serial port index, while allowing for wildcard of -1


tridge:
incorrect fix, closing

PR #28346, opened 2024-10-08
Plane: evaluate assistance requirements on mode change

peterbarker:
this avoid the AHRS being told we are flying forward - because we are no longer in a VTOL mode - and instantly being told we are not flying forward - because we are providing assistance

The following two plots are plotting the fly-forward flag being shoved into the Replay log.

Before:
image

After:
image

This is particularly important when an external wind estimate has been supplied to the vehicle, as setting the FF flag causes EKF3 to completely nuke the wind estimates.  The following graphs are from a different test where the wind estimate is set externally:

Before:
image

After:
image

... the EKF wind estimate is still suffering when the FF flag is finally set after FF is *correctly* set - but at least we have transitioned with the flag set!

### DIFF
--- a/ArduPlane/mode.cpp
+++ b/ArduPlane/mode.cpp
@@ -140,6 +140,14 @@ bool Mode::enter()
 
         // Make sure the flight stage is correct for the new mode
         plane.update_flight_stage();
+
+#if HAL_QUADPLANE_ENABLED
+        if (quadplane.enabled()) {
+            float aspeed;
+            bool have_airspeed = quadplane.ahrs.airspeed_estimate(aspeed);
+            quadplane.assisted_flight = quadplane.assist.should_assist(aspeed, have_airspeed);
+        }
+#endif
     }
 
     return enter_result;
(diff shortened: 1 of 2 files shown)


priseborough (approved):

ArduPlane/mode.cpp
```diff
@@ -140,6 +140,12 @@ bool Mode::enter()
 
         // Make sure the flight stage is correct for the new mode
         plane.update_flight_stage();
+
+#if HAL_QUADPLANE_ENABLED
+        float aspeed;
+        bool have_airspeed = quadplane.ahrs.airspeed_estimate(aspeed);
```
tridge, line 146: should have quadplane.enabled() check

tridge (changes requested):
thanks Peter!

ArduPlane/mode.cpp
peterbarker: Added the check.

tridge (approved):

Issue #20754, opened 2022-05-13
Plane Auto TakeOff : Idle Throttle for Hand/Bungee/Catapult launch


romarysonrier:
## Feature request

**ESC may delay startup when using folding propeler or heavy propeller**
If the AUW is near the max supported by a airframe, delayed spinning will cause crash

**My proposal is to implement a param TKOFF_IDLE_MODE with 3 values possibles:**
MODE=0 throttle when idle is equal to 0%
MODE=1 throttle when idle is equals to param TKOFF_IDLE_THR (static value, in % of motor PWM) . The motor should start to spin when RC MOTOR INPUT >1500 for safety reason.
MODE=2 throttle equals RC MOTOR INPUT (eg channel 3), limited by the upper value define by TKOFF_IDLE_THR (in % of motor PWM). My prefered mode.

So basicaly, MODE=0 is the current behaviour, MODE=1 is the INAV feature, MODE=2 allow the pilot to take manual control with its RC on the motor startup, with a predefined “security” limit .

To sumup my main feature request is to have a “balanced folding propeller” BEFORE launch time… and the only known way is to have a 10-30% throttle when the plane is IDLE. INAV have this feature for this specific prupose..

**Known Alternative : FBWA**
_But FBWA launch requires at least 2 people to be securly executed., And it is not as convenient as AUTO TakeOFF._

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[X] Plane
[  ] Rover
[  ] Submarine

**Additional context**
Take-off is kind of **very risky** with that kind of frame using arduplane Auto-Takeoff because if motor is not spinning before launching the plane (Idle Throttle is equal to 0% by default), there is a high risk that the ESC has to retry its “startup procedure” due to the torque required at startup-time ( if propeler is folded or unbalanced) . causing the propeler to have a 1-2 sec delay before really spinning… leading to some crashes… especially with wings that have fewer attitude authority than t-tail plane at low speed.

What’s-more if the propeller itself is heavy or just big, the torque required to speed-up the motor will generate some roll at launch-time, what is very unwanted at takeoff, causing loss of lift… leading to some crashes…

[INAV has a parameter for that issue](https://inavfixedwinggroup.com/guides/tuning/how-to-setup-nav-launch-correctly/#6-nav_fw_launch_idle_thr) : nav_fw_launch_idle_thr
Here is the associated documentation for folding propeler :

[...]

PR #28351, opened 2024-10-08
Implement TKOFF_THR_IDLE

Georacer:
I thought I'd tackle https://github.com/ArduPilot/ardupilot/issues/20754.

A new parameter `TKOFF_THR_IDLE` has been added and a new option bit 1 to `TKOFF_OPTIONS` to enable it (off by default).

In TAKEOFF and AUTO modes, while in `FlightStage::TAKEOFF` and while the throttle is suppressed, `TKOFF_THR_IDLE` is output to the throttle.

Here's how it looks like in a catapult takeoff, for `TKOFF_THR_IDLE=20`:
image

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -158,6 +158,15 @@ const AP_Param::Info Plane::var_info[] = {
     // @User: Standard
     ASCALAR(takeoff_throttle_min,       "TKOFF_THR_MIN",    0),
 
+    // @Param: TKOFF_THR_IDLE
+    // @DisplayName: Takeoff idle throttle
+    // @Description: The idle throttle to hold after arming and before a takeoff. Applicable in TAKEOFF and AUTO modes.
+    // @Units: %
+    // @Range: 0 100
+    // @Increment: 1
+    // @User: Standard
+    ASCALAR(takeoff_throttle_idle,       "TKOFF_THR_IDLE",    0),
+
     // @Param: TKOFF_OPTIONS
     // @DisplayName: Takeoff options
     // @Description: This selects the mode of the takeoff in AUTO and TAKEOFF flight modes. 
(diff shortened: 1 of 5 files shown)


ArduPlane/Parameters.cpp
```diff
@@ -158,10 +158,20 @@ const AP_Param::Info Plane::var_info[] = {
...
+    // @Range: 0 100
+    // @Increment: 1
+    // @User: Standard
+    ASCALAR(takeoff_throttle_idle,       "TKOFF_THR_IDLE",    0),
+
     // @Param: TKOFF_OPTIONS
     // @DisplayName: Takeoff options
     // @Description: This selects the mode of the takeoff in AUTO and TAKEOFF flight modes. 
     // @Bitmask: 0: When unset the maximum allowed throttle is always used (THR_MAX or TKOFF_THR_MAX) during takeoff. When set TECS is allowed to operate between a minimum (THR_MIN or TKOFF_THR_MIN) and a maximum (THR_MAX or TKOFF_THR_MAX) limit. Applicable only when using an airspeed sensor.
+    // @Bitmask: 1: When set the throttle will run at TKOFF_THR_IDLE before takeoff.
```
tridge, line 174: don't need the option bit, just use when non-zero

tridge (changes requested):

ArduPlane/servos.cpp
```diff
@@ -619,6 +619,11 @@ void Plane::set_throttle(void)
             // throttle is suppressed (above) to zero in final flare in auto mode, but we allow instead thr_min if user prefers, eg turbines:
             SRV_Channels::set_output_scaled(SRV_Channel::k_throttle, aparm.throttle_min.get());
 
+        } else if ((flight_stage == AP_FixedWing::FlightStage::TAKEOFF)
+                    && !(aparm.takeoff_throttle_idle == 0)
```
tridge, line 623: is_positive(aparm.takeoff_throttle_idle.get())

tridge (changes requested):

Georacer:
Closes #20754 .

tridge (approved):

rtm7777:
I see this PR in 4.7.0 release notes, did you test it on a plane? I've tried to implement something similar and I had to change much more things and have no time to finish and properly test

Georacer:
It has been tested in simulation. I'm not aware of real-life testing.
We discovered last week that it works only on TAKEOFF mode, not on AUTO, and [the fix](https://github.com/ArduPilot/ardupilot/pull/33784) will likely make its way into 4.7.1.

PR #28353, opened 2024-10-08
GCS_MAVLink: Support requesting MESSAGE_INTERVAL via REQUEST_MESSAGE cmd

nexton-winjeel:
Updates the `MAV_CMD_REQUEST_MESSAGE` handler with a special case for requesting the `MESSAGE_INTERVAL` message. The old method of using `MAV_CMD_GET_MESSAGE_INTERVAL` is deprecated.

Raised in this Discord message: https://discord.com/channels/674039678562861068/728017546313466047/1293342302261219329

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -3200,6 +3200,12 @@ uint8_t GCS::get_channel_from_port_number(uint8_t port_num)
 MAV_RESULT GCS_MAVLINK::handle_command_request_message(const mavlink_command_int_t &packet)
 {
     const uint32_t mavlink_id = (uint32_t)packet.param1;
+
+    if (mavlink_id == MAVLINK_MSG_ID_MESSAGE_INTERVAL) {
+        const mavlink_command_int_t msg_interval_cmd = { .param1 = packet.param2, };
+        return handle_command_get_message_interval(msg_interval_cmd);
+    }
+
     const ap_message id = mavlink_id_to_ap_message_id(mavlink_id);
     if (id == MSG_LAST) {
         return MAV_RESULT_FAILED;
--- a/Tools/autotest/rover.py
+++ b/Tools/autotest/rover.py
@@ -6884,6 +6884,32 @@ Brakes have negligible effect (with=%0.2fm without=%0.2fm delta=%0.2fm)
         self.context_pop()
         self.reboot_sitl()
 
+    def GetMessageInterval(self):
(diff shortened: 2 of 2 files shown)


nexton-winjeel:
FYI: @peterbarker

peterbarker (changes requested):
Please create an autotest for this.

This won't be the final shape of this functionality - it should be inverted, with handle_command_get_message_interval implemented in terms of handle_command_request_message and eventually going away.

The method you're modifying will need to be a lot more involved to support the "instance" parameter generally - but it will be nice to be able to request a specific battery instance, for example.

... and per-instance message rates are going to be "interesting".

I'm happy with this implementation, it just needs a regression test - which is going to be maybe 3 lines in rover.py or something.

nexton-winjeel:
>Please create an autotest for this.

Added an autotest to Rover.

peterbarker (commented):
Test is good.

Just needs to pass Python cleanliness.  You can run `flake8` on the command-line against the file.  I run it as part of my git commit hooks to avoid pushing stuff which doesn't work.  And recently using fly-check.

nexton-winjeel:
@peterbarker: Can I get you to review this again?

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3195,6 +3195,12 @@ uint8_t GCS::get_channel_from_port_number(uint8_t port_num)
 MAV_RESULT GCS_MAVLINK::handle_command_request_message(const mavlink_command_int_t &packet)
 {
     const uint32_t mavlink_id = (uint32_t)packet.param1;
+
+    if (mavlink_id == MAVLINK_MSG_ID_MESSAGE_INTERVAL) {
+        const mavlink_command_int_t msg_interval_cmd = { packet.param2, };
```
peterbarker, line 3200:
```suggestion
        const mavlink_command_int_t msg_interval_cmd { command : packet.param2, };
```

Sorry, but I don't think we should rely on the magic packing stuff that mavlink does in its messages to do this!

peterbarker (changes requested):

libraries/GCS_MAVLink/GCS_Common.cpp
nexton-winjeel: Updated.

peterbarker:
Can be merged once conflict is fixed.

nexton-winjeel:
> Can be merged once conflict is fixed.

@peterbarker: Rebased.

peterbarker:
Merged, thanks @nexton-winjeel

PR #28354, opened 2024-10-09
Copter: stop using global ap variable as bitmask 

peterbarker:
Table edited for new iterate-over-cast-structure scheme:
```
CubeOrange-periph-heavy  *                                                                     
CubeRedPrimary                      *      *           -448    -408              *      *      *
Durandal                            *      *           -272    -128              *      *      *
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     *      *           -424    -360              *      *      *
MatekF405                           *      *           -376    -352              *      *      *
Pixhawk1-1M-bdshot                  *                  -416    -360              *      *      *
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           *      *           -384    -352              *      *      *
skyviper-v2450                                         -464                                    
```

Tested in SITL by generating a simple log-disarmed log and looking at the message, and using `du32_change.py` to see bits are changing appropriately

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -688,12 +688,29 @@ void Copter::three_hz_loop()
     low_alt_avoidance();
 }
 
+// ap_value calculates a 32-bit bitmask representing various pieces of
+// state about the Copter.  It replaces a global variable which was
+// used to track this state.
+uint32_t Copter::ap_value() const
+{
+    uint32_t ret = 0;
+
+    const bool *b = (const bool *)&ap;
+    for (uint8_t i=0; i<sizeof(ap); i++) {
+        if (b[i]) {
+            ret |= 1U<<i;
+        }
+    }
+
+    return ret;
+}
+
 // one_hz_loop - runs at 1Hz
 void Copter::one_hz_loop()
 {
 #if HAL_LOGGING_ENABLED
     if (should_log(MASK_LOG_ANY)) {
-        Log_Write_Data(LogDataID::AP_STATE, ap.value);
+        Log_Write_Data(LogDataID::AP_STATE, ap_value());
     }
 #endif
 
--- a/ArduCopter/Copter.h
+++ b/ArduCopter/Copter.h
@@ -347,45 +347,47 @@ private:
 # include USERHOOK_VARIABLES
 #endif
 
-    // Documentation of GLobals:
-    typedef union {
-        struct {
-            uint8_t unused1                 : 1; // 0
-            uint8_t unused_was_simple_mode  : 2; // 1,2
-            uint8_t pre_arm_rc_check        : 1; // 3       // true if rc input pre-arm checks have been completed successfully
-            uint8_t pre_arm_check           : 1; // 4       // true if all pre-arm checks (rc, accel calibration, gps lock) have been performed
-            uint8_t auto_armed              : 1; // 5       // stops auto missions from beginning until throttle is raised
-            uint8_t logging_started         : 1; // 6       // true if logging has started
-            uint8_t land_complete           : 1; // 7       // true if we have detected a landing
-            uint8_t new_radio_frame         : 1; // 8       // Set true if we have new PWM data to act on from the Radio
-            uint8_t usb_connected_unused    : 1; // 9       // UNUSED
-            uint8_t rc_receiver_present_unused : 1; // 10      // UNUSED
-            uint8_t compass_mot             : 1; // 11      // true if we are currently performing compassmot calibration
-            uint8_t motor_test              : 1; // 12      // true if we are currently performing the motors test
-            uint8_t initialised             : 1; // 13      // true once the init_ardupilot function has completed.  Extended status to GCS is not sent until this completes
-            uint8_t land_complete_maybe     : 1; // 14      // true if we may have landed (less strict version of land_complete)
-            uint8_t throttle_zero           : 1; // 15      // true if the throttle stick is at zero, debounced, determines if pilot intends shut-down when not using motor interlock
-            uint8_t system_time_set_unused  : 1; // 16      // true if the system time has been set from the GPS
-            uint8_t gps_glitching           : 1; // 17      // true if GPS glitching is affecting navigation accuracy
-            uint8_t using_interlock         : 1; // 20      // aux switch motor interlock function is in use
-            uint8_t land_repo_active        : 1; // 21      // true if the pilot is overriding the landing position
-            uint8_t motor_interlock_switch  : 1; // 22      // true if pilot is requesting motor interlock enable
-            uint8_t in_arming_delay         : 1; // 23      // true while we are armed but waiting to spin motors
-            uint8_t initialised_params      : 1; // 24      // true when the all parameters have been initialised. we cannot send parameters to the GCS until this is done
-            uint8_t unused3                 : 1; // 25      // was compass_init_location; true when the compass's initial location has been set
-            uint8_t unused2                 : 1; // 26      // aux switch rc_override is allowed
-            uint8_t armed_with_airmode_switch : 1; // 27      // we armed using a arming switch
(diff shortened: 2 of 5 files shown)


ArduCopter/Copter.h
```diff
@@ -347,46 +347,38 @@ class Copter : public AP_Vehicle {
...
+    // ap_value calculates a 32-bit bitmask representing various pieces of
+    // state about the Copter.  It replaces a global variable which was
+    // used to track this state.
+    uint32_t ap_value() const;
 
-    ap_t ap;
+    // These variables are essentially global variables.  These should
+    // be removed over time.
+    struct {
+        uint8_t pre_arm_rc_check;               // true if rc input pre-arm checks have been completed successfully
```
tridge, line 358: why make these uint8_t instead of bool?

ArduCopter/Copter.cpp
```diff
@@ -688,12 +688,73 @@ void Copter::three_hz_loop()
...
+        IN_ARMING_DELAY = (1U << 23),           // true while we are armed but waiting to spin motors
+        INITIALISED_PARAMS = (1U << 24),        // true when the all parameters have been initialised. we cannot send parameters to the GCS until this is done
+        // unused3 = (1U << 25),                // was compass_init_location; true when the compass's initial location has been set
+        // unused2 = (1U << 26),                // aux switch rc_override is allowed
+        ARMED_WITH_AIRMODE_SWITCH = (1U << 27), // we armed using a arming switch
+        PREC_LAND_ACTIVE = (1U << 28),          // true if precland is active
+    };
+
+    uint32_t ret = 0;
+
```
tridge, line 726:
```
uint32_t bmask = 0;
const bool *b = (const uint8_t *)&ap;
for (uint16_t i=0; i<sizeof(ap)) {
 if (b[i]) {
  bmask |= 1U<<i;
 }
}
```

nexton-winjeel: This doesn't give the same result: there are gaps in the bitmask (e.g. the first 3 bits are unused).

peterbarker: > This doesn't give the same result: there are gaps in the bitmask (e.g. the first 3 bits are unused).

Yep, sorry, we didn't capture all of the fun on the call last night - tridge made it abundantly clear he was only sketching :-)

I did actually implement it while on the call last night, but discovered the more serious issue that our *existing* implementation is flawed - if you have a look at the current bitmask commnets you will note there's a gap between 17 and 20.  I need to do some archeology to work out what the correct fix there is.

I've force-pushed this branch (complete with the debug which showed this problem - I was expecting iteration up to 28 but only got to 26)

tpwrules: This is cute but is totally undefined behavior.

peterbarker: > This is cute but is totally undefined behavior.

Oh?  How so?  Pretty sure we need to `PACK` this structure, but past that - do you see other undefined behaviour?

tpwrules: Explained e.g. [here](https://wiki.sei.cmu.edu/confluence/display/c/ARR37-C.+Do+not+add+or+subtract+an+integer+to+a+pointer+to+a+non-array+object) . 

Each member is a separate object, which is an array of length 1, and simply generating a pointer more than 1 past the end of an array is undefined. `PACK` would make it clearer, and it's probably okay in practice, and other parts of the code use a similar pattern, but it's still undefined.

tpwrules: A more C++ solution would be to have methods that return references so you could do e.g. `ap.motor_interlock_switch() = true`.

peterbarker: > Explained e.g. [here](https://wiki.sei.cmu.edu/confluence/display/c/ARR37-C.+Do+not+add+or+subtract+an+integer+to+a+pointer+to+a+non-array+object) .
> 
> Each member is a separate object, which is an array of length 1, and simply generating a pointer more than 1 past the end of an array is undefined. `PACK` would make it clearer, and it's probably okay in practice, and other parts of the code use a similar pattern, but it's still undefined.

I'm pretty sure the packed attribute makes it just fine and defined.  As you say, if it doesn't then the hundreds of places we checksum by iterating across uint8_t-*-cast structures is going to be a problem :-)

tpwrules: It's about the pointer arithmetic, not the packing. And I don't like those places either ;)

A union would be another idea I think.

ArduCopter/Copter.h
peterbarker: Fixed

ArduCopter/Copter.cpp
peterbarker: New structure is in place.

peterbarker:
For a simple up-and-down flight, setting log_disarmed first:

```
pbarker@fx:~/rc/ardupilot(pr/ap-bitmask-stop-it)$ ./Tools/scripts/du32_change.py logs/00000033.BIN 
1728616069: Creating connection
Original armed_with_airmode_switch: 0
Original auto_armed: 0
Original compass_mot: 0
Original gps_glitching: 0
Original in_arming_delay: 0
Original initialised: 1
Original initialised_params: 1
Original land_complete: 1
Original land_complete_maybe: 1
Original land_repo_active: 0
Original logging_started: 0
Original motor_interlock_switch: 0
Original motor_test: 0
Original new_radio_frame: 1
Original pre_arm_check: 1
Original pre_arm_rc_check: 1
Original prec_land_active: 0
Original rc_receiver_present_unused: 0
Original system_time_set_unused: 0
Original throttle_zero: 1
Original unused1: 0
Original unused2: 0
Original unused3: 0
Original unused_was_simple_mode bit1: 0
Original unused_was_simple_mode bit2: 0
Original usb_connected_unused: 0
Original using_interlock: 0
2024-10-11 14:06:46.23:  +in_arming_delay
2024-10-11 14:06:48.23:  -in_arming_delay
2024-10-11 14:06:49.23:  +auto_armed -throttle_zero
2024-10-11 14:06:50.23:  -land_complete -land_complete_maybe
2024-10-11 14:07:14.23:  +land_complete_maybe
2024-10-11 14:07:15.23:  +land_complete
2024-10-11 14:07:16.23:  -auto_armed
pbarker@fx:~/rc/ardupilot(pr/ap-bitmask-stop-it)$ 
```

peterbarker:
... and the same thing but arming the vehicle with an airmode switch (aux function 154)
```
pbarker@fx:~/rc/ardupilot(pr/ap-bitmask-stop-it)$ ./Tools/scripts/du32_change.py logs/00000034.BIN 
1728616318: Creating connection
Original armed_with_airmode_switch: 1
Original auto_armed: 1
Original compass_mot: 0
Original gps_glitching: 0
Original in_arming_delay: 1
Original initialised: 1
Original initialised_params: 1
Original land_complete: 1
Original land_complete_maybe: 1
Original land_repo_active: 0
Original logging_started: 0
Original motor_interlock_switch: 0
Original motor_test: 0
Original new_radio_frame: 1
Original pre_arm_check: 1
Original pre_arm_rc_check: 1
Original prec_land_active: 0
Original rc_receiver_present_unused: 0
Original system_time_set_unused: 0
Original throttle_zero: 0
Original unused1: 0
Original unused2: 0
Original unused3: 0
Original unused_was_simple_mode bit1: 0
Original unused_was_simple_mode bit2: 0
Original usb_connected_unused: 0
Original using_interlock: 0
2024-10-11 14:11:28.25:  -in_arming_delay
2024-10-11 14:11:31.25:  -auto_armed
2024-10-11 14:11:44.25:  +auto_armed +in_arming_delay
2024-10-11 14:11:46.25:  -in_arming_delay
2024-10-11 14:11:47.25:  -auto_armed
```

... so I think the offsets are currently correct.

tridge (approved):

rmackay9 (approved):

PR #28355, opened 2024-10-09
Copter: remove global ap.new_radio_frame state 

peterbarker:
we have better means to determine if you have valid input

Rebase/push after https://github.com/ArduPilot/ardupilot/pull/28354 is merged - in the meantime only the most recent patch is interesting

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -772,13 +772,15 @@ void Copter::update_simple_mode(void)
 {
     float rollx, pitchx;
 
-    // exit immediately if no new radio frame or not in simple mode
-    if (simple_mode == SimpleMode::NONE || !ap.new_radio_frame) {
+    // exit immediately if not in simple mode:
+    if (simple_mode == SimpleMode::NONE) {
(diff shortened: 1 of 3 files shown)


ArduCopter/Copter.cpp
```diff
@@ -772,13 +772,15 @@ void Copter::update_simple_mode(void)
...
-    if (simple_mode == SimpleMode::NONE || !ap.new_radio_frame) {
+    // exit immediately if not in simple mode:
+    if (simple_mode == SimpleMode::NONE) {
         return;
     }
 
-    // mark radio frame as consumed
-    ap.new_radio_frame = false;
+    // exit immediately if RC input is invalid:
+    if (!rc().has_valid_input()) {
```
peterbarker, line 781: This is not the same; RC input would come in at maybe 50Hz, this method is called at the main loop rate, so we'd start to do these transforms much faster than required in the case someone's using simple mode.  The method is also not idempotent, so pretty much completely broken.

We probably want to unconditionally call this method when a new radio frame is received instead.

PR #28359, opened 2024-10-09
AP_PiccoloCAN: use 32 bit microsecond timeouts

tpwrules:
For consistency with other parts of the code. Saves a handful of bytes.

Please see commit messages for caveats with this change; this does change protocol semantics ever so slightly. Needs testing with real hardware.


SchrodingersGat:
@tridge all looks good from my end

PR #28361, opened 2024-10-09
defaults_periph.h: Removed duplication for AP_STATS_ENABLED

TOTON95:
Removed duplication for AP_STATS_ENABLED in AP_Periph defaults


peterbarker (approved):
LGTM

PR #28366, opened 2024-10-09
AP_RangeFinder: initial support for the DFRobot07 lidar

landswellsong:
Support for [DFRobot07](https://www.dfrobot.com/product-2281.html) which is also sold as [VSemi Atom](https://vsemi.io/products/single-point)

Documentation: https://dfimg.dfrobot.com/nobody/wiki/1840a7b7b14e02f3566e0cef5b51e9ba.pdf

Implementation caveats:
- UART only for now;
- 50 ms fixed data rate, can be synced up with `update()` but it involves waiting for the result since the sensor does take a bit of time to process commands;
- there are a bunch of values coming from the sensor that don't have a clear mapping to anything in AP so they are ignored for now, although potentially can be used to estimate the signal quality;

Code quality and style questions that I'd love to have a senior developer feedback:
- Should the initialization be done like I did in `update()` or should I rather have overriden `init_serial`? Given how sensor works it would mean the thread that runs it needs to be waiting for a while.
- Is the offset macros fine or should I rather make structures/unions for a more clear picture of the data layout?
- The CRC32 in the sensor seems to work in opposite bit order reflection mode in comparison to the AP implementation (and frankly any standard implementation). Is it okay to put extra CRC32 utils hidden like I did or should I move that to AP_Math / optimize it etc?


landswellsong:
SITL test in the room conditions:
image

landswellsong:
https://github.com/ArduPilot/ardupilot/actions/runs/11257361975/job/31301422237?pr=28366#step:9:1282

Right, guess I'll rewrite the offset part to work with ARM and test that locally too.

libraries/AP_RangeFinder/AP_RangeFinder_DFRobot_Lidar07.cpp
```diff
@@ -0,0 +1,241 @@
...
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "AP_RangeFinder_DFRobot_Lidar07.h"
+
+#if AP_RANGEFINDER_DFROBOT_LIDAR07_ENABLED
```
peterbarker, line 18:
```suggestion
#include "AP_RangeFinder_config.h"

#if AP_RANGEFINDER_DFROBOT_LIDAR07_ENABLED

#include "AP_RangeFinder_DFRobot_Lidar07.h"
```

... I know this isn't what all of the other headers in here do, but this is the pattern we're trying to use in ArduPilot now where config files are involved.

libraries/AP_RangeFinder/AP_RangeFinder_DFRobot_Lidar07.cpp
```diff
@@ -0,0 +1,241 @@
...
+    cmd_buf[LIDAR07_REG_OFFSET] = reg;
+    *reinterpret_cast<uint32_t*>(&cmd_buf[LIDAR07_VALUE_OFFSET]) = htole32(value);
+    *reinterpret_cast<uint32_t*>(&cmd_buf[LIDAR07_CRC_OFFSET]) = htole32(crc32_unreflected(cmd_buf, LIDAR07_CRC_OFFSET));
+    uart->write(cmd_buf, sizeof(cmd_buf));
+
+    // the sensor doesn't quite like when we lump commands together, give it piecewise input
+    uart->flush();
+
+    // as well as yield to the scheduler to give the sensor some pause to think
+    hal.scheduler->delay(50);
```
peterbarker, line 122: Can't sleep here - this is the main thread!

libraries/AP_RangeFinder/AP_RangeFinder_DFRobot_Lidar07.cpp
```diff
@@ -0,0 +1,241 @@
...
+{
+    if (uart == nullptr) {
+        return;
+    }
+
+    uint8_t cmd_buf[LIDAR07_CONTROLLER_LEN];
+    cmd_buf[LIDAR07_SOF_OFFSET] = LIDAR07_SOF_CONTROLLER;
+    cmd_buf[LIDAR07_REG_OFFSET] = reg;
+    *reinterpret_cast<uint32_t*>(&cmd_buf[LIDAR07_VALUE_OFFSET]) = htole32(value);
+    *reinterpret_cast<uint32_t*>(&cmd_buf[LIDAR07_CRC_OFFSET]) = htole32(crc32_unreflected(cmd_buf, LIDAR07_CRC_OFFSET));
```
peterbarker, line 115: Either use a packed structure (preferred) or zero that buffer on the stack!

libraries/AP_RangeFinder/AP_RangeFinder_DFRobot_Lidar07.cpp
```diff
@@ -0,0 +1,241 @@
...
+    uart->write(cmd_buf, sizeof(cmd_buf));
+
+    // the sensor doesn't quite like when we lump commands together, give it piecewise input
+    uart->flush();
```
peterbarker, line 119: Not sure you can/should be doing this on the main thread in ChibiOS.

peterbarker (changes requested):
Sorry, very brief review only.

libraries/AP_RangeFinder/AP_RangeFinder_DFRobot_Lidar07.cpp
landswellsong: @peterbarker got it. I took the NMEA as a template, will change.

libraries/AP_RangeFinder/AP_RangeFinder_DFRobot_Lidar07.cpp
landswellsong: Hmm is it? I mean I specifically avoided doing that in `init_serial` for that reason. 
I mean here it's scheduled as a separate task: https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/ArduPlane.cpp#L85

So what I assumed would happen is when I do `delay()` the rangefinder thread yields before the next rescheduling. Am I wrong in this assumption?

This is also happening with MAVLink FTP hence why I thought it'd be acceptable: https://github.com/ArduPilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_FTP.cpp#L152

landswellsong: Deleted my previous message after having re-read how scheduling work. I'll make a state machine here then.

libraries/AP_RangeFinder/AP_RangeFinder.h
```diff
@@ -182,6 +182,9 @@ class RangeFinder
 #if AP_RANGEFINDER_RDS02UF_ENABLED
         RDS02UF = 43,
 #endif
+#if AP_RANGEFINDER_DFROBOT_LIDAR07_ENABLED
+        DFRobot_Lidar07 = 44,
```
Hwurzburg, line 186: need to add to metadata for rf type param

Hwurzburg (changes requested):

landswellsong:
Tested the code against a real device, corrected a few mistakes. Works alright indoors. Outdoors it seem to be getting shorter distances alright, but when pointed to the horizon it apparently times out while waiting for the laser bounce and thus the message rate drops dramatically, ultimately spamming 65536 outputs. 

Will also test on a plane on whether this is suitable for landing, but for indoor use I'd say it's ready to be used.

Hwurzburg:
@landswellsong status?

landswellsong:
@Hwurzburg flight-testing over this weekend. SITL operation stable and as expected.

landswellsong:
@Hwurzburg sorry took me a while to get airborne.
I have 2 news about the code in this PR:
1. It's operational as intended. No holes in the log, no lock ups, no crashes. Gives out a 65.535 m reading when out of range which is very far from it's practical ceiling of 10 m so easy to set the bad data filter with `_MAX_CM`. I caught a bunch of near-zero readings but I was literally flying inside a cloud so I guess that's expected.
RFND Dist
2. Now the bad news: [It has just been discontinued by the manufacturer](https://www.dfrobot.com/product-2281.html)

I'll be keeping the patch for myself until I crash all of the units I've bought, but very unsure if this needs to be merged. Basically dead code since it's not a good recommendation for new builds. 

If this is still somehow a good idea to keep it, I can rebase and update. What do you think?

Hwurzburg:
Since its not made, probably no reason to keep this active....thanks

Hwurzburg (commented):
should close this based on author's comment that the lidar is not being made

peterbarker:
Thanks anyway, @landswellsong

PR #28368, opened 2024-10-10
SITL: avoid floating point exception around rangefinder distance

peterbarker:
projecting onto an infinite plane can cause exceptionally long rangefinder distances - for now jsut cap the distance that the simulated rangefinder can return to avoid floating point exceptions.

the FPE is caused in the Plane FlyEachFrame autotest when flying quadplane-copter_tailsitter - which ends up with a rangefinder at yaw-minus-180.

This is the FPE:
image

### DIFF
--- a/libraries/SITL/SIM_Aircraft.cpp
+++ b/libraries/SITL/SIM_Aircraft.cpp
@@ -595,6 +595,15 @@ float Aircraft::rangefinder_range() const
             return INFINITY;
         }
         altitude /= v.z;
+
+        // this is awful, but there are drawbacks to assuming an
+        // infinite plane.  If we don't do this here then we end up
+        // with a ridiculous rangefinder range, and that can cause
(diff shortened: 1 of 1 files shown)


libraries/SITL/SIM_Aircraft.cpp
```diff
@@ -1058,6 +1067,9 @@ void Aircraft::update_external_payload(const struct sitl_input &input)
 
     {
         const float range = rangefinder_range();
+        if (!isinf(range) && range > 100000) {
+            AP_HAL::panic("Bad rangefinder calculation");
```
muramura, line 1071: What happens if a PANIC occurs during flight?
I think it's unacceptable if it becomes uncontrollable.

khancyr: that is SITL, panic will just block the simulation to allow to catch the bug

PR #28370, opened 2024-10-10
AP_HAL_ESP32: RCOutput ported to new mcpwm driver

arg7:
Ported RCOutput code to new MCPWM driver

Best regards,
A. Romanov


davidbuzz:
What's wrong with the old code that it needed all these changes?

arg7:
Old driver was deprecated and will be removed in next idf 6.x release.
 
I don't like new driver, it hides actual hardware implementation too much, to my taste.

The only positive is that it is possible to use one timer for group of six rc out pins, so they work synchronized and may be consume less power.

davidbuzz (approved):
looks good to me

PR #28373, opened 2024-10-10
CADDX gimbal backend

Hwurzburg:
similar to storm32 without gimbal replies....used as template and will fine tune the packet data and rc control and gimbal mode control later...


libraries/AP_Mount/AP_Mount.cpp
```diff
@@ -280,6 +290,18 @@ void AP_Mount::set_yaw_lock(uint8_t instance, bool yaw_lock)
     backend->set_yaw_lock(yaw_lock);
 }
 
+// set mount operating mode
+void AP_Mount::set_mount_mode( uint8_t mount_mode)
```
rmackay9, line 294: nitpick: let's remove the space after the braken, "( "

libraries/AP_Mount/AP_Mount_CADDX.h
```diff
@@ -0,0 +1,84 @@
...
+#define AP_MOUNT_CADDX_RESEND_MS   1000    // resend angle targets to gimbal once per second
+
+#define AXIS_MAX 2047
+#define AXIS_MIN -2048
+
+typedef enum {
+    GIMBAL_MODE_FOLLOW = 0,
+    GIMBAL_MODE_TILT_LOCK = (1<<0),
+    GIMBAL_MODE_ROLL_LOCK = (1<<1),
+    GIMBAL_MODE_PAN_LOCK  = (1<<2),
```
rmackay9, line 23: if possible let's use "Yaw" instead of "Pan".  I think we should avoid synonyms where we can and I think I've gotten rid of most of them

rmackay9: I think this overlaps with the Siyi mount's "FPV" mode which we've had users asking for for quite some time.  Siyi's FPV mode means all the angle targets are in vehicle body frame.

Siyi's FPV mode is only valid while the mount is in RC_TARGETING mode so if this mount is the same we need to be sure this is enfoced somehow.  We don't want the user's aux switch leading to the mount not working while (for example) it tries to point at a POI.

I wonder how much flexibility the CADDX supports.  For example can roll be locked (e.g. earth frame) while pitch is not (e.g. body-frame)?  What would happen if the vehicle rolled 90 degrees (e.g. knife edge)?

re naming, I think we will probably need to replace the existing "yaw lock" boolean with a "lock_mode" that includes the new options.  this should probably be visible to all gimbals even if they don't all support it.  E.g. this enum should be moved up to AP_Mount.

Hwurzburg: we will need to discuss

rmackay9:
We now have confirmation that this PR is basically capable of controlling the CADDX gimbal but the gimbal must be running version 3.4.

We have some other changes from me in this WIP branch https://github.com/rmackay9/rmackay9-ardupilot/commits/caddx-mount-serial/

rmackay9:
I've created an updated version of this PR here https://github.com/ArduPilot/ardupilot_wiki/pull/6441

The commit in the new PR is co-authored by me so attribution for Henry's work remains intact

PR #28375, opened 2024-10-11
Remove stray log_init declaration

peterbarker:
This method went away


IamPete1 (approved):
Builds so must be OK.

PR #28376, opened 2024-10-11
AP_NavEKF3: break circular takeoff-expected loop in Replay

peterbarker:
if NavEKF3 modifies the AHRS state then the next time we start a DAL frame we will record that new state in our Replay log file.

When we replay the log file we will make that value set - but that's *bad* because we're now priming the Replayed EKF3 with state output from the original run-through of the EKF


priseborough:
I have approved https://github.com/ArduPilot/ardupilot/pull/28389 as a better alternative.

PR #28381, opened 2024-10-11
AC_Autorotaton: Format using astyle

muramura:
Since it's a new file, I formatted it using astyle.


peterbarker (changes requested):
@muramura you've formatted Copter's GCS_MAVLink in this PR as well, and that one we can't auto-format!

peterbarker:
Sorry, this is now no longer a new file, so we can't auto-format it.  And we could never auto-format (the now-conflicting) ArduCopter/GCS_MAVLink.cpp file.

Closing this one.

PR #28383, opened 2024-10-11
AP_InertialSensor: added IIM42653

dakejahl:
Tested on the ARKV6X rev5.

This part is functionally equivalent to the iim42652 but with higher full scale range of 32g and 4000dps.

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor_Backend.h
+++ b/libraries/AP_InertialSensor/AP_InertialSensor_Backend.h
@@ -136,6 +136,7 @@ public:
         DEVTYPE_INS_ICM42670 = 0x3A,
         DEVTYPE_INS_ICM45686 = 0x3B,
         DEVTYPE_INS_SCHA63T  = 0x3C,
+        DEVTYPE_INS_IIM42653 = 0x3D,
     };
 
 protected:
--- a/Tools/scripts/decode_devid.py
+++ b/Tools/scripts/decode_devid.py
@@ -104,6 +104,7 @@ imu_types = {
     0x3A : "DEVTYPE_INS_ICM42670",
     0x3B : "DEVTYPE_INS_ICM45686",
     0x3C : "DEVTYPE_INS_SCHA63T",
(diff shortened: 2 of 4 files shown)


libraries/AP_InertialSensor/AP_InertialSensor_Backend.h
```diff
@@ -131,11 +131,12 @@ class AP_InertialSensor_Backend
         DEVTYPE_INS_ICM42605 = 0x35,
         DEVTYPE_INS_ICM40605 = 0x36,
         DEVTYPE_INS_IIM42652 = 0x37,
-        DEVTYPE_BMI270       = 0x38,
-        DEVTYPE_INS_BMI085   = 0x39,
-        DEVTYPE_INS_ICM42670 = 0x3A,
-        DEVTYPE_INS_ICM45686 = 0x3B,
-        DEVTYPE_INS_SCHA63T  = 0x3C,
+        DEVTYPE_INS_IIM42653 = 0x38,
```
peterbarker, line 134: I don't think we can renumber these - they're logged and tools like `decode_devid.py` would insta-break.

peterbarker (changes requested):

peterbarker (changes requested):
OK, looks right now.

Will mark for DevCall, but:
 - please add to the `decode_devid.py` metadata
 - we should only have two commits, one in AP_InertialSensor and one in Tools (for `decode_devid.py`)
 -

IamPete1:
> * please add to the `decode_devid.py` metadata

Also https://github.com/ArduPilot/WebTools/blob/main/Libraries/DecodeDevID.js please.

dakejahl:
Also https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/APM/APMSensorIdDecoder.qml

tridge (approved):

andyp1per (approved):

peterbarker:
Merged, thanks!

PR #15810, opened 2020-11-14
Improve operation of simple hand launched planes

priseborough:
These changes have been developed following on from code inspection and flying of a strix nano goblin 250gram flying wing with Matek F765-wing controller and M8Q GPS with no airspeed sensor or magnetometer.

1) Airspeed estimate used to scale the pitch and roll loops was too low when launching into a strong wind resulting in limit cycling of the loops until the vehicle turned and the wind speed estimates improved. This PR uses the ARSPD_FBW_MIN and ARSPD_FBW_MAX parameter values to set an airspeed estimate halfway between the two limits and with an uncertainty of half the limit interval. This is used as a 'fake' airspeed measurement to enable faster wind speed estimation as start of flight. Also a zero sideslip assumption is used by default unless inhibited by parameter. These tow measures enable the EKF's to provide a robust wind speed measurement. Only wind speed states are updated by these measurements during this mode of operation.

2) Low ground speed during a hand launch into a strong wind in TAKEOFF mode would result in a nose down pitch angle after launch and flight into ground. When a hand launch is detected by setting of takeoff acceleration of speed threshold limits, a fixed lower pitch angle limit of 5 degrees is used during launch.

3)  Flying without a magnetomer relies on correct operation of the EKF-GSF yaw estimator. A vulnerability in the GSF yaw estimator calculation that could result in underflow of the weighting calculation if large innovations effected all hypothesis filters has been encountered and fixed in the PX4/ecl implementation of this algorithm. This algorithm fix has been incorporated here.

4) When flying without a magnetometer and performing repeat flights on a single power cycle, the EKF yaw drifts on the ground and was not being reset on the following takeoff/launch which can cause loss of navigation and attitude reference. 

**Testing**

Log from flight test at commit https://github.com/ArduPilot/ardupilot/pull/15810/commits/5d47abb0be76f4adfd0707ce039bf3b172f3429a flown on a Strix Nano Goblin with MatekF765-Wing board, GPS and no magnetomer or airspeed sensor.

https://drive.google.com/file/d/15TIeZ61IHtgrbUQXkiPNGzEGRBzMZmFg/view?usp=sharing

PR #28389, opened 2024-10-12
AP_NavEKF3: remove takeoff-detection code

peterbarker:
the current implementation pokes data back into the DAL, which is in the not-good category.

This is an alternative to https://github.com/ArduPilot/ardupilot/pull/28376 .  Instead of breaking the loop we just remove the offending code.

I'm not sure about this one, but talking with @tridge he thought it a reasonable step.  OTOH, this code wasn't added just for fun, it might have solved a real problem.  tridge and I both thought the vehicle probably has more information to do this sort of calculation...

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
@@ -894,15 +894,6 @@ void NavEKF3_core::calcOutputStates()
         posOffsetNED.zero();
     }
 
-    // Detect fixed wing launch acceleration using latest data from IMU to enable early startup of filter functions
-    // that use launch acceleration to detect start of flight
-    if (!inFlight && !dal.get_takeoff_expected() && assume_zero_sideslip()) {
-        const ftype launchDelVel = imuDataNew.delVel.x + GRAVITY_MSS * imuDataNew.delVelDT * Tbn_temp.c.x;
-        if (launchDelVel > GRAVITY_MSS * imuDataNew.delVelDT) {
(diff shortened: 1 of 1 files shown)


rmackay9:
I agree that the flight code (and not the EKF) should not be responsible for determining if the vehicle is flying or not.  Maybe the flight code needs to tell the EKF but the EKF shouldn't be trying to figure this out on its own

priseborough (approved):
Agree - this is a decision best made outside the EKF

LupusTheCanine:
For reference, I found the PR #15810 in which the feature was added to the mainline AP and the comment that indicates the likely cause of that detection being added in the first place https://github.com/ArduPilot/ardupilot/pull/15810#issuecomment-750650213 and the commit adding the logic https://github.com/ArduPilot/ardupilot/pull/15810/commits/204dff27b3da43cbf0cffed761dd8f97507df779

peterbarker:
Ping @Hwurzburg .

I think the information that @LupusTheCanine points out this was added for compass-less Planes to get the GSF working ASAP.  Could you confirm that's your recollection, and what the scenario was?  It seems like it was something along the lines of "throw a compass-less Plane off in FBWA (with no throttle-up) and expect yaw to start working instantly."

Hwurzburg:
I usually auto takeoff, not FBWA takeoff, but the problem would have occurred in both...without Paul's changes it could veer horribly from throw direction in takeoff....we need to be sure any changes to this area of the code are thoroughly tested in a compass less plane takeoff, both manual and auto

Issue #28295, opened 2024-10-03
DroneCAN: RC link statistics


olliw42:
DroneCAN supports RC receivers via the https://github.com/dronecan/DSDL/blob/master/dronecan/sensors/rc/1140.RCInput.uavcan message. 

However, sadly, it only has a "quality" field which by looking into the AP code is in fact RSSI in units of 0 ... 255. 

Today however most people in the FPV arena also want to get LQ, and possibly SNR, and especially want to put them on their OSD. There is no support yet for this. As much as I know support for extending DroneCAN messages definitions while retaining compatibility with existing implementations is not yet existing or in the making. So, extending the existing RCInput seems not a viable path.

This leads to a pretty ugly situation and massively reduces the usefullness of DroneCAN RC (the situation is actually similarly sad as with the MAVLink receiver link stats not being there). To give a sense of it as I perceive it based on user responses: This makes folks wanting CRSF in addition ... just for the purpose of also getting the link stats. Obviously this is a bummer for using DroneCAN since the promise of simplified wiring cannot be kept and the advantages over normal serial largely evaporates. This therefore makes folks wanting CRSF via DroneCAN serial ... which appears to not work with AP, but if it would, doing such work arounds is hardly in the intention of DroneCAN, and also ugly as one then consumes two of the DroneCAN serial ports.

I have no idea what the "best" path and approach would be to correct the situation, but I feel I should raise it as an issue. 

Three approaches would come to my mind:
1. Add a DroneCAN link statistics message. In view of the slow and difficult process with the similar MAVLink link statistics message this may not look as the best path.
2. Add a revised RCInput version 2 message, with few little more fields for the link stats.
3. Add flags to teh existing RCInput message so that the quality field could host different metrics (like RSSI, LQ, SNR), Thuis would at least allow to send the data round robin, i.e. packet #n send RSSI, #n+1 sends LQ, #n+2 send SNR, and so on

tridge:
> 3\. Add flags to teh existing RCInput message so that the quality field could host different metrics (like RSSI, LQ, SNR),

my preference is to add a STATUS flag in the existing message that indicates quality is LQ or SNR

tridge:
maybe assign 3 of the status bits as what type the quality is?

peterbarker:
@olliw42 tear off three bits and use them to indicate what the RSSI field *actually* represents - that's basically your option (3), I think?

olliw42:
ok. Will do that :)
THX!

[...]

PR #28390, opened 2024-10-12
DroneCAN: Add support for few more statistics to DroneCAN RC

olliw42:
This PR adds support for few more link statistics to the DroneCAN RC.

Specifically, it adds the support of three status flags which are propsoed to be added to the DroneCAN RCInput message (https://github.com/dronecan/DSDL/pull/56), which allows to also send LQ, RSSI_DBM, and SNR, in addition to RSSI, via that message. These data are then made available to the OSD.

This PR is in principle dependent on https://github.com/dronecan/DSDL/pull/56, but it doesn't need it, i.e. compiles and works fine without this being merged, since the code uses bit fields, and none of the flags defined in https://github.com/dronecan/DSDL/pull/56.

I could test this only very lightly myself on my bench setup, but @jlpoltrack did test this carefully on a matekh743 with plane firmware. He found the CRSF to still work and also the new DroneCAN support to work.

The PR has a slight change in behavior as compared to before, namely that then no STATUS flags is set it now reports -1 to the add_input() rssi field, whereas before it was reporting 0. (old behavior: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_RCProtocol/AP_RCProtocol_DroneCAN.cpp#L97).

The OSD was using so far only the crsf() singleton, so it needed now some abstraction so either the DroneCAN or the crsf class is used by the OSD, dependent on which RC protocol is configured by the user. I found that all a bit strange code, and as result of a struct being used the abstraction doesn't look like it is for other frontedn/backend examples. Bit it's aligned with how the crsf() singleton works.


jlpoltrack:
Related Issue: https://github.com/ArduPilot/ardupilot/issues/28295

Tested with both CRSF and DroneCAN RC - OSD elements behaved as expected when using RC_PROTOCOLS bitmask to switch between the two protocols.

libraries/AP_RCProtocol/AP_RCProtocol.h
```diff
@@ -250,6 +250,17 @@ class AP_RCProtocol {
...
 #endif
 
+    // some backends have a struct LinkStatus and a field _link_status, avoid name clash
+    struct RcLinkStatus {
+        int16_t rssi = -1;
+        int16_t link_quality = -1;
+        int8_t rssi_dbm = -1;
+        int8_t snr = INT8_MIN;
+    };
+    volatile struct RcLinkStatus _rc_link_status;
```
peterbarker, line 260:
```suggestion
    volatile struct RCLinkStatus _rc_link_status;
```

Why `volatile`?

libraries/AP_RCProtocol/AP_RCProtocol_DroneCAN.cpp
```diff
@@ -90,12 +90,38 @@ void AP_RCProtocol_DroneCAN::update()
         }
         last_receive_ms = rcin.last_sample_time_ms;
 
+        if (rcin.bits.QUALITY_VALID) {
+            frontend._rc_link_status.rssi = rcin.quality;
+        } else
+        if (rcin.bits.QUALITY_LQ) {
```
peterbarker, line 96:
```suggestion
        switch (QualityType)rcin.bits.QUALITY_TYPE)) {
        case QualityType::RSSI:  // this would be value zero
```

similarly elsewhere

libraries/AP_RCProtocol/AP_RCProtocol_DroneCAN.h
```diff
@@ -37,6 +37,9 @@ class AP_RCProtocol_DroneCAN : public AP_RCProtocol_Backend {
             struct {
                 uint8_t QUALITY_VALID : 1;
                 uint8_t FAILSAFE : 1;
+                uint8_t QUALITY_LQ : 1;
+                uint8_t QUALITY_RSSI_DBM : 1;
+                uint8_t QUALITY_SNR : 1;
```
peterbarker, line 42:
```suggestion
                uint8_t QUALITY_TYPE : 3;
```

libraries/AP_RCProtocol/AP_RCProtocol_DroneCAN.cpp
```diff
@@ -90,12 +90,38 @@ void AP_RCProtocol_DroneCAN::update()
         }
         last_receive_ms = rcin.last_sample_time_ms;
 
+        if (rcin.bits.QUALITY_VALID) {
+            frontend._rc_link_status.rssi = rcin.quality;
+        } else
```
peterbarker, line 95:
```suggestion
```

libraries/AP_RCProtocol/AP_RCProtocol_DroneCAN.cpp
```diff
@@ -90,12 +90,38 @@ void AP_RCProtocol_DroneCAN::update()
...
+            // AP rssi: -1 for unknown, 0 for no link, 255 for maximum link
+            if ( rcin.quality < 50) {
+                frontend._rc_link_status.rssi = 255;
+            } else if ( rcin.quality > 120) {
+                frontend._rc_link_status.rssi = 0;
+            } else {
+                frontend._rc_link_status.rssi = int16_t(roundf((120.0f - rcin.quality) * (255.0f / 70.0f) ));
+            }
+        } else
+        if (rcin.bits.QUALITY_SNR) {
```
peterbarker, line 110: Our style does not permit a linefeed between these two lines

peterbarker (changes requested):
Sorry, we must not have communicated our thoughts well on how we'd like this shaped.

We wanted three bits set aside which would specify a value out of an enumeration, not three bits, one for each of the types (and re-use of valid as RSSI, it seems?)

olliw42:
> We wanted three bits set aside which would specify a value out of an enumeration, not three bits, one for each of the types (and re-use of valid as RSSI, it seems?)

I don't understand the language, but form the suggestion I infeer you say you want it as said by @tpwrules in https://github.com/dronecan/DSDL/pull/56 :) I'll do

libraries/AP_RCProtocol/AP_RCProtocol.h
olliw42: this I didn't understand too ... so I was just closily following what is done for the CRSF class 
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_RCProtocol/AP_RCProtocol_CRSF.h#L310-L314
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_RCProtocol/AP_RCProtocol_CRSF.h#L361
assuming that someone had a good reason which is above my paygrade

pl say clear "yes volatile" or "no voloatile" and I will do

libraries/AP_RCProtocol/AP_RCProtocol_DroneCAN.cpp
olliw42: understood

libraries/AP_RCProtocol/AP_RCProtocol_DroneCAN.h
olliw42: understood

olliw42:
made the changes as suggested - I hope at least.

also added the option for tx_power (mLRS doesn't use it but orther systems may)

changes were tested by @jlpoltrack and found to work for both normal CRSF and DroneCAN rc.

I'd like to reiterate the question if it would be acceptable to pack both LQ and active_antenna into the quality field

peterbarker:
Please do add the active antenna bit.

We need to know what older versions of ArduPilot (eg. 4.5) will shows when a RC receiver sending this new style of packet is used.

olliw42:
> Please do add the active antenna bit.

great, will do.

> We need to know what older versions of ArduPilot (eg. 4.5) will shows when a RC receiver sending this new style of packet is used.

in the OSD it will show nothing since there is no code path for how the data received by DroneCAN can go to the OSD. The OSD uses the crsf() singleton.  E.g. https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_RCProtocol/AP_RCProtocol_DroneCAN.cpp#L97-L98, and similarly for the other ext stats fields.

as regards AP_RCpororocol_DroneCAN it will set rssi to 0 if the QUALITY bit is not set: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_RCProtocol/AP_RCProtocol_DroneCAN.cpp#L97-L98. So, if in the new message other bits are set rssi will be set to zero - like before.

IMHO I did consider backwards compatibilioty properly when I made that suggestion

libraries/AP_RCProtocol/AP_RCProtocol.h
olliw42: has the "volatile" thing been sorted out? It should stay?

timtuxworth:
> active_antenna bit added

This only allows for 2 antennas. Note that the RadioMaster DBR4 has 4 antennas!
https://www.radiomasterrc.com/products/dbr4-dual-band-xross-gemini-expresslrs-receiver

olliw42:
it needs to be understood that the mechanism used here isn't and can't be a full fledged link statistics
if that is wanted a link stats message needs to come

if 4 antenna is really intended with the mechanism here, one could do LQ in units of 2, to get the additionally required bit

olliw42:
> We need to know what older versions of ArduPilot (eg. 4.5) will shows when a RC receiver sending this new style of packet is used.

while the explanation given in the above based on the code might be compelling enough, @jlpoltrack has explicitely tested this with result "AP 4.5.7 + mLRS Latest w/ if(1) works fine too.  RC Works and OSD Elements don't show (as expected)"

olliw42:
@peterbarker 
I'd like to give this another try,
mLRS is going to officially release DroneCAN support in the very, very near future, and it would be great to have this thing in ArduPilot. I believe users will appreciate having this statistics.
Rereading the above, I believe all comments were properly adressed.

jlpoltrack:
Tested with a build that has Extended Link Stats enabled - all five link stats populate as expected:

RC_LQ, RSSIDBM, RC_ANT, RC_SNR, RC_PWR

image

peterbarker (approved):
LGTM - thanks!

tpwrules (approved):
Thanks for your patience, this looks reasonable and it's great to hear that it's been tested.

Future work:
* make GHST use this too
* remove CRSF and GHST specific get_link_status backend methods.
* do a real lock
* maybe migrate some of the more specific getters to this

PR #28391, opened 2024-10-12
Copter: Share flags across common areas

muramura:
The `failsafe_enabled` flag can be shared across common areas. 
By sharing it, I can group them together.

### DIFF
--- a/ArduCopter/failsafe.cpp
+++ b/ArduCopter/failsafe.cpp
@@ -48,26 +48,28 @@ void Copter::failsafe_check()
         return;
     }
 
-    if (!in_failsafe && failsafe_enabled && tnow - failsafe_last_timestamp > 2000000) {
-        // motors are running but we have gone 2 second since the
-        // main loop ran. That means we're in trouble and should
-        // disarm the motors->
-        in_failsafe = true;
-        // reduce motors to minimum (we do not immediately disarm because we want to log the failure)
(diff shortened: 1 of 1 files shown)


ArduCopter/failsafe.cpp
```diff
@@ -48,26 +48,28 @@ void Copter::failsafe_check()
...
-        // disarm the motors->
-        in_failsafe = true;
-        // reduce motors to minimum (we do not immediately disarm because we want to log the failure)
-        if (motors->armed()) {
-            motors->output_min();
-        }
-
-        LOGGER_WRITE_ERROR(LogErrorSubsystem::CPU, LogErrorCode::FAILSAFE_OCCURRED);
-    }
+    if (failsafe_enabled) {
```
peterbarker, line 51: Why not just early-return instead?

peterbarker (changes requested):
Please state what testing you have done for this.

This code is safety-critical and we don't cross this code very much at all, as we usually rely on a hardware watchdog.

However this would still be crossed on Linux-based autopilots if bad things happened, so we do need to know it still work!

peterbarker:
Ping @muramura - I wanted to know what testing you've done on this.

Still want to chase it?

peterbarker:
No answer from author, and much testing required.

Please re-open if you want to pursue this one.

PR #28395, opened 2024-10-13
AP_HAL_ChibiOS: increase the number of memory regions for crashdump

bugobliterator:
* We were running out of memory regions and hence losing dump data even though we had enough flash to record, this PR increases the number of regions and also checks num region overruns for bss and heap


tridge (approved):

tridge:
Thanks Sid!

PR #28396, opened 2024-10-13
AP_NavEKF3: Remove plane gps check from EKF3 init

paulerikf:
Allow EKF3 to initialize w/out GPS for non-VTOL fixed wings.

The previous behavior stopped non-VTOL planes from being used completely without GPS. As NavEKF3::setOriginLLH returns false if EKF3 is unitialized - which makes it impossible to set the EKF Origin from a GCS before a flight.

This check originates in EKF2, and I'm unable to see a reason for it to remain in the code. The arming checks already provide configurable GPS checks to prevent accidentally takeoff w/out GPS unless that is intended.  I've tested in SITL w/out this check removed and got the behavior I expected - able to set EKF Origin from a GCS & take off w/out GPS if configured w/out GPS arming checks. 

Please don't hesitate to push back if I'm missing something major here.


libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
```diff
@@ -472,15 +472,6 @@ bool NavEKF3_core::InitialiseFilterBootstrap(void)
     // update sensor selection (for affinity)
     update_sensor_selection();
 
-    // If we are a plane and don't have GPS lock then don't initialise
-    if (assume_zero_sideslip() && dal.gps().status(preferred_gps) < AP_DAL_GPS::GPS_OK_FIX_3D) {
```
tridge, line 476: this could be added as an EK3_OPTIONS bit, but not by default.

tridge (changes requested):

libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
```diff
@@ -165,7 +165,7 @@ bool NavEKF3_core::setup_core(uint8_t _imu_index, uint8_t _core_index)
         return false;
     }
 #endif
- 
+
```
tridge, line 168: please remove the white space changes :-)

libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
paulerikf: Will give that a try!

paulerikf:
Thanks for the review @tridge

Added an EK3_OPTIONS bit: PlaneRequiresGPS - It's set true by default. If set false the GPS check for planes is skipped.

Does this work?

libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -738,10 +738,10 @@ const AP_Param::GroupInfo NavEKF3::var_info2[] = {
 
     // @Param: OPTIONS
     // @DisplayName: Optional EKF behaviour
-    // @Description: This controls optional EKF behaviour. Setting JammingExpected will change the EKF nehaviour such that if dead reckoning navigation is possible it will require the preflight alignment GPS quality checks controlled by EK3_GPS_CHECK and EK3_CHECK_SCALE to pass before resuming GPS use if GPS lock is lost for more than 2 seconds to prevent bad
-    // @Bitmask: 0:JammingExpected
+    // @Description: This controls optional EKF behaviour. Setting JammingExpected will change the EKF behaviour such that if dead reckoning navigation is possible it will require the preflight alignment GPS quality checks controlled by EK3_GPS_CHECK and EK3_CHECK_SCALE to pass before resuming GPS use if GPS lock is lost for more than 2 seconds to prevent bad GPS measurements from getting fused. If PlaneRequiresGPS is true, EKF3 will not initialize unless the preferred GPS has at least GPS_OK_FIX_3D.
+    // @Bitmask: 0:JammingExpected,1:PlaneRequiresGPS
     // @User: Advanced
-    AP_GROUPINFO("OPTIONS",  11, NavEKF3, _options, 0),
+    AP_GROUPINFO("OPTIONS",  11, NavEKF3, _options, 2),
```
peterbarker, line 744: We can't really change the default on a bitmask, sorry.  If someone has set any other bits in their options then this new bit wouldn't be set in their parameter and thus their vehicle would change behaviour underneath them.

This has to be inverted, so something like `1:SuppressGPSFixCheck`

libraries/AP_NavEKF3/AP_NavEKF3.h
```diff
@@ -455,6 +455,7 @@ class NavEKF3 {
     // enum for processing options
     enum class Options {
         JammingExpected     = (1<<0),
+        PlaneRequiresGPS    = (1<<1),
```
peterbarker, line 458: Nothing about this is `Plane`-specific AFAICS.

peterbarker (changes requested):

libraries/AP_NavEKF3/AP_NavEKF3.cpp
paulerikf: Gotcha, o.k. will do.

libraries/AP_NavEKF3/AP_NavEKF3.h
paulerikf: Correct me if I'm wrong, but I think it is Plane-specific implicitly.

The check blocking initialization only triggers if assume_zero_sideslip() is true:
 https://github.com/ArduPilot/ardupilot/blob/145cc4bb260c56fb92632886a624bc06b56b0e4d/libraries/AP_NavEKF3/AP_NavEKF3_core.cpp#L476
 
 `assume_zero_sidslip` is only true if `get_fly_forward()` is true & the vehicle is not a ground vehicle.
 https://github.com/ArduPilot/ardupilot/blob/145cc4bb260c56fb92632886a624bc06b56b0e4d/libraries/AP_NavEKF3/AP_NavEKF3_Control.cpp#L665-L671
 
AFAICT only the Rover (a ground vehicle) & Plane (in non-VTOL mode) ever return `get_fly_forward()` - `true`.
 
Obviously if another non-ground vehicle w/ fly_forward `true` gets added this check would apply to it as well.

peterbarker (commented):
This accords with my understanding of the outcome from the last DevCall.

It does occur to me - just now, sorry! - that we could instead predicate this check on any of your EKF3 source sets requiring a GPS.  Possibly in addition to this new flag.

paulerikf:
> This accords with my understanding of the outcome from the last DevCall.
> 
> It does occur to me - just now, sorry! - that we could instead predicate this check on any of your EKF3 source sets requiring a GPS. Possibly in addition to this new flag.

@peterbarker That approach makes a lot of sense to me. If we go that direction, I assume we drop the check of `assume_zero_sideslip()`?

peterbarker:
> Just to run my plan by you @peterbarker :
> 
> I will add a `anySourceSetUsingGPS()` function to AP_NavEKF_Source.cpp, and check that (contingent on the `SuppressGPSFixCheck` bit not being set).
> 
> Side Note:
> 
> It feels to me like this is a somewhat confusing replication of parts of the pre_arm_check functionality in AP_NavEKF_Source, and there isn't really a reason (that I can see) for the EKF to not initialize w/out GPS.
> 
> So I'm still somewhat in favor of just deleting the check in question entirely. But if you guys aren't comfortable w/ that I'm happy to make the aforementioned changes (adding `anySourceSetUsingGPS()`).

It was just a thought, don't adjust this PR based on that :-)

I've marked this for DevCall but it might be a while before we have one of those - skipping two due to ArduPilot Developer Conference.

tridge (approved):

tridge:
@peterbarker looks OK, but please check

tridge (changes requested):
I'd like to see an auto-test showing this working
also, do you have a flight log showing this working on a real aircraft? preferably a replay log
and is there a reason not to just fly with DCM for this use case?

rmackay9:
I wonder if we even need the check.  If it's safe to bypass the check then why don't we just remove it?

tridge:
another option is to set EK3_SRC1_POSXY=0

rmackay9:
After a discussion on the dev call I think we need to take a deeper look at the Plane arming checks and safely allow arming without a GPS instead of adding an option which basically bypasses the checks.

Ryanf55:
Which modes do you want to use without GPS, and how are you getting position data? ExternalOdom or other custom position source? We can improve the arming checks and avoid you needing a new option bit.

paulerikf:
Apologies for the slow reply & thanks all for taking a look. I will be very busy the next few weeks but wanted to reply.

@Ryanf55 
> Which modes do you want to use without GPS, and how are you getting position data? ExternalOdom or other custom position source? We can improve the arming checks and avoid you needing a new option bit.

Unsure on exact modes, but would like to be able to use "most of them" - (specifically we usually use FBWA, Guided/Loiter & Auto).

@rmackay9 
I haven't looked into the arming checks too closely yet, but there is already a GPS check in there.

When testing the earlier version of this PR (removing the check entirely) the GPS arming checks worked as expected, and were nicely configurable to let me disable them if I think I know what I'm doing :P

PR #28397, opened 2024-10-14
AP_DroneCAN: force DroneCAN zero throttle when disarmed

tridge:
if a user has set CAN_D1_UC_ESC_RV which is the mask of ESCs that are reversible we were sending -8191 when disarmed, which is full reverse throttle. This is the correct output when armed as it is treated as full reverse at "PWM" 1000 and stopped at 1500, but when disarmed we should always send zero or the user may find all ESCs spin up at full reverse when disarmed if the ESC supports reverse throttle (which is rare in DroneCAN ESCs)

### DIFF
--- a/libraries/AP_DroneCAN/AP_DroneCAN.cpp
+++ b/libraries/AP_DroneCAN/AP_DroneCAN.cpp
@@ -844,9 +844,9 @@ void AP_DroneCAN::SRV_send_esc(void)
     // if at least one is active (update) we need to send to all
     if (active_esc_num > 0) {
         k = 0;
-
+        const bool armed = hal.util->get_soft_armed();
         for (uint8_t i = esc_offset; i < max_esc_num && k < 20; i++) {
-            if ((((uint32_t) 1) << i) & _ESC_armed_mask) {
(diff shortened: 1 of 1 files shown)


tridge:
> I think this will break motor test on quadplane (it just stets motor armed state not global armed state).

I have tested this and it works fine, this is a motor test of front right motor:
image


> I don't understand the situation that this would happen in. You have to have reversible ESCs and set the reverse bitmask but not have configured AP to use reversible motors?

I had a reversible AM32 DroneCAN ESC. I set CAN_D1_UC_ESC_RV to 15 and ArduPilot instantly demanded full reverse.

IamPete1:
> I have tested this and it works fine, this is a motor test of front right motor:

I was not expecting that, I had a deeper look, plane forces armed in this function when in motor test. https://github.com/ArduPilot/ardupilot/blob/2524583ddacbf95af3b88f15beecdb72794e3ba2/ArduPlane/AP_Arming.cpp#L390-L392

So we do endup setting the global armed state, just not directly from the motor test function per the other vehicles.

IamPete1 (approved):
LGTM, sound like this brings it inline with what we do for DShot.

PR #28399, opened 2024-10-14
AP_CANManager: use 32 bit timeout

tpwrules:
Saves a handful of bytes. 71 minutes ought to be enough for anybody!


tpwrules:
Tested that DroneCAN peripherals and scripting CAN sensors still work properly on CubeOrange.

tpwrules:
A note about safety: this doesn't change any of the arithmetic, just changes the types so that specifying a timeout of longer than 71 minutes is not possible. Nowhere that I could find calculates timeouts, they are always hardcoded, so this is safe.

I did update a variable name to make clear that it's actually a 64 bit deadline instead of a timeout, but not the arithmetic.

bugobliterator (approved):

tridge (approved):

PR #28400, opened 2024-10-14
AP_DroneCAN: optimize message subscription process

tpwrules:
* remove unnecessary nullptr check, these are always called from an initialized AP_DroneCAN so if it's nullptr something has gone horrifically wrong

* pass in driver index instead of repeatedly calling function to get

* simplify error handling; knowing exactly which callback failed is not super helpful and one failing likely means subsequent ones will as they can only fail due to being out of memory


tpwrules:
Tested that a DroneCAN GPS still works on CubeOrange.

tridge:
added DelayMerge for after 4.6 forks

libraries/AP_Baro/AP_Baro_DroneCAN.h
```diff
@@ -15,7 +15,7 @@ class AP_Baro_DroneCAN : public AP_Baro_Backend {
 
     void update() override;
 
-    static void subscribe_msgs(AP_DroneCAN* ap_dronecan);
+    static bool subscribe_msgs(AP_DroneCAN* ap_dronecan, uint8_t driver_index);
```
tridge, line 18: could it be a reference?

tpwrules: Technically possible, but doesn't save anything and means everything has to be `*this` and `&ap_dronecan`. I don't like it.

libraries/AP_Airspeed/AP_Airspeed_DroneCAN.cpp
```diff
@@ -13,21 +13,13 @@ extern const AP_HAL::HAL& hal;
 AP_Airspeed_DroneCAN::DetectedModules AP_Airspeed_DroneCAN::_detected_modules[];
 HAL_Semaphore AP_Airspeed_DroneCAN::_sem_registry;
 
-void AP_Airspeed_DroneCAN::subscribe_msgs(AP_DroneCAN* ap_dronecan)
+bool AP_Airspeed_DroneCAN::subscribe_msgs(AP_DroneCAN* ap_dronecan, uint8_t driver_index)
 {
-    if (ap_dronecan == nullptr) {
```
tridge, line 18: i don't think the removal of the driver_index helps us move towards having the subscribes happen directed by the library subsystems instead of by AP_DroneCAN

tridge: having: 
```
const auto driver_index = ap_dronecan->get_driver_index();
```
may make things more readable while keeping the interface simple

libraries/AP_DroneCAN/AP_DroneCAN.cpp
```diff
@@ -368,45 +368,50 @@ void AP_DroneCAN::init(uint8_t driver_index, bool enable_filters)
...
-    AP_TemperatureSensor_DroneCAN::subscribe_msgs(this);
+    subscribed = subscribed && AP_TemperatureSensor_DroneCAN::subscribe_msgs(this, driver_index);
 #endif
 #if AP_RPM_DRONECAN_ENABLED
-    AP_RPM_DroneCAN::subscribe_msgs(this);
+    subscribed = subscribed && AP_RPM_DroneCAN::subscribe_msgs(this, driver_index);
 #endif
 
+    if (!subscribed) {
+        AP_BoardConfig::allocation_error("DroneCAN callback");
```
tridge, line 412: I do like this change

tridge (changes requested):
happy with this except for driver_index change

PR #28402, opened 2024-10-14
AP_Temperature: fix MCP9600 i2c address and TEMP9 index

magicrub:
fix MCP9600 i2c address and TEMP9 index.


CraigElder:
This has been tested on real hardware with 8 Temperature sensors connected

tridge (approved):

PR #28406, opened 2024-10-15
Plane: fixed guided VTOL command in terrain mode

tridge:
if the GCS commands a guided re-position with an AGL frame and quadplane is in Q_GUIDED_MODE=1 then this prevents us adding the home altitude to the target terrain altitude
we really need auto tests for these

### DIFF
--- a/ArduPlane/GCS_MAVLink_Plane.cpp
+++ b/ArduPlane/GCS_MAVLink_Plane.cpp
@@ -759,7 +759,7 @@ bool GCS_MAVLINK_Plane::handle_guided_request(AP_Mission::Mission_Command &cmd)
 void GCS_MAVLINK_Plane::handle_change_alt_request(AP_Mission::Mission_Command &cmd)
 {
     plane.next_WP_loc.alt = cmd.content.location.alt;
-    if (cmd.content.location.relative_alt) {
+    if (cmd.content.location.relative_alt && !cmd.content.location.terrain_alt) {
         plane.next_WP_loc.alt += plane.home.alt;
     }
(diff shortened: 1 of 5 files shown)


peterbarker:
> What we really need to do is change to using an alt_frame attribute on Location instead of using these booleans.

Yes, we do, I've had patches around it before.

There some subtleties around the terrain stuff in particularly - there's some weird persistence with terrain following which you have to be really careful around.

Nothing impossible, of course, but terrain is unforgiving :-)

The trick is to make `.alt` private in Location and fix what breaks.

Ryanf55:
> > What we really need to do is change to using an alt_frame attribute on Location instead of using these booleans.
> 
> Yes, we do, I've had patches around it before.
> 
> There some subtleties around the terrain stuff in particularly - there's some weird persistence with terrain following which you have to be really careful around.
> 
> Nothing impossible, of course, but terrain is unforgiving :-)
> 
> The trick is to make `.alt` private in Location and fix what breaks.

@timtuxworth  Wanna figure this out at dev conference this week?

timtuxworth:
Absolutely

peterbarker (changes requested):
Needs autotests

tridge:
> Needs autotests

test added

peterbarker (approved):
LGTM

rmackay9:
I was unable to immediately include this in 4.6.0-beta5 due to merge conflicts

rmackay9:
this is included in 4.6.0-beta5 via PR https://github.com/rmackay9/rmackay9-ardupilot/pull/40

PR #28410, opened 2024-10-15
AP_Networking: make connector loopback test priority to be PRIORITY_IO-1

bugobliterator:


magicrub:
This thread is faster so it makes sense but I'm not sure amount will change. Do you have any metrics to show that this makes any difference?

bugobliterator:
> This thread is faster so it makes sense but I'm not sure amount will change. Do you have any metrics to show that this makes any difference?

We need this for Testing the modules. Lua scripts don't run properly until we set priority higher than loopback test thread.

tridge:
Maybe just set lua thread priority higher with the parameter for that?

bugobliterator:
> Maybe just set lua thread priority higher with the parameter for that?

I think the loopback test needs to be lower than IO anyways. It completely chokes out all the remaining threads. As with loopback its just pushing and receiving data at a really high rate.

tridge (approved):

PR #28413, opened 2024-10-15
Copter: Change enum values to int type

muramura:
AP_Enum is an ENUM variable.
The FLOAT type is incorrect.
Change it to the INT type.

### DIFF
--- a/ArduCopter/Parameters.cpp
+++ b/ArduCopter/Parameters.cpp
@@ -690,7 +690,7 @@ const AP_Param::Info Copter::var_info[] = {
     // @Description: Used by Throw mode. Controls whether motors will run at the speed set by MOT_SPIN_MIN or will be stopped when armed and waiting for the throw.
     // @Values: 0:Stopped,1:Running
     // @User: Standard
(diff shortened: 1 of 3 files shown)


IamPete1:
All defaults are stored as floats. https://github.com/ArduPilot/ardupilot/blob/81768b26ff447c7a8684c75f03fe6db941846b92/libraries/AP_Param/AP_Param.h#L205

muramura:
@IamPete1 san. Comments thanks.
All the values are floating-point numbers, aren't they?
I think it would be better if the parameter definitions matched the type of AP_XXXX.
Enums are integer values.

peterbarker:
@muramura as @IamPete1 points out, the underlying type in the structure is float, not integer.  The cast to float would have been added to fix the compilation error which said pretty much exactly that.

Casting it to an integer first won't prevent the cast to float but will (a) make the code more ugly and (b) hide the fact that the value is stored as a float.

Sorry, closing this one.

muramura:
I understand that the parameter values are float values.

PR #28414, opened 2024-10-15
Copter: fixed landing gear deploys and retracts when rangefinder is out of range

4dri4anxD:
When the drone height exceeds the working range of the rangefinder, the status of the range finder is set to OutOfRangeLow even if the drone height is greater than the maximum distance the rangefinder can detect, this causes the drone height to be set to 0 and the landing gear to deploy and then retract and so on infinitely. In the modified part, the value detected by the rangefinder when it goes out of the working range defined by the parameters RNGFND_MIN_CM and RNGFND_MAX_CM is ignored to take into account other values.

### DIFF
--- a/ArduCopter/landing_gear.cpp
+++ b/ArduCopter/landing_gear.cpp
@@ -18,15 +18,10 @@ void Copter::landinggear_update()
     switch (rangefinder.status_orient(ROTATION_PITCH_270)) {
     case RangeFinder::Status::NotConnected:
     case RangeFinder::Status::NoData:
-        // use altitude above home for non-functioning rangefinder
-        break;
-
-    case RangeFinder::Status::OutOfRangeLow:
-        // altitude is close to zero (gear should deploy)
-        height_cm = 0;
-        break;
-
     case RangeFinder::Status::OutOfRangeHigh:
+    case RangeFinder::Status::OutOfRangeLow:
+        // use altitude above home for non-functioning or out of range rangefinder
+        break;
     case RangeFinder::Status::Good:
         // use last good reading
         height_cm = rangefinder_state.alt_cm_filt.get();


IamPete1:
Do you have a log? Which rangefinder are you using. It sounds to me like it should be reporting `OutOfRangeHigh` not `OutOfRangeLow` if it is too high?

4dri4anxD:
Sure.
I agree it should be reporting OutOfRangeHigh, but its measurement becomes unpredictable over its maximum detection distance (which is 40 mts).
I'm using a Garmin LidarLiteV3-I2C
Landing gear is on RCOU9 on both logs
I'm attaching as well some screenshots of the graphs:

Rangefinder graph, in the middle part it flies higher than 40 meters and its measurement alternates between 0 and the actual value.
Capture
Landing gear graph (RCOU9), in deploys and retracts in a loop because it depends on the value of the rangefinder.
landing_Gear
This is the altitude of the drone (this is how rangefinder graph should look like):
altitude

After making the modifications, this is the result:
Rangefinder is the same, becoming unpredictable if flying higher than its maximum working distance:
rnffnd_corrected
 This is landing gear, it deploys and retracts only when it should, because it is not depending on Rangefinder measurements when it goes out of range:
lg_corrected
And this is the real altitude of the drone:
real_altitude

Here are the logs:

Landing gear issue:
[LandingGear_issue.zip](https://github.com/user-attachments/files/17397531/LandingGear_issue.zip)

After code changes:
[LandingGear_fixed.zip](https://github.com/user-attachments/files/17397528/LandingGear_fixed.zip)

peterbarker (commented):
This doesn't look right to me.

A RangeFinder returning out-of-range-low should definitely be deploying landing gear!

You might propose using above-home for out-of-range-high, but it seems unusual too.

4dri4anxD:
But RangeFinder is alternating between out-of-range-low and out-of-range-high when going higher than its maximum detection distance (40 mts). 
Using above-home for out-of-range-high won't solve the problem since it will keep reporting out-of-range-low erroneously deploying landing gear when it should not. 

Another solution could be only setting height_cm = 0 when reporting out-of-range-low AND height_cm is less than RangeFinder min_distance_cm

rmackay9:
Hi @4dri4anxD,

So the problem was that the rangefinder itself was unable to correctly report if it is out-of-range-low or out-of-range-high.  Some lidars are like this including, it seems the garmin.

PR #28209, opened 2024-09-24
Heli autorotation restructure mode and improve speed controller

MattKear:
This PR is the next step in bringing on the autorotation flight mode.  It builds on the work done in #27786 and is dependant upon it.  Hence, the diff looks very large at the moment.

The following changes have been made:
- A much needed re-structure to move more of the code out of the mode and down into the autorotation library
- Add pre-arm checks specific to the autorotation mode
- Improvement of the speed controller by moving to a full PID instead of just P and FF.  Also, the accel limiting that I had previously implemented was a bit weird, so I have improved that.
- Move from a stabilise style control where pilot has to do both yaw and roll control, to a lateral acceleration controller, whereby pilot requests a lateral accel component and the controller figures out the roll angle and yaw rate needed to do a co-ordinated turn.

Testing has been in Real flight only.

The majority of this PR is still behind the SITL only wall.

**Special attention should be paid to the changes in AP_MotorsHeli_Single, _Dual & _Quad** whereby I have slightly modified the way we detect the collective is on land col min.  A modification to the autorotation auto test highlighted that this flag was not tripping when collective was being constrained to land col min. **This bit is not SITL only**

PR #28420, opened 2024-10-16
Heli: Add Autorotation Flare and Touch Down Phases

MattKear:
This is a very WIP PR that I am opening for some discussion.

It incorporates the flare and touchdown phases.

I have tried to keep changes to other libraries as minimal as possible.  I may split those changes out into a separate PR.

This PR depends on #28209, and does contain the commits from that PR.

### DIFF
--- a/libraries/AC_Autorotation/AC_Autorotation.cpp
+++ b/libraries/AC_Autorotation/AC_Autorotation.cpp
@@ -115,6 +115,7 @@ AC_Autorotation::AC_Autorotation() :
         AP_Param::setup_object_defaults(this, var_info);
     }
 
+
 // Initialisation of head speed controller
 void AC_Autorotation::init_hs_controller()
 {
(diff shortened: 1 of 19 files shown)


libraries/AC_Autorotation/AC_Autorotation.cpp
```diff
@@ -2,375 +2,957 @@
...
-            _healthy_rpm_counter = 0;
-        }
+    // Estimate inflow velocity at beginning of flare
+    float entry_inflow = - speed_module * sinf(glide_angle + radians(AP_ALPHA_TPP));
+
+    const float k_1 = safe_sqrt(_hover_thrust / CR); // TODO: discuss: in the initial estimate k1 will always equal climb rate, see definition of CR above
+
+    // Protect against div by 0 case
+    if (is_zero(climb_rate + k_1)) {
+        climb_rate -= 0.05;
```
MattKear, line 664: something I noticed in the rework/simplification.  climb_rate always equals k_1 because of the way k_1 is calculated from rotor drag (CR) and hover thrust.  See lines 660 and 650.  It was less obvious before because the calc for CR was always called externally to this function.  
This means that the denominator in the calc for "a" (line 674) is always this arbitrary -0.05 value.

srmainwaring:
@MattKear I've got a version of this PR running in simulation and can help with analysis if you have time to take it forward. Leonard's unit changes in master make it time consuming to rebase all the separate commits, so I squashed them down and rebased in this branch: https://github.com/srmainwaring/ardupilot/tree/wips/pr-heli-autorotation-flare-squashed. It also includes a quick prototype for a reversible DDFP tail rotor which can be dropped.

I have not gone through all the changes in detail yet, just a high level inspection. It may be useful to be able to enable autorotation without a rangefinder (retain the capabilities of https://github.com/ArduPilot/ardupilot/pull/28209), with the additional features in this PR available if one is present?

I have not yet made a quantitative comparison of the simulated helicopter in Gazebo with standard rotor model, so the behaviour is mainly qualitative but appears reasonable.

I am seeing a pitch oscillation in the autorotation glide phase - any suggestions which controller(s) might be responsible for this? It could be that the target rotor head speed is too high, so their is insufficient headroom in the swashplate control to respond? 

Oscillations start around 20s after the rotor speed drops from 2300 rpm to around 1500 rpm which is the AROT set point.

https://vimeo.com/1101472139

MattKear:
@srmainwaring thanks for this.  Please be aware that the autorotation code is still getting constant development at the moment.  This PR does represent the latest code at this point.  I moved to a separate branch to stop constantly pushing commits to this repo and pinging everyone all of the time.

Regarding rangefinder: We need to be able to detect the ground to flare and land.  I have a TODO in the code to implement SRTM terrain data for HAGL measurement.  My key focus at the moment is to prevent users from trying to send the aircraft into an autorotation with it clearly miss-configured.  What is your use case/thoughts about doing it without a rangefinder?

Re Pitch oscillation: without seeing a log I speculate that it is because your pitch rate and attitude controllers are tuned for a head speed of 2300 rpm and then your autorotation head speed is at 1500 rpm so you are essentially flying with pitch rate gains too low in the autorotation condition.  We do not currently do gain scheduling by head speed, but this is certainly something that @bnsgeyer and want to add.

Happy to have a call if its easier to chat.

PR #23368, opened 2023-03-30
A Lua Script to provide callouts when near or over 400 feet

timtuxworth:
So many countries have a limit of 400 feet when flying RPAS/UAV (rc planes). This script provides groundstation messages which can be "called out" by QGroundControll, Mission Planner. etc. to advise the pilot what their current altitude is, warn them when approaching 400 feet (or any other set limit) and call out when they have exceeded thee limit.

This may not be required on an OpenTX/EdgeTX radio with something like Crossfire/Express LRS as the radio can be configured to do callouts and Yaapu telemetry has a max altitude alert, but EdgeTX won't provide ongoing altitude updates the way this script can do.

PR #28426, opened 2024-10-17
Scripting example to provide altitude callouts in feet defaulting to 400ft warnings

timtuxworth:
Replaces #23368 

So many countries have a limit of 400 feet when flying RPAS/UAV (rc planes). This script provides groundstation messages which can be "called out" by QGroundControl, Mission Planner. etc. to advise the pilot what their current altitude is, warn them when approaching 400 feet (or any other set limit) and call out when they have exceeded the limit.

This may not be required on an OpenTX/EdgeTX radio with something like Crossfire/Express LRS as the radio can be configured to do callouts and Yaapu telemetry has a max altitude alert, but EdgeTX won't provide ongoing altitude updates the way this script can do, and this will work with any controller, not just EdgeTX/Yaapu.

After @peterbarker  challenged me on this one, I rebased and then had to clean up/improve the example to reflect the code I'm flying now which has changed a bit since the original PR.

Hope it's still ok to merge Peter.

### DIFF
--- a/libraries/AP_Scripting/examples/plane-callout-alt.lua
+++ b/libraries/AP_Scripting/examples/plane-callout-alt.lua
@@ -1,0 +1,219 @@
+--[[
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


timtuxworth:
I messed up GitHub somehow @peterbarker so I've created this new PR for the altitude callouts, I also renamed the script and updated it based on the one I'm actually flying.

peterbarker (changes requested):
scripting linter detected a problem which needs fixing

timtuxworth:
> scripting linter detected a problem which needs fixing

Looks like it's fixed now @peterbarker

peterbarker:
Merged it - thanks!

PR #28428, opened 2024-10-18
AP_Precland::added two backend

rahulOCR:
Hey everyone,
I have experimentally found that IR_Lock is helpful in low altitude and vision based precland works fine for higher altitude but as it reaches close to aruco-marker, it lost as it goes out of FOV of camera. Hence tried both backend and found good results hence putting PR here.

I have made option to have two backend type running as per choice [condition].

### DIFF
--- a/ArduCopter/precision_landing.cpp
+++ b/ArduCopter/precision_landing.cpp
@@ -13,8 +13,8 @@ void Copter::init_precland()
 
 void Copter::update_precland()
 {
-    // alt will be unused if we pass false through as the second parameter:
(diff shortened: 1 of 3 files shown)


amilcarlucas:
If you rename "backend" to "backend1" then you do need to remove all references to "backend".
If you rename "type" to type1 then you do need to remove all references to "type".

rahulOCR:
I have not renamed the backed to backend1, instead I created two new pointer backend1 and backend2

amilcarlucas:
It is confusing to have duplicated pointers _backend and _backend1

peterbarker:
Maybe backend_high and backend_low?

We will want this functionality behind a `#ifdef`

PR #28434, opened 2024-10-19
AP_DDS: Support compile-time configurable rates for each publisher

Ryanf55:
Instead of hard-coded rates, allow devs to set the rates at compile time for their specific application. Rates are left the same as before, so this is reverse compatible.


Ryanf55:
FYI @tizianofiorenzani I thought you might like this.

srmainwaring (approved):
@Ryanf55 Looks ok - backwards compatible.

Is the intention for the override to be set in a custom hdef.dat, or on the ./waf build line? Perhaps add an example usage to the README?

Ryanf55:
> @Ryanf55 Looks ok - backwards compatible.
> 
> Is the intention for the override to be set in a custom hdef.dat, or on the ./waf build line? Perhaps add an example usage to the README?

Either will work. I have a wiki I'm working on for compile time configuration and will add specifics there.

Ryanf55:
Here's my favorite way:

Add a HWDEF file in `ardu_ws/src/ardupilot/dds_custom_rates.dat` with contents:
```bash
AP_DDS_ENABLED 0
```

Build
```bash
colcon build --packages-select ardupilot_sitl --cmake-args -DARDUPILOT_WAF_CONFIGURE_ARGS="--extra-hwdef=dds_custom_rates.dat" --event-handlers=console_direct+
```

Launch
```bash
source install/setup.bash
ros2 launch ardupilot_sitl sitl_dds_udp.launch.py
```

I also tried sim_vehicle.py
```bash
./Tools/autotest/sim_vehicle.py -v ArduPlane  --console --enable-dds --waf-configure-arg="--extra-hwdef=dds_custom_rates.dat"
```

Sadly, none of these seem to affect the build.

Ryanf55:
WIKI is here: https://ardupilot.org/dev/docs/ros2-interfaces.html#configuring-interfaces-at-compile-time

PR #28439, opened 2024-10-20
AP_DDS: Wrap services in defines

Ryanf55:
Allow disabling service interfaces at compile time.


srmainwaring (approved):
Looks fine @Ryanf55. Confirmed SITL build on macOS.

PR #28291, opened 2024-10-02
AP_DDS: battery topic to report all the available batteries

tizianofiorenzani:
# Issue

https://github.com/ArduPilot/ardupilot/issues/28288

# What Changed

- Battery topic renamed to `/ap/battery`
- All the detected batteries are reported to the same topic with different `frame_id` from `'0'` to `'<AP_BATT_MONITOR_MAX_INSTANCES>'`

# Test

```
colcon test --packages-select ardupilot_dds_tests \
--event-handlers=console_cohesion+ --pytest-args -k test_battery_msg_received

colcon test-result --all --verbose
```

PR #28441, opened 2024-10-21
AP_DDS: Set GPS instance ID in the GPS frame ID

Ryanf55:
# Purpose

Match battery behavior by setting the GPS instance number in the frame ID.

# Other changes

Remove the `0` from the topic. We are doing multiple GPS's on the same topic now.

# Demo

```bash
ros2 topic echo /ap/navsat --once
```
```yaml
header:
  stamp:
    sec: 1729471218
    nanosec: 41276000
  frame_id: '0'
status:
  status: 2
  service: 1
latitude: -35.36326217651367
longitude: 149.1652374267578
altitude: 584.0899658203125
position_covariance:
- 0.09000000357627869
- 0.0
- 0.0
- 0.0
- 0.09000000357627869
- 0.0
- 0.0
- 0.0
- 0.09000000357627869
position_covariance_type: 2
---
```

# Based on 

https://github.com/ArduPilot/ardupilot/pull/28291/files#diff-21a2f7a8c86c0d8903e0c8c3e48a7a8ae29fe5afa11878a8cd4b58224f31d94bR307

^ This does the same thing but allows two digits of instance number. I assume we will have less than 10 GPS's.

FYI @tizianofiorenzani

### DIFF
--- a/Tools/ros2/README.md
+++ b/Tools/ros2/README.md
@@ -29,7 +29,7 @@ To see all current options, use the `-s` argument:
 ros2 launch ardupilot_sitl sitl.launch.py -s
 ```
 
-#### `ardupilot_dds_test`
+#### `ardupilot_dds_tests`
 
 A `colcon` package for testing communication between `micro_ros_agent` and the
 ArduPilot `AP_DDS` client library.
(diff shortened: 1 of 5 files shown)


tizianofiorenzani:
Looks good to me.

tizianofiorenzani (approved):
Looks good to me

PR #28442, opened 2024-10-21
APM_Control: Use deceleration for stopping distance

stephendade:
For Rover, a correction for calculating the stopping distance - it should use deceleration, not acceleration.

Tested in SITL.

### DIFF
--- a/libraries/APM_Control/AR_AttitudeControl.cpp
+++ b/libraries/APM_Control/AR_AttitudeControl.cpp
@@ -1058,15 +1058,14 @@ float AR_AttitudeControl::get_desired_speed_accel_limited(float desired_speed, f
 float AR_AttitudeControl::get_stopping_distance(float speed) const
 {
     // get maximum vehicle deceleration
-    const float accel_max = get_accel_max();
+    const float decel_max = get_decel_max();
 
-    // avoid divide by zero
-    if ((accel_max <= 0.0f) || is_zero(speed)) {
+    if ((decel_max <= 0.0f) || is_zero(speed)) {
         return 0.0f;
     }
 
     // assume linear deceleration
-    return 0.5f * sq(speed) / accel_max;
+    return 0.5f * sq(speed) / decel_max;
 }
 
 // relax I terms of throttle and steering controllers


stephendade:
Turns out there's a bunch of other parts of Rover that just use acceleration instead of deceleration. Will add them to this PR...

stephendade:
Digging further into this, the s-curve navigation assumes that acceleration and deceleration are the same value.

There's some fairly scary-looking maths in the s-curve navigation that I don't think I'd be able to extend to different acceleration/deceleration values.

So I'll keep this patch to the current fixing of the APM_Control portion of the code. Any flight mode that uses s-curves will continue to use the minimum of (atc_accel_max, atc_decel_max).

Tested in SITL.

peterbarker:
This looks right, but @lthall might have words to add.

@stephendade can you explain the reasoning behind the speed_prev change, please?  It's not covered in your comments.

stephendade:
> @stephendade can you explain the reasoning behind the speed_prev change, please? It's not covered in your comments.

It appeared to be a (deliberate?) typo/bug. The original ``if (fabsf(desired_speed) < fabsf(desired_speed) && is_positive(_throttle_decel_max))`` would _always_ return ``false``, thus limiting the max speed change to ``_throttle_accel_max``.

I fixed the statement to ``if (fabsf(desired_speed) < fabsf(speed_prev) && is_positive(_throttle_decel_max))``, so the correct value of (``_throttle_decel_max``, ``_throttle_accel_max``) would be used.

stephendade:
> It appeared to be a (deliberate?) typo/bug. The original if (fabsf(desired_speed) < fabsf(desired_speed) && is_positive(_throttle_decel_max)) would always return false, thus limiting the max speed change to _throttle_accel_max.

After some testing and re-reading of the code (I missed the _desired_speed!), turns out the original code was fine, so I've removed that change.

stephendade:
And as an addedum to this, Rover uses min(accel_max, decel_max) in all the higher-level navigation (guided, auto, rtl, etc).

Would be worth clarifying this in doco

rmackay9 (approved):

PR #28443, opened 2024-10-21
AP_BattMonitor: Add the battery critical action delay time

muramura:
The battery voltage decreases along a curve.
Therefore, we want to set the action delay times for the battery warning and battery critical separately.

Screenshot from 2024-10-22 05-26-16

Screenshot from 2024-10-22 05-31-09


IamPete1:
I don't understand why you would want a different time for each in the real world?

muramura:
@IamPete1 san.
There are times when the voltage drops suddenly—for example, due to poor contact in the battery connector. In such cases, the FC (Flight Controller) may not be able to execute emergency actions within the same time frame as warnings.

I believe it's acceptable for the durations of warnings and emergencies to be separate.

Issue #28444, opened 2024-10-21
AP_DDS: Rally service interface


tizianofiorenzani:
It would be great for the Autonomy interface to command the place to set the Recovery or to Land to. That is possible by setting one or more RALLY points.

## Proposal

We could have 2 services, to `get` and `set` ardupilot by calling the `AP_Rally` functions `get_rally_point_with_index` and `set_rally_point_with_index`.

- `/ap/get_rally`, provides the rally with the requested index.
- `/ap/set_rally`, sets the rally with the provided index.

**Platform**
[  ] All
[  ] AntennaTracker
[ x ] Copter
[ x ] Plane
[ x ] Rover
[ x ] Submarine

PR #28449, opened 2024-10-22
AP_DDS: Rally Get and Set

tizianofiorenzani:
# Issue

https://github.com/ArduPilot/ardupilot/issues/28444

# What Changed

- Service `/ap/rally_set` appends a Rally Point to the list or clears the list (using the `clear` flag). Replies with:
  - `success` is the action is successfull;
  - `size` is the list's size.  
- Service `/ap/rally_get` gets the Rally Point at a given index. Returns:
  - `success`
  - `rally` the rally point as `ardupilot_msgs/Rally` type.
 - New `ardupilot_msgs/Rally` type is defined, that copies the structure definition in `AP_Rally`.

# Test
I plan to add a automatic test: For now it can be tested with SITL.

- run the SITL
- Call the service to clear the rally list `ros2 service call /ap/rally_set ardupilot_msgs/srv/RallySet "{clear: true}"`
- Open up Mavproxy with map and make sure the rally list is clear.
- Append a new Rally Point: `ros2 service call /ap/rally_set ardupilot_msgs/srv/RallySet "{rally: {point: {latitude: -35.36, longitude: 149.17, altitude: 400}}}"`
- In mavproxy, update the rally list and make sure the rally point appears on the map

image

- use the service call to get the Rally at index 0: `ros2 service call /ap/rally_get ardupilot_msgs/srv/RallyGet "index: 0"`

### DIFF
--- a/Tools/ros2/ardupilot_msgs/msg/Rally.msg
+++ b/Tools/ros2/ardupilot_msgs/msg/Rally.msg
@@ -1,0 +1,13 @@
+geographic_msgs/GeoPoint point
+float32 break_altitude # when autolanding, break out of loiter at this alt (meters)
+float32 land_dir # when the time comes to auto-land, try to land in this direction (degrees)
+
+# uint8_t flags; 
+# struct {
+#     uint8_t favorable_winds : 1; // bit 0 = seek favorable winds when choosing a landing poi
+#     uint8_t do_auto_land    : 1; // bit 1 = do auto land after arriving
+#     uint8_t alt_frame_valid : 1; // bit 2 = true if following alt frame value should be used, else Location::AltFrame::ABOVE_HOME
+#     uint8_t alt_frame       : 2; // Altitude frame following Location::AltFrame enum
+#     uint8_t unused          : 3;
+# };
+uint8 flags
--- a/Tools/ros2/ardupilot_msgs/srv/RallySet.srv
+++ b/Tools/ros2/ardupilot_msgs/srv/RallySet.srv
@@ -1,0 +1,13 @@
+
+# This service appends a Rally Point to the list or clears the whole list.
+
+# Rally Point.
+ardupilot_msgs/Rally rally
+
+# True to clear the whole list (rally entry is ignored).
+bool clear
+---
+# True if the action is successful.
+uint8 success
+# Size of Rally Points list.
+uint8 size
--- a/libraries/AP_DDS/AP_DDS_Client.cpp
+++ b/libraries/AP_DDS/AP_DDS_Client.cpp
@@ -15,9 +15,12 @@
 #include <AP_Arming/AP_Arming.h>
 #include <AP_Vehicle/AP_Vehicle.h>
 #include <AP_ExternalControl/AP_ExternalControl_config.h>
+#include <AP_Rally/AP_Rally.h>
 
 #include "ardupilot_msgs/srv/ArmMotors.h"
 #include "ardupilot_msgs/srv/ModeSwitch.h"
+#include "ardupilot_msgs/srv/RallyGet.h"
+#include "ardupilot_msgs/srv/RallySet.h"
 
 #if AP_EXTERNAL_CONTROL_ENABLED
 #include "AP_DDS_ExternalControl.h"
@@ -787,6 +790,107 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
 
         uxr_buffer_reply(uxr_session, reliable_out, replier_id, sample_id, reply_buffer, ucdr_buffer_length(&reply_ub));
         GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Request for Mode Switch : %s", msg_prefix, mode_switch_response.status ? "SUCCESS" : "FAIL");
+        break;
+    }
+    case services[to_underlying(ServiceIndex::GET_RALLY)].rep_id: {
+        ardupilot_msgs_srv_RallyGet_Request rally_get_request;
+        ardupilot_msgs_srv_RallyGet_Response rally_get_response;
+        const bool deserialize_success = ardupilot_msgs_srv_RallyGet_Request_deserialize_topic(ub, &rally_get_request);
+        if (deserialize_success == false) {
+            break;
+        }
+
+        const uxrObjectId replier_id = {
+            .id = services[to_underlying(ServiceIndex::GET_RALLY)].rep_id,
+            .type = UXR_REPLIER_ID
+        };
+
+        RallyLocation rally_location {};
+        const AP_Rally *rally = AP::rally();
+        if (rally->get_rally_point_with_index(rally_get_request.index, rally_location)) {
+            rally_get_response.success = true;
+            rally_get_response.rally.point.latitude = rally_location.lat * 1e-7;
+            rally_get_response.rally.point.longitude = rally_location.lng * 1e-7;
+            rally_get_response.rally.point.altitude = rally_location.alt;
+            rally_get_response.rally.break_altitude = rally_location.break_alt;
+            rally_get_response.rally.land_dir = rally_location.land_dir * 1e-2f;
+            rally_get_response.rally.flags = rally_location.flags;
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Request Rally index %u is valid: lat: %f  lon: %f",
+                          msg_prefix, rally_get_request.index, rally_get_response.rally.point.latitude, rally_get_response.rally.point.longitude);
+        } else {
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Request Rally index %u is invalid", msg_prefix, rally_get_request.index);
+            rally_get_response.success = false;
+        }
+
+        uint8_t reply_buffer[ardupilot_msgs_srv_RallyGet_Response_size_of_topic(&rally_get_response, 0)] {};
+        ucdrBuffer reply_ub;
+
+        ucdr_init_buffer(&reply_ub, reply_buffer, sizeof(reply_buffer));
+        const bool serialize_success = ardupilot_msgs_srv_RallyGet_Response_serialize_topic(&reply_ub, &rally_get_response);
+        if (serialize_success == false) {
+            break;
+        }
+
+        uxr_buffer_reply(uxr_session, reliable_out, replier_id, sample_id, reply_buffer, ucdr_buffer_length(&reply_ub));
+
+        break;
+    }
+    case services[to_underlying(ServiceIndex::SET_RALLY)].rep_id: {
+
+        ardupilot_msgs_srv_RallySet_Request rally_set_request;
+        ardupilot_msgs_srv_RallySet_Response rally_set_response;
+        const bool deserialize_success = ardupilot_msgs_srv_RallySet_Request_deserialize_topic(ub, &rally_set_request);
+        if (deserialize_success == false) {
+            break;
+        }
+
+        const uxrObjectId replier_id = {
+            .id = services[to_underlying(ServiceIndex::SET_RALLY)].rep_id,
+            .type = UXR_REPLIER_ID
+        };
+
+        RallyLocation rally_location{};
+        AP_Rally *rally = AP::rally();
+
+        if (rally_set_request.clear) {
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Clear Rally List", msg_prefix);
+            rally->truncate(0);
+            rally_set_response.success = rally->get_rally_total() == 0;
+        } else {
+            rally_location.lat = static_cast<int32_t>(rally_set_request.rally.point.latitude * 1e7);
+            rally_location.lng = static_cast<int32_t>(rally_set_request.rally.point.longitude * 1e7);
+            rally_location.alt = static_cast<int16_t>(rally_set_request.rally.point.altitude);
+            rally_location.break_alt = static_cast<int16_t>(rally_set_request.rally.break_altitude);
+            rally_location.land_dir = static_cast<int16_t>(rally_set_request.rally.land_dir);
+            rally_location.flags = rally_set_request.rally.flags;
+
+            if (rally_location.lat != 0 && rally_location.lng != 0) {
+                if (rally->append(rally_location)) {
+                    rally_set_response.success = true;
+                    GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Rally Point appended", msg_prefix);
+                } else {
+                    GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Failed to append Rally Point", msg_prefix);
+                    rally_set_response.success = false;
+                }
+            } else {
+                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Invalid Rally requested", msg_prefix);
+                rally_set_response.success = false;
+            }
+        }
+
+        rally_set_response.size = rally->get_rally_total();
+
+        uint8_t reply_buffer[ardupilot_msgs_srv_RallySet_Response_size_of_topic(&rally_set_response, 0)] {};
+        ucdrBuffer reply_ub;
+
+        ucdr_init_buffer(&reply_ub, reply_buffer, sizeof(reply_buffer));
+        const bool serialize_success = ardupilot_msgs_srv_RallySet_Response_serialize_topic(&reply_ub, &rally_set_response);
+        if (serialize_success == false) {
+            break;
+        }
+
+        uxr_buffer_reply(uxr_session, reliable_out, replier_id, sample_id, reply_buffer, ucdr_buffer_length(&reply_ub));
+
         break;
     }
     }
--- a/libraries/AP_DDS/AP_DDS_Service_Table.h
+++ b/libraries/AP_DDS/AP_DDS_Service_Table.h
@@ -2,7 +2,9 @@
 
 enum class ServiceIndex: uint8_t {
     ARMING_MOTORS,
-    MODE_SWITCH
+    MODE_SWITCH,
+    GET_RALLY,
+    SET_RALLY
 };
 
 static inline constexpr uint8_t to_underlying(const ServiceIndex index)
@@ -37,5 +39,25 @@ constexpr struct AP_DDS_Client::Service_table AP_DDS_Client::services[] = {
         .reply_type = "ardupilot_msgs::srv::dds_::ModeSwitch_Response_",
         .request_topic_name = "rq/ap/mode_switchRequest",
         .reply_topic_name = "rr/ap/mode_switchReply",
+    },
+    {
+        .req_id = to_underlying(ServiceIndex::GET_RALLY),
+        .rep_id = to_underlying(ServiceIndex::GET_RALLY),
+        .service_rr = Service_rr::Replier,
+        .service_name = "rs/ap/rally_getService",
+        .request_type = "ardupilot_msgs::srv::dds_::RallyGet_Request_",
+        .reply_type = "ardupilot_msgs::srv::dds_::RallyGet_Response_",
+        .request_topic_name = "rq/ap/rally_getRequest",
+        .reply_topic_name = "rr/ap/rally_getReply",
+    },
+    {
+        .req_id = to_underlying(ServiceIndex::SET_RALLY),
+        .rep_id = to_underlying(ServiceIndex::SET_RALLY),
+        .service_rr = Service_rr::Replier,
+        .service_name = "rs/ap/rally_setService",
+        .request_type = "ardupilot_msgs::srv::dds_::RallySet_Request_",
+        .reply_type = "ardupilot_msgs::srv::dds_::RallySet_Response_",
+        .request_topic_name = "rq/ap/rally_setRequest",
+        .reply_topic_name = "rr/ap/rally_setReply",
     },
 };
--- a/libraries/AP_DDS/Idl/ardupilot_msgs/msg/Rally.idl
+++ b/libraries/AP_DDS/Idl/ardupilot_msgs/msg/Rally.idl
@@ -1,0 +1,32 @@
+// generated from rosidl_adapter/resource/msg.idl.em
+// with input from ardupilot_msgs/msg/Rally.msg
+// generated code does not contain a copyright notice
+
+#include "geographic_msgs/msg/GeoPoint.idl"
+
+module ardupilot_msgs {
+  module msg {
+    struct Rally {
+      geographic_msgs::msg::GeoPoint point;
+
+      @verbatim (language="comment", text=
+        "when autolanding, break out of loiter at this alt (meters)")
+      float break_altitude;
+
+      @verbatim (language="comment", text=
+        "when the time comes to auto-land, try to land in this direction (degrees)")
+      float land_dir;
+
+      @verbatim (language="comment", text=
+        "uint8_t flags;" "\n"
+        "struct {" "\n"
+        "    uint8_t favorable_winds : 1; // bit 0 = seek favorable winds when choosing a landing poi" "\n"
+        "    uint8_t do_auto_land    : 1; // bit 1 = do auto land after arriving" "\n"
+        "    uint8_t alt_frame_valid : 1; // bit 2 = true if following alt frame value should be used, else Location::AltFrame::ABOVE_HOME" "\n"
+        "    uint8_t alt_frame       : 2; // Altitude frame following Location::AltFrame enum" "\n"
+        "    uint8_t unused          : 3;" "\n"
+        "};")
+      uint8 flags;
+    };
+  };
+};
--- a/libraries/AP_DDS/README.md
+++ b/libraries/AP_DDS/README.md
@@ -209,6 +209,8 @@ nanosec: 729410000
 $ ros2 service list
 /ap/arm_motors
 /ap/mode_switch
+/ap/rally_get
+/ap/rally_set
 ---
 ```
 
@@ -234,6 +236,8 @@ List the available services:
 $ ros2 service list -t
 /ap/arm_motors [ardupilot_msgs/srv/ArmMotors]
 /ap/mode_switch [ardupilot_msgs/srv/ModeSwitch]
+/ap/rally_get [ardupilot_msgs/srv/RallyGet]
+/ap/rally_set [ardupilot_msgs/srv/RallySet]
 ```
 
 Call the arm motors service:
@@ -255,6 +259,37 @@ requester: making request: ardupilot_msgs.srv.ModeSwitch_Request(mode=4)
 response:
 ardupilot_msgs.srv.ModeSwitch_Response(status=True, curr_mode=4)
 ```
+
+Append a Rally Point:
+
+```bash
+ros2 service call /ap/rally_set ardupilot_msgs/srv/RallySet "{rally: {point: {latitude: -35, longitude: 151, altitude: 400}}}"
+requester: making request: ardupilot_msgs.srv.RallySet_Request(rally=ardupilot_msgs.msg.Rally(point=geographic_msgs.msg.GeoPoint(latitude=-35.0, longitude=151.0, altitude=400.0), break_altitude=0.0, land_dir=0.0, flags=0), clear=False)
+
+response:
+ardupilot_msgs.srv.RallySet_Response(success=1, size=2)
+```
+
+Clear the Rally Points list:
+
+```bash
+ros2 service call /ap/rally_set ardupilot_msgs/srv/RallySet "{clear: true}"
+requester: making request: ardupilot_msgs.srv.RallySet_Request(rally=ardupilot_msgs.msg.Rally(point=geographic_msgs.msg.GeoPoint(latitude=-35.0, longitude=150.0, altitude=400.0), break_altitude=0.0, land_dir=0.0, flags=0), clear=True)
+
+response:
+ardupilot_msgs.srv.RallySet_Response(success=1, size=0)
+```
+
+Get a specific Rally Point:
+
+```bash
+ros2 service call /ap/rally_get ardupilot_msgs/srv/RallyGet "index: 0"
+requester: making request: ardupilot_msgs.srv.RallyGet_Request(index=0)
+
+response:
+ardupilot_msgs.srv.RallyGet_Response(success=True, rally=ardupilot_msgs.msg.Rally(point=geographic_msgs.msg.GeoPoint(latitude=-35.0, longitude=150.0, altitude=400.0), break_altitude=0.0, land_dir=0.0, flags=0))
+```
+
 
 ## Commanding using ROS 2 Topics
 
--- a/Tools/ros2/ardupilot_msgs/CMakeLists.txt
+++ b/Tools/ros2/ardupilot_msgs/CMakeLists.txt
@@ -6,6 +6,7 @@ project(ardupilot_msgs)
 
 find_package(ament_cmake REQUIRED)
 find_package(geometry_msgs REQUIRED)
+find_package(geographic_msgs REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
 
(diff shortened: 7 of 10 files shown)


Tools/ros2/ardupilot_msgs/msg/Rally.msg
```diff
@@ -0,0 +1,13 @@
+geographic_msgs/GeoPoint point
+float32 break_altitude # when autolanding, break out of loiter at this alt (meters)
```
Ryanf55, line 2: What if a platform doesn't fly? What should this be set to?

Tools/ros2/ardupilot_msgs/msg/Rally.msg
```diff
@@ -0,0 +1,13 @@
...
+float32 land_dir # when the time comes to auto-land, try to land in this direction (degrees)
```
Ryanf55, line 3: What coordinate system? I assume 0 is south to north, and 45 is south west to north east (compass bearing coordinates)

Tools/ros2/ardupilot_msgs/srv/RallySet.srv
```diff
@@ -0,0 +1,13 @@
...
+# Rally Point.
+ardupilot_msgs/Rally rally
+
+# True to clear the whole list (rally entry is ignored).
+bool clear
+---
+# True if the action is successful.
+uint8 success
+# Size of Rally Points list.
+uint8 size
```
Ryanf55, line 13:
```suggestion
uint8 size

```

Missing newline

Tools/ros2/ardupilot_msgs/srv/RallySet.srv
```diff
@@ -0,0 +1,13 @@
...
+# This service appends a Rally Point to the list or clears the whole list.
+
+# Rally Point.
+ardupilot_msgs/Rally rally
+
+# True to clear the whole list (rally entry is ignored).
+bool clear
+---
+# True if the action is successful.
+uint8 success
```
Ryanf55, line 11: Why is this not a bool if you call it `True` above?

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -789,6 +792,107 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+            rally_get_response.rally.point.latitude = rally_location.lat * 1e-7;
+            rally_get_response.rally.point.longitude = rally_location.lng * 1e-7;
+            rally_get_response.rally.point.altitude = rally_location.alt;
+            rally_get_response.rally.break_altitude = rally_location.break_alt;
+            rally_get_response.rally.land_dir = rally_location.land_dir * 1e-2f;
+            rally_get_response.rally.flags = rally_location.flags;
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Request Rally index %u is valid: lat: %f  lon: %f",
+                          msg_prefix, rally_get_request.index, rally_get_response.rally.point.latitude, rally_get_response.rally.point.longitude);
+        } else {
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Request Rally index %u is invalid", msg_prefix, rally_get_request.index);
```
Ryanf55, line 821: If it's invalid, should this be a warning level?

Ryanf55 (changes requested):
Nice addition, great to see some more interfaces like this.

Tools/ros2/ardupilot_msgs/msg/Rally.msg
tizianofiorenzani: that is just a copy/paste from the `AP_Rally` library. You can add rally points at all times.

Tools/ros2/ardupilot_msgs/msg/Rally.msg
tizianofiorenzani: I think I need to add more documentation from the library to the comments there.

Tools/ros2/ardupilot_msgs/srv/RallySet.srv
tizianofiorenzani: good catch, it is supposed to be a bool.

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: Definitely.

Tools/ros2/ardupilot_msgs/msg/Rally.msg
Ryanf55: That could be good. AP libraries can be a bit light on docs.

tizianofiorenzani: I should probably unpack the flags in the message, to make it easier to set from ROS.

Ryanf55: Might be nice for ROS, but I would double check the flash impact before comitting to it. 

An alternative is to supply some C++/python functions in ardupilot_msgs to unpack/pack the data.

timtuxworth:
The altitude needs a frame.

Tools/ros2/ardupilot_msgs/msg/Rally.msg
tizianofiorenzani: The altitude is respect to Home, added to the comment.

Tools/ros2/ardupilot_msgs/msg/Rally.msg
tizianofiorenzani: I have unpacked the flags to make it easier from ROS to interact with those settings.

Tools/ros2/ardupilot_msgs/srv/RallySet.srv
tizianofiorenzani: manually applied

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: Updated to Warning level.

tizianofiorenzani:
@timtuxworth I have set the altitude frame explicit now.
I don't have a clear answer for what to do with the `break_altitude`

Tools/ros2/ardupilot_msgs/msg/Rally.msg
tizianofiorenzani: @Ryanf55  Question for developers: should we keep this flag in Above home altitude (and if so, I'd like to change name of the flag to `break_altitude_above_home`) or to get the same reference as the Rally altitude (and internally convert it to Above Home Altitude)

tizianofiorenzani: from dev call: ignore this interface as it is not supported by the code.

tizianofiorenzani:
As there is no code associated with the Rally flags, the DDS interface has been limited to only the Geopoint and the altitude frame

Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_rally_point.py
```diff
@@ -0,0 +1,204 @@
...
+        assert clear_flag, "Could not clear"
+        node.send_rally(RALLY_0)
+
+        send_flag = node.add_rally_event.wait(10)
+        assert send_flag, "Could not send Rally"
+        
+        node.get_and_compare_rally(RALLY_0)
+        send_flag = node.get_rally_event.wait(10)
+        assert send_flag, "Wrong Rally point back"        
+    except Exception as e:
```
Ryanf55, line 199: Remove exception handling. Don't catch them, especially like this. It's going to swallow all issues in the code, including stuff that should fail the test.

Ryanf55 (changes requested):

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -800,6 +805,116 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+            rally->truncate(0);
+            rally_set_response.success = rally->get_rally_total() == 0;
+            if (rally_set_response.success) {
+                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Rally Point List cleared", msg_prefix);
+            }
+        } else {
+            rally_location.lat = static_cast<int32_t>(rally_set_request.rally.point.latitude * 1e7);
+            rally_location.lng = static_cast<int32_t>(rally_set_request.rally.point.longitude * 1e7);
+            rally_location.alt = static_cast<int16_t>(rally_set_request.rally.point.altitude);
+            rally_location.flags = 0;
```
Ryanf55, line 883: Is this necessary? You just constructed it as a local above.

Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_rally_point.py
tizianofiorenzani: that was a leftover of my debugging, thanks for pointing it out.

Tools/ros2/ardupilot_msgs/msg/Rally.msg
```diff
@@ -0,0 +1,16 @@
...
+
+# do auto land after arriving
+bool flag_do_auto_land
+
+# Altitude reference frame
+uint8 ALTITUDE_FRAME_ABSOLUTE = 0
+uint8 ALTITUDE_FRAME_HOME = 1
+uint8 ALTITUDE_FRAME_ORIGIN = 2
+uint8 ALTITUDE_FRAME_TERRAIN = 3
+uint8 altitude_frame
```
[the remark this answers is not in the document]
tizianofiorenzani: applied

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: Maybe not, but I like to be certain.

tizianofiorenzani:
Rebased and pushed again.

libraries/AP_DDS/AP_DDS_Service_Table.h
```diff
@@ -51,6 +55,28 @@ constexpr struct AP_DDS_Client::Service_table AP_DDS_Client::services[] = {
...
+        .reply_type = "ardupilot_msgs::srv::dds_::RallyGet_Response_",
+        .request_topic_name = "rq/ap/rally_getRequest",
+        .reply_topic_name = "rr/ap/rally_getReply",
+    },
+    {
+        .req_id = to_underlying(ServiceIndex::SET_RALLY),
+        .rep_id = to_underlying(ServiceIndex::SET_RALLY),
+        .service_rr = Service_rr::Replier,
+        .service_name = "rs/ap/rally_setService",
+        .request_type = "ardupilot_msgs::srv::dds_::RallySet_Request_",
```
timtuxworth, line 74: It's probably a good idea to construct a string like this from "ardupilot_msgs::srv::dds_::" + the request - this will save a significant amount of flash because of how many times this substring seems to be used. 
You might also want to do the same thing with "rs/ap/rally_"

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -866,6 +871,116 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+            rally_get_response.rally.point.longitude = rally_location.lng * 1e-7;
+            rally_get_response.rally.point.altitude = rally_location.alt;
+            bool alt_frame_valid = rally_location.flags & (1 << 2);
+            if (alt_frame_valid) {
+                rally_get_response.rally.altitude_frame = (rally_location.flags >> 3) & ((1 << 2) - 1);
+            } else {
+                rally_get_response.rally.altitude_frame = static_cast<uint8_t>(Location::AltFrame::ABOVE_HOME);
+            }
+            // rally_get_response.rally.flags = rally_location.flags;
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Request Rally index %u is valid: lat: %f  lon: %f",
```
timtuxworth, line 902: This is going to wrap and probably overrun the display limits on the GCS. Suggest something like
"%s Req Rally %u is valid: lat: %.2f  lon: %.2f".

libraries/AP_DDS/AP_DDS_Service_Table.h
tizianofiorenzani: that seems like a good idea. I can apply that to both service and topic file.

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: sounds good, I will shorten that.

libraries/AP_DDS/AP_DDS_Service_Table.h
tizianofiorenzani: Those are not stings, so it would make the code pretty unreadable. I would push this optimization to a dedicated PR.

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -874,6 +879,116 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+            rally_get_response.rally.point.latitude = rally_location.lat * 1e-7;
+            rally_get_response.rally.point.longitude = rally_location.lng * 1e-7;
+            rally_get_response.rally.point.altitude = rally_location.alt;
+            bool alt_frame_valid = rally_location.flags & (1 << 2);
+            if (alt_frame_valid) {
+                rally_get_response.rally.altitude_frame = (rally_location.flags >> 3) & ((1 << 2) - 1);
+            } else {
+                rally_get_response.rally.altitude_frame = static_cast<uint8_t>(Location::AltFrame::ABOVE_HOME);
+            }
+            // rally_get_response.rally.flags = rally_location.flags;
```
Ryanf55, line 909:
```suggestion
```

Remove commented code

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -874,6 +879,116 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+        };
+
+        RallyLocation rally_location{};
+        AP_Rally *rally = AP::rally();
+
+        if (rally_set_request.clear) {
+            rally->truncate(0);
+            rally_set_response.success = rally->get_rally_total() == 0;
+            if (rally_set_response.success) {
+                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "%s Rally Point List cleared", msg_prefix);
```
Ryanf55, line 951: Remove

tizianofiorenzani:
@Ryanf55 suggestions applied, rebased and tested.

tizianofiorenzani:
@Ryanf55 the expected behavior with the `clear` flag is that the rally list is  cleared and the attached waypoint is ignored. I might need to clarify that better in the readme and the service message definition.

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -944,6 +949,106 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+        RallyLocation rally_location{};
+        AP_Rally *rally = AP::rally();
+
+        if (rally_set_request.clear) {
+            rally->truncate(0);
+            rally_set_response.success = rally->get_rally_total() == 0;
+        } else {
+            rally_location.lat = static_cast<int32_t>(rally_set_request.rally.point.latitude * 1e7);
+            rally_location.lng = static_cast<int32_t>(rally_set_request.rally.point.longitude * 1e7);
+            rally_location.alt = static_cast<int16_t>(rally_set_request.rally.point.altitude);
```
timtuxworth, line 1019: We are implicitly losing precision and accuracy here. What should we do if the altitude is > 16 bits? Just silently truncate the value?

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -944,6 +949,106 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+            rally_location.flags = 0;
+            rally_location.flags |= (1 << 2); // Use the provided frame.
+            rally_location.flags |= (rally_set_request.rally.altitude_frame << 3);
+
+            if (rally_location.lat != 0 && rally_location.lng != 0
+                && rally_set_request.rally.altitude_frame <= static_cast<uint8_t>(Location::AltFrame::ABOVE_TERRAIN)) {
```
timtuxworth, line 1025: I don't understand this test for altitude_frame being valid. Could you code it in a way that's clearer to understand what you are actually doing?

libraries/AP_DDS/README.md
```diff
@@ -272,6 +276,51 @@ or
...
+requester: making request: ardupilot_msgs.srv.RallySet_Request(rally=ardupilot_msgs.msg.Rally(point=geographic_msgs.msg.GeoPoint(latitude=0.0, longitude=0.0, altitude=0.0), break_altitude=0.0, land_dir=0.0, flag_favorable_winds=False, flag_do_auto_land=False, altitude_frame=0), clear=True)
+
+response:
+ardupilot_msgs.srv.RallySet_Response(success=True, size=0)
+```
+
+Append a Rally Point:
+
+```bash
+ros2 service call /ap/rally_set ardupilot_msgs/srv/RallySet "{rally: {point: {latitude: -35, longitude: 151, altitude: 400}}}"
```
timtuxworth, line 306: It would be good to have an example with a frame supplied.

libraries/AP_DDS/README.md
```diff
@@ -272,6 +276,51 @@ or
...
+
+response:
+ardupilot_msgs.srv.RallySet_Response(success=True, size=1)
+```
+
+Get a specific Rally Point:
+
+```bash
+ros2 service call /ap/rally_get ardupilot_msgs/srv/RallyGet "index: 0"
+requester: making request: ardupilot_msgs.srv.RallyGet_Request(index=0)
```
timtuxworth, line 317: Maybe it's something I need to understand about ROS, but how would you know if there is more than one rally point, there doesn't seem to be a request rally count message, do you just keep requesting indexes until you get a fail?

libraries/AP_DDS/README.md
```diff
@@ -272,6 +276,51 @@ or
...
+
+```bash
+ros2 service call /ap/rally_set ardupilot_msgs/srv/RallySet "{clear: true}"
+requester: making request: ardupilot_msgs.srv.RallySet_Request(rally=ardupilot_msgs.msg.Rally(point=geographic_msgs.msg.GeoPoint(latitude=0.0, longitude=0.0, altitude=0.0), break_altitude=0.0, land_dir=0.0, flag_favorable_winds=False, flag_do_auto_land=False, altitude_frame=0), clear=True)
+
+response:
+ardupilot_msgs.srv.RallySet_Response(success=True, size=0)
+```
+
+Append a Rally Point:
```
timtuxworth, line 303: It would probably be good to show fractional lat/long and be clear that the altitude is meters.

libraries/AP_DDS/Idl/ardupilot_msgs/msg/Rally.idl
```diff
@@ -0,0 +1,23 @@
...
+      const uint8 ALTITUDE_FRAME_ABSOLUTE = 0;
+      const uint8 ALTITUDE_FRAME_HOME = 1;
+      const uint8 ALTITUDE_FRAME_ORIGIN = 2;
+      const uint8 ALTITUDE_FRAME_TERRAIN = 3;
+    };
+    struct Rally {
+      geographic_msgs::msg::GeoPoint point;
+
+      uint8 altitude_frame;
+    };
```
timtuxworth, line 21: Where are break_altitude, land_dir, flag_favorable_winds, flag_do_auto_land?

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -944,6 +949,106 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+            rally_get_response.success = true;
+            rally_get_response.rally.point.latitude = rally_location.lat * 1e-7;
+            rally_get_response.rally.point.longitude = rally_location.lng * 1e-7;
+            rally_get_response.rally.point.altitude = rally_location.alt;
+            bool alt_frame_valid = rally_location.flags & (1 << 2);
+            if (alt_frame_valid) {
+                rally_get_response.rally.altitude_frame = (rally_location.flags >> 3) & ((1 << 2) - 1);
+            } else {
+                rally_get_response.rally.altitude_frame = static_cast<uint8_t>(Location::AltFrame::ABOVE_HOME);
+            }
```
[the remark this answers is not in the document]
tizianofiorenzani: They are not used anywhere in the code and have been removed. We discussed about that in the dev call.

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: Location is set to be integer, I need to pass values like this, the same as you do if it was the GCS to provide it.

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: That means the altitude enum is valid. I can re write it to check for each individual value, but I don't think it makes it more readable.

libraries/AP_DDS/README.md
tizianofiorenzani: When you set a rally, it responds with the length of the Rally list. When you request a particular index, it fails if it does not exist. You can also use the service set call with an invalid point to obtain the length of the Rally list without adding any.

libraries/AP_DDS/README.md
tizianofiorenzani: Do we have to show units of measure when in SI?

libraries/AP_DDS/Idl/ardupilot_msgs/msg/Rally.idl
tizianofiorenzani: They have been removed, as not used. It was agreed in the dev call.

libraries/AP_DDS/README.md
timtuxworth: I think so because unfortunately some ArduPilot metrics are not SI (it's a work in progress).

libraries/AP_DDS/README.md
timtuxworth: An AP might have previously y been configured with zero one or more rally points. After booting, the only way to find out if there are any rally points would be to start requesting them at the first and then incrementally keep requesting until you get a fail, or to request some really large index number (what the largest valid number +1?) to get the number.  

I'm not sure I understand how that's ok.

libraries/AP_DDS/README.md
Ryanf55: I think this README needs an update now that you removed some of the flags and optional features.

Regarding units, the DDS interface follows ROS REP-103. Units are SI unless otherwise stated.

libraries/AP_DDS/README.md
Ryanf55: I agree that we should ensure this API works if there are pre-existing rally points. A sequence diagram could be helpful

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: @timtuxworth check it out now: I have explicitly compared it against every value in the enum.

tizianofiorenzani:
Rebased, ready for another round.

libraries/AP_DDS/README.md
tizianofiorenzani: @Ryanf55 the Readme has been updated.

libraries/AP_DDS/README.md
tizianofiorenzani: Would adding a field to the `set_rally` response that provides the number of rally points solve this issue?

tizianofiorenzani: @timtuxworth the `rally_get` has been extended with a `size` field. Now it returns also the size of the list.

libraries/AP_DDS/README.md
tizianofiorenzani: Readme has been updated.

tizianofiorenzani:
@Ryanf55 @timtuxworth I applied your suggestions. It should be good to go now.

libraries/AP_DDS/README.md
tizianofiorenzani: @timtuxworth Readme has been updated with fractional lat/lon

libraries/AP_DDS/README.md
```diff
@@ -274,14 +278,44 @@ or
...
+
+The `altitude_frame` specifies the datum for the altitude, among Home, Absolute, Terrain or EKF Origin
+(use the enumerator specified in the message definition).
 
 ```bash
-$ ros2 service call /ap/experimental/takeoff ardupilot_msgs/srv/Takeoff "{alt: 10.5}"
-requester: making request: ardupilot_msgs.srv.Takeoff_Request(alt=10.5)
+ros2 service call /ap/rally_set ardupilot_msgs/srv/RallySet "{rally: {altitude_frame: 1,  point: {latitude: -35.12345, longitude: 151.12345, altitude: 400.5}}}"
+
+requester: making request: ardupilot_msgs.srv.RallySet_Request(rally=ardupilot_msgs.msg.Rally(point=geographic_msgs.msg.GeoPoint(latitude=-35.12345, longitude=151.12345, altitude=400.5), altitude_frame=1), clear=False)
```
timtuxworth, line 304: At line 296 it says altitude has no decimals. Also doesn't this return a size now?

libraries/AP_DDS/README.md
timtuxworth: Good answer. Should Set also return size? Oh I see it does, so at least one example needs the size return added.

timtuxworth (changes requested):
A couple of minor documentation/example fixes. I'm confused why altitude is documented as having 0 digits (which I would question), whereas several examples who altitudes with at least 1 decimal place. For small vehicles 1 meter is a long way.

libraries/AP_DDS/README.md
tizianofiorenzani: Set does return size already

libraries/AP_DDS/README.md
tizianofiorenzani: I fixed that

timtuxworth:
LGTM - thanks for your work on this!

tizianofiorenzani:
@timtuxworth @Ryanf55 could you approve this PR?

Ryanf55 (approved):

timtuxworth (approved):
LGTM

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -980,6 +985,110 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+
+        RallyLocation rally_location {};
+        const AP_Rally *rally = AP::rally();
+        if (rally->get_rally_point_with_index(rally_get_request.index, rally_location)) {
+            rally_get_response.success = true;
+            rally_get_response.size = rally->get_rally_total();
+            rally_get_response.rally.point.latitude = rally_location.lat * 1e-7;
+            rally_get_response.rally.point.longitude = rally_location.lng * 1e-7;
+            rally_get_response.rally.point.altitude = rally_location.alt;
+            bool alt_frame_valid = rally_location.flags & (1 << 2);
```
tridge, line 1010: should be rally_location.alt_frame_valid

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -980,6 +985,110 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+            if (alt_frame_valid) {
+                rally_get_response.rally.altitude_frame = (rally_location.flags >> 3) & ((1 << 2) - 1);
```
tridge, line 1012: again, use the union fields

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -980,6 +985,110 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+
+        if (rally_set_request.clear) {
+            rally->truncate(0);
+            rally_set_response.success = rally->get_rally_total() == 0;
+        } else {
+            rally_location.lat = static_cast<int32_t>(rally_set_request.rally.point.latitude * 1e7);
+            rally_location.lng = static_cast<int32_t>(rally_set_request.rally.point.longitude * 1e7);
+            rally_location.alt = static_cast<int16_t>(rally_set_request.rally.point.altitude);
+            rally_location.flags = 0;
+            rally_location.flags |= (1 << 2); // Use the provided frame.
```
tridge, line 1058: use union fields

tridge (changes requested):
needs to use the union fields:
image

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -980,6 +985,110 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+            .id = services[to_underlying(ServiceIndex::GET_RALLY)].rep_id,
+            .type = UXR_REPLIER_ID
+        };
+
+        RallyLocation rally_location {};
+        const AP_Rally *rally = AP::rally();
+        if (rally->get_rally_point_with_index(rally_get_request.index, rally_location)) {
+            rally_get_response.success = true;
+            rally_get_response.size = rally->get_rally_total();
+            rally_get_response.rally.point.latitude = rally_location.lat * 1e-7;
```
tridge, line 1007: need to cast the 1e-7 to double or you will round to 1m, so do double(1.0e-7)
same in other places

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -980,6 +985,110 @@ void AP_DDS_Client::on_request(uxrSession* uxr_session, uxrObjectId object_id, u
...
+        };
+
+        RallyLocation rally_location{};
+        AP_Rally *rally = AP::rally();
+
+        if (rally_set_request.clear) {
+            rally->truncate(0);
+            rally_set_response.success = rally->get_rally_total() == 0;
+        } else {
+            rally_location.lat = static_cast<int32_t>(rally_set_request.rally.point.latitude * 1e7);
```
tridge, line 1054: need double(1e7)

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: updated as suggested.

libraries/AP_DDS/AP_DDS_Client.cpp
tizianofiorenzani: defined a double variable.

tizianofiorenzani:
@tridge the PR is ready for a last review.

tizianofiorenzani:
@tridge @Ryanf55 During my tests I noticed that multiplying latitude/longitude x double(1e-7) was actually less accurate than using float(1e-7). When multiplying by double(1e-7) I had errors of the order of 5e-6.

I solved by dividing by 1e7, which passes the test now and makes the service responds with the very same values I set:

```
ros2 service call /ap/rally_set ardupilot_msgs/srv/RallySet "{rally: {point: {latitude: -35.1234567, longitude: 149.1234567, altitude: 400}, altitude_frame: 1}}"
waiting for service to become available...
requester: making request: ardupilot_msgs.srv.RallySet_Request(rally=ardupilot_msgs.msg.Rally(point=geographic_msgs.msg.GeoPoint(latitude=-35.1234567, longitude=149.1234567, altitude=400.0), altitude_frame=1), clear=False)

response:
ardupilot_msgs.srv.RallySet_Response(success=True, size=1)
```
```
ros2 service call /ap/rally_get ardupilot_msgs/srv/RallyGet "index: 0"
requester: making request: ardupilot_msgs.srv.RallyGet_Request(index=0)

response:
ardupilot_msgs.srv.RallyGet_Response(success=True, size=1, rally=ardupilot_msgs.msg.Rally(point=geographic_msgs.msg.GeoPoint(latitude=-35.1234567, longitude=149.1234567, altitude=400.0), altitude_frame=1))
```

tpwrules:
The problem is that `1./1.e7f` is a slightly different number than `1.*1.e-7f`. However, `1.f/1.e7f` equals `1.f*1.e-7f`, and `1./1.e7` equals `1.*1.e-7`. The rounding is not working right because you are mixing float and double numbers.

Unfortunately, with our compiler options, I'm not sure it's possible to specify the double versions of these numbers.

We probably need to check for other round trip problems like this in the codebase, but I think merging this is fine so long as you use the two different operators and the same constant.

PR #28450, opened 2024-10-23
DroneCAN: added FlexDebug support

tridge:
This allows lua scripts to receive FlexDebug DroneCAN messages, allowing for flexible arbitrary debug data from peripheral nodes
This was developed for debugging AM32 ESCs, but any vendor can reserve ranges of IDs

### DIFF
--- a/libraries/AP_DroneCAN/AP_DroneCAN.cpp
+++ b/libraries/AP_DroneCAN/AP_DroneCAN.cpp
@@ -128,7 +128,7 @@ const AP_Param::GroupInfo AP_DroneCAN::var_info[] = {
     // @Param: OPTION
     // @DisplayName: DroneCAN options
     // @Description: Option flags
-    // @Bitmask: 0:ClearDNADatabase,1:IgnoreDNANodeConflicts,2:EnableCanfd,3:IgnoreDNANodeUnhealthy,4:SendServoAsPWM,5:SendGNSS,6:UseHimarkServo,7:HobbyWingESC,8:EnableStats
+    // @Bitmask: 0:ClearDNADatabase,1:IgnoreDNANodeConflicts,2:EnableCanfd,3:IgnoreDNANodeUnhealthy,4:SendServoAsPWM,5:SendGNSS,6:UseHimarkServo,7:HobbyWingESC,8:EnableStats,9:EnableFlexDebug
     // @User: Advanced
     AP_GROUPINFO("OPTION", 5, AP_DroneCAN, _options, 0),
     
@@ -1502,6 +1502,62 @@ bool AP_DroneCAN::is_esc_data_index_valid(const uint8_t index) {
     }
     return true;
 }
+
+#if AP_SCRIPTING_ENABLED
+/*
+  handle FlexDebug message, holding a copy locally for a lua script to access
+ */
+void AP_DroneCAN::handle_FlexDebug(const CanardRxTransfer& transfer, const dronecan_protocol_FlexDebug &msg)
+{
+    if (!option_is_set(Options::ENABLE_FLEX_DEBUG)) {
+        return;
+    }
+
+    // find an existing element in the list
+    const uint8_t source_node = transfer.source_node_id;
+    for (auto *p = flexDebug_list; p != nullptr; p = p->next) {
+        if (p->node_id == source_node && p->msg.id == msg.id) {
+            p->msg = msg;
+            p->timestamp_us = uint32_t(transfer.timestamp_usec);
+            return;
+        }
+    }
+
+    // new message ID, add to the list. Note that this gets called
+    // only from one thread, so no lock needed
+    auto *p = NEW_NOTHROW FlexDebug;
+    if (p == nullptr) {
+        return;
+    }
+    p->node_id = source_node;
+    p->msg = msg;
+    p->timestamp_us = uint32_t(transfer.timestamp_usec);
+    p->next = flexDebug_list;
+
+    // link into the list
+    flexDebug_list = p;
+}
+
+/*
+  get the last FlexDebug message from a node
+ */
+bool AP_DroneCAN::get_FlexDebug(uint8_t node_id, uint16_t msg_id, uint32_t &timestamp_us, dronecan_protocol_FlexDebug &msg) const
+{
+    for (const auto *p = flexDebug_list; p != nullptr; p = p->next) {
+        if (p->node_id == node_id && p->msg.id == msg_id) {
+            if (timestamp_us == p->timestamp_us) {
+                // stale message
+                return false;
+            }
+            timestamp_us = p->timestamp_us;
+            msg = p->msg;
+            return true;
+        }
+    }
+    return false;
+}
+
+#endif // AP_SCRIPTING_ENABLED
 
 /*
   handle LogMessage debug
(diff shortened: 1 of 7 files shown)


modules/DroneCAN/DSDL
tpwrules, line 1: I think `u8` for the data array should be like `data` or `payload`.

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -1237,4 +1237,39 @@ int lua_GCS_command_int(lua_State *L)
...
+/*
+  get FlexDebug from a DroneCAN node
+ */
+int lua_DroneCAN_get_FlexDebug(lua_State *L)
+{
+    binding_argcheck(L, 4);
+
+    const uint8_t bus = get_uint8_t(L, 1);
+    const uint8_t node_id = get_uint8_t(L, 2);
+    const uint8_t msg_id = get_uint8_t(L, 3);
```
tpwrules, line 1250:
```suggestion
    const uint16_t msg_id = get_uint16_t(L, 3);
```

libraries/AP_DroneCAN/AP_DroneCAN.cpp
```diff
@@ -1503,6 +1503,58 @@ bool AP_DroneCAN::is_esc_data_index_valid(const uint8_t index) {
...
+        if (p->node_id == source_node && p->msg.id == msg.id) {
+            p->msg = msg;
+            p->timestamp_us = uint32_t(transfer.timestamp_usec);
+            return;
+        }
+    }
+
+    // new message ID, add to the list. Note that this gets called
+    // only from one thread, so no lock needed
+    auto *p = NEW_NOTHROW FlexDebug;
```
tpwrules, line 1524: Concerned that DroneCAN devices can cause arbitrary amounts of flight controller memory to be consumed that is never released. I think there should be a Lua call that allocates a fixed size buffer (or at least sets a limit on the number of allowed allocations) like with the scripting CAN buffers.

libraries/AP_DroneCAN/AP_DroneCAN.cpp
```diff
@@ -1503,6 +1503,58 @@ bool AP_DroneCAN::is_esc_data_index_valid(const uint8_t index) {
...
+  get the last FlexDebug message from a node
+ */
+bool AP_DroneCAN::get_FlexDebug(uint8_t node_id, uint8_t msg_id, uint32_t &timestamp_us, dronecan_protocol_FlexDebug &msg) const
+{
+    for (const auto *p = flexDebug_list; p != nullptr; p = p->next) {
+        if (p->node_id == node_id && p->msg.id == msg_id) {
+            if (timestamp_us == p->timestamp_us) {
+                // stale message
+                return false;
+            }
```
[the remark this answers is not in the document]
tridge: we try to minimise the amount of memory allocation we do in flight, allocating on every packet would lead to possibly nasty memory allocation

libraries/AP_DroneCAN/AP_DroneCAN.h
```diff
@@ -409,6 +418,16 @@ class AP_DroneCAN : public AP_CANDriver, public AP_ESC_Telem_Backend {
     void handle_param_get_set_response(const CanardRxTransfer& transfer, const uavcan_protocol_param_GetSetResponse& rsp);
     void handle_param_save_response(const CanardRxTransfer& transfer, const uavcan_protocol_param_ExecuteOpcodeResponse& rsp);
     void handle_node_info_request(const CanardRxTransfer& transfer, const uavcan_protocol_GetNodeInfoRequest& req);
+
+#if AP_SCRIPTING_ENABLED
+    void handle_FlexDebug(const CanardRxTransfer& transfer, const dronecan_protocol_FlexDebug& msg);
+    struct FlexDebug {
+        struct FlexDebug *next;
+        uint64_t timestamp_us;
```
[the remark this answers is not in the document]
tridge: fixed, thanks

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -1237,4 +1237,39 @@ int lua_GCS_command_int(lua_State *L)
...
+        return 0;
+    }
+
+    luaL_Buffer b;
+
+    uint8_t *data = (uint8_t *)luaL_buffinitsize(L, &b, msg.u8.len);
+    memcpy(data, msg.u8.data, msg.u8.len);
+
+    *new_uint32_t(L) = tstamp_us;
+    luaL_pushresultsize(&b, msg.u8.len);
```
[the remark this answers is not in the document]
tridge: thanks! fixed

libraries/AP_DroneCAN/AP_DroneCAN.cpp
tridge: I don't think it will be a problem in practice, as amount of memory for a scripting capable flight controller isn't much
we could reduce the ram by only allocating up to the actual length of the message

libraries/AP_Scripting/lua_bindings.cpp
tridge: fixed

modules/DroneCAN/DSDL
tridge: i'm not really worried about that name

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -1237,4 +1237,34 @@ int lua_GCS_command_int(lua_State *L)
...
+  get FlexDebug from a DroneCAN node
+ */
+int lua_DroneCAN_get_FlexDebug(lua_State *L)
+{
+    binding_argcheck(L, 4);
+
+    const uint8_t bus = get_uint8_t(L, 1);
+    const uint8_t node_id = get_uint8_t(L, 2);
+    const uint16_t msg_id = get_uint16_t(L, 3);
+    uint32_t tstamp_us = get_uint32(L, 4, 0, UINT32_MAX);
```
IamPete1, line 1251:
```suggestion
    uint32_t tstamp_us = coerce_to_uint32_t(L, 4);
```
Don't need the range check.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -414,6 +414,16 @@ function CAN:get_device(buffer_len) end
...
 
+
+-- get latest FlexDebug message from a CAN node
+---@param bus number -- CAN bus number, 0 for first bus, 1 for 2nd
+---@param node number -- CAN node
+---@param id number -- FlexDebug message ID
+---@param last_us uint32_t_ud|integer|number -- timestamp of last received message, new message will be returned if timestamp is different
+---@return uint32_t_ud|nil -- timestamp of message (first frame arrival time)
+---@return string|nil -- up to 255 byte buffer
+function DroneCAN_get_FlexDebug(bus,node,id,last_us) end
```
IamPete1, line 425: Should we make a `DroneCAN` singleton? I think it makes sense if we would ever want another method.

libraries/AP_DroneCAN/AP_DroneCAN.cpp
tpwrules: Fixed by adding the option bit so it's not possible for this to cause problems with ordinary users.

libraries/AP_Scripting/docs/docs.lua
tpwrules: Might get changed in the future, understanding in the dev call is that we have the liberty to change this as it's a special debugging thing.

PR #28453, opened 2024-10-23
ArduPlane: clean up guided heading type initialization

timtuxworth:
CMD_GUIDED_CHANGE_HEADING was initializing heading values even if not using HEADING_TYPE_HEADING.


peterbarker:
Could you explain why https://github.com/ardupilot/ardupilot/blob/master/ArduPlane/mode_guided.cpp#L65 is not crossed when using cog, please?

Some broad explanation in here of why you believe this is unused would be nice :-)

timtuxworth:
> Could you explain why https://github.com/ardupilot/ardupilot/blob/master/ArduPlane/mode_guided.cpp#L65 is not crossed when using cog, please?

It seems like it would be crossed. I'm not sure why it matters, this PR doesn't change plane.guided_state.target_heading_limit

timtuxworth:
fixed

timtuxworth:
This change is wrong.

PR #28455, opened 2024-10-23
AP_Mount: Changed the type of gimbal angle acquisition and converted …

laozhou-fujian:
…the lowercase characters in the command to uppercase

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Topotek.cpp
+++ b/libraries/AP_Mount/AP_Mount_Topotek.cpp
@@ -31,7 +31,7 @@ extern const AP_HAL::HAL& hal;
 # define AP_MOUNT_TOPOTEK_ID3CHAR_START_TRACKING "LOC"          // start image tracking
 # define AP_MOUNT_TOPOTEK_ID3CHAR_LRF           "LRF"           // laser rangefinder control, data bytes: 00:ranging stop, 01:ranging start, 02:single measurement, 03:continuous measurement
 # define AP_MOUNT_TOPOTEK_ID3CHAR_PIP           "PIP"           // set picture-in-picture setting, data bytes: // 00:main only, 01:main+sub, 02:sub+main, 03:sub only, 0A:next
-# define AP_MOUNT_TOPOTEK_ID3CHAR_GIMBAL_ATT    "GAA"           // get gimbal attitude, data bytes: 00:stop, 01:start
+# define AP_MOUNT_TOPOTEK_ID3CHAR_GIMBAL_ATT    "GIA"           // get gimbal attitude, data bytes: 00:stop, 01:start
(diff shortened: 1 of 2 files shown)


peterbarker (changes requested):
Does this break compatibility with older mount firmwares?

Should be a single commit, not three....

laozhou-fujian:
> Does this break compatibility with older mount firmwares?
> 
> Should be a single commit, not three....

This will not compromise compatibility; it simply changes the attitude information transmitted from the gimbal to the flight controller from the previous internal gimbal frame angles to pitch and roll angles relative to the ground, while the heading remains as the previous internal gimbal frame angle. This is because the flight controller controls the gimbal using pitch and roll angles relative to the ground.

Since most of the data bits in the commands we send to control the gimbal are uppercase letters and numbers, although lowercase letters are also compatible, we prefer to standardize them as uppercase letters.

Previously, the reason for repeated submissions was that the automatic testing verification did not pass, but this will not happen again in the future.

peterbarker:
> > Does this break compatibility with older mount firmwares?
> This will not compromise compatibility; it simply changes the attitude information transmitted from the gimbal to the flight controller from the previous internal gimbal frame angles to pitch and roll angles relative to the ground, while the heading remains as the previous internal gimbal frame angle. This is because the flight controller controls the gimbal using pitch and roll angles relative to the ground.

Great - so older gimbal firmwares can send in either frame, which is what I was asking about :-)

peterbarker:
I've fixed the commit message and marked for DevCall.

timtuxworth:
> Previously, the reason for repeated submissions was that the automatic testing verification did not pass, but this will not happen again in the future.

It's ok to submit multiple times until the PR is acceptable @laozhou-fujian , but you should "squash" the individual commits into one commit which keeps the GitHub history clean. https://ardupilot.org/dev/docs/git-interactive-rebase.html

laozhou-fujian:
> > Previously, the reason for repeated submissions was that the automatic testing verification did not pass, but this will not happen again in the future.
> 
> It's ok to submit multiple times until the PR is acceptable @laozhou-fujian , but you should "squash" the individual commits into one commit which keeps the GitHub history clean. https://ardupilot.org/dev/docs/git-interactive-rebase.html

Okay, understood, thank you

rmackay9:
Great to see this!  support for this gimbal is new for 4.6 and I have that I can also use to help test.

I think we should merge this

peterbarker:
@rmackay9 sooo... did you want to test this before merging?  I'd intended to just merge it....

rmackay9:
yes, I'll test this..

rmackay9 (approved):
I've tested this and it works fine, thanks!

PR #28456, opened 2024-10-23
AP_HAL_Linux: Enforce line buffering

landswellsong:
Helpful when run via systemd or other tool that captures stdout for logging purpose
Similar code already exists in SITL

### DIFF
--- a/libraries/AP_HAL_Linux/HAL_Linux_Class.cpp
+++ b/libraries/AP_HAL_Linux/HAL_Linux_Class.cpp
@@ -340,6 +340,10 @@ void HAL_Linux::run(int argc, char* const argv[], Callbacks* callbacks) const
         CMDLINE_SERIAL9,
     };
 
+    // set explicit line buffering for e.g. logging via systemd
+    setvbuf(stdout, (char *)0, _IOLBF, 0);
+    setvbuf(stderr, (char *)0, _IOLBF, 0);
+
     int opt;
     const struct GetOptLong::option options[] = {
         {"uartA",         true,  0, 'A'},


landswellsong:
Kindly tagging @lucasdemarchi for the review (since it's Linux)

libraries/AP_HAL_Linux/HAL_Linux_Class.cpp
```diff
@@ -361,6 +361,10 @@ void HAL_Linux::run(int argc, char* const argv[], Callbacks* callbacks) const
         CMDLINE_SERIAL9,
     };
 
+    // set explicit line buffering for e.g. logging via systemd
```
lucasdemarchi, line 364: systemd should already be setting the buffer depending on where it's sending stdout/stderr to. I'm confused what this is fixing.

libraries/AP_HAL_Linux/HAL_Linux_Class.cpp
```diff
@@ -361,6 +361,10 @@ void HAL_Linux::run(int argc, char* const argv[], Callbacks* callbacks) const
...
+    setvbuf(stdout, (char *)0, _IOLBF, 0);
+    setvbuf(stderr, (char *)0, _IOLBF, 0);
```
lucasdemarchi, line 366: NULL or nullptr could be used.

lucasdemarchi (changes requested):

libraries/AP_HAL_Linux/HAL_Linux_Class.cpp
landswellsong: I do not think it does. By default it's sent to `journal` and it's the executable's responsibility to set the buffering. Since `journal` pipes the streams across, the libc runtime does an `isatty()` check somewhere deep in itself, determines that the output isn't a TTY and thus switches full buffering on. This means that systemd doesn't get any logs until a relatively large buffer fills up.

At least this is my understanding of how systemd works maybe I'm missing something, but I hit this in numerous projects and the solution was always the same.

We can go with line buffering here too, but I'm carrying over the logic from other parts of ArduPilot,

libraries/AP_HAL_Linux/HAL_Linux_Class.cpp
landswellsong: It's just copied from elsewhere to keep things uniform:
1. https://github.com/ArduPilot/ardupilot/blob/f0900bd119562dffe4cafabc1d2903e6951c128f/libraries/AP_HAL_SITL/SITL_cmdline.cpp#L369
2. https://github.com/ArduPilot/ardupilot/blob/f0900bd119562dffe4cafabc1d2903e6951c128f/libraries/AP_HAL_SITL/SITL_Periph_State.cpp#L63

Should I change to `nullptr` in those instances as well? I also am a little unsure if we really need unbuffered output or maybe line buffering would suffice. It definitely would for HAL Linux case, but with SITL it's not super clear as it may be doing some printing without a line feed which I'm not aware of.

libraries/AP_HAL_Linux/HAL_Linux_Class.cpp
lucasdemarchi: I'm not sure if ardupilot is closing/opening the fd given or something like that... I will have to double check.

It  always worked for me as line-buffered in other projects using stdout/stderr. Just tested this:

```
$ cat <<EOF > cat foo.sh 
#!/bin/bash

echo foo
sleep 10
echo -n bar
sleep 10
echo
echo foobar
$ systemd-run --wait -u logtest --user ./foo.sh
```

In the other terminal I was running:
```
journalctl --user -e -u logtest --follow
```

It shows the expected behavior: "foo" is print right away. "bar" is not (since echo -n omits the newline). And after another 10s we have both "bar" and "foobar". Maybe it's something in your service config? Can you share it?

landswellsong: Try this:
```C
#include <unistd.h>
#include <stdio.h>

int main() {
    setvbuf(stdout, (char *)0, _IOLBF, 0);
    printf("Hello\n");
    sleep(10);
    printf("cruel\n");
    sleep(10);
    printf("world\n");
    return 0;
}
```

And compile it with and without the `setvbuf` call. 

Here's what I'm getting in exact setup as you've mentioned:
```
$ journalctl --user -e -u logtest --follow
Nov 07 19:25:46 hobontica systemd[912]: Started logtest.service.
Nov 07 19:26:06 hobontica a.out[445559]: Hello
Nov 07 19:26:06 hobontica a.out[445559]: cruel
Nov 07 19:26:06 hobontica a.out[445559]: world

Nov 07 19:26:37 hobontica systemd[912]: Started logtest.service.
Nov 07 19:26:37 hobontica a.out[445833]: Hello
Nov 07 19:26:47 hobontica a.out[445833]: cruel
Nov 07 19:26:57 hobontica a.out[445833]: world
```

First run is without `setvbuf`, second run is with it. I would think that your BASH example has some internal `setvbuf` logic to it which differs from what stdlib does by default.

lucasdemarchi: Ok, and this is why did I never have this problem in other projects (e.g. https://github.com/mavlink-router/mavlink-router):  because I normally use stderr for logging, not stdout. Using your snippet like this:

```
#include <unistd.h>
#include <stdio.h>

int main() {
//    setvbuf(stdout, (char *)0, _IOLBF, 0);
    fprintf(stderr, "Hello\n");
    sleep(10);
    fprintf(stderr, "cruel\n");
    sleep(10);
    fprintf(stderr, "world\n");
    return 0;
}
```

Sorry for the back and forth.... Now I think what you're doing here does make sense.

lucasdemarchi (approved):

libraries/AP_HAL_Linux/HAL_Linux_Class.cpp
landswellsong: Heh, this is shaping to be an amazing C interview question :D

landswellsong:
Rebased over 4.7. @lucasdemarchi can we get this merged?

PR #28459, opened 2024-10-24
AP_HAL_ESP32: Add custom SDMMC configuration

sblaksono:
Add custom configuration for SDMMC slot. In my case, I'm using Freenove ESP32S3 board which has onboard SD card slot and different pin configuration.


davidbuzz:
this is not a correct PR, as it might work now for forr your specific board, but you've broken it for everyone else who is using the normal pinout.     i think it needs a else block that has the original pin definitions or something like that.

sblaksono:
The current SDMMC code hardcoded everything, even for pullups by GPIO pin number.  I change them by using existing slot_config variable and change it assignment to a new HAL_ESP32_SDMMC_SLOT_CONFIG so we can define it in board configuration or using SDMMC_SLOT_CONFIG_DEFAULT if not defined. 

Here is my configuration:

#define HAL_ESP32_SDMMC_SLOT_CONFIG() {\
    .clk = GPIO_NUM_39, \
    .cmd = GPIO_NUM_38, \
    .d0 = GPIO_NUM_40, \
    .d1 = GPIO_NUM_NC, \
    .d2 = GPIO_NUM_NC, \
    .d3 = GPIO_NUM_NC, \
    .d4 = GPIO_NUM_NC, \
    .d5 = GPIO_NUM_NC, \
    .d6 = GPIO_NUM_NC, \
    .d7 = GPIO_NUM_NC, \
    .cd = GPIO_NUM_NC, \
    .wp = GPIO_NUM_NC, \
    .width = 1, \
    .flags = 0, \
}

Try to compile your esp32buzz board also and it can be compiled. Not tested because I don't have other board which has an SD card slot except that s3 board :)

davidbuzz:
If this requires changes to the other board.h file/s as you describe,then this PR should also include those changes to those (eg changes to esp32buzz.h ) so that it continues to work as before.

sblaksono:
I found the problem, the default configuration for esp32 needs the harcoded pullups because they are not defined in default config. I've added if else block to preserve that. Please check

PR #28460, opened 2024-10-24
AP_HAL_ESP32: Switch WIFI task from FASTCPU to SLOWCPU

sblaksono:
Switching WIFI task from FASTCPU to SLOWCPU seems to bring more balance between CPUs and thus increasing connection reliabiiity


tpwrules (approved):
Flew and it seems to work well. I noticed faster parameter downloads as well.

davidbuzz (approved):
looks good to me if tested ok

peterbarker:
Well and truly flown

PR #28463, opened 2024-10-24
Real-time data logging

amitggadag:
Logging all rotor RPMs and quaternions for research.


peterbarker:
@amitggadag this one needs a rebase

It seems that at least some of this is already done in master - logging of more than 2 RPMS.

Are you going to rebase this on master?

andyp1per:
Right, this is not doing what he thinks it is doing

peterbarker:
OK, so this is conflicting and it's unclear what the purpose was.

Please re-open if there's still worth here, @amitggadag

PR #28464, opened 2024-10-24
AP_ICEngine: If surfaces are disabled do not allow a disarmed start

WickedShell:
We don't want to allow AP_ICEngine to start an engine while disarmed if the surfaces are locked, as in this case we should have little to no control over the engine. This is particularly evident when using relay control for starter/ignition control, as it's now easy to enable both of these without having active throttle control.

Bench tested with relay outputs from AP_ICE, with a relay control for ignition power, and a scope connected to the throttle signal.

### DIFF
--- a/libraries/AP_ICEngine/AP_ICEngine.cpp
+++ b/libraries/AP_ICEngine/AP_ICEngine.cpp
@@ -357,7 +357,8 @@ void AP_ICEngine::update(void)
             allow_single_start_while_disarmed = false;
         } else {
             // check if we are blocking disarmed starts
-            if (!allow_single_start_while_disarmed) {
+            if ((hal.util->safety_switch_state() == AP_HAL::Util::SAFETY_DISARMED) ||
+                !allow_single_start_while_disarmed) {
                 should_run = false;
             }
         }


libraries/AP_ICEngine/AP_ICEngine.cpp
```diff
@@ -357,7 +357,8 @@ void AP_ICEngine::update(void)
             allow_single_start_while_disarmed = false;
         } else {
             // check if we are blocking disarmed starts
-            if (!allow_single_start_while_disarmed) {
+            if ((hal.util->safety_switch_state() == AP_HAL::Util::SAFETY_DISARMED) ||
```
IamPete1, line 360: Won't this still be a issue if `NO_RUNNING_WHILE_DISARMED` is not set?

tridge, line 360: a fairly common setup is to have safety on, but allow control surfaces or throttle though with BRD_SAFETY_MASK, this change looks like it would break that workflow

tridge (changes requested):

libraries/AP_ICEngine/AP_ICEngine.cpp
WickedShell: The intention is to leverage the allow a disarmed start MAVLink command for those cases. That is actually how I'm using it.

The problem I'm trying to fix is I managed to start an engine and have 0 control of the throttle, which is a bad situation as well. The surfaces were not enabled, and I had a running engine with no throttle control. We need a solution that protects people in this case. I can add another option bit somewhere, but I think the default expectation would be this.

WickedShell: I mean yes, but presumably you at least requested that? I think it's still safer then otherwise.

PR #28465, opened 2024-10-24
AP_DDS: Change the comparison of float value zero to IS_XXXX

muramura:
I want to use IS_POSITIVE and IS_NEGATIVE instead of comparing with a float value of zero.

### DIFF
--- a/libraries/AP_DDS/AP_DDS_Client.cpp
+++ b/libraries/AP_DDS/AP_DDS_Client.cpp
@@ -347,9 +347,9 @@ void AP_DDS_Client::update_topic(sensor_msgs_msg_BatteryState& msg, const uint8_
     if (battery.current_amps(current, instance)) {
         if (percentage == 100) {
             msg.power_supply_status = 4;   //POWER_SUPPLY_STATUS_FULL
-        } else if (current < 0.0) {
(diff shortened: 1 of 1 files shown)


libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -340,9 +340,9 @@ void AP_DDS_Client::update_topic(sensor_msgs_msg_BatteryState& msg, const uint8_
     if (battery.current_amps(current, instance)) {
         if (percentage == 100) {
             msg.power_supply_status = 4;   //POWER_SUPPLY_STATUS_FULL
-        } else if (current < 0.0) {
+        } else if (is_negative(current)) {  // TODO
```
Ryanf55, line 343:
```suggestion
        } else if (is_negative(current)) {
```

What is the TODO, this looks like a good change to me

Ryanf55 (commented):
Thank you for the contribution

tpwrules:
Note that this is not exactly the same now, it will compare with approximately 1e-7 instead of 0. Should be fine for this application though.

peterbarker:
Fixed the commit list, marking as MergeOnCIPass - thanks!

PR #28468, opened 2024-10-25
HAL_QURT: build deb packages in CI

tridge:
and fixed make_package.sh to not use sudo, which is not allowed on the build server
ping @katzfey


katzfey:
Yep, looks good, thanks!!!

PR #28470, opened 2024-10-25
Fix metadata

amilcarlucas:


Hwurzburg:
what functional impact does this have?

peterbarker (approved):
LGTM

peterbarker:
Marking this for DevCall to make sure people are OK with the concept of a "range" for the main loop rate.

@Hwurzburg these obviously don't affect the flight code - but will be important for MethodicConfigurator and other places.  It may affect the interface in MissionPlanner poorly for the range check - perhaps a slider rather than a dropdown for the main loop rate.  @amilcarlucas have you checked that?

amilcarlucas:
I did not test MP, but it will improve the Methodic configurator GUI.
And this is a better fix than adding extra exception code on the GCS side.

tridge (approved):

rmackay9 (approved):
we could increase the LOOP_RATE range to 800 but this is fine, txs!

PR #28472, opened 2024-10-25
Plane: Add a feed forward from flaps to pitch

WickedShell:
The intent is to approximate the effect of what is typically done on the RC transmitter. While the integrator will typically catch up to this, it still is a reactive system rather then proactive, and can have negative impacts on things like servo auto trim.

This is currently only applied to auto flaps, and thus autopilot controlled modes. It's done within APM_Control as this allows the integrator to still wind up and allow us to use full surface deflection if needed, something that can be lost when just doing a direct mix, or applying the compensation after calculating an elevator command.

I originally did this outside of APM_Control and it also works there, however the loss of servo travel range makes it much easier to get into an accident, and doesn't feel great. You can view that version [here](https://github.com/WickedShell/ardupilot/tree/wickedshell/kff-flap) if you prefer.

Tagged as a draft pending real flight testing next week.


libraries/APM_Control/AP_PitchController.cpp
```diff
@@ -199,10 +208,13 @@ float AP_PitchController::_get_rate_out(float desired_rate, float scaler, bool d
         rate_pid.set_integrator(old_I);
     }
 
+    // calculate feed forward for auto flaps
+    const float scaled_ff_flap = (SRV_Channels::get_output_norm(SRV_Channel::k_flap_auto) + 1) * _flap_ff * scaler * 0.5f;
```
IamPete1, line 212: Might need to use  `get_slew_limited_output_scaled` (with the right maths for scale.)

IamPete1 (commented):
One issue is that it won't apply in manual, so you will get a step change in elevator output when you switch into and out of manual. Having it work in manual should make it easier for those more used to doing it in transmitter.

libraries/APM_Control/AP_PitchController.cpp
WickedShell: I'd actually added: `SRV_Channel::get_slew_limited_output_norm()` in this PR for this exact purpose, but apparently pushed the one not using it (I was switching back and forth in testing).

I wanted to avoid having to do the extra scaling here, because it makes it harder to handle changing the scaling in the future.

WickedShell:
> One issue is that it won't apply in manual, so you will get a step change in elevator output when you switch into and out of manual. Having it work in manual should make it easier for those more used to doing it in transmitter.

So I agree, but there's also an interesting thing here. The only way it worked well before was to do it as a mix on the transmitter that was aware of the flightmode switch, because if you applied it in say FBWA it would be an increase in demanded pitch, causing the aircraft to be even more off of where it wants to be.

The problem I'm running into is the way this is normally mixed on the transmitter tends to lead to reducing throws in one direction (or having asymmetric stick response), neither of which is great, albeit tolerable. The biggest win of doing it inside the pitch controller is the vehicle is still allowed full elevator travel if it happens to need it. It's also nice to pick up some speed scaling out of this as well.

I'm very open to suggestions to get it applied in manual as well, as long as we can make it not limit the travel, but I've been struggling to come up with something that covers both bases equally well. (There's also a degenerate case of a mix of auto flaps + manually controlled flaps, and I've chosen to only look at the auto ones).

I know I'm seeming very hung up on this full travel comment, but I have a couple aircraft we work with that really want to use that last bit of travel to flare at the end at slow speeds where we need more stick input, and the normal transmitter mix tends to make this hard as we lose a bit of travel (even if it was otherwise great for general flight) in manual.

WickedShell:
@IamPete1 I've done a seperate patch to let you have direct flap to pitch mix in manual. It means the user has 2 parameters to set up, but it does work well. In manual you lose some travel, but that's fairly typical, and in anything using a pitch controller you can get the benefit here. I'll get that PR'd shortly.

LupusTheCanine:
You could only adjust the center point. This would require that pitch trim to be fairly close at low speeds but would preserve throws, fading it only above 50% would mitigate this limitation at a cost of a bit less linear response.

libraries/APM_Control/AP_PitchController.cpp
```diff
@@ -158,6 +159,15 @@ const AP_Param::GroupInfo AP_PitchController::var_info[] = {
 
     AP_SUBGROUPINFO(rate_pid, "_RATE_", 11, AP_PitchController, AC_PID),
 
+    // @Param: 2SRV_FLAP
+    // @DisplayName: Flap compensation
+    // @Description: Gain added to pitch to keep aircraft from descending or ascending when deploying flaps. Decrease to compensate for pitching up when deploying flaps, increase if the vehicle pitches down when deploying flaps.
```
tridge, line 164: I think what it is actually doing is changing the elevator trip with flaps. I think this could be done a lot more simply in Plane::stabilize_pitch() with the parameter in the main plane code.

tridge (changes requested):

PR #28473, opened 2024-10-26
AP_Mission: exclude more code based on defines

peterbarker:


IamPete1:
I think if someone has a mission with a command which is being removed here the mission will still work, just the command will be skipped?

tridge:
@peterbarker i think we do need to know what happens with an unsupported mission item - user can presumably still upload the item? It is probably fine that we just skip unsupported items if we can confirm that behaviour

peterbarker:
Because the code used to pack/unpack the stored "Mission_Command" to/from "mavlink_mission_item_int" goes away, transfer of the mission to/from the autopilot no longer works.

We could probably change this to convert to an object which is essentially empty except for the command identifier?

We should probably also include a pre-arm check that we don't try to execute a mission which includes an item we don't actually know how to execute...

tridge (changes requested):
i think we need to do 2 things:
 - work out what MissionPlanner and QGC does with bad mission download
 - add an arming check (only scan once, remember scan result) that checks all items are valid

PR #28474, opened 2024-10-26
add RTL as a short failsafe option

peterbarker:
This PR adds the ability to enter RTL as a short-failsafe action.  From the request:

```
We still want both a short and a long failsafe action. To clarify, this is what we want to do:
When connection is lost continue for X seconds
After X seconds: RTL
If connection still not recovered after Y seconds: Glide
If FS_Short could do RTL, it could take care of #1&2, and FS_Long could take care of #3
```


tridge:
I suspect the request could be satisfied with https://ardupilot.org/plane/docs/advanced-failsafe-configuration.html or a lua script
ping @bengterik

bengterik:
From the docs:
> This makes the AFS options only really appropriate for AUTO missions. If you are primarily flying Plane in CRUISE mode or other modes then you should use the standard failsafe options.

Since we are primarily flying in GUIDED mode (except for takeoff and landing), these AFS options may not be applicable.

tridge:
@bengterik I think this really seems like something that is very specific to your use case, and the logic for not losing state on the RTL so it can go back to what it was doing on recovery of is quite complex.
In lua for GCS failsafe detection you can use gcs:last_seen(), which I think gives you what you need to implement the required behaviour. Peter or I could put together an example script if that would be helpful

tridge:
@bengterik if you would like us to write a script, please let us know what failsafe types are needed (GCS, R/C, battery etc)

bengterik:
Thank you for having a look at this again!

If it simplifies implementation it is not necessary for the mission to continue from where it was should connection be regained between FS_SHORT and FS_LONG. It can continue in RTL until the pilot enters a new mission command, just like when FS_SHORT is set to CIRCLE.

If it remains problematic we'll go the scripting route instead!

PR #28475, opened 2024-10-26
Fix clang compiler warning

katzfey:
Fix for compiler warning from clang about unknown warning group -Wswitch-unreachable. This warning is prevalent in both Qurt and MacOS SITL builds.


peterbarker (approved):
LGTM.

I'm assuming this doesn't actually introduce the warning the pragma is meant to fix :-)

PR #28478, opened 2024-10-27
AP_RCTelemetry: Fix Baro and Vario values

Frank0587:
Add the missing byte swapping for 16bit values


peterbarker:
@andyp1per this is pretty recent code - has to be one or the other, how did you test the original code?

andyp1per:
> @andyp1per this is pretty recent code - has to be one or the other, how did you test the original code?

It requires the TX to specifically understand the encoding, so it was tested with great difficulty by others. I'm quite prepared to believe the encoding is still wrong, but would need to see evidence of success in order to approve

Frank0587:
I had used the original code on my plane and got always about -586,5m instead of 0 as altitude. As I calculate the values back to hex and look at the formula used at elrs and opentx, i found the wrong byteorder. Compare the code in AP between baro and vario against other telemetry values, i found the missing byte swapping. I test the new code on my plane, and the telemetry values looks now like expected.

andyp1per:
@Frank0587 you need to squash your commits

andyp1per (approved):
LGTM

PR #26328, opened 2024-02-26
Plane: handle guided path requests in GUIDED mode

srmainwaring:
Add a new method `handle_guided_path_request` to the plane GUIDED mode and extra update function `update_path ` in `AP_Navigation`. A preliminary implementation is provided for `AP_L1_Control`. The guided path mode is used to handle `SET_POSITION_TARGET_GLOBAL_INT` mavlink commands.

This is a simplified version of:

- https://github.com/ArduPilot/ardupilot/pull/26262

It excludes the NPFG controller and modifies the existing `ModeGuided` rather than introduce a new mode.

### Details

- The implementation follows the approach used in Rover's `ModeGuided` which supports sub-modes.
- Two modes are provided: `WP` for the existing `handle_guided_request`, and `Path` for the new `handled_guided_path_request`.
- A new virtual update method `update_path` is added to `AP_Navigation`.
- An implementation in `AP_L1_Control` uses the existing `update_waypoint` and `update_loiter` functions.

### Testing

See the tests using the [`terrain_navigation`](https://github.com/ethz-asl/terrain-navigation/tree/ros2) package in:

- https://github.com/ArduPilot/ardupilot/pull/26262

PR #28479, opened 2024-10-28
Plane Guided Alt Command: Use frame instead of bools for setting alt frame

Ryanf55:
# Purpose

When a plane receives an altitude command from mavlink [SET_POSITION_TARGET_GLOBAL_INT](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_GLOBAL_INT), we were using the booleans in the Location field to set the frame. This switches to using the altitude setter with the frame which is more explicit. 

We now convert the frames with `mavlink_coordinate_frame_to_location_alt_frame`.

Conveniently, deleting code saves 40b flash.

# Tests performed

`./Tools/autotest/autotest.py test.Plane`. 

It seems like this might be covered by `SetpointGlobalPos`, but that test is not run by Plane.

I also used the `changealt`  and `changealt_abs` command in `GUIDED` mode while flying in SITL to verify we can change altitude. Those commands do not use this interface. 

I do not have a great way to test this as is. Suggestions welcome!

# Size Compare Durandal

```
Binary Name      Text [B]        Data [B]     BSS (B)      Total Flash Change [B] (%)      Flash Free After PR (B)
---------------  --------------  -----------  -----------  ----------------------------  -------------------------
arducopter-heli  0 (0.0000%)     0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      14[28](https://github.com/ArduPilot/ardupilot/actions/runs/11563094404/job/32185663719?pr=28479#step:10:29)72
antennatracker   0 (0.0000%)     0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      615404
arducopter       0 (0.0000%)     0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      142344
ardurover        0 (0.0000%)     0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      [29](https://github.com/ArduPilot/ardupilot/actions/runs/11563094404/job/32185663719?pr=28479#step:10:30)6528
arduplane        -40 (-0.0022%)  0 (0.0000%)  0 (0.0000%)  -40 (-0.0022%)                                   145824
blimp            0 (0.0000%)     0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      593368
ardusub          0 (0.0000%)     0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      352236
```

### DIFF
--- a/ArduPlane/GCS_Mavlink.cpp
+++ b/ArduPlane/GCS_Mavlink.cpp
@@ -1372,43 +1372,27 @@ void GCS_MAVLINK_Plane::handle_set_position_target_global_int(const mavlink_mess
 
         mavlink_set_position_target_global_int_t pos_target;
         mavlink_msg_set_position_target_global_int_decode(&msg, &pos_target);
+
+        Location::AltFrame frame;
+        if (!mavlink_coordinate_frame_to_location_alt_frame((MAV_FRAME)pos_target.coordinate_frame, frame)) {
+            gcs().send_text(MAV_SEVERITY_WARNING, "Invalid coord frame in SET_POSTION_TARGET_GLOBAL_INT");
+            // Even though other parts of the command may be valid, reject the whole thing.
+            return;
+        }
+
         // Unexpectedly, the mask is expecting "ones" for dimensions that should
         // be IGNORNED rather than INCLUDED.  See mavlink documentation of the
         // SET_POSITION_TARGET_GLOBAL_INT message, type_mask field.
         const uint16_t alt_mask = 0b1111111111111011; // (z mask at bit 3)
             
-        bool msg_valid = true;
         AP_Mission::Mission_Command cmd = {0};
         
         if (pos_target.type_mask & alt_mask)
         {
-            cmd.content.location.alt = pos_target.alt * 100;
-            cmd.content.location.relative_alt = false;
-            cmd.content.location.terrain_alt = false;
-            switch (pos_target.coordinate_frame) 
-            {
-                case MAV_FRAME_GLOBAL:
-                case MAV_FRAME_GLOBAL_INT:
-                    break; //default to MSL altitude
-                case MAV_FRAME_GLOBAL_RELATIVE_ALT:
-                case MAV_FRAME_GLOBAL_RELATIVE_ALT_INT:
-                    cmd.content.location.relative_alt = true;
-                    break;
-                case MAV_FRAME_GLOBAL_TERRAIN_ALT:
-                case MAV_FRAME_GLOBAL_TERRAIN_ALT_INT:
-                    cmd.content.location.relative_alt = true;
-                    cmd.content.location.terrain_alt = true;
-                    break;
-                default:
-                    gcs().send_text(MAV_SEVERITY_WARNING, "Invalid coord frame in SET_POSTION_TARGET_GLOBAL_INT");
-                    msg_valid = false;
-                    break;
-            }    
-
-            if (msg_valid) {
-                handle_change_alt_request(cmd);
-            }
-        } // end if alt_mask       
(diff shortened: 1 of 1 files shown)


Ryanf55:
@IamPete1  Mind taking a look? This is starting to move towards treating altitude frame booleans as private.

It will make this PR smaller: https://github.com/ArduPilot/ardupilot/pull/26328#discussion_r1522656642

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -1382,22 +1382,20 @@ void GCS_MAVLINK_Plane::handle_set_position_target_global_int(const mavlink_mess
...
         {
-            cmd.content.location.alt = pos_target.alt * 100;
-            cmd.content.location.relative_alt = false;
-            cmd.content.location.terrain_alt = false;
+            const int32_t alt_cm = pos_target.alt * 100;
             switch (pos_target.coordinate_frame) 
             {
                 case MAV_FRAME_GLOBAL:
                 case MAV_FRAME_GLOBAL_INT:
+                    cmd.content.location.set_alt_cm(alt_cm, Location::AltFrame::ABSOLUTE);
```
peterbarker, line 1390: If we're playing in here we really should move to `mavlink_coordinate_frame_to_location_alt_frame` instead.

Ryanf55: Can do!

Ryanf55:
Switched to `mavlink_coordinate_frame_to_location_alt_frame` , now there's much less code. `changealt` still works.

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -1382,33 +1382,19 @@ void GCS_MAVLINK_Plane::handle_set_position_target_global_int(const mavlink_mess
...
-                    break;
-            }    
+            const int32_t alt_cm = pos_target.alt * 100;
+            Location::AltFrame frame;
+            if (mavlink_coordinate_frame_to_location_alt_frame((MAV_FRAME)pos_target.coordinate_frame, frame)) {
+                gcs().send_text(MAV_SEVERITY_WARNING, "Invalid coord frame in SET_POSTION_TARGET_GLOBAL_INT");
+                msg_valid = false;
+            } else {
+                cmd.content.location.set_alt_cm(alt_cm, frame);
+            }
```
peterbarker, line 1392:
```suggestion
                return;
            }
            cmd.content.location.set_alt_cm(alt_cm, frame);
```

also remove variable `msg_valid` entirely

peterbarker (changes requested):
Please outline what testing you have done on this PR.

Ryanf55:
> Please outline what testing you have done on this PR.

I did. Do you want me to upload logs, copy the exact terminal output, a video, or something more specific?

Ryanf55:
I'm working on adding an autotest in `master` to preserve existing behavior. Once that's in, we can revisit this PR.
See https://github.com/ArduPilot/ardupilot/pull/28487

peterbarker:
So PH and I weren't going to scope creep you on that bit :-)

OTOH, now that you've started - why not early-return on the alt-mask check too?  It's going to be a nice function after that.

Ryanf55:
> So PH and I weren't going to scope creep you on that bit :-)
> 
> OTOH, now that you've started - why not early-return on the alt-mask check too? It's going to be a nice function after that.

If you have a way to easily test it and are comfortable. 

The reason I don't early return on the alt mask is there will be other masks to test against coming up. The alt can be controlled in the upcoming path guidance mode at the same time as other fields in this function.

IamPete1 (approved):
This looks correct to me.

Issue #28435, opened 2024-10-19
Make Location.alt private to make altitude datums explicit


Ryanf55:
> What we really need to do is change to using an alt_frame attribute on Location instead of using these booleans.

Yes, we do, I've had patches around it before.

There some subtleties around the terrain stuff in particularly - there's some weird persistence with terrain following which you have to be really careful around.

Nothing impossible, of course, but terrain is unforgiving :-)

The trick is to make `.alt` private in Location and fix what breaks.

_Originally posted by @peterbarker in https://github.com/ArduPilot/ardupilot/issues/28406#issuecomment-2416165852_

Let's make the `alt` member of the Location structure private. After spending about 2 hours converting to getters/setters on https://github.com/Ryanf55/ardupilot/tree/private-alt-in-location, I have some ideas:

We frequently subtract/add altitudes from location structures. They are usually in the same datum. Perhaps we can find a way to save on comparisons and still shortcut when they are the same frame if we make an AP_Location::Altitude class that carries the altitude value, has the frame, and supports operators such as `-`, `+`, `+=`, `-=`. 

Thoughts on the approach I've taken so far of adding a getter with no frame? The getter could be a workaround as we phase reading `.alt` in the codebase.

PR #28481, opened 2024-10-28
Start using altitude getter - towards private alt in Location

Ryanf55:
# Purpose

We want to move towards having the altitude in AP_Location private and force using frames when getting/setting it to make it easier to spot frame bugs. This PR gets us 1/3 of the way to making the altitude field private. Once that's private, we can start to use grep to migrate off `get_alt_cm()` where we are worried about using the wrong frame.

Also, this could get us closer towards geoid-datum aware altitudes in AP.

# Context

https://github.com/ArduPilot/ardupilot/pull/28406#issuecomment-2416165852

# Issue

Relates to #28435

# Testing

None, hopefully this is narrowly scoped we can merge via inspection and manual testing in any areas of higher concern.

### DIFF
--- a/libraries/AP_Common/Location.h
+++ b/libraries/AP_Common/Location.h
@@ -42,6 +42,11 @@ public:
     // - above-home and home is not set
     // - above-origin and origin is not set
     bool get_alt_cm(AltFrame desired_frame, int32_t &ret_alt_cm) const WARN_IF_UNUSED;
+
+    // retrieve the altitude (in cm) in the stored frame.
+    int32_t get_alt_cm() const WARN_IF_UNUSED {
+        return alt;
+    }
     // same as get_alt_cm but in metres:
     bool get_alt_m(AltFrame desired_frame, float &ret_alt) const WARN_IF_UNUSED;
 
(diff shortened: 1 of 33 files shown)


libraries/AP_Common/Location.h
```diff
@@ -42,6 +42,11 @@ class Location
     // - above-home and home is not set
     // - above-origin and origin is not set
     bool get_alt_cm(AltFrame desired_frame, int32_t &ret_alt_cm) const WARN_IF_UNUSED;
+
+    // retrieve the altitude (in cm) in the stored frame.
+    int32_t get_alt_cm() const WARN_IF_UNUSED {
+        return alt;
+    }
```
peterbarker, line 49: Sorry, but I don't think this is acceptable.

The API requires you to specify a frame for a reason, and adding this is a significant step backwards.

peterbarker (changes requested):

libraries/AP_Common/Location.h
Ryanf55: Currently  You can just call `myLocation.alt = 42.0` and now you just set the value without any ability for us to find and stop it. 

This PR gives us ability to make .alt private, add a deprecation to this getter on a branch, and use the compiler to tell us all the places we need to phase it out of the codebase. 
Eventually, this function gets removed.

Ryanf55: I did discuss the approach with Pete Hall before starting, he thought this approach was valuable because we can grep the code. I'm totally open to alternatives, and am only 1/3 of the way through at the current state.

PR #28484, opened 2024-10-28
AP_L1_Control: Fix missing include

Ryanf55:


khancyr (approved):
LGTM

libraries/AP_L1_Control/AP_L1_Control.h
```diff
@@ -13,6 +13,7 @@
  *  - Ability to use loiter radius smaller than L1 length
  */
 
+#include <AP_AHRS/AP_AHRS.h>
```
peterbarker, line 16: We don't need this include - we don't need to know the shape of the class, only that it exists.

Instead, either pre-declare the class with `class AP_AHRS`, or modify the first use in the file (in the constructor) to be `class AP_AHRS &`

peterbarker (changes requested):

PR #28486, opened 2024-10-28
Copter: Configurable yaw offset for DO_SET_ROI

Ttl:
Currently DO_SET_ROI command always points nose of the vehicle towards the ROI. This PR adds parameter `ROI_YAW_OFFSET` that can be used to control yaw angle that points towards ROI.

Earlier forum question about this feature: https://discuss.ardupilot.org/t/offset-drone-heading-with-do-set-roi/88787

### DIFF
--- a/ArduCopter/Parameters.cpp
+++ b/ArduCopter/Parameters.cpp
@@ -1232,6 +1232,14 @@ const AP_Param::GroupInfo ParametersG2::var_info2[] = {
     // @User: Advanced
     AP_GROUPINFO("FS_EKF_FILT", 8, ParametersG2, fs_ekf_filt_hz, FS_EKF_FILT_DEFAULT),
 
+    // @Param: ROI_YAW_OFFSET
+    // @DisplayName: ROI yaw offset
+    // @Description: Offset applied to yaw when pointing to region of interest. Zero offset points nose at ROI.
+    // @Range: -18000 18000
+    // @Units: cdeg
+    // @User: Advanced
+    AP_GROUPINFO("ROI_YAW_OFFSET", 9, ParametersG2, roi_yaw_offset, 0),
+
     // ID 62 is reserved for the AP_SUBGROUPEXTENSION
 
     AP_GROUPEND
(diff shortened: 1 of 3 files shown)


NDevDrone:
@Ttl Hi Henrik, thank you for the contribution. 

The blog post is what caught my attention the other day. I asked for input from knowledgable devs, who took a look. @IamPete1  recommended "You might be able to get the desired behaviour by setting up a 2 axis servo gimbal. We could add the yaw offset param there."

If you'd like to take a crack at that, we could take a look and then and consider merging it.

rmackay9:
Hi @Ttl,

Thanks for the PR.  Can I ask why would you want to point the side of the vehicle towards the ROI?  I wonder if perhaps there is another solution.  For example, we don't currently have a parameter to allow people to specify how the gimbal is mounted on the frame.

Ttl:
I have [radar mounted on a drone](https://hforsten.com/homemade-polarimetric-synthetic-aperture-radar-drone.html) that is looking sideways and I want the radar antennas to keep pointing towards ROI as it flies. This should also be useful for fixed mounted cameras that are not pointing towards front.

rmackay9:
@Ttl,

Thanks for the feedback.  I must say though that this sounds like a real edge case so I wonder if it could be handled using a Lua script somehow.  We have to balance the addition of new parameters (which everyone pays a price for) against flexibility and how often features are likely to be used.

Ttl:
I'm fine with maintaining this patch myself for my own use. I submitted this PR in case there is interest for this feature in other applications, but it seems that there is very little demand for it.

If additional parameter is an issue and this is a desired feature, it could be a part of the gimbal logic instead. Initially, I tried specifying a dummy gimbal with YAW_MIN = YAW_MAX = 90 degrees. When gimbal is defined ROI logic points gimbal instead of drone front towards it. I assumed that if the gimbal yaw range is not enough it would yaw the drone as needed. However, this is not the case and gimbal yaw limits are currently not considered in the ROI logic.

rmackay9:
Hi @Ttl,

thanks very much for submitting it in any case.  Probably keeping it "out of tree" for now is best.

I completely agree that Copter's ROI handling should consider the gimbal's yaw min/max and point the nose of the vehicle only as much as it needs to.  That's a patch we would definitely accept but it's tricky.

Thanks!

PR #28490, opened 2024-10-29
Update clang pragma that is causing trouble

katzfey:
Update a clang pragma to check for the proper version of clang that introduces the warning.

### DIFF
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -1,7 +1,7 @@
 #include "Copter.h"
 
 #pragma GCC diagnostic push
-#if defined(__clang__)
+#if defined(__clang_major__) && __clang_major__ >= 14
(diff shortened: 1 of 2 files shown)


ArduCopter/AP_Arming.cpp
```diff
@@ -1,10 +1,5 @@
 #include "Copter.h"
 
-#pragma GCC diagnostic push
-#if defined(__clang__)
```
nexton-winjeel, line 4: This warning was added in Clang v14. Better to update the check rather than remove it, e.g.:
```
#if defined(__clang_major__) && __clang_major__ >= 14
```

katzfey: Okay, thanks, I will update the check then instead of removing it.

peterbarker (approved):
LGTM

tpwrules:
Is there code we should fix? What is the warning?

tridge (approved):

PR #14768, opened 2020-07-07
Allow retrieval of bootloader's git hash from bootloader

peterbarker:

PR #28492, opened 2024-10-29
Bootloader: Add Reading of Git SHA to bootloader

joshanne:
Allows the bootloader build to be identified with `uploader.py`

### DIFF
--- a/Tools/AP_Bootloader/bl_protocol.cpp
+++ b/Tools/AP_Bootloader/bl_protocol.cpp
@@ -83,7 +83,7 @@
 // RESET		finalise flash programming, reset chip and starts application
 //
 
-#define BL_PROTOCOL_VERSION 		5		// The revision of the bootloader protocol
+#define BL_PROTOCOL_VERSION 		6		// The revision of the bootloader protocol
 // protocol bytes
 #define PROTO_INSYNC				0x12    // 'in sync' byte sent before status
 #define PROTO_EOC					0x20    // end of command
@@ -128,6 +128,7 @@
 #define PROTO_DEVICE_FW_SIZE	4	// size of flashable area
 #define PROTO_DEVICE_VEC_AREA	5	// contents of reserved vectors 7-10
 #define PROTO_DEVICE_EXTF_SIZE  6   // size of available external flash
+#define PROTO_DEVICE_BL_GIT_HASH 7  // git hash of bootloader build
 // all except PROTO_DEVICE_VEC_AREA and PROTO_DEVICE_BOARD_REV should be done
 #define CHECK_GET_DEVICE_FINISHED(x)   ((x & (0xB)) == 0xB)
 
@@ -601,6 +602,10 @@ bootloader(unsigned timeout)
 
             case PROTO_DEVICE_EXTF_SIZE:
                 cout((uint8_t *)&board_info.extf_size, sizeof(board_info.extf_size));
+                break;
+            
+            case PROTO_DEVICE_BL_GIT_HASH:
+                cout((uint8_t *)&board_info.git_hash, sizeof(board_info.git_hash));
                 break;
 
             default:
--- a/Tools/scripts/uploader.py
+++ b/Tools/scripts/uploader.py
@@ -239,11 +239,12 @@ class uploader(object):
 
     INFO_BL_REV     = b'\x01'        # bootloader protocol revision
     BL_REV_MIN      = 2              # minimum supported bootloader protocol
-    BL_REV_MAX      = 5              # maximum supported bootloader protocol
+    BL_REV_MAX      = 6              # maximum supported bootloader protocol
     INFO_BOARD_ID   = b'\x02'        # board type
     INFO_BOARD_REV  = b'\x03'        # board revision
     INFO_FLASH_SIZE = b'\x04'        # max firmware size in bytes
     INFO_EXTF_SIZE  = b'\x06'        # available external flash size
+    INFO_GIT_HASH   = b'\x07'        # git hash of bootloader build
 
     PROG_MULTI_MAX  = 252            # protocol max is 255, must be multiple of 4
     READ_MULTI_MAX  = 252            # protocol max is 255
@@ -737,6 +738,10 @@ class uploader(object):
         self.board_rev = self.__getInfo(uploader.INFO_BOARD_REV)
         self.fw_maxsize = self.__getInfo(uploader.INFO_FLASH_SIZE)
 
+        # Git SHA introduced added in v6
+        if self.bl_rev > 5: 
+            self.git_hash = self.__getInfo(uploader.INFO_GIT_HASH)
+
     def dump_board_info(self):
         # OTP added in v4:
         print("Bootloader Protocol: %u" % self.bl_rev)
@@ -838,6 +843,9 @@ class uploader(object):
         else:
             print("  board_type: %u" % self.board_type)
         print("  board_rev: %u" % self.board_rev)
+
+        git_hash = "%x" % (self.git_hash) if self.bl_rev > 5 else "UNKNOWN"
+        print("  git hash: %s" % git_hash)
 
         print("Identification complete")
 
--- a/Tools/AP_Bootloader/AP_Bootloader.cpp
+++ b/Tools/AP_Bootloader/AP_Bootloader.cpp
@@ -39,6 +39,10 @@
 #include <AP_CheckFirmware/AP_CheckFirmware.h>
 #include "network.h"
 
+#define FORCE_VERSION_H_INCLUDE
+#include "ap_version.h"
+#undef FORCE_VERSION_H_INCLUDE
+
 extern "C" {
     int main(void);
 }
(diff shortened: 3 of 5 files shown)


Tools/AP_Bootloader/bl_protocol.cpp
```diff
@@ -128,6 +128,7 @@
 #define PROTO_DEVICE_FW_SIZE	4	// size of flashable area
 #define PROTO_DEVICE_VEC_AREA	5	// contents of reserved vectors 7-10
 #define PROTO_DEVICE_EXTF_SIZE  6   // size of available external flash
+#define PROTO_DEVICE_BL_GIT_HASH 7  // git hash of bootloader build
 // all except PROTO_DEVICE_VEC_AREA and PROTO_DEVICE_BOARD_REV should be done
 #define CHECK_GET_DEVICE_FINISHED(x)   ((x & (0xB)) == 0xB)
```
joshanne, line 133: I'm unsure if this line would need to be updated to `0x4B`?

joshanne:
This is helpful for users building their own bootloaders, and allowing the user to identify what commit the build was built at.

Feedback welcome.

Tools/AP_Bootloader/bl_protocol.cpp
```diff
@@ -128,6 +128,7 @@
 #define PROTO_DEVICE_FW_SIZE	4	// size of flashable area
 #define PROTO_DEVICE_VEC_AREA	5	// contents of reserved vectors 7-10
 #define PROTO_DEVICE_EXTF_SIZE  6   // size of available external flash
+#define PROTO_DEVICE_BL_GIT_HASH 7  // git hash of bootloader build
```
joshanne, line 131: I've also got changes that add reading off the Application Git SHA if `AP_CHECK_FIRMWARE_ENABLED` is enabled in the bootloader (and application)

Tools/AP_Bootloader/bl_protocol.cpp
```diff
@@ -122,14 +122,28 @@
...
+#define PROTO_DEVICE_FW_SIZE	  4  // size of flashable area
+#define PROTO_DEVICE_VEC_AREA	  5  // contents of reserved vectors 7-10
+#define PROTO_DEVICE_EXTF_SIZE    6  // size of available external flash
+#define PROTO_DEVICE_BL_GIT_HASH  7  // git hash of bootloader build (requires > v5)
+
 // all except PROTO_DEVICE_VEC_AREA and PROTO_DEVICE_BOARD_REV should be done
-#define CHECK_GET_DEVICE_FINISHED(x)   ((x & (0xB)) == 0xB)
+
+// Convert the Device Info to a bitmask
+#define REQUIRE_DEVICE_INFO(x) (1<<(x-1))
```
nexton-winjeel, line 136: This should be used below (line 627) where `done_get_device_flags` is updated, i.e. change:
```
            done_get_device_flags |= (1<<(arg-1)); // set the flags for use when resetting timeout 
```
to:
```
            done_get_device_flags |= REQUIRE_DEVICE_INFO(arg); // set the flags for use when resetting timeout 
```

nexton-winjeel, line 136: This would be clearer if it was named `DEVICE_INFO_MASK`.

Tools/scripts/uploader.py
```diff
@@ -839,6 +844,9 @@ def dump_board_info(self):
             print("  board_type: %u" % self.board_type)
         print("  board_rev: %u" % self.board_rev)
 
+        git_hash_bl = "%x" % (self.git_hash_bl) if self.bl_rev > 5 else "UNKNOWN"
```
tpwrules, line 847:
```suggestion
        git_hash_bl = "%x" % (self.git_hash_bl) if self.bl_rev > 5 else "UNAVAILABLE"
```

tridge:
if we can get away without a protocol change, by checking for a NACK then that would be less risky, otherwise we need to be sure existing bootloader clients are OK

peterbarker:
Wrote this about 4 years ago - glad someone's pushing through with it (https://github.com/ArduPilot/ardupilot/pull/14768/files)

Tools/scripts/uploader.py
joshanne: removed completely.

joshanne:
If this implementation stands, probably need to include an update to the comments above, here: 

https://github.com/ArduPilot/ardupilot/blob/f0900bd119562dffe4cafabc1d2903e6951c128f/Tools/AP_Bootloader/bl_protocol.cpp#L549-L557

joshanne:
> Wrote this about 4 years ago - glad someone's pushing through with it (https://github.com/ArduPilot/ardupilot/pull/14768/files)

@peterbarker - Your implementation was adding a new command. This implementation makes use of the `GET_DEVICE` (_INFO) command.

Do you have a feeling either way? I'm less inclined to burn a full command index unnecessarily.

tridge:
> given this is based off PX4-Bootloader, what requirement is there to keep it in sync?

doesn't need to be completely in sync, but we do need to ensure that flashing ArduPilot when px4 bootloader is used works and vice-versa
some flight controllers that usually run ArduPilot come with the px4 bootloader

Tools/AP_Bootloader/bl_protocol.cpp
```diff
@@ -602,11 +623,57 @@ bootloader(unsigned timeout)
...
+            case PROTO_DEVICE_APP_SOFTWARE:
+            {
+#if AP_CHECK_FIRMWARE_ENABLED
+                uint32_t git_hash;
+                const auto *ad = get_app_descriptor();
+                
+                if (ad != nullptr) {
+                    // app_descriptor is valid, grab the hash and output
+                    git_hash = ad->git_hash;
+                    cout((uint8_t *)&git_hash, sizeof(git_hash));
```
peterbarker, line 663: This probably needs to be 64 bits.

I've seen problems in our tooling which I think comes down to us now having more commits and increased chance of collisions, so a 32-bit subset of the sha yields collisions.  Sadly I haven't actually chased this problem down, so only half-remembering; the workaround is to slightly change a commit...

peterbarker:
We should probably PR into the PX4 bootloader the additional "info" IDs so as not to cause collisions down the road.  Would require a chunk of documentation of the shape of the relevant data...

joshanne:
> We should probably PR into the PX4 bootloader the additional "info" IDs so as not to cause collisions down the road. Would require a chunk of documentation of the shape of the relevant data...

Documented size and shape of data for Build SHAs - both BL and App SHAs are 64 bits each (16 hex characters each).

Also pushed a PR to PX4 to reserve and maintain compatability with PX4 bootloader.

Only catch at this stage is `app_descriptor` is a `uint32_t` sha - so for the time being, we reply with the SHA, with succeeding zeroes, and handle display in `uploader.py`.

joshanne:
> > given this is based off PX4-Bootloader, what requirement is there to keep it in sync?
> 
> doesn't need to be completely in sync, but we do need to ensure that flashing ArduPilot when px4 bootloader is used works and vice-versa some flight controllers that usually run ArduPilot come with the px4 bootloader

Tested flashing AP to PX4 and back to AP works as expected. The display of the SHAs and bootloader options only occurs when using `--identify`.

peterbarker:
Apparently PX4 might have added an equivalent feature into their bootloader.  Might be worth checking how they worked it into the protocol and whether we can be compatible.

joshanne:
> Apparently PX4 might have added an equivalent feature into their bootloader.

Yeah, I had noticed that late yesterday too. grumble grumble.

I'm reducing the scope of this PR to solely implement the bootloader version reporting using the same command as PX4. 

Works as is, ArduPilot reports the 16 char sha (eg: `1d1e8e6cbb906865`), PX4 reports a custom string (eg: `PX4BLv1.16.0g1abc27dd1f`).

Should we prefix the sha to look something like the following: `APBL.1d1e8e6cbb906865`?
If so, is that preferred to be setup at build time (in `boards.py`)? or run time (in `bootloader.cpp`)?

julianoes (commented):
I think this looks correct. If it was me, I'd probably add a `g` before the git hash but that's up to you.

And I will remove the confusing `b'\x07'` on the PX4 side which looks like a mistake.

peterbarker:
```
Board,bootloader
CubeOrange,80
MatekF405,72
```

joshanne:
Great to see this is getting traction again - i'll be there in the dev call. :smile:

tpwrules:
Does this fit in all bootloaders? It leaves 36 bytes free in `DrotekP3Pro`. Not that I think we shouldn't accept this, just that optimizations to bootloader code are needed.

tridge:
The Tools/bootloaders/MatekF405-Wing_bl.bin build is 16368 bytes, so only 16 bytes free
I think we need to turn this off on boards with a 16k bootloader sector (eg. F4xx) 
even if it does fit, it would give us no room to grow

tridge (changes requested):
I'd like to see this disabled on boards with a 16k bootloader sector, we're just too close, I think disable on F4xx

peterbarker:
So checking PROGRAM_SIZE > 16384 in a new define

joshanne:
Rebased and added checks for program size as requested.

Adding to the EU Dev Call list for tomorrow.

tridge (approved):

PR #28494, opened 2024-10-29
GCS_MAVLink: add warning if guided-mode-mission-requests are used

peterbarker:
image


peterbarker:
```
###### Enabling GUIDED_MISSION_REQUESTS(HAL_GCS_GUIDED_MISSION_REQUESTS_ENABLED) on copter costs -600 bytes
###### Enabling GUIDED_MISSION_REQUESTS(HAL_GCS_GUIDED_MISSION_REQUESTS_ENABLED) on plane costs -552 bytes
###### Enabling GUIDED_MISSION_REQUESTS(HAL_GCS_GUIDED_MISSION_REQUESTS_ENABLED) on rover costs -496 bytes
###### Enabling GUIDED_MISSION_REQUESTS(HAL_GCS_GUIDED_MISSION_REQUESTS_ENABLED) on antennatracker costs -432 bytes
###### Enabling GUIDED_MISSION_REQUESTS(HAL_GCS_GUIDED_MISSION_REQUESTS_ENABLED) on sub costs -536 bytes
###### Enabling GUIDED_MISSION_REQUESTS(HAL_GCS_GUIDED_MISSION_REQUESTS_ENABLED) on blimp costs -432 bytes
```

tridge:
@peterbarker we discussed on the dev call and think we can just remove support now for 4.7

rmackay9:
I think we should just remove the code.  Our mavlink interface wiki page spells out much better ways to achieve the same thing

Another option would be to merge this to 4.6 and then remove the code from master (aka 4.7) almost immediately

peterbarker:
> I think we should just remove the code. Our mavlink interface wiki page spells out much better ways to achieve the same thing

Insta-breaking people's setups isn't a great idea IMO.  Removing this code would insta-break all existing MAVProxy installs, for example.
 
> Another option would be to merge this to 4.6 and then remove the code from master (aka 4.7) almost immediately

That would be nice!

rmackay9 (approved):

PR #28495, opened 2024-10-29
AP_NavEKF3: fix computation of rho

jschall:
I noticed this calculation was inconsistent with get_air_density_ratio in AP_Baro, so I checked which one is theoretically correct and AP_Baro is the winner.

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_AirDataFusion.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_AirDataFusion.cpp
@@ -485,7 +485,7 @@ void NavEKF3_core::FuseDragForces()
     ZERO_FARRAY(Kfusion);
     Vector24 Hfusion; // Observation Jacobians
     const ftype R_ACC = sq(fmaxF(frontend->_dragObsNoise, 0.5f));
-    const ftype density_ratio = sqrtF(dal.get_EAS2TAS());
+    const ftype density_ratio = 1.0f/sq(dal.get_EAS2TAS());
     const ftype rho = fmaxF(1.225f * density_ratio, 0.1f); // air density
 
     // get latest estimated orientation


rmackay9:
Thanks for this.

Not sure I'm qualified to review this but I guess this change introduces the possibility of a divide by zero?

I wonder what level of testing has been done.  How much of a difference does this make?

peterbarker:
> Not sure I'm qualified to review this but I guess this change introduces the possibility of a divide by zero?

We're assuming get_EAS2TAS() is non-zero in many, many places already.

peterbarker (approved):
This looks right to me.

This moves it to be the same calculation as `AP_AHRS::get_air_density_ratio` and `AP_Baro::_get_air_density_ratio` (see also `tailsitter.cpp:718` calculating rho from `get_air_density_ratio`).

The common code seems to be the correct inverse of the calculations made in `AP_Baro::get_EAS2TAS_simple`.

libraries/AP_NavEKF3/AP_NavEKF3_AirDataFusion.cpp
```diff
@@ -485,7 +485,7 @@ void NavEKF3_core::FuseDragForces()
     ZERO_FARRAY(Kfusion);
     Vector24 Hfusion; // Observation Jacobians
     const ftype R_ACC = sq(fmaxF(frontend->_dragObsNoise, 0.5f));
-    const ftype density_ratio = sqrtF(dal.get_EAS2TAS());
+    const ftype density_ratio = 1.0f/sq(dal.get_EAS2TAS());
```
Georacer, line 488: Perhaps keep the `sqrtF` to respect EKF operations number types?

Georacer:
I have also verified this change, with reference to:
US Standard Atmosphere, 1976. (1976). National Oceanic and Atmospheric Administration, National Aeronautics and Space Adminisration, United States Air Force.

Georacer (approved):

priseborough (approved):

peterbarker:
Merged, thanks @jschall !

PR #28498, opened 2024-10-30
ArduPlane: Remove unused radio_in_rssi function

Ryanf55:
This is not used anywhere


tpwrules (approved):
Don't see any other references.

PR #28499, opened 2024-10-30
AP_HAL_ESP32: ADC timer tick moved into scheduler timer process

sblaksono:
Move the ADC timer tick into timer process to make scheduler code looks tidier


davidbuzz (approved):

davidbuzz:
does this affect the performance, eg loop rate of the scheduler, at all?

peterbarker:
@sblaksono while Buzz has approved this, he also asked a question.  I'm hesitant to merge this without an answer, sorry :-)

PR #28502, opened 2024-10-30
.github: update actions/cache to v4 to remove node version warning

khancyr:
The following actions use a deprecated Node.js version and will be forced to run on node20: actions/cache@v3.

No change to be expected, only nodejs update according to changelog https://github.com/actions/cache?tab=readme-ov-file#whats-new


Ryanf55 (approved):
v4 is a stable release. LGTM.

https://github.com/actions/cache/releases

PR #28506, opened 2024-10-31
Fix Documentation Generating Parameters Incorrectly

joshanne:
Parameters are being generated as `INS_` as the group, so `INS_USE` is actually documented as `INS__USE` (note the double underscore). (See: [INS__USE: Use first IMU for attitude, velocity and position estimates](https://ardupilot.org/dev/docs/AP_Periph-Parameters.html#ins-use-use-first-imu-for-attitude-velocity-and-position-estimates)

Also fixes documentation in the battery library where the generated referenced parameter contains double underscores: (see: [BATT2_ARM_MAH: Required arming remaining capacity](https://ardupilot.org/plane/docs/parameters.html#batt2-arm-mah-required-arming-remaining-capacity))

### DIFF
--- a/Tools/AP_Periph/Parameters.cpp
+++ b/Tools/AP_Periph/Parameters.cpp
@@ -732,7 +732,7 @@ const AP_Param::Info AP_Periph_FW::var_info[] = {
     // @User: Standard
     GSCALAR(imu_sample_rate, "INS_SAMPLE_RATE", 0),
 
-    // @Group: INS_
+    // @Group: INS
     // @Path: ../libraries/AP_InertialSensor/AP_InertialSensor.cpp
     GOBJECT(imu, "INS", AP_InertialSensor),
 #endif
--- a/libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
@@ -143,7 +143,7 @@ const AP_Param::GroupInfo AP_BattMonitor_Params::var_info[] = {
 
     // @Param: ARM_MAH
     // @DisplayName: Required arming remaining capacity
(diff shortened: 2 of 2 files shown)


joshanne:
@Hwurzburg this is a parameter documentation generation fix - two instances were generating parameter names with double underscores.

Tools/AP_Periph/Parameters.cpp
```diff
@@ -732,7 +732,7 @@ const AP_Param::Info AP_Periph_FW::var_info[] = {
     // @User: Standard
     GSCALAR(imu_sample_rate, "INS_SAMPLE_RATE", 0),
 
-    // @Group: INS_
+    // @Group: INS
```
joshanne, line 735: Follows the same standard as: [ArduPlane/Parameters.cpp](https://github.com/ArduPilot/ardupilot/blob/be5c68d74db69dd2f3e5226ff576e1b9b8166c79/ArduPlane/Parameters.cpp#L910)

joshanne:
Tested with: `./Tools/autotest/param_metadata/param_parse.py --vehicle AP_Periph`

Hwurzburg (approved):
LGTM, but I dont see the correction here for the BATT2 double underscore....which I only find in the metadata description of the periph BATT2_ARM_MAH param

joshanne:
> I only find in the metadata description of the periph BATT2_ARM_MAH param

It's not only `BATT2`, it is in the description of _**all**_ `BATTn_ARM_MAH` - it incorrectly references `BATTn__ARM_VOLT` (note the double underscore) - and it affects all vehicles. The correct way (as has been done in other places in `AP_BattMonitor_Params`) is to use `@PREFIX@` without the trailing underscore. See [source code for `BATTn_LOW_VOLT`](https://github.com/ArduPilot/ardupilot/blob/be5c68d74db69dd2f3e5226ff576e1b9b8166c79/libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp#L82)

joshanne:
> LGTM, but I dont see the correction here for the BATT2 double underscore....which I only find in the metadata description of the periph BATT2_ARM_MAH param

Quite right, apparently I forgot to push the second commit.

amilcarlucas (approved):

IamPete1 (approved):
LGTM

tridge (approved):

PR #28508, opened 2024-10-31
Feature/rangefinder: Cheap generic sonar

khancyr:
This is the first part of the code for https://discuss.ardupilot.org/t/cheap-waterproof-ultrasonic-sensor-modules/125049

Need to add lua and autotest and try on real device


geofrancis:
you will need to add a polled mode, most of these sonars can work in 2 modes, they either ping continuously every 100ms or they have to be polled by sending 0x55 it can usually be set with a jumper on the board but that isn't accessible if it's been encapsulated.  so for it to be fully compatible with all of them you need to send 0x55, you can send this to all of them, it will be ignored by the sonars running continuously .

rmackay9:
Thanks for this.  Instead of "generic" I wonder if we could name it after one of the more common chips or sonar models?  So "A02YYUW" or "ME007YS".  The only reason is that "generic" could be misinterpreted by users as supporting nearly all sonars and could lead to support issues.  If we're more specific with the name we could avoid that.

geofrancis:
> Thanks for this. Instead of "generic" I wonder if we could name it after one of the more common chips or sonar models? So "A02YYUW" or "ME007YS". The only reason is that "generic" could be misinterpreted by users as supporting nearly all sonars and could lead to support issues. If we're more specific with the name we could avoid that.

JSN-SR04T and A02YYUW are the 2 most common versions air versions, but this would also cover the GL041MT underwater sonar.

khancyr:
@geofrancis do you have a docs about a sensor using the polling ?

The two sensors I got use low RX pin to get realtime measurements and high RX to get processed measurements. No idea what is processed but from the doc the rate is random between 100-300ms which can be an issue. So sending 0x55 can be an issue on such device

khancyr:
@rmackay9 Indeed, it could be a bit confusing, I will try to dig a bit more to find a another name or ask our Chinese partner if they got an idea what this protocol is called

geofrancis:
> @geofrancis do you have a docs about a sensor using the polling ?
https://dronebotworkshop.com/waterproof-ultrasonic

peterbarker:
I've rebased this and cleaned it up a little.

We do need a better name than "SinoSonar", sorry :-)  That's a real company with real soundy products.

peterbarker:
I've tested this here on a JSN-SR04T.  I bridged the jumper to set it into "request serial" mode and added a patch here to send the request as @geofrancis suggested.

@khancyr did you want to pursue this or shall I get this ready for merge?

khancyr:
Hello,

You can take over if you work on this! 
For the name, maybe AliExpSonar, NOIDSonar, CLNESonar, BULKSonar ?

rmackay9:
Hi @khancyr, @peterbarker, I hope it's OK, I've slightly updated the title.  I think we should change the commit titles as well at a convenient time.  I think we should describe the sonar in a more functional sense without directly referring to where it's made because that comes with baggage

khancyr:
AP_RangeFinder_SerialFF
AP_RangeFinder_UARTFF
AP_RangeFinder_FFFrame

AP_RangeFinder_Serial4B
AP_RangeFinder_Fixed4B
AP_RangeFinder_SerialFrame4

for name so we can merge this ?

libraries/AP_RangeFinder/AP_RangeFinder_SinoSonar_Serial.cpp
```diff
@@ -0,0 +1,71 @@
...
+            break;
+        }
+        if (buf_len == 0 && b != SINOSONAR_HEADER) {
+            // discard
+            continue;
+        }
+        buf[buf_len++] = b;
+        if (buf_len == sizeof(buf)) {
+            buf_len = 0;
+            const uint16_t crc = (buf[0] + buf[1] + buf[2]) & 0x00FF;
```
peterbarker: Any reason to not to just sum into a uint8_t?

libraries/AP_RangeFinder/AP_RangeFinder_config.h
```diff
@@ -63,6 +63,10 @@
 #define AP_RANGEFINDER_DRONECAN_ENABLED (HAL_ENABLE_DRONECAN_DRIVERS && AP_RANGEFINDER_BACKEND_DEFAULT_ENABLED)
 #endif
 
+#ifndef AP_RANGEFINDER_SINOSONAR_SERIAL_ENABLED
+#define AP_RANGEFINDER_SINOSONAR_SERIAL_ENABLED AP_RANGEFINDER_BACKEND_DEFAULT_ENABLED
```
peterbarker: Perhaps off by default?

khancyr: yup, that is probably usefull only for rover and boat considering the capabilities

libraries/AP_RangeFinder/AP_RangeFinder_SinoSonar_Serial.cpp
khancyr: old things, but I would say we can indeed reduce to uint8_t directly.  will update

khancyr:
rebased and renamed. 
Need final testing

PR #28512, opened 2024-10-31
AP_Scripting: mount-djirs2: freshen driver

tpwrules:
In particular allows the use on a DroneCAN bus. Please see commits for details.

Tested on a DJI RS2 using firmware 01.05.00.20.

### DIFF
--- a/libraries/AP_Scripting/drivers/mount-djirs2-driver.lua
+++ b/libraries/AP_Scripting/drivers/mount-djirs2-driver.lua
@@ -1,5 +1,4 @@
 -- mount-djirs2-driver.lua: DJIRS2 mount/gimbal driver
--- luacheck: only 0
 
 --[[
   How to use
@@ -71,8 +70,6 @@
 
 --]]
 
----@diagnostic disable: cast-local-type
-
 
 -- global definitions
 local INIT_INTERVAL_MS = 3000           -- attempt to initialise the gimbal at this interval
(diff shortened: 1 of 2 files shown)


rmackay9 (approved):
Great, thanks very much!  nice to have someone else also poking away at this driver!

libraries/AP_Scripting/drivers/mount-djirs2-driver.lua
```diff
@@ -71,16 +76,16 @@
 
 --]]
 
----@diagnostic disable: cast-local-type
-
+-- user definitions
+local MOUNT_INSTANCE = 0                -- default to MNT1
```
joshanne, line 80: If mount is being exposed to a `User` should be `1-indexed` and calculated below to point to the `n-1` driver? (noting that `mnt_type_name` uses `MOUNT_INSTANCE+1` so would need to change)

or should we better document it so that users know this is `0-indexed`?

Does AP have a standard? I think in Periph any mention of external ports are `0-indexed`, but only because the serial driver is `0-indexed`?

Given the first mount is `MNT1`, my assumption would be this should be `MOUNT_INSTANCE = 1` for the first mount.

tpwrules: I'm not sure to be honest, the CAN parameter is also in some sense 1-indexed. I had hoped the comment would be enough clarity, and I think it's extremely unlikely that this will ever get used.

rmackay9: For better or worse, we normally use zero indexed within the C++ and Lua code and one indexed when exposing parameters to the user.  I think this is a definition within the code (e.g. not a parameter) so it should stay as zero

PR #28513, opened 2024-11-01
AP_Scripting: ESC_slew_rate: fix lua warning

tpwrules:
Docs say that logger must be called with `:`.

Not sure why this didn't get caught on the initial merge. Perhaps the checker upgraded.


tridge (approved):

tridge:
We should also reset the CHAN parameter on startup

tpwrules:
This also needs to go in 4.6 so that CI passes.

PR #28515, opened 2024-11-02
AP_HAL_ChibiOS: add hwdef MFE_POS3_CAN

mikefenghao:
This is the hwdef file for MFE_POS3_CAN. We've done a lot of testing and it's working fine.


libraries/AP_HAL_ChibiOS/hwdef/MFE_POS3_CAN/hwdef.dat
```diff
@@ -0,0 +1,108 @@
...
+# MCU class and specific type
+MCU STM32F4xx STM32F427xx
+
+FLASH_RESERVE_START_KB 36
+
+STORAGE_FLASH_PAGE 16
+define HAL_STORAGE_SIZE 8192
+
+# board ID for firmware load
+APJ_BOARD_ID 6101
```
Hwurzburg, line 15: use logical id

Hwurzburg (changes requested):

libraries/AP_HAL_ChibiOS/hwdef/MFE_POS3_CAN/hwdef.dat
peterbarker, line 15: You need to reserve this board ID in `Tools/AP_Bootloader/board_types.txt` and then use the name you use in that file in here.

peterbarker (dismissed):
Grab 10 IDs in board_types.txt while you're there - you can see how others are doing that in there.

libraries/AP_HAL_ChibiOS/hwdef/MFE_POS3_CAN/hwdef.dat
tridge: fixed

tridge (approved):

PR #28520, opened 2024-11-04
AP_Scripting: Applets: MissionSelector: fix error on file open and checker errors

IamPete1:
Ages ago file open would always succeed, we fixed the other scripts and missed this one. Now file can be `nil` so we need to check for that. This fixes the script so we can run the static checkers.

### DIFF
--- a/libraries/AP_Scripting/applets/MissionSelector.lua
+++ b/libraries/AP_Scripting/applets/MissionSelector.lua
@@ -84,4 +84,6 @@ function update()
   return update, 1000
 end
 
+gcs:send_text(5,"Loaded MissionSelector.lua")
+
 return update, 5000
--- a/libraries/AP_Scripting/applets/MissionSelector.lua
+++ b/libraries/AP_Scripting/applets/MissionSelector.lua
@@ -2,9 +2,6 @@
 -- Must have Mission Reset switch assigned, it will function normally when armed or disarmed
 -- but also on the disarm to arm transition, it will load (if file exists) a file in the root named
 -- missionH.txt, missionM.txt, or missionH.txt corresponding to the the Mission Reset switch position of High/Mid/Low
--- luacheck: only 0
----@diagnostic disable: need-check-nil
-
 
 local mission_loaded = false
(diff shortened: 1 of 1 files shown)


libraries/AP_Scripting/applets/MissionSelector.lua
```diff
@@ -64,7 +61,6 @@ local function read_mission(file_name)
     end
     index = index + 1
   end
-  file:close()
```
IamPete1, line 67: This is unreachable.

Hwurzburg (approved):
thanks...while we are at it could you add:
gcs:send_text(5,"Loaded MissionSelector.lua")
 before the last line....I think a notification at boot of what scripts actually are running is worthwhile and have been asking submitters to add this when I catch it....

IamPete1:
> thanks...while we are at it could you add: gcs:send_text(5,"Loaded MissionSelector.lua") before the last line....I think a notification at boot of what scripts actually are running is worthwhile and have been asking submitters to add this when I catch it....

Done.

timtuxworth:
Why do you wait till armed to load the mission? Surely this could happen (on Plane):

1. user selects AUTO
2. user arms the vehicle
3. vehicle tries to run the mission (there is none)
4. script loads the mission

resulting in the vehicle just sitting on the ground, instead of taking off to fly the selected mission. No?

Hwurzburg:
if there is no mission and you are in AUTO, you cant arm.....if there is, the existing mission wont be cleared(because its actually in progress when you switched mode to AUTO, even disarmed), it will run normally, and the script will send GCS text warnings to inform the user....one would first arm in some other mode, like FBWA(loading the desired mission....or not if there is no file), then switch to AUTO to run auto missions

timtuxworth:
The comments at the top of the file say this 
`it will function normally when armed or disarmed`
but as far as I can see, it will not function while disarmed. I'm thinking it would be better if the comment was correct, I'd want to be able to switch missions while disarmed and verify I have the correct one while still on the ground before arming.

Hwurzburg:
the first senteence applies to the Mission Reset switch....it indeed functions normally as Mission Reset whether armed or disarrmed

Hwurzburg (approved):

PR #28526, opened 2024-11-05
ArduPlane: make Guided update frequency limit configurable

timtuxworth:
in Guided mode, the current code throttles updates from externally received messages that updates roll, pitch or throttle. allowing a maximum of one update every 3 seconds for any external message 

For Plane Follow, the lua script runs at 10Hz, so this limit prevents it from working properly. 

So as not to change existing behavior, I made it a new parameter GUIDED_UPD_LIM - in milliseconds, default to 3000.

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1288,7 +1288,14 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("RNGFND_LND_ORNT", 36, ParametersG2, rangefinder_land_orient, ROTATION_PITCH_270),
 #endif
-    
+
+    // @Param: GUIDED_UPD_LIM
+    // @DisplayName: Guided Update Limit
+    // @Description: The maximum frequency that an guided mode commands sent by external system such as lua or mavlink can update roll, pitch and throttle.
+    // @Units: ms
+    // @User: Standard
+    AP_GROUPINFO("GUIDED_UPD_LIM", 37, ParametersG2, guided_update_frequency_limit, 3000),
+
     AP_GROUPEND
 };
 
--- a/ArduPlane/Parameters.h
+++ b/ArduPlane/Parameters.h
@@ -566,6 +566,8 @@ public:
     AP_Follow follow;
 #endif
 
+    AP_Int32        guided_update_frequency_limit;
+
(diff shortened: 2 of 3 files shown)


timtuxworth:
I found this was required for plane follow to work. Does this make sense @peterbarker @IamPete1 ?

IamPete1:
You can update roll/pitch/throttle as fast as you like. I think the issue your having is that you can't swap from roll/pitch/throttle commands to a navigation command quickly because the timeout has to elapse?

timtuxworth:
> You can update roll/pitch/throttle as fast as you like. I think the issue your having is that you can't swap from roll/pitch/throttle commands to a navigation command quickly because the timeout has to elapse?

That might be what's happening @IamPete1 - the one that impacts me most is throttle, as in follow I am trying to navigate and also potentially adjust speed every loop. My use case is always level flight.

IamPete1:
How about a update to `handle_set_attitude_target` such that if the bits are not set the time stamps are reset? This has the advantage that you can reset instantly rather than just having to wait for a smaller timeout.

timtuxworth:
> How about a update to `handle_set_attitude_target` such that if the bits are not set the time stamps are reset? This has the advantage that you can reset instantly rather than just having to wait for a smaller timeout.

I'm not using SET_ATTITUDE_TARGET - I'm using GUIDED_CHANGE_SPEED.

IamPete1:
> I'm not using SET_ATTITUDE_TARGET - I'm using GUIDED_CHANGE_SPEED.

`GUIDED_CHANGE_SPEED` commands don't use the timeouts your changing here? You must be using `SET_ATTITUDE_TARGET` somewhere or this change would make no difference?

Maybe you can explain the exact commands your using.

timtuxworth:
> > I'm not using SET_ATTITUDE_TARGET - I'm using GUIDED_CHANGE_SPEED.
> 
> `GUIDED_CHANGE_SPEED` commands don't use the timeouts your changing here? You must be using `SET_ATTITUDE_TARGET` somewhere or this change would make no difference?
> 
> Maybe you can explain the exact commands your using.

I'm sure I found that this code was being hit by putting a breakpoint there, bit I can't remember the details now, so I'll see if I can duplicate the problem.

timtuxworth:
@Ryanf55 has a similar PR #28943

timtuxworth:
Closed in favor of #28943

PR #28527, opened 2024-11-05
implement GUIDED_HEADING_TYPE default mavlink message

timtuxworth:
The MAVLink GUIDED_CHANGE_HEADING command will switch from the default to either HEADING_TYPE_HEADING or HEADING_TYPE_COG, but did not have a way to switch back to "default" this adds the code to implement HEADING_TYPE_DEFAULT which will switch the ArduPilot internal plane.guided_state.target_heading_type back to GUIDED_HEADING_NONE, which is the default value with it's own behavior.

Follow up to the MAVLink PR https://github.com/ArduPilot/mavlink/pull/372 which adds the implementation.

This PR also updates the mavlink submodule reference.

### DIFF
--- a/ArduPlane/GCS_Mavlink.cpp
+++ b/ArduPlane/GCS_Mavlink.cpp
@@ -863,6 +863,9 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_do_reposition(const mavlink_com
     if (((int32_t)packet.param2 & MAV_DO_REPOSITION_FLAGS_CHANGE_MODE) ||
         (plane.control_mode == &plane.mode_guided)) {
         plane.set_mode(plane.mode_guided, ModeReason::GCS_COMMAND);
+#if AP_PLANE_OFFBOARD_GUIDED_SLEW_ENABLED
+        plane.guided_state.target_heading_type = GUIDED_HEADING_NONE;
+#endif
 
         // add home alt if needed
         if (requested_position.relative_alt) {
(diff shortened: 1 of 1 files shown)


timtuxworth:
Follow up to the MAVLink PR https://github.com/ArduPilot/mavlink/pull/372 which adds the implementation. How does this look @peterbarker , @IamPete1 ?

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -973,22 +976,28 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands(const mavl
...
+        default:
             //  MAV_RESULT_DENIED  means Command is invalid (is supported but has invalid parameters).
             return MAV_RESULT_DENIED;
         }
 
         plane.g2.guidedHeading.reset_I();
 
         plane.guided_state.target_heading = new_target_heading;
-        plane.guided_state.target_heading_accel_limit = MAX(packet.param3, 0.05f);
+        plane.guided_state.target_heading_accel_limit = MAX(is_zero(packet.param3)? 10.0f : packet.param3 , 10.0f); // the, previous limit of 0.05 was 0.29 degrees, not very useful
```
IamPete1, line 1000: Why this change?

timtuxworth: As per the comment @iampete // the, previous limit of 0.05 was 0.29 degrees, not very useful
The value is in radians so It was throttling the rate at which heading would change to very, very tiny increments that prevented the follow from  being able to track the target plane.

ArduPlane/GCS_Mavlink.cpp
```diff
@@ -973,22 +976,28 @@ MAV_RESULT GCS_MAVLINK_Plane::handle_command_int_guided_slew_commands(const mavl
 
         float new_target_heading = radians(wrap_180(packet.param2));
 
-        // course over ground
-        if ( int(packet.param1) == HEADING_TYPE_COURSE_OVER_GROUND) { // compare as nearest int
+        switch(HEADING_TYPE(packet.param1)) {
+        case HEADING_TYPE_COURSE_OVER_GROUND:
+            // course over ground
```
Ryanf55, line 981: I think it's safe to remove these comments, it's obvious from the case what this is for.

ArduPlane/GCS_Mavlink.cpp
IamPete1: It's a max not a min, it's saying the lowest it can be is  0.05 but it can be larger.

timtuxworth: > It's a max not a min, it's saying the lowest it can be is 0.05 but it can be larger.

Read that again - does it make sense to you?

IamPete1: > Read that again - does it make sense to you?

Yes, but then I'm dyslexic. Let me try again. 

Its a `MAX` operation. The output can be anything larger than 0.05.

Your complaint seems to be that 0.05 is too small? The `MAX` does not prevent you using a larger value.

ArduPlane/GCS_Mavlink.cpp
timtuxworth: I would but reviewers tend to not to like those kind of changes

IamPete1 (approved):
I think this looks good.

tridge (approved):

PR #28528, opened 2024-11-05
Tools: scripts: esp32_get_idf: Use break_system_packages

patrickelectric:
The only way to install the libraries on newer systems is using break-system-packages argument or env variable


tpwrules:
I do not like this, there should be a virtual environment active at this point I thought. But I use Nix to set up ESP-IDF rather than these scripts.

peterbarker (dismissed):
> The only way to install the libraries on newer systems is using break-system-packages argument or env variable

As @tpwrules says - modern systems should have a `venv` active into which these packages will be installed.

Not saying I like that, but this is where things have gone.

davidbuzz (approved):

peterbarker:
@patrickelectric thoughts on the "you should be in a venv" thing?

davidbuzz:
patrick , peter an i have all said this is ok to go in, despite it not explicity handling the venv situation, as it will work as-is for the default case of users-who-dont-have-a-venv, and users inside a venv can/should find their own solution.    please go ahead an merge it.    If someone fixes this script in the future to be smarter, that would be ok too, but shouldnt stop this going in.

tpwrules:
I still don't think it's okay to unsuspectingly break the user's system.

If I understand the install instructions correctly, and if they are followed correctly, a venv should be active at this point.

davidbuzz:
sorry, i tried to update this PR agains master ( 'python' vs 'python3' ) in the github gui / website  and it made a merge commit.  i think someones gonna have to squash and rebase this before CI will pass again.

davidbuzz:
@tpwrules  - are you volunteering to fix this so its venv aware and works under both situations.?   sounds to me like you are. :-)

peterbarker:
> I still don't think it's okay to unsuspectingly break the user's system.
> 
> If I understand the install instructions correctly, and if they are followed correctly, a venv should be active at this point.

I don't think I ended up with a venv on the VM I use for building esp32.

I'd be just as happy to add an assertion in here that you are in a `venv` if that's what our instructions specify!

I've rebased this anyway in case we just decide to hit merge!  @davidbuzz is maintainer, if he wants to ram this in... he gets to apologise to each and every owner of a broken packaging system ;-)

davidbuzz:
my current box, recently rebuilt , i ran the normal usbuntu prerequisites script and said yes to everything nd ended up with a venv, so i now this that this script should *not* go in as-is.  If it was able to detect teh presence of a virtual env and handndle both case, that would be the right move.   I'm removing my approval tickbox
(venv-ardupilot) buzz@foo:

peterbarker:
@patrickelectric thoughts?

khancyr:
I would also add that those package shouldn't be needed as they are already part of ardupilot dev env and not a requirement for idf (see the CI container : https://github.com/ArduPilot/ardupilot_dev_docker/blob/master/docker/Dockerfile_dev-esp32)

Note also that espressif add a .deb install, migrating to this may be a better move !

PR #28529, opened 2024-11-05
libraries: AP_HAL_ESP32: README: Add info about ESPPORT

patrickelectric:


tpwrules:
This is usually automatically determined but it's good to know it can be manually specified. Perhaps it should be a separate aside? I've also found that ESPBAUD is not necessary on ESP32-S3.

patrickelectric:
> This is usually automatically determined but it's good to know it can be manually specified. Perhaps it should be a separate aside? I've also found that ESPBAUD is not necessary on ESP32-S3.

I got a weird problem on my setup where it was necessary to setup the board device. I can break in a different example.

davidbuzz (approved):

tpwrules:
I changed the example because the given port won't work with ESP32S3 (it's usually `/dev/ttyACM0`).

peterbarker:
Note that we wouldn't ordinarily include "libraries: " in the commit message here.

PR #28530, opened 2024-11-05
AP_Mount: fix camera glitch when in sysid follow mode

timtuxworth:
### Summary

When using the MNTx_DEFLT_MODE = SysID Target, if the telemetry from the target vehicle isn't reliable or has high latency, the camera will follow the old location of the target vehicle, and then glitch as it catches up with a new location fairly violently. Here's a short video showing the problem.
https://youtu.be/DTtoUS021WQ

New Update

Fix AP_Mount's SYSID_TARGET mode pointing at a stale/lagged target location, and if relevant use AP_Follow's kinematic estimate to keep tracking smoothly through telemetry gaps when it's tracking the same vehicle.

### Classification & Testing (check all that apply and add your own)

- [ ] Non-functional change
- [ ] No-binary change
- [ ] Infrastructure change (e.g. unit tests, helper scripts)
- [x] Automated test(s) verify changes (e.g. unit test, autotest)
- [ ] Tested manually, description below (e.g. SITL)
- [ ] Tested on hardware
- [ ] Logs attached
- [ ] Logs available on request

Two new SITL autotest subtests were added to `test.Copter.Mount` (ArduCopter). Each was verified as a genuine regression test by reverting its corresponding part of the `AP_Mount` change and confirming the subtest fails, then restoring it and confirming it passes again.

Real-board flash cost was measured on CubeOrangePlus by diffing `./waf copter`/`./waf sub` binary sizes before/after the AP_Follow-integration part of the change: **+68 bytes** on ArduCopter (already links AP_Follow for its own Follow mode), **+292 bytes** on ArduSub (which doesn't otherwise use AP_Follow at all — confirmed via `nm` that only a small amount of reachable code gets newly linked in, not the whole library). Both negligible.

### Description

This supersedes the original (Nov 2024) version of this PR, which used a bare timeout only. rmackay9 and IamPete1 pointed out that a bare timeout helps when telemetry lags then catches up, but could make a target that is genuinely slow-moving (not just lagging) worse, by freezing instead of continuing to track — and suggested hooking into AP_Follow's velocity-based estimate instead. peterbarker asked in Oct 2025 to rebase this onto AP_Follow's kinematic work once it existed in master; that's now in master (since the 4.6 release), so this PR does that.

`get_angle_target_to_sysid()` previously always pointed at the last-received `GLOBAL_POSITION_INT` location for the followed vehicle, no matter how old, causing the mount to keep pointing at a stale location and then snap violently once fresh data arrived. This PR adds a 600ms staleness check as a baseline safety net: the mount now holds its last commanded angle instead of tracking a stale location.

On top of that, when `AP_Follow` is enabled and tracking the same sysid as the mount's SYSID_TARGET, the mount now prefers `AP_Follow`'s jerk/accel-limited kinematic extrapolation over the bare timeout above. This addresses the slow-moving-target case directly: instead of freezing, the mount keeps smoothly tracking through a telemetry gap using AP_Follow's velocity estimate. It falls back to the timeout behavior whenever AP_Follow isn't enabled, is tracking a different vehicle, or has no estimate yet.

AI-assisted: implementation and both autotest subtests were drafted with Claude Code (Anthropic). The author reviewed the diff, directed the design decisions (including scoping the AP_Follow integration and rejecting an initial flawed version of the tracking-vs-frozen test), and personally verified the testing and flash-size claims above by running the SITL autotests and board builds.

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Backend.cpp
+++ b/libraries/AP_Mount/AP_Mount_Backend.cpp
@@ -16,6 +16,7 @@ extern const AP_HAL::HAL& hal;
 #define AP_MOUNT_POI_REQUEST_TIMEOUT_MS 30000   // POI calculations continue to be updated for this many seconds after last request
 #define AP_MOUNT_POI_RESULT_TIMEOUT_MS  3000    // POI calculations valid for 3 seconds
 #define AP_MOUNT_POI_DIST_M_MAX         10000   // POI calculations limit of 10,000m (10km)
+#define AP_MOUNT_SYSID_TIMEOUT_MS       3000    // sysid target location considered stale if not updated within this many ms (matches AP_Follow's own default FOLL_TIMEOUT)
 
 // Default init function for every mount
 void AP_Mount_Backend::init()
@@ -396,6 +397,12 @@ void AP_Mount_Backend::set_roi_target_wpnext_offset(const Vector3f &rpy)
 // set_sys_target - sets system that mount should attempt to point towards
 void AP_Mount_Backend::set_target_sysid(uint8_t sysid)
 {
+    if (sysid != _target_sysid) {
+        // forget the previous target's location so it isn't briefly
+        // reported as the new target's (still-fresh) position
+        _target_sysid_location.zero();
+        _target_sysid_update_ms = 0;
+    }
     _target_sysid = sysid;
 
     // set the mode to sysid tracking mode
@@ -657,6 +664,7 @@ bool AP_Mount_Backend::handle_global_position_int(uint8_t msg_sysid, const mavli
     _target_sysid_location.lng = packet.lon;
     // global_position_int.alt is *UP*, so is location.
     _target_sysid_location.set_alt_cm(packet.alt*0.1, Location::AltFrame::ABSOLUTE);
+    _target_sysid_update_ms = AP_HAL::millis();
 
     return true;
 }
@@ -1223,6 +1231,10 @@ bool AP_Mount_Backend::get_angle_target_to_sysid(MountAngleTarget& angle_rad) co
         return false;
     }
     if (!_target_sysid) {
+        return false;
+    }
+    // exit if we haven't heard from the target recently, to avoid snapping to a stale location
+    if (AP_HAL::millis() - _target_sysid_update_ms > AP_MOUNT_SYSID_TIMEOUT_MS) {
         return false;
     }
     return get_angle_target_to_location(_target_sysid_location, angle_rad);
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -8116,6 +8116,44 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
                 constrained=constrain_sysid_target,
             )
 
+            self.progress("Testing mount holds last angle once sysid target telemetry goes stale")
+            pre_stale_pitch = self.get_mount_roll_pitch_yaw_deg()[1]
+            self.delay_sim_time(3.5, reason="let sysid target telemetry go stale")
+            # move a long way in a direction that would swing the
+            # elevation angle a lot if the mount were still (incorrectly)
+            # tracking the stale target location as we move; it should
+            # instead hold the angle last commanded above
+            startpos = self.assert_receive_message('LOCAL_POSITION_NED')
+            orig_x, orig_y, orig_z = startpos.x, startpos.y, startpos.z
+            self.fly_guided_move_local(orig_x, orig_y + 80, -orig_z, timeout=60)
+            self.test_mount_pitch(
+                pre_stale_pitch,
+                3,
+                mavutil.mavlink.MAV_MOUNT_MODE_SYSID_TARGET,
+                timeout=5,
+                hold=2,
+                constrained=False,
+            )
+
+            # move back to the original position, then confirm fresh
+            # telemetry resumes tracking immediately (ie. not stuck forever)
+            self.fly_guided_move_local(orig_x, orig_y, -orig_z, timeout=60)
+            self.mav.mav.global_position_int_send(
+                0, # time boot ms
+                int(roi_lat * 1e7),
+                int(roi_lon * 1e7),
+                670 * 1000, # mm alt amsl
+                100 * 1000, # mm UP!
+                0, # vx
+                0, # vy
+                0, # vz
+                0 # heading
+            )
+            self.test_mount_pitch(
+                68, 5, mavutil.mavlink.MAV_MOUNT_MODE_SYSID_TARGET, hold=1,
+                constrained=constrain_sysid_target,
+            )
+
             self.set_mount_mode(mavutil.mavlink.MAV_MOUNT_MODE_NEUTRAL)
             self.test_mount_pitch(0, 0.1, mavutil.mavlink.MAV_MOUNT_MODE_NEUTRAL)
 
--- a/libraries/AP_Mount/AP_Mount_Backend.h
+++ b/libraries/AP_Mount/AP_Mount_Backend.h
@@ -485,6 +485,7 @@ private:
 
     uint8_t _target_sysid;          // sysid to track
     Location _target_sysid_location;// sysid target location
+    uint32_t _target_sysid_update_ms;// system time (ms) _target_sysid_location was last updated
 
     uint32_t _last_warning_ms;      // system time of last warning sent to GCS
 


timtuxworth:
This is the mount glitch fix I mentioned at the conference @peterbarker. This definitely works (I've tested it).

I haven't been able to test setting the mount to rate mode yet, but I think it should make for an even smoother ride.

Could I ask you to give it a once over too please @IamPete1 ?

timtuxworth:
This is the mount glitch fix I mentioned at the conference @peterbarker. This definitely works (I've tested it).

I haven't been able to test setting the mount to rate mode yet, but I think it should make for an even smoother ride.

Could I ask you to give it a once over too please @IamPete1 ?

rmackay9:
Thanks for this.  I don't think setting MNTx_RC_RATE will help because that parameter only affects how the pilot's inputs are interpreted when the mount is in RC_Targeting mode.

I think to improve this we should improve how AP_Mount is estimating the position of the thing it is following.  At the moment it just stores the last known location.  AP_Follow has a better method where it also records the velocity and then updates the target location using that velocity between updates.

So perhaps we should change AP_Mount to re-use AP_Follow's calculations somehow.  E.g. duplicate the calcs or somehow get the location from AP_Follow

timtuxworth:
> Thanks for this. I don't think setting MNTx_RC_RATE will help because that parameter only affects how the pilot's inputs are interpreted when the mount is in RC_Targeting mode.
> 
> I think to improve this we should improve how AP_Mount is estimating the position of the thing it is following. At the moment it just stores the last known location. AP_Follow has a better method where it also records the velocity and then updates the target location using that velocity between updates.
> 
> So perhaps we should change AP_Mount to re-use AP_Follow's calculations somehow. E.g. duplicate the calcs or somehow get the location from AP_Follow

Yes that makes sense and will make a lot more sense when @lhall and @peterbarker add their fancy new "kinetic interpolation" to the follow code. Can we include this and then add the follow integration in a future PR?

This PR now uses AP_Follow when the MNTx_SYSID_DFLT or the ROI sysid sent by the GCS matches the FOLL_SYSID.

IamPete1:
I'm not sure this is a improvement in all cases, it really depends on how the target is moving. If the target is moving slowly then it makes sense to keep pointing at the last known location. 

In your example the following vehicle rapidly catches up with the target resulting in rapid angle change as the follower flys past the targets last location. 

If you were following a stationary vehicle whilst flying round it in a circle then this would be much worse. 

Hopefully hooking into the new follow stuff will give good results for both use-cases.

timtuxworth:
> I'm not sure this is a improvement in all cases, it really depends on how the target is moving. If the target is moving slowly then it makes sense to keep pointing at the last known location.

Have you tried it? I have - I saw the same glitching with an EDU-650 following a QAV-250.

> In your example the following vehicle rapidly catches up with the target resulting in rapid angle change as the follower flys past the targets last location.

Not just that - you also see it if the vehicles are flying at the same speed, but telemetry messages get dropped.
 
> If you were following a stationary vehicle whilst flying round it in a circle then this would be much worse.

I guess - but wouldn't you just use a POI in that case?

> Hopefully hooking into the new follow stuff will give good results for both use-cases.

I think this is an improvement now, and it can improve even more when the new follow stuff arrives.

peterbarker:
Please re-rest on top of the kinematic stuff in AP_Follow now.

timtuxworth:
I had Codex review it which surfaced three further gaps in the fallback path, all fixed:
- freshness was tracked by local packet-arrival time rather than the data's own timestamp, so a delayed/backlogged transport could keep "refreshing" a stale position indefinitely — now uses the target's jitter-corrected timestamp, and rejects an update outright if its timestamp isn't actually newer than what's already held (falling back to arrival time when a sender doesn't populate `time_boot_ms`);
- retargeting to a different sysid didn't clear the previous target's location, so it could briefly be reported as the new target's position;
- the original 600ms timeout was too tight for legitimate low-rate telemetry (e.g. 1Hz), causing visible freeze-then-step stutter — bumped to 3s to match `AP_Follow`'s own default.

tridge:

<details><summary>Previous review</summary>

Full report: https://uav.tridgell.net/DevCallReviews/2026_08_30/devcall_pr_reviews.html#pr28530

Reviewed at head `f19a4b5b55`. **Verdict: COMMENT** — the code is correct and well-tested (CI 101/101 green, and the new autotest subtests are genuinely good coverage), with one fixable issue and two items that want an explicit maintainer decision:

1. **ISSUE** — `libraries/AP_Mount/AP_Mount_Backend.cpp:399-407`: `set_target_sysid()` zeroes the location and update time but not `_target_sysid_jitter`, so switching to a target whose `time_boot_ms` is behind the previous one's inherits the old clock offset — the max-lag clamp stamps the new target's updates ~3 s in the past and the raw-location path reports stale until the correction re-converges (potentially 100+ packets). Cheap fix: reset the jitter corrector there too. https://github.com/ArduPilot/ardupilot/blob/f19a4b5b55338d630b7435d403688d4a90859e8d/libraries/AP_Mount/AP_Mount_Backend.cpp#L399
2. **Design sign-off wanted** — `AP_Mount_Backend.cpp:1277`: after 3 s without updates and no AP_Follow match, the mount now holds last *angle* rather than tracking last *location* forever — worse for a stationary target with lapsed telemetry while the camera vehicle moves (IamPete1's original 2024 concern). Related: when AP_Follow matches, it is preferred even over a just-received raw update, which also routes altitude through AP_Follow's frame handling. Both are defensible and the in-code rationale is good, but they change behaviour for existing users — worth an explicit ack from the mount maintainers plus a wiki/release note.
3. **NOTE** — timestamp edge cases, all self-limiting: a tracked-vehicle reboot freezes the mount ~3 s; a sender with a constant non-zero `time_boot_ms` is rejected until the clamp catches up (~3 s) then rides the staleness boundary; `AP_MOUNT_SYSID_TIMEOUT_MS` is hardcoded 3000 so it won't track a user-shortened FOLL_TIMEOUT.

Verified clean: all claimed API signatures, side-effect-free false-returns (frontend ignores the return), retarget location clearing, wraparound-safe compares, all mount drivers benefiting via the shared backend, non-SYSID modes untouched. An independent Codex cross-check ran; its escalation to REQUEST CHANGES was not adopted (the confirmed defect is a transient with a trivial fix; the rest are policy calls).

</details>

libraries/AP_Mount/AP_Mount_Backend.cpp
```diff
@@ -1218,11 +1247,34 @@ bool AP_Mount_Backend::get_angle_target_to_wpnext_offset(MountAngleTarget& angle
...
+    }
+#endif
+
+    // exit immediately if no location available
     if (!_target_sysid_location.initialised()) {
         return false;
     }
-    if (!_target_sysid) {
+    // exit if we haven't heard from the target recently, to avoid snapping to a stale location
+    if (AP_HAL::millis() - _target_sysid_update_ms > AP_MOUNT_SYSID_TIMEOUT_MS) {
```
rmackay9, line 1237: Thanks for this.  I think there are two separate changes here:

1. handle timeout of the global-position-int message
2. use Follow mode's target

Could we reduce the scope of this PR to just cover the 1st and then create a separate PR that adds support for using AP_Follow?

For the 1st fix, the changes here are very close but just need a few changes:

1. remove JitterCorrection and calculation of "update_ms".  This is overkill for a simple timeout
2. remove the use of AP_Follow

timtuxworth: > Thanks for this. I think there are two separate changes here:
> 
> 1. handle timeout of the global-position-int message
> 2. use Follow mode's target
> 
> Could we reduce the scope of this PR to just cover the 1st and then create a separate PR that adds support for using AP_Follow?
> 
> For the 1st fix, the changes here are very close but just need a few changes:
> 
> 1. remove JitterCorrection and calculation of "update_ms".  This is overkill for a simple timeout
> 2. remove the use of AP_Follow

Thanks for the review @rmackay9 . I've reduced the scope of this PR as you suggested and created a new pr #34237  for the Foliow mode Target function.

tridge:
Full report: https://uav.tridgell.net/DevCallReviews/2026_09_01_AIReview/devcall_pr_reviews.html#pr28530

Reviewed at head `0c21728f92` — verdict **COMMENT**. The code is correct. The PR description, however, describes work that is not in this branch - it belongs to the stacked follow-on - and states a timeout value five times smaller than the one actually added.

1. **ISSUE** — **The description does not match the diff.** It says the PR adds "a **600ms** staleness check" (the code defines `AP_MOUNT_SYSID_TIMEOUT_MS 3000`), describes AP_Follow kinematic extrapolation at length (there is none here — that is #34237, stacked on top), claims "**Two** new SITL autotest subtests" (the diff adds one), and reports a flash cost measured "for the AP_Follow-integration part of the change", which is likewise not in this branch. It reads as the description of the combined work written before the split. Worth correcting before merge: a reviewer looking for the AP_Follow path will not find it, and the 600 vs 3000 ms discrepancy invites a wrong conclusion about the constant.

Non-blocking notes:

2. **NOTE** — Staleness is measured from packet _arrival_ (`_target_sysid_update_ms = AP_HAL::millis()`), not from `GLOBAL_POSITION_INT.time_boot_ms`, so a continuously delayed or replayed stream arriving faster than every 3 s keeps refreshing the timestamp and never trips the timeout. That is the lagging-telemetry case rather than the loss case. The stacked #34237 addresses it via AP_Follow's jitter-corrected estimate, so this is a limitation of this PR alone rather than of the pair.

3. **NOTE** — The timeout constant's comment ("matches AP_Follow's own default FOLL_TIMEOUT") is accurate — `AP_FOLLOW_TIMEOUT` is 3 seconds. But AP_Follow's own check pairs that timeout with an explicit `_last_location_update_ms == 0` guard, which the mount's version omits. No bug is reachable today, because the pre-existing `!_target_sysid_location.initialised()` guard covers the never-updated case (and `set_target_sysid()` now keeps it true by zeroing the location). Adding the `== 0` test would make the check self-sufficient rather than dependent on a separate guard holding.

CI at this head: 84 pending, 1 passing.

rmackay9 (approved):
This looks good to me now, thanks!

peterbarker (approved):
LGTM

peterbarker:
Your opening comment appears to be woefully out of date - it's talking about stuff simply not present in this PR.

peterbarker:
@timtuxworth so you force-pushed over a failure in `clear_roi`.  I asked claude:

```
  The CI answer: clear_roi is not failing upstream

  Over 1000 master runs of the copter workflow, 2025-12-04 to 2026-09-01 — 367 successes, 585 cancelled (superseded pushes), 48 failures.

  Of the failures I've annotated so far, the sitltest-copter-tests1e shard failed 6 times in the most recent ~3 months, and none was 
  clear_roi:
```

It *may* just be a rare flake, but suspicious it popped up in a mount-related change.

I do have a fix which *might* explain the flake: https://github.com/ArduPilot/ardupilot/pull/34255

rmackay9:
Hi @peterbarker, I can definitely imagine this would cause the test to fail is the autotest doesn't send the Location regularly and instead relies on being able to just send it once.

timtuxworth:
All checks pass @rmackay9 @peterbarker

rmackay9:
Merged, thanks!

PR #28540, opened 2024-11-06
AP_Boootloader: board_types.txt: correct company name for V-UAV reser…

peterbarker:
…vations


LCaiJie:
Thank you for your correction

PR #28541, opened 2024-11-06
hwdef: use V-UAV in place of VIEWPRO for link text

peterbarker:


LCaiJie:
Thank you for your correction,Can it be merged into other branches?

PR #28544, opened 2024-11-07
Ros jazzy support

pedro-fuoco:
Chore from ardupilot/ardupilot_wiki#6378


Tools/ros2/README.md
```diff
@@ -36,163 +36,7 @@ ArduPilot `AP_DDS` client library.
...
-```
-
-The remaining steps 1 - 5 are the same as for Ubuntu. You may need to
-install MAVProxy if it is not available on the container.
-
-
-```bash
-pip install -U MAVProxy
-```
+Follow the installation instructions in the [ArduPilot Wiki](https://ardupilot.org/dev/docs/ros2.html)
```
pedro-fuoco, line 39: Keeping two documentations might cause confusion/out-of-sync problems. Now that we have a Wiki Page, shouldn't we link to it ?

Tools/ros2/jazzy_macos.repos
```diff
@@ -0,0 +1,17 @@
+repositories:
+  ardupilot:
+    type: git
+    url: https://github.com/srmainwaring/ardupilot.git
+    version: pr/pr-dds-launch-tests
```
pedro-fuoco, line 5: The macOS `.repos` file in `ardupilot_gz` is using `master` instead. Is this specific version still required ? @srmainwaring

Tools/ros2/jazzy_macos.repos
```diff
@@ -0,0 +1,17 @@
...
+  microxrcedds_gen:
+    type: git
+    url: https://github.com/eProsima/Micro-XRCE-DDS-Gen.git
+    version: develop
+  micro_ros_agent:
+    type: git
+    url: https://github.com/srmainwaring/micro-ROS-Agent.git
+    version: cmake-macos
```
pedro-fuoco, line 13: Will the fork still be required in `Jazzy` ? @srmainwaring 
I've also notices that this branch is different than the one in [`ardupilot_gz`](https://github.com/ArduPilot/ardupilot_gz/blob/main/ros2_gz_macos.repos).

PR #28546, opened 2024-11-07
Plane follow script Applet

timtuxworth:
This creates a new Lua Script applet that implements "plane follow" - a fixed wing follow/chase functionality that runs in guided mode. The code uses the core AP_Follow library functionality and therefore uses most of the documented FOLL_ parameters that are used in Copter and Rover FOLLOW mode.

This PR depends on 
https://github.com/ArduPilot/ardupilot/pull/28526 
https://github.com/ArduPilot/mavlink/pull/372
https://github.com/ArduPilot/ardupilot/pull/28527


libraries/AP_Scripting/modules/speedpi.lua
```diff
@@ -0,0 +1,156 @@
...
+    local _current = 0
+    local nowPI = millis():tofloat() * 0.001
+ 
+    -- update the controller.
+    function self.update(target, current)
+       local now = millis():tofloat() * 0.001
+       if not _last_t then
+          _last_t = now
+       end
+       local dt = now - _last_t
```
IamPete1, line 75:
```suggestion
       local now = millis()
       if not _last_t then
          _last_t = now
       end
       local dt = (now - _last_t):tofloat() * 0.001
```
Better to do the conversion to float on the dt rather than the millis, dt is a small magnitude so you get good float resolution. If you do it on the millis it gets worse resolution the longer you run. 

Also if your running more than about 10Hz its worth swapping to `micros` and `* 0.000001`

timtuxworth: This change breaks the script due to other uses of "now" as a float lower down. I think I've done it right but could you LMK if I've made any obvious mistakes. I've run it in SITL and it seems to be working.

timtuxworth:
Could you take a look @peterbarker and @tridge ?

timtuxworth:
I wonder if you'd be able to take a look at my Lua @yuri-rage - any suggestions appreciated.

yuri-rage:
> I wonder if you'd be able to take a look at my Lua @yuri-rage - any suggestions appreciated.

I'd be happy to take a look, Tim, but I have some external commitments upcoming that may prevent me from diving in right away. Please remind me either here, Discuss, or Discord if I don't reply in a week.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -129,7 +132,7 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
     // @Param: _ALT_TYPE
     // @DisplayName: Follow altitude type
     // @Description: Follow altitude type
-    // @Values: 0:absolute, 1:relative
+    // @Values: 0:absolute, 1:relative, 2:origin (not used), 3:terrain (plane)
```
IamPete1, line 135: Why add a unused option?

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -141,6 +144,13 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("_OPTIONS", 11, AP_Follow, _options, 0),
 
+    // @Param: _TIMEOUT
+    // @DisplayName: Follow timeout
+    // @Description: Follow position update from lead - timeout after x milliseconds
```
IamPete1, line 149: might as well go to seconds?

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -340,19 +351,45 @@ bool AP_Follow::handle_global_position_int_message(const mavlink_message_t &msg)
...
         _target_location.lng = packet.lon;
 
         // select altitude source based on FOLL_ALT_TYPE param 
+#if APM_BUILD_TYPE(APM_BUILD_ArduPlane)
+        switch(_alt_type) {
+            case AP_FOLLOW_ALT_TYPE_DEFAULT:
+                _target_location.set_alt_cm(packet.alt * 0.1, Location::AltFrame::ABSOLUTE);
+                break;
+            case AP_FOLLOW_ALTITUDE_TYPE_RELATIVE:
+            case AP_FOLLOW_ALTITUDE_TYPE_ORIGIN:
```
IamPete1, line 370: Remove origin case?

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -559,11 +596,41 @@ bool AP_Follow::have_target(void) const
...
+   local xy_dist = follow:get_distance_to_target() -- this value is set by get_target_dist_and_vel_ned() - why do I have to know this?
+   local target_heading = follow:get_target_heading_deg()
+*/
+bool AP_Follow::get_target_info(Vector3f &dist_ned, Vector3f &dist_with_offs,
+                                Vector3f &target_vel_ned, Vector3f &target_vel_ned_ofs,
+                                Location &target_loc, Location &target_loc_ofs,
+                                float &target_dist_ofs
+                                )
+{
+    // The order here is VERY important. This needs to called first because it updates a lot of internal variables
```
IamPete1, line 620: If the ordering is important does that mean the existing bindings make it possible for lua users to get it wrong?

The lua calls really should not be changing the internal follow state. Maybe if we fixed that this new method would not be needed?

libraries/AP_Follow/AP_Follow.h
```diff
@@ -26,6 +26,8 @@
 #include <AC_PID/AC_P.h>
 #include <AP_RTC/JitterCorrection.h>
 
+#include <AP_Vehicle/AP_Vehicle_Type.h>
```
IamPete1, line 29: Should be the CPP.

libraries/AP_Follow/AP_Follow.cpp
timtuxworth: So that the numbers match Location:AltFrame

libraries/AP_Follow/AP_Follow.cpp
timtuxworth: As per above - this means the ordinal values match Location:AltFrame, which is kind of useful.

libraries/AP_Follow/AP_Follow.cpp
IamPete1: Sure, but you don't need to call out 2 in the description. That will populate a drop down that will allow people to select it.

libraries/AP_Follow/AP_Follow.cpp
timtuxworth: > If the ordering is important does that mean the existing bindings make it possible for lua users to get it wrong?

Absolutely yes.

> The lua calls really should not be changing the internal follow state. Maybe if we fixed that this new method would not be needed?

I think that's a good idea, but outside the scope of this PR? Happy to refactor later.

ArduPlane/Parameters.cpp
```diff
@@ -1325,6 +1325,13 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     AP_SUBGROUPINFO(systemid, "SID", 38, ParametersG2, AP_SystemID),
 #endif
     
+    // @Param: GUIDED_UPD_LIM
+    // @DisplayName: Guided Update Limit
+    // @Description: The maximum frequency that an guided mode commands sent by external system such as lua or mavlink can update roll, pitch and throttle.
```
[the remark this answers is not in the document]
timtuxworth: I'd be happy of @Ryanf55 's PR https://github.com/ArduPilot/ardupilot/pull/28943 was merged then I'd use that.

libraries/AP_Follow/AP_Follow.cpp
IamPete1: Sure, but it doesn't do relative to origin, it should return false out of the function as a invalid setting. Or it should come under the a default case in the switch. Adding a enum value for origin but then using it for relative home is just confusing.

libraries/AP_Follow/AP_Follow.cpp
IamPete1: If we did the refactor would that remove the need for the new method?

ArduPlane/commands.cpp
```diff
@@ -86,8 +86,11 @@ void Plane::set_guided_WP(const Location &loc)
     next_WP_loc = loc;
 
     // used to control FBW and limit the rate of climb
+    // cannot be applied in Guided as this prevents the guided set altitude from working correctly
```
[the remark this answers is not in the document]
timtuxworth: I guess it's because this code assumes that GUIDED is always running in "location" sub-mode (i.e. flying towards a location), but my follow code needs to switch between "location" sub-mode and "heading" sub mode. This line prevents the target altitude being set when switching between sub-modes.

libraries/AP_Follow/AP_Follow.cpp
timtuxworth: Copter (Rover?) uses a hardcoded timeout of 600ms for this value.

libraries/AP_Follow/AP_Follow.cpp
timtuxworth: Well it still leaves the case that if you call multiple methods Lua might get values from different loops because of the lower priority scheduling in Lua. One method means you are guaranteed to get a consistent set of values. 

One method is also a lower flash cost for the binding I think.

ArduPlane/Parameters.cpp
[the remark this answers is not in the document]
tridge: also, it is milliseconds (default 3000), which is not a frequency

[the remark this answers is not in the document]
tridge: what value are you using?

libraries/AP_Follow/AP_Follow.cpp
tridge: I'd prefer seconds too

libraries/AP_Follow/AP_Follow.h
```diff
@@ -153,10 +165,12 @@ class AP_Follow
     AP_Int8     _alt_type;          // altitude source for follow mode
```
peterbarker, line 165: Maybe change this to `AP_Enum<Location::AltType>` - it should clean up a bunch of code.

ArduPlane/Parameters.cpp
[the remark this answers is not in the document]
timtuxworth: Changed to GUIDED_TIMEOUT and updated the comments to reflect that it's a timeout not a frequency.

timtuxworth (commented):
Implemented changes from the Dev call on 31 March 2025

ArduPlane/Parameters.h
```diff
@@ -574,6 +574,8 @@ class ParametersG2 {
     AP_Follow follow;
 #endif
 
+    AP_Int32        guided_timeout;
```
IamPete1, line 577: `AP_Float` please.

ArduPlane/commands.cpp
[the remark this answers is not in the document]
tridge: I think this would prevent a simple mode change to GUIDED from locking the current alt

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -263,7 +273,8 @@ bool AP_Follow::get_target_heading_deg(float &heading) const
     }
 
     // check for timeout
-    if ((_last_heading_update_ms == 0) || (AP_HAL::millis() - _last_heading_update_ms > AP_FOLLOW_TIMEOUT_MS)) {
+    if ((_last_heading_update_ms == 0) || (AP_HAL::millis() - _last_heading_update_ms > (uint32_t)_timeout_ms)) {
+        gcs().send_text(MAV_SEVERITY_NOTICE, "gthd timeout %d", (int)(AP_HAL::millis() - _last_location_update_ms));
```
tridge, line 277: could flood the link with msgs?

tridge (changes requested):

libraries/AP_Follow/AP_Follow.h
```diff
@@ -150,13 +160,15 @@ class AP_Follow
     AP_Int8     _offset_type;       // offset frame type (0:North-East-Down, 1:RelativeToLeadVehicleHeading)
     AP_Vector3f _offset;            // offset from lead vehicle in meters
     AP_Int8     _yaw_behave;        // following vehicle's yaw/heading behaviour (see YAW_BEHAVE enum)
-    AP_Int8     _alt_type;          // altitude source for follow mode
+    AP_Enum<Location::AltFrame>    _alt_type;          // altitude source for follow mode
     AC_P        _p_pos;             // position error P controller
     AP_Int16    _options;           // options for mount behaviour follow mode
+    AP_Int32    _timeout_ms;        // position estimate timeout after x milliseconds
```
IamPete1, line 166: `AP_Float` and update comment for seconds.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -340,19 +351,51 @@ bool AP_Follow::handle_global_position_int_message(const mavlink_message_t &msg)
         if ((packet.lat == 0 && packet.lon == 0)) {
             return false;
         }
+        // ignore message if we already received a more recent one
+        if(_target_location_last_time_boot_ms != 0 && packet.time_boot_ms <= _target_location_last_time_boot_ms ) {
```
IamPete1, line 355: What happens if the follow target vehicle is rebooted? Its boot time will go backwards so it will be ignored. Maybe if the time is more than 5 or 10 seconds in the past it would reset the boot time.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -559,12 +603,42 @@ bool AP_Follow::have_target(void) const
...
+#if APM_BUILD_TYPE(APM_BUILD_ArduPlane)
+// create a single method to retrieve all the relevant values in one shot for Lua
+/* replaces the following Lua calls
+   target_distance, target_distance_offsets, target_velocity = follow:get_target_dist_and_vel_ned()
+   target_location, target_velocity = follow:get_target_location_and_velocity()
+   target_location_offset, target_velocity = follow:get_target_location_and_velocity_ofs()
+   local xy_dist = follow:get_distance_to_target() -- this value is set by get_target_dist_and_vel_ned() - why do I have to know this?
+   local target_heading = follow:get_target_heading_deg()
+*/
+bool AP_Follow::get_target_info(Vector3f &dist_ned, Vector3f &dist_with_offs,
```
tridge, line 621: I find it odd to be fetching both with and without the offset, why would a consumer want both?

tridge (changes requested):

libraries/AP_Follow/AP_Follow.cpp
tridge: I think we need to fix the API so you don't need to call function A in order to get the correct value for function B

timtuxworth:
Replaced by #30654

PR #28547, opened 2024-11-07
Tools: Checksum for gcc-arm download on arch prereqs

jmcudd:
This will check to see if the tar.bz2 file exists and if it does it will run a checksum and skip redownloading the file if its already there. If the checksum fails or the file doesn't exist it will redownload the file.

I ran into issues with the download taking so long that my sudo permissions timed out and the install failed to complete. When rerunning the script it would redownload the file even if the file was already there. This change solves this issue.

### DIFF
--- a/Tools/environment_install/install-prereqs-arch.sh
+++ b/Tools/environment_install/install-prereqs-arch.sh
@@ -33,6 +33,7 @@ PYTHON_PKGS="future lxml pymavlink MAVProxy pexpect argparse matplotlib pyparsin
 ARM_ROOT="gcc-arm-none-eabi-10-2020-q4-major"
 ARM_TARBALL="$ARM_ROOT-x86_64-linux.tar.bz2"
 ARM_TARBALL_URL="https://firmware.ardupilot.org/Tools/STM32-tools/$ARM_TARBALL"
+ARM_TARBALL_CHECKSUM="21134caa478bbf5352e239fbc6e2da3038f8d2207e089efc96c3b55f1edcd618" 
 
 # Ardupilot Tools
 ARDUPILOT_TOOLS="ardupilot/Tools/autotest"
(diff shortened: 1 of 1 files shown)


peterbarker (changes requested):
Please prefix the commit message with `Tools: `

jmcudd:
@peterbarker Done

Tools/environment_install/install-prereqs-arch.sh
```diff
@@ -33,6 +33,7 @@ PYTHON_PKGS="future lxml pymavlink MAVProxy pexpect argparse matplotlib pyparsin
 ARM_ROOT="gcc-arm-none-eabi-10-2020-q4-major"
 ARM_TARBALL="$ARM_ROOT-x86_64-linux.tar.bz2"
 ARM_TARBALL_URL="https://firmware.ardupilot.org/Tools/STM32-tools/$ARM_TARBALL"
+ARM_TARBALL_CHECKSUM="21134caa478bbf5352e239fbc6e2da3038f8d2207e089efc96c3b55f1edcd618"
```
khancyr, line 36: @peterbarker that is totally something we will forget to update. Can't we add a checksum file on the server instead ?

peterbarker: @khancyr I really did consider doing that - including getting a shell on the server to create the file.

Then I thought again.  We will *not* be updating any package on the server to have a different checksum - we never have, and I can't see why we would.  Everything's a mirror of something someone else has done, and we're not stripping version information from the file we're downloading or anything along those lines.

I decided on balance that dropping the expected checksum in here was fine, and that I didn't need to complicate things!

peterbarker: ... I also considered how often we update our release compiler and decided it was unlikely it would be a massive problem.

peterbarker:
I've merged this as solving an immediate problem, but happy to discuss further the option of downloading a sha checksum file instead.

PR #28548, opened 2024-11-07
AP_NavEKF3: ignore VelZ of body odometry if SRCn_VELZ is set to None

Williangalvani:
This was found by using a dvl which was outputting bad VISION_POSITION_DELTA messages. it was losing locks and outputting bogus (0,0,0) readings fox xyz speed.
Cosuming the bad data caused the baro depth to be somewhat ignored and the vehicle to bounce a lot.

I'm actually not sure of how right this is. is the time of stateStrct the same of bodyOdmDelayed?


rmackay9 (changes requested):
thanks for finding this and having the courage to dig into the EKF.

I think though that it would be better to add some arguments to the FuseBodyVel() method which specifies whether XY and/or Z velocities are fused.  The reason is that I think there will be some effects on the EKF's estimate fusing the Z-axis even if it is exactly the same as its current estimate.  It will probably result in the EKF becoming overly confident of its estimate

Williangalvani:
done better in https://github.com/ArduPilot/ardupilot/pull/29213

PR #28549, opened 2024-11-07
AP_Volz_Protocol: bugfix with scaling integer

DjMixMasterDragon:
Bugfix to the recent Volz protocol implementation. _motor_temp_deg_ and _pcb_temp_deg_ are both defined as uint16_t which is valid pre-scaling (ie. receiving the message from the buffer) but with scaling being done below (range -49 to +204) this will need to account for signed values.

```
case CMD_ID::TEMPERATURE_RESPONSE:
      // Temperature is reported relative to -50 deg C
      telem.data[index].motor_temp_deg = -50 + cmd.arg1;
      telem.data[index].pcb_temp_deg = -50 + cmd.arg2;
      break;
```

### DIFF
--- a/libraries/AP_Volz_Protocol/AP_Volz_Protocol.cpp
+++ b/libraries/AP_Volz_Protocol/AP_Volz_Protocol.cpp
@@ -405,7 +405,7 @@ void AP_Volz_Protocol::update()
                 "TimeUS,I,Dang,ang,pc,sc,pv,sv,mt,pt",
                 "s#ddAAvvOO",
                 "F000000000",
-                "QBffffffHH",
+                "QBffffffhh",
                 AP_HAL::micros64(),
                 i + 1, // convert to 1 indexed to match actuator IDs and SERVOx numbering
                 telem.data[i].desired_angle,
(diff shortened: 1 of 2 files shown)


IamPete1:
@DjMixMasterDragon Well spotted, it not been cold enough for me too see this IRL. Have you been testing with hardware? How have you found the updated library? I have only been able to test on one model number. 

The logging format type also need changing to match here:

https://github.com/ArduPilot/ardupilot/blob/874d268aa636eb5cadc0bf0de9d9ffa2d289075d/libraries/AP_Volz_Protocol/AP_Volz_Protocol.cpp#L408

From `H` to `h`.

DjMixMasterDragon:
@IamPete1 Yes - specifically testing with the DA15-N High Toque model. Like you, I have yet to get the temperature cold enough. We are planning on getting our Drone into a environmental chamber that gets down to -70degC, so hoping to get some colder temperatures reported! 

Updated logging format as advised - good catch.

IamPete1 (approved):
Looks good!

IamPete1:
Thanks!

PR #28557, opened 2024-11-08
Tools: Recommend what to do when astyle fails

Ryanf55:
# Purpose

Tell developers what to do if the astyle check fails. 

# Motivation

https://github.com/ArduPilot/ardupilot_wiki/pull/6292#discussion_r1777897717

# Demo

This is what a developer will see if they push code and CI fails. 

Specifically, we went from:
```
****** Files needing formatting found.
```
to:
```
****** Files needing formatting found. Please run ./Tools/scripts/run_astyle.py
```

In the screenshot below, I intentionally broke formatting in AP_DDS_Client.h. This is what the script will display in CI.
image

### DIFF
--- a/Tools/scripts/run_astyle.py
+++ b/Tools/scripts/run_astyle.py
@@ -60,7 +60,7 @@ class AStyleChecker(object):
             self.progress("astyle check failed: (%s)" % (ret.stdout))
             self.retcode = 1
         if "Formatted" in ret.stdout:
-            self.progress("Files needing formatting found")
+            self.progress("Files needing formatting found.")
             print(ret.stdout)
             self.retcode = 1
 
--- a/Tools/scripts/build_ci.sh
+++ b/Tools/scripts/build_ci.sh
@@ -473,7 +473,11 @@ for t in $CI_BUILD_TARGET; do
 
     if [ "$t" == "astyle-cleanliness" ]; then
         echo "Checking AStyle code cleanliness"
+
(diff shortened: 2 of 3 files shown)


Tools/scripts/run_astyle.py
```diff
@@ -60,7 +60,7 @@ def check(self):
             self.progress("astyle check failed: (%s)" % (ret.stdout))
             self.retcode = 1
         if "Formatted" in ret.stdout:
-            self.progress("Files needing formatting found")
+            self.progress("Files needing formatting found. Please run ./Tools/scripts/run_astyle.py")
```
peterbarker, line 63: This is a really strange thing to be printing in `run_astyle.py`!

If this is intended to make CI clearer, grepping the output of the dry-run might be a better idea.

peterbarker (changes requested):

Tools/scripts/run_astyle.py
Ryanf55: Is your expectation that we only print this in `build_ci.sh`?

peterbarker: Yes, that's my expectation - don't confuse the user who is running this manually.

Ryanf55: Please look again. I've moved this message to build-ci.sh and check the error code.

peterbarker (approved):
LGTM

PR #28561, opened 2024-11-08
Tools: Add astyle dependency

Ryanf55:
Experimenting to see if we can install `astyle` by default and start enabling astyle cleanlyness on Ubuntu.


Ryanf55:
To the dev team: What do you want tested before we would merge this?

Ryanf55:
@khancyr  We're one step closer to getting pre-commit easier to use with regular workflow. The next step is to add support for this in our pre-commit file. Just FYI, if you wanted to help out.

PR #28565, opened 2024-11-09
Use RC_Channel to populate IOMCU mappings

peterbarker:
Bench-tested using an OrangeRX and a CubeOrangePlus and a servo.  All seems right compared to master.

```
Board,plane
CubeOrangePlus,-32
```

### DIFF
--- a/ArduPlane/ArduPlane.cpp
+++ b/ArduPlane/ArduPlane.cpp
@@ -329,7 +329,7 @@ void Plane::one_second_loop()
     set_control_channels();
 
 #if HAL_WITH_IO_MCU
-    iomcu.setup_mixing(&rcmap, g.override_channel.get(), g.mixing_gain, g2.manual_rc_mask);
+    iomcu.setup_mixing(g.override_channel.get(), g.mixing_gain, g2.manual_rc_mask);
 #endif
 
 #if HAL_ADSB_ENABLED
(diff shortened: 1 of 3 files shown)


amilcarlucas:
How close are we to remove RC_MAP ?

peterbarker:
> How close are we to remove RC_MAP ?

I think my `die-rcmap-die` branch shows most of that.  Parameter conversion, scripting binding changes and simulink changes.

The other big thing is GCS adjustment.  QGC is going to require some surgery to understand `RCn_OPTION` 201 means roll, that sort of thing, rather than looking at `RCMAP_ROLL`, likewise with MissionPlanner.  I'd imagine several closed-source GCSs would also need updating.  These GCSs will need to understand both mechanisms for several years, of course.

I'm wondering about some interim arrangement whereby we magically update the `RCMAP` parameters from the `RCn_OPTION` parameters and vice-versa, throwing a warning when we update `RCn_OPTION` from `RCMAP`.

IamPete1:
> I'm wondering about some interim arrangement whereby we magically update the `RCMAP` parameters from the `RCn_OPTION` parameters and vice-versa, throwing a warning when we update `RCn_OPTION` from `RCMAP`.

We could make RCMAP read only, so legacy GCSs can still show the correct thing on there RC calibration screen. I'm not sure if any try and change them, Mission Planner doesn't as far as I know.

IamPete1 (approved):
LGTM

tridge (approved):

PR #28568, opened 2024-11-09
QuadPlane: Fix QLand getting stuck in pilot repositioning

IamPete1:
This fixes a bug where Qland will never land if it is entered when pilot repositioning is active.

Qland uses `landing_descent_rate_cms` which is called here: 

https://github.com/ArduPilot/ardupilot/blob/863b6222de8a3a2f8b9ffe44ab64e148a337d64e/ArduPlane/mode_qloiter.cpp#L146

This function checks `poscontrol.pilot_correction_active` and prevents descent if true here:

https://github.com/ArduPilot/ardupilot/blob/863b6222de8a3a2f8b9ffe44ab64e148a337d64e/ArduPlane/quadplane.cpp#L1279-L1282

However, `poscontrol.pilot_correction_active` is not reset on mode change, it is normally updated by `update_land_positioning` however this is never called in Qland because it does not support land repositioning.

The fix is to make sure the pilot correction is reset on entry to QLand.

### DIFF
--- a/ArduPlane/mode_qland.cpp
+++ b/ArduPlane/mode_qland.cpp
@@ -9,7 +9,6 @@ bool ModeQLand::_enter()
     quadplane.throttle_wait = false;
     quadplane.setup_target_position();
     poscontrol.set_state(QuadPlane::QPOS_LAND_DESCEND);
-    poscontrol.pilot_correction_done = false;
     quadplane.last_land_final_agl = plane.relative_ground_altitude(plane.g.rangefinder_landing);
     quadplane.landing_detect.lower_limit_start_ms = 0;
     quadplane.landing_detect.land_start_ms = 0;
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -4596,6 +4596,11 @@ void QuadPlane::mode_enter(void)
     poscontrol.last_velocity_match_ms = 0;
     poscontrol.set_state(QuadPlane::QPOS_NONE);
 
+    // Clear any pilot corrections
+    poscontrol.pilot_correction_done = false;
+    poscontrol.pilot_correction_active = false;
+    poscontrol.target_vel_cms.zero();
+
     // clear guided takeoff wait on any mode change, but remember the
     // state for special behaviour
     guided_wait_takeoff_on_mode_enter = guided_wait_takeoff;


ArduPlane/mode_qloiter.cpp
```diff
@@ -132,25 +132,10 @@ void ModeQLoiter::run()
                                                                   quadplane.get_desired_yaw_rate_cds());
 
     if (plane.control_mode == &plane.mode_qland) {
-        if (poscontrol.get_state() < QuadPlane::QPOS_LAND_FINAL && quadplane.check_land_final()) {
```
IamPete1, line 135: QLand is always in `QPOS_LAND_DESCEND`, this will never be true.

ArduPlane/mode_qloiter.cpp
```diff
@@ -132,25 +132,10 @@ void ModeQLoiter::run()
...
-            // cut IC engine if enabled
-            if (quadplane.land_icengine_cut != 0) {
-                plane.g2.ice_control.engine_control(0, 0, 0, false);
-            }
-#endif  // AP_ICENGINE_ENABLED
-        }
         float height_above_ground = plane.relative_ground_altitude(plane.g.rangefinder_landing);
         float descent_rate_cms = quadplane.landing_descent_rate_cms(height_above_ground);
-
-        if (poscontrol.get_state() == QuadPlane::QPOS_LAND_FINAL && !quadplane.option_is_set(QuadPlane::OPTION::DISABLE_GROUND_EFFECT_COMP)) {
```
IamPete1, line 148: QLand is always in QPOS_LAND_DESCEND, this will never be true.

ArduPlane/mode_qloiter.cpp
```diff
@@ -132,25 +132,10 @@ void ModeQLoiter::run()
...
-            ahrs.set_touchdown_expected(true);
-        }
-
         pos_control->land_at_climb_rate_cm(-descent_rate_cms, descent_rate_cms>0);
-        quadplane.check_land_complete();
```
IamPete1, line 153: `check_land_complete` returns if not in `QPOS_LAND_FINAL`, Qland never is.

https://github.com/ArduPilot/ardupilot/blob/863b6222de8a3a2f8b9ffe44ab64e148a337d64e/ArduPlane/quadplane.cpp#L3468-L3471

ArduPlane/mode_qland.cpp
```diff
@@ -9,7 +9,12 @@ bool ModeQLand::_enter()
     quadplane.throttle_wait = false;
     quadplane.setup_target_position();
     poscontrol.set_state(QuadPlane::QPOS_LAND_DESCEND);
+
+    // Clear any pilot corrections
```
IamPete1, line 13: There could be a argument to do this on every mode change, not just into QLand.

tridge: i think every mode change makes sense

IamPete1: Now on every mode change.

IamPete1:
Updated to clear reposition state on every mode change.

tridge (approved):

PR #26383, opened 2024-03-03
Copter: support `DO_RETURN_PATH_START` in AUTO RTL mode

IamPete1:
This adds support for a new mission item (https://github.com/ArduPilot/mavlink/pull/212, https://github.com/mavlink/mavlink/pull/2088) that acts as a marker to the autopilot that it can rejoin a mission at any point after this marker and before a landing or MAV_CMD_DO_LAND_START.

Demo: https://youtu.be/SoHXRTD0vEs?si=pHTvRDMcOsStv4l6

This was originally done on 4.1.0 where it has a lots of testing on real vehicles this rebase has not had the same level of testing.

Along with the new mission item this also adds support for passing in a location to the various `DO_LAND_...` AP mission functions. Copter then uses this to pass in the stopping point rather than the current location. Helpers are added to so the current method without location still works for plane. Passing in the stopping point prevents copter going back on itself when a rejoin is commanded as it is about to pass a way point at speed. Where the next waypoint is still in front of the vehicles current location but it can't stop in time to hit it and has to go back.

Both `DO_LAND_START` and `DO_RETURN_PATH_START` can be sent as commands to trigger there respective behavior and enter Auto RTL mode. Entering Auto RTL mode manually will first look for a `DO_RETURN_PATH_START` and if it cannot find one it will look for a `DO_LAND_START`. We might want to add a parameter like plane's `RTL_AUTOLAND` to control which we do when switching to Auto RTL mode.

I have added a auto test, but some manual testing is probably also a good idea.

PR #28571, opened 2024-11-10
ArduPlane: add failsafe auto variant which skips loops 

peterbarker:
Sometimes a mission is specifically crafted to avoid terrain and other obstacles.

In the event of a failsafe, it may be better for the vehicle to continue on the plotted course, but consider any loops to be satisfied.  Naturally this may lead to the vehicle flying further in the case of a battery failsafe.

This PR adds a new failsafe type for Plane which pokes the AP_Mission library to indicate all loops should be satisfied.

A similar effect can be achieved by creating a "shadow" mission containing waypoints identical to the original mission and adding many  `DO_LAND_START` mission items to ensure the correct next waypoint is chosen.

I'm not 100% convinced this is the correct way to approach things.  One obvious issue is that if an RTL is issues then we won't continue to follow the mission, rather jump straight to the `DO_LAND_START`.

The original work was done on top of Plane-4.5, which was missing @IamPete1 's [`DO_RETURN_PATH_START`](https://github.com/ArduPilot/ardupilot/pull/26383) code.  It might be that we would be better-off adding an option into param 1 of that mission item type which specifies that all loop counters should be considered satisfied?  Would also need to add support to Plane for `DO_RETURN_PATH_START`.

@MichelleRos

### DIFF
--- a/ArduPlane/defines.h
+++ b/ArduPlane/defines.h
@@ -45,6 +45,7 @@ enum failsafe_action_long {
     FS_ACTION_LONG_GLIDE = 2,
     FS_ACTION_LONG_PARACHUTE = 3,
     FS_ACTION_LONG_AUTO = 4,
+    FS_ACTION_LONG_AUTO_SATISFY_JUMP_REPEATS = 5,
 };
 
 // type of stick mixing enabled
--- a/ArduPlane/events.cpp
+++ b/ArduPlane/events.cpp
@@ -145,6 +145,10 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
             set_mode(mode_fbwa, reason);
         } else if (g.fs_action_long == FS_ACTION_LONG_AUTO) {
             set_mode(mode_auto, reason);
+        } else if (g.fs_action_long == FS_ACTION_LONG_AUTO_SATISFY_JUMP_REPEATS) {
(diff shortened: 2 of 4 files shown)


IamPete1:
I think this is a bit to specific. One problem is that I might want to jump to a `DO_LAND_START` or a `DO_RETURN_PATH_START` and skip jumps. All the combinations mean its not really practical as a fail safe action. 

`DO_RETURN_PATH_START` is a nice solution here, although as you mention you would have to plan a "shadow" mission. I would like to bring it to plane, that bit is not too hard, what I really wanted was to bring Auto RTL mode across too which is a larger task.

How about a new command the allows completing or resetting either all or a individual jump way-point. So you could set it when you jump to a particular way-point to restart all your jumps as if it were the first time through, or you could use it as you need to here to set them all as completed.

ArduPlane/defines.h
```diff
@@ -45,6 +45,7 @@ enum failsafe_action_long {
     FS_ACTION_LONG_GLIDE = 2,
     FS_ACTION_LONG_PARACHUTE = 3,
     FS_ACTION_LONG_AUTO = 4,
+    FS_ACTION_LONG_AUTO_SATISFY_JUMP_REPEATS = 5,
```
Hwurzburg, line 48: needs to be added to param metadata

Hwurzburg (changes requested):

PR #26751, opened 2024-04-10
hwdef: FPV/OSD boards only get one GPS

peterbarker:
notably cuts out blended GPS support

```
Board                    AP_Periph  blimp  bootloader  copter  heli   iofirmware  plane  rover  sub
CubeOrange-periph-heavy  *                                                                      
Durandal                            *      *           *       *                  *      *      *
Hitec-Airspeed           *                                                                      
KakuteH7-bdshot                     *      *           *       *                  *      *      *
MatekF405                           -1488  *           -1800   -1720              -1640  -1872  -1608
Pixhawk1-1M-bdshot                  *                  *       *                  *      *      *
f103-QiotekPeriph        *                                                                      
f303-Universal           *                                                                      
iomcu                                                                 *                         
revo-mini                           -1616  *           -1504   -1536              -1688  -1824  -1696
skyviper-v2450                                         *                                        
```

PR #28572, opened 2024-11-10
Add specific defines for sending of GPS mavlink messages

peterbarker:
This will allow us to remove these messages from the code eventually.

We can also adjust the defines (in a future PR) to remove the ability to send RTK mavlink messages based on which GPS backends are compiled in (only supported on 2 uncommon backends)

```
CubeOrange-periph-heavy  *                                                                     
Hitec-Airspeed           *                                                                     
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
```


peterbarker:
> do any need to be included in the build server?

Not at this time, I think.

As I mused above, we could potentially remove the RTK sending stuff in a future PR - and then we would need it in the custom build server.  Maybe.  I'm a bit surprised these aren't gated on moving baseline being compield in or similar, but that might mean the compiler is eliding anyway.

I wanted a no-compiler-output change :-)

libraries/AP_GPS/AP_GPS_config.h
```diff
@@ -104,3 +104,20 @@
 #ifndef HAL_GPS_COM_PORT_DEFAULT
 #define HAL_GPS_COM_PORT_DEFAULT 1
 #endif
+
+
+#ifndef AP_GPS_GPS_RAW_INT_SENDING_ENABLED
```
tridge, line 109: I don't think we want or need to be able to control these individually. Instead a single define AP_GPS_LEGACY_MESSAGE_SENDING_ENABLED would be a lot simpler

tridge (changes requested):

libraries/AP_GPS/AP_GPS_config.h
peterbarker: I've pushed up a patch which uses the new more-details defines to save 300-odd bytes on boards that don't have either of ERB or SBF:

```
CubeOrange-periph-heavy  *                                                                     
Hitec-Airspeed           *                                                                     
MatekF405                           -312   *           -312    -312              -312   -320   -312
Pixhawk1-1M-bdshot                  -312               -312    -312              -312   -304   -312
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           -312   *           -312    -312              -312   -312   -312
skyviper-v2450                                         -280                                    
```

peterbarker: .... also, if you only have a single GPS (https://github.com/ArduPilot/ardupilot/pull/26751) we can save a few hundred more bytes for a year or two.

PR #28573, opened 2024-11-10
AP_Scripting: script to disable subset of motors in fwd flight

tridge:
this script allows quadplanes to disable a subset of tilt motors in forward flight


IamPete1:
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/applets/forward_flight_motor_shutdown.md

PR #28584, opened 2024-11-11
AP_VideoTX: Fix _configuration_finished indication for Tramp VTX

menschel:
This patch fixes a minor bug in the Tramp VTX status propagation.

The flag `_configuration_finished` in class AP_VideoTX is not set by AP_Tramp class.
Therefore VTX Power blinks in OSD forever.


menschel:
I just tested this change on my setup [SpeedyBee F405 Wing Mini](https://www.speedybee.com/speedybee-f405-wing-mini-fixed-wing-flight-controller) + [HGLRC Zeus Nano VTX](https://www.hglrc.com/products/hglrc-zeus-nano-vtx-350mw-for-fpv-racing-drone)

VTX_PWR starts blinking when I change VTX Power via RC Switch and stops blinking one second later.
This is the expected behavior IMO.

menschel:
Rebased on Master.

andyp1per (approved):
Nice catch, thanks!

PR #28588, opened 2024-11-11
AP_AHRS: correct get_accel() to use primary accel rather than first usable for scripting

andyp1per:


libraries/AP_AHRS/AP_AHRS.h
```diff
@@ -620,7 +620,7 @@ class AP_AHRS {
 
     // return primary accels, for lua
```
IamPete1, line 621: We should probably remote the "for lua". This is used in loads of places now.

IamPete1: https://github.com/ArduPilot/ardupilot/blob/da69e2267392788b709eefa24395595a98301921/ArduPlane/pullup.cpp#L149

https://github.com/ArduPilot/ardupilot/blob/da69e2267392788b709eefa24395595a98301921/ArduPlane/servos.cpp#L643

https://github.com/ArduPilot/ardupilot/blob/da69e2267392788b709eefa24395595a98301921/libraries/AP_AHRS/AP_AHRS_Logging.cpp#L107

https://github.com/ArduPilot/ardupilot/blob/da69e2267392788b709eefa24395595a98301921/libraries/AP_IBus_Telem/AP_IBus_Telem.cpp#L599

https://github.com/ArduPilot/ardupilot/blob/da69e2267392788b709eefa24395595a98301921/libraries/AP_IBus_Telem/AP_IBus_Telem.cpp#L603

https://github.com/ArduPilot/ardupilot/blob/da69e2267392788b709eefa24395595a98301921/libraries/AP_IBus_Telem/AP_IBus_Telem.cpp#L607

andyp1per: Fixed thanks

IamPete1 (approved):
I think this is correct.

We might want to audit what is using ahrs vs ins at some point.

PR #28591, opened 2024-11-12
AP_Scripting: add Plane autoland applet

Hwurzburg:
Facilitates having an autoland FS mission setup for flying from fields without prior planning(or computers).....sets  up an autolanding sequence based on takeoff direction....

https://youtu.be/Y8Hn5h5h-3s?si=d7UuL5IL1rAcW1ZM

### DIFF
--- a/libraries/AP_Scripting/applets/UniversalAutoLand.lua
+++ b/libraries/AP_Scripting/applets/UniversalAutoLand.lua
@@ -1,0 +1,178 @@
+--[[ Upon Arming , creates a four item mission consisting of: NAV_TAKEOFF, DO_LAND_START,Final Approach WP opposite bearing from HOME of heading used during takeoff, at TKOFF_ALT or SCR_USER3 above home, SCR_USER2 or 2X TKOFF_DIST, and a LAND waypoint at HOME and stops until next disarm/boot. SCR_USER1 is used to enable or disable it.
+--]]
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+
+local PARAM_TABLE_KEY = 80
+local PARAM_TABLE_PREFIX = "AUTOLAND_"
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 4), 'could not add param table')
+
+-- add a parameter and bind it to a variable
+function bind_add_param(name, idx, default_value)
+    assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), string.format('could not add param %s', name))
+    return Parameter(PARAM_TABLE_PREFIX .. name)
+end
+
+--[[
+  // @Param: ENABLE
+  // @DisplayName: AUTOLAND ENABLE
+  // @Description: enable AUTOLAND script action
+  // @Values: 0:Disabled,1:Enabled
+  // @User: Standard
+--]]
+local AULND_ENABLE = bind_add_param('ENABLE', 1, 1)
+local enable = AULND_ENABLE:get()
+
+--[[
+  // @Param: WP_ALT
+  // @DisplayName: Final approach waypoint alt
+  // @Description: Altitude of final approach waypoint created by script
+  // @Range: 1 100
+  // @Units: m
+  // @User: Standard
+--]]
+local AULND_ALT = bind_add_param('WP_ALT', 2, 0)
+local final_wp_alt = AULND_ALT:get()
+--[[
+  // @Param: WP_DIST
+  // @DisplayName: Final approach waypoint distance
+  // @Description: Distance from landng point (HOME) to final approach waypoint created by script in the opposite direction of initial takeoff
+  // @Range: 0 1000
+  // @Units: m
+  // @User: Standard
+--]]
+local AULND_DIST = bind_add_param('WP_DIST', 3, 0)
+local final_wp_dist = AULND_DIST:get()
+
+FRAME_GLOBAL = 3
+NAV_WAYPOINT = 16
+NAV_TAKEOFF = 22
+NAV_LAND = 21
+DO_LAND_START = 189
+
+TAKEOFF_PITCH = 15
+
+local function wrap_360(angle)
+   local res = math.fmod(angle, 360.0)
+    if res < 0 then
+        res = res + 360.0
+    end
+    return res
+end
+
+local function wrap_180(angle)
+    local res = wrap_360(angle)
+    if res > 180 then
+       res = res - 360
+    end
+    return res
+end
+
+function create_final_approach_WP(i,bearing,dist,alt) --index,degs,m,m
+   local item = mavlink_mission_item_int_t()
+   local loc = ahrs:get_home()
+   loc:offset_bearing(bearing,dist) ---degs and meters
+
+   item:seq(i)
+   item:frame(FRAME_GLOBAL)
+   item:command(NAV_WAYPOINT)
+   item:param1(0)
+   item:param2(0)
+   item:param3(0)
+   item:param4(0)
+   item:x(loc:lat())
+   item:y(loc:lng())
+   item:z(alt)
+   return item
+end
+
+function create_takeoff_WP(alt)
+   local item = mavlink_mission_item_int_t()
+   local loc = ahrs:get_home()
+   
+   item:seq(1)
+   item:frame(FRAME_GLOBAL)
+   item:command(NAV_TAKEOFF)
+   item:param1(TAKEOFF_PITCH)
+   item:param2(0)
+   item:param3(0)
+   item:param4(0)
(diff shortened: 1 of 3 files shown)


libraries/AP_Scripting/applets/UniversalAutoLand.lua
```diff
@@ -0,0 +1,121 @@
...
+    return update,1000
+  end
+  
+  if not mission_loaded then --if first time after arm and enabled is valid then create  mission 
+    local home = ahrs:get_home()
+    local location = ahrs:get_location()
+    if location:alt() > (home:alt() + param:get("TKOFF_LVL_ALT")*100) then
+        local yaw = ahrs:get_yaw()
+        mission:set_item(2,create_do_land_start_WP())
+        mission:set_item(3,create_final_approach_WP(3,math.deg(yaw)+180,FIN_WP_DIST,param:get("TKOFF_ALT")))
```
timtuxworth, line 102: You might want to wrap the yaw after adding 180 (e.g. 359 degrees + 180 = 539), Tridges scripted aerobatics has some wrap functions that will do it.

Hwurzburg:
yeah, I thought about that but forgot....will wrap it

    On Monday, November 11, 2024 at 10:17:39 PM CST, Tim Tuxworth ***@***.***> wrote:  
 
 


@timtuxworth commented on this pull request.

In libraries/AP_Scripting/applets/UniversalAutoLand.lua:
> +   return item
+end
+
+function update()
+  if not arming:is_armed() then --if disarmed, wait until armed
+    mission_loaded = false
+    return update,1000
+  end
+  
+  if not mission_loaded then --if first time after arm and enabled is valid then create  mission 
+    local home = ahrs:get_home()
+    local location = ahrs:get_location()
+    if location:alt() > (home:alt() + param:get("TKOFF_LVL_ALT")*100) then
+        local yaw = ahrs:get_yaw()
+        mission:set_item(2,create_do_land_start_WP())
+        mission:set_item(3,create_final_approach_WP(3,math.deg(yaw)+180,FIN_WP_DIST,param:get("TKOFF_ALT")))

You might want to wrap the yaw after adding 180 (e.g. 359 degrees + 180 = 539), Tridges scripted aerobatics has some wrap functions that will do it.

libraries/AP_Scripting/applets/UniversalAutoLand.lua
Hwurzburg: had planned on it but forgot while debugging...done

Hwurzburg:
@timtuxworth yeah I lost the mission reset somewhere along the way...made a few tweaks also...felt I needed the user to consciously set the SCR_USER params instead of providing a default if they did not...

Rolf-G:
I have read the instructions 10 times, but have not yet understood how exactly the landing direction is determined ? (I don't mean in the LUA script, but actually in flight operations)

Hwurzburg:
Ahrs has a get yaw function that returns vehicle heading

libraries/AP_Scripting/applets/UniversalAutoLand.lua
```diff
@@ -0,0 +1,139 @@
+--[[ Upon Arming , creates a four item mission consisting of: NAV_TAKEOFF, DO_LAND_START,Final Approach WP opposite bearing from HOME of heading used during takeoff, at TKOFF_ALT or SCR_USER3 above home, SCR_USER2 or 2X TKOFF_DIST, and a LAND waypoint at HOME and stops until next disarm/boot. SCR_USER1 is used to enable or disable it.
+--]]
+
+--local variablees
+
+SCR_USER_EN = Parameter("SCR_USER1") -- enable this script
```
tridge, line 6: should create a param table for an applet I think, with param docs

libraries/AP_Scripting/applets/UniversalAutoLand.lua
```diff
@@ -0,0 +1,139 @@
...
+        local yaw = ahrs:get_yaw()
+        mission:set_item(3,create_final_approach_WP(3,wrap_180(math.deg(yaw)+180),FIN_WP_DIST,FIN_WP_ALT))
+        mission:set_item(4,create_land_WP())
+              mission_loaded = true
+    end
+  end
+  return update, 1000
+end 
+
+gcs:send_text(5,"Loaded TKOFF/RTL/Land Mission'.lua")
```
tridge, line 122: use MAV_SEVERITY table, as per other scripts, also typo with the ' ?

tridge (changes requested):

libraries/AP_Scripting/applets/UniversalAutoLand.lua
tridge: maybe should say UniversalAutoland ?

libraries/AP_Scripting/applets/UniversalAutoLand.lua
```diff
@@ -0,0 +1,139 @@
...
+  if not arming:is_armed() then --if disarmed, wait until armed
+    mission_loaded = false
+    return update,1000
+  end
+  
+  if not mission_loaded then --if first time after arm and enabled is valid then create  mission 
+    local home = ahrs:get_home()
+    local location = ahrs:get_location()
+    if location and home and location:alt() > (home:alt() + param:get("TKOFF_LVL_ALT")*100) then
+        local yaw = ahrs:get_yaw()
```
tridge, line 113: likely want ground course

tridge: actually, should be based on launch pt and pt when it has climbed up

Hwurzburg: I disagree...I want a 180 reciprocal waypoint from the initial takeoff heading, which in 99% of the cases should be into the wind or down the runway...so not ground course which could be heavily skewed by crosswinds...

libraries/AP_Scripting/applets/UniversalAutoLand.lua
Hwurzburg: will change

libraries/AP_Scripting/applets/UniversalAutoLand.lua
Hwurzburg: ground course on takeoffs with crosswind will set the final approach askew to the runway....

libraries/AP_Scripting/applets/UniversalAutoLand.lua
Hwurzburg: done

libraries/AP_Scripting/applets/UniversalAutoLand.lua
Hwurzburg: done

libraries/AP_Scripting/applets/UniversalAutoLand.lua
Hwurzburg: my point:
image

Hwurzburg: @tridge After our conversation, I have switched to setting bearing quickly using GPS ground course at .5 AIRSPEED_MIN and it seems to work well, acquiring bearing before rolling lift off and well before TKOFF_LVL_ALT, so should suffice for both compass and compass-less takeoffs...will flight test this week

Hwurzburg:
I test flew on compass-less plane with strong wind...both into the wind and at 90deg crosswind....in both cases the created final approach waypoint was place in line with throw direction....much better crosswind performance than waiting for LVL_ALT, since by then the yaw has weathervaned into the wind rather than being the initial throw/roll-out direction...

https://youtu.be/nd9rq2448Hw

tridge (changes requested):
I think an autotest that checks we end up close to home and pointing in the reverse direction (within a margin) of the takeoff direction would be worthwhile

Hwurzburg:
An autotest is gating a Lua applet?

tridge (approved):

libraries/AP_Scripting/applets/UniversalAutoLand.lua
```diff
@@ -0,0 +1,178 @@
...
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 4), 'could not add param table')
+
+-- add a parameter and bind it to a variable
+function bind_add_param(name, idx, default_value)
+    assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), string.format('could not add param %s', name))
+    return Parameter(PARAM_TABLE_PREFIX .. name)
+end
+
+--[[
+  // @Param: ENABLE
```
tridge, line 16: need full name of params in lua scripts, or you end up with just "ENABLE" as the param

tridge (changes requested):

libraries/AP_Scripting/applets/UniversalAutoLand.lua
Hwurzburg: done

tridge (approved):

PR #28599, opened 2024-11-12
AP_VideoTX / AP_OSD: Make VTX temperature available for OSD

menschel:
This patch adds VTX temperature forwarding from AP_Tramp through AP_VideoTX to AP_OSD.

I would appreciate feedback about style and correct use of methods because I merely copied code sections based on assumptions without digging too deep into the macros etc.

### DIFF
--- a/libraries/AP_OSD/AP_OSD_Screen.cpp
+++ b/libraries/AP_OSD/AP_OSD_Screen.cpp
@@ -1045,6 +1045,23 @@ const AP_Param::GroupInfo AP_OSD_Screen::var_info[] = {
 	AP_SUBGROUPINFO(rrpm, "RPM", 62, AP_OSD_Screen, AP_OSD_Setting),
 #endif
 
+
+    // @Param: VTX_TEMP_EN
+    // @DisplayName: VTX_TEMP_EN
+    // @Description: Displays the VTX Temperature. A Value of '-1' indicates that temperature is not available.
+    // @Values: 0:Disabled,1:Enabled
+
+    // @Param: VTX_TEMP_X
+    // @DisplayName: VTX_TEMP_X
+    // @Description: Horizontal position on screen for VTX_TEMP item
+    // @Range: 0 59
+
+    // @Param: VTX_TEMP_Y
+    // @DisplayName: VTX_TEMP_Y
+    // @Description: Vertical position on screen for VTX_TEMP item
+    // @Range: 0 21
+    AP_SUBGROUPINFO(vtx_temp, "VTX_TEMP", 63, AP_OSD_Screen, AP_OSD_Setting),
+
     AP_GROUPEND
 };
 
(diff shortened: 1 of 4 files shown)


menschel:
OK, I have tested and I can see the OSD element but the temperature keeps at 80°C the whole time.
Has this been tested on AP_Tramp, i.e. does the value saved there is accurate ?

libraries/AP_OSD/AP_OSD_Screen.cpp
```diff
@@ -927,6 +927,23 @@ const AP_Param::GroupInfo AP_OSD_Screen::var_info[] = {
...
+    // @Param: VTX_TEMP_X
+    // @DisplayName: VTX_TEMP_X
+    // @Description: Horizontal position on screen
+    // @Range: 0 59
+
+    // @Param: VTX_TEMP_Y
+    // @DisplayName: VTX_TEMP_Y
+    // @Description: Vertical position on screen
+    // @Range: 0 21
+    AP_SUBGROUPINFO(vtx_temp, "VTX_TEMP", 56, AP_OSD_Screen, AP_OSD_Setting),
```
andyp1per, line 946: You can't just change the index of an existing parameter. This needs to go in an empty slot and there are only 64 per group so you need to pick a different group

libraries/AP_VideoTX/AP_VideoTX.h
```diff
@@ -158,6 +158,11 @@ class AP_VideoTX {
     bool get_enabled() const { return _enabled; }
     bool update_enabled() const { return _defaults_set && _enabled != _current_enabled; }
 
+    // set current vtx temperature (called by backend)
+    void set_temperature(uint16_t temperature) { _current_temperature = temperature; }
```
andyp1per, line 162: you should record this in appropriate units and document, probably oC recorded as a float

libraries/AP_VideoTX/AP_VideoTX.h
```diff
@@ -204,6 +209,9 @@ class AP_VideoTX {
     AP_Int8 _enabled;
     bool _current_enabled;
 
+    // vtx temperature, not all backends provide it
+    uint16_t _current_temperature;
```
andyp1per, line 213: You need a way of showing this is unavailable (-1) and use it to disable the OSD element

andyp1per (changes requested):
Main problem is parameter changes

libraries/AP_OSD/AP_OSD_Screen.cpp
menschel: I went for index 63 which should be last in group.

libraries/AP_VideoTX/AP_VideoTX.h
menschel: changed to float

libraries/AP_VideoTX/AP_VideoTX.h
menschel: introduced -1.0f as invalid, it will be handled in the next step to show something like '----' .

menschel:
Added changes to the bottom of each location instead of mixing it inside other items.
Do I really need to open up a new group for my parameter? I assume Index 63 is allowed, yes?

menschel:
OK, I get it, now I have cross talk with TER_HGT. I'll try with a new Group.

menschel:
I'm getting the impression, that the referenced temperature in AP_Tramp is a thermal protection threshold and not a measurement.

menschel:
I'll see that I get a second Tramp VTX to solder into some betaflight setup for comparism.

https://betaflight.com/docs/wiki/guides/current/irc-tramp

Just checked that INAV does not have this feature via the stick OSD.
They threw it out in 2020
https://github.com/iNavFlight/inav/commit/91594314c6dbd6bd08446c191548bccaf7c1368b

Betaflight should still have it but they did not have a firmware image for the speedybee f405 wing mini
https://github.com/betaflight/betaflight/blob/master/src/main/cms/cms_menu_vtx_tramp.c#L89C5-L89C17

menschel:
This PR is void.

I received feedback from ImmersionRC / Orqa FPV that this value is a thermal setpoint, means it's stupid to view on OSD.

I'll do something else in another PR to write that setpoint and provide a setpoint value from ardupilot parameters.

PR #28602, opened 2024-11-13
AP_Compass: add IIS2MDC driver

dakejahl:
Adds a driver for the ST IIS2MDC magnetometer
https://www.st.com/resource/en/datasheet/iis2mdc.pdf


libraries/AP_Compass/AP_Compass_IIS2MDC.cpp
```diff
@@ -0,0 +1,181 @@
...
+ *
+ */
+#include "AP_Compass_IIS2MDC.h"
+
+#include <stdio.h>
+
+#include <AP_HAL/AP_HAL.h>
+#include <AP_Math/AP_Math.h>
+
+#if AP_COMPASS_IIS2MDC_ENABLED
```
peterbarker, line 23:
```suggestion
#include "AP_Compass_config.h"

#if AP_COMPASS_IIS2MDC_ENABLED

#include "AP_Compass_IIS2MDC.h"

#include <stdio.h>

#include <AP_HAL/AP_HAL.h>
#include <AP_Math/AP_Math.h>

```

libraries/AP_Compass/AP_Compass_IIS2MDC.h
```diff
@@ -0,0 +1,68 @@
...
+ * (at your option) any later version.
+ *
+ * This file is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
```
peterbarker, line 14: Please drop a link to the datasheet here

libraries/AP_Compass/AP_Compass_IIS2MDC.h
```diff
@@ -0,0 +1,68 @@
...
+ */
+#pragma once
+
+#include <AP_HAL/AP_HAL.h>
+#include <AP_HAL/I2CDevice.h>
+#include <AP_Math/AP_Math.h>
+
+#include "AP_Compass_config.h"
+
+#ifdef AP_COMPASS_IIS2MDC_ENABLED
```
peterbarker, line 24:
```suggestion
#include "AP_Compass_config.h"

#ifdef AP_COMPASS_IIS2MDC_ENABLED

#include <AP_HAL/AP_HAL.h>
#include <AP_HAL/I2CDevice.h>
#include <AP_Math/AP_Math.h>
```

libraries/AP_Compass/AP_Compass_IIS2MDC.cpp
```diff
@@ -0,0 +1,181 @@
...
+        enum Rotation rotation)
+    : _dev(std::move(dev))
+    , _rotation(rotation)
+    , _force_external(force_external)
+{
+}
+
+bool AP_Compass_IIS2MDC::init()
+{
+    _dev->get_semaphore()->take_blocking();
```
peterbarker, line 76:
```suggestion
    WITH_SEMAPHORE(_dev->get_semaphore());
```

... and remove the `goto fail` calls (and the label), just return false.

THis is a new pattern - we've worked out there's no problem holding the thread across the `register_periodic_callback` call.

libraries/AP_Compass/AP_Compass_IIS2MDC.cpp
```diff
@@ -0,0 +1,181 @@
...
+        return;
+    }
+
+    if (!_dev->read_registers(IIS2MDC_ADDR_OUTX_L_REG, (uint8_t *) &buffer, sizeof(buffer))) {
+        return;
+    }
+
+    int16_t x = int16_t((buffer.xout1 << 8) | buffer.xout0);
+    int16_t y = int16_t((buffer.yout1 << 8) | buffer.yout0);
+    int16_t z = -int16_t((buffer.zout1 << 8) | buffer.zout0);
```
peterbarker, line 169:
```suggestion
    const int16_t x = buffer.xout1 << 8) | buffer.xout0;
    const int16_t y = (buffer.yout1 << 8) | buffer.yout0;
    const int16_t z = ((buffer.zout1 << 8) | buffer.zout0);
```

libraries/AP_Compass/AP_Compass_IIS2MDC.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    Vector3f field = Vector3f{x * range_scale, y * range_scale, z * range_scale };
```
peterbarker, line 171:
```suggestion
    const Vector3f field{x * range_scale, y * range_scale, z * range_scale };
```

libraries/AP_Compass/AP_Compass_IIS2MDC.h
```diff
@@ -0,0 +1,68 @@
...
+
+    bool check_whoami();
+    void timer();
+    bool init();
+
+    AP_HAL::OwnPtr<AP_HAL::Device> _dev;
+
+    enum Rotation _rotation;
+    uint8_t _instance;
+    bool _force_external:1;
```
peterbarker, line 65:
```suggestion
    bool _force_external;
```

peterbarker (changes requested):

libraries/AP_Compass/AP_Compass_IIS2MDC.cpp
dakejahl: this can't be const because `accumulate_sample(field, _instance, 20)` takes a field reference and modifies it

peterbarker (changes requested):
Please fold the commits.

Please also outline the testing you've done.

dakejahl:
> Please fold the commits.

Done. Should I do a separate commit for Tools/? I assumed a single commit since this PR should be atomic.

> Please also outline the testing you've done.

Bench tested on the [ARK FPV](https://github.com/ArduPilot/ardupilot/pull/28603). I've uploaded the [log file to Drive](https://drive.google.com/file/d/1vqSb8G2AlR1bQXTrS2D6daDBdQa7rP-2/view?usp=sharing).

**QGC**
Screenshot from 2024-11-17 19-29-13
**PlotJuggler**
Screenshot from 2024-11-17 20-56-37

dakejahl:
@peterbarker can you let me know if there's anything else here you'd like to see?

dakejahl:
Is there anything else needed to get this merged? The ARK FPV board support depends on it.

I've also uploaded a flight log in position mode demonstrating functionality
https://github.com/ArduPilot/ardupilot/pull/28603#issuecomment-2568583875

libraries/AP_Compass/AP_Compass_IIS2MDC.h
peterbarker: This wasn't actually done and the conversation resolved?

peterbarker (changes requested):
LGTM apart from the minor include issue

libraries/AP_Compass/AP_Compass_IIS2MDC.h
dakejahl: I think I missed that it was an `#if` and not an `#ifdef`. I cleaned up the includes and fixed it

dakejahl:
BTW, this driver will also work for the LIS2MDL. Just discovered this while doing a betaflight port board. ST says that this is basically just a better version of the LIS2MDL

> LIS2MDL, IIS2MDC, LSM303AGR and LSM303AH are firmware and pin-to-pin compatible solutions

https://www.st.com/resource/en/design_tip/dt0131-digital-magnetometer-and-ecompass-efficient-design-tips--stmicroelectronics.pdf

andyp1per:
```
arducopter-heli  996 (+0.0546%)  0 (0.0000%)  4 (+0.0015%)   996 (+0.0545%)                                   137060
antennatracker   996 (+0.0739%)  0 (0.0000%)  -4 (-0.0015%)  996 (+0.0738%)                                   614956
arducopter       996 (+0.0546%)  0 (0.0000%)  -4 (-0.0015%)  996 (+0.0544%)                                   135436
ardurover        996 (+0.0596%)  0 (0.0000%)  4 (+0.0015%)   996 (+0.0595%)                                   292164
arduplane        996 (+0.0550%)  0 (0.0000%)  4 (+0.0015%)   996 (+0.0549%)                                   149348
blimp            996 (+0.0727%)  0 (0.0000%)  -4 (-0.0015%)  996 (+0.0726%)                                   592608
ardusub          996 (+0.0617%)  0 (0.0000%)  -4 (-0.0015%)  996 (+0.0616%)                                   348032
```
This adds 1k to all the boards that include all the compasses, but none of them are using it. I think we should consider having this off by default or disabling by default one of the other compass drivers that is now little used.

dakejahl:
@andyp1per does ardupilot have a mechanism to enable only certain drivers for a target? For example in PX4 uses Kconfig 
https://github.com/PX4/PX4-Autopilot/blob/main/boards/ark/fpv/default.px4board#L17

andyp1per:
> @andyp1per does ardupilot have a mechanism to enable only certain drivers for a target? For example in PX4 uses Kconfig https://github.com/PX4/PX4-Autopilot/blob/main/boards/ark/fpv/default.px4board#L17

Yes, just define the backends you need - https://github.com/ArduPilot/ardupilot/blob/609551212afc02f0c0b30fda165f90211eed23df/libraries/AP_HAL_ChibiOS/hwdef/skyviper-v2450/hwdef.dat#L109

dakejahl:
@andyp1per hmm it seems that I need to change both of these
```
Feature('Compass', 'IIS2MDC', 'AP_COMPASS_IIS2MDC_ENABLED', 'Enable IIS2MDC compasses', 0, None),
```
and
```
#ifndef AP_COMPASS_IIS2MDC_ENABLED
#define AP_COMPASS_IIS2MDC_ENABLED 0
#endif
```
which is a little strange since I would expect the latter to only be necessary if the former is missing. I noticed only changing the "Feature" to 0 (default disabled) still results in the .cpp being compiled

dakejahl:
Re-tested on hardware functional.

andyp1per:
Feature only affects custom build server.
File will be compiled regardless, but the lack of linkage means it will be elided

dakejahl:
Okay, should I change one of these back to 1? Or does this look good to go?

andyp1per:
Looks good to me

andyp1per (approved):

peterbarker:
Merged, thanks!

rmackay9:

PR #26876, opened 2024-04-23
AP_InertialSensor: Murata SCH16T IMU 

dakejahl:
Still need to validate on hardware. I can't tell if the IMU driver is running or not, is there anyway to quickly validate sensors on the bench? Using `hal.console->printf()` and `printf()` doesn't output anything on the debug uart. 

Any suggestions for the best way to validate a new IMU? 
- validate data
- validate loop rate 
- validate missing samples (failed CRC, missed DRDY)
- validate driver runs without restarting (no failed register config, no failed data reads)

You can see where I added printfs at. I'm taking off for the day but when I'm back I'll keep looking around.

PR #28603, opened 2024-11-13
AP_HAL_ChibiOS: hwdef: ARK_FPV board support

dakejahl:
Adds board support for the ARK FPV.
https://arkelectron.com/product/ark-fpv-flight-controller/

### Testing Checklist
From https://ardupilot.org/dev/docs/porting.html
- [x] the board boots and connects to a GCS via USB (Mission Planner, QGC, MAVProxy,etc.)
- [x] the GPS and Compass is recognized.
- [x] the RC is recognized and RC input follows the TX
- [x] the pitch and roll reports (ie HUD horizon) follows autopilot movement correctly (ie IMU is oriented properly and working). test each IMU individually using the INSx_USE parameters.
- [x] the board arms with default arming checks (may need to force arm if GPS is indoors)
- [x] attach a battery either directly if onboard power sensors or via external power module and make sure voltage is correctly displayed and current indication present
- [x] attach a test servo to each output after setting the output to a normal function like elevator, and exercise each one with TX while in MANUAL mode to check output functionality. Bdshot capable outputs should be tested with a BLHeli32 esc for passthrough mode communication (non-IOMCU outputs).
- [x] move the GPS to each UART output after setting all other UART protocols to NONE and the tested UART to GPS to assure the UART is functioning.
- [x] UARTs with CTS/RTS lines should use a telemetry radio with those connected to be sure they function with BRD_SERx__CTSRTS=1

From my own notes
- [x] FW upload on USB
- [x] FW upload on serial port (GPS)
- [x] Flash based params functional
- [x] Telem port test (mavlink)
- [x] GPS port test (mavlink)
- [x] RCIN port test (SBUS)
- [x] PWM 1-4 test (Bidir-DSHOT)
- [x] ESC Telem test 
- [x] Analog current measurement (ESC pin CURR_IN)
- [x] CAN port test (DroneCAN via MAVCAN over USB)
- [ ] VTX OSD port test (MSP Displayport)
- [x] PWM AUX (5 - 9) port test
    - **Servo9 doesn't go back to 0 when disabled, it stays at the disarmed PWM value**
- [ ] SPI port test (SCH16T IMU https://github.com/ArduPilot/ardupilot/pull/26876) 
- [x] 12V ADC measurement

### Requires
IIS2MDC compass driver PR
https://github.com/ArduPilot/ardupilot/pull/28602


dakejahl:
I want to reduce the clock speed to 320MHz to reduce heat but adding `MCU_CLOCKRATE_MHZ 320
` to the hwdef causes the build to fail. Any ideas?

AlexKlimaj:
Just built and flashed. Looks like it is working.

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef-bl.dat
```diff
@@ -0,0 +1,41 @@
...
+
+include common.inc
+
+# Bootloader is installed at zero offset
+FLASH_RESERVE_START_KB 0
+
+# Location where the bootloader will put the firmware
+FLASH_BOOTLOADER_LOAD_KB 128
+
+env OPTIMIZE -Os
```
andyp1per, line 11: This is the default

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef-bl.dat
```diff
@@ -0,0 +1,41 @@
...
+# 12V peripheral enable
+PG4 nVDD_5V_PERIPH_EN OUTPUT HIGH
+
+# LEDs
+PE3 LED_ACTIVITY OUTPUT OPENDRAIN HIGH # red
+PE5 LED_BOOTLOADER OUTPUT OPENDRAIN HIGH # blue
+define HAL_LED_ON 0
+
+# No param flash storage for bootloader
+define HAL_USE_EMPTY_STORAGE 1
```
andyp1per, line 40: no need for these in the bl

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,219 @@
...
+PB7 USART1_RX USART1
+define DEFAULT_SERIAL3_PROTOCOL SerialProtocol_GPS
+
+# Telem3 - VTX (DJI Air Unit, RX only)
+PA3 USART2_RX USART2
+PD5 USART2_TX USART2
+define DEFAULT_SERIAL4_PROTOCOL SerialProtocol_None # TODO: DJI HDL?
+
+# ESC telem RX only
+PH13 UART4_TX UART4
```
andyp1per, line 59: you can make this NODMA to save on DMA allocation

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,219 @@
...
+define HAL_BATT_VOLT_PIN 9
+define HAL_BATT_CURR_PIN 12
+define HAL_BATT_VOLT_SCALE 21
+define HAL_BATT_CURR_SCALE 120
+define HAL_BATT_MONITOR_DEFAULT 4
+
+# ESC Telem on UART4
+define HAL_SERIAL_ESC_COMM_ENABLED 1
+define HAL_WITH_ESC_TELEM 1
+define DEFAULT_SERIAL4_PROTOCOL SerialProtocol_ESCTelemetry
```
andyp1per, line 88: Don't think this is right - its on serial5

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,219 @@
...
+PA6  SPI6_MISO SPI6
+PG14 SPI6_MOSI SPI6
+PI10 EXT1_CS CS
+PD11 DRDY_ADIS16507 INPUT GPIO(93)
+
+# GPIO(93) for data ready on ADIS16507
+# define ADIS_DRDY_PIN 93
+
+# Motors, see STM32H743xx.py
+PI0  TIM5_CH4 TIM5 PWM(1) GPIO(50) BIDIR
```
andyp1per, line 108: remove BIDIR

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,219 @@
...
+PH12 TIM5_CH3 TIM5 PWM(2) GPIO(51) BIDIR
+PH11 TIM5_CH2 TIM5 PWM(3) GPIO(52) BIDIR
```
andyp1per, line 110: remove BIDIR

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,219 @@
...
+PH10 TIM5_CH1 TIM5 PWM(4) GPIO(53) BIDIR
+PI5  TIM8_CH1 TIM8 PWM(5) GPIO(54) BIDIR
+PI6  TIM8_CH2 TIM8 PWM(6) GPIO(55) BIDIR
```
andyp1per, line 113: remove bidir

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,219 @@
...
+PI7  TIM8_CH3 TIM8 PWM(7) GPIO(56) BIDIR
+PI2  TIM8_CH4 TIM8 PWM(8) GPIO(57) BIDIR
```
andyp1per, line 115: remove bidir

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,219 @@
...
+DMA_PRIORITY SDMMC* SPI*
+
+# Bidir DShot timers cannot share DMA
+DMA_NOSHARE SPI* TIM5* TIM8*
+
+# Enable FAT filesystem support (needs a microSD defined via SDMMC)
+define HAL_OS_FATFS_IO 1
+
+# Bootloader embedding / bootloader flashing not available
+define AP_BOOTLOADER_FLASHING_ENABLED 0
```
andyp1per, line 215: Given the capabilities of the board I am not sure this is a good idea

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
dakejahl: how did you determine the correct ones to select for BIDIR? I was reading through some forum posts where you explained it, I thought as long as they didn't share DMA it was okay. I'm also surprised the group is split across two timers. I recently worked on bidir dshot in PX4 so I would love to hear the technical reasons and understand how it works in ardupilot.

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
dakejahl: this came in from copy/paste from ARKV6X, which is what I started from. This embeds the bootloader in flash? Doesn't this use more memory? I'm not sure why it's disabled in ARKV6X

dakejahl: Just did a compare before and after, only uses 16KB :+1:

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,215 @@
...
+define HAL_SERIAL_ESC_COMM_ENABLED 1
+define HAL_WITH_ESC_TELEM 1
+
+# RC Input
+PC6 USART6_TX USART6
+PC7 USART6_RX USART6
+define DEFAULT_SERIAL6_PROTOCOL SerialProtocol_RCIN
+
+# TODO: build was failing
+# define DEFAULT_SERIAL7_PROTOCOL SerialProtocol_SLCAN
```
dakejahl, line 67: not sure if this needs to be here or not

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,215 @@
...
+
+# ADCs
+PA0 SCALED1_V3V3 ADC1 SCALE(2)
+PB1 VDD_5V_SENS ADC1 SCALE(2)
+# TODO: 12V monitor
+# PA4 SCALED2_V3V3 ADC1 SCALE(7.66)
```
dakejahl, line 73: Do non int values work here? Is this the right way to do it?
Screenshot from 2024-12-11 17-26-15

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,215 @@
...
+# Motors, see STM32H743xx.py
+PI0  TIM5_CH4 TIM5 PWM(1) GPIO(50)
+PH12 TIM5_CH3 TIM5 PWM(2) GPIO(51) BIDIR
+PH11 TIM5_CH2 TIM5 PWM(3) GPIO(52)
+PH10 TIM5_CH1 TIM5 PWM(4) GPIO(53) BIDIR
+PI5  TIM8_CH1 TIM8 PWM(5) GPIO(54) BIDIR
+PI6  TIM8_CH2 TIM8 PWM(6) GPIO(55)
+PI7  TIM8_CH3 TIM8 PWM(7) GPIO(56) BIDIR
+PI2  TIM8_CH4 TIM8 PWM(8) GPIO(57)
+PD12 TIM4_CH1 TIM4 PWM(9) GPIO(58)
```
dakejahl, line 112: All 8 channels are DMA capable. Is it possible to allow usage of all of them, but only have 4 at a time configured for use? How does the user figure out which channels they need to use?

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,215 @@
...
+PB9 I2C1_SDA I2C1
+PB8 I2C1_SCL I2C1
+
+# I2C2 (BMP390 baro, internal)
+PF1 I2C2_SCL I2C2
+PF0 I2C2_SDA I2C2
+
+# I2C4 (IIS2MDC compass, internal)
+PF14 I2C4_SCL I2C4
+PF15 I2C4_SDA I2C4
```
dakejahl, line 131: Both of these busses are "internal". I noticed most other boards only have 1 internal bus, is this okay to have 2 internal busses?  I assume so since the I2C_INTERNAL_MASK bitmask exists, but wanted to double check.

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,215 @@
...
+
+# IMU heater
+PB10 HEATER_EN OUTPUT LOW GPIO(80)
+define HAL_HEATER_GPIO_PIN 80
+define HAL_HAVE_IMU_HEATER 1
+define HAL_IMU_TEMP_DEFAULT 45
+define HAL_IMUHEAT_P_DEFAULT 50
+define HAL_IMUHEAT_I_DEFAULT 0.07
+# TODO: should we use this?
+# define HAL_IMU_TEMP_MARGIN_LOW_DEFAULT 5
```
dakejahl, line 145: Good practice to set an IMU temp low margin? Or just pre-arm annoyance?

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
```diff
@@ -0,0 +1,215 @@
...
+COMPASS IIS2MDC I2C:2:0x1E false ROTATION_NONE
+define HAL_PROBE_EXTERNAL_I2C_COMPASSES
+
+# IIM42653 IMU on SPI1
+SPIDEV iim42653  SPI1 DEVID1  IMU1_CS MODE3  2*MHZ  8*MHZ
+IMU Invensensev3 SPI:iim42653 ROTATION_YAW_270
+
+# SPI6 external bus
+# SPIDEV adis16507 SPI6 DEVID1  EXT1_CS MODE3  1*MHZ  2*MHZ
+# IMU ADIS1647x    SPI:adis16507 ROTATION_NONE ADIS_DRDY_PIN
```
dakejahl, line 197: Should I have this driver active by default? Along with any other IMU drivers we want to support on external SPI?

Tools/AP_Bootloader/board_types.txt
```diff
@@ -57,6 +58,7 @@ Reserved "PX4 [BL] FMU v6X.x"          53
 Reserved "PX4 [BL] FMU v6C.x"          56
 
 Reserved "ARK [BL] FMU v6X.x"          57
+Reserved "ARK [BL] FPV"                59
```
dakejahl, line 61: this causes CI to fail

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
dakejahl: should I set any new default params for bidir dshot? Or better to let the user configure these?

MOT_PWM_TYPE
SERVO_BLH_BDMASK
SERVO_BLH_OTYPE
SERVO_DSHOT_ESC
SERVO_DSHOT_RATE

andyp1per: Unless you know what ESC the user is going to use I would leave these alone

dakejahl: Just tested this and it works
```
PI0  TIM5_CH4 TIM5 PWM(1) GPIO(50) BIDIR
PH12 TIM5_CH3 TIM5 PWM(2) GPIO(51) BIDIR
PH11 TIM5_CH2 TIM5 PWM(3) GPIO(52) BIDIR
PH10 TIM5_CH1 TIM5 PWM(4) GPIO(53) BIDIR
PI5  TIM8_CH1 TIM8 PWM(5) GPIO(54)
PI6  TIM8_CH2 TIM8 PWM(6) GPIO(55)
PI7  TIM8_CH3 TIM8 PWM(7) GPIO(56)
PI2  TIM8_CH4 TIM8 PWM(8) GPIO(57)
PD12 TIM4_CH1 TIM4 PWM(9) GPIO(58)
```

dakejahl:
ESC Bidir DShot passthrough works using BIDIR on Channels 1 - 4
image

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
andyp1per: We rotate the inbound DMA channel between CH1-2 and CH3-4, this is because the IC is organized in this way. I suspect that what yo uhave above will simply mean that the first channel in the IC group gets used. Removing one makes it clear which one is being used because sometimes it matters.

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
andyp1per: DMA is allocated at build time, but the actual runtime allocation of the channel only happens when you select a protocol needing DMA

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
andyp1per: yes

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
andyp1per: yes, otherwise it won't probe for these

Tools/AP_Bootloader/board_types.txt
andyp1per: yeah seems like a pretty low id

Hwurzburg:
@dakejahl readme will need images of board with connectors labeled

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
dakejahl: > We rotate the inbound DMA channel between CH1-2 and CH3-4

Are you saying that the dshot RX frame is being read round robin? I am looking at the hwdef.dat and this is how it appears since they all are configured use the same DMA stream when I group them all on a single timer. So the eRPM read rate would be 1/4th.
```
#define STM32_TIM_TIM5_CH1_DMA_STREAM  STM32_DMA_STREAM_ID(2, 2)
#define STM32_TIM_TIM5_CH2_DMA_STREAM  STM32_DMA_STREAM_ID(2, 2)
#define STM32_TIM_TIM5_CH3_DMA_STREAM  STM32_DMA_STREAM_ID(2, 2)
#define STM32_TIM_TIM5_CH4_DMA_STREAM  STM32_DMA_STREAM_ID(2, 2)
```

Tools/AP_Bootloader/board_types.txt
AlexKlimaj: The ID is already set in PX4 and QGC. Why would the board ID make the CI fail?

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
andyp1per: For each dshot pulse that goes out using the timer _UP channel only one is read on the return. So yes in this case the rate is 1/4 of the output rate. This is the main reason that the SERVO_DSHOT_RATE parameter exists.

Tools/AP_Bootloader/board_types.txt
andyp1per: IDK, one for @peterbarker

dakejahl: looks like bl/fmu don't share IDs
https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/validate_board_list.py#L55-L66

dakejahl:
> @dakejahl readme will need images of board with connectors labeled

@Hwurzburg I've added top/bottom images of the silkscreen which has the pin 1 indicator next to each connector
Screenshot from 2024-12-13 11-11-33

Hwurzburg:
@dakejahl ping me when testing is complete and I will reveiw

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/hwdef.dat
dakejahl: yes this works

dakejahl:
@Hwurzburg testing is complete. I've uploaded a log of a flight test here.
https://drive.google.com/file/d/1FsKDafLm5o2UE7yZUbidiRbH3OhbslD0/view?usp=sharing

Hwurzburg (changes requested):
you should consider adding SD card to bootloader for SD card firmware updating

libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/README.md
```diff
@@ -0,0 +1,184 @@
...
+| 6   | GND                  | GND     |
+
+#### VTX - 6 Pin JST-GH
+| Pin | Signal Name           | Voltage |
+|-----|-----------------------|---------|
+| 1   | 12.0V                | 12.0V   |
+| 2   | GND                  | GND     |
+| 3   | UART5_TX_TELEM2_EXT  | 3.3V    |
+| 4   | UART5_RX_TELEM2_EXT  | 3.3V    |
+| 5   | USART2_RX_TELEM3_EXT | 3.3V    |
```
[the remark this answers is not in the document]
dakejahl: I see what you mean. It does match the schematic though. @AlexKlimaj 
image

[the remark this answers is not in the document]
Hwurzburg: then the board is incorrect and will not function with DJI remote controllers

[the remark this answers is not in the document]
AlexKlimaj: The connectors are not the same, jst gh instead of jst sh. So an off the shelf cable can't be used anyway. The board is not incorrect.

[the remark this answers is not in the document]
Hwurzburg: then at least user inconvenient requiring cross over in cabling, and confusing...needs this note after VTX header:
"Note: connector pinout not in same order as standard HD VTX calling"

Hwurzburg (changes requested):

Tools/AP_Bootloader/board_types.txt
peterbarker: > looks like bl/fmu don't share IDs https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/validate_board_list.py#L55-L66

I note that you've fixed the CI test by adding an exception into our validation script, and that's fine, all good, happy to merge that.


OTOH, why are you adding two lines to this file?  What does `Reserved "ARK [BL] FPV"` actually achieve?  Perhaps there's a process which I'm unaware of, but I'd guess these "Reserved" line should possibly have been removed once the "TARGET_" lines were added, but at some stage one wasn't and now it's just a pattern people are following....

AlexKlimaj: The STM32H7 using the in tree bootloader in PX4. I forgot about the separate bootloader repo with the IDs. 

https://github.com/PX4/PX4-Autopilot/blob/main/boards/ark/fpv/nuttx-config/nsh/defconfig#L79
https://github.com/PX4/PX4-Autopilot/blob/main/boards/ark/fpv/firmware.prototype#L2

dakejahl: I opened a PR to have it reserved
https://github.com/PX4/PX4-Bootloader/pull/242

> OTOH, why are you adding two lines to this file? What does Reserved "ARK [BL] FPV" actually achieve? Perhaps there's a process which I'm unaware of, but I'd guess these "Reserved" line should possibly have been removed once the "TARGET_" lines were added, but at some stage one wasn't and now it's just a pattern people are following....

That's probably right, large projects with a lot of history... just follow the patterns...

peterbarker: This issue has been resolved,thanks!

Hwurzburg:
@dakejahl I have made a few tweaks to the readme and squashed and library split into the form we need for merging...marked for devcall....still needs @andyp1per to give approval.

Hwurzburg (approved):

dakejahl:
@Hwurzburg Thanks for that. Appreciate all of the help so far

dakejahl:
IIS2MDC is merged. This is ready for merge!

andyp1per (approved):

PR #28604, opened 2024-11-13
draft: Fix MAV_CMD_DO_JUMP skipping to end of mission when Jump points to itself

joshanne:
In the case where a unlimited jump is used, if the jump item points to itself, it will result in the AV skipping the rest of the mission and triggering RTL.

This handles JUMPs that point to themselves by skipping them.

First commit shows the issue occuring, second commit shows the fix to the issue.


libraries/AP_Mission/AP_Mission.cpp
```diff
@@ -2175,6 +2175,14 @@ bool AP_Mission::get_next_cmd(uint16_t start_index, Mission_Command& cmd, bool i
                 return false;
             }
 
+            // handle jump item that points to itself
+            if ((temp_cmd.content.jump.target == cmd_index)) {
+                // This jump points to itself - move on to next mission item
+                // To-Do: log an error? 
+                cmd_index++;
```
rmackay9, line 2182: Thanks for this.
We could perhaps use a SEND_TEXT to alert the user as to why we're skipping the jump.

rmackay9: I guess one thing that'd I'd check is that there's no change in behaviour if there are two jumps in a row.  the 1st one points to itself and the 2nd one points to the 1st one.

joshanne: > I guess one thing that'd I'd check is that there's no change in behaviour if there are two jumps in a row. the 1st one points to itself and the 2nd one points to the 1st one.

* If both waypoints are infinite jumps, it will skip both.
* If jump waypoint 1 is infinite and jump waypoint 2 is finite, then mission will result in "skip-then-jump" `n-jumps` (where `n-jumps` is jump waypoint 2 count).

peterbarker (changes requested):
Nice test.

libraries/AP_Mission/AP_Mission.cpp
joshanne: > I guess one thing that'd I'd check is that there's no change in behaviour if there are two jumps in a row. the 1st one points to itself and the 2nd one points to the 1st one.

To further clarify:

If there are two infinite jump loops in a row, both pointing to the first JUMP command in the sequence, it will still bail into RTL.

This is existing logic in ArduPilot that handles endless loops in this way.

The `max_loops` counter is decremented for each jump performed, once it reaches zero it bails and vehicle falls into RTL.

Tools/autotest/arduplane.py
```diff
@@ -6289,6 +6289,41 @@ def MAV_CMD_NAV_LOITER_TO_ALT(self):
         self.wait_current_waypoint(4)
         self.fly_home_land_and_disarm()
 
+    def MAV_CMD_DO_JUMP_to_self(self):
+        '''Test MAV_CMD_DO_JUMP to Jumping To Self and skipping rest of mission'''
+        offset = 500
+        alt = self.get_parameter("RTL_ALTITUDE")
+        waypoints = self.create_simple_relhome_mission([
```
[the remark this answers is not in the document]
joshanne: Updated.

Tools/autotest/arduplane.py
```diff
@@ -6289,6 +6289,41 @@ def MAV_CMD_NAV_LOITER_TO_ALT(self):
...
+            (mavutil.mavlink.MAV_CMD_NAV_WAYPOINT, offset, -offset, alt),
+            self.create_MISSION_ITEM_INT(
+                mavutil.mavlink.MAV_CMD_DO_JUMP,
+                p1=4,  # waypoint to jump to
+                p2=-1  # number of jumps (-1: infinite)
+            ),
+            (mavutil.mavlink.MAV_CMD_NAV_WAYPOINT, -offset, -offset, alt),
+            (mavutil.mavlink.MAV_CMD_NAV_WAYPOINT, -offset, offset, alt)
+        ])
+        self.check_mission_upload_download(waypoints)
```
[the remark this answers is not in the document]
joshanne: Updated.

libraries/AP_Mission/AP_Mission.cpp
joshanne: > Thanks for this. We could perhaps use a SEND_TEXT to alert the user as to why we're skipping the jump.

I've can modify to send a message if the get_next_cmd is called with `send_gcs_msg == true`:
```c++
if (send_gcs_msg) {
    GCS_SEND_TEXT(MAV_SEVERITY_NOTICE, "Mission: Waypoint %u jumps to itself", cmd_index);
}
```

But that message is coming out more than once. Seems like it might be for each of the calls through `get_next_cmd` is sending out.

joshanne: The question is whether we want these self pointing waypoints to be considered bad when decrementing the max_loop jump count or not.

[decrementing `max_loops`](https://github.com/ArduPilot/ardupilot/pull/28604/files#diff-1500d3a42eeb51ad59732e97da1aa22b07cf37676664f4b5a0e21ea2f80cadd2R2168) before this check essentially halves the amount of jumps in this two infinite-jump-waypoint case (of course those jumps need to include the self-pointing jump waypoint).

Now that may be fine, but I can't say my mission plans have "advanced" jumps in them...

rmackay9: as long as it's not getting stuck or crashing then I'm happy.  We probably should have a mission pre-arm check that checks for all kinds of potential issues including jump commands that jump to itself

rmackay9: If the send-text is spamming then I'm ok removing it

rmackay9 (approved):

rmackay9:
Thanks for this, the change means that jumps to itself will be treated the same as jumps that have reached their max number so it should be safe and correct

rmackay9:
As discussed on the dev call we think the behaviour should be consistent so a new time should be added around line 2173.

Ideally we should catch these issues as part of an arming check but that's too much scope creep for this PR.

magicrub:
This should have the same behavior as the line above it - return false;
- I suggest adding it to the above invalid check. Jumping to itself is also an invalid index so treat it the same.
- Add in a GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Mission: Invalid Jump from %u to %u", cmd.index, temp_cmd.content.jump.target);

PR #28607, opened 2024-11-13
AP_DDS: status topic to report RC failsafe with callback function

tizianofiorenzani:
# What Changed

Follow Up to https://github.com/ArduPilot/ardupilot/pull/28337.

The Status topic is now populated with the function `in_rc_failsafe()`, which is implemented as virtual in the `RC_Channel` class.

# Test

Tested in SITL: set the parameter `SIM_RC_FAIL` to 1 and then 0. 
You should see the following texts on the GCS:

image

And the following result when the status topic is echoed with `ros2 topic echo /ap/status`

image


peterbarker (approved):
LGTM

PR #28608, opened 2024-11-13
AP_Mission: `DO_RETURN_PATH_START` limit worst case runtime

IamPete1:
If you fill your mission with `DO_RETURN_PATH_START` way-points the best rejoin location calculation will take so long that the flight controller watchdogs.

This adds a worst case limit of 1000 search points. The limit used to be 254 points per `DO_RETURN_PATH_START`, the the new limit is equivalent to four maximum length return path options. This change allows a single path to be longer. 

https://github.com/ArduPilot/ardupilot/pull/28574

### DIFF
--- a/libraries/AP_Mission/AP_Mission.cpp
+++ b/libraries/AP_Mission/AP_Mission.cpp
@@ -2446,15 +2446,21 @@ bool AP_Mission::jump_to_closest_mission_leg(const Location &current_loc)
     uint16_t landing_start_index = 0;
     float min_distance = -1;
 
+    // This defines the maximum number of waypoints that will be searched, this limits the worst case runtime
+    uint16_t search_remaining = 1000;
+
     // Go through mission and check each DO_RETURN_PATH_START
     for (uint16_t i = 1; i < num_commands(); i++) {
-        Mission_Command tmp;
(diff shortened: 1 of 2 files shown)


libraries/AP_Mission/AP_Mission.cpp
```diff
@@ -2446,16 +2446,22 @@ bool AP_Mission::jump_to_closest_mission_leg(const Location &current_loc)
...
     float min_distance = -1;
 
+    // This defines the maximum number of waypoints that will be searched, the limits the worst case runtime
+    uint16_t search_remaining = 1000;
+
     // Go through mission and check each DO_RETURN_PATH_START
     for (uint16_t i = 1; i < num_commands(); i++) {
-        Mission_Command tmp;
-        if (read_cmd_from_storage(i, tmp) && (tmp.id == MAV_CMD_DO_RETURN_PATH_START)) {
+        if (get_command_id(i) == uint16_t(MAV_CMD_DO_RETURN_PATH_START)) {
```
IamPete1, line 2454: This is a efficiency saving, we were only checking the ID, so we don't need the full command.

tridge (approved):

PR #28609, opened 2024-11-13
Dds arm through external control

Ryanf55:
# Purpose

Make arming from DDS go through external control. That way, we can prevent a companion computer from arming using https://github.com/ArduPilot/ardupilot/pull/28429

We want all services that control the state of the vehicle to go through external control so we can gate them in a single place.

# Size (Durandal)

```
arducopter-heli  72 (+0.0040%)  0 (0.0000%)  0 (0.0000%)  72 (+0.0039%)                                    141248
antennatracker   0 (0.0000%)    0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      613396
arducopter       72 (+0.0040%)  0 (0.0000%)  0 (0.0000%)  72 (+0.0039%)                                    140792
ardurover        72 (+0.0043%)  0 (0.0000%)  0 (0.0000%)  72 (+0.0043%)                                    [29](https://github.com/ArduPilot/ardupilot/actions/runs/11826461468/job/32952516727?pr=28609#step:10:30)4832
arduplane        72 (+0.0040%)  0 (0.0000%)  0 (0.0000%)  72 (+0.0040%)                                    143680
blimp            0 (0.0000%)    0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      591472
ardusub          0 (0.0000%)    0 (0.0000%)  0 (0.0000%)  0 (0.0000%)                                      350564
```

# Motivation

https://github.com/ArduPilot/ardupilot/pull/28429#discussion_r1829759857

### DIFF
--- a/libraries/AP_ExternalControl/AP_ExternalControl.h
+++ b/libraries/AP_ExternalControl/AP_ExternalControl.h
@@ -8,6 +8,7 @@
 
 #if AP_EXTERNAL_CONTROL_ENABLED
 
+#include <AP_Arming/AP_Arming.h>
 #include <AP_Common/Location.h>
 #include <AP_Math/AP_Math.h>
 
@@ -31,6 +32,16 @@ public:
     virtual bool set_global_position(const Location& loc) WARN_IF_UNUSED {
         return false;
     }
+
+    /*
+        Arm the vehicle
+    */
+    virtual bool arm(AP_Arming::Method method, bool do_arming_checks) WARN_IF_UNUSED;
+
+    /*
+        Disarm the vehicle
+    */
+    virtual bool disarm(AP_Arming::Method method, bool do_disarm_checks) WARN_IF_UNUSED;
 
     static AP_ExternalControl *get_singleton(void) WARN_IF_UNUSED {
         return singleton;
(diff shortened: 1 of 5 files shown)


tizianofiorenzani (approved):
LGTM

libraries/AP_ExternalControl/AP_ExternalControl.h
```diff
@@ -32,6 +33,16 @@ class AP_ExternalControl
         return false;
     }
 
+    /*
+        Arm the vehicle
+    */
+    virtual bool arm(AP_Arming::Method method, bool do_arming_checks=true);
```
IamPete1, line 39: Should these be `WARN_IF_UNUSED`? Everything else is in this lib. 

Personally I don't like the default arguments, it can cause confusion. I would wither pass the bool all the way through or just hard code that external control must do the checks. But I'm happy to defer to @Ryanf55 on this.

Ryanf55: Done, I got distracted by the lack of const in `AP_Arming` and forgot to add it here. 

I removed the defaults and will pass the bools through. We should probably avoid them in use cases like this:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c140-do-not-provide-different-default-arguments-for-a-virtual-function-and-an-overrider

IamPete1 (approved):
LGTM

Ryanf55:
Hi @Hwurzburg , this PR has no change in user behavior. I don't think a wiki is necessary. Are you ok if I remove the label?

Hwurzburg:
np

PR #28613, opened 2024-11-14
AP_HAL_QURT: Fix SPI transfer special case

katzfey:
Fix the SPI transfer special case where a send buffer is passed in even though it is a read transaction. This happens in the IMU
driver now that it has been optimized. The send buffer is the same as the read buffer even though it's just doing a read. This is
perfectly fine by SPI standards but needs special case handling in the Qurt SPI driver.


katzfey:
Change was flight tested successfully.

PR #26413, opened 2024-03-05
Plane: correct RTL_AUTOLAND=1 behaviour

peterbarker:
we are calling "reached_loiter_target" as part of our checks as to whether to fly home or not.
    
    We need to call update_loiter so the L1 controller can update its internal state for the new waypoint which do_RTL has set.  Depending on location (but typically), that will mean that L1's reached_loiter_target() will then return false, so we fly home.
    
    This bug was affected by f8d7be5e43686e5a20a1a59420c308ac1fab4c70 .  Any sort of altitude error greater than 10m would delay us entering the landing sequence, allowing the L1 controller to update its state.

autotest added for same

PR #28614, opened 2024-11-14
Plane: Fix MAV_CMD_NAV_LOITER_TURNS with zero turns skipping waypoint location

joshanne:
In the case where MAV_CMD_NAV_LOITER_TURNS waypoint is used the AV can skip travelling to the waypoint and jump directly to the next waypoint (if the conditions are correct).

This ensures loiter turns cannot be verified as complete until both loiter target is reached and turn count is met (the latter half is unchanged).

First commit shows the issue occurring, and is an example use case where you might use 0-turn loiters.
Second commit shows the fix to the issue.

### DIFF
--- a/ArduPlane/commands_logic.cpp
+++ b/ArduPlane/commands_logic.cpp
@@ -726,8 +726,10 @@ bool Plane::verify_loiter_turns(const AP_Mission::Mission_Command &cmd)
     // LOITER_TURNS makes no sense as VTOL
     auto_state.vtol_loiter = false;
 
-    if (condition_value != 0) {
-        // primary goal, loiter time
+    if (!reached_loiter_target()) {
+        result = false;
+    } else if (condition_value != 0) {
+        // primary goal, loiter turns
         if (loiter.sum_cd > loiter.total_cd && loiter.sum_cd > 1) {
             // primary goal completed, initialize secondary heading goal
             condition_value = 0;
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -6317,6 +6317,65 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
         self.wait_current_waypoint(4)
         self.fly_home_land_and_disarm()
 
+    def MAV_CMD_NAV_LOITER_TURNS_zero_turn(self):
+        '''Ensure air vehicle achieves loiter target before exiting'''
+        offset = 500
+        alt = self.get_parameter("RTL_ALTITUDE")
+        waypoint_radius = 100
+
+        loiter_turns_loc_ccw = self.home_relative_loc_ne(offset, offset)
+        loiter_turns_loc_cw = self.home_relative_loc_ne(offset, -offset)
+
+        # upload a mission plan containing zero-turn loiters
+        self.upload_simple_relhome_mission([
+            (mavutil.mavlink.MAV_CMD_NAV_TAKEOFF, 0, 0, 10),
+            self.create_MISSION_ITEM_INT(
+                mavutil.mavlink.MAV_CMD_NAV_LOITER_TURNS,
+                p1=0,                # Turns
+                p3=-waypoint_radius, # Radius (If positive loiter clockwise, else counter-clockwise)
+                x=int(loiter_turns_loc_ccw.lat*1e7),
+                y=int(loiter_turns_loc_ccw.lng*1e7),
+                z=alt,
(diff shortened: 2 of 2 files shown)


joshanne:
@peterbarker @tridge I suspect a similar issue of short circuiting a waypoint can occur with `LOITER_TO_ALT`, given it is not checking for reaching of the loiter target either.

joshanne:
> @peterbarker @tridge I suspect a similar issue of short circuiting a waypoint can occur with `LOITER_TO_ALT`, given it is not checking for reaching of the loiter target either.

Although, i haven't been able to prove this through a separate autotest.

ArduPlane/commands_logic.cpp
```diff
@@ -726,8 +726,10 @@ bool Plane::verify_loiter_turns(const AP_Mission::Mission_Command &cmd)
     // LOITER_TURNS makes no sense as VTOL
     auto_state.vtol_loiter = false;
 
-    if (condition_value != 0) {
-        // primary goal, loiter time
+    if (!reached_loiter_target()) {
```
IamPete1, line 729: This can be a little tricky. If you set a new loiter target and your already loitering `reached_loiter_target` doesn't go false until the nav controller has run on the new target. It would be worth checking what happens with two consecutive loiter turns way points.

joshanne: This new autotest already contains two consecutive zero turn loiters. 

AP had a bug that was causing issues with the RTL AUTOLAND. When it was configured to fly home first. If you were already on a loiter, it would skip the fly home loiter because it had already deemed as reaching the loiter target. 

See: https://github.com/ArduPilot/ardupilot/pull/26413

I suspect this is the behaviour you're eluding to?

joshanne: I would need to check later today, but Im leaning towards that fix actually being the reason this fix works.

joshanne: Now I revisit that fix, it is explicitly for mode RTL.
I've modified the autotest to add in a guided orbit, and this fix works as expected too.

image

joshanne: The same function is called in `verify_loiter_time()` in much the same way.

I'll bring this to the dev call for broader discussion.

IamPete1: I'm convinced. 

BTW, you could early return false. 

```suggestion
    if (!reached_loiter_target()) {
        return false;
    }
```

IamPete1 (approved):
Change looks good and the consecutive loitering trap has been tested.

tridge:
@joshanne I'd like to look at this a bit more

robertlong13:
@tridge have you had a chance to look at this yet?

I think this PR is good. It's not addressing the underlying issue, but it's applying the same fix that other places in the code have employed and is the simplest correction with the lowest chance to hurt anything. And this is, I think, a pretty critical bug.

The root of the issue seems to stem from here. In `loiter_angle_update`, we specifically try to stop 
the angle sum from accumulating until we reach the loiter target, but between clearing the angle, and calling `update_loiter` (where the `reached_loiter_target` finally switches back to false), this update occurs, so this protection doesn't always work. https://github.com/ArduPilot/ardupilot/blob/b722ea8f2f88719f7c1ee92d5257f215fee0b68e/ArduPlane/navigation.cpp#L26-L36

This snippet in `loiter_angle_update` would work correctly if `loiter_angle_reset` also reset the flag in the navigation controller that tracks if we have reached the loiter target. I made a quick patch that added a function to clear the `_WPcircle` flag and that does work to fix this bug, but it's also less clear to me that it doesn't come with other unintended consequences.

tridge:
@joshanne @robertlong13 I've rebased this PR on master

tridge (approved):

joshanne:
No worries, what's your thoughts on it being an actual issue?

tridge:
@joshanne my apologies for forgetting about this one!

tridge:
> No worries, what's your thoughts on it being an actual issue?

yes, I think it is, we'll merge, thanks!

peterbarker:
Merged, thanks!

rmackay9:
I've pulled the code change into 4.6.2-beta1 but was unable to pull in the autotest change due to a merge conflict

rmackay9:
This is included in 4.6.2-beta1 via https://github.com/ArduPilot/ardupilot/pull/30466

PR #28615, opened 2024-11-14
Fix build for CubeNode ETH

bugobliterator:
* build was broken as a mandatory virtual method was added under GCS_MAVLink, without update AP_Periph/GCS_Mavlink
* also adds CubeNode-ETH to the CI to ensure that we have a periph firmware built with GCS_MAVLink and scripting enabled.


peterbarker (approved):
LGTM

peterbarker:
Note the commit message should be `Tools:` rather than `.github:` but I can't bring myself to require it :-)

PR #28618, opened 2024-11-14
AP_HAL_ChibiOS: correct logging of MAV_POWER_STATUS_CHANGED

peterbarker:
need to exclude the changed flag from the comparison!


IamPete1 (approved):
LGTM

PR #28621, opened 2024-11-14
AP_Rely: allow RELAYn_DEFAULT values on DroneCAN Periphs

magicrub:
When using DroneCAN Hardpoint on a vehicle to control GPIO/Relay pins on a periph, the periph's default pin param value is ignored. This enables that.


IamPete1 (approved):
Looks good.

PR #28623, opened 2024-11-15
Rename Location get-vector-from-origin methods

peterbarker:
Moves us closer to using metres in this interface

Several places `*= 0.01`the result of this method already.

```
CubeOrange-periph-heavy  *                                                                     
Durandal                            0      *           8       0                 8      8      8
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     0      *           8       8                 8      8      0
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
```

Bindings have changed for vehicles that run scripting.

Need compatibility code for scripts.  Somehow.

### DIFF
--- a/libraries/AP_Scripting/generator/description/bindings.desc
+++ b/libraries/AP_Scripting/generator/description/bindings.desc
@@ -14,8 +14,13 @@ userdata Location method get_distance float Location
 userdata Location method offset void float'skip_check float'skip_check
 userdata Location method offset_bearing void float'skip_check float'skip_check
 userdata Location method offset_bearing_and_pitch void float'skip_check float'skip_check float'skip_check
+userdata Location method get_vector_from_origin_NEU_m boolean Vector3f'Null
+userdata Location method get_vector_from_origin_NEU_m depends AP_AHRS_ENABLED
+userdata Location method get_vector_from_origin_NEU_cm boolean Vector3f'Null
+userdata Location method get_vector_from_origin_NEU_cm depends AP_AHRS_ENABLED
 userdata Location method get_vector_from_origin_NEU boolean Vector3f'Null
 userdata Location method get_vector_from_origin_NEU depends AP_AHRS_ENABLED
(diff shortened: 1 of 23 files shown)


libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -14,8 +14,8 @@ userdata Location method get_distance float Location
 userdata Location method offset void float'skip_check float'skip_check
 userdata Location method offset_bearing void float'skip_check float'skip_check
 userdata Location method offset_bearing_and_pitch void float'skip_check float'skip_check float'skip_check
-userdata Location method get_vector_from_origin_NEU boolean Vector3f'Null
-userdata Location method get_vector_from_origin_NEU depends AP_AHRS_ENABLED
+userdata Location method get_vector_from_origin_NEU_cm boolean Vector3f'Null
+userdata Location method get_vector_from_origin_NEU_cm depends AP_AHRS_ENABLED
```
IamPete1, line 18: You could rename this back to the original method for use in scripts. Currently this is a breaking change.

peterbarker: > You could rename this back to the original method for use in scripts. Currently this is a breaking change.

Can we create a clone (or shim?) with the old method name that scripts can use but get a deprecation warning?

At some stage in the (far-flung) future it would be nice to re-introduce the method but working in metres rather than centimetres.

IamPete1: > Can we create a clone (or shim?) with the old method name that scripts can use but get a deprecation warning?

Yeah, but the binding generator doesn't currently support two names for the same function, you could shim or do a manual binding.

peterbarker: > > Can we create a clone (or shim?) with the old method name that scripts can use but get a deprecation warning?
> 
> Yeah, but the binding generator doesn't currently support two names for the same function, you could shim or do a manual binding.

Thanks, will look at that.

IamPete1 (approved):
LGTM

rmackay9 (approved):

PR #28624, opened 2024-11-15
Fix for Qurt DSP to Apps Proc link message sequence checking

katzfey:
Make the receive sequence number checking on Qurt DSP per link to match what is done on the application processor side so that we don't get annoying sequence number mismatch warnings in the mini-dm output. This has no affect on functionality, just cleans up the warning messages.

### DIFF
--- a/libraries/AP_HAL_QURT/UARTDriver.cpp
+++ b/libraries/AP_HAL_QURT/UARTDriver.cpp
@@ -150,9 +150,19 @@ UARTDriver_MAVLinkUDP::UARTDriver_MAVLinkUDP(uint8_t instance) : inst(instance)
     register_mavlink_data_callback(instance, _mavlink_data_cb, (void *) this);
 }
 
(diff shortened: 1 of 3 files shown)


libraries/AP_HAL_QURT/UARTDriver.cpp
```diff
@@ -150,9 +150,19 @@ UARTDriver_MAVLinkUDP::UARTDriver_MAVLinkUDP(uint8_t instance) : inst(instance)
     register_mavlink_data_callback(instance, _mavlink_data_cb, (void *) this);
 }
 
+void UARTDriver_MAVLinkUDP::check_rx_seq(uint32_t seq)
+{
+	if (seq != rx_seq)
+	{
```
peterbarker, line 156:
```suggestion
	if (seq != rx_seq) {
```

peterbarker (approved):

PR #28625, opened 2024-11-15
AP_Scripting: quadplane terrain avoidance with "can't make that climb"

timtuxworth:
This is a Lua script that runs continually while a quadplane is in AUTO, GUIDED, RTL and QRTL mode (typically running a mission) and monitors LIDAR readings for dangerously close terrain. If the terrain gets too close, the script will take over the quadplane performing one of three avoidance maneouvers:

1. "Pitching" the plane will pitch up at the maximum rate/speed possible (subject to TECS limits)
2. "Quading" the plane will switch to QLOITER mode and use the VTOL motors to gain altitude as quickly as possible
3. "Can't make that Climb" - the plane will circle left or right up to a safe altitude above approaching terrain.

This has been extensively tested in the wilds of northern Canada and Alaska. There is an autotest which flies a challenging mission in Alaska.

Here's a video of a RealFlight demo which demonstrates each kind of avoidance https://youtu.be/wbSNbGx1tDM

Here's a video of a MAVProxy Cesium demo at "Top of the World" which is the same mission used in the autotest. https://youtu.be/531qnKnrR5k


timtuxworth:
Hi @IamPete1  and @peterbarker - I wonder if you could take a look at this. It's mostly a Lua script with some added bindings. 
Is there anyone else who could/should review it?

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3011,6 +3011,12 @@ function terrain:status() end
 ---@return boolean
 function terrain:enabled() end
 
+--  Calculate lookahead rise in terrain. This returns extra altitude needed to clear upcoming terrain in meters
+--- @param bearing number -- direction to look ahead 
+--- @param distance number -- distance to look ahead
+--- @param climb_ratio number - assumed rate of climb of the vehicle
+--- @return number -- the hightest extra altitude reired to clear the terrain
+function terrain:lookahead(bearing, distance, climb_ratio) end
```
IamPete1, line 3019: You don't seem to be using this new method. You already have a implementation in lua? My vote would be to save flash and not have the binding.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3043,7 +3043,6 @@ function terrain:status() end
 ---@return boolean
 function terrain:enabled() end
 
-
```
IamPete1, line 3046: Left over white space change.

libraries/AP_Scripting/modules/speedpi.lua
```diff
@@ -0,0 +1,155 @@
...
+    local _current = 0
+    local nowPI = millis():tofloat() * 0.001
+ 
+    -- update the controller.
+    function self.update(target, current)
+       local now = millis():tofloat() * 0.001
+       if not _last_t then
+          _last_t = now
+       end
+       local dt = now - _last_t
```
IamPete1, line 75: Do the DT first and then convert to float. That prevent lose of precision  at millis get larger.

libraries/AP_Scripting/modules/speedpi.lua
```diff
@@ -0,0 +1,155 @@
...
+    -- log the controller internals
+    function self.log(name, add_total)        
+       -- allow for an external addition to total
+       -- Targ = Current + error ( target airspeed )
+       -- Curr = Current airspeed input to the controller
+       -- P = calculated Proportional component
+       -- I = calculated Integral component
+       -- Total = calculated new Airspeed 
+       -- Add - passed in as 0 
+       logger.write(name,'Targ,Curr,P,I,Total,Add','ffffff',_target,_current,_P,_I,_total,add_total)
```
IamPete1, line 129:
```suggestion
       logger:write(name,'Targ,Curr,P,I,Total,Add','ffffff',_target,_current,_P,_I,_total,add_total)
```

libraries/AP_Scripting/modules/mavlink_cmdint.lua
```diff
@@ -0,0 +1,145 @@
...
+                               p3 = velocity,
+                               z = alt })
+    if mavlink_result > 0 then
+         gcs:send_text(MAVLink.SEVERITY.ERROR, SCRIPT_NAME_SHORT .. string.format(": MAVLink GUIDED_CHANGE_ALTITUDE returned %d", mavlink_result))
+         return false
+    end
+    return true
+end
+
+return MAVLink
```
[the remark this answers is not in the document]
timtuxworth: Is that a standard/recommended style? I like it like this if I don't _have_ to change it.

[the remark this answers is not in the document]
IamPete1: Yeah, standard to have a blank line at the end of the file. You will notice guthub puts a little stop sign there to warn you.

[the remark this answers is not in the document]
timtuxworth: Oh I thought you were talking about the blank line _before_ the return. No worries. I'll fix it.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1168 @@
...
+    local altitude_difference =  (highest_location:alt() * 0.01) + clearance - (current_location_amsl:alt() * 0.01)
+    if altitude_difference > 0 then
+        -- what is the pitch up requried to acheive that altitude?
+        local highest_distance = current_location_amsl:get_distance(highest_location)
+        pitch_required = math.deg(math.atan(altitude_difference,highest_distance))
+        -- the target location we need to hit needs to be AMSL to ensure we fly above terrain
+        alt_required_amsl = highest_location:alt() * 0.01
+        highest_alt_difference = altitude_difference
+    end
+
```
Ryanf55, line 886: What if you are currently in a dive? You ignore the time it takes to recover from the dive, pitch up, and how much altitude you lose to switch to the max pitch? I guess handling dynamic constraints could be added as a follow up.

libraries/AP_Scripting/modules/speedpi.lua
```diff
@@ -0,0 +1,156 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+   SpeedPI
+   A simple "PI" controller for airspeed. Copied from Andrew Tridgell's original
+   work on the ArduPilot Aerobatics Lua scripts.
```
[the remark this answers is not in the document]
timtuxworth: It could be. For now this just makes it available.

libraries/AP_Scripting/modules/mavlink_cmdint.lua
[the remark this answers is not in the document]
timtuxworth: No longer applies.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.md
```diff
@@ -0,0 +1,106 @@
...
+ This script will detect if a quadplane following an Auto mission is likely to hit elevated terrain, such as
+ a small hill, cliff edge, high trees or other obstacles that might not show up in the OOTB STRM terrain model.
+ The code will attempt to avoid the impact by:-
+ - "Pitching" up if the plane can safely fly over the obstacle 
+ - "Quading" up by switching to QuadPlane loiter mode (Quading) and gaining altitude using VTOL motors 
+ - "CMTC" - can't make that climb. If the approaching terrain is higher than the plane can climb based on the 
+    configured PTCH_LIM_MAX_DEG (divided by 2.0), then loiter to altitude before continuing.
+
+This code works best long range rangefinders such as the LightWare long range lidars that can measure 
+   distances up to 90-95 meters away.
```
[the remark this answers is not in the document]
timtuxworth: Yes we certainly could. Rangefinder is only required for the pitching/ quading options. It really doesn't do anything for CMTC, but everything in this script is integrated because the partner I'm working with wants the functions to cooperate. CMTC only would be possible by disabling the pitching and quading options by setting their respective DWN and FWD parameter values to zero.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.md
```diff
@@ -0,0 +1,106 @@
...
+ The terrain avoidance will be on by default but will not function at "home" or within ZTA_HOME_DIST meters
+ of home. The scripting function ZTA_ACT_FN can be used to disable terrain folling at any time
+ Terrain following will operate in modes Auto, Gukded, RTL and QRTL.
+
+The "Can't make that climb" (CMTC) feature will prevent ArduPlane from flying into terrain it does know about
+by calculating the required pitch to avoid terrain between the current location and the next waypoint including
+all points in between. If the pitch required is > PTCH_LIM_MAX_DEG / 2 then the code will perform a loiter to
+altitude, using fixed wing loiter, to acheive a safe altitude to avoid the terrain before continuing the mission. 
+
+Note: Q_ASSIST should also be configured for best results ZTA_PTCH_DWN_MIN > Q_ASSIST_ALT > ZTA_QUAD_DWN_MIN
```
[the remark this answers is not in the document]
timtuxworth: I think it would work by setting the ZTA_QUAD_DWN_MIN and ZTA_QUAD_FWD_MIN values to zero.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
timtuxworth: It would have to be. I currently assume level flight.

timtuxworth:
Made this a draft because it's pending testing from a partner.

timtuxworth:
The DO_CHANGE_SPEED in plane is failing @peterbarker - any idea why this might be happening? I'm pretty sure nothing in this PR should do that. Seems to be fixed now.

libraries/AP_Scripting/modules/speedpi.lua
IamPete1: Bump on this. Should be:

```
       local now = millis()
       if not _last_t then
          _last_t = now
       end
       local dt = (now - _last_t):tofloat() * 0.001

```

libraries/AP_Scripting/modules/speedpi.lua
```diff
@@ -0,0 +1,156 @@
...
+    local _min = min
+    local _max = max
+    local _last_t = nil
+    local _I = 0
+    local _P = 0
+    local _total = 0
+    local _counter = 0
+    local _target = 0
+    local _current = 0
+    local nowPI = millis():tofloat() * 0.001
```
IamPete1: This seems unused.

libraries/AP_Scripting/modules/speedpi.lua
```diff
@@ -0,0 +1,156 @@
...
+    -- the new instance. You can put public variables inside this self
+    -- declaration if you want to
+    local self = {}
+
+    -- private fields as locals
+    local _kP = kP or 0.0
+    local _kI = kI or 0.0
+    local _iMax = iMax
+    local _min = min
+    local _max = max
```
IamPete1: you actually don't need to shadow these, you can use the passed variables directly.

libraries/AP_Scripting/modules/speedpi.lua
```diff
@@ -0,0 +1,156 @@
...
+    -- update the controller.
+    function self.update(target, current)
+       local now = millis():tofloat() * 0.001
+       if not _last_t then
+          _last_t = now
+       end
+       local dt = now - _last_t
+       _last_t = now
+       local err = target - current
+       _counter = _counter + 1
```
IamPete1: Dosn't do anything?

libraries/AP_Scripting/modules/speedpi.lua
```diff
@@ -0,0 +1,156 @@
...
+ 
+    -- log the controller internals
+    function self.log(name, add_total)        
+       -- allow for an external addition to total
+       -- Targ = Current + error ( target airspeed )
+       -- Curr = Current airspeed input to the controller
+       -- P = calculated Proportional component
+       -- I = calculated Integral component
+       -- Total = calculated new Airspeed 
+       -- Add - passed in as 0
```
IamPete1: Might as well remove this field?

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+RANGEFINDER_ORIENT = {DOWNWARD = 25, FORWARD = 0}
+RANGEFINDER_STATUS = {NOTCONNECTED = 0, NODATA = 1, OUTOFRANGELOW = 2, OUTOFRANGEHIGH = 3, GOOD = 4}
+
+PARAM_TABLE_KEY = 106
+PARAM_TABLE_PREFIX = "TA_"
+
+-- bind a parameter to a variable
+function bind_param(name)
+    return Parameter(name)
+end
```
IamPete1: Might as well remove the helper.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+
+-- add a parameter and bind it to a variable
+function bind_add_param(name, idx, default_value)
+    assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), SCRIPT_NAME_SHORT .. string.format('could not add param %s', name))
+    return bind_param(PARAM_TABLE_PREFIX .. name)
+end
+
+-- setup follow mode specific parameters need 2wo tables because there are > 10 parameters
```
IamPete1: `need 2wo tables because there are > 10 parameters` ?

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+        quad_down_min = TA_QUAD_DWN_MIN:get()
+        quad_forward_min = TA_QUAD_FWD_MIN:get()
+
+        altitude_max = TA_ALT_MAX:get()
+
+        groundspeed_max = TA_GSP_MAX:get() or -1
+        groundspeed_airbrake_limit = TA_GSP_AIRBRAKE:get() or 0
+        airspeed_min = AIRSPEED_MIN:get() or 10
+        airspeed_cruise = AIRSPEED_CRUISE:get() or 18
+        airspeed_max = AIRSPEED_MAX:get() or 30
```
IamPete1: Why not just use them directly? You might cache them at the top of a single function, but I don't think you need to go this far.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+        gcs:send_text(MAV_SEVERITY.INFO, SCRIPT_NAME_SHORT .. ": away from home")
+    end
+    close_to_home = false
+
+    return true
+end
+
+-- main update function called by protected_wrapper REFRESH_RATE times per second
+function Update()
+    now = millis():tofloat() * 0.001
```
IamPete1: Same as the PI controller, you almost certainly don't want to do this.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+end
+
+-------------------------------------------------
+-- Mode functions
+-------------------------------------------------
+
+-------------------------------------------------
+-- wrap vehicle:set_mode to prevent accidentally setting invalid modes or manual mode
+-------------------------------------------------
+---@diagnostic disable-next-line
```
IamPete1: What is the error here?

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+function set_vehicle_mode(new_mode)
+    if new_mode <= 0 or vehicle_mode == new_mode then
+        return
+    end
+    vehicle:set_mode(new_mode)
+    vehicle_mode = new_mode
+end
+
+-- save the current mode and then enter the new mode
+---@diagnostic disable-next-line
```
IamPete1: What is the error?

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+    ---@diagnostic disable-next-line
+    Get_Roll_Function = ahrs.get_roll
+end
+function Get_Roll_Deg()
+    return math.deg(Get_Roll_Function(ahrs))
+end
+
+Get_Pitch_Function = ahrs.get_pitch_rad
+if Get_Pitch_Function == nil then
+    ---@diagnostic disable-next-line
```
IamPete1: Its better to only disable the exact error, but I don't know the error name off the top of my head, sorry.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+        return
+    end
+    vehicle:set_mode(new_mode)
+    vehicle_mode = new_mode
+end
+
+-- save the current mode and then enter the new mode
+---@diagnostic disable-next-line
+function set_avoid_mode(new_mode)
+    if vehicle:get_mode() ~= new_mode then
```
IamPete1, line 368: The vehicle call will already check if your already in a mode, so you could skip this to simplify the script.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+        avoid_enter_mode == FLIGHT_MODE.QHOVER or avoid_enter_mode == FLIGHT_MODE.QLOITER or avoid_enter_mode == FLIGHT_MODE.QRTL or
+        avoid_enter_mode == FLIGHT_MODE.RTL then
+       new_mode= avoid_enter_mode
+    end
+
+    set_vehicle_mode(new_mode)
+    avoid_enter_mode = -1
+
+    if new_mode == FLIGHT_MODE.AUTO then
+        PITCH_CHANNEL:set_override(PITCH_FORWARD_PWM)
```
IamPete1: Not a fan of using RC overrides. I think we concluded that this was the only way (for now), but I think it warrants a fairly big warning in the readme.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+    end
+    return airspeed_new
+end
+
+local function set_altitude_target(new_altitude)
+    current_location_target = vehicle:get_target_location()
+    if current_location_target ~= nil then
+        new_location_target = current_location_target:copy()
+        if new_location_target ~= nil and new_altitude ~= nil then
+            local target_alt_cm = math.floor(math.min(new_altitude * 100, 319000))
```
IamPete1: 319000?

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+
+    while search_distance > 0 do
+        search_location:offset_bearing(search_bearing, terrain_spacing)
+        climb = climb + search_ratio * terrain_spacing
+
+        height = terrain:height_amsl(search_location, true)
+        if height ~= nil then
+            local rise = (height - base_height) - climb
+            if rise > highest_rise then
+                local highest_alt_cm = math.floor(math.min(height * 100.0, 320000))
```
IamPete1: 320000?

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+    local direction
+    local loiter_center_left = current_location:copy()
+    local loiter_center_right = current_location:copy()
+    local radius_scaled_m = cmtc_rad_m * (1.0 + current_altitude_m/3000.0)
+    -- calculate a loiter location (center of the circle) 90 degrees left and 90 degrees right
+    -- note we are using the bearing to the current waypoint, not the current heading which might be different
+    -- the fudge factor on the radius is becuase AP scales he loiter radius internally based on altitude
+    -- gcs:send_text(MAV_SEVERITY.WARNING, SCRIPT_NAME_SHORT .. string.format(": CMTC hdg left %.0f right %.0f rad %.0f", wrap_360(current_wp_bearing_deg - 90), wrap_360(current_wp_bearing_deg + 90), radius_scaled_m) )
+    loiter_center_left:offset_bearing(wrap_360(current_wp_bearing_deg - 90), radius_scaled_m)
+    loiter_center_right:offset_bearing(wrap_360(current_wp_bearing_deg + 90), radius_scaled_m)
```
IamPete1, line 1073: You don't need to wrap here, the trig will come out the same.

IamPete1: But the wrap might make it easier to understand.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+    current_location = ahrs:get_location()
+    if current_location ~= nil then
+        current_altitude_m = current_location:alt() * 0.01
+    end
+    vehicle_mode = vehicle:get_mode()
+    current_location_target = vehicle:get_target_location()
+    current_wp_bearing_deg = vehicle:get_wp_bearing_deg() or 0
+    current_heading_deg = math.deg(get_yaw_function(ahrs) or 0)
+    groundspeed_vector = ahrs:groundspeed_vector()
+    groundspeed_current = ahrs:groundspeed_vector():length()
```
IamPete1:
```suggestion
    groundspeed_current = groundspeed_vector:length()
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1375 @@
...
+    return protected_wrapper, 1000 * REFRESH_RATE
+end
+
+function Delayed_Startup()
+    gcs:send_text(MAV_SEVERITY.INFO, string.format("%s %s script loaded", SCRIPT_NAME, SCRIPT_VERSION) )
+    return protected_wrapper()
+end
+
+-- wait a bit for AP to come up then start running update loop
+if FWVersion:type() == 3 and Q_ENABLE:get() == 1 and TERRAIN_ENABLE:get() == 1 then
```
IamPete1, line 1315: Where you only use a param once you could use `param:get('Q_ENABLE')` and save the local.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1168 @@
...
+    local spacing_degrees = (terrain_spacing * 360.0) / (2.0 * math.pi * wp_loiter_rad)
+
+    local highest_left_terrain = highest_arc_terrain(loiter_center_left, current_bearing, -spacing_degrees, 180)
+    local highest_right_terrain = highest_arc_terrain(loiter_center_right, current_bearing, spacing_degrees, 180)
+    local roll = ahrs:get_roll()
+
+    set_avoid_mode(FLIGHT_MODE.GUIDED)
+
+    -- loiter up to the requjired AMSL height, in the direction of lowest terrain
+    if highest_left_terrain < highest_right_terrain or roll < -30 then
```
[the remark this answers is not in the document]
timtuxworth: It's a bit arbitrary, just gut feel and it seems to work. Could be adjusted after further testing.

libraries/AP_Scripting/modules/speedpi.lua
timtuxworth: nowPI is Used at line 91 and 92? Or if you mean _counter - I've removed it. I've removed nowPI too.

libraries/AP_Scripting/modules/speedpi.lua
timtuxworth: No idea, I just copied from the aerobatics lua and didn't want to break anything. Do you think it needs to change, or just a "good idea"?

libraries/AP_Scripting/modules/speedpi.lua
timtuxworth: I see your point.

libraries/AP_Scripting/modules/speedpi.lua
timtuxworth: Done

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
timtuxworth: I've found if you try to send an altitude above about 3200 meters to vehicle:update_target_location() it breaks, so I put this limit in so I don't try to go that high.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
timtuxworth: It does for me, I'll leave it in.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
timtuxworth: I guess so, but it's not a local it's a GLOBAL, so is it a problem?

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
timtuxworth: Thanks

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
timtuxworth: I've found if you try to send an altitude above about 3200 meters to vehicle:update_target_location() it breaks, so I put this limit in so I don't try to go that high.

IamPete1: Breaks how?

libraries/AP_Scripting/modules/speedpi.lua
timtuxworth: Ok done (I think)

libraries/AP_Scripting/applets/quadplane_terrain_avoid.md
```diff
@@ -0,0 +1,106 @@
+# QuadPlane Terrain Avoidance
+
+ This script will detect if a quadplane following an Auto mission is likely to hit elevated terrain, such as
+ a small hill, cliff edge, high trees or other obstacles that might not show up in the OOTB STRM terrain model.
+ The code will attempt to avoid the impact by:-
+ - "Pitching" up if the plane can safely fly over the obstacle 
+ - "Quading" up by switching to QuadPlane loiter mode (Quading) and gaining altitude using VTOL motors 
+ - "CMTC" - can't make that climb. If the approaching terrain is higher than the plane can climb based on the 
+    configured PTCH_LIM_MAX_DEG (divided by 2.0), then loiter to altitude before continuing.
```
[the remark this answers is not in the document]
timtuxworth: Well for now I'm not using that.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
timtuxworth: I'm pretty sure it throws an error, I do forget, it's been a long time.

IamPete1: There are some places in the world where this constrain will cause a dive into terrain rather than avoiding it. I think we need to understand the error and either fix it or add a pre-arm from the script if its likely to trigger.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.md
```diff
@@ -0,0 +1,115 @@
...
+
+# Operation
+
+Good TECS tuning of your aircraft is essential. The script relies
+on TECS to do all the work. Some parameters it refers to directly.
+
+This script operates by default and can be turned off in flight with
+a switch, or will disable it self if close to home (TA_HOME_DIST meters).
+
+Install the script in the APM/scripts folder on the SD card on your autopilot. Install the speedpi.lua
```
IamPete1: Just a refence to the speedpi that might need removing.

pompecukor:
@timtuxworth how would this handle any false positive from Lidar? Is there an option for user to set to ignore lidar and only use terrain.
I have seen so many faults and false positives on lidar (clouds, dirt, even a bird), so I personally would never use it for such. Also there is not look ahead, the range is limited (90-100m) so not a good buffer so to speak.

timtuxworth:
> @timtuxworth how would this handle any false positive from Lidar? Is there an option for user to set to ignore lidar and only use terrain. I have seen so many faults and false positives on lidar (clouds, dirt, even a bird), so I personally would never use it for such. Also there is not look ahead, the range is limited (90-100m) so not a good buffer so to speak.

Terrain (in the places this has been tested) may not be correct. The use of the rangefinder is intended to save the plane when the real life terrain doesn't match the terrain data. 

You can take the risk if you fly where the terrain data is reliable, but what if it's not?

Yes, the rangefinder can sometimes be unreliable. This uses the AP_RangeFinder library and in particular RANGEFINDER_MIN, so if you are getting false positives, increase this. For example, I recently flew over an area with some small ponds, when flying over the pond at 80m altitude, the rangefinder would return values of 1-2 meters. I set RANGEFINDER_MIN to 5 to deal with this.

timtuxworth:
Responded to all the comments/suggestions (I think). The only failing tests are the flapping colcon and FRSky tests.

timtuxworth:
> @timtuxworth how would this handle any false positive from Lidar? Is there an option for user to set to ignore lidar and only use terrain. I have seen so many faults and false positives on lidar (clouds, dirt, even a bird), so I personally would never use it for such. Also there is not look ahead, the range is limited (90-100m) so not a good buffer so to speak.

The script allows for a forward facing LIDAR (probably pointing at about 45 degrees) to try to give some lookahead. There are different TA_FWD parameters for limits for a forward facing lidar (if installed).

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1331 @@
...
+local ptch_lim_max_deg = PTCH_LIM_MAX_DEG:get() or 25
+local tecs_climb_max = TECS_CLMB_MAX:get() or 5.0
+
+if cmtc_rad_m <= 0 then
+    cmtc_rad_m = wp_loiter_rad_m
+end
+
+THROTTLE_CHANNEL = rc:get_channel(Parameter('RCMAP_THROTTLE'):get()) -- The RC channel used for throttle
+THROTTLE_UP_PWM = 1800
+THROTTLE_HOVER_PWM = 1500
```
IamPete1: Better to pull this out from params. 1500 may not be hover on all vehicles. It would be the average of RC min and max.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1331 @@
...
+local terrain_spacing = TERRAIN_SPACING:get() or 100
+local ptch_lim_max_deg = PTCH_LIM_MAX_DEG:get() or 25
+local tecs_climb_max = TECS_CLMB_MAX:get() or 5.0
+
+if cmtc_rad_m <= 0 then
+    cmtc_rad_m = wp_loiter_rad_m
+end
+
+THROTTLE_CHANNEL = rc:get_channel(Parameter('RCMAP_THROTTLE'):get()) -- The RC channel used for throttle
+THROTTLE_UP_PWM = 1800
```
IamPete1: Calculate from endpoints.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1331 @@
...
+THROTTLE_HOVER_PWM = 1500
+PITCH_CHANNEL = rc:get_channel(Parameter('RCMAP_PITCH'):get())
+PITCH_FORWARD_PWM = 1350
+PITCH_NEUTRAL_PWM = 1500
```
IamPete1: Should be trim for the channel.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1331 @@
...
+
+if cmtc_rad_m <= 0 then
+    cmtc_rad_m = wp_loiter_rad_m
+end
+
+THROTTLE_CHANNEL = rc:get_channel(Parameter('RCMAP_THROTTLE'):get()) -- The RC channel used for throttle
+THROTTLE_UP_PWM = 1800
+THROTTLE_HOVER_PWM = 1500
+PITCH_CHANNEL = rc:get_channel(Parameter('RCMAP_PITCH'):get())
+PITCH_FORWARD_PWM = 1350
```
IamPete1: Calculate from params.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1331 @@
...
+
+local vehicle_mode = vehicle:get_mode()
+
+local current_location = ahrs:get_location()
+local current_wp_bearing_deg = 0.0
+local previous_location
+local current_altitude_m = 0.0
+local current_location_target
+local current_heading_deg = -1
+local new_location_target = Location()
```
IamPete1: Maybe leave this as nil rather than a unintalised location?

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1331 @@
...
+
+local location_tracker  -- forward declaration. See below for definition and instantiation.
+
+-------------------------------------------------
+-- deal with deprecations in 4.7
+-------------------------------------------------
+version47orhigher = true
+get_yaw_function = ahrs.get_yaw_rad
+if get_yaw_function == nil then
+    ---@diagnostic disable:deprecated
```
IamPete1: I think you still need the next line bit, currently this disables for the whole file I think.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1331 @@
...
+-------------------------------------------------
+-- Mode functions
+-------------------------------------------------
+
+-------------------------------------------------
+-- wrap vehicle:set_mode to prevent accidentally setting invalid modes or manual mode
+-- and also keep track of "prior" mode
+-------------------------------------------------
+---@diagnostic disable:lowercase-global
+function set_vehicle_mode(new_mode)
```
IamPete1: Why not just use upper case? Or make it a local?

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1331 @@
...
+        Q_WVANE_ENABLE:set(q_wvane_enable_save)
+    else
+        Q_WVANE_ENABLE:set(0)
+    end
+    q_wvane_enable_save = -1
+end
+
+local airbrake_trigger = false
+local airbrake_on = false
+local airbrake_triggered_now_ms = millis() --:tofloat() * 0.001
```
IamPete1:
```suggestion
local airbrake_triggered_now_ms = millis()
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+if cmtc_rad_m <= 0 then
+    cmtc_rad_m = wp_loiter_rad_m
+end
+
+THROTTLE_CHANNEL_NO = Parameter('RCMAP_THROTTLE'):get() or 3
+THROTTLE_CHANNEL = rc:get_channel(Parameter('RCMAP_THROTTLE'):get()) -- The RC channel used for throttle
+THROTTLE_CHANNEL_PREFIX = string.format("RC%.0f_", THROTTLE_CHANNEL_NO)
+THROTTLE_CHANNEL_MIN = Parameter(THROTTLE_CHANNEL_PREFIX .. "MIN"):get() or 1000
+THROTTLE_CHANNEL_MAX = Parameter(THROTTLE_CHANNEL_PREFIX .. "MAX"):get() or 1000
+THROTTLE_CHANNEL_TRIM = Parameter(THROTTLE_CHANNEL_PREFIX .. "TRIM"):get() or 1000
```
IamPete1: This is unused.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+local terrain_spacing = TERRAIN_SPACING:get() or 100
+local ptch_lim_max_deg = PTCH_LIM_MAX_DEG:get() or 25
+local tecs_climb_max = TECS_CLMB_MAX:get() or 5.0
+
+if cmtc_rad_m <= 0 then
+    cmtc_rad_m = wp_loiter_rad_m
+end
+
+THROTTLE_CHANNEL_NO = Parameter('RCMAP_THROTTLE'):get() or 3
+THROTTLE_CHANNEL = rc:get_channel(Parameter('RCMAP_THROTTLE'):get()) -- The RC channel used for throttle
```
IamPete1:
```suggestion
THROTTLE_CHANNEL = rc:get_channel(THROTTLE_CHANNEL_NO) -- The RC channel used for throttle
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+local cmtc_rad_m = TA_CMTC_RAD:get() or wp_loiter_rad_m
+local terrain_spacing = TERRAIN_SPACING:get() or 100
+local ptch_lim_max_deg = PTCH_LIM_MAX_DEG:get() or 25
+local tecs_climb_max = TECS_CLMB_MAX:get() or 5.0
+
+if cmtc_rad_m <= 0 then
+    cmtc_rad_m = wp_loiter_rad_m
+end
+
+THROTTLE_CHANNEL_NO = Parameter('RCMAP_THROTTLE'):get() or 3
```
IamPete1:
```suggestion
THROTTLE_CHANNEL_NO = Parameter('RCMAP_THROTTLE'):get()
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+PTCH_LIM_MAX_DEG = Parameter("PTCH_LIM_MAX_DEG")
+TECS_CLMB_MAX = Parameter("TECS_CLMB_MAX")
+local airspeed_min = AIRSPEED_MIN:get() or 10
+local airspeed_cruise = AIRSPEED_CRUISE:get() or 18
+local airspeed_max = AIRSPEED_MAX:get() or 30
+local wp_loiter_rad_m = WP_LOITER_RAD:get() or 150
+local cmtc_rad_m = TA_CMTC_RAD:get() or wp_loiter_rad_m
+local terrain_spacing = TERRAIN_SPACING:get() or 100
+local ptch_lim_max_deg = PTCH_LIM_MAX_DEG:get() or 25
+local tecs_climb_max = TECS_CLMB_MAX:get() or 5.0
```
IamPete1: Don't need the `or x` Parameters will error out if not found when init is done in the constructor.

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+    // @DisplayName: CMTC loiter radius
+    // @Description: Use this radius for the loiter when trying to gain altitude. If not set or <=0 use WP_LOITER_RAD
+    // @Units: m
+--]]
+TA_CMTC_RAD = bind_add_param("CMTC_RAD", 15, 0)
+
+local pitch_down_min = TA_PTCH_DWN_MIN:get()
+local pitch_forward_min = TA_PTCH_FWD_MIN:get()
+local pitch_timeout = TA_PTCH_TIMEOUT:get()
+local home_distance_max = TA_HOME_DIST:get() or 75
```
IamPete1:
```suggestion
local home_distance_max = TA_HOME_DIST:get()
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+
+local quad_down_min = TA_QUAD_DWN_MIN:get()
+local quad_forward_min = TA_QUAD_FWD_MIN:get()
+
+local altitude_max = TA_ALT_MAX:get()
+
+local groundspeed_max = TA_GSP_MAX:get() or -1
```
IamPete1:
```suggestion
local groundspeed_max = TA_GSP_MAX:get()
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+local groundspeed_airbrake_limit = TA_GSP_AIRBRAKE:get() or 0
```
IamPete1:
```suggestion
local groundspeed_airbrake_limit = TA_GSP_AIRBRAKE:get()
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+
+local cmtc_height_m = TA_CMTC_HGT:get()
+if cmtc_height_m == 0 then
+    cmtc_height_m = pitch_down_min
+end
+local cmtc_enable = TA_CMTC_ENABLE:get() or 0
```
IamPete1:
```suggestion
local cmtc_enable = TA_CMTC_ENABLE:get()
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+
+MIN_ALT_MAX = 20
+
+REFRESH_RATE = 1.0 / (TA_UPDATE_RATE:get() or 10)
```
IamPete1:
```suggestion
REFRESH_RATE = 1.0 / TA_UPDATE_RATE:get()
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+local mavlink = require("mavlink_wrappers")
+
+local location_tracker  -- forward declaration. See below for definition and instantiation.
+
+-------------------------------------------------
+-- deal with deprecations in 4.7
+-------------------------------------------------
+version47orhigher = true
+---@diagnostic disable-next-line: lowercase-global`
+get_yaw_function = ahrs.get_yaw_rad
```
IamPete1:
```suggestion
Get_Yaw_Function = ahrs.get_yaw_rad
```
This brings it inline with `Get_Roll_Function` and `Get_Pitch_Function`

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+        else
+            gcs:send_text(MAV_SEVERITY.ERROR, SCRIPT_NAME_SHORT .. ": failed to set alt: " .. new_altitude)
+        end
+    end
+    return new_altitude
+end
+
+-- This method forces the plane to target a very high altitude to force it to gain altitude quicky (fixed wing only)
+local function set_altitude_high()
+    -- local target_altitude = current_altitude_m + 2000
```
IamPete1:
```suggestion
```

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+        quad_down_min = TA_QUAD_DWN_MIN:get()
+        quad_forward_min = TA_QUAD_FWD_MIN:get()
+
+        altitude_max = TA_ALT_MAX:get()
+
+        groundspeed_max = TA_GSP_MAX:get() or -1
+        groundspeed_airbrake_limit = TA_GSP_AIRBRAKE:get() or 0
+        airspeed_min = AIRSPEED_MIN:get() or 10
+        airspeed_cruise = AIRSPEED_CRUISE:get() or 18
+        airspeed_max = AIRSPEED_MAX:get() or 30
```
IamPete1: Don't need the `or`

libraries/AP_Scripting/applets/quadplane_terrain_avoid.lua
```diff
@@ -0,0 +1,1324 @@
...
+
+        cmtc_enable = TA_CMTC_ENABLE:get()
+        if cmtc_enable then
+            cmtc_height_m = TA_CMTC_HGT:get() or 0
```
IamPete1:
```suggestion
            cmtc_height_m = TA_CMTC_HGT:get()
```

libraries/AP_Scripting/modules/mavlink_command_int.lua
```diff
@@ -0,0 +1,80 @@
...
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+   MAVLink_command_int
+   A MAVlink message manager for COMMAND_INT messages specifically
+   Only send is implemented at the moment, but its intended that receive could be added
+--]]
+
+local MAVLink_command_int = {}
```
IamPete1: Is this file still used?

IamPete1 (dismissed):

Tools/autotest/quadplane.py
```diff
@@ -2752,6 +2752,196 @@ def RudderArmingWithARMING_CHECK_THROTTLEUnset(self) -> None:
...
+        self.set_parameters({
+            "SCR_ENABLE": 1,
+            "SIM_SPEEDUP": 20, # need to give some cycles to lua
+            "RC7_OPTION": 305,
+            "RTL_AUTOLAND": 2,
+            "RNGFND1_TYPE": 100,
+        })
+
+        self.install_applet_script_context("quadplane_terrain_avoid.lua")
+        self.install_script_module(self.script_modules_source_path("mavlink_wrappers.lua"), "mavlink_wrappers.lua")
```
peterbarker, line 2822: Could you make sure these modules are removed after the test has run, please?

Tools/autotest/quadplane.py
```diff
@@ -2752,6 +2752,196 @@ def RudderArmingWithARMING_CHECK_THROTTLEUnset(self) -> None:
...
+        })
+
+        # This mission triggers an intersting selection of "Pitching", "Quading" and "CMTC" events
+        # it's not always consistent, perhaps due to wind, so the tests try to accomodate variances.
+        filename = "TopOfTheWorldShort.waypoints"
+        self.progress("Flying mission %s" % filename)
+        num_wp = self.load_mission(filename)
+        self.progress("Mission items %d" % num_wp)
+
+        self.send_cmd_do_set_mode("AUTO")
```
peterbarker: Why not `self.change_mode`?

Tools/autotest/quadplane.py
```diff
@@ -2752,6 +2752,196 @@ def RudderArmingWithARMING_CHECK_THROTTLEUnset(self) -> None:
...
+        expected_terrain_height = 1101
+        if abs(report.terrain_height - expected_terrain_height) > 1.0:
+            raise NotAchievedException("Expected terrain height=%f got=%f" %
+                                       (expected_terrain_height, report.terrain_height))
+
+        self.set_rc(7, 1000)
+        self.wait_text("TerrAvoid: activated", check_context=True)
+        self.set_rc(7, 2000)
+        self.wait_text("TerrAvoid: deactivated", check_context=True)
+        self.set_rc(7, 1000)
```
peterbarker, line 2890: Isn't high usually activated?

Tools/autotest/quadplane.py
```diff
@@ -2752,6 +2752,196 @@ def RudderArmingWithARMING_CHECK_THROTTLEUnset(self) -> None:
...
+
+        # wait for 1 more CMTC
+        self.wait_text("TerrAvoid: CMTC Done", check_context=True, regex=True)
+
+        self.wait_statustext('Land complete', timeout=600)
+        self.wait_disarmed(timeout=120) # give quadplane a long time to land
+
+        # autotest doesn't like this location, so need to move back to Dalby before finishing
+        self.customise_SITL_commandline(
+            ["--home", "-27.274439,151.290064,343.0,0"]
```
peterbarker, line 2975: Sounds like something I need to fix in the harnness, really.

Tools/autotest/quadplane.py
peterbarker: To clarify, let me know, don't necessarily remove them yourself.  It's just I was expecting something to do with "context"s here somewhere, so I'm worried we're leaving stuff lying around

timtuxworth:
Based on the Dev Call, I wonder if I could get an approval @tpwrules , @IamPete1 @peterbarker , @Ryanf55 ?

tpwrules (approved):
LGTM, though the Lua is not quite as beautiful as it could be. But if it works that's great!

cclauss:
Two typos added...
```
Tools/autotest/quadplane.py:2838: intersting ==> interesting
Tools/autotest/quadplane.py:2839: accomodate ==> accommodate
```
Fixed in
* #30907

PR #28630, opened 2024-11-15
Enable BATTx_options for AP_Periph builds if useful - Battery summing on a node

AlfieLockrey:
This PR adds a condition for the options parameter to be included in the build for AP_Periph firmware if AP_BATTERY_SUM_ENABLED is true. 

This is a niche use case where a user is using a battery monitor sum of other monitor(s) and may want to be able to set the voltage source as minimum rather than average. Current builds do not include this options parameter as it is kept out of AP_Periph builds.

This has been tested on a MATEK L431 Periph CAN Node with the hwdef.dat file modified. I have verified that the options are included when `define AP_BATTERY_SUM_ENABLED 1` is included and are not present when removed from the hwdef.dat.

Current hardware definitions for the L431 don't have these in by default however it may be beneficial to include them if they don't hinder other functionality. The inclusion of 4 adcs allows for voltage and current to be measured from two battery sources and combining them may be helpful for more complex or non-standard vehicles.

I have included the `defined(HAL_BUILD_AP_PERIPH)` as I was unsure of any consequences if the parameter is defined twice.


libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
```diff
@@ -169,6 +169,15 @@ const AP_Param::GroupInfo AP_BattMonitor_Params::var_info[] = {
     AP_GROUPINFO("ESC_INDEX", 22, AP_BattMonitor_Params, _esc_telem_outbound_index, 0),
 #endif
 
+#if AP_BATTERY_SUM_ENABLED && defined(HAL_BUILD_AP_PERIPH)
+    // @Param: OPTIONS
+    // @DisplayName: Battery monitor options
+    // @Description: This sets options to change the behaviour of the battery monitor
+    // @Bitmask: 0:Ignore DroneCAN SoC, 1:MPPT reports input voltage and current, 2:MPPT Powered off when disarmed, 3:MPPT Powered on when armed, 4:MPPT Powered off at boot, 5:MPPT Powered on at boot, 6:Send resistance compensated voltage to GCS, 7:Allow DroneCAN InfoAux to be from a different CAN node, 9:Sum monitor measures minimum voltage instead of average
+    // @User: Advanced
+    AP_GROUPINFO("OPTIONS", 21, AP_BattMonitor_Params, _options, 0),
```
IamPete1, line 178: Better to keep this in one place. Edit the defines so it will work on line 159.

IamPete1: You have done the right thing, thanks.

libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
```diff
@@ -150,14 +150,16 @@ const AP_Param::GroupInfo AP_BattMonitor_Params::var_info[] = {
     AP_GROUPINFO("ARM_MAH", 19, AP_BattMonitor_Params, _arming_minimum_capacity, 0),
 
     // 20 was BUS
+#endif // HAL_BUILD_AP_PERIPH
 
+#if !defined(HAL_BUILD_AP_PERIPH) || AP_BATTERY_SUM_ENABLED
```
peterbarker, line 155: Please create a fresh define name in `AP_Battery_config.h`.

libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
```diff
@@ -150,14 +150,16 @@ const AP_Param::GroupInfo AP_BattMonitor_Params::var_info[] = {
...
     // @Param: OPTIONS
     // @DisplayName: Battery monitor options
     // @Description: This sets options to change the behaviour of the battery monitor
     // @Bitmask: 0:Ignore DroneCAN SoC, 1:MPPT reports input voltage and current, 2:MPPT Powered off when disarmed, 3:MPPT Powered on when armed, 4:MPPT Powered off at boot, 5:MPPT Powered on at boot, 6:Send resistance compensated voltage to GCS, 7:Allow DroneCAN InfoAux to be from a different CAN node, 9:Sum monitor measures minimum voltage instead of average
     // @User: Advanced
     AP_GROUPINFO("OPTIONS", 21, AP_BattMonitor_Params, _options, 0),
-#endif // HAL_BUILD_AP_PERIPH
+#endif
```
peterbarker, line 162: include the new define name on this line too

peterbarker (changes requested):
Please squash all commits together and prefix with `AP_Battery:`

So perhaps

`AP_Battery: include OPTIONS parameter on boards with SUM backend`

It *might* be we actually just want to set it to (!defined(HAL_BUILD_AP_PERIPH)` and let the person creating the Perioph hwdef set the  option in that hwdef.

PR #28633, opened 2024-11-15
Fixed an explicit initializer to get rid of Qurt compiler warning

katzfey:
Added another set of braces to an initializer for a local variable to get rid of the following Qurt compiler warning:

../../libraries/AP_RCProtocol/spm_srxl.cpp:1307:26: warning: suggest braces around initialization of subobject [-Wmissing-braces]
    SrxlFullID retVal = {0};
                         ^
                         {}


katzfey:
Take a look at PR28643: https://github.com/ArduPilot/ardupilot/pull/28643

If that is a more reasonable way to deal with these compiler warnings then I can close this PR.

tridge (approved):

PR #28634, opened 2024-11-15
Remove extraneous zero initializers to clean up Qurt compiler warnings

katzfey:
Extraneous zero initializers for global / static variables were removed to get rid of the following Qurt compiler warnings:

../../libraries/AP_RCProtocol/spm_srxl.cpp:78:36: warning: suggest braces around initialization of subobject [-Wmissing-braces]
SrxlTelemetryData srxlTelemData = {0};
                                   ^
                                   {}
../../libraries/AP_RCProtocol/spm_srxl.cpp:83:34: warning: suggest braces around initialization of subobject [-Wmissing-braces]
static SrxlDevice srxlThisDev = {0};
                                 ^
                                 {}
../../libraries/AP_RCProtocol/spm_srxl.cpp:91:36: warning: suggest braces around initialization of subobject [-Wmissing-braces]
static SrxlReceiverStats srxlRx = {0};
                                   ^
                                   {}
../../libraries/AP_RCProtocol/spm_srxl.cpp:91:36: warning: suggest braces around initialization of subobject [-Wmissing-braces]
static SrxlReceiverStats srxlRx = {0};
                                   ^
                                   {}

Since this data will end up in .bss and be zeroed out it does not need explicit zero initializers.

### DIFF
--- a/libraries/AP_RCProtocol/spm_srxl.cpp
+++ b/libraries/AP_RCProtocol/spm_srxl.cpp
@@ -74,27 +74,27 @@ const uint16_t srxlCRCTable[] =
 
 /// PUBLIC VARIABLES ///
 
-SrxlChannelData srxlChData = {0, 0, 0, {0}};
-SrxlTelemetryData srxlTelemData = {0};
+SrxlChannelData srxlChData;
+SrxlTelemetryData srxlTelemData;
 SrxlVtxData srxlVtxData = {0, 0, 1, 0, 0, 1};
 
 /// LOCAL VARIABLES ///
 
-static SrxlDevice srxlThisDev = {0};
+static SrxlDevice srxlThisDev;
 static SrxlBus srxlBus[SRXL_NUM_OF_BUSES];
 static bool srxlChDataIsFailsafe = false;
 static bool srxlTelemetryPhase = false;
 #ifdef SRXL_INCLUDE_MASTER_CODE
-static uint32_t srxlFailsafeChMask = 0;  // Tracks all active channels for use during failsafe transmission
+static uint32_t srxlFailsafeChMask;  // Tracks all active channels for use during failsafe transmission
 #endif
-static SrxlBindData srxlBindInfo = {0, 0, 0, 0};
-static SrxlReceiverStats srxlRx = {0};
-static uint16_t srxlTelemSuppressCount = 0;
+static SrxlBindData srxlBindInfo;
+static SrxlReceiverStats srxlRx;
+static uint16_t srxlTelemSuppressCount;
 
 #ifdef SRXL_INCLUDE_FWD_PGM_CODE
-static SrxlFullID srxlFwdPgmDevice = {0, 0};  // Device that should accept Forward Programming connection by default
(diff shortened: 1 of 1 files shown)


andyp1per:
This is an external library, modifying it makes it hard to integrate newer versions - that's why these have not been fixed previously

katzfey:
Shoot! I have 3 PR's with changes to the spm_srxl files. Unfortunately, it's a C file that was copied over and given the cpp extension. Then the C++ compiler gives warnings because it has C specific stuff in it.

katzfey:
@andyp1per Take a look at PR28643: https://github.com/ArduPilot/ardupilot/pull/28643

If that is a more reasonable way to deal with these compiler warnings then I can close this PR.

tridge:
> This is an external library, modifying it makes it hard to integrate newer versions - that's why these have not been fixed previously

I think we just need to wear that cost next time we update, more important to get rid of clang warnings I think

tridge:
@katzfey what is a bit surprising is this doesn't seem to save any flash space. I would have expected variables to move from DATA to BSS with a corresponding flash saving

peterbarker:
Alternate forumation, same as we use in LUA:
```
--- a/libraries/AP_RCProtocol/spm_srxl.cpp
+++ b/libraries/AP_RCProtocol/spm_srxl.cpp
@@ -22,6 +22,11 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
+#if ARDUPILOT_BUILD
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-braces"
+#endif
+
 #include <string.h>
 #include <stdint.h>
 
@@ -1428,3 +1433,7 @@ bool srxlUpdateCommStats(bool isFade)
     // Return true while we're in hold condition (failsafe)
     return srxlRx.lossCountdown == 0;
 }
+
+#if ARDUPILOT_BUILD
+#pragma GCC diagnostic pop
+#endif
```

.... but the compiler warning is pointing out a bug, so I reckong we just accept this PR.

If we lock the error in with a `-Werror` then the next person to freshen this file from its source gets to fix the bugs again :-)

tridge:
@andyp1per my preference is not to use the ifdefs, and just fix the code, I think the cost of updating these quite small files at some point in the future will be low, and clang is correct about the initialisers being wrong for these nested structures

peterbarker (approved):

tridge (approved):

tridge:
I'd like @andyp1per to have a chance to comment before merge

katzfey:
@tridge My understanding is that the compiler sees that you are initializing with 0 and puts them in bss anyways so shouldn't change anything there.

andyp1per (commented):
This looks ok to me - what testing have you done?

katzfey:
I have only tested on my Qurt based drone so it isn't actually exercising any of the changed code. But, in theory, this doesn't actually change anything.

katzfey:
In fact, the object file is identical after the change.

andyp1per (approved):

PR #28638, opened 2024-11-15
Cast enum to int to fix Qurt compiler warning

katzfey:
Cast enum to an int to get rid of Qurt compiler warning when max enum value is less than the constant it is being compared to. Here is the warning:

../../libraries/AP_GPS/AP_GPS.cpp:1622:14: warning: result of comparison of constant 8 with expression of type 'mavlink_channel_t' is always false [-Wtautological-constant-out-of-range-compare]
    if (chan >= MAVLINK_COMM_NUM_BUFFERS) {
        ~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -1619,7 +1619,7 @@ void AP_GPS::handle_gps_rtcm_data(mavlink_channel_t chan, const mavlink_message_
  */
 bool AP_GPS::parse_rtcm_injection(mavlink_channel_t chan, const mavlink_gps_rtcm_data_t &pkt)
 {
-    if (chan >= MAVLINK_COMM_NUM_BUFFERS) {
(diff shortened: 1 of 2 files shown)


peterbarker:
We went the other way for a similar thing:
```
libraries/GCS_MAVLink/GCS_MAVLink.h:#pragma clang diagnostic ignored "-Wtautological-constant-out-of-range-compare"
```

katzfey:
If that's what you guys prefer then I can certainly change it to that. Was the int cast discussed when that one was put in?

katzfey:
Casting as an int certainly looks nicer than all that pragma gunk. C99 section 6.7.2.2 §2: "The expression that defines the value of an enumeration constant shall be an integer constant expression that has a value representable as an int."; compilers are free to use a different integer type to represent the enumeration, but conversion to int is always lossless

peterbarker (approved):
LGTM

peterbarker:
Merged, thanks!

It would be nice to adjust our compiler flags so these don't creep back in - even in non-QURT code.

PR #28642, opened 2024-11-15
Replace fabsF macro with std::abs call to avoid having float / double mismatch warnings

katzfey:
The fabsF macro is defined as either the float version or double version in ftype.h depending on HAL_WITH_EKF_DOUBLE. If it is defined as the float version and a double is used as the argument then we get a compiler warning in the Qurt build. If those are replaced with std:abs calls then that clears up the warnings since it is overloaded for both floats and doubles.

### DIFF
--- a/ArduCopter/mode_guided.cpp
+++ b/ArduCopter/mode_guided.cpp
@@ -733,7 +733,7 @@ void ModeGuided::pos_control_run()
 
     float pos_offset_z_buffer = 0.0; // Vertical buffer size in m
     if (guided_pos_terrain_alt) {
-        pos_offset_z_buffer = MIN(copter.wp_nav->get_terrain_margin() * 100.0, 0.5 * fabsF(guided_pos_target_cm.z));
+        pos_offset_z_buffer = MIN(copter.wp_nav->get_terrain_margin() * 100.0, 0.5 * std::abs(guided_pos_target_cm.z));
     }
     pos_control->input_pos_xyz(guided_pos_target_cm, terr_offset, pos_offset_z_buffer);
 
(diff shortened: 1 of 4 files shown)


katzfey:
Perhaps fabsF should be removed everywhere and replaced by std::abs?

ArduCopter/mode_guided.cpp
```diff
@@ -733,7 +733,7 @@ void ModeGuided::pos_control_run()
 
     float pos_offset_z_buffer = 0.0; // Vertical buffer size in m
     if (guided_pos_terrain_alt) {
-        pos_offset_z_buffer = MIN(copter.wp_nav->get_terrain_margin() * 100.0, 0.5 * fabsF(guided_pos_target_cm.z));
+        pos_offset_z_buffer = MIN(copter.wp_nav->get_terrain_margin() * 100.0, 0.5 * std::abs(guided_pos_target_cm.z));
```
tridge, line 736: if the variable is a float, will it promote to double by making this change? That would slow things a lot on older boards

tridge (changes requested):

tridge:
cppreference shows only integer versions?
image

tridge:
we could add a float cast to the float version of fabsF()
also note that we generally trying to avoid the std library, it often causes more pain than it fixes

katzfey:
The float versions are here: https://en.cppreference.com/w/cpp/numeric/math/fabs

katzfey:
But if you want to avoid using std library then let me try with the float cast

katzfey:
Closing this PR. Going with https://github.com/ArduPilot/ardupilot/pull/28897 instead with the cast method instead of the switch to std::abs proposed in this PR.

PR #28643, opened 2024-11-15
Wrap the imported spm_srxl files to allow using pragmas to disable compiler warnings

katzfey:
This is another attempt to deal with compiler warnings from the spm_srxl files brought in from an external library. Rather than modifications to the files themselves they are wrapped to allow pragmas to be included to disable the compiler warnings in the wrapper. This PR would be in lieu of PRs 28632, 28633, and 28634.

https://github.com/ArduPilot/ardupilot/pull/28632
https://github.com/ArduPilot/ardupilot/pull/28633
https://github.com/ArduPilot/ardupilot/pull/28634


katzfey:
Will close this one since the alternate ones were accepted.

PR #28652, opened 2024-11-16
AP_Scripting: offset_bearing fixed in an example

landswellsong:
I was playing with Lua examples and I've noticed that `offset_bearing` wasn't called with correct arguments in the spiral test. And the autotest would still pass since it wasn't comparing the shape per se and just the fact that what was written would match the generated points.

This should be helpful for the next person who tries to write scripts by copying parts of examples.


landswellsong:
Rebased over 4.7.

PR #28653, opened 2024-11-16
Plane: slew limit all throttles in one place

IamPete1:
Rather than do the slew calculation for each throttle we can do it once and then set all throttles. Previously throttle left and right were done in a different function, this change means all throttles behave the same in all cases.

### DIFF
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -1159,7 +1159,7 @@ private:
     void servos_twin_engine_mix();
     void force_flare();
     void throttle_watt_limiter(int8_t &min_throttle, int8_t &max_throttle);
-    void throttle_slew_limit(SRV_Channel::Aux_servo_function_t func);
+    void throttle_slew_limit();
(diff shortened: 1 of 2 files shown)


Georacer (approved):
This would prevent us from applying separate slew rates between `k_throttle` and `k_throttleLeft/k_throttleRight`. But we aren't doing that now and premature optimization is a sin anyways, so all good.

I would ask if there's a case of an error if we call `set_slew_rate` on an unconfigured servo type, but apparently the autotests are happy, so that's covered too.

tridge (approved):

rmackay9:
BTW, I think this has no functional impact so I've left it out of the 4.7 release notes.  I comment here because I'm not actually sure if I've made the correct decision

IamPete1:
> BTW, I think this has no functional impact so I've left it out of the 4.7 release notes. I comment here because I'm not actually sure if I've made the correct decision

Your correct, no change in behaviour.

PR #28654, opened 2024-11-16
Plane: tailsitter: tidy tilt left and right handling

IamPete1:
Tidy up the the tailsitter output method. The scope of the tilt left and right variables can be reduced making the code easier to follow.

### DIFF
--- a/ArduPlane/tailsitter.cpp
+++ b/ArduPlane/tailsitter.cpp
@@ -289,9 +289,6 @@ void Tailsitter::output(void)
         return;
     }
 
-    float tilt_left = 0.0f;
-    float tilt_right = 0.0f;
-
     // throttle 0 to 1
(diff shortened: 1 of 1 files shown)


ArduPlane/tailsitter.cpp
```diff
@@ -438,8 +442,10 @@ void Tailsitter::output(void)
         tailsitter_motors->set_min_throttle(0.0);
     }
 
-    tilt_left = 0.0f;
-    tilt_right = 0.0f;
```
Georacer, line 442: Agreed. The fact that we were nulling tilt here basically creates a notional separation between its previous usages.

Georacer (approved):

peterbarker (approved):
LGTM.

PR #28670, opened 2024-11-18
AP_Arming: typo fix in magnetic model check

landswellsong:
Judging from the context of the code and the error message, the Z component should be checked instead of X which is already checked in the clause above. This would lead to Z component never actually checked and X checked against two differing thresholds.

The correction should not affect the autotests since none seem to be testing for this exact message.


landswellsong:
Tagging @rmackay9 as the author of the original code to make sure I understood it correctly.

rmackay9 (approved):
Oh my, thanks for the fix

peterbarker (approved):

lthall (approved):
This looks like an obvious bug fix.

PR #28675, opened 2024-11-19
TECS: Correctly handle Home altitude change

Georacer:
## Summary

In case where the Home altitude is changed, TECS will now not reset, but offset the relevant attributes instead.

## Background

Recently one of our partners flying a quadplane reported that while doing a ship landing, their plane would continuously lose altitude.
image

## Details

The reason is that the most minute home altitude changes [trigger a TECS reset](https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/altitude.cpp#L47). This behaviour is about 1 year old.

In the meantime, the ship landing lua script continuously [updates the home location](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/applets/plane_ship_landing.lua#L432).
This code is ~2years old.

The combination results in the TECS constantly switching into a state of rest and becoming effectively paralyzed.

## Solution

I added a new method `AP_TECS::offset_altitude(const float alt_offset)` where TECS is made aware of the change in home altitude and offsets all internal attributes accordingly.

This is a pretty wide-reaching change, so all comments and test proposals are welcome.

## Testing

An autotest was created for Plane (`SetHomeAltChange2`) that, once the plane comes to altitude, it wiggles the home altitude over 10s.
The MAVLink command is set at 10Hz, but it seems to get parsed only at 2Hz.

Still, the results are conclusive:

**Before**:
- When the wiggle starts, TECS starts resetting (bottom right)
- The actual altitude starts climbing, since the trims cause the plane to climb (bottom middle).
- All TECS quantities start wiggling too, including throttle and pitch. Still, this is an exaggerated case, since home is wiggled by +-40m.

Screenshot from 2024-11-25 16-16-34

**After**:
- When the wiggle starts, TECS does not reset (bottom right).
- The actual altitude stays on target (bottom middle).
- All TECS quantities are stable, except for altitude (middle left). This is expected, since its frame is home-relative.

Screenshot from 2024-11-25 16-18-47

## Alternatives

An alternative would be to just not reset TECS as often in case of home altitude change. But what is a "good" interval? Are we happy with TECS resetting every 10s? 20s?

Another alternative would be to just not allow the user to update home while TECS is running, but that sounds a bit silly to me.

### DIFF
--- a/ArduPlane/altitude.cpp
+++ b/ArduPlane/altitude.cpp
@@ -44,7 +44,7 @@ void Plane::check_home_alt_change(void)
             next_WP_loc.alt += alt_change_cm;
         }
         // reset TECS to force the field elevation estimate to reset
-        TECS_controller.reset();
+        TECS_controller.offset_altitude(alt_change_cm * 0.01f);
     }
     auto_state.last_home_alt_cm = home_alt_cm;
 }
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -5,6 +5,7 @@ AP_FLAKE8_CLEAN
 '''
 
 from __future__ import print_function
+import copy
 import math
 import os
 import signal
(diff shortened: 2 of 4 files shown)


tridge:
@Georacer really nice work tracking this down!
as I mentioned on discord, I think we need a test for this

PR #28678, opened 2024-11-19
AP_Battery: Enable BATTx_options for AP_Periph builds if battery summing is used

AlfieLockrey:
**Updated PR as wrong branch was originally used (#28630). @peterbarker @IamPete1 **

Adds `AP_BATTERY_OPTIONS_PARAM_ENABLED` that is true when the build is NOT a Periph build OR if the build has battery summing enabled. If enabled, the Battx_options bitmask is present, allowing battery summing behaviour to be modified on a periph node.

This is a niche use case where a user is using a battery monitor sum of other monitor(s) and may want to be able to set the voltage source as minimum rather than average. Current builds do not include this options parameter as it is kept out of AP_Periph builds.

This has been tested on a MATEK L431 Periph CAN Node with the hwdef.dat file modified. I have verified that the options are included when define AP_BATTERY_SUM_ENABLED 1 is included and are not present when removed from the hwdef.dat.

Current hardware definitions for the L431 don't have these in by default however it may be beneficial to include them if they don't hinder other functionality. The inclusion of 4 adcs allows for voltage and current to be measured from two battery sources and combining them may be helpful for more complex or non-standard vehicles.


libraries/AP_BattMonitor/AP_BattMonitor_config.h
```diff
@@ -119,3 +119,7 @@
 #ifndef AP_BATTERY_SCRIPTING_ENABLED
 #define AP_BATTERY_SCRIPTING_ENABLED (AP_SCRIPTING_ENABLED && AP_BATTERY_BACKEND_DEFAULT_ENABLED)
 #endif
+
+#ifndef AP_BATTERY_OPTIONS_PARAM_ENABLED
+#define AP_BATTERY_OPTIONS_PARAM_ENABLED (!defined(HAL_BUILD_AP_PERIPH) || AP_BATTERY_SUM_ENABLED)
+#endif
```
IamPete1, line 125: Just need a new line at the end of the file to get rid of the stop sign thing.
```suggestion
#endif

```

IamPete1 (approved):
LGTM, Thanks

peterbarker:
Merged, thanks!

PR #28682, opened 2024-11-19
AP_InertialSensor: constrain ICM45686 to 800Hz-6.4KHz

andyp1per:
Simpler alternative to https://github.com/ArduPilot/ardupilot/pull/27573


bugobliterator:
Hi @andyp1per I don't like this way of doing it actually. This PR changes the ODR of the sensor to a seriously low value, which is quite a regression, and changes filter behaviour and possibly have affect on measurement quality as well. I would really like to continue to find a way to get the initial PR in.

andyp1per:
@bugobliterator I changed this so the ODR is unchanged for the settings you are concerned about. To get the other PR inplease can you make it this simple - I had to write this PR simply to get straight in my head what your PR is actually trying to do - there are so many non-core changes that its really hard to parse and its the non-core changes that I am really objecting to. The actual change you want should be  a matter of a few lines.

Issue #23927, opened 2023-05-30
Set Servo options triggered by failsafe


alexanderever:
## Feature request

For UAS which perform a function controlled by Set_servo outputs, such as crop spraying, it would be desirable for the UAS to stop doing the function if a failsafe mode is triggered. It is not clear if the Sprayer functionality includes this but it may be useful for other  applications such as vectored thrust VTOLs etc

Currently the consequence of RC signal loss while spraying would be the UAS performing a RTL while still spraying. This would be VERY BAD as the spray would be released at a high altitude and subsequently drift to unintended locations. 

**Describe the solution you'd like**
_A clear and concise description of what you want to happen._

A parameter to set servo outputs should failsafe be triggered Eg SERVOXX_FS_VALUE = 1000 

**Describe alternatives you've considered**
_A clear and concise description of any alternative solutions or features you've considered._
Lua scripting?

checked the Sprayer functionality in ardupilot, it is unclear if this issue is addressed. 

(Also documentation on "SPRAY_SPINNER Spinner rotation speed: Spinner's rotation speed in PWM (a higher rate will disperse the spray over a wider area horizontally)" is inaccurate and could cause serious issues, increased RPM actually creates smaller droplets.  see https://en.wikipedia.org/wiki/Rotary_atomizers)

**Platform**
[ x ] All
[x  ] AntennaTracker
[ x ] Copter
[  x] Plane
[ x ] Rover
[  x] Submarine

**Additional context**
_Add any other context or screenshots about the feature request here._

rmackay9:
Thanks for the feature request.  Turning off the sprayer during any kind of failsafe makes a lot of sense.

alexanderever:
Hi Randy,

thanks for considering the feature!

 I work for an agricultural company with a long history of spraying in ground based and aerial settings. We also contribute to international discussions on drones as spraying systems. There are a number of features we can think of that we feel would improve Arducopter equipped aircraft for this role.

i'd be very interested in helping the development of Arducopter.

Please feel free to contact me at ***@***.***

Thanks,

Alexander

________________________________

peterbarker:
@alexanderever a script has been offered up to accomplish what you wanted here.  Would you be happy with such a solution?

alexanderever:
Hi Peter, 

Its great that my idea has been considered. thanks to everyone involved! 

[...]

PR #28684, opened 2024-11-20
Turn off sprayer on RTL or mission complete

VICTUSE:
**Reason**
Almost all the crop sprayers are used in an auto mission with survay plan (takeoff, survay the field, RTL). Current AC_Sprayer library  manages spraying according to the speed of the vehicle. Imagine a field is gonna spraying, wp1 is 100m far away from home, copter wastes it's fertilizer. Same thing for the RTL too. 
**Method**
In a survay plan, field would probably start from waypoint index 3 so spraying won't start until reach to the field starting point on AUTO mode. `if ((current_wp_index >= 3) && !((current_cmd.id == MAV_CMD_NAV_RETURN_TO_LAUNCH) || (mission_state == 2)))` in this condition `current_cmd.id == MAV_CMD_NAV_RETURN_TO_LAUNCH) || (mission_state == 2` will make sure `should_be_spraying` won't be true so spraying wouldn't start again after mission complete or RTL command.

This condition `if (vehicle->get_mode() == COPTER_MODE_AUTO)` should make sure `if ((current_wp_index >= 3) && !((current_cmd.id == MAV_CMD_NAV_RETURN_TO_LAUNCH) || (mission_state == 2)))` cond. gets checked only on AUTO so user freely spray in other modes. 
**SITL Test**
wp2
RTL
I printed aux pwm during mission
on 1) Copter starts spraying
on 2) Where RTL command comes so spraying stops 
**Conclusion**
SITL sim shows that start and stop method works on AUTO. I don't know if `if (vehicle->get_mode() == COPTER_MODE_AUTO)` method works and we can spray on manual modes. 
I only made SITL sim. I didn't test it on hardware.
This is my first contribution so there might be huge mistakes please guide me.
Thanks.

### DIFF
--- a/libraries/GCS_MAVLink/MissionItemProtocol_Fence.cpp
+++ b/libraries/GCS_MAVLink/MissionItemProtocol_Fence.cpp
@@ -110,10 +110,16 @@ MAV_MISSION_RESULT MissionItemProtocol_Fence::convert_MISSION_ITEM_INT_to_AC_Pol
     switch (mission_item_int.command) {
     case MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION:
         ret.type = AC_PolyFenceType::POLYGON_INCLUSION;
+        if (mission_item_int.param1 > 255) {
+            return MAV_MISSION_INVALID_PARAM1;
+        }
         ret.vertex_count = mission_item_int.param1;
         break;
     case MAV_CMD_NAV_FENCE_POLYGON_VERTEX_EXCLUSION:
         ret.type = AC_PolyFenceType::POLYGON_EXCLUSION;
+        if (mission_item_int.param1 > 255) {
+            return MAV_MISSION_INVALID_PARAM1;
+        }
         ret.vertex_count = mission_item_int.param1;
         break;
     case MAV_CMD_NAV_FENCE_RETURN_POINT:


rmackay9:
Hi @VICTUSE,

Thanks very much for this.  How do you feel about implementing this as a lua script instead?

Note that we have a request for this feature here on the issues list https://github.com/ArduPilot/ardupilot/issues/23927

VICTUSE:
Hello @rmackay9 ,

I believe I have addressed all requested changes, including the prefix update and removal of the merge commit. Please let me know if further adjustments are needed.

I cheked #23927 feature request. I can work on implementing that request and my idea as a lua script.

rmackay9:
@VICTUSE,

Thanks very much for the changes.  We can see what other developers say but I suspect the consensus will be that we shouldn't have a low level driver reliant on the higher level AP_Vehicle or AP_Mission libraries.  AP_Vehicle in particular is basically the top of the software stack while AC_Sprayer is down at the bottom.  The check of the vehicle mode vs the copter specific number is also probably a no-no.  It will incorrectly match on vehicles that are not copters... of course AC_Sprayer is probably only used by Copter anyway, but still, it highlights an architectural issue.

If you're happy to re-implement this as a lua script then I think it will be much easier to merge.  Lua script take a little more effort for the user to load onto their vehicles of course.

IamPete1:
We have a mission command for sprayer start/stop, so I think that covers the mission case. Stopping spraying in RTL and land seems reasonable, that would be a very easy script. you can use:
```
local sprayer_function = 15
local sprayer_run = 2
local sprayer_stop = 0

rc:run_aux_function(sprayer_function, sprayer_run)

```

VICTUSE:
@IamPete1  thank you very much,
@rmackay9 Hello again,

I have written the lua script and sitl tested couldn't find any problem. No parameter or something else user will only activate the scripting 
Should I put it in AP_Scripting/applets and create PR or do something else ?

IamPete1:
> Should I put it in AP_Scripting/applets and create PR or do something else ?

A PR to add it as a applet would be great! Only requirement for apples is a little readme file that explains what it does, thanks!

PR #28687, opened 2024-11-20
hwdef: new board CSKY-PMU

jstc4ll:
Power Management Unit by CSKY with INA239 current sensor onboard and two additional 12V and 5V GaN-FET based power sources.
image


Hwurzburg:
@jstc4ll is this still active or should we close it?

Hwurzburg:
@jstc4ll you have not split out the Pixhawk and SpeedyBee board changes as separate PRs.....next time I review pending board submissions I will close this if there is no activity

Hwurzburg:
still need to remove the Pixhawk6 and speedbee changes....if you want to make them, do separate PRs....also you must squash all commits and then split into a single commit per library

PR #28690, opened 2024-11-20
Tradheli: fix yaw attitude error corrections while rotors turning on the ground

bnsgeyer:
This PR addresses an issue with the handling of yaw attitude error corrections being fed to the rate controller when the aircraft is on the ground with rotors turning.  Currently the yaw attitude error correction is being fed to the rate controller with both integrator management schemes.  This PR keeps the heading error correction when using the leaky integrator but removes it when using the integrator management that triggers on the landing detector. 

This also cleans up some inconsistencies in the logic with landed and pretakeoff althold states in each of the non-manual throttle flight modes. 

It also addresses the definition of behaviors for Spool up/down which was not addressed at all in the manual throttle modes. The only one that affects copter is Drift

Needs testing.


MattKear:
@bnsgeyer why not apply this to the leaky I case as well? I appreciate it is arguably not needed, but there is also no harm in applying it, as far as I can tell. By not applying it would reduce the amount of checks in the if statements and make the code easier to read.

bnsgeyer:
@MattKear Leaky I allows the I term to grow as soon as motor interlock is enabled.  This PR assumes no change to the leaky I integrator management case.  Thus the heading error corrections are allowed with motor interlock enabled if the user is using leaky I.  

Are you referring to the added logic in the Drift mode?  I realized as I reviewed the current code that drift mode did not conform to how heli stabilize and heli acro handled the I term.  There were some other inconsistencies that I fixed as well.

Ferruccio1984:
Hi @bnsgeyer code looks good to me, will potentially solve the behavior reported for the tandem heli setup. There is a little window of non-safety though for traditional tail rotor helis when taking off/landing on concrete helipads. In the video attached I poked a little the "light on the skids" condition: the situation can potentially evolve into dynamic roll-over if the pilot keep the collective up but not enough to take-off. Could be problematic also for auto takeoff and landing. I would think if adjusting this implementation into a bitmask option would be a win-win for both trad-helis and tandems?
https://www.youtube.com/watch?v=Y5TvBSDiLH8

bnsgeyer:
After working through the details of the changes, here are the desired behaviors for the yaw axis.

The goal is to allow helicopters sensitive to heading corrections on the ground to operate without these corrections.  The heading error correction by the attitude controller is responsible for the heading corrections on the ground.  

This change was only intended for the non-leaky I  integrator scheme and leaky I only affected manual collective modes.  There should be no change when using the leaky i integrator scheme.

The H_OPTIONS Bitmask was used to allow users to choose between attitude controller using heading error correction or not.  

When using heading error correction, behaviors should not have changed in manual and non-manual collective modes. 
landed, motor interlock disabled (althold state: Landed_Ground_Idle) - targets heading reset and zero I term
    this is required to keep stick inputs from moving target heading
    zeroing I term was added for non-manual collective modes - this was missed when spool logic was added
landed, motor interlock enabled (althold state: Landed_Pre_Takeoff) - target heading released but still zero I term

When NOT using heading error correction, only yaw rate feedback is used on the ground.  This serves two purposes. 1) to provide pilot control of yaw axis and 2) allowing autopilot to help fight gusts. 
landed, motor interlock disabled (althold state: Landed_Ground_Idle) - targets heading reset and zero I term
    this is required to keep stick inputs from moving target heading 
landed, motor interlock enabled (althold state: Landed_Pre_Takeoff) - target heading reset and zero I term

ArduCopter/mode_loiter.cpp
```diff
@@ -125,17 +125,21 @@ void ModeLoiter::run()
 
     case AltHoldModeState::MotorStopped:
         attitude_control->reset_rate_controller_I_terms();
-        attitude_control->reset_yaw_target_and_rate();
+        attitude_control->reset_yaw_target_and_rate(true);
```
MattKear, line 128: this is not a change in behaviour, default value is true.

ArduCopter/mode_loiter.cpp
```diff
@@ -125,17 +125,21 @@ void ModeLoiter::run()
...
         loiter_nav->init_target();
         attitude_control->input_thrust_vector_rate_heading(loiter_nav->get_thrust_vector(), target_yaw_rate, false);
         break;
 
     case AltHoldModeState::Landed_Ground_Idle:
-        attitude_control->reset_yaw_target_and_rate();
-        FALLTHROUGH;
+        attitude_control->reset_yaw_target_and_rate(false);
+        attitude_control->reset_rate_controller_I_terms_smoothly();
+        break;
```
MattKear, line 137: This is a fairly significant change in behaviour by removing the fall through.  We would at least want to keep the loiter_nav->init_target()?

ArduCopter/mode_loiter.cpp
```diff
@@ -125,17 +125,21 @@ void ModeLoiter::run()
...
+        attitude_control->reset_yaw_target_and_rate(true);
         pos_control->relax_z_controller(0.0f);   // forces throttle output to decay to zero
         loiter_nav->init_target();
         attitude_control->input_thrust_vector_rate_heading(loiter_nav->get_thrust_vector(), target_yaw_rate, false);
         break;
 
     case AltHoldModeState::Landed_Ground_Idle:
-        attitude_control->reset_yaw_target_and_rate();
-        FALLTHROUGH;
+        attitude_control->reset_yaw_target_and_rate(false);
```
MattKear, line 135: Why wouldn't we want to reset_rate on landed ground idle?

MattKear (changes requested):
Generally speaking, this makes sense, as we do need to improve the heading target handling on the ground.

I do not understand why we need the separation of using leaky I or not in this though.  In my mind, regardless of whether leaky I is enabled, we do not want to allow heading or yaw rate errors to build when we are on the ground.

ArduCopter/mode_acro_heli.cpp
```diff
@@ -45,26 +45,29 @@ void ModeAcro_Heli::run()
     switch (motors->get_spool_state()) {
     case AP_Motors::SpoolState::SHUT_DOWN:
         // Motors Stopped
-        attitude_control->reset_target_and_rate(false);
```
lthall, line 48: Changing this from false to true here will remove all counter talk on the tail rotor. If you can be still spinning down in this phase of flight this will let the tail move more than it would do now.

Low risk but something to consider.

ArduCopter/mode_loiter.cpp
lthall: This provides some damping to the tail if it gets pushed by a gust or something but will stay close to zero as long as we keep the integrator zero.

ArduCopter/mode_loiter.cpp
lthall: Yep, we need to add the loiter_nav->init_target();

I think we can move this to the bottom:

attitude_control->input_thrust_vector_rate_heading(loiter_nav->get_thrust_vector(), target_yaw_rate, false);

libraries/AP_Motors/AP_Motors_Class.h
```diff
@@ -258,6 +258,10 @@ class AP_Motors {
     // This function required for tradheli. Tradheli initializes targets when going from unarmed to armed state.
     // This function is overriden in motors_heli class.   Always true for multicopters.
     virtual bool init_targets_on_arming() const { return true; }
+    // use leaking integrator management scheme is always false for multirotors
+    virtual bool using_leaky_integrator() const { return false; }
+    // use heading error correction which is always true for multirotors
+    virtual bool using_hdg_error_correction() const { return true; }
```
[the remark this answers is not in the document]
lthall: We have been trying to remove the hash defines in the shared code and instead make it flexible enough to cover all the hovering aircraft.

The aspirational goal is to not have any Heli specific code in the shared code base and instead have optional behaviour that is defined by the vehicle.

lthall (commented):
Looks like it is close!

ArduCopter/mode_acro_heli.cpp
bnsgeyer: @lthall What do you mean by counter talk?  I think the issue here is that when disarmed, there won't be any target reset that would cause an issue with trying to fly disarmed.  I know this sounds crazy but I believe that this was intentional by my predecessor to allow users to fly in acro if the aircraft disarms unexpectedly, like due to code bug.  I think that this should true so that the attitude errors are reset.  However I may make this change in a separate PR to make this more visible as an intentional change.

ArduCopter/mode_loiter.cpp
bnsgeyer: Ok.  I wasn't sure that I needed to do this.  I saw that it was done in one place so I thought I needed to do it in all places. I can remove it.

ArduCopter/mode_poshold.cpp
```diff
@@ -116,15 +116,20 @@ void ModePosHold::run()
...
         init_wind_comp_estimate();
-        FALLTHROUGH;
+        break;
 
     case AltHoldModeState::Landed_Pre_Takeoff:
+        if (!motors->using_hdg_error_correction()) {
+            attitude_control->reset_target_and_rate(false);
+        }
         attitude_control->reset_rate_controller_I_terms_smoothly();
+        init_wind_comp_estimate();
```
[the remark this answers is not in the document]
bnsgeyer: @MattKear maybe I was wrong here.  I thought that it was needed because I removed the fallthrough.  I will remove it unless told otherwise.

ArduCopter/mode_poshold.cpp
```diff
@@ -116,15 +116,20 @@ void ModePosHold::run()
...
         pos_control->relax_z_controller(0.0f);   // forces throttle output to decay to zero
+        loiter_nav->clear_pilot_desired_acceleration();
+        loiter_nav->init_target();
```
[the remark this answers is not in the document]
bnsgeyer: @MattKear @lthall This does not follow the same flow as Loiter.  I changed it because loiter is done this way.  Not sure why they would be different.  Again I can change it back if you think it should not change.  Or I can just do a separate PR.

ArduCopter/mode_acro_heli.cpp
bnsgeyer: @Ferruccio1984 What do you think of having the attitude error correction (pitch and roll) removed for the disarmed state in acro.  This means that when you move the pitch and roll stick that the swashplate will not react while disarmed which is what you probably saw in your flight today.  I think it is a good idea because it would keep users from moving the swashplate while disarmed and then being surprised on spool up if the swash was not centered.  I think the idea of ever having to fly in a disarmed state is very low risk.

ArduCopter/mode_loiter.cpp
bnsgeyer: @lthall I think I need to move all of the lines in the pretakeoff state up to the landed ground idle since I removed the fall through.

ArduCopter/mode_poshold.cpp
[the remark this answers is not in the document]
bnsgeyer: fixed.  Thanks,

ArduCopter/mode_loiter.cpp
bnsgeyer: fixed. Thanks.

libraries/AP_Motors/AP_Motors_Class.h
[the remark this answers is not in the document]
MattKear: ok, that is fair enough.  I can certainly get on board with that!
If that is the case, I would still like to see the heli_options param move up to copter level.  I think it is weird that we ask motors about leaky I and heading error correction, both of which effect the rate/attitude controllers which are a "layer" above motors.

ArduCopter/mode_acro_heli.cpp
MattKear: I personally like that we can do "control surface" checks when on the ground.  Making sure that all of the servos and swash are moving and in the correct direction is a good pre-flight check.

bnsgeyer: @MattKear i think my comment was not completely accurate. In acro currently, if you move the sticks, the  commanded attitude is tracked and you could end up having 90 deg commanded while on the ground and disarmed. With ths new implementation, you should still see movement of rhe swash because you are commanding rates and it will always center itself when you center the stick (zero rates) where it does not do that currently.  So you can stick check controls on the ground while disarmed. Again this is only in acro. Stabilize should not change except for yaw and that should react to rate requests also.

ArduCopter/mode_zigzag.cpp
```diff
@@ -356,10 +356,17 @@ void ModeZigZag::manual_control()
...
+        attitude_control->reset_yaw_target_and_rate(false);
+        attitude_control->reset_rate_controller_I_terms_smoothly();
+        loiter_nav->init_target();
+        attitude_control->input_thrust_vector_rate_heading(loiter_nav->get_thrust_vector(), target_yaw_rate);
+        pos_control->relax_z_controller(0.0f);   // forces throttle output to decay to zero
+        break;
 
     case AltHoldModeState::Landed_Pre_Takeoff:
+        if (!motors->using_hdg_error_correction()) {
+            attitude_control->reset_target_and_rate(false);
```
IamPete1, line 368: I would have expected this to be `reset_yaw_target_and_rate`?

ArduCopter/mode_stabilize_heli.cpp
```diff
@@ -52,26 +52,29 @@ void ModeStabilize_Heli::run()
...
             attitude_control->reset_yaw_target_and_rate(false);
             attitude_control->reset_rate_controller_I_terms_smoothly();
         }
         break;
+    case AP_Motors::SpoolState::SPOOLING_UP:
     case AP_Motors::SpoolState::THROTTLE_UNLIMITED:
+    case AP_Motors::SpoolState::SPOOLING_DOWN:
         if (copter.ap.land_complete && !motors->using_leaky_integrator()) {
+            if (!motors->using_hdg_error_correction()) {
+                attitude_control->reset_target_and_rate(false);
```
IamPete1, line 73: `reset_yaw_target_and_rate`?

ArduCopter/mode_sport.cpp
```diff
@@ -76,15 +76,20 @@ void ModeSport::run()
...
-        attitude_control->reset_yaw_target_and_rate();
-        FALLTHROUGH;
+        attitude_control->reset_yaw_target_and_rate(false);
+        attitude_control->reset_rate_controller_I_terms_smoothly();
+        pos_control->relax_z_controller(0.0f);   // forces throttle output to decay to zero
+        break;
 
     case AltHoldModeState::Landed_Pre_Takeoff:
+        if (!motors->using_hdg_error_correction()) {
+            attitude_control->reset_target_and_rate(false);
```
IamPete1, line 91: `reset_yaw_target_and_rate`?

ArduCopter/mode_poshold.cpp
```diff
@@ -116,13 +116,22 @@ void ModePosHold::run()
...
-        FALLTHROUGH;
+
+        // set poshold state to pilot override
+        roll_mode = RPMode::PILOT_OVERRIDE;
+        pitch_mode = RPMode::PILOT_OVERRIDE;
+        break;
 
     case AltHoldModeState::Landed_Pre_Takeoff:
+        if (!motors->using_hdg_error_correction()) {
+            attitude_control->reset_target_and_rate(false);
```
IamPete1, line 133: `reset_yaw_target_and_rate`?

ArduCopter/mode_loiter.cpp
```diff
@@ -132,10 +132,17 @@ void ModeLoiter::run()
...
+        attitude_control->reset_rate_controller_I_terms_smoothly();
+        loiter_nav->init_target();
+        attitude_control->input_thrust_vector_rate_heading(loiter_nav->get_thrust_vector(), target_yaw_rate, false);
+        pos_control->relax_z_controller(0.0f);   // forces throttle output to decay to zero
+        break;
 
     case AltHoldModeState::Landed_Pre_Takeoff:
+        if (!motors->using_hdg_error_correction()) {
+            attitude_control->reset_target_and_rate(false);
```
IamPete1, line 144: `reset_yaw_target_and_rate`?

ArduCopter/mode_flowhold.cpp
```diff
@@ -288,10 +288,15 @@ void ModeFlowHold::run()
...
-        attitude_control->reset_yaw_target_and_rate();
-        FALLTHROUGH;
+        attitude_control->reset_yaw_target_and_rate(false);
+        attitude_control->reset_rate_controller_I_terms_smoothly();
+        pos_control->relax_z_controller(0.0f);   // forces throttle output to decay to zero
+        break;
 
     case AltHoldModeState::Landed_Pre_Takeoff:
+        if (!motors->using_hdg_error_correction()) {
+            attitude_control->reset_target_and_rate(false);
```
IamPete1, line 298: `reset_yaw_target_and_rate`?

ArduCopter/mode_althold.cpp
```diff
@@ -50,15 +50,20 @@ void ModeAltHold::run()
...
-        attitude_control->reset_yaw_target_and_rate();
-        FALLTHROUGH;
+        attitude_control->reset_yaw_target_and_rate(false);
+        attitude_control->reset_rate_controller_I_terms_smoothly();
+        pos_control->relax_z_controller(0.0f);   // forces throttle output to decay to zero
+        break;
 
     case AltHoldModeState::Landed_Pre_Takeoff:
+        if (!motors->using_hdg_error_correction()) {
+            attitude_control->reset_target_and_rate(false);
```
IamPete1, line 65: `reset_yaw_target_and_rate`?

IamPete1 (commented):
It would be nice to move all the attitude control handling down to the attitude control lib. Maybe a new function that takes motor the spool state. There is currently lots of duplication,

Issue #27729, opened 2024-08-01
EKF3: switching sources between ExtNav and OptFlow leads to position jump if GPS enabled


rmackay9:
If GPS/Non-GPS transitions are setup using External Nav (e.g. Visual Odometry) and Optical flow then transitions from optical flow to External Nav may result in a large position jump to the GPS location even though the GPS is not defined in any of the source sets (e.g. the GPS is not defined in any EK3_SRCx_ parameters)

Below is an image from SITL testing where GCS_SEND_TEXT() debug messages were added to [NavEKF3_core::ResetPosition()](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_NavEKF3/AP_NavEKF3_PosVelFusion.cpp#L74) to determine the position source used for the position reset.
![image](https://github.com/user-attachments/assets/134a0dcf-dab1-41c8-ac56-77bd890de538)

The issue can be reproduced in SITL by doing the following the [SITL instructions for Vicon](https://ardupilot.org/dev/docs/adding_simulated_devices.html#testing-vicon-aka-vision-positioning) and [optical flow](https://ardupilot.org/dev/docs/adding_simulated_devices.html#adding-an-optical-flow-sensor).  To be more specific follow these instructions:

- start SITL
- param load ../Tools/autotest/default_params/[copter-vicon.parm](https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/default_params/copter-vicon.parm)
- param fetch (and repeat above again)
- param load ../Tools/autotest/default_params/[copter-optflow.parm](https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/default_params/copter-optflow.parm)
- param fetch (and repeat above again)
- param set SIM_VICON_YAW 90
- param set EK3_SRC1_POSXY  6
- param set EK3_SRC1_POSZ    6
- param set EK3_SRC1_VELXY   6
- param set EK3_SRC1_VELZ    6
- param set EK3_SRC1_YAW     6
- param set EK3_SRC2_POSXY   0
- param set EK3_SRC2_POSZ    1
- param set EK3_SRC2_VELXY   5
- param set EK3_SRC2_VELZ    0
- param set EK3_SRC2_YAW     0
- param set RC9_OPTION 90
- param set GPS1_TYPE 1
- exit SITL
- `../Tools/autotest/sim_vehicle.py --map --console -A "--serial5=sim:vicon:"`

Fly the vehicle in Guided mode first using External nav

[...]

PR #28693, opened 2024-11-21
AP_NavEKF3: pos and vel resets default to user defined source

rmackay9:
This resolves issue https://github.com/ArduPilot/ardupilot/issues/27729 by using the user defined position or velocity source (e.g. EK3_SRCx_POSXY, EK3_SRCx_VELXY) whenever the caller does not specify the source.

This solution is a minimal-change solution to the problem.  Another possible solution would be to remove ResetVelocity and ResetPosition's xxxResetSoiurce argument completely and instead always just use the user defined source.

This has been lightly tested in SITL to confirm that it resolves the issue.  Below are before and after screen shots of SITL testing where the Vicon has been turned 90degrees from reality.

Before we see terrible behaviour (even worse than described in [the issue](https://github.com/ArduPilot/ardupilot/issues/27729 )) where the lane switches every 7 seconds and the vehicle resets to the GPS location.  This results in the vehicle flying an ever expanding circle around the target at full speed
before-map-and-console

After we see nearly twitch-less transitions between Optical flow and External Nav.  The two vehicles shown are the EKF's estimated position and the actual position and their positions are stable
after-map-and-console

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_PosVelFusion.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_PosVelFusion.cpp
@@ -13,6 +13,26 @@
 // Do not reset vertical velocity using GPS as there is baro alt available to constrain drift
 void NavEKF3_core::ResetVelocity(resetDataSource velResetSource)
 {
+    // if reset source is not specified then use user defined velocity source
+    if (velResetSource == resetDataSource::DEFAULT) {
+        switch (frontend->sources.getVelXYSource()) {
+        case AP_NavEKF_Source::SourceXY::GPS:
+            velResetSource = resetDataSource::GPS;
+            break;
+        case AP_NavEKF_Source::SourceXY::BEACON:
+            velResetSource = resetDataSource::RNGBCN;
+            break;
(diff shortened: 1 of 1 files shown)


priseborough:
TXS @rmackay9

libraries/AP_NavEKF3/AP_NavEKF3_PosVelFusion.cpp
```diff
@@ -73,6 +93,26 @@ void NavEKF3_core::ResetVelocity(resetDataSource velResetSource)
 // resets position states to last GPS measurement or to zero if in constant position mode
 void NavEKF3_core::ResetPosition(resetDataSource posResetSource)
 {
+    // if reset source is not specified thenn use the user defined position source
```
peterbarker, line 96:
```suggestion
    // if reset source is not specified then use the user defined position source
```

tridge (approved):

libraries/AP_NavEKF3/AP_NavEKF3_PosVelFusion.cpp
rmackay9: thanks, I'll try to remember to fix this on my next pass through the code

PR #28695, opened 2024-11-21
Correct documentation on PM and RTC messages

peterbarker:


amilcarlucas (approved):

rmackay9 (approved):
We don't normally put underscores in log message names (see "old_utc") instead we tend to use pascal case (aka camel case).  Still, I think we should merge this if only to get CI passing again.

rmackay9:
I'm including this to help resolve CI failures for 4.6.0-beta2

Issue #28685, opened 2024-11-20
Takeoff not completing under TKOFF_OPTIONS[0]=0


Georacer:
User @menschel reported that
- in a Plane
- without an airspeed sensor -> running under `TKOFF_OPTIONS[0]=0`
- in an AUTO mission with a takeoff item to 50m
the aircraft would climb to 47m and just hang there, not completing the mission item.
![image](https://github.com/user-attachments/assets/2123dc18-1988-45f9-9de6-74d6da6251d9)
The final reach of 50m was accidental.

He was flying a small ZOHD flying wing.
He kindly provided a flight log for us [here](https://github.com/ArduPilot/ardupilot/pull/28352#issuecomment-2482853648).

He attempted to file a fix here: https://github.com/ArduPilot/ardupilot/pull/28352, but what he proposes isn't the solution.

This affects our 4.6.0 beta.

---

While it was unable to climb the last 3m, the airframe was at full throttle and at ~6deg pitch up (3deg nav pitch + 3deg `PTCH_TRIM_DEG`)
Still, that wasn't enough to make the last few meters.

tridge:
we discussed on the EU dev call, and think we should gradually bring in TECS control during the takeoff, with a widening pitch window available to TECS, so in initial part of takeoff the TECS min/max pitch would be completely constrained, then it widens during the takeoff so that at the end of the takeoff TECS has full control

Georacer:
> Just for clarification.
It's a plane with a tail [1](https://www.zohd.net/zohd-drift), [2](https://oscarliang.com/zohd-drift-dark-breeze/) not a flying wing, customized sub 250g build (246g)

Thanks!

Georacer:
This is now addressed by #28792.

PR #28707, opened 2024-11-22
Plane: Fix inability to climb the last few meters of takeoff

Georacer:
## Summary

This PR addresses #28685.

## Details

This affects takeoffs without an airspeed sensor.

### Cause

Some airframes will not climb at 0 pitch, regardless of throttle setting.
Thus, the current takeoff level-off strategy of asymptotically forcing pitch to 0, results in the aircraft to never climb the final few meters.

### Reproduction

The behaviour was reproduced in SITL, by setting:
- `PTCH_TRIM_DEG`: -10.0
- `TKOFF_THR_MAX`: 75.0

### Solution

After the expected time to complete the takeoff has elapsed, drive the desired pitch gradually back to takeoff pitch, over the next 5 seconds.

### Testing

**Before**:
image

**After**:
image

### Alternatives

The proposed solution of widening the TECS pitch limits and letting TECS take over and regulate the last few meters of the climb was not feasible:
During an airspeedless takeoff, TECS uses throttle to correct for altitude errors.
It would use pitch (if it was allowed) to correct for airspeed errors).
Thus letting TECS command the pitch angle would not help.


menschel:
Weather is currently not favorable for a test flight but I'll try to test in case it gets better.

menschel:
Two short flights.
[24-11-23_08-37-52.bin](https://github.com/menschel/ardupilot/blob/stash/build/bin/Arduplane/24-11-23_08-37-52.bin)
[24-11-23_08-42-26.bin](https://github.com/menschel/ardupilot/blob/stash/build/bin/Arduplane/24-11-23_08-42-26.bin)
Fun fact, the overshoot now happens based on the workaround.

IMO the whole thing is connected to a false climb-rate estimation. I have logs from a landing test session with 4.5.7 and I see level-off happening at 3m below target altitude, not 8m.
It may be a good idea to stop limiting the pitch at level-off and see what TECS is doing out of it. If I'm not mistaken, the waypoint is 1km away and TECS should linearly climb.

ArduPlane/takeoff.cpp
```diff
@@ -304,6 +304,14 @@ int16_t Plane::get_takeoff_pitch_min_cd(void)
         // if height-below-target has been initialized then use it to create and apply a scaler to the pitch_min
         if (auto_state.height_below_takeoff_to_level_off_cm != 0) {
             float scalar = remaining_height_to_target_cm / (float)auto_state.height_below_takeoff_to_level_off_cm;
+
+            const int32_t takeoff_end_delay_ms = AP_HAL::millis() - takeoff_state.level_off_start_time_ms - g.takeoff_pitch_limit_reduction_sec*1000;
```
peterbarker, line 308:
```suggestion
            const uint32_t takeoff_end_delay_ms = AP_HAL::millis() - takeoff_state.level_off_start_time_ms - g.takeoff_pitch_limit_reduction_sec*1000;
```

peterbarker (changes requested):

tridge:
@Georacer I'm surprised you didn't implement the expanding pitch window using TECS like we discussed, did it not work out?

Georacer:
> @Georacer I'm surprised you didn't implement the expanding pitch window using TECS like we discussed, did it not work out?

As I wrote in the description, the proposed solution of widening the TECS pitch limits and letting TECS take over and regulate the last few meters of the climb was not feasible:
During an airspeedless takeoff, TECS uses throttle to correct for altitude errors.
It would use pitch (if it was allowed) to correct for airspeed errors; this is not reliable in an airspeedless case, which we are in.
Thus letting TECS command the pitch angle would not help.

I did try it and it didn't give effective results.
If you're really curious I can repeat the experiment.

ArduPlane/takeoff.cpp
```diff
@@ -304,6 +304,14 @@ int16_t Plane::get_takeoff_pitch_min_cd(void)
...
+            if (takeoff_end_delay_ms > 0) {
```
peterbarker, line 309: This might be a `linear_interpolate`?

peterbarker (changes requested):

tridge:
discussed on EU dev call, decided just to declare the takeoff complete if we are in the taper off region and we timeout

menschel:
Can we also do something about the now forced 100% throttle?

I argue that it's a wrong design decision not to fall back to TECS controlled throttle here.
https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/takeoff.cpp#L282

It also adds to the issue here because it influences the climb-rate estimation.

Georacer:
> Can we also do something about the now forced 100% throttle?
> 
> I argue that it's a wrong design decision not to fall back to TECS controlled throttle, means the airspeed sensor presence should not be evaluated here. https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/takeoff.cpp#L282
> 
> It also adds to the issue here because it influences the climb-rate estimation.

For this PR we definitely won't change the throttle strategy.

Having a constant THR_MAX was a request from other developers, so you'd have to make your case on why this is a good idea.
Preferably by opening a new ticket.

menschel:
I've been spinning the numbers, with level off at 42m, distance 8m, climb_rate 4mps, I came up with this calculation that results in the observed behavior.

```
dh = climb_rate * t

climb_rate = (pitch/pitch_max) * climb_max

pitch = pitch_initial * (takeoff_pitch_limit_reduction_sec - t)

dh = (((pitch_initial * (takeoff_pitch_limit_reduction_sec - t))  /pitch_max) * climb_max) * t

dh = (((12 * (2 - t))  /18) * 3) * t

dh = 3t - 2*t²

Integral(dh) = (3/2)t² - (2/3)t³
from 0 to takeoff_pitch_limit_reduction_sec = 2

(3/2)takeoff_pitch_limit_reduction_sec² - (2/3)takeoff_pitch_limit_reduction_sec³

(3/2)*4 - 2/3*8

6-(1/12) = 5.916

--> 47.916m
```

I'm still working on the math but maybe there is some sqrt(g.takeoff_pitch_limit_reduction_sec)) missing, maybe here.

```
float sec_to_target = (remaining_height_to_target_cm * 0.01f * sqrt(g.takeoff_pitch_limit_reduction_sec)) / (-auto_state.sink_rate);
```

menschel:
I have created alternative PR https://github.com/ArduPilot/ardupilot/pull/28776
Mea Culpa, Mea Culpa.

menschel:
I believe I know what it is

`float sec_to_target = (remaining_height_to_target_cm * 0.01f) / (-auto_state.sink_rate);`

the interval i.e. the altitude gained is only half in what that time because of the linear decreasing pitch and consequently climb-rate. 

`float sec_to_target = (remaining_height_to_target_cm * 0.01f * 2) / (-auto_state.sink_rate);`
is the correct time to target.

If I may do so I'll include your commit d698139f0c6dfe3ce965d722b8c73be607de6d44 into my [PR](https://github.com/ArduPilot/ardupilot/pull/28776)

Georacer:
Superseded by #28792

PR #28710, opened 2024-11-22
Add CBUnmanned CBU-H7-LC-Stamp

andyp1per:


Hwurzburg:
@andyp1per if this is just a subset of the regular stamp, why not do the hwdef as based on it with undefs?

andyp1per:
@CBUnmanned basing it on the existing config seems like a reasonable idea - what do you think?

CBUnmanned:
@andyp1per @Hwurzburg Sorry for the slow reply.

The only real thing it shares in common with the other stamp is the mounting footprint. The difference is that it uses a different H7 MCU so having a dedicated hwdef rather than 'undef'ing nearly everything seems appropriate. 

Both Stamps will have there own AP_Periph based firmwares at some point that will use undefs so each being dedicated makes that easier down the line!

andyp1per:
Ok fair enough - seems fine to me

Hwurzburg:
so it seems the current wiki entry is no longer available (https://ardupilot.org/plane/docs/common-StampH743.html) correct? it matches nothing on your website...
I will try to review before the DevCall

CBUnmanned:
@Hwurzburg Yes this is the old information which should be superseded by this wiki PR:
https://github.com/ArduPilot/ardupilot_wiki/pull/6412 that brings the information inline with the products available from my site. (Still for me to make the changes requested by you)

PR #28723, opened 2024-11-24
AntennaTracker: add push command for AntennaTracker servos

eppravitra:
Add push command for AntennaTracker servos

This is necessary because "smart actuators" were not working. 

Specifically, I have a Robotis servo, and the change is required to make it work with AntennaTracker.

My guess is, it was not already there because we were not serious that AntennaTracker output all actuators at the same time. I couldn't completely understand RCOutput.cpp, but it looks like this change should not break anything because there is no cork command for AntennaTracker.

### DIFF
--- a/AntennaTracker/mode_servotest.cpp
+++ b/AntennaTracker/mode_servotest.cpp
@@ -30,8 +30,10 @@ bool ModeServoTest::set_servo(uint8_t servo_num, uint16_t pwm)
         SRV_Channels::constrain_pwm(SRV_Channel::k_tracker_pitch);
     }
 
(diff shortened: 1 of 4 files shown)


peterbarker:
The pattern in other vehicles is to cork then push.  Any reason you didn't follow that pattern in here?

eppravitra:
The changes was tested with Pixhawk6x

eppravitra:
> The pattern in other vehicles is to cork then push. Any reason you didn't follow that pattern in here?

It was simply that it was working without the need to call cork().

Do you want me to call cork() so it matches with other vehicles?

PR #28724, opened 2024-11-24
Support for scripting in flash using littlefs

andyp1per:
This completely replaces the flash based support, including logging with a littlefs backend. This results in much more reliable FS access as well as support for things like scripting.

With thanks to @ntamas for the original code


tpwrules:
Oh excellent!! Have you tried with both NOR and NAND?

andyp1per:
> Oh excellent!! Have you tried with both NOR and NAND?

No, still a lot of work to do as there is lots of duplication currently. The code is from @ntamas but needs a bunch of rework to get into master. I have a plan, but the main thing is that it now works for me locally and I know how to proceed from here.

ntamas:
Let me know if I can help with reworking this PR for inclusion. This is something that we at @skybrush-io would also be interested in.

jpkh:
Oh yes, there are a lot small H743 based FPVDrone FCs that only have dataflash. This will be important update to all those (and others). Thanks Andy.

robustini:
Thanks Andy, this will be most useful!

andyp1per:
@ntamas actually it's not so bad - I think what you have done is correct. I really want the block logger to die, but I don't think that's going to be possible on boards with very small flash chips

tridge:
it will be very interesting to see how this goes! and some stats on storage overhead (and flash cost) would be useful

andyp1per:
Working quite nicely for me now. Not yet flown it.

andyp1per:
Unfortunately I have yet to be convinced that the architecture of littlefs is suitable for logging, there are a lot of unaddressed issues which particularly bit hard.

ntamas:
Are you experiencing performance issues with LittleFS? If so, how about splitting the flash memory into two partitions, one for logging with the old block logger, and one for storing files (scripts etc) that are assumed to be mostly static?

andyp1per:
> Are you experiencing performance issues with LittleFS? If so, how about splitting the flash memory into two partitions, one for logging with the old block logger, and one for storing files (scripts etc) that are assumed to be mostly static?

That's obviously the fallback, but at an even higher flash cost and would require modifying either lfs or the block logger to cope with some kind of partitioning scheme. I'm going to see what is achievable with LFS first before going that route.

andyp1per:
We don't have a block logger based build to check the size against

andyp1per:
Actually this is looking pretty promising

libraries/AP_Logger/AP_Logger_File.cpp
```diff
@@ -980,6 +982,33 @@ void AP_Logger_File::io_timer(void)
         write_fd_semaphore.give();
         return;
     }
+
+#if AP_FILESYSTEM_LITTLEFS_ENABLED
+    // see https://github.com/littlefs-project/littlefs/issues/564#issuecomment-2363032827
```
tridge, line 987: it would be nice for this to be inside the AP_Filesystem LittleFS code somehow, could even be a static function on that class?
or just return short from the write() ?

libraries/AP_HAL_ChibiOS/Storage.cpp
```diff
@@ -42,8 +42,12 @@ extern const AP_HAL::HAL& hal;
 #endif
 
 #ifndef HAL_STORAGE_BACKUP_COUNT
+#if AP_FILESYSTEM_LITTLEFS_ENABLED
+#define HAL_STORAGE_BACKUP_COUNT 5
```
tridge, line 46: better off with 0 for littlefs

MichelleRos:
Finished the size compare for the boards in the PR:
```
Board,copter,plane
FoxeerF405v2,22892,23404
KakuteH7Mini-Nand,26028,23516
KakuteH7v2,25844,23324
MambaH743v4,335292,330580
MatekH7A3,-10828,-15620
```

So, MatekH7A3 reduces in flash cost - looks like it had the full filesystem stuff included even though it doesn't have an SD card?

The other boards increase, but I don't understand why the Mamba is showing such a massive flash usage...

andyp1per:
The block logger costs about 40k, this PR is 15k cheaper. The Mamba increases due to the inclusion of scripting which was originally turned off

jpkh:
@andyp1per some of the MatekH743 (like -SLIM) has SD card. Its included on most of their boards as default. Then again JHEMCU-H743 does not have SD only flash

MichelleRos:
>The block logger costs about 40k, this PR is 15k cheaper. The Mamba increases due to the inclusion of scripting which was originally turned off

All of the boards except for the H7A3 increase, not just the Mamba.

The Foxeer and both Kakutes increase by about 21k for both copter and plane, so looks like littlefs uses about 21k more flash than the block logger.

libraries/AP_Logger/AP_Logger_File.cpp
andyp1per: Fixed

libraries/AP_HAL_ChibiOS/Storage.cpp
andyp1per: Fixed

andyp1per:
Yeah I think 21k is the most accurate estimate of the cost. Some of the others implicitly get things or already have things which make the estimate difficult.

andyp1per:
Yeah I think 21k is the most accurate estimate of the cost. Some of the others implicitly get things or already have things which make the estimate difficult.

andyp1per:
I've added SITL support, but can't get it to work. It seems there are some issues with the filebd provider and littlefs provides no working examples. There is an open issue asking how to make it work. A project for someone else in the future maybe :)

.gitmodules
```diff
@@ -42,3 +42,6 @@
 [submodule "modules/lwip"]
 	path = modules/lwip
 	url = https://github.com/ArduPilot/lwip.git
+[submodule "modules/littlefs"]
+	path = modules/littlefs
+	url = https://github.com/ArduPilot/littlefs.git
```
tpwrules, line 47: Would much prefer to just vendor the two relevant files. Git submodules are always a pain though admittedly one more doesn't significantly increase the level.

libraries/AP_HAL_ChibiOS/hwdef/scripts/defaults_normal.h
```diff
@@ -11,13 +11,18 @@
...
 #endif
 
+#ifndef HAL_OS_LITTLEFS_IO
+#define HAL_OS_LITTLEFS_IO 0
+#endif
+
 #ifndef AP_TERRAIN_AVAILABLE
 // enable terrain only if there's an SD card available:
-#define AP_TERRAIN_AVAILABLE HAL_OS_FATFS_IO
+#define AP_TERRAIN_AVAILABLE (HAL_OS_FATFS_IO || (HAL_OS_LITTLEFS_IO && (BOARD_FLASH_SIZE>1024)))
```
tpwrules, line 25: Is LittleFS big enough for a meaningful amount? Is it okay with the terrain access patterns?

libraries/AP_Logger/AP_Logger.cpp
```diff
@@ -56,6 +56,14 @@ extern const AP_HAL::HAL& hal;
 #define HAL_LOGGER_ARM_PERSIST 15
 #endif
 
+#ifndef HAL_LOGGER_MIN_MB_FREE
+#if AP_FILESYSTEM_LITTLEFS_ENABLED
+#define HAL_LOGGER_MIN_MB_FREE 1
```
tpwrules, line 61: Probably needs to be bigger so that a meaningful amount can be logged. Also to match the parameter range annotation.

libraries/AP_Logger/AP_Logger_File.cpp
```diff
@@ -980,6 +985,11 @@ void AP_Logger_File::io_timer(void)
         write_fd_semaphore.give();
         return;
     }
+
+#if AP_FILESYSTEM_LITTLEFS_ENABLED && CONFIG_HAL_BOARD != HAL_BOARD_SITL
+    bool sync_block = AP::littlefs().sync_block(_write_fd, _write_offset, nbytes);
```
tpwrules, line 990: Could be done later, but it would be nicer to have a `bytes_until_sync` method on every filesystem, coupled with some logic to sync after a minimum number of bytes logged.

I would prefer not to add another singleton.

libraries/AP_Logger/AP_Logger_File.cpp
```diff
@@ -999,16 +1009,20 @@ void AP_Logger_File::io_timer(void)
...
         _write_offset += nwritten;
         _writebuf.advance(nwritten);
+
         /*
-          the best strategy for minimizing corruption on microSD cards
-          seems to be to write in 4k chunks and fsync the file on each
-          chunk, ensuring the directory entry is updated after each
-          write.
+          fsync on littlefs is extremely expensive (20% CPU on an H7) and not
+          required since the whole point of the filesystem is to avoid corruption
```
tpwrules, line 1015: Needs updating/combining.

libraries/AP_Logger/AP_Logger_File.h
```diff
@@ -125,7 +129,11 @@ class AP_Logger_File : public AP_Logger_Backend
     // data and failures-to-boot.
     uint32_t _free_space_last_check_time; // milliseconds
     const uint32_t _free_space_check_interval = 1000UL; // milliseconds
+#if AP_FILESYSTEM_LITTLEFS_ENABLED
+    const uint32_t _free_space_min_avail = 4096; // bytes
```
tpwrules, line 133: Probably need more on NAND.

tpwrules (commented):
Really excited to se this; sorry for not properly testing yet. Some miscellaneous notes.

Is this meant to be mutually exclusive with FATFS support? Could a board have both NAND/NOR and SD?

.gitmodules
andyp1per: Yeah, I hear you but that ship has sailed and the extra files in the repo are actually somewhat useful. The repo is not big.

libraries/AP_Logger/AP_Logger.cpp
andyp1per: Actually, this needs to change so that the oldest file is deleted if there is not enough space, not sure that works right now

andyp1per: Ok the code exists, I have massaged this to be more appropriate

libraries/AP_Logger/AP_Logger_File.cpp
andyp1per: Fixed

libraries/AP_Logger/AP_Logger_File.h
andyp1per: Fixed

libraries/AP_HAL_ChibiOS/hwdef/scripts/defaults_normal.h
andyp1per: a 10km radius is 2.5Mb of data, so perfectly fine on the W25NXX chips

timtuxworth: The problem is that the "Local Area" would need to be flashed in the firmware right? I usually load about 200km2 of STRM30 data around Edmonton when I load terrain on a new AP which works out to about 4G.
How would I do that?

ntamas:
> It seems there are some issues with the filebd provider and littlefs provides no working examples

I figured out how to use the `filebd` provider in another project of mine; I think the trick was to provide a pointer to the `lfs_filebd_t` struct as the context of the filesystem itself. Can't remember if there were any other tricks other than this one, but I can start working on it once this PR is merged.

andyp1per:
> > It seems there are some issues with the filebd provider and littlefs provides no working examples
> 
> I figured out how to use the `filebd` provider in another project of mine; I think the trick was to provide a pointer to the `lfs_filebd_t` struct as the context of the filesystem itself. Can't remember if there were any other tricks other than this one, but I can start working on it once this PR is merged.

Thanks. I have added the sitl build option now, so its all ready to go when someone has some time to get it working

andyp1per:
Definitely something still not right here. Performance slows down over time and there are a lot of reads going on.

andyp1per:
> Definitely something still not right here. Performance slows down over time and there are a lot of reads going on.

Ok think I have resolved this, a combination of excess reads and calling disk_free() the whole time. Need another proper test flight to confirm, but looking really promising now.

andyp1per:
Full flight test was great, no dropped packets in acro with 4KHz fast rate and fast attitude logging. Some dropped in loiter, but that's going to be down to how hard I am pushing the CPU. So I think this is a winner now.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+    WITH_SEMAPHORE(fs_sem);
+
+    lfs_info info;
+    lfs_dir_entry_pair *pair = static_cast<lfs_dir_entry_pair*>(ptr);
+    if (!pair) {
+        errno = EINVAL;
+        return nullptr;
+    }
+
+    if (!lfs_dir_read(&fs, &pair->dir, &info)) {
```
tpwrules, line 344: Needs an error check. For some strange reason this tries to return `LFS_ERR_CORRUPT` when using the clear log button in QGroundControl but it gets stuck forever since the error gets ignored here. Needs further investigation.

tpwrules (changes requested):
Been running some experiments and this looks pretty good! Glad all my random memories of the guts of LittleFS were helpful. A bunch of little things I noted in comments. I can run the aerobatics in sim on hardware, so I can run a script and the script can access files and I can log to the flash in the meantime. I'm logging at like an order of magnitude less rate, haven't tried to turn that up yet.

Terrain is a no-go for this first cut; it needs more studying.

Also encountered a weird issue where the "erase all logs" button in QGC hangs the autopilot forever without watchdogging. It gets stuck in an error loop with an EXPECT_DELAY_MS reading the directory. I marked where we're accidentally discarding the error in the review but there shouldn't be an error in the first place...

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+    fp->cfg.attrs = fp->attrs;
+    fp->cfg.attr_count = 1;
+    fp->attrs[0] = {
+        .type = LFS_ATTR_MTIME,
+        .buffer = &fp->mtime,
+        .size = sizeof(fp->mtime)
+    };
+    // this is a cheat, in theory it could be dynamically allocated but is only used to find the descriptor 
+    // for mtime and we know that's called with a name that doesn't disappear
+    fp->filename = pathname;
```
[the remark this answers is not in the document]
andyp1per: Fair, fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+        printf("Unknown SPI Flash 0x%08x\n", id);
+        return 0;
+    }
+
+    fs_cfg.read_size = page_size;
+    fs_cfg.prog_size = page_size;
+    fs_cfg.block_size = flash_block_size * LFS_FLASH_BLOCKS_PER_BLOCK;
+    fs_cfg.block_count = flash_block_count / LFS_FLASH_BLOCKS_PER_BLOCK;
+#if AP_FILESYSTEM_LITTLEFS_FLASH_TYPE == AP_FILESYSTEM_FLASH_W25NXX
+    fs_cfg.metadata_max = page_size;
```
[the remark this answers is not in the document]
andyp1per: page size is 2k. I tried 8k before and it was slower because of the extra page copies when un-inlining. One option would be to specify unlined for logs. It exposes more flags at the logging layer but might be worth it.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+    fs_cfg.metadata_max = 2048;
+
+    char lfsname[L_tmpnam];
+    uint32_t id = 0;
+    if (std::tmpnam(lfsname)) {
+        lfs_filebd_create(&fs_cfg, lfsname, &fbd_config);
+        id = 0xFAFF;
+    }
+#endif // CONFIG_HAL_BOARD != HAL_BOARD_SITL
+    fs_cfg.block_cycles = 75000;
```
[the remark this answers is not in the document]
andyp1per: I copied this from someone else's issue where wear-levelling was causing a performance slowdown. Right now we do no wear levelling, so this is not making things worse and its probably better because of the allocate on demand of blocks. So the point here *is* to disable wear-levelling. We can experiment with it, but the performance is the gating issue right now.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+    lfs_block_t block, lfs_off_t off, void* buffer, lfs_size_t size
+) {
+    uint32_t address;
+    lfs_off_t read_off = 0;
+
+    if (dead) {
+        return LFS_ERR_IO;
+    }
+
+    address = lfs_block_and_offset_to_raw_flash_address(block, off);
```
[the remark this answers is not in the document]
andyp1per: Maybe, but the current code is correct and tested and I am keen not to make structural changes that might break that.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.h
```diff
@@ -0,0 +1,137 @@
...
+
+    static AP_Filesystem_FlashMemory_LittleFS* singleton;
+
+    int allocate_fd();
+    int free_fd(int fd);
+    void free_all_fds();
+    file_descriptor* lfs_file_from_fd(int fd) const;
+
+    uint32_t lfs_block_and_offset_to_raw_flash_address(lfs_block_t block, lfs_off_t off = 0, lfs_off_t flash_block = 0);
+    uint32_t lfs_block_to_raw_flash_page_index(lfs_block_t block, lfs_off_t flash_block = 0);
```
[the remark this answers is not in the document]
andyp1per: It's private and the name of the function is literally what it does, so leaving it as is.

libraries/AP_HAL_ChibiOS/hwdef/MatekH7A3/hwdef.dat
```diff
@@ -156,12 +160,10 @@ ROMFS_WILDCARD libraries/AP_OSD/fonts/font*.bin
 
 # --------------------- SD & FLASH ----------------------
 SPIDEV sdcard    SPI2 DEVID3 SD_CS      MODE0 400*KHZ 25*MHZ
-SPIDEV dataflash SPI2 DEVID1 FLASH_CS   MODE3 32*MHZ  32*MHZ
-
-define HAL_LOGGING_DATAFLASH_ENABLED 1
-define HAL_OS_FATFS_IO 1
-define HAL_LOGGING_DATAFLASH_DRIVER AP_Logger_W25NXX
+SPIDEV dataflash SPI2 DEVID1 FLASH_CS   MODE3 104*MHZ  104*MHZ
```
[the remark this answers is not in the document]
andyp1per: The chips support this and it makes a bit of a difference. Most boards with this flash use this speed.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+#endif
+
+        // sleep so that othher processes get the CPU cycles that the 4ms erase cycle needs.
+        hal.scheduler->delay(4);
+    }
+
+    return LFS_ERR_OK;
+}
+
+int AP_Filesystem_FlashMemory_LittleFS::_flashmem_sync() {
```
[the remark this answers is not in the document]
andyp1per: This is already done in wait_until_device_is_ready()

libraries/AP_Logger/AP_Logger_config.h
```diff
@@ -48,7 +48,7 @@
 #endif
 
 #ifndef HAL_LOGGER_FILE_CONTENTS_ENABLED
-#define HAL_LOGGER_FILE_CONTENTS_ENABLED HAL_LOGGING_FILESYSTEM_ENABLED
+#define HAL_LOGGER_FILE_CONTENTS_ENABLED HAL_LOGGING_FILESYSTEM_ENABLED && !AP_FILESYSTEM_LITTLEFS_ENABLED
```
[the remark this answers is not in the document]
andyp1per: It's causes a big burst on startup. It's more the performance rather than the size that's the issue.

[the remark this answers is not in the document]
tpwrules: Okay fair enough. The aerobatics script needs updating to check for this method before using it; otherwise it crashes.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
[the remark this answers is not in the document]
tpwrules: Could also reduce the inline file size.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+
+int AP_Filesystem_FlashMemory_LittleFS::free_fd(int fd)
+{
+    file_descriptor* fp = lfs_file_from_fd(fd);
+    if (!fp) {
+        return -1;
+    }
+
+    free(fp);
+    open_files[fd] = fp = nullptr;
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+
+    return 0;
+}
+
+void AP_Filesystem_FlashMemory_LittleFS::free_all_fds()
+{
+    int fd;
+
+    for (fd = 0; fd < MAX_OPEN_FILES; fd++) {
+        if (open_files[fd] == nullptr) {
```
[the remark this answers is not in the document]
andyp1per: Good catch! Fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+    }
+
+    if (flags & O_APPEND) {
+        outflags |= LFS_O_APPEND;
+    }
+
+    return outflags;
+}
+
+// get_singleton for scripting
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.h
```diff
@@ -0,0 +1,137 @@
...
+    int _flashmem_prog(lfs_block_t block, lfs_off_t off, const void* buffer, lfs_size_t size);
+    int _flashmem_erase(lfs_block_t block);
+    int _flashmem_sync();
+
+    // get_singleton for scripting
+    static AP_Filesystem_FlashMemory_LittleFS *get_singleton(void);
+
+private:
+    // JEDEC ID of the flash memory, JEDEC_ID_UNKNOWN if not known or not supported
+    uint32_t jedec_id;
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+
+    char lfsname[L_tmpnam];
+    uint32_t id = 0;
+    if (std::tmpnam(lfsname)) {
+        lfs_filebd_create(&fs_cfg, lfsname, &fbd_config);
+        id = 0xFAFF;
+    }
+#endif // CONFIG_HAL_BOARD != HAL_BOARD_SITL
+    fs_cfg.block_cycles = 75000;
+    fs_cfg.lookahead_size = fs_cfg.block_count/8;
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+    // disable write protection
+    write_status_register(W25NXX_PROT_REG, 0);
+
+    // enable ECC and buffer mode
+    write_status_register(W25NXX_CONF_REG, W25NXX_CONFIG_ECC_ENABLE | W25NXX_CONFIG_BUFFER_READ_MODE);
+#else
+    if (use_32bit_address) {
+        WITH_SEMAPHORE(dev_sem);
+
+        const uint8_t cmd = 0xB7;
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -943,34 +950,45 @@ def write_mcu_config(self, f):
...
             f.write('#define HAL_SDCARD_SPI_HOOK TRUE\n')
             self.build_flags.append('USE_FATFS=yes')
             self.env_vars['WITH_FATFS'] = "1"
+        elif self.has_dataflash_spi():
+            f.write('// Dataflash memory via SPI available, enable POSIX filesystem support\n')
+            f.write('#define USE_POSIX\n')
+            f.write('#define USE_POSIX_LITTLEFS\n\n')
+            f.write('#define HAL_USE_SDC FALSE\n')
+            self.build_flags.append('USE_FATFS=no')
+            self.env_vars['WITH_LITTLEFS'] = "1"
```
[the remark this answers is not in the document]
andyp1per: Good catch! Fixed now

wscript
```diff
@@ -303,6 +303,11 @@ submodules at specific revisions.
     g.add_option('--enable-dronecan-tests', action='store_true',
                  default=False,
                  help="Enables DroneCAN tests in sitl")
+
+    g.add_option('--sitl-littlefs', action='store_true',
+                 default=False,
+                 help="Enable littlefs for filesystem accesson SITL")
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+        }
+
+        WITH_SEMAPHORE(dev_sem);
+#if AP_FILESYSTEM_LITTLEFS_FLASH_TYPE == AP_FILESYSTEM_FLASH_W25NXX
+        send_command_addr(JEDEC_BLOCK_ERASE, lfs_block_to_raw_flash_page_index(block, fblock));
+#else
+        send_command_addr(JEDEC_SECTOR4_ERASE, lfs_block_and_offset_to_raw_flash_address(block, 0, fblock));
+#endif
+
+        // sleep so that othher processes get the CPU cycles that the 4ms erase cycle needs.
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.h
```diff
@@ -0,0 +1,137 @@
...
+    // format flash.  This is async, monitor get_format_status for progress
+    bool format(void) override;
+    AP_Filesystem_Backend::FormatStatus get_format_status() const override;
+
+    int _flashmem_read(lfs_block_t block, lfs_off_t off, void* buffer, lfs_size_t size);
+    int _flashmem_prog(lfs_block_t block, lfs_off_t off, const void* buffer, lfs_size_t size);
+    int _flashmem_erase(lfs_block_t block);
+    int _flashmem_sync();
+
+    // get_singleton for scripting
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
andyp1per: Good catch. Fixed

tpwrules:
The issue with QGC is a bug in LittleFS: https://github.com/littlefs-project/littlefs/issues/1061 It's easily replicable by running `log erase ; log list` with mavproxy assuming there's at least 3 or 4 logs in memory. The problem is that the main thread races to count the number of logs using readdir while the I/O thread works on deleting them. This is okay according to POSIX but LittleFS's readdir apparently explodes under that type of concurrent modification.

I don't know if it's actually causing real corruption or damaging LittleFS's internal state. The flight controller won't hang forever now that the readdir result is checked properly but the error will be ignored. We need to confirm there's no real issue or fix the race before I'd be okay merging this.

libraries/AP_HAL_ChibiOS/hwdef/scripts/defaults_normal.h
andyp1per: @tpwrules where is this happening - I don't understand where this is happening in the terrain library? mavftp for upload is fine.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+
+#define W25NXX_CONFIG_ECC_ENABLE         (1 << 4)
+#define W25NXX_CONFIG_BUFFER_READ_MODE   (1 << 3)
+
+#define W25NXX_TIMEOUT_PAGE_READ_US        60   // tREmax = 60us (ECC enabled)
+#define W25NXX_TIMEOUT_PAGE_PROGRAM_US     700  // tPPmax = 700us
+#define W25NXX_TIMEOUT_BLOCK_ERASE_MS      10   // tBEmax = 10ms
+#define W25NXX_TIMEOUT_RESET_MS            500  // tRSTmax = 500ms
+
+bool AP_Filesystem_FlashMemory_LittleFS::sync_block(int _write_fd, uint32_t _write_offset, uint32_t& nbytes)
```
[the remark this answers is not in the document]
andyp1per: ok

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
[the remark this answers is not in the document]
andyp1per: Default is 4k (block size) on NOR chips, so 2k doesn't seem massively problematic. Compaction threshold only affects explicit GC which we do at startup, so I am inclined to set this to 88% of the metadata_max.

libraries/AP_Logger/AP_Logger_File.h
```diff
@@ -124,8 +128,16 @@ class AP_Logger_File : public AP_Logger_Backend
...
+#if AP_FILESYSTEM_LITTLEFS_ENABLED
+#if AP_FILESYSTEM_LITTLEFS_FLASH_TYPE == AP_FILESYSTEM_FLASH_W25NXX
+    const uint32_t _free_space_min_avail = 1024 * 1024; // bytes
+#else
+    const uint32_t _free_space_min_avail = 1024 * 256; // bytes
+#endif
+#else
     const uint32_t _free_space_min_avail = 8388608; // bytes
+#endif
+    const uint32_t _free_space_check_interval = 1000UL; // milliseconds
```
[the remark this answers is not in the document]
andyp1per: Fixed thanks

libraries/AP_Scripting/AP_Scripting_config.h
```diff
@@ -10,7 +10,7 @@
 #if AP_SCRIPTING_ENABLED
     #include <AP_Filesystem/AP_Filesystem_config.h>
     // enumerate all of the possible places we can read a script from.
-    #if !AP_FILESYSTEM_POSIX_ENABLED && !AP_FILESYSTEM_FATFS_ENABLED && !AP_FILESYSTEM_ESP32_ENABLED && !AP_FILESYSTEM_ROMFS_ENABLED
+    #if !AP_FILESYSTEM_POSIX_ENABLED && !AP_FILESYSTEM_FATFS_ENABLED && !AP_FILESYSTEM_ESP32_ENABLED && !AP_FILESYSTEM_ROMFS_ENABLED && !AP_FILESYSTEM_LITTLEFS_ENABLED
```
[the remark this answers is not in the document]
andyp1per: Thanks, fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1227 @@
...
+    }
+    return LFS_ERR_OK;
+}
+
+int AP_Filesystem_FlashMemory_LittleFS::_flashmem_erase(lfs_block_t block) {
+    if (dead) {
+        return LFS_ERR_IO;
+    }
+
+    EXPECT_DELAY_MS(2 * LFS_FLASH_BLOCKS_PER_BLOCK);
```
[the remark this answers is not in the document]
andyp1per: Thanks, removed

libraries/AP_HAL_ChibiOS/hwdef/scripts/defaults_normal.h
tpwrules: Open for write here: https://github.com/ArduPilot/ardupilot/blob/bdacfdc76d72820a97e2392e242dbae0b4c49211/libraries/AP_Terrain/TerrainIO.cpp#L203

Failed seek here: https://github.com/ArduPilot/ardupilot/blob/bdacfdc76d72820a97e2392e242dbae0b4c49211/libraries/AP_Terrain/TerrainIO.cpp#L243

Retry every 5 seconds here: https://github.com/ArduPilot/ardupilot/blob/bdacfdc76d72820a97e2392e242dbae0b4c49211/libraries/AP_Terrain/TerrainIO.cpp#L343

andyp1per: This is only true if there is no terrain data correct? But I suppose letting the GCS update the terrain data is the most convenient way and that will currently fail with the current design.

tpwrules: Correct. There are several possible fixes but I'd prefer to just turn it off and come back to it later.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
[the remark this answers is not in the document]
andyp1per: I changed to 4k and the performance seems ok

libraries/AP_HAL_ChibiOS/hwdef/scripts/defaults_normal.h
andyp1per: Ok, think I have implemented a fix. Please check this works for you.

andyp1per:
> The issue with QGC is a bug in LittleFS: [littlefs-project/littlefs#1061](https://github.com/littlefs-project/littlefs/issues/1061) It's easily replicable by running `log erase ; log list` with mavproxy assuming there's at least 3 or 4 logs in memory. The problem is that the main thread races to count the number of logs using readdir while the I/O thread works on deleting them. This is okay according to POSIX but LittleFS's readdir apparently explodes and returns `LFS_ERR_CORRUPT` under that type of concurrent modification.
> 
> I don't know if it's actually causing real corruption or damaging LittleFS's internal state. The flight controller won't hang forever now that the readdir result is checked properly but the error will be ignored. We need to confirm there's no real issue or fix the race before I'd be okay merging this.

We have a file system semaphore, so each access should be atomic, feels like we are missing a lock somewhere. Any ideas?

tpwrules:
> We have a file system semaphore, so each access should be atomic, feels like we are missing a lock somewhere. Any ideas?

We have a lock and it's working properly; the problem is deeper in LittleFS. It happens because we `lfs_dir_open` a particular directory, then start modifying it (by deleting files). After the modifications, `lfs_dir_read` can return `LFS_ERR_CORRUPT`. LittleFS has some mitigations for this case but they are evidently not complete. It doesn't matter whether the directory access and modification are in the same thread or not.

I think we could mitigate the problem by having an unsigned int "directory epoch" variable. Increment it on each write operation (open, close, sync, mkdir, unlink), store its current value in the `lfs_dir_entry_pair` (please change the name, it is not an lfs object!), then close the directory and don't call into LittleFS if the current value doesn't match the value when the directory is opened inside our `readdir`.

Would you like me to implement that on top of this PR?

tpwrules:
We should upgrade to the latest littlefs before merge too. It fixes some issues with slashes in paths that I was running into in testing, and also problems with seeking which might bite terrain data.

andyp1per:
> > We have a file system semaphore, so each access should be atomic, feels like we are missing a lock somewhere. Any ideas?
> 
> We have a lock and it's working properly; the problem is deeper in LittleFS. It happens because we `lfs_dir_open` a particular directory, then start modifying it (by deleting files). After the modifications, `lfs_dir_read` can return `LFS_ERR_CORRUPT`. LittleFS has some mitigations for this case but they are evidently not complete. It doesn't matter whether the directory access and modification are in the same thread or not.
> 
> I think we could mitigate the problem by having an unsigned int "directory epoch" variable. Increment it on each write operation (open, close, sync, mkdir, unlink), store its current value in the `lfs_dir_entry_pair` (please change the name, it is not an lfs object!), then close the directory and don't call into LittleFS if the current value doesn't match the value when the directory is opened inside our `readdir`.
> 
> Would you like me to implement that on top of this PR?

Ok, that's pretty lame - so what you are saying is while a directory is open you can't mess with it. We only ever do opendir/readdir/closedir - so as long as unlink returns an error while a directory is open that seems ok. Presumably it is only a problem if you are modifying files inside the directory, not in other directories? Did you try taking out the lock across the whole opendir/readdir/closedir that would certainly fix it, and we don't appear to be doing much when iterating through all the directories.

tpwrules:
> Ok, that's pretty lame - so what you are saying is while a directory is open you can't mess with it.

Correct. Not exactly sure what constitutes messing with it. Unrelated accesses might break a directory access if there is an internal GC?

> We only ever do opendir/readdir/closedir - so as long as unlink returns an error while a directory is open that seems ok.

The Lua engine opens files for read within the readdir loop.

> Presumably it is only a problem if you are modifying files inside the directory, not in other directories? Did you try taking out the lock across the whole opendir/readdir/closedir that would certainly fix it, and we don't appear to be doing much when iterating through all the directories.

I can try this, it would only allow one directory traversal to be going on at a time. I think it would be simpler than my epoch idea. Recursive semaphores should handle the Lua case, though we have to be mindful that things will break if we introduce other accesses into a readdir loop.

andyp1per:
ok @tpwrules I did the locking and it seems to work fine. Please check. Think this is good to go now.

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -853,6 +853,7 @@ int lua_dirlist(lua_State *L) {
     int i;
     const char *path = luaL_checkstring(L, 1);
     
+    WITH_SEMAPHORE(AP::FS().get_semaphore(path));
     /* open directory */
     auto dir = AP::FS().opendir(path);
```
peterbarker, line 858: `WITH_DIRHANDLE(....)`

that's a random idea to avoid the WITH_SEMAPHORE and opendir being intrisically linked (oir you get readdir failing randomly)

tpwrules:
Please consider this alternate approach for locking: https://github.com/ArduPilot/ardupilot/compare/cc679cc1b779...tpwrules:ardupilot:xpr/littlefs-alt

It's two code lines and avoids touching the rest of the code. It should be simple to remove later if and when the underlying issue is fixed as well.

Did some more testing and it all still looks good. I can't crash the board and scripting restart in flight works properly.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1259 @@
...
+    }
+
+
+    return fd;
+}
+
+int AP_Filesystem_FlashMemory_LittleFS::close(int fileno)
+{
+    FileDescriptor* fp;
+    int retval;
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1259 @@
...
+
+    // populate the file config with the mtime attribute, will get written out on first sync or close
+    fp->cfg.attrs = fp->attrs;
+    fp->cfg.attr_count = 1;
+    fp->attrs[0] = {
+        .type = LFS_ATTR_MTIME,
+        .buffer = &fp->mtime,
+        .size = sizeof(fp->mtime)
+    };
+    fp->filename = strdup(pathname);
```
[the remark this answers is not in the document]
andyp1per: Fixed thanks

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1259 @@
...
+    if (!pair) {
+        errno = EINVAL;
+        return nullptr;
+    }
+
+    int retval = lfs_dir_read(&fs, &pair->dir, &info);
+    if (retval == 0) {
+        /* no more entries */
+        return nullptr;
+    } else if (retval < 0) {
```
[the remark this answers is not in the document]
andyp1per: Removed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1259 @@
...
+
+        /* try mounting again */
+        if (lfs_mount(&fs, &fs_cfg) < 0) {
+            /* cannot mount after formatting */
+            mark_dead();
+            return false;
+        }
+    }
+
+#ifdef HAL_BOARD_STORAGE_DIRECTORY
```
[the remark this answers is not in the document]
andyp1per: ok

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1259 @@
...
+            return LFS_ERR_IO;
+        }
+
+        WITH_SEMAPHORE(dev_sem);
+
+#if AP_FILESYSTEM_LITTLEFS_FLASH_TYPE == AP_FILESYSTEM_FLASH_W25NXX
+        send_command_addr(JEDEC_BLOCK_ERASE, lfs_block_to_raw_flash_page_index(block, fblock));
+#else
+        send_command_addr(JEDEC_SECTOR4_ERASE, lfs_block_and_offset_to_raw_flash_address(block, 0, fblock));
+#endif
```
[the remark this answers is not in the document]
andyp1per: This is intended to replace the JEDEC block logger which also pulls this trick. The Nand version can initially be removed after this goes. Long-term we should refactor Sid's generic JEDEC driver to be used here, but that was a much more complicated change

libraries/AP_HAL_ChibiOS/hwdef/common/malloc.c
```diff
@@ -585,7 +585,7 @@ void __wrap__free_r(void *rptr, void *ptr)
     return free(ptr);
 }
 
-#ifdef USE_POSIX
+#if HAL_OS_POSIX_IO
```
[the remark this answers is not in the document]
andyp1per: Fixed thanks

libraries/AP_Scripting/lua_bindings.cpp
andyp1per: Went with @tpwrules suggestion

Tools/ardupilotwaf/littlefs.py
```diff
@@ -0,0 +1,25 @@
...
+    ])
+
+
+@conf
+def littlefs(bld, **kw):
+    kw.update(
+        name='littlefs',
+        source=['modules/littlefs/lfs.c', 'modules/littlefs/lfs_util.c', 'modules/littlefs/bd/lfs_filebd.c'],
+        target='littlefs',
+        defines=['LFS_NO_DEBUG'],
```
tpwrules, line 22:
```suggestion
        defines=['LFS_NO_DEBUG', 'LFS_NO_WARN', 'LFS_NO_ERROR', 'LFS_NO_ASSERT'],
```

Saves about 1K and disabling asserts is mandatory to avoid crashing the autopilot.

Future work: define reasonable equivalents for these functions.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -1780,6 +1804,32 @@ def write_AIRSPEED_config(self, f):
...
+        seen = set()
+        for dev in self.dataflash_list:
+            if not self.has_dataflash_spi():
+                self.error("Missing DATAFLASH device: %s" % self.seen_str(dev))
+            if self.seen_str(dev) in seen:
+                self.error("Duplicate DATAFLASH: %s" % self.seen_str(dev))
+            seen.add(self.seen_str(dev))
+            a = dev[0].split(':')
+            if a[0].startswith('block'):
+                if len(a[1]) > 1 and a[1].startswith('w25nxx'):
```
tpwrules, line 1820:
```suggestion
                if len(a) > 1 and a[1].startswith('w25nxx'):
```

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -1780,6 +1804,32 @@ def write_AIRSPEED_config(self, f):
         if len(devlist) > 0:
             f.write('#define HAL_AIRSPEED_PROBE_LIST %s\n\n' % ';'.join(devlist))
 
+    def write_DATAFLASH_config(self, f):
+        '''write dataflash config defines'''
+        # DATAFLASH block|littlefs:<w25nxx>
```
tpwrules, line 1809: Do you plan to add the other flash types for block logging and migrate everything over fully before merging? Or keep the old defines for the NOR flashes?

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1254 @@
...
+    }
+#endif
+
+    return true;
+}
+
+#ifdef AP_LFS_DEBUG
+static uint32_t block_writes;
+static uint32_t last_write_msg_ms;
+uint32_t page_reads;
```
tpwrules, line 1013:
```suggestion
static uint32_t page_reads;
```

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1254 @@
...
+}
+
+void AP_Filesystem_FlashMemory_LittleFS::unmount(void)
+{
+    WITH_SEMAPHORE(fs_sem);
+
+    if (mounted && !dead) {
+        if (lfs_unmount(&fs) >= 0) {
+            mounted = false;
+        }
```
tpwrules, line 456:
```suggestion
        lfs_unmount(&fs);
        mounted = false;
```

Not sure what to do with the error code, but the unmount can't fail. The filesystem is unmounted after the function is called no matter the error returned.

tpwrules: Should we call `free_all_fds()`? Not exactly sure how this works under fatfs, but it seems dangerous either way.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1254 @@
...
+    }
+#endif // CONFIG_HAL_BOARD != HAL_BOARD_SITL
+    fs_cfg.block_cycles = 75000;
+    // cache entire flash state in RAM (8 blocks = 1 byte of storage) to
+    // avoid scanning while logging
+    fs_cfg.lookahead_size = fs_cfg.block_count/8;
+    // non-inlined files require their own block, but must be copie. Generally we have requirements for tiny files
+    // (scripting) and very large files (e.g. logging), but not much in-between. Setting the cache size will also
+    // limit the inline size.
+    fs_cfg.cache_size = 2*fs_cfg.prog_size;
```
tpwrules, line 820:
```suggestion
    fs_cfg.cache_size = fs_cfg.prog_size;
```

Offers >50% better read performance and equivalent write performance in my testing.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -0,0 +1,1254 @@
...
+    if (ret == LFS_ERR_OK) {
+        ret = lfs_mount(&fs, &fs_cfg);
+    }
+
+#ifdef HAL_BOARD_STORAGE_DIRECTORY
+    // try to create the root storage folder. Ignore the error code in case
+    // the filesystem is corrupted or it already exists.
+    if (strlen(HAL_BOARD_STORAGE_DIRECTORY) > 0) {
+        lfs_mkdir(&fs, HAL_BOARD_STORAGE_DIRECTORY);
+    }
```
tpwrules, line 937:
```suggestion
    if (ret == LFS_ERR_OK && strlen(HAL_BOARD_STORAGE_DIRECTORY) > 0) {
        ret = lfs_mkdir(&fs, HAL_BOARD_STORAGE_DIRECTORY);
    }
```

tpwrules (changes requested):
A few more super minor things and I think this is ready!

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
andyp1per: Plan to keep the old defines for now, but eventually I would like to move stuff over.

Tools/ardupilotwaf/littlefs.py
andyp1per: Thanks, fixed

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -1780,6 +1804,32 @@ def write_AIRSPEED_config(self, f):
...
+            if a[0].startswith('block'):
+                if len(a[1]) > 1 and a[1].startswith('w25nxx'):
+                    f.write('#define HAL_LOGGING_DATAFLASH_DRIVER AP_Logger_W25NXX')
+                f.write('#define HAL_LOGGING_DATAFLASH_ENABLED TRUE')
+            elif a[0].startswith('littlefs'):
+                f.write('#define USE_POSIX\n')
+                f.write('#define HAL_USE_SDC FALSE\n')
+                f.write('#define HAL_OS_LITTLEFS_IO TRUE\n')
+                f.write('#define HAL_OS_POSIX_IO TRUE\n')
+                if len(a[1]) > 1 and a[1].startswith('w25nxx'):
```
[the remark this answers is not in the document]
andyp1per: Thanks, fixed

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
andyp1per: Thanks, fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
andyp1per: Thanks, fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
andyp1per: Yes. Thanks, fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
andyp1per: Thanks, fixed

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
andyp1per: Thanks, fixed

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -1780,6 +1804,32 @@ def write_AIRSPEED_config(self, f):
...
+                self.error("Duplicate DATAFLASH: %s" % self.seen_str(dev))
+            seen.add(self.seen_str(dev))
+            a = dev[0].split(':')
+            if a[0].startswith('block'):
+                if len(a[1]) > 1 and a[1].startswith('w25nxx'):
+                    f.write('#define HAL_LOGGING_DATAFLASH_DRIVER AP_Logger_W25NXX')
+                f.write('#define HAL_LOGGING_DATAFLASH_ENABLED TRUE')
+            elif a[0].startswith('littlefs'):
+                f.write('#define USE_POSIX\n')
+                f.write('#define HAL_USE_SDC FALSE\n')
```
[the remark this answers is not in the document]
andyp1per: Thanks, fixed

tpwrules (approved):
Looks great, excited to see this go in!

andyp1per:
Flight tested ok

tridge:
@andyp1per was this resolved?
image

tridge:
I think we need at least one of the boards that uses littlefs to be tested in CI, so we can see regressions. Maybe add one to the size compare table?

tridge:
@peterbarker is doing a full build of all targets, results soon!

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -224,3 +224,15 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
 #endif // HAL_LOGGING_ENABLED
 #endif // HAL_UART_STATS_ENABLED
+
+const extern AP_HAL::HAL& hal;
+
+extern "C" {
+    void hal_console_printf(const char *, ...);
+}
+void hal_console_printf(const char *fmt, ...) {
+    va_list ap;
```
tridge, line 234: we already do this with __wrap_printf(), why another copy?

tridge (changes requested):

peterbarker:
image

Needs fixing?

peterbarker:
```
Aocoda-RC-H743Dual,,21764,*,27228,27228,29492,22884,26924
```
Why the difference between Plane and Rover (third-from-last, last)

libraries/AP_HAL/UARTDriver.cpp
tpwrules: This is indeed unnecessary, I tested and bare `printf` works fine in the LittleFS code because of that hook. Could revert commit a39ad0681094c2f2923aac0108be8bf0d4eef1e3.

tpwrules:
> image
> 
> Needs fixing?

https://github.com/ArduPilot/ardupilot/blob/8e00a1e3a6c93e9fccc36f00f2ed4f568a6287e2/libraries/AP_Logger/AP_Logger_File.cpp#L997-L1002 Due to this clause being new even when LittleFS isn't enabled.



> ```
> Aocoda-RC-H743Dual,,21764,*,27228,27228,29492,22884,26924
> ```
> 
> Why the difference between Plane and Rover (third-from-last, last)

Guessing this is due to the extent of terrain data support. It's enabled by default on 2M boards with LittleFS enabled. Rover includes only part of the library and I don't think it can navigate with terrain.

libraries/AP_HAL/UARTDriver.cpp
andyp1per: Dropped

andyp1per:
> I think we need at least one of the boards that uses littlefs to be tested in CI, so we can see regressions. Maybe add one to the size compare table?

Added MambaH743v4

andyp1per:
> Would be nice for users to be able to customise their mavlink message rates by writing files into the littlefs; see definition of `HAL_MAVLINK_INTERVALS_FROM_FILES_ENABLED` (which is kind of crap at the moment...)

Added

andyp1per:
> ```
> Aocoda-RC-H743Dual,,21764,*,27228,27228,29492,22884,26924
> ```
> 
> Why the difference between Plane and Rover (third-from-last, last)

Why are they different sizes? It will be a combination of scripting and terrain and other bits and bobs all of which were off before

peterbarker:
Love this.  Didn't see anything on quick looks through!

tridge (approved):

andyp1per:
yay!

hnz101:
This option isn't selectable from the web build system or am I missing something?
(I mean https://custom.ardupilot.org/ )

Issue #28534, opened 2024-11-06
AP_IOMCU: status read error counter never resets


RickReeser:
## Bug report
Copter 4.5.6
If I leave my vehicle turned on for a long time, I inevitably encounter this prearm error:
```
AP: PreArm: Internal errors 0x1000 1:469 iomcu_reset error
```
Which points to this line: https://github.com/ArduPilot/ardupilot/blob/Copter-4.5/libraries/AP_IOMCU/AP_IOMCU.cpp#L469

An overnight disarmed log of IOMCU errors looks like this:
![image](https://github.com/user-attachments/assets/6ee96eb8-c7bb-4f2d-89e2-d49197c0edfe)

So I guess the status read errors are accruing until they hit the limit of 20, triggering the prearm error. The comments in [the code](https://github.com/ArduPilot/ardupilot/blob/Copter-4.5/libraries/AP_IOMCU/AP_IOMCU.cpp#L473-L476) indicate that they are supposed to reset upon a successful read:
```
    if (read_status_ok == 0) {
        // reset error count on first good read
        read_status_errors = 0;
    }
```
This never happens because `read_status_ok` is never 0. I guess this line is supposed to be `read_status_ok > 0`? This would cause the errors to reset (maybe the conditional is not needed at all since we return early if the read is not successful).

However, this would cause another issue. `read_status_errors` is [used elsewhere](https://github.com/ArduPilot/ardupilot/blob/Copter-4.5/libraries/AP_IOMCU/AP_IOMCU.cpp#L1138):
```
bool AP_IOMCU::healthy(void)
{
    return crc_is_ok && protocol_fail_count == 0 && !detected_io_reset && read_status_errors < read_status_ok/128U;
}
```
Resetting the read errors will make the last check useless, because `read_status_ok` is always incrementing and will be some arbitrarily large number. It should be compared to the total read errors ever seen, so I guess we should track the total errors and errors-since-last-success separately.

This check was added by [this commit](https://github.com/ArduPilot/ardupilot/commit/ab0768871b6dc636be6688a81d49d97b9a957e0f).

**Version**
Copter 4.5.6

**Platform**
[ x ] All

peterbarker:
@RickReeser did you want to put together a PR with your proposal?

RickReeser:
> @RickReeser did you want to put together a PR with your proposal?
https://github.com/ArduPilot/ardupilot/pull/28725

[...]

Issue #30, opened 2013-01-17
Rover should be able to learn both speed and position when learning waypoints


tridge:
Should add LEARN_WP_SPEED boolean option

gmorph:
I believe this feature means when learning waypoints we also want to record the speed.  Not sure why we would record the position as this will already be recorded by the learning of the waypoint.  Assigning to myself for actioning.

rmackay9:
is it OK if I close this?  Rover-3.2 (and higher) allows learning the cruise speed and cruise throttle with the ch7 switch.  Its also documented on the wiki.

Actually i think I'll close this so I dont forget but we can re-open if people object.

Issue #28, opened 2013-01-17
add LEARNING mode to mavproxy and mission planner


tridge:
The ground stations need to understand LEARNING mode (mode 2). Currently mavproxy displays it as STABILIZE mode.

PR #28725, opened 2024-11-24
AP_IOMCU: Fix status read error counter not resetting

RickReeser:
See issue: https://github.com/ArduPilot/ardupilot/issues/28534


andyp1per (approved):

peterbarker (approved):
Log message documentation hints that read_status_errors should be being zeroed - but it's being used elsewhere for other things.

libraries/AP_IOMCU/AP_IOMCU.cpp
```diff
@@ -486,16 +486,16 @@ void AP_IOMCU::read_status()
...
             INTERNAL_ERROR(AP_InternalError::error_t::iomcu_reset);
         }
         return;
     }
-    if (read_status_ok == 0) {
-        // reset error count on first good read
-        read_status_errors = 0;
-    }
+
+    // reset recent error count on first good read
```
peterbarker, line 497:
```suggestion
```

tridge:
rebuilt IO firmware, rebased

tridge:
I'm fine with this PR once it is tested

peterbarker:
I flew this on a sparknavi-blue today without incident.

tridge:

<details><summary>Previous review (2026-09-06)</summary>

Reviewed at head `3adb56b27d`. Full report, including what was checked and cleared: https://uav.tridgell.net/DevCallReviews/2026_09_06/devcall_pr_reviews.html#pr28725

**COMMENT — no blockers.** The split into a cumulative `read_status_errors` and a consecutive `read_status_errors_since_last_success` is exactly what issue #28534 proposed, and it correctly fixes the reported symptom. Two consequences are worth a deliberate decision rather than a silent one, and one thing @peterbarker already asked for is still outstanding.

---

**ISSUE — deleting the first-good-read reset makes `healthy()` permanently stricter for anything seen before the first successful status read.**

The old code zeroed the cumulative counter once, at the first good read; removing that block means `read_status_errors` is now never reset. It still feeds `AP_IOMCU.cpp:1188` (`read_status_errors < read_status_ok/128U`), which is the only thing behind the "IOMCU is unhealthy" prearm failure at `AP_Arming.cpp:1093`.

Because the division is integer, `E` pre-first-success errors need `S >= 128(E+1)` successes, and status reads run at 20 Hz (`AP_IOMCU.cpp:328-331`). That is 6.4 s of unhealthy at `E=0` — which is pre-existing — **plus a further 6.4 s per startup error**: 12.8 s at `E=1`, 38.4 s at `E=5`.

I could not establish that `E` is ever non-zero on real hardware: `check_crc()` runs synchronously in `init()` before the thread is created, and `read_status()` is only reached once `IOEVENT_INIT` has succeeded, so `E` is plausibly 0 or 1. But the deleted block's own comment and the `last_iocmu_timestamp_ms != 0` guard both show startup failures were expected by whoever wrote it. Keeping `if (read_status_ok == 0) { read_status_errors = 0; }` alongside the new reset costs nothing; dropping it deliberately is also a defensible reading of the issue ("compared to the total read errors ever seen"). Worth saying which you meant.

**ISSUE — moving RSErr to the consecutive counter makes isolated errors invisible in the log, because the log runs 20× slower than the counter is cleared.**

`write_log()` is rate-limited to 1 Hz, `read_status()` runs at 20 Hz, and the new counter is zeroed on *every* successful read. So an isolated failure is only recorded if it lands in the one cycle per second that gets logged. The counter it replaced was monotonic, so 1 Hz sampling still caught every error eventually.

That matters because the cumulative `read_status_errors` is now logged nowhere (at this head it appears only at `:492`, `:1188` and in the header) — and it is the exact value that decides `healthy()`. Diagnosing "IOMCU is unhealthy" from a log gets materially harder; not impossible, since `Nerr` still logs protocol failures. Logging both is one line, and keeps the field documentation honest either way.

**NOTE — the new comment at `AP_IOMCU.cpp:501` is stale, and this is a re-raise, not a new point.** It says "reset recent error count on first good read", but the reset now runs on *every* good read. @peterbarker left an empty ```suggestion``` block on that line on 2025-01-29; GitHub still reports it unresolved and not outdated against this head.

**NOTE — two merge-time tidy-ups.** The first commit's subject ends `(#1)`, a leftover from your fork that would read as ArduPilot PR #1 in history. And the description links #28534 without a `Fixes` keyword, so the issue will not auto-close.

---

The 17 rebuilt IO firmware images were checked rather than assumed, since a 7-line fix touching 19 files invites suspicion. They are corrective: each keeps its exact byte length and differs in only 4 bytes (10 for CubeRed), the differing bytes are real Thumb code (`subs r3,r5,#2`/`cmp r3,#30` → `subs r3,r5,#4`/`cmp r3,#28`), and building `--board iomcu` from *current master* source reproduces the PR's variant and not master's committed one. Master's checked-in binaries were stale with respect to master's own source. All 17 padded CRCs change, so affected boards re-flash their IOMCU once on upgrade — worth a release-note line.

Also verified as fine: the `== 20` trip is safe and is not redundant against `thread_main`'s `last_reg_access_ms > 1000` check (that timestamp is refreshed by any successful register access, so a fault confined to `PAGE_STATUS` reads is only caught by `== 20`); both counters are written only in the IOMCU thread and read from the main thread, with single-instruction aligned access on Cortex-M; `protocol_fail_count` is already a since-last-success counter, so no double-counting; the new member needs no initialiser; every consumer of both counters is accounted for; and the change makes the `@Field: RSErr` documentation true for the first time. CI: 100 passing, 1 pending, 0 failing.


</details>

AP-Review:

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_19_AIReview/devcall_pr_reviews.html#pr28725

**Re-reviewed at the same head `3adb56b27d` — my earlier comment above is superseded.** Nothing has changed on the branch; this pass re-read it to check the findings I had already published, and what follows is the part that is new or corrected. I have not restated the findings that still stand unchanged — they are in the linked report.

### Verdict: COMMENT

**Re-read at the same head.** The source fix is correct and does exactly what issue #28534 asked for, and the human review is effectively complete and positive — andyp1per and peterbarker both approved, tridge asked for hardware testing and peterbarker flew it on a sparknavi-blue on 2026-06-14 "without incident". The one genuinely open item is peterbarker's 2025-01-29 empty-`suggestion` block on `AP_IOMCU.cpp:501` (delete the stale comment), which GraphQL confirms is unresolved, not outdated, and pinned to the current head. What is new this round is what the firmware rebuild actually contains.

CI at this head: ✓ 101 passing.

### Findings

- **ISSUE** — `Tools/IO_Firmware/iofirmware_cubered.bin:1` — **The firmware rebuild ships an undisclosed, unrelated IOMCU behaviour change: the SUMD minimum frame length goes from 2 channels to 4.** The source change in this PR is FMU-side and is **not compiled into the IOMCU firmware at all** — the `iomcu` build produces only `iofirmware/{iofirmware,analog,mixer,rc}.cpp.o` from `libraries/AP_IOMCU`, no `AP_IOMCU.cpp.o` — so the `.bin` churn is entirely independent of the stated fix. Decoding the diff: all 17 images keep their exact length and differ in only 4 bytes (10 for `iofirmware_cubered.bin`). Three of those, in all 17, are the Thumb sequence `1e ab / 1e 2b` (`subs r3,r5,#2`; `cmp r3,#30`) becoming `1f 2b / 1c 2b` (`subs r3,r5,#4`; `cmp r3,#28`), located by disassembly inside `AP_RCProtocol_SUMD::_process_byte` — i.e. master commit **`7420edcb43` "AP_RCProtocol: bump min SUMD channels to 4 to avoid corruption"** (2026-05-05). Net effect: master's checked-in IO firmware is **stale and missing a memory-corruption fix**, and this rebuild is corrective — which is good, but it means merging this PR changes IOMCU-side SUMD decoding (frames declaring 2 or 3 channels are now rejected) for every board with an IOMCU, under a commit message that says only "rebuild IO firmware". No code change needed: expand that commit message to name what the rebuild picks up, and add a release-note line that IOMCU firmware re-flashes once on upgrade.

- **ISSUE** — `Tools/scripts/build_ci.sh:446` — CI structurally cannot catch stale IO firmware: the `iofirmware` branch runs `Tools/scripts/build_iofirmware.py` and then `git checkout Tools/IO_Firmware/` — it builds the images and throws them away without ever diffing against the committed ones. That is why master's binaries could drift out of sync with master's own source for four months. Out of scope for this PR, but worth a follow-up: replace the `git checkout` with `git diff --exit-code Tools/IO_Firmware/`, or a warn-only variant, since toolchain differences make byte-exactness fragile.

- **NOTE** — `libraries/AP_IOMCU/AP_IOMCU.cpp:501` — The comment "reset recent error count on first good read" is stale — the reset now runs on **every** good read — and this is exactly peterbarker's unresolved request. Either delete the line, as his empty suggestion asks, or reword it.

- **NOTE** — `libraries/AP_IOMCU/AP_IOMCU.cpp:1188` — Two smaller points. `healthy()` becomes marginally stricter, but only for errors occurring before the **first** successful status read: the deleted block zeroed the cumulative counter once, and after that it was monotonic in both versions, so with `E` such errors recovery needs `read_status_ok >= 128*(E+1)`, i.e. 6.4 s per E at the 20 Hz status rate (E is likely 0 in practice, since `thread_main` skips the timed-events block entirely on any `IOEVENT_INIT` failure). And the monotonic status-read error count is no longer logged — `RSErr` now carries the consecutive counter, zeroed at 20 Hz while `write_log()` emits at 1 Hz — though monotonic visibility is **not** lost, because every `read_status()` failure also bumps `protocol_fail_count` inside `read_registers()` and `total_errors += protocol_fail_count` on the next successful access, so `Nerr` still counts them; you simply cannot separate status-read failures from other register-access failures. `read_status_errors` itself, the exact input to `healthy()`, is now logged nowhere.

- **NOTE** — `libraries/AP_IOMCU/AP_IOMCU.cpp:1` — Commit hygiene: the first commit's subject is `AP_IOMCU: Fix status read error counter not resetting (#1)` with a body of "Update AP_IOMCU.h / Update AP_IOMCU.cpp" — a squash artifact from the author's fork, and the `(#1)` will read as ArduPilot PR #1 in `git log`. The description links #28534 without a `Fixes` keyword, so the issue will not auto-close.

Issue #25846, opened 2023-12-29
There is a bug in AP_AHRS_DCM::drift_correction()


luweiagi:
There is two bug in this function:
===========================================================
(1) first bug:
from here 
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_AHRS/AP_AHRS_DCM.cpp#L1024

```c++
if (now - _last_wind_time > 2000 && airspeed_sensor_enabled()) {
    // when flying straight use airspeed to get wind estimate if available
    const Vector3f airspeed = _dcm_matrix.colx() * AP::airspeed()->get_airspeed();
    const Vector3f wind = velocity - (airspeed * get_EAS2TAS());
    _wind = _wind * 0.92f + wind * 0.08f;
}
```

we can see that `AP::airspeed()->get_airspeed()` means EAS NOT TAS, because `TAS = EAS * get_EAS2TAS()`.
but,  at here
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_AHRS/AP_AHRS_DCM.cpp#L710

```c++
        float airspeed = _last_airspeed;
#if AP_AIRSPEED_ENABLED
        if (airspeed_sensor_enabled()) {
            airspeed = AP::airspeed()->get_airspeed();
        }
#endif

        // use airspeed to estimate our ground velocity in
        // earth frame by subtracting the wind
        velocity = _dcm_matrix.colx() * airspeed;

        // add in wind estimate
        velocity += _wind;
```

we can see that `AP::airspeed()->get_airspeed()` is treated as TAS, because it is added with wind directly.
So the above code snippet can be rewritten as:

```c++
        float airspeed = _last_airspeed;
#if AP_AIRSPEED_ENABLED
        if (airspeed_sensor_enabled()) {
            airspeed = AP::airspeed()->get_airspeed();
        }
#endif

        // use airspeed to estimate our ground velocity in
        // earth frame by subtracting the wind
        velocity = _dcm_matrix.colx() * airspeed  * get_EAS2TAS();

        // add in wind estimate
        velocity += _wind;
```

So, the question is AP::airspeed()->get_airspeed() means TAS or EAS on earth?

===========================================================
(2) second bug:
from here 
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_AHRS/AP_AHRS_DCM.cpp#L489

```
    float yaw_deltat;
```
should be fixed as:
```
    float yaw_deltat = 0.0f;
```
[...]

PR #28727, opened 2024-11-25
Rename airspeed estimate members in AP_AHRS

peterbarker:
This is a no-compiler-output change designed to clarify which sort of airspeed is being used within DCM.

It shows up three places where we use the wrong sort of airspeed - within DCM for drift correction and externally via `airspeed_estimate`

```
Board                    AP_Periph  blimp  bootloader  copter  heli  iofirmware  plane  rover  sub
CubeOrange-periph-heavy  *                                                                     
Durandal                            *      *           *       *                 *      *      *
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     *      *           *       *                 *      *      *
MatekF405                           *      *           *       *                 *      *      *
Pixhawk1-1M-bdshot                  *                  *       *                 *      *      *
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
iomcu                                                                *                         
revo-mini                           *      *           *       *                 *      *      *
skyviper-v2450                                         *                                       
```

PR #28728, opened 2024-11-25
Correct use of TAS/EAS in DCM backend

peterbarker:
This is based on https://github.com/ArduPilot/ardupilot/pull/28727 .  ~Until that is merged only the top two commits should be looked at in this PR.~ That PR has been merged.

The drift correction code was using EAS when it should be using TAS.

We were returning TAS instead of EAS when no airspeed present.

Ping @luweiagi who dug up these issues in https://github.com/ArduPilot/ardupilot/issues/25846


peterbarker:
Closes https://github.com/ArduPilot/ardupilot/issues/25846

PR #28729, opened 2024-11-25
AC_Fence: fixed 2 regressions from 4.5.x

tridge:
This fixes 2 regressions in fence behaviour since 4.5.x.
- we had lost the pre-arm check on being inside the polygon fence for FENCE_AUTOENABLE=3
- if the user has both a RCn_OPTION=11 (FENCE) and FENCE_AUTOENABLE=3 then the autoenable did not happen

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -1728,6 +1728,7 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
             "FENCE_ALT_MIN": 20,
             "FENCE_ACTION": 0,
         })
+        self.reboot_sitl()
 
         self.takeoff(30, mode="LOITER")
         """Hold loiter position."""
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -1719,8 +1719,6 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
     # Also check that the vehicle will not try and ascend too fast when trying to backup from a min alt fence due to avoidance
     def MinAltFenceAvoid(self):
         '''Test Min Alt Fence Avoidance'''
-        self.takeoff(30, mode="LOITER")
-        """Hold loiter position."""
 
         # enable fence, only min altitude
         # No action, rely on avoidance to prevent the breach
@@ -1730,6 +1728,9 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
             "FENCE_ALT_MIN": 20,
             "FENCE_ACTION": 0,
         })
+
+        self.takeoff(30, mode="LOITER")
+        """Hold loiter position."""
 
         # Try and fly past the fence
         self.set_rc(3, 1120)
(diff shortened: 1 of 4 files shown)


andyp1per:
@tridge I fully rely on the autotests to tell me whether a change is ok or not - you can see that this breaks one so will need some investigation. The change looks ok on the face of it.

tridge:
@andyp1per I had to modify the autotest, the test was relying on having the RC option causing the fence to not enable, which is what this PR fixes

andyp1per:
> @andyp1per I had to modify the autotest, the test was relying on having the RC option causing the fence to not enable, which is what this PR fixes

@tridge I don't use fences on a switch so don't particularly care, but I was trying to preserve behavior from 4.5 - so on 4.5 this test would fail?

tridge:
> I don't use fences on a switch so don't particularly care, but I was trying to preserve behavior from 4.5 - so on 4.5 this test would fail?

the presence of an RC switch for fence in 4.5 doesn't prevent the FENCE_AUTOENABLE=2 or 3 from working. In 4.6 before this PR if you have the RC switch then neither autoenable=2 or 3 works to enable the fence.

tridge:
@andyp1per can you approve?

andyp1per:
> * f the user has both a RCn_OPTION=11 (FENCE) and FENCE_AUTOENABLE=3 then the autoenable did not happen

Needs some more discussion because the test you fixed is for FENCE_AUTOENABLE=2

tridge:
> I would really like to see the test work in 4.5 and fail in 4.6 before I approve.

do you realise that the test currently in master fails in 4.5.x?

tridge:
@andyp1per looking into this further, I think there are much deeper problems with the fence code in master. It looks like the auto tests aren't written to check for correct behaviour, but to pass the current behaviour.
I don't actually know why this particular test is passing in master at the moment. When I run what I think is the same steps in sim_vehicle then it fails (gets fence breach soon after the takeoff completes). Oddly enough, that is the right behaviour (to get a fence breach), but somehow when run as an autotest it doesn't.

andyp1per:
> do you realise that the test currently in master fails in 4.5.x?

Right, I did this branch to investigate - https://github.com/andyp1per/ardupilot/tree/test-fence-4.5 - but it fails in a different way. The reason I wanted to discuss is its not clear to me what the behaviour should be. The test changes you made completely invert the test - they may be correct, but I would like to at least understand why. I had thought I had written this test to replicate either yours or Henry's manual testing so I'd be a little surprised if its completely upside down

andyp1per:
> I don't actually know why this particular test is passing in master at the moment. When I run what I think is the same steps in sim_vehicle then it fails (gets fence breach soon after the takeoff completes). Oddly enough, that is the right behaviour (to get a fence breach), but somehow when run as an autotest it doesn't.

Make sure that when you run it in sim_vehicle you use the same speedup as the autotests - the speedup plays havoc with some of these testss.

tridge:
@Hwurzburg I'd like to discuss with you too

andyp1per:
The changes look good to me, but as I say - I really rely on the autotests to catch problems. Not sure what testing we want to do with this.

tridge:
@andyp1per I really need an approval if we follow our rules. Any chance you can approve?

andyp1per (approved):
LGTM

rmackay9:

rmackay9:
Sorry, these were already included in 4.6.0-beta2

Issue #28681, opened 2024-11-19
regression in plane takeoff behaviour in 4.6


tridge:
in 4.6.x we have introduced a bug in takeoff handling. We are not correctly handling an incorrect yaw estimate on takeoff. This is demonstrated in the following flight using CMAC_South, so the plane is pointing approximately south, with COMPASS_USE disabled:
![image](https://github.com/user-attachments/assets/395c20e1-573c-44f9-94d4-1ed07b55ef82)
here is the same test with 4.5.7:
![image](https://github.com/user-attachments/assets/4cc7238f-5a6b-40a1-840c-b75763305363)

menschel:
As reported following https://github.com/ArduPilot/ardupilot/pull/28352 there are three observed regressions in takeoff behaviour.

- The not excepted yaw reset by GPS leading to the course corrections mentioned above.
- The continuous use of 100% throttle after `TKOFF_THR_MAX_T` is exceeded until `TKOFF_ALT` or the mission items takeoff altitude is reached. Previously it fell back to TECS controlled throttle based on the pitch angle.
- The level-off functionality can level out the plane below target altitude thus preventing takeoff to complete.

menschel:
The YAW issue is not present in auto takeoff mission item. 
Measurements are referenced in the PR for the items I mentioned.
https://github.com/ArduPilot/ardupilot/pull/28776

PR #28730, opened 2024-11-25
Plane: fixed takeoff direction with no yaw source

tridge:
in TAKEOFF mode with either very poor yaw source or no yaw source we need to use ground vector and wait for sufficient ground speed
fixes https://github.com/ArduPilot/ardupilot/issues/28681

### DIFF
--- a/ArduPlane/mode_takeoff.cpp
+++ b/ArduPlane/mode_takeoff.cpp
@@ -88,7 +88,7 @@ void ModeTakeoff::update()
         const float groundspeed = groundspeed2d.length();
 
         // see if we will skip takeoff as already flying
-        if (plane.is_flying() && (millis() - plane.started_flying_ms > 10000U) && ahrs.groundspeed() > 3) {
+        if (plane.is_flying() && (millis() - plane.started_flying_ms > 10000U) && groundspeed > 3) {
             if (altitude >= alt) {
                 gcs().send_text(MAV_SEVERITY_INFO, "Above TKOFF alt - loitering");
                 plane.next_WP_loc = plane.current_loc;
--- a/ArduPlane/mode_takeoff.cpp
+++ b/ArduPlane/mode_takeoff.cpp
@@ -83,7 +83,10 @@ void ModeTakeoff::update()
     if (!takeoff_mode_setup) {
         plane.auto_state.takeoff_altitude_rel_cm = alt * 100;
         const uint16_t altitude = plane.relative_ground_altitude(false,true);
-        const float direction = degrees(ahrs.get_yaw());
+        const Vector2f &groundspeed2d = ahrs.groundspeed_vector();
+        const float direction = wrap_360(degrees(groundspeed2d.angle()));
+        const float groundspeed = groundspeed2d.length();
+
         // see if we will skip takeoff as already flying
         if (plane.is_flying() && (millis() - plane.started_flying_ms > 10000U) && ahrs.groundspeed() > 3) {
             if (altitude >= alt) {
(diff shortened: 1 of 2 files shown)


ArduPlane/mode_takeoff.cpp
```diff
@@ -115,7 +118,15 @@ void ModeTakeoff::update()
...
-            if (!plane.throttle_suppressed) {
+            /*
+              don't lock in the takeoff loiter location until we reach
+              a ground speed of AIRSPEED_MIN*0.3 to cope with aircraft
+              with no compass or poor compass. If flying in a very
+              strong headwind then the is_flying() check above will
+              eventually trigger
+             */
+            if (!plane.throttle_suppressed &&
+                groundspeed > plane.aparm.airspeed_min*0.3) {
```
Georacer, line 129: So basically you will not allow the heading to "lock-in" until some groundspeed is gained, deeming the heading valid.
Sounds reasonable.

Tools/autotest/arduplane.py
```diff
@@ -4804,6 +4805,40 @@ def TakeoffTakeoff4(self):
...
+
+        bearing_margin = 35
+        loc = self.mav.location()
+        bearing_from_home = self.get_bearing(start_loc, loc)
+        if bearing_from_home < 0:
+            bearing_from_home += 360
+        if abs(bearing_from_home - start_loc.heading) > bearing_margin:
+            raise NotAchievedException(f"Did not takeoff in the right direction {bearing_from_home}")
+
+        self.fly_home_land_and_disarm()
```
Georacer, line 4840: Since the test is only about the takeoff, does it make sense to not wait for the landing, but end the test early here? This would alleviate the test server a bit.

Georacer (approved):

Hwurzburg (changes requested):
shouldn't the course lock based on GPS course be based on when the GPS speed is valid? and this is not really airspeeed dependentat all?
- in Auto takeoffs, its set 2s after exceeding 5m/s ground speed from GPS (which I think could be set immediately after exceeding this since modern GPS's have been accurate for years, by then)
- in my pending Autoland Lua script PR , I use 5m/s  with a modern  cheap GPS (M8) and it works great capturing takeoff direction, even in strong crosswinds in hand launches
- your criteria is 2.7m/s in the default case
I think that it should just be set to a magic number of X (3-5m/s?) since the quality of the GPS heading is not really related to anything to do with airspeed

menschel:
Off-Topic Question - Does takeoff compensate for cross-wind ?

Because until now I was convinced that the heading error is also present in auto takeoff mission item.
After reviewing the code,
https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/ArduPlane.cpp#L190C5-L190C34
https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/commands_logic.cpp#L566
it appears to be legit and all that is left for the massive course deviation is the crosswind of ~15km/h here. Approx 100m over 300m take-off.
[24-11-23_08-42-26.bin](https://github.com/menschel/ardupilot/blob/stash/build/bin/Arduplane/24-11-23_08-42-26.bin)

Hwurzburg:
No navigation other than heading lick

ArduPlane/mode_takeoff.cpp
```diff
@@ -83,7 +83,10 @@ void ModeTakeoff::update()
     if (!takeoff_mode_setup) {
         plane.auto_state.takeoff_altitude_rel_cm = alt * 100;
         const uint16_t altitude = plane.relative_ground_altitude(false,true);
-        const float direction = degrees(ahrs.get_yaw());
+        const Vector2f &groundspeed2d = ahrs.groundspeed_vector();
+        const float direction = wrap_360(degrees(groundspeed2d.angle()));
+        const float groundspeed = groundspeed2d.length();
+
         // see if we will skip takeoff as already flying
         if (plane.is_flying() && (millis() - plane.started_flying_ms > 10000U) && ahrs.groundspeed() > 3) {
```
IamPete1, line 91:
```suggestion
        if (plane.is_flying() && (millis() - plane.started_flying_ms > 10000U) && groundspeed > 3) {
```

IamPete1: If there a chance that the ground speed check here will trigger before the new one on line 129. If your min airspeed is larger than 10m/s (3/0.3) and it takes longer than 10 seconds to get there then its will skip the full takeoff every time.  Really this skip takeoff path should only be done once. Maybe it should be moved to the mode enter?

Hwurzburg (approved):

IamPete1 (approved):
Looks good for a 4.6 fix.

Future work to move the is-flying check into the mode enter and unify the methods between Auto takeoff and mode takeoff.

PR #28732, opened 2024-11-25
AP_Scripting: Auto turn off sprayer

VICTUSE:
I wrote this script according to this PR #28684 and #23927 request.
Purpose of it is turning off Sprayer (water pump) after completion of auto missions, 
on RTL, LAND or when RC Failsafe triggered. 

I think code comments are pretty revealing.

It is tested only on SİTL sim. The printed servo PWM is to check the exact output.
rcfailsafe


peterbarker (changes requested):
I don't think this one can go in as an applet with the current formulation; it can't just use aux-function 300 if it is an applet as it may conflict with other things on the system.

I'd be happy to merge this if you move it into the "examples" directory!

PR #28735, opened 2024-11-25
Add tabulate dependency

Ryanf55:
* Used in filter_size_compare_branches_csv.py


khancyr (approved):
LGTM

Used in some tools by CI too and not a lot of deps so that is as fine as junitparser

khancyr (changes requested):
rectification : need to update commit style

Ryanf55:
> rectification : need to update commit style

Fixed

peterbarker (approved):
LGTM

PR #28737, opened 2024-11-25
AP_Logger: warn on high dropped log message rate

andyp1per:
It's easy to miss this


IamPete1 (commented):
I have seen logs that would just result in sending this message on every check at 1hz. I think it has to be rate limited down. Typically logging rates done change that much during flight, maybe just one warning per opend log is enough.

But do agree that it is often missed, it's a common bug report on filter review that the results look odd when really there is just no data to go on.

andyp1per:
> maybe just one warning per opend log is enough.

Tricky because there is usually a burst at startup which then settles down

IamPete1:
> > maybe just one warning per opend log is enough.
> 
> Tricky because there is usually a burst at startup which then settles down

@peterbarker has fixed that now, or at least on master I get 0 dropped messages. As you say we regularly got a few at log start before that.

peterbarker:
You can now graph the dropped message count in real time - it's in `SYS_STATUS`.

I think that means we can make this a GCS issue :-)

andyp1per:
> I think that means we can make this a GCS issue :-)

Nah, that requires updating all the GCS' and also assumes you have a GCS

rmackay9:
I'm pretty sure that @IamPete1 is correct that this could lead to a lot of messages being sent to the GCS.  It should probably be a warning and only be sent at most every 30sec.

I think it might be helpful for users if the message gave advice on what to do.  E.g. lower parameter X.

PR #28739, opened 2024-11-25
AP_ADSB: add ability to force Mode3AC Only

magicrub:
Add ability to force Mode3AC Only (disable Mode-S aka ADSB). Technically this can be controlled by the GCS via MAVLink but none implement it and it can totally ground your aircraft if you can't do it immediately, no matter what GCS you're running.

variable description says it all:
```
// @Description: Transmit in traditional Mode 3A/C only and inhibit Mode-S and ES (ADSB) transmissions. This is useful when an ATC (Air Traffic control) requests it in certain scenarios.
// @Values: 0:Transmit all Modes (3A 3C S ES), 1:Transmit Mode 3A/C only (no ADSB).

```
I also added a minor refactor.

### DIFF
--- a/libraries/AP_ADSB/AP_ADSB.cpp
+++ b/libraries/AP_ADSB/AP_ADSB.cpp
@@ -172,7 +172,7 @@ const AP_Param::GroupInfo AP_ADSB::var_info[] = {
     // @Param: OPTIONS
     // @DisplayName: ADS-B Options
     // @Description: Options for emergency failsafe codes and device capabilities
-    // @Bitmask: 0:Ping200X Send GPS,1:Squawk 7400 on RC failsafe,2:Squawk 7400 on GCS failsafe,3:Sagetech MXS use External Config
+    // @Bitmask: 0:Ping200X Send GPS,1:Squawk 7400 on RC failsafe,2:Squawk 7400 on GCS failsafe,3:Sagetech MXS use External Config,4:Transmit in traditional Mode 3A/C only and inhibit Mode-S and ES (ADSB) transmissions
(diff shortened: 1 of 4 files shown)


libraries/AP_ADSB/AP_ADSB.cpp
```diff
@@ -176,6 +176,13 @@ const AP_Param::GroupInfo AP_ADSB::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("OPTIONS",  15, AP_ADSB, _options, 0),
 
+    // @Param: MODE3_ONLY
```
peterbarker, line 179: Why didn't you use an option bit for this?

peterbarker (changes requested):
... also, the refactor is completely unrelated and shouldn't be in here.  Shift it to a new PR, it looks like an easy merge.

libraries/AP_ADSB/AP_ADSB.cpp
magicrub: Hmm. yeah, thats a good idea

magicrub:
refactor is in PR https://github.com/ArduPilot/ardupilot/pull/28763

libraries/AP_ADSB/AP_ADSB.cpp
magicrub: I've updated this PR to be an options bit. Thanks for the idea!

magicrub:
I don't think a wiki update is needed. It's just an option and all the text that is needed is already captured in the options param

PR #28743, opened 2024-11-25
.github: remove mantic test from environnement testing EOL

khancyr:
mantic is EOL since July but the apt repo aren't working anymore


peterbarker (approved):
LGTM

PR #28744, opened 2024-11-25
Feature/GitHub clean

khancyr:
fix a pip warning on CI
use a tag to not catch false issue on github. in this case, it prevents spamming from the log symlink failure.

### DIFF
--- a/Tools/autotest/vehicle_test_suite.py
+++ b/Tools/autotest/vehicle_test_suite.py
@@ -3318,7 +3318,7 @@ class TestSuite(ABC):
             os.link(self.logfile, self.buildlog)
         except OSError as error:
             self.progress("OSError [%d]: %s" % (error.errno, error.strerror))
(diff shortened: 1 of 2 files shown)


Tools/autotest/vehicle_test_suite.py
```diff
@@ -3318,7 +3318,7 @@ def try_symlink_tlog(self):
             os.link(self.logfile, self.buildlog)
         except OSError as error:
             self.progress("OSError [%d]: %s" % (error.errno, error.strerror))
-            self.progress("WARN: Failed to create link: %s => %s, "
+            self.progress("ExcludedError WARN: Failed to create link: %s => %s, "
```
peterbarker, line 3321: We could just say `Problem:` rather than `WARN`?

khancyr: yes, no problem for this.

peterbarker:
Merged, thanks!

PR #28745, opened 2024-11-25
Feature/ci output

khancyr:
move pretty_diff_size to Tools/scripts/build_tests as that is only use on CI (cleanup scripts a bit)

Send the ccache and diff size output to Github Action Summary : 
image


peterbarker (approved):
LGTM

PR #28749, opened 2024-11-26
SITL: Add simulator for tethered vehicle

rishabsingh3003:
This is an extremely simplified simulator for a tethered vehicle. The dynamics aren't accurate, but it allows a fairly reasonable simulation for a "stuck tether".

I want to enhance this simulator to have better dynamics, but also make the tether on the ground movable (perhaps on top of a rover).


tridge:
I'd like to see a simple autotest for this to ensure it doesn't bitrot

tridge (approved):

PR #28750, opened 2024-11-27
AP_NavEKF3: added two more EK3_OPTION bits

tridge:
these allow for easier GPS-denied testing of EKF3, and allows for replay testing without the SetLatLng code being active.

Note that the gaps in the bitmask are for pending range beacon features, we don't want to fill in the gaps so that existing replay logs for that feature remain valid


libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -738,8 +738,8 @@ const AP_Param::GroupInfo NavEKF3::var_info2[] = {
 
     // @Param: OPTIONS
     // @DisplayName: Optional EKF behaviour
-    // @Description: This controls optional EKF behaviour. Setting JammingExpected will change the EKF nehaviour such that if dead reckoning navigation is possible it will require the preflight alignment GPS quality checks controlled by EK3_GPS_CHECK and EK3_CHECK_SCALE to pass before resuming GPS use if GPS lock is lost for more than 2 seconds to prevent bad
-    // @Bitmask: 0:JammingExpected
+    // @Description: This controls optional EKF behaviour. Setting JammingExpected will change the EKF nehaviour such that if dead reckoning navigation is possible it will require the preflight alignment GPS quality checks controlled by EK3_GPS_CHECK and EK3_CHECK_SCALE to pass before resuming GPS use if GPS lock is lost for more than 2 seconds to prevent bad. The DisableSetLatLng option disables processing of SetLatLng commands provided by an external estimator. The DisableGPSAtStartup option allows for easier GPS denied testing to ensure that no GPS data is processed before an AUX function to disable GPS can be processed.
```
timtuxworth, line 741: Seems like a combo between what was there before and what you added doesn't make sense. "if GPS lock is lost for more than 2 seconds to prevent bad. The DisableSetLatLng option disables".

To prevent bad what?

tridge: fixed (existing issue)

rmackay9:
I wonder why we couldn't just use the EK3 source switching feature?  there are 3 slots and you can setup one to not use GPS

tridge:
> I wonder why we couldn't just use the EK3 source switching feature? there are 3 slots and you can setup one to not use GPS

if this was being used for end users, yes, but as a development option to test the behaviour of EKF3 under GPS denied environments it doesn't produce equivalent behaviour.

priseborough (approved):

rmackay9:
@tridge,

This seems like a very rare edge case so I wonder why we need a parameter to set it.  How about an auxiliary switch and/or a mavlink command?

priseborough:
> @tridge,
> 
> This seems like a very rare edge case so I wonder why we need a parameter to set it. How about an auxiliary switch and/or a mavlink command?

With parameters it is possible to change the behaviour when doing tuning and comparison of processing options via EKF replay.

rmackay9:
Hi @priseborough,

OK, but can't auxiliary switches also be changed easily as part of tuning?  I think the recording of the EKF state in the DAL is separate from how the user configures the feature.

Sorry, I don't mean to be difficult but I saw another one of those usability complaints on our YouTube channel and it reminded me that we need to be careful when we expose additional complexity to our users.

rmackay9 (changes requested):
sorry, I'm adding a "request changes" because I just want to be sure we're thinking through the UI exposed to the user

rmackay9 (approved):
I've been meaning to investigate what parts of the EKF3 will use GPS even when the EKF source sets have been configured to NOT use them but I've run out of time and I don't want to hold this up anymore

PR #28755, opened 2024-11-27
AP_Mount: privatise various fields and methods

peterbarker:
We are hoping to pull more target calculations into the frontend.

Having these non-private threatens for calculations to move into the separate backends instead, so privatise them

I'm relying on the autotest suite (compilation...) and inspection for this.

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Backend.h
+++ b/libraries/AP_Mount/AP_Mount_Backend.h
@@ -271,25 +271,9 @@ protected:
     // returns true if mavlink heartbeat should be suppressed for this gimbal (only used by Solo gimbal)
     virtual bool suppress_heartbeat() const { return false; }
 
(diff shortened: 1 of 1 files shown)


libraries/AP_Mount/AP_Mount_Backend.h
```diff
@@ -337,6 +335,8 @@ class AP_Mount_Backend
         Location poi_loc;         // location of the POI
         Quaternion att_quat;      // attitude quaternion of the gimbal
     } poi_calculation;
+    // calculate the Location that the gimbal is pointing at
+    void calculate_poi();
```
rmackay9, line 339: maybe also keep calculate_poi() up higher as well.  I totally understand the intention is to move the methods to be with the members that they use but that's now how the rest of the file is organised and I'd rather not mix styles

peterbarker: Done

PR #28734, opened 2024-11-25
AP_Terrain: Add const to locals

Ryanf55:
# Purpose

Add local const to the AP terrain libraries. Pulled out from another branch. 

# Why

* https://isocpp.org/wiki/faq/const-correctness#retrofitting-const
* https://isocpp.org/wiki/faq/const-correctness#const-and-type-safety

PR #28756, opened 2024-11-27
Add topographic lift library and initial example implementation to SITL

Ryanf55:
# Purpose

Demonstrate slope soaring in SITL using a dead-simple lift estimator. 

# longer term

I want to investigate the idea of feeding in much more complex wind data from CFD models using this interface. 
Can ArduPilot be extensible to heavy models that run ONLY on Linux, and perhaps can we feed in data to AP's FDM from another library (python), or can we link in shared libraries that use things like the STL and Boost just to get fancy CDF models? 

AP SITL is awesome, but it doesn't have all the capabilities. I don't expect this to run on embedded.

### DIFF
--- a/libraries/AP_Terrain/AP_Terrain.cpp
+++ b/libraries/AP_Terrain/AP_Terrain.cpp
@@ -185,6 +185,51 @@ bool AP_Terrain::height_amsl(const Location &loc, float &height, bool corrected)
     return true;
 }
 
+bool AP_Terrain::gradient(const Location &loc, Vector2f& gradient)
+{
+    if (!allocate()) {
+        return false;
+    }
+
+    struct grid_info info;
+
+    calculate_grid_info(loc, info);
+
+    // find the grid
+    const struct grid_block &grid = find_grid_cache(info).grid;
+
+    /*
+      note that we rely on the one square overlap to ensure these
+      calculations don't go past the end of the arrays
+     */
+    ASSERT_RANGE(info.idx_x, 0, TERRAIN_GRID_BLOCK_SIZE_X-2);
+    ASSERT_RANGE(info.idx_y, 0, TERRAIN_GRID_BLOCK_SIZE_Y-2);
+
+
+    // check we have all 4 required heights
+    if (!check_bitmap(grid, info.idx_x,   info.idx_y) ||
+        !check_bitmap(grid, info.idx_x,   info.idx_y+1) ||
+        !check_bitmap(grid, info.idx_x+1, info.idx_y) ||
+        !check_bitmap(grid, info.idx_x+1, info.idx_y+1)) {
+        return false;
+    }
+
+    // hXY are the heights of the 4 surrounding grid points
+    int16_t h00, h01, h10;
+
+    h00 = grid.height[info.idx_x+0][info.idx_y+0];
+    h01 = grid.height[info.idx_x+0][info.idx_y+1];
+    h10 = grid.height[info.idx_x+1][info.idx_y+0];
+    // h11 = grid.height[info.idx_x+1][info.idx_y+1];
+
+    // Positive X gradient means terrain faces east.
+    gradient.x = float(h00 - h10) / grid.spacing;
+    // Positive Y gradient means terrain faces north.
+    gradient.y = float(h01 - h01) / grid.spacing;
+    
+    return true;
+}
+
 
 /* 
    find difference between home terrain height and the terrain
--- a/libraries/AP_Terrain/AP_Terrain.h
+++ b/libraries/AP_Terrain/AP_Terrain.h
@@ -125,6 +125,12 @@ public:
      */
     bool height_amsl(const Location &loc, float &height, bool corrected = true);
 
+    /*
+      find the terrain gradient (slope) for a location
+
+      return false if not available
+     */
+    bool gradient(const Location &loc);
     /* 
        find difference between home terrain height and the terrain
        height at the current location in meters. A positive result
--- a/libraries/AP_Terrain/AP_Terrain.h
+++ b/libraries/AP_Terrain/AP_Terrain.h
@@ -126,11 +126,12 @@ public:
     bool height_amsl(const Location &loc, float &height, bool corrected = true);
 
     /*
-      find the terrain gradient (slope) for a location
+      find the terrain gradient (slope) for a location.
+      Gradient is [x, y] in m/m
 
       return false if not available
      */
-    bool gradient(const Location &loc);
+    bool gradient(const Location &loc, Vector2f& gradient);
     /* 
        find difference between home terrain height and the terrain
        height at the current location in meters. A positive result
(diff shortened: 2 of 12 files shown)


libraries/AP_Terrain/AP_Terrain.cpp
```diff
@@ -185,6 +185,51 @@ bool AP_Terrain::height_amsl(const Location &loc, float &height, bool corrected)
...
+    // check we have all 4 required heights
+    if (!check_bitmap(grid, info.idx_x,   info.idx_y) ||
+        !check_bitmap(grid, info.idx_x,   info.idx_y+1) ||
+        !check_bitmap(grid, info.idx_x+1, info.idx_y) ||
+        !check_bitmap(grid, info.idx_x+1, info.idx_y+1)) {
+        return false;
+    }
+
+    // hXY are the heights of the 4 surrounding grid points
+    int16_t h00, h01, h10;
```
peterbarker, line 218: `const` these and define as part of declaration

libraries/AP_Terrain/AP_Terrain.h
```diff
@@ -422,6 +429,7 @@ class AP_Terrain {
 
     // cache the home altitude, as it is needed so often
     float home_height;
+    Vector2f home_gradient;
```
peterbarker, line 432: Not needed?

libraries/AP_TopoLift/AP_TopoLiftSinglePointEstimate.cpp
```diff
@@ -0,0 +1,90 @@
...
+    // status() handles enable check.
+    if (terrain->status() != AP_Terrain::TerrainStatusOK) {
+        return false;
+    }
+
+    const auto corrected = true;
+    return terrain->height_amsl(loc, height_amsl, corrected);
+
+
+    return true;
```
peterbarker, line 23:
```suggestion
```

libraries/AP_TopoLift/AP_TopoLiftSinglePointEstimate.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return ahrs.wind_estimate(wind);
+}
+
+bool AP_TopoLiftSinglePointEstimate::calc_lift(const Location loc, float& lift, const Vector2f& env_wind) {
+    float terrain_height_amsl;
+    if (!get_terrain_height(loc, terrain_height_amsl)) {
+        return false;
+    }
+
+    [[maybe_unused]] const float veh_height_agl = loc.alt * 1E-2 - terrain_height_amsl;
```
peterbarker, line 53: We have a define for unused stuff

libraries/AP_TopoLift/AP_TopoLiftSinglePointEstimate.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    // status() handles enable check.
+    if (terrain->status() != AP_Terrain::TerrainStatusOK) {
+        return false;
+    }
+
+    return terrain->gradient(loc, grad);
+}
+
+bool AP_TopoLiftSinglePointEstimate::get_wind_at_current_loc(Vector3f& wind) {
```
peterbarker, line 41:
```suggestion
bool AP_TopoLiftSinglePointEstimate::get_wind_at_current_loc(Vector3f& wind)
{
```

etc

libraries/AP_TopoLift/AP_TopoLift_config.h
```diff
@@ -0,0 +1,3 @@
+#ifndef AP_TOPO_LIFT_ENABLED
```
peterbarker, line 1:
```suggestion
#include <AP_HAL/AP_HAL_Boards.h>

#ifndef AP_TOPO_LIFT_ENABLED
```

libraries/SITL/SIM_Aircraft.cpp
```diff
@@ -853,7 +853,10 @@ void Aircraft::update_wind(const struct sitl_input &input)
                        sinf(radians(input.wind.direction))*cosf(radians(input.wind.dir_z)), 
                        sinf(radians(input.wind.dir_z))) * input.wind.speed;
 
-    wind_ef.z += get_local_updraft(position + home.get_distance_NED_double(origin));
+    auto const curr_pos = position + home.get_distance_NED_double(origin);
+    wind_ef.z += get_local_updraft(curr_pos);
+    wind_ef.z += get_topo_lift(location);
```
peterbarker, line 858: Definitely want this parameterized; don't want more CPU generally - even in SITL :-)

libraries/SITL/SIM_Aircraft.cpp
```diff
@@ -1244,6 +1247,29 @@ float Aircraft::get_local_updraft(const Vector3d &currentPos)
...
+    // A SITL wind of 0 is N wind (from N to S).
+    // A SITL wind of 90 is E wind (from E to W).
+
+    // This dir will be 0 with east wind.
+    const float dir = radians(AP::sitl()->wind_direction - 90.0);
+
+    // Wind in the topographic frame. X points east, Y points north.
+    // East wind will be [speed, 0]
+    // North wind will be [0, speed]
+    const Vector2f wind_topo(cosf(dir) * speed, sinf(dir) * speed);
```
peterbarker, line 1263:
```suggestion
    const Vector2f wind_topo{cosf(dir) * speed, sinf(dir) * speed};
```

Ryanf55:
Check `--osd` that links to an external library if we want, or use mavproxy for a python implementation.

libraries/AP_TopoLift/AP_TopoLiftSinglePointEstimate.cpp
```diff
@@ -0,0 +1,90 @@
...
+    const AP_AHRS &ahrs = AP::ahrs();
+
+    return ahrs.wind_estimate(wind);
```
peterbarker, line 44: Should we *really* be using `AP::ahrs()` here?  Your simulation shouldn't really depend on what NavEKF3 is up to - if your fix is wandering your simulation is going to follow that wandering vehicle rather than represent the true lift value the simulated vehicle should be experiencing.

I'm hoping to have a set of AHRS backend results available from the SITL backend which might allow for using the "perfect" AHRS for determining the simulatedwind and a stnadrd backend's results when using on a real vehicle.

peterbarker (changes requested):

libraries/AP_Terrain/AP_Terrain.cpp
Ryanf55: Like the PR from yesterday:) https://github.com/ArduPilot/ardupilot/pull/28734

libraries/AP_Terrain/AP_Terrain.h
Ryanf55:
```suggestion
```

libraries/SITL/SIM_Aircraft.cpp
Ryanf55: Should I parameterize `get_local_updraft` to establish the pattern?

Issue #5672, opened 2017-02-06
Plane: FBWB, Cruise, Loiter modes should be referenced AMSL


WickedShell:
#### Issue details
FBWB, Cruise, Loiter flight modes are all subject to changes in altitude if the home position is changed. These are all expected to maintain their current altitude rather then suddenly diving. (If anyone disagrees then you should post so here, but the current thought is these should all switch to useing AMSL altitudes).

#### Version
Master

#### Platform
[  ] All
[  ] AntennaTracker
[  ] Copter
[ x ] Plane
[  ] Rover

peterbarker:
Still a problem in 2024.  Added a disabled autotest: https://github.com/ArduPilot/ardupilot/pull/27721

Georacer:
This can now be closed. This is fixed twice, by https://github.com/ArduPilot/ardupilot/commit/2e2c7b50ad2a6fb4fde9c27d9511ddfc862db794 and https://github.com/ArduPilot/ardupilot/pull/28675

@peterbarker please reopen if you disagree.

PR #24487, opened 2023-08-02
Plane: cope with home altitude change while navigating

tridge:
this fixes a bug where a change of home altitude would cause a sudden height demand change. This copes with 3 situations:

 - flying with AMSL alt demand. Changing home altitude makes for no change
 - flying with AGL alt demand. Changing home altitude requires update of next_WP_loc
 - flying with home relative alt demand. Changing home altitude changes demand at end of current navigation leg

PR #28758, opened 2024-11-27
autotest: Enabled old test and added new test on home alt reset

Georacer:
# Summary

Regarding Plane tests on Home Alt change during flight:
- Enabled old disabled test that was already passing, but was malformed.
- Created new test variant, where the altitude change occurs during stable level flight.

# Details

## Old test

The existing test should be passing since https://github.com/ArduPilot/ardupilot/pull/24487, but it was malformed. It should be checking for the absolute altitude to be constant. The relative one is supposed to change.

Additionally, before it would perform an FBWA takeoff and would switch to FBWB/CRUISE/LOITER at an altitude that might not be exactly takeoff alt. Since these modes reset the altitude setpoint upon entry, it would be hard to test for a specific altitude.
Now the takeoff happens upon TAKEOFF mode, waits for the altitude to settle and then switches the mode.
@peterbarker let me know if you had intended for this autotest to test something different.

## New test

The new test tries to make the check a bit more explicit, by doing the Home alt change during a stable, level flight, instead of the end of the takeoff climb.

# Testing

## Before

The new test was passing even before https://github.com/ArduPilot/ardupilot/pull/28675, as expected.
The cursor is at the moment of Home Alt change.

The TECS altitude was correctly changed:
Screenshot from 2024-11-27 12-19-15

TECS was getting reset, as expected:
Screenshot from 2024-11-27 12-19-23

The absolute altitude was roughly stable, but the TECS reset caused a dampened oscillation:
Screenshot from 2024-11-27 12-19-19

## After

After https://github.com/ArduPilot/ardupilot/pull/28675 TECS is shifting again:
Screenshot from 2024-11-27 12-20-48

but TECS does not reset:
Screenshot from 2024-11-27 12-20-55

and the altitude does not have a transient anymore:
Screenshot from 2024-11-27 12-20-51

### DIFF
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -6185,14 +6185,18 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
     def SetHomeAltChange(self):
         '''check modes retain altitude when home alt changed'''
         for mode in 'FBWB', 'CRUISE', 'LOITER':
+            self.set_rc(3, 1000)
             self.wait_ready_to_arm()
             home = self.home_position_as_mav_location()
-            self.takeoff(20)
-            higher_home = home
+            target_alt = 20
+            self.takeoff(target_alt, mode="TAKEOFF")
+            self.delay_sim_time(20)  # Give some time to altitude to stabilize.
+            self.set_rc(3, 1500)
+            self.change_mode(mode)
+            higher_home = copy.copy(home)
             higher_home.alt += 40
             self.set_home(higher_home)
-            self.change_mode(mode)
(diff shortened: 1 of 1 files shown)


amilcarlucas:
So does this close #5672? if yes you can add that to the PR description.

Georacer:
> So does this close #5672? if yes you can add that to the PR description.

I closed this manually already, as it was not an issue anymore anyways.

Georacer:
Apparently this fails on `SetHomeAltChange`, I'll have to look more carefully.

tridge (approved):

PR #28764, opened 2024-11-28
Tools: remove duplication in ardupilotwaf.py

robot-to-society:
I don't think two airspeeds are necessary.


tpwrules (approved):
Does create slightly different build results since the linker order is different.

PR #28771, opened 2024-11-29
Plane: add Autoland mode

Hwurzburg:
Captures ground course when gnd speed reaches 5m/s as takeoff direction in NAV_TAKEOFF and Mode Takeoff (if a reliable way to obtain a takeoff direction in other modes is determined, this can be expanded in the future)

Currently valid only for fixed wing configuration, quadplane wont enter mode...

Upon entry creates a final approach waypoint at parameterized distance and alt from home in the opposite direction of autotakeoff direction, flies to it and does a land exactly like NAV_LAND using full landing code for aborts. (Emulates my UniversalAutoLand LUA script #28591)

~~If a DO_LAND_START sequence(s) exist, will jump to them instead~~

If takeoff direction has not been captured then it will just not enter the mode

If not flying it will not enter the mode

Adds it as a long failsafe action (will RTL if no takeoff direction has been captured)...qplane ignores it

Includes an autotest

edit: can now be used in QPlane if fw landing option bit set and is a build option, added landing direction offset parameter for when you want the landing to be different direction than takeoff

PS: once merged I have a PR to add a mode param to allow the direction to be set on arm if a compass is in use instead of autotakeoffs, or to allow setting takeoff direction in FBWA/MANUAL/STABILIZE takeoffs in addition to autotakeoffs

### DIFF
--- a/ArduPlane/Parameters.h
+++ b/ArduPlane/Parameters.h
@@ -363,6 +363,8 @@ public:
 
         k_param_pullup = 270,
         k_param_quicktune,
+        k_param_mode_autoland, 
+
     };
 
     AP_Int16 format_version;
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -172,7 +172,9 @@ public:
     friend class ModeTakeoff;
     friend class ModeThermal;
     friend class ModeLoiterAltQLand;
-
+#if MODE_AUTOLAND_ENABLED
+    friend class ModeAutoLand;
+#endif
 #if AP_EXTERNAL_CONTROL_ENABLED
     friend class AP_ExternalControl_Plane;
 #endif
@@ -326,6 +328,9 @@ private:
 #endif  // QAUTOTUNE_ENABLED
 #endif  // HAL_QUADPLANE_ENABLED
     ModeTakeoff mode_takeoff;
+#if MODE_AUTOLAND_ENABLED
+    ModeAutoLand mode_autoland;
+#endif
 #if HAL_SOARING_ENABLED
     ModeThermal mode_thermal;
 #endif
@@ -454,6 +459,13 @@ private:
         float throttle_lim_min;
         uint32_t throttle_max_timer_ms;
         uint32_t level_off_start_time_ms;
+        // Good candidate for keeping the initial time for TKOFF_THR_MAX_T.
+#if MODE_AUTOLAND_ENABLED
+       struct {
+            float heading; // deg
+            bool initialized;
+        } initial_direction;
+#endif
     } takeoff_state;
 
     // ground steering controller state
--- a/ArduPlane/events.cpp
+++ b/ArduPlane/events.cpp
@@ -65,7 +65,11 @@ void Plane::failsafe_short_on_event(enum failsafe_state fstype, ModeReason reaso
         break;
 #endif // HAL_QUADPLANE_ENABLED
 
-    case Mode::Number::AUTO: {
+    case Mode::Number::AUTO:
+#if MODE_AUTOLAND_ENABLED
+    case Mode::Number::AUTOLAND:
+#endif
+        {
         if (failsafe_in_landing_sequence()) {
             // don't failsafe in a landing sequence
             break;
@@ -145,6 +149,12 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
             set_mode(mode_fbwa, reason);
         } else if (g.fs_action_long == FS_ACTION_LONG_AUTO) {
             set_mode(mode_auto, reason);
+#if MODE_AUTOLAND_ENABLED
+        } else if (g.fs_action_long == FS_ACTION_LONG_AUTOLAND) {
+            if (!set_mode(mode_autoland, reason)) {
+               set_mode(mode_rtl, reason);
+            }
+#endif
         } else {
             set_mode(mode_rtl, reason);
         }
@@ -194,14 +204,23 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
             set_mode(mode_fbwa, reason);
         } else if (g.fs_action_long == FS_ACTION_LONG_AUTO) {
             set_mode(mode_auto, reason);
+#if MODE_AUTOLAND_ENABLED
+        } else if (g.fs_action_long == FS_ACTION_LONG_AUTOLAND) {
+            if (!set_mode(mode_autoland, reason)) {
+               set_mode(mode_rtl, reason);
+            } 
+#endif           
         } else if (g.fs_action_long == FS_ACTION_LONG_RTL) {
             set_mode(mode_rtl, reason);
         }
         break;
-
     case Mode::Number::RTL:
         if (g.fs_action_long == FS_ACTION_LONG_AUTO) {
             set_mode(mode_auto, reason);
+#if MODE_AUTOLAND_ENABLED
+        } else if (g.fs_action_long == FS_ACTION_LONG_AUTOLAND) {
+            set_mode(mode_autoland, reason);
+#endif
         }
         break;
 #if HAL_QUADPLANE_ENABLED
@@ -210,6 +229,9 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
     case Mode::Number::LOITER_ALT_QLAND:
 #endif
     case Mode::Number::INITIALISING:
+#if MODE_AUTOLAND_ENABLED
+    case Mode::Number::AUTOLAND:
+#endif
         break;
     }
     gcs().send_text(MAV_SEVERITY_WARNING, "%s Failsafe On: %s", (reason == ModeReason:: GCS_FAILSAFE) ? "GCS" : "RC Long", control_mode->name());
--- a/ArduPlane/mode.h
+++ b/ArduPlane/mode.h
@@ -897,8 +897,8 @@ public:
 
 protected:
     bool _enter() override;
-    AP_Mission::Mission_Command cmd;
-    bool land_started;
+    AP_Mission::Mission_Command cmd[3];
+    uint8_t stage;
 };
 #endif
 #if HAL_SOARING_ENABLED
--- a/ArduPlane/mode.h
+++ b/ArduPlane/mode.h
@@ -15,6 +15,10 @@
 #define AP_QUICKTUNE_ENABLED HAL_QUADPLANE_ENABLED
 #endif
 
+#ifndef MODE_AUTOLAND_ENABLED
+#define MODE_AUTOLAND_ENABLED 1
+#endif
+
 #include <AP_Quicktune/AP_Quicktune.h>
 
 class AC_PosControl;
@@ -60,6 +64,9 @@ public:
         THERMAL       = 24,
 #if HAL_QUADPLANE_ENABLED
         LOITER_ALT_QLAND = 25,
+#endif
+#if MODE_AUTOLAND_ENABLED
+        AUTOLAND      = 26,
 #endif
 
     // Mode number 30 reserved for "offboard" for external/lua control.
@@ -859,7 +866,41 @@ private:
     bool have_autoenabled_fences;
 
 };
+#if MODE_AUTOLAND_ENABLED
+class ModeAutoLand: public Mode
+{
+public:
+    ModeAutoLand();
 
+    Number mode_number() const override { return Number::AUTOLAND; }
+    const char *name() const override { return "AUTOLAND"; }
+    const char *name4() const override { return "ALND"; }
+
+    // methods that affect movement of the vehicle in this mode
+    void update() override;
+
+    void navigate() override;
+
+    bool allows_throttle_nudging() const override { return true; }
+
+    bool does_auto_navigation() const override { return true; }
+
+    bool does_auto_throttle() const override { return true; }
+    
+
+    // var_info for holding parameter information
+    static const struct AP_Param::GroupInfo var_info[];
+
+    AP_Int16 final_wp_alt;
+    AP_Int16 final_wp_dist;
+    AP_Int16 landing_dir_off;
+
+protected:
+    bool _enter() override;
+    AP_Mission::Mission_Command cmd;
+    bool land_started;
+};
+#endif
 #if HAL_SOARING_ENABLED
 
 class ModeThermal: public Mode
--- a/ArduPlane/mode_autoland.cpp
+++ b/ArduPlane/mode_autoland.cpp
@@ -64,21 +64,62 @@ bool ModeAutoLand::_enter()
         return false;
     }
 
-
-    //setup final approach waypoint
     plane.prev_WP_loc = plane.current_loc;
-    const Location &home = ahrs.get_home();
     plane.set_target_altitude_current();
-    plane.next_WP_loc = home;
-    const float bearing = wrap_360(plane.takeoff_state.initial_direction.heading + 180);
-    plane.next_WP_loc.offset_bearing(bearing, final_wp_dist);
-    plane.next_WP_loc.set_alt_m(final_wp_alt, Location::AltFrame::ABOVE_HOME);
-    
-    // create a command to fly to final approach waypoint and start it
-    cmd.id = MAV_CMD_NAV_WAYPOINT;
-    cmd.content.location = plane.next_WP_loc;
-    plane.start_command(cmd);
-    land_started = false;
+
+    /*
+      landing is in 3 steps:
+        1) a base leg waypoint
+        2) a land start WP, with crosstrack
+        3) the landing WP, with crosstrack
+
+      the base leg point is 90 degrees off from the landing leg
+     */
+    const Location &home = ahrs.get_home();
+
+    /*
+      first calculate the starting waypoint we will use when doing the
+      NAV_LAND. This is offset by final_wp_dist from home, in a
+      direction 180 degrees from takeoff direction
+     */
+    Location land_start_WP = home;
+    land_start_WP.offset_bearing(wrap_360(plane.takeoff_state.initial_direction.heading + 180), final_wp_dist);
+    land_start_WP.set_alt_m(final_wp_alt, Location::AltFrame::ABOVE_HOME);
+    land_start_WP.change_alt_frame(Location::AltFrame::ABSOLUTE);
+
+    /*
+      now create the initial target waypoint for the base leg. We
+      choose if we will do a right or left turn onto the landing based
+      on where we are when we enter the landing mode
+     */
+    const float bearing_to_current_deg = wrap_180(degrees(land_start_WP.get_bearing(plane.current_loc)));
+    const float bearing_err_deg = wrap_180(wrap_180(plane.takeoff_state.initial_direction.heading) - bearing_to_current_deg);
+    const float bearing_offset_deg = bearing_err_deg > 0? -90 : 90;
+    const float base_leg_length = final_wp_dist * 0.333;
+    cmd[0].id = MAV_CMD_NAV_WAYPOINT;
+    cmd[0].content.location = land_start_WP;
+    cmd[0].content.location.offset_bearing(plane.takeoff_state.initial_direction.heading + bearing_offset_deg, base_leg_length);
+    // set a 1m acceptance radius, we want to fly all the way to this waypoint
+    cmd[0].p1 = 1;
+
+    /*
+      create the WP for the start of the landing
+     */
+    cmd[1].content.location = land_start_WP;
+    cmd[1].id = MAV_CMD_NAV_WAYPOINT;
+
+    // and the land WP
+    cmd[2].id = MAV_CMD_NAV_LAND;
+    cmd[2].content.location = home;
+
+    // start first leg
+    stage = 0;
+    plane.start_command(cmd[0]);
+
+    // don't crosstrack initially
+    plane.auto_state.crosstrack = false;
+    plane.auto_state.next_wp_crosstrack = true;
+
     return true;
 }
 
@@ -96,23 +137,22 @@ void ModeAutoLand::update()
 }
 
 void ModeAutoLand::navigate()
-{    
-    // check to see if if we have reached final approach waypoint, switch to NAV_LAND command and start it once if so
-    if (!land_started){
-        if (plane.verify_nav_wp(cmd)){
-            const Location &home_loc = ahrs.get_home();
-            cmd.id = MAV_CMD_NAV_LAND;
-            cmd.content.location = home_loc;
-            land_started = true;
-            plane.prev_WP_loc = plane.current_loc;
-            plane.next_WP_loc = home_loc;
-            plane.start_command(cmd);
-            plane.set_flight_stage(AP_FixedWing::FlightStage::LAND);
-        }
+{
+    if (stage == 2) {
+        // we are on final landing leg
+        plane.set_flight_stage(AP_FixedWing::FlightStage::LAND);
+        plane.verify_command(cmd[stage]);
         return;
-    //otherwise keep flying the current command   
-    } else {
-       plane.verify_command(cmd);
+    }
+
+    // see if we have completed the leg
+    if (plane.verify_nav_wp(cmd[stage])) {
+        stage++;
+        plane.prev_WP_loc = plane.next_WP_loc;
+        plane.auto_state.next_turn_angle = 90;
+        plane.start_command(cmd[stage]);
+        plane.auto_state.crosstrack = true;
+        plane.auto_state.next_wp_crosstrack = true;
     }
 }
 #endif
--- a/ArduPlane/mode_autoland.cpp
+++ b/ArduPlane/mode_autoland.cpp
@@ -1,0 +1,118 @@
+#include "mode.h"
+#include "Plane.h"
+#include <GCS_MAVLink/GCS.h>
+
+#if MODE_AUTOLAND_ENABLED
+
+/*
+  mode AutoLand parameters
+ */
+const AP_Param::GroupInfo ModeAutoLand::var_info[] = {
+    // @Param: WP_ALT
+    // @DisplayName: Final approach WP altitude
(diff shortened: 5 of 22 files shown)


timtuxworth (commented):
Love this idea - some suggestions.

Hwurzburg:
@robustini many people never use MP after initial setup and arent comfortable with mission planning....I never use missions other than autotakeoff/do-land-start/landing sequence with my mission selector app to select land direction for my field based on wind direction....
this is the C++ implementation of my UniversalAutoLand lua script which I use when traveling and just stop at a location and fly...no laptop used..

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "Plane.h"
+#include <GCS_MAVLink/GCS.h>
+
+/*
+  mode AutoLand parameters
+ */
+const AP_Param::GroupInfo ModeAutoLand::var_info[] = {
+    // @Param: WP_ALT
+    // @DisplayName: Final approach WP altitude
+    // @Description: This is the target altitude for final approach waypoint
```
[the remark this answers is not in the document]
Hwurzburg: virtually every other alt param we have implies above home,but it doesnt hurt, so done

timtuxworth:
Should this be disabling fences (autoenabled, and floor) when doing a landing?

Hwurzburg:
It does a NAV_LAND....so whatever it does or does not

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    //setup final approach waypoint
+    plane.prev_WP_loc = plane.current_loc;
+    const Location &home = ahrs.get_home();
+    plane.set_target_altitude_current();
+    plane.next_WP_loc = home;
+    uint16_t bearing_cd = wrap_360((plane.takeoff_state.takeoff_initial_direction + 180));
+    plane.next_WP_loc.offset_bearing(bearing_cd, final_wp_dist);
+    plane.next_WP_loc.alt = home.alt + final_wp_alt*100;
```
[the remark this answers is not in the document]
Hwurzburg: done

Tools/autotest/arduplane.py
```diff
@@ -4200,6 +4200,19 @@ def FlyEachFrame(self):
             self.fly_mission(mission_file, strict=False, quadplane=quadplane, mission_timeout=400.0)
             self.wait_disarmed()
 
+    def AutoLandMode(self):
+        '''Test AUTOLAND mode'''
+        self.customise_SITL_commandline(["--home", "-35.362938,149.165085,585,173"])
+        self.change_mode("TAKEOFF")
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
```
peterbarker, line 4208:
```suggestion
        self.context_collect('STATUSTEXT')
        self.takeoff(mode='TAKEOFF')
```

Tools/autotest/arduplane.py
```diff
@@ -4200,6 +4200,19 @@ def FlyEachFrame(self):
...
+        self.wait_text("Takeoff initial direction")
```
peterbarker, line 4209:
```suggestion
        self.wait_text("Takeoff initial direction", check_context=True)
```

ArduPlane/events.cpp
```diff
@@ -145,6 +146,8 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
             set_mode(mode_fbwa, reason);
         } else if (g.fs_action_long == FS_ACTION_LONG_AUTO) {
             set_mode(mode_auto, reason);
+        } else if (g.fs_action_long == FS_ACTION_LONG_AUTOLAND) {
+            set_mode(mode_autoland, reason);
```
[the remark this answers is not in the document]
Hwurzburg: it does in the mode

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,119 @@
...
+    //if do_land_start exists, jump to it
+    if( (plane.mission.contains_item(MAV_CMD_DO_LAND_START)) && (plane.arming.is_armed())) {
+        if (plane.have_position && plane.mission.jump_to_landing_sequence(plane.current_loc)) {
+            // switch to AUTO
+            plane.mission.set_force_resume(true);
+            if (plane.set_mode(plane.mode_auto, ModeReason::FIXED_WING_AUTOLAND)) {
+                 return true;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
Hwurzburg: dont know what you mean, Tridge wanted this

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,119 @@
...
+    if (quadplane.available()) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "AutoLand is fixed wing only,changing to RTL"); 
+        plane.set_mode(plane.mode_rtl, ModeReason::FIXED_WING_AUTOLAND);
+        return true; 
+    }
+#endif
+    if (!plane.takeoff_state.takeoff_direction_initialized) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "Takeoff initial direction not set,must use autotakeoff");
+        gcs().send_text(MAV_SEVERITY_WARNING, "Changing to RTL"); 
+        plane.set_mode(plane.mode_rtl, ModeReason::FIXED_WING_AUTOLAND);
```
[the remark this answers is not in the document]
Hwurzburg: done

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,119 @@
...
+    uint16_t bearing_cd = wrap_360_cd((plane.takeoff_state.takeoff_initial_direction + 180)*100);
+    plane.next_WP_loc.offset_bearing(bearing_cd * 0.01f, final_wp_dist);
+    plane.next_WP_loc.set_alt_m(final_wp_alt, Location::AltFrame::ABOVE_HOME);
+    
+    // create a command to fly to final approach waypoint and start it
+    cmd.id = MAV_CMD_NAV_WAYPOINT;
+    cmd.content.location = plane.next_WP_loc;
+    plane.start_command(cmd);
+    land_started = false;
+    gcs().send_text(MAV_SEVERITY_WARNING, "AutoLanding");
```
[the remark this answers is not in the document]
Hwurzburg: okay

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,119 @@
...
+    //also must be flying to enter
+    if (!plane.is_flying()) { 
+        return false;
+    }
+    
+    //takeoff direction must be set and must not be a quadplane, otherwise since flying switch to RTL so this can be used as FS action
+#if HAL_QUADPLANE_ENABLED
+    if (quadplane.available()) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "AutoLand is fixed wing only,changing to RTL"); 
+        plane.set_mode(plane.mode_rtl, ModeReason::FIXED_WING_AUTOLAND);
```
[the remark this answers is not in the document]
Hwurzburg: there are 3 reasons to fail...two result in switch to RTL...its a lot less code to do the switch in the enter rather than hack around system.cpp and change the architecture of set_mode, IMHO

ArduPlane/mode.h
```diff
@@ -835,6 +836,45 @@ class ModeTakeoff: public Mode
...
+protected:
+    bool _enter() override;
+    AP_Mission::Mission_Command cmd {};
+    bool land_started;
+ 
+
+private:
+
+
+
```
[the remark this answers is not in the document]
Hwurzburg: done

ArduPlane/Parameters.h
```diff
@@ -100,7 +100,7 @@ class Parameters {
         k_param_sonar_old, // unused
         k_param_log_bitmask,
         k_param_BoardConfig,
-        k_param_rssi_range,     // unused, replaced by rssi_ library parameters
+        k_param_mode_autoland,     // was rssi_range
```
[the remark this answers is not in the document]
Hwurzburg: moved it

ArduPlane/mode_autoland.cpp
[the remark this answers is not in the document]
IamPete1: Well, here your assuming that the user wants RTL. They might prefer to do something else. If you refuse the mode enter they can decide. Maybe they want QRTL or QLAND instead of RTL. These eventualities result in us having to add more Q_OPTIONS in the future, better to just reject the mode change.

ArduPlane/events.cpp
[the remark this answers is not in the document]
IamPete1: Set mode can stay the same, the failsafe case is the only one that should fall-through to RTL. RC or GCS mode changes should just fail. This is how copter does it:

https://github.com/ArduPilot/ardupilot/blob/da4fee5a9a94511ce778a3b740990f8bcf0e12fb/ArduCopter/events.cpp#L424-L437

It trys auto RTL, if that fails it trys RTL, if that fails it trys land.

ArduPlane/mode_autoland.cpp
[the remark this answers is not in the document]
Hwurzburg: okay no mode changes in enter

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,112 @@
...
+ModeAutoLand::ModeAutoLand() :
+    Mode()
+{
+    AP_Param::setup_object_defaults(this, var_info);
+}
+
+bool ModeAutoLand::_enter()
+{ 
+    //if do_land_start exists, jump to it
+    if( (plane.mission.contains_item(MAV_CMD_DO_LAND_START)) && (plane.arming.is_armed())) {
```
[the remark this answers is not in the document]
tridge: i'd prefer just to remove this

tridge (changes requested):

ArduPlane/mode.h
```diff
@@ -835,6 +836,45 @@ class ModeTakeoff: public Mode
...
+    // var_info for holding parameter information
+    static const struct AP_Param::GroupInfo var_info[];
+
+    AP_Int16 final_wp_alt;
+    AP_Int16 final_wp_dist;
+
+protected:
+    bool _enter() override;
+    AP_Mission::Mission_Command cmd {};
```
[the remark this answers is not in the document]
Hwurzburg: done

ArduPlane/mode_autoland.cpp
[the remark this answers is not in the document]
Hwurzburg: removed

Hwurzburg:
Added AUTOLAND_DIR_OFS param to rotate the landing direction from takeoff direction as requested....

Test flown today without any issues...

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,114 @@
...
+    //must be flying to enter
+    if (!plane.is_flying()) { 
+        gcs().send_text(MAV_SEVERITY_WARNING, "Must already be flying!"); 
+        return false;
+    }
+    
+    //takeoff direction must be set and must not be a quadplane, otherwise since flying switch to RTL so this can be used as FS action
+#if HAL_QUADPLANE_ENABLED
+    if (quadplane.available()) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "AutoLand is fixed wing only mode");
```
peterbarker, line 56: We should permit the use of this mode if the Q_OPTION bit is set to allow fixed-wing landings.

Hwurzburg: fine by me and easy to do, but @tridge was the one who asked me to prevent Qplanes from using this

peterbarker: He reminded me the Q_OPTION bit existed :-)

Hwurzburg: @peterbarker thanks for the flight testing! added the ability to use this mode in QPlane if option is set....also landing auto-flaps WAS working, although incorrectly: they were activated immediately upon mode entry while flying to final waypoint...the last commit (unsquashed so changes are easily reviewed) fixes this....SITL tested for flaps
my namespace mojo is weak and still get a compile error on the Qoption addition even adding the QPlane friend class...need help fixing that

tridge:
@Hwurzburg Peter and I tested yesterday, looked pretty good!
One question is if flaps were correctly deployed in landing, @peterbarker can you check the bin log?

Hwurzburg:
@tridge read my comment above yours

Hwurzburg:
added all requests and fixed auto-flaps to be active only in final landing segment instead of entire mode duration...should be good to go now

ArduPlane/Plane.h
```diff
@@ -454,6 +459,11 @@ class Plane : public AP_Vehicle {
         float throttle_lim_min;
         uint32_t throttle_max_timer_ms;
         uint32_t level_off_start_time_ms;
+        // Good candidate for keeping the initial time for TKOFF_THR_MAX_T.
+#if MODE_AUTOLAND_ENABLED
+        int32_t takeoff_initial_direction; //deg
+        bool takeoff_direction_initialized;
+#endif
```
IamPete1, line 466: Comment is out of date and int32 is a bit unexpected for direction, why not float?
```suggestion
#if MODE_AUTOLAND_ENABLED
        struct {
            float heading; // deg
            bool initialized;
        } initial_direction;
#endif
```
You could also add a little structure to keep the variable names shorter.

ArduPlane/events.cpp
```diff
@@ -145,6 +149,12 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
             set_mode(mode_fbwa, reason);
         } else if (g.fs_action_long == FS_ACTION_LONG_AUTO) {
             set_mode(mode_auto, reason);
+#if MODE_AUTOLAND_ENABLED
+        } else if (g.fs_action_long == FS_ACTION_LONG_AUTOLAND) {
+            if(!set_mode(mode_autoland, reason)) {
```
IamPete1, line 154:
```suggestion
            if (!set_mode(mode_autoland, reason)) {
```

ArduPlane/events.cpp
```diff
@@ -194,14 +204,23 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
...
```
IamPete1, line 209:
```suggestion
            if (!set_mode(mode_autoland, reason)) {
```

ArduPlane/events.cpp
```diff
@@ -194,14 +204,23 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
...
+               set_mode(mode_rtl, reason);
+            }            
         } else if (g.fs_action_long == FS_ACTION_LONG_RTL) {
             set_mode(mode_rtl, reason);
+#endif
```
IamPete1, line 214: Should this be before the else if above?

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,118 @@
...
+    
+    //takeoff direction must be set and must not be a quadplane, otherwise since flying switch to RTL so this can be used as FS action
+#if HAL_QUADPLANE_ENABLED
+    if (quadplane.available() && !quadplane.option_is_set(QuadPlane::OPTION::ALLOW_FW_LAND)) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "Option not set to allow fixed wing autoland"); 
+        return false; 
+    }
+#endif
+    if (!plane.takeoff_state.takeoff_direction_initialized) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "Takeoff initial direction not set,must use autotakeoff");
```
IamPete1, line 63:
```suggestion
        gcs().send_text(MAV_SEVERITY_WARNING, "Takeoff initial direction not set, must use autotakeoff");
```

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+
+    //setup final approach waypoint
+    plane.prev_WP_loc = plane.current_loc;
+    const Location &home = ahrs.get_home();
+    plane.set_target_altitude_current();
+    plane.next_WP_loc = home;
+    uint16_t bearing_cd = wrap_360_cd((plane.takeoff_state.takeoff_initial_direction + 180)*100);
+    plane.next_WP_loc.offset_bearing(bearing_cd * 0.01f, final_wp_dist);
```
IamPete1, line 74:
```suggestion
    plane.next_WP_loc.offset_bearing(plane.takeoff_state.takeoff_initial_direction, -final_wp_dist);
```
Negative offset distance remvoe the step adding 180.

IamPete1 (changes requested):
Basically looks good, just a couple of things.

ArduPlane/mode_autoland.cpp
timtuxworth, line 63: This message is a bit long. I'm sure it's going to wrap on Mission Planner. Make two messages?

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,118 @@
...
+    //must be flying to enter
+    if (!plane.is_flying()) { 
+        gcs().send_text(MAV_SEVERITY_WARNING, "Must already be flying!"); 
+        return false;
+    }
+    
+    //takeoff direction must be set and must not be a quadplane, otherwise since flying switch to RTL so this can be used as FS action
+#if HAL_QUADPLANE_ENABLED
+    if (quadplane.available() && !quadplane.option_is_set(QuadPlane::OPTION::ALLOW_FW_LAND)) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "Option not set to allow fixed wing autoland");
```
timtuxworth, line 58: How about "fixed wing autoland not enabled" (shorter)

ArduPlane/mode_autoland.cpp
Hwurzburg: no this is required to put the waypoint behind the home waypoint

ArduPlane/mode_autoland.cpp
Hwurzburg: it will wrap in the MP messages tab, like many other messages...but making it two will make the first one not show in HUB and OSD where last GCS send text always wins

ArduPlane/mode_autoland.cpp
Hwurzburg: change to "Fixed wing autoland option disabled"

ArduPlane/mode_autoland.cpp
Hwurzburg: done

ArduPlane/events.cpp
Hwurzburg: yes...my bad

ArduPlane/events.cpp
Hwurzburg: done

ArduPlane/events.cpp
Hwurzburg: done

ArduPlane/Plane.h
Hwurzburg: done

ArduPlane/mode_autoland.cpp
IamPete1: Using a negative offset is the same as moving the bearing by 180 degrees.

ArduPlane/mode_autoland.cpp
IamPete1: Yes, lets not try and fix Mission Planner issues in AP. It should re-assemble messages if it needs to.

ArduPlane/mode_autoland.cpp
Hwurzburg: actually that does not work...the offset method apparently expects -180 t0 +360 degs for the arguement ...attempted in SITL and it wraps at -180 so it wouldn't really be saving much since a  conversion would have to be done anyway I think

IamPete1: You should not have to wrap, the function will take any value.

ArduPlane/mode_autoland.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+
+    //setup final approach waypoint
+    plane.prev_WP_loc = plane.current_loc;
+    const Location &home = ahrs.get_home();
+    plane.set_target_altitude_current();
+    plane.next_WP_loc = home;
+    uint16_t bearing_cd = wrap_360_cd((plane.takeoff_state.initial_direction.heading + 180)*100);
+    plane.next_WP_loc.offset_bearing(bearing_cd * 0.01f, final_wp_dist);
```
IamPete1, line 74: At least don't go to cd and back.
```suggestion
    const float bearing = wrap_360(plane.takeoff_state.initial_direction.heading + 180);
    plane.next_WP_loc.offset_bearing(bearing, final_wp_dist);
```

Hwurzburg: fair enough...left over from previous cd based stuff

ArduPlane/mode_autoland.cpp
Hwurzburg: not -270....it wraps to -90 which offsets the wrong direction...but will fix the cd vs deg as suggested

tridge:
@peterbarker need the bin log of your tests

peterbarker:
Log from testing on crnch: https://drive.google.com/file/d/1Hwq9iGn1AlVooi7GgTMGQud1k-Sf-NF-/view?usp=sharing

Hwurzburg:
this looks 100% correct and matches what master would do if that final waypoint and nav_land waypoint were the content of a do_land_start sequence....verified that in master in SITL

IamPete1 (changes requested):
New mode needs adding to this array for the available modes message.

https://github.com/ArduPilot/ardupilot/blob/5a8bc161709aba97a8b7a6da3ce0ec1a49df2718/ArduPlane/GCS_Mavlink.cpp#L1592

Sorry, this was added after you started your PR. I can push a commit for it if you prefer.

Hwurzburg:
@IamPete1 if you could, it would help...I tried adding it and get an error message that MODE_AUTOLAND_ENABLED is not defined around the addition

IamPete1:
> @IamPete1 if you could, it would help...I tried adding it and get an error message that MODE_AUTOLAND_ENABLED is not defined around the addition

@Hwurzburg  Fix here: https://github.com/IamPete1/ardupilot/tree/autoland I also rebased.

Hwurzburg:
@IamPete1 fix added...thanks....can you approve now?

IamPete1 (approved):
Looks good.

tridge:
I think this needs a base leg of 1/4 of the approach distance, like this:
image
the base leg will keep the aircraft in the part of the field the pilot is flying in, otherwise we end up with a path like the black line above which is what we got with the bixler

tridge:
@Hwurzburg I've added a base leg to the PR. This is with the base leg:
image

IamPete1:
> I think this needs a base leg of 1/4 of the approach distance, like this: image the base leg will keep the aircraft in the part of the field the pilot is flying in, otherwise we end up with a path like the black line above which is what we got with the bixler

I don't understand how the shown path related to the red(?) line which I guess was the points this added as it was when it was flown? If it was no where near the set path then how does adding the extra leg help?

Hwurzburg:
@IamPete1 the base leg addition was to minimize the overshoot of the final approach waypoint by forcing the approach to it to be no more than 90 degs before the turn onto final....otherwise the approach to the final approach waypoint could have been from 180degs to landing direction resulting in larger overshoot and more track error to correct as it lined back up on final leg...
depending on plane config, roll max, and wp-rad you can get a large overshoot otherwise without base leg and Tridge has fences to the south of the landing approach that could breach....adding a base leg minimizes the overshoot amounts which can be further tunedd with wp_rad,max bank, etc.

@trdige SITL looks good....will test fly this weekend

tridge:
@Hwurzburg tested on Peters plane, worked well:

tridge:
> I don't understand how the shown path related to the red(?) line which I guess was the points this added as it was when it was flown? If it was no where near the set path then how does adding the extra leg help?

there were a few reasons it didn't do what it should do without the base leg:
 - the up to 180 degree turnaround as Henry mentioned
 - it wasn't flying all the way to the autoland start point as it was using WP_RADIUS, and as there is no mission it can't know the turn angle for the next wp, so can't judge when to turn
 - it uses a lot (sometimes nearly all) of the autoland distance getting back onto the landing track
 - the geometry meant it always did the landing from the opposite side of the field that the pilot was flying in (as it guaranteed an overshoot due to the above reasons)

A normal fixed wing landing has a base leg for all size aircraft. The only time you can really avoid it is if you happen to start the landing sequence when already lined up for the landing. We could add a special case for that if we think it is important

Hwurzburg:
During testing I found a corner case bug.....multiple mode takeoffs in a row sometimes triggers the clearing of takeoff_state structure...before I added the bearing into that structure I had accounted for this...but when we went to a substructure including the bearing, it was not included in code that prevents its clearing....fixed....I did not squash, but will clean the commit tree once we agree to add @tridge  base leg....(which I do)

LupusTheCanine:
I personally use LOITER_TO_ALT with tangent exit with great success for landing entry with great success.

I did a bit more SITL testing and it has ugly entry for approaches that are 90° to the wrong side. but it can be fixed by preceding it with LOITER_TURNS set to 0.125 (45°).

image

Here is example waypoint list for EPWA.
```
3	DO_LAND_START	0	0	0	0	0	0	50	Relative	X			0	0	0	0
4	LOITER_TURNS	0.125	0	50	1	52.1719446	20.9432673	35	Relative	X			0.0	0.0	273.3	154
5	LOITER_TO_ALT	0	50	0	1	52.1719446	20.9432673	35	Relative	X			NaN	NaN	0.0	180
6	LAND	0	0	0	1	52.1697995	20.9527194	0	Relative	X			-5.1	-2.9	688.2	110
```

tridge (approved):

PR #28772, opened 2024-11-29
AP_Scripting: update HFE EFI driver script

tridge:
added option to log all CAN frames for debugging new protocols

user has reported new HFE firmware doesn't work, logging will help us work out why


Gromitvt10:
Driver is working on my HFE 215 Genpod EFI engine.  I had to use the latest driver to make it work with Arduplane 4.6.1.Setup was per the instructions.  HFE engine has latest firmware as of 14 June 2025 from HFE..  Fields i am displaying are "EFI Head Temp", "EFI RPM", "EFI Fuel Flow" and "EFI Fuel Consumed".

PR #28774, opened 2024-11-29
Copter: change LOG_BITMASK to include PID logging by default

tridge:
and suggest using LOG_FILE_RATEMAX

### DIFF
--- a/ArduCopter/Parameters.cpp
+++ b/ArduCopter/Parameters.cpp
@@ -306,7 +306,7 @@ const AP_Param::Info Copter::var_info[] = {
 
     // @Param: LOG_BITMASK
     // @DisplayName: Log bitmask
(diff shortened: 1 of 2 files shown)


rmackay9 (approved):

rmackay9:
BTW, I think one of the concerns I had (a long time ago) about turning on more logging was the increased size of logs and the complaints we would get in the support forums.  Those concerns are mostly gone now though due to these advances:

- addition of LOG_RATE_MAX parameter (as mentioned in the description modified in this PR)
- [Peter Hall's hardware report web tool](https://firmware.ardupilot.org/Tools/WebTools/HardwareReport/) that shows what percentage of each log is taken by each message

Re the 2nd point, the web tool, here's a graph of a typical log on the left, and then on the right is one where something has gone wrong :-)

image

andyp1per (approved):

rmackay9:
So this CI failure looks like a real issue although the issue may be in the autotests.

rmackay9:
We can close this because it was included in the follow-up PR https://github.com/ArduPilot/ardupilot/pull/29547

Issue #28775, opened 2024-11-29
Plane 4.6 beta - Takeoff throttle behavior for planes without airspeed sensor changed from 4.5.7


menschel:
## Bug report

In my application forced max throttle regardless of climb angle results in
- Over-speeding of the plane if climb angle is low
- Over-stressing the battery if there are constraints on the discharge curve.

**Issue details**
There have been design changes with Plane 4.6 that change takeoff throttle behavior for planes without an airspeed sensor.
Throttle is forced to max throttle while it was TECS controlled based on climb angle in Plane 4.5.7.

With the ZOHD Drift that I fly for recreational purposes and development, this breaks at least 2 constraints of the Plane.

- It has a recommended max speed of 50km/h or about 15m/s. This is well exceeded to 60km/h or 17m/s. There is also the suspicion that this over-speeds the wing profile, causing it to loose lift.

- It has a custom Li-Ion battery based on 18350 cells that are rated for 3A (5A burst for 10s).
With Plane 4.5.7, it was possible to adjust the current during takeoff by adjusting the takeoff pitch, i.e. 12° lead to 3A.
This is void with Plane 4.6 because it draws 5.5A for more then 30 seconds, thus violates the discharge profile.

**Version**
4.6beta

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[x] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
ZOHD Drift, i.e. Plane, sub 250g 

**Hardware type**
SpeedyBee F405 Wing Mini

**Logs**
![TkOff_Thr_Max_Overspeed_Overcurrent](https://github.com/user-attachments/assets/c0a27c8c-626c-4efe-8009-d21f31ad4482)

[24-11-18_12-43-12.bin](https://github.com/menschel/ardupilot/blob/stash/build/bin/Arduplane/24-11-18_12-43-12.bin)

Georacer:
We intentionally changed the throttle profile in airspeedless takeoff here: https://github.com/ArduPilot/ardupilot/pull/27174/files#diff-499a0c85162572f911764247ca98ae60042d4b6e52ba693b7fdcf9173532810b

And will document it here: https://github.com/ArduPilot/ardupilot_wiki/pull/6149/files#

[...]

PR #28776, opened 2024-11-29
Plane: Fix level-off and throttle without airspeed sensor

menschel:
This PR does intend to fix Issues https://github.com/ArduPilot/ardupilot/issues/28685 and https://github.com/ArduPilot/ardupilot/issues/28775

This patch intends to fix two issues with auto takeoff mission item for
planes without an airspeed sensor.

Level-off fix:
- Correct the level-off condition by factor 2 because during level-off,
  the pitch and therefore the climb-rate decreses linearly.
  Mathmatically the plane therefore only climbs half of the altitude in the expected time
  and is not guaranteed to converge to target altitude by itself.
- Add flag for level-off stage and let TECS handle the final climb.

Throttle without airspeed sensor:
- Remove conditional evaluation of airspeed sensor to enable traditional
  TECS-controlled throttle range also without airspeed sensor.
  This restores the behavior of ArduPlane 4.5.7.

It is to be discussed if the flag THROTTLE_RANGE in parameter TKOFF_OPTIONS is to be reversed to keep
the behavior from Plane 4.5.7 consistent with Plane 4.6 onwards.
It could be something like FORCE_MAX_THROTTLE instead.

Latest measurements with this change cherry-picked onto ArduPilot-4.6:

[24-12-01_09-12-18.bin](https://github.com/menschel/ardupilot/blob/stash/build/bin/Arduplane/24-12-01_09-12-18.bin)
[24-12-01_09-16-52.bin](https://github.com/menschel/ardupilot/blob/stash/build/bin/Arduplane/24-12-01_09-16-52.bin)
[24-12-01_09-21-56.bin](https://github.com/menschel/ardupilot/blob/stash/build/bin/Arduplane/24-12-01_09-21-56.bin)
level_off_fix
throttle_fix

### DIFF
--- a/ArduPlane/takeoff.cpp
+++ b/ArduPlane/takeoff.cpp
@@ -208,7 +208,7 @@ void Plane::takeoff_calc_pitch(void)
     bool pitch_clipped_max = false;
 
     // If we're using an airspeed sensor, we consult TECS.
-    if (ahrs.using_airspeed_sensor()) {
+    if (ahrs.using_airspeed_sensor() || (takeoff_state.in_level_off_phase == true)) {
         calc_nav_pitch();
         // At any rate, we don't want to go lower than the minimum pitch bound.
         if (nav_pitch_cd < pitch_min_cd) {
@@ -279,7 +279,6 @@ void Plane::takeoff_calc_throttle() {
     // Set the minimum throttle limit.
     const bool use_throttle_range = (aparm.takeoff_options & (uint32_t)AP_FixedWing::TakeoffOption::THROTTLE_RANGE);
     if (!use_throttle_range // We don't want to employ a throttle range.
-        || !ahrs.using_airspeed_sensor() // We don't have an airspeed sensor.
         || below_lvl_alt // We are below TKOFF_LVL_ALT.
         ) { // Traditional takeoff throttle limit.
         takeoff_state.throttle_lim_min = takeoff_state.throttle_lim_max;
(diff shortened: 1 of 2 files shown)


menschel:
I would appreciate feedback on the final meter issue. The plane is above target altitude of 50m but exits takeoff mission item 5-6 seconds later when 51m is reached. Am I testing wrong?

menschel:
After a couple of iterations, I came to the conclusion, that Navigation / TECS iterates much faster on the target altitude, if the target altitude of the waypoint is elevated by 1m.
The other conclusion is that the calculation of the level-off switch time is flawed.
Linearly raping out pitch, also linearly ramps out climb-rate. Therefore it cannot reach the target-altitude by math. It can only gain half of the altitude in that time. I will replace the measurements and the figures later.

menschel:
One remark, I have absolutely no clue why with that mission, I get 4A current during takeoff. The exact same mission yields 3A in Plane 4.5.7

menschel:
I'm satisfied with the takeoff behavior now. There is no visible difference from 4.5.7. The parameter needs to be discussed though. I argue the parameter should be "inverted", so any user migrating from 4.5.7 to 4.6 experiences the known behavior.

ArduPlane/takeoff.cpp
```diff
@@ -309,13 +308,17 @@ int16_t Plane::get_takeoff_pitch_min_cd(void)
 
         // are we entering the region where we want to start levelling off before we reach takeoff alt?
         if (auto_state.sink_rate < -0.1f) {
-            float sec_to_target = (remaining_height_to_target_cm * 0.01f) / (-auto_state.sink_rate);
+            float sec_to_target = (remaining_height_to_target_cm * 0.01f * 2.0f) / MIN(TECS_controller.get_max_climbrate(), -auto_state.sink_rate);
```
Georacer, line 311: The decrease of pitch is not linear; that much we've seen from the SITL tests. Can you provide some justification for this formula?

Georacer (commented):
Thank you for your effort @menschel !

As you have noticed, we chose to address the level-off bug in a much simpler way, implemented here: https://github.com/ArduPilot/ardupilot/pull/28792

Regarding the throttle behaviour, we intentionally changed it from 4.5 to 4.6 here: https://github.com/ArduPilot/ardupilot/pull/27758
In 4.5, the throttle during takeoff would be managed by TECS in TAKEOFF mode but set fixed to `TKOFF_THR_MAX` in AUTO mode takeoffs.
We opted to keep the 2nd behaviour.

If you really want the old throttle behaviour for TAKEOFF mode back, we would probably need to make use of another `TKOFF_OPTIONS` bit. That would need to be done carefully, testing all corner cases.

It's not high on my priority list. If you want to implement that, you're very welcome of course.
If you believe the 4.6 behaviour is dangerous for some reason, naturally we'd want to hear your argument.

ArduPlane/takeoff.cpp
menschel: I make the case that the scalar creates the linear dependency between remaining height and pitch.
Integrating A*X results in (A/2)*X² and that's where the factor 2 comes from.
Setting both equal results in X=1/2 for the gained altitude.

The MIN() is a sanity check to limit climb-rate to the max value known to TECS.
It is not dependent on angle but on the power to weight ratio.

I don't know what happens in 4.5.7 but there it is not linear.

menschel:
Ahem, you have mixed Takeoff-mode and Auto-mode takeoff mission item.
I fly Auto-mode missions to have reproducible results between tests.
In 4.5 TECS manages throttle in Auto-mode during takeoff mission item and it works great.

menschel:
This PR has been stale for 2 weeks. I assume it is not welcome then.

rmackay9:
Hi @menschel,

Sorry for the delay, it takes us a while to get to the bottom of issues some times though.  This item is still on our 4.6 issues list so we're going to keep looking into it a bit more.

PR #28778, opened 2024-11-29
AP_InertialSensor: fixed check for changes to notch filters

tridge:
if the configured freq changes on any type of notch then A and Q change, so init must be called. This does not affect only Fixed notches
related to this PR for 4.5:
https://github.com/ArduPilot/ardupilot/pull/28765
but the impact in master is much less significant, it only matters if a user changes the configured freq


andyp1per (approved):
Hopefully passes CI, but this looks right to me.

PR #28779, opened 2024-11-30
CANManager: support CAN frame logging

tridge:
This adds support for logging all incoming CAN frames in a form that can be played back by a python script. It is enabled using a CAN_Pn_OPTIONS bit.
I have been doing this for quite some time with a lua script in examples/CAN_logger.lua, but that only works for can interfaces set to scripting, and it also can miss frames, plus does not support CANFD.
The utility of this functionality is:
 - writing drivers for new CAN devices. We can ask a user to capture all CAN frames then playback in SITL to develop the driver
 - debug hard to reproduce CAN issues on busy CAN buses

The overhead of the logging approach is quite low, as is flash cost.

I've marked this for 4.6 as it will be so useful for supporting users, but it may be more suitable for a bit later 4.6.x release


bugobliterator (approved):
Looks good!

PR #28781, opened 2024-11-30
hwdef: added baro in HolybroG4_GPS

tridge:
new units have a ICP201XX baro and IMU

disable baro by default to avoid issues with airflow and GPS modules installed on the outside of the aircraft
set IMU rate to zero by default

This PR should also be merged for DSDL units fix:
https://github.com/dronecan/DSDL/pull/62


vincentpoont2:
Tested FW can recognize both IMU and baro. 
Please merge @tridge @peterbarker 

The IMU can be recognized after setting the parameter "INS_SAMPLE_RATE". 
(It default to zero to disable the IMU set it to the rate of IMU data you want. For a peripheral like this a value of around 50Hz would be typical)

PR #28783, opened 2024-11-30
AP_InertialSensor: stop sensors converging if motors arm

tridge:
if the user arms within 30s of startup then stop the re-init of the sensors. This can give less accurate frequency as the sample rate may not have settled yet, but it is better than doing init of the filters while the vehicle may be flying

also fix a 32 bit millis wrap

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp
@@ -55,6 +55,15 @@ void AP_InertialSensor_Backend::_set_gyro_oversampling(uint8_t instance, uint8_t
 }
 
 /*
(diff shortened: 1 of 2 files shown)


andyp1per (approved):

IamPete1:
Wouldn't it be better to add a pre-arm that the sensors have converged? Could we stash the result of the convergence from the last time in a hidden param that would short cut the convergence?

tridge:
> Wouldn't it be better to add a pre-arm that the sensors have converged?

I think FPV pilots often do want to do quick takeoffs, and waiting the full 30s is unlikely to be important enough to hold up the flight.

andyp1per:
> I think FPV pilots often do want to do quick takeoffs, and waiting the full 30s is unlikely to be important enough to hold up the flight.

Exactly

PR #28787, opened 2024-12-01
Filter: allow for PID filtering on low flash boards

tridge:
we have a user with a 1M flash board that has some fixed wing structural modes that are different on roll and pitch. They will benefit from being able to setup PID notches on those structural modes and should end up with a tighter tune


andyp1per (approved):
LGTM - thanks!

PR #28789, opened 2024-12-01
MissionRotation LUA - Load one of up to 10 mission files

robustini:
This LUA script is an evolution on the MissionSelector.LUA.
Allows you to select up to 10 missions, either while arming or in the unarmed state.
Requires that an RCx_OPTION be set to 24 (Mission Reset).
The scripts should be installed in the SCRIPTS folder on the MicroSD, and the mission files in the root.
The mission file called mission0.txt (the default mission) must exist and is loaded at boot, if the script does not find it it stops working.
To load to the next mission file just bring AUX switch from low to high for no more than three seconds, each switch loads the next mission file in numerical order.
If the next mission file numerically does not exist load the next one, at the end of the rotation it returns to load mission0.txt.
If AUX is held at high state for more than three seconds the script will restart by loading mission0.txt.
If a mission switch is attempted in AUTO flight mode, a warning “Could not clear current mission” will be shown and no operation will be performed
Provides for sending messages under any condition, load or error.

Classic example of use, i.e. how do i use this rotation:

- mission0.txt is my automatic takeoff mission with home setting at a different point from arming followed by RTL, this way after manually launching my fpv plane I can put on my goggles, exit RTL and execute my flight
- mission1.txt is a full circuit mission for automatic landing, I load it if I want Ardupilot to perform an automated landing from the R side of the runway. This comes in handy, for example, if video transmission is interrupted and visibility on the ground is not optimal
- mission2.txt is like mission1.txt but executes the circuit for landing runway L
- the other mission files I use to perform specific automated flights

In my case if any failsafe intervenes the chosen mission is then executed.
In this way by pre scheduling the appropriate missions for the location where I fly I can afford to handle any emergencies without using GCS, simply choosing with the script the most appropriate mission, and in case changing it on the fly.

### DIFF
--- a/libraries/AP_Scripting/applets/MissionRotation.lua
+++ b/libraries/AP_Scripting/applets/MissionRotation.lua
@@ -4,14 +4,16 @@
 
 local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
 local rc_switch = rc:find_channel_for_option(24)
+
 if not rc_switch then
     gcs:send_text(MAV_SEVERITY.ERROR, "Mission Reset switch not assigned")
     return
 end
 
 local current_mission_index = 0
-local max_missions = 9
+local max_missions = 6  -- Aggiornato a 6 poiché hai fino a mission6.txt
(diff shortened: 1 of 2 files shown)


robustini:
Need better debug.

PR #28791, opened 2024-12-01
Tools: Add scripts for Fedora systems

radityankn:
This script is basically the modification from OpenSUSE Tumbleweed environment install script. The only changes was about the python package names with awk and some renaming to match fedora packages. Tested and the resulting environment is able to compile ArduCopter for Pixhawk1 (STM32) and ESP32S3 Devkit without error. The resulting binaries have not been tested yet


Ryanf55:
Always nice to see more of these! Do you mind amending your commit title to prefix it with "Tools: "

radityankn:
> Always nice to see more of these! Do you mind amending your commit title to prefix it with "Tools: "

No problem, I will add it

Ryanf55:
By commit, I mean here:
image


In CLI:
`git commit --amend`

change your commit title

Then, 

`git push -f`

radityankn:
sorry, accidentally deleted the branch while cleaning up my repo

Ryanf55:
No worries. Looks like you also added a merge commit. ArduPilot uses a rebase workflow. I'll drop that off your branch and force push. Looks fine otherwise!

Ryanf55:
Alrighty, all cleaned up. Approved! I requested final review from a maintainer.

If you have time, mind trying to add a job in CI to enforce this script can continue to work as you intended as a follow up PR?
We have run through many of the setup scripts in this workflow. 
It can be amended to support fedora.
https://github.com/ArduPilot/ardupilot/blob/master/.github/workflows/test_environment.yml

Ryanf55 (approved):

Ryanf55:
Eh, maybe not, I don't think we can run these setup scripts tests on fedora in CI:
https://github.com/actions/runner-images/issues/10802

Maybe you know a way?

radityankn:
I'd like to help, but I don't think I have time to do the CI part. Still new to github, still have to learn more about the testing and CI part

anyway, thank you for reviewing the pr

peterbarker:
I'm a bit concerned by the fact that this is just the fedora install script which has been copied sideways and modified.

By contrast, our `install-prereqs-ubuntu.sh` script handles Ubuntu, debian and `Mint`.

Copying-sideways-and-modifying increases maintenance load on the project, so I would strongly prefer a modification (and perhaps rename) of the existing script to be able to handle both.  

Something like
```
REMOVE_PACKAGE="sudo dnf remove"  # OpenSUSE
if $IS_FEDORA; then
    REMOVE_PACKAGE="sudo zypper rm"
fi
```

would make it all reasonable clean, I think.

radityankn:
@peterbarker sorry for the late reply. So from your suggestion, it's better to merge the OpenSUSE and Fedora scripts into a single env install file to make maintenance easier, correct?

radityankn:
it can be done with some if statements though, to handle differences in package names and others

peterbarker:
> @peterbarker sorry for the late reply. So from your suggestion, it's better to merge the OpenSUSE and Fedora scripts into a single env install file to make maintenance easier, correct?

That's correct.

peterbarker (changes requested):
Please rejig as a modification to the opensuse script.

I wouldn't mind if you started by `git mv` to a different filename first which is collective of both Fedora and OpenSUSE.  I'll note that our Ubuntu install script also handles base Debian and Mint, 'though, so it's already a matter of knowing which one to run from some other source, really....

peterbarker:
> Since it has been quite old, I probably should start with a new script modification, though probably not soon as I'm also quite busy at the moment. Probably better to close this PR before I make a new one with a better script. Thank you for your input @peterbarker

No worries.  Sorry if it seems I'm being too finicky, but whatever goes in has to be maintained!

Linjieqiang:
Any idea? I use the fedora system now. And want to build the ardupilot firmware.

Landrash:
If this could be added without being included in the tests it would be a great addition for those who choose to run Fedora.

Runner image of fedora seems to be a nogo with Github not listening to community with Fedora and others.

PR #28792, opened 2024-12-01
Plane: Fix inability to climb the last few meters of takeoff

Georacer:
## Summary

This PR addresses #28685 and supersedes #28707.

## Details

This affects takeoffs without an airspeed sensor.

### Cause

Some airframes will not climb at 0 pitch, regardless of throttle setting.
Thus, the current takeoff level-off strategy of asymptotically forcing pitch to 0, results in the aircraft to never climb the final few meters.

### Reproduction

The behaviour was reproduced in SITL, by setting:
- `PTCH_TRIM_DEG`: -10.0
- `TKOFF_THR_MAX`: 75.0

### Solution

After the expected time to complete the takeoff has elapsed, drive the desired pitch gradually back to takeoff pitch, over the next 5 seconds.

### Testing

**Before**:
The plane starts level off but manages to climb only up to 45m and hovers there, stuck in takeoff stage.
Screenshot from 2024-12-01 19-49-33

**After**:
The takeoff times out after 2s (`TKOFF_PLIM_SEC`) of being in the level-off maneuver.
Screenshot from 2024-12-01 19-50-37

### Alternatives

The proposed solution of widening the TECS pitch limits and letting TECS take over and regulate the last few meters of the climb was not feasible:
During an airspeedless takeoff, TECS uses throttle to correct for altitude errors.
It would use pitch (if it was allowed) to correct for airspeed errors).
Thus letting TECS command the pitch angle would not help.

The previous PR #28707 was deemed too complicated and too heavy-handed. This present PR prefers to hand over altitude control to TECS.


ArduPlane/takeoff.cpp
```diff
@@ -400,3 +401,17 @@ bool Plane::check_takeoff_timeout(void)
...
 
+/*
+ check if the pitch level-off time has expired; returns true if timeout has occurred
+*/
+bool Plane::check_takeoff_timeout_level_off(void)
+{
+    if (takeoff_state.level_off_start_time_ms > 0) {
+        // A takeoff is in progress.
+        uint32_t now = AP_HAL::millis();
+        if ((now - takeoff_state.level_off_start_time_ms) > (uint32_t)(1000U * g.takeoff_pitch_limit_reduction_sec)) {
```
Georacer, line 412: @peterbarker is this the right way to check a timeout in the face of time folding?

ArduPlane/takeoff.cpp
menschel, line 1: Please investigate if sec_to_target needs to be multiplied by a factor of 2.

In case of linear decrease of pitch / climb-rate, the plane ends up half-way and mathematically does not converge to target altitude.

That is what I came up with during a dozen test flights with 4 different SW changes over the weekend.

I have no clue why this works in Plane 4.5.7.

tridge (approved):

tridge:
@Georacer we need a backport of this for 4.6.x

Georacer:
Alright @tridge . Which branch do I create a PR against?

PR #28793, opened 2024-12-01
AP_Airspeed: don't send AIRSPEED_AUTOCAL message when disabled

tridge:
we were wasting bandwidth sending this all the time

also fixed to support calibration messages for 2nd airspeed sensor

### DIFF
--- a/libraries/AP_Airspeed/Airspeed_Calibration.cpp
+++ b/libraries/AP_Airspeed/Airspeed_Calibration.cpp
@@ -171,23 +171,34 @@ void AP_Airspeed::update_calibration(const Vector3f &vground, int16_t max_airspe
 #if HAL_GCS_ENABLED
 void AP_Airspeed::send_airspeed_calibration(const Vector3f &vground)
 {
+    /*
+      the AIRSPEED_AUTOCAL message doesn't have an instance number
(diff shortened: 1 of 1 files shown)


libraries/AP_Airspeed/Airspeed_Calibration.cpp
```diff
@@ -171,23 +171,33 @@ void AP_Airspeed::update_calibration(const Vector3f &vground, int16_t max_airspe
 #if HAL_GCS_ENABLED
 void AP_Airspeed::send_airspeed_calibration(const Vector3f &vground)
 {
+    /*
+      the AIRSPEED_AUTOCAL message doesn't have an instance number
+      so we can only send it for one sensor at a time
+     */
+    for (uint8_t i=0; i<AIRSPEED_MAX_SENSORS; i++) {
```
IamPete1, line 178: Should we send instances in turn? This doubles the bandwidth if you have two sensors.

tridge: I just added a break, it doesn't have an instance number, so can't usefully send more than one

IamPete1 (approved):

PR #28795, opened 2024-12-01
esp32s3devkit fix adc define name

nkruzan:
allows the board to build


libraries/AP_HAL_ESP32/boards/esp32s3devkit.h
```diff
@@ -88,17 +88,12 @@
...
-	{ADC1_GPIO39_CHANNEL, 11, 39},\
-	{ADC1_GPIO36_CHANNEL, 11, 36}\
+	{ADC_CHANNEL_4, 11, 1},\
+	{ADC_CHANNEL_3, 11, 2},\
+	{ADC_CHANNEL_1, 11, 3},\
+	{ADC_CHANNEL_0, 11, 4}\
 }
+
 // pick one:
 //#define HAL_ESP32_ADC_PINS HAL_ESP32_ADC_PINS_OPTION1
```
davidbuzz, line 98: the correct fix would have been to fix the comment-ed out line here so it ends in OPTION2, and leave both options available to the user to pick between them, no.?

nkruzan: The compiler complaint was about the name of adc defines, I suppose leaving the option to choose is fine, at the time I felt rather than editing all the define names drop the option I had  never tested

nkruzan: The girhub diff looks weird

davidbuzz:
a fix that doesnt delete one of the options like this PR would have been the right approach.

PR #28800, opened 2024-12-04
AC_AutoTune: fix tracking of maximum angular acceleration

### LINKED ISSUE #28799: Copter: AutoTune produces ridiculously small ACCEL_MAX values
## Bug report

**Issue details**

In my 2" whoop-sized quad, running AutoTune produces nice enough rate and angle PIDs, but too small maximum accelerations.

Example: `ATC_ANG_RLL_P` of 25+, but `ATC_ACCEL_R_MAX` of around 1000, written as 4000 because it was clamped from below.

These values are not safe to fly.

**Version**

GitHub master, commit 3ed6ada2e316c52ee9ebae4202c99c681e891693.

**Platform**
[  ] All
[  ] AntennaTracker
[X ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**

Quadcopter.

**Hardware type**

FlywooF405S-AIO

**Logs**

Two logs are attached, for roll and yaw. They are filtered to 5Hz because the flash chip is very small, but should likely be enough for analysis.

[24-12-03_22-19-27_roll.bin.gz](https://github.com/user-attachments/files/18000663/24-12-03_22-19-27_roll.bin.gz)
[24-12-03_22-56-20_yaw.bin.gz](https://github.com/user-attachments/files/18000664/24-12-03_22-56-20_yaw.bin.gz)

tpwrules:
Issue introduced in https://github.com/ArduPilot/ardupilot/pull/27370 and partially fixed in https://github.com/ArduPilot/ardupilot/pull/27762, though evidently not properly tested. Fixes #28799 .

Failing to track the maximum can result in dangerously low values being calculated for `ATC_ACCEL_[RPY]_MAX` and the vehicle becoming unflyable.

Make the variable a reference so that the maximum value is preserved between function calls.

I have tested tuning in SITL, though I have not tested flyability and I have not tested on a real vehicle. I also have not searched for similar missing reference sigils.

I observed the following tuning result differences after this fix, as compared to Copter 4.5.7 (non-listed results are virtually identical):
* roll/pitch rate D is 50% lower
* yaw E rate P/I are 10% lower
* yaw E angle P is 30% higher
* yaw E max accel is 35% lower
* no more warnings during yaw of "please tune manually"

Would appreciate a more thorough look at the code, and confirmation that these value changes are expected improvements and not another (less severe) problem with autotune.

### DIFF
--- a/libraries/AC_AutoTune/AC_AutoTune_Multi.cpp
+++ b/libraries/AC_AutoTune/AC_AutoTune_Multi.cpp
@@ -644,7 +644,7 @@ void AC_AutoTune_Multi::twitching_test_angle(float angle, float rate, float angl
 }
 
 // twitching_measure_acceleration - measure rate of change of measurement
-void AC_AutoTune_Multi::twitching_measure_acceleration(float &accel_average, float rate, float rate_max) const
+void AC_AutoTune_Multi::twitching_measure_acceleration(float &accel_average, float rate, float &rate_max) const
 {
     if (rate_max < rate) {
         rate_max = rate;
--- a/libraries/AC_AutoTune/AC_AutoTune_Multi.h
+++ b/libraries/AC_AutoTune/AC_AutoTune_Multi.h
@@ -156,7 +156,7 @@ private:
     void twitching_test_angle(float angle, float rate, float angle_target, float &meas_angle_min, float &meas_angle_max, float &meas_rate_min, float &meas_rate_max);
 
     // measure acceleration during twitch test
-    void twitching_measure_acceleration(float &accel_average, float rate, float rate_max) const;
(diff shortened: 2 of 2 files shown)


mbuzdalov:
Tested on a 2" quad where #28799 was found. This fix works perfectly!

tpwrules:
Confirmed that #27370 did not introduce those other changes to turning results, just the ones fixed by this PR.

andyp1per (approved):

andyp1per:
Nice!

IamPete1 (approved):
LGTM

lthall (approved):
Great catch. Thank you very much!

PR #28803, opened 2024-12-04
AP_HAL_ESP32: auto-regenerate sdkconfig and speed up CPU

srmainwaring:
Restore default post upgrade to esp-idf 5.3.1.


tpwrules (approved):
Gives a couple dozen extra Hz on the stampfly. Seems to fly fine.

I had to do some munging in waf to be sure the sdkconfig file got regenerated with the new defaults and that users would actually get the new setting.

davidbuzz (approved):
re-reviewing after waf changes - yes, this looks like a reasonable soution - ie turning sdkconfig into a fully managed file now.

PR #21005, opened 2022-06-18
Tramp VTX Protocol

andyp1per:
Implement support for the IRC tramp protocol to allow control of VTX pitmode, power and frequency.
This completes our support for VTX control as we already support SmartAudio and CRSF

https://github.com/ArduPilot/ardupilot/issues/20438

PR #28806, opened 2024-12-04
AP_VideoTX: Tramp: Add half-duplex to port config

menschel:
This Patch fixes a minor annoyance with Tramp.
The protocol is half-duplex, yet the corresponding bit needs to be set manually in options.
This has already been done for Smart Audio.

BTW: Documentation does not even tell about this bit.
https://ardupilot.org/plane/docs/common-vtx.html#immersionrc-tramp

### DIFF
--- a/libraries/AP_VideoTX/AP_Tramp.cpp
+++ b/libraries/AP_VideoTX/AP_Tramp.cpp
@@ -532,8 +532,11 @@ bool AP_Tramp::init(void)
         port->configure_parity(0);
         port->set_stop_bits(1);
         port->set_flow_control(AP_HAL::UARTDriver::FLOW_CONTROL_DISABLE);
-        port->set_options((port->get_options() & ~AP_HAL::UARTDriver::OPTION_RXINV));
-
+        port->set_options((port->get_options() & ~AP_HAL::UARTDriver::OPTION_RXINV)
+                          | AP_HAL::UARTDriver::OPTION_HDPLEX
+                          | AP_HAL::UARTDriver::OPTION_PULLDOWN_TX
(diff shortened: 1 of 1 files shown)


andyp1per (commented):
Thanks! Can I ask that you set the correct baud rate as well? Pretty sure its fixed for tramp

menschel:
You did that already, Andy.

https://github.com/ArduPilot/ardupilot/pull/21005/files#diff-fb699e6f83e01febd7de07fbf9e7d45899e384afff32823ce259be6f0380c597R25

But I had a bracket issue none the less. In my setup, pull-down on tx and rx are not necessary. Should I add them none the less to keep consistency with SmartAudio?

menschel:
-2

I have some InternalError 0x10000000 with this patch, have to reverse engineer where this comes from.

andyp1per:
> -2
> 
> I have some InternalError 0x10000000 with this patch, have to reverse engineer where this comes from.
> 
> `params_restored = (1U << 28), //0x10000000 268435456`
> 
> Seems to be some fighting between tramp trying to set options and parameter storage.
> 
> I don't understand it yet, will keep digging. https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Param/AP_Param.cpp#L381

That indicates some kind of hardware failue or crash. Maybe there is some npe in your code.

menschel:
It may have something to do with me changing between this SW, https://github.com/ArduPilot/ardupilot/pull/28810 and AP 4.5.7 .

I have to get another tramp vtx.

I'm currently violating engineers code by working on the system that is actually flying instead of the technical mock-up.

PR #28810, opened 2024-12-05
AP_VideoTX: Feat: Add RC Switch to VTX Power Mapping

menschel:
This patch does both enhance and simplify the vtx power selection via RC switch.

The RC Option VTX Power expects a 6 position switch and is therefore useless if any other sort of switch is used.

The typical setup found on a radio is a 3 position switch and up until now, position 1 results in pit mode which is not accessible if armed and typically pit mode is tied to arm state through option bits. So Position 1 is lost for no reason!
Position 3 is the MAX POWER value and Position 2 is some arbitrary value in between.

This patch makes use of the same concept as the flight-mode switch with configurable positional values.

It also removes obscure code to find power levels that also seems to have code duplicates.

NOTE: Betaflight features the concept of a "Label" / Value array. Is this feasible for ardupilot?
In that case I'll rework to some sort of VTX_POWER_1_LABEL / VTX_POWER_1_VALUE concept.
It would also open the door to import the VTX table json files in betaflight format.

### DIFF
--- a/libraries/AP_VideoTX/AP_VideoTX.cpp
+++ b/libraries/AP_VideoTX/AP_VideoTX.cpp
@@ -75,6 +75,42 @@ const AP_Param::GroupInfo AP_VideoTX::var_info[] = {
     // @Range: 25 1000
     AP_GROUPINFO("MAX_POWER", 7, AP_VideoTX, _max_power_mw, 800),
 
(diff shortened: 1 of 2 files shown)


menschel:
I have not yet tested this feature.

PR #28812, opened 2024-12-05
AP_HAL_ChibiOS: Refactor Flash Options and add Brownout Reset

magicrub:
Refactor Flash Options and add Brownout Reset. Upon abusing the power on/off requirements, I can "corrupt" my h757 based periph where it gets an ECC error and wipes the app. When enabling the Brownout reset (BOR) feature of the STM32, the problem goes away.

I'm reluctant to add this right now to any existing hwdef.dat but on my own boards the comment in the hwdef.dat is

```
# set Brownout Reset level to highest voltage rising:2.7V, falling:2.61V
define HAL_FLASH_SET_BOR_LEVEL 3
```

### DIFF
--- a/libraries/AP_HAL_ChibiOS/hwdef/common/board.c
+++ b/libraries/AP_HAL_ChibiOS/hwdef/common/board.c
@@ -303,6 +303,10 @@ void __late_init(void) {
   // ensure NRST_MODE is set correctly
   stm32_flash_set_NRST_MODE(HAL_FLASH_SET_NRST_MODE);
 #endif
+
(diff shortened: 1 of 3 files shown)


magicrub:
@bugobliterator  asked me to check if level 2 works and If so then that might be a better default

tridge:
@magicrub this would be good to be able to enable with BRD_OPTIONS, and also be an option bit in AP_Periph

peterbarker:
@magicrub tridge asked for a bit in BRD_OPTIONS... still want a define gating its presence, 'though.

PR #28813, opened 2024-12-05
AP_ADSB: refactor IDENT and add gcs_sendtext feedback

magicrub:
refactor IDENT so that the two methods cmd_int MAV_CMD_DO_ADSB_OUT_IDENT and handle_out_control() plus add gcs_sendtext feedbacl

This also removes the rfselect check which is not used by uAvionix driver(s) and is unnecessary anyways.


rmackay9:
Hi @magicrub,

Normally we are requiring that every PR is approved by a developer besides the dev who wrote the PR.  I don't see an approval on this PR.  Did that happen some other way?

magicrub:
Nope. It was just a refector and no one commented for a week. I'll make sure to bring it up at a dev call to consume dev call time next time

rmackay9:
Hi @magicrub, OK, well, that's very much against our rules

PR #28815, opened 2024-12-06
AP_Network/AP_DDS: default IP address to 192.168.144.x

rmackay9:
I suspect people won't like this change because it is a break changing to existing setups.  Still, I think it's early enough that we could change the default IP address without upsetting too many people as long as we do it as part of a major release (e.g. 4.6).

While I know there will be push back on this change, I really would like it if those pushing back consider usability and how we can make things easier for users.  It's easy to say, "no", it's more difficult to think of ways to make things better


Ryanf55:
Why is 144 better than 13?

timtuxworth:
> Why is 144 better than 13?

It seems that that most existing ethernet equipped devices that can connect to ArduPilot seem to have settled on the 192.168.144  subnet. It seems the HereLink may have set the precedent and there are now a number of other devices such as gimbal/cameras and AI modules from companies like SiYi and TopoTek that are all using 144 addresses. There's some good background here https://ardupilot.org/plane/docs/common-ethernet-vehicle.html

This is going to be one of the first things that I and others will want to do - connect the AP to a BotBlox switch, either using the PPP interface or ethernet.  AP will need to communicate with one or more gimbal/cameras, AI module and companion computer using the same physical port.

This is why I previously suggested that DDS should not be doing it's own networking - it should use the IP addresses configured in the NET* parameters and should be able to run as a NET_Pn_PROTOCOL, so all these devices can be connected over a single physical port to the AP.

Ryanf55:
I'm fine with this change if there's a convention. Our DDS users that use Ethernet understand networking and we have good log messages already. 

Tim - please file a new issue if you have concerns. We are using direct access to the network stack for a few reasons, and we need convincing performance data to remove that support.

Ryanf55 (approved):

timtuxworth:
> Tim - please file a new issue if you have concerns. We are using direct access to the network stack for a few reasons, and we need convincing performance data to remove that support.

https://github.com/ArduPilot/ardupilot/issues/28819

tridge (approved):

Issue #28113, opened 2024-09-15
AP_Camera: FOV quaternion should be Earth-Frame


robertlong13:
Gimbal attitude is reported in a weird "hypothetical earth-frame quaternion if you pointed the aircraft north", which I'm guessing is important for some historical compatibility reason. The gimbal attitude status message at least has a flag indicating if it's using this. The camera FOV status message does not have such a flag, as such I believe the quaternion should be the proper Earth-frame quaternion. You could check the gimbal attitude message, to see if the gimbal is using the "yaw is body-frame", but I think the FOV message should contain all the information needed by a GCS to turn pixel coordinates into a 3D ray vector.

In my testing during Mission Planner development, I'm using this dirty hack as a workaround:
```
    // getting corresponding mount instance for camera
    AP_Mount* mount = AP::mount();
    if (mount == nullptr) {
        return;
    }
    Quaternion quat;
    Location loc;
    Location poi_loc;
    if (!mount->get_poi(get_mount_instance(), quat, loc, poi_loc)) {
        return;
    }

    float roll, pitch, yaw;
    mount->get_attitude_euler(get_mount_instance(), roll, pitch, yaw);
    quat.from_euler(radians(roll), radians(pitch), radians(yaw) + AP::ahrs().get_yaw());
```

**Platform**
[ x ] All

PR #28818, opened 2024-12-06
Camera: CAMERA_FOV_STATUS always sent and attitude is earth-frame

rmackay9:
This makes two improvements to the CAMERA_FOV_STATUS message sent to the GCS:

1. the message is always sent even the gimbal is pointing upwards meaning the image location cannot be calculated
2. the camera's attitude is sent in earth-frame resolving issue https://github.com/ArduPilot/ardupilot/issues/28113

This has been lightly tested in SITL.  I'm quite confident that the first enhancement is all correct and below are some screen shots
fov-status-before-vs-after

I tested the 2nd change using printf debug and it seemed correct but SITL's gimbal simulator is a simple servo gimbal and reports even its roll and pitch in body-frame which made debugging slightly difficult.  So I'd like @robertlong13 to confirm he's happy.

### DIFF
--- a/libraries/AP_Camera/AP_Camera_Backend.cpp
+++ b/libraries/AP_Camera/AP_Camera_Backend.cpp
@@ -295,16 +295,25 @@ void AP_Camera_Backend::send_camera_settings(mavlink_channel_t chan) const
 void AP_Camera_Backend::send_camera_fov_status(mavlink_channel_t chan) const
 {
     // getting corresponding mount instance for camera
-    const AP_Mount* mount = AP::mount();
+    AP_Mount* mount = AP::mount();
     if (mount == nullptr) {
         return;
     }
+
+    // get latest POI from mount
     Quaternion quat;
-    Location loc;
+    Location camera_loc;
     Location poi_loc;
-    if (!mount->get_poi(get_mount_instance(), quat, loc, poi_loc)) {
-        return;
+    const bool have_poi_loc = mount->get_poi(get_mount_instance(), quat, camera_loc, poi_loc);
+
+    // if failed to get POI, get camera location directly from AHRS
+    // and attitude directly from mount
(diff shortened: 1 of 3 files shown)


robertlong13 (approved):
Looks good to me. Tested the quaternion output with my gimbal video testbed. I also validated that your math returns identical results to the dirty workaround I posted in my issue ticket:
```
mount->get_attitude_euler(get_mount_instance(), roll, pitch, yaw);
quat.from_euler(radians(roll), radians(pitch), radians(yaw) + AP::ahrs().get_yaw());
```

libraries/AP_Camera/AP_Camera_Backend.cpp
```diff
@@ -295,32 +295,46 @@ void AP_Camera_Backend::send_camera_settings(mavlink_channel_t chan) const
...
-    if (!mount->get_poi(get_mount_instance(), quat, loc, poi_loc)) {
-        return;
+    const bool have_poi_loc = mount->get_poi(get_mount_instance(), quat, camera_loc, poi_loc);
+
+    // if failed to get POI, get camera location directly from AHRS
+    // and attitude directly from mount
+    bool have_camera_loc = have_poi_loc;
+    if (!have_camera_loc) {
+        have_camera_loc = AP::ahrs().get_location(camera_loc);
+        mount->get_attitude_quaternion(get_mount_instance(), quat);
```
IamPete1, line 314: Surely this method should be const? I was just surprised you had to un-const the singleton.

rmackay9: Hi @IamPete1, yes it should be const except for some bad behaviour inside the SToRM32 gimbal driver which uses the "get" to trigger another request to the gimbal for the attitude.  I'd like to fix that but not in this PR

rmackay9:

PR #28820, opened 2024-12-06
GCS_Common: fix send_sys_status() to send correct current

magicrub:
Current behavior: If you have multiple batteries and ANY of them are unhealthy then the SYS_STATUS msg shows a current of -1A. All the other battery member functions have a default index == AP_BATT_PRIMARY_INSTANCE except healthy() which applies to all instances.

New behavior: battery index [AP_BATT_PRIMARY_INSTANCE] (usually 0) current is always shown for SYS_STATUS regardless of the state of other battery indexes

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -5766,7 +5766,7 @@ void GCS_MAVLINK::send_sys_status()
     float battery_current;
     const int8_t battery_remaining = battery_remaining_pct(AP_BATT_PRIMARY_INSTANCE);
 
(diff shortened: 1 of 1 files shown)


peterbarker:
> New behavior: battery index [AP_BATT_PRIMARY_INSTANCE] (usually 0) current is always shown for SYS_STATUS regardless of the state of other battery indexes

You haven't called out the change in behaviour of the healthy flag here.  Did you also mean to change the meaning of that bit?

magicrub:
Yes, in fact that's the root cause of my problem. I had multiple batteries (mppts and payload sensors) and if ANY of them were unhealthy then the main battery would report 0.0A.

PR #28821, opened 2024-12-06
AP_Scripting: Examples: message_interval: give checker a hint about types

IamPete1:
A new [release](https://github.com/LuaLS/lua-language-server/releases) of the lua language server results in this invalid CI failure:
```
Diagnosis complete, 2 problems found, see /__w/ardupilot/ardupilot/lualogs/check.json
/__w/ardupilot/ardupilot/libraries/AP_Scripting/examples/message_interval.lua:51:30
	Code: param-type-mismatch
	Cannot assign `number|uint32_t_ud` to parameter `integer`.
	- `uint32_t_ud` cannot match `integer`
	- Type `uint32_t_ud` cannot match `integer`
/__w/ardupilot/ardupilot/libraries/AP_Scripting/examples/message_interval.lua:49:32
	Code: param-type-mismatch
	Cannot assign `uint32_t_ud` to parameter `number`.
	- `uint32_t_ud` cannot match `number`
	- Type `uint32_t_ud` cannot match `number`
```

Previously it would not try and guess the types from a table unpack, so we got the benefit of the doubt. Now it seems to try but gets it a bit wrong in this case. Adding the cast annotation means the variables have the correct type so the rest of the file can be checked as normal.

We maybe able to remove the checks again at some point in the future if the checker is improved so it can extract the correct types.

### DIFF
--- a/libraries/AP_Scripting/examples/message_interval.lua
+++ b/libraries/AP_Scripting/examples/message_interval.lua
@@ -44,6 +44,12 @@ gcs:send_text(MAV_SEVERITY.INFO, "Loaded message_interval.lua")
 function update() -- this is the loop which periodically runs
   for i = 1, #intervals do -- we want to iterate over every specified interval
     local channel, message, interval_hz = table.unpack(intervals[i]) -- this extracts the channel, MAVLink ID, and interval
+
+    -- Lua checks get the unpacked types wrong, these are the correct types
+    ---@cast channel integer
+    ---@cast message uint32_t_ud
+    ---@cast interval_hz number
+
     local interval_us = -1
     if interval_hz > 0 then
       interval_us = math.floor(1000000 / interval_hz) -- convert the interval to microseconds


rmackay9 (approved):

rmackay9:
I don't think I've ever seen the "cast" keyword before in Lua

IamPete1:
> I don't think I've ever seen the "cast" keyword before in Lua

Strictly is not lua itself, just the markup that the checker uses. There is no change to how the script works, this just means the checks do the right thing. Were trying to use the checks to add type safety to lua which not a type safe language. 

The full documentation for the various annotations is here: https://luals.github.io/wiki/annotations/. I have been tying to annotate any scripts I write with them too. Certainty helps catch silly mistakes.

Hwurzburg:
@rmackay9 @peterbarker can we get this merged??

rmackay9:
Merged, thanks!

rmackay9:
I'm including this to help resolve CI failures for 4.6.0-beta2

PR #28822, opened 2024-12-07
AP_HAL: protect against invalid values when calculating Jain's estimator

andyp1per:
For some reason I see this on 4.6 when using the FFT

Without it you get cnstring_nan on boot

Tested on my 5"


IamPete1 (approved):
This protection looks good, but I wonder if there is a underlying issue we should fix too.

rmackay9:
We have this report from 4.5.3 which looks like it may be the same issue https://discuss.ardupilot.org/t/copter-4-5-3-released/118508/3

Issue #28823, opened 2024-12-07
New cork check fails on dev4.7 build on autopilots


Hwurzburg:
#27029 causes flow_of_control errors on dev Plane build for KakutH7Wing

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[ x ] Plane
[  ] Rover
[  ] Submarine

andyp1per:
Hi @Hwurzburg - can you possibly post your parameters. Did this occur in flight or on the ground?

robustini:
Same on Matek H743 Wing.

Hwurzburg:
> Hi @Hwurzburg - can you possibly post your parameters. Did this occur in flight or on the ground?

prearm fail...your new PR fixes (on the bench)

andyp1per:
#28828 is an alternative

timtuxworth:
I'm getting this on the ground on a QioTek Zealot H743. Here's a log:
https://www.dropbox.com/scl/fi/qpnslilgvkrxnoom6i8tr/log_68_2024-12-9-09-59-34.bin?rlkey=5aswmo6eawgg5n6xj0kwfoc7r&st=5d9b3go5&dl=0

PR #28825, opened 2024-12-08
Plane: match cork() and push()

andyp1per:
Hopefully closes https://github.com/ArduPilot/ardupilot/issues/28823

I'm kind of in two minds about this internal error now. We could just remove it, but lets see how far the rabbit hole goes.

### DIFF
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -1163,7 +1163,7 @@ private:
     void dspoiler_update(void);
     void airbrake_update(void);
     void landing_neutral_control_surface_servos(void);
-    void servos_output(void);
+    void servos_output(bool cork = true);
     void servos_auto_trim(void);
     void servos_twin_engine_mix();
     void force_flare();
--- a/ArduPlane/servos.cpp
+++ b/ArduPlane/servos.cpp
@@ -961,7 +961,7 @@ void Plane::set_servos(void)
 #endif  // AP_ICENGINE_ENABLED
 
     // run output mixer and send values to the hal for output
-    servos_output();
+    servos_output(false);
 }
 
 /*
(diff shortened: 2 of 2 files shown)


ArduPlane/Plane.h
```diff
@@ -1163,7 +1163,7 @@ class Plane : public AP_Vehicle {
     void dspoiler_update(void);
     void airbrake_update(void);
     void landing_neutral_control_surface_servos(void);
-    void servos_output(void);
+    void servos_output(bool cork = true);
```
rmackay9, line 1166: Thanks for this.  It's not my area but it might be safer to remove the default from the argument and force the caller to make a choice.. maybe add a comment to help them make the choice.  Just an idea though

IamPete1: I agree, the default hides the other users (here on github anyway) which are important to understand the scope of the change.

IamPete1:
I'm not sure its really fair to call this a bug. I agree its a bit odd, but plane had been working fine like this for years. I would like to tidy up the oddness at some point. I'm slightly hesitant to change plane to fix a check that was added for a feature that plane does not use, seems backwards. 

We do have `APM_BUILD_TYPE` defines in other parts of the hal. Before we rush this fix in it might be worth double checking the other vehicle types, we may need a similar fix elsewhere at which point a define to error on copter only might be a better option.

IamPete1:
I wonder if we could add the same error hal SITL. That would make testing easier at least.

andyp1per:
@IamPete1 I am happy to instead remove the internal error if @tridge and @peterbarker agree - it was added because I was assured it should be safe 😄 - I don't really have strong opinions either way.

andyp1per:
Here's the alternative if you prefer @IamPete1 - https://github.com/ArduPilot/ardupilot/pull/28828

IamPete1:
We merged the other one.

PR #28828, opened 2024-12-09
AP_HAL_ChibiOS: don't throw an internal error on mismatched cork/push

andyp1per:
Please decide if you want this or #28825

### DIFF
--- a/libraries/AP_HAL_ChibiOS/RCOutput.cpp
+++ b/libraries/AP_HAL_ChibiOS/RCOutput.cpp
@@ -1329,9 +1329,6 @@ bool RCOutput::get_output_mode_banner(char banner_msg[], uint8_t banner_msg_len)
  */
 void RCOutput::cork(void)
 {
-    if (corked) {
-        INTERNAL_ERROR(AP_InternalError::error_t::flow_of_control);
-    }
     corked = true;
 #if HAL_WITH_IO_MCU
     if (iomcu_enabled) {


libraries/AP_HAL_ChibiOS/RCOutput.cpp
```diff
@@ -1329,9 +1329,6 @@ bool RCOutput::get_output_mode_banner(char banner_msg[], uint8_t banner_msg_len)
  */
 void RCOutput::cork(void)
 {
-    if (corked) {
```
tridge, line 1332: i'd remove this internal error, but not the other one
calling cork(),cork(),push() isn't great, but is not actually dangerous. Calling push() without cork() is the key issue

tridge (changes requested):

libraries/AP_HAL_ChibiOS/RCOutput.cpp
andyp1per: Fixed

tridge (approved):

tridge:
@IamPete1 I think this is reasonable, but would like you to look too

IamPete1 (approved):
LGTM

timtuxworth:
Now I'm getting a panic 
https://www.dropbox.com/scl/fi/a6jypd23rnt9dpigamv3q/log_89_2024-12-11-13-44-32.bin?rlkey=bc34pt8or1pa0qy9whw3792ct&st=9ekhb8yx&dl=0

PR #28829, opened 2024-12-09
AP_HAL_ChibiOS:add board support for CrazyF405

Srivenkateshwar:
Hardware Description file for CrazyF405.

PR #28830, opened 2024-12-09
AP_HAL_ChibiOS:added bootloader support for CrazyF405

Srivenkateshwar:
Added bootloader support for CrazyF405 board


peterbarker (changes requested):
I don't see any corresponding hwdef for these files.

We would usually have this commit as part of a wider PR which adds support for the CrazyF405 board.  Do you have such a PR?

Srivenkateshwar:
> I don't see any corresponding hwdef for these files.
> 
> We would usually have this commit as part of a wider PR which adds support for the CrazyF405 board. Do you have such a PR?

This is the PR were I have added the support for CrazyF405
https://github.com/ArduPilot/ardupilot/pull/28829

Hwurzburg:
This should be included in https://github.com/ArduPilot/ardupilot/pull/28829 instead of a separate PR

Srivenkateshwar:
> This should be included in #28829 instead of a separate PR

Should I include now?

Hwurzburg:
Yes

peterbarker:
@Srivenkateshwar how are we going on this PR?  I believe you were going to add something here?

peterbarker:
Looks like we got these as part of the other PR.

Thanks!

PR #28831, opened 2024-12-09
Tuning indication on OSD(and GCS)

Yury-MonZon:
This PR includes:

Displaying the current value of the parameter being tuned
image

When tuning knob is in the middle (note `>` instead of `:`):
image

Adding PIDFF tuning sets for FixedWing mode:

1. D and P for Roll and Pitch (for quick post-autotune adjustments)
2. PIDFF for Roll
3. PIDFF for Pitch

These additions are primarily designed for tuning via FPV without requiring MAVLink modems.

I'll add PR for wiki once everything is settled, currently it is here https://github.com/Yury-MonZon/ardupilot_wiki/tree/patch-1

### DIFF
--- a/ArduPlane/tuning.h
+++ b/ArduPlane/tuning.h
@@ -94,6 +94,9 @@ private:
         TUNING_SET_RATE_PITCHDP =            8,
         TUNING_SET_RATE_ROLLDP =             9,
         TUNING_SET_RATE_YAWDP =             10,
+        TUNING_SET_DP_ROLL_PITCH =          11,
+        TUNING_SET_PIDFF_ROLL =             12,
+        TUNING_SET_PIDFF_PITCH =            13,
     };
 
     AP_Float *get_param_pointer(uint8_t parm) override;
@@ -112,6 +115,9 @@ private:
     static const uint8_t tuning_set_rate_pitchDP[];
     static const uint8_t tuning_set_rate_rollDP[];
     static const uint8_t tuning_set_rate_yawDP[];
+    static const uint8_t tuning_set_dp_roll_pitch[];
+    static const uint8_t tuning_set_pidff_roll[];
+    static const uint8_t tuning_set_pidff_pitch[];
 
     // mask of what params have been set
     uint64_t have_set;
--- a/libraries/AP_Tuning/AP_Tuning.cpp
+++ b/libraries/AP_Tuning/AP_Tuning.cpp
@@ -218,6 +218,7 @@ void AP_Tuning::check_input(uint8_t flightmode)
     last_channel_value = chan_value;
 
     float new_value;
+    static float old_value;
     if (chan_value > 0) {
         new_value = linear_interpolate(center_value, range*center_value, chan_value, 0, 1);
     } else {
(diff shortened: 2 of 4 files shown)


andyp1per:
Nice idea! @Hwurzburg may have an opinion

Hwurzburg:
not sure what you are talking about "0.5f"
....but 5% seems reasonable for any PID related value as a threshold to display a new message (remember OSD messages persist a while (10s is default)....and ultrafine adjustment of PIDs is a pointless exercise since noticeable changes only occur with larger changes and you need a wide margin of stability in order to prevent corner case flight points resulting in oscillations...for example, we take PIDS, to oscillation, when tuning in level flight and cut them by 1/3 to prevent instability in other flight extremes in our autotune algorithms...

Hwurzburg:
the wiki has lots of details on tuning process for each vehicle

the TX tuning is really a bit dated and for special cases (like default pids results in an unstable vehicle...highly unlikely)....in Plane the autotune for fixed wing and quicktune for quadplanes is the easy path, Copter: Autotune, Rover:is pretty straightforward....using manual PID tuning is really beyond most users and can even be dangerous...

most users will have a lot of difficulty using manual tuning and getting good results....sometimes even having a dangerous tune on big vehicles...

Hwurzburg (approved):

libraries/AP_Tuning/AP_Tuning.cpp
```diff
@@ -202,23 +202,23 @@ void AP_Tuning::check_input(uint8_t flightmode)
...
         }
         // starting tuning
         mid_point_wait = false;
-        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Tuning: mid-point %s", get_tuning_name(current_parm));
         AP_Notify::events.tune_started = 1;
     }
     last_channel_value = chan_value;
 
     float new_value;
+    static float old_value;
```
tridge, line 221: the static doesn't look right for when we change which parameter we are tuning, we'd need the variable in the class and reset on param change

libraries/AP_Tuning/AP_Tuning.cpp
```diff
@@ -228,6 +228,15 @@ void AP_Tuning::check_input(uint8_t flightmode)
     need_revert |= (1U << current_parm_index);
     set_value(current_parm, new_value);
 
+    if ( fabsf(new_value-old_value) > (0.05 * old_value) ) {
+        old_value = new_value;
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, 
+            "Tuning %s%s%0.4f",
```
tridge, line 234: if the intention is this comes out as speech in mavproxy it needs to start with "Tuning: " or an "@"

tridge (changes requested):

libraries/AP_Tuning/AP_Tuning.cpp
peterbarker, line 221: Move initialisation of value to `check_selector` switch and `old_value` into the tuning class

ArduPlane/tuning.h
```diff
@@ -94,6 +94,9 @@ class AP_Tuning_Plane : public AP_Tuning
         TUNING_SET_RATE_PITCHDP =            8,
         TUNING_SET_RATE_ROLLDP =             9,
         TUNING_SET_RATE_YAWDP =             10,
+        TUNING_SET_DP_ROLL_PITCH =          11,
```
tridge, line 97: on what aircraft does it make sense to link D and P?

tridge (approved):

PR #28834, opened 2024-12-09
AP_HAL_QURT: Add support for the ModalAI IO board to support PWM ESCs

katzfey:
Adding in support for the ModalAI IO board. This can be used to interface to PWM ESCs since the VOXL boards have no PWM capability on their own. This code will attempt to detect whether a ModalAI ESC is attached or the ModalAI IO board. Depending on what it detects it will send the appropriate PWM commands.

This has been flight tested with ModalAI ESC to verify that the original functionality has not been impacted.
This has been bench tested with a ModalAI IO board and 4 PWM ESCs. Can arm and spin up motors. RC throttle will control motor spin speed. Still needs an actual flight test but I think the PR is good to go at this time.

### DIFF
--- a/libraries/AP_HAL_QURT/RCOutput.cpp
+++ b/libraries/AP_HAL_QURT/RCOutput.cpp
@@ -8,19 +8,27 @@ extern const AP_HAL::HAL& hal;
 
 using namespace QURT;
 
+// ESC specific definitions
 #define ESC_PACKET_TYPE_PWM_CMD 1
 #define ESC_PACKET_TYPE_FB_RESPONSE 128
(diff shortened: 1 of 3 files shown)


katzfey:
WIP label can be removed. It's complete, tested, and ready to go.

tridge (commented):
nicely done!
have you checked the feedback from the ESCs still works in ESC mode?

katzfey:
Yep, in fact there was a bug with ESC feedback before that I just noticed and fixed. It uses the low bit to request feedback. It wasn't being cleared from all 4 ESCs before choosing the one to request and setting that bit. So often multiple requests would be sent at once which would result in no feedback because they share the UART and the feedback packets would collide.

PR #28840, opened 2024-12-11
Fence warn at the margins

andyp1per:
This PR allows you to get a repeated warning when flying outside FENCE_MARGIN but inside the fence itself. Can be enabled by setting FENCE_OPTIONS=4. Multiple people have requested this because they are required to have a hard fence with failsafe action for regulatory reasons but want to know when they are close so they can avoid the action (typically when flying manually)


libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -876,12 +905,48 @@ void AC_Fence::record_breach(uint8_t fence_type)
...
 
+/// clear_margin_breach - update margin reach bitmask
+void AC_Fence::clear_margin_breach(uint8_t fence_type)
+{
+    if (_breached_fence_margins & fence_type) {
+        const uint32_t now = AP_HAL::millis();
+
+        // emit a message indicated we're newly-breached, but not too often
+        if (option_enabled(OPTIONS::MARGIN_BREACH)
+            && AP_HAL::timeout_expired(_last_margin_breach_notify_sent_ms, now, 1000U)) {
```
timtuxworth, line 941: parameterize "not too often"?

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3906,6 +3906,14 @@ function fence:get_breach_time() end
 ---| 8 # Minimum altitude
 function fence:get_breaches() end
 
+-- Returns the type bitmask of any fence whose margins have been crossed
+---@return integer fence_type bitmask
+---| 1 # Maximim altitude
+---| 2 # Circle
+---| 4 # Polygon
```
timtuxworth, line 3913: It might be helpful to know it it was an include or exclude fence.

andyp1per: Not easy to do and we don't currently do that for breaches

libraries/AC_Fence/AC_Fence.cpp
andyp1per: I'm loathe to add another parameter unless its really required - we have many other repeating messages in warning situations that do not have a configurable repeat interval

timtuxworth: Fair enough, but this seems to be every second. (Do I have that wrong?) That seems pretty often to me.

andyp1per: Yeah, open to a better value - although you obviously can turn it off. You only get the warning when you are inside the fence but outside the margin - its supposed to be naggy.

enok71: As one of the requestors of this feature: I think it looks like the solution to our problem, which @andyp1per described perfectly above.
1 message per second is fine with me.

andyp1per: @enok71 note that the polyfence support is very approximate. There is no margin support for polyfences so what I did was check a square box, centered on the current position with edges at the margin and see if any corner was outside the fence. Obviously lots of situations in which this is not quite accurate, but checking the polyfence is quite expensive and it wasn't clear to me if there was a better way without writing lots of code.

Ryanf55:
> This PR allows you to get a repeated warning when flying outside FENCE_MARGIN but inside the fence itself. Can be enabled by setting FENCE_OPTIONS=4. Multiple people have requested this because they are required to have a hard fence with failsafe action for regulatory reasons but want to know when they are close so they can avoid the action (typically when flying manually)

Do you have any references to the specific regulatory requirements for fences?

libraries/AC_Fence/AC_Fence.cpp
enok71: Ok. I realize computing actual smallest distance to polygon exterior would be complicated.

Perhaps allowing two geofences would be a way to eliminate that? One warning-geofence and one failsafe action geofence.

Anyway, what you already have implemented is much better than nothing, so thanks once again.

andyp1per: @timtuxworth pretty sure the avoid code already uses the margin - this is for the specific case where you are flying manually without avoid.

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -665,7 +673,24 @@ bool AC_Fence::check_fence_polygon()
             return true;
         }
         return false;
+    } else if (option_enabled(OPTIONS::MARGIN_BREACH)) {    // check is relatively expensive
+        Location loc;
+        if (AP::ahrs().get_location(loc)) {
```
tridge, line 678: can use Polygon_closest_distance_point()

tridge (changes requested):

Tools/autotest/arducopter.py
```diff
@@ -1905,6 +1905,93 @@ def FenceFloorAutoEnableOnArming(self):
...
+        self.set_rc(2, 1100)
+
+        self.wait_statustext("Circle fence outside margin")
+        self.progress("Waiting for fence breach")
+        tstart = self.get_sim_time()
+        while not self.mode_is("RTL"):
+            if self.get_sim_time_cached() - tstart > 30:
+                raise NotAchievedException("Did not breach fence")
+
+            m = self.mav.recv_match(type='GLOBAL_POSITION_INT', blocking=True)
```
peterbarker, line 1959:
```suggestion
            m = self.assert_receive_message('GLOBAL_POSITION_INT')
```

... and elsewhere

rmackay9:
Let's be careful about spamming the user when in Loiter mode and the user pushes the vehicle against the fence

Tools/autotest/arducopter.py
```diff
@@ -1905,6 +1905,93 @@ def FenceFloorAutoEnableOnArming(self):
...
+        # fence requires home to be set:
+        m = self.poll_home_position(quiet=False)
+
+        ## 110m polyfence
+        home_loc = self.mav.location()
+        locs = [
+            mavutil.location(home_loc.lat - 0.0013, home_loc.lng - 0.0013, 0, 0),
+            mavutil.location(home_loc.lat - 0.0013, home_loc.lng + 0.0013, 0, 0),
+            mavutil.location(home_loc.lat + 0.0013, home_loc.lng + 0.0013, 0, 0),
+            mavutil.location(home_loc.lat + 0.0013, home_loc.lng - 0.0013, 0, 0),
```
peterbarker, line 1932:
```suggestion
            self.offset_location_ne(home_loc, -110, -110,
            self.offset_location_ne(home_loc, 110, -110,
            self.offset_location_ne(home_loc, 110, 110,
            self.offset_location_ne(home_loc, -110, 110,
```

peterbarker (changes requested):

libraries/AC_Fence/AC_Fence.cpp
timtuxworth: Sure it does - but your calculation of the margin is different. I'm saying you should use the same calculation. Your approximation will give different results.

Tools/autotest/arducopter.py
andyp1per: Fixed

Tools/autotest/arducopter.py
andyp1per: Fixed

libraries/AC_Fence/AC_Fence.cpp
andyp1per: Fixed

libraries/AP_Math/polygon.cpp
```diff
@@ -118,12 +118,33 @@ bool Polygon_complete(const Vector2<T> *V, unsigned n)
     return (n >= 4 && V[n-1] == V[0]);
 }
 
+/*
+  return the closest distance that point p comes to an edge of closed
+  polygon V, defined by N points
+ */
+template <typename T>
```
peterbarker, line 125: Shouldn't move this in the file

andyp1per:
Getting the maths right on this was actually quite hard. Think I might have it now.

Ryanf55:
> Getting the maths right on this was actually quite hard. Think I might have it now.

Thoughts on unit tests to prove they are right?

andyp1per:
> Thoughts on unit tests to prove they are right?

I added a unit test, did you not see it?

libraries/AP_Math/polygon.cpp
andyp1per: Fixed

andyp1per:
> Let's be careful about spamming the user when in Loiter mode and the user pushes the vehicle against the fence

So @rmackay9 what would you suggest here? If they have the option on and avoidance on and are in loiter mode, don't send the message?

Ryanf55:
> > Thoughts on unit tests to prove they are right?
> 
> I added a unit test, did you not see it?

Totally missed it. Those look great. I'm guessing the choice of sphere approximation over something like ellipsoid with Vincenty's distance is because fences are usually small enough and with not enough accuracy that we can just neglect the difference?

andyp1per:
> > > Thoughts on unit tests to prove they are right?
> > 
> > 
> > I added a unit test, did you not see it?
> 
> Totally missed it. Those look great. I'm guessing the choice of sphere approximation over something like ellipsoid with Vincenty's distance is because fences are usually small enough and with not enough accuracy that we can just neglect the difference?

Yeah, I was relying on the reference @tpwrules  sent me which shows that for small angles the haversine approximation is totally fine (the unit test demonstrates meter resolution). What wasn't fine was the original implementation which did not work at all for lat/lng 😛
It does make me wonder how accurate the current polyfence calculation is, but I haven't checked it.

Ryanf55:
> > > > Thoughts on unit tests to prove they are right?
> > > 
> > > 
> > > I added a unit test, did you not see it?
> > 
> > 
> > Totally missed it. Those look great. I'm guessing the choice of sphere approximation over something like ellipsoid with Vincenty's distance is because fences are usually small enough and with not enough accuracy that we can just neglect the difference?
> 
> Yeah, I was relying on the reference @tpwrules sent me which shows that for small angles the haversine approximation is totally fine (the unit test demonstrates meter resolution). What wasn't fine was the original implementation which did not work at all for lat/lng 😛 It does make me wonder how accurate the current polyfence calculation is, but I haven't checked it.

Yep, we just might want a note to not try to use fences to define boundaries the size of a country because you can be far enough off that you'll make the wrong people mad.

andyp1per:
I think you would have to have a margin the size of a country to run into this - which would be an odd thing to doSent from my iPhoneOn 26 Jan 2025, at 19:44, Ryan ***@***.***> wrote:﻿




Thoughts on unit tests to prove they are right?

I added a unit test, did you not see it?

Totally missed it. Those look great. I'm guessing the choice of sphere approximation over something like ellipsoid with Vincenty's distance is because fences are usually small enough and with not enough accuracy that we can just neglect the difference?

Yeah, I was relying on the reference @tpwrules sent me which shows that for small angles the haversine approximation is totally fine (the unit test demonstrates meter resolution). What wasn't fine was the original implementation which did not work at all for lat/lng 😛 It does make me wonder how accurate the current polyfence calculation is, but I haven't checked it.

Yep, we just might want a wiki note to not try to use fences to define boundaries the size of a country because you can be far enough off that you'll make the wrong people mad.

libraries/AP_Math/polygon.cpp
```diff
@@ -196,21 +196,121 @@ float Polygon_closest_distance_line(const Vector2f *V, unsigned N, const Vector2
...
+}
+
 /*
-  return the closest distance that point p comes to an edge of closed
-  polygon V, defined by N points
+  return the closest distance in meters that point p in lat/lng 1e7 comes to an edge of closed
+  polygon V, defined by N points of lat/lng 1e7
  */
-float Polygon_closest_distance_point(const Vector2f *V, unsigned N, const Vector2f &p)
+bool Polygon_closest_distance_point(const Vector2l *V, unsigned N, const Vector2l &p, ftype& closest)
```
tridge, line 290: should be able to use: ```
float Polygon_closest_distance_point(const Vector2f *V, unsigned N, const Vector2f &p)
```

**Review by tridge — COMMENTED** [review]
see if you can use the exiting function

File: libraries/AC_Fence/AC_Fence.cpp [hunk 8]
```diff
@@ -876,17 +895,54 @@ void AC_Fence::record_breach(uint8_t fence_type)
...
+{
+    // if we haven't already breached a margin limit, update the breach time
+    if (!(_breached_fence_margins & fence_type)) {
+        const uint32_t now = AP_HAL::millis();
+
+        // emit a message indicated we're newly-breached, but not too often
+        if (option_enabled(OPTIONS::MARGIN_BREACH)
+            && AP_HAL::timeout_expired(_last_margin_breach_notify_sent_ms, now, 1000U)) {
+            _last_margin_breach_notify_sent_ms = now;
+            print_fence_message("outside margin", _breached_fence_margins | fence_type);
```
tridge, line 909: i'm not sure "outside margin" and "inside margin" are clear, when I first read it I wondered if "outside margin" is "almost breached" or "all OK"
@rmackay9 ?

rmackay9: yes, we should try and be clear and we support both include and exclude fences.  Maybe we should used "breach margin" and "cleared margin"?

timtuxworth: > yes, we should try and be clear and we support both include and exclude fences. Maybe we should used "breach margin" and "cleared margin"?

Perhaps the other way around - "margin breach", "margin clear".

nexton-winjeel:
It's probably worth putting this new information in the [`FENCE_STATUS`](https://mavlink.io/en/messages/common.html#FENCE_STATUS) message.

andyp1per:
Using the Cartesian version of Polygon_closest_distance_point() saves about 1k, the overall cost is still 1k since this was not used before.

libraries/AP_Math/polygon.cpp
andyp1per: Done.

libraries/AC_Fence/AC_Fence.cpp
andyp1per: Ok I have changed to "xxx fence close" and "xxx fence cleared margin breach"

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -659,13 +665,23 @@ bool AC_Fence::check_fence_polygon()
     }
 
     const bool was_breached = _breached_fences & AC_FENCE_TYPE_POLYGON;
-    if (_poly_loader.breached()) {
+
+    Location loc;
+    bool have_location = AP::ahrs().get_location(loc);
```
peterbarker, line 670:
```suggestion
    const bool have_location = AP::ahrs().get_location(loc);
```

libraries/AP_Math/polygon.cpp
```diff
@@ -196,21 +196,154 @@ float Polygon_closest_distance_line(const Vector2f *V, unsigned N, const Vector2
...
+// Converts (latitude, longitude) in radians to 3D Cartesian coordinates on a unit sphere
+static void latlng_to_cartesian(ftype latRad, ftype lonRad, Vector3F& v)
+{
+    v.x = cosF(latRad) * cosF(lonRad);
+    v.y = cosF(latRad) * sinF(lonRad);
+    v.z = sinF(latRad);
+}
+
+// Haversine distance between two points given in 1e7 degrees
+static ftype haversine(long lat1_1e7, long lon1_1e7, long lat2_1e7, long lon2_1e7)
```
tridge, line 216: unused code?

peterbarker (changes requested):
Need to remove the new unused code

Need performance numbers

Move the in-margin stuff to a thread?

tridge:
we need the worst case CPU cost, and quite possibly move to a IO callback, maybe at 20Hz or so

libraries/AP_Math/polygon.cpp
andyp1per: Removed

libraries/AC_Fence/AC_Fence.cpp
andyp1per: Fixed

andyp1per:
Performance testing on a CubeOrange:
```
|--|--|--|
|None|4us|25Hz|
|20 point|22us|25Hz|
|40 point|30us|25Hz|
|60 point|45us|25Hz|
```
Could not go any higher because of storage space

tridge:
@andyp1per I think we should consider an IO callback, users can set BRD_SD_FENCE to get up to 8k fence points

andyp1per:
I retested with 250 points and the io callback. Fence check time is 130-230us but main loop delay is a few us.

libraries/AP_Vehicle/AP_Vehicle.cpp
```diff
@@ -1149,6 +1150,13 @@ bool AP_Vehicle::block_GCS_mode_change(uint8_t mode_num, const uint8_t *mode_lis
 }
 #endif
 
+#if AP_FENCE_ENABLED
+void AP_Vehicle::fence_init()
+{
+    hal.scheduler->register_io_process(FUNCTOR_BIND_MEMBER(&AP_Vehicle::fence_run_checks, void));
+}
+#endif
```
peterbarker, line 1158:
```suggestion
#endif  // AP_FENCE_ENABLED
```

tridge (commented):
thanks for implementing the IO callback. Do we need a mutex for it?

Rover/fence.cpp
```diff
@@ -1,26 +1,51 @@
...
+    fence_breaches.last_check_ms = now;
     const uint8_t orig_breaches = fence.get_breaches();
+    // check for new breaches; new_breaches is bitmask of fence types breached
+    fence_breaches.new_breaches = fence.check();
 
-    // check for a breach
-    new_breaches = fence.check();
+    if (!fence_breaches.new_breaches && orig_breaches && fence.get_breaches() == 0) {
+        // record clearing of breach
+        LOGGER_WRITE_ERROR(LogErrorSubsystem::FAILSAFE_FENCE, LogErrorCode::ERROR_RESOLVED);
```
peterbarker, line 25:
```suggestion
        LOGGER_WRITE_ERROR(LogErrorSubsystem::FAILSAFE_FENCE, 0);
```

... the enumeration value should be considered opaque in terms of its value.

peterbarker (dismissed):

Rover/fence.cpp
andyp1per: This is not code I wrote, it pre-exists this PR - I'd really like to avoid scope-creeping PR's if possible.

ArduCopter/fence.cpp
```diff
@@ -4,29 +4,57 @@
...
 #if AP_FENCE_ENABLED
 
-// fence_check - ask fence library to check for breaches and initiate the response
-// called at 1hz
-void Copter::fence_check()
+// async fence checking io callback at 1Khz
+void Copter::fence_run_checks()
 {
-    const uint8_t orig_breaches = fence.get_breaches();
+    uint32_t now = AP_HAL::millis();
```
peterbarker, line 10:
```suggestion
    const uint32_t now = AP_HAL::millis();
```

andyp1per: fixed

libraries/AC_Fence/AC_Fence.h
```diff
@@ -178,6 +184,7 @@ class AC_Fence
     enum class OPTIONS {
         DISABLE_MODE_CHANGE = 1U << 0,
         INCLUSION_UNION = 1U << 1,
+        MARGIN_BREACH = 1U << 2,
```
rmackay9, line 187: maybe this should be "NOTIFY_MARGIN_BREACH"

ArduCopter/fence.cpp
```diff
@@ -4,29 +4,57 @@
...
+    fence_breaches.new_breaches = fence.check(is_landing_or_landed);
+
+    if (!fence_breaches.new_breaches && orig_breaches && fence.get_breaches() == 0) {
+        if (!copter.ap.land_complete) {
+            GCS_SEND_TEXT(MAV_SEVERITY_NOTICE, "Fence breach cleared");
+        }
+        // record clearing of breach
+        LOGGER_WRITE_ERROR(LogErrorSubsystem::FAILSAFE_FENCE, LogErrorCode::ERROR_RESOLVED);
+    }
+    fence_breaches.have_updates = true; // new breache status latched so main loop will now pick it up
```
rmackay9, line 34: nitpick: "breache"

rmackay9 (approved):

tridge:
looks pretty good, I think it needs more testinging. @peterbarker, @MichelleRos and I should test this weekend, last weekend it was rained out

ArduCopter/fence.cpp
```diff
@@ -4,29 +4,57 @@
 
 #if AP_FENCE_ENABLED
 
-// fence_check - ask fence library to check for breaches and initiate the response
-// called at 1hz
-void Copter::fence_check()
+// async fence checking io callback at 1Khz
+void Copter::fence_run_checks()
```
rmackay9, line 8: I'm not a huge fan of the names, "fence_run_checks" vs "fence_checks".  these two names are different but they don't really make it clear what's different about them.  So for example, "fence_checks_async" would be better.  Other names might also be fine, this is just a suggestion

andyp1per: Fixed thanks

ArduCopter/fence.cpp
andyp1per: Fixed thanks

andyp1per:
I have changed the warning to print out the distance. There were also some wrong checks which meant the warning would only be printed out on margin breach rather than continuously

PR #28842, opened 2024-12-11
AP_HAL_ESP32: add targets to measure static memory sizes

srmainwaring:
Add targets from `${IDF_PATH}/tools/cmake/project.cmake` to measure the allocation of static memory.

These targets: `size`, `size-components`, `size-files`, may be used in the board build directory. For example if `board=esp32s3empty`:

```bash
$ cd ./build/esp32s3empty/esp-idf_build
$ ninja -v -v size

Total sizes:
Used stat D/IRAM:  243738 bytes (  98022 remain, 71.3% used)
      .data size:   14796 bytes
      .bss  size:  130256 bytes
      .text size:   97659 bytes
   .vectors size:    1027 bytes
Used Flash size : 1122994 bytes
           .text:  925938 bytes
         .rodata:  196800 bytes
Total image size: 1236476 bytes (.bin may be padded larger)
```

### DIFF
--- a/libraries/AP_HAL_ESP32/targets/esp32/esp-idf/CMakeLists.txt
+++ b/libraries/AP_HAL_ESP32/targets/esp32/esp-idf/CMakeLists.txt
@@ -123,3 +123,76 @@ idf_build_executable(${elf_file})
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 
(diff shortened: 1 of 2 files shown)


tpwrules (approved):
Does what it says on the tin. Uploading still works.

davidbuzz (commented):
i'm a bit late but i did look at this.  :-)

Issue #28841, opened 2024-12-11
Heli should not allow users to configure throttle-based dynamic notch


MattKear:
## Feature request

**Is your feature request related to a problem? Please describe.**
Heli should not allow users to configure throttle-based dynamic notch.  "Throttle" is not related to motor RPM in anyway in Heli.  "Throttle" from AP_MotorsHeli is actually collective position (necessary for the integration into the generic AP_Motors Class).  Most users do not understand this, so we should protect them against this miss-config.

**Describe the solution you'd like**
Throw a warning message when throttle-based dynamic notch is configured on Heli.

**Platform**
[  ] All
[  ] AntennaTracker
[ x ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

PR #28843, opened 2024-12-11
Do not allow users to set throttle-based notch in Heli 

MattKear:
My attempt at fixing issue: #28841

Throttle is not "Throttle" in AP_MotorsHeli, it is actually collective position.  Therefore, throttle-based tracking does not make any sense for Heli.

This is a subtly that most users will not appreciate, therefore we should help users to not make this mistake by throwing a config error if selected.

Have also changed the default notch mode for heli.

Tested in SITL only.  This is how it presents to the user upon miss-configuration:

image

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor.cpp
@@ -1006,6 +1006,16 @@ AP_InertialSensor::init(uint16_t loop_rate)
 #endif
 
 #if AP_INERTIALSENSOR_HARMONICNOTCH_ENABLED
+
+#if APM_BUILD_TYPE(APM_BUILD_Heli)
+    // Throttle tracking does not make sense with heli because "throttle" is actually collective position in AP_MotorsHeli
+    for (auto &notch : harmonic_notches) {
+        if (notch.params.enabled() && notch.params.tracking_mode() == HarmonicNotchDynamicMode::UpdateThrottle) {
+            AP_BoardConfig::config_error("Throttle notch unavailable with heli");
+        }
+    }
+#endif
+
     // the center frequency of the harmonic notch is always taken from the calculated value so that it can be updated
     // dynamically, the calculated value is always some multiple of the configured center frequency, so start with the
     // configured value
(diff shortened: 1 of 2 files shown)


libraries/Filter/HarmonicNotchFilter.cpp
```diff
@@ -48,6 +49,13 @@
  */
 #define NOTCHFILTER_ATTENUATION_CUTOFF 0.25
 
+#if APM_BUILD_TYPE(APM_BUILD_Heli)
+    // We cannot use throttle based notch on helis
+    #define NOTCHFILTER_DEFAULT_MODE 0 // fixed
+#else
+    #define NOTCHFILTER_DEFAULT_MODE 1 // throttle based
```
andyp1per, line 56: Think this should be int8_t(HarmonicNotchDynamicMode::UpdateThrottle) etc

andyp1per (changes requested):
LGTM, just would like to see the types called out explicitly

libraries/Filter/HarmonicNotchFilter.cpp
IamPete1: All defaults are floats.

bnsgeyer (approved):
@MattKear Thanks for this PR.  LGTM

IamPete1 (approved):
LGTM

rmackay9 (approved):

tridge (approved):

libraries/AP_InertialSensor/AP_InertialSensor.cpp
```diff
@@ -1006,6 +1006,16 @@ AP_InertialSensor::init(uint16_t loop_rate)
 #endif
 
 #if AP_INERTIALSENSOR_HARMONICNOTCH_ENABLED
+
+#if APM_BUILD_TYPE(APM_BUILD_Heli)
+    // Throttle tracking does not make sense with heli because "throttle" is actually collective position in AP_MotorsHeli
+    for (auto &notch : harmonic_notches) {
+        if (notch.params.enabled() && notch.params.tracking_mode() == HarmonicNotchDynamicMode::UpdateThrottle) {
+            AP_BoardConfig::config_error("Throttle notch unavailable with heli");
```
rmackay9, line 1014: Upon further though, I think it might have been better if this had been made an arming check.  With a config_error the main loop won't run and the user will find the HUD doesn't update.  Sadly this will only happen the next time that they reboot the board so they may not remember that they changed the harmonic notch method

PR #28848, opened 2024-12-12
SITL: add ENABLED defines for remaining simulated I2C devices

srmainwaring:
Adds conditional compilation defines for the remainder of the I2C sensors included in  `SIM_I2C.cpp`.

### Motivation

Aside from consistency, this reduces DRAM use for SIM on HW (esp. esp32empty and esp32s3empty where RAM limits are causing WiFi to not initialise correctly).

```bash
cd ./build/esp32s3empty/esp-idf_build
ninja -v -v size-files | grep SIM_I2C.cpp
```

#### Before:

| Object File | DRAM .data | .bss | IRAM .text | RAM Total | Flash .text | .rodata | Flash Total |
| --- | --- | --- | --- | --- | --- | --- | --- | 
| SIM_I2C.cpp.0.o | 116 | 20620 | 0 | 20736 | 867 | 184 | 1167 |

#### After:

| Object File | DRAM .data | .bss | IRAM .text | RAM Total | Flash .text | .rodata | Flash Total |
| --- | --- | --- | --- | --- | --- | --- | --- | 
| SIM_I2C.cpp.0.o | 36 | 3984 | 0 | 4020 | 71 | 24 | 131 |

See also:

- https://github.com/ArduPilot/ardupilot/pull/28842


libraries/SITL/SIM_Airspeed_DLVR.h
```diff
@@ -2,6 +2,14 @@
 
 #include "SIM_I2CDevice.h"
 
+#include <AP_HAL/AP_HAL_Boards.h>
+
+#ifndef AP_SIM_AIRSPEEDDLVR_ENABLED
```
peterbarker, line 7: These should go in `SIM_config.h`

Also, please add underscores in each of them:

```suggestion
#ifndef AP_SIM_AIRSPEED_DLVR_ENABLED
```

peterbarker (changes requested):

libraries/SITL/SIM_Airspeed_DLVR.h
srmainwaring: Thanks for the review. Updated in 92b8e86eb1546be8c7ff28839fe0905c642b802d

libraries/SITL/SIM_Temperature_TSYS01.cpp
```diff
@@ -1,5 +1,7 @@
 #include "SIM_Temperature_TSYS01.h"
 
+#if AP_SIM_TEMPERATURE_TSYS01_ENABLED
```
peterbarker, line 3: Our pattern is to include `SIM_config.h` and then check this, including the .h corresponding to the .cpp after that check has passed.

peterbarker (approved):

peterbarker:
Merged, thanks!

libraries/SITL/SIM_Temperature_TSYS01.cpp
srmainwaring: Fixed in follow up here: https://github.com/ArduPilot/ardupilot/pull/28871

PR #28852, opened 2024-12-13
AP_BattMonitor: make INA3221 parameters visible for metadata gen

tpwrules:
Not sure if more files need to be added for other battery monitor types? I suppose as long as the file list has the union of all possible parameters... It doesn't look like the metadata generator deduplicates anyway, c.f. `BATTn_I2C_BUS`.

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor.cpp
@@ -69,6 +69,12 @@ const AP_Param::GroupInfo AP_BattMonitor::var_info[] = {
     // @Path: AP_BattMonitor_INA2xx.cpp
     // @Group: _
     // @Path: AP_BattMonitor_ESC.cpp
+    // @Group: _
+    // @Path: AP_BattMonitor_INA239.cpp
+    // @Group: _
+    // @Path: AP_BattMonitor_INA3221.cpp
+    // @Group: _
+    // @Path: AP_BattMonitor_AD7091R5.cpp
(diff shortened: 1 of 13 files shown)


Hwurzburg (approved):

IamPete1 (changes requested):
Sorry, should have spotted this on the original PR. The new monitors param indexes conflict with the AD7091 eg: https://github.com/ArduPilot/ardupilot/blob/684d95a8260f2d9adf694a06873911f90e0e4b6c/libraries/AP_BattMonitor/AP_BattMonitor_AD7091R5.cpp#L87

https://github.com/ArduPilot/ardupilot/blob/684d95a8260f2d9adf694a06873911f90e0e4b6c/libraries/AP_BattMonitor/AP_BattMonitor_INA3221.cpp#L70

One of the odd things about battery monitor is that all the backends are loaded as though they share and single param table so none the params in the sub tables can use the same index. It seemed link a nice simplification at the time but now were running out of indexes.....

tpwrules:
I thought all monitors were supposed to use the same index range cause only one would be active at a time?

tpwrules:
Reallocated the indices and fixed up to make a conflict hopefully less likely in the future. It seems we are still decent on absolute count for a while yet as long as we do not allocate blocks.

Needs to be backported to 4.6 now so the INA3221 IDs do not desync.

IamPete1 (approved):
Great, thanks!

rmackay9:

tpwrules:
Turns out the INA3221 driver is not in 4.6, so that explains it! Removing the request for now, but this might come up again if any other battery driver stuff gets backported.

PR #28857, opened 2024-12-14
AP_Scripting: add bindings for servo telemetry

IamPete1:
Adds bindings for servo telem, this allows servo checks to be done in lua.

### DIFF
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -1395,9 +1395,6 @@ function camera:take_picture(instance) end
 ---@class (exact) AP_Camera__camera_state_t_ud
 local AP_Camera__camera_state_t_ud = {}
 
----@return AP_Camera__camera_state_t_ud
-function AP_Camera__camera_state_t() end
-
 -- get field
 ---@return Vector2f_ud
 function AP_Camera__camera_state_t_ud:tracking_p1() end
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -4015,7 +4015,6 @@ function networking:get_netmask_active() end
 function networking:get_ip_active() end
 
 -- visual odometry object
---@class visual_odom
 visual_odom = {}
 
 -- visual odometry health
@@ -4025,3 +4024,59 @@ function visual_odom:healthy() end
 -- visual odometry quality as a percentage from 1 to 100 or 0 if unknown
 ---@return integer
 function visual_odom:quality() end
+
+-- servo telemetry class
+servo_telem = {}
+
+-- get servo telem for the given servo number
+---@param servo_index integer -- 0 indexed servo number
+---@return AP_Servo_Telem_Data_ud|nil
+function servo_telem:get_telem(servo_index) end
+
+-- Servo telemtry userdata object
+---@class AP_Servo_Telem_Data_ud
+local AP_Servo_Telem_Data_ud = {}
+
+-- Get timestamp of last telem update
+---@return uint32_t_ud -- milliseconds since boot
+function AP_Servo_Telem_Data_ud:last_update_ms() end
+
+-- Get type spesfic status flags
+---@return integer|nil -- flags or nil if not available
+function AP_Servo_Telem_Data_ud:status_flags() end
+
+-- Get pcb temprature in centidegrees
+---@return integer|nil -- temperature in centidegrees or nil if not available
+function AP_Servo_Telem_Data_ud:pcb_temperature_cdeg() end
+
+-- Get motor temprature in centidegrees
+---@return integer|nil -- temperature in centidegrees or nil if not available
+function AP_Servo_Telem_Data_ud:motor_temperature_cdeg() end
+
+-- Get duty cycle
+---@return integer|nil -- duty cycle 0% to 100% or nil if not available
+function AP_Servo_Telem_Data_ud:duty_cycle() end
+
+-- get current
+---@return number|nil -- current in amps or nil if not available
+function AP_Servo_Telem_Data_ud:current() end
+
+-- get voltage
+---@return number|nil -- voltage in volts or nil if not available
+function AP_Servo_Telem_Data_ud:voltage() end
+
+-- get speed
+---@return number|nil -- speed in degrees per second or nil if not available
+function AP_Servo_Telem_Data_ud:speed() end
+
+-- get force
+---@return number|nil -- force in newton meters or nil if not available
+function AP_Servo_Telem_Data_ud:force() end
+
+-- get measured position
+---@return number|nil -- measured position in degrees or nil if not available
+function AP_Servo_Telem_Data_ud:measured_position() end
+
+-- get commanded position
+---@return number|nil -- comanded position in degrees or nil if not available
+function AP_Servo_Telem_Data_ud:command_position() end
--- a/libraries/AP_Scripting/generator/description/bindings.desc
+++ b/libraries/AP_Scripting/generator/description/bindings.desc
@@ -1051,14 +1051,14 @@ singleton AP_Servo_Telem method get_telem boolean uint8_t'skip_check AP_Servo_Te
 
 userdata AP_Servo_Telem::TelemetryData depends AP_SERVO_TELEM_ENABLED
 userdata AP_Servo_Telem::TelemetryData rename AP_Servo_Telem_Data
-userdata AP_Servo_Telem::TelemetryData field command_position float'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::COMMANDED_POSITION
-userdata AP_Servo_Telem::TelemetryData field measured_position float'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::MEASURED_POSITION
-userdata AP_Servo_Telem::TelemetryData field force float'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::FORCE
-userdata AP_Servo_Telem::TelemetryData field speed float'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::SPEED
-userdata AP_Servo_Telem::TelemetryData field voltage float'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::VOLTAGE
-userdata AP_Servo_Telem::TelemetryData field current float'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::CURRENT
-userdata AP_Servo_Telem::TelemetryData field duty_cycle uint8_t'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::DUTY_CYCLE
-userdata AP_Servo_Telem::TelemetryData field motor_temperature_cdeg int16_t'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::MOTOR_TEMP
-userdata AP_Servo_Telem::TelemetryData field pcb_temperature_cdeg int16_t'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::PCB_TEMP
-userdata AP_Servo_Telem::TelemetryData field status_flags uint8_t'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::STATUS
+userdata AP_Servo_Telem::TelemetryData field command_position float'skip_check read valid_mask present_types AP_Servo_Telem::TelemetryData::Types::COMMANDED_POSITION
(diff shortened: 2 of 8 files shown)


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4013,3 +4012,58 @@ function visual_odom:healthy() end
 -- visual odometry quality as a percentage from 1 to 100 or 0 if unknown
 ---@return integer
 function visual_odom:quality() end
+
+
+-- servo telemetry class
+---@class servo_telem
```
IamPete1, line 4018: Don't need this.

tridge (dismissed):

IamPete1:
For refence attempt 1 cost 1500 bytes or so. 
image

IamPete1:
This now get the structure and adds auto generation of a bitmask check and set. This means the script writer does not need to know there is a validity bitmask in the structure. This is currently read only, but it also adds auto setting of the mask when a value is set. The only slight problem is that once set there is no way to un-set the mask from lua. 

One disadvantage to this method is that were stuck with the types in the structure, previously I had returned the temp in degrees by applying the scale in the getter, now were stuck with centi degrees.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -1395,9 +1395,6 @@ function camera:take_picture(instance) end
 ---@class (exact) AP_Camera__camera_state_t_ud
 local AP_Camera__camera_state_t_ud = {}
 
----@return AP_Camera__camera_state_t_ud
-function AP_Camera__camera_state_t() end
-
```
IamPete1, line 1400: This should not have been in the docs generation, the function is not generated on the C++ side. We do not need it because all fields are read only and there are no methods or operators. The docs generator has been fixed to match the C++ generator.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4011,7 +4008,6 @@ function networking:get_netmask_active() end
 function networking:get_ip_active() end
 
 -- visual odometry object
---@class visual_odom
```
IamPete1, line 4014: This is incorrect markup. It should be `---@class`, however the class mark up is for user data objects only and should not be used on singletons. The error was overlooked due to the typo.

IamPete1:
Example of the new docs markup:
```
userdata AP_Servo_Telem::TelemetryData field command_position float'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::COMMANDED_POSITION
```
First bit is the same, the stuff added is:
```
valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::COMMANDED_POSITION
```
"valid_mask" is the keyword to enable the bitmask validity check.
"valid_types" is the variable name on the "AP_Servo_Telem::TelemetryData" object. 
"AP_Servo_Telem::TelemetryData::Types::COMMANDED_POSITION" is the bitmask value that the mask should be checked against.

IamPete1:
Now only costs  about 1K, so saves 400 bytes over the individual getters. 

image

tpwrules:
I haven't looked in detail, but it seems this can't absolutely fix the race condition because there isn't a lock protecting the telemetry data. The race window is much shorter now however, so maybe it's fixed enough? It was said in the call that there's servos which output different data at different rates anyway, so maybe comparing commanded vs. achieved just isn't possible in general.

Also still curious what the size impact would be if you had one mega function which returned eleven values.

rmackay9 (approved):
Looks OK to me but @tpwrules will review more thoroughly

tpwrules (changes requested):
A couple little things in the generator but should be ready to go after!

I tested in SITL using quadplane-can and the position from four servos comes through but it's the same for all four of them. I checked that the data I could get from Lua matched the data logged in CSRV, and it all did so this PR is good, but there's still probably something funky here.

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -1041,3 +1041,23 @@ include AP_TemperatureSensor/AP_TemperatureSensor.h
...
+
+include AP_Servo_Telem/AP_Servo_Telem.h depends AP_SERVO_TELEM_ENABLED
+include AP_Servo_Telem/AP_Servo_Telem_config.h
+singleton AP_Servo_Telem depends AP_SERVO_TELEM_ENABLED
+singleton AP_Servo_Telem rename servo_telem
+singleton AP_Servo_Telem method get_telem boolean uint8_t'skip_check AP_Servo_Telem::TelemetryData'Null
+
+userdata AP_Servo_Telem::TelemetryData depends AP_SERVO_TELEM_ENABLED
+userdata AP_Servo_Telem::TelemetryData rename AP_Servo_Telem_Data
+userdata AP_Servo_Telem::TelemetryData field command_position float'skip_check read valid_mask valid_types AP_Servo_Telem::TelemetryData::Types::COMMANDED_POSITION
```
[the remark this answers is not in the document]
IamPete1: Yeah, they should all be skip. But it doesn't matter in this case because there read only. Otherwise I think we should have got a error. 

I will fix.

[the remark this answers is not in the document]
IamPete1: fixed.

libraries/AP_Scripting/generator/src/main.c
```diff
@@ -2578,20 +2606,14 @@ void emit_loaders(void) {
...
 
-void emit_userdata_new_funcs(void) {
-  struct userdata *data = parsed_userdata;
-  fprintf(source, "const struct userdata {\n");
-  fprintf(source, "    const char *name;\n");
-  fprintf(source, "    const lua_CFunction fun;\n");
-  fprintf(source, "} new_userdata[] = {\n");
-  while (data) {
+int should_emit_creation(struct userdata * data) {
     // Dont expose creation function for all read only items
```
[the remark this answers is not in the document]
IamPete1: done.

libraries/AP_Scripting/generator/src/main.c
```diff
@@ -2601,8 +2623,18 @@ void emit_userdata_new_funcs(void) {
...
 
-    if (expose_creation) {
+void emit_userdata_new_funcs(void) {
+  struct userdata *data = parsed_userdata;
+  fprintf(source, "const struct userdata {\n");
+  fprintf(source, "    const char *name;\n");
+  fprintf(source, "    const lua_CFunction fun;\n");
+  fprintf(source, "} new_userdata[] = {\n");
+  while (data) {
+    // Dont expose creation function for all read only items
```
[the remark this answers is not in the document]
IamPete1: removed

libraries/AP_Scripting/generator/src/main.c
```diff
@@ -2885,32 +2917,33 @@ void emit_docs(struct userdata *node, int is_userdata, int emit_creation) {
...
-      // local userdata
-      fprintf(docs, "local %s = {}\n\n", name);
-
-      int creation_disabled = (node->creation && node->creation_args == -1);
-      if (emit_creation && (!node->creation || !creation_disabled)) {
-        // creation function
-        if (node->creation != NULL) {
-          for (int i = 0; i < node->creation_args; ++i) {
-            fprintf(docs, "---@param param%i UNKNOWN\n", i+1);
+      if (should_emit_creation(node)) {
```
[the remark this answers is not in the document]
IamPete1: Yeah, I have re-worked to simplify.

libraries/AP_Scripting/generator/src/main.c
```diff
@@ -2922,11 +2955,14 @@ void emit_docs(struct userdata *node, int is_userdata, int emit_creation) {
     if (node->fields != NULL) {
       struct userdata_field *field = node->fields;
       while(field) {
+          int valid_mask = (field->type.valid_mask.name != NULL) & (field->type.valid_mask.value != NULL);
```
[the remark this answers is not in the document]
IamPete1: done

libraries/AP_Scripting/generator/src/main.c
```diff
@@ -2922,11 +2955,14 @@ void emit_docs(struct userdata *node, int is_userdata, int emit_creation) {
...
+          const char * return_postfix = valid_mask ? "|nil\n" : "*\n";
```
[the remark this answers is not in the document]
IamPete1: typo.

libraries/AP_Scripting/docs/docs.lua
tpwrules: I found another tiny similar bug where we were accidentally generating `lua_new_X` functions that were never called. I folded that fix into this commit.

tpwrules (approved):
Tested and still works, thanks for the fixes.

I cleaned up the history slightly and folded in another really tiny fix to the generator. Figured it was easier to do it myself then to go through review again.

tpwrules:
Concerned whether servo telemetry needs a fix similar to https://github.com/ArduPilot/ardupilot/pull/28999 but that doesn't impact this PR at all.

libraries/AP_Scripting/docs/docs.lua
IamPete1: @tpwrules I disagree with your changes. We can emit them and rely on the linker to not include them if there not used. It seems it does still build, but you might want to use one of the functions in a manual binding which the generator does not know about. What is the benefit of not generating them?

tpwrules: It was my understanding that those functions were specifically designed to be called by Lua, e.g. because of the warnings and the use of the stack. If a manual binding wanted to create a new object, it would use the `new_X` function.

You are right that there is no grave harm in generating them though but I don't think a manual binding using them makes a lot of sense.

tpwrules: In case it wasn't clear here are the two places I added a new `should_emit_creation()` call to cause the change in behavior:
* https://github.com/ArduPilot/ardupilot/blob/e72720f2bffd5abda7908e927032f1adfcf707b2/libraries/AP_Scripting/generator/src/main.c#L1346
* https://github.com/ArduPilot/ardupilot/blob/e72720f2bffd5abda7908e927032f1adfcf707b2/libraries/AP_Scripting/generator/src/main.c#L1431

IamPete1: Your right, we don't need those.

libraries/AP_Servo_Telem/AP_Servo_Telem.cpp
```diff
@@ -152,6 +152,38 @@ void AP_Servo_Telem::write_log()
...
+    }
+
+    // Check if data has ever been received for the servo index provided
+    if ((active_mask & (1U << servo_index)) == 0) {
+        return false;
+    }
+
+    const volatile TelemetryData &telem_data = _telem_data[servo_index];
+
+    // Because the structure is volatile we have to copy element wise
```
tridge, line 170: we could just cast to override the volatile?

tridge: or could memcpy?

IamPete1: I couldn't get casting to work. The compiler was too smart. `memcpy` may work....

tridge: try this: 
```
    telem = *const_cast<TelemetryData*>(&telem_data);
```

tpwrules: `memcpy(&telem, (const void*)&telem_data, sizeof(telem));` will work too. But it's not exactly kosher.

IamPete1: > try this:
> 
> ```
>     telem = *const_cast<TelemetryData*>(&telem_data);
> ```

I have changed to this. Thanks.

robertlong13:
> Concerned whether servo telemetry needs a fix similar to #28999 but that doesn't impact this PR at all.

AP_Servo_Telem works subtly differently and does not suffer the same issue. The getter functions return success if that telemetry field has _ever_ been reported, and it is up to the caller to check whether that data is stale. AP_ESC_Telem's getters return failure if the data has timed out (or if a race condition leads it to falsely conclude there has been a timeout).

`AP_Servo_Telem::stale` suffers the same potential race condition, but that function is only used by `AP_Servo_Telem::valid` and _that_ function is unused. The lua bindings use the `valid_mask` instead of `AP_Servo_Telem::valid`, which should probably be named to `present_mask` since it's the mask checked by `AP_Servo_Telem::present` and `AP_Servo_Telem::valid` requires things to be both present and fresh.

Perhaps that rename should be part of this PR though.

IamPete1:
> `AP_Servo_Telem::stale` suffers the same potential race condition, but that function is only used by `AP_Servo_Telem::valid` and _that_ function is unused. The lua bindings use the `valid_mask` instead of `AP_Servo_Telem::valid`, which should probably be named to `present_mask` since it's the mask checked by `AP_Servo_Telem::present`. `AP_Servo_Telem::valid` requires things to be both present and fresh.

Yep, your right. I have removed the unused functions and renamed `valid_types` to `present_types`

tridge (approved):

PR #28858, opened 2024-12-14
RC_Channel: add another 8 scripting aux functions

IamPete1:
Adds some more scripting aux functions for those with complicated scripts. Now you can send aux functions from the GCS were no longer limited the number of RC channels users have free.


peterbarker (approved):
LGTM

PR #28865, opened 2024-12-15
AP_HAL_ESP32: RCOutput: add support for brushed PWM mode

tpwrules:
This is needed to fly the Stampfly. It's a significant re-engineering from the in-master state and the brushed hack that was used previously (though I've now rebased my stampfly branch to include this for those who want to try).

I tried to match the (somewhat vague) semantics of the ChibiOS HAL with regards to groups and frequencies, and I tested that the PWM output in "normal" mode still doesn't work the same way on the Stampfly, but someone needs to test it more thoroughly.

### DIFF
--- a/libraries/AP_HAL_ESP32/RCOutput.cpp
+++ b/libraries/AP_HAL_ESP32/RCOutput.cpp
@@ -50,6 +50,8 @@ gpio_num_t outputs_pins[] = {};
 #endif
 
 #define MAX_CHANNELS ARRAY_SIZE(outputs_pins)
+static_assert(MAX_CHANNELS < 12, "overrunning _pending and safe_pwm"); // max for current chips
+static_assert(MAX_CHANNELS < 32, "overrunning bitfields");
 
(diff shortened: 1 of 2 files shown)


andyp1per (approved):
LGTM, although whether it works or not is really the key

tpwrules:
Pretty convinced that this is good now. Added a hobby servo to one of the Grove connectors on the stamp fly and was able to use it and the brushed flight motors with no problems. Even could control it in flight!

davidbuzz (approved):
logically this looks good to me, although if the testing was inadequate, I'm sure we'll find out soon enough.    I like this if for no other reason than its got comments that the old code didnt really have.      there's a few probably-not-necessary variable renames, but nothing I'm going to stop it going in for.    merge it!

srmainwaring (approved):
Tested on ESP32 S3 DevKitC with PWM ESC and servos (MOT_PWM_TYPE=0). These are working fine.

Am not familiar enough with MCPWM peripheral to carry out in depth review.

PR #28866, opened 2024-12-15
add sanity check for command-int conversion

peterbarker:
Replaces https://github.com/ArduPilot/ardupilot/pull/28790

This provides a useful return value to the GCS that they've passed in invalid lat/lng in their command-long positions.  Which they shouldn't be using anyway.

```
Board,plane
CubeOrange,0
Pixhawk1-1M,40
```


libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -5170,10 +5158,27 @@ void GCS_MAVLINK::convert_COMMAND_LONG_to_COMMAND_INT(const mavlink_command_long
...
-    out.x = convert_COMMAND_LONG_loc_param(in.param5, stores_location);
-    out.y = convert_COMMAND_LONG_loc_param(in.param6, stores_location);
+
+    // we can't convert NaN/INF to an integer:
+    if (!isfinite(in.param5) || !isfinite(in.param6)) {
+        return MAV_RESULT_DENIED;
+    }
+
+    // check parameters are valid latitude/longitude:
+    if (!check_lat(in.param5) || !check_lng(in.param6)) {
```
IamPete1, line 5168: Don't we only want to do this if the command stores in a location? Possibly that it is also in a global frame, These param are not always lat/lon.

peterbarker: Well, that was a brain outage.

Fixed, thanks!

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -5170,10 +5158,27 @@ void GCS_MAVLINK::convert_COMMAND_LONG_to_COMMAND_INT(const mavlink_command_long
     out.param2 = in.param2;
     out.param3 = in.param3;
     out.param4 = in.param4;
-    const bool stores_location = command_long_stores_location((MAV_CMD)in.command);
-    out.x = convert_COMMAND_LONG_loc_param(in.param5, stores_location);
-    out.y = convert_COMMAND_LONG_loc_param(in.param6, stores_location);
+
+    // we can't convert NaN/INF to an integer:
```
IamPete1, line 5162: We are supposed to deal with NaNs, it might be fine if the command is not using those params. We could convert from NaN which is no value for float Int32 max which is no value for int32 according to the spec.

peterbarker: Oof.  Hmmm.  Will think about this more in the morning.

Problem is that if we just squash them to zero them we may end up triggering the sanitise-using-a-default-location code....

IamPete1: The spec calls for INT32 max for invalid. (https://mavlink.io/en/messages/common.html#COMMAND_INT), so NaN should be converted to that. Then your new range check will bounce if it is a location.

peterbarker: OK, squashed to uint32_t as you suggest - seems the best way forward here.

Note that this will be a behavioural change - previously we would have treated these as 0.

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -5170,10 +5158,37 @@ void GCS_MAVLINK::convert_COMMAND_LONG_to_COMMAND_INT(const mavlink_command_long
...
+    if (command_long_stores_location((MAV_CMD)in.command)) {
+        // check parameters are valid latitude/longitude:
+        if (!check_lat(param5) || !check_lng(param6)) {
+            return MAV_RESULT_DENIED;
+        }
+
+        out.x = param5 * 1e7;
+        out.y = param6 * 1e7;
+    } else {
+        out.x = param5;
```
tridge, line 5186: we need to see if there may be any commands that relied on the previous zero, and now will get INT32_MAX, motor test is a possible issue

tridge (changes requested):
need to see if the change of behaviour is safe

IamPete1 (commented):
Code changes look good now. We do generally need to be better handling incoming commands with invalid values.

peterbarker:
The scope of checking whether supplied parameters which could be passed in as NaN from the GCS was too large / boring.

So I've experimented here with adding a more comprehensive mavlink-command parameter vetting mechanism here.

I've converted two commands over to using this mechanism, `MAV_CMD_DO_GRIPPER` and `MAV_CMD_DO_SPRAYER`.

The concept is that we don't call simple functions to handle these any more.  We instead allocate little objects which subclass `CommandIntHandler`.  The "run" method on that object (1) vets the parameters and then (2) runs the guts of the old handler function.  The vetting is simple - if a parameter is used by the sub-class (as indicated by a method returning a mask of used parameters) then it must NOT be NaN.  If a parameter is NOT used by the subclass then it MUST be NaN.  Naturally there's scope for adding a mask to indicate it's OK for a used parameter to be NaN (and then the handling function would need to handle that case).

Total cost at the moment is ~650 bytes.  Incremental cost per command moved to this arrangement appears to be ~60 bytes.

We seem to handle 145 mavlink commands ('though some of these might be mission-only):
```
pbarker@crun:~/rc/ardupilot(pr/convert-better-errors)$ git grep -o 'MAV_CMD_[A-Z0-9_]*'  | cut -f 2 -d ':' | sort | uniq | wc
    145     145    3755
pbarker@crun:~/rc/ardupilot(pr/convert-better-errors)$ 
```

```
pbarker@crun:~/rc/ardupilot(pr/convert-better-errors)$ git grep -o 'MAV_CMD_[A-Z0-9_]*'  | grep -v 'AP_Mission/' | cut -f 2 -d ':' | sort | uniq | wc
    140     140    3620
pbarker@crun:~/rc/ardupilot(pr/convert-better-errors)$ 
```

So for ~7.4kB we could start to reject commands for which we do not understand all of the field values (excepting bits in bitmasks we don't support).  However, that leads to a problem - how do you give new meaning to previously unused command parameters without making any GCS which uses those new meanings useless on older versions of the firmware which have this code in it?

PR #28867, opened 2024-12-15
waf: add defaults.parm support to esp32 using ROMFS

tpwrules:
And drop broken remnants of code using apj_tool to do it. Now just load the file at `libraries/AP_HAL_ESP32/hwdef/<board>/defaults.parm` until we get hwdef.dat support in.

According to tridge, apj_tool is deprecated anyway. It's difficult to use because the .bin has an XOR checksum and SHA256 checksum that would have to be patched up. So just switch to using the already-working ROMFS.

Tested on the stampfly.


rmackay9 (approved):
I'm not really qualified but it looks harmless so I think we should merge it

davidbuzz (commented):
i did review this, after it was merged. and was just gonna say merge-it.

PR #28868, opened 2024-12-15
AP_Volz_Protocol: move serial port init of Volz protocol into Volz library

peterbarker:
Avoids spreading information about the Volz protocol throughout the serialmanager code.

```
Durandal                       16     *           16      16                16     16     16
Hitec-Airspeed      0                                                                     
KakuteH7-bdshot                40     *           40      40                40     40     40
MatekF405                      -8     *           0       -8                0      -8     0
Pixhawk1-1M-bdshot             -8                 -8      0                 0      -8     -8
f103-QiotekPeriph   -8                                                                    
f303-Universal      -8                                                                    
revo-mini                      0      *           0       0                 -8     0      0
skyviper-v2450                                    0                                       
```


libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -676,6 +672,15 @@ uint32_t AP_SerialManager::find_baudrate(enum SerialProtocol protocol, uint8_t i
     return _state->baudrate();
 }
 
+void AP_SerialManager::set_default_baud(enum SerialProtocol protocol, uint8_t instance, uint32_t _baud)
```
IamPete1, line 675: I think this should be `set_and_default_baud`. That is exactly the AP Param call. It has both `set_default` which won't change the current value and `set_and_default` which does.

peterbarker: Yep, fixed.

IamPete1 (approved):
LGTM

PR #28871, opened 2024-12-15
SITL: follow standard pattern for including ENABLED guards

srmainwaring:
Follow up to https://github.com/ArduPilot/ardupilot/pull/28848 to use standard include pattern.


peterbarker (approved):
LGTM, thanks!

PR #28873, opened 2024-12-15
Stop making python an alias for python3 in environment setups

peterbarker:
These were equivalents for installing python-is-python3 packages on Ubuntu.

We really shouldn't need these now - esp. since we don't do this on Ubuntu any more.

If we are invoking Python now we should be doing so as "python3".

### DIFF
--- a/.github/workflows/cygwin_build.yml
+++ b/.github/workflows/cygwin_build.yml
@@ -187,7 +187,6 @@ jobs:
           PATH: /usr/bin:$(cygpath ${SYSTEMROOT})/system32
         shell: C:\cygwin\bin\bash.exe -eo pipefail '{0}'
         run: >-
-          ln -sf /usr/bin/python3.7 /usr/bin/python && ln -sf /usr/bin/pip3.7 /usr/bin/pip &&
           python3 -m pip install --progress-bar off empy==3.3.4 pexpect &&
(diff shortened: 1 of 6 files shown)


Ryanf55 (approved):
We should update the shebangs in mavproxy too for all the ones that have main()

peterbarker:
> We should update the shebangs in mavproxy too for all the ones that have main()

Sounds good - but this isn't the MAVProxy repo :-)

Ryanf55:
> > We should update the shebangs in mavproxy too for all the ones that have main()
> 
> Sounds good - but this isn't the MAVProxy repo :-)

Please clarify then: https://github.com/ArduPilot/ardupilot_wiki/pull/6371#issuecomment-2480734514

The wiki ambiguous

peterbarker:
> > > We should update the shebangs in mavproxy too for all the ones that have main()
> > 
> > 
> > Sounds good - but this isn't the MAVProxy repo :-)
> 
> Please clarify then: [ArduPilot/ardupilot_wiki#6371 (comment)](https://github.com/ArduPilot/ardupilot_wiki/pull/6371#issuecomment-2480734514)
> 
> The wiki ambiguous

https://github.com/ArduPilot/MAVProxy/pull/1497

PR #17804, opened 2021-06-18
Rover/GCS: send water depth and temperature

rmackay9:
This PR adds support for bots with downward facing rangefinders (assumed to be underwater sonar) to send the WATER_DEPTH message each time a new reading is received.  This message contains the depth, water temperature, location and attitude of the vehicle.  This is conceptually similar to camera feedback messages which are sent to the GCS each time a picture is taken

This has been lightly tested in SITL to ensure the message is only sent for Rovers with downward facing rangefinders.

PR #28875, opened 2024-12-15
Rover: allow WATER_DEPTH mavlink message rate to be specified

peterbarker:
... and reduce the default rate

this is currently unconditionally streamed at 50Hz, chewing up all available bandwidth on some telemetry radios.

### DIFF
--- a/Rover/GCS_Mavlink.cpp
+++ b/Rover/GCS_Mavlink.cpp
@@ -589,6 +589,7 @@ static const ap_message STREAM_EXTRA3_msgs[] = {
     MSG_WIND,
 #if AP_RANGEFINDER_ENABLED
     MSG_RANGEFINDER,
+    MSG_WATER_DEPTH,
 #endif
     MSG_DISTANCE_SENSOR,
     MSG_SYSTEM_TIME,
--- a/Rover/GCS_Mavlink.cpp
+++ b/Rover/GCS_Mavlink.cpp
@@ -199,9 +199,14 @@ void GCS_MAVLINK_Rover::send_water_depth() const
 
     RangeFinder *rangefinder = RangeFinder::get_singleton();
 
-    if (rangefinder == nullptr || !rangefinder->has_orientation(ROTATION_PITCH_270)){
+    if (rangefinder == nullptr) {
         return;
-    } 
+    }
+
+    // depth can only be measured by a downward-facing rangefinder:
+    if (!rangefinder->has_orientation(ROTATION_PITCH_270)) {
+        return;
+    }
 
     // get position
     const AP_AHRS &ahrs = AP::ahrs();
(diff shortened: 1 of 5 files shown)


rmackay9:
Here's the original PR that added support https://github.com/ArduPilot/ardupilot/pull/17804

rmackay9:
Hi @peterbarker,

This looks good, what testing has been done?

peterbarker:
> Hi @peterbarker,
> 
> This looks good, what testing has been done?

I found a convenient autotest, `Rover.DepthFinder` which seems to do all the testing required.  It actually picked up the fact my original patch wasn't filtering it to send only for boat frames!

I also did a general size-check for sanity:
```
Durandal                       *      *           *       *                 *      -32    *
Hitec-Airspeed      *                                                                     
KakuteH7-bdshot                *      *           *       *                 *      8      *
MatekF405                      *      *           *       *                 *      8      *
Pixhawk1-1M-bdshot             *                  *       *                 *      0      *
f103-QiotekPeriph   *                                                                     
f303-Universal      *                                                                     
revo-mini                      *      *           *       *                 *      8      *
skyviper-v2450                                    *                                       
```

rmackay9 (approved):

Rover/GCS_Mavlink.cpp
```diff
@@ -185,6 +185,65 @@ void GCS_MAVLINK_Rover::send_rangefinder() const
...
+    if (!rangefinder->has_orientation(ROTATION_PITCH_270)) {
+        return;
+    }
+
+    // get position
+    const AP_AHRS &ahrs = AP::ahrs();
+    Location loc;
+    IGNORE_RETURN(ahrs.get_location(loc));
+
+    for (uint8_t i=0; i<rangefinder->num_sensors(); i++) {
```
IamPete1, line 216: Should we cycle backends? Before this fix I have have the link be completely flooded on a vehicle with several rangefinders.

peterbarker: Yes, we should.

See last commit on here: https://github.com/ArduPilot/ardupilot/pull/28877

peterbarker:
This PR fixed the partner's problem.

Rover/GCS_Mavlink.cpp
```diff
@@ -1206,3 +1273,6 @@ uint8_t GCS_MAVLINK_Rover::send_available_mode(uint8_t index) const
 
     return mode_count;
 }
+
+#if AP_RANGEFINDER_ENABLED
+#endif
```
IamPete1, line 1278: Left over from a rebase?

peterbarker: Not in this case - I moved the send-water-depth-code to be next to the send-rangefinder-code as the `#if`s already existed and the methods are related....

Removed these lines.

PR #28879, opened 2024-12-16
Tradheli: offset collective from landed min when landed in non-manual collective modes

bnsgeyer:
This PR adds the ability for the user to offset the collective from the landed min collective position when landed in non-manual collective modes.  This will enable users to minimize vibrations when landed by being able to have the collective raised after the land complete flag is true.

### DIFF
--- a/ArduCopter/heli.cpp
+++ b/ArduCopter/heli.cpp
@@ -76,18 +76,23 @@ void Copter::check_dynamic_flight(void)
 void Copter::update_heli_control_dynamics(void)
 {
 
+    if (ap.land_complete) {
+        motors->set_land_complete(true);
+    } else {
+        motors->set_land_complete(false);
+    }
+    if (ap.land_complete_maybe) {
+        motors->set_land_complete_maybe(true);
+    } else {
+        motors->set_land_complete_maybe(false);
+    }
+    
     if (!motors->using_leaky_integrator()) {
         //turn off leaky_I
         attitude_control->use_leaky_i(false);
-        if (ap.land_complete || ap.land_complete_maybe) {
-            motors->set_land_complete(true);
-        } else {
-            motors->set_land_complete(false);
-        }
     } else {
         // Use Leaky_I if we are not moving fast
         attitude_control->use_leaky_i(!heli_flags.dynamic_flight);
-        motors->set_land_complete(false);
     }
 
     if (ap.land_complete || (is_zero(motors->get_desired_rotor_speed()))) {
--- a/libraries/AP_Motors/AP_MotorsHeli.cpp
+++ b/libraries/AP_Motors/AP_MotorsHeli.cpp
@@ -152,6 +152,15 @@ const AP_Param::GroupInfo AP_MotorsHeli::var_info[] = {
     // @Path: ../AC_Autorotation/RSC_Autorotation.cpp
     AP_SUBGROUPINFO(_main_rotor.autorotation, "RSC_AROT_", 33, AP_MotorsHeli, RSC_Autorotation),
 
+    // @Param: COL_LAND_OFSET
+    // @DisplayName: Offset of Collective Blade Pitch when Landed
+    // @Description: Offset of the collective blade pitch angle when landed in degrees for non-manual collective modes (i.e. modes that use altitude hold).
+    // @Range: -2 2
+    // @Units: deg
+    // @Increment: 0.1
+    // @User: Standard
+    AP_GROUPINFO("COL_LAND_OFSET", 34, AP_MotorsHeli, _collective_land_offset_deg, 0.0),
+
     AP_GROUPEND
 };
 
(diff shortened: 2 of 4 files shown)


ArduCopter/heli.cpp
```diff
@@ -76,18 +76,23 @@ void Copter::check_dynamic_flight(void)
 void Copter::update_heli_control_dynamics(void)
 {
 
+    if (ap.land_complete) {
+        motors->set_land_complete(true);
+    } else {
+        motors->set_land_complete(false);
+    }
```
MattKear, line 83:
```suggestion
        motors->set_land_complete(bool(ap.land_complete));
    }
```

libraries/AP_Motors/AP_MotorsHeli_Single.cpp
```diff
@@ -388,13 +390,23 @@ void AP_MotorsHeli_Single::move_actuators(float roll_out, float pitch_out, float
     }
 
     // ensure not below landed/landing collective
-    if (_heliflags.landing_collective && collective_out < _collective_land_min_pct && !_main_rotor.in_autorotation()) {
+    if (_heliflags.land_complete && _heliflags.landing_collective && collective_out < _collective_land_min_pct + _collective_offset_landed) {
+        _collective_offset_landed += 0.0001 * (_collective_land_offset_pct - _collective_offset_landed);
+        collective_out = _collective_land_min_pct + _collective_offset_landed;
```
MattKear, line 395: We should constrain this against zero thrust collective as a miss config this could lead to a collective oscillation on the ground with the aircraft taking off and landing constantly.....

libraries/AP_Motors/AP_MotorsHeli_Single.cpp
```diff
@@ -388,13 +390,23 @@ void AP_MotorsHeli_Single::move_actuators(float roll_out, float pitch_out, float
     }
 
     // ensure not below landed/landing collective
-    if (_heliflags.landing_collective && collective_out < _collective_land_min_pct && !_main_rotor.in_autorotation()) {
+    if (_heliflags.land_complete && _heliflags.landing_collective && collective_out < _collective_land_min_pct + _collective_offset_landed) {
+        _collective_offset_landed += 0.0001 * (_collective_land_offset_pct - _collective_offset_landed);
```
MattKear, line 394: As you have already identified in your comments, this needs to be a low pass filter object.

MattKear, line 394: Please add a comment to make it clear that this increases collective away from landed collective position.

MattKear (changes requested):
Based on my first look at this, I really don't like this feature.

I would like to understand more about the use case.  Currently I cannot understand why an operator would want the collective to reduce to land_col_min just to have it raise up again.  If someone asked me for this feature my initial reaction would be that they have simply set the landing collective position wrong if it causes undue vibration/loads on when landing in auto-collective modes.

I am fairly sure this introduces a circular dependency.  By adding _heliflags.land_complete into these checks where we update the motors limit.throttle_lower and the below_land_min_coll flags.  Both of these flags are used in the motor_at_lower_limit check in the landing detector for auto-collective modes.  See here:
https://github.com/MattKear/ardupilot/blob/2ce6532698477d7f3e7bc0ae9e18bf9178a94e0e/ArduCopter/land_detector.cpp#L79-L81

If it doesn't actually introduce a circular dependency then it makes it a really tricky piece of code that is hard to maintain to ensure that we don't in the future.  (I have been staring at this for a couple of hours now and I can't convince myself definitively either way that the is/isn't a circular dependency).

libraries/AP_Motors/AP_MotorsHeli.cpp
```diff
@@ -393,7 +402,7 @@ void AP_MotorsHeli::output_logic()
             // Servos should exhibit normal flight behavior.
 
             // set limits flags
-            if (_heliflags.land_complete && !using_leaky_integrator()) {
+            if ((_heliflags.land_complete || _heliflags.land_complete_maybe) && !using_leaky_integrator()) {
```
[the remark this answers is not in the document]
bnsgeyer: The reason did this was that the original implementation lumped land complete maybe and land complete into the same flag for the motors class.  You can see the change where I broke them out so they could be tracked separately.  Thus to keep the behavior the same, I had to add the extra conditional here.

libraries/AP_Motors/AP_MotorsHeli_Single.cpp
```diff
@@ -284,13 +284,15 @@ void AP_MotorsHeli_Single::calculate_scalars()
...
     _collective_land_min_deg.set(constrain_float(_collective_land_min_deg, _collective_min_deg, _collective_max_deg));
+    _collective_land_offset_deg.set(constrain_float(_collective_land_offset_deg, _collective_min_deg, _collective_max_deg));
 
     if (!is_equal((float)_collective_max_deg, (float)_collective_min_deg)) {
         // calculate collective zero thrust point as a number from 0 to 1
         _collective_zero_thrust_pct = (_collective_zero_thrust_deg-_collective_min_deg)/(_collective_max_deg-_collective_min_deg);
 
         // calculate collective land min point as a number from 0 to 1
         _collective_land_min_pct = (_collective_land_min_deg-_collective_min_deg)/(_collective_max_deg-_collective_min_deg);
+        _collective_land_offset_pct = _collective_land_offset_deg/(_collective_max_deg-_collective_min_deg);
```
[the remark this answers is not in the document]
bnsgeyer: This is just an offset.  Just calculating it as a percent of the entire range

PR #21861, opened 2022-10-03
remove Copter auto_trim functionality

peterbarker:

PR #28881, opened 2024-12-16
Copter: use Auxiliary function to control adjusting of AHRS trim from RC inputs

peterbarker:
This PR replaces https://github.com/ArduPilot/ardupilot/pull/21861 , which removed the "auto_trim" function entirely.  Turns out people do use the feature....

From that PR:

> Move this feature to an RC switch and remove the stick gesture initiation

Done, new RC aux option 182

>     Ask Michael Oborne to add an easy aux function page to MP to allow users to easily trigger this function (and any other > available function) without needing to setup and RC switch (https://github.com/ArduPilot/MissionPlanner/issues/2957)

Done long ago

>    Allow this feature to be run from AltHold

Done

>    maybe change the feature to be run in other flight modes (like Loiter) and take the trim adjust from the lean angle of the vehicle instead of from the pilot input. This would need to not be an instantaneous capture of the lean angles but instead a filtered value taken over 10sec

That's a significant change in behaviour, virtually unrelated to the RC-based option.  There's many questions around doing this sort of thing in loiter - estimatesd-position drift over time and wind come to mind.... perhaps a future enhancement :-)

New semantics are that if you flick the switch on while you are flying it will start to trim.  If the vehicle  lands, or you change modes to something other than STABILIZE/ALT_HOLD the trimming is stopped.  If you flick the switch back to low while flying then the trimming is stopped and the current trims saved.

I've also allowed the feature to be compiled out (about 384 bytes).  We should probably compiled it out of toy mode (skyviper) as it already has some magic stick calibration stuff in it.

I've written an autotest for it.

```
CubeRedPrimary                 *      *           -24     -16               *      *      *
Durandal                       *      *           -32     -64               *      *      *
Hitec-Airspeed      *                                                                     
KakuteH7-bdshot                *      *           8       8                 *      *      *
MatekF405                      *      *           0       -24               *      *      *
Pixhawk1-1M-bdshot             *                  -24     -24               *      *      *
f103-QiotekPeriph   *                                                                     
f303-Universal      *                                                                     
revo-mini                      *      *           8       -64               *      *      *
skyviper-journey                                  -16                                     
skyviper-v2450                                    24                                      
```

Needs a flight test.


ArduCopter/RC_Channel.cpp
```diff
@@ -703,68 +709,78 @@ void RC_Channel_Copter::do_aux_function_change_force_flying(const AuxSwitchPos c
...
+    float roll_trim = ToRad((float)get_roll_channel().get_control_in()/100.0f);
+    float pitch_trim = ToRad((float)get_pitch_channel().get_control_in()/100.0f);
+    AP::ahrs().add_trim(roll_trim, pitch_trim);
     LOGGER_WRITE_EVENT(LogEvent::SAVE_TRIM);
     gcs().send_text(MAV_SEVERITY_INFO, "Trim saved");
 }
 
+#if AP_COPTER_AHRS_AUTO_TRIM_ENABLED
+
+void RC_Channels_Copter::do_aux_function_ahrs_auto_trim(const RC_Channel::AuxSwitchPos ch_flag)
```
rmackay9, line 727: let's put a comment above this function to explain what it's doing.  Somethjing like, "start/stop ahrs auto trim"

ArduCopter/RC_Channel.h
```diff
@@ -46,6 +46,17 @@ class RC_Channels_Copter : public RC_Channels
     // returns true if throttle arming checks should be run
     bool arming_check_throttle() const override;
 
+    // support for trimming AHRS using RC stick inputs:
```
rmackay9, line 49: need to update the comment?  perhaps just replace "RC stick inputs:" with "an auxiliary function"

ArduCopter/RC_Channel.cpp
peterbarker: Comment added.

ArduCopter/RC_Channel.h
peterbarker: The comment is accurate - you do adjust the AHRS trim using the RC stick inputs!

I've added some words to indicate that the functionality is triggered by using an aux switch.

peterbarker:
@MichelleRos tested this on her 5 inch quad on the weekend.  Worked as expected.

ArduCopter/motors.cpp
```diff
@@ -51,15 +51,6 @@ void Copter::arm_motors_check()
             }
         }
 
-        // arm the motors and configure for flight
-        if (arming_counter == AUTO_TRIM_DELAY && motors->armed() && flightmode->mode_number() == Mode::Number::STABILIZE) {
```
IamPete1, line 55: `AUTO_TRIM_DELAY` define can be removed too.

rmackay9 (approved):

ArduCopter/motors.cpp
peterbarker: Done

MichelleRos (approved):
I'm ok with removing it from Blimp.

ArduCopter/RC_Channel_Copter.cpp
```diff
@@ -706,68 +712,78 @@ void RC_Channel_Copter::do_aux_function_change_force_flying(const AuxSwitchPos c
...
+        // only trim in certain modes:
+        if (copter.flightmode != &copter.mode_stabilize &&
+            copter.flightmode != &copter.mode_althold) {
+            auto_trim_cancel();
+            return;
         }
 
-        if (ap.land_complete) {
-            // landed again.
+        // must be started and stopped mid-air:
```
Hwurzburg, line 773: @peterbarker so this cannot be activated before takeoff? I assume land_complete_maybe is true while disarmed and thru arm until liftoff??

PR #28883, opened 2024-12-16
Tools: ardupilotwaf: esp32 does not have hardware double

srmainwaring:
There is no hardware double on esp32.


tpwrules:
I would prefer to see this in `AP_HAL/boards/esp32.h` like Linux and QURT.

davidbuzz (approved):
logincally correct, but its also ok to do it in AP_HAL/boards/esp32.h, i dont mind which, do we just merge this PR and move on .?

PR #28886, opened 2024-12-16
AP_Scripting, AC_PosControl: add follow-target-send.lua applet

rmackay9:
This adds a follow-target-send.lua applet which can send the [FOLLOW_TARGET](https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/common.xml#L5532) mavlink message to other vehicles.  This will be used in the cooperative lift and/or copter-ship-landing work.

This has been lightly tested in SITL and below is a screenshot from MP's MAVLink inspector sent from a vehicle flying North at 10m/s in Guided mode
follow-target-send-16Dec2024

I was hoping to test [Follow mode](https://ardupilot.org/copter/docs/follow-mode.html) with [multiple vehicles in SITL](https://ardupilot.org/dev/docs/using-sitl-for-ardupilot-testing.html#swarming-with-sitl) but it looks like mavlink messages are not passed between them.

BTW the mavlink spec for the [FOLLOW_TARGET](https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/common.xml#L5532) message is not too clear on the frame of each field.  I've assumed everything is NED

### DIFF
--- a/libraries/AP_Scripting/applets/follow-target-send.lua
+++ b/libraries/AP_Scripting/applets/follow-target-send.lua
@@ -1,0 +1,121 @@
+-- Send the FOLLOW_TARGET mavlink message to allow other vehicles to follow this one
+--
+-- How To Use
+-- 1. copy this script to the autopilot's "scripts" directory
+-- 2. within the "scripts" directory create a "modules" directory
+-- 3. copy the MAVLink/mavlink_msgs_xxx files to the "scripts" directory
+-- 4. the FOLLOW_TARGET message will be published at 10hz
+
+-- load mavlink message definitions from modules/MAVLink directory
+local mavlink_msgs = require("MAVLink/mavlink_msgs")
+
+-- global definitions
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+local UPDATE_INTERVAL_MS = 100                  -- update at about 10hz
+local FOLLOW_TARGET_CAPABILITIES = {POS=2^0, VEL=2^1, ACCEL=2^2, ATT_RATES=2^3}
+
+ -- setup script specific parameters
+local PARAM_TABLE_KEY = 88
+local PARAM_TABLE_PREFIX = "FOLT_"
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 2), 'could not add param table')
+
+-- add a parameter and bind it to a variable
+local function bind_add_param(name, idx, default_value)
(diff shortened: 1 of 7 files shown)


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3785,6 +3785,18 @@ function poscontrol:set_posvelaccel_offset(pos_offset_NED, vel_offset_NED, accel
 ---@return Vector3f_ud|nil
 function poscontrol:get_posvelaccel_offset() end
 
+-- get position controller's target velocity in m/s in NED frame
+---@return Vector3f_ud|nil
+---@return Vector3f_ud|nil
+---@return Vector3f_ud|nil
+function poscontrol:get_vel_target() end
+
+-- get position controller's acceleration velocity in m/s/s in NED frame
```
IamPete1, line 3794:
```suggestion
-- get position controller's target acceleration in m/s/s in NED frame
```

libraries/AP_Scripting/applets/follow-target-send.lua
```diff
@@ -0,0 +1,119 @@
...
+    -- get vehicle rotation rates
+    local curr_rot_rate = ahrs:get_gyro()
+    if curr_rot_rate == nil then
+        curr_rot_rate = Vector3f()
+    end
+    local curr_rot_rate_NED = ahrs:body_to_earth(curr_rot_rate)
+
+    -- prepare FOLLOW_TARGET message
+    local follow_target_msg = {}
+    follow_target_msg.timestamp = millis():toint() * 1000
```
IamPete1, line 85: I think millis is correct, no need for the * 1000?

We do support 64bit numbers now. We could add a binding for millis64.

rmackay9: thanks @IamPete1 for the review.  I thought of this but then wondered if the millis64 wrapped more quickly than millis()?  Not that most vehicles are powered up long enough to wrap but..

libraries/AP_Scripting/docs/docs.lua
rmackay9: thanks, fixed!

libraries/AP_Scripting/applets/follow-target-send.lua
rmackay9: oh my, you're right!  fixed!

rmackay9:
I've fixed up issues found by @IamPete1 (thanks!).

I've also change the script slightly so that the capabilities are filled in dynamically and change with the flight mode.  E.g. in Stabilize the capability is "9" meaning position, attitude and rates while in Guided it is 15 meaning position, vel target, accel target, attitude and rates

Finally I've also rebased on master

tridge (approved):

PR #28889, opened 2024-12-16
scripting: Added generator cutoff for EFI Skypower

Georacer:
This PR adds a parameter in the script of the Skypower EFI driver.

The parameter will cut off the generator when the engine target load is above that threshold.

This is meant to alleviate engine load when maximum mechanical power is required for thrust.


peterbarker:
Has this been tested and confirmed working?

peterbarker:
Ping @Georacer - we can put this in if it has been tested...

PR #28890, opened 2024-12-16
AP_Scripting: added QUIK_ANGLE_MAX

tridge:
This brings the same attitude error abort protection to the lua VTOL_quicktune script as we have in the C++ version

### DIFF
--- a/libraries/AP_Scripting/docs/docs.lua
+++ b/libraries/AP_Scripting/docs/docs.lua
@@ -3761,6 +3761,10 @@ AC_AttitudeControl = {}
 ---@return number -- yaw slew rate
 function AC_AttitudeControl:get_rpy_srate() end
 
+-- Return the angle between the target thrust vector and the current thrust vector in degrees.
+---@return number -- attitude error
+function AC_AttitudeControl:get_att_error_angle_deg() end
+
 -- desc
 AR_AttitudeControl = {}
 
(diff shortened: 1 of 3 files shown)


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3761,6 +3761,10 @@ AC_AttitudeControl = {}
 ---@return number -- yaw slew rate
 function AC_AttitudeControl:get_rpy_srate() end
 
+-- return current VTOL total attitude error in degrees
+---@return number -- attitude error
+function get_att_error_angle_deg() end
```
IamPete1, line 3766:
```suggestion
function AC_AttitudeControl:get_att_error_angle_deg() end
```

tridge: thanks, fixed

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -3761,6 +3761,10 @@ AC_AttitudeControl = {}
 ---@return number -- yaw slew rate
 function AC_AttitudeControl:get_rpy_srate() end
 
+-- return current VTOL total attitude error in degrees
```
IamPete1, line 3764: This is only roll and pitch, I wonder if we should just copy the function description exactly from attitude control.  That would be:

```
Return the angle between the target thrust vector and the current thrust vector in degrees.
```

IamPete1 (approved):
New binding looks good. Script changes look OK but I have not tested.

rmackay9:

PR #28891, opened 2024-12-16
ArduPlane: parameterize lookahead climb ratio

timtuxworth:
Plane altitude terrain lookahead currently uses a fairly arbitrary but safe 50% factor when calculating the maximum pitch to assume when calculating the if the plane can safely fly over the approaching terrain.

Some users want to be more aggressive than the default. so this makes the 50% factor into a parameter. The default is no-change to current behavior, but by changing TERRAIN_LKA_CLMB users can now chose a more (or less) aggressive factor for the percentage of maximum climb factor to use when calculating terrain lookahead.

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1309,7 +1309,15 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("RNGFND_LND_ORNT", 36, ParametersG2, rangefinder_land_orient, ROTATION_PITCH_270),
 #endif
-    
+
+    // @Param: TERRAIN_LKA_CLMB
+    // @DisplayName: terrain lookahead climb pitch ratio
+    // @Description: The ratio of climb rate to use when doing a terrain lookahead so the plane is not constantly at 100% throttle. 100% is the maximum climb rate which should give maximum ground clearance, and 0% should give the minimum climb rate and the minimum ground clearance, the default is 50% to try to balance safety vs battery usage. User TERRAIN_LOOKAHD=0 to disable lookahead.
+    // @Range: 0 100
+    // @Units: %
+    // @User: Standard
+    AP_GROUPINFO("TERRAIN_LKA_CLMB", 37, ParametersG2, terrain_lookahead_pitch_climb_rate, 50),
+
     AP_GROUPEND
 };
 
--- a/ArduPlane/altitude.cpp
+++ b/ArduPlane/altitude.cpp
@@ -602,10 +602,13 @@ float Plane::lookahead_adjustment(void)
         // we're not moving
         return 0;
     }
-    // we need to know the climb ratio. We use 50% of the maximum
+    // we need to know the climb ratio. We use TERRAIN_LKA_CLMB % of the maximum
     // climb rate so we are not constantly at 100% throttle and to
     // give a bit more margin on terrain
-    float climb_ratio = 0.5f * TECS_controller.get_max_climbrate() / groundspeed;
+    // we inverse the provided ratio because it gives the best result and is easiest for the user to understand
+    // maximum rate = maximum climb rate = maximum ground clearance
+    float pitch_ratio = constrain_float(1.0 - g2.terrain_lookahead_pitch_climb_rate * 0.01, .01, 1.0);
+    float climb_ratio = pitch_ratio * TECS_controller.get_max_climbrate() / groundspeed;
 
     if (climb_ratio <= 0) {
         // lookahead makes no sense for negative climb rates
--- a/ArduPlane/Parameters.h
+++ b/ArduPlane/Parameters.h
@@ -590,6 +590,8 @@ public:
     // orientation of rangefinder to use for landing
     AP_Int8 rangefinder_land_orient;
 #endif
+
+    AP_Float        terrain_lookahead_pitch_climb_rate;
 };
 
 extern const AP_Param::Info var_info[];


timtuxworth:
Hi George - could you take a look? Any suggestions?

ArduPlane/altitude.cpp
```diff
@@ -605,7 +605,7 @@ float Plane::lookahead_adjustment(void)
     // we need to know the climb ratio. We use 50% of the maximum
     // climb rate so we are not constantly at 100% throttle and to
     // give a bit more margin on terrain
-    float climb_ratio = 0.5f * TECS_controller.get_max_climbrate() / groundspeed;
+    float climb_ratio = g2.pitch_lookahead_climb_ratio * TECS_controller.get_max_climbrate() / groundspeed;
```
Georacer, line 608: By default, this will replace 0.5 with 50. That's not what we want.

Georacer (changes requested):
Hi Tim!

I've tried to read and internalize the lookahead code, but it's quite opaque; in the way I understand it, it's plainly wrong.
So I probably understand it wrong.

But that means that I shouldn't just expect your intervention to have no side effects.

Instead, do you have any test results (logs), where before the plane wouldn't manage a climb and now it does?

ArduPlane/altitude.cpp
timtuxworth: That sounds like a bug - thanks.

timtuxworth:
> Hi Tim!
> 
> I've tried to read and internalize the lookahead code, but it's quite opaque; in the way I understand it, it's plainly wrong. So I probably understand it wrong.

No you have it right. I had to rewrite terrain lookahead in my terrain avoidance Lua script #28625  because of that. In fact the only thing I really want out of it is the parameter value :-), so that my Lua based lookahead can use the same ratio.

ArduPlane/altitude.cpp
timtuxworth: Fixed.

timtuxworth:
> But that means that I shouldn't just expect your intervention to have no side effects.

The code should have no side effects iff the default value is used, which should give the same results as the previous hard coded value.

Georacer:
@timtuxworth I've spent some time parsing the code.

To me, it looks like if the ratio is 0, only then will the full climb rate be asked by the lookahead.

But I want to see comparative tests.
If you beat me to that, please post those logs.

Georacer:
Hello @timtuxworth !

I've done some testing with your code today.

I built a mission that I think is a repeatable stress-test for the lookahead code:
[dadi.txt](https://github.com/user-attachments/files/18587447/dadi.txt)

You can launch your SITL at those locations with
```bash
sim_vehicle.py -v ArduPlane --frame quadplane --console --map -D -G -l 38.638235,22.595485,388,0
```
and then load the mission.

I've run this mission with various settings and you can see cumulative results in the following figure.

- With purple, you see the terrain height.
- With blue, you see a run with `TERRAIN_LOOKAHD=0`. It crashes into the mountain ~90s in.
- With red, you see a run with `PTCH_LKAHD_CLMB=0`. This should provide the biggest clearance, but because of the way the code is structured, it returns early and is effectively the same as `TERRAIN_LOOKAHD=0`. This is a bug and I'll make a note of it later.
- With green, you see a run with `PTCH_LKAHD_CLMB=1`. This is actually what gives the greatest clearance. In theory this means that TECS capabilities will not be taken into account.
- With orange, you see the default (and currently applied on master) case of `PTCH_LKAHD_CLMB=50`. This is almost the same as the one prior, except for the end of the climb, where it flies lower.
- With pink, you see the case of `PTCH_LKAHD_CLMB=100`. In theory, this will reduce the lookahead altitude by 100% of the TECS capabilities. You see this is the setting that flies closest to the ground.

image

That said, I'd like to ask you for the following things:
1. Please repeat my experiments. Sometimes I'm getting flaky results.
2. Change the parameter behaviour, so that you get the most clearance at `PTCH_LKAHD_CLMB=100`. I think this is more intuitive.
3. Make it so that when setting `PTCH_LKAHD_CLIMB=0` the lookahead behaviour isn't disabled. Limiting to a minimum of 1 should do the trick, but feel free to do what you think best.

Georacer (changes requested):
<mandatory text to re-request for changes>

timtuxworth:
This is great. Thanks so much @Georacer . I should be able to look at this in the next couple of days. 

One big question. HOW did you get this graph of multiple flights on the same graph?

Georacer:
> One big question. HOW did you get this graph of multiple flights on the same graph?

Heh...

meme_08

timtuxworth:
This is interesting @Georacer - I'm getting Q_ASSIST kicking in at 0 altitude. According to the description of Q_ASSIST_ALT,  0 should mean disabled, but I'm definitely seeing it firing.

Georacer:
> This is interesting @Georacer - I'm getting Q_ASSIST kicking in at 0 altitude. According to the description of Q_ASSIST_ALT, 0 should mean disabled, but I'm definitely seeing it firing.

You know how it goes @timtuxworth : Log or didn't happen! :)

timtuxworth:
Made the changes requested by @Georacer and ran a suite of tests, to compare the results with 
a. no lookahead
b. TERRAIN_LKA_CLMB = 0, 25, 50 (the default), 75, 100
The plane crashes into the side of he mountain with no lookahead and with 0% ratio. It's successively "safer" with 25 ... 100.

Logs are here: (for as long as the links last). https://www.dropbox.com/scl/fo/pohi2wssydjefod689khp/APSSKzsSW3mkgHw29A0Ysjw?rlkey=seedu7gzeslzw9l6dfbdcnzpe&dl=0

ArduPlane/Parameters.cpp
```diff
@@ -1309,7 +1309,15 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("RNGFND_LND_ORNT", 36, ParametersG2, rangefinder_land_orient, ROTATION_PITCH_270),
 #endif
-    
+
+    // @Param: TERRAIN_LKA_CLMB
+    // @DisplayName: terrain lookahead climb pitch ratio
+    // @Description: The ratio of pitch to use when doing a terrain lookahead so the plane is not constantly at 100% throttle. 100% is the maximum  pitch which should give maximum ground clearance, and 0% should give the minimum pitch and the minimum ground clearance, the default is 50% to try to balance safety vs battery usage. User TERRAIN_LOOKAHD=0 to disable lookahead.
```
Georacer, line 1315: I would prefer you replace "pitch" with "climb rate" in this description.
Also you have a double space after the phrase "100% is the maximum".

ArduPlane/altitude.cpp
```diff
@@ -602,10 +602,13 @@ float Plane::lookahead_adjustment(void)
         // we're not moving
         return 0;
     }
-    // we need to know the climb ratio. We use 50% of the maximum
+    // we need to know the climb ratio. We use PTCH_LKAHD_CLMB % of the maximum
```
Georacer, line 605: Use the new name of the parameter.

ArduPlane/altitude.cpp
```diff
@@ -602,10 +602,13 @@ float Plane::lookahead_adjustment(void)
...
     // climb rate so we are not constantly at 100% throttle and to
     // give a bit more margin on terrain
-    float climb_ratio = 0.5f * TECS_controller.get_max_climbrate() / groundspeed;
+    // we inverse the provided ratio because it gives the best result and is easiest for the user to understand
+    // maximum rate = maximum pitch = maximum ground clearance
```
Georacer, line 609: Again, replace "pitch" with "climb rate".
Or even better
```
maximum climb rate = maximum ground clearance
```

Georacer (changes requested):

timtuxworth:
> @timtuxworth how is it possible that with lookahead the plane ends up **below** the terrain altitude? 

The green line is with lookahead disabled.

ArduPlane/Parameters.cpp
timtuxworth: done

timtuxworth:
Currently waiting on a partner to do some real-world testing.

Georacer:
No rush on my end ;)

timtuxworth:
I think this probably doesn't achieve the goal.

PR #28895, opened 2024-12-17
SITL: make SIM_ENGINE_FAIL a mask

robertlong13:
Peter Barker accidentally misdocumented the SIM_ENGINE_FAIL parameter. Rather than fix the documentation, I'd much rather change the code to align with it. I need a mask-based version of this for something I'm working on.

Normally, I'm guessing it's frowned upon to just change what a parameter does without renaming it, but given that this is SITL, and the docs suggest that this is what we want that param to do, I'm hoping this is okay.

I've also make SIM_ENGINE_MUL default to 0, so you just have to set SIM_ENGINE_FAIL to fail a motor, which I think is more intuitive. I can drop that part if there are objections.


libraries/SITL/SITL.cpp
```diff
@@ -246,7 +246,7 @@ const AP_Param::GroupInfo SIM::var_info[] = {
     // @Param: ENGINE_FAIL
     // @DisplayName: Engine Fail Mask
     // @Description: mask of motors which SIM_ENGINE_MUL will be applied to
-    // @Bitmask: 0: Servo 1, 1: Servo 2, 2: Servo 3, 3: Servo 4, 4: Servo 5, 5: Servo 6, 6: Servo 7, 7: Servo 8
+    // @Bitmask: 0: Servo 1, 1: Servo 2, 2: Servo 3, 3: Servo 4, 4: Servo 5, 5: Servo 6, 6: Servo 7, 7: Servo 8, 8: Servo 9, 9: Servo 10, 10: Servo 11, 11: Servo 12, 12: Servo 13, 13: Servo 14, 14: Servo 15, 15: Servo 16, 16: Servo 17, 17: Servo 18, 18: Servo 19, 19: Servo 20, 20: Servo 21, 21: Servo 22, 22: Servo 23, 23: Servo 24, 24: Servo 25, 25: Servo 26, 26: Servo 27, 27: Servo 28, 28: Servo 29, 29: Servo 30, 30: Servo 31, 31: Servo 32
```
peterbarker, line 249: ... we can't actually set 32 bits with the traditional protocol.  *shrug*

peterbarker (changes requested):
I'm OK with the semantics change, but we will need to call it out in release notes as others out there might have other tests relying on current semantics

libraries/SITL/SITL.cpp
robertlong13: Good point, but this makes it consistent with many other cases of servo/esc bitmask params we have the go up to 32.

peterbarker (approved):
LGTM

Tools/autotest/quadplane.py
```diff
@@ -1475,7 +1474,10 @@ def VTOLQuicktune_CPP(self):
                     (p, new_values[p], threshold))
 
         self.progress("ensure we are not overtuned")
-        self.set_parameter('SIM_ENGINE_MUL', 0.9)
+        self.set_parameters({
+            'SIM_ENGINE_MUL', 0.9,
+            'SIM_ENGINE_FAIL', 1 << 0,
```
peterbarker, line 1479:
```suggestion
            'SIM_ENGINE_MUL': 0.9,
            'SIM_ENGINE_FAIL': 1 << 0,
```

peterbarker (changes requested):

Tools/autotest/quadplane.py
robertlong13: 🤦🏻

peterbarker (approved):

PR #28896, opened 2024-12-17
AP_Periph: Allow configuring parity and stop bits for network passthrough

rishabsingh3003:


amilcarlucas:
I know this is late, but wouldn't it be better if the parameters where called `1_PAR`, `2_PAR`. `1_BT` and `2_BT` ?

PR #28897, opened 2024-12-17
AP_Math: squelch fabsf warnings on qurt

katzfey:
Add float cast to float version of fabsF to get rid of compiler warnings about demoting doubles to floats. Also removes the duplicate definitions of the fabsF macro.

### DIFF
--- a/ArduCopter/mode_guided.cpp
+++ b/ArduCopter/mode_guided.cpp
@@ -765,7 +765,14 @@ void ModeGuided::pos_control_run()
 
     float terrain_margin_m = 0.0; // Vertical buffer size in m
     if (guided_is_terrain_alt) {
+#if CONFIG_HAL_BOARD == HAL_BOARD_QURT
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wabsolute-value"
+#endif
         terrain_margin_m = MIN(copter.wp_nav->get_terrain_margin_m(), 0.5 * fabsF(guided_pos_target_ned_m.z));
+#if CONFIG_HAL_BOARD == HAL_BOARD_QURT
+#pragma clang diagnostic pop
+#endif
     }
     pos_control->input_pos_NED_m(guided_pos_target_ned_m, terrain_d_m, terrain_margin_m);
 
--- a/libraries/AP_Math/ftype.h
+++ b/libraries/AP_Math/ftype.h
@@ -28,7 +28,6 @@ typedef double ftype;
 #define ceilF(x) ceil(x)
 #define fminF(x,y) fmin(x,y)
 #define fmodF(x,y) fmod(x,y)
-#define fabsF(x) fabs(x)
 #define toftype todouble
 #else
 typedef float ftype;
@@ -47,7 +46,6 @@ typedef float ftype;
 #define ceilF(x) ceilf(x)
 #define fminF(x,y) fminf(x,y)
 #define fmodF(x,y) fmodf(x,y)
-#define fabsF(x) fabsf(x)
 #define toftype tofloat
 #endif
 
--- a/libraries/AP_Math/matrix_alg.cpp
+++ b/libraries/AP_Math/matrix_alg.cpp
@@ -77,9 +77,16 @@ static void mat_pivot(const T* A, T* pivot, uint16_t n)
     for(uint16_t i = 0;i < n; i++) {
         uint16_t max_j = i;
         for(uint16_t j=i;j<n;j++){
+#if CONFIG_HAL_BOARD == HAL_BOARD_QURT
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wabsolute-value"
+#endif
             if(fabsF(A[j*n + i]) > fabsF(A[max_j*n + i])) {
                 max_j = j;
             }
+#if CONFIG_HAL_BOARD == HAL_BOARD_QURT
+#pragma clang diagnostic pop
+#endif
         }
 
         if(max_j != i) {
--- a/libraries/AP_Math/quaternion.cpp
+++ b/libraries/AP_Math/quaternion.cpp
@@ -694,7 +694,14 @@ void QuaternionT<T>::zero(void)
 template <typename T>
 bool QuaternionT<T>::is_unit_length(void) const
 {
+#if CONFIG_HAL_BOARD == HAL_BOARD_QURT
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wabsolute-value"
+#endif
     if (fabsF(length_squared() - 1) < 1E-3) {
+#if CONFIG_HAL_BOARD == HAL_BOARD_QURT
+#pragma clang diagnostic pop
+#endif
         return true;
     }
 
--- a/libraries/AP_Math/vector2.h
+++ b/libraries/AP_Math/vector2.h
@@ -250,6 +250,10 @@ struct Vector2
         const T expected_run = seg_end.x-seg_start.x;
         const T intersection_run = point.x-seg_start.x;
         // check slopes are identical:
+#if CONFIG_HAL_BOARD == HAL_BOARD_QURT
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wabsolute-value"
+#endif
         if (::is_zero(expected_run)) {
             if (fabsF(intersection_run) > FLT_EPSILON) {
                 return false;
@@ -261,6 +265,9 @@ struct Vector2
                 return false;
             }
         }
+#if CONFIG_HAL_BOARD == HAL_BOARD_QURT
+#pragma clang diagnostic pop
+#endif
         // check for presence in bounding box
         if (seg_start.x < seg_end.x) {
             if (point.x < seg_start.x || point.x > seg_end.x) {


peterbarker:
(skyviper uses toy_mode, so it's probably something in there.

This is the command-line used to generatae the above:
```
git fetch github && git checkout master && git pull && git branch -D test ; git checkout -b test $GITHUB_BRANCH && git rebase master && rm -rf /tmp/tmp* && time size-check --parallel=6 -j 32 --board=CubeRedPrimary  --board=bebop
```

rmackay9:
We discussed this on the dev call (no Tridge though this week) and while it looks OK I was concerned that this is hiding an issue in the Copter guided mode code.  It seems that this change would cause us to cast a double to a float and lose accuracy.

katzfey:
Yes, it will cast a double to a float. That is the intention. But without the cast, if you pass a double to a function that takes a float argument it will do the same thing and generate a warning at the same time. No idea why this would cause a compiled code change in Skyviper though.

katzfey:
@rmackay9 Would the PR be acceptable if I put an "#if CONFIG_HAL_BOARD == HAL_BOARD_QURT" around the macro version with the float cast? Or is someone going to look into the Skyviper issue?

katzfey:
I rebased this one. AI analysis shows that there is no longer any difference in the skyviper build, possibly due to the more recent addition of the `-frandom-seed=4` compiler option. Is this one good to go now?

rmackay9:
Hi @katzfey, I've added it to next week's dev call so it will be reviewed by then if not before. txs!

tridge:

Verdict: **COMMENT** — minor points, not merge blockers.

Each point below was produced by one reviewer and then independently re-checked against the diff at the current head by a second, which either confirmed it or corrected it; items that did not survive that check have been dropped.

- (corrected) The title is stale because no cast is added, but `fabsF` is not removed entirely; definitions remain at `libraries/AP_Math/ftype.h:27` and `:45`, while selected calls become `std::abs`, e.g. `libraries/AP_Math/matrix_alg.cpp:80`.
- `libraries/AP_Math/matrix_alg.cpp:20` includes `AP_Math.h` and uses `std::abs` at `:80`; `<cmath>` is provided transitively by `libraries/AP_Math/AP_Math.h:3`, not included directly.
- (corrected) The build-break withdrawal is correct, but the callers were not all migrated; for example, `ArduPlane/navigation.cpp:61` still calls `fabsF`, which remains defined at `libraries/AP_Math/ftype.h:27` and `:45`.

Full report: https://uav.tridgell.net/DevCallReviews/2026_08_04/devcall_pr_reviews.html

peterbarker:
Of real concern here is that our unit tests didn't pick up the loss of precision.

tridge:
@katzfey need to fix the comment from the AI comment above

peterbarker:
@katzfey just to be clear, these functions are supposed to call the correct underlying function, depending on what precision the compilation is working with - so `fabsF` should use `fabs` if you're compiling with double-precision or `fabsf` if you're compiling with single-precision.

rmackay9:
I suspect that we actually need to fix the code that is causing the compile warnings.

For those interested in resolving this the errors can be seen by opening the [Checks tab](https://github.com/ArduPilot/ardupilot/pull/28897/checks) just above the PR description, then expand QURT Build and then further expand the "Waf Build" above "'copter' finished successfully".  Below are at least some of the failures

```
[77](https://github.com/ArduPilot/ardupilot/actions/runs/29965103821/job/89074763976?pr=28897#step:3:179)
  ../../libraries/AC_PrecLand/AC_PrecLand_StateMachine.cpp:219:13: warning: absolute value function 'fabsf' given an argument of type 'double' but has parameter of type 'float' which may cause truncation of value [-Wabsolute-value]
          if (fabsf(pos.z - retry_pos_m.z) < MAX_POS_ERROR_M) {
              ^
  ../../libraries/AC_PrecLand/AC_PrecLand_StateMachine.cpp:219:13: note: use function 'std::abs' instead
  [ 55/965] Compiling libraries/AP_AHRS/AP_AHRS_NavEKF2.cpp
          if (fabsf(pos.z - retry_pos_m.z) < MAX_POS_ERROR_M) {
              ^~~~~
              std::abs
  1 warning generated.
```

```
  In file included from ../../libraries/AP_GPS/AP_GPS_UBLOX_CFGv2.cpp:21:
  ../../libraries/AP_GPS/AP_GPS_UBLOX.h:748:16: warning: static_assert with no message is a C++17 extension [-Wc++17-extensions]
      UBLOX_GNSS(UBLOX_GNSS_TYPE_ASSERT)
```

Plane's warnings are
```
[4](https://github.com/ArduPilot/ardupilot/actions/runs/29965103821/job/89074763976?pr=28897#step:3:1167)
  In file included from ../../libraries/APM_Control/AP_FW_Controller.cpp:20:
  ../../libraries/APM_Control/AP_FW_Controller.h:128:14: warning: private field 'last_run_ticks' is not used [-Wunused-private-field]
      uint32_t last_run_ticks;
               ^
  1 warning generated.
```

katzfey:
Okay, thanks, I'll review the comments and try to update accordingly

rmackay9:
@katzfey,

Here's an attempted fix for the AC_PrecLand warning https://github.com/ArduPilot/ardupilot/pull/33857

.. and here's another one https://github.com/ArduPilot/ardupilot/pull/33858

katzfey:
@rmackay9 I just pushed the updates. Hopefully this resolves the fabsf issue appropriately.

tridge:
@katzfey does clang on the qurt compiler have a pragma to hide this class of warning?

peterbarker:
Sorry - we don't use std::abs - https://ardupilot.org/dev/docs/style-guide.html#no-standard-library

Feel free to squash this with a compiler directive in waf for now.

rmackay9:
I don't really understand the issue but @peterbarker and perhaps @tpwrules do.  Here's a Claude description of the issue.

The root cause is that ArduPilot has two independent "generic precision" typedefs that are easy to conflate:

- ftype (and the *F macros like fabsF, sqrtF, etc. in ftype.h) — resolves to float or double based on HAL_WITH_EKF_DOUBLE, which in turn mirrors HAL_HAVE_HARDWARE_DOUBLE (does this CPU have a fast hardware double FPU?).
- postype_t (used by Vector3p/Vector2p, e.g. position state) — resolves to float or double based on HAL_WITH_POSTYPE_DOUBLE, which is purely a flash-size heuristic (HAL_PROGRAM_SIZE_LIMIT_KB > 1024), unrelated to FPU capability.

These two flags usually agree, but not always. On QURT, HAL_HAVE_HARDWARE_DOUBLE is 0 (so ftype = float), while HAL_PROGRAM_SIZE_LIMIT_KB is 2048 (so postype_t = double). Any code that calls fabsF()/fabsf() directly on a postype_t value (or a Vector3p/Vector2p component) on that board ends up passing a double to a function expecting float, triggering a -Wabsolute-value truncation warning.

The fix is to never mix the two: convert postype_t/Vector3p values to float explicitly (e.g. via .tofloat()) before using float-only helpers like fabsf, rather than assuming fabsF/ftype macros track postype_t's precision — they don't, since they're driven by a different config flag.

timtuxworth:
> I don't really understand the issue but @peterbarker and perhaps @tpwrules do. Here's a Claude description of the issue.
> 
> The root cause is that ArduPilot has two independent "generic precision" typedefs that are easy to conflate:
> 
> * ftype (and the *F macros like fabsF, sqrtF, etc. in ftype.h) — resolves to float or double based on HAL_WITH_EKF_DOUBLE, which in turn mirrors HAL_HAVE_HARDWARE_DOUBLE (does this CPU have a fast hardware double FPU?).
> * postype_t (used by Vector3p/Vector2p, e.g. position state) — resolves to float or double based on HAL_WITH_POSTYPE_DOUBLE, which is purely a flash-size heuristic (HAL_PROGRAM_SIZE_LIMIT_KB > 1024), unrelated to FPU capability.
> 
> These two flags usually agree, but not always. On QURT, HAL_HAVE_HARDWARE_DOUBLE is 0 (so ftype = float), while HAL_PROGRAM_SIZE_LIMIT_KB is 2048 (so postype_t = double). Any code that calls fabsF()/fabsf() directly on a postype_t value (or a Vector3p/Vector2p component) on that board ends up passing a double to a function expecting float, triggering a -Wabsolute-value truncation warning.
> 
> The fix is to never mix the two: convert postype_t/Vector3p values to float explicitly (e.g. via .tofloat()) before using float-only helpers like fabsf, rather than assuming fabsF/ftype macros track postype_t's precision — they don't, since they're driven by a different config flag.

Seems to me the real problem is that HAL_WITH_POSTYPE_DOUBLE based on HAL_PROGRAM_SIZE_LIMIT_KB is wrong. If it was replaced with HAL_HAVE_HARDWARE_DOUBLE it would probably compile without an error and avoid using slow double math on a board without an FPU.

katzfey:
As per comments above I removed std::abs and went back to fabsF and just added a pragma for QURT HAL only that disables that warning. @rmackay9 Not sure if you want to change #33857 to use this pragma also?

peterbarker (approved):
I'm fine with this in the short term.

I think this is something we can solve correctly long-term, but this is a pain point for @katzfey so I'm going to approve and merge it.

PR #28898, opened 2024-12-17
AP_Volz_Protocol: Moved member variable definition to proper block

katzfey:
Moved member variable sent_count definition to proper #define block to prevent unused variable warning.


IamPete1 (approved):
LGTM

PR #28903, opened 2024-12-18
Bendy Ruler Lua Script Added

TeteJK:
# Bendy Ruler Lua Script Implementation

Problem framing from @NDevDrone  and @tridge

This implementation aimed to integrate the Bendy Ruler obstacle avoidance algorithm into the latest version of ArduPilot by using a Lua script for evaluation and calculations.

SingleDAA MultiDAA

---

## 1. Enhancements to the AP_Avoidance Module

Bindings and functions related to obstacle detection using ADS-B were implemented within the AP_Avoidance module to interact with Lua. The following methods were added:

> num_obstacles()
> Returns the total number of obstacles detected via ADS-B (requires ADS-B to be enabled).
> 
> get_obstacle_loc(instance)
> Retrieves the last known position of a detected obstacle as a Location userdata type. The instance parameter specifies the obstacle.
> 
> get_obstacle_vel(instance)
> Provides the velocity of a detected obstacle as a 3D vector (Vector3f) in NED (North-East-Down) coordinates, measured in meters per second.
> 
> get_obstacle_id(instance)
> Returns the unique ID of an obstacle as reported by ADS-B, using the instance parameter.
> 
> get_obstacle_time(instance)
> Returns the timestamp of the ADS-B message for a specific obstacle as an integer value.
> 
> These methods, accessible via the AP_Avoidance singleton instance, enable real-time interaction with ADS-B data. Where existing ArduPilot methods sufficed for Lua script interactions, they were reused. Missing or underperforming methods were hardcoded into the Lua script.

---

## 2. Lua Implementation of the Bendy Ruler Logic

The Bendy Ruler avoidance algorithm was implemented in Lua, with modifications to adapt to the scripting environment:

- Exclusions: Logic for fences and margin areas was omitted for simplicity. The current focus is solely on avoiding obstacles.
- Simplified Task Logic: Instead of using threads and schedulers, the avoidance logic is invoked within the update() function. This ensures computations only occur when a potential threat is detected, optimizing limited computational resources when working with Lua scripts.
- Navigation Commands: The Lua script does not interact directly with navigation and command libraries. Instead, it sends commands to fly to a new position when avoidance is required using the already implemented Lua methods.

---

## 3. Addressing Altitude Handling Issues

The GenObstacles library, used for terrain altitude calculations, had a bug preventing accurate obstacle generation. To enable horizontal avoidance testing:

- All altitude-dependent logic was temporarily commented out.
- Obstacles were generated without considering altitude.

This workaround allows testing of horizontal avoidance logic but requires future fixes to handle altitude data correctly.

---

## 4. STIL Simulation for Testing the Lua Script

Follow these steps to test the Lua script implementation for the Bendy Ruler obstacle avoidance algorithm:

# 1. Prepare the Simulation Environment

- Navigate to the directory where the simulation will be run.
- Create a folder named `scripts` in this directory (if not already present).
- Copy the latest version of the Lua script containing the Bendy Ruler code (`Bendy_ruler.lua`) into the `scripts` folder.

# 2. Start the Simulation

Run the simulation using the following command, replacing `IP` with the appropriate IP address:

```bash
python sim_vehicle.py -w -v ArduPlane -f plane-elevon --console --map -C --load-module asterix,genobstacles --out=udp:IP:14550
```

# 3. Set Bendy Ruler Parameters

```
param set SCR_ENABLE 1 
param set AVD_ENABLE 1 
param set ADSB_TYPE 1
param set SCR_VM_I_COUNT 1000000
param set AVD_W_ACTION 2 
param set AVD_W_DIST_XY 5000
```

# 4. Clear Initial Obstacles 
Delete all obstacles generated during startup to avoid overloading the Lua script.

- Right-click on the map.
- Select Obstacles > Clear All.

# 5. Reset Bendy Ruler Lua Script 
Something like 

```
script stop bendy-ruler.lua
script load bendy-ruler.lua
```

### DIFF
--- a/libraries/AP_Scripting/applets/Bendy_ruler.lua
+++ b/libraries/AP_Scripting/applets/Bendy_ruler.lua
@@ -1,0 +1,896 @@
+-- MAV_COLLISION_THREAT_LEVEL
+MAV_COLLISION_THREAT_LEVEL = {
+    NONE = 0,         -- Not a threat
+    LOW = 1,          -- Mild concern about this threat
+    HIGH = 2,         -- Craft is panicking and may take action to avoid
+    ENUM_END = 3      -- End of enum
+}
+-- MAV_COLLISION_SRC
+MAV_COLLISION_SRC = {
+    ADSB = 0,                         -- Source is ADSB_VEHICLE packets
+    MAVLINK_GPS_GLOBAL_INT = 1,        -- Source is MAVLink GPS_GLOBAL_INT
+    ENUM_END = 2                      -- End of enum
+}
+MAV_COLLISION_ACTION={
+    NONE=0, -- Ignore any potential collisions 
+    REPORT=1, -- Report potential collision 
+    ASCEND_OR_DESCEND=2, -- Ascend or Descend to avoid threat 
+    MOVE_HORIZONTALLY=3, -- Move horizontally to avoid threat 
+    MOVE_PERPENDICULAR=4, -- Aircraft to move perpendicular to the collision's velocity vector 
+    RTL=5, -- Aircraft to fly directly back to its launch point 
+    HOVER=6, -- Aircraft to stop in place 
+ } ;
+
+
+local gcs_threat = {
+    src = nil,  
+    src_id = 0,  
+    timestamp_ms = 0,  
+    _location = nil,  
+    _velocity = nil,  
+    threat_level = nil,  
+    closest_approach_xy = 0.0,  
+    closest_approach_z = 0.0,  
+    time_to_closest_approach = 0.0,  
+    distance_to_closest_approach = 0.0  
+}
+
+
+local avoidance_request = {
+    current_loc = nil,
+    target_loc = nil,
+    groundspeed = 0.0,
+    airspeed = 0.0,
+    wind_dir_rad = 0.0,
+    wind_speed = 0.0,
+    request_time_ms = 0
+}
+local avoid_req2 = {
+    current_loc = nil,
+    target_loc = nil,
+    groundspeed = 0.0,
+    airspeed = 0.0,
+    wind_dir_rad = 0.0,
+    wind_speed = 0.0,
+    request_time_ms = 0
+}
+local avoidance_result = {
+    target_loc = nil,
+    new_target_loc = nil,
+    result_time_ms = 0,
+    avoidance_needed = nil
+}
+
+local PARAM_TABLE_KEY = 7
+local PARAM_TABLE_PREFIX = "BENDY_"
+local OPTION_IGNORE_HEIGHT=1
+local ALT_FRAME_ABSOLUTE = 0
+
+-- bind a parameter to a variable
+function bind_param(name)
+   local p = Parameter()
+   assert(p:init(name), string.format('could not find %s parameter', name))
+   return p
+end
+
+-- add a parameter and bind it to a variable
+function bind_add_param(name, idx, default_value)
+   assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), string.format('could not add param %s', name))
+   return bind_param(PARAM_TABLE_PREFIX .. name)
+end
+
+-- setup SHIP specific parameters
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 7), 'could not add param table')
+--[[
+  // @Param: BENDY_OPTIONS
+  // @DisplayName: ADS-B avoidance options
+  // @Description: Bitmask of behaviour options
+  // @Values: 0:Disabled,1:Enabled
+  // @User: Advanced
+--]]
+BENDY_OPTIONS     = bind_add_param('OPTIONS', 1, 1)
+
+--[[
+  // @Param: BENDY_MARGIN_FENCE
+  // @DisplayName: fence margin
+  // @Description: Avoidance margin for fence
+  // @User: Advanced
+--]]
+BENDY_MARGIN_FENCE = bind_add_param('MARGIN_FENCE', 2, 50)
+
+--[[
+  // @Param: BENDY_MARGIN_DYN
+  // @DisplayName: dynamic margin
+  // @Description: Avoidance margin for dynamic objects
+  // @User: Advanced
+--]]
+BENDY_MARGIN_DYN   = bind_add_param('MARGIN_DYN', 3, 20)
+
+--[[
+  // @Param: BENDY_MARGIN_EXCL
+  // @DisplayName: exclusion zone margin
+  // @Description: Avoidance margin for exclusion zones
+  // @User: Advanced
+--]]
+BENDY_MARGIN_EXCL   = bind_add_param('MARGIN_EXCL', 4, 20)
+
+--[[
+  // @Param: BENDY_MARGIN_WIDE
+  // @DisplayName: wide avoidance margin
+  // @Description: Avoidance margin for wide avoidance
+  // @User: Advanced
+--]]
+BENDY_MARGIN_WIDE   = bind_add_param('MARGIN_WIDE', 5, 30)
+
+--[[
+  // @Param: BENDY_MARGIN_HGT 
+  // @DisplayName: height avoidance margin
+  // @Description: Avoidance margin for height avoidance
+  // @User: Advanced
+--]]
+BENDY_MARGIN_HGT   = bind_add_param('MARGIN_HGT', 6, 60)
+
+--[[
+  // @Param: BENDY_LOOKAHEAD
+  // @DisplayName: avoidance lookahead distance
+  // @Description: Avoidance lookahead distance
+  // @Units: metres
+  // @User: Advanced
+--]]
+BENDY_LOOKAHEAD  = bind_add_param('LOOKAHEAD', 7, 500)
+
+WARN_DIST_XY  = bind_param("AVD_W_DIST_XY")
+WARN_ACTION  = bind_param("AVD_W_ACTION")
+local warn_act= WARN_ACTION:get()
+AVD_ENABLE  = bind_param("AVD_ENABLE")
+local enabled=AVD_ENABLE:get()
+ROLL_LIMIT_DEG = bind_param("ROLL_LIMIT_DEG")
+local roll_limit_cd=ROLL_LIMIT_DEG:get() *100
+local lookahead_param=BENDY_LOOKAHEAD:get()
+local margin_fence=BENDY_MARGIN_FENCE:get()
+local enabled=AVD_ENABLE:get()
+local current_loc=Location()
+local target_loc=Location()
+local thread_created = false
+local current_lookahead =0
+local obstacle_count
+local collision_detected
+local GRAVITY_MSS=9.80665
+local LOCATION_SCALING_FACTOR_INV=89.83204953368922
+local new_target_loc = Location() 
+local current_target_s
+local flag=0
+
+--Auxiliary functions
+function wrap_180(angle)
+    local res = wrap_360(angle)
+    if res > 180 then
+       res = res - 360
+    end
+    return res
+end
+
+function wrap_360(angle)
+    local res = math.fmod(angle, 360.0)
+     if res < 0 then
+         res = res + 360.0
+     end
+     return res
+ end
+
+function length_squared(w)
+    return (w:x() * w:x()) + (w:y() * w:y())
+end
+
+function dot_product_2vector(p,w)
+    return (w:x() * p:x()) + (w:y() * p:y())
+end
+
+function closest_point(p, w)
+    local l2 = length_squared(w)
+    if l2 < 1e-6 then
+        return w  -- case v == w
+    end
+    local t = dot_product_2vector(p,w) / l2
+    if t <= 0 then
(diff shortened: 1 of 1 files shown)


libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+    local new_target_loc = mission_avoidance(current_loc, orig_target_loc, groundspeed)
+
+    if new_target_loc then
+        -- tell the vehicle to fly to the new calculated target
+        vehicle:update_target_location(target_loc, new_target_loc)
+        last_updated_target_loc = new_target_loc:copy()
+
+        local avoid_dist = orig_target_loc:get_distance(new_target_loc)
+
+        gcs:send_named_float("AVD_DIST", avoid_dist)
```
timtuxworth, line 852: What does this do?

timtuxworth:
It would be really useful to have some logging.

libraries/AP_Avoidance/AP_Avoidance.h
```diff
@@ -106,6 +106,23 @@ class AP_Avoidance {
...
+    Vector3f get_obstacle_vel(uint8_t instance) const {
+        return _obstacles[instance]._velocity;
+    }    
+    int32_t get_obstacle_id(uint8_t instance) const {
+        return _obstacles[instance].src_id;
+    }    
+    int32_t get_obstacle_time(uint8_t instance) const {
+        return _obstacles[instance].timestamp_ms;
+    }
+
```
timtuxworth, line 125: I'm thinking it might be better to have this as a single method that returns all these values in a single call right @tpwrules ? 
Apart from the simplicity and flash savings isn't there also the risk of a race condition?

Tools/autotest/locations.txt
```diff
@@ -104,3 +104,5 @@ Egge=60.215720,10.324071,198,303
 Gundaroo=-35.02349196,149.26496411,576.8,0
 Kaga=36.3268982,136.3316638,44,0
 UCSB=34.413963,-119.848946,0,0
+California=37.4275449,-122.1697,0,0
```
timtuxworth, line 107: California is a big place - just saying.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4013,3 +4013,24 @@ function visual_odom:healthy() end
 -- visual odometry quality as a percentage from 1 to 100 or 0 if unknown
 ---@return integer
 function visual_odom:quality() end
+
+-- Returns the number of detected obstacles via ADS-B
+-- the ADS-B Must be enabled.
+---@return integer -- number of obstacles
+function avoid:num_obstacles() end-- Returns a Location userdata for the last position of the obstacle observed.
```
timtuxworth, line 4020: The comment does't seem to match the function.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4013,3 +4013,24 @@ function visual_odom:healthy() end
...
+
+---@param instance integer -- instance number
+---@return Location_ud --ADSB location
+function avoid:get_obstacle_loc(instance) end-- Returns a Vector3f that contains the velocity as obtained from ADS-B.
+
+---@param instance integer -- instance number
+---@return Vector3f_ud -- 3D velocity in m/s, in NED format
+function avoid:get_obstacle_vel(instance) end-- Obstacle id as obtained from ADS-B.
```
timtuxworth, line 4028: Can't this be other obstacles? Not just ADS-B? What about Remote ID vehicles or something identified by an off board AI/camera?

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+MAV_COLLISION_THREAT_LEVEL = {
+    NONE = 0,         -- Not a threat
+    LOW = 1,          -- Mild concern about this threat
+    HIGH = 2,         -- Craft is panicking and may take action to avoid
+    ENUM_END = 3      -- End of enum
+}
+-- MAV_COLLISION_SRC
+MAV_COLLISION_SRC = {
+    ADSB = 0,                         -- Source is ADSB_VEHICLE packets
+    MAVLINK_GPS_GLOBAL_INT = 1,        -- Source is MAVLink GPS_GLOBAL_INT
```
timtuxworth, line 13: Remote ID?

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+}
+local avoidance_result = {
+    target_loc = nil,
+    new_target_loc = nil,
+    result_time_ms = 0,
+    avoidance_needed = nil
+}
+
+local PARAM_TABLE_KEY = 7
+local PARAM_TABLE_PREFIX = "BENDY_"
```
timtuxworth, line 69: I'm pretty sure we are going to need more than 9 parameters, using "BENDY". I've started using this convention:
prefix = 
Z - identifies a scripting parameter 
XX - scripting function - e.g. PF = Plane Follow - so BR = Bendy Ruler
9 - bank number when you get beyond 9 parameters so

ZBR_ (or maybe ZBR1) for the first bank of 9 bendy ruler parameters
ZBR2_ for the second bank etc.

This means 

- its easy to find _any_ scripted parameters by just searching or "Z" (or Z* in mavproxy).
- the maximum hit for bendy ruler (for up to 99 parameters) is 4 characters.

Which leaves room for coding some reasonable meaning into the remaining characters.

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+--]]
+BENDY_OPTIONS     = bind_add_param('OPTIONS', 1, 1)
+
+--[[
+  // @Param: BENDY_MARGIN_FENCE
+  // @DisplayName: fence margin
+  // @Description: Avoidance margin for fence
+  // @User: Advanced
+--]]
+BENDY_MARGIN_FENCE = bind_add_param('MARGIN_FENCE', 2, 50)
```
timtuxworth, line 103: How about MRG - jus thinking we might need the characters later so
MRG_FEN
MRG_DYN
MRG_EXC

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
+local UPDATE_RATE_HZ = 10
```
timtuxworth, line 1: Aren't CAPITALs supposed to be reserved for GLOBALS?

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+    --print(string.format("   time_horizon: (%d)", time_horizon))
+    --print(string.format("   delta pos: (y=%f, x=%f)", delta_pos_ne.x, delta_pos_ne.y))
+    --print(string.format("   delta vel: (y=%f, x=%f)", delta_vel_ne.x, delta_vel_ne.y))
+    --print(string.format("   line segment: (y=%f, x=%f)", line_segment_ne.x, line_segment_ne.y))
+    --print(string.format("   closest: (%f)", ret))    
+    return ret
+end
+
+function longitude_scale(lat)
+    local DEG_TO_RAD = math.pi / 180
```
timtuxworth, line 232: there is already a math.deg() function in Lua

libraries/AP_Scripting/applets/Bendy_ruler.lua
tpwrules: Where did you get the idea we can't have more than nine? A table can have 62 or so.

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+function location_project(loc1, bearing_deg, distance)
+    -- Create a copy of the location
+    local loc2 = Location()
+    --loc2=loc1
+    --loc2:offset_bearing(bearing_deg, distance)
+    local lat, lon=offset_bearing(loc1:lat(), loc1:lng(), bearing_deg, distance)
+    loc2:lat(lat)
+    loc2:lng(lon)
+    loc2:alt(loc1:alt())
+    --gcs:send_text(0, string.format("IN: Lat: %.2f, Lon: %.2f", loc1:lat(),loc1:lng()))
```
timtuxworth, line 277: Need the frame

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+   return bind_param(PARAM_TABLE_PREFIX .. name)
+end
+
+-- setup SHIP specific parameters
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 7), 'could not add param table')
+--[[
+  // @Param: BENDY_OPTIONS
+  // @DisplayName: ADS-B avoidance options
+  // @Description: Bitmask of behaviour options
+  // @Values: 0:Disabled,1:Enabled
```
timtuxworth, line 92: If this is enabled/disabled it should be _ENABLE

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+    -- get the current ground course
+    local ground_course_deg
+    ground_course_deg = wrap_180(math.deg(ahrs:groundspeed_vector():angle()))
+    --gcs:send_text(0, string.format("ground_course_deg : %.2f",ground_course_deg ))  
+       
+    -- If the full distance is less than 20m, no avoidance is needed
+    if full_distance < 20 then
+        return nil
+    end    
+    -- Try 5 degree increments around a circle, alternating left and right. Check each one to see if flying in that direction would avoid all obstacles.
```
timtuxworth, line 655: Seems to be 15 degrees (or am I reading it wrong)?
bearing_inc_cd = 1500

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+
+    local groundspeed = ahrs:groundspeed_vector():length()
+
+    -- run avoidance calculation
+    local new_target_loc = mission_avoidance(current_loc, orig_target_loc, groundspeed)
+
+    if new_target_loc then
+        -- tell the vehicle to fly to the new calculated target
+        vehicle:update_target_location(target_loc, new_target_loc)
+        last_updated_target_loc = new_target_loc:copy()
```
timtuxworth, line 848: update_target_location() is probably going go fail sometimes, so maybe last_updated_target_loc should only get updated if it returns true?

libraries/AP_Scripting/applets/Bendy_ruler.lua
tpwrules: It’s a global considering it’s in the outermost scope of the script and a setting at the top. Marking it as a local is probably marginally better for performance.

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+
+        local avoid_dist = orig_target_loc:get_distance(new_target_loc)
+
+        gcs:send_named_float("AVD_DIST", avoid_dist)
+    else
+        -- revert to the original target location, no avoidance needed
+        vehicle:update_target_location(target_loc, orig_target_loc)
```
timtuxworth, line 855: It's not clear why you are doing this. It seems like it would already be going here, so why tell it to go where it's already going?

libraries/AP_Scripting/applets/Bendy_ruler.lua
```diff
@@ -0,0 +1,871 @@
...
+
+--[[
+    return true if two locations are identical
+--]]
+local function locations_equal(loc1, loc2)
+    return loc1:lat() == loc2:lat() and loc1:lng() == loc2:lng() and loc1:alt() == loc2:alt()
+end
+
+local orig_target_loc = nil
+local last_updated_target_loc = nil
```
timtuxworth, line 810: What is this for?

libraries/AP_Scripting/applets/Bendy_ruler.lua
timtuxworth: I hit this in writing quadplane_terrain_avoid.lua. An error definitely gets thrown if you try to add more than 9 parameters to a table. I think it might be a Lua only restriction.

timtuxworth: I think its in AP_Param.h - AP_PARAM_MAX_DYNAMIC = 10, even though there is 

```
 if (num_params > 63) {
        return false;
    }

```
at line 2999, I think all the code in add_table() uses AP_PARAM_MAX_DYNAMIC as the limit, and I have found that you definitely can't add more than 10 parameters in a table.

timtuxworth: Turns out I was wrong about the 9 parameter limit. Thanks for setting me straight on this.

PR #28904, opened 2024-12-19
Plane: FlyEachFrame Fixes

robertlong13:
Mainly adds ICE frames back to FlyEachFrame, but I noticed a couple other issues while I was here.

For the ICE frames, all I needed to do was add engine-start commands to the beginning of the missions. The RPM on `plane-ice` dips under 100 on the landing approach, so I dropped the rpm threshold to 50 in its param file to fix that.

Other issues:
https://github.com/ArduPilot/ardupilot/pull/26839 fixed the default param loading on copter, but the same issue was present in plane (didn't cause any problems, because we don't override "model" in any plane frames in vehicleinfo... yet).

And I was going to try to fix the tailsitter issue, but it looks like tailsitter is just utterly unflyable right now, and I ran out of time. Instead, I have updated the reason to reflect what is currently wrong with it.

image
Large position-hold instability in hover, and pretty immediate loss of attitude control in forward flight. I might come back and tackle this one in a future PR.


peterbarker (approved):
LGTM

peterbarker:
Merged, thanks!

PR #28905, opened 2024-12-19
SITL: fix SIM_BATT_CAP_AH

robertlong13:
Previously did nothing. The only way to set the capacity was to set it within your model's json.

### DIFF
--- a/libraries/SITL/SIM_Battery.cpp
+++ b/libraries/SITL/SIM_Battery.cpp
@@ -128,6 +128,12 @@ void Battery::init_voltage(float voltage)
     set_initial_SoC(voltage);
 }
 
+void Battery::init_capacity(float capacity)
+{
+    capacity_Ah = capacity;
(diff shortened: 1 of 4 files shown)


peterbarker:
@robertlong13 how *does* this interact with someone setting the capacity in the json?

robertlong13:
> @robertlong13 how _does_ this interact with someone setting the capacity in the json?

After the json is loaded, it [sets the default for this parameter to whatever capacity the json set it to](https://github.com/ArduPilot/ardupilot/blob/c29815644f8c28b5bbf9f5a341013cea6b78c48a/libraries/SITL/SIM_Frame.cpp#L519). So, unless you go out of your way to set this parameter, this PR doesn't impact anything (tested with Callisto).

robertlong13:
@peterbarker reminder on this one. Fell off my radar too

peterbarker:
Merged, thanks!

PR #28906, opened 2024-12-19
Plane: log whether throttle is suppressed or not 

peterbarker:
This adds logging of whether the throttle is suppressed or not into the `STAT` message.

I've created a new field, "Bit" to hold a bitmask of boolean information about the vehicle - saves 40 bytes per message past me adding the new field.

I suggest we remove the old equivalent uint8_t fields in 4.8 - just have them in here for a bit of cross-over.

The new LogStatus class is perhaps overkill - but I think we might be able to templatise it and have a nice, compact way of doing similar bitmasks in log messages.

For this entire PR:
```
CubeOrange-periph-heavy  *                                                                     
CubeRedPrimary                      *      *           *       *                 120    *      *
Durandal                            *      *           *       *                 72     *      *
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     *      *           *       *                 120    *      *
MatekF405                           *      *           *       *                 120    *      *
Pixhawk1-1M-bdshot                  *                  *       *                 120    *      *
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           *      *           *       *                 120    *      *
```

Without the new column, just with the weird class and consolidating into status, removing the old fields:
```
CubeOrange-periph-heavy  *                                                                     
CubeRedPrimary                      *      *           *       *                 -32    *      *
Durandal                            *      *           *       *                 -24    *      *
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     *      *           *       *                 -32    *      *
MatekF405                           *      *           *       *                 -32    *      *
Pixhawk1-1M-bdshot                  *                  *       *                 -32    *      *
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           *      *           *       *                 -32    *      *
```

### DIFF
--- a/ArduPlane/Log.cpp
+++ b/ArduPlane/Log.cpp
@@ -210,6 +210,7 @@ struct PACKED log_Status {
     bool is_still;
     uint8_t stage;
     bool impact;
+    bool throttle_supressed;
 };
 
 void Plane::Log_Write_Status()
@@ -225,6 +226,7 @@ void Plane::Log_Write_Status()
         ,is_still    : AP::ins().is_still()
         ,stage       : static_cast<uint8_t>(flight_stage)
         ,impact      : crash_state.impact_detected
+        ,throttle_supressed : throttle_suppressed
         };
 
     logger.WriteBlock(&pkt, sizeof(pkt));
@@ -382,8 +384,9 @@ const struct LogStructure Plane::log_structure[] = {
 // @Field: Still: True when vehicle is not moving in any axis
 // @Field: Stage: Current stage of the flight
 // @Field: Hit: True if impact is detected
+// @Field: Sup: True if throttle is suppressed
     { LOG_STATUS_MSG, sizeof(log_Status),
-      "STAT", "QBfBBBBBB",  "TimeUS,isFlying,isFlyProb,Armed,Safety,Crash,Still,Stage,Hit", "s--------", "F--------" , true },
+      "STAT", "QBfBBBBBBB",  "TimeUS,isFlying,isFlyProb,Armed,Safety,Crash,Still,Stage,Hit,Sup", "s---------", "F---------" , true },
 
 // @LoggerMessage: QTUN
 // @Description: QuadPlane vertical tuning message


Georacer:
I like the idea of a more compact logging of logic flags. Smaller logs is always better.

But as it stands, it would make reading the logs (e.g. via MAVExplorer) much harder.
How do you propose decoding the new `status` field to extract and plot individual bits?

peterbarker:
> I like the idea of a more compact logging of logic flags. Smaller logs is always better.
> 
> But as it stands, it would make reading the logs (e.g. via MAVExplorer) much harder. How do you propose decoding the new `status` field to extract and plot individual bits?

MissionPlanner already has a good interface for this - as do the web tools IIUC.

MAVExplorer you can log the bits with `graph STAT.Bit&1<<0` for the IS_FLYING bit.

ArduPlane/Log.cpp
```diff
@@ -210,22 +210,65 @@ struct PACKED log_Status {
...
+    status.set(LogStatus::Bit::IMPACT, crash_state.impact_detected);
+    status.set(LogStatus::Bit::THROTTLE_SUPPRESSED, throttle_suppressed);
+
+    const struct log_STAT pkt {
+        LOG_PACKET_HEADER_INIT(LOG_STATUS_MSG),
+        time_us   : AP_HAL::micros64(),
+        is_flying   : status.get(LogStatus::Bit::IS_FLYING),
+        is_flying_probability : isFlyingProbability,
+        armed       : status.get(LogStatus::Bit::ARMED),
+        safety      : static_cast<uint8_t>(hal.util->safety_switch_state()),
```
IamPete1, line 265: Why not have safety state in the bitmask?

ArduPlane/Log.cpp
```diff
@@ -382,8 +425,10 @@ const struct LogStructure Plane::log_structure[] = {
 // @Field: Still: True when vehicle is not moving in any axis
 // @Field: Stage: Current stage of the flight
 // @Field: Hit: True if impact is detected
-    { LOG_STATUS_MSG, sizeof(log_Status),
-      "STAT", "QBfBBBBBB",  "TimeUS,isFlying,isFlyProb,Armed,Safety,Crash,Still,Stage,Hit", "s--------", "F--------" , true },
+// @Field: Bit: bitmask of vehicle state information
+// @FieldBitmaskEnum: Bit: LogStatus::Bit
+    { LOG_STATUS_MSG, sizeof(log_STAT),
+      "STAT", "QBfBBBBBBH",  "TimeUS,isFlying,isFlyProb,Armed,Safety,Crash,Still,Stage,Hit,Bit", "s---------", "F---------" , true },
```
tridge, line 431: I think a 'Sup' field is a better approach

tridge (changes requested):

ArduPlane/Log.cpp
peterbarker: Three states and comes from an enumeration.  We've done that before in copter's `ap` and it sucked

peterbarker:
image

ArduPlane/Log.cpp
peterbarker: Done

peterbarker:
Merging as agreed at DevCall.

PR #28908, opened 2024-12-19
AP_GPS: handle switching for more than 2 gpses

sellauth:
Instead of instantly switching after finding a better GPS, check all GPSes first then switch to ensure there isn't any other better GPS, This will only improve performance on setups with more than 2 GPSes, This commit also changes the behavior of status lock switches, previously they switched without any timeouts, in the new code the 5000ms timeout applies to them as well.

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -1197,35 +1197,62 @@ void AP_GPS::update_primary(void)
         return;
     }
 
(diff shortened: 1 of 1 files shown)


Ryanf55:
Nice! Looks reasonable. How'd you test the changes?

sellauth:
> Nice! Looks reasonable. How'd you test the changes?

I haven't personally tested the changes yet

PR #28909, opened 2024-12-19
Github: add Lua problem matcher

IamPete1:
This adds a problem matcher for the lua CI check. 

image

Its not perfect, the lua language server check regex only gets the first line of the message. There is also a limit of 10 annotations per step, so we only get 10 errors. But hopefully its a little easier than what we have now.


peterbarker (approved):
LGTM

PR #28910, opened 2024-12-19
AP_BattMonitor: Handle allocating too many analog channels

WickedShell:
If you over allocate the number of analog channels this results in a crash. It's easy to trigger this if you have voltage only monitors as we still eat up a current input channel, regarless of if we use it. There are only 16 channels at this time on ChibiOS, so if you have 9 voltage only battery monitors you are out.

This PR improves that situation by only allocating channels when needed, and in the case where we run out we now set a ConfigError, which on a flight controller is much more friendly then a instant segfault the moment we read a battery monitor. NOTE: on AP_Periph this takes the node off the bus, rather then just sitting in the bootloader. This was consideted acceptable as the current behaviour was to segfault and then sit in the bootloader, unless you made new firmware that limited the number of channels allocated it wasn't possible to recover in this situation anyways.

Tested on CubeOrangePlus with plane firmware, as well as on AP_Periph.

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor_Analog.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor_Analog.cpp
@@ -5,6 +5,7 @@
 #include <AP_HAL/AP_HAL.h>
 #include <AP_Common/AP_Common.h>
 #include <AP_Math/AP_Math.h>
+#include <AP_BoardConfig/AP_BoardConfig.h>
 
 #include "AP_BattMonitor_Analog.h"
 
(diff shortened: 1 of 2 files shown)


andyp1per (approved):
LGTM

peterbarker:
How do you recover the periph if it never comes up on the bus?

IamPete1 (changes requested):
This line needs deleting from `AP_BattMonitor_Synthetic_Current`which inherits from `AP_BattMonitor_Analog`

https://github.com/ArduPilot/ardupilot/blob/613b835d67f817bdf9a1b864e9d988fd819efa03/libraries/AP_BattMonitor/AP_BattMonitor_Synthetic_Current.cpp#L62-L63

Looks good apart from that.

IamPete1 (approved):
LGTM

PR #28911, opened 2024-12-20
Heli AutoTune: fix rate and accel limiting

bnsgeyer:
The original implementation of this relied on the attitude controller to provide the rate and accel limiting. The rate limiting worked but it can cause square wave behavior as it clips the rate at the rate limit. As for the accel limiting, the attitude controller wasn’t designed with the feedforward turned off to limit accel. It actually reduces the angle P to reduce acceleration.  

So a different approach was taken. Using a sinusoidal input makes it easy to limit the acceleration and rate without causing any clipping.   The attitude amplitude was modified based on the first and second derivative of the amplitude sinusoid chirp. This ensures the desired rate and acceleration are limited. However, the target values maybe smaller or larger based on the error and angle P. In most cases I have seen the target rates are lower. 

This has been tested in SITL.

### DIFF
--- a/libraries/AC_AutoTune/AC_AutoTune_Heli.cpp
+++ b/libraries/AC_AutoTune/AC_AutoTune_Heli.cpp
@@ -528,9 +528,9 @@ void AC_AutoTune_Heli::load_test_gains()
             rate_test_max = orig_roll_rate;
             accel_test_max = tune_roll_accel;
         } else {
-            // have attitude controller use accel and rate limit parameter
-            rate_test_max = rate_max;
-            accel_test_max = accel_max;
+            // have attitude controller not perform rate limiting and angle P scaling based on accel max
+            rate_test_max = 0.0;
+            accel_test_max = 0.0;
         }
         if (tune_type == SP_UP || tune_type == TUNE_CHECK) {
             rate_i = tune_roll_rff*AUTOTUNE_FFI_RATIO_FINAL;
@@ -552,9 +552,9 @@ void AC_AutoTune_Heli::load_test_gains()
             rate_test_max = orig_pitch_rate;
             accel_test_max = tune_pitch_accel;
         } else {
-            // have attitude controller use accel and rate limit parameter
-            rate_test_max = rate_max;
-            accel_test_max = accel_max;
+            // have attitude controller not perform rate limiting and angle P scaling based on accel max
+            rate_test_max = 0.0;
+            accel_test_max = 0.0;
         }
         if (tune_type == SP_UP || tune_type == TUNE_CHECK) {
             rate_i = tune_pitch_rff*AUTOTUNE_FFI_RATIO_FINAL;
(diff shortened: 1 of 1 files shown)


bnsgeyer:
@rmackay9 @MattKear @lthall can i get someone to review this.  I would like to get it merged so I can backport it to 4.6. I have tested it on a real heli and it works as designed.

libraries/AC_AutoTune/AC_AutoTune_Heli.cpp
```diff
@@ -811,7 +811,22 @@ void AC_AutoTune_Heli::dwell_test_run(sweep_info &test_data)
     }
 
     if (settle_time == 0) {
-        target_angle_cd = -chirp_input.update((now - dwell_start_time_ms) * 0.001, degrees(tgt_attitude) * 100.0f);
+        dwell_freq = chirp_input.get_frequency_rads();
+        float tgt_att_limited = tgt_attitude;
+        if (is_positive(dwell_freq)) {
+            float tgt_att_temp = tgt_attitude;
```
lthall, line 817: Why do we need the extra variable?

libraries/AC_AutoTune/AC_AutoTune_Heli.cpp
```diff
@@ -577,9 +577,9 @@ void AC_AutoTune_Heli::load_test_gains()
             rate_test_max = orig_yaw_rate;
             accel_test_max = tune_yaw_accel;
         } else {
-            // have attitude controller use accel and rate limit parameter
-            rate_test_max = rate_max;
-            accel_test_max = accel_max;
+            // have attitude controller not perform rate limiting and angle P scaling based on accel max
+            rate_test_max = 0.0;
```
lthall, line 581: If we are limiting the chirp then we shouldn't be hitting these limits unless there is a plot input. 

I am not very familiar with how you handle pilot input here but it may be dangerous if you don't have acceleration limiting and do accept pilot input.

lthall (approved):
I have posed some questions but it all looks correct to me.

libraries/AC_AutoTune/AC_AutoTune_Heli.cpp
bnsgeyer: @lthall Pilot input is not allowed during the autotune input.  It works just like multi autotune where once the pilot makes inputs, the autotune maneuver is stopped.

libraries/AC_AutoTune/AC_AutoTune_Heli.cpp
bnsgeyer: @lthall I think it is needed to hold the rate limited attitude.  If the rate limit is not set then it will skip setting tgt_att_temp variable.  I guess I could just use the tgt_att_limited variable throughout and keep setting it from the MIN function.

bnsgeyer:
@lthall I think you should consider adding the limiting the way i have done it here to the chirp function and SysID mode.   Once it is added then I would modify my code to just use it.  I did it in the Heli autotune just because I wanted to ensure this got into 4.6 since the other technique was not effective.

bnsgeyer:
@rmackay9 If there are no issues here from your end, I would like to merge this since I got Leonard's approval.

rmackay9 (approved):
Hi @bnsgeyer, it's OK to me to merge, thanks!

rmackay9:

PR #28912, opened 2024-12-20
AP_Scripting: add a simple serial dump script

peterbarker:
the existing example seems to do *way* too much work so has trouble capturing at higher baud rates.


tpwrules:
Ports have `readstring` as of 4.6 which would make things massively more efficient.

libraries/AP_Scripting/examples/Simple_Serial_Dump.lua
```diff
@@ -0,0 +1,35 @@
...
+function update() -- this is the loop which periodically runs
+
+  local n_bytes = math.min(port:available(), 512)
+  file = io.open(file_name, "a")
+  while n_bytes > 0 do
+      c = port:read()
+      if c == nil then
+          break
+      end
+      file:write(string.char())
```
tpwrules, line 27:
```suggestion
      file:write(string.char(c))
```

peterbarker: Question.

How was this working with this problem in here?!

The data coming out looked valid to me (it's in my LD06 PR).

peterbarker:
> Ports have `readstring` as of 4.6 which would make things massively more efficient.

I'm guessing we're not running that....

peterbarker:
We could potentially look to see if we have the readstring binding and use it if available, 'though.

I'll need to remove "Simple" from the script name then....

libraries/AP_Scripting/examples/Simple_Serial_Dump.lua
tpwrules: I don't know, that would imply a really interesting bug in the Lua interpreter. I couldn't find the PR you mentioned, but I assume you fixed it and forgot to update the code in Git.

Also `read` returns -1 on error, not `nil`.

And it's strange to open and close the file every loop. Probably better to simply flush it every n loops.

PR #28913, opened 2024-12-20
Create separate option for bridging mcast and hw can

bugobliterator:
* Bridging between mcast CAN and HW CAN needs to be turned off by default. The reason is if there are more than 2 devices on the same HW CAN and Ethernet network that will cause network loops and potentially choke the CAN bus.
* Bridging needs to be always turned off for Bootloaders, and in the case of AP_Periph and ardupilot, it should be enabled by the user.

Changes done:
* Change option name from  CAN1_MCAST_GATEWAY to CAN1_MCAST_ENDPOINT in NET_OPTIONS.
* Adds another option CAN1_MCAST_BRIDGED to enable bridging between MCAST and Hw CAN.
* Bridging option can be changed and will take effect on the same run, reboot is not required.


libraries/AP_HAL/CANIface.h
```diff
@@ -262,19 +262,20 @@ class AP_HAL::CANIface
...
     virtual int8_t get_iface_num() const = 0;
-    virtual bool add_to_rx_queue(const CanRxItem &rx_item) = 0;
-
     struct {
 #ifndef HAL_BOOTLOADER_BUILD
         HAL_Semaphore sem;
 #endif
         // allow up to 3 callbacks per interface
         FrameCb cb[3];
+        bool rx_cb_disable[3];
```
tridge, line 278: why do we need these disable flags? if we want to disable, why don't we just call unregister_frame_callback()?

bugobliterator: @tridge the disable is only to remove calling on rx, but we still call the callback on tx from CANIface. Basically we don't forward what we receive from CAN Bus, but we still transmit frames that are to be sent out. That's how we leave the interaction with the device itself functional. Unregistering will remove the call from both tx and rx.

tridge:
@bugobliterator could this be handled in the callback in the AP_Networking layer?

bugobliterator:
> @bugobliterator could this be handled in the callback in the AP_Networking layer?

I don't think there's a way to do this without changes to CANIface, there's a common callback for both RX and TX. We have to either pass the direction as callback parameter, make it part of each frame or the way I have done which is to have the direction selection per callback. First method will require wider changes, second one costs more memory and third one I think is a happy middle.

libraries/AP_Networking/AP_Networking.cpp
```diff
@@ -200,13 +200,13 @@ void AP_Networking::init()
 #endif
 
 #if AP_NETWORKING_CAN_MCAST_ENABLED && !defined(HAL_BOOTLOADER_BUILD)
```
peterbarker, line 202:
```suggestion
#if AP_NETWORKING_CAN_MCAST_BRIDGING_ENABLED
```

libraries/AP_HAL/CANIface.cpp
```diff
@@ -64,9 +64,9 @@ int16_t AP_HAL::CANIface::receive(CANFrame& out_frame, uint64_t& out_ts_monotoni
 #ifndef HAL_BOOTLOADER_BUILD
     WITH_SEMAPHORE(callbacks.sem);
 #endif
-    for (auto &cb : callbacks.cb) {
-        if (cb != nullptr) {
-            cb(get_iface_num(), out_frame);
```
tridge, line 69: adding an bool is_send as an additional field in the callback
maybe change
```
    FUNCTOR_TYPEDEF(FrameCb, void, uint8_t, const AP_HAL::CANFrame &);
```
to this:
```
    FUNCTOR_TYPEDEF(FrameCb, void, uint8_t, const AP_HAL::CANFrame &, uint8_t flags);
```

libraries/AP_Networking/AP_Networking.h
```diff
@@ -157,14 +157,23 @@ class AP_Networking
...
+        CAN2_MCAST_ENDPOINT=(1U<<2),
+        CAN1_MCAST_BRIDGED=(1U<<3),
+        CAN2_MCAST_BRIDGED=(1U<<4),
 #endif
     };
     bool option_is_set(OPTION option) const {
         return (param.options.get() & int32_t(option)) != 0;
     }
 
+#if AP_NETWORKING_CAN_MCAST_ENABLED
```
peterbarker, line 170: This would make it onto the bootloader?  Maybe elided?

Separate define as suggested would make this clearly not a problem.

libraries/AP_HAL/CANIface.cpp
tridge: or could be a bool

libraries/AP_Networking/AP_Networking_CAN.cpp
```diff
@@ -122,6 +121,14 @@ void AP_Networking_CAN::mcast_server(void)
             GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "CAN_MCAST[%u]: failed to register", unsigned(bus));
             goto de_allocate;
         }
+#ifndef HAL_BOOTLOADER_BUILD
```
peterbarker, line 124:
```suggestion
#if AP_NETWORKING_CAN_MCAST_BRIDGING_ENABLED
```

libraries/AP_Networking/AP_Networking_CAN.cpp
```diff
@@ -180,15 +188,28 @@ void AP_Networking_CAN::mcast_server(void)
             */
             AP_HAL::CANFrame frame;
             const uint16_t timeout_us = 2000;
-
+#ifndef HAL_BOOTLOADER_BUILD
```
peterbarker, line 191: ... use new define...

peterbarker (dismissed):

libraries/AP_HAL/CANIface.cpp
bugobliterator: Moved to using CANIOFlags actually to keep things consistent.

Also changed IsMAVCAN flag to IsForwardedFrame. This flag is then used to check if the frame was forwarded from different physical iface.

libraries/AP_Networking/AP_Networking.cpp
bugobliterator: done

libraries/AP_Networking/AP_Networking.h
bugobliterator: done using AP_NETWORKING_CAN_MCAST_BRIDGING_ENABLED

libraries/AP_Networking/AP_Networking_CAN.cpp
bugobliterator: not needed

bugobliterator (commented):
requested changes made

libraries/AP_Networking/AP_Networking_CAN.cpp
```diff
@@ -203,12 +214,27 @@ void AP_Networking_CAN::mcast_server(void)
...
+#if AP_NETWORKING_CAN_MCAST_BRIDGING_ENABLED
+    // check if bridged mode is enabled
+    const bool bridged = AP::network().is_can_mcast_ep_bridged(bus);
+#else
+    // never bridge in bootloader, as we can cause loops if multiple
+    // bootloaders with mcast are running on the same network and CAN Bus
+    const bool bridged = false;
+#endif
+
+    if (flags & AP_HAL::CANIface::IsForwardedFrame && !bridged) {
```
tridge, line 232: I think we need parentheses around the bitmask test, or some compilers will complain

tridge (changes requested):
this looks much better, thank you!
just the one trivial fix needed then should be merged

libraries/AP_Networking/AP_Networking_CAN.cpp
bugobliterator: fixed

tridge (approved):

libraries/AP_Networking/AP_Networking.h
```diff
@@ -157,14 +157,23 @@ class AP_Networking
     enum class OPTION {
         PPP_ETHERNET_GATEWAY=(1U<<0),
 #if AP_NETWORKING_CAN_MCAST_ENABLED
-        CAN1_MCAST_GATEWAY=(1U<<1),
-        CAN2_MCAST_GATEWAY=(1U<<2),
+        CAN1_MCAST_ENDPOINT=(1U<<1),
+        CAN2_MCAST_ENDPOINT=(1U<<2),
+        CAN1_MCAST_BRIDGED=(1U<<3),
```
peterbarker, line 162: These could also be protected by `AP_NETWORKING_CAN_MCAST_ENABLED` to catch places that are trying to handle bridged stuff when they shouldn't.

libraries/AP_Networking/AP_Networking_CAN.cpp
```diff
@@ -180,18 +180,29 @@ void AP_Networking_CAN::mcast_server(void)
...
                 if (cbus == nullptr) {
                     break;
                 }
-                auto retcode = cbus->send(frame,
-                                                      AP_HAL::micros64() + timeout_us,
-                                                      AP_HAL::CANIface::IsMAVCAN);
-                if (retcode == 0) {
-                    break;
+                if (bridged) {
+                    auto retcode = cbus->send(frame, AP_HAL::micros64() + timeout_us,
```
peterbarker, line 195: This is dead code on bootloader... I wonder if the compiler is smart enough to elide..
.

libraries/AP_Networking/AP_Networking_Config.h
```diff
@@ -107,6 +107,10 @@
 #define AP_NETWORKING_CAN_MCAST_ENABLED 0
 #endif
 
+#ifndef AP_NETWORKING_CAN_MCAST_BRIDGING_ENABLED
+#define AP_NETWORKING_CAN_MCAST_BRIDGING_ENABLED AP_NETWORKING_CAN_MCAST_ENABLED && !defined(HAL_BOOTLOADER_BUILD)
```
peterbarker, line 111: The default for the bootloader should be in `libraries/AP_HAL_ChibiOS/hwdef/scripts/defaults_bootloader.h`

libraries/AP_Networking/AP_Networking_CAN.cpp
bugobliterator: probably, but going to play it safe and forcing it out.

libraries/AP_Networking/AP_Networking_Config.h
bugobliterator: fixed

libraries/AP_Networking/AP_Networking.h
peterbarker: > did you mean AP_NETWORKING_CAN_MCAST_BRIDGING_ENABLED ?

Yes, sorry

peterbarker (approved):
LGTM

rmackay9:
I was unable to include this in AP-4.6.0-beta4 due to merge conflicts.  If we're keen to get this backported could @bugobliterator or someone else perhaps create a backport PR?

bugobliterator:
> I was unable to include this in AP-4.6.0-beta4 due to merge conflicts. If we're keen to get this backported could @bugobliterator or someone else perhaps create a backport PR?

@rmackay9 backport pr is here https://github.com/ArduPilot/ardupilot/pull/29297

PR #28914, opened 2024-12-20
AP_RangeFinder: Hexsoon 77G 40M/80M MWW support

rmackay9:
This adds support for two of the Hexsoon radar as a rangefinder:

- [HS 77G 40M MMW](http://www.hexsoon.com/en/product/product-87-23.html) (40m range)
- [HS 77G 80M MMW](http://www.hexsoon.com/en/product/product-7-973.html) (80m range)

hexsoon-hs77g-autopilot-option2

This has been lightly tested on real hardware.  If anyone else wishes to test it the parameter settings are:

- CAN_P1_DRIVER = 1
- CAN_P1_BITRATE = 500000
- CAN_D1_PROTOCOL = 15
- RNGFND1_TYPE = 44
- RNGFND1_MAX_CM = 4000

A few questions that will likely come up as part of the review:

1. The radar provides distances for 3 sectors similar to some other drivers should this sensors driver also be a Proximity sensor?  Should we try and support using this radar as both a RangeFinder and a Proximity sensor?
2. Does it need to be written in C++ or would Lua be sufficient?  I think Hexsoon would prefer it was written in C++ to reduce the barrier for users

This work was sponsored by Hexsoon


rmackay9:
Closed because the alternative has been merged

PR #28916, opened 2024-12-20
 Add new boards ZeroOne_Air flight controller

ZeroOne-Aero:
AP_HAL_ChibiOS: Added support for ZeroOne_Air flight controller


Hwurzburg:
@ZeroOne-Aero this is going to need a bootloader in Tools/bootloaders
just a quick scan, not a detailed review: use the board test name instead of board number and use it as the folder names so that board naming is consistent  ie ZeroOne_X6 rather than ZeroOneX6_Air

When you are ready I will review

ZeroOne-Aero:
@Hwurzburg Thank you very much for your reply during the holidays, we will add a bootloader in Tools/bootloaders later, ZeroOneX6_Air is the name of our new board, I don't fully understand what  build_types name means

PR #28920, opened 2024-12-20
autotest: Added Copter Smart RTL repeat detection test

MichelleRos:
This test is to demonstrate a bug that was found in Copter's Smart RTL where if it's left in circle mode for a long time it apparently doesn't always detect the loop, so Smart RTL still runs out of space after a while.

Added as a disabled test, to be enabled by whoever fixes the bug.

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -9976,6 +9976,15 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
         self.change_mode('ALT_HOLD')
         self.change_mode('SMART_RTL')
 
+    def SMART_RTL_Repeat(self):
+        '''Test whether Smart RTL catches the repeat'''
+        self.takeoff(alt_min=10, mode='GUIDED')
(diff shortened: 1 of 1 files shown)


Tools/autotest/arducopter.py
```diff
@@ -9976,6 +9976,15 @@ def SMART_RTL_EnterLeave(self):
         self.change_mode('ALT_HOLD')
         self.change_mode('SMART_RTL')
 
+    def SMART_RTL_Repeat(self):
+        '''Test whether Smart RTL catches the repeat'''
+        self.takeoff(alt_min=10, mode='GUIDED')  
+        self.set_rc(3, 1500)
```
peterbarker, line 9982:
```suggestion
        self.hover()
```
... that could be useful if you're running a frame other than the default, boring one.

We don't do this in many places yet, and it doesnt' actually do anything but set it to 1500... but aspirationally...

peterbarker (approved):
LGTM!

PR #28923, opened 2024-12-21
AP_Periph: release notes for 1.8.0

rishabsingh3003:


peterbarker:
Did these release notes get used?

1.8.0 went out the door, and this is conflicting now.

I understand in the future we will be syncing AP_Periph releases with ArduPilot's.....

rmackay9:
We should consider adding the AP_Periph 1.8 release to releases section here: https://github.com/ArduPilot/ardupilot/releases

PR #28930, opened 2024-12-21
Plane: remove mode guided pointer comparisons. 

IamPete1:
This is a step towards allowing custom modes in plane. Custom modes should inherit behaviour from guided, so all comparisons by pointer or number need to be removed. Removing number comparisons will be a future PR (or PRs)


peterbarker (approved):
LGTM

tridge:
as discussed on the dev call, we will create a new OFFBOARD mode for offboard control, and GUIDED will be only for fly-to functionality under human remote pilot control

PR #28933, opened 2024-12-23
Avoid including AP_InertialSensor.h in headers

peterbarker:
increases amount of work compiler needs to do for no good reason

... particularly when including in AP_AHRS.h, which is used a lot.

```
CubeOrange-periph-heavy  *                                                                     
CubeRedPrimary                      8      *           0       0                 -16    0      16
Durandal                            -8     *           -24     -24               -32    -24    0
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     16     *           0       -8                -16    -8     16
MatekF405                           0      *           8       16                0      16     0
Pixhawk1-1M-bdshot                  0                  8       16                -8     8      0
bebop                               32                 16      48                -4     -16    -16
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           0      *           8       8                 0      8      0
skyviper-v2450                                         8                                       
```

### DIFF
--- a/libraries/AP_AHRS/AP_AHRS_Backend.h
+++ b/libraries/AP_AHRS/AP_AHRS_Backend.h
@@ -23,7 +23,6 @@
 #include <AP_Math/AP_Math.h>
 #include <inttypes.h>
 #include <AP_Airspeed/AP_Airspeed.h>
-#include <AP_InertialSensor/AP_InertialSensor.h>
 #include <AP_Common/Location.h>
 #include <AP_NavEKF/AP_NavEKF_Source.h>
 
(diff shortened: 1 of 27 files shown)


tpwrules:
Is this a measurable time save? Allowing inline isn't a bad reason. Of course the difference doesn't appear huge, and I am assuming these are not methods on the hot path.

peterbarker:
> Is this a measurable time save? Allowing inline isn't a bad reason. Of course the difference doesn't appear huge, and I am assuming these are not methods on the hot path.

They are also methods of dubious implementation ("first usable" vs "primary")....  We're also not particularly good at being consistent about whether the accel vector is a product of the AHRS or ins library.

I have not attempted to measure time deltas.  I think it's pretty clear that we won't form needless header dependencies during compilation, on top of any time savings from not including them - and that makes cache hits more likely.

tpwrules:
Okay, it seems those methods are very rarely used anyway, maybe it would be better to just kill them?

I assume you are testing all the build options, that's the only other possibility of issue I can see.

libraries/AP_GyroFFT/AP_GyroFFT.h
```diff
@@ -221,9 +220,7 @@ class AP_GyroFFT
     // whether analysis can be run again or not
     bool start_analysis();
     // return samples available in the gyro window
-    uint16_t get_available_samples(uint8_t axis) {
```
andyp1per, line 224: Please don't do this, this is in the hot path and inlining helps

libraries/AP_AHRS/AP_AHRS_Backend.h
```diff
@@ -76,22 +75,10 @@ class AP_AHRS_Backend
...
-#if AP_INERTIALSENSOR_ENABLED
-        return AP::ins().get_first_usable_accel();
-#else
-        return 0;
-#endif
-    }
+    virtual uint8_t get_primary_accel_index(void) const;
 
     // get the index of the current primary gyro sensor
-    virtual uint8_t get_primary_gyro_index(void) const {
```
andyp1per, line 88: Simply un-inlining things requires a fair amount of testing, I suggest you don't do that in this PR

andyp1per (changes requested):
We use inline for a reason

PR #28934, opened 2024-12-23
Rename vehicle-specific RC_Channel and GCS_MAVLink files

peterbarker:
We are somewhat inconsistent with our file naming for these derived classes.

So, for example, we have GCS_Plane and GCS_Copter which contain the classes `GCS_Plane` and `GCS_Copter`, but we use `GCS_Mavlink.h` for both Plane and Copter to hold `GCS_MAVLINK_Plane` and `GCS_MAVLINK_Copter`.

This PR renames the RC_Channel and GCS_MAVLink files for the vehicles to be closer to the class names they contain.  The class name defines in GCS_MAVLink_Plane.cpp, however, is still `GCS_MAVLINK_Plane` rather than `GCS_MAVLink_Plane`.  There are 1500 lines to change to correct the case, and that's going to make backports difficult.  Probably something to do closer to the middle of a release cycle.


tpwrules:
I like this. It would be nice to interleave the commits to avoid so much of a bisect pit.

I assume the copter case would also require the same order of magnitude of changes?

rmackay9 (approved):
The renaming of files will make backports tricky but besides that the changes themselves are pretty small, thanks!

peterbarker:
> I like this. It would be nice to interleave the commits to avoid so much of a bisect pit (i.e. do each vehicle's commit in sequence rather than two blocks for all vehicles).

Good point, I've interleaved the commits now.

> I assume the copter case would also require the same order of magnitude of changes?

Not sure what you mean here.

The "1500" was a `git grep 'GCS_MAVLINK'` in the base directory of the checkout.  That gets the base class and its derivatives.

tpwrules:
> > I assume the copter case would also require the same order of magnitude of changes?
> 
> Not sure what you mean here.

You mentioned only Plane in the context of that 1500 number. It looks like the problem affects both Plane and Copter so that number is for both? Not quibbling about the count, just want to accurately know what vehicles would be affected.

peterbarker:
> > > I assume the copter case would also require the same order of magnitude of changes?
> > 
> > 
> > Not sure what you mean here.
> 
> You mentioned only Plane in the context of that 1500 number. It looks like the problem affects both Plane and Copter so that number is for both? Not quibbling about the count, just want to accurately know what vehicles would be affected.

The entire codebase.

Primarily 14 files containing the 7 classes, but there are going to be statics floating around, probably.

tridge (approved):

PR #28939, opened 2024-12-24
Offboard reserve mode ID

Ryanf55:
Reserve mode 30 for offboard so it's the same on all vehicles.

### DIFF
--- a/AntennaTracker/mode.h
+++ b/AntennaTracker/mode.h
@@ -13,6 +13,7 @@ public:
         GUIDED=4,
         AUTO=10,
         INITIALISING=16
+        // Mode number 30 reserved for "offboard" for external/lua control.
     };
 
     Mode() {}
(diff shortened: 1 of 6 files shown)


rmackay9:
Thanks for this. I'd vote for using mode 29 or 30.  29 is an odd number but that's OK.

Ryanf55:
> Thanks for this. I'd vote for using mode 29 or 30. 29 is an odd number but that's OK.

I'll do 30!

timtuxworth:
Does this have to line up with @IamPete1 's work on "scripted modes", where it will be possible to create new modes in Lua script that inherit from Guided?

Ryanf55:
> Does this have to line up with @IamPete1 's work on "scripted modes", where it will be possible to create new modes in Lua script that inherit from Guided?

Yea that's the idea. We need to work together on a long term design, this is just a reservation if we choose to stop using guided for offboard on plane, and make it consistent between all the vehicles.

rmackay9 (approved):
thanks!

tridge (approved):

PR #28942, opened 2024-12-24
AP_Mount: add CADDX gimbal support

rmackay9:
This is a simplified version of @Hwurzburg's previous PR https://github.com/ArduPilot/ardupilot/pull/28373 to add support for the [CADDX GM3 gimbal](https://caddxfpv.com/products/caddxfpv-gm1-gm2-gm3?variant=48924891611438)

This simply adds support for the gimbal without trying to support the extra lock types that the gimbal provides.  These extra lock types add the ability to control each axis (roll, pitch and yaw) in earth or body frame independently.  For example roll control could be in body frame while pitch is in earth frame.  This new ability is **not** supported in this PR.

This has been tested on real hardware and seems OK.

The corresponding wiki PR is here https://github.com/ArduPilot/ardupilot_wiki/pull/6441


Hwurzburg:
why drop the lock types and control?

rmackay9:
Hi @Hwurzburg,

I dropped the lock types and control because I couldn't get them to work and also they are really an enhancement so they naturally fall into a follow-up PR that could potentially affect other gimbals as well.  The Siyi gimbal for example also supports FPV mode and users have been asking for that for a long time so we could do the two together

Lalbabu2001:

Also add support for PPM .
Thanks.

rmackay9:
Hi @Lalbabu2001,

I've created an issue here to capture this enhancement request https://github.com/ArduPilot/ardupilot/issues/28945.  No promises on when/if it will be done though.  I'm not personally very experienced porting AP to new boards.

libraries/AP_Mount/AP_Mount_CADDX.h
```diff
@@ -0,0 +1,54 @@
+/*
+  CADDX mount using serial protocol backend class
+ */
+#pragma once
+
+#include "AP_Mount_Backend_Serial.h"
```
peterbarker, line 6:
```suggestion
#include "AP_Mount_Backend_config.h"
```

libraries/AP_Mount/AP_Mount_CADDX.h
```diff
@@ -0,0 +1,54 @@
...
+
+#if HAL_MOUNT_CADDX_ENABLED
+
+#include <AP_HAL/AP_HAL.h>
```
peterbarker, line 10: #include "AP_Mount_Backend_Serial.h"

libraries/AP_Mount/AP_Mount_CADDX.h
```diff
@@ -0,0 +1,54 @@
...
+protected:
+
+    // get attitude as a quaternion.  returns true on success
+    bool get_attitude_quaternion(Quaternion& att_quat) override;
+
+private:
+
+    // lock mode enum
+    enum class LockMode {
+        GIMBAL_MODE_FOLLOW = 0,
```
peterbarker, line 42:
```suggestion
        FOLLOW = 0,
```

etc

peterbarker (dismissed):
partial on-call review only :-)

libraries/AP_Mount/AP_Mount_CADDX.h
rmackay9: done, thanks for the review!

libraries/AP_Mount/AP_Mount_CADDX.h
rmackay9: txs,done

libraries/AP_Mount/AP_Mount_CADDX.h
rmackay9: done, txs

libraries/AP_Mount/AP_Mount_CADDX.h
```diff
@@ -0,0 +1,55 @@
...
+  CADDX mount using serial protocol backend class
+ */
+#pragma once
+
+#include "AP_Mount_config.h"
+
+#if HAL_MOUNT_CADDX_ENABLED
+
+#include "AP_Mount_Backend_Serial.h"
+#include <AP_HAL/AP_HAL.h>
```
peterbarker, line 11:
```suggestion
```

libraries/AP_Mount/AP_Mount_CADDX.h
```diff
@@ -0,0 +1,55 @@
...
+#include <AP_Math/AP_Math.h>
```
peterbarker, line 12:
```suggestion
```

libraries/AP_Mount/AP_Mount_CADDX.h
```diff
@@ -0,0 +1,55 @@
...
+    // has_roll_control - returns true if this mount can control its roll
+    bool has_roll_control() const override { return roll_range_valid(); };
+
+    // has_pitch_control - returns true if this mount can control its tilt
+    bool has_pitch_control() const override { return pitch_range_valid(); };
+
+protected:
+
+    // get attitude as a quaternion.  returns true on success
+    bool get_attitude_quaternion(Quaternion& att_quat) override;
```
peterbarker, line 37:
```suggestion
    bool get_attitude_quaternion(class Quaternion& att_quat) override;
```

Not actually including the right header to get this - just getting it transitively somehow

libraries/AP_Mount/AP_Mount_CADDX.h
```diff
@@ -0,0 +1,55 @@
...
+
+private:
+
+    // lock mode enum
+    enum class LockMode {
+        FOLLOW = 0,
```
peterbarker, line 43: This doesn't make a lot of sense!
```suggestion
```

peterbarker (changes requested):
Looks pretty good.

libraries/AP_Mount/AP_Mount_CADDX.h
rmackay9: removed include, thanks

libraries/AP_Mount/AP_Mount_CADDX.h
rmackay9: removed include, thanks

libraries/AP_Mount/AP_Mount_CADDX.h
rmackay9: Hi @peterbarker, thanks for the review.  I've removed this line because we don't use it but I can imagine that we could use it as a default value so not sure why it doesn't makes sense?

libraries/AP_Mount/AP_Mount_CADDX.h
rmackay9: I'm going to skip this change I think because it leads to a compile failure and would make it inconsistent with the virtual method that it is overriding and also all other drivers.  Apologies, I'm probably misunderstanding something

libraries/AP_Mount/AP_Mount_CADDX.h
rmackay9: Actually I've put this back because AP_Math.h is where "quaternion" is included

libraries/AP_Mount/AP_Mount_CADDX.h
peterbarker: Generally we don't use the same enumeration which is giving names to bit-offsets for *combinations* of those bits (AKA a "mask").

It would be nice to have some type-safety around this sort of thing, but I've never managed to propose a PR to add that....

libraries/AP_Mount/AP_Mount_CADDX.h
peterbarker:
```
diff --git a/libraries/AP_Mount/AP_Mount_CADDX.h b/libraries/AP_Mount/AP_Mount_CADDX.h
index fe423f55459..3d831209bf7 100644
--- a/libraries/AP_Mount/AP_Mount_CADDX.h
+++ b/libraries/AP_Mount/AP_Mount_CADDX.h
@@ -8,8 +8,7 @@
 #if HAL_MOUNT_CADDX_ENABLED
 
 #include "AP_Mount_Backend_Serial.h"
-#include <AP_Math/AP_Math.h>
-#include <AP_Common/AP_Common.h>
+#include <AP_Math/quaternion.h>
 
 class AP_Mount_CADDX : public AP_Mount_Backend_Serial
 {
```

libraries/AP_Mount/AP_Mount_CADDX.cpp
```diff
@@ -0,0 +1,172 @@
...
+    // ensure we have enough space to send the packet
+    if (_uart->txspace() < SET_ATTITUDE_BUF_SIZE) {
+        return;
+    }
+
+    // calculate roll, pitch, yaw angles in range 0 to 4096
+    const float scalar = AXIS_MAX / M_2PI;
+    uint16_t roll_target_cmd = constrain_uint16(wrap_2PI(angle_target_rad.roll) * scalar, AXIS_MIN, AXIS_MAX);
+    uint16_t pitch_target_cmd = constrain_uint16(wrap_2PI(angle_target_rad.pitch) * scalar, AXIS_MIN, AXIS_MAX);
+    uint16_t yaw_target_cmd = constrain_uint16(wrap_2PI(angle_target_rad.get_bf_yaw()) * scalar, AXIS_MIN, AXIS_MAX);
```
peterbarker, line 125:
```suggestion
    const uint16_t roll_target_cmd = constrain_uint16(wrap_2PI(angle_target_rad.roll) * scalar, AXIS_MIN, AXIS_MAX);
    const uint16_t pitch_target_cmd = constrain_uint16(wrap_2PI(angle_target_rad.pitch) * scalar, AXIS_MIN, AXIS_MAX);
    const uint16_t yaw_target_cmd = constrain_uint16(wrap_2PI(angle_target_rad.get_bf_yaw()) * scalar, AXIS_MIN, AXIS_MAX);
```

peterbarker (approved):
LGTM

libraries/AP_Mount/AP_Mount_CADDX.cpp
rmackay9: ha ha, thanks.  I was wondering if someone would notice the missing const .. added!

libraries/AP_Mount/AP_Mount_CADDX.h
rmackay9: ok thanks.  Updated to remove AP_Common.h and change to <AP_Math/quaternion.h>.  I'm surprised we don't need AP_Common.h in there.  We include that in every other AP_Mount backend so I guess we could remove it from many places

rmackay9:
I've updated with @peterbarker's requests and marked as "MergeOnCIPass"

libraries/AP_Mount/AP_Mount_CADDX.h
peterbarker: > ok thanks. Updated to remove AP_Common.h and change to <AP_Math/quaternion.h>. I'm surprised we don't need AP_Common.h in there. We include that in every other AP_Mount backend so I guess we could remove it from many places

Probably, yes - and that will save compile time, which everybody likes.

We're getting away with some shenanigans by having those all around the place - things like the stdint.h include come to mind :-)

peterbarker:
Merged it!

rmackay9:

rmackay9:

PR #28943, opened 2024-12-24
ArduPlane: add guided timeout

Ryanf55:
# Purpose

Allow configurable guided timeout. For an application expecting to send 50hz guided velocity commands, if the offboard controller dies, ArduPlane should fall back to loiter much sooner. This allows a user to bump up the expected controller rate.

# Difference from copter

The parameter is the same, but this one calls `constrain_float` on the parameter. Copter only enforces the minimum value.
I also assumed that the plane will fall back to a guided loiter when control ends. Perhaps copter should enforce the max? 

I also bumped the minimum period down to 0.01s. Copter doesn't seem to expect anything greater than 100mS between commands, but we know people operate controllers much faster than that. Perhaps copter's minimum should be bumped down too.

# Testing Performed

* N/A


ArduPlane/Parameters.cpp
```diff
@@ -1303,6 +1303,14 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
...
     AP_GROUPINFO("RNGFND_LND_ORNT", 36, ParametersG2, rangefinder_land_orient, ROTATION_PITCH_270),
 #endif
+
+    // @Param: GUID_TIMEOUT
+    // @DisplayName: Guided mode timeout
+    // @Description: Guided mode timeout after which vehicle will return to guided loiter if no updates are received from caller. Only applicable during any combination of velocity, acceleration, angle control, and/or angular rate control
+    // @Units: s
+    // @Range: 0.01 5
+    // @User: Advanced
+    AP_GROUPINFO("GUID_TIMEOUT", 37, ParametersG2, guided_timeout, 3.0),
```
magicrub, line 1313: This param should live in mode_guided.cpp. See mode_takeoff as an example

magicrub (changes requested):

IamPete1:
We were talking about adding offboard mode. That would leave guided with only commands that don't timeout (ie positions), so maybe the param is better referring to the new mode.

ArduPlane/Parameters.cpp
```diff
@@ -1303,6 +1303,14 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("RNGFND_LND_ORNT", 36, ParametersG2, rangefinder_land_orient, ROTATION_PITCH_270),
 #endif
+
+    // @Param: GUID_TIMEOUT
```
timtuxworth, line 1307: All the existing guided parameters use the prefix GUIDED_,

timtuxworth:
This seems to be the same as my #28526

ArduPlane/Parameters.cpp
Ryanf55: That came from Copter. Looks like Plane and Copter chose different prefixes for the same mode.

ArduPlane/Parameters.cpp
Ryanf55: Done.

Ryanf55:
> We were talking about adding offboard mode. That would leave guided with only commands that don't timeout (ie positions), so maybe the param is better referring to the new mode.

How would you like it done? Can we call it GUIDED_TIMEOUT when backported to 4.6, and OFFBOARD_TIMEOUT on 4.7+ assuming we get offboard mode written?

ArduPlane/Parameters.cpp
Ryanf55: I matched `GUIDED_` prefix. For some reason the other GUIDED_ parameters are part of plane and not ModeGuided.

timtuxworth:
It's not really a "timeout" though. Pretty sure it doesn't do this:

> Allow configurable guided timeout. For an application expecting to send 50hz guided velocity commands, if the offboard controller dies, ArduPlane should fall back to loiter much sooner. This allows a user to bump up the expected controller rate.

I used it in Plane Follow because from my testing it acts to throttle the messages, so that it won't accept new guided commands if they arrive < this value after the previous message. That's why I called it "UPD_LIM" - it acts to limit the frequency of updates received.

Lalbabu2001:

Also add support for PPM .
Thanks.

Ryanf55:
> This seems to be the same as my #28526

It's different because
* I copied the implementation from Copter, reducing cognitive load on our users by keeping the param named the same and described  the same as Copter and Rover
* I enforce the limits of the parameter in the getter, you don't
* My param is part of guided mode instead of plane so if you were to make guided mode optional, Plane doesn't have the flash cost.
* You call the variable frequency_limit, but I prefer Copter's naming of calling it a timeout becuase that's what it does. If you don't send guided commands fast enough, it will fall back to a position loiter point.  

I agree with Pete's assessment here: https://github.com/ArduPilot/ardupilot/pull/28526#issuecomment-2456573675

Functionally, they achieve similar purposes to allow the plane to either
* Recognize the companion computer stopped sending low level commands and fall back to the safer position loiter more quickly
* Provide a way to reduce the timeout if you are sending multiple different low level control commands (Pete's optimization here make sense to add as a follow up for our offboard mode: https://github.com/ArduPilot/ardupilot/pull/28526#issuecomment-2460563783)

IamPete1:
Needs a re-base. 

My only concern here is that the param lives in guided mode and we were talking about removing support for any commands that need a timeout in guided and instead using a new off board mode. Once we have made that change then it make sense for the param to live in that mode.

IamPete1:
> Once we start moving things to offboard mode, is it possible to do a parameter conversion seamless for the users?

Yes, we can do the param conversion, does cost flash tho...

ArduPlane/Parameters.cpp
```diff
@@ -1028,6 +1028,10 @@ const AP_Param::Info Plane::var_info[] = {
     // @Path: mode_takeoff.cpp
     GOBJECT(mode_takeoff, "TKOFF_", ModeTakeoff),
 
+    // @Group: GUIDED_
+    // @Path: mode_guided.cpp
+    GOBJECT(mode_guided, "GUIDED_", ModeGuided),
```
tridge, line 1033: maybe OFBD_ ?

ArduPlane/mode_guided.cpp
```diff
@@ -1,6 +1,29 @@
 #include "mode.h"
 #include "Plane.h"
 
+/*
+  mode guided parameters
+ */
+const AP_Param::GroupInfo ModeGuided::var_info[] = {
+
+    // @Param: _TIMEOUT
```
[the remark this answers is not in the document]
Ryanf55: Hmm, that didn't solve the rest of these:
```
2025-01-11T19:49:08.0832958Z AT-0029.5: GUIDED_P not in XML
2025-01-11T19:49:08.0833366Z AT-0029.5: GUIDED_I not in XML
2025-01-11T19:49:08.0833769Z AT-0029.5: GUIDED_D not in XML
2025-01-11T19:49:08.0834167Z AT-0029.5: GUIDED_FF not in XML
2025-01-11T19:49:08.0834575Z AT-0029.5: GUIDED_IMAX not in XML
2025-01-11T19:49:08.0835003Z AT-0029.5: GUIDED_FLTT not in XML
2025-01-11T19:49:08.0835411Z AT-0029.5: GUIDED_FLTE not in XML
2025-01-11T19:49:08.0835824Z AT-0029.5: GUIDED_FLTD not in XML
2025-01-11T19:49:08.0836232Z AT-0029.5: GUIDED_SMAX not in XML
2025-01-11T19:49:08.0836846Z AT-0029.5: GUIDED_PDMX not in XML
2025-01-11T19:49:08.0837267Z AT-0029.5: GUIDED_D_FF not in XML
2025-01-11T19:49:08.0837677Z AT-0029.5: GUIDED_NTF not in XML
2025-01-11T19:49:08.0838077Z AT-0029.5: GUIDED_NEF not in XML
2025-01-11T19:49:10.2642142Z AT-0029.5: Exception caught: Downloaded parameters missing in XML
2025-01-11T19:49:10.2642884Z Traceback (most recent call last):
2025-01-11T19:49:10.2643958Z   File "/__w/ardupilot/ardupilot/Tools/autotest/vehicle_test_suite.py", line 8885, in run_one_test_attempt
2025-01-11T19:49:10.2645123Z     test_function(**test_kwargs)
2025-01-11T19:49:10.2646150Z   File "/__w/ardupilot/ardupilot/Tools/autotest/vehicle_test_suite.py", line 12458, in Parameters
2025-01-11T19:49:10.2647279Z     self.test_parameter_documentation()
2025-01-11T19:49:10.2648522Z   File "/__w/ardupilot/ardupilot/Tools/autotest/vehicle_test_suite.py", line 12451, in test_parameter_documentation
2025-01-11T19:49:10.2650006Z     self.test_parameter_documentation_get_all_parameters()
2025-01-11T19:49:10.2651437Z   File "/__w/ardupilot/ardupilot/Tools/autotest/vehicle_test_suite.py", line 2792, in test_parameter_documentation_get_all_parameters
2025-01-11T19:49:10.2652960Z     raise NotAchievedException("Downloaded parameters missing in XML")
2025-01-11T19:49:10.2654149Z vehicle_test_suite.NotAchievedException: Downloaded parameters missing in XML
2025-01-11T19:49:10.2654807Z 
```

PR #28944, opened 2024-12-24
feat: add send_named_int function and Lua binding

uxduck:
This is useful to send NAMED_VALUE_INT mavlink messages in the same way as NAMED_VALUE_FLOAT, including in lua scripts:
```lua
local an_integer = 1
gcs:send_named_int("EXAMPLE", an_integer)
```

I've confirmed this works by building Arduplane in Docker and then running it my lua script.


Lalbabu2001:

Also add support for PPM .
Thanks.

rmackay9:
Hi @Lalbabu2001, please do not spam the pull requests.  That's definitely not a good way to achieve what you'd like.

peterbarker (approved):
LGTM, but would be nice to have an example script and autotest.

peterbarker (changes requested):
Gah.  Just realised the commit messages need splitting.

Please run`./Tools/gittools/git-subsystems-split` and force-push your branch.

That example would still be nice ;-)

ben-xD:
That script doesn't seem to work on macOS (sed is slightly different?) but it worked on Debian.

I've also edited a simple script to include `send_named_int`

ben-xD:
@peterbarker I've updated the PR please have a look :)

ben-xD:
@peterbarker, I rebased onto the latest `master`, all checks have passed, could you take a look?

khancyr:
replaced by https://github.com/ArduPilot/ardupilot/pull/33409. thanks

PR #28948, opened 2024-12-26
AP_Bootloader: cope with very high latency DroneCAN links

tridge:
This fixes a round trip time calculation for timed out messages, which allows for updates of firmware over very high latency links (eg. fw update over support.ardupilot.org)


andyp1per (approved):
LGTM

PR #28951, opened 2024-12-27
Copter: remove stale parameter conversions

peterbarker:
this removes stale parameter conversions.

These patches are all in the Copter-4.0.0 tag, meaning that they are all for upgrading from a version prior to that to a version prior-or-equal-to Copter-4.0.0

Past this a user running a version of ArduCopter < 4.0.0 to something >= 4.7.0 will not have smooth parameter conversions

```
CubeOrange-periph-heavy  *                                                                      
CubeRedPrimary                      *      *           -1720   -3072              *      *      *
Durandal                            *      *           -1872   -3488              *      *      *
Hitec-Airspeed           *                                                                      
KakuteH7-bdshot                     *      *           -1720   -3080              *      *      *
MatekF405                           *      *           -1712   -3072              *      *      *
Pixhawk1-1M-bdshot                  *                  -1704   -3072              *      *      *
bebop                               *                  -556    -5108              *      *      *
f103-QiotekPeriph        *                                                                      
f303-Universal           *                                                                      
revo-mini                           *      *           -1712   -3072              *      *      *
skyviper-v2450                                         -1192                                    
```


ArduCopter/Parameters.cpp
```diff
@@ -1413,103 +1370,26 @@ void Copter::load_parameters(void)
...
-        { "PHLD_BRAKE_RATE", 4.0f },
-        { "PSC_ACCZ_P", 0.28f },
-        { "PSC_VELXY_D", 0.0f },
-        { "PSC_VELXY_I", 0.5f },
-        { "PSC_VELXY_P", 1.0f },
-        // PARAMETER_CONVERSION - Added: Jan-2019
-        { "RC8_OPTION", 32 },
-        // PARAMETER_CONVERSION - Added: Aug-2018
+        // PARAMETER_CONVERSION - Added: Aug-2020
         { "RC_OPTIONS", 0 },
```
rmackay9, line 1386: optionally we could probably drop this RC_OPTIONS linte too which is from Aug 2020

rmackay9 (approved):
LGTM but @bnsgeyer should approve as well because these are mostly trad heli changes

ArduCopter/Parameters.cpp
peterbarker: > optionally we could probably drop this RC_OPTIONS line too which is from Aug 2020

Yeah, I  really did consider it.  But they I couldn't use the pretty straight-forward 3.6.9 -> 4.x.x argument as to why it was time for these to go away :-)

I was also a little surprised at how much of the conversion code did go away.  I wonder if we're doing less of it now-adays...

peterbarker:
@rmackay9 I've fixed an error in this PR.  Currently in master there's a block of code which is responsible for setting parameter defaults on the heli frame - and it's within a method called `convert_pid_parameters`  (mostly not actually PID parameters either...).

I'd like to move that code out in a future PR - it's in the wrong place IMO.  Probably move it up, called around the comment "now setup some frame-class specific defaults"

bnsgeyer (approved):
I have no problem with this if the Lead Devs agree that this is the way we want to move out on parameter conversions.  I agree that someone upgrading from 3.6 to 4.5 is a large move.  Maybe we need to put a warning in the wiki?  Just a thought.

PR #28952, opened 2024-12-27
Revert "CI: disable ccache for MacOS builds"

peterbarker:
This reverts commit c4327161bce5e4ee8b68b6ff4cdc8d19c0d9afc5.

### DIFF
--- a/.github/workflows/macos_build.yml
+++ b/.github/workflows/macos_build.yml
@@ -161,6 +161,22 @@ jobs:
           fi
           Tools/environment_install/install-prereqs-mac.sh -y
           source ~/.bash_profile
+      # Put ccache into github cache for faster build
+      - name: Prepare ccache timestamp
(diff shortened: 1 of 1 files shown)


tpwrules:
Just checked that an old PR is now fixed. Homebrew or whomever must have fixed the dependency and this should pass.

tridge:
> Just checked that an old PR is now fixed. Homebrew or whomever must have fixed the dependency and this should pass.

it wasn't a dependency issue, ccache was segfaulting

tpwrules:
I figured it might be because of something wrong with a dependency because it's working now and its version hasn't changed recently.

peterbarker:
I'm marking this one as MergeOnCIPass on the basis that the segfaulting was probably something bad which github have fixed in their CI.

It has already passed CI once, I've force-pushed it.

PR #28955, opened 2024-12-27
autotest: add test for return-ot-initial-yaw RTL behaviour

peterbarker:


peterbarker:
@tpwrules autotest for that autoyaw behaviour we were discussing.

You have to change the `WP_YAW_BEHAVIOR` to get the functionality - the default value explicitly doesn't do it for RTL.

I have a feeling the comment in the code around pilot-input might be misleading/incorrect, but haven't checked.

tpwrules:
So you have to set `WP_YAW_BEHAVIOR` to 1 to get that behavior? Any idea why it's 2 by default?

Tools/autotest/arducopter.py
```diff
@@ -12311,6 +12311,22 @@ def ScriptingFlipMode(self):
         # Test done
         self.land_and_disarm()
 
+    def RTLYaw(self):
+        '''test that vehicle yaws to original heading on RTL'''
+        self.set_parameters({
+            'WP_YAW_BEHAVIOR': 3,  # 3 is WP_YAW_BEHAVIOR_LOOK_AHEAD
```
rmackay9, line 12317: Thanks for this.  I think WP_YAW_BEHAVIOR = 1 (Face next waypoint) could be a better choice because this is closer to the default behaviour which is 2 (Face next waypoint except RTL)

rmackay9:
@tpwrules,

2 (Face next waypoint except RTL) is the default because we thought that would be easier for new pilots who can be confused by the vehicle's heading changing.  A sudden RTL (caused by a failsafe) plus a heading change could make it hard for new pilots to control the vehicle

Tools/autotest/arducopter.py
peterbarker: Ah ,well.  Why not both? :-)

rmackay9 (approved):

Issue #28957, opened 2024-12-27
Plane: quadplane tilt rate issues


tridge:
There are 2 issues with the quadplane tilt down rate for tilt-enabled quadplanes.
- the quadplane.assisted_flight flag oscillates during the TRANSITION_TIMER phase of the fwd transition
- if Q_TRANSITION_MS is smaller than the time it takes to go from Q_TILT_MAX to full tilt with the configured Q_TILT_RATE parameters then the last part will happen at 90 deg/s

timtuxworth:
Is this what makes transitioning on a tilt rotor so hmmm - frustrating? It feels like it's trying hard to transition but won't quite get there? I haven't flow my VTBird for a while this is the last flight but I always had this feeling. 
BTW it was always with manually transitioning to FBWA - the auto mode transition always seems to be much smoother and I always wished I could duplicate it. Here's the log:
https://www.dropbox.com/scl/fi/tv8sl0va1xhdti691zrv2/log_154_2024-8-31-18-24-52.bin?rlkey=b7ujssije9iulhdgho9ky603m&dl=0

priseborough:
I think there is also a potential issue with the way that TECS can dip the throttle demand at the time that the tilting rotor throttle control hand over to TECS from the motors library. For vehicles where the propwash over the front wing contributes significantly to lift, we should be limiting throttle to no less than the trim value initially.

priseborough:
I've created a branch which contains a modification to the forward transition completion criteria to make it wait until rotor tilt forward has completed.

https://github.com/priseborough/ardupilot/tree/pr-quadplane-tilt-rate

I'll have a look at what we can do to prevent a step in throttle for those setups with a front wing  that benefits from prop-wash energisation at lower speeds.

priseborough:
See https://github.com/ArduPilot/ardupilot/pull/29696 for fix to throttle glitch.

PR #28960, opened 2024-12-29
Plane: tilt quadplane fixes

tridge:
this fixes some issues with tilt quadplanes. 
The first fix is to ensure we don't complete the transition till the rotors are fully tilted forward. The demanded VTOL throttle during the 2nd stage of the transition is also fixed to be based on the tilt angle, not the time.
The 2nd fix is to prevent oscillation of the assisted_flight flag

fixes some of the issues in https://github.com/ArduPilot/ardupilot/issues/28957


tridge:
@IamPete1 I've added a patch to allow THR_SLEWRATE to apply to tilted motors, by adding a new k_throttle_tilt actuator function

ArduPlane/tiltrotor.cpp
```diff
@@ -229,7 +231,9 @@ void Tiltrotor::continuous_update(void)
 
         max_change = tilt_max_change(false);
 
-        float new_throttle = constrain_float(SRV_Channels::get_output_scaled(SRV_Channel::k_throttle)*0.01, 0, 1);
+        SRV_Channels::set_output_scaled(SRV_Channel::k_throttle_tilt, SRV_Channels::get_output_scaled(SRV_Channel::k_throttle));
```
IamPete1, line 234: I think this should constrain to not be negative before the set. Normal throttle can be negative. 
```suggestion
        SRV_Channels::set_output_scaled(SRV_Channel::k_throttle_tilt, MAX(0, SRV_Channels::get_output_scaled(SRV_Channel::k_throttle)));
```

ArduPlane/tiltrotor.cpp
```diff
@@ -229,7 +231,9 @@ void Tiltrotor::continuous_update(void)
...
+        float new_throttle = constrain_float(SRV_Channels::get_slew_limited_output_scaled(SRV_Channel::k_throttle_tilt)*0.01, 0, 1);
+
         if (current_tilt < get_fully_forward_tilt()) {
             current_throttle = constrain_float(new_throttle,
```
IamPete1, line 238: This constrain could be achieved with the new slew limited function.

ArduPlane/tiltrotor.cpp
```diff
@@ -245,7 +249,7 @@ void Tiltrotor::continuous_update(void)
         }
         if (!quadplane.motor_test.running) {
             // the motors are all the way forward, start using them for fwd thrust
-            const uint16_t mask = is_zero(current_throttle)?0U:tilt_mask.get();
+            const uint16_t mask = tilt_mask.get();
```
IamPete1, line 252: I think will mean you get spin min in motor test rather than no motors spinning?

ArduPlane/tiltrotor.cpp
```diff
@@ -489,6 +494,9 @@ void Tiltrotor::tilt_compensate_angle(float *thrust, uint8_t num_motors, float n
             thrust[i] *= scale;
         }
     }
+    if (!fully_fwd()) {
+        SRV_Channels::set_output_scaled(SRV_Channel::k_throttle_tilt, largest_tilted*100);
+    }
```
IamPete1, line 499: I guess this is to make sure the slew starts off in the right place? I would prefer to see it done with the throttle output value from motors. You should be able to use motors thrust linearisation like we do for tailsitter here:

https://github.com/ArduPilot/ardupilot/blob/5f8a655f35a2a7901a56e898609a2d7c488a99dd/ArduPlane/tailsitter.cpp#L321-L323

libraries/SRV_Channel/SRV_Channel.h
```diff
@@ -189,6 +189,7 @@ class SRV_Channel {
         k_rcin15_mapped         = 154,
         k_rcin16_mapped         = 155,
         k_lift_release          = 156,
+        k_throttle_tilt         = 157,
```
IamPete1, line 192: Should we enforce that users don't assign this to a actual output? It does not abide by safety state for example (or E-stop).

IamPete1:
https://github.com/ArduPilot/ardupilot/pull/29011 is a alternate approach to get slew limiting for tiltrottor throttle outputs.

rmackay9:
@tridge This item is still on our 4.6.0 issues list so I wonder if we should try again to get this into master and then 4.6?

tridge:
> #29011 is a alternate approach to get slew limiting for tiltrottor throttle outputs.

I think that will impact the VTOL tune on tilt rotors?

libraries/SRV_Channel/SRV_Channel.h
tridge: i don't think that is necessary, for any particular vehicle type most of these are not available, and I didn't add it to the doc list, so won't appear in the dropdown options

ArduPlane/tiltrotor.cpp
tridge: it is inside a if() checking for !quadplane.motor_test.running, so I don't think it could impact motor test

ArduPlane/tiltrotor.cpp
tridge: done

ArduPlane/tiltrotor.cpp
```diff
@@ -150,6 +150,8 @@ void Tiltrotor::setup()
     }
     quadplane.transition = transition;
 
+    SRV_Channels::set_range(SRV_Channel::k_throttle_tilt, 100); // match k_throttle
```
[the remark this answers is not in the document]
tridge: yep, this is not the PR to make such a global change

ArduPlane/tiltrotor.cpp
tridge: sorry, I don't know what you mean

IamPete1:
> I think that will impact the VTOL tune on tilt rotors?

No, it will only affect forward flight when the `output_motor_mask` method is used.

PR #28965, opened 2024-12-30
ardupilotwaf: properly configure double precision tasks

tpwrules:
There might be multiple copies of the flags, so remove all of them. Also pull the setup into a helper function to avoid duplicating the logic. On ChibiOS boards, this fixes SITL (which had issues with position quantization) and affects the SBF GPS driver (which was probably fine before but does change slightly).

Sim on hardware now flies smoothly instead of jittering about all over the place. SBF requires a handful more bytes but I don't think this meaningfully affects the function based on a brief read-through; the important lat/lon casts were already done manually.

Many thanks to tridge for working with me to debug this. Assuming it should be backported so that SBF works as written even if the change in behavior is negligible.

### DIFF
--- a/Tools/ardupilotwaf/ap_library.py
+++ b/Tools/ardupilotwaf/ap_library.py
@@ -277,13 +277,7 @@ def double_precision_check(tasks):
             double_library = t.env.DOUBLE_PRECISION_LIBRARIES.get(src[0],False)
 
             if double_library or src in double_tasks:
-                t.env.CXXFLAGS = t.env.CXXFLAGS[:]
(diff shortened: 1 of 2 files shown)


andyp1per (approved):
Nice! Having worked through a couple of these I know how dark the magic is.

khancyr (approved):
LGTM, python code is correct.

can probably squash the commit in one.

For testing, you can probably look at target with double enabled and compare if both version produce the same binaries. The second test should be to look at a lib that got double enable and both version produce the same compiled code

tridge (approved):

rmackay9:

PR #28966, opened 2024-12-30
Correct compilation VOLZ_FEEDBACK is enabled 

peterbarker:
- lacked dependencies in `build_options.py`
 - flat-out syntax error when it does build

```
./Tools/autotest/test_build_options.py --define-match-glob=AP_DRONECAN_VOLZ_FEEDBACK_ENABLED
```


IamPete1 (approved):
LGTM

PR #28967, opened 2024-12-30
ArduPlane: Clarify behavior with small WP_LOITER_RAD

Ryanf55:
https://aviation.stackexchange.com/a/89380

If you set `WP_LOITER_RAD` too small, then your achieved loiter radius is `g * tan(ROLL_LIMIT_DEG) / v`

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -273,7 +273,7 @@ const AP_Param::Info Plane::var_info[] = {
 
     // @Param: WP_LOITER_RAD
     // @DisplayName: Waypoint Loiter Radius
-    // @Description: Defines the distance from the waypoint center, the plane will maintain during a loiter. If you set this value to a negative number then the default loiter direction will be counter-clockwise instead of clockwise.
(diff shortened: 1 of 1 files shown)


ArduPlane/Parameters.cpp
```diff
@@ -273,7 +273,7 @@ const AP_Param::Info Plane::var_info[] = {
 
     // @Param: WP_LOITER_RAD
     // @DisplayName: Waypoint Loiter Radius
-    // @Description: Defines the distance from the waypoint center, the plane will maintain during a loiter. If you set this value to a negative number then the default loiter direction will be counter-clockwise instead of clockwise.
+    // @Description: Defines the distance from the waypoint center, the plane will maintain during a loiter. If you set this value to a negative number then the default loiter direction will be counter-clockwise instead of clockwise. If this value is too close to zero, the achieved loiter radius will be driven by ROLL_LIMIT_DEG.
```
Hwurzburg, line 276:
```suggestion
    // @Description: Defines the distance from the waypoint center, the plane will maintain during a loiter. If you set this value to a negative number then the default loiter direction will be counter-clockwise instead of clockwise. If this value is too close to zero, the achieved loiter radius will be determined by ROLL_LIMIT_DEG.
```

Hwurzburg (changes requested):

Hwurzburg (approved):

PR #28971, opened 2024-12-30
AP_Filesystem: FATFS: remove redundant delay expectations

tpwrules:
The delays will be canceled on return by the EXPECT_DELAY_MS(3000) destructor at the start of the function. The current behavior will unexpectedly cancel delays from higher levels up the stack and is likely not what was intended.

Tested that filesystem operations still work on SD card on Cube Orange.

Is it possible to do filesystem ops from the main thread? Should that other EXPECT_DELAY_MS be removed too?

### DIFF
--- a/libraries/AP_Filesystem/AP_Filesystem_FATFS.cpp
+++ b/libraries/AP_Filesystem/AP_Filesystem_FATFS.cpp
@@ -254,7 +254,6 @@ static bool remount_file_system(void)
     }
     if (!sdcard_retry()) {
         remount_needed = true;
-        EXPECT_DELAY_MS(0);
         return false;
     }
     remount_needed = false;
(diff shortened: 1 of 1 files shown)


peterbarker (approved):
LGTM.

peterbarker:
> Is it possible to do filesystem ops from the main thread? Should that other EXPECT_DELAY_MS be removed too?

No to filesystem ops on the main thread.

You're going to have to be clearer on which "other EXPECT_DELAY_MS" you're referring to :-)

tpwrules:
The other one in this `remount_file_system` function. `EXPECT_DELAY_MS` only does anything on the main thread so by construction it should be removed as well.

tridge:
file ops can happen on main thread while disarmed
see AP_Filesystem_Backend::file_op_allowed()

tridge (approved):

PR #20442, opened 2022-04-04
Fixed lockups on H743 with flash storage

tridge:
This is the result of a long investigation led by @peterbarker and myself into a watchdog reset of a Lumineer H7 in GUIDED mode.
The issue turned out to be the result of several factors:
 - timing changes in chVTDoTickI() due to flash writes blocking one bank
 - monitor thread stack overflows
 - dshot DMA timeout calculation errors

This reverts the MatekH743 back to using a 16 bit timer, which restores the alarm function. The change to 32 bit timer was made as we thought it would fix the 68ms delays, but we now have a much better understanding of those delays, and can reproduce them at-will without these patches by loading large waypoint files. So we can be quite confident that going back to a 16 bit timer is safe

This PR includes patches from #20427

PR #14063, opened 2020-04-12
Support STM32H7 flash storage

tridge:
This adds support for AP_FlashStorage for STM32H7. This is needed for a new H743 based Matek board
The H7 is tricky as:
 - you must write 32 bytes at a time
 - you can't write to a block that isn't all 0xFF without risking ECC corruption
 - sectors are 128k in size

This PR abstracts AP_FlashStorage to allow for the 3 different types of storage more cleanly

This has been tested on:
- f303-GPS for F3 support
- MatekF405-Wing for F4 support
- MatekH743 for H7 support

PR #28972, opened 2024-12-30
AP_HAL_ChibiOS: Scheduler: clean up expected delay functions

tpwrules:
The feature to expect delays from the main thread never worked (since introduction in #14063 ) and hasn't been used since #20442 . Remove it to gain back some flash and reduce complexity.

Checked that parameter and mission ops still work on Kakute H7 Mini.


peterbarker:
Won't this cause issues when we erase flash pages?  We'll no longer be expecting the long delay that that introduces?

tpwrules:
It never worked in the first place, there's no change in behavior here :)

tridge (approved):

PR #28976, opened 2024-12-31
Plane:mode AUTOLAND enhancements

Hwurzburg:
This PR enhances the new AUTOLAND mode. It adds the following new features:

1. Allows takeoff direction capture, and hence autoland direction and use, in ACRO, STABILIZE, FBWA, TRAINING, and MANUAL modes, in addition to Mode TAKEOFF and AUTO NAV_TAKEOFF, using the same criteria as currently used, the first time after arming.

2. Adds an AUTOLAND_OPTIONS option that will set this direction upon arming using a compass, if in use, instead of ground course, allowing the user to set it on the ground arbitrarily as desired, independent of the actual takeoff direction. The option is in the autoland module for future expansion
3.  the "base" leg is actually a WP_LOITER_RAD loiter to alt with tangent exit directly into the final approach waypoint and landing with proper loiter direction.

image

changes to current behavior, besides the new features, is that once the direction is captured it persists until a disarm occurs. Therefore multiple auto takeoffs while armed no longer reset the direction captured by the first, on each subsequent takeoff. Originally, I thought this behavior was desirable, but I now believe it's not of much benefit and the new behavior avoids the current corner case issue of resetting the direction while in flight if a NAV_TAKEOFF autotakeoff is interrupted and then resumed. Also, this is not available for QudPlanes, even if fixed wing landing option  is used. We thought the possibility of getting the direction inappropriately wrong on transitions from a VTOL takeoff just too complex to get correct.

In addition, it removes the AUTOLAND code segments in the commands_logic and mode_takeoff modules, reducing the flash impact overall...

These changes have been tested in SITL and flight-tested in all configurations I could think of.

### DIFF
--- a/ArduPlane/AP_Arming.cpp
+++ b/ArduPlane/AP_Arming.cpp
@@ -321,6 +321,10 @@ bool AP_Arming_Plane::arm(const AP_Arming::Method method, const bool do_arming_c
     // rising edge of delay_arming oneshot
     delay_arming = true;
 
+#if MODE_AUTOLAND_ENABLED
+    plane.mode_autoland.arm_check();
+#endif
+
     send_arm_disarm_statustext("Throttle armed");
 
     return true;
--- a/ArduPlane/Plane.cpp
+++ b/ArduPlane/Plane.cpp
@@ -514,6 +514,10 @@ void Plane::update_control_mode(void)
     update_fly_forward();
 
     control_mode->update();
+
+#if MODE_AUTOLAND_ENABLED
+    mode_autoland.check_takeoff_direction();
+#endif
 }
 
 
--- a/ArduPlane/mode_autoland.cpp
+++ b/ArduPlane/mode_autoland.cpp
@@ -28,12 +28,19 @@ const AP_Param::GroupInfo ModeAutoLand::var_info[] = {
 
     // @Param: DIR_OFF
     // @DisplayName: Landing direction offset from takeoff
-    // @Description: The captured takeoff direction (at arming,if TKOFF_OPTION bit1 is set, or after ground course is established in autotakeoffs)is offset by this amount to create a different landing direction and approach.
+    // @Description: The captured takeoff direction after ground course is established in autotakeoffsis offset by this amount to create a different landing direction and approach.However,if TKOFF_OPTION bit1 is set, the takeoff(landing) direction is captured immediately via compass heading upon arming, then this offset is NOT applied.
     // @Range: -360 360
     // @Increment: 1
     // @Units: deg
     // @User: Standard
     AP_GROUPINFO("DIR_OFF", 3, ModeAutoLand, landing_dir_off, 0),
+
+    // @Param: OPTIONS
+    // @DisplayName: Autoland mode options
+    // @Description: Enables optional autoland mode behaviors
+    // @Bitmask: 1: When set if there is a healthy compass in use the compass heading will be captured at arming and used for the AUTOLAND mode's initial takeoff direction instead of capturing ground course in NAV_TAKEOFF or Mode TAKEOFF or other modes.
+    // @User: Standard
+    AP_GROUPINFO("OPTIONS", 4, ModeAutoLand, options, 0),
 
     AP_GROUPEND
 };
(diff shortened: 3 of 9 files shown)


ArduPlane/AP_Arming.cpp
```diff
@@ -321,6 +321,15 @@ bool AP_Arming_Plane::arm(const AP_Arming::Method method, const bool do_arming_c
     // rising edge of delay_arming oneshot
     delay_arming = true;
 
+#if MODE_AUTOLAND_ENABLED
+    //capture heading for autoland mode if option is set and using a compass
+    if (plane.ahrs.use_compass() && plane.tkoff_option_is_set(AP_FixedWing::TakeoffOption::AUTOLAND_DIR_ON_ARM)) {
+        plane.takeoff_state.initial_direction.heading = wrap_360(plane.ahrs.yaw_sensor * 0.01f);
+        plane.takeoff_state.initial_direction.initialized = true;
+        gcs().send_text(MAV_SEVERITY_INFO, "Autoland direction= %u",int(plane.takeoff_state.initial_direction.heading));
```
peterbarker, line 329:
```suggestion
        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Autoland direction= %u",int(plane.takeoff_state.initial_direction.heading));
```

ArduPlane/Plane.cpp
```diff
@@ -1058,7 +1062,30 @@ void Plane::update_quicktune(void)
...
+#if MODE_AUTOLAND_ENABLED
+/*
+  In certain pilot controlled modes other than NAV_TAKEOFF or Mode TAKEOFF,takeoff direction is initialized after arm when sufficient altitude and ground speed is obtained, then captured takeoff direction + offset used as landing direction in AUTOLAND
+*/
+void Plane::check_takeoff_direction()
+{
+    if (takeoff_direction_check_completed || takeoff_state.initial_direction.initialized) {
+        return;
+    }
+    //set autoland direction if for GPS heading
```
peterbarker, line 1074:
```suggestion
    // consider setting autoland direction from GPS CoG
```

ArduPlane/mode_autoland.cpp
```diff
@@ -63,7 +63,6 @@ bool ModeAutoLand::_enter()
         gcs().send_text(MAV_SEVERITY_WARNING, "Takeoff initial direction not set,must use autotakeoff");
         return false;
     }
-
```
peterbarker, line 66: Random

libraries/AP_Vehicle/AP_FixedWing.h
```diff
@@ -56,5 +56,6 @@ struct AP_FixedWing {
     // Bitfields of TKOFF_OPTIONS
     enum class TakeoffOption {
         THROTTLE_RANGE          = (1U << 0), // Unset: Max throttle. Set: Throttle range.
+        AUTOLAND_DIR_ON_ARM     = (1U << 1), // set dir for autoland on arm if compass in use.
```
peterbarker, line 59: If this is *really* only for autoland then it should be a parameter in the autoland mode.

Capturing the takeoff direction could be useful elsewhere, however, so I'm not that fussed.

peterbarker (dismissed):

ArduPlane/AP_Arming.cpp
Hwurzburg: done

ArduPlane/mode_autoland.cpp
Hwurzburg: done

ArduPlane/Plane.cpp
Hwurzburg: done

ArduPlane/Plane.cpp
```diff
@@ -1058,7 +1062,30 @@ void Plane::update_quicktune(void)
     quicktune.update(control_mode->supports_quicktune());
 }
 #endif
+#if MODE_AUTOLAND_ENABLED
+/*
+  In certain pilot controlled modes other than NAV_TAKEOFF or Mode TAKEOFF,takeoff direction is initialized after arm when sufficient altitude and ground speed is obtained, then captured takeoff direction + offset used as landing direction in AUTOLAND
+*/
+void Plane::check_takeoff_direction()
```
IamPete1, line 1069: These function could live in the mode.

ArduPlane/Plane.cpp
```diff
@@ -514,6 +514,10 @@ void Plane::update_control_mode(void)
     update_fly_forward();
 
     control_mode->update();
+
+#if MODE_AUTOLAND_ENABLED
+    check_takeoff_direction();
```
IamPete1, line 519:
```suggestion
    mode_autoland.check_takeoff_direction();
```
move function into mode and then call via the refence.

ArduPlane/AP_Arming.cpp
```diff
@@ -321,6 +321,15 @@ bool AP_Arming_Plane::arm(const AP_Arming::Method method, const bool do_arming_c
     // rising edge of delay_arming oneshot
     delay_arming = true;
 
+#if MODE_AUTOLAND_ENABLED
+    //capture heading for autoland mode if option is set and using a compass
+    if (plane.ahrs.use_compass() && plane.tkoff_option_is_set(AP_FixedWing::TakeoffOption::AUTOLAND_DIR_ON_ARM)) {
+        plane.takeoff_state.initial_direction.heading = wrap_360(plane.ahrs.yaw_sensor * 0.01f);
```
IamPete1, line 327: I would have expected `landing_dir_off` to be added here as with the ground course method.

ArduPlane/Plane.cpp
```diff
@@ -1058,7 +1062,30 @@ void Plane::update_quicktune(void)
     quicktune.update(control_mode->supports_quicktune());
 }
 #endif
+#if MODE_AUTOLAND_ENABLED
+/*
+  In certain pilot controlled modes other than NAV_TAKEOFF or Mode TAKEOFF,takeoff direction is initialized after arm when sufficient altitude and ground speed is obtained, then captured takeoff direction + offset used as landing direction in AUTOLAND
+*/
+void Plane::check_takeoff_direction()
+{
+    if (takeoff_direction_check_completed || takeoff_state.initial_direction.initialized) {
```
IamPete1, line 1071: Do we need both `takeoff_direction_check_completed` and `takeoff_state.initial_direction.initialized`?

ArduPlane/AP_Arming.cpp
Hwurzburg: offset should not be added to the armed set....the user knows the exact direction he wants

ArduPlane/Plane.cpp
Hwurzburg: no...fixed

ArduPlane/Plane.cpp
Hwurzburg: seems cleaner to me here

Hwurzburg: fixed

ArduPlane/Plane.cpp
Hwurzburg: done

ArduPlane/mode_autoland.cpp
```diff
@@ -129,30 +139,97 @@ void ModeAutoLand::update()
...
-        plane.auto_state.next_wp_crosstrack = true;
+    if (plane.takeoff_state.initial_direction.initialized) {
+        return;
+    }
+    //set autoland direction to GPS course over ground
+    if (plane.control_mode->allows_autoland_direction_capture() &&
+        plane.is_flying() &&
+        hal.util->get_soft_armed() &&
+        plane.gps.ground_speed() > GPS_GND_CRS_MIN_SPD) {
+        set_autoland_direction();
```
IamPete1, line 202:
```suggestion
        set_autoland_direction(plane.gps.ground_course() + landing_dir_off);
```
If you passing in the heading the `set_autoland_direction` can be used for both this and the compass case saving some flash.

ArduPlane/mode_autoland.cpp
```diff
@@ -129,30 +139,97 @@ void ModeAutoLand::update()
...
     }
 }
-#endif
+
+// Sets autoland direction using ground course + offest parameter
+void ModeAutoLand::set_autoland_direction()
```
IamPete1, line 207:
```suggestion
void ModeAutoLand::set_autoland_direction(const float heading)
```

ArduPlane/mode_autoland.cpp
```diff
@@ -129,30 +139,97 @@ void ModeAutoLand::update()
...
+{
+    plane.takeoff_state.initial_direction.heading = wrap_360(plane.gps.ground_course() + landing_dir_off);
```
IamPete1, line 209:
```suggestion
    plane.takeoff_state.initial_direction.heading = wrap_360(heading);
```

ArduPlane/mode_autoland.cpp
```diff
@@ -129,30 +139,97 @@ void ModeAutoLand::update()
...
+    return plane.mode_loiter.isHeadingLinedUp_cd(cmd[0].content.location.get_bearing(cmd[1].content.location)*100);
+}
+
+// possibly capture heading on arming for autoland mode if option is set and using a compass
+void ModeAutoLand::arm_check(void)
+{
+    if (plane.ahrs.use_compass() && autoland_option_is_set(ModeAutoLand::AutoLandOption::AUTOLAND_DIR_ON_ARM)) {
+        plane.takeoff_state.initial_direction.heading = wrap_360(plane.ahrs.yaw_sensor * 0.01f);
+        plane.takeoff_state.initial_direction.initialized = true;
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Autoland direction= %u",int(plane.takeoff_state.initial_direction.heading));
```
IamPete1, line 230:
```suggestion
        set_autoland_direction(plane.ahrs.yaw_sensor * 0.01);
```

ArduPlane/mode_autoland.cpp
```diff
@@ -129,30 +139,97 @@ void ModeAutoLand::update()
...
+}
+
+/*
+  return true when the LOITER_TO_ALT is lined up ready for the landing, used in commands_logic verify loiter to alt
+ */
+bool ModeAutoLand::landing_lined_up(void)
+{
+    // use the line between the center of the LOITER_TO_ALT on the base leg and the
+    // start of the landing leg (land_start_WP)
+    return plane.mode_loiter.isHeadingLinedUp_cd(cmd[0].content.location.get_bearing(cmd[1].content.location)*100);
```
IamPete1, line 221: Why not use `isHeadingLinedUp(cmd[0].content.location, cmd[1].content.location)`? It has lots of extra handling that should make it more reliable.

ArduPlane/mode_autoland.cpp
```diff
@@ -88,38 +99,37 @@ bool ModeAutoLand::_enter()
...
      */
     const float bearing_to_current_deg = wrap_180(degrees(land_start_WP.get_bearing(plane.current_loc)));
     const float bearing_err_deg = wrap_180(wrap_180(plane.takeoff_state.initial_direction.heading) - bearing_to_current_deg);
     const float bearing_offset_deg = bearing_err_deg > 0? -90 : 90;
-    const float base_leg_length = final_wp_dist * 0.333;
-    cmd[0].id = MAV_CMD_NAV_WAYPOINT;
+
+    // make the "base leg" the smaller of the waypoint loiter radius or 1/3 of the final WP distance
+    // this results in a neat landing no matter the loiter radius
+    const float base_leg_length = MIN(final_wp_dist * 0.333, fabsf(plane.aparm.loiter_radius));
```
IamPete1, line 112: This hits the loiter radius varying with altitude issue. Loiters get bigger with altitude, what happens if the radius is larger than `final_wp_dist * 0.333`. 

You could use `plane.nav_controller->loiter_radius(fabsf(plane.loiter.radius))` to get the corrected radius, but that will be fore the altitude at which home is entered not the altitude at which the waypoint is.

ArduPlane/mode_autoland.cpp
IamPete1: It should also better deal with the case that the loiter radius is larger than the base leg.

Hwurzburg: I had that and Tridge did not like the way it behaved

ArduPlane/mode_autoland.cpp
Hwurzburg: this assures that the base leg and loiter radius never gets bigger than 1/3 the final no matter what the WP_LOITER_RAD is...Tridge added this

ArduPlane/mode_autoland.cpp
Hwurzburg: done

ArduPlane/mode_autoland.cpp
Hwurzburg: done

ArduPlane/mode_autoland.cpp
Hwurzburg: done

ArduPlane/mode_autoland.cpp
Hwurzburg: done

ArduPlane/mode_autoland.cpp
IamPete1: That is not the case because of the altitude of loiter radius is scaled in the L1 controller. For example do a normal RTL in SITL, you will find the distance to home is 30% larger than the RTL radius. With defaults `RTL_RADIUS` is zero and `WP_LOITER_RAD` is 80. So we would expect the loiter radius to be 80m However, if you try it you will find that it is ~96m. The extra 16m are due to the altitude, at sea level it would loiter at the expected altitude. 

In this case if your base leg is short enough the second waypoint will be inside the loiterer circle which will cause the breakout logic to do funny things. (that is one of the things the other heading lined up function handled better)

Hwurzburg: up to Tridge...his code change

IamPete1:
@Hwurzburg @tridge I have done a PR to address the issues with loiter radius scaling, it does get quite bad at higher altitudes.

https://github.com/Hwurzburg/ardupilot/pull/5 

In that messing about I have found two other issues. The first is pre-existing and that is what happens if you have `LAND_TYPE` set to deep stall rather than glide slope. I guess no one has tested? Maybe it is not too bad, but we should either test it or refuse mode entry if the not glide slope i'm not sure the base entry setup thing makes any sense for it in any case. The other issue is that this re-uses some functions from missions, for example this uses `plane.start_command` to run the waypoints. This is a issue because of this:

https://github.com/ArduPilot/ardupilot/blob/230f8248c953a6cdbd8b7035948de300623d2fea/ArduPlane/commands_logic.cpp#L29-L37

Its looking in the mission at what the next command is, of course here were not in the mission so that will be nonsense. Maybe it would be enough to warp a check that were in auto for that bit. In the longer term it would be nice to rework more comprehensively so both mission and this could use a common base method.

Hwurzburg:
Added all of Pete Hall's changes to cope with high altitude

tridge (approved):

PR #28979, opened 2025-01-01
AP_DroneCAN: correct narrowing-conversion errors

peterbarker:


tpwrules (approved):
Compiles properly now, conversions won't overflow and match the message descriptions.

There is an error value when `msg.motor_temperature == 255` but it doesn't overflow and hopefully the user will notice so whatever.

PR #28980, opened 2025-01-01
hwdef: Add GPS moving baseline to MatekH743-periph

rishabsingh3003:
I couldn't get moving baseline working without this on the periph


peterbarker (approved):
LGTM.  We do this on a lot of other peripherals, and this one can certainly afford the flash.


We probably should have separate defines for the simple agrica-style heading-from-GPS and the multi-backend ublox stuff.  Given the restriction here on the number of receivers supported this define pulls in a lot of unnecessary code.

Issue #28969, opened 2024-12-30
Floating Point Exception in the position controller


peterbarker:
CI run produced an FPE while running `test.Plane.FlyEachFrame`.  I've never seen this before.

The PR in question is going absolutely nowhere near the relevant code (should actually be a no-compiler-output change in SITL?).  https://github.com/ArduPilot/ardupilot/pull/28966

Link to output is below, also excerpted below.

Seems to be something wrong with the maths in `limit_accel_xy` - we're taking the square root of a negative number; something wrong with taking the cross product, perhaps?!  Thankfully the output below contains values for all of the variables required...

https://productionresultssa10.blob.core.windows.net/actions-results/4e63b159-021f-4b97-a167-86fc7dcf4f13/workflow-job-run-6b0240de-2d48-582d-e68b-9616ef328464/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-12-30T07%3A30%3A15Z&sig=CO5BGQOMT4%2FEO2n%2F2x77s%2FBvK54kxyiUgE7qSlhKxUs%3D&ske=2024-12-30T19%3A10%3A12Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-12-30T07%3A10%3A12Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-11-04&sp=r&spr=https&sr=b&st=2024-12-30T07%3A20%3A10Z&sv=2024-11-04

```
2024-12-30T06:13:01.8519750Z AT-0413.1: AP: VTOL position1 v=21.3 d=169.1 h=49.8 dc=19.6
2024-12-30T06:13:01.8520094Z arduplane: ERROR: Floating point exception - aborting
2024-12-30T06:13:01.8520719Z arduplane: Running: sh Tools/scripts/dumpstack.sh 8040 >dumpstack.sh_arduplane.8040.out 2>&1
2024-12-30T06:13:01.8521189Z arduplane: dumpstack.sh has been run.  Output was:
2024-12-30T06:13:01.8521547Z arduplane: -------------- begin dumpstack.sh output ----------------
2024-12-30T06:13:01.8521872Z arduplane: [New LWP 8041]
2024-12-30T06:13:01.8522144Z arduplane: [Thread debugging using libthread_db enabled]
[...]

PR #28981, opened 2025-01-01
AP_Math: prevent FPE in SITL when limitting accel, vectors

peterbarker:
Co-authored-by: Leonard Hall <leonardthall@gmail.com>

the  cross-product code can produce something slightly negative.  Stop passing that into safe_sqrt; it isn't *that* safe on SITL!

### DIFF
--- a/libraries/AP_Math/AP_Math.cpp
+++ b/libraries/AP_Math/AP_Math.cpp
@@ -67,14 +67,17 @@ template float safe_asin<short>(const short v);
 template float safe_asin<float>(const float v);
 template float safe_asin<double>(const double v);
 
+// sqrt which takes any type and returns 0 if the input is NaN or less than zero
 template <typename T>
 float safe_sqrt(const T v)
 {
-    float ret = sqrtf(static_cast<float>(v));
-    if (isnan(ret)) {
-        return 0;
+    // cast before checking so we sqrtf the same value we check
+    const float val = static_cast<float>(v);
+    // use IEEE-754 compliant function which returns false if val is NaN
+    if (isgreaterequal(val, 0)) {
+        return sqrtf(val);
     }
-    return ret;
+    return 0;
 }
 
 template float safe_sqrt<int>(const int v);
--- a/libraries/AP_Math/control.cpp
+++ b/libraries/AP_Math/control.cpp
@@ -387,6 +387,8 @@ bool limit_accel_xy(const Vector2f& vel, Vector2f& accel, float accel_max)
             if (accel_cross.limit_length(accel_max)) {
                 accel_dir = 0.0;
             } else {
+                // limit_length can't absolutely guarantee this subtraction
+                // won't be slightly negative, so safe_sqrt is used
                 float accel_max_dir = safe_sqrt(sq(accel_max) - accel_cross.length_squared());
                 accel_dir = constrain_float(accel_dir, -accel_max_dir, accel_max_dir);
             }
--- a/libraries/AP_Math/tests/test_control.cpp
+++ b/libraries/AP_Math/tests/test_control.cpp
@@ -5,6 +5,9 @@
 #include <AP_Math/vector3.h>
 #include <AP_Math/control.h>
 
+#include <fenv.h>
+#include <setjmp.h>
+
 TEST(Control, test_control)
 {
     postype_t pos_start = 17;
@@ -465,6 +468,58 @@ TEST(Control, test_control)
     EXPECT_FLOAT_EQ(velxy.y, 0.0);
 }
 
+// catch floating point exceptions
+sigjmp_buf avert_your_eyes_children;
+static void _tc_sig_fpe(int signum)
+{
+    siglongjmp(avert_your_eyes_children, 1);
+}
+
+TEST(Control, test_limit_accel)
+{
+    // reproduction of FPE (https://github.com/ArduPilot/ardupilot/issues/28969)
+    // FPE will only be raised in SITL HAL, so compiling for linux HAL
+    // isn't useful.
+    const Vector2f vel{
+        0.984285712, 0.176583186
+    };
+    Vector2f accel{99.9008408, -557.304077};
+    const float accel_max = 566.187256;
+
+    struct sigaction old_sa_fpe = {};
+
+    struct sigaction sa_fpe = {};
+    sigemptyset(&sa_fpe.sa_mask);
+    sa_fpe.sa_handler = _tc_sig_fpe;
+    if (sigaction(SIGFPE, &sa_fpe, &old_sa_fpe) == -1) {
+        abort();
+    }
+    const int excepts = FE_UNDERFLOW | FE_OVERFLOW | FE_INVALID;
+    fexcept_t old_except_flags;
+    if (fegetexceptflag(&old_except_flags, excepts) == -1) {
+        abort();
+    }
+
+    feenableexcept(excepts);
+
+    bool signal_caught = false;
+    if (sigsetjmp(avert_your_eyes_children, 1)) {
+        // we come through here if an FPE is triggered (via a goto in
+        // our custom signal handler, _tc_sig_fpe)
+        signal_caught = true;
+    } else {
+        // we come through here normally
+        EXPECT_FALSE(limit_accel_xy(vel, accel, accel_max));
+    }
+
+    EXPECT_TRUE(signal_caught);
+
+    // now restore the original fpe handling
+    if (fesetexceptflag(&old_except_flags, excepts) == -1) {
+        abort();
+    }
+    sigaction(SIGFPE, &old_sa_fpe, nullptr);
+}
 
 AP_GTEST_MAIN()
 int hal = 0;
--- a/libraries/AP_Math/tests/test_control.cpp
+++ b/libraries/AP_Math/tests/test_control.cpp
@@ -509,10 +509,10 @@ TEST(Control, test_limit_accel)
         signal_caught = true;
     } else {
         // we come through here normally
-        EXPECT_FALSE(limit_accel_xy(vel, accel, accel_max));
+        EXPECT_TRUE(limit_accel_xy(vel, accel, accel_max));
     }
 
-    EXPECT_TRUE(signal_caught);
+    EXPECT_FALSE(signal_caught);
 
     // now restore the original fpe handling
     if (fesetexceptflag(&old_except_flags, excepts) == -1) {


khancyr:
Shouldn't be this done in safe_sqrt ? I would assume that we have some other place that can have this issue

peterbarker:
> Shouldn't be this done in safe_sqrt ? I would assume that we have some other place that can have this issue

Well... we really want to know about where this actually does happen.  This PR isn't really a satisfactory solution to what we're seeing IMO - it would be nice to trace back the maths to work out how the rhs ens up larger than the lhs in here.  https://github.com/ArduPilot/ardupilot/issues/28969 is the issue if you missed it

IamPete1:
I made a Geogebra to visualise the maths.  

https://www.geogebra.org/calculator/ktxy2uyf

I don't see any issue with the maths. Must be floating point stuff.

If I were to guess it would be `vel.is_zero()`. It checks the individual X and Y components are not zero. This does not mean that the length is none zero (when using floats). Squaring a very small number which is not zero might result in a number which is zero. So you get zeros before the addition and square root that would make the number larger again.

Maybe we need a new method `length_is_zero` we could compare components against `sqrt(FLT_EPSILON)` rather than `FLT_EPSILON`.

tpwrules:
Do we have the actual numbers that were given to `limit_accel_xy`? Are they unreasonably small?

tpwrules:
Pete is close, the problem is imprecision in `Vector2f.limit_length()` on line 387.

The vector is `(99.9791489, 557.290039)` and the length limit is `566.187256`. `length()` of the vector is `566.187256` so `limit_length()` doesn't do anything and the sqrtf runs. But `length_squared()` is `320568.031250` while `accel_max*accel_max` is `320568.000000`. The difference in these quantities is exactly the value that was passed to `safe_sqrt`. Lucky shot!

* We could implement a MAX with 0, that seems okay
* We could adjust the safe_sqrt call to not square accel_max, but that would require an extra square root
* We could adjust `limit_length` to check the squared length, or have an alternate method for where we care about the squared limit

I'm not sure how to guarantee the vector length quantity is actually <= the limit after the math in `limit_length()` no matter how it's done. I am guessing that could have errors too if they are very close.

tpwrules:
Doing some research and fiddling suggests this is the correct solution. We need to audit other uses of `limit_length()` to make sure they won't explode if the vector ends up slightly over the limit.

It's not feasible due to the vagaries of floating point math to guarantee it's always equal or under.

libraries/AP_Math/control.cpp
```diff
@@ -387,7 +387,7 @@ bool limit_accel_xy(const Vector2f& vel, Vector2f& accel, float accel_max)
             if (accel_cross.limit_length(accel_max)) {
                 accel_dir = 0.0;
             } else {
-                float accel_max_dir = safe_sqrt(sq(accel_max) - accel_cross.length_squared());
+                float accel_max_dir = safe_sqrt(MAX(0, sq(accel_max) - accel_cross.length_squared()));
```
tpwrules, line 390:
```suggestion
                // limit_length can't absolutely guarantee this subtraction won't be slightly negative
                float accel_max_dir = safe_sqrt(MAX(0, sq(accel_max) - accel_cross.length_squared()));
```

tpwrules: Actually, it would probably be more efficient to just calculate the two terms of the subtraction and compare them.

IamPete1:
> * We could adjust `limit_length` to check the squared length, or have an alternate method for where we care about the squared limit

We could limit to the smaller of the the existing and the squared so we always conservative. 

On the whole I tend to think know we know what is going on we probably don't need to panic. I'd even be tempted not to fix this case so we can see if a recent change had made this more likely.

We do have a test for the function, maybe we should add some cases there. 
https://github.com/ArduPilot/ardupilot/blob/5726c03e385a7290377c7a4c19b8c9ec289ac2a1/libraries/AP_Math/tests/test_vector2.cpp#L125-L131

peterbarker:
Closes https://github.com/ArduPilot/ardupilot/issues/28969

tpwrules:
I don't actually think `limit_length()` can be entirely fixed unfortunately. I think it's better to just be mindful of its limitations.

And it has happened again here: https://github.com/ArduPilot/ardupilot/actions/runs/12572260649/job/35044375709?pr=28857

peterbarker:
I've written a test for this now so other people can play with things in gdb easily.

I've also pushed up a patch reverting the original fix and proposing an alternative fix.  Basically stop using the apples comparison to determine if we do the oranges maths - use a single fruit in both cases.

peterbarker:
> And it has happened again here: https://github.com/ArduPilot/ardupilot/actions/runs/12572260649/job/35044375709?pr=28857

Well, that's coincidence... third is presumed to be enemy action...

I've had a quick look through the stack and code changes and haven't found any smoking gun.  @robertlong13 changed things up in this test a bit recently be adding in the ICE frames.  Is it possible that the slightly different mission is causing this?  https://github.com/ArduPilot/ardupilot/commit/3982d576eb5d714ac62ac91386f1ec9450d54800

tpwrules:
So this both fixes the problem and doesn't. The problem is fixed because we won't execute a square root with a negative number; we now trust that the limit function says it wanted to limit the vector and then set `accel_dir` to 0 instead of calling `safe_sqrt`.

But it doesn't fix the problem because the vector is still in some sense too long. Calling `accel_cross.limit_length_squared(accel_max)` again will still return true. The test `(len_squared <= max_length_squared)` is still false, but `len == max_length` so `x` and `y` get multiplied by precisely one and the vector doesn't change at all despite the function returning that it tried to change it.

I think adding a `MAX(0, <old value>)` to the `safe_sqrt` now is the best solution to stop CI flapping, then do some more detailed studies of who uses `limit_length` and what are the problems for going over.

> Is it possible that the slightly different mission is causing this?

Yes, I expect so. It just happens to sometimes hit this exact point now. Here's the enemy action: https://github.com/ArduPilot/ardupilot/actions/runs/12581405966/job/35065761383?pr=28991 . The numbers are again identical.

tpwrules:
Another failure: https://github.com/ArduPilot/ardupilot/actions/runs/12591859248/job/35097592628?pr=28996 

Going to try to replicate locally and see if that mission change was the instigator.

Still prefer the `MAX` fix.

tpwrules:
I was able to replicate it (running only the FlyEachFrame test and quadplane-tilt frame) but sometimes it took 50-60 tries and sometimes it was one or two. It would be hard to say what particular commit did it. I imagine it's kind of dependent on system load and stuff too.

tpwrules:
Now it's infected copter: https://github.com/ArduPilot/ardupilot/actions/runs/12616584539/job/35158079748?pr=28857

`accel_cross = {x = 290.641205, y = -485.89682}`, `accel_max = 566.187256`

peterbarker:
> Now it's infected copter: https://github.com/ArduPilot/ardupilot/actions/runs/12616584539/job/35158079748?pr=28857

We're probably going to need fire by the end of this.

peterbarker:
I've popped off my alternative fix now.

tpwrules (approved):
Fix looks good to me, thanks! Tested that the test fails when the fix is reverted.

Some super minor nits:
* never quite sure what "limiting accel, vectors" meant, that could be clearer
* squash last two commits into one as they are both needed together to fix the test
* `Co-authored-by:` tag needs to be the last line in the commit message, after two blank lines for GitHub to pick it up

peterbarker:
> Fix looks good to me, thanks! Tested that the test fails when the fix is reverted.
> 
> Some super minor nits:
> 
>     * never quite sure what "limiting accel, vectors" meant, that could be clearer
> 
>     * squash last two commits into one as they are both needed together to fix the test
> 
>     * `Co-authored-by:` tag needs to be the last line in the commit message, after two blank lines for GitHub to pick it up

All fixed, thanks!

libraries/AP_Math/control.cpp
peterbarker: Done

lthall (approved):

tridge:
I'd prefer to restore the original intention of safe_sqrt() which is what a negative input is not an error

IamPete1:
Looks like this is the same problem and used the same fix as proposed here:

https://github.com/ArduPilot/ardupilot/blob/01345e5a38c942f6fce4d92d3f1fcfecdc190833/libraries/AC_AttitudeControl/AC_PosControl.cpp#L1415-L1416

tpwrules:
A little clever but should serve the purpose :)

libraries/AP_Math/AP_Math.cpp
```diff
@@ -67,14 +67,18 @@ template float safe_asin<short>(const short v);
...
 float safe_sqrt(const T v)
 {
-    float ret = sqrtf(static_cast<float>(v));
-    if (isnan(ret)) {
+    // cast before checking so we sqrtf the same value we check
+    float val = static_cast<float>(v);
+    // use IEEE-754 compliant function which returns false if val is NaN
+    if (isgreaterequal(val, 0)) {
+        return sqrtf(val);
+    } else {
```
peterbarker, line 79:
```suggestion
    }
```

libraries/AP_Math/AP_Math.cpp
```diff
@@ -67,14 +67,17 @@ template float safe_asin<short>(const short v);
...
 
+// sqrt which takes any type and returns 0 if the input is NaN or less than zero
 template <typename T>
 float safe_sqrt(const T v)
 {
-    float ret = sqrtf(static_cast<float>(v));
-    if (isnan(ret)) {
-        return 0;
+    // cast before checking so we sqrtf the same value we check
+    float val = static_cast<float>(v);
```
peterbarker:
```suggestion
    const float val = static_cast<float>(v);
```

libraries/AP_Math/AP_Math.cpp
```diff
@@ -67,14 +67,17 @@ template float safe_asin<short>(const short v);
...
 
+// sqrt which takes any type and returns 0 if the input is NaN or less than zero
 template <typename T>
 float safe_sqrt(const T v)
 {
-    float ret = sqrtf(static_cast<float>(v));
-    if (isnan(ret)) {
-        return 0;
+    // cast before checking so we sqrtf the same value we check
+    const float val = static_cast<float>(v);
```
peterbarker, line 75: Wow, we went through this entire discussion without noting that we're squashing any double which might be passed in....

tpwrules: I had noticed that. But we do it in several places, I'm not too concerned specifically here.

rmackay9:
@peterbarker @tpwrules @lthall,

Do we want this backported to 4.6?

tpwrules:
> @peterbarker @tpwrules @lthall,
> 
> Do we want this backported to 4.6?

I say it would be nice but not necessary. The change is safe but we haven't seen that CI breakage in the 4.6 branch AFAIK.

rmackay9:
OK great, thanks.  I've removed this PR from the 4.6 backport project for now at least

PR #28982, opened 2025-01-01
Baro thrust scaling

andyp1per:
This allows a user to apply a linear pressure compensation to a baro based on thrust output. On a small whoop there is a very causal relationship between pressure and throttle:

image

By applying this offset the effect can be normalized:

image

Set by using ```BAROx_THST_SCALE"

### DIFF
--- a/libraries/AP_Baro/AP_Baro.cpp
+++ b/libraries/AP_Baro/AP_Baro.cpp
@@ -56,6 +56,9 @@
 #include <AP_Logger/AP_Logger.h>
 #include <AP_GPS/AP_GPS.h>
 #include <AP_Vehicle/AP_Vehicle.h>
+#if AP_BARO_THST_COMP_ENABLED
+#include <AP_Motors/AP_Motors.h>
+#endif
 
 #define INTERNAL_TEMPERATURE_CLAMP 35.0f
 
@@ -253,7 +256,15 @@ const AP_Param::GroupInfo AP_Baro::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("_OPTIONS", 24, AP_Baro, _options, 0),
 #endif
-    
+
+#if AP_BARO_THST_COMP_ENABLED
+    // @Param: 1_THST_SCALE
+    // @DisplayName: Thrust compensation
+    // @Description: Thrust scaling in Pascals. This value scaled by the normalized thrust is subtracted from the barometer pressure. This is used to adjust linearly based on the thrust output for local pressure difference induced by the props.
+    // @Range: -300 300
+    // @User: Advanced
+    AP_GROUPINFO("1_THST_SCALE", 25, AP_Baro, sensors[0].mot_scale, 0),
+#endif  // AP_BARO_THST_COMP_ENABLED
     AP_GROUPEND
 };
 
@@ -872,6 +883,12 @@ void AP_Baro::update(void)
 #if HAL_BARO_WIND_COMP_ENABLED
                 corrected_pressure -= wind_pressure_correction(i);
 #endif
+#if AP_BARO_THST_COMP_ENABLED
+                corrected_pressure -= thrust_pressure_correction(i);
+#endif
+#if (HAL_BARO_WIND_COMP_ENABLED || AP_BARO_THST_COMP_ENABLED)
+                sensors[i].corrected_pressure = corrected_pressure;
+#endif
                 altitude = get_altitude_difference(sensors[i].ground_pressure, corrected_pressure);
 
                 // the ground pressure is references against the field elevation
@@ -987,6 +1004,22 @@ void AP_Baro::update_field_elevation(void)
 #endif
 }
 
+#if AP_BARO_THST_COMP_ENABLED
+// scale the baro linearly with thrust
+float AP_Baro::thrust_pressure_correction(uint8_t instance)
+{
+#if APM_BUILD_TYPE(APM_BUILD_ArduPlane) || APM_BUILD_COPTER_OR_HELI
+    const AP_Motors* motors = AP::motors();
+    if (motors == nullptr) {
+         return 0.0f;
+    }
+    const float motors_throttle = MAX(0,motors->get_throttle_out());
+    return sensors[instance].mot_scale * motors_throttle;
+#else
+    return 0.0f;
+#endif
+}
+#endif
 
 /* register a new sensor, claiming a sensor slot. If we are out of
    slots it will panic
--- a/libraries/AP_Baro/AP_Baro_config.h
+++ b/libraries/AP_Baro/AP_Baro_config.h
@@ -103,3 +103,7 @@
 // this allows for using the simple model with the --ekf-single configure option
 #define AP_BARO_1976_STANDARD_ATMOSPHERE_ENABLED HAL_WITH_EKF_DOUBLE || AP_SIM_ENABLED
 #endif
+
+#ifndef AP_BARO_THST_COMP_ENABLED
+#define AP_BARO_THST_COMP_ENABLED 0
+#endif
--- a/libraries/AP_Baro/LogStructure.h
+++ b/libraries/AP_Baro/LogStructure.h
@@ -18,7 +18,8 @@
 // @Field: SMS: time last sample was taken
 // @Field: Offset: raw adjustment of barometer altitude, zeroed on calibration, possibly set by GCS
 // @Field: GndTemp: temperature on ground, specified by parameter or measured while on ground
-// @Field: Health: true if barometer is considered healthy
+// @Field: H: true if barometer is considered healthy
+// @Field: CPress: compensated atmospheric pressure
 struct PACKED log_BARO {
     LOG_PACKET_HEADER;
     uint64_t time_us;
@@ -32,6 +33,7 @@ struct PACKED log_BARO {
     float   drift_offset;
     float   ground_temp;
     uint8_t healthy;
+    float   corrected_pressure;
 };
 
 // @LoggerMessage: BARD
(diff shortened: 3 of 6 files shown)


libraries/AP_Baro/AP_Baro.cpp
```diff
@@ -56,6 +56,9 @@
 #include <AP_Logger/AP_Logger.h>
 #include <AP_GPS/AP_GPS.h>
 #include <AP_Vehicle/AP_Vehicle.h>
+#if HAL_BARO_THST_COMP_ENABLED
```
peterbarker, line 59:
```suggestion
#if AP_BARO_THST_COMP_ENABLED
```

libraries/AP_Baro/AP_Baro_config.h
```diff
@@ -103,3 +103,7 @@
 // this allows for using the simple model with the --ekf-single configure option
 #define AP_BARO_1976_STANDARD_ATMOSPHERE_ENABLED HAL_WITH_EKF_DOUBLE || AP_SIM_ENABLED
 #endif
+
+#ifndef HAL_BARO_THST_COMP_ENABLED
+#define HAL_BARO_THST_COMP_ENABLED 1
```
peterbarker, line 108: Need to know how useful this might be on other vehicles.  And flash cost...

andyp1per: Cost 160 bytes avg on Durandal

libraries/AP_Baro/AP_Baro.cpp
andyp1per: Fixed thanks

tpwrules:
The parameter is error at max thrust?

andyp1per:
> The parameter is error at max thrust?

Not sure I follow?

tpwrules:
> Not sure I follow?

The parameter description says "User provided thrust scaling in Pascals. This is used to adjust linearly based on the thrust output for local pressure difference induced by the props.".

If it's a linear scaling, it should be Pascals per something. The code uses the scale as `sensors[instance].mot_scale * motors_throttle;`, so it's the Pascal scaling to apply when `motors_throttle` is 1. When is that? Can you clarify the parameter description?

libraries/AP_Baro/AP_Baro.cpp
```diff
@@ -225,6 +228,32 @@ const AP_Param::GroupInfo AP_Baro::var_info[] = {
 #endif
 #endif  // HAL_BARO_WIND_COMP_ENABLED
 
+#if AP_BARO_THST_COMP_ENABLED
+    // @Param: 1_THST_SCALE
+    // @DisplayName: Thrust compensation
+    // @Description: User provided thrust scaling in Pascals. This is used to adjust linearly based on the thrust output for local pressure difference induced by the props.
+    // @Range: -300 300
+    // @User: Advanced
+    AP_GROUPINFO("1_THST_SCALE", 25, AP_Baro, sensors[0].mot_scale, 0),
```
IamPete1, line 237: I find it confusing that index 25 is here. Why not at the end? If you want to keep them here please add a comment at the end of the table noting that 25,26 and 27 are used here.

libraries/AP_Baro/AP_Baro.cpp
```diff
@@ -871,6 +900,10 @@ void AP_Baro::update(void)
             if (sensors[i].type == BARO_TYPE_AIR) {
 #if HAL_BARO_WIND_COMP_ENABLED
                 corrected_pressure -= wind_pressure_correction(i);
+#endif
+#if AP_BARO_THST_COMP_ENABLED
+                corrected_pressure -= thrust_pressure_correction(i);
+                sensors[i].corrected_pressure = corrected_pressure;
```
IamPete1, line 906: This additional logging should be enabled for wind comp or thrust comp. 
```suggestion
#endif
#if (HAL_BARO_WIND_COMP_ENABLED || AP_BARO_THST_COMP_ENABLED)
                sensors[i].corrected_pressure = corrected_pressure;
```

libraries/AP_Baro/LogStructure.h
```diff
@@ -32,6 +33,7 @@ struct PACKED log_BARO {
     float   drift_offset;
     float   ground_temp;
     uint8_t healthy;
+    float   corrected_pressure;
```
IamPete1, line 36: Should log NaN if the feature is not avalalbe.

IamPete1: Or I guess you could define it out of the structure completely.

libraries/AP_Baro/AP_Baro.cpp
IamPete1: I wonder if it would be more intuitive to log the correction itself rather than the corrected value. Would make it easier to look at of the course of a log. Could calculate the difference with MAVExplorer I guess.

libraries/AP_Baro/AP_Baro.cpp
```diff
@@ -225,6 +228,32 @@ const AP_Param::GroupInfo AP_Baro::var_info[] = {
 #endif
 #endif  // HAL_BARO_WIND_COMP_ENABLED
 
+#if AP_BARO_THST_COMP_ENABLED
+    // @Param: 1_THST_SCALE
+    // @DisplayName: Thrust compensation
+    // @Description: User provided thrust scaling in Pascals. This is used to adjust linearly based on the thrust output for local pressure difference induced by the props.
```
rmackay9, line 234: Txs for this. I guess we could remove the "User provided" at the beginning of the Description.  I think most parameters are probably "user provided"

lthall:
This is also a combination of roll and yaw:
image


It looks like the tune was on the edge and got into a corner of the flight evelope that was properly unstable with some cross compling between roll and yaw.

libraries/AP_Baro/AP_Baro.cpp
```diff
@@ -225,6 +228,32 @@ const AP_Param::GroupInfo AP_Baro::var_info[] = {
...
+    // @Range: -300 300
+    // @User: Advanced
+    AP_GROUPINFO("1_THST_SCALE", 25, AP_Baro, sensors[0].mot_scale, 0),
+
+#if BARO_MAX_INSTANCES > 1
+    // @Param: 2_THST_SCALE
```
tridge, line 240: i think we should only put it on the first baro, I think this issue can only happen on really tiny vehicles, and fixing just one baro should be enough (just to reduce the params). We could keep the variable in the common structure, but it would just be zero

tridge (dismissed):
I still wonder if this could be a voltage supply, ie motor load drawing down the voltage, impacting the sensor. Could you try props off, but friction load on the motors (eg. your fingers) and see if baro is impacted?

libraries/AP_Baro/AP_Baro.cpp
tridge: @IamPete1 also suggested upside down props for the test, see if the effect is changed

rmackay9:
It'd be nice if we could get other devs with whoops to give it a try as well

libraries/AP_Baro/AP_Baro.cpp
```diff
@@ -871,6 +900,10 @@ void AP_Baro::update(void)
             if (sensors[i].type == BARO_TYPE_AIR) {
 #if HAL_BARO_WIND_COMP_ENABLED
                 corrected_pressure -= wind_pressure_correction(i);
+#endif
+#if AP_BARO_THST_COMP_ENABLED
+                corrected_pressure -= thrust_pressure_correction(i);
```
Georacer, line 905: It would be nice to make the effect of the scale parameter explicit, in terms of what direction it does the correction.
I see you subtract the correction here probably to keep form with the lines above, but that doesn't necessarily make it clear to the one who sets the parameter.

libraries/AP_Baro/AP_Baro.cpp
andyp1per: I suspect part of the reason is that the props and motors are already on the underside of the vehicle, pushing away from the FC

libraries/AP_Baro/AP_Baro.cpp
andyp1per: Fixed

libraries/AP_Baro/AP_Baro.cpp
andyp1per: I fixed the #define. Don't know about the logging - I feel its better this way as you can easily see the pressure that is actually being used.

libraries/AP_Baro/LogStructure.h
andyp1per: Fixed

libraries/AP_Baro/AP_Baro.cpp
andyp1per: Fixed

libraries/AP_Baro/AP_Baro.cpp
andyp1per: Added to description of param

libraries/AP_Baro/AP_Baro_config.h
```diff
@@ -103,3 +103,7 @@
 // this allows for using the simple model with the --ekf-single configure option
 #define AP_BARO_1976_STANDARD_ATMOSPHERE_ENABLED HAL_WITH_EKF_DOUBLE || AP_SIM_ENABLED
 #endif
+
+#ifndef AP_BARO_THST_COMP_ENABLED
+#define AP_BARO_THST_COMP_ENABLED 1
```
rmackay9, line 108: I think it would be good if we only enabled this on the boards/vehicles that really need it.  It's just a small part of keeping complexity for the user down

andyp1per:
Props upside down the effect is reversed, so pretty clear this is being caused by the ducting

image

libraries/AP_Baro/AP_Baro_config.h
andyp1per: Ok fixed

libraries/AP_Baro/AP_Baro.cpp
andyp1per: Tested and fixed

libraries/AP_Baro/AP_Baro_config.h
andyp1per: Thanks, done

tridge (approved):

peterbarker (approved):

peterbarker:
Could have used a shorter field name for CPress - when we add another field to this mesage we will need to break existing pre-canned graphs which include CPRess...

alioguzdemir:
Based on my understanding of the code, the underlying formulation looks something like this:
BARO[1].Press + (CTUN.ThO * sensors[0].mot_scale)

Therefore, to define the sensors[0].mot_scale variable in the backend, we need to assign a value to the BARO1_THST_SCALE parameter. Did I understand this correctly?

In other words, is the relationship strictly:
sensors[0].mot_scale = BARO1_THST_SCALE ?

andyp1per:
Yes that's what happens

PR #28989, opened 2025-01-02
AP_HAL_ChibiOS: add hwdef MFE_AirSpeed_CAN

mikefenghao:
This is the hwdef file for MFE_AirSpeed_CAN. Please help me merge. Thanks。


peterbarker:
Is there some reason this hasn't been done by `include`ing eg. `f103-periph/hwdef.inc` (and the bootloader equivalent) and then adding only what needs to be added?

Copy-and-pasting entire hwdefs around like this increases maintenance load on the project and can lead to bugs.  For example, several of the dfefines around serial use are changing (and have changed) in master since this copy/paste was done.

As an example of how this might be structured - `AeroFox-Airspeed` is doing what I describe.

mikefenghao:
> Is there some reason this hasn't been done by `include`ing eg. `f103-periph/hwdef.inc` (and the bootloader equivalent) and then adding only what needs to be added?
> 
> Copy-and-pasting entire hwdefs around like this increases maintenance load on the project and can lead to bugs. For example, several of the dfefines around serial use are changing (and have changed) in master since this copy/paste was done.
> 
> As an example of how this might be structured - `AeroFox-Airspeed` is doing what I describe.

As you said, I have made changes.

peterbarker (approved):
LGTM

PR #28992, opened 2025-01-02
GCS_MAVLink: emit warning when MISSION_SET_CURRENT received

peterbarker:
slated to be removed - everybody should be using the command which was added 5 years ago...

Needs MissionPlanner modification to send command in place of message.

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -747,7 +747,7 @@ void GCS_MAVLINK::send_mission_current(const class AP_Mission &mission, uint16_t
  */
 void GCS_MAVLINK::handle_mission_set_current(AP_Mission &mission, const mavlink_message_t &msg)
 {
-    // send_received_message_deprecation_warning("MISSION_SET_CURRENT");
+    send_received_message_deprecation_warning("MISSION_SET_CURRENT");
 
     // decode
     mavlink_mission_set_current_t packet;
@@ -4998,11 +4998,7 @@ MAV_RESULT GCS_MAVLINK::handle_command_run_prearm_checks(const mavlink_command_i
 #endif  // AP_ARMING_ENABLED
 
 #if AP_MISSION_ENABLED
-// changes the current waypoint; at time of writing GCS
-// implementations use the mavlink message MISSION_SET_CURRENT to set
-// the current waypoint, rather than this DO command.  It is hoped we
-// can move to this command in the future to avoid acknowledgement
-// issues with MISSION_SET_CURRENT
+// changes the current waypoint
 MAV_RESULT GCS_MAVLINK::handle_command_do_set_mission_current(const mavlink_command_int_t &packet)
 {
     AP_Mission &mission = AP::mission();
--- a/libraries/GCS_MAVLink/GCS_config.h
+++ b/libraries/GCS_MAVLink/GCS_config.h
@@ -36,6 +36,11 @@
 // removed.  It has signficant deficiencies vs MAV_CMD_DO_SET_CURRENT.
 // The command was added to the spec in January 2019 and to MAVLink in
 // ArduPilot in 4.1.x
+// CODE_REMOVAL
+// ArduPilot 4.8 warns if the message is received
+// ArduPilot 4.9 warns if the message is received
+// ArduPilot 4.10 compiles the code out
+// ArduPilot 4.11 removes the code entirely
 #ifndef AP_MAVLINK_MISSION_SET_CURRENT_ENABLED
 #define AP_MAVLINK_MISSION_SET_CURRENT_ENABLED AP_MISSION_ENABLED
 #endif


peterbarker:
This PR for QGC adds support for setting the current sequence number via command, falling back to the old message: https://github.com/mavlink/qgroundcontrol/pull/12285

tridge:

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_03/devcall_pr_reviews.html#pr28992

Reviewed at head `ecbe827b65`. **COMMENT** — the deprecation schedule is right; two problems in the machinery this switches on.

Uncommenting that line makes this PR the **first and only** caller of `send_received_message_deprecation_warning()` — the only other occurrence in the tree is the commented-out line you are uncommenting. So the helper has never run before, and two things in it don't work.

### The 30-second rate limit is inert

`GCS_Common.cpp:6561` reads `if (last_deprecation_warning_send_time_ms - now_ms < 30000)`. The operands are the wrong way round. I ran the unsigned arithmetic rather than reasoning about it: with `last` zero-initialised and calls at 5000, 6000, 7000, 40000 and 100000 ms, **5 of 5 are SENT**. It only ever suppresses if the clock goes backwards.

What actually prevents spam today is the *pointer* comparison at line 6556 (`message == last_deprecation_message`), which with one call site and one string literal matches forever after the first warning. So the real behaviour is **one warning per `GCS_MAVLINK` instance per boot, then silence** — not spam, but not the 30 s cadence intended either, and arguably thin for a campaign meant to span 4.8 and 4.9.

**The obvious fix introduces a new bug**, which is why this is worth more than "swap the operands". With the corrected test and `last` starting at 0, any call before `millis() == 30000` is suppressed — measured: 5000, 6000 and 7000 ms all return silently. That is exactly when a GCS connects after boot. It needs a "never sent" sentinel or an explicit first-time case.

### The message is 52 characters and STATUSTEXT holds 50

`"Received message (MISSION_SET_CURRENT) is deprecated"` is 52 chars, so it chunks as `"...is deprecat"` + `"ed"`, and the onboard log keeps only the first 50. Since this string is the PR's entire user-facing output it is worth shortening — and there is room to name the replacement, which it currently doesn't. `"MISSION_SET_CURRENT is deprecated"` is 33.

### Minor

`GCS_config.h:36`, just above your new `CODE_REMOVAL` block, says `MAV_CMD_DO_SET_CURRENT`; the real replacement is `MAV_CMD_DO_SET_MISSION_CURRENT`. Pre-existing, but you're editing that comment block.

CI is fully green (99 passing). Nothing here blocks merging.

AP-Review:

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_19_AIReview/devcall_pr_reviews.html#pr28992

**Re-reviewed at the same head `ecbe827b65` — my earlier comment above is superseded.** Nothing has changed on the branch; this pass re-read it to check the findings I had already published, and what follows is the part that is new or corrected. I have not restated the findings that still stand unchanged — they are in the linked report.

### Verdict: COMMENT

**Re-read at the same head.** All 99 checks pass. Nothing is blocking this on technical objection — there are no human reviews and no review comments at all — it is blocked on the dependency the author states in the body, and that dependency is **still unresolved 21 months on**: `MissionPlanner ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs:2469` and `:2481` still send `MISSION_SET_CURRENT` from `setWPCurrentAsync`, called from routine UI paths. MAVProxy and QGC have both migrated; Mission Planner is the sole remaining holdout, which is why the warning is wanted — and why it needs to actually be visible.

CI at this head: ✓ 99 passing.

### Findings

- **NOTE** — `libraries/GCS_MAVLink/GCS_Common.cpp:6556` — **Correction to the 2026-09-03 automated review on this PR.** It claimed this change makes `send_received_message_deprecation_warning()` "the first and only caller" and that "the helper has never run before". That is **wrong**: the helper shipped with live callers for years — `SET_HOME_POSITION` (`29a469432e`, 2022-02-08, the commit that introduced it), `MOUNT_CONFIGURE`/`MOUNT_CONTROL` (`ff1624d4a0`, 2022-08-23) and `FENCE_FETCH_POINT`/`RALLY_FETCH_POINT` (`9c42a5d9dd`, 2024-06-06) — until they were removed in `bacda78e0e` (2026-04-28). What **is** true is that master today has zero live callers, so this PR re-activates code dormant since April 2026. Also worth recording: the commented-out line was added *already commented out* in `068443718e` (2023-03-16), so this PR simply enables a hook pre-staged three years ago.

- **BUG** — `libraries/GCS_MAVLink/GCS_Common.cpp:6561` — **The 30-second rate limit is inert; its operands are reversed.** `if (last_deprecation_warning_send_time_ms - now_ms < 30000) { return; }` — both are `uint32_t`, so with `last == 0` and `now == 5000`, `0 - 5000` wraps to 4294962296, which is not `< 30000` and does not suppress. It only ever suppresses if the clock runs backwards. Pre-existing since 2022, but this PR is what makes it live again, so it becomes this PR's behaviour. **The naive fix introduces a second bug**, which is why this is worth more than "swap the operands": with `now_ms - last < 30000` and `last` zero-initialised, calls at 5000, 6000 and 7000 ms were all measured suppressed — every warning in the first 30 s after boot swallowed, which is exactly when a GCS connects. It needs a never-sent sentinel: `if (last_deprecation_warning_send_time_ms != 0 && now_ms - last_deprecation_warning_send_time_ms < 30000) { return; }`. The 30000 is the constant already in the source, and the members at `GCS.h:845-846` are zero at construction because every backend is `NEW_NOTHROW`-allocated.

- **ISSUE** — `libraries/GCS_MAVLink/GCS_Common.cpp:6567` — **The message is 52 characters, STATUSTEXT holds 50, and Mission Planner renders the overflow as a separate garbage line.** `common.xml:6796` defines `char[50]`; `"Received message (%s) is deprecated"` with `MISSION_SET_CURRENT` measures **52**. `GCS::send_textv` (`GCS_Common.cpp:2558`) assigns a chunk id when `strlen > 50` and splits into `"...is deprecat"` plus `"ed"`. Mission Planner does **not** reassemble chunks — `MAVLinkInterface.cs:5379-5389` decodes `msg.text`, truncates at the first NUL and appends it directly; the only `chunk_seq` reference in that repo is the generated struct field. So an MP user sees two lines, the second reading `ed`. The dataflash log also keeps only the first 50 characters. Measured alternatives that fit: `"Deprecated message: %s"` → 39, `"%s is deprecated"` → 33, `"Deprecated msg: %s, use command"` → 48. The first reads best and stays generic, which matters since this is now the only caller.

- **ISSUE** — `libraries/GCS_MAVLink/GCS_Common.cpp:6567` — **`MAV_SEVERITY_INFO` is below Mission Planner's default display threshold, so the warning will not reach the users it is aimed at.** `MAV_SEVERITY_INFO` is 6 and `MAV_SEVERITY_WARNING` is 4 (`common.xml:2891-2899`); Mission Planner gates display with `if (sev <= Settings.Instance.GetInt32("severity", 4))` at `MAVLinkInterface.cs:5400`, default 4. `6 <= 4` is false, so `printit` stays false: no HUD message, no speech, and the text only lands in the scrolling Messages list. Given Mission Planner is the **only** remaining sender, a warning it hides by default does not advance the campaign. `MAV_SEVERITY_WARNING` passes that filter. The severity is pre-existing in the helper, but this PR is what makes it matter.

- **NOTE** — `libraries/GCS_MAVLink/GCS_Common.cpp:6556` — In practice this warns at most once per link per boot, not every 30 s: the pointer dedup `if (message == last_deprecation_message)` compares against a single string literal, so once set it matches forever, and combined with the inert timer the real behaviour is one STATUSTEXT per `GCS_MAVLINK` instance per boot. Good news for the spam question — an MP user clicking "set current WP" repeatedly gets exactly one warning, and MP's five-retry loop cannot amplify it — but it is thin for a campaign meant to run across 4.8 and 4.9, and it is an accident of the broken timer rather than a decision.

- **NOTE** — `Tools/autotest/vehicle_test_suite.py:7633` — ArduPilot's own test suite is still a `MISSION_SET_CURRENT` consumer and will break at the 4.10 step: `set_current_waypoint()` defaults to `set_current_waypoint_using_mission_set_current()` and is called from 40-plus sites across `arducopter.py`, `arduplane.py` and friends. Harmless today, but the new schedule's "4.10 compiles the code out" line breaks every one. `set_current_waypoint_using_mav_cmd_do_set_mission_current()` already exists at `:7608`; flipping the default there is the natural companion commit and would give the command path real CI coverage.

- **NOTE** — `libraries/GCS_MAVLink/GCS_config.h:37` — The comment directly above the new block names a command that does not exist: "It has signficant deficiencies vs `MAV_CMD_DO_SET_CURRENT`" — that identifier has **zero** occurrences in `common.xml`; the real replacement is `MAV_CMD_DO_SET_MISSION_CURRENT` (224), which is also what `common.xml:5427` gives as `replaced_by` on the `` tag. ("signficant" is also misspelled.) Pre-existing, but this is the exact comment block being edited. The block's claim that the command "was added to the spec in January 2019" also does not match the XML's own 2022-08 supersession date. Separately, the declared schedule starts at 4.8 and `ArduCopter/version.h` reads `V4.8.0-dev`, so if this slips past the 4.8 branch point all four lines need bumping — with no objections on the thread, merging or renumbering is the actionable item.

PR #28994, opened 2025-01-02
LD06/LD19 measured distance stability improvement

boa-pe:
Fix for Lidar measured distance fluctuations by little transplant from RPLidar driver.

So far tested only on desk with LD19. Test flight as soon as weather is good enough.

### DIFF
--- a/libraries/AP_Proximity/AP_Proximity_LD06.cpp
+++ b/libraries/AP_Proximity/AP_Proximity_LD06.cpp
@@ -116,8 +116,6 @@ void AP_Proximity_LD06::get_readings()
 
                 // Updates the temporary boundary and passes off the completed data
                 parse_response_data();
-                _temp_boundary.update_3D_boundary(state.instance, frontend.boundary);
-                _temp_boundary.reset();
 
                 // Resets the bytes read and whether or not we are reading data to accept a new payload
                 _byte_count = 0;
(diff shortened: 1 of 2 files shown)


peterbarker:
@boa-pe is this a duplicate of https://github.com/ArduPilot/ardupilot/pull/27045 ?

boa-pe:
> @boa-pe is this a duplicate of #27045 ?

It seems to solve the same issue and I see also some parts of code from RPLidar too. I just started looking into the code just because my LD19 was going crazy.

peterbarker:
Sorry, this PR has been made against 4.5 - we don't merge things directly into stable branches, things have to go via master first.

I was unable to distinguish the behaviour of your PR vs that other PR.  I've tested that other PR (again) and merged it.

Please create a PR against master with the changes you'd like to make - please make the changes minimal and explain how I should test to see an improvement.

Interestingly my device does not have a cable coming out of the back of the device - but there *is* an arrow on top of the device and that does indicate the correct orientation.  So I'm not sure our documentation for this device is correct (@Hwurzburg)  The phrase "lidar should be mounted horizontally on the top or bottom of the vehicle with the black cable pointing towards the rear of the vehicle" was probably copied from the rplidara2 page.  The correct phrase should be, "lidar should be mounted horizontally on the top or bottom of the vehicle with the arrow on top of the unit pointing towards the front of the vehicle"

Again, please open with a PR against master.  If there's a problem with the logic in here then the same problem may exist in the RPiLIDAR code too...

Hwurzburg:
done https://github.com/ArduPilot/ardupilot_wiki/pull/6528

PR #28998, opened 2025-01-03
AP_Soaring: Move define to config file

Ryanf55:
* Prepare for more compile time additional features in soaring that will be enabled/disabled in the config header


peterbarker (approved):
LGTM

PR #23896, opened 2023-05-25
AP_ESC_Telem: Improve handling of time wrap with RPM data

WickedShell:
The primary target here is to close a window where after time wraps we would ignore the last reported RPM data even if it was valid (such as in the are_motors_running function). The secondary thing this helps us with is that by zeroing out the time we can ensure that if we lose telemetry after wrapping around we don't have a window of time where we consider it to be valid again.

This PR is worth reviewing in separate commits. The second commit is a combination of me kibitzing, and attempting some drive by improvements. They're detailed in the commit message and should be reviewed carefully.

Thus far this has been run through the autotest sequence, but needs to be tested on real hardware still. I'm hoping to get back to that shortly, and do some explicit testing of the behavior on wraps.

PR #28999, opened 2025-01-03
AP_ESC_Telem: fix RPM timeout race

robertlong13:
Alternative approach to #28987.

We don't need to worry about explicitly checking the timestamp in most places; they can just use the `data_valid` flag directly. This eliminates the race without harming the performance of `get_rpm` at all (in fact, it improves it, even if just barely). The only places that need to actually check the timestamp are `update`, where the `data_valid` gets cleared on timeout, and `are_motors_running`, which has a stricter timeout than `data_valid` does. These two run much less frequently than `get_rpm`, so the extra copying shouldn't be a problem.

This might change the timing of when the data is considered stale by a few microseconds, since it's checking the timestamp at 100Hz instead of every time we get RPM, but that should be absolutely fine. The other change you might notice is that previously `rpm_data_within_timeout` explicitly checked if `last_update_us` was zero and I don't. That check was unnecessary in there though, because `data_valid` cannot possibly be true if `last_update_us` is zero.

I actually really like this approach. I'll probably do something similar to solve the `TelemetryData::stale` race. And if we don't mind the signed-int stuff in my other PR, then I'll take this idea over to that PR.

This approach does not solve the potential race in the `slew` calculation, but I'm guessing that's probably fine. If it was a problem, we probably would have noticed by now. My other approach didn't fully fix the slew calculation either (if you interrupted the backend after it updated `prev_rpm` and `rpm`, but before it updates `last_update_us`, then you'll get a little jump; that should be very rare though).

### DIFF
--- a/libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
+++ b/libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
@@ -105,14 +105,12 @@ uint8_t AP_ESC_Telem::get_motor_frequencies_hz(uint8_t nfreqs, float* freqs) con
 uint32_t AP_ESC_Telem::get_active_esc_mask() const {
     uint32_t ret = 0;
     const uint32_t now = AP_HAL::millis();
-    uint32_t now_us = AP_HAL::micros();
     for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
         if (_telem_data[i].last_update_ms == 0 && !was_rpm_data_ever_reported(_rpm_data[i])) {
             // have never seen telem from this ESC
             continue;
         }
-        if (_telem_data[i].stale(now)
-            && !rpm_data_within_timeout(_rpm_data[i], now_us, ESC_RPM_DATA_TIMEOUT_US)) {
+        if (_telem_data[i].stale(now) && !_rpm_data[i].data_valid) {
             continue;
         }
         ret |= (1U << i);
@@ -126,14 +124,12 @@ uint8_t AP_ESC_Telem::get_max_rpm_esc() const
     uint32_t ret = 0;
     float max_rpm = 0;
     const uint32_t now = AP_HAL::millis();
-    const uint32_t now_us = AP_HAL::micros();
     for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
         if (_telem_data[i].last_update_ms == 0 && !was_rpm_data_ever_reported(_rpm_data[i])) {
             // have never seen telem from this ESC
             continue;
         }
-        if (_telem_data[i].stale(now)
-            && !rpm_data_within_timeout(_rpm_data[i], now_us, ESC_RPM_DATA_TIMEOUT_US)) {
+        if (_telem_data[i].stale(now) && !_rpm_data[i].data_valid) {
             continue;
         }
         if (_rpm_data[i].rpm > max_rpm) {
@@ -153,13 +149,12 @@ uint8_t AP_ESC_Telem::get_num_active_escs() const {
 // return the whether all the motors in servo_channel_mask are running
 bool AP_ESC_Telem::are_motors_running(uint32_t servo_channel_mask, float min_rpm, float max_rpm) const
 {
-    const uint32_t now = AP_HAL::micros();
 
     for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
         if (BIT_IS_SET(servo_channel_mask, i)) {
             const volatile AP_ESC_Telem_Backend::RpmData& rpmdata = _rpm_data[i];
             // we choose a relatively strict measure of health so that failsafe actions can rely on the results
-            if (!rpm_data_within_timeout(rpmdata, now, ESC_RPM_CHECK_TIMEOUT_US)) {
+            if (!rpm_data_within_timeout(rpmdata, ESC_RPM_CHECK_TIMEOUT_US)) {
                 return false;
             }
             if (rpmdata.rpm < min_rpm) {
@@ -201,7 +196,7 @@ bool AP_ESC_Telem::get_rpm(uint8_t esc_index, float& rpm) const
     }
 
     const uint32_t now = AP_HAL::micros();
-    if (rpm_data_within_timeout(rpmdata, now, ESC_RPM_DATA_TIMEOUT_US)) {
+    if (rpmdata.data_valid) {
         const float slew = MIN(1.0f, (now - rpmdata.last_update_us) * rpmdata.update_rate_hz * (1.0f / 1e6f));
         rpm = (rpmdata.prev_rpm + (rpmdata.rpm - rpmdata.prev_rpm) * slew);
 
@@ -225,9 +220,7 @@ bool AP_ESC_Telem::get_raw_rpm(uint8_t esc_index, float& rpm) const
 
     const volatile AP_ESC_Telem_Backend::RpmData& rpmdata = _rpm_data[esc_index];
 
-    const uint32_t now = AP_HAL::micros();
-
-    if (!rpm_data_within_timeout(rpmdata, now, ESC_RPM_DATA_TIMEOUT_US)) {
+    if (!rpmdata.data_valid) {
         return false;
     }
 
@@ -413,7 +406,6 @@ void AP_ESC_Telem::send_esc_telemetry_mavlink(uint8_t mav_chan)
     }
 
     const uint32_t now = AP_HAL::millis();
-    const uint32_t now_us = AP_HAL::micros();
 
     // loop through groups of 4 ESCs
     const uint8_t esc_offset = constrain_int16(mavlink_offset, 0, ESC_TELEM_MAX_ESCS-1);
@@ -435,8 +427,7 @@ void AP_ESC_Telem::send_esc_telemetry_mavlink(uint8_t mav_chan)
         for (uint8_t j=0; j<4; j++) {
             const uint8_t esc_id = (i * 4 + j) + esc_offset;
             if (esc_id < ESC_TELEM_MAX_ESCS &&
-                (!_telem_data[esc_id].stale(now) ||
-                 rpm_data_within_timeout(_rpm_data[esc_id], now_us, ESC_RPM_DATA_TIMEOUT_US))) {
+                (!_telem_data[esc_id].stale(now) || _rpm_data[esc_id].data_valid)) {
                 all_stale = false;
                 break;
             }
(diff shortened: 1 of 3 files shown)


libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -153,13 +149,19 @@ uint8_t AP_ESC_Telem::get_num_active_escs() const {
...
     for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
         if (BIT_IS_SET(servo_channel_mask, i)) {
             const volatile AP_ESC_Telem_Backend::RpmData& rpmdata = _rpm_data[i];
-            // we choose a relatively strict measure of health so that failsafe actions can rely on the results
-            if (!rpm_data_within_timeout(rpmdata, now, ESC_RPM_CHECK_TIMEOUT_US)) {
+            // copy the last_updated_us timestamp to avoid any race issues
+            const uint32_t last_updated_us = rpmdata.last_update_us;
+            const uint32_t now = AP_HAL::micros();
+            // we choose a stricter timeout than the data_valid flag so that failsafe actions can rely on the results
+            if (now - last_updated_us > ESC_RPM_CHECK_TIMEOUT_US) {
```
andyp1per, line 160: Please use AP_HAL::timeout_expired() here

andyp1per: Surely this is just rpm_data_within_timeout() in which case better to leave it in.

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -808,29 +806,17 @@ void AP_ESC_Telem::update()
...
 #endif  // HAL_LOGGING_ENABLED
 
-    const uint32_t now_us = AP_HAL::micros();
     for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
+        // copy the last_updated_us timestamp to avoid any race issues
+        const uint32_t last_updated_us = _rpm_data[i].last_update_us;
+        const uint32_t now_us = AP_HAL::micros();
         // Invalidate RPM data if not received for too long
-        if ((now_us - _rpm_data[i].last_update_us) > ESC_RPM_DATA_TIMEOUT_US) {
+        if ((now_us - last_updated_us) > ESC_RPM_DATA_TIMEOUT_US) {
```
andyp1per, line 814: Please use AP_HAL::timeout_expired() here

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -808,29 +806,17 @@ void AP_ESC_Telem::update()
...
             _rpm_data[i].data_valid = false;
         }
     }
 }
 
-bool AP_ESC_Telem::rpm_data_within_timeout(const volatile AP_ESC_Telem_Backend::RpmData &instance, const uint32_t now_us, const uint32_t timeout_us)
```
andyp1per, line 820: I think leave this in for the check above

andyp1per (changes requested):
@robertlong13 I am happier with this approach, but I need to point out that this code is very, very subtle with all sorts of bad things happening (crashed aircraft) if we get it wrong. @magicrub originally did the rpm_data_within_timeout() changes which fundamentally relies on the 0 value for the timeout so we should get him to review and also check his PR. We should also be striving for the smallest change possible that is demonstrable correct.

andyp1per:
We also need a comment about the 100Hz update in AP_Vehicle as we are now heavily reliant on that. Probably also needs a comment in AP_Vehicle.

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
robertlong13: Huh, cool helper. Done.

robertlong13:
> We also need a comment about the 100Hz update in AP_Vehicle as we are now heavily reliant on that. Probably also needs a comment in AP_Vehicle.

I wouldn't say we're heavily reliant on 100Hz. It **definitely** needs to be faster than 0.0002Hz, but even if it was called as slow as 10Hz (which is what it used to be, but I doubt we're going back to that, and we definitely won't be going slower than it), our arbitrary 1s timeout would be worst-case 1.1s, and average 1.05s.

I'll chuck a comment in AP_Vehicle anyway.

robertlong13:
> @magicrub originally did the rpm_data_within_timeout() changes which fundamentally relies on the 0 value for the timeout so we should get him to review and also check his PR. 

Really? I only saw the [PR](https://github.com/ArduPilot/ardupilot/pull/23896) from @WickedShell, which is why I marked him for review (thought forgot to mark him for this second one).

andyp1per:
> Really? I only saw the [PR](https://github.com/ArduPilot/ardupilot/pull/23896) from @WickedShell, which is why I marked him for review (though I forgot to mark him for this second one, thanks for reminding me).

Sorry my mistake - yes @WickedShell - there were issues with motors in transition IIRC

andyp1per (approved):
This looks ok to me, but I would like @WickedShell to verify

WickedShell (approved):
It's been awhile since I've dove through this, so I'm not 100% confident, but it generally seems fine.

It's worth noting that all the other accessors (get_active_esc_mask/get_max_rpm_esc etc) all got more expensive as they are now all getting their own copy of time. (I know getting the timestamp in milliseconds was surprisingly expensive, I'm not sure if the micros one is as expensive though).

As an aside all of the time management stuff is a here be dragons area, I don't think this makes it any worse then it was.

andyp1per:
Given the sensitivity of the code I think we will need to fly before merging

robertlong13:
> It's worth noting that all the other accessors (get_active_esc_mask/get_max_rpm_esc etc) all got more expensive as they are now all getting their own copy of time.

It's the other way around. Those accessors got less expensive, as now they don't need a copy of time, they just check the `data_valid` flag.

The only functions that got any more expensive are `are_motors_running` and `update`

WickedShell:
> > It's worth noting that all the other accessors (get_active_esc_mask/get_max_rpm_esc etc) all got more expensive as they are now all getting their own copy of time.
> 
> It's the other way around. Those accessors got less expensive, as now they don't need a copy of time, they just check the `data_valid` flag.
> 
> The only functions that got any more expensive are `are_motors_running` and `update`

Right, I was referring to anything leaning against `rpm_data_within_timeout` but then I managed to get confused while jumping around through things here, and thought it had been left in more places then it was. Sorry!

libraries/AP_Vehicle/AP_Vehicle.cpp
```diff
@@ -624,6 +624,8 @@ const AP_Scheduler::Task AP_Vehicle::scheduler_tasks[] = {
 #endif
     SCHED_TASK(send_watchdog_reset_statustext,         0.1,     20, 225),
 #if HAL_WITH_ESC_TELEM
+    // This update function is responsible for checking timeouts and invalidating the ESC telemetry data.
+    // Be mindful of this if you are planning to reduce the frequency from 100Hz.
```
IamPete1, line 628: On plane/rover this will run at 50Hz, I assume that is still OK?

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -126,14 +124,12 @@ uint8_t AP_ESC_Telem::get_max_rpm_esc() const
...
     const uint32_t now = AP_HAL::millis();
-    const uint32_t now_us = AP_HAL::micros();
     for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
         if (_telem_data[i].last_update_ms == 0 && !was_rpm_data_ever_reported(_rpm_data[i])) {
             // have never seen telem from this ESC
             continue;
         }
-        if (_telem_data[i].stale(now)
-            && !rpm_data_within_timeout(_rpm_data[i], now_us, ESC_RPM_DATA_TIMEOUT_US)) {
+        if (_telem_data[i].stale(now) && !_rpm_data[i].data_valid) {
```
IamPete1, line 132:
```suggestion
        if (_telem_data[i].stale(now) || !_rpm_data[i].data_valid) {
```
If data stale or not valid then don't used it?

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
```diff
@@ -105,14 +105,12 @@ uint8_t AP_ESC_Telem::get_motor_frequencies_hz(uint8_t nfreqs, float* freqs) con
...
     const uint32_t now = AP_HAL::millis();
-    uint32_t now_us = AP_HAL::micros();
     for (uint8_t i = 0; i < ESC_TELEM_MAX_ESCS; i++) {
         if (_telem_data[i].last_update_ms == 0 && !was_rpm_data_ever_reported(_rpm_data[i])) {
             // have never seen telem from this ESC
             continue;
         }
-        if (_telem_data[i].stale(now)
-            && !rpm_data_within_timeout(_rpm_data[i], now_us, ESC_RPM_DATA_TIMEOUT_US)) {
+        if (_telem_data[i].stale(now) && !_rpm_data[i].data_valid) {
```
IamPete1, line 113:
```suggestion
        if (_telem_data[i].stale(now) || !_rpm_data[i].data_valid) {
```
If data stale or not valid then don't used it?

libraries/AP_Vehicle/AP_Vehicle.cpp
robertlong13: Yeah, that's perfectly okay. The timeout is configured to be 1s, and has a worst case actual time of 1s + 1/freq. Even as low as 10Hz should be perfectly fine IMO, especially on platforms with low main-loop rates.

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
robertlong13: That would be a change to the current behavior. I don't understand why, but this logic was very clearly intentional. The function's description even mentions "and/or" instead of "and"

robertlong13:
Tested on a 15-minute flight on Volanti, and ground-tested the timeout behavior. Unplugging ESCs registers a timeout after 1 second as expected.

libraries/AP_Vehicle/AP_Vehicle.cpp
IamPete1: Thanks

libraries/AP_ESC_Telem/AP_ESC_Telem.cpp
IamPete1: It seems to be because there are two timeouts. One for "everything else" and one for "RPM". So the assumption is that if we got "everything else" recently then a timed out RPM is better than nothing. I'm not sure that really makes sense, it would mean we use RPM from more than one second ago. I wonder if this is a bodge to make up for the issue your fixing.

Since we also have a timeout for "everything else" I think we also need a fix the same as the one you have done for RPM for it?

robertlong13: > we use RPM from more than one second ago

Not really. This logic is only in `get_active_esc_mask` and `get_max_rpm_esc`, the latter is only used for OSD. `get_rpm` checks the timeout of RPM only, not the other telem, so we don't have an issue. Arguably, the "other telem" timeout check is unnecessary in `get_max_rpm_esc` and should be dropped entirely, not just changed to `||`.

> Since we also have a timeout for "everything else" I think we also need a fix the same as the one you have done for RPM for it?

Yup: https://github.com/ArduPilot/ardupilot/pull/29028

IamPete1: OK, its a bit odd but I'm convinced. Thanks!

PR #29000, opened 2025-01-03
Copter: switch off fast rate while doing temperature calibration

andyp1per:
Tested on an F405 with an ICM42688.

This *may* not be required, but are not running the fast rate code when the gyros are initialized/calibrated and it makes me nervous that the increased CPU load may warp the results.

### DIFF
--- a/ArduCopter/rate_thread.cpp
+++ b/ArduCopter/rate_thread.cpp
@@ -215,7 +215,11 @@ void Copter::rate_controller_thread()
 #endif
 
         // allow changing option at runtime
-        if (get_fast_rate_type() == FastRateType::FAST_RATE_DISABLED) {
+        if (get_fast_rate_type() == FastRateType::FAST_RATE_DISABLED
+#if HAL_INS_TEMPERATURE_CAL_ENABLE
+            || ins.temperature_cal_running()
+#endif
+            ) {
             if (was_using_rate_thread) {
                 disable_fast_rate_loop(rates);
                 was_using_rate_thread = false;


tridge (approved):

rmackay9:
Sorry to be nitpicky but to help the review go more smoothly we really should always be providing a description clarify what testing has been done

tridge:
@andyp1per can you also explain why this is needed?

peterbarker:
@andyp1per tridge asked why this is required.

Some convincing graphs would be good IMO :-)

tridge:

<details><summary>Previous review (2026-09-12)</summary>

Reviewed at head `dfff83a3a3`.

Full report (with the evidence behind each point): https://uav.tridgell.net/DevCallReviews/2026_09_12_AIReview/devcall_pr_reviews.html#pr29000

**Verdict: APPROVE**

One-hunk change at ArduCopter/rate_thread.cpp:218-222: extends the existing runtime "should the rate thread be running" guard so that ins.temperature_cal_running() also parks the fast rate loop, on the theory (PR body) that the extra CPU load could warp the tcal results.

**CI at this head:** 85 passing / 1 failing / 0 pending (86 rows). Failing job: build-test from the colcon build/test workflow (run 33803778220, job 100809242494). The log ends at "ERROR: the following packages/stacks could not have their rosdep keys resolved" during rosdep update — a ROS2 dependency-resolution infrastructure failure, entirely unrelated to a change in ArduCopter/rate_thread.cpp.

<details><summary>Lower-severity notes (4)</summary>

- NOTE | VERIFIED | ArduCopter/rate_thread.cpp:219-221 | The #if HAL_INS_TEMPERATURE_CAL_ENABLE guard is redundant. bool AP_InertialSensor::temperature_cal_running() const is declared unconditionally at libraries/AP_InertialSensor/AP_InertialSensor.h:142 and its body (AP_InertialSensor.cpp:2322-2331) already #ifs internally and returns false when the feature is off — which the compiler will constant-fold. Dropping the two preprocessor lines would make the condition a single readable expression. Cosmetic only.
- NOTE | VERIFIED | tcal can in principle be started WHILE ARMED, bypassing the interlock. temperature_cal_running() reads tcal(i).enable == LearnCalibration live from the parameter, with no arming gate on the write. So setting INS_TCAL1_ENABLE=2 in flight would trip this new condition and tear down the fast rate loop mid-flight. However this is NOT a regression introduced here: the identical transition is already reachable at runtime by setting RATE_THREAD/att_enable to 0, which is what the pre-existing comment "allow changing option at runtime" (rate_thread.cpp:217) is for. Worth a sentence in the commit message.
- NOTE | UNCONFIRMED | No autotest; would not fail on revert. Only ArduCopter/rate_thread.cpp is touched. This is also hard to test meaningfully — the claim is about tcal ACCURACY, not about a state transition, and the author himself says "This *may* not be required".
- NOTE | VERIFIED | The thread asks for evidence and it has not been supplied. tridge asked "can you also explain why this is needed?" (2025-01-07), rmackay9 asked for a description of testing done, and peterbarker asked for "some convincing graphs" (2025-05-01). The body says only "This *may* not be required... it makes me nervous". tridge's APPROVED review predates all three questions. Since the change costs nothing if unnecessary and the risk is low, the graphs are the only thing gating it.

</details>

**What was checked** (this is the evidence the clearance rests on — please push back if something important is missing):
- DISABLE/RE-ENABLE IS SYMMETRIC ON EVERY PATH. The condition is re-evaluated at the top of every loop iteration (rate_thread.cpp:218). While it holds, disable_fast_rate_loop(rates) runs once (guarded by was_using_rate_thread) and the thread spins on a 500 us delay. The instant temperature_cal_running() goes false — normal completion, an aborted calibration, or the user writing INS_TCALn_ENABLE back to any other value — execution falls through to if (!using_rate_thread) enable_fast_rate_loop(...) at rate_thread.cpp:229-230. There is no exit path that can leave the loop disabled. On reboot the state is rebuilt from scratch: Copter.cpp:824 only starts the thread at all if get_fast_rate_type() != FAST_RATE_DISABLED, and was_using_rate_thread is initialised false at rate_thread.cpp:190, so disable_fast_rate_loop() is correctly NOT called on a first iteration where enable_fast_rate_loop() never ran.
- Read in the right place: the flag is read in the rate thread's own top-of-loop guard, alongside the existing FAST_RATE_DISABLED check — the same place the existing runtime teardown is driven from.
- Atomicity: temperature_cal_running() reads an AP_Enum<Enable> (single byte) per IMU; a torn read is not possible and a one-iteration-stale value is harmless given the 500 us poll. using_rate_thread is non-volatile but that is pre-existing (ArduCopter/Copter.h:1115) and not touched here.
- THE ARMING INTERLOCK EXISTS AND IS REAL: libraries/AP_Arming/AP_Arming.cpp:576-580 — if (ins.temperature_cal_running()) { check_failed(Check::INS, report, "temperature cal running"); return false; }. The vehicle cannot be armed while tcal is running, so the normal case of disabling the fast rate loop is always on the ground and disarmed.
- Completion path confirmed: AP_InertialSensor.cpp:2010-2017 clears tcal_learning and emits "TCAL finished all IMUs" once temperature_cal_running() returns false, so the transition back is a real observable event.
- disable_fast_rate_loop() / enable_fast_rate_loop() (rate_thread.cpp:446-461) are exact mirrors: ins.enable/disable_fast_rate_buffer(), rate_controller_set_rates(), hal.rcout->force_trigger_groups(true/false), using_rate_thread = true/false.

The change is correct, symmetric, and low-risk, and the one CI failure is unrelated infrastructure. Merge it
once the "why" question from tridge/peterbarker is answered — a tcal run with and without the rate thread
would settle it.

_Both an independent Codex pass and this review were run on this head and reconciled; findings that one pass refuted are not listed above._

</details>

AP-Review:

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_19_AIReview/devcall_pr_reviews.html#pr29000

**Re-reviewed at the same head `dfff83a3a3` — my earlier comment above is superseded.** Nothing has changed on the branch; this pass re-read it to check the findings I had already published, and what follows is the part that is new or corrected. I have not restated the findings that still stand unchanged — they are in the linked report.

### Verdict: COMMENT

**Re-read at the same head.** The five-line change is correct, symmetric and safe in both directions, and the single failing check is not caused by it — `build-test` is the ROS2 colcon workflow, dying at `rosdep install` with `Cannot locate rosdep definition for [microxrcedds_agent]`, which a change to `ArduCopter/rate_thread.cpp` cannot cause; all eight copter autotest shards, the heli shard and every board build pass. What blocks it is a 20-month-old unanswered question: tridge asked "can you also explain why this is needed?" on 2025-01-07, rmackay9 asked what testing was done, and peterbarker followed up on 2025-05-01 asking for "some convincing graphs". None has been answered. (tridge's APPROVE predates all three by minutes.) The important item this round is a correction.

CI at this head: ✗ 1 failing, 85 passing.

### Findings

- **NOTE** — `ArduCopter/rate_thread.cpp:219` — **Correction to the 2026-09-12 automated review on this PR: the `#if HAL_INS_TEMPERATURE_CAL_ENABLE` guard is NOT redundant and should be kept as written.** That review said the guard could be dropped because `temperature_cal_running()` already `#if`s internally and "the compiler will constant-fold" it. It cannot: `bool AP_InertialSensor::temperature_cal_running() const` is defined **out-of-line** in `libraries/AP_InertialSensor/AP_InertialSensor.cpp:2322`, declared unconditionally at `AP_InertialSensor.h:142`, and `Tools/ardupilotwaf/boards.py` contains **no `-flto`** — so with no LTO this is a real cross-translation-unit call that the guard elides on small boards. `HAL_INS_TEMPERATURE_CAL_ENABLE` defaults to `HAL_PROGRAM_SIZE_LIMIT_KB > 1024` (`AP_InertialSensor_config.h:16`) and is forced to 0 on esp32, so 1 MB boards genuinely benefit. Acting on the earlier note would have cost flash on exactly those boards. (The macro is also genuinely visible in this TU: `rate_thread.cpp` includes `Copter.h` → `AP_InertialSensor.h` → `AP_InertialSensor_config.h`. Had it not been, `#if` on an undefined identifier would have silently compiled the whole change away with no warning.)

- **NOTE** — `ArduCopter/rate_thread.cpp:223` — A pre-existing window where `was_using_rate_thread` can lag `using_rate_thread`, which this PR gives a second trigger. `enable_fast_rate_loop()` sets `using_rate_thread = true` at `:455`, but `was_using_rate_thread = true` is only reached at `:407`, at the bottom of the loop body — after the `if (!ins.get_next_gyro_sample(gyro)) { continue; }` early-out at `:240`. If that early-out fires on the first iteration after an enable (reachable if `FastRateBuffer` allocation returned nullptr, or a stale `_notifier` signal makes `wait_blocking()` return on an empty window) and tcal starts on the very next iteration, the gate at `:218` is taken with `was_using_rate_thread == false`, so `disable_fast_rate_loop()` is skipped: `using_rate_thread` stays true while the thread parks on the 500 µs delay, and `Attitude.cpp:17` then also skips `rate_controller_run()` — **no rate controller runs at all**. Not armable (`AP_Arming.cpp:576-580` blocks arming during tcal), self-healing the iteration after tcal ends, and the identical hole already exists for the `FAST_RATE_DISABLED` arm of the same `if`, so it is not a regression. Minimal fix would be to set the flag at the point of enable rather than at the bottom of the loop — but that also changes the `|| !was_using_rate_thread` notch re-init at `:332`, so it is a judgement call; flagging, not insisting.

- **NOTE** — `libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp:493` — **A stronger answer to tridge's "why is this needed?" than CPU load exists in the code, and the author should use it.** Beyond raw CPU, enabling the fast rate loop measurably changes the IMU sample path that tcal learns from: `AP_InertialSensor_Invensensev3::set_primary()` (`:493-503`) reacts to `is_dynamic_fifo_enabled()` — true only while `fast_rate_buffer_enabled` — by re-timing the FIFO drain callback for **non-primary** IMUs to `1000000UL / constrain_int16(get_loop_rate_hz() * 2, 400, 1000)`. Coarser drains mean each batch of samples is paired with a staler `_imu.get_temperature(instance)` in the learning calls at `AP_InertialSensor_Backend.cpp:124/161`, which is exactly the temperature-versus-reading pairing tcal fits. That is a concrete, testable mechanism and would satisfy peterbarker's request far better than a CPU-load argument. Caveat: `set_primary()` only runs when the primary actually changes, so the magnitude is board-dependent and was not measured here — this is a suggested line of evidence, not a claim that it matters.

- **NOTE** — `ArduCopter/rate_thread.cpp:218` — Two smaller observations. There is a one-main-loop asymmetry at the end of calibration: the gate reads `temperature_cal_running()` (the parameter state) while the learning path is driven by `tcal_learning`, and `finish_calibration()` writes `tcal.enable.set_and_save_ifchanged(Enabled)` at `AP_InertialSensor_tempcal.cpp:418` while `tcal_learning` is only cleared later at `AP_InertialSensor.cpp:2011-2017` — so for up to one main-loop period the rate thread has re-enabled itself while samples are still being fed to the learner. Against a multi-minute polynomial fit that is noise, and reading the parameter is the *more* conservative choice at the **start** of calibration, so no change is recommended — just worth a sentence if the design is documented. And there is no test coverage: `FSTRATE_ENABLE` appears in autotest only at `arducopter.py:9455` and `:9476`, and `git grep` for `temperature_cal|TCAL` in that file returns nothing, so nothing would fail on a revert. That is genuinely hard to autotest, because the claim is about tcal *accuracy*, not a state transition.

Issue #28807, opened 2024-12-05
Missing/ empty/ extra ESC telemetry log data


MarbleJames:
## Bug report

**Issue details**
I have a Quadplane with 5 esc’s connected. Since updating from 4.4.4, extra empty ESC’s (up to 8 in total) have been appearing in logs and the cruise ESC data is often empty.

On occasion data for 5 esc's has been observed in mission planner but only data for 4 has been logged.

Just verified to be a firmware issue as follows;

Flashed back to 4.4.4 - 5 esc’s correct data
Flashed 4.6.0 - 8 esc’s 4 with data 4 without
Flashed to 4.4.4 - 5 esc’s correct again.
I haven’t tried all the firmware releases in-between but 4.5.4 and 4.5.0 have also been shown to not work.

(https://discuss.ardupilot.org/t/esc-telemetry-phantom-escs/127106)

**Version**
Post 4.4.4

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[ X ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
4+1 quadplane

**Hardware type**
Cube orange plus bdshot, 4x apd80, 1x apd120

**Logs**
Two logs included, both same params one 4.4.4 and one 4.6.0
[ESC telem logs.zip](https://github.com/user-attachments/files/18020578/ESC.telem.logs.zip)
https://drive.google.com/drive/folders/17_GhgOKQxd0-tNSXUfivTVYeqpDNg5mY?usp=sharing

andyp1per:
This is a change in 4.5 to support dshot on IOMCU. ESC numbers are now aligned with servo numbers.

MarbleJames:
> This is a change in 4.5 to support dshot on IOMCU. ESC numbers are now aligned with servo numbers.

Thanks for the reply, I don’t understand how that affects this though, as far as I understand IOMCU is only the first 8 channels of cube orange? I’m running esc’s on 9-13. Anywhere i can read more on this.

I can deal with empty ESC’s in logs but surely this shouldn’t stop recording the telem stream for the cruise ESC? (especially when i'm getting live telemetry for the 5th in mission planner)

v4.4.4
RC9 - ESC[0]
RC10 - ESC[1]
RC11 - ESC[2]
RC12 - ESC[3]
RC13 - ESC[4]

v4.6.0
RC9 - ESC[1]
RC10 - ESC[2]
RC11 - ESC[3]
RC12 - ESC[4]
[...]

PR #29001, opened 2025-01-04
AP_BLHeli: normalize ESC index correctly with iomcu

andyp1per:
Closes https://github.com/ArduPilot/ardupilot/issues/28807

Tested with the original issue's parameters on a CubeOrangePlus

### DIFF
--- a/libraries/AP_BLHeli/AP_BLHeli.cpp
+++ b/libraries/AP_BLHeli/AP_BLHeli.cpp
@@ -1469,7 +1469,7 @@ void AP_BLHeli::read_telemetry_packet(void)
 
     uint8_t normalized_motor_idx = motor_idx - chan_offset;
 #if HAL_WITH_IO_MCU
-    if (AP_BoardConfig::io_dshot()) {
+    if (AP_BoardConfig::io_enabled()) {
         normalized_motor_idx = motor_idx;
     }
 #endif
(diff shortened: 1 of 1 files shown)


tridge:
waiting for user to test

libraries/AP_BLHeli/AP_BLHeli.cpp
```diff
@@ -1469,7 +1469,7 @@ void AP_BLHeli::read_telemetry_packet(void)
 
     uint8_t normalized_motor_idx = motor_idx - chan_offset;
 #if HAL_WITH_IO_MCU
-    if (AP_BoardConfig::io_dshot()) {
+    if (AP_BoardConfig::io_enabled()) {
```
tridge, line 1472: this seems quite convoluted, chan_offset is 8 if io_enabled(), zero otherwise, seems to be undoing what we have in init? can you explain what testing has been done on this, with and without IO enabled?

tridge:
needs backport to 4.6 ?

libraries/AP_BLHeli/AP_BLHeli.cpp
tridge: can the subtraction motor_idx - chan_offset underflow?

andyp1per: You are right, this can be cleaned up more. Since 4.5 the ESC index needs to reflect the servo index. This has been done in the bdshot code but was missed here.

andyp1per: Fixed

rmackay9:
This has been successfully tested by the user (see bottom of https://github.com/ArduPilot/ardupilot/issues/28807)

tridge (approved):

PR #27053, opened 2024-05-13
AP_Mount: add topotek gimbal driver

rmackay9:
This adds support for the [Topotek gimbals](https://www.topotek.com/) and replaces https://github.com/ArduPilot/ardupilot/pull/26666

This has been heavily reworked to reduced its flash size (now 6660 on CubeOrangePlus) and to make it more robust to errors.

This has been tested on real hardware including confirming that:

- angle control works
- rate control works
- camera take-picture and record-video works
- pictures and videos are recorded to the SD card with the correct date
- pictures include EXIF data including accurate location where the picture was taken

Some of the changes compared to [the initial PR](https://github.com/ArduPilot/ardupilot/pull/26666) are:

- display firmware version to user
- switch automatically to RC control
- simplfy set-gimbal-lock
- replace use of ctime
- use unix line endings
- comment improvements

The only known issues are:

- some use of "int" and "long" (see send_location_info() and gimbal_dist_info_analyse()
- the model version is reported as "Unknown"
- during yaw rate control the gimbal always moves in "earth-frame".  Once it stops rotating (e.g. user centers sticks) the gimbal does correctly maintain either earth-frame or body-frame according to the "lock" vs "follow" state.  The DJI RS2/RS3 gimbals also act like this so it's not a blocker

@laozhou-fujian this builds on your PR and the single commit is now Co-authored by both of us.  Can you confirm you are OK with these changes?

Size comparison
image

Wiki PR https://github.com/ArduPilot/ardupilot_wiki/pull/6092

PR #29002, opened 2025-01-04
AP_Mount: Topotek pitch rate direction fix

rmackay9:
This fixes the direction of pitch **rate** control in the Topotek gimbal driver (**angle** control is fine).

This bug has existed since the driver was first created in PR https://github.com/ArduPilot/ardupilot/pull/27053 but we somehow missed it during testing.  If you look closely the bug is fairly obvious from inspection because we can see the [angle control includes the minus](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Mount/AP_Mount_Topotek.cpp#L819) while the [rate control does not](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Mount/AP_Mount_Topotek.cpp#L848).

This has been tested on real hardware and was discovered during testing of the new Mission Planner Gimbal Control interface (see https://github.com/ArduPilot/MissionPlanner/pull/3386)


peterbarker (approved):

rmackay9:
Thanks very much!

rmackay9:

PR #29003, opened 2025-01-04
AP_HAL_ESP32: check UART thread ownership

srmainwaring:
Add checks that UART driver functions are called on the thread the port is initialised on, and handle a change of thread ownership.

### Motivation

- Fix an issue where calling `UARTDriver::_begin` with a baudrate of zero (for example when changing the thread owning the UART while initialising the AP_MSP_Telem_Backend) causes a core panic due to an integer division by zero exception.
- The implementation follows the equivalent AP_HAL_ChibiOS UART driver. 

### Testing

- Configure, build and flash `esp32s3empty`.
- Set `SERIAL3_BAUD 115`.
- Set `SERIAL3_PROTOCOL 32`.
- Without this change there is a core panic on the MSP thread.
- With this change a MSP GPS such as the Mateksys M8Q will work (set GPS1_TYPE = 19).

### DIFF
--- a/libraries/AP_HAL_ESP32/UARTDriver.cpp
+++ b/libraries/AP_HAL_ESP32/UARTDriver.cpp
@@ -38,6 +38,12 @@ void UARTDriver::vprintf(const char *fmt, va_list ap)
 
 void UARTDriver::_begin(uint32_t b, uint16_t rxS, uint16_t txS)
 {
+    if (b == 0 && txS == 0 && rxS == 0 && _initialized) {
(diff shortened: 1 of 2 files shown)


tpwrules (approved):
Looks right and was able to test that the crash is fixed though I don't have an MSP GPS.

This is another strange area of the code I'd like to fix, but for now ESP32 should better match ChibiOS.

davidbuzz (commented):
reviewed ok too

PR #29004, opened 2025-01-04
AP_Filesystem: Don't try and rename between backends

IamPete1:
Extract from https://github.com/ArduPilot/ardupilot/pull/28977. Currently only the old name is passed to `backend_by_path`. This function will remove the backed prefix. This means the two file names are no longer comparable. This also disallows trying to rename between two backends.


peterbarker:
The very much looks right.  What are the consequences of someone attempting this rename before this patch?

peterbarker (approved):

IamPete1:
> The very much looks right. What are the consequences of someone attempting this rename before this patch?

None of the existing `@` backends support renaming. So there is no change. It worked on local because there is no change to the path in the `backend_by_path` function.

PR #29005, opened 2025-01-04
AP_MultiHeap: initialize only if heap allocation succeeded

srmainwaring:
Ensure new_heap is not null before accessing, and initialise current heap usage.

### Motivation

- On ESP32 S3 the heap allocator is failing when attempting to create a new Lua state.
- The variable `current_heap_usage` is not initialised to zero.

### DIFF
--- a/Tools/ardupilotwaf/boards.py
+++ b/Tools/ardupilotwaf/boards.py
@@ -1390,6 +1390,7 @@ class linux(Board):
         ]
 
         # wrap malloc to ensure memory is zeroed
+        # note that this also needs to be done in the CMakeLists.txt files
         env.LINKFLAGS += ['-Wl,--wrap,malloc']
 
         if cfg.options.force_32bit:
--- a/libraries/AP_HAL_ESP32/targets/esp32/esp-idf/CMakeLists.txt
+++ b/libraries/AP_HAL_ESP32/targets/esp32/esp-idf/CMakeLists.txt
@@ -121,6 +121,9 @@ target_link_libraries(${elf_file}
 # linker script generation, partition_table generation, etc.
 idf_build_executable(${elf_file})
 
+# wrap malloc to ensure memory is zeroed
+target_link_options(${elf_file} PRIVATE "-Wl,--wrap,malloc")
+
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 
 # Additional targets for measuring RAM use: size, size-components, size-files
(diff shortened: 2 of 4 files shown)


tpwrules:
What exactly is failing? That malloc call to initialize the heap?

Moving the magic init is correct (poor review on my part, whoops) and probably needs a backport.

But Theoretically™ malloc returns zero-initialized memory on ArduPilot so the clear of the used size is not necessary. Have you seen that be non-zero?

srmainwaring:
> But Theoretically™ malloc returns zero-initialized memory on ArduPilot so the clear of the used size is not necessary. Have you seen that be non-zero?

Yes. The non-zero is the reason the allocator is failing on ESP32. The first pass through the allocator it fails at https://github.com/ArduPilot/ardupilot/blob/5f8a655f35a2a7901a56e898609a2d7c488a99dd/libraries/AP_MultiHeap/MultiHeap_malloc.cpp#L80

tpwrules:
The failure matches my experience, and it makes sense that the heap malloc itself is not failing because then it should crash with a null pointer dereference.

But that's bad, it means the wrapping is not working. Let me see if there's an obvious cause.

srmainwaring:
I don't see anything in https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Common/c%2B%2B.cpp that would wrap malloc for ESP32 to ensure memory is initialised to zero. Where is that meant to occur for ESP32?

tpwrules:
It is supposed to be done here but I suspect that doesn't get passed through to the linker because that is done by ESP-IDF: https://github.com/ArduPilot/ardupilot/blob/5f8a655f35a2a7901a56e898609a2d7c488a99dd/Tools/ardupilotwaf/boards.py#L1064

Adding `target_link_options(${elf_file} PRIVATE "-Wl,--wrap,malloc")` after the `idf_build_executable` in CMakeLists.txt fixes it, or at least I can see `__wrap_malloc` in the binary now.

We really need to convince the team to switch to a different function name.

tpwrules:
I have adjusted to include this fix, please test that it works properly.

srmainwaring:
> I have adjusted to include this fix, please test that it works properly.

Thanks - looks good. Re-tested a `hello world` script on ESP32 S3.

esp32_scripting_v2

davidbuzz:
this PR should really have had the prefix AP_HAL_ESP32 now that its been determined to be a missing --wrap or malloc in the CMake linker, but its a great find, and brings the esp32 builds closer to how chibios does its pre-zero-ing so its a good-thing.

PR #29007, opened 2025-01-05
Compilation fixes when features disabled / feature extraction fixes

peterbarker:
- several missing dependencies
 - kludge to make AP_CAN_LOGGING_ENABLED work removed as not necessary with corrected includes
 - add missing #ifs for harmonic notch
 - several build_options.py features are only applicable on Plane or Copter, so add them into the `test_build_options.py` exemptions list (which should be moved into `build_options.py`
 - remove AP_NETWORKING_CAN_MCAST_ENABLED from build server options as dependent defines are not present (most notably `AP_NETWORKING_ENABLED`.


rmackay9:
Thanks for this.  I think the Copter changes are probably best answered by @andyp1per (who I see you've already tagged)

ArduCopter/Copter.cpp
```diff
@@ -260,7 +260,7 @@ const AP_Scheduler::Task Copter::scheduler_tasks[] = {
 #if HAL_BUTTON_ENABLED
     SCHED_TASK_CLASS(AP_Button,            &copter.button,              update,           5, 100, 168),
 #endif
-#if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED
+#if AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED && AP_INERTIALSENSOR_HARMONICNOTCH_ENABLED
```
andyp1per, line 263: Not necessary, this is already covered by the define of AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED

ArduCopter/rate_thread.cpp
```diff
@@ -407,10 +407,12 @@ void Copter::rate_controller_thread()
 */
 void Copter::rate_controller_filter_update()
 {
+#if AP_INERTIALSENSOR_HARMONICNOTCH_ENABLED
```
andyp1per, line 410: ditto

ArduCopter/rate_thread.cpp
```diff
@@ -477,6 +479,7 @@ void Copter::rate_controller_log_update()
 #endif
 }
 
+#if AP_INERTIALSENSOR_HARMONICNOTCH_ENABLED
```
andyp1per, line 482: ditto

andyp1per (changes requested):
AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED should never be built without harmonic notch support, so we should express the dependency a different way if necessary rather than littering the code with extra defines

peterbarker:
> AP_INERTIALSENSOR_FAST_SAMPLE_WINDOW_ENABLED should never be built without harmonic notch support, so we should express the dependency a different way if necessary rather than littering the code with extra defines

I was indeed wondering if that was the case.  I've added the dependency into `build_options.py` and removed the patches from the Copter directory.

tridge (approved):

PR #29009, opened 2025-01-05
AP_DDS: configuration fixes

srmainwaring:
Miscellaneous fixes to ensure that the various enable options in `AP_DDS_config.h` work as intended.

### Motivation

-  Building for ESP32 S3 with minimal capabilities (i.e. only topics `/ap/clock` and `ap/time` enabled) is failing due to indexing errors in to the topic table and unused variables.

### Details

- Replace use of #ifdef with #if for `*_ENABLED` options.
- Exclude quaternion helper function when not required.
- Additional exclude option for HAL_BOARD_ESP32.
- Resolve uninitialised variable error observed when compiling with the esp-idf toolchain.

### Testing

- Standard SITL build and test ok.
- Excluding all topics and services apart from `/ap/clock` and `/ap/time` in an `esp32s3empty` build compiles correctly.
- Note: DDS is not currently running on ESP32 due to other issues (unrelated to this PR).

_Extract from esp32s3empty.h configured for a minimal DDS build_
```c++
// DDS
#define AP_DDS_VISUALODOM_ENABLED 0

#define AP_DDS_EXPERIMENTAL_ENABLED 0

#define AP_DDS_AIRSPEED_PUB_ENABLED 0
#define AP_DDS_CLOCK_PUB_ENABLED 1
#define AP_DDS_BATTERY_STATE_PUB_ENABLED 0
#define AP_DDS_GEOPOSE_PUB_ENABLED 0
#define AP_DDS_GLOBAL_POS_CTRL_ENABLED 0
#define AP_DDS_GOAL_PUB_ENABLED 0
#define AP_DDS_GPS_GLOBAL_ORIGIN_PUB_ENABLED 0
#define AP_DDS_IMU_PUB_ENABLED 0
#define AP_DDS_LOCAL_POSE_PUB_ENABLED 0
#define AP_DDS_LOCAL_VEL_PUB_ENABLED 0
#define AP_DDS_NAVSATFIX_PUB_ENABLED 0
#define AP_DDS_STATUS_PUB_ENABLED 0
#define AP_DDS_STATIC_TF_PUB_ENABLED 0
#define AP_DDS_TIME_PUB_ENABLED 1
#define AP_DDS_VEL_CTRL_ENABLED 0

#define AP_DDS_JOY_SUB_ENABLED 0
#define AP_DDS_DYNAMIC_TF_SUB_ENABLED 0

#define AP_DDS_ARM_SERVER_ENABLED 0
[...]

PR #29010, opened 2025-01-05
AP_HAL_ESP32: ensure task is deleted if member proc returns

srmainwaring:
Resolve issue where an illegal instruction error occurs if a task member function returns.

- FreeRTOS task functions must not return or exit while the task is active.
- vTaskDelete(NULL) deletes the calling task.

See: 
- https://www.freertos.org/Documentation/02-Kernel/04-API-references/01-Task-creation/01-xTaskCreate
- https://www.freertos.org/Documentation/02-Kernel/04-API-references/01-Task-creation/03-vTaskDelete

### Motivation

To replicate the issue:
- Requires https://github.com/ArduPilot/ardupilot/pull/29009.
- Configure `esp32s3empty` with the `--enable-dds` flag, build and flash.
- If the DDS transport initialisation fails, or the `AP_DDS_Client::main_loop` otherwise returns, an illegal instruction occurs.

### DIFF
--- a/libraries/AP_HAL_ESP32/Scheduler.cpp
+++ b/libraries/AP_HAL_ESP32/Scheduler.cpp
@@ -140,6 +140,9 @@ void IRAM_ATTR Scheduler::thread_create_trampoline(void *ctx)
     AP_HAL::MemberProc *t = (AP_HAL::MemberProc *)ctx;
     (*t)();
     free(t);
+
+    // delete the calling task
+    vTaskDelete(NULL);
 }
 
 /*


tpwrules (approved):
Correct according to the docs, tested that it crashes before and doesn't crash now. Thanks for the replication instructions.

It looks like all our other uses of these task functions have an infinite loop and don't return so they don't need adjustment.

tpwrules:
Sorry, minor github stupidity...

davidbuzz (commented):
reviewed

PR #29011, opened 2025-01-05
Plane: pass throttle slew limit to motors `output_motor_mask` function. 

IamPete1:
This is a alternate way to implement slew limiting of tilt rotor motors in forward flight ( https://github.com/ArduPilot/ardupilot/pull/28960).

This passes the slew limit down to the motor mask function which applies both it and the underlying motors slew rate. Unlike https://github.com/ArduPilot/ardupilot/pull/28960 this results in a per motor slew. This means the slew is included for yaw differential thrust, this is inline with what is done with the left and right throttle outputs in plane. The outputs are also slewed from there position in VTOL flight rather than having a jump.

### DIFF
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -1180,6 +1180,7 @@ private:
     void servos_twin_engine_mix();
     void force_flare();
     void throttle_watt_limiter(int8_t &min_throttle, int8_t &max_throttle);
+    float get_throttle_slewrate() const;
     void throttle_slew_limit();
     bool suppress_throttle(void);
     void update_throttle_hover();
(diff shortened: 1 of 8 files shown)


tridge (commented):
@IamPete1 I am concerned that may impact stability on tilt-vtols when in primarily VTOL flight. Existing tunes won't have factored in this slew limit.

IamPete1:
Rebased. 

> hat may impact stability on tilt-vtols when in primarily VTOL flight. Existing tunes won't have factored in this slew limit.

This will only affect forward flight when the `output_motor_mask` method is used. VTOL flight is unaffected.

PR #29012, opened 2025-01-05
AP_Arming: Allow arming when forcing DCM and no GPS configured

Ryanf55:
# Purpose

If I do not have a GPS and don't use "Set Home Here” in mission planner because I'm just flying FPV without a GCS, I want to be able to keep all arming checks enabled. This is for running AP as a simple stabilizer (FBWA) with minimal peripherals. 

I never expect a home location, and do NOT expect AUTO, RTL, and other GPS-reliant modes to be allowed. 
Seems like having a home location is merely a convenience for the GCS, and I shouldn't rely on global position or local position if I only fly in FBWA and ACRO, so why do we make it required?

# Approach 

This PR introduces mode transition rejection into Plane based on needing a home. I have re-used the mode rejection from Copter, modeled after `ArduCopter/Mode::requires_GPS()`. Mode changes in plane can now be rejected, but only while armed.
While disarmed, like Copter, you can change to any mode, but arming checks in Plane now check whether your mode requires a home. If home is not set, you can't arm.

Note - without changes to the EKF, I recommend using DCM to fly without any position aiding. It won't fly as well.

# Wiki

This change will allow this wiki to be complete: https://github.com/ArduPilot/ardupilot_wiki/pull/6508

# Details

* If we force AP to use DCM, and disable all GPS's, we never expect a home location. Currently, AP will prevent arming.
* Allow users to arm DCM with all pre-arm checks enabled in this special case
* This allows using AP as a simple stabilizer (FBWA) on plane

# Testing performed

This sets up a no-gps plane that is locked to DCM, and then demonstrates we can now arm.

```bash
./Tools/autotest/sim_vehicle.py -v Plane --console --map -D -w
param set STALL_PREVENTION 0
param set GPS1_TYPE 0
param set AHRS_EKF_TYPE 0
# Enable the new no-home-required for arming
param bitmask set AHRS_OPTIONS 3
map set showsimpos 1
reboot
# Move the map to follow the sim position.
mode fbwa
# Wait for "pre-arm good"
arm throttle
rc 2 2000
rc 3 1500
```

image

If you cause an RC failsafe, then you see it will stay in circle, instead of have a flyaway in RTL. This is because the short failsafe is circle, and the long failsafe is RTL. I assert this is an improvement to not fly to 0, 0. 

```
rc 3 900
```
image

# To discuss

* If we do **NOT** have a GPS, but we **DO HAVE** visual odom enabled, but we don't have a home, should we be allowed to arm? I suspect no currently -> Correct
* If we are using EKF3, and **DO NOT** have any GPS's configured, and no other position sources, should we require a home position to arm?  I suspect yes. -> Unless the user enables the no home required option. 
* Do we want to allow plane users to arm without a home, and add protection in mode change to block change from FBWA to AUTO if there's no home? -> Yes, added.

# Follow up

Improve mode transitions to cope with some modes not having a home, such as RTL long action should go to GLIDE rather than stay in circle.

### DIFF
--- a/ArduPlane/mode.h
+++ b/ArduPlane/mode.h
@@ -110,6 +110,9 @@ public:
     // true if mode can have terrain following disabled by switch
     virtual bool allows_terrain_disable() const { return false; }
 
+    // true if mode requires the AHRS to have a home before using this mode.
+    virtual bool requires_home() const { return true; };
+
     // true if automatic switch to thermal mode is supported.
     virtual bool does_automatic_thermal_switch() const {return false; }
 
@@ -204,6 +207,8 @@ public:
     void stabilize();
 
     void stabilize_quaternion();
+
+    bool requires_home() const override { return false; };
 
 protected:
 
@@ -327,6 +332,10 @@ public:
 
     bool does_auto_throttle() const override { return true; }
 
+    // Once offboard mode exists, only global position control modes require home.
+    // Low level velocity and acceleration controls don't require home.
+    bool requires_home() const override { return true; };
+
     // handle a guided target request from GCS
     bool handle_guided_request(Location target_loc) override;
 
@@ -360,6 +369,8 @@ public:
     bool does_auto_navigation() const override { return true; }
 
     bool does_auto_throttle() const override { return true; }
+
+    bool requires_home() const override { return false; };
 
 protected:
 
@@ -442,6 +453,8 @@ public:
     // true if voltage correction should be applied to throttle
     bool use_battery_compensation() const override { return false; }
 
+    bool requires_home() const override { return false; };
+
 };
 
 
@@ -488,6 +501,8 @@ public:
 
     void run() override;
 
+    bool requires_home() const override { return false; };
+
 private:
     void stabilize_stick_mixing_direct();
 
@@ -500,6 +515,8 @@ public:
     Number mode_number() const override { return Number::TRAINING; }
     const char *name() const override { return "TRAINING"; }
     const char *name4() const override { return "TRAN"; }
+
+    bool requires_home() const override { return false; };
 
     // methods that affect movement of the vehicle in this mode
     void update() override;
@@ -525,6 +542,8 @@ public:
 
     bool does_auto_throttle() const override { return true; }
 
+    bool requires_home() const override { return false; };
+
 protected:
     bool _pre_arm_checks(size_t buflen, char *buffer) const override { return false; }
 
@@ -542,6 +561,8 @@ public:
     void update() override;
     
     bool mode_allows_autotuning() const override { return true; }
+
+    bool requires_home() const override { return false; };
 
     void run() override;
 
@@ -565,6 +586,8 @@ public:
     bool does_auto_throttle() const override { return true; }
     
     bool mode_allows_autotuning() const override { return true; }
+
+    bool requires_home() const override { return false; };
 
     void update_target_altitude() override {};
 
@@ -780,6 +803,7 @@ public:
     bool is_vtol_mode() const override { return true; }
     bool is_vtol_man_throttle() const override { return true; }
     virtual bool is_vtol_man_mode() const override { return true; }
+    bool requires_home() const override { return false; };
 
     // methods that affect movement of the vehicle in this mode
     void update() override;
@@ -836,6 +860,8 @@ public:
     bool does_auto_navigation() const override { return true; }
 
     bool does_auto_throttle() const override { return true; }
+
+    bool requires_home() const override { return false; };
 
     // var_info for holding parameter information
     static const struct AP_Param::GroupInfo var_info[];
--- a/ArduPlane/system.cpp
+++ b/ArduPlane/system.cpp
@@ -286,6 +286,16 @@ bool Plane::set_mode(Mode &new_mode, const ModeReason reason)
         return false;
     }
 
+    // Block mode changes to modes that require home if we don't have a home, we need a home, and we're armed.
+    // Mode changes without a home, while disarmed, are always ok.
+    if (new_mode.requires_home() &&
+        !ahrs.home_is_set() && 
+        hal.util->get_soft_armed()
+        ) {
+        mode_change_failed(&new_mode, "requires home");
+        return false;
+    }
+
     // backup current control_mode and previous_mode
     Mode &old_previous_mode = *previous_mode;
     Mode &old_mode = *control_mode;
--- a/libraries/AP_Arming/AP_Arming.cpp
+++ b/libraries/AP_Arming/AP_Arming.cpp
@@ -636,7 +636,8 @@ bool AP_Arming::gps_checks(bool report)
             }
         }
 
-        if (!AP::ahrs().home_is_set()) {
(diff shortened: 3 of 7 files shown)


IamPete1:
I think I would argue that having home is so fundamental to the way AP works that trying to fly without it is going to cause lots of unusual issues. All of planes altitude handling is relative to home for example.

If we did want to make the decision to allow flying without home we would need to update all the modes than need it to refuse entry. So you would not even be allowed to enter RTL for example.

Ryanf55:
> I think I would argue that having home is so fundamental to the way AP works that trying to fly without it is going to cause lots of unusual issues. All of planes altitude handling is relative to home for example.
> 
> If we did want to make the decision to allow flying without home we would need to update all the modes than need it to refuse entry. So you would not even be allowed to enter RTL for example.

Yea, if we could refuse entry, it would prevent the flyaway risk if we allow our users to fly without home/gps. I would be happy to make that change if the other devs are ok allow flying in some modes without home.

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -636,7 +636,13 @@ bool AP_Arming::gps_checks(bool report)
             }
         }
 
-        if (!AP::ahrs().home_is_set()) {
+        // Only check we have an AHRS home when we have at least 1 GPS enabled.
```
tridge, line 639: we'd need a FLIGHT_OPTIONS bit, or a AHRS_OPTIONS bit

tridge (changes requested):
you'd also need some changes in Plane::calc_speed_scaler() and some failsafe checking
I also think you will find situations where the attitude goes badly off, really does need an airspeed estimate based on drag+throttle

libraries/AP_Arming/AP_Arming.cpp
Ryanf55: I've added an AHRS options bit to allow arming without a home.

ArduPlane/system.cpp
```diff
@@ -286,6 +286,12 @@ bool Plane::set_mode(Mode &new_mode, const ModeReason reason)
         return false;
     }
 
+    if (new_mode.requires_home() &&
+        !ahrs.home_is_set()) {
```
Ryanf55, line 290: Do I need to use `quadplane.ahrs` in the condition, or is this sufficient?

Ryanf55:
I've rescoped just to cover no-home (instead of no-gps no-airspeed) to make it a smaller scope change. inverse TECS and improving no-airspeed no-gps can come later. This now includes mode change rejection. Additionally, you can now arm without home set in modes that don't require home if you set AHRS_OPTIONS bit 3.

Because previous ardupilot required home to be set in all modes, this is a NFC for any current user.

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -636,7 +636,8 @@ bool AP_Arming::gps_checks(bool report)
             }
         }
 
-        if (!AP::ahrs().home_is_set()) {
+        // Plane uses an option to allow users to fly without a home.
+        if (!AP::ahrs().allows_arming_without_home() && !AP::ahrs().home_is_set()) {
```
Ryanf55, line 640: Perhaps we should move this check to the vehicle. On plane, I would like should check whether our current mode requires home. If it does, and we don't have a home yet, arming should be blocked.

Ryanf55: image

As implemented, you can switch to LOITER while disarmed, but won't be allowed to arm till you set home.

ArduPlane/mode.h
```diff
@@ -395,6 +410,8 @@ class ModeLoiter : public Mode
     
     bool mode_allows_autotuning() const override { return true; }
 
+    bool requires_home() const override { return true; };
```
peterbarker, line 413: Either make the method pure-virtual or rely on a default value from the base class.

ArduPlane/mode.h
```diff
@@ -837,6 +879,8 @@ class ModeTakeoff: public Mode
 
     bool does_auto_throttle() const override { return true; }
 
+    bool requires_home() const override { return true; };
```
peterbarker, line 882: Why does TAKEOFF mode need home? This is an example question applicable to other modes you've marked as "true".

libraries/AP_Arming/AP_Arming.cpp
peterbarker: >     * AHRS home is the RTL point, position estimate is from EKF origin. It should be safe to treat having a home and having an ekf origin as the same.

... or, you know, you could check to see if you have a navigation origin instead....

If your failsafes include "RTL", then you might go further and require home to be set then.

ArduPlane/mode.h
Ryanf55: Done! I use the default from the base class, gps is required.

ArduPlane/mode.h
Ryanf55: TAKEOFF surprised me. When I tested it, the plane did not take off without home. It just sits armed on the runway drawing 0A. 

This is because takeoff mode relies on having home here in the `update()` call: https://github.com/ArduPilot/ardupilot/blob/20c77ae762951fe4410959b94cfe22f44bf07583/ArduPlane/mode_takeoff.cpp#L74

It would be nice if Takeoff could just climb to the altitude using baro alt above arm height, but that doesn't work yet.

I need to test each of the Q* modes; they are all biased to requiring home right now till I confirm they are safe without it.

libraries/AP_Arming/AP_Arming.cpp
Ryanf55: Are you saying we might want to add two methods to the mode class to prevent mode entry:
1. The one I currently have of `requires_home`, which uses `AP_AHRS::home_is_set()`
1. Another of `requires_navigation_origin`, which uses `AP_AHRS::get_origin()`

I agree that some of the modes only need the navigation origin, and I'm happy to add that to that check to mode entry if we are ok with the flash.

Ryanf55:
> calc_speed_scaler

I attempted an AP_Airspeed backend for synthetic airspeed. For the throttle effect on airspeed, I wanted to use the plane parameters `throttle_cruise`, `airspeed_min` and `airspeed_max`, but I can't get the `Plane`  parameters in the airspeed sensor class. I don't see any reasonable code structure to make it an airspeed backend

Branch: https://github.com/Ryanf55/ardupilot/tree/airspeed-synthetic

ArduPlane/mode.h
peterbarker: OK, bad example; TAKEOFF is home-relative, so that one makes sense :-)

Ryanf55:
The dev team is aware of too many footguns flying a FPV plane without global position in MANUAL and FBWA.

* If we block RTL entry, you're hosed 
* If you take off in MANUAL without GPS, and the GPS gets home over a lake while flying, you're hosed if it goes into RTL

We could add more arming checks; we can try to protect against some of them, but it's very risky we'll miss one and someone is going to lose a vehicle over it. The team wants to prioritize safety for the general user over adding potentially dangerous capabilities.

PR #29013, opened 2025-01-05
Tools: fix Cygwin CI build

tpwrules:
There are currently issues where the non-.exe-suffixed files can't be copied into the `artifacts` folder; `cp` claims "File exists". Previously this worked but the suffix was added by Cygwin so all files in `artifacts` had a `.exe` suffix anyway.

This is evidently intended, though non-intuitive, behavior: https://sourceware.org/legacy-ml/cygwin/2009-08/msg00293.html

> On Cygwin, you should avoid having a file "foo" and a file "foo.exe" in the same directory at all cost to avoid puzzeling POSIX borderline behaviour like this.  What you do is essentially in the "not supported" class of problems.

> [...] Cygwin does not check for a file "foo", if the name of the file is explicitely given as "foo.exe".

Apparently something similar was addressed in PR #20926; the current code installs files with both suffixes, but that fix contradicts the info above and now has broken.

This PR changes the code to only install .exe-suffixed files, as opposed to only non-.exe-suffixed files, which was the behavior before that PR. The set of installed files should not actually change.

### DIFF
--- a/.github/workflows/cygwin_build.yml
+++ b/.github/workflows/cygwin_build.yml
@@ -156,7 +156,7 @@ jobs:
           NAME_DASHED=${WORKFLOWNAME//+( )/_}
           echo "cache-key=${NAME_DASHED}" >> $GITHUB_OUTPUT
 
-      - uses: cygwin/cygwin-install-action@master
+      - uses: cygwin/cygwin-install-action@v5
         with:
           packages: cygwin64 gcc-g++=10.2.0-1 ccache python37 python37-future python37-lxml python37-pip python37-setuptools python37-wheel git procps gettext
           add-to-path: false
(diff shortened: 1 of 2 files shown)


robertlong13 (approved):
Yup, good find on that mailing list.

tpwrules:
Looks like the root cause for the change in observed behavior was us using the master branch of the Cygwin install action, which had a bug for a little while that caused it to install the testing versions of packages by default. I've added another commit to use the latest released version (which includes the fix for that bug) instead.

tridge:
cygwin CI populates this: https://firmware.ardupilot.org/Tools/MissionPlanner/sitl/

tridge (approved):

PR #29014, opened 2025-01-05
Dds goal pub scripting dependency

Ryanf55:
https://github.com/ArduPilot/ardupilot/pull/29009#discussion_r1903355890

Tests Performed:
```
./waf configure --board sitl --enable-dds --disable-scripting
nice ./waf plane
```

A bit back on `master`, specifically on commit 87b0de79dfacc418ff44e4dd6afe788cfddbac65, if you tried that, you got:
image

Now, it compiles as expected.

### DIFF
--- a/libraries/AP_Vehicle/AP_Vehicle.h
+++ b/libraries/AP_Vehicle/AP_Vehicle.h
@@ -176,6 +176,8 @@ public:
     virtual bool start_takeoff(const float alt) { return false; }
     // Method to control vehicle position for use by external control
     virtual bool set_target_location(const Location& target_loc) { return false; }
+    // Get target location for use by external control
+    virtual bool get_target_location(Location& target_loc) { return false; }
 #endif // AP_SCRIPTING_ENABLED || AP_EXTERNAL_CONTROL_ENABLED
 #if AP_SCRIPTING_ENABLED
     /*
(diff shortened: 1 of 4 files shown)


libraries/AP_Vehicle/AP_Vehicle.h
```diff
@@ -176,6 +176,8 @@ class AP_Vehicle : public AP_HAL::HAL::Callbacks {
     virtual bool start_takeoff(const float alt) { return false; }
     // Method to control vehicle position for use by external control
     virtual bool set_target_location(const Location& target_loc) { return false; }
+    // Get target location for use by external control
+    virtual bool get_target_location(Location& target_loc) { return false; }
```
Ryanf55, line 180: Every function called by DDS that was only enabled by scripting should always be moved into this block.
Otherwise, if you disable scripting and enable DDS, you can't compile.

peterbarker:
I was expecting to see a patch to `AP_DDS_config.h` such that `AP_DDS_GOAL_PUB_ENABLED` defaults to "off" if `!AP_SCRIPTING_ENABLED`.

Without that patch this is not NFC.  Is that intentional?

Ryanf55:
> to see a patch t

The PR today was not following our previous patterns. Upon further inspection, I realized this is the proper patch IF we want to follow the current patterns.

DDS does not depend on scripting. They should always be independent, and AP should compile with neither enabled, with either enabled, or both enabled. 

I do not think making DDS depend on scripting is a good idea.

srmainwaring (changes requested):
Changes also required to `AP_DDS_Client.h` and `AP_DDS_Topic_Table.h`.

See: https://github.com/srmainwaring/ardupilot/tree/dds-goal-pub-scripting-dependency.

(tried force pushing changes but permission denied).

Looks ok with the additional change. Tested build and execution on `esp32empty` with DDS enabled.

srmainwaring (approved):
LGTM

peterbarker:
Merged, thanks!

PR #29020, opened 2025-01-08
Add support for 32 motors, including dodecahexa-octa (two-layers of 8 motors each)

peterbarker:
This PR is a mashup of two other PRs which conflicted with one another.

tridge's PR (https://github.com/ArduPilot/ardupilot/pull/27598) added 32-rotor support and a half-baked "m32" SITL frame to test it.  I've completed the work on the 32-motor demo frame.

@muramura 's PR (https://github.com/ArduPilot/ardupilot/pull/21157) added support for the 16-motor decahexa frame.  It also modified ArduPilot to support 16 motors, which was a subset of tridge's 32-motor support, 'though remembered to patch AP_Periph to match the rest of ArduPilot.

I've reworked both sets of patches significantly to create this combined set, notably trying to bring consistency to the naming of files and frames.

Replaces https://github.com/ArduPilot/ardupilot/pull/21157
Replaces https://github.com/ArduPilot/ardupilot/pull/27598

Both new frames are flown in our "FlyEachFrame" test successfully.


peterbarker:
I've removed a patch which did this:
```
@@ -355,7 +355,7 @@ SRV_Channels::SRV_Channels(void)
 #if NUM_SERVO_CHANNELS > 16
         if (i >= 16) {
             // default to GPIO, this disables the pin and stops logging
-            channels[i].function.set_default(SRV_Channel::k_GPIO);
+            channels[i].function.set_default(SRV_Channel::k_none);
         }
 #endif
     }

```
.... it was killing various autotests.

amilcarlucas:
```
arducopter-heli  -268 (-0.0147%)  0 (0.0000%)  4 (+0.0015%)   -268 (-0.0147%)                                  138324
antennatracker   124 (+0.0092%)   0 (0.0000%)  -4 (-0.0015%)  124 (+0.0092%)                                   615828
arducopter       216 (+0.0118%)   0 (0.0000%)  0 (0.0000%)    216 (+0.0118%)                                   136212
ardurover        184 (+0.0110%)   0 (0.0000%)  0 (0.0000%)    184 (+0.0110%)                                   292980
arduplane        196 (+0.0108%)   0 (0.0000%)  4 (+0.0015%)   196 (+0.0108%)                                   150148
blimp            124 (+0.0090%)   0 (0.0000%)  -4 (-0.0015%)  124 (+0.0090%)                                   593480
ardusub          516 (+0.0320%)   0 (0.0000%)  -4 (-0.0015%)  516 (+0.0319%)                                   348512
```

peterbarker:
```
CubeOrange-periph-heavy  24                                                                    
CubeRedPrimary                      40     *           88      40                48     56     216
Durandal                            120    *           232     -272              216    200    536
Hitec-Airspeed           *                                                                     
KakuteH7-bdshot                     16     *           64      16                24     32     120
MatekF405                           16     *           48      40                48     32     240
Pixhawk1-1M-bdshot                  16                 48      48                48     40     304
bebop                               4096               -12400  0                 -8336  0      -16576
f103-QiotekPeriph        *                                                                     
f303-Universal           *                                                                     
revo-mini                           16     *           48      48                56     40     104
skyviper-v2450                                         40                                      
```

libraries/SRV_Channel/SRV_Channel_aux.cpp
```diff
@@ -544,7 +544,8 @@ bool SRV_Channels::set_aux_channel_default(SRV_Channel::Aux_servo_function_t fun
         // already assigned
         return true;
     }
-    if (channels[channel].function != SRV_Channel::k_none) {
+    if (channels[channel].function != SRV_Channel::k_none &&
+        !(channel >15 && channels[channel].function == SRV_Channel::k_GPIO)) {
```
IamPete1, line 548: This is because if this I guess.

https://github.com/ArduPilot/ardupilot/blob/0c9aaa2ae6b9b70fbf132f733c51bd90d903d56b/libraries/SRV_Channel/SRV_Channels.cpp#L355-L360

Which is in turn because of this:

https://github.com/ArduPilot/ardupilot/blob/0c9aaa2ae6b9b70fbf132f733c51bd90d903d56b/libraries/AP_Logger/LogFile.cpp#L189

Its a bit odd that we have to log the None output type, we really should move them over to a dedicated "servo events function" so none really means disabled.

IamPete1 (dismissed):
The output motor mask stuff in plane is passing uint16's. This is the motors function:

https://github.com/ArduPilot/ardupilot/blob/0c9aaa2ae6b9b70fbf132f733c51bd90d903d56b/libraries/AP_Motors/AP_MotorsMulticopter.cpp#L743C28-L743C45

Unfortunately both tilt rotor and tailsitter store the mask in a AP_Int16 param. 

I can't find any other uint16 issues. I would also be happy if only went to 16 motors on plane and put in a `#error` for the time being and only allowed the full 32 on copter.

@andyp1per Do you foresee any issues on for harmonic notched trying to track 32 motors? I guess the existing maximum number of notches check should be fine.

andyp1per:
> @andyp1per Do you foresee any issues on for harmonic notched trying to track 32 motors? I guess the existing maximum number of notches check should be fine.

Depends on the use of harmonics. I added harmonics to per-motor and its really very useful. I can't see 32 motors with 4 harmonics working well. We probably need a way to limit the number of motors that get tracked.

The other issue is getting data from 32 motors - will be quite a lot of bandwidth if it works at all. I guess you have to assume that not all channels will support telemetry in which case I am not sure what would happen.

peterbarker:
> The output motor mask stuff in plane is passing uint16's. This is the motors function:
> 
> https://github.com/ArduPilot/ardupilot/blob/0c9aaa2ae6b9b70fbf132f733c51bd90d903d56b/libraries/AP_Motors/AP_MotorsMulticopter.cpp#L743C28-L743C45

I've created https://github.com/ArduPilot/ardupilot/pull/29084 to help migrate that mask to at least 32 bits.

> Unfortunately both tilt rotor and tailsitter store the mask in a AP_Int16 param.
> 
> I can't find any other uint16 issues. I would also be happy if only went to 16 motors on plane and put in a `#error` for the time being and only allowed the full 32 on copter.

Not really sure how that would be nicely done.... I could add a runtime check easily enough, I think...

peterbarker:
> The output motor mask stuff in plane is passing uint16's. This is the motors function:

I've now rebased this on top of a pair of commits which make this mask 32-bits wide

IamPete1:
I think the changes now look good, its just a matter of deciding what `AP_MOTORS_MAX_NUM_MOTORS` is by default on each vehicle type.

On sub we have a direction parameter per motor, so supporting more than 12 is a little odd:

https://github.com/ArduPilot/ardupilot/blob/372a1a6e6ec02f617415550f074678408b416467/libraries/AP_Motors/AP_Motors6DOF.cpp#L119

I think the extra flash cost is just that sub has more arrays with size number of motors than copter.

If scripting motor backed is not enabled, since it it is the only way of adding more than 12 motors I think we should set the define to 12.

There is actually a define for each frame type on multirotor, so we could go down from 12 to 4 if only quad is enabled for example. There is a slight trap in that its not the number of motors but the highest number motor that is used, so it would be 4 on tricopter not 3 (we could fix this).

peterbarker:
> I think the changes now look good, its just a matter of deciding what `AP_MOTORS_MAX_NUM_MOTORS` is by default on each vehicle type.

Not sure how you're suggesting we change that on a per-vehicle basis... it's defined in AP_Motors_Class.h, which is definitely not going to be on the vehicle-specific-stuff whitelist!

IamPete1:
> Not sure how you're suggesting we change that on a per-vehicle basis... it's defined in AP_Motors_Class.h, which is definitely not going to be on the vehicle-specific-stuff whitelist!

If not vehicle type then at least based on scripting. No point in have 20 motors that are impossible to use.

peterbarker:
> > Not sure how you're suggesting we change that on a per-vehicle basis... it's defined in AP_Motors_Class.h, which is definitely not going to be on the vehicle-specific-stuff whitelist!
> 
> If not vehicle type then at least based on scripting. No point in have 20 motors that are impossible to use.

Seems reasonable.  Pushed a patch.

IamPete1 (approved):
LGTM

amilcarlucas:
```
arducopter-heli  -248 (-0.0136%)  0 (0.0000%)  0 (0.0000%)    -248 (-0.0136%)                                  137340
antennatracker   124 (+0.0092%)   0 (0.0000%)  -4 (-0.0015%)  124 (+0.0092%)                                   614924
arducopter       248 (+0.0136%)   0 (0.0000%)  0 (0.0000%)    248 (+0.0135%)                                   135220
ardurover        204 (+0.0122%)   0 (0.0000%)  -4 (-0.0015%)  204 (+0.0122%)                                   291996
arduplane        236 (+0.0130%)   0 (0.0000%)  -4 (-0.0015%)  236 (+0.0130%)                                   150836
blimp            124 (+0.0090%)   0 (0.0000%)  -4 (-0.0015%)  124 (+0.0090%)                                   592576
ardusub          544 (+0.0337%)   0 (0.0000%)  0 (0.0000%)    544 (+0.0336%)                                   347560
```

peterbarker:
```
CubeOrange-periph-heavy  24                *
CubeRedPrimary                      40     *           112     40                80     64     232
Durandal                            120    *           248     -272              232    208    536
KakuteH7-bdshot                     16     *           88      16                64     32     136
MatekF405                           16     *           40      48                56     40     40
Pixhawk1-1M-bdshot                  16                 48      48                56     40     40
revo-mini                           16     *           40      40                56     32     40
skyviper-journey                                       48
skyviper-v2450                                         40

real    7m40.615s
user    185m33.436s
sys     24m0.595s
```

libraries/AP_Motors/AP_Motors_Class.h
```diff
@@ -20,8 +20,34 @@
...
+#define AP_MOTORS_MOT_27 26U
+#define AP_MOTORS_MOT_28 27U
+#define AP_MOTORS_MOT_29 28U
+#define AP_MOTORS_MOT_30 29U
+#define AP_MOTORS_MOT_31 30U
+#define AP_MOTORS_MOT_32 31U
+
+#ifndef AP_MOTORS_MAX_NUM_MOTORS
+#if AP_SCRIPTING_ENABLED
+#define AP_MOTORS_MAX_NUM_MOTORS 32
```
andyp1per, line 46: I think this needs to be constrained to NUM_SERVO_CHANNELS. We lower this on certain boards to save space and it constrains the size of arrays in certain circumstances

andyp1per (changes requested):
Think we need to observe NUM_SERVO_CHANNELS

libraries/AP_Motors/AP_Motors_Class.h
peterbarker: I've added another patch to constrain it to NUM_SERVO_CHANNELS - is this what you had in mind?

andyp1per: LGTM

andyp1per (approved):
Seems ok now. Not sure about the flash cost

peterbarker:
> Seems ok now. Not sure about the flash cost

Pasted results for the usual sets of boards in another comment.

Also had to add a patch to set a minimum of 12 motor outputs or various frame classes don't compile and kill the compilation.  This is the case for AP_Motors6DOF on CubeOrange-periph-heavy (which has 6 outputs).

We can come back later and remove this clamping, but it will require a whole bunch of new `#ifdef`s around the motor class code (and where it's used).

PR #29027, opened 2025-01-09
SRV_Channel: build bitmasks thread safe

robertlong13:
Alternative to #29021

This costs a couple hundred bytes, and loops a worst-case of NUM_SERVO_CHANNELS^2 (better than NUM_FUNCTIONS*NUM_SERVO_CHANNELS though, so I'm pretty proud of that).

The advantage is that it doesn't impact the performance of any getters.

### DIFF
--- a/libraries/SRV_Channel/SRV_Channel_aux.cpp
+++ b/libraries/SRV_Channel/SRV_Channel_aux.cpp
@@ -211,23 +211,41 @@ void SRV_Channels::update_aux_servo_function(void)
     if (!channels) {
         return;
     }
-    function_mask.clearall();
 
-    for (uint16_t i = 0; i < SRV_Channel::k_nr_aux_servo_functions; i++) {
-        functions[i].channel_mask = 0;
-    }
-    invalid_mask = 0;
-
-    // set auxiliary ranges
+    // We build up new masks and set them at the end so that they are never
+    // in the middle of rebuilding if another thread tries to read them.
+    static Bitmask<SRV_Channel::k_nr_aux_servo_functions> new_function_mask;
+    new_function_mask.clearall();
+    uint32_t new_invalid_mask = 0;
+    
+    // Build up the new function and invalid masks, and set auxiliary ranges
     for (uint8_t i = 0; i < NUM_SERVO_CHANNELS; i++) {
         if (!channels[i].valid_function()) {
-            invalid_mask |= 1U<<i;
+            new_invalid_mask |= 1U<<i;
             continue;
         }
         const uint16_t function = channels[i].function.get();
         channels[i].aux_servo_function_setup();
-        function_mask.set(function);
-        functions[function].channel_mask |= 1U<<i;
+        new_function_mask.set(function);
+    }
+    function_mask = new_function_mask;
+    invalid_mask = new_invalid_mask;
+
+    // Set the channel masks for each function
+    // This nested loop runs worst case O(NUM_SERVO_CHANNELS^2), but this only
+    // runs once per second.
+    for (uint16_t i = 0; i < SRV_Channel::k_nr_aux_servo_functions; i++) {
+        if (!new_function_mask.get(i)) {
+            functions[i].channel_mask = 0;
+            continue;
+        }
+        SRV_Channel::servo_mask_t new_mask = 0;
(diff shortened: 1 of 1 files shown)


robertlong13:
Did some very-basic profiling on a CubeOrange on the bench. Adds about 40us to the max time, 7us to the minimum time, and generally more spikes (split between min/mean time is larger in new code).

Costs 300-400 bytes on Durandal, and 76 bytes on Pixhawk-1M.

I think that should all be fine.

old-5-servos
old-32-servos
new-5-servos
new-32-servos

andyp1per:
> Adds about 40us to the max time, 7us to the minimum time

These are large numbers. The whole of motor output only takes 50us currently, so that's a 15% increase. With 4k loops there is only 250us to do everything, and typically I am looking to do everything in 80-90us.

andyp1per:
I suppose if the access is unaffected then this should be ok.

libraries/SRV_Channel/SRV_Channel_aux.cpp
```diff
@@ -211,23 +211,41 @@ void SRV_Channels::update_aux_servo_function(void)
...
-    for (uint16_t i = 0; i < SRV_Channel::k_nr_aux_servo_functions; i++) {
-        functions[i].channel_mask = 0;
-    }
-    invalid_mask = 0;
-
-    // set auxiliary ranges
+    // We build up new masks and set them at the end so that they are never
+    // in the middle of rebuilding if another thread tries to read them.
+    static Bitmask<SRV_Channel::k_nr_aux_servo_functions> new_function_mask;
+    new_function_mask.clearall();
```
IamPete1, line 218:
```suggestion
    static Bitmask<SRV_Channel::k_nr_aux_servo_functions> new_function_mask {};
```

IamPete1 (approved):
Looks good.

libraries/SRV_Channel/SRV_Channel_aux.cpp
IamPete1: Ah, I didn't spot the static. In general i'm not a fan of statics in function for that reason. I don't have a feel for which would give better performance.

libraries/SRV_Channel/SRV_Channel_aux.cpp
```diff
@@ -211,23 +211,41 @@ void SRV_Channels::update_aux_servo_function(void)
...
 
-    for (uint16_t i = 0; i < SRV_Channel::k_nr_aux_servo_functions; i++) {
-        functions[i].channel_mask = 0;
-    }
-    invalid_mask = 0;
-
-    // set auxiliary ranges
+    // We build up new masks and set them at the end so that they are never
+    // in the middle of rebuilding if another thread tries to read them.
+    static Bitmask<SRV_Channel::k_nr_aux_servo_functions> new_function_mask;
```
tridge: why static? why not a stack variable?

libraries/SRV_Channel/SRV_Channel_aux.cpp
```diff
@@ -211,23 +211,40 @@ void SRV_Channels::update_aux_servo_function(void)
...
 
-    for (uint16_t i = 0; i < SRV_Channel::k_nr_aux_servo_functions; i++) {
-        functions[i].channel_mask = 0;
-    }
-    invalid_mask = 0;
-
-    // set auxiliary ranges
+    // We build up new masks and set them at the end so that they are never
+    // in the middle of rebuilding if another thread tries to read them.
+    Bitmask<SRV_Channel::k_nr_aux_servo_functions> new_function_mask {};
```
peterbarker, line 217:
```suggestion
    Bitmask<SRV_Channel::k_nr_aux_servo_functions> new_function_mask;
```

tridge:
its tempting to make this update only when armed, idea from @IamPete1

libraries/SRV_Channel/SRV_Channel_aux.cpp
```diff
@@ -211,23 +211,40 @@ void SRV_Channels::update_aux_servo_function(void)
...
+    uint32_t new_invalid_mask = 0;
+
```
peterbarker, line 219:
```suggestion
```

tridge (changes requested):
I think this has to move to being a register_io_process() callback, from AP_Vehicle, the new code is O(N^2), just too expensive for main thread

PR #29028, opened 2025-01-09
AP_ESC_Telem: fix other timeout race

robertlong13:
Prerequisite PR: #28999

The timeout for non-RPM telemetry is vulnerable to a similar race to the RPM one above. This PR makes the timeout logic consistent between the two, adding a validity flag to track the timeout instead of explicitly subtracting timestamps (one of which is volatile).

I broke this PR into a separate dependent PR in case this is considered too risky of a change. I have an alternative PR ready, mirroring how Pete is binding AP_Servo_Telem in the soon-to-be-merged #28857. That PR would be guaranteed not to change any behavior outside of Lua, however, I think this PR is vastly a better approach.


IamPete1 (approved):
LGTM

robertlong13:
Also, this PR was flight tested together with the already-merged https://github.com/ArduPilot/ardupilot/pull/28999 with that 15-minute flight on Volanti.

PR #29035, opened 2025-01-10
Create ArduRocket

Lioeng:


rmackay9:
This looks like a mistake so I might close it.  We can always re-open if it's not.  By the way, the obvious "dual usage" of rockets has been a barrier in the past at least to AP adding support for rockets

PR #29036, opened 2025-01-10
fix param Increment comment

Huibean:
fix param Increment comment for xml metadata use


Huibean:
@rmackay9 done, thanks for the correction!

rmackay9 (approved):

tpwrules:
Where does this cause problems? There are a few other places in the code with a dot and no leading zero.

peterbarker:
> Where does this cause problems? There are a few other places in the code with a dot and no leading zero.

https://discord.com/channels/674039678562861068/710994174299603055/1327151096380133396

image

peterbarker:
Merged, thanks!

tpwrules:
If that is the standard the generator should probably normalize it.

Huibean:
> If that is the standard the generator should probably normalize it.

For number value is not hard to correct or report invalid, and in the text field just keep displaying the original content from xml

PR #29038, opened 2025-01-10
AP_Mission: get_item: ensure returned item is clean

IamPete1:
Mission does not fill in all fields of the returned item, this means you can get random data. This results in getting bad data via misison ftp or scripting. The mavlink mission item protocol does not use this method. The other mission ftp types are OK because they always fill in the full set of information. 

To reproduce, upload a misison and then read it back again. For example [before.txt](https://github.com/user-attachments/files/18379386/before.txt). Testing shown for SITL but the bug also exists on real hardware.

This is what we upload:
image

This is what we read back on master:
image

With the fix we get the same as what is written:
image

This only shows up in the x,y,z fields of a item without location because the other fields are defaulted here:

https://github.com/ArduPilot/ardupilot/blob/6efe2105496ff9ffeca2ba13c0d80b1f9212c308/libraries/AP_Mission/AP_Mission.cpp#L1574-L1585

A alternate (or complementary) fix would be to also set default values for the x,y and z values there.

### DIFF
--- a/libraries/AP_Mission/AP_Mission.cpp
+++ b/libraries/AP_Mission/AP_Mission.cpp
@@ -736,16 +736,9 @@ bool AP_Mission::get_item(uint16_t index, mavlink_mission_item_int_t& ret_packet
         return false;
     }
 
-    // minimal placeholder values during read-from-storage
-    tmp.target_system = 1;     // unused sysid
-    tmp.target_component =  1; // unused compid
-
-    // 0=home, higher number/s = mission item number.
-    tmp.seq = index;
-
     // retrieve mission from eeprom
     AP_Mission::Mission_Command cmd {};
-    if (!read_cmd_from_storage(tmp.seq, cmd)) {
+    if (!read_cmd_from_storage(index, cmd)) {
         return false;
     }
     // convert into mavlink-ish format for lua and friends.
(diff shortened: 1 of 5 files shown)


IamPete1:
I caved and added the defaulting in the `mission_cmd_to_mavlink_int` function too. There is no change in behavior from that fix because all callers pass in a clean command object.

libraries/AP_Mission/AP_Mission.cpp
```diff
@@ -1577,12 +1577,19 @@ bool AP_Mission::mission_cmd_to_mavlink_int(const AP_Mission::Mission_Command& c
...
+    // Strictly these should be defaulted to NaN
     packet.param1 = 0;
     packet.param2 = 0;
     packet.param3 = 0;
     packet.param4 = 0;
-    packet.frame = 0;
-    packet.autocontinue = 1;
+
+    // Strictly these should be defaulted to INT32_MAX for x and y and NaN for z
+    packet.x = 0;
```
peterbarker, line 1590: Maybe just zero it - that way if extensions fields are added they also get zeroed...

IamPete1: The caller may have filled in the target system and component ID. The changes in this function are not needed in anycase, the all the callers do pass in a zeroed structure (after the first commit). The main thing is at least zeroing these extras manually means that we always set the same fields no matter the way-point type.

IamPete1: In fact the `get_item` fills in the `target_system`, `target_component` and `seq` before passing it in. It probably doesn't need to....  but it currently does....

IamPete1: I have re-worked to zero and remove those unneeded sets.

IamPete1: Hum, that broke everything because mission protocol uses `convert_MISSION_ITEM_INT_to_MISSION_ITEM`, I have reworked some more, but I'm tempted to go back to the just manually zeroing X Y and Z.

peterbarker (approved):

rmackay9 (approved):

IamPete1:
Tested with `--valgrid`,  didn't all pass, but no mission related failures. Scripting ones are because there is junk in my scripts folder. Safety switch one is because of a pymavlink thing. So just two GPS failures.

image

I'm happy those were not caused by this change.

PR #29039, opened 2025-01-10
AP_Mission: don't allow waypoint 0 (home) to be changed

IamPete1:
This highlights what is a tricky problem to fix. The main problem is that this will break most scripts which try and upload missions. 

The problem is that mission allows you to set way-point 0, however when you try and read way point 0 you just get home. 

https://github.com/ArduPilot/ardupilot/blob/6efe2105496ff9ffeca2ba13c0d80b1f9212c308/libraries/AP_Mission/AP_Mission.cpp#L817-L822

This is the key this issue. You can write whatever you like to waypoint 0, but there is no way to read it back or for it to be used in anyway. (The only way I can think is that if you have a eeprom copy you could read it out manualy)

This makes the change that your no longer allowed to write to waypoint 0, this give users a clue that there throwing away there first point, it will never be used.

Maybe the more complete fix is to stop storing home completely, this would require doing param conversion on the total mission item count. 

A nice example of something that does not work as expected on master is the copter save way point switch (you have to make sure your mission is fully cleared and MIS_TOTAL is 0). In this case it carefully checks for there being 0 waypoints and tries to add a takeoff.

https://github.com/ArduPilot/ardupilot/blob/6efe2105496ff9ffeca2ba13c0d80b1f9212c308/ArduCopter/RC_Channel_Copter.cpp#L238-L250  

However, that takeoff is written to index 0 so it is never used. If you read the mission back again you will find only a single mission item and no take off:
image

### DIFF
--- a/libraries/AP_Mission/AP_Mission.cpp
+++ b/libraries/AP_Mission/AP_Mission.cpp
@@ -498,14 +498,18 @@ bool AP_Mission::start_command(const Mission_Command& cmd)
 ///     cmd.index is updated with it's new position in the mission
 bool AP_Mission::add_cmd(Mission_Command& cmd)
 {
+    // Home is always the index 0.
+    // Commands should be added after home.
+    const uint16_t index = MAX(_cmd_total, AP_MISSION_FIRST_REAL_COMMAND);
+
     // attempt to write the command to storage
-    bool ret = write_cmd_to_storage(_cmd_total, cmd);
+    bool ret = write_cmd_to_storage(index, cmd);
 
     if (ret) {
         // update command's index
-        cmd.index = _cmd_total;
+        cmd.index = index;
         // increment total number of commands
-        _cmd_total.set_and_save(_cmd_total + 1);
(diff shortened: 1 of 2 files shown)


robertlong13:
> this will break most scripts which try and upload missions.

Does it? It looks like only scripts that are already broken will be just as broken but now set_item will return false (which I'd imagine many broken scripts aren't checking). I'd say we should go further and actually break those scripts by putting the range check on the binding, and make lua throw an error.

IamPete1:
> Does it? It looks like only scripts that are already broken will be just as broken but now set_item will return false (which I'd imagine many broken scripts aren't checking). I'd say we should go further and actually break those scripts by putting the range check on the binding, and make lua throw an error.

Sot of, currently they would upload a mission but loose the first item, if we made the change they would error out.

tridge:
i think we should auto-create the home wp if set_item() is called when num WPs is zero, and always return true from writing to wp 0

PR #29041, opened 2025-01-11
Disable angle/Alt-hold based avoidance by default

rishabsingh3003:
I have never seen or heard of the "ping pong method" working successfully and I think we should disable/remove it by deafult. The parameter AVOID_ANGLE_MAX is also confusing because it does not work for simple avoidance and is only meant for the "ping-pong" method. 

Recently I encountered a few setups that were trying to get the regular loiter-based avoidance working, and kept alt-hold on a switch for manual control/recovery. However, since the angle based avoidance was enabled by default in that mode which the operator did not know about, it almost led to a crash.  I don't think anyone would mind removing this feature by default but it makes things cleaner and safer IMO.


rmackay9 (approved):
I've approved but after that I'm happy to have it merged.

ArduCopter/mode_althold.cpp
```diff
@@ -79,7 +79,7 @@ void ModeAltHold::run()
     case AltHoldModeState::Flying:
         motors->set_desired_spool_state(AP_Motors::DesiredSpoolState::THROTTLE_UNLIMITED);
 
-#if AP_AVOIDANCE_ENABLED
+#if AP_AVOIDANCE_ENABLED && AP_ALT_HOLD_AVOIDANCE_ENABLED
```
peterbarker, line 82: This doesn't look right; should be a single define being checked here?

rmackay9:
One small thing, I wonder why the build option is AC_AVOID_ALTHOLD and the code definition is AP_ALT_HOLD_AVOIDANCE_ENABLED, maybe make them the same?

rmackay9:
I've updated the contents of this PR here https://github.com/ArduPilot/ardupilot/pull/32360 so I'll close this one.

PR #29042, opened 2025-01-11
Plane Interlock DRAFT

timtuxworth:
On my TX16S I have arming and estop protected by momentary switch. Unless I hold the switch on/up the plane can't arm. This works great so I can't accidentally arm/disarm or emergency stop the motors unless I also hold the switch at the same time. I am setting up a SiYi UniRC7 which doesn't have a way to do this like on the TX16S which uses a whole mess of special functions and global variables to make it work. So I want to implement this protection in ArduPlane. 

This PR implements the MOTOR_INTERLOCK RCx_OPTION for Plane and adds an ILCK_OPTIONS bitmask to allow the user what it applies to. The options are:

-     Arm 
-     Disarm 
-     RudderArm
-     RudderDisarm
-     Emergency Stop

And the interlock applies no matter how it is used, so Arm via RC, rudder, Mission Planner Aux function or mavproxy will all require the interlock RC channel (if set) to be low. 

This would usually be set on a momentary switch so the pilot has to hold the switch or button in order to allow Arming etc.

If there is no RC channel set to MOTOR_INTERLOCK then existing behavior works as it currently does.

### DIFF
--- a/ArduPlane/RC_Channel_Plane.cpp
+++ b/ArduPlane/RC_Channel_Plane.cpp
@@ -208,6 +208,10 @@ void RC_Channel_Plane::init_aux_function(const RC_Channel::AUX_FUNC ch_option,
         // want to startup with reverse thrust
         break;
 
+    case AUX_FUNC::MOTOR_INTERLOCK:
+        plane.interlock_disallowed();
+        break;
+
     default:
         // handle in parent class
         RC_Channel::init_aux_function(ch_option, ch_flag);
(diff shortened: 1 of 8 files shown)


ArduPlane/RC_Channel_Plane.cpp
```diff
@@ -474,6 +478,72 @@ bool RC_Channel_Plane::do_aux_function(const AuxFuncTrigger &trigger)
...
+        case AuxSwitchPos::HIGH:
+            if(plane.interlock_option_allowed(InterlockOptions::Arm)) {
+                // request arm, disable emergency motor stop
+                SRV_Channels::set_emergency_stop(false);
+                AP::arming().arm(AP_Arming::Method::AUXSWITCH, true);
+            } else {
+                GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "Interlock %s %s", "Arm", "denied");
+            }
+            break;
+        case AuxSwitchPos::MIDDLE:
```
Hwurzburg, line 507: you dont need a middle case...

Hwurzburg (changes requested):
-what about the disarm method...nothing done there
-what about the normal arm/disarm switch, the arm/disarm with airmode?

I think given the flash cost of this and that it will be useful for a very small group of users, it should be a build option, defaulted to off normally

timtuxworth:
> -what about the disarm method...nothing done there -what about the normal arm/disarm switch, the arm/disarm with airmode?

Yes it's an option to have the interlock to apply to disarm (or not), the ILCK_OPTIONS can have it apply to Arm, Disarm, Rudder Arm, Rudder Disarm, Motor Emergency stop - or not.

> 
> I think given the flash cost of this and that it will be useful for a very small group of users, it should be a build option, defaulted to off normally

Hwurzburg:
I dont see it blocking anything in the disarm method
I don't see the other aux switches I mentioned superseded in the Plane RC_Channels

timtuxworth:
I have a lua script for this which works great.

PR #29047, opened 2025-01-12
HAL_ChibiOS: Added hardware definition for Spedix F405 / Argus ECO FC F405

KoteSoft:


peterbarker:
Looks like we now have support for this board already merged into master.

Thanks anyway!

PR #29049, opened 2025-01-13
Remove command-line option enabling synthetic clock 

peterbarker:
It's always on


khancyr (approved):
LGTM
This was always confusing so nice to remove

PR #29052, opened 2025-01-13
QuadPlane: add RPM check to takeoff

robertlong13:
I'm taking over #27696. This PR addresses Tridge's requests, moving the check to `QuadPlane::takeoff_controller` like the tiltrotor delay check, and never allows the motors to be blocked for plane.

For clarity, I moved all the takeoff preventers (tilt, rpm, and rudder stick) into the same if-block on the spool state. The tilt check was checking `spool_state != AP_Motors::DesiredSpoolState::THROTTLE_UNLIMITED` and the rudder check was checking `spool_state <= AP_Motors::DesiredSpoolState::GROUND_IDLE`. These are currently identical conditionals, so I picked the former, rather arbitrarily.

I also removed the AP_Motors_config stuff. If we want to add an `AP_MOTORS_ENABLED` define, I think we should do it properly and use it everywhere we're currently using `APM_BUILD_COPTER_OR_HELI || APM_BUILD_TYPE(APM_BUILD_ArduPlane)` around calls to `AP::motors`. I was going to do that as part of this PR, but it got a little tricky (has the same weird thing we had to do for AP_LANDINGGEAR_ENABLED), and I decided to punt it to [another branch](https://github.com/robertlong13/ardupilot/tree/pr/make-and-use-ap-motors-enabled) (for a future PR) so I can get Peter Barker's help with it later.

Also, while I was here, I noticed that the max rpm check (to catch cases where one prop isn't installed) is almost certainly broken for real vehicles. It works in SITL because the simulated arm RPM goes immediately from 0 to the simulated value. Real vehicles will probably pass through a valid RPM value before reaching the invalidly high RPM, and once it passes once, it won't check again. `takeoff_check` is called at 50Hz on Copter, so it's pretty easy to imagine it will read at least one RPM value within bounds on its way. I don't want that to hold up this PR though (I'm not planning to use that check), so I'll fix it afterwards if nobody else takes it. It will take some SITL changes to properly autotest it.


libraries/AP_Vehicle/AP_Vehicle.cpp
```diff
@@ -1111,6 +1111,50 @@ void AP_Vehicle::check_motor_noise()
 #endif
 }
 
+#if HAL_WITH_ESC_TELEM && (APM_BUILD_COPTER_OR_HELI || APM_BUILD_TYPE(APM_BUILD_ArduPlane))
+bool AP_Vehicle::motors_takeoff_check(float rpm_min, float rpm_max)
```
IamPete1, line 1115: Could this whole function live in ESC telem lib?

robertlong13: This was one of the decisions from https://github.com/ArduPilot/ardupilot/pull/27696. It was originally put within AP_Motors, but because it depends on two subsystems, it was decided that it should go higher up to AP_Vehicle.

ArduCopter/takeoff_check.cpp
```diff
@@ -8,49 +8,19 @@
...
-        takeoff_check_warning_ms = now_ms;
-        const char* prefix_str = "Takeoff blocked:";
-        if (!telem_active) {
-            gcs().send_text(MAV_SEVERITY_CRITICAL, "%s waiting for ESC RPM", prefix_str);
-        } else if (!rpm_adequate) {
-            gcs().send_text(MAV_SEVERITY_CRITICAL, "%s ESC RPM out of range", prefix_str);
-        }
-    }
+    // Run the common takeoff checks
+    motors->set_spoolup_block(!motors_takeoff_check(g2.takeoff_rpm_min, g2.takeoff_rpm_max));
```
andyp1per, line 24: I am sure this is the right kind of refactoring, but I am having hard time convincing myself that before and after is equivalent for copter. Will need careful testing

robertlong13: The hard line on `set_spoolup_block` and not blocking manual takeoffs in plane made it tough to split out the common code in a more-obviously-equivalent way. The order of some of these if-checks has shuffled for sure, which is tough.

If you have any loose suggestions for additional edge-cases to cover in the autotests, I can add them.

andyp1per: I think just relying on the autotests and making sure this is flown on copter (where it is used a lot) is sufficient. @rmackay9 should also review since he was the driving force behind most of the copter changes

peterbarker:
@robertlong13 could you rebase this and bring it to a DevCall to get the rework in COpter OK'd, please?

robertlong13:
Rebased. I made a couple changes to how I implemented the copter side which should make it easier to confirm correctness/equivalence.

This PR slightly impacts that CPU load check; previously the CPU load check was skipped if the RPM check was not enabled. @andyp1per, let me know if that was actually intentional and I'll add that back (but we should document it if so).

I have not retested this yet other than running the autotests.

andyp1per:
> This PR slightly impacts that CPU load check; previously the CPU load check was skipped if the RPM check was not enabled. @andyp1per, let me know if that was actually intentional and I'll add that back (but we should document it if so).

It was intentional, but we don't have a lot of experience with how meaningful it is in practice. I think the CPU load check should definitely go in AP_Vehicle - you will want it for quadplane as well.

robertlong13:
> It was intentional, but we don't have a lot of experience with how meaningful it is in practice.

Gotcha, so want me to go back to disabling it when min RPM is zero?

> I think the CPU load check should definitely go in AP_Vehicle - you will want it for quadplane as well.

I agree, but after talking to Peter, I decided to separate that into a future PR instead of adding it to this one.

andyp1per:
> Gotcha, so want me to go back to disabling it when min RPM is zero?

Yes, otherwise there is no way to turn it off

ArduPlane/quadplane.cpp
```diff
@@ -556,6 +556,22 @@ const AP_Param::GroupInfo QuadPlane::var_info2[] = {
     // @User: Standard
     AP_GROUPINFO("APPROACH_DIST", 39, QuadPlane, approach_distance_m, 0),
     
+#if HAL_WITH_ESC_TELEM
+    // @Param: TKOFF_RPM_MIN
+    // @DisplayName: Takeoff Check RPM minimum
+    // @Description: Takeoff is not permitted until motors report at least this RPM.  Set to zero to disable check
```
IamPete1: I think we need to make this clear its auto takeoff only. Copter is all takeoffs I think.

robertlong13: Ah, stale description from the original implementation. Good catch

rmackay9:
I'll give this a test on my aerial photography copter

ArduPlane/quadplane.cpp
robertlong13: Updated. I've also added a comment that disabling the min check also disables the max check.

ArduPlane/quadplane.cpp
```diff
@@ -556,6 +556,22 @@ const AP_Param::GroupInfo QuadPlane::var_info2[] = {
     // @User: Standard
     AP_GROUPINFO("APPROACH_DIST", 39, QuadPlane, approach_distance_m, 0),
     
+#if HAL_WITH_ESC_TELEM
+    // @Param: TKOFF_RPM_MIN
+    // @DisplayName: Auto Takeoff Check RPM minimum
```
andyp1per, line 561: Why only auto takeoff? Surely this is applicable in the same way that it is on copter?

Tools/autotest/arducopter.py
```diff
@@ -1255,6 +1232,7 @@ def TakeoffCheck(self):
         self.test_takeoff_check_mode("POSHOLD")
         # self.test_takeoff_check_mode("SPORT")
 
+        self.start_subtest("Check TKOFF_RPM_MAX works")
```
andyp1per: What about min? Seems like some testing may have been lost on copter here. I think on a change like this it is better to leave the existing tests as-is to verify that you really haven't broken anything

ArduPlane/quadplane.cpp
robertlong13: That's what the original PR did. Tridge didn't want that https://github.com/ArduPilot/ardupilot/pull/27696#discussion_r1719136921

andyp1per: Makes zero sense to me given that this is specifically designed to prevent you flipping if one motor doesn't spin in copter modes - but that's just me.

Tools/autotest/arducopter.py
robertlong13: What? This is an added line that is, essentially, a comment/print-statement. I don't know why Peter added it, and why he didn't add one before the min check too, but this is clearly not deleting anything from the test. I have now dropped that line anyway.

No testing has been lost. A helper method has been moved, almost verbatim (a force-disarm option was added, defaulting to false), to the parent class. It's very easy to compare in an editor the method before/after it moved to the parent class.

ArduPlane/quadplane.cpp
robertlong13: Tridge's concern is that spoolup blocks, no matter how careful we are, could stop a quadplane during back-transition in some edge case or future bug, and he feels better just not having them in quadplane at all. I see where he's coming from, but I'm not totally convinced we can't just be careful; I'm not going to fight it. The auto takeoff is the case I care about anyway.

When I arm in QHover, I can easily check that the props are spinning before I raise the stick. If I arm in Auto, I can't verify before it flips (I believe this is another Plane-vs-Copter difference: we can arm in Auto if the current command is a takeoff)

IamPete1: Yeah, Plane will take off without waiting for pilot throttle in auto. I agree it would be nice to apply the check in all modes, but that can be a future enhancement.

IamPete1:
> I would like to see the original copter test re-instated. Code duplication can be removed in a future PR.

The Copter test is still there, just the helper function is moved up to the common vehicle test file.

andyp1per (approved):
LGTM

rmackay9:
Sorry, I will test this!

rmackay9:
I was not able to test this fully but I did lightly test and confirmed that the "Takeoff blocked" message appeared if my vehicle's RPM were below the TKOFF_MIN parameter value, thanks!
image

PR #29059, opened 2025-01-13
Allow reverse channel positions on stick gesture channels

andyp1per:
This fixes a long standing bug, where TX stick commands do not obey AP reversal settings

### DIFF
--- a/libraries/AP_OSD/AP_OSD_ParamScreen.cpp
+++ b/libraries/AP_OSD/AP_OSD_ParamScreen.cpp
@@ -388,37 +388,13 @@ void AP_OSD_ParamScreen::modify_configured_parameter(uint8_t number, Event ev)
     }
 }
 
-// return radio values as LOW, MIDDLE, HIGH
-// this function uses different threshold values to RC_Chanel::get_aux_switch_pos()
-// to avoid glitching on the stick travel
-RC_Channel::AuxSwitchPos AP_OSD_ParamScreen::get_channel_pos(const RC_Channel &chanref) const
-{
-    const auto *chan = &chanref;
-
-    const uint16_t in = chan->get_radio_in();
-    if (in <= 900 || in >= 2200) {
-        return RC_Channel::AuxSwitchPos::LOW;
-    }
-
-    // switch is reversed if 'reversed' option set on channel and switches reverse is allowed by RC_OPTIONS
-    bool switch_reversed = chan->get_reverse() && rc().option_is_enabled(RC_Channels::Option::ALLOW_SWITCH_REV);
-
-    if (in < RC_Channel::AUX_PWM_TRIGGER_LOW) {
-        return switch_reversed ? RC_Channel::AuxSwitchPos::HIGH : RC_Channel::AuxSwitchPos::LOW;
-    } else if (in > RC_Channel::AUX_PWM_TRIGGER_HIGH) {
-        return switch_reversed ? RC_Channel::AuxSwitchPos::LOW : RC_Channel::AuxSwitchPos::HIGH;
-    } else {
-        return RC_Channel::AuxSwitchPos::MIDDLE;
-    }
-}
-
 // map rc input to an event
 AP_OSD_ParamScreen::Event AP_OSD_ParamScreen::map_rc_input_to_event() const
 {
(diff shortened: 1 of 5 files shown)


libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -1952,6 +1952,28 @@ RC_Channel::AuxSwitchPos RC_Channel::get_aux_switch_pos() const
...
+    // switch is reversed if 'reversed' option set on channel and switches reverse is allowed by RC_OPTIONS
+    bool switch_reversed = get_reverse();
+
+    if (in < RC_Channel::AUX_PWM_TRIGGER_LOW) {
+        return switch_reversed ? RC_Channel::AuxSwitchPos::HIGH : RC_Channel::AuxSwitchPos::LOW;
+    } else if (in > RC_Channel::AUX_PWM_TRIGGER_HIGH) {
+        return switch_reversed ? RC_Channel::AuxSwitchPos::LOW : RC_Channel::AuxSwitchPos::HIGH;
+    } else {
+        return RC_Channel::AuxSwitchPos::MIDDLE;
+    }
```
IamPete1, line 1974:
```suggestion
    if (in < RC_Channel::AUX_PWM_TRIGGER_LOW) {
        return switch_reversed ? RC_Channel::AuxSwitchPos::HIGH : RC_Channel::AuxSwitchPos::LOW;
    } 
    if (in > RC_Channel::AUX_PWM_TRIGGER_HIGH) {
        return switch_reversed ? RC_Channel::AuxSwitchPos::LOW : RC_Channel::AuxSwitchPos::HIGH;
    } 
    return RC_Channel::AuxSwitchPos::MIDDLE;
```

libraries/AP_OSD/AP_OSD_ParamScreen.cpp
```diff
@@ -388,37 +388,13 @@ void AP_OSD_ParamScreen::modify_configured_parameter(uint8_t number, Event ev)
...
-{
-    const auto *chan = &chanref;
-
-    const uint16_t in = chan->get_radio_in();
-    if (in <= 900 || in >= 2200) {
-        return RC_Channel::AuxSwitchPos::LOW;
-    }
-
-    // switch is reversed if 'reversed' option set on channel and switches reverse is allowed by RC_OPTIONS
-    bool switch_reversed = chan->get_reverse() && rc().option_is_enabled(RC_Channels::Option::ALLOW_SWITCH_REV);
```
IamPete1, line 404: There is a change in behavior for OSD it used to check the switch reversal option.

andyp1per: Yeah, but it didn't used to - this was broken when that option or some related change was introduced. It's been broken for a long time.

rmackay9 (approved):
This looks good to me, thanks!

I agree with PeterH's suggestion re the if/else stucture

rmackay9:
Ah, as per usual, some clarification on the testing that's been done would be good

peterbarker:
So this isn't really "rcmap" so much as "roll/pitch/yaw channels".  How those are determined is opaque now - thankfully.

I like the concept, not looking deeply as Randy's already approved.

My only concern is that we're going to be magically changing the meaning of that parameter.  We might end up changing people's setups rather significantly.  Just noting it, I'm not all that fussed until someone points out just how bad that fact is....

rmackay9:
@peterbarker, Txs for the feedback.  I don't think there's much danger to this change because RunCam simply uses the stick gestures to control the camera's OSD.  This allows users to set various options on the camera.

What would be scary is if this affected the stick gesture for arming but that's not the case

andyp1per:
Tested on a RunCam tracer both the menu commands and the OSD parameter change commands

peterbarker (approved):

rmackay9:

rmackay9:
With help from @andyp1per this is not included in 4.6.0-beta3, thanks!

Hwurzburg:
@peterbarker why do you think this needs wiki? it just affects how stick gestures affect RunCam setup and was backwards if user had used chan reversing to work normally for AP, correct? seems just to make bug fixed...does not affect anything other than RunCam users, correct? @andyp1per ?

andyp1per:
Yes agree @Hwurzburg

PR #29061, opened 2025-01-14
AP_Periph: Add restart acknowledgement for UAVCAN

haydendonald:
There was no acknowledgement for UAVCAN so we add it here.

Tested on: CubeNode, AP_Periph

### DIFF
--- a/Tools/AP_Periph/can.cpp
+++ b/Tools/AP_Periph/can.cpp
@@ -797,15 +797,23 @@ void AP_Periph_FW::onTransferReceived(CanardInstance* canard_instance,
         handle_begin_firmware_update(canard_instance, transfer);
         break;
 
-    case UAVCAN_PROTOCOL_RESTARTNODE_ID:
-        printf("RestartNode\n");
-        hal.scheduler->delay(10);
-        prepare_reboot();
-#if CONFIG_HAL_BOARD == HAL_BOARD_CHIBIOS
-        NVIC_SystemReset();
-#elif CONFIG_HAL_BOARD == HAL_BOARD_SITL
-        HAL_SITL::actually_reboot();
-#endif
+    case UAVCAN_PROTOCOL_RESTARTNODE_ID: {
+            printf("RestartNode\n");
+            uavcan_protocol_RestartNodeResponse pkt {
+                ok: true,
+            };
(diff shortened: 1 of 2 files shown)


IamPete1 (commented):
This is not millis wrap safe. If the +1000 causes a wrap it will reboot instantly. Better to store the reboot request time. Then you can do `(AP_HAL::millis() - reboot_request_ms) > 1000` which should deal with the wrap correctly.

haydendonald:
> This is not millis wrap safe. If the +1000 causes a wrap it will reboot instantly. Better to store the reboot request time. Then you can do `(AP_HAL::millis() - reboot_request_ms) > 1000` which should deal with the wrap correctly.

Done, built, and tested working :)

IamPete1 (approved):
LGTM

bugobliterator (approved):

Tools/AP_Periph/can.cpp
```diff
@@ -1950,6 +1958,16 @@ void AP_Periph_FW::can_update()
 
         }
     }
+
+    // if there is a reboot scheduled, do it 1 second after request to allow the acknowledgement to be sent
```
peterbarker, line 1962: 1 second seems a little arbitrary.  And perhaps long.

It would be nice to have some sort of trigger (ie. triggered when we know the packet's been successfully sent), with the timeout as a secondary trigger. 

Probably too much work, really.

Tools/AP_Periph/can.cpp
```diff
@@ -1950,6 +1958,16 @@ void AP_Periph_FW::can_update()
...
+    if (reboot_request_ms != 0 && (AP_HAL::millis() - reboot_request_ms > 1000)) {
+        prepare_reboot();
+#if CONFIG_HAL_BOARD == HAL_BOARD_CHIBIOS
+        NVIC_SystemReset();
+#elif CONFIG_HAL_BOARD == HAL_BOARD_SITL
+        HAL_SITL::actually_reboot();
+#endif
```
peterbarker, line 1969: pre-existing, and doesn't need fixing here, but seems to say we should have an `actually_reboot` method on the HAL.  Preferably with a better name.

peterbarker (approved):

PR #29066, opened 2025-01-14
Rover: add arm-once-after-prearms-pass functionality

peterbarker:
Add option to Rover to arm the vehicle *once*, once prearm (and arm!) checks have passed.

This is intended to be an alternative to having an `ARMING_REQUIRE` of 0, which can cause all sorts of fun which is best avoided.

Also adds an "REQUIRE_POSITION" as a DriveOptions setting for Rover, the same as Copter's "REQUIRE_POSITION" functionality.

Tested in SITL.

autotests coming, assuming this doesn't get insta-killed...


IamPete1:
Doesn't have the advantage of your `get_last_prearm_checks_result` but this does work:

```
-- Arm the vehicle one time once arming checks have passed
local function update()

    if arming:is_armed() then
        -- Kill script once armed
        return
    end

    if arming:pre_arm_checks() and arming:arm() then
        -- Kill the script once it works
        return
    end

    return update, 5000
end

return update, 5000
```

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -96,7 +96,8 @@ const AP_Param::GroupInfo AP_Arming::var_info[] = {
     // @Param{Plane, Rover}: REQUIRE
     // @DisplayName: Require Arming Motors 
     // @Description: Arming disabled until some requirements are met. If 0, there are no requirements (arm immediately).  If 1, sends the minimum throttle PWM value to the throttle channel when disarmed. If 2, send 0 PWM (no signal) to throttle channel when disarmed. On planes with ICE enabled and the throttle while disarmed option set in ICE_OPTIONS, the motor will always get THR_MIN when disarmed. Arming will occur using either rudder stick arming (if enabled) or GCS command when all mandatory and ARMING_CHECK items are satisfied. Note, when setting this parameter to 0, a reboot is required to immediately arm the plane.
-    // @Values: 0:Disabled,1:minimum PWM when disarmed,2:0 PWM when disarmed
+    // @Values{Plane}: 0:Disabled,1:minimum PWM when disarmed,2:0 PWM when disarmed
```
rmackay9, line 99: It's an existing issue but just as a drive-by comment I'd say we could improve the wording of "1:minimum PWM when disarmed" because PWM doesn't clarify if its for RC input or servo output.  It should perhaps be changed to be "1:RC throttle at minimum" or something similar

Rover/Parameters.cpp
```diff
@@ -694,6 +694,12 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Path: mode_circle.cpp
     AP_SUBGROUPINFO(mode_circle, "CIRC", 57, ParametersG2, ModeCircle),
 
+    // @Param: DRIVE_OPTIONS
```
rmackay9, line 697: I'd like to avoid this new parameter if possible.  I think the user could achieve essentially the same thing by setting the INITIAL_MODE parameter to the mode that requires a position estimate?

rmackay9:
Thanks for this, it would be great to remove the arming-required =0 option.

libraries/AP_Arming/AP_Arming.cpp
peterbarker: Definitely need tridge's input on that one.  It's the latter.

Rover/Parameters.cpp
peterbarker: The point here is to require that the vehicle has position in all modes.  So the user can't arm the vehicle at all until it knows how to get home.

libraries/AP_Arming/AP_Arming.cpp
rmackay9: yes, it's outside the scope of this PR, I'm just bringing it up as a drive-by comment

Rover/Parameters.cpp
rmackay9: yes, I definitely understand the need for this now and we have the same option in Copter.  I just wonder why we put it into FLIGHT_OPTIONS instead of ARMING_OPTIONS as part of PR https://github.com/ArduPilot/ardupilot/pull/27367

peterbarker: > yes, I definitely understand the need for this now and we have the same option in Copter. I just wonder why we put it into FLIGHT_OPTIONS instead of ARMING_OPTIONS as part of PR #27367

I think we *did* discuss at DevCall - but I can't find any notes on it.

It probably should be in arming options.  Should I move it?  This will require some awkward code in Copter if we want people to start using  the bit in ARMING_OPTIONS.  But I think we should.

rmackay9: Hi @peterbarker yes, I'm sorry but I agree that we should move it.  It's new for 4.6 so I think we could get away without the parameter conversion if we backported it before the official release

peterbarker: .... actually, I think the problems revolved around Plane.

If we ever allow Plane to arm without position (@Ryanf55 ...) we would have issues around retaining the default for Plane.  parameter defaults and bitmasks are a mess - if someone already has set ARMING_OPTION in their parameter storage then you can't have sensible default values on a per-vehicle basis (or indeed any non-zero bit default).

rmackay9: For copter and rover the default is zero (e.g. require GPS for arming is false) but I guess plane would like that to default to 1.  It's sad to put the option in the wrong place because of this though.  I understand the issue but.. it's hard to accept doing the wrong thing for a small technical reason

Ryanf55: Good details, I need time to digest, but am swamped at work. If you aren't in a rush, I can discuss in 2 days.

peterbarker: PR to move the bit into AP_Arming: https://github.com/ArduPilot/ardupilot/pull/29076

peterbarker:
Works as expected on my Rover

peterbarker:
Note that Rover doesn't actually use the `ZERO_PWM` garbage, but values added for both cases anyway.

rmackay9 (approved):
great, thanks!

PR #28991, opened 2025-01-02
Plane: ATT.DesPitch now reports actual target, not nav output

Georacer:
## Summary

This PR changes the logged quantity in `ATT.DesPitch` from `Plane::nav_pitch_cd` to the target value that actually gets passed to the pitch controller.

## Details

Logging `Plane::nav_pitch_cd` in `ATT.DesPitch` gives the false impression that there's steady-state error in the pitch controller.

This PR improves log readability, as it already includes `PTCH_TRIM_DEG` and the contribution of `KFF_THR2PTCH`.

`Plane::nav_pitch_cd` is already logged in `CTUN`.

## Testing

`sim_vehicle.py` was spun up and `PTCH_TRIM_DEG=3` was set.
TAKEOFF mode was armed.

**Before**:
`ATT.DesPitch` and `CTUN.NavPitch` overlap.
Steady-state error is falsely logged.
Screenshot from 2025-01-02 11-59-11

**After**:
The three pitch quantites are clearly discernible.
Screenshot from 2025-01-02 11-58-48

---
**EDIT**

The PR was updated to write NaN on every loop, when a mode does not employ `demanded_pitch_cd` to control pitch. The log viewers handle this gracefully:
Screenshot from 2025-01-14 12-55-15
Screenshot from 2025-01-14 12-56-22

PR #29067, opened 2025-01-14
Plane: ATT.DesPitch now reports actual target, not nav output (Float version)

Georacer:
Alternative to https://github.com/ArduPilot/ardupilot/pull/28991

It converts `Plane::demanded_pitch_cd` into a `float`, and can thus save a lot of headache: It is no longer necessary to have a second variable to track the value validity.

Still tested to work in SITL, by changing MANUAL->FBWA->MANUAL.


ArduPlane/Plane.h
```diff
@@ -668,9 +668,12 @@ class Plane : public AP_Vehicle {
     // The instantaneous desired bank angle.  Hundredths of a degree
     int32_t nav_roll_cd;
 
-    // The instantaneous desired pitch angle.  Hundredths of a degree
+    // The instantaneous navigator-commanded pitch angle. Hundredths of a degree
     int32_t nav_pitch_cd;
 
+    // The instantaneous desired pitch angle. Hundredths of a degree.
```
peterbarker, line 674:
```suggestion
    // The instantaneous desired pitch angle in degrees
```

peterbarker (changes requested):

ArduPlane/Attitude.cpp
```diff
@@ -209,7 +209,7 @@ float Plane::stabilize_pitch_get_pitch_out()
     const bool quadplane_in_frwd_transition = false;
 #endif
 
-    int32_t demanded_pitch = nav_pitch_cd + int32_t(g.pitch_trim * 100.0) + SRV_Channels::get_output_scaled(SRV_Channel::k_throttle) * g.kff_throttle_to_pitch;
+    demanded_pitch = nav_pitch_cd*0.01f + g.pitch_trim + SRV_Channels::get_output_scaled(SRV_Channel::k_throttle)*0.01f*g.kff_throttle_to_pitch;
```
IamPete1, line 212: I think this also needs setting on 196 for the quadplane case. `use_fw_attitude_controllers` is not quite the same as `show_vtol_view` that we use for switching over logging, there maybe some overlap where this would be a change in behavior.

IamPete1 (commented):
There are some modes where we swap between angle controller and not. It does look like we currently just leave the angle targets hanging in those cases. Training is one case:
https://github.com/ArduPilot/ardupilot/blob/429ed5cdd26e857491a4c2c26c885bd5da70df46/ArduPlane/mode_training.cpp#L55-L58

Nav scripting is the other:

https://github.com/ArduPilot/ardupilot/blob/429ed5cdd26e857491a4c2c26c885bd5da70df46/ArduPlane/Attitude.cpp#L405-L423

It would be nice to reflect that in the log, but it is a existing issue, so it doesn't need to be resolved here.

Hwurzburg:
This impacts tuning instructions for level trim, so I put a wiki needed label on it .....but it will remove the offset in logs when comparing Desired vs Actual looking at pitch attitude tuning....we need to be sure that it does NOT affect the artificial horizon, but I think it uses ATT.Pitch which isn't affected by this

IamPete1:
I think I prefer https://github.com/ArduPilot/ardupilot/pull/29152, we might as well make use of the redundant pitch logged in CTUN. It also makes it match the HUD which I think is more intuitive.

Georacer:
> There are some modes where we swap between angle controller and not. It does look like we currently just leave the angle targets hanging in those cases. Training is one case [...]

Indeed the target pitch would get stuck. Now that it's being set to NaN on every loop it will be evident when it's not actually being used/followed.

Georacer:
I've updated with Peter's suggestions and tested on quadplane as well.

The top plot is the 3 pitches.
The two bottom ones are to give an indication of the plane/VTOL switch.

In FW flight the CTUN.NavPitch and ATT.DesPitch separate. In VTOL they merge back together.
I'm not sure what causes the slow merge, but I've noticed the same behaviour in master as well.

Screenshot from 2025-01-28 23-29-51

ArduPlane/Plane.cpp
```diff
@@ -511,6 +511,10 @@ void Plane::update_control_mode(void)
     takeoff_state.throttle_lim_max = 100.0f;
     takeoff_state.throttle_lim_min = -100.0f;
 
+    // Preemptively set demanded pitch to NaN, so that it properly doesn't
+    // get logged. Modes ovewriting it (most modes do) will change it.
+    demanded_pitch_logged = logger.quiet_nanf();
```
peterbarker, line 516:
```suggestion
    demanded_pitch_logged = AP::logger().quiet_nanf();
```

tridge:
I'd like input from @Hwurzburg on this, Henry and I discussed, and Henry suggested that CTUN should have the vehicle specific trims applied, but ATT should only have the AHRS trims applied
note that CTUN for change between the fixed wing pitch trim and Q_TRIM_PITCH depending on the Q view mode

peterbarker (changes requested):

Hwurzburg:
My thoughts on Pitch logging:
1. Any log group that reports desired/target pitch and attained/actual pitch should have both overlap when in steady state "level" flight if the vehicle has been appropriately tuned, ie in FBWA hands off mid- throttle steady state, both should be the same mean value.
2. We should have at least one log group that shows attained/actual pitch is ZERO under steady state, level altitude, cruise conditions, ie artificial horizon level.

When we mount the autopilot,we use IMU cal to set the AHRS trims for "level AHRS". This should be in line with the fuselage line being earth frame horizontal.
We then, as desired, use PITCH_TRIM_DEG or Q_PITCH_TRIM to adjust the "flight trim" such that the artificial horizon indication in fixed wing flight is level under desired cruise conditions, or that Qplanes generally hover in place in no wind. We'll call this the level trimmed attitude.

I think that ATT pitch logs should reflect the angle departure from the level cal attitude, reflecting IMU calibration trimmed outputs (ie independent of other trimming params) and that CTUN pitch logs should reflect angle departure from level trimmed attitude (current artificial horizon level attitude indication which includes pitch trim params), whether in FW or VTOL. In both cases, the desired should match the attained pitch if steady state on a well tuned vehicle.

I think @IamPete1 agrees with me on this?

ArduPlane/Plane.cpp
IamPete1: @peterbarker Why? I always assumed if we have access its better to go direct.

ArduPlane/Attitude.cpp
IamPete1: bump on this^

PR #29068, opened 2025-01-14
Add decapenta frameclass

tsg03:
Hi guys, I tried my best and added a new frame class which is basically similar to the “OctaQuad” frame class but with two additional motors. To be honest, I don't really know what I need to change and where, or how communication with MAVLink works. I oriented myself on the documentation and searched for additional files all over the repo, hoping to make all the necessary changes. I would greatly appreciate the help of knowledgeable developers.


IamPete1:
The easiest thing to do is load it at runtime using scripting. See our fault tolerant hex example:

https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/examples/MotorMatrix_fault_tolerant_hex.lua

You need a flight controller which supports scripting of course. Then you set `FRAME_CLASS` to "15:Scripting Matrix".

peterbarker:
https://github.com/ArduPilot/ardupilot/pull/29020 adds similarly ridiculous frames, you might model things on that PR.

Georacer:
Linguistics police here! Decapenta means "fifteen". For 10 motors, you probably want something like "deca frameclass".

Georacer:
> Linguistics police here! Decapenta means "fifteen". For 10 motors, you probably want something like "deca frameclass".

On second thought, "decapente" is fifteen in modern Greek. Since you're a cool kid and you probably mean to use ancient Greek, fifteen there is "pente-ke-deca". So there is no conflict in meaning.

And I understand that you want to differentiate between the flat-10 configuration and you tried to write "ten-five".
So writing "deca-penta" in an ancient Greek context would indeed mean "ten-five".

Thanks for giving me the trigger to brush that corner of my memory a bit!

YoubiDev:
> > Linguistics police here! Decapenta means "fifteen". For 10 motors, you probably want something like "deca frameclass".
> 
> On second thought, "decapente" is fifteen in modern Greek. Since you're a cool kid and you probably mean to use ancient Greek, fifteen there is "pente-ke-deca". So there is no conflict in meaning.
> 
> And I understand that you want to differentiate between the flat-10 configuration and you tried to write "ten-five". So writing "deca-penta" in an ancient Greek context would indeed mean "ten-five".
> 
> Thanks for giving me the trigger to brush that corner of my memory a bit!

Hi Georacer,

Your linguistic insights are appreciated and certainly add depth to the discussion. However, I would suggest presenting corrections with a tone that prioritizes respect and fosters constructive collaboration. While precision is important, maintaining a polite and professional approach ensures a more inclusive and supportive environment for all contributors.

Thank you for understanding, and I look forward to your continued contributions to this discussion

peterbarker (changes requested):
Must be implemented as a LUA script.

We have previously bounced pull requests adding these esoteric frames to the C++ code, requiring the user to run a script to set up the motors matrix.  I did not agree with this decision and argued against it at the time, but that's what we have decided as a project.

I have already given a link to a PR which adds some scripts.

If you want the frame to continue to be tested you can add a SITL frame for it and an entry into `vehicleinfo.py`.  I *doubt* it would stop working.

On the subject of naming... it's difficult, but the schema appears to be working out as "{number_of_props}_{sublayout}_{orientation}".  So "dotriaconta_octaquad_x" is 32-props with enough octaquad layouts to add up to 32".  So a slight tweak to naming, please - "deca_penta" rather than "decapenta".  And maybe an "_x" or "_plus" depending on whether there's a motor at yaw-0 or not.

Incidentally, IMO it would have been better overall had you *not* responded to Georacer.  Their comment was not out-of-line.

peterbarker:
@tsg03 are you interested in pursuing this one as a LUA script?

peterbarker:
No answer; redo as LUA script and re-open PR if you want to pursue.

PR #29071, opened 2025-01-14
AP_Mission: better deal with home being in index 0.

IamPete1:
This is a result of the discussion on https://github.com/ArduPilot/ardupilot/pull/29039. The two key things are:

- `AP_Mission::add_cmd` will add home if needed and then add the passed in command so that the passed in command is always added and not lost in index 0.
- `AP_Mission::replace_cmd` will not replace the command in index 0, but it does still return true in that case.
- `AP_Mission::write_home_to_storage` will set `_cmd_total` to 1 if it is 0.

There are also a number of clean ups and bug fixes:

- new `present()` method on mission which consolidates all the places that were trying to tell if there was a mission. We had been using `mission.num_commands() > 1` and `mission.num_commands() <= 1` or ` mission.num_commands() < 2`  for not present. This also replaces `mission.num_commands() == 0` which copter was using incorrectly for `AUX_FUNC::SAVE_WP`
- Fixes the lack of a workaround in copter `AUX_FUNC::SAVE_WP` and remove the need for the workaround in rover `AUX_FUNC::SAVE_WP`
- AHRS now directly calls mission `write_home_to_storage` so it will always be upto date and we don't have to do it in rover and plane manually.

The AHRS auto set of the home mission item is the only one users might notice, this was only done on plane and rover previously. Because it now also updates `_cmd_total` it is now more likely to be 1 than 0 with no mission loaded.

### DIFF
--- a/libraries/AP_Mission/AP_Mission.cpp
+++ b/libraries/AP_Mission/AP_Mission.cpp
@@ -971,6 +971,11 @@ void AP_Mission::write_home_to_storage()
     home_cmd.id = MAV_CMD_NAV_WAYPOINT;
     home_cmd.content.location = AP::ahrs().get_home();
     write_cmd_to_storage(0,home_cmd);
+
+    // Make sure command total reflects that home has been added
+    if (_cmd_total < 1) {
+        _cmd_total.set_and_save(1);
+    }
 }
 
 MAV_MISSION_RESULT AP_Mission::sanity_check_params(const mavlink_mission_item_int_t& packet)
--- a/libraries/AP_Mission/AP_Mission.cpp
+++ b/libraries/AP_Mission/AP_Mission.cpp
@@ -498,6 +498,11 @@ bool AP_Mission::start_command(const Mission_Command& cmd)
 ///     cmd.index is updated with it's new position in the mission
 bool AP_Mission::add_cmd(Mission_Command& cmd)
 {
+    // Add home if its not already present
+    if (_cmd_total < 1) {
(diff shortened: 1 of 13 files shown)


IamPete1:
I have updated the copter `RecordThenPlayMission` test to expect  the takeoff as the first item (the test was enforcing that adding the takeoff did not work).

IamPete1:
I have updated mission write to not change `_last_change_time_ms` if index 0 is being written, not of the things checking it look at waypoint 0 in an case so this ensures we not triggering things to re-fetch the mission now we do write to index 0 more often.

libraries/AP_Mission/AP_Mission.cpp
```diff
@@ -498,6 +498,11 @@ bool AP_Mission::start_command(const Mission_Command& cmd)
 ///     cmd.index is updated with it's new position in the mission
 bool AP_Mission::add_cmd(Mission_Command& cmd)
 {
+    // Add home if its not already present
+    if (_cmd_total < 1) {
```
tridge, line 502: i think we need a check for index 0 in set_item, triggering add of home, return true, otherwise it could break currently working scripts

tridge (changes requested):

libraries/AP_Mission/AP_Mission.cpp
IamPete1: Fixed

tridge (approved):

rmackay9 (approved):

PR #29073, opened 2025-01-14
Rover: add require-position-for-arming option

peterbarker:
Extracted from https://github.com/ArduPilot/ardupilot/pull/29066 as it's really a separate feature

Tested in SITL, including autotest

### DIFF
--- a/Rover/AP_Arming_Rover.cpp
+++ b/Rover/AP_Arming_Rover.cpp
@@ -34,7 +34,9 @@ bool AP_Arming_Rover::rc_calibration_checks(const bool display_failure)
 // performs pre_arm gps related checks and returns true if passed
 bool AP_Arming_Rover::gps_checks(bool display_failure)
 {
-    if (!rover.control_mode->requires_position() && !rover.control_mode->requires_velocity()) {
+    if (!require_location &&
+        !rover.control_mode->requires_position() &&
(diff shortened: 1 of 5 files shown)


Rover/Parameters.cpp
```diff
@@ -694,6 +694,12 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Path: mode_circle.cpp
     AP_SUBGROUPINFO(mode_circle, "CIRC", 57, ParametersG2, ModeCircle),
 
+    // @Param: DRIVE_OPTIONS
+    // @DisplayName: Drive Options
+    // @Description: Rnadom behavioural changes for the vehicle
```
Hwurzburg, line 699: spelling

peterbarker: Fixed, thanks

rmackay9:
Hi @peterbarker, thanks for this.  I've already commented on the other PR but shouldn't this be in ARMING_OPTIONS?  I see in Copter we've put it in flight options but I wonder why..

Rover/Parameters.cpp
peterbarker: We've moved to a different parameter now

rmackay9 (approved):
thanks!

Issue #29072, opened 2025-01-14
Inconsistency in SITL JSON interface example


Tuxliri:
https://github.com/ArduPilot/ardupilot/blob/4088ab91c62d7ad1f4472df461061ff84ede1433/libraries/SITL/examples/JSON/C%2B%2B/minimal.cpp#L71-L76

In the example above the IMU is supposed to have a FRD (x-axis pointing forward, y-axis pointing right and z-axis pointing down) orientation. In this case the expected measured acceleration in the line above should be `9.81` rather than `-9.81`. Is this correct or did I understand this incorrectly?

Tuxliri:
Okay I have figured this out. When the drone is on a surface the accelerometer of the imu senses the counteracting force of the flat surface pushing up against gravity, thus an upward acceleration of `9.81 m/s^2`. In the FRD body reference frame (*z-axis pointing down*) this is an acceleration of `-9.81 m/s^2`.

Tuxliri:
Added a clarification in the code.

PR #29074, opened 2025-01-14
Added comment to clarify IMU acceleration value

Tuxliri:
Clarifying IMU reading value, as noted in #29072.

### DIFF
--- a/libraries/SITL/examples/JSON/C++/minimal.cpp
+++ b/libraries/SITL/examples/JSON/C++/minimal.cpp
@@ -67,6 +67,12 @@ int main() {
         ap.setWindvane(1 , 1);
         ap.setRangefinder(rangefinder_example, 6);
 
+        /* 
+        When the drone is on a surface the accelerometer of the imu senses the counteracting 
+        force of the flat surface pushing up against gravity, thus an upward acceleration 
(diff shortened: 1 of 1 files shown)


rmackay9:
Hi @Tuxliri,

Txs for this.  Could you update the commit title to prefix it with "SITL: "?  We like to prefix all our commits with the subsystem or folder name where the change is made because it makes backporting easier.  To be clear, I meant the **commit** title, not the PR title.

Tuxliri:
@rmackay9 done!

IamPete1:
Did you find the readme? https://github.com/ArduPilot/ardupilot/blob/master/libraries/SITL/examples/JSON/readme.md Do you think we should add the clarification there too?

Tuxliri:
@IamPete1 this is example-specific not related to the JSON protocol per-se. But it would be useful to add a section to the README with a description of the reference frames used and this could be an example there. I have written a Jupyter notebook to explain this [link](https://docs.duckietown.com/ente/devmanual-duckiematrix/developing/physics/rotorpy-ardupilot-conventions.html), I can sum it up in the readme.

peterbarker:
@IamPete1 are you happy with this?

PR #29076, opened 2025-01-15
Move REQUIRE_POSITION_FOR_ARMING bit to AP_arming 

peterbarker:
... also rename it to `REQUIRE_LOCATION_FROM_ARMING` to be more in-line with ArduPilot's distinction between "position" (local position) and "location" (absolute lat/lng)

### DIFF
--- a/ArduCopter/Copter.h
+++ b/ArduCopter/Copter.h
@@ -632,7 +632,6 @@ private:
         DISABLE_THRUST_LOSS_CHECK     = (1<<0),   // 1
         DISABLE_YAW_IMBALANCE_WARNING = (1<<1),   // 2
         RELEASE_GRIPPER_ON_THRUST_LOSS = (1<<2),  // 4
-        REQUIRE_POSITION_FOR_ARMING =   (1<<3),   // 8
     };
 
     // type of fast rate attitude controller in operation
--- a/libraries/AP_Arming/AP_Arming.cpp
+++ b/libraries/AP_Arming/AP_Arming.cpp
@@ -85,6 +85,26 @@
   #define ARMING_RUDDER_DEFAULT         (uint8_t)RudderArming::ARMDISARM
 #endif
 
+// find a default value for ARMING_NEED_POS parameter, and determine
+// whether the parameter should be shown:
+#ifndef AP_ARMING_NEED_LOC_PARAMETER_ENABLED
+// determine whether ARMING_NEED_POS is shown:
+#if APM_BUILD_COPTER_OR_HELI
+#define AP_ARMING_NEED_LOC_PARAMETER_ENABLED 1
+#else
+#define AP_ARMING_NEED_LOC_PARAMETER_ENABLED 0
+#endif  // build types
+#endif  // AP_ARMING_NEED_LOC_PARAMETER_ENABLED
+
+// if ARMING_NEED_POS is shown, determine what its default should be:
+#if AP_ARMING_NEED_LOC_PARAMETER_ENABLED
+#if APM_BUILD_COPTER_OR_HELI
+#define AP_ARMING_NEED_LOC_DEFAULT 0
+#else
+#error "Unable to find value for AP_ARMING_NEED_LOC_DEFAULT"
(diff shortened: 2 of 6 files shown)


libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -146,6 +146,7 @@ const AP_Param::GroupInfo AP_Arming::var_info[] = {
     // @DisplayName: Arming options
     // @Description: Options that can be applied to change arming behaviour
     // @Bitmask: 0:Disable prearm display,1:Do not send status text on state change
+    // @Bitmask{Copter}: 0:Disable prearm display,1:Do not send status text on state change,2:Require position for arming
```
amilcarlucas, line 149:
```suggestion
    // @Bitmask{Copter}: 0:Disable prearm display,1:Do not send status text on state change,2:Require location for arming
```

amilcarlucas: Why not add it to plane also?

peterbarker: Plane *already* requires position for arming, all the time.

That's actually why we put this into Copter's flight options in the first place - we're going to have an awkward time when Plane no longer requires position for arming....

But we want this bit for Rover, so @rmackay9 asked for it to be moved (https://github.com/ArduPilot/ardupilot/pull/29073)

rmackay9 (approved):
Thanks for this.  I'm not sure we really needed to replace "position" with "location" but it's OK in either case.

More importantly I guess the reason we put it in FLIGHT_OPTIONS in the first place was because once someone sets this bit they will never get the default changes to the ARMING_OPTIONS parameter?  that hardly seems like a good enough reason so I suspect I'm missing something

peterbarker:
> Thanks for this. I'm not sure we really needed to replace "position" with "location" but it's OK in either case.
> 
> More importantly I guess the reason we put it in FLIGHT_OPTIONS in the first place was because once someone sets this bit they will never get the default changes to the ARMING_OPTIONS parameter? that hardly seems like a good enough reason so I suspect I'm missing something

I should have taken better notes :-)

ArduCopter/Copter.h
```diff
@@ -629,7 +629,6 @@ class Copter : public AP_Vehicle {
         DISABLE_THRUST_LOSS_CHECK     = (1<<0),   // 1
         DISABLE_YAW_IMBALANCE_WARNING = (1<<1),   // 2
         RELEASE_GRIPPER_ON_THRUST_LOSS = (1<<2),  // 4
-        REQUIRE_POSITION_FOR_ARMING =   (1<<3),   // 8
```
IamPete1, line 632: Should we keep the enum value to prevent the bit being used in the future? EG "REQUIRE_POSITION_FOR_ARMING_UNUSED". 

It might be a good idea even if we don't do the conversion, it really depends on how exciting the next option we want to add it.

IamPete1 (commented):
Was added https://github.com/ArduPilot/ardupilot/pull/27367, do we need conversion?

ArduCopter/Copter.h
peterbarker: @rmackay9 mentioned this hasn't gone out in a stable release.  It's a reasonable question, 'though.

peterbarker:
> Was added #27367, do we need conversion?

Randy suggested we could make this change with relatively little fuss as it hasn't gone out in stable.

Do you disagree?

IamPete1:
I'm OK with not doing the conversion. but I think we should keep the enum value reserved. We could decide to re-use it when we know what the next option is and how critical it would be it if was unexpectedly enabled.

peterbarker:
I've now adjusted this according to our DevCall meeting outcomes.

That's a new parameter, `ARMING_NEED_POS`, with two valid values, yes and no, 1 and 0.

This PR only moves it for Copter, https://github.com/ArduPilot/ardupilot/pull/29073 adds it for Rover (based on this PR)

IamPete1 (approved):
LGTM

rmackay9:

PR #29077, opened 2025-01-15
Plane: Add climb slope minimum height parameter

rubenp02:
# Plane: Add climb slope minimum height parameter

Description
-----------
### [Plane: Add climb slope minimum height parameter](https://github.com/ArduPilot/ardupilot/pull/29077/commits/ae51af99de443375df7805abd5bc112b18d81eee):

Slope treatment for a flight leg was conditioned to a minimum height of 20m (which was a hardcoded value and only documented in a comment in the code), below which an immediate altitude change was performed for that leg. This caused unpredictable results for flight plans with waypoints at exactly 20m followed by a climb, as the aircraft could, depending on chance, be above or below the height demand, resulting in very different trajectories.

This commit addresses that issue with the following changes:
- Added a parameter to control the minimum height at which a gradual climb slope can be performed. This documents the original feature and retains the safety aspect of it.
- Changed the behavior when below said minimum height to only immediately climb up to the safe height, regaining the intended climb slope afterwards. This leverages existing code for recalculating climb slopes, provides a good balance between safety and following the flight plan as intended, and fixes the core issue of different trajectories being taken based on random external factors.

Testing flight plan and logs
-----------
[testing-logs.zip](https://github.com/user-attachments/files/18421787/testing-logs.zip)

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1267,6 +1267,15 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     AP_SUBGROUPINFO(systemid, "SID", 38, ParametersG2, AP_SystemID),
 #endif
     
+    // @Param: CLIMB_SLOPE_HGT
+    // @DisplayName: Climb slope mininum height
+    // @Description: Sets the minimum height above home at which the aircraft will apply a climb slope between waypoints. Below it, the aircraft will ascend immediately, and will only resume the requested trajectory upon reaching this height. This prevents unsafe behavior such as attempting to slowly gain altitude near obstacles. The default value ensures safe operations in most environments, but it can be adjusted based on specific terrain or operational needs.
+    // @Units: m
+    // @Range: 0 50
+    // @Increment: 1
+    // @User: Advanced
+    AP_GROUPINFO("CLIMB_SLOPE_HGT", 39, ParametersG2, waypoint_climb_slope_height_min, 25),
+
     AP_GROUPEND
 };
 
--- a/ArduPlane/altitude.cpp
+++ b/ArduPlane/altitude.cpp
@@ -83,16 +83,9 @@ void Plane::setup_alt_slope(void)
             break;
         }
 
-        // we only do glide slide handling in AUTO when above 20m or
-        // when descending. The 20 meter threshold is arbitrary, and
-        // is basically to prevent situations where we try to slowly
-        // gain height at low altitudes, potentially hitting
-        // obstacles.
-        if (adjusted_relative_altitude_cm() > 2000 || above_location_current(next_WP_loc)) {
-            set_offset_altitude_location(prev_WP_loc, next_WP_loc);
-        } else {
-            reset_offset_altitude();
-        }
+        // otherwise we set up an altitude slope for this leg
+        set_offset_altitude_location(prev_WP_loc, next_WP_loc);
+
         break;
     default:
         reset_offset_altitude();
@@ -311,6 +304,17 @@ void Plane::change_target_altitude(int32_t change_cm)
 void Plane::set_target_altitude_proportion(const Location &loc, float proportion)
 {
     set_target_altitude_location(loc);
+
+    // Only do altitude slope handling when above CLIMB_SLOPE_HGT or when
+    // descending. This is meant to prevent situations where the aircraft tries
+    // to slowly gain height at low altitudes, potentially hitting obstacles.
+    if (target_altitude.offset_cm > 0 &&
+        (adjusted_relative_altitude_cm() <
+         (g2.waypoint_climb_slope_height_min * 100))) {
+        // Early return to ensure a full-rate climb past CLIMB_SLOPE_HGT
+        return;
+    }
+
     proportion = constrain_float(proportion, 0.0f, 1.0f);
     change_target_altitude(-target_altitude.offset_cm*proportion);
 
--- a/ArduPlane/Parameters.h
+++ b/ArduPlane/Parameters.h
@@ -522,6 +522,8 @@ public:
 
     AP_Int32 flight_options;
 
+    AP_Int16 waypoint_climb_slope_height_min;
+
     AP_Int8 takeoff_throttle_accel_count;
     AP_Int8 takeoff_timeout;
 


ArduPlane/Parameters.cpp
```diff
@@ -1309,7 +1309,16 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("RNGFND_LND_ORNT", 36, ParametersG2, rangefinder_land_orient, ROTATION_PITCH_270),
 #endif
-    
+
+    // @Param: CLMB_SLOPE_HGT
+    // @DisplayName: Climb slope mininum height
+    // @Description: Sets the minimum height at which the aircraft will apply a climb slope between waypoints. Below it, the aircraft will ascend immediately, and will only resume the requested trajectory upon reaching this height. This prevents unsafe behavior such as attempting to slowly gain altitude near obstacles. The default value of 20m ensures safe operations in most environments, but it can be adjusted based on specific terrain or operational needs.
```
Hwurzburg, line 1315: what height? above home, above terrain?

Hwurzburg:
needs proper library commit splits

ArduPlane/Parameters.cpp
```diff
@@ -88,23 +88,23 @@ const AP_Param::Info Plane::var_info[] = {
...
     // @Increment: 1
     // @Units: m
     // @User: Advanced
-    GSCALAR(glide_slope_min, "GLIDE_SLOPE_MIN", 15),
+    GSCALAR(alt_slope_min, "ALT_SLOPE_MIN", 15),
 
-    // @Param: GLIDE_SLOPE_THR
-    // @DisplayName: Glide slope threshold
-    // @Description: This controls the height above the glide slope the plane may be before rebuilding a glide slope. This is useful for smoothing out an autotakeoff
+    // @Param: ALT_SLOPE_THR
```
timtuxworth, line 100: Since the param is being changed anyway by the PR can we do something about THR usually meaning Throttle? How about THD?

timtuxworth:
Probably needs parameter conversion for the changed params no? https://ardupilot.org/dev/docs/code-overview-adding-a-new-parameter.html#parameter-conversion-expiration

Ryanf55:
> This seems to make sense, what do you think @Ryanf55?


Many thanks for the contribution!

I like the naming change on first glance, perhaps that's worth splitting to it's own PR?

The other change also makes sense. If we could get some exact steps to reproduce the poor behavior in SITL, so I can see a before/after from this PR, that would be fantastic. Some recommendation on which plots you are looking at would be helpful. At least for me, who's less experienced, understanding what the steps you take to reproduce the behavior and what graphs/logs you check are very useful.

ArduPlane/Parameters.cpp
Hwurzburg: ALT_SLOPE_THRES is clearer and fits the name length restriction

ArduPlane/altitude.cpp
```diff
@@ -79,25 +79,15 @@ void Plane::setup_glide_slope(void)
...
-        // when descending. The 20 meter threshold is arbitrary, and
-        // is basically to prevent situations where we try to slowly
-        // gain height at low altitudes, potentially hitting
-        // obstacles.
-        if (adjusted_relative_altitude_cm() > 2000 || above_location_current(next_WP_loc)) {
-            set_offset_altitude_location(prev_WP_loc, next_WP_loc);
-        } else {
-            reset_offset_altitude();
-        }
+        set_offset_altitude_location(prev_WP_loc, next_WP_loc);
```
Hwurzburg, line 89: good to keep comment (update)

Hwurzburg (changes requested):

ArduPlane/altitude.cpp
Hwurzburg: I find it helpful to say why you call a function whose actions arent clearly described in its name, as well as complete comment on the function definition itself....not a blocker, just might  help someone in the future understand whats going on

peterbarker:
As mentioned by Henry, separate PRs for a name change and functionality addition would be a very good idea.

Hwurzburg (approved):

ArduPlane/Parameters.cpp
```diff
@@ -1309,7 +1309,16 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("RNGFND_LND_ORNT", 36, ParametersG2, rangefinder_land_orient, ROTATION_PITCH_270),
 #endif
-    
+
+    // @Param: CLMB_SLOPE_HGT
+    // @DisplayName: Climb slope mininum height
+    // @Description: Sets the minimum height above home at which the aircraft will apply a climb slope between waypoints. Below it, the aircraft will ascend immediately, and will only resume the requested trajectory upon reaching this height. This prevents unsafe behavior such as attempting to slowly gain altitude near obstacles. The default value of 20m ensures safe operations in most environments, but it can be adjusted based on specific terrain or operational needs.
```
timtuxworth, line 1315: It seems like using set_altitude_offset_location() in setup_glide_slope() means this will now work correctly if terrain following. Do I have that right?
IMO that would be a very good thing, so should this comment be changed to reflect that?

timtuxworth (approved):
I'd love to see this in master. LGTM.

IamPete1 (changes requested):
Its not clear to me exactly what the expected change of behavior is here. Adding a param to replace the hard coded value is clear. But then the logic is moved into a different function? That change expands the scope from just auto mode to lots of other modes. Its also called directly in landing via a functor.

I would prefer to just update the hard coded value to the param. Otherwise we will need loads of testing.

ArduPlane/altitude.cpp
```diff
@@ -315,8 +307,19 @@ void Plane::set_target_altitude_proportion(const Location &loc, float proportion
 {
     set_target_altitude_location(loc);
     proportion = constrain_float(proportion, 0.0f, 1.0f);
+
+    // We only do altitude slope handling when above CLMB_SLOPE_HGT or when
+    // descending. This is meant to prevent situations where we try to slowly
+    // gain height at low altitudes, potentially hitting obstacles.
+    if (target_altitude.offset_cm > 0 &&
+        adjusted_relative_altitude_cm() < g2.waypoint_climb_slope_height_min * 100) {
+        plane.set_target_altitude_location(plane.next_WP_loc);
```
IamPete1, line 316: Should use the passed in location here. I also think its odd that your setting to the destination altitude here. That means that the altitude target will jump up to the destination, then when you climb above the min height it jumps back down to min height and then ramps from there. Better to constrain up to min height so that the altitude is constant before the ramp rather than having step changes. (you will also have to constrain down to the altitude of the passed in location).

IamPete1 (commented):
I would like to see testing in lots of modes, RTL, QRTL, Guided, auto with all altitude types with special care for terrain relative.

IamPete1:
> More importantly, do you think what I was saying before makes sense in the first place?

Yeah, it makes sense. I do wonder if we should bump up the min height a bit to make up for the shorter "max climb" it will do now. That is to say if you have your first point at 10 the next at 100. We might have missed a obstacle at 25 meters before and hit it now. 

If you want to write tests that would be great!, but I don't think its required. The key thing is just to make sure we don't get it completely wrong in some edge case. Tests often are not great at proving a new feature works, but they a great to make sure we don't break the feature in the future.

ArduPlane/Parameters.cpp
IamPete1: We could go to `CLIMB` we have two characters free.

IamPete1:
The test logs (I looked at two) show this odd altitude demand spike (MAVExplorer: `graph TECS.hin POS.RelHomeAlt<20:2`):
image

We see the demanded altitude go to 100 and then back to the ramp value when the altitude is above 20m but then there is a little spike. Any idea what is going on?

IamPete1:
To answer my own question, its overshooting by 5m and triggering the glide slope-recalculation. 

image

`graph NTUN.TAT abs(NTUN.AltE)>5:2`

IamPete1:
> Yes, it should be fixed by constraining the altitude for the immediate climb like you suggested, I think

Yeah, I think I would prefer to increase the default threshold by 5m and have less overshoot than keep the current threshold with the overshoot. 

Maybe we need to look at the glide slope recalc in the future, I would have expected it to recalculate up but not come down again afterwards as it did in those logs.

ArduPlane/Parameters.cpp
tridge, line 1315: probably should be above ground level, if we are doing terrain following currently, but should not use lidar

ArduPlane/altitude.cpp
```diff
@@ -315,8 +307,19 @@ void Plane::set_target_altitude_proportion(const Location &loc, float proportion
 {
     set_target_altitude_location(loc);
     proportion = constrain_float(proportion, 0.0f, 1.0f);
+
+    // We only do altitude slope handling when above CLMB_SLOPE_HGT or when
+    // descending. This is meant to prevent situations where we try to slowly
+    // gain height at low altitudes, potentially hitting obstacles.
+    if (target_altitude.offset_cm > 0 &&
+        adjusted_relative_altitude_cm() < g2.waypoint_climb_slope_height_min * 100) {
```
tridge, line 315: maybe use above ground?

tridge (changes requested):

tridge:
@rubenp02 the reason for relative ground level is users doing long distance flights from point A to point B where B is much higher/lower than A. I think using relative_ground_altitude(false) would be correct if we are currently terrain following

ArduPlane/altitude.cpp
IamPete1: Hum, I guess the problem is using the true height as feedback. We could remove that and just shape the desired altitude.

IamPete1:
I can't think of anyway we can keep the new "only climb to min height" behavior that was the original intent of this PR and make the height threshold evaluated continuously relative to terrain.  It would result into forcing everything to be terrain following at low altitude. We could evaluate the relative to terrain height when we start the leg and convert it to a above home alt for use in the remainder of the leg.

IamPete1:
Thinking about this and https://github.com/ArduPilot/ardupilot/pull/29142, they both have the same problem. They want a max rate climb, but if its relative to terrain they don't know at what altitude the climb will be completed. This means you have to essentially make up a height target to pass to TECS and hope that it will be high enough to trick TECS into doing a max rate climb.

What we really need is a way to just tell TECS to do the climb directly without needing to pass a altitude. What we would do on copter is pass the climb rate and then continually check if the stopping point has reached the required altitude at the current location.

rubenp02:
Just rebased this and updated the full-rate climb to use the target WP altitude again. Using the CLIMB_SLOPE_HGT is not correct since it causes a slow climb, and if the aircraft doesn't overshoot it at all it will actually get stuck at around that height. Since https://github.com/ArduPilot/ardupilot/pull/29142 got merged and used a 10m offset which is a bit arbitrary yet functionally identical to simply using the target WP altitude, this approach is validated. Rate management is TECS job, its height demand input often isn't continuous.

Currently, the 20m threshold below which the aircraft performs an immediate climb to the next waypoint height is still based on relative altitude, hardcoded, and nearly undocumented. Maybe this solution isn't perfect, but sometimes the perfect is the enemy of the good :)

As @IamPete1 said, using terrain based heights means making all low altitude climbs terrain-following, and I'd add that this would need a fallback altitude system anyway since terrain data isn't always available. I reckon that as it stands, this is already better in every front than the previous solution. The original behavior is quite unpredictable and dangerous.

ArduPlane/altitude.cpp
```diff
@@ -294,6 +287,18 @@ void Plane::change_target_altitude(int32_t change_cm)
...
 {
+    // Only do altitude slope handling when above CLIMB_SLOPE_HGT or when
+    // descending. This is meant to prevent situations where the aircraft tries
+    // to slowly gain height at low altitudes, potentially hitting obstacles.
+    if (target_altitude.offset_cm > 0 &&
+        (adjusted_relative_altitude_cm() <
+         (g2.waypoint_climb_slope_height_min * 100))) {
+        // Use the destination waypoint's altitude to ensure a full-rate climb
+        // past CLIMB_SLOPE_HGT
+        set_target_altitude_location(plane.next_WP_loc);
```
IamPete1: Should this use the passed in location `loc`?

rubenp02: Fixed, forgot that loc was next_WP_loc

ArduPlane/altitude.cpp
```diff
@@ -295,6 +288,17 @@ void Plane::change_target_altitude(int32_t change_cm)
 void Plane::set_target_altitude_proportion(const Location &loc, float proportion)
 {
     set_target_altitude_location(loc);
+
+    // Only do altitude slope handling when above CLIMB_SLOPE_HGT or when
+    // descending. This is meant to prevent situations where the aircraft tries
+    // to slowly gain height at low altitudes, potentially hitting obstacles.
+    if (target_altitude.offset_cm > 0 &&
+        (adjusted_relative_altitude_cm() <
+         (g2.waypoint_climb_slope_height_min * 100))) {
```
tridge, line 313: i think we want a special case for CLIMB_SLOPE_HGT==0, to skip this code (and in param docs)

tridge (dismissed):

tridge:
@IamPete1 will be doing some testing on this in SITL

timtuxworth:
I had thought this was supposed work when terrain following. Based on the discussion at the dev call it won't do that. Is this supposed to work with terrain @rubenp02 ?

rubenp02:
> @IamPete1 will be doing some testing on this in SITL

Great, thanks! I also did [some tests](https://github.com/ArduPilot/ardupilot/pull/29077#issuecomment-2630982148) myself a while back, and the current revision should be functionally identical to the one used for them. Just in case it's of any use.

rubenp02:
> I had thought this was supposed work when terrain following. Based on the discussion at the dev call it won't do that. Is this supposed to work with terrain @rubenp02 ?

To be honest I'm not familiar with the terrain following feature as I haven't used it, but this PR is unrelated. It is meant as an adjustment to the waypoint to waypoint climb slope handling when the relative altitude of the aircraft at the time it reaches the first waypoint is low (below CLIMB_SLOPE_HGT). The point is to prevent 2 issues with the current logic: 1) the rel. altitude when an immediate rather than a gradual height is triggered is hardcoded and undocumented and 2) the altitude demand during the mission is sometimes tied to the altitude at which the aircraft happens to be when entering a leg, rather than what's in the flight plan, which can be very unpredictable. IIRC there's a pretty clear example of this in the logs attached in the PR message.

So back to your question, during the leg it will interact with terrain following as follows:

- While the aircraft is under CLIMB_SLOPE_HGT it will act as it would during a normal climb leg with bit 14 of FLIGHT_OPTIONS (climb to next waypoint altitude immediately instead of linear climb) set.
- After it reaches CLIMB_SLOPE_HGT (and if that bit is unset, i.e. a linear climb is wanted if possible), it will resume the normal behavior, which is climbing linearly and only reaching the target altitude at the end of the leg.

Sorry I'm not being more specific because again, I'm not sure what the terrain following behavior is supposed to be under each of those conditions. But if it normally works in those cases, then it will certainly work here, too.

Where it doesn't apply and instead always defaults to an immediate climb is when the first waypoint of the leg is terrain alt. based. This is not related to this PR, it's just how it has always been with terrain alt. waypoints.

I have some testing notes of this feature in different modes and conditions. They're on the ZIP file I uploaded to this thread a while back, which also contains all of the SITL logs, but I'll also leave them here for convenience:

<details>
<summary>Notes</summary>
<br>

```
Tests performed and results:

- RTL ("rtl.*"):

    Navigation to a waypoint at 10 m. (below climb slope min. height), followed
    by an RTL.

    Result: No changes between this PR and master: immediate climb behavior is
    used, as a normal RTL doesn't use alt. slope handling. If RTL_AUTOLAND is
    set to 2, the Auto behavior (tested later) is used.

- QRTL ("qrtl.*"):

    Navigation to a waypoint at 10 m. (below climb slope min. height), followed
    by a manual mode change to QRTL.

    Result: No changes between this PR and master: immediate climb behavior is
    used.

- Guided ("guided.*"):

    Navigation to a waypoint at 10 m. (below climb slope min. height), followed
    by a MAV_CMD_DO_REPOSITION to a point at 100 m. (sent manually).

    Result: No changes between this PR and master: immediate climb behavior is
    used, as Guided navigation doesn't use alt. slope handling.

- Auto w/ relative altitudes ("auto-rel-alts.*"):

    Navigation to a waypoint at 10 m. (below climb slope min. height), followed
    by a waypoint at 100 m.

    Result: Immediate climb up until CLMB_SLOPE_HGT is reached. The aircraft
    then levels off until it gets to the planned slope, and then it follows it
    as normal (at this point, the trajectory is the same as if the leg had been
    entered above CLMB_SLOPE_HGT).

- Auto w/ absolute altitudes ("auto-abs-alts.*"):

    Navigation to a waypoint at 594 m, followed by a waypoint at 684 m, with
    Home and EKF origin being at 584 m.

    Result: Identical as when using relative altitudes, as this flight plan is
    equivalent to the one before in terms of height/relative altitudes.

- Auto w/ terrain-relative altitudes in both waypoints ("auto-terr-alts.*"):

    Navigation to a waypoint 10 m. above the terrain, followed by a waypoint 100
    m. above the terrain.

    Result: No changes between this PR and master: immediate climb behavior is
    used, as slope handling isn't applied to legs starting with a
    terrain-relative waypoint.

- Auto w/ relative to terrain altitude ("auto-rel-to-terr-alt.*"):

    Navigation to a waypoint at 10 m. (below climb slope min. height), followed
    by a waypoint 100 m. above the terrain.

    Result: Since the leg was entered through a relative alt. waypoint, an
    immediate climb up until CLMB_SLOPE_HGT is performed, followed by a climb
    slope to the next waypoint.
```

</details>

ArduPlane/altitude.cpp
rubenp02: But relative altitude might be 0 or even negative while the aircraft is flying, if it was launched from an elevated platform or in a point to point flight where the landing point is at a significantly lower altitude AMSL. And even when rel. alt. 0 is the ground, if CLIMB_SLOPE_HGT==0, the code will be skipped anyway since we will be flying higher than that.

IamPete1: I guess the argument is that you could set a large negative value to disable rather than having a special case 0 to disable. Since 0 is a valid value. 0 or -1 to disable is a fairly common pattern in AP tho. There is no reason a -2 ect could not still be treated as valid.

ArduPlane/Parameters.cpp
```diff
@@ -1270,6 +1270,15 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     AP_SUBGROUPINFO(systemid, "SID", 38, ParametersG2, AP_SystemID),
 #endif
     
+    // @Param: CLIMB_SLOPE_HGT
+    // @DisplayName: Climb slope mininum height
+    // @Description: Sets the minimum height above home at which the aircraft will apply a climb slope between waypoints. Below it, the aircraft will ascend immediately, and will only resume the requested trajectory upon reaching this height. This prevents unsafe behavior such as attempting to slowly gain altitude near obstacles. The default value of 25 m. ensures safe operations in most environments, but it can be adjusted based on specific terrain or operational needs.
```
IamPete1:
```suggestion
    // @Description: Sets the minimum height above home at which the aircraft will apply a climb slope between waypoints. Below it, the aircraft will ascend immediately, and will only resume the requested trajectory upon reaching this height. This prevents unsafe behavior such as attempting to slowly gain altitude near obstacles. The default value, ensures safe operations in most environments, but it can be adjusted based on specific terrain or operational needs.
```
Users can see the defualt in Mission Planner, including it in the text usualy just means we forget to update it if we change the defualt. People could also have loaded a custom defualt value in a custom build.

rubenp02: Good call. Done!

PR #29078, opened 2025-01-15
gitignore: Add correct crash dump patterns

rubenp02:
Adds pattern for core dumps and makes existing pattern for stack dumps more lenient, as depending on the environment they could be generated with a filename the old pattern didn't match


Ryanf55 (approved):
I've noticed these too, good idea.

khancyr (approved):
LGTM

PR #29079, opened 2025-01-15
Plane: Remove takeoff recalculation dead code

rubenp02:
The takeoff calculation changes introduced in c15fa7b and 6ce6ef8 mean the heading no longer gets recalculated at TKOFF_LVL_ALT. Removed the related code as it never runs.

### DIFF
--- a/ArduPlane/mode_takeoff.cpp
+++ b/ArduPlane/mode_takeoff.cpp
@@ -142,26 +142,11 @@ void ModeTakeoff::update()
         plane.set_flight_stage(AP_FixedWing::FlightStage::NORMAL);
         takeoff_mode_setup = false;
     }
-    // We update the waypoint to follow once we're past TKOFF_LVL_ALT or we
-    // pass the target location. The check for target location prevents us
-    // flying towards a wrong location if we can't climb.
-    // This will correct any yaw estimation errors (caused by EKF reset
-    // or compass interference from max throttle), since it's using position bearing.
-    const float altitude_cm = plane.current_loc.alt - start_loc.alt;
-    if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF
-        && plane.steer_state.hold_course_cd == -1 // This is the enter-once flag.
-        && (altitude_cm >= (level_alt * 100.0f) || start_loc.get_distance(plane.current_loc) >= dist)
-    ) {
-        const float direction = start_loc.get_bearing_to(plane.current_loc) * 0.01;
-        plane.next_WP_loc = start_loc;
-        plane.next_WP_loc.offset_bearing(direction, dist);
-        plane.next_WP_loc.alt += alt*100.0;
-        plane.steer_state.hold_course_cd = wrap_360_cd(direction*100); // Necessary to allow Plane::takeoff_calc_roll() to function.
-    }
-        
+
     // We finish the initial level takeoff if we climb past
     // TKOFF_ALT or we pass the target location. The check for
     // target location prevents us flying forever if we can't climb.
+    const float altitude_cm = plane.current_loc.alt - start_loc.alt;
     if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF &&
         (altitude_cm >= (alt*100 - 200) ||
         start_loc.get_distance(plane.current_loc) >= dist)) {


Georacer:
Hello @rubenp02 !

Thanks for this code inspection.

So basically after https://github.com/ArduPilot/ardupilot/commit/6ce6ef8fffb83ba69847dcdcb1d65e9f3ff27f22, the segment
```c++
    if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF
        && plane.steer_state.hold_course_cd == -1 // This is the enter-once flag.
        && (altitude_cm >= (level_alt * 100.0f) || start_loc.get_distance(plane.current_loc) >= dist)
    ) {
        const float direction = start_loc.get_bearing_to(plane.current_loc) * 0.01;
        plane.next_WP_loc = start_loc;
        plane.next_WP_loc.offset_bearing(direction, dist);
        plane.next_WP_loc.alt += alt*100.0;
        plane.steer_state.hold_course_cd = wrap_360_cd(direction*100); // Necessary to allow Plane::takeoff_calc_roll() to function.
    }
```
never gets executed.

This means that the loiter target direction for the TAKEOFF mode is set once during the first motion of the plane on the track and is never re-set once the plane reaches `TKOFF_LVL_ALT`.
@Hwurzburg does it sound like this is a behaviour we want?

Ryanf55:
Have you tried adding an AP_InternalError in the code block and run the entire plane autotest suite to better ensure it's never called currently?

Hwurzburg:
>  @Hwurzburg does it sound like this is a behaviour we want?

sounds like it

Hwurzburg (approved):

Georacer (changes requested):
To be honest, I think there's merit in re-calculating the loiter waypoint when crossing past `TKOFF_LVL_ALT`.
All sorts of things could make the original, on the ground speed vector inaccurate.

So, instead of deleting this code, I'd rather see it fixed so that it triggers the recalculation one time after `TKOFF_LVL_ALT` is reached.

@rubenp02 , I do see your point, but I still think it's a good safety net.

> I think it's desirable because the main point of the heading recalc was to correct yaw errors which isn't needed with the new changes.

Indeed yaw errors are prominent, so groundspeed and its direction is a better choice...

> Also, silently recalculating the heading like this caused issues. For example, planes with no rudder and some propeller induced yaw veer slightly before TKOFF_LVL_ALT as they can't roll to compensate, which meant that in this case the actual takeoff heading never matched the one reported to the GCS.

... but up to `TKOFF_LVL_ALT`, the plane is limited in roll action, so (as you mention), it can have drifted (in yaw and sideways) towards another direction.
If that's the case, isn't it better to re-set the plane on its new course, rather than have it correct to the original one?

So, to re-cap:
1. Without recalculation: Plane pursues the runway direction.
  - PRO: This is an intuitive result.
  - CON: It might need a lot of heading and roll correction past `TKOFF_LVL_ALT`, if yaw has drifted by then.
2. With recalculation: Plane will re-target the direction where it find itself upon `TKOFF_LVL_ALT`.
  - PRO: It results in the least amount of roll/yaw input possible.
  - CON: It might end up in a direction other than where the pilot placed it on the runway.

I think I'm a fan of No2. If the user **really** wants a straight takeoff, they can opt for designing an AUTO mission.

Hwurzburg:
I think we need to be consistent....we already lock heading for AUTOLAND and for STEERING upon reaching a ground speed....which is almost always the runway heading for a rolling takeoff or the initial hand launch direction....I would not change it upon reaching LVL_ALT

Georacer:
I seem to disagree with @Hwurzburg .
I've marked the ticket as Dev Call topic, so that the maintainers can arbiter.

ArduPlane/mode_takeoff.cpp
```diff
@@ -155,26 +155,10 @@ void ModeTakeoff::update()
...
     }
 #endif
-    // We update the waypoint to follow once we're past TKOFF_LVL_ALT or we
-    // pass the target location. The check for target location prevents us
-    // flying towards a wrong location if we can't climb.
-    // This will correct any yaw estimation errors (caused by EKF reset
-    // or compass interference from max throttle), since it's using position bearing.
-    const float altitude_cm = plane.current_loc.alt - start_loc.alt;
-    if (plane.flight_stage == AP_FixedWing::FlightStage::TAKEOFF
-        && plane.steer_state.hold_course_cd == -1 // This is the enter-once flag.
```
tridge, line 165: if hold_course_cd isn't set at this point in the code it would be very bad I think. I know in nearly every case this code isn't an advantage, but can we really guarantee that hold_course_cd has been set in all wind conditions?

Georacer: I've had a hard time understanding this comment.
Do you mean that there's a chance when:
- starting a TAKEOFF in very strong headwind
where the condition `(plane.is_flying() && (millis() - plane.started_flying_ms > 10000U) && groundspeed > 3)` will pass immediately and thus set `takeoff_mode_setup` early based on a bad groundspeed vector?

If yes, than it is indeed a basis to leave this marked code in.

Hwurzburg: we are very inconsistent in our handling of the initial heading during takeoff:
- mode takeoff uses gndspd thresholds of one value, nav_takeoff and AUTOLAND mode uses others in their code
- mode takeoff initially uses gndcrs if above gndspd threshold to set loiter waypoint, If threshold is not reached then above LVL_ALT check will set it in  gndcrs direction or if flying more than 10s  and below LVL_ALT it will be set if >3m/s gndspd
- nav_takeoff compensates grndcrs heading with integrated ahrs yaw since takeoff start and waits 2s after gnspd threshold to do so, but does not cope with high head wind , only continues level flight
- AUTOLAND just uses gndspd threshold to set heading used for landing approach

I already have a WIP PR to try to reconcile all these to use a consistent gndspd threshold and a failsafe heading lock after LVL_ALT if not obtained. In dev call I committed to try to finish this during the week.

Hwurzburg: see #29104 
I think removing this makes sense after I analyzed the code and did a lot of SITL with headwinds almost the same as the vehicle climb speed (which only can occur if someone is trying to launch in wind that could lead to a flyaway anyway)

rubenp02:
Where does this stand now? I see the file has changed quite a bit but this block is still in, and running the Plane autotest, it's still never called, so I just rebased.

What's a "proper" takeoff supposed to do, reset the target heading after TKOFF_LVL_ALT, or define it as soon as a GPS velocity vector is obtained and then keep it? In my opinion, a decision should be made between the two options, and for clarity and consistency, only the chosen one should appear in the code.

I've had problems in the past with rudder-less aircraft that will drift when the wings are forced level (since they lack any mechanism to change heading except roll) and therefore the takeoff heading ends up being completely different to the launch heading if it gets recalculated at TKOFF_LVL_ALT.

PR #29080, opened 2025-01-15
Plane: Add fwd. throttle cutoff voltage parameter

rubenp02:
# Plane: Add fwd. throttle cutoff voltage parameter

Updated the forward throttle battery voltage compensation feature to disable the throttle entirely when the sag-compensated voltage drops below the new parameter FWD_BAT_THR_CUT.

Key changes:
- Added new parameter FWD_BAT_THR_CUT to control the voltage threshold for this feature. The default value of 0 matches the original behavior of never cutting the throttle due to low voltage.
- Modified forward throttle battery voltage compensation logic in the servos code to cut off the throttle if the resting voltage estimate of the FWD_BAT_IDX battery is under FWD_BAT_THR_CUT.

### DIFF
--- a/ArduPlane/servos.cpp
+++ b/ArduPlane/servos.cpp
@@ -440,6 +440,19 @@ void ParametersG2::FWD_BATT_CMP::update()
 // Apply throttle scale to min and max limits
 void ParametersG2::FWD_BATT_CMP::apply_min_max(int8_t &min_throttle, int8_t &max_throttle) const
 {
+    // Cut off throttle if FWD_BAT_IDX battery resting voltage is below
+    // FWD_THR_CUTOFF_V (if set), to preserve battery life for the electronics
+    // and actuators. Only applies when the battery monitor is working and the
+    // current mode does auto-throttle.
+    if (is_positive(batt_voltage_throttle_cutoff) &&
+        plane.control_mode->does_auto_throttle() && AP::battery().healthy(batt_idx) &&
+        (AP::battery().voltage_resting_estimate(batt_idx) < batt_voltage_throttle_cutoff)) {
+        min_throttle = 0;
+        max_throttle = 0;
+
+        return;
+    }
+
     // return if not enabled
     if (!enabled) {
         return;
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1309,7 +1309,16 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("RNGFND_LND_ORNT", 36, ParametersG2, rangefinder_land_orient, ROTATION_PITCH_270),
 #endif
-    
+
+    // @Param: FWD_BAT_THR_CUT
+    // @DisplayName: Forward throttle cutoff battery voltage
+    // @Description: The estimated battery resting voltage below which the throttle is cut in auto-throttle modes. Measured on the battery used for forward throttle compensation (FWD_BAT_IDX). If set to zero, the throttle will not be cut due to low voltage, allowing the motor(s) to continue running until the battery is depleted. This should be set to the minimum operating voltage of you motor(s) or to a voltage level where minimal thrust is produced, to conserve the remaining battery power for the electronics and actuators.
+    // @Range: 0 35
+    // @Units: V
+    // @Increment: 0.1
+    // @User: Standard
+    AP_GROUPINFO("FWD_BAT_THR_CUT", 37, ParametersG2, fwd_batt_cmp.batt_voltage_throttle_cutoff, 0.0f),
+
     AP_GROUPEND
 };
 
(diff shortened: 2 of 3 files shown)


Hwurzburg:
Needs library commit split

IamPete1:
Have you considered using scripting for this?

IamPete1:
> Would it work as cleanly? This has the advantage that it keeps the aircraft armed.

You could trigger the E-Stop  aux function which will only stop the motors. You could also force override the throttle PWM to 0. Or set the throttle max param to some low value such that you still get a little throttle. 

The nice thing about scripting is that it keeps it very contained, you only have to consider your use case. For example this PR would be very bad on a hovering quadplane, to get it into the main code you need to deal with that case (and I'm sure we will think of some others). If its just a script then that is not a problem. 

Another option could be to support glide as a battery fail safe action. We support this already for RC/GCS failsafe. That means you don't have to touch the throttle suppression code which we have to be very careful with.

IamPete1:
> What do you think is best? Should I just submit this as a script, or should I flesh out this implementation and get it into the main code?

Your welcome carry on with the C++ implementation just be warned that we will be quite picky with failsafes and things touching throttle suppression ;)

Hwurzburg:
this is now an ESC voltage cut-off equivalent, I think...

IamPete1 (commented):
I'm really not a fan of using throttle suppression here. It could cause all sorts of odd stuff, this also needs to use a battery index, just using the first may not be valid. 

How about linking into the existing battery voltage compensation code path. Here:

https://github.com/ArduPilot/ardupilot/blob/25e1701447e582bc378c9cd0c7f3f554f424d2de/ArduPlane/servos.cpp#L408

It has a index parameter already. There is a function for setting min and max here:
 https://github.com/ArduPilot/ardupilot/blob/25e1701447e582bc378c9cd0c7f3f554f424d2de/ArduPlane/servos.cpp#L441C34-L441C47. 

So in this new case you could just set both min and max to 0 if under the threshold.

IamPete1:
> Makes sense but what sort of issues does this cause?

I have no idea off the top of my head, the throttle suppression stuff is used in lots of places. Although I guess your not actually setting the `throttle_suppressed` variable but you do early return before other code that might set it. This part of plane is not very clean it is easy to get unexpected edge cases so we need to be very careful. The `FWD_BATT_CMP` stuff is more contained so its easier to be sure we have not broken other things.

ArduPlane/servos.cpp
```diff
@@ -440,6 +440,14 @@ void ParametersG2::FWD_BATT_CMP::update()
 // Apply throttle scale to min and max limits
 void ParametersG2::FWD_BATT_CMP::apply_min_max(int8_t &min_throttle, int8_t &max_throttle) const
 {
+    // cut off throttle if FWD_BAT_IDX battery is below FWD_THR_CUTOFF_V resting voltage
+    if (AP::battery().voltage_resting_estimate(batt_idx) < batt_voltage_throttle_cutoff) {
```
IamPete1, line 444:
```suggestion
    if (is_positive(batt_voltage_throttle_cutoff) && (AP::battery().voltage_resting_estimate(batt_idx) < batt_voltage_throttle_cutoff)) {
```
Just to be sure that some battery bug where it returns less than zero for resting voltage doesn't cut throttle if the feature is not enabled.

IamPete1 (commented):
Basically looks good. I guess it still works as expected for your use case?

ArduPlane/servos.cpp
```diff
@@ -440,6 +440,15 @@ void ParametersG2::FWD_BATT_CMP::update()
 // Apply throttle scale to min and max limits
 void ParametersG2::FWD_BATT_CMP::apply_min_max(int8_t &min_throttle, int8_t &max_throttle) const
 {
+    // cut off throttle if FWD_BAT_IDX battery is below FWD_THR_CUTOFF_V resting voltage
+    if (is_positive(batt_voltage_throttle_cutoff) &&
```
tridge, line 444: does this apply in manual throttle modes, such as FBWA? It would be nice to be able to switch to FBWA to recover. Maybe check the auto-throttle flag in the mode?

tridge (changes requested):

ArduPlane/servos.cpp
tridge: maybe also check healthy() on the battery index?

IamPete1: FBWA will return true for that case `throttle_passthru_stabilize` is 0 by default. So throttle limits will still apply. You can use `plane.control_mode->does_auto_throttle()`

IamPete1: I think its fine. The idea being in manual throttle mode its up to the pilot to decide what to do. In this case you could well be in the "very low battery" code path that turns off the normal battery compensation in anycase, https://github.com/ArduPilot/ardupilot/blob/09c8ad81bb1624fb049967ab2987dc0a2aafd0b0/ArduPlane/servos.cpp#L419-L422

IamPete1 (approved):

tridge (approved):

PR #29081, opened 2025-01-15
AP_Scripting:Add disarm on low copter flipping LUA example

Hwurzburg:
Contribution from a user who wanted to share it

### DIFF
--- a/libraries/AP_Scripting/examples/disarm_on_flip.lua
+++ b/libraries/AP_Scripting/examples/disarm_on_flip.lua
@@ -1,0 +1,38 @@
+--Example For disarming the drone when roll or pitch exceeds n degrees below n altitude.
+--used to hand recover quads and to prevent prop strikes in case of low altitude collision
+--Must have a downward facing rangefinder
+--Made by shroxgh
+
+local rotation_downward = 25
+local ALT_THRESH = 250 -- Altitude in which in centimeters 
+local ANGLE_THRESH = 80 -- Roll/Pitch angle threshold in degrees
+
+function update()
+    local sensor_count = rangefinder:num_sensors()
+    
+
+    -- Check altitude from downward rangefinder
+    if rangefinder:has_data_orient(rotation_downward) then
+        local distance_cm = rangefinder:distance_cm_orient(rotation_downward)
+        if distance_cm and distance_cm < ALT_THRESH then
+            -- Check roll and pitch
+            local roll = ahrs:get_roll() * 57.2958 -- Convert radians to degrees
+            local pitch = ahrs:get_pitch() * 57.2958 -- Convert radians to degrees
+
+           
+
+            if math.abs(roll) > ANGLE_THRESH or math.abs(pitch) > ANGLE_THRESH then
+                -- Disarm the drone
+                arming:disarm() -- 
+                gcs:send_text(0, "Disarming: Roll or Pitch exceeded limits below altitude threshold!")
+            end
+        end
+    else
+        gcs:send_text(1, "No valid data from downward rangefinder!")
+    end
+
+    -- Schedule the next update in milliseconds
+    return update, 10
+end
+
+return update()


libraries/AP_Scripting/examples/disarm_on_flip.lua
```diff
@@ -0,0 +1,38 @@
...
+    
+
+    -- Check altitude from downward rangefinder
+    if rangefinder:has_data_orient(rotation_downward) then
+        local distance_cm = rangefinder:distance_cm_orient(rotation_downward)
+        if distance_cm and distance_cm < ALT_THRESH then
+            -- Check roll and pitch
+            local roll = ahrs:get_roll() * 57.2958 -- Convert radians to degrees
+            local pitch = ahrs:get_pitch() * 57.2958 -- Convert radians to degrees
+
```
rmackay9, line 21: no big deal but maybe we could remove these extra blank lines

libraries/AP_Scripting/examples/disarm_on_flip.lua
```diff
@@ -0,0 +1,38 @@
...
+           
+
+            if math.abs(roll) > ANGLE_THRESH or math.abs(pitch) > ANGLE_THRESH then
+                -- Disarm the drone
+                arming:disarm() --
```
rmackay9, line 26: extra "--" at the end of this line that can probably be deleted

libraries/AP_Scripting/examples/disarm_on_flip.lua
```diff
@@ -0,0 +1,38 @@
+--Example For disarming the drone when roll or pitch exceeds n degrees below n altitude.
+--used to hand recover quads and to prevent prop strikes in case of low altitude collision
+--Must have a downward facing rangefinder
+--Made by shroxgh
+
+local rotation_downward = 25
+local ALT_THRESH = 250 -- Altitude in which in centimeters 
+local ANGLE_THRESH = 80 -- Roll/Pitch angle threshold in degrees
+
```
rmackay9, line 9: Not a blocker but normally we print a message to tell the user that the script has been loaded.

gcs:send_text(1, "copter-disarnm-on-flip script loaded")

libraries/AP_Scripting/examples/disarm_on_flip.lua
```diff
@@ -0,0 +1,38 @@
...
+           
+
+            if math.abs(roll) > ANGLE_THRESH or math.abs(pitch) > ANGLE_THRESH then
+                -- Disarm the drone
+                arming:disarm() -- 
+                gcs:send_text(0, "Disarming: Roll or Pitch exceeded limits below altitude threshold!")
+            end
+        end
+    else
+        gcs:send_text(1, "No valid data from downward rangefinder!")
```
rmackay9, line 31: This is going to spam the user at 100hz so perhaps add a "return update(), 1000" to slow it down

rmackay9:
thanks for adding this.  I've made a couple of comments but it's pretty close to merge-able.  Could we rename it to start with "copter-"?  I think the script is only valid for Copter so having the vehicle at the front makes it more likely that users who can use it will find it and others can quickly ignore it

Hwurzburg:
@rmackay9 I dont think this actually works, right? unless somehow the RF is updating so slowly that it still shows close to ground as it rotates toward the sky to get the attitude limits evoked...
perhaps, we could make a slowly updating (at least on alt increase, fast on decrease) baro tracking var that is updated via RF so that the alt limit is still active if the attitude is not pointing down any more?

peterbarker:
Ping @rmackay9 - Henry wanted clarification on your queries.

@Hwurzburg you could probably attack some of the other things that Randy raised...

rmackay9:
Hi @Hwurzburg,

Yes, you're right about the rangefinder.  My guess is that the user found it worked because they had the lean angles set very low.

maintaining a good estimate of the alt above ground sounds like a good idea.  We have some code in Copter's cpp code that does this but it's quite simple.  https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/mode.cpp#L605

PR #29083, opened 2025-01-16
Global: fix SCHED_TASK rate_hz name

robertlong13:
Confused me for a minute. Figured it's worth a quick rename.

A couple places got missed during the original refactor ages ago, and Copter got accidentally reverted to the old name later.


peterbarker (approved):
Zero-compiler-output change

PR #29084, opened 2025-01-16
Add and use AP_MotorMask

peterbarker:
@IamPete1 suggested on https://github.com/ArduPilot/ardupilot/pull/29020 that the output mask had to be adjusted for 32 motors.

This PR is two phases - the first moves the API to use a typedef rather than uint16_t for the mask.  It typedefs that new type to a class, so that a type error is raised if anyone tries to pass a uint16_t into the function.

The second patch just typedefs it back to 16, making this PR a no-compiler-output change:
```
```

This will allow us to move to 32 (or 64?! as hinted at in a DevCall) relatively easily.

### DIFF
--- a/libraries/AP_Motors/AP_Motors_Class.h
+++ b/libraries/AP_Motors/AP_Motors_Class.h
@@ -52,17 +52,18 @@
 // motor update rate
 #define AP_MOTORS_SPEED_DEFAULT     490 // default output rate to the motors
 
-class AP_MotorMask16 {
-public:
-    AP_MotorMask16() {}
-    AP_MotorMask16(const short int value) { mask = value; }
-    bool operator &(uint16_t value) const { return mask & value; }
-    void operator =(uint16_t value) { mask = value; }
-    void operator |=(uint16_t value) { mask |= value; }
-    uint16_t mask;
-};
+// a class that AP_MotorMask can be typedef'd to to ensure type-safety:
+// class AP_MotorMask16 {
+// public:
+//     AP_MotorMask16() {}
+//     AP_MotorMask16(const short int value) { mask = value; }
+//     bool operator &(uint16_t value) const { return mask & value; }
+//     void operator =(uint16_t value) { mask = value; }
+//     void operator |=(uint16_t value) { mask |= value; }
+//     uint16_t mask;
+// };
 
-typedef AP_MotorMask16 AP_MotorMask;
+typedef uint16_t AP_MotorMask;
 
 /// @class      AP_Motors
 class AP_Motors {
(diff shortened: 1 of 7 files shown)


libraries/AP_Motors/AP_Motors_Class.h
```diff
@@ -52,6 +52,19 @@
 // motor update rate
 #define AP_MOTORS_SPEED_DEFAULT     490 // default output rate to the motors
 
+// a class that AP_MotorMask can be typedef'd to to ensure type-safety:
```
rmackay9, line 55: I'm sure it's on purpose but it's a bit odd to have commented out code like this

andyp1per:
Not sure I see the point or agree with this. Seems like over-complexifying for little gain.

peterbarker:
> The problem this fixes is that if we try and cast a int16 form the mask parameters to a uint32 we get the wrong answer? I was expecting to just change all the masks to uint32. Currently we cast from int16 to uint16 which comes out right I guess. I suppose if we did int16 -> uint16 -> uint32 that would be OK too?

There are rumours of people requiring a 64-bit mask here.  We probably do not want to use 64-bits on stm32 where we don't have to.  This will give us a compile-time method of changing it.

libraries/AP_Motors/AP_Motors_Class.h
peterbarker: The comment there explains what it's for.  I thought it better to leave it available but commented out so it was easily available.

I could leave it commented in - or remove it.

IamPete1:
> There are rumours of people requiring a 64-bit mask here. We probably do not want to use 64-bits on stm32 where we don't have to. This will give us a compile-time method of changing it.

I don't think this is worth it just for future 64 bit support, there are tens of other bit masks that are currently 32 bit that will have to be updated, but we could stash this away for when we do want to do that. TBH, for the motors example, I don't see there being a significant benefit to actually mixing for 64 motors, you could just have 32 groups of 2. 32 actuators really is a lot for any vehicle.

peterbarker:
> example, I don't see there being a significant benefit to actually mixing for 64 motors, you could just have 32 groups of 2. 32 actuators really is a lot for any vehicle.

..... in which case you'd probably definitely want an abstraction like a real class.

Obviously not a lot of love for this PR, I'll make up a trivial patch to move the mask to 32 bits instead.

peterbarker:
> example, I don't see there being a significant benefit to actually mixing for 64 motors, you could just have 32 groups of 2. 32 actuators really is a lot for any vehicle.

..... in which case you'd probably definitely want an abstraction like a real class.

Obviously not a lot of love for this PR, I've made up a trivial patch to move the mask to 32 bits instead in the 32-motors PR.

PR #29085, opened 2025-01-16
Rename SRV_Channel::Aux_servo_function_t to SRV_Channel::Function

peterbarker:
Not only is the new name shorter, it's also more in-keeping with the way we do enum-class names elsewhere in the code.

Additionally, the word "aux" in the old name is completely misleading - this covers all servo functions, not just "auxiliary" functions!

This is a no-compiler-output-change PR:

```
```

### DIFF
--- a/libraries/SRV_Channel/SRV_Channel.h
+++ b/libraries/SRV_Channel/SRV_Channel.h
@@ -210,10 +210,10 @@ public:
         k_motor31               = 178,
         k_motor32               = 179,
         k_nr_aux_servo_functions         ///< This must be the last enum value (only add new values _before_ this one)
-    } Aux_servo_function_t;
+    } Function;
 
     // check if a function is valid for indexing into functions
-    static bool valid_function(Aux_servo_function_t fn) {
+    static bool valid_function(Function fn) {
         return fn >= k_none && fn < k_nr_aux_servo_functions;
     }
     bool valid_function(void) const {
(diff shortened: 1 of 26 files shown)


rmackay9 (approved):
LGTM.  As always there are some concerns about difficulties doing backports once this is merged but in that case we could consider backporting this as well I guess

IamPete1 (approved):
Looks good, maybe we can remove the `k_` on all the functions in the future too.

peterbarker:
> Looks good, maybe we can remove the `k_` on all the functions in the future too.

I did play around with making this *actually* an enum-class, rather than just looking like one.  It involves inserting `Function::` into a *lot* of places where we use this enumeration!  Presumably we'd do that at the same to as removing the `k_` thing - but you are looking at 1000 patches!

libraries/SRV_Channel/SRV_Channel.h
```diff
@@ -43,8 +43,7 @@ class SRV_Channel {
 
     static const struct AP_Param::GroupInfo var_info[];
 
-    typedef enum
-    {
+    enum class Function : uin8_t {
```
tridge, line 46: needs to be signed, and needs to be a real type

tridge (changes requested):

tridge:
happy with the rename, no enum class change in this PR please

libraries/SRV_Channel/SRV_Channel.h
peterbarker: This was a bad patch, now removed

peterbarker:
I've rebased to resolve conflicts.

Still no compiler output change

PR #24509, opened 2023-08-04
ArduPlane: Fixed formulas for load factor and roll, also corrected _TASmin in AP_TECS accordingly

abaghiyan:

Issue #24320, opened 2023-07-15
Issues related to formula in Plane::update_load_factor


abaghiyan:
## Bug report

**Issue details**

In method [Plane::update_load_factor](https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/Attitude.cpp#L582) to estimate aerodynamic load factor the following formula is used `aerodynamic_load_factor = 1.0f / safe_sqrt(cosf(radians(demanded_roll)));`. In general when there the pitch and roll exist the formula for the aerodynamic load factor should be as follows: `aerodynamic_load_factor = cosf(radians(_ahrs.pitch))/cosf(radians(demanded_roll))`, which in case of turns (in horizontal plane) becomes `aerodynamic_load_factor = 1.0f/cosf(radians(demanded_roll))`. Why square root is used in the existing code. If proposed formulas are accepted shouldn't the code `int32_t roll_limit = degrees(acosf(sq(1.0f / max_load_factor)))*100;` also be changed accordingly (without square and taking into account pitch).

**Version**
4.3.7

**Platform**
Plane

**Airframe type**
Plane airframes

**Hardware type**
Pixhawk

IamPete1:
I think all our maths give the correct results, just that were just doing the sqrt in the wrong spot?

I think we would accept a PR to fix it, remove the sqrt from the load factor and the square from the roll limit cal. We would also need to add a sqrt here in that case:

https://github.com/ArduPilot/ardupilot/blob/bde43f167ef535c1c301c90528d91b4fa0815a91/libraries/AP_TECS/AP_TECS.cpp#L418

Is that something you could do?

Once that is corrected we could think about including the pitch. That will reduce the load factor, so we might be better to ignore to stay on the conservative side.

abaghiyan:
> I think all our maths give the correct results, just that were just doing the sqrt in the wrong spot?
> 
> I think we would accept a PR to fix it, remove the sqrt from the load factor and the square from the roll limit cal. We would also need to add a sqrt here in that case:
> 
> https://github.com/ArduPilot/ardupilot/blob/bde43f167ef535c1c301c90528d91b4fa0815a91/libraries/AP_TECS/AP_TECS.cpp#L418
> 
> Is that something you could do?
> 
> Once that is corrected we could think about including the pitch. That will reduce the load factor, so we might be better to ignore to stay on the conservative side.

[...]

PR #29092, opened 2025-01-16
Plane: Attitude stall prevention adjustments

rubenp02:
# Plane: Attitude stall prevention adjustments

Description
-----------
### [Plane: Use stall speed for roll stall prevention](https://github.com/ArduPilot/ardupilot/pull/29092/commits/6f8f1ae984c23078405f0c64cfa08e338db4d4ea):

Applied the new AIRSPEED_STALL parameter to roll stall prevention, to enable tighter turns at lower airspeeds when this parameter is set. For accurate behavior, it should be set at the stall IAS on level flight at sea level, plus the desired safety margin.

### DIFF
--- a/ArduPlane/Attitude.cpp
+++ b/ArduPlane/Attitude.cpp
@@ -659,7 +659,12 @@ void Plane::update_load_factor(void)
     }
 #endif
 
-    float max_load_factor = sq(smoothed_airspeed / MAX(aparm.airspeed_min, 1));
+    float stall_airspeed_1g = is_positive(aparm.airspeed_stall)
+                                  ? aparm.airspeed_stall
+                                  : aparm.airspeed_min;
+
+    float max_load_factor =
+        sq(smoothed_airspeed / MAX(stall_airspeed_1g, 1));
 
     if (max_load_factor <= 1) {
         // our airspeed is below the minimum airspeed. Limit roll to
@@ -671,7 +676,7 @@ void Plane::update_load_factor(void)
         // load limit. Limit our roll to a bank angle that will keep
         // the load within what the airframe can handle. We always
         // allow at least 25 degrees of roll however, to ensure the
-        // aircraft can be manoeuvred with a bad airspeed estimate. At
+        // aircraft can be manoeuvered with a bad airspeed estimate. At
         // 25 degrees the load factor is 1.1 (10%)
         int32_t roll_limit = degrees(acosf(1.0f / max_load_factor))*100;
         if (roll_limit < 2500) {
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -323,7 +323,7 @@ const AP_Param::Info Plane::var_info[] = {
 
     // @Param: AIRSPEED_STALL
     // @DisplayName: Stall airspeed
-    // @Description: If stall prevention is enabled this speed is used to calculate the minimum airspeed while banking. If this is set to 0 then the stall speed is assumed to be the minimum airspeed speed. Typically set slightly higher then true stall speed. Value is as an indicated (calibrated/apparent) airspeed.
+    // @Description: If stall prevention is enabled this speed is used for its calculations. If set to 0 then AIRSPEED_MIN is used instead. Typically set slightly higher than the true stall speed. Value is as an indicated (calibrated/apparent) airspeed.
     // @Units: m/s
     // @Range: 5 75
     // @User: Standard


IamPete1:
There is some pre-existing work here that we never did merge, but I think your proposed fix is similar. https://github.com/ArduPilot/ardupilot/issues/24320 https://github.com/ArduPilot/ardupilot/pull/24509

Re the second commit, I think swapping stall to speed if set is fine. Less keen on the 0.85 scale factor on airspeed min, the vast majority have no idea what there stall speed is, so we need to be conservative. I think if you do make that change you would need to remove the correction that is done in TECS here because the load factor would be lower already: https://github.com/ArduPilot/ardupilot/blob/25e1701447e582bc378c9cd0c7f3f554f424d2de/libraries/AP_TECS/AP_TECS.cpp#L418

IamPete1:
I felt bad we had overlooked that original PR for so long so I have rebased it here: https://github.com/ArduPilot/ardupilot/pull/29101

Do you agree with all the changes there? Once that is in you could rebase this to bring in the use of the stall speed param.

libraries/AP_TECS/AP_TECS.h
```diff
@@ -77,6 +77,13 @@ class AP_TECS {
         return _TAS_dem_adj / _ahrs.get_EAS2TAS();
     }
 
+    // Return stall airspeed, adjusted for bank angle. This is based on
+    // AIRSPEED_MIN as a very conservative estimate if AIRSPEED_STALL is not
+    // set.
+    float get_stall_airspeed(void) const {
+        return _TASmin / _ahrs.get_EAS2TAS();
+    }
+
```
Georacer, line 86: I'm not a fan of calling this the stall airspeed. `AIRSPEED_MIN` is explicitly meant to NOT be the stall airspeed, as per its parameter definition.

As a side-note, the resulting maximum loading factor is the maximum **allowed** loading factor, not the maximum loading factor supported by the airframe. That is what would correspond to the stall airspeed.

Georacer (changes requested):

libraries/AP_TECS/AP_TECS.h
Georacer: @rubenp02 I think you're onto something.
But first, some clarifications.

Currently, this is what happens (to my understanding):
1. `Attitude.cpp::update_load_factor()` will calculate a load factor using `AIRSPEED_MIN`.
2. It will then communicate a load factor to TECS.
3. In turn, TECS will increase `_TASmin` (its minimum airspeed limit) by:
  1. `AIRSPEED_MIN*_load_factor`
  2. (or if defined) `max(`AIRSPEED_MIN, AIRSPEED_STALL*_load_factor)`.

If I read your code correctly, you want to take the final `TECS::_TASmin` and feed it back into `update_load_factor()`.
Won't this create a feedback loop, constantly increasing `_TASmin`?
I don't think that's right.

---

I think you have some more errors in your logic:

> Since the idea behind AIRSPEED_STALL is to optionally allow for maneuvers that are closer to the stall speed

This is not correct. The idea is the allow maneuvers at `AIRSPEED_MIN`, as introduced [here](https://github.com/ArduPilot/ardupilot/commit/edaddc043166189938aa7654e9a6e72c06a0b6d0).

> But the min. airspeed increase based on bank angle isn't something that makes any sense for an airspeed that isn't a stall speed.

Indeed. And as far as I understand, the PR linked above fixed that.

---

However, I think I see the merit in your idea.

Are you proposing switching the load factor calculation based on `AIRSPEED_STALL`, when that's available?
If yes, I think it's fine to repeat a similar `if (is_positive(aparm.airspeed_stall))` structure in `update_load_factor()`.
And I understand that will have the effect of not exaggerating the load factor.

Georacer: > Take this with a grain of salt as I'm still learning the codebase, but I don't think that's true. The load factor is based on the demanded roll so it's entirely independent from the airspeed and thus no feedback loop is possible.

Let's see:

`Attitude.cpp` indeed calculates `aerodynamic_load_factor` based on `demanded_roll`. And you also propose
```
float max_load_factor = sq(smoothed_airspeed / MAX(TECS_controller.get_stall_airspeed(), 1));
if (max_load_factor <= 1) {
// our airspeed is below the minimum airspeed. Limit roll to 25 degrees.
[...]
```

Consider the following reasoning:
1. You fly level at $V_c$ that equal to `AIRSPEED_MIN`.
2. A 45deg bang is requested and `aerodynamic_load_factor=1.41`.
3. Currently, `max_load_factor=1`
4. At this point you're not allowed to bank more than 25deg, because of the above code.
5. You bank at 25deg.
6. TECS raises `_TASmin` by `sqrt(1.41)`.
7. [next iteration]
8. The `demanded_roll` is the same, so is `aerodynamic_load_factor`.
9. `max_load_factor` is now even smaller, because the denominator has grown. Thus the roll will keep getting limited at 25deg. This is a steady-state point for this algorithm.
10. Thus, the desired `aerodynamic_load_factor` **does** affect the calculation of `max_load_factor`. Albeit, in a more conservative way, which I think is the opposite of what you wanted to achieve?

Can you check my logic above?
In any case, one of your best arguments is going to be SITL simulations with a before/after comparison.

Georacer: Okay, let's see it!

IamPete1:
> Any news about this and #29101?

@Georacer  Is going to do a review on #29101 but he has been busy.

libraries/AP_TECS/AP_TECS.h
```diff
@@ -77,6 +77,20 @@ class AP_TECS {
         return _TAS_dem_adj / _ahrs.get_EAS2TAS();
     }
 
+    // Return stall airspeed, adjusted for bank angle. This is based on
+    // AIRSPEED_MIN as a very conservative estimate if AIRSPEED_STALL is not
+    // set.
+    float get_stall_airspeed(void) const {
+        return _TASmin / _ahrs.get_EAS2TAS();
+    }
```
IamPete1, line 85: this is unused ?

IamPete1: remove please.

libraries/AP_TECS/AP_TECS.h
```diff
@@ -77,6 +77,13 @@ class AP_TECS {
         return _TAS_dem_adj / _ahrs.get_EAS2TAS();
     }
 
+    // Return stall airspeed on level flight. This is based on AIRSPEED_MIN as a
+    // very conservative estimate if AIRSPEED_STALL is not set.
+    float get_stall_airspeed_1g(void) const {
```
IamPete1, line 82: Is there any reason this needs to be in TECS? Plane can see these parameters directly.

IamPete1: I can't find it anywhere else?

IamPete1: Remove for now, simple PRs get in easier :)

IamPete1 (approved):
LGTM thanks.

ArduPlane/Attitude.cpp
```diff
@@ -659,7 +659,12 @@ void Plane::update_load_factor(void)
     }
 #endif
 
-    float max_load_factor = sq(smoothed_airspeed / MAX(aparm.airspeed_min, 1));
+    float stall_airspeed_1g = is_positive(aparm.airspeed_stall)
```
peterbarker, line 662:
```suggestion
    const float stall_airspeed_1g = is_positive(aparm.airspeed_stall)
```

ArduPlane/Attitude.cpp
```diff
@@ -659,7 +659,12 @@ void Plane::update_load_factor(void)
...
+                                  ? aparm.airspeed_stall
+                                  : aparm.airspeed_min;
+
+    float max_load_factor =
```
peterbarker, line 666:
```suggestion
    const float max_load_factor =
```

PR #29094, opened 2025-01-17
Add support for running IOMCU FW on CubeRedSecondary 

bugobliterator:
Tested working DSHOT, PWM modes and RC Input on CubredSecondary with IOFirmware.

No size changes for older iomcus, checked using sizecompare for both with and without dshot.

Allows updates to the secondary MCU's firmware even after the IOFirmware is flashed.


libraries/AP_HAL_ChibiOS/Scheduler.cpp
```diff
@@ -58,6 +58,14 @@
 extern AP_IOMCU iomcu;
 #endif
 
+#ifdef IOMCU_FW
```
andyp1per, line 61: This is only with H7 I think? Probably should add that to the define.

libraries/AP_IOMCU/AP_IOMCU.cpp
```diff
@@ -100,6 +104,34 @@ void AP_IOMCU::init(void)
     initialised = true;
 }
 
+/*
+  disable the IOMCU operation by erasing iofirmware
```
andyp1per, line 108: I think this should have a different name given the fairly radical function of its operation

libraries/AP_IOMCU/fw_uploader.cpp
```diff
@@ -281,7 +290,7 @@ bool AP_IOMCU::erase()
     debug("erase...");
     send(PROTO_CHIP_ERASE);
     send(PROTO_EOC);
-    return get_sync(10000);
+    return get_sync(20000);
```
andyp1per, line 293: Is this a timeout? Maybe add a comment

libraries/AP_IOMCU/iofirmware/analog.cpp
```diff
@@ -157,3 +159,54 @@ OSAL_IRQ_HANDLER(STM32_ADC1_HANDLER) {
...
+
+AP_HAL::AnalogSource *vrssi;
+
+void adc_init(void)
+{
+    vrssi = hal.analogin->channel(HAL_IOMCU_RSSI_ADC_CHANNEL);
+}
+
+/*
+  capture VSERVO in mV
```
andyp1per, line 176: Comments are wrong. Should add a comment as to why these are empty

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
```diff
@@ -73,9 +103,17 @@ static void setup_rx_dma(hal_uart_driver* uart)
...
+#if defined(STM32H7)
+    dmaStreamSetPeripheral(uart->dmarx, &(uart->usart->RDR));
+#else
     dmaStreamSetPeripheral(uart->dmarx, &(uart->usart->DR));
+#endif
     dmaStreamSetMode(uart->dmarx, uart->dmarxmode    | STM32_DMA_CR_DIR_P2M |
-                     STM32_DMA_CR_MINC | STM32_DMA_CR_TCIE);
+                     STM32_DMA_CR_MINC | STM32_DMA_CR_TCIE
+#if defined(STM32H7)
+                     | (1<<20)
```
andyp1per, line 114: needs a define

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
```diff
@@ -87,11 +125,23 @@ static void setup_tx_dma(hal_uart_driver* uart)
...
+#if defined(STM32H7)
+    dmaStreamSetPeripheral(uart->dmatx, &(uart->usart->TDR));
+#else
     dmaStreamSetPeripheral(uart->dmatx, &(uart->usart->DR));
+#endif
     dmaStreamSetMode(uart->dmatx, uart->dmatxmode    | STM32_DMA_CR_DIR_M2P |
-                     STM32_DMA_CR_MINC | STM32_DMA_CR_TCIE);
+                     STM32_DMA_CR_MINC | STM32_DMA_CR_TCIE
+#if defined(STM32H7)
+                     | (1<<20)
```
andyp1per, line 136: ditto

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
```diff
@@ -147,22 +207,43 @@ static void dma_tx_end_cb(hal_uart_driver *uart)
...
-        (void)uart->usart->DR;  /* SR reset step 2 - clear ORE | FE.*/
-
+        CLEAR_ERRORS(); /* SR reset step 2 - clear ORE | FE.*/
         /* send a line break - this will abort transmission/reception on the other end */
         chSysLockFromISR();
-        uart->usart->SR = ~USART_SR_LBD;
-        uart->usart->CR1 = cr1 | USART_CR1_SBK;
-
+        CLEAR_ISR_FLAG(LBD);
+        SEND_BREAK();
```
andyp1per, line 246: this is a slight change in behaviour because this was a local variable for threading reasons. Ideally would preserve the current semantics

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
```diff
@@ -176,7 +257,7 @@ static void idle_rx_handler(hal_uart_driver *uart)
 
     if ((sr & USART_SR_TC) && (cr1 & USART_CR1_TCIE)) {
         /* TC interrupt cleared and disabled.*/
-        uart->usart->SR &= ~USART_SR_TC;
+        CLEAR_ISR_FLAG(TC);
```
andyp1per, line 260: Might be better to use the MODIFY_REG() macro that you used in bdshot

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
```diff
@@ -214,11 +298,11 @@ static void uart_lld_serve_tx_end_irq(hal_uart_driver *uart, uint32_t flags)
...
 void AP_IOMCU_FW::tx_dma_allocate(Shared_DMA *ctx)
 {
-    hal_uart_driver *uart = &UARTD2;
+    hal_uart_driver *uart = &HAL_IO_FMU_COMMS;
     chSysLock();
     if (uart->dmatx == nullptr) {
-        uart->dmatx = dmaStreamAllocI(STM32_UART_USART2_TX_DMA_STREAM,
-                                        STM32_UART_USART2_IRQ_PRIORITY,
+        uart->dmatx = dmaStreamAllocI(HAL_IO_FMU_COMMS_TX_DMA_STREAM,
+                                        12,  //IRQ Priority
```
andyp1per, line 305: This is bypassing the config, thing it should use the define or put a define above

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
```diff
@@ -447,7 +567,7 @@ void AP_IOMCU_FW::update()
     loop_counter++;
 
     if (do_reboot && (last_ms > reboot_time)) {
-        hal.scheduler->reboot(true);
+        hal.scheduler->reboot(false);
```
andyp1per, line 570: why?

andyp1per (changes requested):
Looking good. Some small changes. Think you need to think about how it looks if someone adds F4 support (which has been requested)

libraries/AP_HAL_ChibiOS/Scheduler.cpp
bugobliterator: it gets removed for other boards that don't use it by compiler. As shown by no size difference.

libraries/AP_IOMCU/iofirmware/analog.cpp
bugobliterator: fixed

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
bugobliterator: I am going to change it back to original as it doesn't really matter in this context.

But the change is actually correct. Firstly a thread cannot interrupt an interrupt, another higher priority interrupt can. Secondly, only bit we want to set in CR1 is Line Break, rest of the bits we don't want to touch. With the older method there's a chance we can end up doing just that.
Basically if there was another interrupt or peripheral event (between where we store cr1 and call SEND_BREAK) where the CR1 is changed we will end up changing it again unknowingly. Fortuntely we don't mess with CR1 outside of init and this interrupt, so it is a non-issue. But in general this can cause issue.

libraries/AP_IOMCU/fw_uploader.cpp
bugobliterator: done

libraries/AP_IOMCU/AP_IOMCU.cpp
bugobliterator: thoughts on name, erase is already taken.

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
bugobliterator: yeah, I was trying to avoid adding yet another define that's going to remain untouched. Will add this to define as well.

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
bugobliterator: fixed

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
bugobliterator: fixed through define and also in UARTDriver

libraries/AP_IOMCU/AP_IOMCU.cpp
bugobliterator: changing it to wipe

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
andyp1per: ok

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
andyp1per: ok sounds reasonable. Probably needs testing on other boards that have iofirmware update issues - particularly Pixhawk6X

andyp1per (approved):
LGTM, just needs testing iofirmware upgrades on some of the other boards (e.g. Pixhawk6X)

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -1903,11 +1914,19 @@ def write_UART_config(self, f):
                 self.error("Need io_firmware.bin in ROMFS for IOMCU")
 
             self.write_defaulting_define(f, 'HAL_WITH_IO_MCU', 1)
-            f.write('#define HAL_UART_IOMCU_IDX %u\n' % len(serial_list))
-            f.write(
-                '#define HAL_UART_IO_DRIVER ChibiOS::UARTDriver uart_io(HAL_UART_IOMCU_IDX)\n'
-            )
-            serial_list.append(self.config['IOMCU_UART'][0])
+            if self.config['IOMCU_UART'][0].startswith('HAL_SERIAL'):
```
tridge, line 1917: if we made AP_IOMCU take a pointer to a AP_HAL::UARTDriver instead of a reference then we could have an AP::iomcu()->set_uart() call that could be made from AP_SerialManager. That would enable runtime configuration.

libraries/AP_Vehicle/AP_Vehicle.cpp
```diff
@@ -356,6 +356,15 @@ void AP_Vehicle::setup()
 #endif
 
 #if AP_SERIALMANAGER_ENABLED
+#if HAL_WITH_IO_MCU && (HAL_UART_IOMCU_IDX < HAL_UART_NUM_SERIAL_PORTS)
+    if (BoardConfig.io_enabled()) {
+        // If IO is enabled, we need to set the protocol to None
+        // as the IO MCU UART is handled by the AP_IOMCU class
+        serial_manager.set_protocol_and_baud(HAL_UART_IOMCU_IDX, AP_SerialManager::SerialProtocol_None, 0);
```
tridge, line 363: SerialProtocol_IOMCU would make more sense

libraries/AP_Vehicle/AP_Vehicle.cpp
```diff
@@ -356,6 +356,15 @@ void AP_Vehicle::setup()
...
+    } else {
+        iomcu.wipe();
```
tridge, line 365: why are we wiping the fw if IO not enabled?

tridge (changes requested):
I'm glad to see this becoming an option, just need to work out how to properly dynamically choose if UART is IOMCU or some other protocol, the method in this PR doesn't look right
also, this does change the size of iofirmware_f103_lowpolh.bin by 8 bytes, need to work out why

libraries/AP_IOMCU/iofirmware/analog.cpp
```diff
@@ -22,6 +22,8 @@
 #include "hal.h"
 #include "analog.h"
 
+#ifdef HAL_DISABLE_ADC_DRIVER
```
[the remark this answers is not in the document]
bugobliterator: no, that is correct. If AP::HAL ADC Driver is disabled analog.cpp handles ADC periph

Otherwise we use hal.analogin

libraries/AP_Vehicle/AP_Vehicle.cpp
bugobliterator: This is to allow flashing a different firmware to secondary. We have to reduce the boot time if IOMCU is enabled to 200ms, which is not long enough for it to be caught over USB. To allow updating Secondary MCU over USB again we need wipe iofirmware.

Tools/AP_Bootloader/support.cpp
```diff
@@ -240,6 +240,17 @@ bool flash_write_buffer(uint32_t address, const uint32_t *v, uint8_t nwords)
     return true;
 }
 
+/*
+  detect if the IOMCU firmware is present
+ */
+#if defined(HAL_BL_IOMCU_FW_DETECT) && HAL_BL_IOMCU_FW_DETECT
+#define HAL_IOMCU_MAGIC 0x937E638A
+uint32_t is_iomcu_fw_present(void)
```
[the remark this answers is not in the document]
bugobliterator: This is so we can reduce the boot time if IOMCU is present to 200ms,. This is done by reading the IOMCU_FW_DETECT magic written to the end of flash. 

I think I am going to add a boot time field to AP_CheckFirmware instead. Was looking for a shortcut but this hasn't turned out to be a clean implementation.

Tools/AP_Bootloader/AP_Bootloader.cpp
```diff
@@ -87,6 +87,12 @@ int main(void)
     bool try_boot = false;
     uint32_t timeout = HAL_BOOTLOADER_TIMEOUT;
 
+#if defined(HAL_BL_IOMCU_FW_DETECT) && HAL_BL_IOMCU_FW_DETECT
+    if (is_iomcu_fw_present()) {
```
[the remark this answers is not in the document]
bugobliterator: Current iofirmware bootloader has 200ms timeout. Also if iofirmware doesn't boot faster than when AP_IOMCU tries to detect it, it will simply try to reflash.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
bugobliterator: I think I am going to do this in a separate PR, as that will require quite a decent amount of change to AP_IOMCU. A separate PR for this change will be much easier to review.

For this PR I can go part of the way and use SerialProtocol and hal.serial instead of current method of declairing a separate serial driver for all iomcu users.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -1052,6 +1052,17 @@ def write_mcu_config(self, f):
         else:
             self.env_vars['IOMCU_FW'] = 0
 
+        if self.get_config('IOMCU_FW', required=False):
+            self.env_vars['IOMCU_FW'] = self.get_config('IOMCU_FW')
+        else:
+            self.env_vars['IOMCU_FW'] = 0
+
+        # check if heater pin defined
+        if 'HEATER' in self.bylabel.keys():
```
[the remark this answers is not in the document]
bugobliterator: This is needed to disable multiple builds for highpol and lowpol heaters. With hwdef without Heaters we were still doing that.

libraries/AP_IOMCU/AP_IOMCU.cpp
bugobliterator: removed completely

bugobliterator:
@tridge I have pushed changes per your comments.

* got rid of the bootloader changes, and using bootloader CRC to verify IOMCU at boot.
* Added option SerialProtocol_IOMCU, currently serialmanager is not doing anything with it.
* moved to using serialxDriver for IOMCU as well.

libraries/AP_Vehicle/AP_Vehicle.cpp
bugobliterator: Not relevant anymore, moved to different procedure to allow updates other than iofirmware.

libraries/AP_Vehicle/AP_Vehicle.cpp
bugobliterator: changed to SerialProtocol_IOMCU

Tools/AP_Bootloader/support.cpp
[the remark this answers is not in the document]
bugobliterator: not relevant anymore, this has been removed now.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -1052,6 +1052,17 @@ def write_mcu_config(self, f):
         else:
             self.env_vars['IOMCU_FW'] = 0
 
+        if self.get_config('IOMCU_FW', required=False):
```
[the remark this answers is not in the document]
bugobliterator: fixed

libraries/AP_IOMCU/iofirmware/iofirmware.cpp
tridge: still pending?

libraries/AP_HAL_ChibiOS/hwdef/CubeRedSecondary-IO/hwdef.dat
```diff
@@ -0,0 +1,170 @@
...
+
+define HAL_LED_ON 0
+
+define IOMCU_FW TRUE
+define AP_FASTBOOT_ENABLED 1
+IOMCU_FW 1
+
+define HAL_USE_RTC FALSE
+define HAL_NO_UARTDRIVER TRUE
+define HAL_LOGGING_ENABLED 0
```
peterbarker, line 35: Why are these in here?  These are the default for io firmware... (defaults_periph.h)

libraries/AP_HAL_ChibiOS/hwdef/CubeRedSecondary-IO/hwdef.dat
```diff
@@ -0,0 +1,170 @@
...
+define STM32_SERIAL_USE_USART3 TRUE
+define STM32_SERIAL_USE_USART6 TRUE
+
+define HAL_IOMCU_RCIN_SERIAL_DRIVER SD4
+define HAL_IOMCU_DSM_SERIAL_DRIVER SD3
+define HAL_IOMCU_SBUS_OUT_SERIAL_DRIVER SD6
+define HAL_GPIO_PIN_SPEKTRUM_OUT PAL_LINE(GPIOC,11U)
+
+
+define AP_NETWORKING_BACKEND_PPP 1
```
peterbarker, line 148: This is odd?

libraries/AP_IOMCU/iofirmware/rc.cpp
```diff
@@ -27,27 +27,57 @@
...
+// listen for parity errors on sbus out input
+static event_listener_t serial_rcin_listener;
 
-static uint8_t sd3_config;
+static uint8_t sbus_out_config;
+
+#define STR(x) XSTR(x)
+#define XSTR(x) #x
+
+
```
peterbarker, line 68: Unused?

libraries/AP_IOMCU/fw_uploader.cpp
```diff
@@ -417,12 +429,13 @@ bool AP_IOMCU::verify_rev3(uint32_t fw_size_local)
...
     if (sum != crc) {
         debug("CRC wrong: received: 0x%x, expected: 0x%x", (unsigned)crc, (unsigned)sum);
+        get_sync();
         return false;
     }
 
     crc_is_ok = true;
 
-    return true;
+    return get_sync();
```
tridge, line 438: ifdef

peterbarker (changes requested):

libraries/AP_HAL_ChibiOS/Scheduler.cpp
```diff
@@ -95,6 +95,14 @@ THD_WORKING_AREA(_monitor_thread_wa, MONITOR_THD_WA_SIZE);
 #define AP_HAL_CHIBIOS_IN_EXPECTED_DELAY_WHEN_NOT_INITIALISED 1
 #endif
 
+#if defined(STM32H7) && defined(IOMCU_FW)
```
peterbarker, line 98:
```suggestion
#if AP_REBOOT_ON_MEMGUARD_ERROR
```
... and default it off and define to true in the hwdef.

bugobliterator:
> i'd like to see binary cmp showing no change in IO fw builds

@tridge I did a binary comparision for both iomcu and iomcu-dshot firmware. The difference is only a single byte (1 -> 0) , which is related to a simple change of calling reboot with argument false instead of previously true here:

https://github.com/ArduPilot/ardupilot/pull/29094/files#diff-31df3802edd794b59d66393f1064d74590190327e6423b2ab743e97067afe4f6R102

libraries/AP_IOMCU/fw_uploader.cpp
```diff
@@ -281,7 +293,7 @@ bool AP_IOMCU::erase()
     debug("erase...");
     send(PROTO_CHIP_ERASE);
     send(PROTO_EOC);
-    return get_sync(10000);
+    return get_sync(20000); // 20s timeout for erase
```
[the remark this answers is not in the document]
bugobliterator: done

libraries/AP_IOMCU/fw_uploader.cpp
bugobliterator: done

libraries/AP_IOMCU/fw_uploader.cpp
```diff
@@ -417,12 +429,13 @@ bool AP_IOMCU::verify_rev3(uint32_t fw_size_local)
     /* compare the CRC sum from the IO with the one calculated */
     if (sum != crc) {
         debug("CRC wrong: received: 0x%x, expected: 0x%x", (unsigned)crc, (unsigned)sum);
+        get_sync();
```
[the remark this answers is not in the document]
bugobliterator: fixed

libraries/AP_IOMCU/fw_uploader.cpp
bugobliterator: fixed

libraries/AP_IOMCU/fw_uploader.cpp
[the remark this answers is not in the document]
bugobliterator: fixed

libraries/AP_IOMCU/fw_uploader.cpp
```diff
@@ -84,11 +88,11 @@ bool AP_IOMCU::upload_fw(void)
...
     }
 
-    uint32_t bl_rev;
-    ret = get_info(INFO_BL_REV, bl_rev);
-
-    if (!ret) {
-        debug("Err: failed to contact bootloader");
+    uint32_t bl_rev, unused;
+    if (!get_info(INFO_BL_REV, bl_rev) ||
+        !get_info(INFO_BOARD_ID, unused) ||
```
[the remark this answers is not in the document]
bugobliterator: done

libraries/AP_HAL_ChibiOS/hwdef/CubeRedSecondary-IO/hwdef.dat
bugobliterator: They are not actually, only HAL_LOGGING_ENABLED is. In anycase, I copied this from iomcu/hwdef.inc

libraries/AP_HAL_ChibiOS/Scheduler.cpp
bugobliterator: I think that can be a bit deceptive If I do it the way you mentioned. 

As the normal AP_memory_guard_error is defined inside AP_InternalError. This is actually a special case for IOMCU_FW running  on STM32H7.

libraries/AP_IOMCU/iofirmware/rc.cpp
bugobliterator: fixed

libraries/AP_HAL_ChibiOS/hwdef/CubeRedSecondary-IO/hwdef.dat
bugobliterator: fixed

libraries/AP_HAL_ChibiOS/hwdef/CubeRedSecondary-IO/hwdef.dat
peterbarker: > They are not actually, only HAL_LOGGING_ENABLED is. In anycase, I copied this from iomcu/hwdef.inc

`HAL_NO_UARTDRIVER` no longer exists.

If `HAL_USE_RTC` needs a default to "off" for IOMCU (and it does!) that default should go into `defaults_iofirmware.h`.  The HAL_LOGGING_ENABLED in `iomcu/hwdef.inc` is redundant as it already appears in `defaults_iofirmware.h`

libraries/AP_HAL_ChibiOS/hwdef/CubeRedSecondary-IO/hwdef.dat
```diff
@@ -0,0 +1,166 @@
...
+define HAL_IOMCU_RCIN_SERIAL_DRIVER SD4
+define HAL_IOMCU_DSM_SERIAL_DRIVER SD3
+define HAL_IOMCU_SBUS_OUT_SERIAL_DRIVER SD6
+define HAL_GPIO_PIN_SPEKTRUM_OUT PAL_LINE(GPIOC,11U)
+
+# only use pulse input for PPM, other protocols
+# are on serial
+define HAL_RCIN_PULSE_PPM_ONLY
+
+
```
peterbarker, line 150: This is unused.

Perhaps it would be better to create a `CubeRedSecondary/hwdef.inc` so we're not making existing problems worse?

Note this would be the first time sharing hwdef information between iomcu and a normal board, so it might not work.

libraries/AP_HAL_ChibiOS/hwdef/common/common_mixf.ld
```diff
@@ -74,6 +74,7 @@ SECTIONS
         *(.ap_romfs*)
         /* moving GCS_MAVLink to ext_flash */
         lib/lib*.a:GCS*.*(.text* .rodata*)
+        lib/lib*.a:tables.*(.text* .rodata*) /* tables for AP_Declination */
```
peterbarker, line 77: We already have patches for this in master - they're in the libraries themselves rather than hidden in the ld scripts

libraries/AP_IOMCU/AP_IOMCU.cpp
```diff
@@ -1092,10 +1096,14 @@ void AP_IOMCU::send_rc_protocols()
 /*
   check ROMFS firmware against CRC on IOMCU, and if incorrect then upload new firmware
  */
+#ifndef HAL_IOMCU_FW_FLASH_SIZE
```
peterbarker, line 1099: It would obviously be nicer to always get this from the build / embedding system

Not really needed for this PR.

libraries/AP_IOMCU/fw_uploader.cpp
```diff
@@ -60,13 +60,21 @@
...
+#endif
+
 /*
   upload a firmware to the IOMCU
  */
 bool AP_IOMCU::upload_fw(void)
 {
     // set baudrate for bootloader
-    uart.begin(115200, 256, 256);
+    uart.begin(HAL_IOMCU_BOOTLOADER_BAUDRATE, 256, 256);
```
peterbarker, line 77:
```suggestion
    uart.begin(AP_IOMCU_BOOTLOADER_BAUDRATE, 256, 256);
```

... similarly for other defines introduced.  `AP_IOMCU` isn't part of the HAL.

libraries/AP_IOMCU/fw_uploader.cpp
```diff
@@ -281,7 +307,11 @@ bool AP_IOMCU::erase()
     debug("erase...");
     send(PROTO_CHIP_ERASE);
     send(PROTO_EOC);
-    return get_sync(10000);
+#if HAL_IOMCU_CHIBIOS_BOOTLOADER
+    return get_sync(20000); // 20s timeout for erase
+#else
+    return get_sync(10000); // 10s timeout for erase
+#endif
```
peterbarker, line 314: Does the erase take longer just because of `HAL_IOMCU_CHIBIOS_BOOTLOADER`?

Should this instead be `HAL_IOMCU_FW_FLASH_SIZE > (0x10000 - 0x1000)`?

libraries/AP_Vehicle/AP_Vehicle.cpp
```diff
@@ -356,6 +356,13 @@ void AP_Vehicle::setup()
 #endif
 
 #if AP_SERIALMANAGER_ENABLED
+#if HAL_WITH_IO_MCU
+    if (BoardConfig.io_enabled()) {
+        // If IO is enabled, we need to set the protocol to None
+        // as the IO MCU UART is handled by the AP_IOMCU class
```
peterbarker, line 362: What happens if we *don't* do this?

peterbarker (changes requested):

libraries/AP_HAL_ChibiOS/hwdef/CubeRedSecondary-IO/hwdef.dat
bugobliterator: sharing hwdef.inc between IO and Secondary would be quite complex actually, except for pins almost everything will need to be disabled or changed.

libraries/AP_HAL_ChibiOS/hwdef/common/common_mixf.ld
bugobliterator: can you please share the link to the PRs?

libraries/AP_IOMCU/fw_uploader.cpp
bugobliterator: the reason is flash size, so I guess it can be changes to be based on that.

libraries/AP_Vehicle/AP_Vehicle.cpp
bugobliterator: there's a typo that it needs to say IOMCU not None

libraries/AP_HAL_ChibiOS/hwdef/common/common_mixf.ld
peterbarker: Showed Sid the code on DevCallEU

tridge (approved):

libraries/AP_IOMCU/fw_uploader.cpp
bugobliterator: fixed

libraries/AP_Vehicle/AP_Vehicle.cpp
bugobliterator: comment removed altogether, as its pretty clear from the code what is happening.

libraries/AP_IOMCU/fw_uploader.cpp
bugobliterator: fixed

libraries/AP_IOMCU/AP_IOMCU.cpp
bugobliterator: ok, will defer it for another time.

PR #29096, opened 2025-01-17
Prop/motor blimp and waypoint navigation

MichelleRos:
Adds propeller/motor-based blimp frame and SCurve-based waypoint navigation.
Also moves some parameters from the main blimp parameter group to the Loiter class.


Blimp/config.h
```diff
@@ -135,5 +135,5 @@
 #endif
 
 #ifndef HAL_FRAME_TYPE_DEFAULT
-#define HAL_FRAME_TYPE_DEFAULT Fins::MOTOR_FRAME_TYPE_AIRFISH
+#define HAL_FRAME_TYPE_DEFAULT Fins::MOTOR_FRAME_TYPE_FISHBLIMP
```
Hwurzburg, line 138: where is this even used?

Blimp/Fins.h
```diff
@@ -14,12 +16,13 @@ class Fins
 
     enum motor_frame_class {
         MOTOR_FRAME_UNDEFINED = 0,
-        MOTOR_FRAME_AIRFISH = 1,
-    };
-    enum motor_frame_type {
-        MOTOR_FRAME_TYPE_AIRFISH = 1,
+        MOTOR_FRAME_FISHBLIMP = 1,
```
Hwurzburg, line 19: MOTOR_FRAME_TYPE_FISHBLIMP???

MichelleRos: Changed the variable to motor_frame instead - there's actually only one frame identifier as I don't have the frame_type + frame_class thing that Copter and other vehicles have.

Blimp/Blimp.cpp
```diff
@@ -213,7 +213,14 @@ void Blimp::one_hz_loop()
...
 
-    blimp.pid_pos_yaw.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
+    loiter->pid_vel_x.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
+    loiter->pid_vel_y.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
+    loiter->pid_vel_z.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
+    loiter->pid_vel_yaw.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
+    loiter->pid_pos_x.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
+    loiter->pid_pos_y.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
+    loiter->pid_pos_z.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
+    loiter->pid_pos_yaw.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
```
peterbarker, line 223:
```suggestion
    auto& {
        pid_vel_x,
        pid_vel_y,
        pid_vel_z,
   } pids[];
   const auto loop_rate_hz = AP::scheduler().get_filtered_loop_rate_hz();
   for (auto &pid : pids) {
       pid..set_notch_sample_rate(filtered_loop_rate_hz);
   }
```

may need a couple of lists depending on the type of the pid (e.g. 1d vs 2d)

Syntax might not be quite right, but concept is to avoid the duplication.

This change isn't vital, it just makes things neater/easier to maintain

Blimp/Blimp.cpp
```diff
@@ -225,9 +232,14 @@ void Blimp::read_AHRS(void)
     IGNORE_RETURN(ahrs.get_relative_position_NED_origin_float(pos_ned));
 
     vel_yaw = ahrs.get_yaw_rate_earth();
-    Vector2f vel_xy_filtd = vel_xy_filter.apply({vel_ned.x, vel_ned.y});
-    vel_ned_filtd = {vel_xy_filtd.x, vel_xy_filtd.y, vel_z_filter.apply(vel_ned.z)};
-    vel_yaw_filtd = vel_yaw_filter.apply(vel_yaw);
+
+    if (g2.frame_class.get() == Fins::MOTOR_FRAME_FISHBLIMP) {
```
peterbarker:
```suggestion
    switch (Fins(g2.frame_class.get())) {
    case Fins::MOTOR_FRAME_FISHBLIMP:
```

... and add entries for all frames here, even if most of them just fall through to what's in the `else` clause.

This makes users that add a frame consider what they need to do here.

peterbarker: Consider doing this somewhere outside `read_AHRS`.  I doubt the frame class doesn't require a reboot to have effect.  You don't want to be changing these values filters when the user changes the frame, just set them when dealing with the frame class setup.

Blimp/Fins.cpp
```diff
@@ -32,26 +33,57 @@ Fins::Fins(uint16_t loop_rate) :
...
+        case Fins::MOTOR_FRAME_FISHBLIMP:
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Setting up FishBlimp.");
+            setup_fins();
+            break;
+        case Fins::MOTOR_FRAME_FOUR_MOTOR:
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Setting up FourMotor.");
+            setup_motors();
+            break;
+        default:
+            GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "ERROR: Wrong frame class.");
```
peterbarker:
```suggestion
            GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "ERROR: Bad frame class.");
```

Blimp/Fins.h
```diff
@@ -1,25 +1,28 @@
...
 #pragma once
 #include <AP_Notify/AP_Notify.h>
+#include <AP_AHRS/AP_AHRS.h>
 
 extern const AP_HAL::HAL& hal;
 
-#define NUM_FINS 4 //Current maximum number of fins that can be added.
-#define RC_SCALE 1000
+#define NUM_FINS 4 //Current maximum number of fins or motors that can be added.
+#define RC_SCALE 1000 //Input and output scaling
```
peterbarker:
```suggestion
#define INPUT_AND_OUTPUT_SCALING 1000
```

Blimp/Fins.cpp
```diff
@@ -84,12 +126,35 @@ void Fins::output()
     blimp.Write_FINI(right_out, front_out, down_out, yaw_out);
 #endif
 
+    if (!is_equal(float(blimp.g.stream_rate), 0.0f) && AP_HAL::millis() % int((1 / blimp.g.stream_rate) * 1000) < 30){
```
peterbarker: At least add a comment that this is debug - or people might start depending on these values being emitted.

Blimp/Fins.cpp
```diff
@@ -84,12 +126,35 @@ void Fins::output()
...
+
+    switch (_frame) {
+        case Fins::MOTOR_FRAME_FISHBLIMP:
+            output_fins();
+            break;
+        case Fins::MOTOR_FRAME_FOUR_MOTOR:
+            output_motors();
+            break;
+        default:
+            GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "ERROR: Wrong frame class.");
```
peterbarker:
```suggestion
            GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "ERROR: Bad frame class.");
```

Blimp/Fins.cpp
```diff
@@ -145,3 +222,18 @@ void Fins::output_min()
     yaw_out   = 0;
     Fins::output();
 }
+
+const char* Fins::get_frame_string()
+{
+    switch (_frame) {
+        case Fins::MOTOR_FRAME_FISHBLIMP:
+            return "FISHBLIMP";
+            break;
```
peterbarker:
```suggestion
```

similarly elsewhere

Blimp/GCS_MAVLink_Blimp.cpp
```diff
@@ -95,7 +95,9 @@ void GCS_MAVLINK_Blimp::send_pid_tuning()
         PID_SEND::POSX,
         PID_SEND::POSY,
         PID_SEND::POSZ,
-        PID_SEND::POSYAW
+        PID_SEND::POSYAW,
+        PID_SEND::LVLRLL,
+        PID_SEND::LVLPIT
```
peterbarker:
```suggestion
        PID_SEND::LVLPIT,
```

Blimp/Log.cpp
```diff
@@ -79,14 +79,16 @@ struct PACKED log_Control_Tuning {
...
 {
-    logger.Write_PID(LOG_PIVN_MSG, pid_vel_xy.get_pid_info_x());
-    logger.Write_PID(LOG_PIVE_MSG, pid_vel_xy.get_pid_info_y());
-    logger.Write_PID(LOG_PIVD_MSG, pid_vel_z.get_pid_info());
-    logger.Write_PID(LOG_PIVY_MSG, pid_vel_yaw.get_pid_info());
-    logger.Write_PID(LOG_PIDN_MSG, pid_pos_xy.get_pid_info_x());
-    logger.Write_PID(LOG_PIDE_MSG, pid_pos_xy.get_pid_info_y());
-    logger.Write_PID(LOG_PIDD_MSG, pid_pos_z.get_pid_info());
-    logger.Write_PID(LOG_PIDY_MSG, pid_pos_yaw.get_pid_info());
+    logger.Write_PID(LOG_PIVN_MSG, loiter->pid_vel_x.get_pid_info());
```
peterbarker, line 82: Consider a structure / loop over structure entries here, too

Blimp/Loiter.cpp
```diff
@@ -1,182 +1,1015 @@
...
+    // @Description: Z axis position target notch filter index
+    // @Range: 1 8
+    // @User: Advanced
+
+    // @Param: POSZ_NEF
+    // @DisplayName: Z axis position error notch filter index
+    // @Description: Z axis position error notch filter index
+    // @Range: 1 8
+    // @User: Advanced
+    AP_SUBGROUPINFO(pid_pos_z, "POSZ_", 6, Loiter, AC_PID),
```
peterbarker, line 654: Are you *sure* you want these position-control PIDs inside your loiter object?  It will make it harder to re-use the same tune across modes that are similar.

Blimp/Loiter.cpp
```diff
@@ -1,182 +1,1015 @@
...
+    //Just for the sake of clarity...
+    Vector3f target_vel_bf = target_vel_ef;
+
+    //New value for scaler
+    float scaler_x_n = 1;
+    float scaler_y_n = 1;
+    float scaler_z_n = 1;
+    float scaler_yaw_n = 1;
+
+    if ((Fins::motor_frame)blimp.g2.frame_class.get() == Fins::MOTOR_FRAME_FISHBLIMP) {
```
peterbarker: `switch`

... just generally, if you ever feel the need to `if` on an enumeration you should *probably* instead use a `switch`.  It helps future you because you have to come in and make a decision, even if that decision is, "it's just like all of these others".

Blimp/Loiter.cpp
```diff
@@ -187,7 +1020,7 @@ void Loiter::run_vel(Vector3f& target_vel_ef, float& target_vel_yaw, Vector4b ax
         blimp.motors->right_out = 0;
     } else if (axes_disabled.y);
     else {
```
peterbarker, line 1017: This is a bit weird!

Blimp/Loiter.cpp
```diff
@@ -203,8 +1036,10 @@ void Loiter::run_vel(Vector3f& target_vel_ef, float& target_vel_yaw, Vector4b ax
     }
 
 #if HAL_LOGGING_ENABLED
-    AC_PosControl::Write_PSCN(0.0, 0.0, blimp.pos_ned.x * 100.0, 0.0, target_vel_ef_c.x * 100.0, blimp.vel_ned_filtd.x * 100.0, 0.0, 0.0, 0.0);
-    AC_PosControl::Write_PSCE(0.0, 0.0, blimp.pos_ned.y * 100.0, 0.0, target_vel_ef_c.y * 100.0, blimp.vel_ned_filtd.y * 100.0, 0.0, 0.0, 0.0);
-    AC_PosControl::Write_PSCD(0.0, 0.0, -blimp.pos_ned.z * 100.0, 0.0, -target_vel_ef_c.z * 100.0, -blimp.vel_ned_filtd.z * 100.0, 0.0, 0.0, 0.0);
+if(log) {
```
peterbarker:
```suggestion
if (log) {
```

Blimp/mode.cpp
```diff
@@ -180,6 +185,22 @@ bool Mode::is_disarmed_or_landed() const
     return false;
 }
 
+void Mode::yaw_forward()
+{
+    const float speed_sq = blimp.vel_ned_filtd.xy().length_squared();
+    // Make sure the blimp is keeping up with the target
+    const bool close_yaw = fabsf(wrap_PI(target_yaw-ahrs.get_yaw())) < (loiter->max_pos_yaw*blimp.loiter->pos_lag);
+    if (blimp.position_ok() && (speed_sq > sq(g.wp_yaw_min_vel)) && close_yaw) {
```
peterbarker: Consider using an early-exit pattern here (you can check position before doing *any* work in this method)

Blimp/mode.h
```diff
@@ -344,3 +351,126 @@ class ModeRTL : public Mode
...
+protected:
+
+    const char *name() const override
+    {
+        return "AUTO";
+    }
+    const char *name4() const override
+    {
+        return "AUTO";
+    }
```
peterbarker:
```suggestion
    const char *name() const override { return "AUTO"; }
    const char *name4() const override { return "AUTO"; }
```

Blimp/mode_auto.cpp
```diff
@@ -0,0 +1,222 @@
...
+    return ret;
+}
+
+Vector3p ModeAuto::vec_from_loc(const Location& loc)
+{
+    Vector3p vec;
+    if(loc.get_vector_from_origin_NEU(vec)){
+        vec.x = vec.x * 0.01;
+        vec.y = vec.y * 0.01;
+        vec.z = - vec.z * 0.01;
```
peterbarker:
```suggestion
```

This is equivalent.

All this method does is throw away the return value of the `loc.` call.  Doesn't even wish the caller well.

Blimp/mode_auto.cpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+    return vec;
+}
+
+Vector3p ModeAuto::vec_from_cmd(const AP_Mission::Mission_Command& cmd, const Location& default_loc){
```
peterbarker: Consider making this `bool` and `WARN_IF_UNUSED`.

Blimp/mode_auto.cpp
```diff
@@ -0,0 +1,222 @@
...
+    origin = destination;
+    destination = vec_from_cmd(cmd, blimp.current_loc);
+
+    scurve_prev_leg = scurve_this_leg;
+    scurve_this_leg = scurve_next_leg;
+    scurve_next_leg.init();
+
+    scurve_this_leg_origin = origin;
+
+    if(!mission_started) {
```
peterbarker:
```suggestion
    if (!mission_started) {
```

similarly elsewhere

Blimp/mode_auto.cpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+
+    AP_Mission::Mission_Command next_cmd;
+    if (!mission.get_next_nav_cmd(cmd.index+1, next_cmd)) {
+        fast_wp = false;
+        return;
+    }
+    const Location dest_loc = loc_from_cmd(cmd, blimp.current_loc);
+    const Vector3p next_dest = vec_from_cmd(next_cmd, dest_loc);
```
peterbarker, line 178: If these calls fail you should be able to handle it here.

Blimp/mode_auto.cpp
```diff
@@ -0,0 +1,222 @@
...
+{
+    const float dt = blimp.scheduler.get_last_loop_time_s();
+    bool s_finished = false;
+    if (blimp.loiter->target_within(g.wp_rad*0.5)){
+        s_finished = scurve_this_leg.advance_target_along_track(scurve_prev_leg, scurve_next_leg, g.wp_rad, g.wp_accel, fast_wp, dt, scurve_this_leg_origin, target_vel, target_accel);
+        target_pos = Vector3f{(float)scurve_this_leg_origin.x,(float)scurve_this_leg_origin.y, (float)scurve_this_leg_origin.z};;
+        scurve_this_leg_origin = origin;
+    }
+
+    if (s_finished) {
```
peterbarker: Much pointless nesting here

Blimp/mode_land.cpp
```diff
@@ -3,14 +3,24 @@
  * Init and run calls for land flight mode
  */
 
+bool ModeLand::init(bool ignore_checks)
+{
+    targ_vel = {0.0,0.0,0.5*blimp.loiter->max_vel_z};
```
peterbarker: This is a little opaque.  Split across multiple lines?

Blimp/mode_land.cpp
```diff
@@ -3,14 +3,24 @@
...
+
 // Runs the main land controller
 void ModeLand::run()
 {
-    //Stop moving
-    motors->right_out = 0;
-    motors->front_out = 0;
-    motors->yaw_out = 0;
-    motors->down_out = 0;
+    if (blimp.position_ok()) {
```
peterbarker: Use an early-return pattern based on
```suggestion
    if (!blimp.position_ok()) {
```

Blimp/mode_manual.cpp
```diff
@@ -9,8 +9,8 @@ void ModeManual::run()
     Vector3f pilot;
     float pilot_yaw;
     get_pilot_input(pilot, pilot_yaw);
-    motors->right_out = pilot.y;
-    motors->front_out = pilot.x;
-    motors->yaw_out = pilot_yaw;
-    motors->down_out = pilot.z;
+    motors->right_out = pilot.y*g.max_man_thr;
```
peterbarker:
```suggestion
    motors->right_out = pilot.y * g.max_man_thr;
```

spaces around operators is the norm (except where it isn't, typically in expressions-as-parameters (e.g. `fo_funct(x*y);`)

libraries/SITL/SIM_Blimp.cpp
```diff
@@ -50,232 +57,275 @@ void Blimp::calculate_forces(const struct sitl_input &input, Vector3f &body_acc,
...
-  //all fin setup
-  for (uint8_t i=0; i<4; i++) {
-    fin[i].last_angle = fin[i].angle;
-    if (input.servos[i] == 0) {
-        fin[i].angle = 0;
-        fin[1].servo_angle = 0;
-    } else {
-        fin[i].angle = filtered_servo_angle(input, i)*radians(45.0f)+radians(13.5); //for servo range of -75 deg to +75 deg
-        fin[i].servo_angle = filtered_servo_angle(input, i);
+  if (!motorblimp) { //Finned blimp
```
peterbarker, line 60: This isn't going to scale.  Suggest making an enumeration and `switch` on it

peterbarker (commented):
I suggest creating a `LoiterEachFrame` test.

Arm the vehicle in loiter , put in a very slight simulated wind, make sure the vehicle stationary for prolonged period of time (30 seconds?)

Needs a small auto-mode autotest, too.

libraries/SITL/SIM_Blimp.cpp
MichelleRos: How would that work, given that with that "-(something)" notation one could add multiple items?

Blimp/mode_auto.cpp
MichelleRos: How else?

Blimp/mode_auto.cpp
MichelleRos: What do you mean? I have to *0.01 else it'll be in cm.

Blimp/Blimp.cpp
MichelleRos: I've changed it to use the variable in motors, which is only set on class creation, hence it will only change on reboot.

Blimp/Fins.cpp
MichelleRos: It's actually not debug. It's a stream rate. Useful for tuning, so I put it under a parameter so people can enable it when they're busy tuning. Maybe needs a more precise parameter name?

Blimp/Loiter.cpp
MichelleRos: The loiter object is basically the equivalent to AC_PosControl. It's used by Auto mode as well, for example. Thus all of these modes do use the same tune/PIDs.

Blimp/mode_auto.cpp
MichelleRos: Done.

Blimp/Blimp.cpp
MichelleRos: I tried this and couldn't get it to compile. Doesn't seem to like the `auto& {` syntax. Nor did it like it when I put the `pids[]` bit before `{`. It was also more lines than what I currently have and looks more complex.

Blimp/Blimp.cpp
MichelleRos: Done.

Blimp/Log.cpp
MichelleRos: See my reply above.

Blimp/mode.cpp
MichelleRos: I've put the position check as the early-exit, but the rest still has its own if statement. Hope that's what you meant.

Blimp/Blimp.cpp
peterbarker:
```
-    loiter->pid_vel_x.set_notch_sample_rate(AP::scheduler().get_filtered_loop_rate_hz());
+    AC_PID *pids[] {
+        &loiter->pid_vel_x,
+    };
+
+    const auto loop_rate_hz = AP::scheduler().get_filtered_loop_rate_hz();
+    for (auto &pid : pids) {
+        pid->set_notch_sample_rate(loop_rate_hz);
+    }
+
```

Blimp/Blimp.cpp
```diff
@@ -225,9 +232,21 @@ void Blimp::read_AHRS(void)
...
-    vel_ned_filtd = {vel_xy_filtd.x, vel_xy_filtd.y, vel_z_filter.apply(vel_ned.z)};
-    vel_yaw_filtd = vel_yaw_filter.apply(vel_yaw);
+
+    switch (motors->_frame) {
+        case Fins::MOTOR_FRAME_FISHBLIMP:
+            vel_ned_filtd = {vel_x_filter.apply(vel_ned.x), vel_y_filter.apply(vel_ned.y), vel_z_filter.apply(vel_ned.z)};
+            vel_yaw_filtd = vel_yaw_filter.apply(vel_yaw);
+            break;
+        case Fins::MOTOR_FRAME_FOUR_MOTOR:
+            FALLTHROUGH;
```
peterbarker:
```suggestion
```

Blimp/Blimp.cpp
```diff
@@ -225,9 +232,21 @@ void Blimp::read_AHRS(void)
...
+        case Fins::MOTOR_FRAME_UNDEFINED:
+            FALLTHROUGH;
```
peterbarker:
```suggestion
```

Blimp/Fins.cpp
```diff
@@ -32,26 +33,57 @@ Fins::Fins(uint16_t loop_rate) :
...
+        case Fins::MOTOR_FRAME_FISHBLIMP:
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Setting up FishBlimp.");
+            setup_fins();
+            break;
+        case Fins::MOTOR_FRAME_FOUR_MOTOR:
+            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Setting up FourMotor.");
+            setup_motors();
+            break;
+        default:
+            GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "ERROR: Bad frame class.");
```
peterbarker, line 70: Make sure this works appropriately

Blimp/Fins.cpp
```diff
@@ -69,7 +101,17 @@ void Fins::add_fin(int8_t fin_num, float right_amp_fac, float front_amp_fac, flo
     }
 }
 
-//B,F,R,L = 0,1,2,3
+void Fins::add_motor(int8_t fin_num, float right_amp_fac, float front_amp_fac, float down_amp_fac, float yaw_amp_fac)
+{
+    // ensure valid fin number is provided
+    if (fin_num >= 0 && fin_num < NUM_FINS) {
```
peterbarker, line 127: early return - or maybe don't silently ignore this.

Blimp/Fins.cpp
peterbarker:
```suggestion
    if (!is_equal(float(blimp.g.stream_rate), 0.0f) && AP_HAL::millis() % int((1 / blimp.g.stream_rate) * 1000) < 30) {
```

Blimp/Fins.cpp
```diff
@@ -84,12 +126,35 @@ void Fins::output()
...
+        case Fins::MOTOR_FRAME_FISHBLIMP:
+            output_fins();
+            break;
+        case Fins::MOTOR_FRAME_FOUR_MOTOR:
+            output_motors();
+            break;
+        default:
+            GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "ERROR: Bad frame class.");
+            break;
+    }
```
peterbarker, line 169:
```suggestion
    switch (_frame) {
        case Fins::MOTOR_FRAME_FISHBLIMP:
            output_fins();
            return;
        case Fins::MOTOR_FRAME_FOUR_MOTOR:
            output_motors();
            return;
        case Fins::MOTOR_FRAME_UNDEFINED:
            break;
    }
    GCS_SEND_TEXT(MAV_SEVERITY_ERROR, "ERROR: Bad frame class.");
```

similarly for other `switch` statements

Blimp/Fins.cpp
```diff
@@ -145,3 +222,15 @@ void Fins::output_min()
...
+
+const char* Fins::get_frame_string()
+{
+    switch (_frame) {
+        case Fins::MOTOR_FRAME_FISHBLIMP:
+            return "FISHBLIMP";
+        case Fins::MOTOR_FRAME_FOUR_MOTOR:
+            return "FOURMOTOR";
+        default:
+            return "NOFRAME";
```
peterbarker: remove default

Blimp/Loiter.cpp
```diff
@@ -1,182 +1,1021 @@
...
+    }
+    case Fins::MOTOR_FRAME_FOUR_MOTOR: {
+        float xyaw_out = fabsf(blimp.motors->front_out) + fabsf(blimp.motors->yaw_out);
+        if (xyaw_out > 1.0f) {
+            scaler_x_n = 1.0f / xyaw_out;
+            scaler_yaw_n = 1.0f / xyaw_out;
+        }
+        break;
+    }
+    default:
```
peterbarker: bad :-)

Blimp/Loiter.cpp
```diff
@@ -1,182 +1,1021 @@
...
+#endif
+    if (!is_equal(float(blimp.g.stream_rate), 0.0f) && AP_HAL::millis() % int((1 / blimp.g.stream_rate) * 1000) < 30){
+        gcs().send_named_float("BSCXN", scaler_x_n);
+        gcs().send_named_float("BSCYN", scaler_y_n);
+        gcs().send_named_float("BSCZN", scaler_z_n);
+        gcs().send_named_float("BSCYAWN", scaler_yaw_n);
+        gcs().send_named_float("BSCX", scaler_x);
+        gcs().send_named_float("BSCY", scaler_y);
+        gcs().send_named_float("BSCZ", scaler_z);
+        gcs().send_named_float("BSCYAW", scaler_yaw);
```
peterbarker: Needs documenting (same as the other named-value-floats)

Blimp/Loiter.cpp
```diff
@@ -187,7 +1026,7 @@ void Loiter::run_vel(Vector3f& target_vel_ef, float& target_vel_yaw, Vector4b ax
         blimp.motors->right_out = 0;
     } else if (axes_disabled.y);
     else {
```
peterbarker, line 1017:
```suggestion
} else if (axes_disabled.y) {
    // explain what is going on here
} else {
```

Blimp/mode.cpp
```diff
@@ -180,6 +185,25 @@ bool Mode::is_disarmed_or_landed() const
...
 
+void Mode::yaw_forward()
+{
+    if (!blimp.position_ok()) {
+        return;
+    }
+    const float speed_sq = blimp.vel_ned_filtd.xy().length_squared();
+    // Make sure the blimp is keeping up with the target
+    const bool close_yaw = fabsf(wrap_PI(target_yaw-ahrs.get_yaw())) < (loiter->max_pos_yaw*blimp.loiter->pos_lag);
+    if ((speed_sq > sq(g.wp_yaw_min_vel)) && close_yaw) {
```
peterbarker: use early-return pattern?

Blimp/mode.h
```diff
@@ -312,35 +215,102 @@ class ModeRTL : public Mode
...
+    bool requires_GPS() const override { return true; }
+    bool has_manual_throttle() const override { return false; }
+    bool allows_arming(bool from_gcs) const override { return true; };
+    bool is_autopilot() const override { return true; }
+
+    Vector3f target_vel;
+    Vector3f target_accel;
+    bool waiting_to_start;
+
+    SCurve scurve_prev_leg;            // previous scurve trajectory used to blend with current scurve trajectory
```
peterbarker, line 250: public?!

Blimp/Loiter.cpp
```diff
@@ -1,182 +1,1021 @@
...
+    // @Description: Number of seconds' worth of travel that the actual position can be behind the target position.
+    // @Units: s
+    // @Range: 0 0.999
+    // @User: Standard
+    AP_GROUPINFO("POS_LAG", 19, Loiter, pos_lag, 1),
+
+    AP_GROUPEND
+};
+
+void Loiter::run(Vector3f& target_pos, float& target_yaw, Vector4b axes_disabled)
```
peterbarker, line 838:
```suggestion
void Loiter::run(Vector3f& target_pos, float& target_yaw, uint8_t disabled_axis_mask)
```

Blimp/Loiter.cpp
```diff
@@ -1,182 +1,1021 @@
...
+    if ((fabsf(err_yaw)   < pid_dz) || !blimp.motors->_armed || (dis_mask & (1<<(4-1)))) {
         zero.yaw = true;
     }
 
     //Disabled means "don't update PIDs or output anything at all". Zero means actually output zero thrust. I term is limited in either case."
     Vector4b limit = zero || axes_disabled;
 
     Vector3f target_vel_ef;
-    if (!axes_disabled.x && !axes_disabled.y) target_vel_ef = {blimp.pid_pos_xy.update_all(target_pos, blimp.pos_ned, dt, {(float)limit.x, (float)limit.y, (float)limit.z}), 0};
+    if (!axes_disabled.x)
```
peterbarker:
```suggestion
    if (axes_disabled & BLIMP_LOITER_AXIS_X)
```

```
static constexpr BLIMP_LOITER_AXIS_X { 1<<0; }
static constexpr BLIMP_LOITER_AXIS_Y { 1<<1; }
static constexpr BLIMP_LOITER_AXIS_Z { 1<<2; }
static constexpr BLIMP_LOITER_AXIS_YAW { 1<<3; }
```

Blimp/mode_auto.cpp
```diff
@@ -0,0 +1,222 @@
...
+                    GCS_SEND_TEXT(MAV_SEVERITY_CRITICAL, "Auto mission changed but failed to restart command.");
+                }
+            }
+        }
+
+        mission.update();
+    }
+
+    yaw_forward();
+    blimp.loiter->run(target_pos, target_yaw, Vector4b{false,false,false,false});
```
peterbarker, line 58:
```suggestion
    blimp.loiter->run(target_pos, target_yaw, 0);
```

or consider setting up the loiter controller in init?

Blimp/mode_auto.cpp
```diff
@@ -0,0 +1,222 @@
...
+    gcs().send_named_float("TarX", target_pos.x);
+    gcs().send_named_float("TarY", target_pos.y);
```
peterbarker, line 60: All of these need to be documented :-)

Blimp/mode_auto.cpp
peterbarker: Replace with `get_vector_from_origin_NED_m`

Blimp/mode_auto.cpp
```diff
@@ -0,0 +1,222 @@
...
+        } else {
+            // regular waypoints also require the copter to be within the waypoint radius
+            //Vector3f{destination.x,destination.y,destination.z}
+            const Vector3p dist_to_dest = Vector3p{blimp.pos_ned.x,blimp.pos_ned.y,blimp.pos_ned.z} - destination;
+            if (dist_to_dest.length_squared() <= sq(g.wp_rad)) {
+                return true;
+            }
+        }
+    }
+    return false;
```
peterbarker:
```suggestion
    if (!s_finished) {
        // explain what's going on here - scurve engine not done yet?
        return false;
    }
    // "fast" waypoints are complete once the intermediate point reaches the destination
    if (fast_wp) {
        // really?!
        return true;
    }
    // regular waypoints also require the copter to be within the waypoint radius
    //Vector3f{destination.x,destination.y,destination.z}
    const Vector3p dist_to_dest = Vector3p{blimp.pos_ned.x,blimp.pos_ned.y,blimp.pos_ned.z} - destination;
    return (dist_to_dest.length_squared() <= sq(g.wp_rad)) {
```

Blimp/Parameters.h
```diff
@@ -282,7 +276,7 @@ class ParametersG2
     // acro exponent parameters
     AP_Float acro_y_expo;
 
-    // frame class
+    // frame type
```
peterbarker: umm?

peterbarker (commented):
Consider renaming Fins to Actuators?  MovementCreatorOutputs?
Consider renaming Loiter to PosControl? BlimpPosControl?

peterbarker:
My branch pr-claude/MichelleRos/pr/blimp8-prop-blimp contains a new autotest and the markdown document explain the named-value-floats

Blimp/Fins.h
Hwurzburg: Since there is only one motor frame type (fish blimps use actuators). shouldn't it be MOTOR_FRAME_TYPE_BLIMP for consistency?

rmackay9:
I've restarted the flapping colcon test :-)

MichelleRos:
@rmackay9 Thank you very much. I've been repushing to try to get them all to pass and it's not worked well. 😅

MichelleRos:
> My branch pr-claude/MichelleRos/pr/blimp8-prop-blimp contains a new autotest and the markdown document explain the named-value-floats

@peterbarker 
I ended up removing the named-value-floats so the markdown document is not necessary anymore. Except for that I just realised TarX and Y are still there, but I really don't want to repush 'cause it's taken me multiple days repushing just trying to get all the flapping autotests to pass...

For the autotest part of your branch I think I'll only end up having one manual mode test because of how easily it turns into a flapping test. I can test the rest of the code that's different between the frames with other modes instead. I'll also be adding autotests for the automode in this PR, but I'd really prefer to do the whole rejigging of the autotest after this PR is merged because this PR is just such a large and unwieldy PR already...

peterbarker (approved):
LGTM

PR #29097, opened 2025-01-17
AP_HAL_ChibiOS: add hwdef files for NarinFC-H7

vololand:
AP_HAL_ChibiOS: add hwdef files for NarinFC-H7
base: Copter-4.5


vololand:
Why is this a failure?
"test chibios / build (configure-all, chibios, 10) (pull_request)"

PR #29099, opened 2025-01-17
AP_RangeFinder: add snr check for USD1 serial rangefinder

magate:


libraries/AP_RangeFinder/AP_RangeFinder_USD1_Serial.h
```diff
@@ -47,6 +50,7 @@ class AP_RangeFinder_USD1_Serial : public AP_RangeFinder_Backend_Serial
     bool     _version_known;
     uint8_t  _header;
     uint8_t  _version;
+    AP_Int32 snr_min;
```
peterbarker, line 53: `Int32` is too much, surely?

libraries/AP_RangeFinder/AP_RangeFinder_USD1_Serial.cpp
```diff
@@ -22,6 +22,29 @@
 #define USD1_HDR 254   // Header Byte from USD1_Serial (0xFE)
 #define USD1_HDR_V0 72 // Header Byte for beta V0 of USD1_Serial (0x48)
 
+const AP_Param::GroupInfo AP_RangeFinder_USD1_Serial::var_info[] = {
+    // @Param: SNR_MIN
+    // @DisplayName: Minimum signal strength
+    // @Description: Minimum signal strength (SNR) to accept distance
+    // @Range: 0 65535
+    // @User: Advanced
+    AP_GROUPINFO("SNR_MIN", 12, AP_RangeFinder_USD1_Serial, snr_min, 0),
```
peterbarker, line 31: I believe this number has to be chosen very carefully to avoid conflicts - how did you choose 12?

peterbarker (changes requested):
I've rebased this PR - it's been a while since it was pushed...

Could you strip the logging component of this out, please?

It' a separate issue to filtering the input based on the SNR, and there are questions around flag values and whatnot.  We log quality, which is a catch-all for SNR and whatever other metrics backends might have, too.

magate:
> I've rebased this PR - it's been a while since it was pushed...
> 
> Could you strip the logging component of this out, please?
> 
> It' a separate issue to filtering the input based on the SNR, and there are questions around flag values and whatnot. We log quality, which is a catch-all for SNR and whatever other metrics backends might have, too.

We've moved on from this. The docs specify that an SNR value of 13 dB or lower are considered erroneous. I am not sure if this is a 1:1 comparison with quality.

https://ainstein.ai/wp-content/uploads/US-D1-Technical-User-Manual-D00.02.05.docx.pdf
image

PR #20408, opened 2022-03-31
add cyphal (uavcan v1) support

PonomarevDA:
This PR adds a minimal implementation of Cyphal (uavcan v1.0) for ESC controlling.
It has the following required for any application functions:
- Node heartbeat (the node publishes uavcan.node.Heartbeat)
- Generic node information (the node response on uavcan.node.GetInfo.1.0 request)
- Register interface (response on uavcan.register.Access.1.0 and uavcan.register.List.1.0)

I also implemented templates for the following data types, but they do nothing yet. They might be completed if someone needs them or even removed for a while:
- uavcan.node.port.List (sends all supported port id)
- uavcan.node.ExecuteCommand (handles requests on restart or save parameters commands)

It also interacts with all necessary for ESC controlling data types, such as:
- reg.udral.service.actuator.common.sp.Vector4.0.1
- reg.udral.service.common.Readiness.0.1
- reg.udral.physics.electricity.PowerTs_0_1
- reg.udral.service.actuator.common.Feedback_0_1
- reg.udral.service.actuator.common.Status_0_1
- reg.udral.physics.dynamics.rotation.PlanarTs_0_1

All ESC related data types have non-fixed subject identifier. It is handled by the register interface that is based on AP_Param library. Since the implementation doesn't support automatically port id enumeration, you need to either manually or using a script from external server set their id (for ardupilot and for esc's).

Typical set of parameters and corresponded registers might be as on the picture below: 
ardupilot_params

I also forwarded esc telemetry to the corresponded MAVLink message:
esc_telem

The implementation is based on `libcanard`, `o1heap`, `public_regulated_data_types` repositories which are added as submodules. It also uses `nunavut` for generating the code from DSDL. I tried to implement it in the similar way as AP_UAVCAN is implemented. You may use both versions of UAVCAN simultaneously, but on different CAN buses only.

According to the build summary comparison, on CUAV v5 the current branch used 14244 bytes of flash memory more than master branch. Although free flash on this branch on this hardware is 317088 bytes.

[...]

PR #29100, opened 2025-01-17
Cyphal

failway:


failway:
> See #20408 and https://discuss.ardupilot.org/t/discussion-of-cyphal-support/98016

This is great, but I need Cyphal, I will still be compiling it. If you have the opportunity and the desire, you can help me. :)

andyp1per:
As per #20408 we won't be merging this.

PR #29101, opened 2025-01-17
Plane: Fixed formulas for load factor and roll, also corrected _TASmin in AP_TECS accordingly

IamPete1:
This is a rebase of https://github.com/ArduPilot/ardupilot/pull/24509, see also https://github.com/ArduPilot/ardupilot/issues/24320 and https://github.com/ArduPilot/ardupilot/pull/29092

Were were passing about the sqrt of the laod factor. This changes to the standard definition.

There is no change in behavior.

### DIFF
--- a/ArduPlane/Attitude.cpp
+++ b/ArduPlane/Attitude.cpp
@@ -633,7 +633,9 @@ void Plane::update_load_factor(void)
         // limit to 85 degrees to prevent numerical errors
         demanded_roll = 85;
     }
-    aerodynamic_load_factor = 1.0f / safe_sqrt(cosf(radians(demanded_roll)));
+
+    // loadFactor = liftForce / gravityForce, where gravityForce = liftForce * cos(roll) on balanced horizontal turn
+    aerodynamic_load_factor = 1.0f / cosf(radians(demanded_roll));
 
 #if HAL_QUADPLANE_ENABLED
     if (quadplane.available() && quadplane.transition->set_FW_roll_limit(roll_limit_cd)) {
@@ -657,7 +659,8 @@ void Plane::update_load_factor(void)
     }
 #endif
 
-    float max_load_factor = smoothed_airspeed / MAX(aparm.airspeed_min, 1);
+    float max_load_factor = sq(smoothed_airspeed / MAX(aparm.airspeed_min, 1));
+
     if (max_load_factor <= 1) {
         // our airspeed is below the minimum airspeed. Limit roll to
         // 25 degrees
@@ -670,11 +673,11 @@ void Plane::update_load_factor(void)
         // allow at least 25 degrees of roll however, to ensure the
         // aircraft can be manoeuvred with a bad airspeed estimate. At
         // 25 degrees the load factor is 1.1 (10%)
-        int32_t roll_limit = degrees(acosf(sq(1.0f / max_load_factor)))*100;
+        int32_t roll_limit = degrees(acosf(1.0f / max_load_factor))*100;
         if (roll_limit < 2500) {
             roll_limit = 2500;
         }
         nav_roll_cd = constrain_int32(nav_roll_cd, -roll_limit, roll_limit);
         roll_limit_cd = MIN(roll_limit_cd, roll_limit);
-    }    
+    }
 }
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -672,7 +672,7 @@ private:
     int32_t nav_pitch_cd;
 
     // the aerodynamic load factor. This is calculated from the demanded
-    // roll before the roll is clipped, using 1/sqrt(cos(nav_roll))
+    // roll before the roll is clipped, using 1/cos(nav_roll)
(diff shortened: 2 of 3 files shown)


ArduPlane/Attitude.cpp
```diff
@@ -657,7 +659,8 @@ void Plane::update_load_factor(void)
     }
 #endif
 
-    float max_load_factor = smoothed_airspeed / MAX(aparm.airspeed_min, 1);
+    float max_load_factor = sq(smoothed_airspeed / MAX(aparm.airspeed_min, 1));
+
```
Georacer, line 663: This doesn't look NFC to me.

Previously `aerodynamic_load_factor` and `max_load_factor` were comparable quantities:
```
    } else if (max_load_factor < aerodynamic_load_factor) {
```

But now `max_load_factor` is applied a square root, whereas `aerodynamic_load_factor` is raised to ^2.

Georacer (changes requested):

ArduPlane/Attitude.cpp
IamPete1: I think your right, here is some working out If we concentrate on just the squaring and ignore the rest

Existing code:

$$aerodynamicLoadFactor = \sqrt{x}$$ 
$$maxLoadFactor = y$$

That makes the comparison:

$$y < \sqrt{x}$$

Which we can re-write as

$$y -  \sqrt{x} < 0$$

Then we square it:

$$(y - \sqrt{x})^2 < 0$$
$$x - 2 \sqrt{x} y + y^2 < 0$$

The new equality is:

$$aerodynamicLoadFactor = x$$ 
$$maxLoadFactor = y^2$$
$$y^2 < x$$
$$y^2 - x < 0$$

The difference is:

$$2x - 2 \sqrt{x} y$$

So as you say the equality is not the same. I guess the next question is which is the correct value. I will have to do some more reading I think.

IamPete1: @Georacer @rubenp02, I decided to just plot both, this is the python:
```

from matplotlib import pyplot as plt
import math
import numpy

airspeed = numpy.linspace(1,40,num=1000).tolist()

AIRSPEED_MIN = 9

demanded_roll = 85

# Old method
def get_roll_lim_old(speed):
    max_load_factor = speed / AIRSPEED_MIN
    aerodynamic_load_factor = 1.0 / math.sqrt(math.cos(math.radians(demanded_roll)))

    if max_load_factor <= 1:
        return 25

    if max_load_factor < aerodynamic_load_factor:
        return max(25, math.degrees(math.acos((1/max_load_factor)**2)))

    return demanded_roll

# new method
def get_roll_lim(speed):
    max_load_factor = (speed / AIRSPEED_MIN)**2
    aerodynamic_load_factor = 1.0 / math.cos(math.radians(demanded_roll))

    if max_load_factor <= 1:
        return 25

    if max_load_factor < aerodynamic_load_factor:
        return max(25, math.degrees(math.acos(1/max_load_factor)))

    return demanded_roll


old = []
new = []
for speed in airspeed:
    old.append(get_roll_lim_old(speed))
    new.append(get_roll_lim(speed))


fig = plt.figure()
plt.title('Comparison of old and new load factor calcs')
plt.plot(airspeed, old, label = "old")
plt.plot(airspeed, new, label = "new")
plt.legend()
plt.xlabel('Airspeed (m/s)')
plt.ylabel('Roll (deg)')

plt.show()

```
It results in:
image

So no change in behaviour. Did I make a mistake in my python?

At least I suspect the code is more complex than it needs to be. I don't know why we don't just calculate the max bank angle and constrain it, the result is the same. EG:
```
def get_roll_lim_easy(speed):
    max_load_factor = (speed / AIRSPEED_MIN)**2

    if max_load_factor <= 1:
        return 25

    return min(max(25, math.degrees(math.acos(1/max_load_factor))), demanded_roll) 
```
Then the aero load factor is not needed at all for this calculation.

IamPete1: > However, this PR is still a functional change because of the TECS side of things

@rubenp02 you have the new load factor calculation with the old method in TECS. As they are in this PR there is no change in behaviour because TECS is also updated. Your `calculate_tecs_limits`  should be this:
```
def calculate_tecs_limits():
    load_factor_old = 1 / np.sqrt(np.cos(np.radians(BANK_ANGLES)))
    load_factor_new = 1 / np.cos(np.radians(BANK_ANGLES))

    tas_min_old = TAS_MIN_LEVEL * load_factor_old
    tas_min_new = TAS_MIN_LEVEL * np.sqrt(load_factor_new)

    return tas_min_old, tas_min_new
``` 

So, I still think there is no change in behaviour as a result of this PR as it currently is. @Georacer @rubenp02 do you agree?

Of course there is a discussion to be had if we want to change the behaviour, but that can be a future PR.

Georacer: @IamPete1 @rubenp02 sorry for the slow replies, it's been a little busy.
I'll try to reply when I can, but mind you I haven't blocked this PR.

Georacer (approved):

PR #29104, opened 2025-01-18
ArduPlane:make heading lock upon gnd speed consistent

Hwurzburg:
we use GPS_GND_CRS_MIN_SPD  in AUTOLAND, this makes ModeTakeoff and NAV_TAKEOFF consistent with this

### DIFF
--- a/ArduPlane/commands_logic.cpp
+++ b/ArduPlane/commands_logic.cpp
@@ -377,7 +377,6 @@ void Plane::do_takeoff(const AP_Mission::Mission_Command& cmd)
     auto_state.takeoff_altitude_rel_cm = next_WP_loc.alt - home.alt;
     next_WP_loc.lat = home.lat + 10;
     next_WP_loc.lng = home.lng + 10;
-    auto_state.takeoff_speed_time_ms = 0;
     auto_state.takeoff_complete = false; // set flag to use gps ground course during TO. IMU will be doing yaw drift correction.
     auto_state.height_below_takeoff_to_level_off_cm = 0;
     // Flag also used to override "on the ground" throttle disable
@@ -553,21 +552,15 @@ bool Plane::verify_takeoff()
     trust_ahrs_yaw |= ahrs.dcm_yaw_initialised();
 #endif
     if (trust_ahrs_yaw && steer_state.hold_course_cd == -1) {
-        const float min_gps_speed = GPS_GND_CRS_MIN_SPD;
-        if (auto_state.takeoff_speed_time_ms == 0 && 
-            gps.status() >= AP_GPS::GPS_OK_FIX_3D && 
-            gps.ground_speed() > min_gps_speed &&
+        // once we reach sufficient speed for good GPS course
+        // estimation we save our current GPS ground course
+        // corrected for summed yaw to set the take off
+        // course. This keeps wings level until we are ready to
+        // rotate, and also allows us to cope with arbitrary
+        // compass errors for auto takeoff
+        if (gps.status() >= AP_GPS::GPS_OK_FIX_3D && 
+            gps.ground_speed() > GPS_GND_CRS_MIN_SPD &&
             hal.util->safety_switch_state() != AP_HAL::Util::SAFETY_DISARMED) {
-            auto_state.takeoff_speed_time_ms = millis();
-        }
-        if (auto_state.takeoff_speed_time_ms != 0 &&
-            millis() - auto_state.takeoff_speed_time_ms >= 2000) {
-            // once we reach sufficient speed for good GPS course
-            // estimation we save our current GPS ground course
-            // corrected for summed yaw to set the take off
-            // course. This keeps wings level until we are ready to
-            // rotate, and also allows us to cope with arbitrary
-            // compass errors for auto takeoff
             float takeoff_course = wrap_PI(radians(gps.ground_course())) - steer_state.locked_course_err;
             takeoff_course = wrap_PI(takeoff_course);
             steer_state.hold_course_cd = wrap_360_cd(degrees(takeoff_course)*100);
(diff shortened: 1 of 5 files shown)


IamPete1:
I think we need to merge https://github.com/ArduPilot/ardupilot/pull/29104 first. There are is some overlap.

ArduPlane/commands_logic.cpp
```diff
@@ -553,26 +552,15 @@ bool Plane::verify_takeoff()
...
             hal.util->safety_switch_state() != AP_HAL::Util::SAFETY_DISARMED) {
-            auto_state.takeoff_speed_time_ms = millis();
-#if MODE_AUTOLAND_ENABLED
-            plane.takeoff_state.initial_direction.heading = wrap_360(plane.gps.ground_course() + plane.mode_autoland.landing_dir_off);
-            takeoff_state.initial_direction.initialized = true;
-            gcs().send_text(MAV_SEVERITY_INFO, "Autoland direction= %u",int(takeoff_state.initial_direction.heading));
-#endif
-        }
-        if (auto_state.takeoff_speed_time_ms != 0 &&
-            millis() - auto_state.takeoff_speed_time_ms >= 2000) {
```
IamPete1, line 569: The key thing is were removing this 2 second delay?

Hwurzburg:
I see no need for the delay before capturing heading...gpses have had good velocity and heading since the m8n appeared years ago

Yahoo Mail: Search, Organize, Conquer 
 
  On Sun, Jan 19, 2025 at 2:00 PM, Peter ***@***.***> wrote:   


@IamPete1 commented on this pull request.

In ArduPlane/commands_logic.cpp:
>              hal.util->safety_switch_state() != AP_HAL::Util::SAFETY_DISARMED) {
-            auto_state.takeoff_speed_time_ms = millis();
-#if MODE_AUTOLAND_ENABLED
-            plane.takeoff_state.initial_direction.heading = wrap_360(plane.gps.ground_course() + plane.mode_autoland.landing_dir_off);
-            takeoff_state.initial_direction.initialized = true;
-            gcs().send_text(MAV_SEVERITY_INFO, "Autoland direction= %u",int(takeoff_state.initial_direction.heading));
-#endif
-        }
-        if (auto_state.takeoff_speed_time_ms != 0 &&
-            millis() - auto_state.takeoff_speed_time_ms >= 2000) {

The key thing is were removing this 2 second delay?

Hwurzburg:
going to try to improve failsafe heading code as per Devcall

Hwurzburg:
After reviewing the code in mode_takeoff and commands_logic, and a lot of SITL with headwinds = cruise speed,  I do not think a failsafe for not having enough groundspeed to break out of roll limit constraints in the face of extreme headwind is required....eventually the altitude target will be reached in either case and the limit eliminated....no fly away will occur, because the vehicle is moving VERY slowly...it will almost vertically ascend and then switch to next mission item or initiate the mode_takeoff loiter, and takeoff has to be in headwind conditions within 5m/s of cruise speed....so just capturing heading based on groundspeed is valid in my opinion....the other PR removing the mode_takeoff failsafe after LEVEL_ALT is obtained can be merged in my opinion also....currently NAV_TAKEOFF has no equivalent and requires no removal 
I have changed the differing ground speed gates where its used to validate the use of GPS heading and used the common define of GPS_GND_CRS_MIN_SPD  which is 5m/s currently...it replaces values ranging from 2.7m/s to 5m/s in these areas...in my opinion it could be changed to 3m/s instead of its current 5m/s without impact given modern GPS...

ArduPlane/servos.cpp
```diff
@@ -104,7 +104,7 @@ bool Plane::suppress_throttle(void)
         return false;
     }
 
-    bool gps_movement = (gps.status() >= AP_GPS::GPS_OK_FIX_2D && gps.ground_speed() >= 5);
+    bool gps_movement = (gps.status() >= AP_GPS::GPS_OK_FIX_2D && gps.ground_speed() >= GPS_GND_CRS_MIN_SPD);
```
IamPete1, line 107: This is not quite the same use case, I don't think we should change to the new define.

ArduPlane/defines.h
```diff
@@ -10,7 +10,7 @@
 
 #define TAKEOFF_RUDDER_WARNING_TIMEOUT 3000 //ms that GCS warning about not returning arming rudder to neutral repeats
 
-#define GPS_GND_CRS_MIN_SPD 5 // m/s, used to set when intial_direction.heading is captured in NAV_TAKEOFF and Mode TAKEOFF
+#define GPS_GND_CRS_MIN_SPD 5 // m/s, used to set when intial_direction.heading is captured or steer_state.hold_course_cd
```
IamPete1, line 13: and cruise mode

ArduPlane/mode_cruise.cpp
```diff
@@ -67,7 +67,7 @@ void ModeCruise::navigate()
         plane.channel_roll->get_control_in() == 0 &&
         plane.rudder_input() == 0 &&
         plane.gps.status() >= AP_GPS::GPS_OK_FIX_2D &&
```
IamPete1, line 69: We don't seem to be consistent about checking for 3D vs 2D fix.

ArduPlane/defines.h
Hwurzburg: added

ArduPlane/servos.cpp
Hwurzburg: okay

ArduPlane/mode_cruise.cpp
Hwurzburg: true, maybe in another PR

ArduPlane/commands_logic.cpp
Hwurzburg: see below

IamPete1 (approved):
LGTM

tridge (approved):

PR #29106, opened 2025-01-18
autotest: allow wsl2 networking bridge to be ignored

andyp1per:
On Windows 11 there is a new networking option  - mirrored - that allows direct access to the NIC which requires that we don't try and connect to the bridge when running simulations
https://learn.microsoft.com/en-us/windows/wsl/wsl-config#main-wsl-settings

### DIFF
--- a/Tools/autotest/sim_vehicle.py
+++ b/Tools/autotest/sim_vehicle.py
@@ -201,7 +201,7 @@ def under_wsl2():
 
 
 def wsl2_host_ip():
-    if not under_wsl2():
+    if not under_wsl2() or cmd_opts.no_wsl2_network:
         return None
 
     pipe = subprocess.Popen("ip route show default | awk '{print $3}'",
@@ -1297,6 +1297,11 @@ group_sim.add_option("", "--no-extra-ports",
                      dest='no_extra_ports',
                      default=False,
                      help="Disable setup of UDP 14550 and 14551 output")
+group_sim.add_option("", "--no-wsl2-network",
+                     action='store_true',
+                     dest='no_wsl2_network',
+                     default=False,
+                     help="Disable setup of WSL2 network for output")
 group_sim.add_option("-Z", "--swarm",
                      type='string',
                      default=None,


timtuxworth:
Great find - thanks @andyp1per !

tridge (approved):

Hwurzburg:
Under what situations would you use this option?

andyp1per:
> Under what situations would you use this option?

It makes the networking behave like WSL1, so much easier from a user perspective. It also considerably improves network performance so recommended by microsoft.

Hwurzburg:
that is what it does, not when touse it....always when you SITL under a WSL2 environment? if so, we change these startup command  here: https://ardupilot.org/dev/docs/sitl-on-windows-wsl.html for all SITL startups
otherwise, needs details on when to use it

andyp1per:
@Hwurzburg you use it if you have configured WSL2 networking to mirrored.

Hwurzburg:
Sorry but I don't have the foggiest idea what that means

andyp1per:
Not asking you to. I think it would be sufficient to document that you need this option if you have configured WSL2 in this way.

timtuxworth:
I think I would agree with @Hwurzburg  that we could be more helpful explaining this to people who just want it to work.

Clearly you are saying @andyp1per  that it's a good idea to configure WSL2 to use mirrored networkingMode, so the wiki should say that - it's our recommendation. I think it might be helpful to say that somewhere.

I've just tried it and it seems to make running SITL (especially with RealFlight) easier and faster.

andyp1per:
> I think I would agree with @Hwurzburg that we could be more helpful explaining this to people who just want it to work.
> 
> Clearly you are saying @andyp1per that it's a good idea to configure WSL2 to use mirrored networkingMode, so the wiki should say that - it's our recommendation. I think it might be helpful to say that somewhere.
> 
> I've just tried it and it seems to make running SITL in WSL2 (especially with RealFlight) easier and faster. I'm getting 300ish frames per second in RF now which is a massive improvement on my hardware.

I would say its too early to tell that it should be the recommended option, but your experience is a good datapoint. If we can get a few more devs giving it the thumbs up then I agree

timtuxworth:
As an aside - this seems to have fixed a "Time moved backwards", problem I was having running SITL in WSL2, now gone after setting networkingMode=mirrored.

PR #29112, opened 2025-01-20
Copter: fix threading issues with motor test and the fast rate thread

andyp1per:
Motortest currently doesn't work with the fast rate thread and I suspect that it was somewhat accidental that it did before. The issue is synchronization between the fast rate thread and the main thread when dealing with the motor test. This fixes the synchronization issues and allows motor test to work normally with the fast rate thread.

Reported here: https://discuss.ardupilot.org/t/unlocking-the-potential-of-faster-attitude-rates-in-copter-control/120743/54

### DIFF
--- a/ArduCopter/motor_test.cpp
+++ b/ArduCopter/motor_test.cpp
@@ -155,14 +155,8 @@ MAV_RESULT Copter::mavlink_motor_test_start(const GCS_MAVLINK &gcs_chan, uint8_t
         } else {
             // start test
             gcs().send_text(MAV_SEVERITY_INFO, "starting motor test");
-            ap.motor_test = true;
 
             EXPECT_DELAY_MS(3000);
-
-            // wait for rate thread to stop running due to motor test
-            while (using_rate_thread) {
-                hal.scheduler->delay(1);
-            }
 
             // enable and arm motors
             if (!motors->armed()) {
@@ -178,6 +172,7 @@ MAV_RESULT Copter::mavlink_motor_test_start(const GCS_MAVLINK &gcs_chan, uint8_t
 
             // turn on notify leds
             AP_Notify::flags.esc_calibration = true;
+            ap.motor_test = true;
         }
     }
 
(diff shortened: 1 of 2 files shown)


tridge:
I think we need @rmackay9 to have a look at this one

ArduCopter/motor_test.cpp
```diff
@@ -155,15 +155,9 @@ MAV_RESULT Copter::mavlink_motor_test_start(const GCS_MAVLINK &gcs_chan, uint8_t
         } else {
             // start test
             gcs().send_text(MAV_SEVERITY_INFO, "starting motor test");
-            ap.motor_test = true;
```
rmackay9, line 158: @andyp1per thanks for this.  Is it really necessary to move "ap.motor_test = true" to be lower in the function?  It's hard for me to imagine why this is helpful.  The purpose of this flag is to stop the function from being called simultaneously by the user pushing the button the GCS rapidly.  moving this line lower should increase the chance that that could happen.

andyp1per: @rmackay9 the move is 100% ncessary - its current position is what causes the threading issue because it is checked concurrently

tridge: removing the while(using_rate_thread) { delay(1); } loop, and removing the flag check in Copter::rate_controller_thread() is sufficient?

andyp1per: @tridge no that does not work

peterbarker:
Randy wants to know *why* movinjg the false assignment makes any difference to the test.

Peter wants to know why it would make a difference since that code won't have run for the first test, which apparently fails...

andyp1per:
The race is with the arming code:
```
    if (!motors->armed()) {
        motors->output_min();  // output lowest possible value to motors
        motors->armed(true);
        hal.util->set_soft_armed(true);
    }
```
if this is moved to the beginning of motor_test_output() then things work ok without moving the flag

rmackay9:
Sorry, I know this is still on my list to investigate

peterbarker (approved):
LGTM

tridge (approved):

PR #29116, opened 2025-01-20
AP_Scripting: patch vendored Lua to 5.3.6

tpwrules:
In particular this fixes some exceedingly rare/impossible use-after-frees. The serious ones aren't actually accessible in our scripting engine.

To maintain our alterations, I applied source patches generated from the source repository manually instead of simply overwriting the code. Please see the commit message for full details.

Tested by flying scripting aerobatics on SITL on HW on Cube Orange.

The size overhead is negligible:
```
CubeOrange-periph-heavy  8                 *                                                   
Durandal                            8      *           8       8                 8      8      8
KakuteH7-bdshot                     48     *           48      48                48     40     48
```

### DIFF
--- a/libraries/AP_Scripting/README.md
+++ b/libraries/AP_Scripting/README.md
@@ -55,3 +55,13 @@ See the [code examples folder](https://github.com/ArduPilot/ardupilot/tree/maste
 
 Edit bindings.desc and rebuild. The waf build will automatically
 re-run the code generator.
+
+## Lua Source Code
+
+The Lua 5.3.6 source code is vendored in `lua/`. This is a customized
+version of the [official
+distribution](https://www.lua.org/ftp/lua-5.3.6.tar.gz). Where possible,
+differences have been marked of the code.
+
+Lua (not including modifications) is distributed under the terms of the
+MIT license.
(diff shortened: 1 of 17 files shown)


IamPete1:
This is missing a patch to `src/Makefile` the freebsd case should be:
```	
$(MAKE) $(ALL) SYSCFLAGS="-DLUA_USE_LINUX -I/usr/include/edit" SYSLIBS="-Wl,-E -ledit" CC="cc"
```
Apart from that this is a exact copy of the 4.5 to 4.6 diff. No idea if everything work but I guess we have to trust the lua maintainers to know what there doing.

tpwrules:
Thanks for the investigation. I didn't catch that makefile; I did the diff based on the Git repo instead of the archives. But I don't think anybody uses our tree to build for freebsd so I'm going to keep it how it is.

For reference, the changes in `lparser.c` are the only ones I think we have a chance of running into.

tpwrules:
Added those clarifications, thanks.

tridge (approved):

tridge:
@IamPete1 looks good to me, but I'd like to give you a chance to review/approve before merging

IamPete1 (approved):
Changes match what lua has done, all we can really do is trust that upstream lua is OK.

PR #29118, opened 2025-01-21
AP_Scripting: only log aerobatic trick file when supported

tpwrules:
LittleFS does not have this function implemented due to logging bandwidth, so `logger:log_file_content` is nil. Don't call it unless it exists, assuming this logging is not critical.


Hwurzburg (approved):
LGTM

IamPete1 (approved):
LGTM

PR #29120, opened 2025-01-21
AP_HAL_ChibiOS: enable scripting on KakuteH7Mini-Nand

tpwrules:
It has LittleFS now so it can easily use scripting.

Missed on the LittleFS PR but has worked great for me.


andyp1per (approved):

Hwurzburg:
The only issue is how does a user know ahead of time that the autopilot has this capability....since its not exposed in features.txt, the only way would be to examine the hwdef, which is beyond a typical user....

tpwrules:
Boards with LittleFS support now all ship scripting enabled by default. Boards which can only run it in ROMFS (should) have scripting disabled by default. So a user can check if scripting is enabled in features.txt, or if the SCR_ENABLE parameter exists (assuming they already flashed it).

Regardless, there is for sure documentation and UX stuff to work out before 4.7. Thanks for the comment.

Issue #19306, opened 2021-11-19
Rover: boats may drift away in Auto if current is too high


rmackay9:
During testing of this PR to add support for "motorboat" to SITL I discovered that in at least one edge case the navigation controller will get stuck pointing the vehicle in the wrong direction with very low throttle if the current is too high and nearly perpendicular to one leg of the mission.

To recreate the issue in SITL do this:

- ../Tools/autotest/sim_vehicle.py --map --console -f motorboat
- param set SIM_TIDE_DIR 90 (current will flow from East to West)
- param set SIM_TIDE_SPEED 3 (3m/s current)
- create a mission that is a square box ([linked here](https://www.dropbox.com/s/97afgmiclv0h22a/rover-42-drift-away-mission.waypoints?dl=0))
- Auto
- arm

You should see the vehicle gets stuck pointing in the wrong direction as it starts the North->South leg of the mission
![image](https://user-images.githubusercontent.com/1498098/142558365-5fa1f8ac-b563-455c-b6a9-f8390ce33799.png)

geofrancis:
is this related to this issue here https://discuss.ardupilot.org/t/improve-steering-if-motor-throttle-is-off/74572

his problem sounds very similar to yours.

PR #29125, opened 2025-01-22
Rover: Update crash_check.cpp to remove stuck check for boat

agregghai:
It is possible for a boat to have a very low GPS speed but be healthy (not stuck) as it fights a strong current. This commit removes the "stuck" check if the type is a boat.

### DIFF
--- a/Rover/crash_check.cpp
+++ b/Rover/crash_check.cpp
@@ -27,7 +27,7 @@ void Rover::crash_check()
     // TODO : Check if min vel can be calculated
     // min_vel = ( CRASH_CHECK_THROTTLE_MIN * g.speed_cruise) / g.throttle_cruise;
 
-    if (!is_balancebot()) {
+    if (!is_balancebot() && !is_boat()) {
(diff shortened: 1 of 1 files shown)


rmackay9:
Hi @agregghai,

Thanks for looking into this.  The Rover crash check was designed some time ago when the code was much simpler than it is today.  I guess you're seeing false positives where the vehicle is being switched into Hold mode?

Instead of just disabling the check I wonder if we could improve the velocity check by pulling it from the position controller.  Another idea is to add CRASH_xxx parameters which allow the user to set some of the values instead of using hard-coded values

rmackay9:
Hi @agregghai,

My concern is that if we simply disable the crash check the motors won't turn off when the boat is beached.  I also think that in practice it will be very difficult for a moving vehicle to have its velocity stay below 0.1 m/s for 2 seconds.  There will be too much noise in the estimated velocity for this to actually happen.

To move forward on this, I think we need a real case where its happening.  That could be an actual vehicle or in a simulator

BTW, we also have [this Auto mode problem](https://github.com/ArduPilot/ardupilot/issues/19306) if the boat is in extremely strong current

PR #29126, opened 2025-01-22
AP_HAL: SIMState: do a single nullptr check in SIMState servo sim

peterbarker:
mirrors if c1016ae52ee4de3181b2c0a07369e51f05eb46f7

### DIFF
--- a/libraries/AP_HAL/SIMState.cpp
+++ b/libraries/AP_HAL/SIMState.cpp
@@ -274,6 +274,10 @@ void SIMState::fdm_input_local(void)
  */
 void SIMState::_simulator_servos(struct sitl_input &input)
 {
+    if (_sitl == nullptr) {
+        return;
+    }
+
     // output at chosen framerate
     uint32_t now = AP_HAL::micros();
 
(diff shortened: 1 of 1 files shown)


davidbuzz:
discussed this in discord with peter, and its an attempt to fix the crash/backtrace report on this closed pr here: https://github.com/ArduPilot/ardupilot/pull/29040   , not sure if its a fix, yet.

peterbarker:
Tested ChibiOS master forCubeOrangePlus-SimOnHW and it works fine.

This PR also works fine for for the same target.

Just need to know it fixes the esp32 problem.

tpwrules (changes requested):
Confirmed this fixes the issue on esp32s3empty, but you missed a couple checks on line 324 and line 340.

Would also prefer to see the battery voltage and FETTec blocks untabbed and `{}` removed; they don't declare any locals so the code doesn't change.

peterbarker:
> Confirmed this fixes the issue on esp32s3empty, but you missed a couple checks on line 324 and line 340.

Thanks for testing!

I've removed those extra checks you spotted.  The second one is strange - I know I've removed both checks in some patch or another; perhaps a bad conflict resolution lost one of them.

> Would also prefer to see the battery voltage and FETTec blocks untabbed and `{}` removed; they don't declare any locals so the code doesn't change.

The braces are there to allow for *not* changing the indenting.  In this case I'm preferring to keep the history on the lines intact than making everything pretty.  There's no flow-of-control ambiguity inserted by having the slightly odd indenting, the basic reason we do indenting in the first place :-)

tpwrules (approved):

tpwrules:
Strange, but if you say so.

Issue #28827, opened 2024-12-09
SITL --add-param-file fails with multiple vehicles


eschutz:
## Bug report

**Issue details**

Passing a param file to `sim_vehicle.py` with `--count` > 1 causes ArduPlane SITL to fail to start.

Reproduce by running the following in the same directory as a valid parameter file `my_plane.parm`:
`sim_vehicle.py -v plane --add-param-file=my_plane.parm --count 2`

Running ArduPlane on its own (executed verbatim from the sim_vehicle.py logs, see below), and connecting with MAVProxy from another terminal, I get:
`<ardupilot directory>/build/sitl/bin/arduplane -S --model plane --speedup 1 --slave 0 --defaults ../../../ardupilot/Tools/autotest/models/plane.parm,my_plane.parm --sim-address=127.0.0.1 -I0`

```
Setting SIM_SPEEDUP=1.000000
Starting sketch 'ArduPlane'
Starting SITL input
Using Irlock at port : 9005
bind port 5760 for SERIAL0
SERIAL0 on TCP port 5760
Waiting for connection ....
Connection on serial port 5760
PANIC: Failed to load defaults from ../../../ardupilot/Tools/autotest/models/plane.parm,my_plane.parm

Running: sh dumpstack.sh 2879426 >dumpstack.sh_arduplane.2879426.out 2>&1
Failed
Running: sh dumpcore.sh 2879426 >dumpcore.sh_arduplane.2879426.out 2>&1
Failed
```

The PANIC line seems to indicate there's a problem with the parameter file path. When running the same command with one level of `..` removed: `<ardupilot directory>/build/sitl/bin/arduplane -S --model plane --speedup 1 --slave 0 --defaults ../../ardupilot/Tools/autotest/models/plane.parm,my_plane.parm --sim-address=127.0.0.1 -I0`, ArduPlane runs as expected.
This suggests the problem lies at this line of code https://github.com/ArduPilot/ardupilot/blob/4d31a7320a1d2c38e2d742ae63c34f914febaa8f/Tools/autotest/sim_vehicle.py#L798

Note that this is in the same region that seems to cause another bug with multiple agents https://github.com/ArduPilot/ardupilot/issues/28826.

**Version**
https://github.com/ArduPilot/ardupilot/commit/4df758f52ad077fd577f0f5a026932642bfb439b

**Logs**
Output of `sim_vehicle.py`:
```
[...]

Issue #28826, opened 2024-12-09
SITL --auto-sysid with a single vehicle causes link failure


eschutz:
## Bug report

**Issue details**

Passing `sim_vehicle.py` --auto-sysid with --count unspecified or --count 1 causes the TCP link to fail.

Reproduce with `sim_vehicle.py -v plane --auto-sysid` or `sim_vehicle.py -v plane --auto-sysid --count 1`. Does not occur when passing `--count 2`.

I'm not familiar with the ArduPilot/SITL source code but I think the fix would be an additional check here: https://github.com/ArduPilot/ardupilot/blob/4d31a7320a1d2c38e2d742ae63c34f914febaa8f/Tools/autotest/sim_vehicle.py#L796

**Version**
Ardupilot: 4df758f

**Logs**
Output of one of the above commands:
```
SIM_VEHICLE: Using defaults from (<ardupilot directory>/Tools/autotest/models/plane.parm)
SIM_VEHICLE: Run ArduPlane
SIM_VEHICLE: "<ardupilot directory>/Tools/autotest/run_in_terminal_window.sh" "ArduPlane" "<ardupilot directory>/build/sitl/bin/arduplane" "-S" "--model" "plane" "--speedup" "1" "--slave" "0" "--defaults" "<ardupilot directory>/Tools/autotest/models/plane.parm" "--sim-address=127.0.0.1" "-I0" "--sysid" "1"
RiTW: Starting ArduPlane : <ardupilot directory>/build/sitl/bin/arduplane -S --model plane --speedup 1 --slave 0 --defaults <ardupilot directory>/Tools/autotest/models/plane.parm --sim-address=127.0.0.1 -I0 --sysid 1
SIM_VEHICLE: Run MavProxy
SIM_VEHICLE: "mavproxy.py" "--out" "127.0.0.1:14550" "--master" "tcp:127.0.0.1:5760" "--sitl" "127.0.0.1:5501"
Connect tcp:127.0.0.1:5760 source_system=255
[Errno 111] Connection refused sleeping

<mavinit logs>

Log Directory: 
Telemetry log: mav.tlog
Waiting for heartbeat from tcp:127.0.0.1:5760
MAV> link 1 down
```

PR #29127, opened 2025-01-23
Fix for ArduPilot/ardupilot#28826 and #28827

eschutz:
Addresses #28826 and #28827 by adding a check for relative param file paths in sim_vehicle.py when simulating multiple agents.

Expected behaviour is that running either
- `sim_vehicle.py` with `--count` > 1 and `--add-param-file=<relative path>`, or
- `sim_vehicle.py --auto-sysid --count 1`

no longer fail.

### DIFF
--- a/Tools/autotest/sim_vehicle.py
+++ b/Tools/autotest/sim_vehicle.py
@@ -792,7 +792,7 @@ def start_vehicle(binary, opts, stuff, spawns=None):
                 print("The parameter file (%s) does not exist" % (x,))
                 sys.exit(1)
         path = ",".join(paths)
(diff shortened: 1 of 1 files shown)


rmackay9:
Hi @eschutz,

Thanks very much for this!  Could you update the commit title to start with "Tools: "?  We do this to make backports easier.  To be clear, it's the **commit** title that should be updated, not the PR title.

Txs again

eschutz:
Thanks - done.

khancyr:
closed in favor of solution here : https://github.com/ArduPilot/ardupilot/pull/29127 Thanks for the work anyway !

PR #28127, opened 2024-09-17
Ensure all peripherals define HAL_PERIPH_ENABLE_GPS to 1 or 0

peterbarker:

PR #29128, opened 2025-01-23
Create and use AP_PERIPH_BATTERY_ENABLED 

peterbarker:
.... similar to https://github.com/ArduPilot/ardupilot/pull/28127/commits

 - convert to new define names
 - ensure it is not just defined but has a value
 - ensure the old commit is no longer used
 -


peterbarker:
This is confirmed as a no-compiler-output change.

PR #29122, opened 2025-01-21
AP_Baro: Support MS5837-02BA #27787

Shaikimram:
The changes include differentiation between multiple MS5xxx sensor variants using the PROM Word 0 bits [11:5], enabling automatic initialization and configuration of the correct driver. Additionally, this update includes specific calculations and second-order compensation logic for the MS5837-02BA sensor, making it more viable for shallow water applications.

Key Changes
Differentiation Logic:

Extracted product ID from PROM Word 0 bits [11:5] to identify sensor variants:
MS5837-30BA26: 0b0011010
MS5837-02BA01: 0b0000000
MS5837-02BA21: 0b0010101
New Sensor Type:

Added a new enum type BARO_MS5837_02BA to differentiate MS5837-02BA sensors.
Calculation Logic:

Implemented _calculate_5837_02ba() for MS5837-02BA-specific pressure and temperature calculations, including second-order compensation for temperatures below 20°C.
Driver Initialization:

Updated _init() to automatically detect and initialize the appropriate sensor type using PROM data.
Added automatic assignment of water pressure type (AP_Baro::BARO_TYPE_WATER) for MS5837 sensors.

fix #27787

PR #29132, opened 2025-01-24
AP_Baro: update MS5837-30BA compensation

ES-Alexander:
### Context
While reviewing #29122 I found that the MS5837_30BA sensor's [current datasheet](https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Data+Sheet%7FMS5837-30BA%7FB1%7Fpdf%7FEnglish%7FENG_DS_MS5837-30BA_B1.pdf%7FCAT-BLPS0017#page=12) has some sections/factors for compensation that seemingly weren't there previously, so aren't yet being considered.

### Changes
This PR adds in sections for both "high temperature" and "very low temperature" compensation regions, which were previously unhandled.

### DIFF
--- a/libraries/AP_Baro/AP_Baro_MS5611.cpp
+++ b/libraries/AP_Baro/AP_Baro_MS5611.cpp
@@ -486,29 +486,37 @@ void AP_Baro_MS56XX::_calculate_5637()
 // Calculate Temperature and compensated Pressure in real units (Celsius degrees*100, mbar*100).
 void AP_Baro_MS56XX::_calculate_5837()
 {
-    int32_t dT, TEMP;
-    int64_t OFF, SENS;
+    int32_t dT, TEMP, T2;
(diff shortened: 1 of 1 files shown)


ES-Alexander:
Rebased over `master`, and updated the commit message to be more descriptive and better match the repo standards.

Williangalvani (approved):
looks good to me. The math matches the datasheet and it worked fine on my vehicle.
I do have mixed feelings regarding the expanded 2^n numbers, but not enough to block merging.

ES-Alexander:
> I do have mixed feelings regarding the expanded 2^n numbers

I would definitely prefer bit-shifts (assuming they're ok for the data-types involved), especially since the datasheet provides powers rather than expanded integers, but I think cleaning that up is a problem for a separate PR, since most of the other drivers in this file also use the expanded values...

For what little consolation we get from "it's a maintained evil, not a new one" 😅

PR #29135, opened 2025-01-24
ArduPlane:add options for minimum alt before turn to Autoland

Hwurzburg:
Last improvement that I can think of....honest


timtuxworth:
This seems a little awkward @Hwurzburg - could it not be a separate parameter with a dropdown, since only one of these is really valid at any time?

Hwurzburg:
lowest flash impact and we have already have bitmask based option params that have internal overrides of bits by bits within themselves....easy to change to an AUTOLAND_CLIMB param, just takes more flash

Hwurzburg:
actually a bit more flash and is less easy to use on a TX GCS to change alts than #29142, so closing

PR #29139, opened 2025-01-24
ROS2: Update from Humble to Jazzy

andrewkwolek:
I made some changes to make the repo compatible with Jazzy Jalisco which is the most recent build of ROS2. Most of these changes involved changing dependencies from `humble` to `jazzy`. I also changed micro_ros_agent to use the jazzy branch instead of the humble branch in Tools/ros2/ros2.repos.

I was able to successfully run the installation and SITL in the documentation (https://ardupilot.org/dev/docs/ros2-sitl.html). This did not take much time, so it makes me curious why this hasn't been done yet? It is possible there are some low level changes that need to be made that I am unaware of. Thanks!

### DIFF
--- a/.github/workflows/colcon.yml
+++ b/.github/workflows/colcon.yml
@@ -178,7 +178,7 @@ jobs:
           # Workaround for flake8 attribute error
           # https://github.com/ArduPilot/ardupilot/pull/24277#issuecomment-1632833433
           python -m pip install flake8==3.7.9
-          source /opt/ros/humble/setup.bash
+          source /opt/ros/jazzy/setup.bash
           rosdep install --from-paths src --ignore-src --default-yes
       - name: Install MAVProxy
         run: |
@@ -192,7 +192,7 @@ jobs:
       - name: Build with colcon
         shell: 'bash'
         run: |
-          source /opt/ros/humble/setup.bash
+          source /opt/ros/jazzy/setup.bash
           colcon build --packages-up-to ardupilot_dds_tests --cmake-args -DBUILD_TESTING=ON
       - name: Test with colcon
         shell: 'bash'
--- a/Tools/ros2/ros2.repos
+++ b/Tools/ros2/ros2.repos
@@ -6,4 +6,4 @@ repositories:
   micro_ros_agent:
     type: git
     url: https://github.com/micro-ROS/micro-ROS-Agent.git
-    version: humble
+    version: jazzy
--- a/Tools/ros2/ros2.repos
+++ b/Tools/ros2/ros2.repos
@@ -1,7 +1,7 @@
 repositories:
   ardupilot:
     type: git
-    url: https://github.com/ArduPilot/ardupilot.git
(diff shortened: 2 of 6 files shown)


andrewkwolek:
Looks like the container used for colcon build/test / build-test (pull_request) does not have jazzy installed on it.

andrewkwolek:
Is there a way to change the container https://github.com/ArduPilot/ardupilot_dev_docker/blob/master/docker/Dockerfile_dev-ros that the `colcon.yml` test uses to have jazzy installed instead of humble?

.github/workflows/colcon.yml
```diff
@@ -178,7 +178,7 @@ jobs:
           # Workaround for flake8 attribute error
           # https://github.com/ArduPilot/ardupilot/pull/24277#issuecomment-1632833433
           python -m pip install flake8==3.7.9
-          source /opt/ros/humble/setup.bash
+          source /opt/ros/jazzy/setup.bash
```
Ryanf55, line 181: Thanks for getting this started. I'd like to see a matrix build - we want to continue to support `humble`. 
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

Do you think you could try that?

Tools/ros2/ros2.repos
```diff
@@ -1,9 +1,9 @@
 repositories:
   ardupilot:
     type: git
-    url: https://github.com/ArduPilot/ardupilot.git
+    url: https://github.com/andrewkwolek/ardupilot.git
```
Ryanf55, line 4: Revert

Ryanf55 (changes requested):
Thanks! It's on our radar to migrate to Jazzy, but we haven't decided how to do that.

Your PR is a proposal to migrate immediately, without warning, and have no support of `humble` anymore, at least from CI. I propose we be a little more conservative, and find a way to phase the migration in.

.github/workflows/colcon.yml
andrewkwolek: Yes! I can give it a shot. I figured we wouldn't want to remove humble support so i kept this as a draft for feedback. I'll try making this a matrix build.

andrewkwolek:
One note. The container `container: image: ardupilot/ardupilot-dev-ros:latest` that colcon.yml runs on is set to test humble. I don't know if there is a way to run two different images from one Docker file so we may need an arupilot-dev-ros-jazzy and ardupilot-ros-dev-humble container. Does that sound okay?

Ryanf55:
Yes that's ok we can add containers like that.

andrewkwolek:
@Ryanf55 I made this PR in `ardupilot_dev_docker`: https://github.com/ArduPilot/ardupilot_dev_docker/pull/33

andrewkwolek:
@Ryanf55 I added the `options` field with the `--build-arg ROS-VERSION=${version}` tag.

PR #29141, opened 2025-01-25
AP_HAL_ChibiOS: add all BETAFPV F405 variants and re-enable baro-less operation

andyp1per:


andyp1per:
> * These have different voltage and current ranges....should have different batt param recommendations?
> 
> * The BEATFPV-F405-I2C uses the same pinout photo as the  V3, since it has different pins that should corrected also, correct? and its readme needs voltages updated possibly?

Betaflight has the same scale for all, so I am assuming the sensor is the same. The original scale I had was definitely wrong because of a bug in my conversion script.

The v3 does not appear to be capable of the UART3 mod, so we should change the photo to the other one

Hwurzburg (approved):

rmackay9:

rmackay9:
If possible it really helps when writing the ReleaseNotes if there's a description.  I'll try to write something but hopefully @andyp1per can check the upcoming PR and provide a description of this PR's impact

PR #29142, opened 2025-01-25
ArduPlane: add climb before turn to AUTOLAND

Hwurzburg:
Alternate to #29135

After test flying using a TX GCS to change climb to altitudes, I agree with Tim that this is easier to use then options in #29135...and it actually uses less flash so I am closing #29135

Now fully terrain following capable

Thanks to Pete Hall for help

Updated autotest for terrain based mission out behind mountain with terrain following autoland return


ArduPlane/mode.h
```diff
@@ -946,6 +946,9 @@ class ModeAutoLand: public Mode
     AP_Int16 final_wp_dist;
     AP_Int16 landing_dir_off;
     AP_Int8  options;
+    AP_Int16 climb_before_turn_alt;
```
timtuxworth, line 949: It seems to be a standard to put _m on altitude variables in meters. This is in meters is it not?

ArduPlane/mode_autoland.cpp
```diff
@@ -38,10 +38,20 @@ const AP_Param::GroupInfo ModeAutoLand::var_info[] = {
...
     AP_GROUPINFO("OPTIONS", 4, ModeAutoLand, options, 0),
 
+    // @Param: CLIMB
+    // @DisplayName: Climb before turning altitude
+    // @Description: Vehicle will climb with limited turn ability (LEVEL_ROLL_LIMIT) upon mode entry to this altitude, before proceeding to loiter-to-alt and landing legs. If a value less than AUTOLAND_WP_ALT is entered, AUTOLAND_WP_ALT will be used.
+    // @Range: 0 500
+    // @Increment: 1
+    // @Units: m
+    // @User: Standard
+    AP_GROUPINFO("CLIMB", 5, ModeAutoLand, climb_before_turn_alt,0),
```
timtuxworth, line 52: ditto climb_before_turn_alt_m

ArduPlane/mode_autoland.cpp
```diff
@@ -237,5 +274,16 @@ void ModeAutoLand::arm_check(void)
     }
 }
 
+// climb before turn altitude check
+void ModeAutoLand::check_altitude(void)
+{
+    const uint16_t altitude = plane.relative_ground_altitude(false,false);
```
timtuxworth, line 280: Sorry clicked the wrong button. this is also meters I think.

timtuxworth (changes requested):
I think that's it just _m on values in meters.

ArduPlane/mode.h
Hwurzburg: my understanding is that SI unit based variables do not need suffix, only those differing, ie _cm,_cd

ArduPlane/mode_autoland.cpp
Hwurzburg: ditto my comment above

ArduPlane/mode_autoland.cpp
Hwurzburg: ditto

ArduPlane/mode.h
timtuxworth: Should it be clarified in the style guide? https://ardupilot.org/dev/docs/style-guide.html

Hwurzburg: Assuming my understanding is correct

ArduPlane/mode_autoland.cpp
```diff
@@ -152,14 +164,39 @@ bool ModeAutoLand::_enter()
 void ModeAutoLand::update()
 {
     plane.calc_nav_roll();
+
+    // Constrain the roll limit if climb before turn options are set when below that altitude,
+    //that way if something goes wrong the plane will
+    // eventually turn back and go to landing waypoint instead of going perfectly straight. This also leaves
+    // some leeway for fighting wind.
+    check_altitude();
```
IamPete1, line 172: your only using this function in one spot, might as a well inline it.

ArduPlane/mode_autoland.cpp
```diff
@@ -152,14 +164,39 @@ bool ModeAutoLand::_enter()
...
+    uint16_t roll_limit_cd;
+    if (!done_climb) {
+        roll_limit_cd = MIN(plane.roll_limit_cd, plane.g.level_roll_limit*100);
+        plane.nav_roll_cd = constrain_int16(plane.nav_roll_cd, -roll_limit_cd, roll_limit_cd);
+        // set CLIMB before turn alt as target if needed
+        if (climb_before_turn_alt > final_wp_alt) {
+            plane.target_altitude.amsl_cm = plane.home.alt + climb_before_turn_alt * 100.0f;
```
IamPete1, line 179: I would have expected this in the enter not update.

ArduPlane/mode_autoland.cpp
```diff
@@ -72,7 +82,7 @@ bool ModeAutoLand::_enter()
         gcs().send_text(MAV_SEVERITY_WARNING, "Takeoff initial direction not set");
         return false;
     }
-
+    climb_before_turn_alt = (climb_before_turn_alt < final_wp_alt)? final_wp_alt : climb_before_turn_alt;
```
IamPete1, line 85: this is directly changing the parameter? I would have expected a compiler error....

IamPete1: Looks like you get away with it because both possible values are `AP_Int16` params so its gets the default operator.

IamPete1: In anycase, we try not to set params like this if we can avoid it.

IamPete1 (commented):
Could we reuse `RTL_CLIMB_MIN`?

ArduPlane/mode.h
timtuxworth: My understanding was that we did not add the suffix to the parameter name, but we did to the internal c++ variable.  Maybe I misunderstood.

Hwurzburg: confirmed with Peter...we dont add suffix unless not SI or its confusing since there is a mismash of similar names in the module with differing units....will update the style guide to make it clear...

ArduPlane/mode_autoland.cpp
Hwurzburg: cant use RTL_CLIMB_MIN.....its a delta...want an absolute like RTL_ALTITUDE + CLIMB before turn option....in case AUTOLAND_WP_ALT (a final approach alt) wont cut it because of obstacles....will inline it and use flaot

ArduPlane/mode_autoland.cpp
Hwurzburg: works there too I think...will try

ArduPlane/mode_autoland.cpp
Hwurzburg: will assign to another new variable...just trying save memory

ArduPlane/mode_autoland.cpp
```diff
@@ -152,14 +167,41 @@ bool ModeAutoLand::_enter()
 void ModeAutoLand::update()
 {
     plane.calc_nav_roll();
+
+    // check if initial climb before turn altitude has been reached
+    const float altitude = plane.relative_ground_altitude(false,false);
```
tridge, line 172: why is this an above home alt? why not use the terrain alt if available?

ArduPlane/mode_autoland.cpp
```diff
@@ -38,10 +38,20 @@ const AP_Param::GroupInfo ModeAutoLand::var_info[] = {
...
     // @DisplayName: Autoland mode options
     // @Description: Enables optional autoland mode behaviors
-    // @Bitmask: 1: When set if there is a healthy compass in use the compass heading will be captured at arming and used for the AUTOLAND mode's initial takeoff direction instead of capturing ground course in NAV_TAKEOFF or Mode TAKEOFF or other modes.
+    // @Bitmask: 0: When set if there is a healthy compass in use the compass heading will be captured at arming and used for the AUTOLAND mode's initial takeoff direction instead of capturing ground course in NAV_TAKEOFF or Mode TAKEOFF or other modes.
     // @User: Standard
     AP_GROUPINFO("OPTIONS", 4, ModeAutoLand, options, 0),
 
+    // @Param: CLIMB
+    // @DisplayName: Climb before turning altitude
+    // @Description: Vehicle will climb with limited turn ability (LEVEL_ROLL_LIMIT) upon mode entry to this altitude, before proceeding to loiter-to-alt and landing legs. If a value less than AUTOLAND_WP_ALT is entered, AUTOLAND_WP_ALT will be used.
```
tridge, line 47: we need to specify what type of altitude it is using

tridge (changes requested):

ArduPlane/mode_autoland.cpp
Hwurzburg: now corrected

Hwurzburg:
@tridge need help in getting glide_slope from point of climb to alt -> base loiter point to follow terrain...it always just use above home alt as it proceeds to the loiter-to-alt point....help would be appreciated

Hwurzburg:
@tridge got it all working now.....full terrain support

ArduPlane/mode_autoland.cpp
```diff
@@ -38,10 +38,20 @@ const AP_Param::GroupInfo ModeAutoLand::var_info[] = {
...
     // @DisplayName: Autoland mode options
     // @Description: Enables optional autoland mode behaviors
-    // @Bitmask: 1: When set if there is a healthy compass in use the compass heading will be captured at arming and used for the AUTOLAND mode's initial takeoff direction instead of capturing ground course in NAV_TAKEOFF or Mode TAKEOFF or other modes.
+    // @Bitmask: 0: When set if there is a healthy compass in use the compass heading will be captured at arming and used for the AUTOLAND mode's initial takeoff direction instead of capturing ground course in NAV_TAKEOFF or Mode TAKEOFF or other modes.
     // @User: Standard
     AP_GROUPINFO("OPTIONS", 4, ModeAutoLand, options, 0),
 
+    // @Param: CLIMB
+    // @DisplayName: Climb before turning altitude
+    // @Description: Vehicle will climb with limited turn ability (LEVEL_ROLL_LIMIT) upon mode entry to this altitude, before proceeding to loiter-to-alt and landing legs. If a value less than AUTOLAND_WP_ALT is entered, AUTOLAND_WP_ALT will be used. If TERRAIN is enabled and healthy, the altitude is above terrain, otherwise above HOME.
```
timtuxworth, line 47: Sorry to seem to be nitpicking @Hwurzburg but should this only apply if TERRAIN_FOLLOW is enabled - either all or for AUTOLAND? (which would mean adding a bit for AUTOLAND to the bit mask).

ArduPlane/mode_autoland.cpp
```diff
@@ -152,14 +168,46 @@ bool ModeAutoLand::_enter()
 void ModeAutoLand::update()
 {
     plane.calc_nav_roll();
+
+    // check if initial climb before turn altitude has been reached
+    const float altitude = plane.relative_ground_altitude(false,true); //above terrain if enabled,otherwise above home
+    // climb to climb_before_turn alt: if terrian enabled its delta above terrain, otherwise above home
```
timtuxworth, line 174: spelling terrian

timtuxworth (commented):
It seems to me that just because TERRAIN_ENABLE is true might not mean you want to use it for AUTOLAND. I guess it could be added to the TERRAIN_FOLLOW bitmask in a follow up PR.

ArduPlane/mode_autoland.cpp
Hwurzburg: not necessarily, but the mode needs a bit in the follow table anyway, so if terrain following, the climb will be above terrain....the final WP alt MUST always be above home, lest you crash into a gully before the runway...

ArduPlane/mode_autoland.cpp
Hwurzburg: fixed

ArduPlane/mode_autoland.cpp
```diff
@@ -138,12 +148,18 @@ bool ModeAutoLand::_enter()
     cmd_loiter.content.location = land_start;
     cmd_loiter.content.location.offset_bearing(plane.takeoff_state.initial_direction.heading + bearing_offset_deg, corrected_loiter_radius);
     cmd_loiter.content.location.loiter_ccw = bearing_err_deg>0? 1 :0;
-
+#if AP_TERRAIN_AVAILABLE
+    if (plane.terrain_enabled_in_current_mode()) {
+        cmd_loiter.content.location.terrain_alt = 1; // this allows terrain following to it as a glide slop destination
```
tridge, line 153: i'm surprised you didn't need to subtract home alt here, I will need to look at this under gdb

ArduPlane/mode_autoland.cpp
```diff
@@ -152,14 +168,46 @@ bool ModeAutoLand::_enter()
...
+    uint16_t roll_limit_cd;
+    if (!done_climb) {
+        // Constrain the roll limit during climb before turn. This also leaves
+        // some leeway for fighting wind.
+        roll_limit_cd = MIN(plane.roll_limit_cd, plane.g.level_roll_limit*100);
+        plane.nav_roll_cd = constrain_int16(plane.nav_roll_cd, -roll_limit_cd, roll_limit_cd);
+        //climb at TECS max until alt reached
+        plane.target_altitude.amsl_cm = plane.current_loc.alt + climb_before_turn_alt * 100.0f;
+#if AP_TERRAIN_AVAILABLE
+        plane.target_altitude.terrain_alt_cm = plane.target_altitude.amsl_cm;
```
IamPete1, line 187: I suspect this is key to how why works. It sets the TECS target to a huge altitude which is enough to get the vehicle to do a max rate climb. That is the only purpose of the target altitude. The key thing is that this target is not used for the comparison of when the climb is done.

The problem is that without terrain, or if your at sea level then the target altitude may be lower than the threshold alt and the climb will never complete. 

We could just set `INT32_MAX` here and the result would be the same. Although I would not recommend that.

Hwurzburg: without terrain it climbs to current amsl + climb alt in cm....so that's correct, right?
how do you suggest I set the max climb , terrain based?

tridge (approved):

PR #29143, opened 2025-01-26
gitignore: ignore .python-version file used by pyenv

shiv-tyagi:


Ryanf55 (commented):
Makes sense you'd want to ignore this if you are using a different python version for AP than what is the default for the OS or higher folder.

Ryanf55 (approved):

shiv-tyagi:
> Makes sense you'd want to ignore this if you are using a different python version for AP than what is the default for the OS or higher folder.

Yes, one can even mention a pyenv virtual environment name in that file and it automatically switches to that environment as soon as we enter the AP directory in your shell.

PR #28780, opened 2024-11-30
Scripting promote video-stream-information to an applet

rmackay9:
This enhances the set_VIDEO_STREAM_INFORMATION.lua to make it usable without any code changes by at least some users.  The script is also promoted to an applet

The purpose of this script is to give the AP_Camera library the information it needs so that it can send [VIDEO_STREAM_INFORMATION](https://mavlink.io/en/messages/common.html#VIDEO_STREAM_INFORMATION) mavlink messages to the ground station so that it can display the live video stream to the user

15 new parameters have been added prefixed with VID1_ that allow the user to set nearly all fields of the VIDEO_STREAM_INFORMATION but for most users all they will only need to set these six:

- VID1_CAMMODEL: Video stream camera model (1:Siyi A8, 2:Siyi ZR10, 3:Siyi ZR30, 4:Siyi ZT30 Zoom, 5:Siyi ZT30 Wide, 6:Siyi ZT30 IR, 7:Siyi ZT6 RGB, 8:Siyi ZT6 IR)
- VID1_IPADDR0: Video stream IP Address first octet
- VID1_IPADDR1: Video stream IP Address second octet
- VID1_IPADDR2: Video stream IP Address third octet
- VID1_IPADDR3: Video stream IP Address fourth octet
- VID1_IPPORT: Video Stream IP Address Port

This has been successfully tested with the Siyi A8, ZR10 and ZT6 cameras and [Mission Planner's new gimbal interface](https://github.com/ArduPilot/MissionPlanner/pull/3386)

Enhancements coming in future PRs:

- default the IP address and Port to the camera defaults
- add support for ViewPro, Topotek and any other Ethernet enabled camera gimbal

PR #29146, opened 2025-01-27
AP_Scripting: video-stream-information param index fix

rmackay9:
This fixes the video-stream-information.lua script's parameter table key so that it doesn't conflict with the [ahrs-set-origin.lua](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/applets/ahrs-set-origin.lua) script

This overlap is my fault but also quite recent (6 weeks ago)

- Nov 29th PR https://github.com/ArduPilot/ardupilot/pull/28770
- Nov 30th PR https://github.com/ArduPilot/ardupilot/pull/28780

I know of 1 partner using the ahrs-set-origin script and at most 2 developers using video-stream-information.lua.  I figure the two developers will be able to handle the change more easily which is why I've chosen to modify the video script.

### DIFF
--- a/libraries/AP_Scripting/applets/video-stream-information.lua
+++ b/libraries/AP_Scripting/applets/video-stream-information.lua
@@ -2,7 +2,7 @@
     Populate the VIDEO_STREAM_INFORMATION message based on user parameters
  --]]
 
-local PARAM_TABLE_KEY = 87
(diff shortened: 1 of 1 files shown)


tridge (approved):

rmackay9:

Issue #29129, opened 2025-01-23
[Feature] Two different modes of OSD working simultaneously


lida2003:
## Feature request

Two different modes of OSD working simultaneously

Details: https://discuss.ardupilot.org/t/is-it-possible-for-two-osd-resolution-working-at-the-same-time/125505

**Platform**
[  ] All
[  ] AntennaTracker
[x] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Additional context**
Copter 4.5.6

```
25/1/23 17:35:56 : ChibiOS: 6a85082c
25/1/23 17:35:56 : ArduCopter V4.5.6 (b867abc8)
25/1/23 17:35:56 : Frame: QUAD/X_REV
25/1/23 17:35:56 : RCOut: DS600:1-4 PWM:5-13
25/1/23 17:35:56 : Aocoda-RC-H743Dual 00290029 30315119 
```

lida2003:
How about add an `OSD_CONCURRENTLY = 3` below for anlog+msp display port as guide says: https://ardupilot.org/copter/docs/common-osd-overview.html#osd

https://github.com/ArduPilot/ardupilot/blob/00c1f1456e8393929ece305b667deea0688aad7d/libraries/AP_OSD/AP_OSD.h#L560-L563

Add code when following conditions are met:
- `switch_method  = OSD_CONCURRENTLY `
- OSD screen is enabled
- OSD_TYPEx is porperly set


https://github.com/ArduPilot/ardupilot/blob/00c1f1456e8393929ece305b667deea0688aad7d/libraries/AP_OSD/AP_OSD.cpp#L555-L605

lida2003:
I also find an issue when OSD init:
- init procedure asked if it's compatible
- MSP Display responses with `false`
- then the `init_backend` returns, there will be NO probe
- finally, `_backend_count` will not be increased in `AP_OSD::init`

I'm NOT sure if it works this way?


https://github.com/ArduPilot/ardupilot/blob/00c1f1456e8393929ece305b667deea0688aad7d/libraries/AP_OSD/AP_OSD.cpp#L321-L328

https://github.com/ArduPilot/ardupilot/blob/00c1f1456e8393929ece305b667deea0688aad7d/libraries/AP_OSD/AP_OSD_MSP_DisplayPort.h#L34-L38

lida2003:
I think I may understand OSD not correctly. 

Here is the case:
[...]

PR #29149, opened 2025-01-27
AP_OSD: add two osd resolution concurrently support

lida2003:


peterbarker:
Closes https://github.com/ardupilot/ardupilot/issues/29129

PR #29150, opened 2025-01-27
Remove vestiges of LearnType::LEARN_INTERNAL

peterbarker:
57a3bc1397239cbd66e5b32d59dcc467003e4bad changed the code from "internal" to "in-flight

It seems the old value of "1" was no longer valid

It also changed things so that the learning system saved the offsets.

We have two sorts of compass learning:
 - in-flight compass learning, type=3
 - copy-values-from-EKF, type=2

Copy-values-from-EKF sets-and-saves offsets on a per-vehicle basis at the moment, eg. https://github.com/ardupilot/ardupilot/blob/master/ArduCopter/AP_Arming.cpp#L814 so it doesn't want or need this code.

The in-flight learning system sets-and-saves the offsets via a called to fixed-mag-cal-yaw code here: https://github.com/ardupilot/ardupilot/blob/master/libraries/AP_Compass/Compass_learn.cpp#L65 (https://github.com/ardupilot/ardupilot/master/libraries/AP_Compass/AP_Compass_Calibration.cpp#L566)

This does seem to have been some sort of back-door for saving compass offsets on Rover, Plane and Tracker, bypassing "forcecal" requirements.

### DIFF
--- a/AntennaTracker/Tracker.cpp
+++ b/AntennaTracker/Tracker.cpp
@@ -53,7 +53,6 @@ const AP_Scheduler::Task Tracker::scheduler_tasks[] = {
     SCHED_TASK(update_tracking,        50,    1000, 15),
     SCHED_TASK(update_GPS,             10,    4000, 20),
     SCHED_TASK(update_compass,         10,    1500, 25),
-    SCHED_TASK(compass_save,            0.02,  200, 30),
     SCHED_TASK_CLASS(AP_BattMonitor,    &tracker.battery,   read,           10, 1500, 35),
     SCHED_TASK_CLASS(AP_Baro,          &tracker.barometer,  update,         10, 1500, 40),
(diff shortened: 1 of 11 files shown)


rmackay9:
I'm personally happy with this.  Looking at our [advanced compass setup wiki page](https://ardupilot.org/copter/docs/common-compass-setup-advanced.html) it says the following which is still correct after this change

`
Setting COMPASS_LEARN to 1 or 2 is not recommended. These modes are deprecated and are either non-functional, or still in development.
`

peterbarker:
> I'm personally happy with this. Looking at our [advanced compass setup wiki page](https://ardupilot.org/copter/docs/common-compass-setup-advanced.html) it says the following which is still correct after this change
> 
> `Setting COMPASS_LEARN to 1 or 2 is not recommended. These modes are deprecated and are either non-functional, or still in development.`

Any idea what the "copy offsets from EKF" code's status is?  Still in dev or is it deprecated?

tridge (approved):

PR #29152, opened 2025-01-27
Plane:adjust CTUN.Pitch to remove PITCH_TRIM

Hwurzburg:
We have ATT.Pitch, CTUN.Pitch, XKF1.Pitch all reporting the same pitch with PITCH_TRIM added, so there are no two log lines that one can graph showing desired/nav pitch vs reported pitch to check if pitch tuning is closely matching to indicate good tune....

removing PITCH_TRIM_DEG bias from one would allow an easy graphical evaluation of Pitch tune....doing it in ATT is problematic since its in AHRS and spans vehicle types....modifying CTUN.Pitch is easy and only affects FW Plane...the other two pitch logs are unaffected.

This also aligns with HUD artificial horizon visual indication

### DIFF
--- a/ArduPlane/Log.cpp
+++ b/ArduPlane/Log.cpp
@@ -102,13 +102,19 @@ void Plane::Log_Write_Control_Tuning()
         synthetic_airspeed = logger.quiet_nan();
     }
 
+    int16_t pitch = ahrs.pitch_sensor - g.pitch_trim * 100;
+#if HAL_QUADPLANE_ENABLED
+    if (quadplane.show_vtol_view()) {
+        pitch = quadplane.ahrs_view->pitch_sensor;
+    }
+#endif      
     struct log_Control_Tuning pkt = {
         LOG_PACKET_HEADER_INIT(LOG_CTUN_MSG),
         time_us         : AP_HAL::micros64(),
         nav_roll_cd     : (int16_t)nav_roll_cd,
         roll            : (int16_t)ahrs.roll_sensor,
         nav_pitch_cd    : (int16_t)nav_pitch_cd,
-        pitch           : (int16_t)ahrs.pitch_sensor,
+        pitch           : pitch,
         throttle_out    : SRV_Channels::get_output_scaled(SRV_Channel::k_throttle),
         rudder_out      : SRV_Channels::get_output_scaled(SRV_Channel::k_rudder),
         throttle_dem    : TECS_controller.get_throttle_demand(),
@@ -312,8 +318,8 @@ const struct LogStructure Plane::log_structure[] = {
 // @Field: TimeUS: Time since system startup
 // @Field: NavRoll: desired roll
 // @Field: Roll: achieved roll
-// @Field: NavPitch: desired pitch
-// @Field: Pitch: achieved pitch
+// @Field: NavPitch: desired pitch assuming pitch trims are already applied
+// @Field: Pitch: achieved pitch assuming pitch trims are already applied,ie "0deg" is level flight trimmed pitch attitude as shown on artifical horizon level line.
 // @Field: ThO: scaled output throttle
 // @Field: RdO: scaled output rudder
 // @Field: ThD: demanded speed-height-controller throttle


Georacer:
@Hwurzburg perhaps you'd like to take a look at https://github.com/ArduPilot/ardupilot/pull/29067

I think it gets the same result as this one (have comparable pitch quantities) by following a complementary path.

Hwurzburg:
I am  against adding Pitch trim in desired to make it match pitch ..then when flying level the desired pitch no longer is zero ...zero  should indicate the vehicle is level on its artificial horizon

ArduPlane/Log.cpp
```diff
@@ -102,13 +102,19 @@ void Plane::Log_Write_Control_Tuning()
         synthetic_airspeed = logger.quiet_nan();
     }
 
+    int16_t pitch = ahrs.pitch_sensor - g.pitch_trim * 100.0f;
+#if HAL_QUADPLANE_ENABLED
+    if (quadplane.in_vtol_mode()) {
+    pitch = ahrs.pitch_sensor;
+    }
+#endif
```
IamPete1, line 110: I don't think you need this case, the existing code is already wrong for quadplanes.

IamPete1 (changes requested):

ArduPlane/Log.cpp
IamPete1: I don't understand the reasoning, both this and the un-offset value are not correct for VTOL. It's just going to be confusing.

ArduPlane/Log.cpp
```diff
@@ -102,13 +102,19 @@ void Plane::Log_Write_Control_Tuning()
         synthetic_airspeed = logger.quiet_nan();
     }
 
+    int16_t pitch = ahrs.pitch_sensor - g.pitch_trim * 100.0f;
```
tridge, line 105: if we are applying TRIM_PITCH_DEG in fixed wing, then we should also be applying Q_TRIM_PITCH in VTOL modes

tridge (dismissed):

ArduPlane/Log.cpp
Hwurzburg: I could be mistake, but I think Q_TRIM_PITCH is already applies to ahrs.pitch_sensor...maybe its in error or maybe me looking at logs to determine this is wrong...here is an SLT loitering in no wind with a Q_TRIM_PITCH of 3 degrees

Hwurzburg:
okay, I think this is good now....here is a SITL log of a quadplane in all different light modes with this logging now:
graph

ArduPlane/Log.cpp
Hwurzburg: looks good to me in the logs now

ArduPlane/Log.cpp
Hwurzburg: nope just out of navpitch

ArduPlane/Log.cpp
```diff
@@ -102,13 +102,19 @@ void Plane::Log_Write_Control_Tuning()
         synthetic_airspeed = logger.quiet_nan();
     }
 
+    int16_t pitch = ahrs.pitch_sensor - g.pitch_trim * 100;
+#if HAL_QUADPLANE_ENABLED
+    if (quadplane.in_vtol_mode()) {
+    pitch = ahrs.pitch_sensor - quadplane.ahrs_trim_pitch * 100;
+    }
```
IamPete1, line 109:
```suggestion
    if (quadplane.show_vtol_view()) {
        pitch = quadplane.ahrs_view->pitch_sensor;
    }
```

IamPete1 (changes requested):

ArduPlane/Log.cpp
IamPete1: For Quadplane we use the ahrs view which includes pitch trim already.

Hwurzburg: changed....graphs in flight are the same, so I am happy

IamPete1 (approved):
Change looks good, should we add a note to the log message documentation here:

https://github.com/ArduPilot/ardupilot/blob/2cd3970e3620a2770fdbe8e0b2d528d9db65dcc8/ArduPlane/Log.cpp#L315-L316

Hwurzburg:
added...a bit long but prevents confusion I think

PR #29153, opened 2025-01-27
Eahrs remove gps serial assumption

Ryanf55:
As discussed in Trimble partner chat, we are looking to add a new EAHRS that uses ethernet connection. The GPS is not connected via serial, there is no serial port. It uses UDP network API instead.

I propose we only check that the EARHS has the GPS sensor enabled.

The call to `get_port` prevents the upcoming Trimble PX-1 from being able to initialize, because `get_port` returns `-1`.

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -662,7 +662,7 @@ AP_GPS_Backend *AP_GPS::_detect_instance(uint8_t instance)
 #endif
 #if HAL_EXTERNAL_AHRS_ENABLED
     case GPS_TYPE_EXTERNAL_AHRS:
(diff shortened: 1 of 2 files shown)


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -662,7 +662,7 @@ AP_GPS_Backend *AP_GPS::_detect_instance(uint8_t instance)
 #endif
 #if HAL_EXTERNAL_AHRS_ENABLED
     case GPS_TYPE_EXTERNAL_AHRS:
-        if (AP::externalAHRS().get_port(AP_ExternalAHRS::AvailableSensor::GPS) >= 0) {
+        if (!AP::externalAHRS().has_sensor(AP_ExternalAHRS::AvailableSensor::GPS)) {
```
peterbarker, line 665: Are you inverting this check?

PR #29154, opened 2025-01-27
waf: append new commands on files to compile_commands.json if different arguments are used

bugobliterator:


MichelleRos:
Tested, and now IntelliSense doesn't seem to work at all - if I just build for a vehicle I still get errors in that vehicle's files.

PR #29158, opened 2025-01-28
Make compass LearnType enum-class and parameter AP_Enum

peterbarker:
- also adds set_and_save to AP_Enum and fixes some problems around that increased type safety.

Removes an old unused method with a bug in it.


MattKear (approved):
A nice tidy up 👍

PR #26157, opened 2024-02-06
AP_Scripting: add ROMFS to lua path

IamPete1:
This allows `require` to find modules in ROMFS. Possibly helps with other file system methods too. Note that this adds the new search path even if we don't have ROMFS compiled in. I think it will be fine, but I have not tested. 

Adding `"./modules/?.lua;" "./modules/?/init.lua"` to get to the same place relative to a script loaded in ROMFS does not seem to work, I'm not sure why. 

This should allow a SD card script to require something from ROMFS, which using relative paths would not. Because of the ordering it should also mean a SD card version of the same file gets priority over a ROMFS copy. Its convenient for development but might give users a very exciting way to break stuff. We could swap the ordering so ROMFS takes priority. Again, I have not tested this.

PR #29161, opened 2025-01-28
AP_HAL_ChibiOS: chibios_hwdef.py tidy embedding of files into ROMFS

peterbarker:
most notably this embed the "modules" directory so you can use modules in ROMFS

I've created a branch which shows this working here: https://github.com/ArduPilot/ardupilot/compare/master...peterbarker:ardupilot:DEBUG/demo-pr/modules-in-romfs?expand=1 .  It creates an OEM board which embeds a script and a library for it to use.  Note that there are several drive-by patches in that branch at this point.

image


IamPete1:
I certainty tested this when I did https://github.com/ArduPilot/ardupilot/pull/26157, it worked then. Could the difference be using `ROMFS_custom` vs hwdef?

peterbarker:
> I certainty tested this when I did #26157, it worked then. Could the difference be using `ROMFS_custom` vs hwdef?

They're completely different paths, both walking the tree.

OTOH, it turns out this "fix" isn't required at all.  No idea what went wrong to make me think it was required.

Changed this to just neaten the function up, rather than modifying its functionality

PR #29163, opened 2025-01-28
AP_HAL_ChibiOS: create and use AP_PERIPH_AHRS_ENABLED

shiv-tyagi:
Part of the series of PRs (like #28127 , #29128) aimed at renaming peripheral defines to the new AP_PERIPH_XX_ENABLED scheme and ensuring that those are assigned some value instead of just being defined.


peterbarker (approved):
LGTM

PR #29164, opened 2025-01-28
Tools: uploader.py: verify correct install

rafasaurus:
- Implement functionality to verify correct firmware installation
- Utilize existing code with an additional argument for verification

### DIFF
--- a/Tools/scripts/uploader.py
+++ b/Tools/scripts/uploader.py
@@ -882,6 +882,20 @@ class uploader(object):
             print("Failed to get name: %s" % str(e))
         return None
 
+    # Verify firmware version on board matches provided version
+    def verify_firmware_is(self, fw, boot_delay=None):
+        if self.bl_rev == 2:
+            self.__verify_v2("Verify ", fw)
+        else:
+            self.__verify_v3("Verify ", fw)
+
+        if boot_delay is not None:
+            self.__set_boot_delay(boot_delay)
+
+        print("\nRebooting.\n")
+        self.__reboot()
+        self.port.close()
+
     # upload the firmware
     def upload(self, fw, force=False, boot_delay=None):
         # Make sure we are doing the right thing
@@ -1120,6 +1134,8 @@ def main():
     )
     parser.add_argument('--download', action='store_true', default=False, help='download firmware from board')
     parser.add_argument('--identify', action="store_true", help="Do not flash firmware; simply dump information about board")
+    parser.add_argument('--verify-firmware-is', action="store_true",
+                        help="Do not flash firmware; verify that the firmware on the board matches the supplied firmware")
     parser.add_argument('--no-extf', action="store_true", help="Do not attempt external flash operations")
     parser.add_argument('--erase-extflash', type=lambda x: int(x, 0), default=None,
                         help="Erase sectors containing specified amount of bytes from ext flash")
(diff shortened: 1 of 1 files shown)


shifter97 (approved):

Tools/scripts/uploader.py
```diff
@@ -1120,6 +1134,7 @@ def main():
     )
     parser.add_argument('--download', action='store_true', default=False, help='download firmware from board')
     parser.add_argument('--identify', action="store_true", help="Do not flash firmware; simply dump information about board")
+    parser.add_argument('--verify', action="store_true", help="Do not flash firmware; simply verify the firmware")
```
peterbarker, line 1137:
```suggestion
    parser.add_argument('--verify', action="store_true", help="Do not flash firmware; verify that the firmware on the board matches the supplied firmware")
```

peterbarker (changes requested):
"verify" is really, "verify that the flash worked".

You're re-using the functionality, but that terminology isn't really applicable in this case.

I'm OK with the current name of the parameter, but if we can find something better that would be good!

`--assert-firmware-is` would work, but a little verbose

In any case, a slightly better description is called for...

peterbarker (changes requested):
... also, please fold all of this into the one commit .  The one starting with `Tools:` looks appropriate.

Tools/scripts/uploader.py
timtuxworth: How about --validate?

rafasaurus:
> ... also, please fold all of this into the one commit . The one starting with `Tools:` looks appropriate.

Should we just squash and merge?

Tools/scripts/uploader.py
rafasaurus: --verify-installed or perhaps --validate-installation

Tools/scripts/uploader.py
```diff
@@ -882,6 +882,20 @@ def board_name_for_board_id(self, board_id):
             print("Failed to get name: %s" % str(e))
         return None
 
+    # verify the firmware
+    def verify(self, fw, boot_delay=None):
```
peterbarker, line 886:
```suggestion
    def verify_firmware_is(self, fw, boot_delay=None):
```

Tools/scripts/uploader.py
```diff
@@ -1120,6 +1134,8 @@ def main():
     )
     parser.add_argument('--download', action='store_true', default=False, help='download firmware from board')
     parser.add_argument('--identify', action="store_true", help="Do not flash firmware; simply dump information about board")
+    parser.add_argument('--verify', action="store_true",
```
peterbarker, line 1137:
```suggestion
    parser.add_argument('--verify-firmware-is', action="store_true",
```

peterbarker (changes requested):
Please avoid merge commits in your history - this needs a rebase again.

I've also requested a change you suggested yourself.  I think the suggestion makes it unambiguous what the new feature does?

rafasaurus:
> Please avoid merge commits in your history - this needs a rebase again.
> 
> I've also requested a change you suggested yourself. I think the suggestion makes it unambiguous what the new feature does?

Changed, rebased and tested on the board.

peterbarker (approved):
The permissions changed on the modified file from 0755 to 0644.  I've fixed this.

I've tested this locally and it works as expected.

Marking as MergeOnCIPass - thanks!

PR #5909, opened 2017-03-20
Navigation: Loiter radius calculations

WickedShell:
This addresses #5809 which is a plane issue relating to loiter radius's scaling with altitude and how that can cause problems.

This takes the approach of allowing the user to specify the bank angle their airframe can safely maintain at sea level. If this is provided we then always fly the currently requested loiter angle, or the lateral acceleration limit from sea level multiplied by the EAS2TAS scaler which ever is larger. This allows us to safely fly a consistent radius until the airframe would start to be excessively stressed by the bank angle at higher altitudes.

The default behavior of setting the parameter uses the current scaling logic, which will keep the aircraft at a reasonable (if excessive) radius at all altitudes.

There are a couple of issues with this at the moment:
1. ~Rover doesn't compile, as a AP_L1_Control now needs a reference to a SpdHgtControl object. I was wondering if anyone had any clever workarounds for this, otherwise I was going to address it by instead providing it with a constant pointer, and checking that the pointer isn't null. (This is straightforward, I was just curious if there was a better way to approach this).~ EDIT: Just did it in this manner.
2. The parameter name is pretty awful. I wanted to call it some variation on LIM_LOITER_BANK but I couldn't fit that in. I fully expect a rename to happen, hoping to source that from the DevCall. (And I need to hide this param from the rover build)

Other changes: (done in their own commits so we can easily drop them if someone doesn't like them)
- Fixed inconsistent whitespace in AP_Navigation, it was really bothering me.
- Updated deepstall to use the loiter radius that will actually be flown, which creates smoother approaches.

PR #29165, opened 2025-01-28
Plane: Fix loiter navigation accuracy

rubenp02:
# Plane: Fix loiter navigation accuracy

Description
-----------
This PR addresses two major issues with loiter navigation in Plane:

1. Incorrect altitude compensation for the loiter radius
2. Lack of steady-state crosstrack error correction in loiter navigation

### Altitude Compensation

The current implementation in master overcompensates for the effects of altitude on turn radius. It applies a correction factor that is only necessary for the tightest possible loiter to loiters of all sizes. This behavior was improved in #5909, but in a way that is specific to loiters and requires parameter setup to take advantage of.

As a result, loiters weren't performed at the correct radius except when flying close to sea level. Moreover, since TECS is based on TAS, no explicit altitude compensation is actually needed. Loiters naturally adjust in size depending on what the aircraft can handle at the current altitude (and always match the requested radius when possible).

This can be seen in the following graph:

latitude_comparison

The explicit altitude correction currently used in master and the implicit correction the TECS naturally provides are nearly identical. They both perform loiters at the requested radius until the altitude increase makes it impossible, and then adjust the radius to be as close as possible to the requested one while avoiding a stall. However, the implicit correction doesn't require any parameter setup, and the difference between the flown radius and the requested one is reflected in the flight logs navigation crosstrack error field, which can be useful.

Therefore, the explicit altitude correction should be removed. But since an approximation of the loiter radius the aircraft will achieve on certain conditions is needed for navigation, the loiter radius compensation has been reworked to provide this information, but without having a direct impact on navigation.

### Steady-State Crosstrack Correction

Even after addressing the altitude compensation issues, the lack of steady-state crosstrack correction in the loiter navigation procedure means that loiter navigation still isn't accurate, particularly in smaller loiters. This PR adds an integral term to the circling PD (now PID) controller.

## Changes

1. Removed explicit loiter radius altitude compensation.
2. Removed the NAVL1_LIM_BANK parameter. Now, the ROLL_LIMIT_DEG is assumed to be safely maintainable at sea level.
3. Reworked the loiter radius compensation to work for any given airspeed and altitude. This allows for more accurate compensation in the AUTOLAND mode, and also to determine if any given loiter is viable, which could be useful to implement a pre-arm warning about a given loiter mission item that is going to be flown with a larger radius, for example.
4. Added an integral term to the loiter navigation PID controller in AP_L1_Control
5. Added navigation mode enum and member to keep track of the current nav. mode

---

<details>
<summary>Screenshot comparison (all default params.)</summary>

### Original navigation

imagen

### Altitude issues fixed

imagen

### Altitude issues fixed (smaller loiters)

imagen

### Integral term added (smaller loiters)

imagen
</details>

### DIFF
--- a/libraries/AP_L1_Control/AP_L1_Control.cpp
+++ b/libraries/AP_L1_Control/AP_L1_Control.cpp
@@ -37,7 +37,7 @@ const AP_Param::GroupInfo AP_L1_Control::var_info[] = {
 // S. Park, J. Deyst, and J. P. How, "A New Nonlinear Guidance Logic for Trajectory Tracking,"
 // Proceedings of the AIAA Guidance, Navigation and Control
 // Conference, Aug 2004. AIAA-2004-4900.
-// Modified to use PD control for circle tracking to enable loiter radius less than L1 length
+// Modified to use PID control for circle tracking to enable loiter radius less than L1 length
 // Modified to enable period and damping of guidance loop to be set explicitly
 // Modified to provide explicit control over capture angle
 
@@ -198,11 +198,47 @@ void AP_L1_Control::_prevent_indecision(float &Nu)
     }
 }
 
+void AP_L1_Control::_update_xtrack_integral(float error, float max_abs_error, float clamp)
+{
+    uint32_t now_us = AP_HAL::micros();
+    float dt = (now_us - _last_update_xtrack_i_us) * 1.0e-6f;
+    if (dt > 1) {
+        // Controller hasn't been called for an extended period of
+        // time. Reset it.
+        _L1_xtrack_i = 0.0f;
+    }
+    if (dt > 0.1) {
+        dt = 0.1;
+    }
+    _last_update_xtrack_i_us = now_us;
+
+    // Reset integral error component if gain parameter has changed. This allows
+    // for much easier tuning by having it re-converge each time it changes.
+    if (!is_positive(_L1_xtrack_i_gain) ||
+        !is_equal(_L1_xtrack_i_gain.get(), _L1_xtrack_i_gain_prev)) {
+        _L1_xtrack_i = 0;
+        _L1_xtrack_i_gain_prev = _L1_xtrack_i_gain;
+
+        return;
+    }
+
+    // Return if error is too large
+    if (fabsf(error) >= max_abs_error) {
+        return;
+    }
+
+    // Compute integral error component to converge to a crosstrack of zero, and
+    // clamp it
+    _L1_xtrack_i += error * _L1_xtrack_i_gain * dt;
+    _L1_xtrack_i = constrain_float(_L1_xtrack_i, -clamp, clamp);
+}
+
 // update L1 control for waypoint navigation
 void AP_L1_Control::update_waypoint(const Location &prev_WP, const Location &next_WP, float dist_min)
 {
     // Update nav. mode
     if (_current_nav_mode != NavMode::WAYPOINT) {
+        _L1_xtrack_i = 0.0f; // Reset integral component on nav. mode change
         _current_nav_mode = NavMode::WAYPOINT;
     }
 
@@ -210,18 +246,6 @@ void AP_L1_Control::update_waypoint(const Location &prev_WP, const Location &nex
     float Nu;
     float xtrackVel;
     float ltrackVel;
-
-    uint32_t now = AP_HAL::micros();
-    float dt = (now - _last_update_waypoint_us) * 1.0e-6f;
-    if (dt > 1) {
-        // controller hasn't been called for an extended period of
-        // time.  Reinitialise it.
-        _L1_xtrack_i = 0.0f;
-    }
-    if (dt > 0.1) {
-        dt = 0.1;
-    }
-    _last_update_waypoint_us = now;
 
     // Calculate L1 gain required for specified damping
     float K_L1 = 4.0f * _L1_damping * _L1_damping;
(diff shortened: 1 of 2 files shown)


IamPete1:
I'm a fan of reworking the loiter radius calculation, the automatic scaling we do by default is confusing and makes mission planning hard. However, we can't just remove it. As your example showed the vehicle will still fly larger radiuses. Plane needs to know what the flown radius will be for navigation. The new autoland is a good example of this, the final waypoint is placed  such that the tangent of the loiter lines up with the runway. If you try autoland mode at high altitude with this branch I suspect you will find a much worse result.

IamPete1:
> As a result, loiters weren't performed at the correct radius except when flying close to sea level. Moreover, since TECS is based on TAS, no explicit altitude compensation is actually needed. Loiters naturally adjust in size depending on what the aircraft can handle at the current altitude (and always match the requested radius when possible).

I don't understand the mechanism here, what has TECS got to do with it? TECS is speed/height, should not affect roll/radius.

If you break out points 4 (formatting) and 5 (typo) into a new PR we could probably merge them relatively quickly and simplify this PR a little.

IamPete1:
> Max roll is altitude independent, it only depends on speed. TECS automatically adjusts the speed for the altitude (it demands a higher TAS in proportion with the altitude), which keeps the flight envelope the same. So the roll limits stay the same, but since we're flying faster, the turn radius is going to be larger (for loiters and in general, as it should).

Makes sense thanks.

> This is still a WIP, but does it look good in regards to what you said about navigation, @IamPete1? And in general

I don't think you need to try and fix the using the current altitude for lookup rather than the target altitude thing in this PR. The messing about with baro is upsetting CI.

I wonder if just using the roll/speed limit radius is the correct thing to do. It means that as soon as you have none zero wind you can't fly the circle anymore. Maybe we could give ourselves some margin so that the radius is a little larger but we will be able to track it better. I not sure. 

There is quite a bit going on in this PR still, if you can break it up anymore that would make it easier to review and faster to get in. Could the addition of the I term be broken out into a new PR?

ArduPlane/Plane.h
```diff
@@ -242,7 +243,7 @@ class Plane : public AP_Vehicle {
 #endif
 
     AP_TECS TECS_controller{ahrs, aparm, landing, MASK_LOG_TECS};
-    AP_L1_Control L1_controller{ahrs, &TECS_controller};
+    AP_L1_Control L1_controller{ahrs, TECS_controller, *AP_Baro::get_singleton(), aparm};
```
IamPete1, line 246: Use the baro singleton directly in L1 (if you need to).

ArduPlane/mode_autoland.cpp
```diff
@@ -129,9 +129,10 @@ bool ModeAutoLand::_enter()
     // Try and minimize loiter radius by using the smaller of the waypoint loiter radius or 1/3 of the final WP distance
     const float loiter_radius = MIN(final_wp_dist * 0.333, fabsf(plane.aparm.loiter_radius));
 
-    // corrected_loiter_radius is the radius the vehicle will actually fly, this gets larger as altitude increases.
-    // Strictly this gets the loiter radius at the current altitude, really we want the loiter radius at final_wp_alt.
-    const float corrected_loiter_radius = plane.nav_controller->loiter_radius(loiter_radius);
+    // corrected_loiter_radius is the radius the vehicle will actually fly, this gets larger as altitude increases
+    const float corrected_loiter_radius =
+        plane.nav_controller->calc_corrected_loiter_radius(loiter_radius, NAN,
```
IamPete1, line 134: Not a fan of passing nans.

libraries/AP_L1_Control/AP_L1_Control.cpp
```diff
@@ -145,36 +136,41 @@ float AP_L1_Control::turn_distance(float wp_radius, float turn_angle) const
     return distance_90 * turn_angle / 90.0f;
 }
 
-float AP_L1_Control::loiter_radius(const float radius) const
+float AP_L1_Control::_calc_min_turn_radius(float indicated_airspeed,
```
IamPete1, line 139: Rather than the NaN, have two finction with a difrenct fingerprint. One takes two floats is the main funciton. Then have one with the same name that takes one float then fetched the alt and passes it into the other function.

libraries/AP_L1_Control/AP_L1_Control.cpp
```diff
@@ -202,27 +198,55 @@ void AP_L1_Control::_prevent_indecision(float &Nu)
...
+        // time. Reset it.
+        _L1_xtrack_i = 0.0f;
+    }
+    if (dt > 0.1) {
+        dt = 0.1;
+    }
+    _last_update_xtrack_i_us = now_us;
+
+    // Reset integral error component if gain parameter has changed. This allows
+    // for much easier tuning by having it re-converge each time it changes.
```
IamPete1, line 216: That introduces steps in the output, better to keep it smooth.

IamPete1: Although I see it is existing code. Seems a little unnecessary....

ArduPlane/mode_autoland.cpp
IamPete1: But your not actually using all those combinations. I think its fine to even not pass altitude to start with, that is a existing issue that could be fixed in a separate PR.

rubenp02:
I'm going to split this into a couple of more specific PRs.

PR #16138, opened 2020-12-26
remove OwnPtr

IamPete1:
This removes the use of OwnPtr, this simplifies the code and paves the way for scripting i2c. This also removes the need for `std::move()`. 

https://github.com/ArduPilot/ardupilot/pull/14276#issuecomment-706670693
 
This has not been tested extensively and the self-deleting functionality of OwnPtr has not been replicated. Possibly some deletes will have to be added.

PR #29166, opened 2025-01-29
Add AP_HAL::I2CDeviceManager->get_device_ptr

peterbarker:
This PR is an alternative path forward for https://github.com/ArduPilot/ardupilot/pull/16138 ; that one's been stalled for many years.

This PR adds an alternate interface for working with I2CDevice objects as allocated by AP_HAL's device manager.

The new method, `get_device_ptr` returns a pointer to an instance of I2CDevice.  This is in contrast to the existing `get_device` method which returns that pointer wrapped in an OwnPtr object.

This new interface allows the code to handle the lifetime in the same way we manage most of our objects in ArduPilot, by `delete`ing what you don't need.

This PR coverts only a single class, the ToshibaLED driver.  This patch saves about 24 bytes.  There are ~150 instances of get_device (both SPI and i2c) in the codebase, so if each use saves this amount of flash that will be about 3kB of flash saved.  Note that the original PR lacked a `delete` call on the I2CDevice member in the ToshibaLED destructor, meaning it would have leaked the memory allocated in `_dev`.  This shows if we are to adopt this pattern we need to be *careful* about object lifetimes!

This entire PR saves ~168 bytes on Pixhawk1-1M`.  I'm assuming the hiding of the virtual function behind a non-virtual function explains that.  

image

Durandal's 1kB saving seems to come from better optimisation made when dealing with the ownptrs.

I'm suggesting we merge this and gradually change users of `get_device` over to use `get_device_ptr`.

Currently the only testing this has had is compilation, CI and flashing onto a Pixhawk-1M to make sure the ToshibaLED still works, and flashing to a Pixhawk6x as representative of a board which doesn't have a ToshibaLED.

```
CubeOrange-periph-heavy  -632              *                                                   
CubeRedPrimary                      -104   *           -216    -216              -216   -216   -296
Durandal                            -744   *           -968    -976              -968   -968   -1088
Hitec-Airspeed           -16               *                                                   
KakuteH7-bdshot                     -144   *           -200    -200              -200   -200   -264
MatekF405                           -72    *           -96     -96               -120   -112   -112
Pixhawk1-1M-bdshot                  -80                -176    -176              -192   -192   -176
f103-QiotekPeriph        16                *                                                   
f303-Universal           0                 *                                                   
iomcu                                                                0                         
revo-mini                           -32    *           -64     -64               -80    -80    -80
skyviper-journey                                       -136                                    
skyviper-v2450                                         -32                                     
```

### DIFF
--- a/libraries/AP_HAL/I2CDevice.h
+++ b/libraries/AP_HAL/I2CDevice.h
@@ -70,11 +70,21 @@ public:
 
 class I2CDeviceManager {
 public:
-    /* Get a device handle */
-    virtual OwnPtr<AP_HAL::I2CDevice> get_device(uint8_t bus, uint8_t address,
-                                                 uint32_t bus_clock=400000,
-                                                 bool use_smbus = false,
-                                                 uint32_t timeout_ms=4) = 0;
+    /* Get a pointer to a newly-allocated I2CDevice handle.  Lifetime
+     * is externally handled */
+    virtual AP_HAL::I2CDevice *get_device_ptr(uint8_t bus, uint8_t address,
+                                              uint32_t bus_clock=400000,
+                                              bool use_smbus = false,
+                                              uint32_t timeout_ms=4) = 0;
+    /* Get a device handle.  Lifetime is handled by OwnPtr system.
+     * Progressively being eliminated */
+    OwnPtr<AP_HAL::I2CDevice> get_device(uint8_t bus, uint8_t address,
+                                         uint32_t bus_clock=400000,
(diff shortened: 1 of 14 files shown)


tridge:
@peterbarker I like this approach, incremental removal of OwnPtr

peterbarker:
Merge if testing on hardware goes well

peterbarker:
I've tested this on Durandal, still get i2c compasses

peterbarker:
I've now tested this on a Navigator and it works just fine.

PR #29170, opened 2025-01-29
SIM_XPlane: avoid leaking memory upon json load failure

peterbarker:
```
../../libraries/SITL/SIM_XPlane.cpp:209:16: warning: Potential leak of memory pointed to by 'fname' [unix.Malloc]
        return false;
               ^~~~~
```


peterbarker:
Now also has a commit to fix:
```
../../libraries/SITL/SIM_XPlane.cpp:239:34: warning: object backing the pointer will be destroyed at the end of the full-expression [-Wdangling-gsl]
  239 |             const char *type_s = d.get("type").to_str().c_str();
```

tridge (approved):

PR #29171, opened 2025-01-29
AP_AIS: remove incorrect use of strncat

peterbarker:
the third argument is space remaining in buffer, not size of buffer...

../../libraries/AP_AIS/AP_AIS.cpp:183:71: warning: Potential buffer overflow. Replace with 'sizeof(multi) - strlen(multi) - 1' or use a safer 'strlcat' API [unix.cstring.BadSizeArg]
                    strncat(multi,_AIVDM_buffer[msg_parts[i]].payload,sizeof(multi));
                                                                      ^~~~~~~~~~~~~
../../libraries/AP_AIS/AP_AIS.cpp:185:49: warning: Potential buffer overflow. Replace with 'sizeof(multi) - strlen(multi) - 1' or use a safer 'strlcat' API [unix.cstring.BadSizeArg]
                strncat(multi,_incoming.payload,sizeof(multi));

### DIFF
--- a/libraries/AP_AIS/AP_AIS.cpp
+++ b/libraries/AP_AIS/AP_AIS.cpp
@@ -32,6 +32,7 @@
 #include <GCS_MAVLink/GCS_MAVLink.h>
 #include <GCS_MAVLink/GCS.h>
 #include <AP_AHRS/AP_AHRS.h>
+#include <AP_Common/ExpandingString.h>
 
 const AP_Param::GroupInfo AP_AIS::var_info[] = {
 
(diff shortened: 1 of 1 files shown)


rmackay9:
Looks like it was a young version of @IamPete1 who wrote the code so I've added him to the reviewers list

IamPete1:
I suspect this bug was why we weren't getting all sorts of errors with the existing code, it was never trying to reassemble. https://github.com/ArduPilot/ardupilot/pull/29195

I would be un-surprised if there are other bugs hiding in AIS, its quite a complex protocol.

PR #29174, opened 2025-01-29
Plane: remove dead store of cruise_speed

peterbarker:
The second assignment, "otherwise cruise" is redundant, so remove it.

```
../../ArduPlane/quadplane.cpp:4260:15: warning: Value stored to 'approach_speed' during its initialization is never read [deadcode.DeadStores]
        float approach_speed = cruise_speed;
              ^~~~~~~~~~~~~~   ~~~~~~~~~~~~
1 warning generated.
```


andyp1per (approved):
There is a similar issue in rover that I noticed on my mac yesterday

IamPete1 (approved):
LGTM

tridge (approved):

PR #29179, opened 2025-01-29
AP_DDS: Added direct actuator control

snktshrma:
This adds a new topic /ap/cmd_mot to allow external controllers control actuators directly. This adds support for external controller and planners (MPC, etc) to handle actuator control directly.

To run:
1. Run Copter SITL
2. On mavproxy, `param set DISARM_DELAY 0`
3. Switch to Guided and arm.
4. Publish: 

```bash
ros2 topic pub /ap/cmd_mot ardupilot_msgs/msg/MotorControl "{timestamp: {sec: 0, nanosec: 0}, actuator: [0.7, 0.7, 0.7, 0.7, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]}"`
```

This addition currently doesn't support case where publisher stops and switches the vehicle to a stable mode. In that case, the actuator output just resets to 0 (Min PWM) if no updates are received.


ArduCopter/mode_guided.cpp
```diff
@@ -584,6 +599,41 @@ void ModeGuided::set_velaccel(const Vector3f& velocity, const Vector3f& accelera
...
 
+// set_actuator_mode - sets guided mode's individual motor mode
+void ModeGuided::set_actuator_mode(float actuator[AP_MOTORS_MAX_NUM_MOTORS]) {
+#if FRAME_CONFIG == MULTICOPTER_FRAME
+    if (guided_mode != SubMode::Actuator) {
+        motor_control_start();
+    }
+
+    guided_motor_state.update_time_ms = millis();
+    memcpy(guided_motor_state.actuator, actuator, AP_MOTORS_MAX_NUM_MOTORS * sizeof(float));
```
Ryanf55, line 610:
```suggestion
    memcpy(guided_motor_state.actuator, actuator, ARRAY_SIZE(guided_motor_state.actuator));
```
If you changed to double in the actuators, then this would not require a change.

ArduCopter/mode_guided.cpp
```diff
@@ -853,6 +903,28 @@ void ModeGuided::velaccel_control_run()
     attitude_control->input_thrust_vector_heading(pos_control->get_thrust_vector(), auto_yaw.get_heading());
 }
 
+// checks for incoming actuator commands
+void ModeGuided::motor_control_run()
+{
+#if FRAME_CONFIG == MULTICOPTER_FRAME
+    // set actuator values to zero if no updates received for 3 seconds
+    uint32_t tnow = millis();
+    if (tnow - guided_motor_state.update_time_ms > get_timeout_ms()) {
```
Ryanf55, line 912:
```suggestion
    if (millis()- guided_motor_state.update_time_ms > get_timeout_ms()) {
```

No need for the local if you only use it once.

Tools/ros2/ardupilot_msgs/msg/MotorControl.msg
```diff
@@ -0,0 +1,2 @@
+builtin_interfaces/Time timestamp # ms
```
Ryanf55, line 1: Can you add documentation what this message is for. Something like

"Control the direct motor outputs of the vehicle"

libraries/AP_DDS/AP_DDS_Topic_Table.h
```diff
@@ -382,4 +385,22 @@ constexpr struct AP_DDS_Client::Topic_table AP_DDS_Client::topics[] = {
...
 #endif // AP_DDS_GLOBAL_POS_CTRL_ENABLED
+#if AP_DDS_MOT_CTRL_ENABLED
+    {
+        .topic_id = to_underlying(TopicIndex::MOTOR_CONTROL_SUB),
+        .pub_id = to_underlying(TopicIndex::MOTOR_CONTROL_SUB),
+        .sub_id = to_underlying(TopicIndex::MOTOR_CONTROL_SUB),
+        .dw_id = uxrObjectId{.id=to_underlying(TopicIndex::MOTOR_CONTROL_SUB), .type=UXR_DATAWRITER_ID},
+        .dr_id = uxrObjectId{.id=to_underlying(TopicIndex::MOTOR_CONTROL_SUB), .type=UXR_DATAREADER_ID},
+        .topic_rw = Topic_rw::DataReader,
+        .topic_name = "rt/ap/cmd_mot",
```
Ryanf55, line 396: Should we make this an experimental topic till we've flown it on a vehicle?

libraries/AP_DDS/AP_DDS_config.h
```diff
@@ -129,6 +129,10 @@
 #define AP_DDS_GLOBAL_POS_CTRL_ENABLED 1
 #endif
 
+#ifndef AP_DDS_MOT_CTRL_ENABLED
+#define AP_DDS_MOT_CTRL_ENABLED 1
```
Ryanf55, line 133: And make this grouped with experimental topics?

Ryanf55 (commented):
Overall looking good. We'll need a review from a copter maintainer such as Randy or Pete Hall.

ArduCopter/mode_guided.cpp
snktshrma: Hi @Ryanf55 ! So ARRAY_SIZE gives number of elements and because of that it will not copy anything after 8th element(32bytes). Either way you suggest using double(float64) instead of float32?

ArduCopter/mode_guided.cpp
snktshrma: Done

libraries/AP_DDS/AP_DDS_Topic_Table.h
snktshrma: Got it

Tools/ros2/ardupilot_msgs/msg/MotorControl.msg
snktshrma: Done!
Also I have modified the actuator array length to 32 in sync with AP_MOTORS_MAX_NUM_MOTORS.

Ryanf55: Nice, maybe later we can figure out how to use the define in the IDL and .msg file without duplicating it. Good enough for now.

ArduCopter/mode_guided.cpp
Ryanf55: I think double precision would be coverkill.

snktshrma: Alright Ryan! Then I'll keep it float32 then

Tools/ros2/ardupilot_msgs/msg/MotorControl.msg
snktshrma: Alright!

PR #29185, opened 2025-01-30
hwdef: support TBS L431 airspeed peripheral

tridge:
This supports a new DLVR airspeed sensor from TBS. Additional features needed were:
 - support for LSE for system clock instead of HSE. This is enabled with "OSCILLATOR_HZ 32768"
 - support for an LED showing CAN terminator state
 - support for both a parameter for CAN termination and an override solder bridge


tridge:
> Will CAN be reliable with LSE?

it should be as we have enabled the auto-trim of MSI using LSE, but it would be good to do some testing across a wide temperature range

libraries/AP_HAL_ChibiOS/hwdef/common/stm32l4_mcuconf.h
```diff
@@ -32,6 +32,18 @@
 #define STM32_PLLM_VALUE                    2
 #define STM32_PLLSRC                        STM32_PLLSRC_HSI16
 
+#elif STM32_HSECLK == 32768U
```
andyp1per, line 35: Slightly holding my nose at the "HSE low frequency means no HSE", but I guess its ok

andyp1per (approved):
LGTM

libraries/AP_HAL_ChibiOS/hwdef/common/stm32l4_mcuconf.h
tridge: yeah, slightly odd, but an awful lot simpler than changing it everywhere

libraries/AP_HAL_ChibiOS/hwdef/TBS-L431-Airspeed/hwdef.dat
```diff
@@ -0,0 +1,24 @@
+include ../TBS-L431/hwdef.inc
+
+USE_BOOTLOADER_FROM_BOARD TBS-L431-Periph
```
peterbarker, line 3: This *should* mean you don't need the `-bl.dat`.  *shrug*

PR #29187, opened 2025-01-30
AR_Motors: fix brushed motor support for omni vehicles

mattp256:
Fixes an issue where omni motors (BrushedWithRelay/BrushedBiplor) are not configured correctly.

I ran into the same problem another user mentioned [here](https://discuss.ardupilot.org/t/mot-pwm-type-3-brushedwithrelay-doesnt-work-with-frame-type-2-omnix/67182):  OmniX frame using MOT_PWM_TYPE = 3 (BrushedWithRelay) only outputs RC servo pulses instead of a duty cycle PWM.

Proposed fix worked great when tested with [Cytron MDD10A](https://www.cytron.io/p-10amp-5v-30v-dc-motor-driver-2-channels) motor driver.

### DIFF
--- a/libraries/AR_Motors/AP_MotorsUGV.cpp
+++ b/libraries/AR_Motors/AP_MotorsUGV.cpp
@@ -132,6 +132,11 @@ void AP_MotorsUGV::init(uint8_t frtype)
 {
     _frame_type = frame_type(frtype);
 
+    // setup for omni vehicles
+    if (_frame_type != FRAME_TYPE_UNDEFINED) {
+        setup_omni();
+    }
+    
     // setup servo output
     setup_servo_output();
 
@@ -141,10 +146,6 @@ void AP_MotorsUGV::init(uint8_t frtype)
     // set safety output
     setup_safety_output();
 
-    // setup for omni vehicles
-    if (_frame_type != FRAME_TYPE_UNDEFINED) {
-        setup_omni();
-    }
 }
 
 bool AP_MotorsUGV::get_legacy_relay_index(int8_t &index1, int8_t &index2, int8_t &index3, int8_t &index4) const


IamPete1:
This looks like a init ordering issue to me. We setup the frame type here:

https://github.com/ArduPilot/ardupilot/blob/a829cd3e60b0c8fec1313082025455ecced6b4a2/libraries/AR_Motors/AP_MotorsUGV.cpp#L146

But we do the PWM setup a few lines above here: 

https://github.com/ArduPilot/ardupilot/blob/a829cd3e60b0c8fec1313082025455ecced6b4a2/libraries/AR_Motors/AP_MotorsUGV.cpp#L139

So `_motors_num` is not set when its used. Can you test if moving `setup_omni` to the top of the function fixes it too.

mattp256:
Yes exactly, _motors_num is not yet set at the time it is used in setup_pwm_type. Moving setup_omni ahead of setup_pwm_type is also confirmed to fix the problem. I was worried there may be some other unintended side effect if I changed the init order so thought forcing the for loop to iterate may be the least disruptive fix - I suppose that carries its own risk of side effects though. Let me know if you prefer changing the init order instead and I will update the PR.

IamPete1:
Sorry for being slow.

> Let me know if you prefer changing the init order instead and I will update the PR.

Yes please, I'm more concerned about setting the PWM type for motors that were not using. As you say we will need to have a look through and see if that change will cause any other issue.

mattp256:
Sure thing. Changed init order and re-tested with latest changes on master. I'm testing by driving the rover around in manual mode. Without this change, the wheels don't spin because the control signal is completely wrong. QGC status messages also show motors 1-4 configured for 'PWM'. With the change, wheels spin normally and I have omni-directional control of the rover. QGC shows the motors configured as 'Brush'.

IamPete1 (approved):
LGTM, Thanks!

tridge:
@rmackay9 you ok with this?

PR #29192, opened 2025-01-30
Scripting: Update Rockblock and MAVLinkHL scripts

stephendade:
Updating the RockBlock and MAVLinkHL scripts to take advantage of the new gcs:run_command_int command.

Also added the gcs:run_command_long lua binding

Trying to debug the merge commits this time....

### DIFF
--- a/libraries/AP_Scripting/applets/RockBlock.lua
+++ b/libraries/AP_Scripting/applets/RockBlock.lua
@@ -10,7 +10,7 @@ Usage:
 Use the MAVLink High Latency Control ("link hl on|off" in MAVProxy) to control
 whether to send or not (or use "force_hl_enable")
 Use the RCK_DEBUG param to view debugging statustexts at the GCS
-Use the RCK_FORCEHL param to force-enable high latency mode, instead of enabling from GCS
+Use the RCK_FORCEHL param to control the mode of operation: 0=Disabled, 1=Enabled, 2=Enabled on 5000ms telemetry loss
 
 Caveats:
 This will *only* send HIGH_LATENCY2 packets via the SBD modem. No heartbeats,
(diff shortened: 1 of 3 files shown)


stephendade:
Having trouble with git here ... I can't find any merge commits in my tree, so not sure why Github thinks there is a merge commit in here.

Sigh .... will leave this for tomorrow.

libraries/GCS_MAVLink/GCS.cpp
```diff
@@ -425,6 +425,20 @@ MAV_RESULT GCS::lua_command_int_packet(const mavlink_command_int_t &packet)
...
+{
+    // for now we assume channel 0. In the future we may create a dedicated channel
+    auto *ch = chan(0);
+    if (ch == nullptr) {
+        return MAV_RESULT_UNSUPPORTED;
+    }
+    // we need a dummy message for some calls
+    mavlink_message_t msg {};
+
+    return ch->try_command_long_as_command_int(packet, msg);
```
IamPete1, line 439: Since this just calls into the command int packet what is the benefit of adding the new method vs doing the conversion in lua?

stephendade: Yep, that works well. I'll modify to do the ``command_long`` to ``command_int`` conversion in the lua script.

peterbarker: > Since this just calls into the command int packet what is the benefit of adding the new method vs doing the conversion in lua?

Doing the conversion in LUA would need to do what we do in the C code, and that's not trivial; we have to come up with a best-guess frame for several incoming location-bearing commands, for example.

stephendade:
A few changes here:
- The ``COMMAND_LONG`` to ``COMMAND_INT`` conversion now takes place in lua, using the conversion code in ``convert_COMMAND_LONG_to_COMMAND_INT()`` in ``libraries/GCS_MAVLink/GCS_Common.cpp``
- I've added a GCS timeout option, where if it's been >5 seconds since a packet (heartbeat?) from a GCS, ``MAV_CMD_CONTROL_HIGH_LATENCY`` will automatically be enabled. So if the main telemetry link is lost, the High Latency link will automatically start (and vice-versa)

@peterbarker and @IamPete1 - not fussed as to whether we do the command_long -> command_int conversion in C or Lua. The Lua version's not _too_ bad.

IamPete1 (approved):
Passes the scripting checks (now without cheating) so I'm happy.

PR #29195, opened 2025-01-30
AP_AIS: fix multi part message decoding

IamPete1:
This bug meant we never re-assembled multi part messages correctly. Tested with the simulator.

### DIFF
--- a/libraries/AP_AIS/AP_AIS.cpp
+++ b/libraries/AP_AIS/AP_AIS.cpp
@@ -146,21 +146,25 @@ void AP_AIS::update()
             } else if (_incoming.num == _incoming.total) {
                 // last part of a multi part message
                 uint8_t index = 0;
-                uint8_t msg_parts[_incoming.num - 1];
+
+                // We have the last part, need to find preceding fragments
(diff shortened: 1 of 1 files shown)


peterbarker:
Needs a rebase...

IamPete1:
> Needs a rebase...

Done, kinda reassuring that the compiler could see the bug you found once it was possible to hit it.

libraries/AP_AIS/AP_AIS.cpp
```diff
@@ -146,21 +146,25 @@ void AP_AIS::update()
             } else if (_incoming.num == _incoming.total) {
                 // last part of a multi part message
                 uint8_t index = 0;
-                uint8_t msg_parts[_incoming.num - 1];
+
+                // We have the last part, need to find preceding fragments
+                const uint8_t parts = _incoming.num - 1;
```
peterbarker, line 151: There appears to be an integer underflow problem here; `_incoming.num` is coming off the wire, and nothing is checking if it is zero or not.

This is pre-existing.

peterbarker (approved):
Tested locally inserting corruption.

PR #29198, opened 2025-01-30
Tools: size_compare_branches.py: mark navigator64 as a Linux board

peterbarker:
... so we don't try to buld bootloaders for it


rmackay9:

PR #29200, opened 2025-01-31
AP_GPS: Fix undulation incorrect convention

### LINKED ISSUE #29199: AP_GPS Undulation is opposite common convention causing incorrect ellipsoid height
## Bug report

**Issue details**

ArduPilot's idea of undulation is not following common convention. It's been this way since it was [first implemented](https://github.com/ArduPilot/ardupilot/pull/21075#pullrequestreview-1065920977), and we have seen [PR's come through to correct device drivers undulation](https://github.com/ArduPilot/ardupilot/pull/23942), when in fact, ArduPilot was using the wrong definition. 

See here for a report:

[Ardupilot GPS Issue - PUBLIC3.pdf](https://github.com/user-attachments/files/18610916/Ardupilot.GPS.Issue.-.PUBLIC3.pdf)

If you use ellipsoid height for an algorithm, you may run into large offsets on some GPS's.

**Version**
Every version of ardupilot that publishes GPS_RAW_INT and dataflash logs to report UND, including `master`.

**Platform**
[x] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
All

**Hardware type**
A few GPS's have been tested

**Logs**
N/A

Ryanf55:
# Purpose

Fixes #29199

# Tests performed

Ground and flight tested on UBX, I would like to request help from the community on checking other GPS's.

# TODO

* We should fix/clarify the comments in get_undulation and spread across the code. The language is either incorrect or imprecise.
* Get other users of GPS's to test this patch - you only need to power on your vehicle, get GPS lock, and send me a BIN and TLOG.
* Consider an autotest and starting to simulate MSL/ellipsoid differences in SITL
* Add documation to the dev wiki on conventions for writing a GPS driver?

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -1377,7 +1377,7 @@ void AP_GPS::send_mavlink_gps_raw(mavlink_channel_t chan)
     float undulation = 0.0;
     int32_t height_elipsoid_mm = 0;
     if (get_undulation(0, undulation)) {
-        height_elipsoid_mm = loc.alt*10 - undulation*1000;
+        height_elipsoid_mm = loc.alt*10 + undulation*1000;
     }
     horizontal_accuracy(0, hacc);
     vertical_accuracy(0, vacc);
@@ -1418,7 +1418,7 @@ void AP_GPS::send_mavlink_gps2_raw(mavlink_channel_t chan)
     float undulation = 0.0;
     float height_elipsoid_mm = 0;
     if (get_undulation(1, undulation)) {
-        height_elipsoid_mm = loc.alt*10 - undulation*1000;
+        height_elipsoid_mm = loc.alt*10 + undulation*1000;
     }
     horizontal_accuracy(1, hacc);
     vertical_accuracy(1, vacc);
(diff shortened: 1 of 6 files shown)


tpwrules (changes requested):
Not tested yet. Needs another sign flip here to avoid breaking AP_Periph: https://github.com/ArduPilot/ardupilot/blob/6bf29eca700120153d7404af1f397c2979715427/Tools/AP_Periph/gps.cpp#L101-L103

I wonder if we should rename dataflash `GPA.Und` to make it clear that the sign has changed for any tools which depend on it. Perhaps `GPA.UndC` for conventional, and possibly keep the old one? I would also support removing it and/or adding an ellipsoid height. Not sure dataflash compatibility is super important.

tpwrules:
Tridge says:
* need to change dataflash name
* could log both, would prefer not to since the existing one is unexpected

NOVA driver might be wrong too, need to test

tpwrules:
From page 245 of (an archived copy of) the NOVA protocol document `om-20000129.pdf` mentioned in the header:
image

The illustration matches Wikipedia and NMEA and is opposite the convention currently in ArduPilot, so the undulation received from the NOVA is almost certainly wrong. If so, this would make a pretty clear case that ArduPilot's convention is backward because it would make all drivers negate the quantity.

tpwrules:
[Geoid information](https://www.unavco.org/software/geodetic-utilities/geoid-height-calculator/geoid-height-calculator.html) using EGM96 from my (approximate) back yard: 

Screenshot from 2025-04-23 20-45-19

Using a NovAtel OEM719 GPS unit and the NOVA driver on Rover `1a0e997a`:
Screenshot from 2025-04-23 20-45-35
Screenshot from 2025-04-23 20-46-53

This proves two things:
* ArduPilot gets the NOVA ellipsoid height wrong because it stores the undulation without negating it.
* Because all drivers indeed need to negate the received undulation (and then negate it again before using it to get the ellipsoid height), ArduPilot's internal sign is wrong.

Ryanf55:
Other PR's have fixed this.

PR #29201, opened 2025-01-31
AP_Filesystem: ROMFS: fix open race conditions

tpwrules:
Lua opens scripts to load them into memory, then the logger opens them after to stream them into the dataflash log. When loading multiple large Lua scripts from ROMFS, decompression takes a significant amount of time. This creates the opportunity for the Lua interpreter and logging threads to both be inside `AP_Filesystem_ROMFS::open()` decompressing a file.

If this happens, the function can return the same `fd` for two different calls as the `fd` is chosen before decompression starts, but only marked as being used after that finishes. The read pointers then stomp on each other, so Lua loads garbled scripts (usually resulting in a syntax error) and the logger dumps garbled data.

Fix the issue by locking before searching for a free record (or marking a record as free). Apply the same fix to directories as well. This doesn't protect against using the same `fd`/`dirp` from multiple threads, but that behavior is to be discouraged anyway and is not the root cause here.

Patch to view the evidence:
```patch
diff --git a/libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp b/libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp
index f481d87255..2fbde94d93 100644
--- a/libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp
+++ b/libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp
@@ -26,6 +26,8 @@
 #include <AP_Math/AP_Math.h>
 #include <AP_ROMFS/AP_ROMFS.h>
 
+#include <stdio.h>
+
 int AP_Filesystem_ROMFS::open(const char *fname, int flags, bool allow_absolute_paths)
 {
     if ((flags & O_ACCMODE) != O_RDONLY) {
@@ -52,6 +54,7 @@ int AP_Filesystem_ROMFS::open(const char *fname, int flags, bool allow_absolute_
         return -1;
     }
     file[idx].ofs = 0;
+    printf("open %d\n", idx);
     return idx;
 }
 
@@ -63,6 +66,7 @@ int AP_Filesystem_ROMFS::close(int fd)
     }
     AP_ROMFS::free(file[fd].data);
     file[fd].data = nullptr;
+    printf("close %d\n", fd);
     return 0;
 }
 ```

The evidence (`fd` 0 is opened twice):
```
AP: Scripting: stopped
AP: Scripting: restarted
Lua: State memory usage: 2796 + 5897
open 0
close 0
open 0
close 0
open 0
open 0
close 0
Lua: Error: @ROMFS/scripts/script2.lua:482: 'end' expected (to close 'function' at line 467) near <eof>
AP: Lua: Error: @ROMFS/scripts/script2.lua:482: 'end' expected (to close 'function' at line 467) near <eof>
```

### DIFF
--- a/libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp
+++ b/libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp
@@ -32,6 +32,8 @@ int AP_Filesystem_ROMFS::open(const char *fname, int flags, bool allow_absolute_
         errno = EROFS;
         return -1;
     }
+
+    WITH_SEMAPHORE(record_sem); // search for free file record
     uint8_t idx;
     for (idx=0; idx<max_open_file; idx++) {
         if (file[idx].data == nullptr) {
@@ -40,10 +42,6 @@ int AP_Filesystem_ROMFS::open(const char *fname, int flags, bool allow_absolute_
     }
     if (idx == max_open_file) {
         errno = ENFILE;
-        return -1;
-    }
-    if (file[idx].data != nullptr) {
-        errno = EBUSY;
         return -1;
     }
     file[idx].data = AP_ROMFS::find_decompress(fname, file[idx].size);
@@ -61,6 +59,8 @@ int AP_Filesystem_ROMFS::close(int fd)
         errno = EBADF;
         return -1;
     }
+
+    WITH_SEMAPHORE(record_sem); // release file record
     AP_ROMFS::free(file[fd].data);
     file[fd].data = nullptr;
     return 0;
@@ -142,6 +142,7 @@ int AP_Filesystem_ROMFS::mkdir(const char *pathname)
 
 void *AP_Filesystem_ROMFS::opendir(const char *pathname)
 {
+    WITH_SEMAPHORE(record_sem); // search for free directory record
     uint8_t idx;
     for (idx=0; idx<max_open_dir; idx++) {
         if (dir[idx].path == nullptr) {
@@ -216,6 +217,8 @@ int AP_Filesystem_ROMFS::closedir(void *dirp)
         errno = EBADF;
         return -1;
     }
+
+    WITH_SEMAPHORE(record_sem); // release directory record
     free(dir[idx].path);
     dir[idx].path = nullptr;
     return 0;
--- a/libraries/AP_Filesystem/AP_Filesystem_ROMFS.h
+++ b/libraries/AP_Filesystem/AP_Filesystem_ROMFS.h
@@ -19,6 +19,8 @@
 
 #if AP_FILESYSTEM_ROMFS_ENABLED
 
+#include <AP_HAL/Semaphores.h>
+
 #include "AP_Filesystem_backend.h"
 
 class AP_Filesystem_ROMFS : public AP_Filesystem_Backend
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by tridge — CHANGES_REQUESTED** [review]
very good find!!
I'd prefer just to put the mutex at the start of open(). We only call open in places where timing is not critical, so I think simplicity is more important than efficiency, and it would be a more obvious patch for 4.6.x

File: libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp [hunk 1]
```diff
@@ -40,17 +51,13 @@ int AP_Filesystem_ROMFS::open(const char *fname, int flags, bool allow_absolute_
     }
     if (idx == max_open_file) {
         errno = ENFILE;
+        AP_ROMFS::free(data);
         return -1;
     }
-    if (file[idx].data != nullptr) {
-        errno = EBUSY;
-        return -1;
-    }
```
robertlong13, line 48: Isn't this check still just as necessary now as before?

robertlong13 (commented):
Tested on CubeOrange and it is working. 80 reloads with no issue. Rolling back I can reproduce every 5 attempts or so.

libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp
tpwrules: No, we know `file[idx].data == nullptr` from the loop above. We now have the lock to protect it from somebody else setting it `!= nullptr`.

This actually might have saved us (i.e. created a more noisy error) if it had been placed after the decompression but  the lock now ensures it can't occur.

robertlong13: Ah, that makes sense, gotcha.

libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp
```diff
@@ -217,6 +218,8 @@ int AP_Filesystem_ROMFS::closedir(void *dirp)
         return -1;
     }
     free(dir[idx].path);
+
+    WITH_SEMAPHORE(record_sem); // release directory record
```
tridge, line 222: I think the WITH_SEMAPHORE() should be above the free()

libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp
```diff
@@ -62,6 +60,8 @@ int AP_Filesystem_ROMFS::close(int fd)
         return -1;
     }
     AP_ROMFS::free(file[fd].data);
+
+    WITH_SEMAPHORE(record_sem); // release file record
```
tridge, line 64: should be above the free() I think

tridge (changes requested):

libraries/AP_Filesystem/AP_Filesystem_ROMFS.cpp
tpwrules: Why? Nobody will re-use the record until the data pointer is nullptr.

This patch doesn’t try to protect using the same fd from multiple threads.

tridge: it needs to either go above the free or can be removed completely. Right now it is protecting a store of a 32 bit value, which is atomic anyway (we assume 32 bit atomic in ArduPilot). So if this is needed at all then it is needed above the free

tpwrules: Fair enough. I really hate that assumption, it is not at all true at the compiler level. I will move it.

tridge (approved):

PR #29186, opened 2025-01-30
Create and use AP_PERIPH_BARO_ENABLED

shiv-tyagi:
Follows #29183

Part of the HAL_PERIPH_ENBALE_XX renaming series.

PR #29203, opened 2025-01-31
Create and use AP_PERIPH_RANGEFINDER_ENABLED

shiv-tyagi:
Follows #29186


libraries/AP_HAL_ChibiOS/hwdef/sw-spar-f407/hwdef.dat
```diff
@@ -91,7 +91,7 @@ define GPS_MAX_RECEIVERS 1
 define GPS_MAX_INSTANCES 1
 
 # rangefinder
-define HAL_PERIPH_ENABLE_RANGEFINDER 2
+define AP_PERIPH_RANGEFINDER_ENABLED 1 2
```
peterbarker, line 94: This doesn't look right.  And does "2" do something?!

peterbarker (changes requested):

libraries/AP_HAL_ChibiOS/hwdef/sw-spar-f407/hwdef.dat
shiv-tyagi: Fixed that. I did check it though, but there are so many of them that this one slipped through my sight again. I'll be more careful next time.

shiv-tyagi: Fixed that. I did check it though but again there are so many of them that this one again escaped my sight. Will be more careful this time.

peterbarker (approved):

peterbarker:
Merged, thanks!

PR #29206, opened 2025-02-01
AP_BattMonitor: remove use of ownptr

peterbarker:
@tpwrules there's no in-tree user of this still :-)

The simulation still finds the device:

image


tpwrules:
Tested on Stampfly, yes I know I need to get that upstream :) Same readings and no crashes.

I assume the leak was an existing problem, it's just now documented? Why did you remove it?

tridge (approved):

PR #29207, opened 2025-02-01
GCS_MAVLink: add capability to send autopilot to config error loop

peterbarker:
another one of our "do nasty thing to autopilot" commands, useful for testing what ground control stations do when the autopilot is in this state

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -3469,6 +3469,11 @@ MAV_RESULT GCS_MAVLINK::handle_preflight_reboot(const mavlink_command_int_t &pac
             send_text(MAV_SEVERITY_WARNING,"deadlock passed");
             return MAV_RESULT_ACCEPTED;
         }
+        if (is_equal(packet.param4, 101.0f)) {
+            // the capital-U and ~ here are actually important for
+            // testing a MissionPlanner bug!
+            AP_BoardConfig::config_error("YOU~RE WELCOME!");
+        }
 #endif  // AP_MAVLINK_FAILURE_CREATION_ENABLED
 
 #if HAL_ENABLE_DFU_BOOT


peterbarker:
Tested in SITL, works as expected.

Tested on a CubeOrangePlus, works as expected.

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3469,6 +3469,9 @@ MAV_RESULT GCS_MAVLINK::handle_preflight_reboot(const mavlink_command_int_t &pac
             send_text(MAV_SEVERITY_WARNING,"deadlock passed");
             return MAV_RESULT_ACCEPTED;
         }
+        if (is_equal(packet.param4, 101.0f)) {
+            AP_BoardConfig::config_error("config_error - YOU~RE WELCOME!");  // the capital-U and ~ here are actually important for testing a MissionPlanner bug!
```
robertlong13, line 3473: The `config_error` method already prepends "Config Error: " to the message, so I guess let's save the 15 bytes
```suggestion
            AP_BoardConfig::config_error("YOU~RE WELCOME!");  // the capital-U and ~ here are actually important for testing a MissionPlanner bug!
```

```
2/2/2025 12:35:36 AM : Config Error: fix problem then reboot
2/2/2025 12:35:36 AM : Config Error: YOU~RE WELCOME!
```

robertlong13 (approved):
I tested that this command correctly borks stuff up in current Planner. The `U~` looks like the start of a rather long MAVLink 0.9 packet and wipes out at least the next two subsequent MAVLink messages.

Of course, after https://github.com/ArduPilot/MissionPlanner/pull/3483, things work fine.

libraries/GCS_MAVLink/GCS_Common.cpp
peterbarker: Done

PR #29210, opened 2025-02-02
Plane: remove unused definitions 

IamPete1:
I found a bunch of things that were not used in `plane.h`. Takeoff had a `tkoff_option_is_set` helper, but it was not used anywhere, this changes to use it.


peterbarker (approved):
LGTM

PR #29215, opened 2025-02-03
Fixed typo in description of RC_PROTOCOL

python36:
Fixed typo in value description of RC_PROTOCOL parameter for enabling all protocols


peterbarker:
Sorry, this is not correct.

These descriptions are for bit offsets in a bitmask.

We just use the bottom-most bit to indicate that all protocols should be scanned.

python36:
@peterbarker 
🤦Sorry and thank you

PR #29216, opened 2025-02-03
Create and use AP_PERIPH_RPM_ENABLED and AP_PERIPH_RPM_STREAM_ENABLED

shiv-tyagi:
Part of HAL_PERIPH_XX_ENABLE renaming series.

Follows #29203


peterbarker (commented):
Any idea why AP_PERIPH_RPM_STREAM_ENABLED requires AP_PERIPH_RPM_ENABLED?

ie. are they two separate features which both happen to require AP_RPM_ENABLED?

shiv-tyagi:
> Any idea why AP_PERIPH_RPM_STREAM_ENABLED requires AP_PERIPH_RPM_ENABLED?

I have no idea I'm afraid. This was pre-existing. Please see [this](https://github.com/ArduPilot/ardupilot/blob/bac78b54bf4cf0b44db425760c9c999c767ce060/Tools/AP_Periph/AP_Periph.h#L95).

peterbarker (changes requested):
@shiv-tyagi I've realized that simply having the RPM library available means that it is available in things like the ESC telemetry library and whatnot.

Pretty sure in this case, since `AP_PERIPH_RPM_ENABLED` simply makes the RPM library available to other libraries the in-between periph define is absolutely pointless.  But I also know that having it is going to be less contentious than removing it.

So I think we should rebase this for the conflict, and add the defines in boards.py to ensure the SITL periphs still compile.

peterbarker:
Merged, thanks!

PR #29219, opened 2025-02-03
Add AP_AirSensor library

Ryanf55:
Looking for initial feedback- see the README.

### DIFF
--- a/libraries/AP_AirSensor/AP_AirSensor.h
+++ b/libraries/AP_AirSensor/AP_AirSensor.h
@@ -1,0 +1,100 @@
+/*
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+#include "AP_AirSensor_config.h"
+
+#if AP_AIRSENSOR_ENABLED
+
+#include <AP_HAL/AP_HAL.h>
+#include <inttypes.h>
+#include <AP_Common/AP_Common.h>
+#include <AP_ExternalAHRS/AP_ExternalAHRS.h>
+
+class AP_AirSensor_Backend;
+
+class AP_AirSensor_Params {
+public:
+    // Constructor
+    AP_AirSensor_Params(void);
+    AP_Int8  type; // TODO forcing scripting for now as default type.
+    // static const struct AP_Param::GroupInfo var_info[];
+};
+
+/// @class AP_AirSensor
+/// Air data sensing class that provides a common interface for 1D AP_Airspeed probes
+/// 2D WindVane sensors, and other 1D-3D capable air data sensor interfaces.
+class AP_AirSensor
+{
+public:
+    friend class AP_AirSensor_Backend;  
+    // static const struct AP_Param::GroupInfo var_info[];
+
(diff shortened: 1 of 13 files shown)


libraries/AP_AirSensor/AP_AirSensor.h
```diff
@@ -0,0 +1,97 @@
...
+    // return the number of air sensor backends
+    uint8_t num_sensors() const { return _num_sensors; }
+
+    /* Do not allow copies */
+    CLASS_NO_COPY(AP_AirSensor);
+
+    enum class Type {
+        NONE,
+#if AP_AIRSENSOR_SCRIPTING_ENABLED
+        SCRIPTING
```
peterbarker, line 50:
```suggestion
        SCRIPTING,
```

libraries/AP_AirSensor/AP_AirSensor_Backend.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+AP_AirSensor_Backend::AP_AirSensor_Backend(AP_AirSensor& frontend, AP_AirSensor::State& state, AP_AirSensor_Params& params) :
+    _frontend(frontend),
+    _state(state),
+    _params(params)
+{
+    _backend_type = (AP_AirSensor::Type )_params.type.get();
+}
+
+bool AP_AirSensor_Backend::get_aoa(float& a) const {
```
peterbarker, line 28: Let's be consistent about where that opening brace is!

ArduPilot convention is on a fresh line, but we differ across libraries

libraries/AP_AirSensor/AP_AirSensor_Backend.h
```diff
@@ -0,0 +1,55 @@
...
+	AP_AirSensor_Backend(AP_AirSensor &frontend, AP_AirSensor::State &state, AP_AirSensor_Params &params);
+    virtual bool init() = 0;
+    virtual ~AP_AirSensor_Backend() {}
+    virtual void update() = 0;
+    // get measured wind vector [m/s]
+    virtual bool get_wind(Vector3f& wind_uvw) const { return false; }
+    // get measured angle of attack [rad]
+    virtual bool get_aoa(float& a) const;
+
+#if AP_SCRIPTING_ENABLED
```
peterbarker, line 34: certical whitespace here somewhere, please!

libraries/AP_AirSensor/AP_AirSensor_config.h
```diff
@@ -0,0 +1,26 @@
...
+#ifndef AP_AIRSENSOR_AIRSPEED_ENABLED
+#define AP_AIRSENSOR_AIRSPEED_ENABLED (AP_AIRSENSOR_ENABLED && AP_AIRSPEED_ENABLED)
+#endif
+
+#ifndef AP_AIRSENSOR_WINDVANE_ENABLED
+#define AP_AIRSENSOR_WINDVANE_ENABLED (AP_AIRSENSOR_ENABLED && AP_WINDVANE_ENABLED)
+#endif
+
+#ifndef AP_AIRSENSOR_SCRIPTING_ENABLED
+#define AP_AIRSENSOR_SCRIPTING_ENABLED (AP_AIRSENSOR_ENABLED && AP_SCRIPTING_ENABLED)
```
peterbarker, line 21: create an AP_AIRSENSOR_BACKEND_DEFAULT_ENABLED

Ryanf55:
* We could use this in EKF3 only if the P matrix is 3D - the state vector is 2F, but the EKF returns a 3D vector
* In the EKF outputs, it returns a 3D, it hard codes AP_navekf3, we would go from 24 to 25 state EKF- we should ask Paul
* We could augment the input side of the EKF with body frame wind
* We seem like we would want move forward with this, we should disable it by default, and can at least gather some data on a real vehicle to see how well the 3D airspeed correlates

libraries/AP_AirSensor/AP_AirSensor_Backend.cpp
Ryanf55: I'll run astyle on all this before it's actually put up for review. I'm more concerned with architecture than code style at this moment.

PR #29222, opened 2025-02-04
AP_Airspeed: disable airspeed auto calibration in specific cases

mduclehcm:
This change disables automatic airspeed calibration when:

airspeed_use == 2 (UseWhenZeroThrottle), and
SRV_Channel::k_throttle != 0.
This prevents calibration issues caused by propeller airflow interfering with airspeed sensors, particularly in gliders.

### DIFF
--- a/libraries/AP_Airspeed/Airspeed_Calibration.cpp
+++ b/libraries/AP_Airspeed/Airspeed_Calibration.cpp
@@ -13,6 +13,7 @@
 #include <AP_Math/AP_Math.h>
 #include <GCS_MAVLink/GCS.h>
 #include <AP_AHRS/AP_AHRS.h>
+#include <SRV_Channel/SRV_Channel.h>
 
 #include "AP_Airspeed.h"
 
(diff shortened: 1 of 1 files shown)


IamPete1:
Whats would the calibration flight look like? My concern is that if you were loitering for example the motor would come on at the same section of the loiter every time and thus bias the result. We could also refuse to start the calibration if the motor was on.

mduclehcm:
> Whats would the calibration flight look like? My concern is that if you were loitering for example the motor would come on at the same section of the loiter every time and thus bias the result. We could also refuse to start the calibration if the motor was on.

I think it should like a normal glider flight. First, climb to a high altitude, then turn off the motor and loiter until reaching the minimum altitude.

I'm not entirely sure how the airspeed auto-calibration algorithm works. I'm just trying to make it function correctly when placed behind the propeller.

tridge (approved):

LupusTheCanine:
We could run tally of headwind component to detect large possible wind bias and warn user that calibration may be inaccurate.

tridge:
marked for 4.6 backport

PR #29223, opened 2025-02-04
GCS_MAVLink: simplify MissionItemProtocol get_item interface

peterbarker:
stop passing through _link and the original msg, move use to the base class instead.

starts fence and rally also using the "correct the GCS's count" code.

This also corrects the error code when correcting the GCS's count to INVALID_SEQUENCE rather than just ERROR

```
CubeOrange-periph-heavy  144               *
CubeRedPrimary                      24     *           24      24                24     24     24
Durandal                            64     *           64      64                64     64     64
KakuteH7-bdshot                     24     *           24      24                24     24     24
MatekF405                           24     *           24      24                24     24     24
Pixhawk1-1M-bdshot                  24                 24      24                24     24     24
revo-mini                           24     *           24      24                24     24     24
skyviper-journey                                       24
skyviper-v2450                                         24
```

This should allow for cleanups in the AP_Filesystem_Mission code in the future.

### DIFF
--- a/libraries/GCS_MAVLink/MissionItemProtocol.cpp
+++ b/libraries/GCS_MAVLink/MissionItemProtocol.cpp
@@ -155,9 +155,19 @@ void MissionItemProtocol::handle_mission_request_int(GCS_MAVLINK &_link,
     }
 
     mavlink_mission_item_int_t ret_packet;
-    const MAV_MISSION_RESULT result_code = get_item(_link, msg, packet, ret_packet);
-
+    const MAV_MISSION_RESULT result_code = get_item(packet.seq, ret_packet);
     if (result_code != MAV_MISSION_ACCEPTED) {
+        if (result_code == MAV_MISSION_INVALID_SEQUENCE) {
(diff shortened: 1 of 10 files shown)


libraries/GCS_MAVLink/MissionItemProtocol.cpp
```diff
@@ -154,8 +154,22 @@ void MissionItemProtocol::handle_mission_request_int(GCS_MAVLINK &_link,
...
+                                           item_count(),
+                                           mission_type());
+        }
+        send_mission_ack(_link, msg, MAV_MISSION_INVALID_SEQUENCE);
+        return;
+    }
+
     mavlink_mission_item_int_t ret_packet;
-    const MAV_MISSION_RESULT result_code = get_item(_link, msg, packet, ret_packet);
+    const MAV_MISSION_RESULT result_code = get_item(packet.seq, ret_packet);
```
tridge, line 172: possible can avoid item_exists() and check return code from get_item()

PR #29226, opened 2025-02-04
ArduPlane: remove unused terrain enum

Hwurzburg:
The AUTO enum is not used in the TerrainFollow table. Terrain following in AUTO is exclusively controlled by the next_WP alt frame


IamPete1:
I'm not sure this is true in every case. Rangefinder based terrain correction might use it in auto. You could put a panic in the `terrain_enabled_in_current_mode` function to see it CI ever trys it.

Hwurzburg:
understand how it could be used in AUTO rangefinder landing now....will update wiki...closing

Issue #18496, opened 2021-08-29
Plane: `RETURN_TO_LAUNCH` mission command should not switch mode


IamPete1:
Currently a `RETURN_TO_LAUNCH` mission command will just switch to RTL mode. 

https://github.com/ArduPilot/ardupilot/blob/a39a948c57b53fb4c36ace731c1a40fea9233638/ArduPlane/commands_logic.cpp#L74-L76

This means that you cannot do a RTL part way through a mission. We also have a verify function that is never used because were no longer in auto to run it. https://github.com/ArduPilot/ardupilot/blob/a39a948c57b53fb4c36ace731c1a40fea9233638/ArduPlane/commands_logic.cpp#L249-L250

Instead plane should behave as copter and the RTL should be a submode of auto.

IamPete1:
This is the lass user of `ModeReason::MISSION_CMD`

PR #29227, opened 2025-02-04
ArduPlane: remove unused verify_RTL code

Hwurzburg:
This code is never used.....do_RTL is, but it changes mode to RTL directly and mission execution is interrupted so is never called by commands_logic....


IamPete1:
I would like to start using it at somepoint. https://github.com/ArduPilot/ardupilot/issues/18496

Hwurzburg:
> I would like to start using it at somepoint. #18496

nice idea, but its been over 3 years since you started....maybe it gets removed until you finish? we can let the group decide, I am not fussed either way

tridge (approved):

PR #29228, opened 2025-02-05
AP_Mount: fix object tracking command for ViewPro

eppravitra:
The command should also take negative values

### DIFF
--- a/libraries/AP_Mount/AP_Mount_Viewpro.cpp
+++ b/libraries/AP_Mount/AP_Mount_Viewpro.cpp
@@ -624,7 +624,7 @@ bool AP_Mount_Viewpro::send_tracking_command(TrackingCommand cmd, uint8_t value)
 }
 
 // send camera command2 and corresponding parameter values
-bool AP_Mount_Viewpro::send_tracking_command2(TrackingCommand2 cmd, uint16_t param1, uint16_t param2)
+bool AP_Mount_Viewpro::send_tracking_command2(TrackingCommand2 cmd, int16_t param1, int16_t param2)
 {
     // fill in packet
     const E2Packet e2_packet {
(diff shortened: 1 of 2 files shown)


eppravitra:
The change was tested with CubeOrangePlus connected to ViewPro A10T

rmackay9:
Hi @eppravitra,

Thanks for your work on fixing the ViewPro driver.  Are you sure this works correctly?  It seems there could be two problems:

1.  send_tracking_command2 uses the htobe16() method which I think takes a uint16_t meaning that param1 and param2 are just going to be cast to be unsigned anyway
2. The existing code that converts p1 and p2 to values between -960~+960 and -540 ~ +540 seems incorrect.  Of course I wrote this code originally and you haven't touched it as part of this PR but I was unable to test this properly.
 
Shouldn't this line:
`send_tracking_command2(TrackingCommand2::SET_POINT, (p1.x - 0.5) * 960, (p1.y - 0.5) * 540));`

.. actually be ..

`send_tracking_command2(TrackingCommand2::SET_POINT, (p1.x - 0.5) * 2.0 * 960, (p1.y - 0.5) * 2.0 * 540));`

rmackay9:
@eppravitra,

Thanks, as long as we think it is correct then I think the 2nd change should be fine.  I suspect nobody has used the tracking function yet so it's likely safe.  It looks like I have added a command to test this to our [MAVLink Interfaces Camera wiki page](https://ardupilot.org/dev/docs/mavlink-camera.html#mav-cmd-camera-track-point-to-start-tracking-a-point-on-the-live-video-stream) but I only tested the very center so this is likely why I didn't notice the problem.  Once we get this fixed, I'll add some more tests to that page.

Ideally when doing testing we should confirm that it's incorrect before and correct aftwards.  So two tests if possible.

eppravitra:
The scaling change has been made @rmackay9 .
I already test it with CubeOrangePlus and ViewPro A10T. It works.

I think this one is ready for a merge.

rmackay9 (approved):
tested and it works well, thanks!

rmackay9:
merged, thanks!

Issue #26932, opened 2024-04-30
ArduPlane Flare Height Cmd During AutoLand


jknight100:
**IF YOU DON'T REMOVE THESE FOUR LINES, THEN YOUR REPORT WILL BE CLOSED AUTOMATICALLY**
Questions and user problems should be directed at the forum (http://discuss.ardupilot.org)
_**Please be very sure you have found a bug when opening this issue**_
If there was a previous discussion in the forum, link to it

## Bug report

**Issue details**
Height demand term is erroneous in flare, when using rangefinder.

_Please describe the problem_
Using a mature simulation with ArduPilot in the loop, and injecting radar altimeter readings from synthesized terrain.
During flight, terrain height was injected at 30ft (~10m).  During autoland, altitude update occurred as expended.  When flare occurs (triggered by 3 seconds to touchdown) TECS target altitude "h_dem" shows 30ft drop followed by 30ft ramp downward. Not sure why the offset is needed, and further, the rampdown seems to be in the wrong direction, causes a pitchdown cmd, and does not agree with altitude rate cmd.
Ultimately, the TECS control law minimizes error between "h_dem" and "height", so these should be in the same altitude frame.  This appears to not be the case.
SEE PICTURE for simulation output.

**Version**
Plane-4.4.4

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[X] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
simulated airplane, powered fixedwing.

**Hardware type**
Simulated CubeOrange

**Logs**
cannot provide due to sensitive information

Suggested fix is to simply remove code in AP_TECS.cpp:
// fade across to the ideal height profile
        _hgt_dem = _flare_hgt_dem_adj * (1.0f - p) + _flare_hgt_dem_ideal * p;
// correct for offset between height above ground and height above datum used by control loops
        _hgt_dem += (_hgt_afe - _height);
When I removed this code, flare altitude and alt cmd were nice and smooth, and agreed with altitude rate terms as expected.
<img width="625" alt="AutoLand_444" src="https://github.com/ArduPilot/ardupilot/assets/114116288/2034e71d-5df8-448e-8ccc-e90b1cb6b1a1">

PR #29230, opened 2025-02-05
AP_TECS: Fix flare height demand when using a rangefinder

rubenp02:
# AP_TECS: Fix flare height demand when using a rangefinder

Addressed an issue in the flaring behavior when using a rangefinder. Previously, the aircraft would pitch down instead of slowing its descent during the flare stage, diverging from the height rate demand considerably and resulting in a hard landing.

The root cause of this issue was an incorrect calculation of the height demand from the sink rate. The rangefinder correction was being accounted for when it shouldn't, causing a mismatch between the intended sink rate and the height demand. Since the latter drives the pitch demand, this led to an unintended pitch down command right after starting the flare. The rangefinder has already been used to adjust the glide slope and to determine when to start the flare, so the aircraft should flare normally.

Key changes:
- Removed invalid rangefinder correction in the height demand calculation during the flare
- Ensured consistent flare behavior regardless of whether a rangefinder is used

### Flare w/o rangefinder in master (good)

imagen

### Flare w/ rangefinder in master (bad, see the nonsensical height demand and the complete divergence in height rate)

imagen

### Flare w/ rangefinder in this PR (identical to the master behavior w/o rangefinder)

imagen

### DIFF
--- a/libraries/AP_TECS/AP_TECS.cpp
+++ b/libraries/AP_TECS/AP_TECS.cpp
@@ -606,7 +606,7 @@ void AP_TECS::_update_height_demand(void)
 
         if (!_flare_initialised) {
             _flare_hgt_dem_adj = _hgt_dem;
-            _flare_hgt_dem_ideal = _hgt_afe;
+            _flare_hgt_dem_ideal = _height;
             _hgt_at_start_of_flare = _hgt_afe;
             _hgt_rate_at_flare_entry = _climb_rate;
             _flare_initialised = true;
@@ -629,9 +629,6 @@ void AP_TECS::_update_height_demand(void)
 
         // fade across to the ideal height profile
         _hgt_dem = _flare_hgt_dem_adj * (1.0f - p) + _flare_hgt_dem_ideal * p;
-
-        // correct for offset between height above ground and height above datum used by control loops
-        _hgt_dem += (_hgt_afe - _height);
     }
 }
 


tridge:
@rubenp02 I'll need to look at this further and discuss with @Georacer - many thanks for reporting this!

tridge:
@rubenp02 can you provide links to the logs you got the above graphs from? I'd like to reproduce. I presume the graphs are from SITL?

tridge:
I suspect this is not related to rangefinder, but instead is related to landing with a target height different from home.

tridge (changes requested):
I'd like to see testing done (SITL is fine) with landing at a runway height that is both above and below home altitude. That is really what this adjustment in TECS is all about. For simplificty, I suggest testing with no rangefinder.
I suspect the change in the PR is a correct one, we just need to ensure we don't regress badly if landing on a runway well above or below home alt

jschall:
KHA is using a heavily modified TECS which includes fixes for landing at locations other than home (which was causing flare to never happen or happen too high depending on whether the target was higher or lower), and fixes flare so that it no longer uses a height target (the controller could slam the plane into the ground correcting for a disturbance) and instead targets a sink rate based on rangefinder distance and estimated ground slope. The sink rate also applies as a limit prior to flare, so it will slow it's descent before the flare. The result is incredibly consistent landings that work on sloped runways and with steep reverse thrust approaches. Getting the ground slope estimation right is pretty dependent on rangefinder latency, though. We had to configure the lightware lidar for lower latency. That part could be left out or made optional.

Additionally our TECS has been adjusted to be more of a speed and height rate controller rather than a speed and height controller, and this eliminates the intermediate height demand that can get out of sync with the actual waypoint path. Throttle control has been completely ripped out and replaced with a model-based controller, but that is extremely specific to us as we use a current-controlled ESC which is unusual.

Happy to share if someone wants to handle PRing it.

rubenp02:
> KHA is using a heavily modified TECS which includes fixes for landing at locations other than home (which was causing flare to never happen or happen too high depending on whether the target was higher or lower), and fixes flare so that it no longer uses a height target (the controller could slam the plane into the ground correcting for a disturbance) and instead targets a sink rate based on rangefinder distance and estimated ground slope. The sink rate also applies as a limit prior to flare, so it will slow it's descent before the flare. The result is incredibly consistent landings that work on sloped runways.
> 
> 
> 
> Additionally our TECS has been adjusted to be more of a speed and height rate controller rather than a speed and height controller, and this eliminates the intermediate height demand that can get out of sync with the actual waypoint path. Throttle control has been completely ripped out and replaced with a model-based controller, but that is extremely specific to us as we use a current-controlled ESC which is unusual.
> 
> 
> 
> Happy to share if someone wants to handle PRing it.

Almost all of that sounds exactly like what TECS should be. I imagine this could also be used to implement a safe pitch stick mixing (#29778) quite trivially.

It would be hard to merge such sweeping changes though. Happy to help in any way I can.

rubenp02:
@jschall Where can I contact you?

jschall:
@rubenp02 discord works.

tridge:
@rubenp02 Henry ( @Hwurzburg ) has reported a similar issue and will test this PR on his plane soon

priseborough:
This is a bug that I have also ran into recently and was about to submit this branch as a fix :https://github.com/priseborough/ardupilot/tree/pr-tecs_flare_bug_fix

@rubenp02 how can I contact you - on Discord?

rubenp02:
> @rubenp02 how can I contact you - on Discord?

Yes, Discord works for me. I'm on the AP server as rubenp02.

rubenp02:
> @rubenp02 discord works.

Can I get your handle? Can't seem to find you on the AP server.

priseborough:
@rubenp02 I've pinged you on Discord. Handle is paulriseborough

priseborough:
Something I would like to preserve is the original intent of the flare height control to fade across to using the height above touchdown point _hgt_afe during the flare manoeuvre as this is going to better adjust to runway gradient when using a range finder. The problem with the original code was that it was fading between a height above home and a height above touchdown point demand, but was not fading the height measurement used by the controller.

Hwurzburg:
@rubenp02 should I hold off test flying this PR until you and Paul discuss??

rubenp02:
> @rubenp02 should I hold off test flying this PR until you and Paul discuss??

My PR should work as it's been running on our production units for months now, at least for the case of landing at home. So feel free to test it IMO.

priseborough:
This PR looks safe to fly to me and fixes the problem of possible large pitch demand transients on flare entry when using a range finder.

After reviewing the changes, where it differs from https://github.com/priseborough/ardupilot/tree/pr-tecs_flare_bug_fix  is that it continues to use the height above field elevation as the feedback measurement for height during flare whereas the more complex fix in https://github.com/priseborough/ardupilot/tree/pr-tecs_flare_bug_fix fades across to using the _hgt_afe as the feedback source which in theory will better handle variations in runway height and EKF height estimation error during flare. For most applications this wont matter as the flare is a short duration manoeuvre that doesn't cover much distance.

rubenp02:
> @rubenp02 I've pinged you on Discord. Handle is paulriseborough

Hi! Sorry I didn't get back to you earlier, didn't see your ping (I checked but I don't know Discord very well).

Just wanted to say that I won't be able to get to this until next Wednesday or Thursday, so while I'd love to help, feel free to take over this if it's urgent. Otherwise, I'll be happy to help next week. I'll shoot you a Discord message then.

priseborough:
@tridge This fixes a bug that has the potential to damage airframes, so and I recommend it be considered  as a hot  fix candidate for inclusion in the next stable release minor update. The blending across to using the _hgt_afe measurement during flare is something that I'm happy tackle after this is merged.

rmackay9:
Based on Paul's comments I've tentatively added this to the Copter-4.6 back ports project but of course it's @tridge's decision as to whether it's included in the next point release

rubenp02:
> Based on Paul's comments I've tentatively added this to the Copter-4.6 back ports project but of course it's @tridge's decision as to whether it's included in the next point release

Copter?

Hwurzburg:
test flew this am and while it did not dive, it landed WAY short, consistently  (>30m) vs autolands before I added the RF (<10m)
here is with this code using a rangefinder https://www.dropbox.com/scl/fi/dedpg36llqbc0znm3p1yj/2025-08-08-08-56-23.bin?rlkey=hssgedfjqm5a8affo9b3c37xi&dl=0
and previous autolands without RF
https://www.dropbox.com/scl/fi/uzd3qfqrpcpliwhjcu99d/2025-01-25-14-26-45.bin?rlkey=22wl1l8ifiygg9v8nrf5vmuah&dl=0

matteotarsi:
Hello, I'm Italian. I'm passionate about autolandings. I have trouble with English so I use translators, and no experience with programming, but I have some experience with flying because my father and friends of my father are private pilot.  I have some experience with RC airplanes and helicopters. I use Ardupilot Plane and Traditional Helicopters.

From Hwurzburg's logs, I don't understand the offset between des pitch and pitch, but the pitch controller works.

The LAND_PITCH_DEG parameter was incorrectly set to 25 degrees, when it should be set around zero or -3 degrees, for example. However, it doesn't cause any problems because it appears to be ignored. However, if the firmware is OK, it should be removed or its operation corrected.

It lands at 30 meters, similar to the 20 meters before using the RF, perhaps because the default LAND_SLOPE_RCALC parameter of 2m is significantly exceeded. We're about 3 meters off when it acquires the RF, so it probably performs glide recalculation, or it doesn't work. In the past, in Plane 4.3, it gave the message, but now in 4.4, 4.5, and 4.6, it no longer does. In any case, the RF is ignored, which is why it lands much earlier than target point 4, because it's over 3 meters lower. I discovered this because my old airfield had an embankment a few meters before the runway, so I was lucky enough to figure it out after many automatic landings.

This happens with almost every aircraft because the static barometric pressure inside the fuselage is always lower than the actual one. The static pitot pressure should be used, but that's not possible because the Matek pitots I use are differential, so this correction must be made with the RF.

During the flare phase, the RF works, the TECS_LAND_PMAX of 10 degrees is bypassed by the LAND_PITCH_DEG of 25, all of which must be corrected.

So this test only works for the flare but not for the glide slope.

As soon as I can, I'll try Plane 4.7.0-dev with the correct parameters to see if it respects the glide slope after RF acquisition and if the error has been fixed, as rubenp02, who seems to be an expert, surely states.

Thanks to George Zogopoulos

Other users who have seen the problem are: https://github.com/ArduPilot/ardupilot/issues/26932

I saw the problem starting with Plane 4.4 around July 2023 and I wrote about it here, but in Italian: https://discuss.ardupilot.org/t/plane-4-6-release/134619/19?u=matteo_tarsi

My log with Plane 4.6.2 and the pitch command error is this: https://www.dropbox.com/scl/fi/hjud02ur3a0klkj8ei4yg/2025-07-12-15-45-59-secondo-arrivo-aridaie-sbatte-gi-il-musso-armare-alto-se-no-si-rischia-aumentare-flare-sci-anche-sul-wing-arriva-a-40-ampere.bin?rlkey=ejcyi9au3tnd6z9yc6cdfnq8v&st=jan5tsdo&dl=0

excuse me for bud englis, for bad expalining, for confusion, thank

tridge (approved):

tridge:
@priseborough based on your comment I've merged this fix now, and we can do the more complex fix as a future enhancement

PR #29235, opened 2025-02-06
AP_Proximity: Hexsoon Radar support

rmackay9:
This is an alternative to https://github.com/ArduPilot/ardupilot/pull/28914 which adds support for the Hexsoon 77G (and similar) radars as a proximity sensor.

I think the only significant question is whether we're happy to have the settings duplication when it is not technically required.  One thin argument for allowing the duplication is that is may allow a divergence of features without requiring users to change their settings.

This has been tested on real hardware to ensure that it works and also that the existing MR72 driver continues to work.  To test the following parameters should be set:

PRX1_TYPE = 18 (Hexsoon Radar)
CAN_P1_DRIVER = 1
CAN_P1_BITRATE = 500000
CAN_D1_PROTOCOL = 15 (Hexsoon)

BTW, there will be a follow-up PR to allow both the MR72 and Hexsoon radar to be used as rangefinders


libraries/AP_Proximity/AP_Proximity_MR72_CAN.cpp
```diff
@@ -34,10 +34,13 @@ const AP_Param::GroupInfo AP_Proximity_MR72_CAN::var_info[] = {
...
 AP_Proximity_MR72_CAN::AP_Proximity_MR72_CAN(AP_Proximity &_frontend,
                                      AP_Proximity::Proximity_State &_state,
-                                     AP_Proximity_Params& _params):
+                                     AP_Proximity_Params& _params,
+                                     Manufacturer manufacture):
     AP_Proximity_Backend(_frontend, _state, _params)
 {
-    multican_MR72 = NEW_NOTHROW MultiCAN{FUNCTOR_BIND_MEMBER(&AP_Proximity_MR72_CAN::handle_frame, bool, AP_HAL::CANFrame &), AP_CAN::Protocol::NanoRadar, "MR72 MultiCAN"};
+    // determine CAN protocol based on manufacturer
+    AP_CAN::Protocol can_protocol = manufacture == Manufacturer::NanoRadar ? AP_CAN::Protocol::NanoRadar : AP_CAN::Protocol::Hexsoon;
```
peterbarker, line 42:
```suggestion
    AP_CAN::Protocol can_protocol = ((manufacture == Manufacturer::NanoRadar) ? AP_CAN::Protocol::NanoRadar : AP_CAN::Protocol::Hexsoon);
```

peterbarker: I'm finding it *really* strange that we would call a CAN protocol "Hexsoon".  I mean... that's a very limiting thing to do!

I'd prefer if we had AP_CAN::Protocol::NanoRadarMR72 and AP_CAN::Protocol::HexsoonMR72 (or at least something else to identify it.

libraries/AP_Proximity/AP_Proximity.cpp
```diff
@@ -229,7 +229,7 @@ void AP_Proximity::init()
 #if AP_PROXIMITY_MR72_ENABLED
         case Type::MR72:
             state[instance].instance = instance;
-            drivers[instance] = NEW_NOTHROW AP_Proximity_MR72_CAN(*this, state[instance], params[instance]);
+            drivers[instance] = NEW_NOTHROW AP_Proximity_MR72_CAN(*this, state[instance], params[instance], AP_Proximity_MR72_CAN::Manufacturer::NanoRadar);
```
peterbarker, line 232: Why not just pass the protocol name directly here?

libraries/AP_Proximity/AP_Proximity_MR72_CAN.cpp
rishabsingh3003: We also call the sensors from NanoRadars CAN protocol, "NanoRadar" and also "Benewake" and so on...

rishabsingh3003 (approved):
If the Hexsoon absolutely must have it own parameter name then I think this is a good approach.

libraries/AP_Proximity/AP_Proximity.cpp
rmackay9: We share the CAN protocol now so the argument in question hasn't been added any more

libraries/AP_Proximity/AP_Proximity_MR72_CAN.cpp
rmackay9: I've renamed the existing "NanoRadar" to "RadarCAN" so no need to worry about the "Hexsoon" can protocol any more

rmackay9:
@rishabsingh3003 what do you think about renaming the "NanoRadar" CAN protocol to "RadarCAN"?

tridge (changes requested):
should combine switch
merge when fixed

libraries/AP_Proximity/AP_Proximity.cpp
```diff
@@ -253,6 +253,12 @@ void AP_Proximity::init()
             }
             break;
 #endif
+#if AP_PROXIMITY_HEXSOONRADAR_ENABLED
+        case Type::Hexsoon_Radar:
```
tridge, line 257: can be part of the switch above

rmackay9: txs for the review, fixed!

rmackay9:

PR #29236, opened 2025-02-06
Re-enable examples in CI

peterbarker:
This gives us a bit more coverage, meaning more chance of detecting breakage in CI.

I've adjusted the suite to cope with examples-that-exit, something we didn't do before.  We don't usually "exit" as that's problematic on real hardware.  But here we are.

### DIFF
--- a/.github/workflows/test_unit_tests.yml
+++ b/.github/workflows/test_unit_tests.yml
@@ -113,8 +113,8 @@ jobs:
         ]
         config: [
             unit-tests,
+            examples,
             sitl
-            # examples,
         ]
     steps:
       # git checkout the PR
--- a/Tools/autotest/autotest.py
+++ b/Tools/autotest/autotest.py
@@ -96,7 +96,7 @@ def build_binaries():
 
 def build_examples(**kwargs):
     """Build examples."""
-    for target in 'Pixhawk1', 'navio', 'linux':
+    for target in 'Pixhawk1', 'navio', 'linux', 'sitl':
         print("Running build.examples for %s" % target)
         try:
             util.build_examples(target, **kwargs)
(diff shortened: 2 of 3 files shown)


rmackay9:
I'm not really qualified but it looks OK to me

Georacer (commented):
As far as I understand, this PR:

- Adds `sitl` as a target for building examples, along with `Pixhawk1`, `navio` and `Linux`.
This will build the examples .cpp files that are found in the various libraries, which are imported by each target.

So I expect a lot more examples to be compiled, since SITL imports a lot of stuff.

- It then runs those examples.
**What I don't understand**: The run step runs the examples found only in the SITL directory, not in the other targets. This makes sense, since you can't simulate/run these.
But if we didn't build `sitl` (as we weren't before), then what was there to run? Did we rely on a non-clean build directory? I think so: the CI script would build the examples first, with `sitl` as its board target.
I don't think `sitl` was being silently built: in my local test `build/sitl` is not updated when I `autotest.py build.examples run.examples`.

That said, I think this change at least has the benefit of building the examples in one step.

Perhaps consider editing `build_ci.sh` to do away with the preliminary build steps, if they're not necessary anymore?

peterbarker:
>       But if we didn't build `sitl` (as we weren't before), then what was there to run? Did we rely on a non-clean build directory? I think so: the CI script would build the examples first, with `sitl` as its board target.

We used to run the Linux binaries.

> That said, I think this change at least has the benefit of building the examples in one step.
> 
> Perhaps consider editing `build_ci.sh` to do away with the preliminary build steps, if they're not necessary anymore?

Still necessary!

Georacer (approved):

PR #29237, opened 2025-02-06
Plane: added systemID as an aux function

tridge:
This adds VTOL systemID as an AUX function to quadplanes. It works in QSTABILIZE, QHOVER and QLOITER modes.
In QLOITER it multiplies the XY position and velocity gains by SID_XY_CTRL_MUL to reduce the impact of the XY controller on the systemID result
The AP_SystemID class is structured in a way to make it fairly easy to move this to a common library in the future if copter wants to adopt the same approach


timtuxworth:
I'm struggling to understand why this is called "System ID". It doesn't seem to have anything to do with identifying the system - which is what we have SYSID_THISMAV for no?

tridge:
> I'm struggling to understand why this is called "System ID"

"system identification" is a control systems term that refers to characterising the dynamics of a system (a vehicle in this case). In copter it is a flight mode called SYSTEMID flight mode.
and yes, this means we have overlap in naming with the MAVLink SYSID. Welcome to the world of UAVs and re-used names.

ArduPlane/systemid.cpp
```diff
@@ -0,0 +1,243 @@
...
+  handle systemid via an auxillary switch
+ */
+
+const AP_Param::GroupInfo AP_SystemID::var_info[] = {
+
+    // @Param: _AXIS
+    // @DisplayName: System identification axis
+    // @Description: Controls which axis are being excited.  Set to non-zero to see more parameters
+    // @User: Standard
+    // @Values: 0:None, 1:Input Roll Angle, 2:Input Pitch Angle, 3:Input Yaw Angle, 4:Recovery Roll Angle, 5:Recovery Pitch Angle, 6:Recovery Yaw Angle, 7:Rate Roll, 8:Rate Pitch, 9:Rate Yaw, 10:Mixer Roll, 11:Mixer Pitch, 12:Mixer Yaw, 13:Mixer Thrust, 14:Measured Lateral Position, 15:Measured Longitudinal Position, 16:Measured Lateral Velocity, 17:Measured Longitudinal Velocity, 18:Input Lateral Velocity, 19:Input Longitudinal Velocity
```
IamPete1, line 18: Should add "VTOL" to these, would be good to support in FW flight in the future.

ArduPlane/systemid.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+    if (!plane.control_mode->supports_systemid()) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: Not supported in mode %s", plane.control_mode->name());
+        return;
+    }
+    if (!hal.util->get_soft_armed()) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: must be armed");
+        return;
+    }
+    if (!plane.quadplane.enabled()) {
```
IamPete1, line 102: This case should be handled by `plane.control_mode->supports_systemid()`

timtuxworth:
> > I'm struggling to understand why this is called "System ID"
> 
> "system identification" is a control systems term that refers to characterising the dynamics of a system (a vehicle in this case). In copter it is a flight mode called SYSTEMID flight mode. and yes, this means we have overlap in naming with the MAVLink SYSID. Welcome to the world of UAVs and re-used names.

I'm aware of this problem hence bringing it up. I've found it helps to provide some way to differentiate overloaded terms.

Could the new (user facing) name be something like "Dynamic System Identification" or "System Model Identification" or DSMI (Dynamic System Model Identification)?

ArduPlane/systemid.cpp
tridge: I was planning to re-use the same axes (where relevant, can't do lateral vel and pos).

bnsgeyer: remove values 14-19.  I did not implement them in my version of sysid.  it doesn't look like you did either.  Lets keep it to the attitude controller.  We can expand to the position controller later.

ArduPlane/systemid.cpp
```diff
@@ -0,0 +1,283 @@
...
+void AP_SystemID::stop()
+{
+    if (running) {
+        running = false;
+        attitude_offset_deg.zero();
+        throttle_offset = 0;
+
+        auto *attitude_control = plane.quadplane.attitude_control;
+        attitude_control->bf_feedforward(restore.att_bf_feedforward);
+        attitude_control->rate_bf_roll_sysid(0);
```
bnsgeyer, line 158: Setting these to zero is not necessary.  These are set to zero in the attitude controller at the end of every loop.
https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/quadplane.cpp#L1948

ArduPlane/systemid.cpp
```diff
@@ -0,0 +1,283 @@
...
+// @Field: F: Instantaneous waveform frequency
+// @Field: Gx: Delta angle, X-Axis
+// @Field: Gy: Delta angle, Y-Axis
+// @Field: Gz: Delta angle, Z-Axis
+// @Field: Ax: Delta velocity, X-Axis
+// @Field: Ay: Delta velocity, Y-Axis
+// @Field: Az: Delta velocity, Z-Axis
+
+// log system id
+void AP_SystemID::log_data() const
```
bnsgeyer, line 256: I see you removed the logging of RATE and ANG from this function.  It is imperative that these all be logged at the same rate.  Why don't you use the same technique as copter to allow logging at multiples of the fast loop rate.  this is important to allow users to choose an appropriate rate for their use case.

bnsgeyer (changes requested):

ArduPlane/systemid.cpp
tridge: > It is imperative that these all be logged at the same rate.

unlike copter, the default setup in plane is to log all these at the full loop rate, so they are all logged at the same rate. We could force the LOG_BITMASK in case a user has changed that, or we could just document that the defaults should be used.

> Why don't you use the same technique as copter to allow logging at multiples of the fast loop rate. this is important to allow users to choose an appropriate rate for their use case.

it would be good to discuss why this is needed. Is it to make logs smaller? The sysid runs for a short time (usually) and the extra complexity to support down sampling didn't seem worthwhile, but let's discuss when you have a chance

ArduPlane/systemid.cpp
tridge: this was paranoia to ensure that even a single loop is not incorrect

ArduPlane/systemid.cpp
lthall: The data logging for system ID is extreemly important if we intend to make this a reliable system. Leaving logging to other parts of the code invites crashed aircraft after small logging problems result in unstable tunes that look great in the too. This is why logging perfection is so important in system ID.

bnsgeyer: I agree with @lthall that the logging is extremely important.  Any time difference between signals can cause issues when either using it to develop a model or tune the aircraft.  In most cases, I use the gyro rates from SIDD because they haven't been filtered by the INS filters and I use the motors inputs from RATE.  Having these signals logged at the same time is very important.  Keeping them all in the SystemID log function also provides higher confidence that future developers won't change how things are logged for System ID.  

As for the various rates that copter allows, I am not as bothered by this.  I think it can be an issue for users if their flight controllers have issues logging at the fast rate.   I know my F7s were unable to keep up at 400hz logging rate.  Thus it was very useful to be able to use 100hz so that I could allow the F7 to keep up with the logging.

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -679,6 +679,10 @@ void AC_PosControl::update_xy_controller()
     // add velocity feed-forward scaled to compensate for optical flow measurement induced EKF noise
     vel_target *= ahrsControlScaleXY;
 
+    if (_xy_control_scale_factor > 0) {
```
lthall, line 682: Why was this added?

bnsgeyer: @lthall 
@tridge added this to be able to scale the velocity and position inputs of the position controller while using System ID feature in a position controlled mode.  Currently he set the scaler to 0.1 to reduce the position and velocity feedback while system ID is done.

lthall: Cool.
We should change this to match the pattern in Attitude Control where we set it to 1.0 by default.

If this is a "Control" scale factor we should also update the accel_target like we do with the ahrsControlScaleXY.
accel_target *= ahrsControlScaleXY;

ArduPlane/systemid.cpp
tridge: done

ArduPlane/systemid.cpp
tridge: changed to 1, and I still prefer to force it to 1, not rely on next loop

libraries/AC_AttitudeControl/AC_PosControl.cpp
tridge: i've changed it to default to 1 and removed the > 0 checks, it already did the accel_target

bnsgeyer:
> needs metadat update to RCx_OPTION params
@Hwurzburg 
Done.  Thanks!

bnsgeyer (approved):
@tridge I think this is ready.  Do we need to take this to devcall or can we merge it?

Hwurzburg (approved):

ArduPlane/systemid.cpp
IamPete1: bump

ArduPlane/systemid.cpp
```diff
@@ -0,0 +1,294 @@
...
+            break;
+        case AxisType::MIX_PITCH:
+            attitude_control->actuator_pitch_sysid(waveform_sample);
+            break;
+        case AxisType::MIX_YAW:
+            attitude_control->actuator_yaw_sysid(waveform_sample);
+            break;
+        case AxisType::MIX_THROTTLE:
+            throttle_offset += waveform_sample;
+            break;
```
IamPete1, line 238: This is the same in copter, but `+=` looks wrong to me. surely it should just be `=` @bnsgeyer?

ArduPlane/systemid.cpp
```diff
@@ -0,0 +1,294 @@
...
+    plane.ins.get_delta_angle(delta_angle, delta_angle_dt);
+
+    Vector3f delta_velocity;
+    float delta_velocity_dt;
+    plane.ins.get_delta_velocity(delta_velocity, delta_velocity_dt);
+
+    if (is_positive(delta_angle_dt) && is_positive(delta_velocity_dt)) {
+        const float dt_ang_inv = 1.0 / delta_angle_dt;
+        const float dt_vel_inv = 1.0 / delta_velocity_dt;
+        AP::logger().WriteStreaming("SIDD", "TimeUS,Time,Targ,F,Gx,Gy,Gz,Ax,Ay,Az",
```
IamPete1, line 275: There might be a argument for this to not be streaming, if you drop a few logs it will mess up the the analysis.

ArduPlane/systemid.h
```diff
@@ -0,0 +1,89 @@
...
+    void update();
+
+    static const struct AP_Param::GroupInfo var_info[];
+
+    const Vector3f &get_attitude_offset_deg(void) const {
+        return attitude_offset_deg;
+    }
+    float get_throttle_offset(void) const {
+        return throttle_offset;
+    }
```
IamPete1, line 28: Because this is not called right next the `update` function like `get_attitude_offset_deg` is I would like it to be clear that it will return 0 if not running. 

```suggestion
    float get_throttle_offset(void) const {
        return running ? throttle_offset : 0.0;
    }
```

ArduPlane/systemid.cpp
IamPete1: The message is still useful I guess. But it need to be above the mode check or it will never be seen.

IamPete1 (commented):
Unlike quicktune this is now enabling inputs not from the user in existing modes. I think I would prefer to have a dedicated `Q_SYSID` mode, at the very least that makes the behavior consistent with copter.

tridge:
@bnsgeyer we should discuss if we can put all the needed data in the single SIDD message instead of spread across ANG, ATT, RATE and SIDD. In SITL at speedup over 2 we lose lots of log messages:
image

ArduPlane/systemid.cpp
bnsgeyer: @IamPete1 you are correct.  because this is done as an offset which is added to the throttle elsewhere, then it should be =.  In Copter, the throttle is pulled into the mode and then modified and thus why it is a += rather than =.

ArduPlane/systemid.cpp
```diff
@@ -0,0 +1,294 @@
...
+                                    waveform_time, waveform_sample, waveform_freq_rads / (2 * M_PI),
+                                    degrees(delta_angle.x * dt_ang_inv),
+                                    degrees(delta_angle.y * dt_ang_inv),
+                                    degrees(delta_angle.z * dt_ang_inv),
+                                    delta_velocity.x * dt_vel_inv,
+                                    delta_velocity.y * dt_vel_inv,
+                                    delta_velocity.z * dt_vel_inv);
+
+        // log attitude controller at the same rate
+        plane.Log_Write_Attitude(true);
```
tridge, line 287: change ATT to be lower rate

ArduPlane/systemid.cpp
bnsgeyer: done.

ArduPlane/systemid.cpp
bnsgeyer: moved above mode check

ArduPlane/systemid.cpp
bnsgeyer: Removed system ID logging of this message.  reverts back to original logging rates

tridge:
> Unlike quicktune this is now enabling inputs not from the user in existing modes. I think I would prefer to have a dedicated `Q_SYSID` mode, at the very least that makes the behavior consistent with copter.

we discussed this on the last dev call and there was support for moving copter to using an AUX function like in this PR, to allow systemid in multiple modes.
Supporting systemid in position and height controlled modes makes it much more accessible to a much wider range of users. In my experience most of our user community do not have the skills to reliably fly in STABILIZE mode

ArduPlane/systemid.cpp
tridge: that won't help, as the other messages needed are streaming too

PR #29238, opened 2025-02-06
Fix clang compiler warnings

peterbarker:
Unused variables and wrong abs function used.

CI-tested only.  And locally for compilation...

.... also now stops scripting using OwnPtr for i2c devices; this was tested on a ZeroOneX6 with an RM3100 and an example script which triggers a self-test on the device.


tridge (approved):

tridge:
ok to merge once testing scripting restart with i2c script

PR #29242, opened 2025-02-06
Tools: Automate the creation of apm.pdef.xml files

amilcarlucas:
Whenever a tag is pushed, create the corresponding parameter metadata documentation
apm.pdef.xml file on the https://autotest.ardupilot.org/Parameters/versioned/ URL


peterbarker:
This should be a triggered action on the autotest server, surely?

I'd really rather do this out of `build_binaryes.py` on the server... which *also* stares at the tags created and acts on them.

PR #29243, opened 2025-02-06
GCS_Common: better handle negative intervals

robertlong13:
Treat all negative interval values like -1. This is preferable to the current behavior, where we treat those like 1000Hz requests.

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -3157,6 +3157,8 @@ MAV_RESULT GCS_MAVLINK::set_message_interval(uint32_t msg_id, int32_t interval_u
     } else if (interval_us == -1) {
         // minus-one is "stop sending"
         interval_ms = 0;
+    } else if (interval_us < 0) {  
+        return MAV_RESULT_DENIED; 
     } else if (interval_us < 1000) {
         // don't squash sub-ms times to zero
         interval_ms = 1;


robertlong13:
GCSs shouldn't send request for negative rates, but this came about because this Mission Planner bug I introduced: https://github.com/ArduPilot/MissionPlanner/pull/3487

timtuxworth:
Tested - looking good now.

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3127,7 +3127,7 @@ MAV_RESULT GCS_MAVLINK::set_message_interval(uint32_t msg_id, int32_t interval_u
             // at.
             interval_ms = 0;
         }
-    } else if (interval_us == -1) {
+    } else if (interval_us < 0) {
```
peterbarker, line 3130:
```suggestion
    } else if (interval_us == -1) {
  ```

File: libraries/GCS_MAVLink/GCS_Common.cpp [hunk 2]
```diff
@@ -3127,7 +3127,7 @@ MAV_RESULT GCS_MAVLINK::set_message_interval(uint32_t msg_id, int32_t interval_u
...
         // minus-one is "stop sending"
         interval_ms = 0;
     } else if (interval_us < 1000) {
```
peterbarker, line 3133:
```suggestion
    } else if (interval_us < 0) {
        return MAV_RESULT_DENIED;
    } else if (interval_us < 1000) {
```

peterbarker (changes requested):

peterbarker (approved):
LGTM

rmackay9:

PR #29244, opened 2025-02-07
Cotper: RTL correct height for `RTL_ALT_FINAL` > 0 (home vs origin)

IamPete1:
I didn't know we had this feature, it allows the vehicle to return to the home location but hover at some given altitude above home.

When doing code review I spotted that the `descent_target` location is setup relative to home and then passed to the position controller which is relative to origin. This converts it to above origin and then updates the comparison function to use inertial nav altitude which is also relative to origin. 

I suspect this has been broken for ages, I wonder if anyone uses the feature? Although admittedly the difference between home and origin is typically small so it may not have been obvious. Maybe we could remove the functionality?

But I can understand why you might want this, maybe we should fix it and add a autotest?

### DIFF
--- a/ArduCopter/mode_rtl.cpp
+++ b/ArduCopter/mode_rtl.cpp
@@ -312,7 +312,7 @@ void ModeRTL::descent_run()
     attitude_control->input_thrust_vector_heading(pos_control->get_thrust_vector(), auto_yaw.get_heading());
 
     // check if we've reached within 20cm of final altitude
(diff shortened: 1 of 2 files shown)


IamPete1:
Master flys to -10 meters relative to home which is the correct altitude relative to origin:
image

This PR as expected, fly to 10 relative home:
image

Test that made the logs is added.

There are still some ways you could trip it up, the path is calculated when RTL is entered so if you change origin on the way then we will have the altitude difference as error. But I suspect the same would be true for normal RTL if you moved home.

rmackay9 (approved):

PR #29247, opened 2025-02-07
AP_Camera_servo: allow focus and zoom control

Williangalvani:
We want to control our camera's focus and pwm through ardupilot.

### DIFF
--- a/libraries/AP_Camera/AP_Camera_Servo.cpp
+++ b/libraries/AP_Camera/AP_Camera_Servo.cpp
@@ -6,6 +6,14 @@
 
 extern const AP_HAL::HAL& hal;
 
+// initialize the AP_Camera_Servo driver
+void AP_Camera_Servo::init()
+{
+    // set the zoom and focus to the trim point
+    SRV_Channels::set_output_scaled(SRV_Channel::k_cam_zoom, 500);
+    SRV_Channels::set_output_scaled(SRV_Channel::k_cam_focus, 500);
+}
+
 // update - should be called at 50hz
 void AP_Camera_Servo::update()
 {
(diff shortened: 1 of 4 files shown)


libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -43,6 +49,28 @@ bool AP_Camera_Servo::trigger_pic()
     return true;
 }
 
+
+bool AP_Camera_Servo::set_zoom(ZoomType zoom_type, float zoom_value) {
```
rmackay9, line 53: Hi there, thanks for this.  Small nitpick is in the cpp files we put the bracket on the next line

libraries/AP_Camera/AP_Camera_Servo.h
```diff
@@ -29,12 +29,20 @@ class AP_Camera_Servo : public AP_Camera_Backend
     // Constructor
     using AP_Camera_Backend::AP_Camera_Backend;
 
+    void init() override;
```
rmackay9, line 32: let's put the init lower like it is in other drivers.  Also if possible can we put a comment above it? I know it's pretty obvious what "init" is for but all our other methods have comments so I'd like to keep it consistent

libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -6,6 +6,12 @@
 
 extern const AP_HAL::HAL& hal;
 
+
+void AP_Camera_Servo::init() {
```
rmackay9, line 10: here too let's put a comment and put the bracket on the next line

rmackay9:
Except for the minor formatting items this looks pretty good to me, thanks!

Williangalvani:
updated, rebased, and retested!

libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -43,6 +51,30 @@ bool AP_Camera_Servo::trigger_pic()
     return true;
 }
 
+
+bool AP_Camera_Servo::set_zoom(ZoomType zoom_type, float zoom_value)
+{
+    if (zoom_type == ZoomType::RATE) {
```
rmackay9, line 57: Hi @Williangalvani,

Thanks for this.  It seems like the absolute zoom output will be slightly incrementing or decrementing each time the set-zoom is called.  This is different (I think) to how other cameras implement the set-zoom command.  The others will continue to zoom until a follow-up set-zoom command is received

Williangalvani: right!
[this](https://mavlink.io/en/messages/common.html#FOCUS_TYPE_STEP) Is what I had in mind. I just updated the code and enums.

Williangalvani: actually I'm just fixing the rate implementation. the step one is likely to cause jitter...

rmackay9 (approved):

rmackay9:
Hi this looks OK to me.

It would be fairly straight forward to also support absolute zoom and focus in the new functions but even as-is this is an improvement so approved!

Williangalvani:
> Hi this looks OK to me.
> 
> It would be fairly straight forward to also support absolute zoom and focus in the new functions

I intend to do that soon, as soon as I have a good way to test it on our gcs side =]

PR #29249, opened 2025-02-08
Tools: explicitly set AP_PERIPH_XX_ENABLED defines to 0 for sitl_periph_gps and sitl_periph_battmon

shiv-tyagi:
Thanks to @peterbarker for finding out that these were broken.
We need to explicitly set these defines to 0 for turning off the features. Just leaving them undefined no longer works.


peterbarker (approved):

peterbarker:
Merged, thanks!

PR #29252, opened 2025-02-08
Add support for multi-GCS operation

Davidsastresas:
This PR adds support for the recently added multi-GCS mavlink subprotocol https://github.com/mavlink/mavlink/pull/2158. This allows graceful change in control ownership on a system with multiple GCS. Please note this subprotocol does not attempt to cover security, it assumes all the operators are in contact between them and they work in a collaborative manner.

For more information about the protocol itself, please read that thread instead.

### How the implementation works
- GCS UI is populated based on CONTROL_STATUS being received. For single GCS operations, it is probably desirable to not show such UI, and not send that message at all. For that reason I left the evaluation to send this message or not at initialization, based on SYSID_ENFORCE parameter. If set to 0 at boot up, this message won't be sent at all, and no UI will be populated.

- When the current GCS in control releases control, it only sets takeover allowed to true, but does not change sysid_mygcs. Mavlink protocol says we should use 0, no one in control, but maybe this is too hardcore at the moment, until we are sure all of this makes total sense. So the current solution of setting takeover allowed to true from my point of view is a balance. Let me know your thoughts.

- As it is now it will assume all the GCS are connected to the same link. Is it worth implementing something so this can work between different links? for example a GCS connected on serialport x to a 900MHz radio telemetry and another GCS possibly connected to a different serial port to a serial to ethernet converter and a digital Ip link.

- On link loss, when we stop seing heartbeats from the GCS in control, we set the allow override, so other GCS can take over automatically.

- Right now, when a GCS that is not our SYSID_MYGCS sends a command it just ignores the command, so the GCS ack mechanism times out and we see a "no response" message. When we were working on this at the mavlink level, @hamishwillee suggested the use of a new MAV_RESULT_PERMISSION_DENIED, but in the end we ditched it due to the extra complexity. Maybe we should come back to this later if this concept works, to make more user friendly when a GCS not in control is exchanging messages with the vehicle.
In any case, I don't think this one in particular is too bad at the moment.

- As we are basing this on sysid_enforce, when a GCS is not in control it will have trouble to download the mission and parameters. Maybe we should allow parameters and mission to be sent to GCS that are not in control?

### Notes
We need this mavlink changes https://github.com/ArduPilot/mavlink/pull/382 for this PR to build

### Pending work to do
- [ ] This PR only adds support for Copter. If everything looks good we should port it to the other kinds of vehicles too.
- [ ] Make this work between different mavlink channels?
- [ ] Allow parameters and mission messages to be sent to a GSC that is not in control?

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -3259,7 +3259,8 @@ MAV_RESULT GCS_MAVLINK::handle_request_operator_control(const mavlink_command_in
                 msg.sysid, // Param1: Sysid of the GCS requesting control
                 1, // Param2: Release/request control. If we are here this should always be 1 (request). 0 would not make sense anyway 
                 packet.param3, // Param3: Allow takeover, this way the GCS in control can prompt the operator with the specific type of control request 
-                0, 0, 0, 0);
+                packet.param4, // Param4: Timeout in seconds before a request to a GCS to allow takeover is assumed to be rejected. This is used to display the timeout graphically on requestor and GCS in control.
+                0, 0, 0);
             // We should answer result failed, see MAV_CMD_REQUEST_OPERATOR_CONTROL documentation for more information
             return MAV_RESULT_FAILED;
         }
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -1151,6 +1151,7 @@ ap_message GCS_MAVLINK::mavlink_id_to_ap_message_id(const uint32_t mavlink_id) c
 #endif
         { MAVLINK_MSG_ID_AVAILABLE_MODES, MSG_AVAILABLE_MODES},
         { MAVLINK_MSG_ID_AVAILABLE_MODES_MONITOR, MSG_AVAILABLE_MODES_MONITOR},
+        { MAVLINK_MSG_ID_CONTROL_STATUS, MSG_CONTROL_STATUS},
             };
 
     for (uint8_t i=0; i<ARRAY_SIZE(map); i++) {
@@ -1515,6 +1516,12 @@ void GCS_MAVLINK::update_send()
 #if HAL_MAVLINK_INTERVALS_FROM_FILES_ENABLED
         initialise_message_intervals_from_config_files();
 #endif
+        // We evaluate SYSID_ENFORCE parameter to send or not CONTROL_STATUS. This message is what
+        // populates GCS UI for multi GCS control
+        if (sysid_enforce()) {
+            set_mavlink_message_id_interval(MAVLINK_MSG_ID_CONTROL_STATUS, 5000);
+        }
+
         deferred_messages_initialised = true;
     }
 
@@ -3202,6 +3209,78 @@ MAV_RESULT GCS_MAVLINK::handle_command_request_message(const mavlink_command_int
 
     send_message(id);
     return MAV_RESULT_ACCEPTED;
+}
+
+MAV_RESULT GCS_MAVLINK::handle_request_operator_control(const mavlink_command_int_t &packet, const mavlink_message_t &msg)
+{
+    // Param2 is 0: Release control, 1: Request control.
+    bool isRequestControl;
+    if (is_equal(packet.param2, 1.0f)) {
+        isRequestControl = true;
+    } else if (is_zero(packet.param2)) {
+        isRequestControl = false;
+    } else {
+        // Malformed packet, return denied
+        return MAV_RESULT_DENIED;
+    }
+
+    // Param3 is 0: Do not allow takeover, 1: Allow takeover.
+    bool setTakeoverAllowed;
+    if (is_equal(packet.param3, 1.0f)) {
+        setTakeoverAllowed = true;
+    } else if (is_zero(packet.param3)) {
+        setTakeoverAllowed = false;
+    } else {
+        // Malformed packet, return denied
+        return MAV_RESULT_DENIED;
+    }
+
+    if (isRequestControl) {
+        // This part would proceed the same if takeover is allowed, or if this is coming from the current GCS in control, 
+        // for example to modify takover allowed
+        if (control_takeover_allowed() || msg.sysid == sysid_my_gcs()) {
+            // Set takeover allowed, so other GCS can get control automatically or they need to ask first to current GCS in control
+            set_control_takeover_allowed(setTakeoverAllowed);
+            // Set current GCS in control if this is coming from a different GCS
+            if (msg.sysid != sysid_my_gcs()) {
+                set_sysid_my_gcs(msg.sysid);
+            }
+            // And send control status inmediatly to notify all GCS
+            send_control_status();
+            return MAV_RESULT_ACCEPTED;
+        // Else send a command to the current GCS in control, so current active operator can be prompted to allow takeover for the GCS asking
+        } else {
+            mavlink_msg_command_long_send(
+                chan,
+                sysid_my_gcs(),
+                0, // Should we retrieve here GCS compid and use it instead?
+                MAV_CMD_REQUEST_OPERATOR_CONTROL,
+                0,
+                msg.sysid, // Param1: Sysid of the GCS requesting control
+                1, // Param2: Release/request control. If we are here this should always be 1 (request). 0 would not make sense anyway 
+                packet.param3, // Param3: Allow takeover, this way the GCS in control can prompt the operator with the specific type of control request 
+                0, 0, 0, 0);
+            // We should answer this on this case, see MAV_CMD_REQUEST_OPERATOR_CONTROL documentation for more information
+            return MAV_RESULT_PERMISSION_DENIED;
+        }
+    // Release control
+    } else {
+        // Sanity check, double check that this is the GCS in control
+        if (msg.sysid == sysid_my_gcs()) {
+            // Set takeover allowed, so other GCS can get control automatically or they need to ask first to current GCS in control
+            set_control_takeover_allowed(true);
+            // As per Mavlink protocol, see MAV_CMD_REQUEST_OPERATOR_CONTROL, we should set sysid_my_gcs to 0, so CONTROL_STATUS emits 0
+            // (nobody in control), so other GCS can display properly lthe current control status of the vehicle. Should we be compliant here
+            // and set sysid_my_gcs to 0?
+            // set_sysid_my_gcs(0)
+            // Finally send a CONTROL_STATUS inmediately to notify the GCSs on the system
+            send_control_status();
+            return MAV_RESULT_ACCEPTED;
+        // A release control command should always be sent from the GCS in control. Return failed otherwise
+        } else {
+            return MAV_RESULT_FAILED;
+        }
+    }
 }
 
 bool GCS_MAVLINK::get_ap_message_interval(ap_message id, uint16_t &interval_ms) const
@@ -5610,6 +5689,8 @@ MAV_RESULT GCS_MAVLINK::handle_command_int_packet(const mavlink_command_int_t &p
     case MAV_CMD_REQUEST_MESSAGE:
         return handle_command_request_message(packet);
 
+    case MAV_CMD_REQUEST_OPERATOR_CONTROL:
+        return handle_request_operator_control(packet, msg);
     }
 
     return MAV_RESULT_UNSUPPORTED;
@@ -5960,6 +6041,18 @@ bool GCS_MAVLINK::send_relay_status() const
     return relay->send_relay_status(*this);
 }
 #endif  // AP_MAVLINK_MSG_RELAY_STATUS_ENABLED
+
+void GCS_MAVLINK::send_control_status() const
+{
+    uint8_t flags = GCS_CONTROL_STATUS_FLAGS_SYSTEM_MANAGER;
+    flags |= control_takeover_allowed() ? GCS_CONTROL_STATUS_FLAGS_TAKEOVER_ALLOWED : 0;
+    mavlink_msg_control_status_send(
+        chan,
+        // System ID of the system currently in control of this MAV
+        sysid_my_gcs(),
+        // Flags: Automatic takeover allowed and this system controlling the rest of components of this system ID ( always true for an autopilot )
+        flags);
+}
 
 void GCS_MAVLINK::send_autopilot_state_for_gimbal_device() const
 {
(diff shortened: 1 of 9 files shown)


timtuxworth:
> As it is now it will assume all the GCS are connected to the same link. Is it worth implementing something so this can work between different links? for example a GCS connected on serialport x to a 900MHz radio telemetry and another GCS possibly connected to a different serial port to a serial to ethernet converter and a digital Ip link.

Absolutely - I do this all the time

hamishwillee:
> When the current GCS in control releases control, it only sets takeover allowed to true, but does not change sysid_mygcs. Mavlink protocol says we should use 0, no one in control, but maybe this is too hardcore at the moment, until we are sure all of this makes total sense. So the current solution of setting takeover allowed to true from my point of view is a balance. Let me know your thoughts.

No, you need to change [`CONTROL_STATUS.sysid_in_control`](https://mavlink.io/en/messages/development.html#CONTROL_STATUS) to zero in this case.

The reason is that this information isn't just for other GCS. It is for components of the system to know what the system in charge is. If you don't set this to 0 then they will refuse commands from anything other than the GCS that used to be in control.

hamishwillee:
> As we are basing this on sysid_enforce, when a GCS is not in control it will have trouble to download the mission and parameters. Maybe we should allow parameters and mission to be sent to GCS that are not in control?

FWIW My intent with this protocol was that parameter and mission data from a controlled system should be available to all GCS, in the same way as telemetry. However a controlled GCS should only accept a mission upload from its controlling GCS.

The protocol wording is something like "a controlled component/system should only accept commands and command-like messages from its controller" so we should perhaps clarify that broadly speaking telemetry and metadata are expected to be available to all GCS.

Davidsastresas:
Thanks for the comments @hamishwillee. I would like to get the Ok from some of the AP dev team before proceeding with the changes.

@rmackay9, could you take a look at this one? Thanks!

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3204,6 +3211,79 @@ MAV_RESULT GCS_MAVLINK::handle_command_request_message(const mavlink_command_int
...
+        setTakeoverAllowed = true;
+    } else if (is_zero(packet.param3)) {
+        setTakeoverAllowed = false;
+    } else {
+        // Malformed packet, return denied
+        return MAV_RESULT_DENIED;
+    }
+
+    if (isRequestControl) {
+        // This part would proceed the same if takeover is allowed, or if this is coming from the current GCS in control,
```
peterbarker, line 3239: Break this out into a method

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3204,6 +3211,79 @@ MAV_RESULT GCS_MAVLINK::handle_command_request_message(const mavlink_command_int
...
+            // Set current GCS in control if this is coming from a different GCS
+            if (msg.sysid != sysid_my_gcs()) {
+                set_sysid_my_gcs(msg.sysid);
+            }
+            // And send control status inmediatly to notify all GCS
+            send_control_status();
+            return MAV_RESULT_ACCEPTED;
+        // Else send a command to the current GCS in control, so current active operator can be prompted to allow takeover for the GCS asking
+        } else {
+            mavlink_msg_command_long_send(
```
peterbarker, line 3253: pointless else

peterbarker, line 3253: Prefer command_int, not command_long

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3204,6 +3211,79 @@ MAV_RESULT GCS_MAVLINK::handle_command_request_message(const mavlink_command_int
...
+                chan,
+                sysid_my_gcs(),
+                0, // Should we retrieve here GCS compid and use it instead?
```
peterbarker, line 3256: No way to tell without a parameter.  I think `compid_my_gcs()` has been put forward before

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3204,6 +3211,79 @@ MAV_RESULT GCS_MAVLINK::handle_command_request_message(const mavlink_command_int
...
+                msg.sysid, // Param1: Sysid of the GCS requesting control
+                1, // Param2: Release/request control. If we are here this should always be 1 (request). 0 would not make sense anyway 
+                packet.param3, // Param3: Allow takeover, this way the GCS in control can prompt the operator with the specific type of control request 
+                packet.param4, // Param4: Timeout in seconds before a request to a GCS to allow takeover is assumed to be rejected. This is used to display the timeout graphically on requestor and GCS in control.
+                0, 0, 0);
+            // We should answer result failed, see MAV_CMD_REQUEST_OPERATOR_CONTROL documentation for more information
+            return MAV_RESULT_FAILED;
+        }
+    // Release control
+    } else {
```
peterbarker, line 3268: Pointless else

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -3204,6 +3211,79 @@ MAV_RESULT GCS_MAVLINK::handle_command_request_message(const mavlink_command_int
...
+            set_control_takeover_allowed(true);
+            // As per Mavlink protocol, see MAV_CMD_REQUEST_OPERATOR_CONTROL, we should set sysid_my_gcs to 0, so CONTROL_STATUS emits 0
+            // (nobody in control), so other GCS can display properly lthe current control status of the vehicle. Should we be compliant here
+            // and set sysid_my_gcs to 0?
+            // set_sysid_my_gcs(0)
+            // Finally send a CONTROL_STATUS inmediately to notify the GCSs on the system
+            send_control_status();
+            return MAV_RESULT_ACCEPTED;
+        // A release control command should always be sent from the GCS in control. Return failed otherwise
+        } else {
```
peterbarker, line 3281: Pointless else, and sysid check should be inverted

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -6823,9 +6920,35 @@ bool GCS_MAVLINK::accept_packet(const mavlink_status_t &status,
         return true;
     }
 
+    // Allow REQUEST_OPERATOR_CONTROL packets. This is needed for a GCS not in control
+    // to communicate with the vehicle, so it can forward the request to GCS in control.
+    if (is_control_request_packet(msg)) {
```
peterbarker, line 6925: Why does the vehicle need to be in the middle of this negotiation?

I'm not saying it's wrong.

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -5961,6 +6043,18 @@ bool GCS_MAVLINK::send_relay_status() const
 }
 #endif  // AP_MAVLINK_MSG_RELAY_STATUS_ENABLED
 
+void GCS_MAVLINK::send_control_status() const
```
peterbarker, line 6046: Everything needs to be behind a feature define

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -6823,9 +6920,35 @@ bool GCS_MAVLINK::accept_packet(const mavlink_status_t &status,
...
+    // Allow REQUEST_OPERATOR_CONTROL packets. This is needed for a GCS not in control
+    // to communicate with the vehicle, so it can forward the request to GCS in control.
+    if (is_control_request_packet(msg)) {
+        return true;
+    } 
+
     return false;
 }
 
+bool GCS_MAVLINK::is_control_request_packet(const mavlink_message_t &msg) const
```
peterbarker, line 6932: This is an *awfully* expensive function to be doing on every incoming packet!

peterbarker (changes requested):
What GCS infrastructure supports this as it currently stands?

I will create a MAVProxy module at some stage to do this of course.

libraries/GCS_MAVLink/GCS_Common.cpp
Davidsastresas: When we thought about this protocol we thought about this paradigm, and some reasons I remember:

- We thought the vehicle should have the ultimate decision to proceed or not on such negotiations. Imagine in the future we add a layer of security to this, and it relies on some kind of key that all parties need to know. And even on current implementation we have certain control over it, for example disabling SYSID_ENFORCE parameter to disable the protocol completely, including GCS UI.

- In the case of mesh radios I guess it would be good to talk GCS - GCS, but if these mesh radios don't work as they are supposed to, or if we are using non mesh radios, we rely on Autopilot mavlink router to make sure the messages arrive to the correct GCS. The initial idea was to handover control between 2 operators in mountain environment, so GCSs would not be having direct contact between them, only through vehicle.

libraries/GCS_MAVLink/GCS_Common.cpp
Davidsastresas: Noted. would MAVLINK_MULTIGCS_PROTOCOL_ENABLED work?

libraries/GCS_MAVLink/GCS_Common.cpp
Davidsastresas: I know it isn't ideal but:

- It is only awfully expensive for command long and int. For the rest of the messages it only checks msgid, similar to what we are doing above with RADIO_STATUS messages.

- What other approach could you think about to do this? We are in a situation where we are ignoring all messages from other GCS, but we really want control request commands to get in. Maybe using a different msgid for these messages we want to let in, so it is less expensive to check for them? We actually talked about this when we were building the protocol and after going back and forth several times we decided to stick with regular commands instead of specific new messages.

Davidsastresas:
Thanks for the review Peter.

**What GCS infrastructure supports this as it currently stands?**
This PR was tested using this QGC master PR https://github.com/mavlink/qgroundcontrol/pull/12410. As soon as this is confirmed I will port it to QGC stable too.

**About my motivation to persisting data**
My idea was to integrate this as seamlessly as possible on the current AP mechanism of SYSID_MYGCS and SYSID_ENFORCE, and that is how it made sense for me. I wasn't aware setting parameters that way had been discarded, I had on my mind we did it eventually for such speed messages, etc. My bad, sorry about it.

**Next steps**
Before starting to work on your suggestions, I think we need to clarify:

- What do we do about the awfully expensive message checks? https://github.com/ArduPilot/ardupilot/pull/29252#discussion_r1988847732

- About persisting data. Should we leave SYSID_MYGCS and SYSID_ENFORCE as they are now, and instead maybe make a new parameter to enable this, or rely on a #define to build the feature? If the feature is enabled and we always boot without any GCS in control, and we rely on a GCS asking control, it would be possible for this feature to work without persisting data. However in this case we could have trouble on initial handshake GCS-AP if SYSID_ENFORCE is set and our GCS id don't match.

I think the rest of the points are clear to me. I will be out a couple of weeks but later in April I will be available again to work on this.

rmackay9:
Hi @Davidsastresas,

Thanks for this, could you rebase on master?  Also if possible if you could address PeterB's review comments?

We're happy with the overall plan though

Davidsastresas:
@rmackay9 thanks for the answer. I will rebase and address the comments Peter made that are clear to me. 

However, I need some guidance/answers to complete all of them, I pointed them on my latest answer. If anybody could give me a hint it would be much appreciated, otherwise I have voids on the plan to complete this PR for it to be suitable for merging. Thanks.

> Next steps
> Before starting to work on your suggestions, I think we need to clarify:
> 
> What do we do about the awfully expensive message checks? https://github.com/ArduPilot/ardupilot/pull/29252#discussion_r1988847732
> 
> About persisting data. Should we leave SYSID_MYGCS and SYSID_ENFORCE as they are now, and instead maybe make a new parameter to enable this, or rely on a #define to build the feature? If the feature is enabled and we always boot without any GCS in control, and we rely on a GCS asking control, it would be possible for this feature to work without persisting data. However in this case we could have trouble on initial handshake GCS-AP if SYSID_ENFORCE is set and our GCS id don't match.
> 
> Should we allow parameter download ( and maybe mission download too? ) for GCS not in control? I think all current GCS implementations out there will misbehave trying to connect a vehicle not answering such requests. In the collaborative environment this is meant to work on, it would be good that all GCS are synced on this front too.

hamishwillee:
> Should we allow parameter download ( and maybe mission download too? ) for GCS not in control? I think all current GCS implementations out there will misbehave trying to connect a vehicle not answering such requests. In the collaborative environment this is meant to work on, it would be good that all GCS are synced on this front too.

FWIW Yes. The intent is to control what can command the vehicle, not reserve its telemetry or hide its mission. Generally you would not allow upload or setting of parameters. 

Note the allowed/denied should be managed by the autopilot, but a GCS aware of this feature should also not even try.

peterbarker:
So now we have https://github.com/ArduPilot/ardupilot/pull/29252 which seeks to allow for multiple system IDs to be "mygcs".

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -6823,9 +6920,35 @@ bool GCS_MAVLINK::accept_packet(const mavlink_status_t &status,
...
+{
+    bool result = false;
+    if (msg.msgid == MAVLINK_MSG_ID_COMMAND_LONG) {
+        mavlink_command_long_t command;
+        mavlink_msg_command_long_decode(&msg, &command);
+        if (command.command == MAV_CMD_REQUEST_OPERATOR_CONTROL) {
+            result = true;
+        }
+    }
```
peterbarker, line 6941:
```suggestion
```

peterbarker:
... oh, should have mentioned this before, but definitely need a timeout in here so if you haven't seen your GCS for a while something else can take control

peterbarker:
Hit an interesting thing where we want multiple GCSs to be considered "mygcs" - but do *not* want more than one to be able to provide RC inputs to the vehicle.

Davidsastresas:
Thanks for the feedback Peter. Would you mind moving the converstation here https://github.com/mavlink/mavlink/pull/2313#issuecomment-3130127517 Thanks!

libraries/GCS_MAVLink/GCS_Common.cpp
peterbarker: We have methods in the bindings which let you extract a single field value from a message - that might be a better option here.

hamishwillee:
> Hit an interesting thing where we want multiple GCSs to be considered "mygcs" - but do _not_ want more than one to be able to provide RC inputs to the vehicle.

Assume you mean that the GCS is providing stick input over MAVLink (i.e. not actual RC) the proposal in https://github.com/mavlink/mavlink/pull/2313#issuecomment-3182624190 addresses this by using operator control to mean "can control RC".  That allows management of the owner on request.

If you don't want to manage ownership of the RC part then you don't need to add any of the proposal. Though I would at least add the changes to the CONTROL_STATUS message because they allow MAVLink components to know who their owner is.

PR #29254, opened 2025-02-08
AP_Scripting: update so mavlink init intent matches behavior

tpwrules:
RX message ID buffer sizes have been shrunk to the amount actually used.

Followup to #29246

### DIFF
--- a/libraries/AP_Scripting/applets/follow-target-send.lua
+++ b/libraries/AP_Scripting/applets/follow-target-send.lua
@@ -43,9 +43,6 @@ local FOLT_ENABLE = bind_add_param("ENABLE", 1, 1)
 --]]
 local FOLT_MAV_CHAN = bind_add_param("MAV_CHAN", 2, 0)
 
--- initialize MAVLink rx with number of messages, and buffer depth
-mavlink:init(2, 5)
-
 -- send FOLLOW_TARGET message
 local function send_follow_target_msg()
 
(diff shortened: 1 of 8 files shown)


libraries/AP_Scripting/applets/follow-target-send.lua
```diff
@@ -43,8 +43,7 @@ local FOLT_ENABLE = bind_add_param("ENABLE", 1, 1)
 --]]
 local FOLT_MAV_CHAN = bind_add_param("MAV_CHAN", 2, 0)
 
--- initialize MAVLink rx with number of messages, and buffer depth
-mavlink:init(2, 5)
+-- mavlink:init(0, 0) -- we never receive, no need to init rx
```
IamPete1, line 46: I would vote to just remove.

libraries/AP_Scripting/applets/mount-poi.lua
```diff
@@ -51,8 +51,7 @@ local last_roi_switch_pos = 0           -- last known rc roi switch position.  U
 local success_count = 0                 -- count of the number of POI calculations (sent to GCS in CAMERA_FEEDBACK message)
 
 -- mavlink message definition
--- initialise mavlink rx with number of messages, and buffer depth
-mavlink.init(1, 10)
+-- mavlink.init(0, 0) -- we never receive, no need to init rx
```
IamPete1, line 54: Maybe just remove?

IamPete1 (commented):
Thanks for taking the time to do this extra fixup. I was slightly on the fence, just because presumably the existing examples have been tested and I didn't want to re-test. But I'm 99% sure this is fine.

tpwrules:
I figure the benefit is to avoid confusing people down the line.

IamPete1 (approved):
Thanks!

PR #27007, opened 2024-05-07
GCS_MAVLink: support HIGHRES_IMU

KoehlerT:
Duplicates #24258 
It allows a companion computer to query the flightcontrollers IMU for navigation and control.
This PR implements all fields for the [HIGHRES_IMU](https://mavlink.io/en/messages/common.html#HIGHRES_IMU) Mavlink message.

PR #29255, opened 2025-02-08
GCS_MAVLink: highres imu: remove all "bitmask" value

IamPete1:
Follow up to https://github.com/mavlink/mavlink/pull/2208. The all item does not add any new information so it can be removed. Support for the message was added in https://github.com/ArduPilot/ardupilot/pull/27007

Our version of MAVLink is still not upto date enough for us to switch to the native enum.


peterbarker (approved):

peterbarker:
@stephendade look OK to you?

stephendade (approved):
That all looks fine to me. LGTM

PR #29256, opened 2025-02-08
GCS_MAVLink: move from `MAV_MODE` enum to `uint8_t`

IamPete1:
Were putting values not in the `MAV_MODE` enum into it, better to use a uint8_t.

See: https://github.com/mavlink/mavlink/pull/2195

### DIFF
--- a/AntennaTracker/GCS_MAVLink_Tracker.cpp
+++ b/AntennaTracker/GCS_MAVLink_Tracker.cpp
@@ -6,7 +6,7 @@ MAV_TYPE GCS_Tracker::frame_type() const
     return MAV_TYPE_ANTENNA_TRACKER;
 }
 
-MAV_MODE GCS_MAVLINK_Tracker::base_mode() const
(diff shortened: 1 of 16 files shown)


AntennaTracker/GCS_MAVLink_Tracker.cpp
```diff
@@ -6,7 +6,7 @@ MAV_TYPE GCS_Tracker::frame_type() const
     return MAV_TYPE_ANTENNA_TRACKER;
 }
 
-MAV_MODE GCS_MAVLINK_Tracker::base_mode() const
+MAV_MODE_FLAG GCS_MAVLINK_Tracker::base_mode() const
```
peterbarker, line 9: This doesn't look right.

The return value from this will often not be an entry in that enumeration. ... `MAV_MODE` also isn't correct for the same reason.

IMO we are going to change this we should move to `uint8`.

peterbarker (approved):
LGTM

PR #29257, opened 2025-02-08
AP_Scripting: mavlink: fix potential deadlock and clean up

tpwrules:
If the second allocation failed, receiving a MAVLink message could deadlock as the script thread would never release the lock yet the main thread would see the partial initialization.

Also clean up while we are in there.

Tested on Cube Orange using the REPL that the functions still work.


IamPete1 (approved):
LGTM

PR #28983, opened 2025-01-01
Rework SPL06 to do background updates compliant with the spec

andyp1per:
Pulls in some bug fixes from betaflight to make the code more understandable.
Enabled background updates by default (which is what the DPS310 driver does)
Uses external temperature sensor as required by the spec.

This results in much more accurate baro updates from this sensor.

Tested on BetaFPV F405

PR #29263, opened 2025-02-09
AP_Baro_SPL06: Work in progress, running but without any good results.

Paku-:
This is a documentation for my attempts to make SPL06 working.

Some improvements over original version like:

- chip reset on init
- state register checks before reading and on workflow
- timing balance between temp and pressure reads
- verified math but found no problems
- some defs added for easier configuration

Can't make it working as expected.
It is possible that my specific chip is either damaged or counterfeit.


peterbarker:
@Paku- I have no reason to believe that the device doesn't work in @radiolinkW 's Pix6 flight controller.

I believe @andyp1per should have one that he can verify with.

Paku-:
Added a 3 main operation modes for a chip. So called Low Power, Standard and High Precision. Non fit into 500ms driver time im AP. Chip is to slow.

peterbarker:
@Paku- we've merged a PR from @andyp1per ( https://github.com/ArduPilot/ardupilot/pull/28983 ) .  Both SPA06 and SP06 were tested as working as part of that PR.

Is this PR still required?

Paku-:
No, it was newer required .

peterbarker:
> No, it was newer required .

OK, so this one is conflicting with some fixes that went into the driver.

I believe this sensor is being used in a lot of boards reasonably successfully.

If you think there are issues, please bring a clean PR and story with minimal changes to the driver to fix whatever needs fixing.

Closing this one now.

PR #29264, opened 2025-02-09
AP_Arming_Copter: Allow arming with unhealthy visodom in non-gps modes

Vabe7:
This is essentially the fixed version of https://github.com/ArduPilot/ardupilot/pull/29258, see more info about that at the bottom.

This allows arming in non-GPS flight modes like Stabilize, even if the visual odometry is unhealthy. The behavior is mostly copied from gps_checks.

I compiled one board with (fmuv3) and one without (Kakute F7 Mini) visual odometry. And for both boards I compiled plane and copter. I also tested it on a barebones fmuv3. With:
- VISO_TYPE=1
- EK3_SRC1_POSXY = 6
- EK3_SRC1_VELXY = 0
- EK3_SRC1_POSZ = 1
- EK3_SRC1_VELZ = 0
on Copter 4.5.7. the prearm message _visodom unhealthy_ is shown in Stabilize. With this branch the message disappears in Stabilize, but appears again when changing to Loiter.

I initially only tested it for copter on sitl and fmuv3, after opening the initial pull request I got all the error messages about compiler errors in other vehicles and boards with only 1MB of storage. Realizing my mistake I closed the pull request. Now due to the rebase and force-push I can't reopen the original pull request. I'm sorry about the inconvenience, but on a positive note, I learned a lot about git today...


rishabsingh3003:
I will test this

ArduCopter/AP_Arming_Copter.cpp
```diff
@@ -399,6 +399,31 @@ bool AP_Arming_Copter::gps_checks(bool display_failure)
...
+{
+    // check if fence requires GPS/position
+    bool fence_requires_pos = false;
+#if AP_FENCE_ENABLED
+    // if circular or polygon fence is enabled we need GPS/position
+    fence_requires_pos = (copter.fence.get_enabled_fences() & (AC_FENCE_TYPE_CIRCLE | AC_FENCE_TYPE_POLYGON)) > 0;
+#endif
+
+    // check if flight mode requires GPS/position
+    bool mode_requires_pos = copter.flightmode->requires_GPS() || fence_requires_pos || (copter.simple_mode == Copter::SimpleMode::SUPERSIMPLE);
```
peterbarker, line 414:
```suggestion
    location_required |= copter.flightmode->requires_GPS() || fence_requires_pos || (copter.simple_mode == Copter::SimpleMode::SUPERSIMPLE);
```

ArduCopter/AP_Arming_Copter.cpp
```diff
@@ -399,6 +399,31 @@ bool AP_Arming_Copter::gps_checks(bool display_failure)
...
+
+    // return true if position is not required
+    if (!mode_requires_pos) {
```
peterbarker, line 417:
```suggestion
    if (!location_required) {
```

ArduCopter/AP_Arming_Copter.cpp
```diff
@@ -399,6 +399,31 @@ bool AP_Arming_Copter::gps_checks(bool display_failure)
...
+        return true;
+    } else {
+        // call parent visodom checks
+        return AP_Arming::visodom_checks(display_failure);
+    }
```
peterbarker, line 422:
```suggestion
    // call parent visodom checks
    return AP_Arming::visodom_checks(display_failure);
```

Vabe7:
The requested changes should be in now.

I want this because I'm currently building a positioning system for my stampfly to use loiter and so on. But I also want to fly it in Stabilize or Althold if I'm somewhere else, so Guided-NoGPS doesn't really fit the case.

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -1986,7 +1986,7 @@ bool AP_Arming::rc_checks_copter_sub(const bool display_failure, const RC_Channe
 
 #if HAL_VISUALODOM_ENABLED
 // check visual odometry is working
-bool AP_Arming::visodom_checks(bool display_failure) const
+bool AP_Arming::visodom_checks(bool display_failure)
```
peterbarker, line 1989: Why do we need to drop `const` here?

rishabsingh3003:
@peterbarker I actually had a competing PR doing the same thing, https://github.com/ArduPilot/ardupilot/pull/30702 , but I didn't need the PR because ARMING_CHECKS already has a bit to turn off the VISO pre-arm check. SO I ended up closing it.

PR #29265, opened 2025-02-10
Plane: Allow lower speeds in landing final

rubenp02:
# Plane: Allow lower speeds in landing final

## Description
This PR allows for lower airspeeds during the landing final while maintaining appropriate limits during other phases of the flight, including the landing approach phase, if AIRSPEED_STALL is set. This enables smoother/slower landings when a conservative AIRSPEED_MIN is used, which previously resulted in fast touchdowns.

## Changes
- Relaxed target airspeed constraints in navigation.cpp to allow for target airspeeds as low as AIRSPEED_STALL if set. This enables the following changes while maintaining safety, as the TECS further constraints the target airspeed as appropriate.
- Added a method to the landing controller to determine if the aircraft is on landing final.
- Adjusted the TECS airspeed constraints during landing final to allow speeds as low as AIRSPEED_STALL, if set.
- Corrected an inaccuracy in the AIRSPEED_STALL parameter description (this can be moved to a separate PR if required, but I thought it was small enough to include here).
- Updated the AIRSPEED_STALL parameter description to explain its new role during landing.

### DIFF
--- a/libraries/AP_Landing/AP_Landing_Slope.cpp
+++ b/libraries/AP_Landing/AP_Landing_Slope.cpp
@@ -390,6 +390,12 @@ bool AP_Landing::type_slope_is_flaring(void) const
     return (type_slope_stage == SlopeStage::FINAL);
 }
 
+bool AP_Landing::type_slope_is_on_final(void) const
+{
+    return (type_slope_stage == SlopeStage::PREFLARE ||
+            type_slope_stage == SlopeStage::FINAL);
+}
+
 bool AP_Landing::type_slope_is_on_approach(void) const
 {
     return (type_slope_stage == SlopeStage::APPROACH ||
@@ -398,8 +404,7 @@ bool AP_Landing::type_slope_is_on_approach(void) const
 
 bool AP_Landing::type_slope_is_expecting_impact(void) const
 {
-    return (type_slope_stage == SlopeStage::PREFLARE ||
-            type_slope_stage == SlopeStage::FINAL);
+    return type_slope_is_on_final();
 }
 
 bool AP_Landing::type_slope_is_complete(void) const
--- a/libraries/AP_TECS/AP_TECS.cpp
+++ b/libraries/AP_TECS/AP_TECS.cpp
@@ -411,6 +411,10 @@ void AP_TECS::_update_speed(float DT)
     _TASmax   = MIN(_TASmax, aparm.airspeed_max * EAS2TAS);
     _TASmin   = aparm.airspeed_min * EAS2TAS;
 
+    if (_landing.is_on_final() && is_positive(aparm.airspeed_stall)) {
+        _TASmin = aparm.airspeed_stall * EAS2TAS;
+    }
+
     if (aparm.stall_prevention) {
         // when stall prevention is active we raise the minimum
         // airspeed based on aerodynamic load factor
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -305,7 +305,7 @@ const AP_Param::Info Plane::var_info[] = {
 
     // @Param: AIRSPEED_STALL
     // @DisplayName: Stall airspeed
-    // @Description: If stall prevention is enabled this speed is used for its calculations. If set to 0 then AIRSPEED_MIN is used instead. Typically set slightly higher than the true stall speed. Value is as an indicated (calibrated/apparent) airspeed.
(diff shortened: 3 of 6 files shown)


Hwurzburg (changes requested):
Since this is a change in behavior, probably should be a flight option with current behavior as default.

Hwurzburg:
Up to @tridge ...many set airspeed_stall, but default for landing airspeed is 1/2 way between min and cruise....so probably not a big deal...

Hwurzburg (approved):

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -411,6 +411,12 @@ void AP_TECS::_update_speed(float DT)
     _TASmax   = MIN(_TASmax, aparm.airspeed_max * EAS2TAS);
     _TASmin   = aparm.airspeed_min * EAS2TAS;
 
+    if (_landing.is_on_final()) {
+        if (is_positive(aparm.airspeed_stall)) {
```
IamPete1, line 415:
```suggestion
    if (_landing.is_on_final() && is_positive(aparm.airspeed_stall)) {
```

IamPete1:
This seems like a reasonable change. Am I correct that this should only do anything is the user has set stall speed and a landing speed less than min airspeed?

What testing has been done? What happens in the case of a go-around?

libraries/AP_Landing/AP_Landing_Slope.cpp
```diff
@@ -390,6 +390,12 @@ bool AP_Landing::type_slope_is_flaring(void) const
     return (type_slope_stage == SlopeStage::FINAL);
 }
 
+bool AP_Landing::type_slope_is_on_final(void) const
+{
+    return (type_slope_stage == SlopeStage::PREFLARE ||
```
tridge, line 395: in practice this will only impact PREFLARE, as after flare (ie. in FINAL) throttle control is disabled
maybe we should allow lower speed in the approach as long as lined up?

tridge (dismissed):
we'd like to see a log showing an improvement with this change

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1156,7 +1160,7 @@ void AP_TECS::_initialise_states(float hgt_afe)
         _pitch_measured_lpf.reset(_ahrs.get_pitch());
 
     } else if (_flight_stage == AP_FixedWing::FlightStage::TAKEOFF || _flight_stage == AP_FixedWing::FlightStage::ABORT_LANDING) {
-
```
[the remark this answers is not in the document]
IamPete1: Code churn, it makes it harder to track the functional changes. Typically we fixup the formatting when we change the lines for some functional reason.

IamPete1:
image

The target airspeed is set with `LAND_PF_ARSPD`, it is 7.5. The target speed starts going down, although it never gets to the set speed of 7.5 before the flare, the target airspeed is 12.5 or so when the flare starts. Since this is above `AIRSPEED_MIN` of 10, I don't think the changes to the constrains made in this PR make any difference in the example log?

IamPete1:
In the new log we see this:
image

The target speed goes down but is constrained to 10 m/s as set by `AIRSPEED_MIN`. The flare message says the flare start speed is 14.3 m/s. The log with this PR starts flare at 14.1 m/s, so there is no significant difference in speed or the other conditions at the start of the flare.

However, as you say there is a difference in the speed at which the ground is hit. Any idea why this is? Is it reproducible over a number of tests?

IamPete1:
> really seemed to make a difference in real world tests 

If you have a longer pre-flare, it could well make more of a difference. 



> I thought there was an explicit interpolation of demanded airspeeds at this stage but it seems like its governed by some maximum negative acceleration rates. These should probably be made higher for this to work as expected, at least during the preflare stage. If you have reverse thrust, this is very important I think. Thoughts?

Stuff like this is a bit tricky, we can't change or bypass the existing limits because we don't want to start breaking existing aircraft, I guess you could have a new parameter for just this phase. If you can prove it helps I think we would accept it. As ever were trying to trade off more code/flash space/parameters/complexity vs real world benefits for users.

IamPete1 (commented):
The code changes look OK. The SITL logs show a marginal benefit, but I could believe that on a better setup vehicle this could make a larger difference. 

Its been a while so I will rebase.

IamPete1:
> Maybe it could be discussed in dev call to see if anyone wants anything else changed before rebasing and merging?

It's marked for this evenings call.

tridge:
@rubenp02 please rebase

IamPete1 (approved):
Changes look good, with a rebase and CI pass were happy to merge.

peterbarker:
@rubenp02 I think you will need to rebase this again on top of some CI fixes.

I just tried to do that for you but your repository denied the push for whatever reason.

IamPete1:
There should be a "allow edits from maintainers" check box on the PR, usually it's checked by default.

IamPete1 (approved):

IamPete1:
Thanks!

PR #25549, opened 2023-11-15
Plane: add param LAND_RNGFND_HLD to holdd-off lidar during landing

magicrub:
Add param LAND_RNGFND_HLD to holdd-off lidar during landing

`    // @Description: Rangefinder hold-off distance from land point. Hold off using the rangefinder's altitude correction until you're this close to the NAV_LAND point. Requires RNGFND_LANDING = 1. This is useful to avoid the rangefinder reacting to objects/trees, or pits, before the level-ish runway starts. Set to zero to disable the hold-off (don't inhibit the rangefinder). Using this feature will cause the rangefinder to correct the baro offset later in the landing which will lead to a larger slope correction. It is suggested to use LAND_ABORT_DEG with this feature so that unacheivable slopes will cause automatic go-arounds while retaining the rangefinder correction in the baro for a smoother landing on the next attempt.
`

Inspired by Ryan Johnson at Applied Aeronautics

An even better solution would be to add it to the NAV_LAND point but it's pretty maxed out for us :(

PR #29266, opened 2025-02-10
Plane: add rangefinder engagement distance parameter

rubenp02:
# Summary

Introduces the RNGFND_LND_DIST parameter, which defines the horizontal distance to the landing point at which the rangefinder engages when RNGFND_LANDING is enabled.

## Testing (more checks increases chance of being merged)

- [X] Checked by a human programmer
- [X] Tested in SITL
- [X] Tested on hardware
- [ ] Logs attached
- [X] Logs available on request
- [ ] Autotest included

## Description

This is useful for landing on top of platforms or small plateaus, and to avoid interference from uneven terrain or obstacles during the approach.

When set to the default value of 0, the rangefinder engages as soon as it is within range, matching existing behavior.

### Key changes
- Adds RNGFND_LND_DIST parameter to control rangefinder engagement horizontal distance during landing.
- Updates rangefinder logic to engage only when within the specified distance.
- Requires rangefinder_state.in_use for rangefinder correction to be applied.
- Requires both in_use and in_range for rangefinder_active to be true.

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1279,6 +1279,17 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("GUIDED_TIMEOUT", 40, ParametersG2, guided_timeout, 3.0f),
 
+#if AP_RANGEFINDER_ENABLED
+    // @Param: RNGFND_LND_DIST
+    // @DisplayName: Rangefinder landing engagement distance
+    // @Description: The horizontal distance to the landing point at which the rangefinder engages when RNGFND_LANDING is enabled. This is useful for landing on platforms or small plateaus, and to avoid interference from uneven terrain or obstacles on approach. A value of 0 engages the rangefinder as soon as it reports valid readings within its range limits. Very small values are not recommended unless required by the landing scenario, as they can force large slope corrections near the ground, increase auto-abort frequency if LAND_ABORT_DEG is set, and provide no slope-correction benefit if the rangefinder is engaged after flare.
+    // @Range: 0 500
+    // @Units: m
+    // @Increment: 1
+    // @User: Standard
+    AP_GROUPINFO("RNGFND_LND_DIST", 41, ParametersG2, rangefinder_land_engage_dist_m, 0),
+#endif
+
     AP_GROUPEND
 };
 
--- a/ArduPlane/altitude.cpp
+++ b/ArduPlane/altitude.cpp
@@ -686,8 +686,10 @@ float Plane::rangefinder_correction(void)
         return 0;
     }
 
-    // for now we only support the rangefinder for landing 
-    bool using_rangefinder = (rangefinder_use(RangeFinderUse::TAKEOFF_LANDING) && flight_stage == AP_FixedWing::FlightStage::LAND);
+    // for now we only support the rangefinder for landing
+    bool using_rangefinder = (rangefinder_use(RangeFinderUse::TAKEOFF_LANDING) &&
+                              flight_stage == AP_FixedWing::FlightStage::LAND &&
+                              rangefinder_state.in_use);
     if (!using_rangefinder) {
         return 0;
     }
@@ -783,9 +785,18 @@ void Plane::rangefinder_height_update(void)
                 flightstage_good_for_rangefinder_landing = true;
             }
 #endif
+
+            // Check if the aircraft is within RNGFND_LND_DIST meters from the
+            // landing point to engage it
+            const int16_t land_engage_dist_m = g2.rangefinder_land_engage_dist_m;
+            const bool is_within_engagement_distance =
+                (land_engage_dist_m <= 0) ||
+                (auto_state.wp_distance <= land_engage_dist_m);
+
             if (!rangefinder_state.in_use &&
                 flightstage_good_for_rangefinder_landing &&
-                rangefinder_use(RangeFinderUse::TAKEOFF_LANDING)) {
+                rangefinder_use(RangeFinderUse::TAKEOFF_LANDING) &&
+                is_within_engagement_distance) {
                 rangefinder_state.in_use = true;
                 gcs().send_text(MAV_SEVERITY_INFO, "Rangefinder engaged at %.2fm", (double)rangefinder_state.height_estimate);
             }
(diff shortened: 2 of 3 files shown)


IamPete1:
What is the use case? I presume to ignore uneven terrain prior to the runway? Would a horizontal distance be more intuitive? I recall we may have had other PRs in this area in the past that may have some useful idea in if we can find them.

LupusTheCanine:
> I did think about using horizontal distance, which has the advantage that it's slope angle agnostic, but I think using a vertical activation distance lends itself better to that last point. Maybe have both? This is is still far from done (doesn't even work) so I'm very open to any suggestions.

IMHO horizontal distance will yield more predictable behaviour. Approaching over a ledge, trees or bushes may cause premature activation of the rangefinder. 

It would be nice if we could trigger abort on unexpectedly low altitude even outside this distance. Though there are a few edge cases that make that complicated:

1. baro drift such that low altitude protection kicks in before we reach use radius.
2. obstacle on path (we can't retry the same path as most obstacles rarely move), mostly an issue for failsafe.
3. baro drift such that we reach flare height/time above landing point without engaging rangefinder (mostly an issue for short (<10m) range rangefinders) and is an issue now too.

Cases 1. and 2. can't be reliably distinguished without overflying the landing point and detecting the ground there.

PS I tested short range rangefinders for landing so I would like to be able to have them keep working. I expect I could get decent landings with 10m  rangefinder.

PSS maybe continue descent after not detecting obstacle underneath for 500-1000ms and abort over landing point after hopefully detecting the ground.

magicrub:
Here's my implementation in PR https://github.com/ArduPilot/ardupilot/pull/25549/files

rubenp02:
Finally gotten around to rebase and update this. I think it's ready, been in our production firmware for ages and it's been working just fine.

ArduPlane/Parameters.cpp
```diff
@@ -1279,6 +1279,17 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
...
 
+#if AP_RANGEFINDER_ENABLED
+    // @Param: RNGFND_LND_DIST
+    // @DisplayName: Rangefinder landing engagement distance
+    // @Description: The horizontal distance to the landing point at which the rangefinder engages if RNGFND_LANDING is enabled. This is useful for landing on top of platforms or small plateaus, and to avoid interference from uneven terrain or obstacles along the flight path. The default value of 0 enables the rangefinder as soon as it is in range.
+    // @Range: 0 250
+    // @Units: m
+    // @Increment: 1
+    // @User: Standard
+    AP_GROUPINFO("RNGFND_LND_DIST", 41, ParametersG2, rangefinder_land_engagement_distance, 0),
```
magicrub: Since this is not a float, please rename the variable to have a _m on the end for meters. Many integer units are cm (or worse!) so best be explicit with it.

ArduPlane/Parameters.cpp
```diff
@@ -1279,6 +1279,17 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("GUIDED_TIMEOUT", 40, ParametersG2, guided_timeout, 3.0f),
 
+#if AP_RANGEFINDER_ENABLED
+    // @Param: RNGFND_LND_DIST
+    // @DisplayName: Rangefinder landing engagement distance
+    // @Description: The horizontal distance to the landing point at which the rangefinder engages if RNGFND_LANDING is enabled. This is useful for landing on top of platforms or small plateaus, and to avoid interference from uneven terrain or obstacles along the flight path. The default value of 0 enables the rangefinder as soon as it is in range.
+    // @Range: 0 250
```
magicrub: It's an int16 so this limit could be up to 32k but that's obviously not practical. 250 feels a little short for larger aircraft so lets bump up the suggested max to 500.

ArduPlane/altitude.cpp
```diff
@@ -783,9 +785,17 @@ void Plane::rangefinder_height_update(void)
                 flightstage_good_for_rangefinder_landing = true;
             }
 #endif
+
+            // Check if the aircraft is within RNGFND_LND_DIST meters from the
+            // landing point to engage it
+            const bool is_within_engagement_distance =
+                (g2.rangefinder_land_engagement_distance <= 0) ||
+                (auto_state.wp_distance <= g2.rangefinder_land_engagement_distance);
```
magicrub: This is a fetch frm the param library that may induce additional work in some cases so better to only fetch it once:

```suggestion
            const int16_t land_engagement_distance_m = g2.rangefinder_land_engagement_distance.get();
            const bool is_within_engagement_distance =
                (land_engagement_distance_m <= 0) ||
                (auto_state.wp_distance <= land_engagement_distance_m);
```

ArduPlane/Parameters.cpp
magicrub: >The default value of 0 enables the rangefinder as soon as it is in range

I can see how someone could be confused with the "as soon as it's in range" terminology because we're talking about a range within the scope of this description. Should say something like vertical range or something to distinguish it.

I also feel like need a warning or at least mention that a value that is short is not recommended unless you have to in the given landing scenario. Upon a ranegfinder "engagement" it will fix/restore/recalibrate its glide slope to better match hitting the land point. Doing that to close to the land location can make things worse.

magicrub (changes requested):

ArduPlane/Parameters.cpp
rubenp02: Very good point. Will take the chance to sorten the variable name a bit.

rubenp02:
> Autotest would be good. Trying to land a plane on top of the KalaupapaCliffs could be interesting.

I've been trying to do this but the terrain data is too smooth to properly test this in a way that's repeatable and results in a clear test pass or fail. Ideally we want a "step" in terrain altitude where the low part is still within LiDAR range so that without this parameter, the aircraft starts to sink below the correct slope and lands very short and hard. I've looked around `KalaupapaCliffs` and `JuanDeLaCierva_*` but the DTM res is just too low.

If it's possible to inject fake terrain into the simulation, then writing an autotest for this is fairly trivial. If not, one can certainly write something that behaves differently depending on the value of this param., but I don't think it'll clearly show the improvement it makes with the large and detailed terrain changes you see IRL.

peterbarker:
> > Autotest would be good. Trying to land a plane on top of the KalaupapaCliffs could be interesting.
> 
> I've been trying to do this but the terrain data is too smooth to properly test this in a way that's repeatable and results in a clear test pass or fail. Ideally we want a "step" in terrain altitude where the low part is still within LiDAR range so that without this parameter, the aircraft starts to sink below the correct slope and lands very short and hard. I've looked around `KalaupapaCliffs` and `JuanDeLaCierva_*` but the DTM res is just too low.
> 
> If it's possible to inject fake terrain into the simulation, then writing an autotest for this is fairly trivial. If not, one can certainly write something that behaves differently depending on the value of this param., but I don't think it'll clearly show the improvement it makes with the large and detailed terrain changes you see IRL.

OK, so let's put that down to, "would be nice to have, but way too hard".

Thanks for looking!

rubenp02:
> I think this basically looks good, just need to get it past CI.

Done!

IamPete1:
@magicrub Can you take another look please.

magicrub (approved):
LGTM, thank you for this! I'll close my PR

PR #29269, opened 2025-02-10
AP_HAL_ESP32: fix up I2C and SPI port drivers

tpwrules:
Lessons learned from the stampfly to make ESP32 more similar to ChibiOS. Please see commits for details. These have been tested on there for a while now.


davidbuzz (approved):
reviewed, this is good to gbe getting us closer to chibios.

srmainwaring (commented):
Thanks @tpwrules - good to have these in master. Have been running these on the stampfly branch and also on a rover build using the same FC hardware.

PR #29274, opened 2025-02-11
Locosys rtk dual

rliou2000:


libraries/AP_GPS/AP_GPS_UBLOX.cpp
```diff
@@ -20,6 +20,8 @@
 //
 #include "AP_GPS_UBLOX.h"
 
+// Test CI CD 0212 #define SUPPORT_LX_RTK_DUAL_PATCH       20240314_20250211_unhealth_gps_patch_branch
+
```
peterbarker, line 24: Doesn't belong here

libraries/AP_GPS/AP_GPS_UBLOX.cpp
```diff
@@ -2279,6 +2284,21 @@ bool AP_GPS_UBLOX::is_healthy(void) const
...
 #if GPS_MOVING_BASELINE
+    
+#ifdef  SUPPORT_LX_RTK_DUAL_PATCH             
+ if (gps._auto_config == AP_GPS::GPS_AUTO_CONFIG_DISABLE) { 
+     // since  GPS_AUTO_CONFIG =0    //  _hardware_generation is not available 
+        return true;                                       
+    }       
+else if ((role == AP_GPS::GPS_ROLE_MB_BASE ||
+           role == AP_GPS::GPS_ROLE_MB_ROVER) 
+          && _hardware_generation ==  LX_RTK_DUAL)
```
peterbarker, line 2295: LX_RTL_DUAL doesn't exist as a symbol, so there's no way this even compiles.

peterbarker (changes requested):
I don't know what you're trying to achieve with this set of patches.

Bare in mind we already support dual-antenna gps-for-yaw in ArduPilot.  Whatever patches you make should generally look like the patches for the existing support.

Please explain what you're trying to accomplish here and why.

rliou2000:
This is a quick patch  to address "unhealthy GPS" messages found in LOCOSYS RTK_Dual module in gps-for-yaw mode.Since we have addressed the issue by updating the GNSS module firmware  to make it compatible with ublox gps-for-yaw mode.This patch is obsoleted.
BR,Roger
    在 2025年3月11日 星期二 上午06:11:21 [GMT+8]， Peter ***@***.***> 寫道：  
 
 


@peterbarker requested changes on this pull request.

I don't know what you're trying to achieve with this set of patches.

Bare in mind we already support dual-antenna gps-for-yaw in ArduPilot. Whatever patches you make should generally look like the patches for the existing support.

Please explain what you're trying to accomplish here and why.

In libraries/AP_GPS/AP_GPS_UBLOX.cpp:
> +// Test CI CD 0212 #define SUPPORT_LX_RTK_DUAL_PATCH       20240314_20250211_unhealth_gps_patch_branch
+

Doesn't belong here

In libraries/AP_GPS/AP_GPS_UBLOX.cpp:
> @@ -2279,6 +2284,21 @@ bool AP_GPS_UBLOX::is_healthy(void) const
     }
 #endif
 #if GPS_MOVING_BASELINE
+    
+#ifdef  SUPPORT_LX_RTK_DUAL_PATCH             
+ if (gps._auto_config == AP_GPS::GPS_AUTO_CONFIG_DISABLE) { 
+     // since  GPS_AUTO_CONFIG =0    //  _hardware_generation is not available 
+        return true;                                       
+    }       
+else if ((role == AP_GPS::GPS_ROLE_MB_BASE ||
+           role == AP_GPS::GPS_ROLE_MB_ROVER) 
+          && _hardware_generation ==  LX_RTK_DUAL)

LX_RTL_DUAL doesn't exist as a symbol, so there's no way this even compiles.

peterbarker:
> This is a quick patch  to address "unhealthy GPS" messages found in LOCOSYS RTK_Dual module in gps-for-yaw mode.Since we have addressed the issue by updating the GNSS module firmware  to make it compatible with ublox gps-for-yaw mode.This patch is obsoleted.

Thanks for getting back to us on this.  Closing this PR now.

Issue #29272, opened 2025-02-11
SlewCalculator2D::get_slew_rate() does not use ylimiter


gorin-s:
## Bug report

**Issue details**

I was checking the code in order to understand the details of Guided mode, and when I got to SlewCalculator2D I found that get_slew_rate uses xlimiter twice in it's calculation, instead of ylimiter

`safe_sqrt(sq(xlimiter.get_slew_rate()) + sq(**xlimiter**.get_slew_rate()))`

**Version**
4.5.7

**Platform**
[  ] All
[  ] AntennaTracker
[ x ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
Quad

**Hardware type**
SITL

**Logs**
Just observing the code...

rmackay9:
Yes, thanks!

PR #23981, opened 2023-06-06
Rover: quik tune lua script

rmackay9:
This implements a "QuickTune" for Rovers and Boats based on the VTOL-quick tune method.  This script is meant to be run with the vehicle in Circle mode running above 20% throttle and turning at at least 10 deg/sec.

The axes tuned are:

- Steering Turn Rate controller (ATC_STR_RAT_P, ATC_STR_RAT_I, ATC_STR_RAT_D, ATC_STR_RAT_FF)
- Speed/Throttle controller (ATC_SPEED_P, ATC_SPEED_I, ATC_SPEED_D, CRUISE_SPEED, CRUISE_THROTTLE)
- Position controller's velocity controller (PSC_VEL_P, PSC_VEL_I, PSC_VEL_D)

The tuning method is basically the same as the VTOL-quick tune in that first D is tuned, then P, then I may be set to the P value (if FF isn't set) then finally FF is tuned (and I is set again to be the same as the FF value).  In the case of D and P, the Slew Rate provided by AC_PID is used to determine when the controllers begin oscillating (se RTUN_OSC_SMAX parameter) after which the tune completes and the gains are reduced according to the RTUN_GAIN_MARGIN parameter.

The feedforward (which VTOL-quicktune does not do) is calculated by capturing the appropriate input and outputs over one rotation of the circle.  So in the case of the steering turn rate control (ATC_STR_RAT_FF) the steering output and the vehicle's actual turn rate are recorded.  The FF is then the sum of the steering / sum of the turn rates.

For speed feedforward (CRUISE_SPEED, CRUISE_THROTTLE) throttle output and actual speed samples are collected.  CRUISE_SPEED is then set to the average speed (e.g. sum of the speed samples / number of samples) and CRUISE_THROTTLE is set to the average throttle (e.g. sum of the throttle samples / number of samples).

Some other differences vs the VTOL-quicktune:

- SlewLimiter2D has been added based on the regular 1D SlewLimiter class.  This seems to work but we need to test that setting PSC_VEL_SRMAX does correctly stop oscillations.
- Only the steering turn rate filters (e.g. ATC_STR_RAT_FLTT, FLTD) are set from the INS_GYRO_FILTER.  I was unsure what to use for the speed and position controller velocity controllers.  Accelerometer?
- GCS_PID_MASK is set automatically so the user can more easily monitor progress.  It's original value is restored at the end of the tune
[...]

PR #29275, opened 2025-02-11
Filter: SlewCalculator2D typo fix

rmackay9:
This fixes a typo in the 2D Slew Calculator used by the AC_PID_2D class which is used in Copter, Rover and Blimp's XY velocity controller.  This means it's also used by QuadPlanes in QLoiter, etc but surely it has nothing to do with the recent questions around quad plane attitude control because the bug was in 4.5 as well.

This was added in PR https://github.com/ArduPilot/ardupilot/pull/23981 but it is only used for reporting so I think this does not cause any actual performance issue.  In fact, since this PR went in we've simplified the Rover autotune and we don't use this is anymore so we could potentially backout the slew reporting completely and save some flash and a very small amount of CPU

This resolves issue https://github.com/ArduPilot/ardupilot/issues/29272

I've done some minor testing in copter in SITL and there is a difference in the logged slew rate.  For example in the "before" we see consistently low slew rate while in "after" it is much much larger.
srate-before-vs-after

I think this is correct though because if I greatly reduce the PSC_VELXY_P and I values I can produce similarly huge numbers with master
image

### DIFF
--- a/libraries/Filter/SlewCalculator2D.cpp
+++ b/libraries/Filter/SlewCalculator2D.cpp
@@ -39,5 +39,5 @@ void SlewCalculator2D::update(const Vector2f& sample, float dt)
 // get last oscillation slew rate
 float SlewCalculator2D::get_slew_rate() const
 {
-    return safe_sqrt(sq(xlimiter.get_slew_rate()) + sq(xlimiter.get_slew_rate()));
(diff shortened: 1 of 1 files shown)


IamPete1 (approved):

MattKear:
Does anyone actually use the slew rate limiter in the position controller?
My vote would be to remove it.

Georacer (approved):

rmackay9:
I've merged this but I'll likely raise another PR to remove the feature completely

PR #29276, opened 2025-02-11
ArduPlane:add AUTOLAND metadata to FS_LONG_ATCN

Hwurzburg:
duh....missed this also when I added AUTOLAND  mode....NFC as far as AP is concerned but needed by GCS


peterbarker (approved):
LGTM

PR #29278, opened 2025-02-11
AP_Frsky_D: Enhanced, add power, time, temperature, acceleration etc

Paku-:
Some enhancements to the very old FrSky D telemetry.

- Few datagrams added (time, power, battery per cell, barometric altitude etc)
- Transmission divided into 3 frames as per documentation
- User can define periods for every frame types to balance transmission payload.
- Some comments on how to manipulate data, as each solution can vary.

### DIFF
--- a/libraries/AP_Frsky_Telem/AP_Frsky_Backend.cpp
+++ b/libraries/AP_Frsky_Telem/AP_Frsky_Backend.cpp
@@ -3,9 +3,15 @@
 #if AP_FRSKY_TELEM_ENABLED
 
 #include <AP_Baro/AP_Baro.h>
+#include <AP_BattMonitor/AP_BattMonitor.h>
+#include <AP_GPS/AP_GPS.h>
 #include <AP_AHRS/AP_AHRS.h>
 #include <AP_RPM/AP_RPM.h>
+#include <AP_Vehicle/AP_Vehicle.h>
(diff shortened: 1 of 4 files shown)


IamPete1:
There is quite a bit of commented out code added that should be removed. 

What, if any, changes are needed on the transmitter side to receive the new messages? Will this break existing setups?

Paku-:
> There is quite a bit of commented out code added that should be removed.
> 
> What, if any, changes are needed on the transmitter side to receive the new messages? Will this break existing setups?

That's good question. I tested it with original screen:

https://www.bing.com/images/search?q=frsky+fld+02&form=HDRSC3&first=1

it works, there is one issue with blinking pack cumulative voltage but could be a screen issue.

I believe messages are read not by frame but by field in the frame. If we add some info it will be read or just discarded.

3 frames construction is a direct copy of original FrSky docs.

But the true is, it's very old code. I work with/on it last time around 2012-2013 :)

My experience is you can mix fields as you want. It may be a good idea to test it with any other "screen".

Hwurzburg:
this is not worth 500 bytes iMO, at most a non defaulted build option

rmackay9:
small thing, the commit title should be prefixed with "AP_Frsky_Telem: "

peterbarker:
Let's make the entire D protocol off by default.  Then we can add this feature whenever you enable the "D" protocol in ArduPilot.

PR #29280, opened 2025-02-12
Add Autoland fence action

Hwurzburg:
And again.....

### DIFF
--- a/ArduPlane/fence.cpp
+++ b/ArduPlane/fence.cpp
@@ -42,6 +42,7 @@ void Plane::fence_check()
             case AC_FENCE_ACTION_GUIDED:
             case AC_FENCE_ACTION_GUIDED_THROTTLE_PASS:
             case AC_FENCE_ACTION_RTL_AND_LAND:
+            case AC_FENCE_ACTION_AUTOLAND_OR_RTL:
                 if (plane.control_mode_reason == ModeReason::FENCE_BREACHED &&
                     control_mode->is_guided_mode()) {
                     set_mode(*previous_mode, ModeReason::FENCE_RETURN_PREVIOUS_MODE);
@@ -81,6 +82,7 @@ void Plane::fence_check()
         case AC_FENCE_ACTION_REPORT_ONLY:
             break;
 
+        case AC_FENCE_ACTION_AUTOLAND_OR_RTL:
         case AC_FENCE_ACTION_RTL_AND_LAND:
             if (control_mode == &mode_auto &&
                 mission.get_in_landing_sequence_flag() &&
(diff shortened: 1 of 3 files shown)


ArduPlane/fence.cpp
```diff
@@ -80,6 +81,13 @@ void Plane::fence_check()
         switch (fence_act) {
         case AC_FENCE_ACTION_REPORT_ONLY:
             break;
+#if MODE_AUTOLAND_ENABLED
+        case AC_FENCE_ACTION_AUTOLAND_OR_RTL:
```
IamPete1, line 85: I think you need the same check as on line 95. Probably easier to fall through to that case and then add the if on 94.

IamPete1: You also need to protect against RLT vs AUTOLAND. It should not switch out of one to go into the other.

Hwurzburg: is this what you mean?

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -66,7 +66,7 @@ const AP_Param::GroupInfo AC_Fence::var_info[] = {
     // @Description: What action should be taken when fence is breached
     // @Values{Copter}: 0:Report Only,1:RTL or Land,2:Always Land,3:SmartRTL or RTL or Land,4:Brake or Land,5:SmartRTL or Land
     // @Values{Rover}: 0:Report Only,1:RTL or Hold,2:Hold,3:SmartRTL or RTL or Hold,4:SmartRTL or Hold
-    // @Values{Plane}: 0:Report Only,1:RTL,6:Guided,7:GuidedThrottlePass
+    // @Values{Plane}: 0:Report Only,1:RTL,6:Guided,7:GuidedThrottlePass,8:AUTOLAND if possible or RTL
```
IamPete1, line 69:
```suggestion
    // @Values{Plane}: 0:Report Only,1:RTL,6:Guided,7:GuidedThrottlePass,8:AUTOLAND if possible else RTL
```
Maybe that is a little clearer that RTL is the backup plan.

ArduPlane/fence.cpp
```diff
@@ -80,17 +81,30 @@ void Plane::fence_check()
...
+#if MODE_AUTOLAND_ENABLED
+|| fence_act == AC_FENCE_ACTION_AUTOLAND_OR_RTL
+#endif
+                                               ) {
                 if (control_mode == &mode_auto &&
                     mission.get_in_landing_sequence_flag() &&
                     (g.rtl_autoland == RtlAutoland::RTL_THEN_DO_LAND_START ||
                      g.rtl_autoland == RtlAutoland::RTL_IMMEDIATE_DO_LAND_START)) {
                     // already landing
                     return;
```
IamPete1, line 101: I think this early return needs a check for being in either RTL or Autoland.

Hwurzburg: I have added a check to not change modes on a breach when AUTOLANDING in is LAND stage

libraries/AC_Fence/AC_Fence.cpp
Hwurzburg: done

tridge (approved):

PR #29281, opened 2025-02-12
Add Autoland batt fs action

Hwurzburg:
Last one....for today...just as a low action....to save some flash...easy add to crt if we think its needed

### DIFF
--- a/ArduPlane/events.cpp
+++ b/ArduPlane/events.cpp
@@ -310,6 +310,11 @@ void Plane::handle_battery_failsafe(const char *type_str, const int8_t action)
                 already_landing = true;
             }
 #endif
+#if MODE_AUTOLAND_ENABLED
+            if (control_mode == &mode_autoland) {
+                already_landing = true;
+            }
+#endif
             if (!already_landing) {
                 // never stop a landing if we were already committed
                 if ((g.rtl_autoland == RtlAutoland::RTL_IMMEDIATE_DO_LAND_START) && plane.have_position && plane.mission.is_best_land_sequence(plane.current_loc)) {
--- a/ArduPlane/events.cpp
+++ b/ArduPlane/events.cpp
@@ -301,7 +301,11 @@ void Plane::handle_battery_failsafe(const char *type_str, const int8_t action)
             }
             FALLTHROUGH;
         }
-        case Failsafe_Action_RTL: {
+        case Failsafe_Action_RTL:
+#if MODE_AUTOLAND_ENABLED
+        case Failsafe_Action_AUTOLAND_OR_RTL:
+#endif
+                                              {
             bool already_landing = flight_stage == AP_FixedWing::FlightStage::LAND;
 #if HAL_QUADPLANE_ENABLED
             if (control_mode == &mode_qland || control_mode == &mode_loiter_qland ||
@@ -316,8 +320,14 @@ void Plane::handle_battery_failsafe(const char *type_str, const int8_t action)
                     plane.mission.set_in_landing_sequence_flag(true);
                     break;
                 }
-                set_mode(mode_rtl, ModeReason::BATTERY_FAILSAFE);
                 aparm.throttle_cruise.load();
+#if MODE_AUTOLAND_ENABLED
+             if (((Failsafe_Action)action == Failsafe_Action_AUTOLAND_OR_RTL) &&
+ set_mode(mode_autoland, ModeReason::BATTERY_FAILSAFE)) {              
+                break;
+            }
+#endif
+                set_mode(mode_rtl, ModeReason::BATTERY_FAILSAFE);
             }
             break;
         }
--- a/ArduPlane/events.cpp
+++ b/ArduPlane/events.cpp
@@ -302,10 +302,7 @@ void Plane::handle_battery_failsafe(const char *type_str, const int8_t action)
             FALLTHROUGH;
         }
         case Failsafe_Action_RTL:
-#if MODE_AUTOLAND_ENABLED
-        case Failsafe_Action_AUTOLAND_OR_RTL:
-#endif
-                                              {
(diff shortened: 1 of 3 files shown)


ArduPlane/events.cpp
```diff
@@ -316,6 +320,14 @@ void Plane::handle_battery_failsafe(const char *type_str, const int8_t action)
                     plane.mission.set_in_landing_sequence_flag(true);
                     break;
                 }
+#if MODE_AUTOLAND_ENABLED
+            if ((Failsafe_Action)action == Failsafe_Action_AUTOLAND_OR_RTL) {
+                if (!set_mode(mode_autoland, ModeReason::BATTERY_FAILSAFE)) {
+                    set_mode(mode_rtl, ModeReason::BATTERY_FAILSAFE);
+                }
+            break;
```
IamPete1, line 328: It will fall through to RTL in anycase. So you can combine the ifs.
```suggestion
            if (((Failsafe_Action)action == Failsafe_Action_AUTOLAND_OR_RTL) && set_mode(mode_autoland, ModeReason::BATTERY_FAILSAFE)) {              
                break;
```

ArduPlane/mode_autoland.cpp
```diff
@@ -175,7 +175,7 @@ bool ModeAutoLand::_enter()
     entry_alt = plane.current_loc.alt;
     
     // start first leg toward the base leg loiter to alt point
-    if (climb_first) {
+    if (climb_first && plane.control_mode_reason != ModeReason::BATTERY_FAILSAFE) {
```
IamPete1, line 178: I'm not sure about this. I would rather run of of battery on the way back than skip the climb and run into the ground. Better to keep the behavior consistent to avoid confusion.

libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
```diff
@@ -112,7 +112,7 @@ const AP_Param::GroupInfo AP_BattMonitor_Params::var_info[] = {
     // @Param: FS_LOW_ACT
     // @DisplayName: Low battery failsafe action
     // @Description: What action the vehicle should perform if it hits a low battery failsafe
-    // @Values{Plane}: 0:None,1:RTL,2:Land,3:Terminate,4:QLand,6:Loiter to QLand
+    // @Values{Plane}: 0:None,1:RTL,2:Land,3:Terminate,4:QLand,5:Parachute release,6:Loiter to QLand,7:AUTOLAND or RTL
```
IamPete1, line 115: Also need to update `FS_CRT_ACT` Parachute is enumerated there already.

IamPete1 (changes requested):

libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
Hwurzburg: did not feel the real need to add this to crt fs....

ArduPlane/mode_autoland.cpp
Hwurzburg: not sure I agree on this

ArduPlane/events.cpp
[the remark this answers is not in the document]
Hwurzburg: fixed

Hwurzburg: changed

ArduPlane/events.cpp
```diff
@@ -316,6 +320,14 @@ void Plane::handle_battery_failsafe(const char *type_str, const int8_t action)
...
+            }
+#endif
                 set_mode(mode_rtl, ModeReason::BATTERY_FAILSAFE);
                 aparm.throttle_cruise.load();
```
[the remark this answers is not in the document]
Hwurzburg: moved

libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
IamPete1: The value will still work whether or not you add it to the description.

ArduPlane/mode_autoland.cpp
```diff
@@ -175,7 +175,11 @@ bool ModeAutoLand::_enter()
     entry_alt = plane.current_loc.alt;
     
     // start first leg toward the base leg loiter to alt point
-    if (climb_first) {
+    if (climb_first && plane.control_mode_reason != ModeReason::BATTERY_FAILSAFE && !(plane.fence.get_breaches()
+#if AP_FENCE_ENABLED
+ & AC_FENCE_TYPE_ALT_MAX)
+#endif
```
IamPete1, line 181: I disagree with this, better to keep behaviour consistent. Mode should behave the same no matter how you enter them.

tridge: I agree with @IamPete1, it should be consistent

Hwurzburg: fixed

libraries/AP_BattMonitor/AP_BattMonitor_Params.cpp
Hwurzburg: done

ArduPlane/events.cpp
```diff
@@ -301,7 +301,11 @@ void Plane::handle_battery_failsafe(const char *type_str, const int8_t action)
             }
             FALLTHROUGH;
         }
-        case Failsafe_Action_RTL: {
+        case Failsafe_Action_RTL:
+#if MODE_AUTOLAND_ENABLED
```
IamPete1, line 305: I wounder if its correct to define out the action here. If someone sets it then we should RTL if auto land is not compiled in.

IamPete1: I have pushed a commit that makes that change.

tridge (approved):

PR #29285, opened 2025-02-13
AC_AttitudeControl: AC_PosControl: do not constrain stopping distance velocity

peterbarker:
constraining our stopping distance against our configured velocity makes no sense; we may be travelling *significantly* faster than our wpnav velocity when asked to calculate a stopping point.

The constraint was added in this commit:
```
commit 92e05e8c4f88335abac014f2d29856297b105656
Author: Leonard Hall <leonardthall@gmail.com>
Date:   Sat Apr 17 01:23:14 2021 +0930

    AC_AttitudeControl: PosControl fixes
```

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -15038,7 +15038,6 @@ return update, 1000
             "CompassMot": "Causes an arithmetic exception in the EKF",
             "SMART_RTL_EnterLeave": "Causes a panic",
             "SMART_RTL_Repeat": "Currently fails due to issue with loop detection",
-            "RTLStoppingDistanceSpeed": "Currently fails due to vehicle going off-course",
         }
 
 


lthall (commented):
The original reason this was added was to limit how far the aircraft would deliberatly travel if traveling much faster than the vel max.

At the time some members of the dev team were conserned that the aircraft would take much longer to stop than the operator expected.

The default acceleration is 1m/ss. Traveling at 20m/s gives us a stopping distance of 200m.

This limits the maximum stopping distance to what is normally seen in RTL or Auto.

rmackay9:
I agree with @lthall's comments.  I wonder if it would help to change the constraint so that it is the maximum of the RTL_SPEED or WPNAV_SPEED.  Maybe @peterbarker saw this issue in a report where the RTL_SPEED was set very low.

PR #29286, opened 2025-02-13
AP_BLHeli: fixed documentation for SERVO_BLH_3DMASK and SERVO_BLH_RVMASK

tridge:
they can be combined on both BLHeli32 and AM32

### DIFF
--- a/libraries/AP_BLHeli/AP_BLHeli.cpp
+++ b/libraries/AP_BLHeli/AP_BLHeli.cpp
@@ -128,7 +128,7 @@ const AP_Param::GroupInfo AP_BLHeli::var_info[] = {
 
     // @Param: 3DMASK
     // @DisplayName: BLHeli bitmask of 3D channels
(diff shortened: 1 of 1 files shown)


rmackay9:
Not complaining of course but just so people are aware, my take is that this means there are two ways to change the direction for AM32 DroneCAN ESCs

- set each ESC's DIR_REVERSED parameter (as mentioned on the [AP AM32 DroneCAN wiki page](https://ardupilot.org/copter/docs/common-am32-escs.html))
- set the SERVO_BLH_RVMASK parameter

tridge:
> > Not complaining of course but just so people are aware, my take is that this means there are two ways to change the direction for AM32 DroneCAN ESCs
> 
> The same is true of BLHeli32 ESCs.

Ahh, sorry, you said DroneCAN. The SERVO_BLH_RVMASK parameter doesn't apply to DroneCAN

PR #29290, opened 2025-02-13
Mark many functions for storage in external flash

peterbarker:
Added new __INITFUNC__ for marking processes, inferring they can be put into external flash segments.

This walks CubeRedPrimary back from the edge a bit.


libraries/AP_Declination/tables.cpp
```diff
@@ -9,7 +9,7 @@ const float AP_Declination::SAMPLING_MAX_LAT = 90;
 const float AP_Declination::SAMPLING_MIN_LON = -180;
 const float AP_Declination::SAMPLING_MAX_LON = 180;
 
-const float AP_Declination::declination_table[LAT_TABLE_SIZE][LON_TABLE_SIZE] = {
+__EXTFLASHFUNC__ const float AP_Declination::declination_table[LAT_TABLE_SIZE][LON_TABLE_SIZE] = {
```
andyp1per, line 12: Typicall data tables might go in a different area - do we know these are only accessed at startup?

libraries/AP_Declination/generate/generate.py
```diff
@@ -24,7 +24,7 @@
 
 def write_table(f,name, table):
     '''write one table'''
-    f.write("const float AP_Declination::%s[%u][%u] = {\n" %
+    f.write("__EXTFLASHFUNC__ const float AP_Declination::%s[%u][%u] = {\n" %
```
andyp1per, line 27: This is different to the PR title

andyp1per (commented):
There's some non-init stuff going on here - would be good to clarify or put in a different PR

peterbarker:
> There's some non-init stuff going on here - would be good to clarify or put in a different PR

Yeah, I was feeling bad about that.  But I wasn't feeling bad enough about it to pull the patches out separately.  Will do so now :-)

libraries/AP_Declination/generate/generate.py
peterbarker: Fixed

libraries/AP_Declination/tables.cpp
peterbarker: These aren't just used at startup.

But I believe they're used at very low rates - of the order of 1 or 2Hz.

And probably quite rapidly while doing mag field calibrations.

peterbarker:
I've flashed this onto a CubeRed, and it seems to continue to function.

I particularly want this for my PROGRAM_FLASH_SIZE PR.

peterbarker:
> There's some non-init stuff going on here - would be good to clarify or put in a different PR

I've fixed the PR title and the commit message on the Declination changes

libraries/AP_Declination/tables.cpp
andyp1per: I mean its probably ok, but I just can't say for sure - it depends what the usage looks like

libraries/AP_HAL/AP_HAL_Boards.h
```diff
@@ -378,6 +378,10 @@
 #define __EXTFLASHFUNC__
 #endif
 
+#ifndef __INITFUNC__
```
rmackay9, line 381: Could we add a comment here to help future developers when they can or cannot use this function?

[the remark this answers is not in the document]
tpwrules: Please also move this commit to the front so half of it doesn't fail to compile.

tridge (approved):

libraries/AP_HAL/AP_HAL_Boards.h
peterbarker: Done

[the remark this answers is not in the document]
peterbarker: > Please also move this commit to the front so half of it doesn't fail to compile.

Done, thanks!

PR #29291, opened 2025-02-13
Tools: AP_Periph: SIM: servo telem move to one indexed

IamPete1:
This matches how Hitec servos report. I don't have any other types to test.

### DIFF
--- a/Tools/AP_Periph/rc_out.cpp
+++ b/Tools/AP_Periph/rc_out.cpp
@@ -206,7 +206,7 @@ void AP_Periph_FW::sim_update_actuator(uint8_t actuator_id)
         }
         const SRV_Channel::Function function = SRV_Channel::Function(SRV_Channel::k_rcin1 + i);
         uavcan_equipment_actuator_Status pkt {};
-        pkt.actuator_id = i;
+        pkt.actuator_id = i + 1;
         // assume 45 degree angle for simulation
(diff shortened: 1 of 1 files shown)


tpwrules:
Looks like this should match the + 1 here, so the change is correct and I think your expectation is not: https://github.com/ArduPilot/ardupilot/blob/15376a4908cc1a0019f692b495085688736f84e0/libraries/AP_DroneCAN/AP_DroneCAN.cpp#L766

Although the DSDL spec implies an actuator ID of 0 is perfectly legal, ArduPilot seems to always send starting at 1 presumably to match the channel indices. What ID is the real servo programmed with in its parameters?

IamPete1:
> What ID is the real servo programmed with in its parameters?

I don't know, I only have a AP log. I found this was needed to set simulated DroneCAN servos to match the real ones.

I would have expected to start at 0 because that is what DroneCAN ESCs do, as you say maybe I was just expecting the wrong thing.

tpwrules:
Yeah it looks like ESCs are defined to start from 0. They don't have quite the same ID system and are in a different namespace anyway.

The rest of the periph code assumes actuators start at 1 including earlier in that same function. So I think this is a simple bug fix. ArduPilot could change how it numbers servos but then everybody would have to reconfigure their vehicles, assuming it is even the right thing.

tpwrules (approved):

tridge (approved):

PR #29295, opened 2025-02-14
AP_Filesystem: littlefs: fix lseek

tpwrules:
lseek must return the current file position. Previously, the littlefs version always returned 0, which broke terrain I/O as it checks that the position returned is the one it seeked to. Fix to return the current position, which is correctly returned from littlefs.

This problem was originally and incorrectly diagnosed as an issue with littlefs seeking past the end of the file, but this functionality works fine and fixing the incorrect return completely fixes terrain.

Terrain functionality was verified using `TERRAIN_DEBUG` on KakuteH7Mini-Nand running sim on HW. Terrain data is correctly downloaded from the GCS and loaded from the filesystem after reboot.


andyp1per (approved):
Good catch!

PR #29296, opened 2025-02-14
Backport CADDX support to 4.6 beta

peterbarker:
Needed to resolve a compilation problem around RC_targetting code.

Only tested for compilation as I don't have one of these units.


rmackay9:
@peterbarker, thanks very much for this!  I'll test this on monday

rmackay9:
I've tested this and it works as well as master does :-).  which is to say that RC rate control never worked so I will create a PR to resolve that shortly

PR #28900, opened 2024-12-18
AP_Networking: fix memory error generated by CAN mcast driver

bugobliterator:

PR #29297, opened 2025-02-14
[Backport 4.6] Create separate option for bridging mcast and hw can

bugobliterator:
backport of this PR https://github.com/ArduPilot/ardupilot/pull/28913


rmackay9:
@bugobliterator,

Thanks very much for this!  I guess it's been tested to some extent?

I'm not really qualified to review this but maybe we could get an approval from @tridge, @tpwrules, @peterbarker or someone else who is more familiar with this part of the code?

tpwrules (approved):
Looks like this also includes #28900 which is why it did not cherry-pick cleanly.

Otherwise LGTM, checked that the bootloaders match.

PR #29299, opened 2025-02-14
AP_NavEKF3: elevate compass change statustext alert

rishabsingh3003:
This shows up on the GCS very silently and often gets missed by the pilot.  When landing in an area with lots of magnetic interference, the primary compass might change automatically. 

I recently saw a crash where the secondary compass wasn't well calibrated, and the switch caused a crash. 
This alertness level is in line with an EKF Lane switch.


tridge:
I think we could use the AP_Declination 3D mag database to calculate the expected innovation for the current compass and the proposed new compass and not switch if the new compass would be worse

PR #29300, opened 2025-02-14
AP_NavEKF3: revert to primary compass on ground

rishabsingh3003:
Revert to the primary compass if it was changed in flight. The compass might have changed because of landing in an area with high mag interference causing a timeout in the magfusion. 
Its more than likely the user wants to go back to flying on the primary mag on subsequent takeoff.  

Pushed on top of #29299

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
@@ -257,27 +257,41 @@ void NavEKF3_core::tryChangeCompass(void)
         if (tempIndex >= maxCount) {
             tempIndex -= maxCount;
         }
-        // if the magnetometer is allowed to be used for yaw and has a different index, we start using it
-        if (compass.healthy(tempIndex) && compass.use_for_yaw(tempIndex) && tempIndex != magSelectIndex) {
-            magSelectIndex = tempIndex;
-            GCS_SEND_TEXT(MAV_SEVERITY_INFO, "EKF3 IMU%u switching to compass %u",(unsigned)imu_index,magSelectIndex);
-            // reset the timeout flag and timer
-            magTimeout = false;
-            lastHealthyMagTime_ms = imuSampleTime_ms;
-            // zero the learned magnetometer bias states
-            stateStruct.body_magfield.zero();
-            // clear the measurement buffer
-            storedMag.reset();
-            // clear the data waiting flag so that we do not use any data pending from the previous sensor
-            magDataToFuse = false;
-            // request a reset of the magnetic field states
-            magStateResetRequest = true;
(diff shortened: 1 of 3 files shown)


libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
```diff
@@ -313,7 +327,16 @@ void NavEKF3_core::readMagData()
     if (maxCount > 1) {
         bool fusionTimeout = magTimeout && !onGround && imuSampleTime_ms - ekfStartTime_ms > 30000 && !(frontend->_affinity & EKF_AFFINITY_MAG);
         bool sensorTimeout = !compass.healthy(magSelectIndex) && imuSampleTime_ms - lastMagRead_ms > frontend->magFailTimeLimit_ms;
-        if (fusionTimeout || sensorTimeout) {
+
+        // check if we have changed the primary compass in flight
+        // get_first_usable() returns the first compass that is healthy and ordered by priority set by user
+        const bool onground_and_changed_primary = onGround && (magSelectIndex != dal.compass().get_first_usable()) && !(frontend->_affinity & EKF_AFFINITY_MAG);
```
tridge, line 333: i think this should be dal.get_armed(), same as forcing first lane

tridge: I think this likely should be in NavEKF3_core::update_mag_selection()

tridge (changes requested):

tridge:
need to check we don't get an issue with the copter option to save offsets from EKF on disarm, make sure that still works, and saves right offsets for right compass

rishabsingh3003:
I have briefly tested in SITL that the offsets from EKF get saved properly. This happens because its saved before the vehicle actually disarms. The mag gets changed back to primary after saving the offsets.

libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
```diff
@@ -257,27 +257,41 @@ void NavEKF3_core::tryChangeCompass(void)
...
+        storedMag.reset();
+        // clear the data waiting flag so that we do not use any data pending from the previous sensor
+        magDataToFuse = false;
+        // request a reset of the magnetic field states
+        magStateResetRequest = true;
+        // declare the field unlearned so that the reset request will be obeyed
+        magFieldLearned = false;
+        // reset body mag variances on next CovariancePrediction
+        needMagBodyVarReset = true;
+        return;
```
tridge, line 294: remove return

tridge (approved):
looks good to me
@priseborough you OK with this? it matches what we do with primary lane

rmackay9 (approved):
looks good to me

PR #13776, opened 2020-03-10
EKF2 and EKF3 operation with no or bad compass data

priseborough:
Replaces https://github.com/ArduPilot/ardupilot/pull/13603

Commit history has been cleaned. See  https://github.com/ArduPilot/ardupilot/pull/13603 for previous testing.

PR #29301, opened 2025-02-15
RC_Channel: Document EKF lane switch and yaw reset aux functions.

IamPete1:
Theses options were added in https://github.com/ArduPilot/ardupilot/pull/13776.

### DIFF
--- a/libraries/RC_Channel/RC_Channel.cpp
+++ b/libraries/RC_Channel/RC_Channel.cpp
@@ -203,6 +203,8 @@ const AP_Param::GroupInfo RC_Channel::var_info[] = {
     // @Values{Copter}: 99:AUTO RTL
     // @Values{Copter, Rover, Plane, Blimp}: 100:KillIMU1, 101:KillIMU2
     // @Values{Copter, Rover, Plane}: 102:Camera Mode Toggle
+    // @Values{Copter, Rover, Plane, Blimp, Sub, Tracker}: 103: EKF lane switch attempt
+    // @Values{Copter, Rover, Plane, Blimp, Sub, Tracker}: 104: EKF yaw reset
     // @Values{Copter, Rover, Plane}: 105:GPS Disable Yaw
     // @Values{Rover, Plane}: 106:Disable Airspeed Use
     // @Values{Plane}: 107:Enable FW Autotune


Hwurzburg:
added to wiki https://github.com/ArduPilot/ardupilot_wiki/pull/6620

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -203,6 +203,8 @@ const AP_Param::GroupInfo RC_Channel::var_info[] = {
     // @Values{Copter}: 99:AUTO RTL
     // @Values{Copter, Rover, Plane, Blimp}: 100:KillIMU1, 101:KillIMU2
     // @Values{Copter, Rover, Plane}: 102:Camera Mode Toggle
+    // @Values{Copter, Rover, Plane, Blimp}: 103: Trigger EKF lane switch attempt
+    // @Values{Copter, Rover, Plane, Blimp}: 104: Trigger EKF yaw reset
```
peterbarker, line 207: What makes these not available on Sub and Tracker?

IamPete1: Nothing, it used to be that they were not enumerated specifically anywhere so they would get all options. Although that seems not to be the case anymore, I think they will currently just get 112.

IamPete1: Yeah, https://ardupilot.org/antennatracker/docs/parameters.html#rc1-option-rc-input-option. Removing the mark-up for 112 doesn't seem to fix tho. I guess there are lots of options that need that mark-up adding. When I first split the descriptions over multiple lines it would export all options in that case. 

Anyway, I will add them here too.

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -203,6 +203,8 @@ const AP_Param::GroupInfo RC_Channel::var_info[] = {
     // @Values{Copter}: 99:AUTO RTL
     // @Values{Copter, Rover, Plane, Blimp}: 100:KillIMU1, 101:KillIMU2
     // @Values{Copter, Rover, Plane}: 102:Camera Mode Toggle
+    // @Values{Copter, Rover, Plane, Blimp, Sub, Tracker}: 103: Trigger EKF lane switch attempt
```
rmackay9, line 206: Let's remove the "Trigger" from the front, sorry!

rmackay9 (approved):

PR #29302, opened 2025-02-15
RC_Channel: add docs for copter inflight ahrs trim

IamPete1:
Missed from https://github.com/ArduPilot/ardupilot/pull/28881

### DIFF
--- a/libraries/RC_Channel/RC_Channel.cpp
+++ b/libraries/RC_Channel/RC_Channel.cpp
@@ -243,6 +243,7 @@ const AP_Param::GroupInfo RC_Channel::var_info[] = {
     // @Values{Plane}: 179:ICEngine start / stop
     // @Values{Copter, Plane}: 180:Test autotuned gains after tune is complete
     // @Values{Plane}: 181: QuickTune
+    // @Values{Copter}: 182: In flight AHRS trim adjust
     // @Values{Rover}: 201:Roll
     // @Values{Rover}: 202:Pitch
     // @Values{Rover}: 207:MainSail


peterbarker (approved):
LGTM

libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -243,6 +243,7 @@ const AP_Param::GroupInfo RC_Channel::var_info[] = {
     // @Values{Plane}: 179:ICEngine start / stop
     // @Values{Copter, Plane}: 180:Test autotuned gains after tune is complete
     // @Values{Plane}: 181: QuickTune
+    // @Values{Copter}: 182: In flight AHRS trim
```
Hwurzburg, line 246:
```suggestion
    // @Values{Copter}: 182: AHRS trim save
```
I think this is better,  "In flight AHRS trim save" would work also, but a bit long since MP cuts off all of these long ones in the dialog box

Hwurzburg (changes requested):

libraries/RC_Channel/RC_Channel.cpp
IamPete1: I didn't want people to confuse it with a "set level" like you might do from the GCS on the bench. The build server description from the original PR is "Enable Copter in-flight AHRS Trim calibration" so this also fits with that.

IamPete1: But I don't have any very strong feelings, just want to get it documented.

peterbarker: This is most definitly not "save trim"!  We have a separate option which does an instantaneous save of your trim, and this is not that!

This could be `In flight calibration of AHRS Level Trim via RC stick input"

Hwurzburg: Pete and I discussed and settled on a name...

Hwurzburg (approved):

PR #29308, opened 2025-02-16
AP_Logger: File: clean up fsync code

tpwrules:
See commits for details.

Costs <200 bytes on ChibiOS in general but simplifies the code and allows for further cleanup. I intend to come back and do this as well, I think some of these assumptions are outdated.

Tested on Cube Orange and KakuteH7Mini-Nand. Log behavior and performance is identical before and after.

### DIFF
--- a/libraries/AP_Filesystem/AP_Filesystem.h
+++ b/libraries/AP_Filesystem/AP_Filesystem.h
@@ -110,6 +110,10 @@ public:
     struct dirent *readdir(DirHandle *dirp);
     int closedir(DirHandle *dirp);
 
+    // return number of bytes that should be written before fsync for optimal
+    // streaming performance/robustness. if zero, any number can be written.
+    uint32_t bytes_until_fsync(int fd);
+
     // return free disk space in bytes, -1 on error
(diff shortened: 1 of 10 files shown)


libraries/AP_Filesystem/AP_Filesystem.h
```diff
@@ -110,6 +110,10 @@ class AP_Filesystem {
     struct dirent *readdir(DirHandle *dirp);
     int closedir(DirHandle *dirp);
 
+    // return number of bytes that should be written before fsync for optimal
+    // streaming performance/robustness. if zero, any number can be written.
+    uint32_t bytes_until_fsync(int fd);
```
peterbarker, line 115: Is there some reason all of these can't be `const`?

tpwrules: This would not let implementations take member semaphores, which LittleFS currently does. So it's a nice idea but not feasible.

tridge (approved):
looks good to me!
@peterbarker you OK with this as logging maintainer?

PR #29309, opened 2025-02-17
AP_Mount: CADDX RC rate control fix

rmackay9:
The CADDX mount driver is fairly new (see https://github.com/ArduPilot/ardupilot/pull/28942) and while it did go through alpha testing we apparently never tested RC rate control.  I think what must have happened is I must have tested rate control using QGC which means rate control was only done using the MAVLINK_TARGETING submode.

This change means that rate targets will always update the angle targets before being passed to the physical gimbal.

I've tested this lightly on real hardware and it works

This bug was found during testing of @peterbarker's backport PR https://github.com/ArduPilot/ardupilot/pull/29296,. thanks!


peterbarker (approved):
LGTM.

PR #29310, opened 2025-02-17
DroneCAN: support sending/receiving DroneCAN messages in lua

tridge:
This adds support for sending and receiving DroneCAN messages from lua scripts.

It includes an example script that subscribes to messages, sends requests and sends broadcast messages

Note that this relies on this libcanard PR: https://github.com/dronecan/libcanard/pull/74


libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4202,3 +4202,50 @@ function crsf:get_menu_event(events) end
...
+
+-- create a DroneCAN_Handle, needed for all other DroneCAN message operations
+-- use close() method when finished
+---@param bus_index number -- DroneCAN bus index, 0 for first bus
+---@return DroneCAN_Handle_ud
+function DroneCAN_Handle(bus_index) end
+
+-- return true if last message was used CANFD
+---@return boolean
+function DroneCAN_Handle_ud:canfd() end
```
IamPete1, line 4218: This is probably just me not understanding CANFD, but why would a script care? I thought it was a property of the whole bus and fd messages don't mix with normal ones.

IamPete1: If it is per message like this should it not be another return on the `check_message` method?

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4202,3 +4202,50 @@ function crsf:get_menu_event(events) end
...
+---@param value boolean
+function DroneCAN_Handle_ud:canfd(value) end
+
+-- set the 64 bit DroneCAN signature for the message
+---@param value uint64_t_ud
+function DroneCAN_Handle_ud:signature(value) end
+
+-- set the DroneCAN data type of the message
+---@param value integer
+function DroneCAN_Handle_ud:data_type(value) end
```
IamPete1, line 4230: Can't we pass these in the constructor? Would you ever change the signature and types or just get a new one?

libraries/AP_Scripting/examples/DroneCAN_test.lua
```diff
@@ -0,0 +1,261 @@
+--[[
+    example of creating and sending DroneCAN messages
+--]]
+---@diagnostic disable: param-type-mismatch
```
IamPete1, line 4: Cheating ;) What is the error? I have seen them with string unpack, in that case you can cast like this:

https://github.com/ArduPilot/ardupilot/blob/1431a08b8d5a8fcaf06f3a9320de5e8a8e81b737/libraries/AP_Scripting/examples/message_interval.lua#L48-L51

The goal is to ideally remove the disabling of the check. If the cast does not work then a per line disable rather than a per file would be preferred.

libraries/AP_Scripting/examples/DroneCAN_test.lua
```diff
@@ -0,0 +1,261 @@
...
+-- table of all nodes
+local node_status = {}
+
+
+--[[
+    send highres mag using a global handle
+--]]
+local function send_mag_highres()
+    dc_handle:signature(MAGNETICFIELDSTRENGTHHIRES_SIGNATURE)
+    dc_handle:data_type(MAGNETICFIELDSTRENGTHHIRES_ID)
```
IamPete1, line 55: Its a little bit odd to have a handle but then change it for each send. I wonder if broadcast would be better to not go via the helper? EG:

```
DroneCAN:broadcast(0, MAGNETICFIELDSTRENGTHHIRES_SIGNATURE, MAGNETICFIELDSTRENGTHHIRES_ID, payload)
```

Or we could establish a pattern of using one helper per message type.

libraries/AP_Scripting/examples/DroneCAN_test.lua
```diff
@@ -0,0 +1,261 @@
...
+--]]
+local nodestatus_handle = DroneCAN_Handle(0)
+nodestatus_handle:data_type(NODESTATUS_ID)
+nodestatus_handle:signature(NODESTATUS_SIGNATURE)
+nodestatus_handle:subscribe()
+
+--[[
+    setup subscription to raw air data
+--]]
+local airspeed_handle = DroneCAN_Handle(0)
```
IamPete1, line 41: Should this be `DroneCAN:handle(0)` so we can add other methods to the `DroneCAN` object in the future.

IamPete1: One example might be to get a list of the "seen" node IDs. Or to reset the saved unique ID in the database for a given  node.

libraries/AP_Scripting/examples/DroneCAN_test.lua
```diff
@@ -0,0 +1,261 @@
...
+    parse a parameter GetSet Value
+--]]
+local function parse_param_Value(payload, byte_offset)
+    local vtype = string.unpack("B", payload, byte_offset[1])
+    byte_offset[1] = byte_offset[1] + 1
+    if vtype == 0 then
+        return nil
+    elseif vtype == 1 then
+        -- integer (treat as 32 bit for now, actually 64 bit)
+        local ret = string.unpack("i", payload, byte_offset[1])
```
IamPete1, line 165: You should be able to unpack it to two ints and the reassemble as a uint64. I think the only thing we would be missing would be the logging of a int64 by casting from a uint64.

Also unpacking as 32bit will mean negative numbers don't work correctly.

libraries/AP_Scripting/examples/DroneCAN_test.lua
```diff
@@ -0,0 +1,261 @@
...
+--]]
+local function check_node_status()
+    local payload, nodeid = nodestatus_handle:check_message()
+    if not payload then
+        return
+    end
+    local uptime_sec, bits, _ = string.unpack("IBH", payload)
+    local health = bits&3
+    local mode = (bits>>2)&7
+    local sub_mode = (bits>>5)&7
```
IamPete1, line 238: I thought that the helper would do all the random size bits for us? I thought the unpack would be of the `uavcan_protocol_NodeStatus` structure it seems like its the "wire" structure?

Tools/AP_Periph/batt_info.cpp
```diff
@@ -0,0 +1,46 @@
...
+const AP_Param::GroupInfo BattInfo::var_info[] {
+    // @Param: _NUM_CYCLES
+    // @DisplayName: Number of cycles
+    // @Description: Number of cycles the battery has been through
+    AP_GROUPINFO("_NUM_CYCLES", 1, BattInfo, num_cycles, 0),
+
+    // @Param: _ARM_HOURS
+    // @DisplayName: Number of armed hours
+    // @Description: Number of hours the battery has been armed
+    AP_GROUPINFO("_ARM_HOURS", 2, BattInfo, num_hours, 0),
```
IamPete1, line 36: This one could be updated locally?

Tools/AP_Periph/batt_info.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#if AP_PERIPH_BATTERY_INFO_ENABLED
+
+extern const AP_HAL::HAL &hal;
+
+const AP_Param::GroupInfo BattInfo::var_info[] {
+    // @Param: _NUM_CYCLES
+    // @DisplayName: Number of cycles
+    // @Description: Number of cycles the battery has been through
+    AP_GROUPINFO("_NUM_CYCLES", 1, BattInfo, num_cycles, 0),
```
IamPete1, line 31: I guess this is a bit more subtle, hense the need for a remote script? Scripting should run fine on the periph node itself. The script could add the params too.

libraries/AP_Scripting/docs/docs.lua
tridge: a script cares as the packing with CANFD is (sometimes!) different. In particular we never do TAO with CANFD, so if there is a variable length array at the end of the top level object then the length is encoded in the packet with CANFD and it is not with bxCAN (this stems from the fact that with CANFD you don't know the exact packet length, as the length is rounded up in an odd fashion, so you can't tell the array length reliably from the packet length)

tridge: I did it as a field for a few reasons. One is to keep as much as possible in auto-generated bindings, another is that nearly everyone won't care about that field (I know they could ignore it), another is that there are other attributes of the transfer that we may also want to add in the future (eg. the arrival timestamp) and I think if we add that then it is cleaner to have these rarely used attributes.
I don't feel strongly about it and I know these are not really strong reasons, so can change if you think it is important

libraries/AP_Scripting/docs/docs.lua
tridge: i think it could be in the constructor, I can't immediately think of a reason to want to change them

libraries/AP_Scripting/examples/DroneCAN_test.lua
tridge: DroneCAN:handle(0) would normally imply DroneCAN is a singleton, which it isn't (there is one per bus). Should we still do it that way? Not sure how to set that up without a singleton

libraries/AP_Scripting/examples/DroneCAN_test.lua
tridge: yes, I know we can, just didn't need to bother for this example, especially as we don't have AP_Int64, but I can add it, negative support is worth it

libraries/AP_Scripting/examples/DroneCAN_test.lua
tridge: it is the payload structure, it ties in at the message level, it doesn't use the compiled DSDL

Tools/AP_Periph/batt_info.cpp
tridge: on a L431? I doubt we'd have the memory

libraries/AP_Scripting/examples/DroneCAN_test.lua
IamPete1: Ah, the dreaded logger one. You should be able to swap to colon and have it pass. 
```
logger:write("ASPL", 'SP,DP,ST,DT,SAT,PIT', 'ffffff', ...
```

Tools/AP_Periph/batt_info.cpp
IamPete1: Maybe, I guess its not actually running anything else? Certainly works on G4's.

libraries/AP_Scripting/examples/DroneCAN_test.lua
IamPete1: I guess we could have a bus object, so you would do a "bus" and then have a "handle" method on the bus. So you wuold endup with:
```
DroneCAN:bus(0):handle()
```

libraries/AP_Scripting/examples/DroneCAN_test.lua
IamPete1: You could get negatives working and still unpack to 32 bit in the script. Maybe return nan if anything other than the sign bit is set in the top 32 bits. That should cover all values that a AP Periph could return.

libraries/AP_Scripting/docs/docs.lua
IamPete1: Isn't there a race condition problem with this getter if there was a CANFD and a bxCAN node sending the message your subscribed too? How do you match up the fd status to the sending node?

tpwrules:
I have to admit I don't understand what getting the internal libcanard semaphore accomplishes. Is it just to avoid deadlocks? I don't think that can happen here.

tpwrules:
> the reason we need the get_semaphore() is mutex ordering. If we don't et that mutex before we get the sem in the scripting code then we deadlock if a message arrives during deletion of a Subscriber in DroneCAN_Handle::check_message(). It happens in SITL if you comment out the get_semaphore() code, DroneCAN thread is in handle_message(), scripting in check_message()

Ah yes, I see that now. But why don't we just release `h->sem` before the `delete`? The function which sets up `h->subscriber` does not even take `h->sem`. It also runs on the same thread as the function doing the delete so I don't think there is a race condition. If we received a message it would just get stored before being thrown away in the destructor.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4202,3 +4202,50 @@ function crsf:get_menu_event(events) end
 ---@param data string -- binary encoded response payload
 ---@return boolean -- true if the repsonse was successfully sent, false otherwise
 function crsf:send_write_response(data) end
+
+-- handle for DroneCAN message operations
+---@class DroneCAN_Handle_ud
+local DroneCAN_Handle_ud = {}
+
+-- create a DroneCAN_Handle, needed for all other DroneCAN message operations
+-- use close() method when finished
```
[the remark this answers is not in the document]
tridge: fixed

libraries/AP_Scripting/docs/docs.lua

libraries/AP_Scripting/examples/DroneCAN_test.lua
tridge: changed

libraries/AP_Scripting/docs/docs.lua
tridge: fixed

Tools/AP_Periph/batt_info.cpp
tridge: done

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4202,3 +4202,50 @@ function crsf:get_menu_event(events) end
...
+---@param bus_index number -- DroneCAN bus index, 0 for first bus
```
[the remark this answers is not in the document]

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4212,3 +4212,45 @@ function crsf:get_menu_event(events) end
...
+-- handle for DroneCAN message operations
+---@class DroneCAN_Handle_ud
+local DroneCAN_Handle_ud = {}
+
+-- create a DroneCAN_Handle, needed for all other DroneCAN message operations
+---@param driver_index number -- DroneCAN driver index, 0 for first driver
+---@param signature uint64_t_ud -- message signature
+---@param data_type number -- message data type ID
+---@return DroneCAN_Handle_ud
+function DroneCAN_Handle(bus_index, signature, data_type) end
```
IamPete1, line 4225:
```suggestion
function DroneCAN_Handle(driver_index, signature, data_type) end
```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4212,3 +4212,45 @@ function crsf:get_menu_event(events) end
...
+
+-- create a DroneCAN_Handle, needed for all other DroneCAN message operations
+---@param driver_index number -- DroneCAN driver index, 0 for first driver
+---@param signature uint64_t_ud -- message signature
+---@param data_type number -- message data type ID
+---@param canfd boolean -- send as CANFD
+---@return DroneCAN_Handle_ud
+function DroneCAN_Handle(bus_index, signature, data_type, canfd) end
```
IamPete1, line 4233:
```suggestion
function DroneCAN_Handle(driver_index, signature, data_type, canfd) end
```

libraries/AP_Scripting/examples/DroneCAN_test.lua
tridge: fixed up to handle negative, still limited to 32 bit though. Pity we can't have an int64_t type and string.format() it

libraries/AP_Scripting/docs/docs.lua

libraries/AP_Scripting/docs/docs.lua

tridge:
@tpwrules I've the get_semaphore() and and per-handle semaphore, replacing it with get_sem_rx() in AP_DroneCAN

libraries/AP_Scripting/AP_Scripting_helpers.cpp
```diff
@@ -195,4 +195,316 @@ bool Parameter::set_default(float value)
...
+        h->subscriber = nullptr;
+    }
+    
+    return 4;
+}
+
+/*
+  cleanup any dangling subscribers
+ */
+void DroneCAN_Handle::destroy_all(void)
```
tpwrules, line 424: This is not needed, Lua guarantees that the `__gc` of all objects will be called when the state is shut down, which we do properly.

libraries/AP_Scripting/AP_Scripting_helpers.cpp
```diff
@@ -195,4 +195,316 @@ bool Parameter::set_default(float value)
...
+    static bool metatable_created;
+    if (!metatable_created) {
+        metatable_created = true;
+        luaL_newmetatable(L, "DroneCAN_Handle");
+    
+        lua_pushcfunction(L, DroneCAN_Handle::__gc);
+        lua_setfield(L, -2, "__gc");
+    
+        lua_pop(L, 1);
+    }
```
tpwrules, line 473:
```suggestion

```

This can be handled by the binding generator with `userdata DroneCAN_Handle manual_operator __gc DroneCAN_Handle::__gc`.

tpwrules (changes requested):
Happy to see the semaphores cleaned up. I think the per-handler semaphore would have worked fine, but using the DroneCAN one probably saves memory (at the cost of CPU and interference risk).

I definitely would want to see the battery tag changes pulled out in a separate PR.

libraries/AP_Scripting/AP_Scripting_helpers.cpp
IamPete1: Even if something very bad happens? Maybe we could internal error if there is anything left for this to clean up and see if we get any reports.

tpwrules: The behavior I mention is handled by `lua_close` which we always call AFAICT: "Destroys all objects in the given Lua state (calling the corresponding garbage-collection metamethods, if any) and frees all dynamic memory used by this state."

If there is memory corruption or something then the metamethod might not be called but then we are already hosed...

libraries/AP_Scripting/AP_Scripting_helpers.cpp
tridge: i thought I'd tried this, but I just tried it again and it does work, thanks! changed

libraries/AP_Scripting/AP_Scripting_helpers.cpp
tridge: I had tested this before with only the __gc and not the destroy_all() and I thought it did leak in some cases, but I now can't find a leak with either a script having an error or a scripting restart, so I've removed the destroy_all

tridge:
@tpwrules i've moved the BatteryTag changes into https://github.com/ArduPilot/ardupilot/pull/29983

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4212,3 +4212,45 @@ function crsf:get_menu_event(events) end
...
+---@param signature uint64_t_ud -- message signature
+---@param data_type number -- message data type ID
+---@return DroneCAN_Handle_ud
+function DroneCAN_Handle(driver_index, signature, data_type) end
+
+-- create a DroneCAN_Handle, needed for all other DroneCAN message operations
+---@param driver_index number -- DroneCAN driver index, 0 for first driver
+---@param signature uint64_t_ud -- message signature
+---@param data_type number -- message data type ID
+---@param canfd boolean -- send as CANFD
```
IamPete1, line 4231: I think the only difference here is this argument, you can mark it as optional with a question mark.

```suggestion
---@param canfd? boolean -- send as CANFD
```

That should allow you to remove the other copy.

IamPete1 (approved):
Scripting stuff looks good, just a small thing in the docs.

libraries/AP_Scripting/AP_Scripting_helpers.cpp
```diff
@@ -195,4 +195,270 @@ bool Parameter::set_default(float value)
...
+/*
+  check for an incoming message
+ */
+int DroneCAN_Handle::check_message(lua_State *L)
+{
+    /*
+      get the receive mutex for this driver
+     */
+    DroneCAN_Handle *h = check_DroneCAN_Handle(L, 1);
+    WITH_SEMAPHORE(h->dc->get_canard_iface().get_sem_rx());
```
tpwrules, line 377: Need to release this before calling into Lua, lest it throws an error and leaves the semaphore locked.

libraries/AP_Scripting/AP_Scripting_helpers.cpp
```diff
@@ -195,4 +195,270 @@ bool Parameter::set_default(float value)
...
+    if (s == nullptr) {
+        return 0;
+    }
+    Subscriber::Payload payload;
+    if (!s->payloads.pop(payload)) {
+        return 0;
+    }
+
+    lua_pushlstring(L, (char *)payload.data, payload.length);
+    free(payload.data);
```
tpwrules, line 390: Memory leak if Lua errors. Perhaps peek the Payload, call Lua, then pop and free if it succeeds?

tpwrules (changes requested):
Some error condition handling but overall looks good.

libraries/AP_Scripting/AP_Scripting_helpers.cpp
tpwrules: We know we only have one consumer so we could re-take the semaphore to do the pop.

libraries/AP_Scripting/AP_Scripting_helpers.cpp
tridge: turns out we don't need the mutex at that time, we only need it for the delete later in the function, so I moved it. For single threaded consumer without using push_force() you don't need the mutex for the pop

libraries/AP_Scripting/AP_Scripting_helpers.cpp
tridge: fixed, thanks

libraries/AP_Scripting/docs/docs.lua
tridge: thanks, changed

libraries/AP_Scripting/AP_Scripting_helpers.cpp
```diff
@@ -195,4 +195,275 @@ bool Parameter::set_default(float value)
...
+int DroneCAN_Handle::new_handle(lua_State *L)
+{
+    lua_Number driver_index = luaL_checknumber(L, 1);
+    const uint64_t sig = *check_uint64_t(L, 2);
+    const uint16_t dtype = luaL_checkinteger(L, 3);
+#if CANARD_ENABLE_CANFD
+    const int args = lua_gettop(L);
+    bool send_canfd = false;
+    if (args > 3) {
+        send_canfd = luaL_checkinteger(L, 4) > 0;
```
tpwrules, line 439:
```suggestion
        send_canfd = (bool)lua_toboolean(L, 4);
```

`luaL_checkinteger` will reject booleans; they do not implicitly convert from numbers like C or Python. `lua_toboolean` is what the binding generator uses for boolean argument, but it treats 0 as `true`.

tridge: good catch, fixed and tested

tpwrules (approved):
The little test example works well on Cube Orange. Great work! I also did some futzing with the REPL and could not immediately break the autopilot or my node.

Future expansion ideas:
* Autotest
* Ability to have autopilot itself/other scripts see script-sent messages
* DroneCAN server support (your idea)
* Priority handler/inhibit mechanism so a script can eat messages
* C++-assisted encoding and decoding

tridge:
@tpwrules thanks for your help with this PR!

PR #29312, opened 2025-02-17
AP_Compass: throw away all samples which aren't acceptable in thinning

peterbarker:
From my 5-year-old notes:
```
+Getting a good list of samples for the fits to run across is probably a good thing to get working well.
+
+I don't think we're doing that at the moment.
+
+At the moment: we test a sample, and if it is no good then we copy a random one in from the end of the list.  We do not test the sample we just copied in, rather move on through the list.
+
+This PR changes things such that we test every sample in the buffer as part of the thinning step.
+
+```
+2019-12-06 10:07:40.92: STATUSTEXT {severity : 6, text : Threw away 247/300 samples}
+2019-12-06 10:07:41.54: STATUSTEXT {severity : 6, text : Threw away 232/300 samples}
+```
+
+There are 576 faces on the default polyhedron.   This test shows that we're only hitting 50-70 of them in the initial collect-samples pass.

+After the initial collect samples pass we fill the buffer up with completely untested samples - I can't imagine we'd do better than the same again.  So overall we'll be hitting 20% of the faces with no guarantee on face distribution, except inasmuch
```

I have no idea how I was going to finish that sentence.

The "threw away line was extra debug in the thinning_samples method recording before/after sample counts.

I've tested this on a Durandal quadcopter here and the calibration result is similar to what was on there.

### DIFF
--- a/libraries/AP_Compass/CompassCalibrator.cpp
+++ b/libraries/AP_Compass/CompassCalibrator.cpp
@@ -513,13 +513,22 @@ void CompassCalibrator::thin_samples()
         _sample_buffer[j] = temp;
     }
 
-    // remove any samples that are close together
-    for (uint16_t i=0; i < _samples_collected; i++) {
+    // remove any samples that are close together.  There are two
+    // cases here - where we throw a sample away and when we keep the
+    // sample.  In the former case we MUST move the end-offset
+    // (_samples_collected) towards the loop iterator (i).  In the
+    // lattter case we MUST move the loop iterator forward (towards
+    // _samples_collected).
+    uint16_t i = 0;
+    while (i != _samples_collected) {
         if (!accept_sample(_sample_buffer[i], i)) {
+            // throw sample away
             _sample_buffer[i] = _sample_buffer[_samples_collected-1];
             _samples_collected--;
             _samples_thinned++;
+            continue;
         }
+        i++;
     }
 
     update_completion_mask();


libraries/AP_Compass/CompassCalibrator.cpp
```diff
@@ -513,13 +513,22 @@ void CompassCalibrator::thin_samples()
...
+    // sample.  In the former case we MUST move the end-offset
+    // (_samples_collected) towards the loop iterator (i).  In the
+    // lattter case we MUST move the loop iterator forward (towards
+    // _samples_collected).
+    uint16_t i = 0;
+    while (i != _samples_collected) {
         if (!accept_sample(_sample_buffer[i], i)) {
+            // throw sample away
             _sample_buffer[i] = _sample_buffer[_samples_collected-1];
             _samples_collected--;
```
andyp1per, line 527: Couldn't this get to 0 meaning the check above would be a free memory read?

peterbarker: Then i would equal samples collected, so we would exit the while loop

libraries/AP_Compass/CompassCalibrator.cpp
```diff
@@ -513,13 +513,22 @@ void CompassCalibrator::thin_samples()
...
-    // remove any samples that are close together
-    for (uint16_t i=0; i < _samples_collected; i++) {
+    // remove any samples that are close together.  There are two
+    // cases here - where we throw a sample away and when we keep the
+    // sample.  In the former case we MUST move the end-offset
+    // (_samples_collected) towards the loop iterator (i).  In the
+    // lattter case we MUST move the loop iterator forward (towards
+    // _samples_collected).
+    uint16_t i = 0;
+    while (i != _samples_collected) {
```
tridge, line 523: maybe i < _samples_collected-1 ? otherwise copies to itself

tridge (changes requested):

tridge:
i think this looks good, but should have a bit more testing

libraries/AP_Compass/CompassCalibrator.cpp
peterbarker: Done, but as a separate `if` statement to avoid potential unsigned underflow here

rnatau:
As discussed on #31152 this fix makes sense and fixes an oversight present in the thinning phase. However, in my understanding, the whole thinning might not be necessary as one could also just start from an empty buffer again in the second phase, as at that point the radius is correctly calibrated and pull_sample() will accept no redundant samples using the accept_sample() call.

In any case I've incorporated the fix as it is in this PR in our codebase and was able to conduct several calibrations without issue.

tridge (approved):
I think this is correct, just a bit hard to read

PR #29314, opened 2025-02-17
AP_Scripting: Better mavlink parsing for MAVLinkHL and Rockblock

stephendade:
Another Rockblock scripting update. Makes the mavlink parsing more robust and adds a configurable timeout for automatically going to High Latency mode on GCS loss.

### DIFF
--- a/libraries/AP_Scripting/applets/RockBlock.lua
+++ b/libraries/AP_Scripting/applets/RockBlock.lua
@@ -54,12 +54,12 @@ function bind_param(name)
  end
 
  -- setup RCK specific parameters
-assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 4), 'could not add param table')
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 5), 'could not add param table')
 --[[
   // @Param: RCK_FORCEHL
   // @DisplayName: Force enable High Latency mode
   // @Description: Automatically enables High Latency mode if not already enabled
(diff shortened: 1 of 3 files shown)


libraries/AP_Scripting/applets/RockBlock.lua
```diff
@@ -54,12 +54,12 @@ function bind_param(name)
  end
 
  -- setup RCK specific parameters
-assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 4), 'could not add param table')
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 5), 'could not add param table')
```
peterbarker, line 57: Isn't this going to break people's working setups?

stephendade: Tested this on a CubeOrangePlus - the params weren't lost when upgrading.

libraries/AP_Scripting/applets/RockBlock.lua
```diff
@@ -92,6 +92,16 @@ RCK_DEBUG     = bind_add_param('DEBUG', 3, 0)
 --]]
 RCK_ENABLE     = bind_add_param('ENABLE', 4, 1)
 
+--[[
+  // @Param: RCK_TIMEOUT
+  // @DisplayName: GCS timeout to start sendin Rockblock messages
```
peterbarker, line 97:
```suggestion
  // @DisplayName: GCS timeout to start sending Rockblock messages
```

peterbarker (approved):

PR #29319, opened 2025-02-17
Tools: tidy build_autotest.sh script

peterbarker:
uses pushd rather than multiple cd calls

indents for clarity

removes unused oldhash assignment

create and make use of intermediate variables


peterbarker:
I've tested this locally on the autotest-server VM

PR #29320, opened 2025-02-18
AP_GPS: fixed the check for disabling CONFIG_RATE_SOL config

tridge:
we should stop looking for SOL as soon as we know we have PVT, instead of waiting for all others to be cleared

may help with
https://discuss.ardupilot.org/t/fc-tries-to-config-gps-constantly-when-trying-to-arm-in-loiter-mode/129930

### DIFF
--- a/libraries/AP_GPS/AP_GPS_UBLOX.cpp
+++ b/libraries/AP_GPS/AP_GPS_UBLOX.cpp
@@ -1123,12 +1123,14 @@ AP_GPS_UBLOX::_parse_gps(void)
                     }
                     break;
                 case MSG_CFG_VALSET:
-                    CFG_Debug("NACK VALSET 0x%x 0x%x", (unsigned)_buffer.nack.msgID,
-                            unsigned(active_config.list[active_config.set_index].key));
-                    if (is_gnss_key(active_config.list[active_config.set_index].key)) {
-                        GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "GPS %u: unable to configure band 0x%02x",
-                            unsigned(state.instance + 1), unsigned(active_config.list[active_config.set_index].key));
+                    if (active_config.list != nullptr) {
+                        CFG_Debug("NACK VALSET 0x%x 0x%x", (unsigned)_buffer.nack.msgID,
+                                  unsigned(active_config.list[active_config.set_index].key));
+                        if (is_gnss_key(active_config.list[active_config.set_index].key)) {
+                            GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "GPS %u: unable to configure band 0x%02x",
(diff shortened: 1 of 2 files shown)


andyp1per (approved):

peterbarker:
I bisected the bug, this was the result:
```
a9dc7b440f7a5b9418733c7b88743767c38623d4 is the first bad commit
commit a9dc7b440f7a5b9418733c7b88743767c38623d4
Author: Andy Piper <github@andypiper.com>
Date:   Mon Apr 29 15:46:53 2024 +0100

    AP_GPS: support GPSx_GNSS_MODE for F9P
    
    support detecting F9P hardware variant
    fix bug in extension buffer management
    support NEO-F9P GNSS configuration
    allow multiple configuration values to be set in one go
    phase F9 configuration to account for GNSS reset
```

andyp1per:
@peterbarker looks right but will need testing on F9P if you did not do that already

libraries/AP_GPS/AP_GPS_UBLOX.cpp
```diff
@@ -498,6 +498,10 @@ AP_GPS_UBLOX::_request_next_config(void)
             const config_list *list = config_M10;
             const uint8_t list_length = ARRAY_SIZE(config_M10);
             Debug("Sending M10 settings");
+
+            // don't mix F9 and M10
+            _unconfigured_messages &= ~CONFIG_F9;
```
peterbarker, line 503: I can't help but wonder whether a tweak to the code affected by the patch which introduced the problems might be better.  In particular:


```
@@ -334,6 +336,12 @@ AP_GPS_UBLOX::_request_next_config(void)
         }
         break;
     case STEP_POLL_GNSS:
+        if (supports_F9_config()) {
+            if (last_configured_gnss != params.gnss_mode) {
+                _unconfigured_messages |= CONFIG_F9;
+            }
+            break;
+        }
         if (!_send_message(CLASS_CFG, MSG_CFG_GNSS, nullptr, 0)) {
             _next_message--;
         }
```

Renaming "supports_f9_config" to "supports_valget_and_valset()` might stop this happening again?

MallikarjunSE:
@peterbarker do you suggest testing on F10N too?

tridge:
@peterbarker @andyp1per I've updated this to allow for GNSS config on M10

tridge:
@andyp1per I have tested GPS1_GNSS_MODE with F9P and it still works fine

peterbarker:
Unrelated to this PR, but it looks like we leak `config_GNSS` when the UBLOX backend is deleted.

peterbarker (approved):

timtuxworth:
This may also have fixed a problem I was seeing with the QioTek M10S GPS.

rmackay9:

PR #29321, opened 2025-02-18
Plane: auto-deploy landing gear on fixed wing landing

tridge:
if LGR_OPTIONS is set to auto-deploy then deploy
this affects all fixed wing auto-landings

### DIFF
--- a/ArduPlane/Plane.cpp
+++ b/ArduPlane/Plane.cpp
@@ -558,7 +558,23 @@ void Plane::set_flight_stage(AP_FixedWing::FlightStage fs)
         return;
     }
 
-    landing.handle_flight_stage_change(fs == AP_FixedWing::FlightStage::LAND);
+    const bool is_landing = (fs == AP_FixedWing::FlightStage::LAND);
+
+    landing.handle_flight_stage_change(is_landing);
+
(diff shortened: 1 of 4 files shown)


Hwurzburg (approved):

Hwurzburg (dismissed):
metadata in LGR_OPTIONS needs a tweak

IamPete1:
We have this case in `servos.cpp`, seems a bit odd to be checking in another place too.

https://github.com/ArduPilot/ardupilot/blob/ab59a00ae17cb9c55fb7949508cec9dfde78465e/ArduPlane/servos.cpp#L754-L769

Maybe just move the `control_mode == &mode_auto` check down into the retract case so they deploy in any mode but only retract in auto.

tridge:
> We have this case in `servos.cpp`, seems a bit odd to be checking in another place too.

fixed

tridge:
> metadata in LGR_OPTIONS needs a tweak

done

IamPete1 (approved):
LGTM

rmackay9:
this is included in 4.6.0-beta5 via PR https://github.com/rmackay9/rmackay9-ardupilot/pull/38

PR #29327, opened 2025-02-19
AP_GPS: correct uBlox config debugging

peterbarker:


tpwrules (approved):
Seems to compile with all the debug options on.

PR #29328, opened 2025-02-19
Create and use AP_PERIPH_NOTIFY_ENABLED

shiv-tyagi:


Tools/AP_Periph/Parameters.cpp
```diff
@@ -472,7 +472,7 @@ const AP_Param::Info AP_Periph_FW::var_info[] = {
     GSCALAR(msp_port, "MSP_PORT", AP_PERIPH_MSP_PORT_DEFAULT),
 #endif
     
-#ifdef HAL_PERIPH_ENABLE_NOTIFY
+#if AP_PERIPH_NOTIFY_ENABLED
```
peterbarker, line 475: We do need to come back and revisit these guards around the objects themselves.

IMO the objects themselves should be guarded with `AP_NOTIFY_ENABLED` - which `AP_PERIPH_NOTIFY_ENABLED` would obviously enable, but other things may as well.

Obviously something for the future...

peterbarker (approved):
LGTM

PR #29332, opened 2025-02-19
AP_Filesystem: FATFS: clean up allocations and error handling

tpwrules:
Saves 100-200 bytes and makes the code a bit more straightforward.

Tested on CubeOrange that scripting and logging still work, that I can remove the SD card and get reasonable errors, and that I can put it back in and the use it without rebooting.

### DIFF
--- a/libraries/AP_Filesystem/AP_Filesystem_FATFS.cpp
+++ b/libraries/AP_Filesystem/AP_Filesystem_FATFS.cpp
@@ -40,7 +40,7 @@ static bool remount_needed;
 static HAL_Semaphore sem;
 
 typedef struct {
-    FIL *fh;
+    FIL fobj; // should be first member; it's the most used
     char *name;
 } FAT_FILE;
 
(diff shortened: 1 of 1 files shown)


libraries/AP_Filesystem/AP_Filesystem_FATFS.cpp
```diff
@@ -40,8 +40,10 @@ static bool remount_needed;
 static HAL_Semaphore sem;
 
 typedef struct {
-    FIL *fh;
-    char *name;
+    // should be first member; it's the most used
+    FIL fobj;
+    // must be last member; it's extended to hold the name
+    char name[1]; // length 1 to always be null terminated
```
tridge, line 46: need to find other examples where we use this trick in our code base. I'm concerned we could get an issue with array bounds checking in compilers or static analysis tools
I'd like @peterbarker to look at this and make sure it doesn't muck up his static analysis efforts

tpwrules: Lua does the same thing (but in .c instead of .cpp):

https://github.com/ArduPilot/ardupilot/blob/5f1d3dac680417a4cc1584524267c2c08d9ae3e4/libraries/AP_Scripting/lua/src/lobject.h#L453-L457

https://github.com/ArduPilot/ardupilot/blob/5f1d3dac680417a4cc1584524267c2c08d9ae3e4/libraries/AP_Scripting/lua/src/lfunc.h#L17-L18

peterbarker: > need to find other examples where we use this trick in our code base. I'm concerned we could get an issue with array bounds checking in compilers or static analysis tools I'd like @peterbarker to look at this and make sure it doesn't muck up his static analysis efforts

We don't build this library in SITL, which is where `clang-scan-build` works.

This deserves a very big comment explaining what we're doing , both on the structure and where we do the allocation.

I can't say I'm a big fan of the approach.

Encapsulation of the allocation with `static bool FAT_FILE::create(const char *fpath)` (just a `memset` and a `strndup`?) and `~FAT_FILE` (just a `delete`) might be a reasonable alternative, but won't save as much flash.

tpwrules: I will probably just revert that portion, I don’t think it saved that much flash.

tpwrules: Reverted, it only saved 16 bytes. If I got clever with short circuiting I could make reversion zero flash penalty but I think clarity does have some merit.

There is still of course the several byte RAM penalty but I think we will be okay there too :)

tridge (approved):

PR #29337, opened 2025-02-20
AP_Scripting: reset aux auths on start

robertlong13:
Allows a scripting restart to reset any auxiliary prearms, and clear out all the old previous aux id allocations.

This is very helpful during development of lua scripts that have prearm checks in them, as without this you can only restart a maximum of three times before you run out of aux auth ids ("PreArm: Too many auxiliary authorisers") and have to reboot. Additionally, if a prearm check was failing before the scripting restart, there was no way to clear it without rebooting.

### DIFF
--- a/libraries/AP_Arming/AP_Arming.cpp
+++ b/libraries/AP_Arming/AP_Arming.cpp
@@ -1464,6 +1464,27 @@ void AP_Arming::set_aux_auth_failed(uint8_t auth_id, const char* fail_msg)
     }
 }
 
+void AP_Arming::reset_all_aux_auths()
+{
+    WITH_SEMAPHORE(aux_auth_sem);
+
+    // clear all auxiliary authorisation ids
+    aux_auth_count = 0;
+    // clear any previous allocation errors
+    aux_auth_error = false;
+
+    // reset states for all auxiliary authorisation ids
+    for (uint8_t i = 0; i < aux_auth_count_max; i++) {
+        aux_auth_state[i] = AuxAuthStates::NO_RESPONSE;
+    }
+
+    // free up the failure message buffer
+    if (aux_auth_fail_msg != nullptr) {
+        free(aux_auth_fail_msg);
+        aux_auth_fail_msg = nullptr;
+    }
+}
+
 bool AP_Arming::aux_auth_checks(bool display_failure)
 {
     // handle error cases
--- a/libraries/AP_Scripting/AP_Scripting.cpp
+++ b/libraries/AP_Scripting/AP_Scripting.cpp
@@ -20,6 +20,7 @@
 #include <AP_Scripting/AP_Scripting.h>
 #include <AP_HAL/AP_HAL.h>
 #include <GCS_MAVLink/GCS.h>
+#include <AP_Arming/AP_Arming.h>
 
 #include "lua_scripts.h"
 
@@ -316,6 +317,10 @@ void AP_Scripting::thread(void) {
             // clear data in serial buffers that the script wasn't ready to
             // receive
             _serialdevice.clear();
+#endif
+#if AP_ARMING_ENABLED && AP_ARMING_AUX_AUTH_ENABLED
+            // Clear any dangling pre-arms from previous script loads
+            AP_Arming::get_singleton()->reset_all_aux_auths();
 #endif
             // run won't return while scripting is still active
             lua->run();
--- a/libraries/AP_Arming/AP_Arming.h
+++ b/libraries/AP_Arming/AP_Arming.h
@@ -129,6 +129,7 @@ public:
     bool get_aux_auth_id(uint8_t& auth_id);
     void set_aux_auth_passed(uint8_t auth_id);
     void set_aux_auth_failed(uint8_t auth_id, const char* fail_msg);
+    void reset_all_aux_auths();
 #endif
 
     static const struct AP_Param::GroupInfo        var_info[];


tpwrules:
This resets them when scripting stops, as opposed to when it starts, contrary to the PR title. Is there a reason to do it either way in particular?

After the reset, is arming possible?

robertlong13:
> This resets them when scripting stops, as opposed to when it starts, contrary to the PR and commit title. Is there a reason to do it either way in particular?

Ah, this stemmed from a misunderstanding on my part. I see now. I can live with it happening on stop, but there may be a case for making it only clear on start.

There is an argument to be made that if a script check is failing, when scripting is stopped, that shouldn't bypass the check. It should continue to fail. Only removing the script and restarting, or restarting and having the script pass all checks, should clear it.

On the other hand, you could argue that if scripting is not running, then you should not get any scripting prearm failures. Which is what this PR does.

I'm happy to go either way.

> After the reset, is arming possible?

Yes, I loaded up the wp1 takeoff prearm script, restarted several times with it failing, then uploaded a mission with a wp1 takeoff and the prearm clears.

tpwrules:
>  After the reset, is arming possible?

Sorry, I was asking for clarification on how the aux auth mechanism works, I've never investigated it before. After the reset function is called, the aux auth mechanism will not prevent arming until a script starts and interacts with it again?

I don't have any opinion on whether or not that's the right move, just trying to clarify how it works.

robertlong13:
> After the reset function is called, the aux auth mechanism will not prevent arming until a script starts and interacts with it again?

Correct

robertlong13:
@tpwrules, if I want to make this "reset on start" instead of stop, then I should put this right after line 320 in AP_Scripting, correct?

I'll bring it up for dev call next week and see if there's a strong preference for one vs the other.

tpwrules:
> @tpwrules, if I want to make this "reset on start" instead of stop, then I should put this right after line 320 in AP_Scripting, correct?

Correct.

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -1464,6 +1464,32 @@ void AP_Arming::set_aux_auth_failed(uint8_t auth_id, const char* fail_msg)
...
+        return;
+    }
+
+    // clear all auxiliary authorisation ids
+    aux_auth_count = 0;
+    // clear any previous allocation errors
+    aux_auth_error = false;
+
+    // reset states for all auxiliary authorisation ids
+    for (uint8_t i = 0; i < aux_auth_count; i++) {
```
timtuxworth, line 1482: You already set this to zero at 1477, so this won't do anything no?

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -1464,6 +1464,32 @@ void AP_Arming::set_aux_auth_failed(uint8_t auth_id, const char* fail_msg)
     }
 }
 
+void AP_Arming::reset_all_aux_auths()
+{
+    WITH_SEMAPHORE(aux_auth_sem);
+
+    // shortcut if this is not needed
+    if (aux_auth_count == 0) {
```
timtuxworth, line 1472: Not sure if this saves much and without this check if there was any floating message, it will also be clear, so I'd remove it.

libraries/AP_Scripting/AP_Scripting.cpp
```diff
@@ -367,6 +368,11 @@ void AP_Scripting::thread(void) {
             }
         }
 
+#if AP_ARMING_ENABLED && AP_ARMING_AUX_AUTH_ENABLED
+        // Clear prearm aux auths
```
timtuxworth, line 372: Maybe make it clear why we need this.

timtuxworth (commented):
Added a couple of thoughts. This PR fixes a problem I found when testing an arming check script. It's mostly a problem for script development in SITL, but I can imagine it happening at the field preventing arming.

libraries/AP_Arming/AP_Arming.cpp
robertlong13: Wow, yeah, meant to make this `aux_auth_count_max` originally, good catch.

libraries/AP_Arming/AP_Arming.cpp
robertlong13: You are absolutely right.

robertlong13:
> I'd go with reset on Start.

Yeah, sounds good to me. I like that one a little more too.

rmackay9:
I agree that reset-on-start is probably the safe approach

tpwrules:
Would it make sense to, on shutdown, reset the state to NO_RESPONSE, but not reset the count? Hopefully the scripts will always request the same number of aux auth handles...

timtuxworth:
> Would it make sense to, on shutdown, reset the state to NO_RESPONSE, but not reset the count? Hopefully the scripts will always request the same number of aux auth handles...

"Hopefully"?

timtuxworth:
> > > Would it make sense to, on shutdown, reset the state to NO_RESPONSE, but not reset the count? Hopefully the scripts will always request the same number of aux auth handles...
> > 
> > 
> > "Hopefully"?
> 
> I can imagine a script that doesn't (e.g. it checks to see if a particular device is attached and doesn't allocate a handle if not found) but it seems likely that most do.
> 
> If they request the same or more, everything will work fine. If they request less, you'll have to reboot the vehicle. But on the other hand there won't be an intermediate period where arming is possible.

I really don't think it's a good idea to design based on "hopefully" and "likely".

IamPete1 (approved):
Looks good. Clears pre-arms just before a scripts are run. So a failure will last until scripting is restarted (which is what we want).

timtuxworth (approved):
Looking forward to getting this in master.

tpwrules (approved):

PR #29338, opened 2025-02-20
hwdef: remove un-needed AP_NOTIFY_TONEALARM_ENABLED

peterbarker:
this is infered by having ALARM set

```
Board,plane
KakuteH7Mini,*
MambaH743v4,*
RADIX2HD,*
SkystarsH7HD,*
TBS_LUCID_H7,*
```


andyp1per:
This was originally to enable dshot alarm when ALARM was not set - did you check that this does not apply?

peterbarker:
> This was originally to enable dshot alarm when ALARM was not set - did you check that this does not apply?

As you can see from my matrix output from `size_compare_branches.py`, , the compiler output is unaffected by this PR for each of the affected hwdefs.

This is probably why that's the case (from `defaults_normal.h`):
```
#ifndef HAL_DSHOT_ALARM_ENABLED
#define HAL_DSHOT_ALARM_ENABLED (HAL_PWM_COUNT>0)
#endif
```
... but that also looks kind of wrong... a little over-enthusiastic as to what we enable it on?!

peterbarker:
... yeah, missing `HAL_DSHOT_ENABLED` on that define

tridge (approved):

PR #29339, opened 2025-02-20
GCS_Common: fix missing #if for CAMERA_FOV_STATUS

robertlong13:
Fixes issue where ArduPilot spams "Sending unknown message" if a GCS asks for CAMERA_FOV_STATUS


rmackay9 (approved):

robertlong13:
Well, this solves one console spam issue (and panic in SITL) and trades it for a much smaller console spam issue. Now, every time the GCS requests this message, it will send a statustext of "No ap_message for mavlink id (271)" (every 38 seconds), which is significantly better than "Sending unknown message" at a very high rate.

I still need to make Planner figure out that the autopilot doesn't support this when it sends a `MAV_RESULT_DENIED` reply (and generally stop it from asking every 38 seconds if the message is already coming in at the expected rate too).

rmackay9:
Hi @robertlong13,

Or we could change AP so that it always handles the message but perhaps does nothing

robertlong13:
> Or we could change AP so that it always handles the message but perhaps does nothing

I think that's worse. Sending a success ACK for a message request but not sending the message feels wrong.

Better would be to make it where we don't compile out the message itself but just compile out the expensive POI calculation (so Planner can still use the other good info in that message). Now that we correctly handle sending the message with missing POI info that's much easier now, but we can make that as a separate PR. I think this one is the right approach for this bug.

rmackay9:

PR #29340, opened 2025-02-20
Remove Linux RST Zynq board

peterbarker:
this can't have compiled since 2019, and nobody's complained.

We would require to do significant surgery on our AP_InertialSensor library if we wanted to support sourcing accel and gyro data from different SPI addresses!


tpwrules (approved):
LGTM

tpwrules:
We should drop its unique IMU driver too.

peterbarker:
> We should drop its unique IMU driver too.

Done.

PR #29341, opened 2025-02-20
Copter: allow auto trim in position hold modes

Hwurzburg:
This:
1. Fixes  bugs in #28881 (aux func was not init'd and auto trim not saved properly)
2. refactors to use an "allows_auto_trim" bool to avoid entering autotrim and then immediately cancelling; and neaten the code a bit
3. Adds Loiter/PosHold to autotrim for hands off trimming
4. Add a warning on attempting to enable autotrim when allowed, but not flying, rather than just entering and exiting immediately leaving a user to wonder why.
5. adjust marginally failing autotest
6. driveby that PeterB requested to fix comment on aux function for autotrim

Flight tested by myself and Peter Barker

### DIFF
--- a/ArduCopter/RC_Channel_Copter.cpp
+++ b/ArduCopter/RC_Channel_Copter.cpp
@@ -114,6 +114,9 @@ void RC_Channel_Copter::init_aux_function(const AUX_FUNC ch_option, const AuxSwi
     case AUX_FUNC::ARMDISARM_AIRMODE:
     case AUX_FUNC::TURBINE_START:
     case AUX_FUNC::FLIGHTMODE_PAUSE:
+#if AP_COPTER_AHRS_AUTO_TRIM_ENABLED
+    case AUX_FUNC::AHRS_AUTO_TRIM:
+#endif
         break;
     case AUX_FUNC::ACRO_TRAINER:
     case AUX_FUNC::ATTCON_ACCEL_LIM:
@@ -717,9 +720,19 @@ void RC_Channel_Copter::do_aux_function_change_force_flying(const AuxSwitchPos c
 // save_trim - adds roll and pitch trims from the radio to ahrs
 void RC_Channels_Copter::save_trim()
 {
+    float roll_trim = 0;
+    float pitch_trim = 0;
+#if AP_COPTER_AHRS_AUTO_TRIM_ENABLED
+    if (auto_trim.running) {
+        auto_trim.running = false;
+    } else {
+#endif
     // save roll and pitch trim
-    float roll_trim = ToRad((float)get_roll_channel().get_control_in()*0.01f);
-    float pitch_trim = ToRad((float)get_pitch_channel().get_control_in()*0.01f);
+    roll_trim = ToRad((float)get_roll_channel().get_control_in()*0.01f);
+    pitch_trim = ToRad((float)get_pitch_channel().get_control_in()*0.01f);
+#if AP_COPTER_AHRS_AUTO_TRIM_ENABLED    
+    }
+#endif
     AP::ahrs().add_trim(roll_trim, pitch_trim);
     LOGGER_WRITE_EVENT(LogEvent::SAVE_TRIM);
     gcs().send_text(MAV_SEVERITY_INFO, "Trim saved");
@@ -731,6 +744,10 @@ void RC_Channels_Copter::do_aux_function_ahrs_auto_trim(const RC_Channel::AuxSwi
 {
     switch (ch_flag) {
     case RC_Channel::AuxSwitchPos::HIGH:
+        if (!copter.flightmode->allows_auto_trim()) {
+            gcs().send_text(MAV_SEVERITY_INFO, "AutoTrim not allowed in this mode");
+            break;
+        }
         gcs().send_text(MAV_SEVERITY_INFO, "AutoTrim running");
         // flash the leds
         AP_Notify::flags.save_trim = true;
@@ -764,23 +781,22 @@ void RC_Channels_Copter::auto_trim_run()
         }
 
         // only trim in certain modes:
-        if (copter.flightmode != &copter.mode_stabilize &&
-            copter.flightmode != &copter.mode_althold) {
+        if (!copter.flightmode->allows_auto_trim()) {
             auto_trim_cancel();
             return;
         }
 
         // must be started and stopped mid-air:
         if (copter.ap.land_complete_maybe) {
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING,"Must be flying to use AUTOTRIM");
             auto_trim_cancel();
             return;
         }
+        // calculate roll trim adjustment, divisor set subjectively to give same "feel" as previous RC input method
+        float roll_trim_adjustment = ToRad(copter.attitude_control->get_att_target_euler_cd().x / 2000.0f);
 
-        // calculate roll trim adjustment
-        float roll_trim_adjustment = ToRad((float)get_roll_channel().get_control_in() / 4000.0f);
-
-        // calculate pitch trim adjustment
-        float pitch_trim_adjustment = ToRad((float)get_pitch_channel().get_control_in() / 4000.0f);
+        // calculate pitch trim adjustment, divisor set subjectively to give same "feel" as previous RC input method
+        float pitch_trim_adjustment = ToRad(copter.attitude_control->get_att_target_euler_cd().y / 2000.0f);
 
         // add trim to ahrs object, but do not save to permanent storage:
         AP::ahrs().add_trim(roll_trim_adjustment, pitch_trim_adjustment, false);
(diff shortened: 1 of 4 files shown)


rmackay9:
Hi @Hwurzburg, I'm not sure Loiter works though does it?  I don't think it should work in Loiter..

Hwurzburg:
> Hi @Hwurzburg, I'm not sure Loiter works though does it? I don't think it should work in Loiter..

you are right....I got ahead of myself......removed the bool in Loiter.....will follow up with a mode that uses nav_controller->get_roll and pitch if in Loiter or PosHold modes instead of RC inputs in the auto_trim_run function....thanks for the catch....

rmackay9 (approved):

Hwurzburg:
@rmackay9 I found another bug and fixed it (was not properly saving trim and terminating auto_trim_run)
also got Loiter and Pos_hold working....checked sensitivities for using sticks or letting AP do it and they seem reasonable

ArduCopter/RC_Channel_Copter.cpp
```diff
@@ -633,7 +636,9 @@ bool RC_Channel_Copter::do_aux_function(const AuxFuncTrigger &trigger)
 
 #if AP_COPTER_AHRS_AUTO_TRIM_ENABLED
         case AUX_FUNC::AHRS_AUTO_TRIM:
+            if (copter.flightmode->allows_auto_trim()) {
             copter.g2.rc_channels.do_aux_function_ahrs_auto_trim(ch_flag);
+            }
             break;
```
peterbarker, line 642:
```suggestion
            if (!copter.flightmode->allows_auto_trim()) {
                break;
            }
            copter.g2.rc_channels.do_aux_function_ahrs_auto_trim(ch_flag);
            break;
```

... means the indenting isn't broken

ArduCopter/RC_Channel_Copter.cpp
```diff
@@ -717,9 +722,19 @@ void RC_Channel_Copter::do_aux_function_change_force_flying(const AuxSwitchPos c
 // save_trim - adds roll and pitch trims from the radio to ahrs
 void RC_Channels_Copter::save_trim()
 {
+    float roll_trim = 0;
+    float pitch_trim = 0;
+#if AP_COPTER_AHRS_AUTO_TRIM_ENABLED
+    if (auto_trim.running){
```
peterbarker, line 728:
```suggestion
    if (auto_trim.running) {
```

ArduCopter/RC_Channel_Copter.cpp
```diff
@@ -764,24 +779,25 @@ void RC_Channels_Copter::auto_trim_run()
...
         // must be started and stopped mid-air:
         if (copter.ap.land_complete_maybe) {
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING,"Must be flying to use AUTOTRIM");
             auto_trim_cancel();
             return;
         }
-
         // calculate roll trim adjustment
         float roll_trim_adjustment = ToRad((float)get_roll_channel().get_control_in() / 4000.0f);
+        roll_trim_adjustment = ToRad(float(copter.attitude_control->get_att_target_euler_cd().x / 2000.0f));
```
peterbarker, line 795: This makes no sense - you're just blatting the previously assigned value.

peterbarker (changes requested):
Great to see us using the controllers to adjust the trim rather than the sticks!

OTOH - we really should make sure that we're using a position-controlled mode before we actually use the attitude control to get the angles... also need to work out whether it is one or the other or a combination of them.  For simplicity maybe say that in a position-controlled mode we only use the controller, for stab/alt-hold we use sticks?

ArduCopter/RC_Channel_Copter.cpp
Hwurzburg: yep...missed removing that...was there in case it didn't work I would not have to remember what current was

ArduCopter/RC_Channel_Copter.cpp
Hwurzburg: done

Hwurzburg:
@peterbarker "OTOH - we really should make sure that we're using a position-controlled mode before we actually use the attitude control to get the angles"
it works fine in any mode, stick input or AP position controller driven.....its from the attitude controller....SITL tested autotrim correction of AHRS_TRIM back to zero(real world it might be non-zero due to frame,CG,or motor imperfections) in STABILIZE/ALT_HOLD using sticks, and in LOITER/POSHOLD hands off....

ArduCopter/RC_Channel_Copter.cpp
```diff
@@ -764,24 +780,23 @@ void RC_Channels_Copter::auto_trim_run()
...
-
         // calculate roll trim adjustment
-        float roll_trim_adjustment = ToRad((float)get_roll_channel().get_control_in() / 4000.0f);
+        float roll_trim_adjustment = ToRad(float(copter.attitude_control->get_att_target_euler_cd().x / 2000.0f));
 
         // calculate pitch trim adjustment
-        float pitch_trim_adjustment = ToRad((float)get_pitch_channel().get_control_in() / 4000.0f);
-
+        float pitch_trim_adjustment = ToRad(float(copter.attitude_control->get_att_target_euler_cd().y / 2000.0f));
+
```
peterbarker, line 799: whitespace

ArduCopter/RC_Channel_Copter.cpp
```diff
@@ -633,7 +636,10 @@ bool RC_Channel_Copter::do_aux_function(const AuxFuncTrigger &trigger)
 
 #if AP_COPTER_AHRS_AUTO_TRIM_ENABLED
         case AUX_FUNC::AHRS_AUTO_TRIM:
+            if (!copter.flightmode->allows_auto_trim()) {
+                break;
             copter.g2.rc_channels.do_aux_function_ahrs_auto_trim(ch_flag);
+            }
```
peterbarker, line 642:
```suggestion
            }
            copter.g2.rc_channels.do_aux_function_ahrs_auto_trim(ch_flag);
```

peterbarker (changes requested):

ArduCopter/RC_Channel_Copter.cpp
Hwurzburg: fixed

ArduCopter/RC_Channel_Copter.cpp
Hwurzburg: oops...done

peterbarker:
```
        AHRS_AUTO_TRIM =     182,  // in-flight trim Copter AHRS using manual levelling
```
Needs fixing

ArduCopter/RC_Channel_Copter.cpp
```diff
@@ -764,23 +780,22 @@ void RC_Channels_Copter::auto_trim_run()
...
         // must be started and stopped mid-air:
         if (copter.ap.land_complete_maybe) {
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING,"Must be flying to use AUTOTRIM");
             auto_trim_cancel();
             return;
         }
-
         // calculate roll trim adjustment
-        float roll_trim_adjustment = ToRad((float)get_roll_channel().get_control_in() / 4000.0f);
+        float roll_trim_adjustment = ToRad(float(copter.attitude_control->get_att_target_euler_cd().x / 2000.0f));
```
IamPete1, line 795:
```suggestion
        float roll_trim_adjustment = ToRad(copter.attitude_control->get_att_target_euler_cd().x / 2000.0f);
```
Its float already.

Why the / 2000?

ArduCopter/RC_Channel_Copter.cpp
```diff
@@ -633,6 +636,9 @@ bool RC_Channel_Copter::do_aux_function(const AuxFuncTrigger &trigger)
 
 #if AP_COPTER_AHRS_AUTO_TRIM_ENABLED
         case AUX_FUNC::AHRS_AUTO_TRIM:
+            if (!copter.flightmode->allows_auto_trim()) {
+                break;
+            }
```
IamPete1, line 641: This should be down in `do_aux_function_ahrs_auto_trim`, just to keep all the logic on one spot.

ArduCopter/RC_Channel_Copter.cpp
Hwurzburg: stick feel and loiter convergence time the same

ArduCopter/RC_Channel_Copter.cpp
Hwurzburg: done

ArduCopter/RC_Channel_Copter.cpp
Hwurzburg: fixed the extra float

peterbarker: > stick feel (and now loiter convergence time) the same as before...it is subjective as before

Since Pete and I asked exactly the same question, it's almost certainly worthy of a comment.

peterbarker (approved):
I tested this

rmackay9 (approved):

PR #29343, opened 2025-02-20
AP_HAL_ChibiOS_hwdef : Modify the LED on the SULILGH7-P1-P2 flight controller

SULILG:
The define HAL_HAVE_PIXRACER_LED works fine in my local Copter 4.5.7 branch, but when I switch to the master branch, it fails to compile and prompts me to replace it with define AP_NOTIFY_GPIO_LED_RGB_ENABLED 1.


peterbarker:
Commit message is mildly off (should be `hwdef:` or `AP_HAL_ChibiOS:`, but near enough :-)

PR #29345, opened 2025-02-20
Create and use AP_PERIPH_SERIAL_OPTIONS_ENABLED

shiv-tyagi:


peterbarker (changes requested):
This PR is changing the size of several peripherals, e.g.

```
BotBloxDroneNet,-1152
```

PR #29352, opened 2025-02-22
RC_Channel: rename rc option to AutoTrim

rmackay9:
This renames the RCx_OPTION parameter description for value 182 from "In flight AHRS trim adjust" to "AHRS AutoTrim".  Either name is fine of course but this makes it consistent with [what we call it on the wiki](https://ardupilot.org/copter/docs/autotrim.html).  We don't need two names for the same feature

As suggested by @IamPete1 below I've also updated the build option description to match


IamPete1:
The name was taken from the build option here:

https://github.com/ArduPilot/ardupilot/blob/297ec3fa7c6e35300c12057d47338b6ad7bb009b/Tools/scripts/build_options.py#L161

No problem with the rename, but we should sync it up there too.

rmackay9:
Only one approval but I think it's OK to merge this

PR #29354, opened 2025-02-22
AP_Filesystem: LittleFS: clarify fsync bytes logic

tpwrules:
Removes some impossible cases and better documents it.

Tested on KakuteH7Mini-Nand that there are no performance regressions or misaligned fsyncs.

### DIFF
--- a/libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
+++ b/libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
@@ -421,10 +421,37 @@ uint32_t AP_Filesystem_FlashMemory_LittleFS::bytes_until_fsync(int fd)
         return 0;
     }
 
-    uint32_t write_amt = fs_cfg.block_size;
+    uint32_t file_pos = fp->file.pos;
+    uint32_t block_size = fs_cfg.block_size;
+
     // calculate how much allowed if a full block would be written
-    sync_block(fd, fp->file.pos, write_amt);
-    return write_amt; // return that amount
+    uint32_t nbytes = fs_cfg.block_size;
+
+    // see https://github.com/littlefs-project/littlefs/issues/564#issuecomment-2363032827
+    // n = (N − w/8 ( popcount( N/(B − 2w/8) − 1) + 2))/(B − 2w/8))
+    // off = N − ( B − 2w/8 ) n − w/8popcount( n )
+#define BLOCK_INDEX(N, B) \
+    (N - sizeof(uint32_t) * (__builtin_popcount(N/(B - 2 * sizeof(uint32_t)) -1) + 2))/(B - 2 * sizeof(uint32_t))
+
+#define BLOCK_OFFSET(N, B, n) \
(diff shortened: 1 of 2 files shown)


andyp1per:
You'll need to walk me through this a bit more to convince me as there are a few things that have changed.

tpwrules:
I've split up the commits and put a lot more clarification in the commit messages.

I tested by making sure the same number of blocks per second were read and written, and that there were no misaligned fsyncs (a diagnostic I added a bit ago).

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -421,10 +421,37 @@ uint32_t AP_Filesystem_FlashMemory_LittleFS::bytes_until_fsync(int fd)
...
+#define BLOCK_INDEX(N, B) \
+    (N - sizeof(uint32_t) * (__builtin_popcount(N/(B - 2 * sizeof(uint32_t)) -1) + 2))/(B - 2 * sizeof(uint32_t))
+
+#define BLOCK_OFFSET(N, B, n) \
+    (N - (B - 2*sizeof(uint32_t)) * n - sizeof(uint32_t) * __builtin_popcount(n))
+
+    uint32_t block_index = BLOCK_INDEX(file_pos, block_size);
+    uint32_t block_offset = BLOCK_OFFSET(file_pos, block_size, block_index);
+    if (block_size - block_offset <= nbytes) {
+        if (block_size == block_offset) {
```
andyp1per: So in this path block_size is returned rather than false for no sync, I don't see how the previous PR can have been correct if this is correct.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -614,34 +641,6 @@ void AP_Filesystem_FlashMemory_LittleFS::mark_dead()
...
-#define BLOCK_OFFSET(N, B, n) \
-    (N - (B - 2*sizeof(uint32_t)) * n - sizeof(uint32_t) * __builtin_popcount(n))
-
-    uint32_t blocksize = fs_cfg.block_size;
-    uint32_t block_index = BLOCK_INDEX(_write_offset, blocksize);
-    uint32_t block_offset = BLOCK_OFFSET(_write_offset, blocksize, block_index);
-    if (blocksize - block_offset <= nbytes) {
-        if (blocksize == block_offset) {
-            // exactly at the end of the block, sync and then write all the data
-            AP::FS().fsync(_write_fd);
```
andyp1per, line 634: nbytes is unset here but the return value is ignored

andyp1per (changes requested):
Need to understand that the logic change the previous PR made was in fact correct

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
tpwrules: That case can never occur, the fsync is never executed.

andyp1per (approved):

PR #29356, opened 2025-02-23
Plane: fence: refactor action cases

IamPete1:
There is no need for the guided and RTL cases to share,  they are independent. This simplifies https://github.com/ArduPilot/ardupilot/pull/29280


tridge (approved):

tridge:
a bit of testing of the cases, with rally pt, ok to merge when tested

IamPete1:
rebased, tested a fence breaches with both RTL and guided. Behaved as expected with and without fence return point, with and without rally points.

PR #23832, opened 2023-05-19
Copter: Add GCS Failsafe Brake option

tatsuy:
I've added a "Brake Mode" option, which allows the copter to hover in place (in Brake Mode) if the connection to the GCS is lost while in Auto or Guided flight modes.

When conducting BVLOS operations, it is crucial to maintain control over the drone via the GCS at all times. In the event of a GCS disconnection, the drone switches to Brake mode, halting flight without GCS control. Upon reestablishing the connection, the drone can be returned to Auto mode via the GCS.

In the case where the connection does not recover, the drone will take RTL or Land actions through a battery failsafe.

This feature follows a similar implementation as in the PR #23110, which added a Brake Mode option to the Radio Failsafe.

I tested this in SITL.
image

PR #23110, opened 2023-03-06
Copter: Add Radio Failsafe Brake Mode option

lthall:
This PR provides an option to change mode to Break Mode in the event of a Radio Failsafe. If GPS is not available a no-gps land is done.

This is required for some industrial operations where the aircraft is connected to tethers and carrying a load. In this case an RTL or payload release could both be dangerous. By changing to Break Mode the operators have the remaining battery capacity to solve the problem. This may be achieved by moving the pilot radio to a better line of sight position, moving to GCS controlled operations, or simply clearing the area of personnel and detaching or securing tethers.

PR #29361, opened 2025-02-23
AP_BattMonitor: add Copter brake/land failsafe action to parameter me…

peterbarker:
…tadata


rmackay9:
Thanks for this.  Just for accountability it looks like:

1. Leonard added the feature for RC failsafe here https://github.com/ArduPilot/ardupilot/pull/23110
2. Tatsuya noticed it was missing for GCS failsafe here https://github.com/ArduPilot/ardupilot/pull/23832
3. PeterB's now noticed it's missing for the battery failsafe

rmackay9 (approved):

lthall:
Yeh, that was my bad. In my defince I noticed the missing battery failsafe and asked Peter if it should have also been added to the battery.

When I read through the code it looks like the failsafes are more seperated than they really are.  What confused me is the hash defines:
#define FS_THR_ENABLED_ALWAYS_RTL
#define FS_GCS_ENABLED_ALWAYS_RTL

And then there are no battery ones at all. It took some going back and forth before I felt confident it was missing.

rmackay9:
I was unable to immediately include this in 4.6.0-beta5 due to merge conflicts

PR #29363, opened 2025-02-24
In RTL, set wp velocity after controller is initialised

peterbarker:
Alternative to https://github.com/ArduPilot/ardupilot/pull/29285/commits/1b29468a89b2a6ebe5cfda9954a0c5f8b05d9489

```
When mode_rtl init() is called, don't explicitly call wp_nav->wp_and_spline_init(g.rtl_speed_cms). I think this is redundant anyway since as soon as we pass the first wapoint to AC_WPNav, it will call wp_and_spline_init on it's own anyway. Critically, we don't send WPNav the RTL speed at this point, because this seems to be crux of the issue. If we pass WPNav a slow speed here then the vehicle always does a harsh stop, regardless of how far away the stopping point actually is.

To actually get the RTL speed to work, we instead set it explicitly ONLY after the vehicle has started the return back home (in return_start()).
```

Not my patch - I need to add the correct author on it yet.

### DIFF
--- a/ArduCopter/mode_rtl.cpp
+++ b/ArduCopter/mode_rtl.cpp
@@ -17,8 +17,7 @@ bool ModeRTL::init(bool ignore_checks)
             return false;
         }
     }
-    // initialise waypoint and spline controller
-    wp_nav->wp_and_spline_init(g.rtl_speed_cms);
+    // do not initialise waypoint and spline controller yet, controller will initialize itself when the first waypoint is passed
     _state = SubMode::STARTING;
     _state_complete = true; // see run() method below
     terrain_following_allowed = !copter.failsafe.terrain;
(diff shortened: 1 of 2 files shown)


rmackay9:
Hi, thanks for this.  I think this is probably not the correct solution though.  The expected use of the navigation controller is that init is called first.  So, I'm sure you're right that this resolves the issue but it's not the way that we should resolve it

peterbarker:
> Hi, thanks for this. I think this is probably not the correct solution though. The expected use of the navigation controller is that init is called first. So, I'm sure you're right that this resolves the issue but it's not the way that we should resolve it

Perhaps, then, we should raise an internal error if `get_wp_destination_loc` (or the spline equivalent) is called and decides *it* needs to do the initialisation?

rmackay9:
@peterbarker,

Sure, that could be a good idea.  The problem may be quadplane that doesn't always follow the rules

PR #29365, opened 2025-02-24
AP_HAL: Socket: check return value of getsockname

peterbarker:
avoids static analysis failure upon this method failing


tridge (approved):
needs testing with a NET_PORT as UDP broadcast

PR #29366, opened 2025-02-24
Fix: Potential Vulnerability in Cloned Function

tabudz:
**Description**
This PR fixes a security vulnerability in singlevar() that was cloned from lua but did not receive the security patch. The original issue was reported and fixed under https://github.com/lua/lua/commit/1f3c6f4534c6411313361697d98d1145a1f030fa.
This PR applies the same patch to eliminate the vulnerability.

**References**
https://nvd.nist.gov/vuln/detail/CVE-2022-28805
https://github.com/lua/lua/commit/1f3c6f4534c6411313361697d98d1145a1f030fa

### DIFF
--- a/libraries/AP_Scripting/lua/src/lparser.c
+++ b/libraries/AP_Scripting/lua/src/lparser.c
@@ -300,6 +300,7 @@ static void singlevar (LexState *ls, expdesc *var) {
     expdesc key;
     singlevaraux(fs, ls->envn, var, 1);  /* get environment variable */
     lua_assert(var->k != VVOID);  /* this one must exist */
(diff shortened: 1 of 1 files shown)


tpwrules:
We are still on Lua 5.3.6. It does not have the `<const>` feature and the CVE indicates it is not vulnerable.

Do you have reason to suspect this will affect us?

tabudz:
I ran a code clone detection tool and found that this function might be a clone of a vulnerable one in my database. Just wanted to give you a heads-up. If you're confident that your case isn't vulnerable, then it likely isn't. Thanks for the clarification! Closed the PR.

IamPete1:
Thanks for the report, were safe on 5.3 for now.

PR #29368, opened 2025-02-24
Scripted CRSF Menus

andyp1per:
Work kindly sponsored by Team Black Sheep and FOSS UAV

This PR adds the ability to create arbitrary custom menus in CRSF using lua scripting. This allows arbitrary scripted commands, including with user feedback, to be driven by transmitter menu interactions without requiring any additional installation on the transmitter. It opens up the possibility of allowing people to add arbitrary menus to individual lua scripts that can be the used easily from a transmitter. (I think this is quite a usability win fyi @rmackay9).
As much of the encoding as possible is done in lua to avoid the additional flash cost, but the menu structure itself still needs to be help in C++ in order to work correctly and efficiently. 

An example menu can be seen in libraries/AP_Scripting/examples/crsf-menu.lua
and the output can be seen below.

image
image
image
image
image


rmackay9:
just FYI, I added this as a bullet point on the Partners slides for this month

libraries/AP_RCProtocol/AP_RCProtocol_CRSF.cpp
```diff
@@ -461,7 +463,24 @@ bool AP_RCProtocol_CRSF::decode_crsf_packet()
...
+#ifdef CRSF_DEBUG_TELEM
+        switch (_frame.type) {
+            case CRSF_FRAMETYPE_RC_CHANNELS_PACKED:
+            case CRSF_FRAMETYPE_LINK_STATISTICS:
+            case CRSF_FRAMETYPE_SUBSET_RC_CHANNELS_PACKED:
+            case CRSF_FRAMETYPE_LINK_STATISTICS_RX:
+            case CRSF_FRAMETYPE_LINK_STATISTICS_TX:
+                break;
+            default:
+                hal.console->printf("CRSF: received %s:", get_frame_type(_frame.type));
```
peterbarker, line 476:
```suggestion
                GCS_SEND_TEXT(MAV_SEVERITY_INFO, "CRSF: unknown frame %s:", get_frame_type(_frame.type));
```

libraries/AP_RCProtocol/AP_RCProtocol_CRSF.cpp
```diff
@@ -461,7 +463,24 @@ bool AP_RCProtocol_CRSF::decode_crsf_packet()
...
+                uint8_t* fptr = (uint8_t*)&_frame;
+                for (uint8_t i = 0; i < _frame.length + 2; i++) {
+                    hal.console->printf(" 0x%x", fptr[i]);
+                }
+                hal.console->printf("\n");
```
peterbarker, line 481: Don't want this in generally

.... also, why _frame.length+2 ?  Something to do with checksums?  Why are we subtracting 2 bytes above, then?

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
```diff
@@ -757,8 +759,18 @@ void AP_CRSF_Telem::process_param_read_frame(ParameterSettingsReadFrame* read_fr
     if (read_frame->destination != 0 && read_frame->destination != AP_RCProtocol_CRSF::CRSF_ADDRESS_FLIGHT_CONTROLLER) {
         return; // request was not for us
     }
+#if AP_CRSF_SCRIPTING
```
peterbarker, line 762:
```suggestion
#if AP_CRSF_SCRIPTING_ENABLED
```

.. and this should be an option in `build_options.py`

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
```diff
@@ -757,8 +759,18 @@ void AP_CRSF_Telem::process_param_read_frame(ParameterSettingsReadFrame* read_fr
...
+    // write parameter in scripted menus
+    if (process_scripted_param_read(read_frame)) {
+        return;
+    }
+#endif
```
peterbarker, line 767:
```suggestion
#endif  // AP_CRSF_SCRIPTING_ENABLED
```

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
```diff
@@ -1168,9 +1187,45 @@ void AP_CRSF_Telem::calc_parameter() {
...
+
+#if AP_CRSF_SCRIPTING
+    // root folder request
+    if (scripted_menus.next_menu != nullptr && _param_request.param_num == PARAMETER_MENU_ID) {
+        _telem.ext.param_entry.header.param_num = PARAMETER_MENU_ID;
+        _telem.ext.param_entry.header.chunks_left = 0;
+        _telem.ext.param_entry.payload[idx++] = 0; // parent folder
+        _telem.ext.param_entry.payload[idx++] = ParameterType::FOLDER; // type
+        // name
+        strncpy((char*)&_telem.ext.param_entry.payload[idx], "Parameters", 16);
```
peterbarker, line 1223:
```suggestion
        strncpy((char*)&_telem.ext.param_entry.payload[idx], "Parameters", ARRAY_SIZE(_telem.ext.param_entry.payload) - idx);
```

... and I'm otherwise concerned with indexing into the payload here without checking that idx isn't walking off the end of the array.  We'd want a really big comment in here explaining how we don't overwrite this buffer...

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
```diff
@@ -1181,60 +1236,122 @@ void AP_CRSF_Telem::calc_parameter() {
...
+    if (menu != nullptr) {
+        debug("menu request %d -> %s[%d]", _param_request.param_num, menu->name, menu->num_params);
+
+        _telem.ext.param_entry.header.param_num = _param_request.param_num;
+        _telem.ext.param_entry.header.chunks_left = 0;  // assume menus fit in a single chunk
+
+        _telem.ext.param_entry.payload[idx++] = menu->parent_id; // parent folder
+        _telem.ext.param_entry.payload[idx++] = ParameterType::FOLDER; // type
+        // name
+        strncpy((char*)&_telem.ext.param_entry.payload[idx], menu->name, 16);
```
peterbarker, line 1251: avoid use of constant 16 here

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
```diff
@@ -1181,60 +1236,122 @@ void AP_CRSF_Telem::calc_parameter() {
...
+#endif // AP_CRSF_SCRIPTING
+
     AP_OSD* osd = AP::osd();
 
     if (osd == nullptr) {
         return;
     }
 
-    AP_OSD_ParamSetting* param = osd->get_setting((_param_request.param_num - 1) / AP_OSD_ParamScreen::NUM_PARAMS,
+    AP_OSD_ParamSetting* setting = osd->get_setting((_param_request.param_num - 1) / AP_OSD_ParamScreen::NUM_PARAMS,
```
peterbarker, line 1307: Pull this name change out as a separate PR and we can get it merged straight away...

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
```diff
@@ -1505,6 +1628,283 @@ void AP_CRSF_Telem::process_param_write_frame(ParameterSettingsWriteFrame* write
...
+
+bool AP_CRSF_Telem::send_write_response(uint8_t length, const char* data)
+{
+    ScriptedParameterWrite spw;
+    if (outbound_params.pop(spw)) {
+        switch (spw.type) {
+        case ScriptedParameterEvents::PARAMETER_READ:
+            // respond with parameter entry updated with the new data
+            spw.param->data = (const char*)hal.util->std_realloc((char*)spw.param->data, length);
+            memcpy((char*)spw.param->data, data, length);
```
peterbarker, line 1672: memcpy's second argument should generally by the length of the first argument, not the second argument!

tridge (dismissed):
a more fleshed out example would be good
suggestion is ability to call an auxiliary function
compass cal?
accel cal?

peterbarker (dismissed):
Would be nice to augment the existing CRSF test to cover some of this code

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
tpwrules: This shouldn’t need to be a realloc either. That function is on my chopping block, please free the old pointer then malloc a new one.

libraries/AP_RCProtocol/AP_RCProtocol_CRSF.cpp
andyp1per: This is deliberate and is compiled in debug, the frame is not unknown - we just want to see it.

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
andyp1per: The name change is because there is another variable with the same name and they need to be different

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
andyp1per: ... at an increase in flash cost, so I define a function to do it and we are back where we started. The future removal of something should not be a precondition for a PR - we can discuss separately. I *do* need to reallocate the memory to a difference size and realloc() is the standard way of doing that.

@peterbarker I don't get your comment - we have just reallocated the target to match the incoming length

libraries/AP_Scripting/AP_Scripting_CRSFMenu.cpp
```diff
@@ -0,0 +1,137 @@
...
+int lua_CRSF_new_menu(lua_State *L)
+{
+    const int args = lua_gettop(L);
+    if (args != 1) {
+        return luaL_argerror(L, args, "wrong number of arguments");
+    }
+    const char * name = luaL_checkstring(L, 1);
+    void *ud = lua_newuserdata(L, sizeof(CRSFMenu));
+
+    AP_CRSF_Telem::ScriptedMenu* menu = AP::crsf_telem()->add_menu(name);
```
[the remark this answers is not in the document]
andyp1per: Fixed

libraries/AP_Scripting/AP_Scripting_CRSFMenu.cpp
```diff
@@ -0,0 +1,137 @@
...
+    lua_setmetatable(L, -2);
+
+    return 1;
+}
+
+int lua_CRSF_get_menu_event(lua_State *L)
+{
+    binding_argcheck(L, 2);
+    const uint8_t raw_events = get_uint8_t(L, 2);
+    const uint8_t events = static_cast<uint8_t>(raw_events);
```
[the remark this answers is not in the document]
andyp1per: fixed

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
andyp1per: Fixed thanks

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
andyp1per: Fixed thanks

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
andyp1per: Fixed

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
andyp1per: Doesn't work:
```
In function ‘char* strncpy(char*, const char*, size_t)’,
    inlined from ‘void AP_CRSF_Telem::calc_parameter()’ at ../../libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp:1252:16:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:95:34: error: ‘char* __builtin_strncpy(char*, const char*, long unsigned int)’ specified bound 54 equals destination size [-Werror=stringop-truncation]
   95 |   return __builtin___strncpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   96 |                                   __glibc_objsize (__dest));
```

libraries/AP_RCProtocol/AP_RCProtocol_CRSF.cpp
```diff
@@ -447,6 +448,7 @@ bool AP_RCProtocol_CRSF::decode_crsf_packet()
             process_link_stats_frame((uint8_t*)&_frame.payload);
             break;
         case CRSF_FRAMETYPE_SUBSET_RC_CHANNELS_PACKED:
+        //case CRSF_FRAMETYPE_RC_CHANNELS_PACKED_11BIT:
```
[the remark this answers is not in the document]
andyp1per: Its an unused frame type for which there is no documentation. We are better off ignoring it,

libraries/AP_Scripting/AP_Scripting_CRSFMenu.cpp
```diff
@@ -0,0 +1,140 @@
...
+{ 
+    return menu->add_menu(menu_name, 2, menu->id);
+}
+
+int lua_CRSF_new_menu(lua_State *L)
+{
+    const int args = lua_gettop(L);
+    if (args != 1) {
+        return luaL_argerror(L, args, "wrong number of arguments");
+    }
```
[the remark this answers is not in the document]
andyp1per: Fixed thanks

libraries/AP_Scripting/AP_Scripting_CRSFMenu.cpp
```diff
@@ -0,0 +1,140 @@
...
+    luaL_checkstack(L, 3, nullptr);
+#endif
+
+    lua_pushinteger(L, param);
+    lua_pushlstring(L, (const char*)payload.payload, payload.payload_length);
+    lua_pushinteger(L, data);
+    return 3;
+}
+
+int lua_CRSF_send_response(lua_State *L)
```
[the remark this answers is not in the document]
andyp1per: No,  the subtlety is the requirement for ```luaL_checklstring```

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4101,3 +4101,67 @@ function AP_Servo_Telem_Data_ud:measured_position() end
 -- get commanded position
 ---@return number|nil -- comanded position in degrees or nil if not available
 function AP_Servo_Telem_Data_ud:command_position() end
+-- desc
```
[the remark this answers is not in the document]
andyp1per: Done thanks

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
andyp1per: Fixed thanks

andyp1per:
Added a crsf-calibrate.lua applet

IamPete1:
Scripting stuff looks OK.

PR #29369, opened 2025-02-24
hwdef: correct compilation of SULILGH7-P1-P2

peterbarker:
This didn't compile when the board was committed to the repository.

@SULILG - is this the correct value for the LED - pin high for LED-is-on?


Hwurzburg:
@peter I really cant help on this one...I dont have one of these....only @SULILG  can comment...

PR #29371, opened 2025-02-25
remove Python print_function import

peterbarker:
won't break Python2, not needed in Python3, just remove the noise

### DIFF
--- a/Tools/Replay/check_replay.py
+++ b/Tools/Replay/check_replay.py
@@ -4,8 +4,6 @@
 check that replay produced identical results
 '''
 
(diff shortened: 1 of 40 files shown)


tpwrules:
To be clear I absolutely support breaking python 2.

peterbarker:
> To be clear I absolutely support breaking python 2.

Ouch.  Vicious! :-)

amilcarlucas:
Use ruff, it can do all these nice things of deprecating and removing old suff.

amilcarlucas (approved):

PR #29375, opened 2025-02-25
AP_Proximity: LD19/LD06 data length sanity check fix

boa-pe:
AP_Rangefinder: LD19/LD06 data length sanity check fix

root cause: number of data points in frame stored only on 5bits. value been used without proper bitmask
to sanity check number of datapoints in frame

### DIFF
--- a/libraries/AP_Proximity/AP_Proximity_LD06.cpp
+++ b/libraries/AP_Proximity/AP_Proximity_LD06.cpp
@@ -107,8 +107,8 @@ void AP_Proximity_LD06::get_readings()
             }
 
             // total_packet_length = sizeof(header) + datalength + sizeof(footer):
-            const uint32_t total_packet_length = 6 + 3*_response[START_DATA_LENGTH] + 5;
-            if (_response[START_DATA_LENGTH] != PAYLOAD_COUNT ||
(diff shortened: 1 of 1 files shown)


rmackay9:
Hi @boa-pe,

thanks for this.  one small thing, could you modify the commit title to start with "AP_Rangefinder: "?  We do this because it makes backporting easier because we can immediately see which subsystem a change affects.  To be clear I mean the commit title, not the PR title. thanks!

boa-pe:
> Hi @boa-pe,
> 
> thanks for this. one small thing, could you modify the commit title to start with "AP_Rangefinder: "? We do this because it makes backporting easier because we can immediately see which subsystem a change affects. To be clear I mean the commit title, not the PR title. thanks!

Hmmm.... Shouldn't that be AP_Proximity?

peterbarker:
> Hmmm.... Shouldn't that be AP_Proximity?

Yes, it definitely should be :-)

rmackay9:
oops :-)

peterbarker (approved):
Tested on my LD06 and it works!

Many thanks for chasing this one down properly.  I should have gone to the datasheet :-/

I've fixed the commit message and marking for merge.

PR #29376, opened 2025-02-25
Plane: climb directly to cruise altitude ignoring glide slope

magate:
Add the ability for quadplanes to climb immediately to the cruise altitude as opposed to using the glide slope calculation.

### Feature disabled under 35 m cutoff altitude
image

### Feature enabled over 35 m cutoff altitude
image

### Example of cruise altitude cutoff
image


peterbarker:
Our existing FlightOptions bit, `FlightOptions::IMMEDIATE_CLIMB_IN_AUTO` isn't suitable for your purposes?

It's bit 14 here: https://ardupilot.org/plane/docs/parameters.html#flight-options-flight-mode-options

magate:
> Our existing FlightOptions bit, `FlightOptions::IMMEDIATE_CLIMB_IN_AUTO` isn't suitable for your purposes?
> 
> It's bit 14 here: https://ardupilot.org/plane/docs/parameters.html#flight-options-flight-mode-options

This will change the behavior for all climbs.

Looks very similar to `GLIDE_SLOPE_MIN` as well except glide_slope_min works both directions (climb and descent).

The PR I have here is just for VTOL Takeoff --> WP. All other waypoints still do the glide slope.

See below where flight options and glide slope min reach WP altitude prior to WP and level off (for both WPs. This PR only reaches WP altitude prior to the WP for the first WP after the VTOL Takeoff.

## Flight options bit 14
image

## Glide slope min zero
image

## This PR Q_FR_CLMBOUT 1
image

ArduPlane/commands_logic.cpp
```diff
@@ -613,6 +613,31 @@ bool Plane::verify_takeoff()
...
+    // If the previous command is takeoff we will do the full rate climbout.
+    if (prev_read_success && quadplane.is_vtol_takeoff(prev_cmd.id)) {
+        quadplane.set_do_full_rate_climbout(true);
+        return;
+    }
+
+    // Default to off.
+    quadplane.set_do_full_rate_climbout(false);
+}
+#endif
```
peterbarker, line 639:
```suggestion
#endif  // HAL_QUADPLANE_ENABLED
```

ArduPlane/commands_logic.cpp
```diff
@@ -613,6 +613,31 @@ bool Plane::verify_takeoff()
...
+  calculating the glide slope based on proportion to the waypoint.
+ */
+#if HAL_QUADPLANE_ENABLED
+void Plane::maybe_do_full_rate_climbout()
+{
+    AP_Mission::Mission_Command prev_cmd;
+
+    // Check the previous mission item.
+    bool prev_read_success = (plane.mission.read_cmd_from_storage(
+        plane.mission.get_current_nav_index() - 1, prev_cmd));
```
peterbarker, line 628: This is most definitely not reliable.  You may have gotten here with a jump of some description.

Please evaluate `get_prev_nav_cmd_id` instead.

ArduPlane/commands_logic.cpp
```diff
@@ -613,6 +613,31 @@ bool Plane::verify_takeoff()
     }
 }
 
+/*
+  When enabled, if the previous waypoint is VTOL Takeoff set the flag to do the full rate climbout.
+  This will give a climbing transition and reach the cruise waypoint immediately as opposed to
+  calculating the glide slope based on proportion to the waypoint.
+ */
+#if HAL_QUADPLANE_ENABLED
+void Plane::maybe_do_full_rate_climbout()
```
peterbarker, line 622: This would probably be neater as `bool Plane::should_do_full_rate_climbout() const` as the early-return when you don't want to do it is trivial.  The caller can `quadplane.set_do_full_rate_climbout(should_do_full_rate_climbout));

Should save a few bytes too.

ArduPlane/mode.cpp
```diff
@@ -201,7 +201,22 @@ void Mode::update_target_altitude()
     } else if (plane.target_altitude.offset_cm != 0 && 
                !plane.current_loc.past_interval_finish_line(plane.prev_WP_loc, plane.next_WP_loc)) {
         // control climb/descent rate
+#if HAL_QUADPLANE_ENABLED
+        // On climbout we will give the ability to do a full rate climb after the VTOL Takeoff. We
+        // will set the glide slope proportion to the destination to achieve this.
+        if (quadplane.get_do_full_rate_climbout_enable() && quadplane.get_do_full_rate_climbout()) {
```
peterbarker, line 207: fold the first part of this expression into the method I proposed?

ArduPlane/quadplane.h
```diff
@@ -396,6 +401,11 @@ class QuadPlane
     // limit applied to back pitch to prevent wing producing excessive lift
     AP_Float q_bck_pitch_lim;
 
+    // This flag controls the ability to disable the glide slope and instead climbout at full rate
+    // once the VTOL Takeoff waypoint altitude is achieved.
+    AP_Int8 enable_full_rate_climbout;
+    bool _do_full_rate_climbout;  // set by the code
```
peterbarker, line 407:
```suggestion
    bool _do_full_rate_climbout;  // set by the code to indicate full-rate-climbout should be done
```

ArduPlane/mode.cpp
```diff
@@ -201,7 +201,22 @@ void Mode::update_target_altitude()
...
+            // Set us to the completed portion to get full rate climb. Here, 0 is completed, see
+            // the docstring for set_target_altitude_proportion.
+            plane.set_target_altitude_proportion(plane.next_WP_loc, 0.0f);
```
peterbarker, line 210: I'm surprised this isn't the same as the Auto code, which does a `reset_offset_altitude`

ArduPlane/mode.cpp
peterbarker, line 207: Why is this quadplane-specific?  Planes do have takeoffs which ytou might want a full-rate climbout?  Those trees can hurt regardless of hte number of props you have.

Tools/autotest/quadplane.py
```diff
@@ -2164,6 +2164,90 @@ def WindEstimateConsistency(self):
...
+        CRUISE_ALTITUDE_CUTOFF = 120
+        WAIT_AFTER_CRUISE_DELAY_S = 10
+
+        # Mission planning items.
+        CRUISE_WP_NUM = 2
+
+        self.progress("Testing full rate climbout.")
+
+        def run_test(flag):
+            filename = "mission.txt"
```
peterbarker, line 2180: Note that a lot of newer tests are using `upload_simple_relhome_mission` now-adays.  It's a bit less work and does keep things together.  It has the disavantage of making it harder to load the mission ino SITL (eg. with `sim_vehicle.py`

Tools/autotest/quadplane.py
```diff
@@ -2164,6 +2164,90 @@ def WindEstimateConsistency(self):
...
+            self.progress("Testing %s with flag %s" % (filename, bool(flag)))
+            self.progress("Setting parameters.")
+
+            self.set_parameter("Q_FR_CLMBOUT", int(flag))
```
peterbarker, line 2184: Technically a float for ArduPilot - at least for transport to the autopilot.



In either case, this cast isn't required

Tools/autotest/quadplane.py
```diff
@@ -2164,6 +2164,90 @@ def WindEstimateConsistency(self):
...
+
+            self.reboot_sitl()
+
+            # Load and fly mission.
+            self.progress(f"Flying mission {filename}")
+            self.load_mission(filename)
+            self.wait_ready_to_arm()
+
+            # Get the starting altitude to convert to agl.
+            starting_alt_m = self.mav.messages['VFR_HUD'].alt
```
peterbarker, line 2194:
```suggestion
            starting_alt_m = self.get_altitude(relative=True)
```

Tools/autotest/quadplane.py
```diff
@@ -2164,6 +2164,90 @@ def WindEstimateConsistency(self):
...
+
+            self.arm_vehicle()
+            self.change_mode('AUTO')
+
+            self.progress("Waiting for Transition done")
```
peterbarker, line 2199: Technically this is racy, a context should be used to collect statustexts.

In this case it's not really a problem - it takes a *long* time to get to the transition, so a user would need an astonishingly fast computer to actually hit the problem.

Tools/autotest/quadplane.py
```diff
@@ -2164,6 +2164,90 @@ def WindEstimateConsistency(self):
...
+            self.wait_statustext("Transition done", timeout=60)
+
+            # Current altitude.
+            altitude_m = self.mav.messages['VFR_HUD'].alt - starting_alt_m
```
peterbarker, line 2203:
```suggestion
            altitude_delta_m = self.get_altitude(relative=True) - starting_alt_m
```

(note the renaming of this variable)

Tools/autotest/quadplane.py
```diff
@@ -2164,6 +2164,90 @@ def WindEstimateConsistency(self):
                 self.progress("Wind estimates correlated")
                 break
 
+    def FullRateClimbout(self):
+        '''test the full rate climbout after VTOL Takeoff functions'''
+        # Set from examining the autotest with enable/disable flag.
+        ALTITUDE_CUTOFF_M = 35
+        CRUISE_ALTITUDE_CUTOFF = 120
+        WAIT_AFTER_CRUISE_DELAY_S = 10
```
peterbarker, line 2172: We usually have constants like this declared and defined just before their first (and typically only...) use.  Saves having to go look stuff up all the time.

Tools/autotest/quadplane.py
```diff
@@ -2164,6 +2164,90 @@ def WindEstimateConsistency(self):
...
+
+        # Mission planning items.
+        CRUISE_WP_NUM = 2
+
+        self.progress("Testing full rate climbout.")
+
+        def run_test(flag):
```
peterbarker, line 2179: Perhaps a better variable name than "flag" here?  :-)

`climbout_enabled`?

Tools/autotest/quadplane.py
```diff
@@ -2164,6 +2164,90 @@ def WindEstimateConsistency(self):
...
+
+            is_over_alt_cutoff = (altitude_m > ALTITUDE_CUTOFF_M)
+
+            # Should be under the cutoff when not doing full rate climb.
+            fail_test = False
+            if not flag and is_over_alt_cutoff:
+                notification_str = "too high"
+                sign_str = ">"
+                fail_test = True
+
```
peterbarker, line 2213: Maybe just call NotAchievedException twice :-)

Also, f-strigns do make these things much more readable.

Tools/autotest/quadplane.py
```diff
@@ -2164,6 +2164,90 @@ def WindEstimateConsistency(self):
...
+            self.delay_sim_time(WAIT_AFTER_CRUISE_DELAY_S, "waypoint altitude change")
+
+            # Check the altitude is under a limit.
+            cruise_altitude_m = self.mav.messages['VFR_HUD'].alt - starting_alt_m
+
+            # Should be under the cutoff when doing cruise climbs.
+            if not (cruise_altitude_m < CRUISE_ALTITUDE_CUTOFF):
+                raise NotAchievedException(
+                    "Cruise altitude too high for feature: %s, test_file: %s: %.2f exceeds %.2f."
+                    % (bool(flag), filename, cruise_altitude_m, CRUISE_ALTITUDE_CUTOFF))
```
peterbarker, line 2242:
```suggestion
        return self.wait_altitude(
            altitude_min=starting_alt_m,
            altitude_max=starting_alt_m + CRUISE_ALTITUDE_CUTOFF,
            relative=True,
            minimum_duration=10,
            timeout=11,
        )
```

peterbarker (changes requested):
Great to see the autotest for the new feature.

Another query - you could get the same effect as this by having a waypoint relatively close to your takeoff point.  I can see a few drawbacks to that - but have you evaluated it as an option and decided this feature was needed?

magate:
Closing because this also does a full rate descent if you are over the WP altitude. Will need to fix.

magate:
> Closing because this also does a full rate descent if you are over the WP altitude. Will need to fix.

Whoops, this is protected.

```if (!above_location_current(next_WP_loc) && (plane.flight_option_enabled(FlightOptions::IMMEDIATE_CLIMB_IN_AUTO) || plane.should_do_full_rate_climbout()))```

PR #29379, opened 2025-02-25
ChibiOS: avoid bdshot fault on null event waiter

### LINKED ISSUE #29370: AP-4.6.0: TradHeli internal errors 0X4000000 and 0X4000020
This issue is to capture the internal errors reported by @Ferruccio1984 during AP-4.6.0-beta3/4 testing ([discussion here](https://discuss.ardupilot.org/t/copter-4-6-0-beta4-released-for-beta-testing/130471/23))

tridge:
This avoids a write to a null pointer in bdshot if the waiter is set to null while a timer is still pending

@andyp1per it would be good to get your input. This fix is more of a bandaid than a real fix

This fixes #29370

### DIFF
--- a/libraries/AP_HAL_ChibiOS/Scheduler.cpp
+++ b/libraries/AP_HAL_ChibiOS/Scheduler.cpp
@@ -623,6 +623,19 @@ void Scheduler::check_low_memory_is_zero()
         }
 #pragma GCC diagnostic pop
     }
+
+#if 0
+    /*
+      enable this on H7 to make writes to the first 1k of RAM on H7
+      produce a hard fault and crash dump
+     */
+    mpuConfigureRegion(MPU_REGION_7,
+                       0x0,
+                       MPU_RASR_ATTR_AP_RO_RO |
+                       MPU_RASR_SIZE_1K |
+                       MPU_RASR_ENABLE);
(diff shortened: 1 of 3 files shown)


tridge:
@andyp1per we need to know when this was introduced too, so we know if this could happen in 4.5.x

libraries/AP_HAL_ChibiOS/Scheduler.cpp
```diff
@@ -623,6 +623,19 @@ void Scheduler::check_low_memory_is_zero()
         }
 #pragma GCC diagnostic pop
     }
+
+#if 0
+    /*
+      enable this on H7 to make writes to the first 1k of RAM on H7
+      produce a hard fault and crash dump
```
tpwrules, line 630:
```suggestion
      enable this on H7 to make writes to the first 1k of RAM on H7
      produce a hard fault and crash dump instead of an internal error.
      do not enable on production aircraft!
```

rmackay9:
I think we can close this now because #29409 has been merged

rmackay9:
Thanks.  BTW, I've added the other PR to the 4.**5** backport project as well (because this one was marked for backporting to 4.5)

PR #29381, opened 2025-02-26
Dynamically generate Linux boards in waf

peterbarker:
Removes the Linux board names from `boards.py`, instead doing as we do for ChibiOS, generating the board classes based on the existence of the hwdef files.

This is a no-compiler output change for Linux targets except for the following:

```
-----------  -----  ------  ----  -----  -----  ---
Board        blimp  copter  heli  plane  rover  sub
erleboard    *      0       *     *      *      *
navigator64  0      0       0     0      0      0
ocpoc_zynq   *      *       *     *      0      *
-----------  -----  ------  ----  -----  -----  ---
```

.... and aero now compiles.

Those "0"s come into elf_diff as no code change, just symbol name changes.


tridge (approved):

tridge:
@peterbarker merge OK when all boards test passes

PR #29389, opened 2025-02-27
 AP_Arming: add a check that AHRS and GPS agree on groundspeed 

peterbarker:


rmackay9:
Hi, no big deal but do we need to make this a new check or can we just include it in the existing GPS check?  I worry that a GPS may be attached but it may not be being used by the EKF.  This could lead to difficulty arming the vehicle.

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -710,6 +710,21 @@ bool AP_Arming::gps_checks(bool report)
...
 #endif  // AP_GPS_ENABLED
 
+bool AP_Arming::groundspeed_checks(bool report)
+{
+#if AP_GPS_ENABLED
+    // check only the first GPS:
+    const float gps_groundspeed = AP::gps().ground_speed();
+    const float ahrs_groundspeed = AP::ahrs().groundspeed();
+    if (fabsf(gps_groundspeed - ahrs_groundspeed) > 1.0f) {
+        check_failed(ARMING_CHECK_GPS, report, "GPS vs AHRS mismatch (%0.2f vs %0.2f)", gps_groundspeed, ahrs_groundspeed);
```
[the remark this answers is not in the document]
peterbarker: Done, thanks!

peterbarker:
> Hi, no big deal but do we need to make this a new check or can we just include it in the existing GPS check? I worry that a GPS may be attached but it may not be being used by the EKF. This could lead to difficulty arming the vehicle.

Done.  Meant to do that :-)

rmackay9:
Hi @peterbarker, maybe the branch hasn't been pushed correctly?  It still looks to me like this is a new separate check.. separate from the existing GPS check

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -710,6 +710,24 @@ bool AP_Arming::gps_checks(bool report)
...
+bool AP_Arming::groundspeed_checks(bool report)
+{
+#if AP_GPS_ENABLED
+    if (!check_enabled(ARMING_CHECK_GPS)) {
+        return true;
+    }
+    // check only the first GPS:
+    const float gps_groundspeed = AP::gps().ground_speed();
+    const float ahrs_groundspeed = AP::ahrs().groundspeed();
+    if (fabsf(gps_groundspeed - ahrs_groundspeed) > 1.0f) {
```
IamPete1, line 722: Would a 2D vector comparison be better? This can pass if there the same magnitude in opposite directions.

```suggestion
    const Vector2f gps_groundspeed = AP::gps().velocity().xy();
    const Vector2f ahrs_groundspeed = AP::ahrs().groundspeed_vector();
    if ((gps_groundspeed - ahrs_groundspeed).length_squared() > 1.0) {
```

IamPete1: Makes the reporting a little harder of course....

peterbarker (commented):
This is probably the wrong approach / too simplistic.

At the very least this would break @andyp1per 's recent work to centering on workflows where GPS is an option but isn't in your current source set in EKF3.

Closing this one as the problem it was supposed to solve (EKF3 velocity estimate divergence) was mitigated in a different way and actually fixed in the EKF3 quite recently.

Issue #21087, opened 2022-06-30
Crash Detection preventing hand launch in Auto


JeyPi:
https://discuss.ardupilot.org/t/crash-detection-preventing-hand-launch-in-auto/86455

## Bug report

**Crash Detection preventing hand launch in Auto**
enabling crash detection prevents hand launch in Auto mode despite setting CRASH_ACC_THRESH higher than maximum detected IMU.AccX during hand launch. It results on a motor starting for half a second and disarming. I think that crash detection should be inactivated during the the early first phase of a mission Takeoff when TKOFF_THR_MINACC is set (or at least that the value of CRASH_ACC_THRESH be augmented by the value of TKOFF_THR_MINACC during that phase) . That said, I don’t understand why in my case crash is detected whereas IMU.AccX max (25,35m/s/s) is much lower than my CRASH_ACC_THRESH setting (50m/s/s).

**Version**
Arduplane 4.2.2

**Plane**

**Airframe type**
Wing

**Hardware type**
Pixracer

**Logs**
[_Please provide a link to any relevant logs that show the issue_]
(https://drive.google.com/file/d/12xrdHlxKanhnPnUB2gk9LsdNnP_Fkc_D/view?usp=sharing)

rmackay9:
Hi @JeyPi,

We think we've perhaps fixed this issue in master now.  I wonder if you'd still be available to re-test with Plane-4.7.0-dev?

rubenp02:
My fix for this was merged a while back (#29390), should this maybe be closed? @rmackay9 

> That said, I don’t understand why in my case crash is detected whereas IMU.AccX max (25,35m/s/s) is much lower than my CRASH_ACC_THRESH setting (50m/s/s).

The takeoff logic intentionally detects a crash when there's a decceleration after the initial acceleration, which directly clashed with the "shake to launch" feature. It didn't use the CRASH_ACC_THRESH at all.

rmackay9:
Hi @rubenp02,

Great, I'll go ahead and close this then, txs!

PR #29390, opened 2025-02-27
Plane: Fix crash detection false positive on "shake to launch"

rubenp02:
# Plane: Fix crash detection false positive on "shake to launch"

Fixed a bug where, if the "shake to launch" functionality is configured, a crash would instantly be detected after the final acceleration event. The issue occurred because if a minimum acceleration threshold was set, a crash was detected if the aircraft wasn't flying immediately after reaching that threshold. This holds for single event acceleration launches but usually not when "shaking to launch".

In addition to this, the CRASH_ACC_THRESH parameter description has been adjusted to fix a small typo, and to clarify that setting it to 0 does not fully disable crash detection, but instead only the impact detection.

### DIFF
--- a/ArduPlane/is_flying.cpp
+++ b/ArduPlane/is_flying.cpp
@@ -261,11 +261,12 @@ void Plane::crash_detection_update(void)
             switch (flight_stage)
             {
             case AP_FixedWing::FlightStage::TAKEOFF:
-                if (g.takeoff_throttle_min_accel > 0 &&
-                        !throttle_suppressed) {
-                    // if you have an acceleration holding back throttle, but you met the
-                    // accel threshold but still not flying, then you either shook/hit the
-                    // plane or it was a failed launch.
+                if (g2.takeoff_throttle_accel_count == 1 && g.takeoff_throttle_min_accel > 0 &&
+                    !throttle_suppressed) {
+                    // if launching requires a single acceleration event and it
+                    // has already happened but the aircraft is still not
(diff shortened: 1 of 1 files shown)


peterbarker (commented):
How does one replicate this bug?

I use a single-throw to launch my vehicles - I haven't used the shake-to-launch stuff in a very long time.  But if it is as simple as "shake the vehicle and wait", then that's a rather large regression that's been introduced!

peterbarker:
> What happened when you tried, did it just work? I can't think of anything outside of a minimal "shake to launch" setup (i.e., TKOFF_THR_MINACC set and TKOFF_ACCEL_CNT > 1) that would be needed to replicate this. This is unrelated to the impact detection (the crash acceleration threshold), so that can be set to zero which I think is the default.
> 
> Check out the diff: the issue is in a hardcoded crash condition during takeoff where if the throttle is unsuppressed and a single acceleration event has happened but the plane still isn't flying then a crash is detected, which of course wouldn't apply for a shake to launch where multiple acceleration events are needed (there's even a comment saying as much).

Well, tridge and I spent a little time today trying to reproduce and failing...

Could you provide a log, please?

peterbarker:
[This](https://github.com/ardupilot/ardupilot/issues/21087) might be the same issue.

tridge (approved):

rmackay9:
This closed issue https://github.com/ArduPilot/ardupilot/issues/21087

PR #29391, opened 2025-02-27
AP_Scripting: Generalize BattEstimate SoC model for more battery chemistries

rubenp02:
# AP_Scripting: Generalize BattEstimate SoC model for more battery chemistries

Made the SoC model used by the BattEstimate script more flexible by turning the previously fixed parameter p0 into a free coefficient C4. This parameter controls the steepness of the SoC curve transition, so this change allows the model to also fit other battery technologies such as Li-ion.

The battery_fit.py script has also been fixed to support logs that use 'Inst' instead of 'Instance' for battery messages.

### DIFF
--- a/Tools/scripts/battery_fit.py
+++ b/Tools/scripts/battery_fit.py
@@ -31,10 +31,13 @@ def SOC_model(cell_volt, c):
     '''simple model of state of charge versus resting voltage.
     With thanks to Roho for the form of the equation
     https://electronics.stackexchange.com/questions/435837/calculate-battery-percentage-on-lipo-battery
+
+    Adjusted to also fit other battery chemistries such as Li-ion.
     '''
-    p0 = 80.0
+    p0 = c[3]
     p1 = c[2]
-    return constrain(c[0]*(1.0-1.0/math.pow(1+math.pow(cell_volt/c[1],p0),p1)),0,100)
+
+    return constrain(c[0] * (1.0 - 1.0 / math.pow(1 + math.pow(cell_volt / c[1], p0), p1)), 0, 100)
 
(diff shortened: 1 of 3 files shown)


peterbarker:
Looks very likely.  Please outline the testing you've given this PR.  In particular, it would b nice to know you've tested against a LiPo and (bot Python and Lua) and can state that it still works Just Fine.

IamPete1 (commented):
Need to bump the param table size to add space for the new params here:
https://github.com/ArduPilot/ardupilot/blob/e352025315b495ffe7d2023ad4bb983455673f5c/libraries/AP_Scripting/applets/BattEstimate.lua#L29

Bonus points for fixing the lus check issues, remove the cheats here:
https://github.com/ArduPilot/ardupilot/blob/e352025315b495ffe7d2023ad4bb983455673f5c/libraries/AP_Scripting/applets/BattEstimate.lua#L7-L8

Run locally with `./Tools/scripts/run_lua_language_check.py`

The count should be constraind to a max of 4 here: https://github.com/ArduPilot/ardupilot/blob/e352025315b495ffe7d2023ad4bb983455673f5c/libraries/AP_Scripting/applets/BattEstimate.lua#L221

IamPete1:
> 
> Maybe I'm wrong but isn't 32 more than enough still?

Yeah, I think there is still room, I'm not sure why it has some much extra space in the first place.

IamPete1:
> Think I should drop it to the actual required size of 25?

No, just leave it as it is.

tridge:
@rubenp02 can you post a lot of the discharge test on your battery?

tridge (approved):

PR #29393, opened 2025-02-27
AP_Mount: Fix to get caddx GM2 smooth operation

boa-pe:
AP_Mount: Fix to get caddx GM2 smooth operation during RC targetting.

root cause: no idea. probably some kind of compiler optimization, or other c++ quirks. Too lazy to dig into that ;)


boa-pe:
Did more testing.... this fix makes no sense. There is some other change between beta4 and master branch that fixes that issue

PR #29397, opened 2025-02-27
AP_Airspeed: avoid use of ownptr

peterbarker:


peterbarker:
Tested using QioTekZealotF427 and QioTek ASP5033 sensor.

Tested using QioTekZealotF427 and ASP5033 on an `f103-QioTekPeriph`

PR #29398, opened 2025-02-27
AP_AccelCal:Remove an unusable return

JulioCesarMatias:
At the end of void `AP_AccelCal::update_status()` there is a `return;` that shouldn't be there.


peterbarker (changes requested):
You've got the right prefix - `AP_AccelCal: `.

But you also need it on the commit, not just on the PR title.

Let me know if you need help fixing this 9 our Wiki has instructions on the git operaitons required for this.

peterbarker (approved):
LGTM

IamPete1 (approved):

PR #29399, opened 2025-02-28
AP_GPS: discard SBF packets with claimed length > 256 bytes

peterbarker:
Many thanks to 3DR for supplying the hardware used to debug this!

    messages coming in with a size of 24320 bytes.  It takes so long to try to read those bytes in (at the device's message rate) that we time out the GCS and try detecting again

    messages coming in with a size of 24320 bytes.  It takes so long to try to read those bytes in (at the device's message rate) that we time out the GCS and try detecting again

@chiara-septentrio this is a MosaicX5 module connected to an ArduPilot periph node.  We're reliably seeing bad bytes in the length field; this fix is necessary in case of true corruption, but it's a bit worrying how reliable this problem is!

### DIFF
--- a/libraries/AP_GPS/AP_GPS_SBF.cpp
+++ b/libraries/AP_GPS/AP_GPS_SBF.cpp
@@ -321,6 +321,14 @@ AP_GPS_SBF::parse(uint8_t temp)
                                      // indicates not enough bytes to do a crc
                 break;
             }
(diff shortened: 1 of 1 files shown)


Ryanf55 (approved):

tridge (approved):

rmackay9:

PR #29406, opened 2025-02-28
Mark more methods as __INITFUNC__

peterbarker:
pixart data and rangefinder init routines can probably go into external flash.


peterbarker:
CubeRedPrimary,-5632

tpwrules (approved):
Confirmed that this builds and boots on CubeRedPrimary.

Issue #28612, opened 2024-11-14
ArduPilot 4.6 issues list


rmackay9:
This is a list of issues discovered during ArduPilot 4.6 testing.  Issues for [Copter/Rover-4.5 issues are here](https://github.com/ArduPilot/ardupilot/issues/26103)

**Reports requiring investigation**

- [ ] https://github.com/ArduPilot/ardupilot/issues/32344
- [ ] https://github.com/ArduPilot/ardupilot/issues/31484
- [ ] Internal Error (flow-of-control) during takeoff in AltHold ([discussion](https://discuss.ardupilot.org/t/copter-4-6-3-released/140596/4))
- [ ] F.Port not working with TradHeli ([discussion](https://discuss.ardupilot.org/t/trad-heli-tuning-fport-can-t-send-telemetry-back/138940/17))
- [ ] https://github.com/ArduPilot/ardupilot/issues/31287
- [ ] https://github.com/ArduPilot/ardupilot/issues/31099
- [ ] EKF reset leads to bad roll estimate ([issue](https://github.com/ArduPilot/ardupilot/issues/31098), [discussion](https://github.com/ArduPilot/ardupilot/issues/31098))
- [ ] BLHeli_S ESC passthrough not working? ([discussion](https://discuss.ardupilot.org/t/copter-4-6-2-released/136822/26))
- [ ] yaapu ver X7 1.8.0 using Taranis QX7 receiver not displaying data ([discussion](https://discuss.ardupilot.org/t/copter-4-6-2-released/136822/12))
- [ ] log download slow? ([discussion](https://discuss.ardupilot.org/t/copter-4-6-1-released/135697))
- [ ] Plane LEDs not working? ([issue](https://github.com/ArduPilot/ardupilot/issues/30331#issuecomment-2988401515), [PR with potential fix](https://github.com/ArduPilot/ardupilot/pull/30469))
- [ ] Bad battery ([discussion](https://discuss.ardupilot.org/t/copter-4-6-0-released/134617/74))
- [ ] Micoair h743aio failing to restart ([discussion](https://discuss.ardupilot.org/t/copter-4-6-0-released/134617/57))
- [ ] Bad Gyro Health on Yaapu ([discussion](https://discuss.ardupilot.org/t/copter-4-6-0-released/134617/50))
- [ ] https://github.com/ArduPilot/ardupilot/issues/30142
[...]

PR #29409, opened 2025-03-02
Fix dshot cancel race

### LINKED ISSUE #29370: AP-4.6.0: TradHeli internal errors 0X4000000 and 0X4000020

andyp1per:
Alternative to https://github.com/ArduPilot/ardupilot/pull/29379 fixes #29370

This ensures dshot_waiter is only manipulated inside a system lock. Tested on a Pavo20Pro

### DIFF
--- a/libraries/AP_HAL_ChibiOS/RCOutput_bdshot.cpp
+++ b/libraries/AP_HAL_ChibiOS/RCOutput_bdshot.cpp
@@ -496,6 +496,13 @@ __RAMFUNC__ void RCOutput::bdshot_finish_dshot_gcr_transaction(virtual_timer_t* 
 #ifdef HAL_GPIO_LINE_GPIO56
     TOGGLE_PIN_DEBUG(56);
 #endif
+    osalDbgAssert(group->dshot_waiter, "No dshot waiter to signal");
+
+    if (group->dshot_waiter == nullptr) {   // transaction was cancelled, leave everything alone
+        chSysUnlockFromISR();
+        return;
+    }
+
     uint8_t curr_telem_chan = group->bdshot.curr_telem_chan;
 
     // the DMA buffer is either the regular outbound one because we are sharing UP and CH
@@ -542,6 +549,8 @@ __RAMFUNC__ void RCOutput::bdshot_finish_dshot_gcr_transaction(virtual_timer_t* 
 
     // tell the waiting process we've done the DMA
     chEvtSignalI(group->dshot_waiter, group->dshot_event_mask);
+    group->dshot_waiter = nullptr;
+
 #ifdef HAL_GPIO_LINE_GPIO56
     TOGGLE_PIN_DEBUG(56);
 #endif
--- a/libraries/AP_HAL_ChibiOS/RCOutput.cpp
+++ b/libraries/AP_HAL_ChibiOS/RCOutput.cpp
@@ -347,7 +347,7 @@ void RCOutput::dshot_collect_dma_locks(rcout_timer_t cycle_start_us, rcout_timer
             if (!mask) {
                 dma_cancel(group);
             }
-            group.dshot_waiter = nullptr;
+            osalDbgAssert(group.dshot_waiter == nullptr, "Dshot waiter was not reset");
 #ifdef HAL_WITH_BIDIR_DSHOT
             // if using input capture DMA then clean up
             if (group.bdshot.enabled) {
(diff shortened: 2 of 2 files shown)


libraries/AP_HAL_ChibiOS/RCOutput_bdshot.cpp
```diff
@@ -496,6 +496,11 @@ __RAMFUNC__ void RCOutput::bdshot_finish_dshot_gcr_transaction(virtual_timer_t*
 #ifdef HAL_GPIO_LINE_GPIO56
     TOGGLE_PIN_DEBUG(56);
 #endif
+    if (group->dshot_waiter == nullptr) {   // transaction was cancelled, leave everything alone
```
tridge, line 499: i think this change won't work, when the bug happens 'this' is nullptr, so we are checking an incorrect memory location

tridge (changes requested):

libraries/AP_HAL_ChibiOS/RCOutput_bdshot.cpp
andyp1per: No, I don't think so. The "this" is a pointer set in the timer, it cannot be null unless the timer is reset in which case the callback is not called. The nullptr is not "this" but this->dshot_waiter which *was* so to nullptr outside of a lock hence the bug.

andyp1per:
This was the crash:
```
#0  0x0813b21a in chEvtSignalI (tp=0x0, events=<optimized out>) at ../../modules/ChibiOS/os/rt/src/chevents.c:300
#1  0x081280e0 in ChibiOS::RCOutput::bdshot_finish_dshot_gcr_transaction (vt=<optimized out>, 
    p=0x24017548 <ChibiOS::RCOutput::pwm_group_list+1392>) at ../../libraries/AP_HAL_ChibiOS/RCOutput_bdshot.cpp:544
#2  0x0813bd38 in chVTDoTickI () at ../../modules/ChibiOS/os/rt/src/chvt.c:537
#3  0x0813b86c in chSysTimerHandlerI () at ../../modules/ChibiOS/os/rt/src/chsys.c:403
#4  0x0813dafc in osalOsTimerHandlerI () at ../../modules/ChibiOS/os/hal/osal/rt-nil/osal.h:669
#5  st_lld_serve_interrupt () at ../../modules/ChibiOS/os/hal/ports/STM32/LLD/SYSTICKv1/hal_st_lld.c:615
#6  0x0813e032 in Vector108 () at ../../modules/ChibiOS/os/hal/ports/STM32/LLD/TIMv1/stm32_tim5.inc:129
#7  <signal handler called>
#8  0x08012500 in Copter::read_inertia (this=0x240033d8 <copter>) at ../../ArduCopter/inertia.cpp:7
#9  0x080798e8 in Functor<void>::operator() (this=0x8147e94 <Copter::scheduler_tasks+120>) at ../../libraries/AP_HAL/utility/functor.h:52
#10 AP_Scheduler::run (this=this@entry=0x24003608 <copter+560>, time_available=2104) at ../../libraries/AP_Scheduler/AP_Scheduler.cpp:264
#11 0x08079d3e in AP_Scheduler::loop (this=this@entry=0x24003608 <copter+560>) at ../../libraries/AP_Scheduler/AP_Scheduler.cpp:393
#12 0x0807b7f6 in AP_Vehicle::loop (this=0x240033d8 <copter>) at ../../libraries/AP_Vehicle/AP_Vehicle.cpp:544
#13 0x080e09b4 in main_loop () at ../../libraries/AP_HAL_ChibiOS/HAL_ChibiOS_Class.cpp:301
#14 HAL_ChibiOS::run (this=<optimized out>, argc=<optimized out>, argv=<optimized out>, callbacks=<optimized out>)
    at ../../libraries/AP_HAL_ChibiOS/HAL_ChibiOS_Class.cpp:354
#15 0x0800c3dc in main (argc=<optimized out>, argv=<optimized out>) at ../../ArduCopter/Copter.cpp:892
```
It's the signal on group->dshot_waiter that is the problem, not the de-reference of group

libraries/AP_HAL_ChibiOS/RCOutput_bdshot.cpp
andyp1per: Incidentally I don't believe this can ever be triggered because we now cancel the timer inside a lock together with setting the value to nullptr

tridge: you're right, sorry!

tridge:
it would be good to run with --enable-asserts, and the MPU protection enabled with @Ferruccio1984 setup

rmackay9:
This is related to the "TradHeli internal errors 0X4000000 and 0X4000020" issue on the 4.6 issues list https://github.com/ArduPilot/ardupilot/issues/28612

tridge:
@andyp1per not building?

andyp1per:
Briefly tested on my fast rate quad without issue

tridge:
@andyp1per @Ferruccio1984 was this tested on the problem setup with asserts enabled?

andyp1per:
@Ferruccio1984 's issue turned out to be flash ECC corruption which I was kind of expecting. Fully wiping the flash and re-flashing fixed the issue.

@tridge when I looked at the assertions the crash was actually happening inside the crashdump code - so the firmware was crashing trying to write a crashdump to flash. I am not sure what we can do about this.

rmackay9:

PR #29411, opened 2025-03-02
matrix_alg: Probable redundancy

JulioCesarMatias:
A comparison is made whether `T` is defined as float or double, but the logic here seems to be redundant, there being no need to use `std::is_same`.


peterbarker (approved):
LGTM

Introduced in 999268cbba2a17daa766e671c9025cc1e6f65d15 (probably a bug at that point, but later fixed by application of `fabsF` in both cases

rmackay9 (approved):

rmackay9:
Normally we'd probably prefix this with AP_Math but that's a nitpick really, thanks!

PR #29413, opened 2025-03-02
AP_Filesystem: LittleFS: support JEDEC NOR and enable KakuteH7Mini

tpwrules:
Various fixes and tweaks to make LittleFS work well and log smooth on my KakuteH7Mini, which has a 16MiB W25Q128 NOR flash chip. These are somewhat different to the NAND W25Nxx chips done previously, and they were written but not tested in the initial PR. Please see the commits for details and tradeoffs.

I can comfortably log up to the full capacity of the chip at 120KB/s without dropping anything, though there are periodic spikes in buffer usage due to metadata compaction. Full speed is something like 150KB/s. Download speeds are >600KB/s, like SD and NAND. Sample log from sim on hardware: [log_littlefs_nor.zip](https://github.com/user-attachments/files/19044796/log_littlefs_nor.zip)

Terrain is still a bugbear, it works but uploading is horrendously slow; I had it disabled for testing. FTP file upload is also breathtakingly slow unless mavproxy's `ftp set write_qsize 1` is used because dropped packets when LittleFS is busy erasing a block cause seeks that cause huge rewrites. Files up to a few dozen kilobytes usually take a couple seconds, but even a 100K file can take nearly 5 minutes. I'm not sure where packets are getting dropped but this might need a revision in the FTP server. We didn't see these issues with NAND because it's generally faster and has less latency.

I only enabled the one board I have to test but a future PR could enable it for all boards with NOR dataflash. The 2MiB chips might feel cramped though. I also tested that scripting works fine, and that nothing broke on KakuteH7Mini-Nand.


libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -810,6 +811,13 @@ uint32_t AP_Filesystem_FlashMemory_LittleFS::find_block_size_and_count() {
     fs_cfg.prog_size = page_size;
     fs_cfg.block_size = flash_block_size * LFS_FLASH_BLOCKS_PER_BLOCK;
     fs_cfg.block_count = flash_block_count / LFS_FLASH_BLOCKS_PER_BLOCK;
+
+    // larger metadata blocks mean less frequent compaction operations, but each
+    // takes longer. however, the total time spent compacting for a given file
+    // size still goes down with larger metadata blocks. 4096 was chosen to
+    // minimize both frequency and log buffer utilization.
+    fs_cfg.metadata_max = 4096;
```
andyp1per, line 819: what was this before?

andyp1per (commented):
Looks good. I am not a fan of "jedec_nor" but can't think of anything better.
I think we should consider doing this for all H7 boards that use the nor chips.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
tpwrules: The block size, i.e. 65536z.

tpwrules:
Yeah I plan to add more targets in a follow-up PR.

libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
andyp1per: I mean I think the lower value is better here, but I recall there being some other default.

andyp1per (approved):

PR #29414, opened 2025-03-02
AP_Generator: FuelCell: increase max data rate

IamPete1:
At some point between when the comment was added on the function and now the generator scheduler rate was dropped to 10Hz. 30 bytes read per call at 10Hz means a max of 300 bytes per second rather than the 600 bytes per second it would have been originally. 

This bumps it up to 255 bytes for a max of 2550 bytes per second. These typically ruin on a serial port at 9600 baud so a max of 960 bytes per second. We could go lower, 50 or 100 bytes per call would be fine for the existing units, but this was a PITA to debug so some headroom is nice.

### DIFF
--- a/libraries/AP_Generator/AP_Generator_IE_FuelCell.cpp
+++ b/libraries/AP_Generator/AP_Generator_IE_FuelCell.cpp
@@ -35,7 +35,7 @@ void AP_Generator_IE_FuelCell::init()
     _health_warn_last_ms = AP_HAL::millis();
 }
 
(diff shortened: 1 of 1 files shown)


IamPete1:
Without, we log exactly the 300 bytes/second limit and the generator is unhealthy:
image

With this fix we get the true data rate of about 350 bytes per second.
image

CraigElder:
I found this issue and have tested this fix on real hardware.

CraigElder (approved):
I found this issue and have tested this fix on real hardware.

Issue #29404, opened 2025-02-28
bug in Tiltrotor.cpp at line 122


jcproulx:
## Bug report

At line 122 of tiltrotor.cpp
Surely a syntax error but you must replace "<<1" by "<<i" if we want a moving bit to compare with "tilt_mask "

Plane Version 4.5.3  but it is also in other versions

VTOL airplane with four motors

Pixhawk 6x

There is no log because I found this error in looking into the code.

IamPete1:
Yep, can you do a PR to fix?

jcproulx:
No because I am not a programmer and not familiar with all the procedures.
I just want to pinpoint this evident syntax error

sanggusti:
I think this issue could be closed since #29415 is already resolving this

PR #29415, opened 2025-03-03
Plane: Tiltrotor: fix `has_vtol_motor` function

IamPete1:
This fixes a by reported by @jcproulx in https://github.com/ArduPilot/ardupilot/issues/29404.

It was always checking the first motor rather than any motor. The `_have_vtol_motor` bool is used in the `has_vtol_motor` function which is only used here: 

https://github.com/ArduPilot/ardupilot/blob/076951f003703f178523cd2e122fd54fffcaefff/ArduPlane/quadplane.cpp#L1591-L1594

The goal of that code is to smooth the handover of throttle control from VTOL to FW in the transition timer phase on vehicles that had all motors tilting.


tridge (approved):

rmackay9:

Issue #29147, opened 2025-01-27
Regression in MCU Volt?


MichelleRos:
**Issue details**

We seem to have had a regression (or unintentional change?) in the logging of the MCU log message.

What it used to look like (v4.4.0):
![Image](https://github.com/user-attachments/assets/c223378e-40d7-44df-9605-60a3e665c599)
Max and min is since the last logged message.

What it is now (recent master):
![Image](https://github.com/user-attachments/assets/e7a94a72-b1d4-4fcb-8f84-5ac7c4da7952)
Max and min are since the start of the flight.

This makes it rather less useful for investigating power issues as we don't see the variation in voltage as clearly.

tridge:
this is a regression, yes, well spotted!

tridge:
could be this one?
```
commit 726dcda24f68da55626cad45b839e9d27d684c04
Author: bugobliterator <siddharthbharatpurohit@gmail.com>
Date:   Tue Jan 17 15:15:15 2023 +1100

    AP_HAL_ChibiOS: add support for using ADC2 and ADC3 on H7
```

MichelleRos:
> could be this one?
> 
> ```
> commit 726dcda24f68da55626cad45b839e9d27d684c04
> Author: bugobliterator <siddharthbharatpurohit@gmail.com>
> Date:   Tue Jan 17 15:15:15 2023 +1100
> 
>     AP_HAL_ChibiOS: add support for using ADC2 and ADC3 on H7
> ```

Yup. I was suspecting that commit, but wanted to check first. I just did tests before and after the PR that added this commit (https://github.com/ArduPilot/ardupilot/pull/22671), and confirmed that this PR is what changed it.

peterbarker:
Fixed by #29417

PR #29417, opened 2025-03-03
AP_HAL_ChibiOS: Fix regression in MCU max and min volt logging

MichelleRos:
This fixes the regression in the max and min MCU voltage in the logs, as described in this issue: https://github.com/ArduPilot/ardupilot/issues/29147

Tested on CubeOrangePlus and MatekH743 with copter firmware.

Master:
00000056 master

This PR:
00000058 my pr

Note, however that this does reset it at 20 Hz, whereas MCU only gets logged at 10 Hz (on copter, at any rate).
AFAICT this was the previous behaviour as well though.


peterbarker (approved):
LGTM

rmackay9:

PR #29418, opened 2025-03-03
Tools: Upgrade tempcal_IMU.py

amilcarlucas:
Upgrade to python > 3.6
Added progress bar support
Use logging best practices
Lint with pylint, mypy, pyright and ruff
Added type annotations
Added pytests assertion tests


amilcarlucas:
@peterbarker can you take a look at the tests? I don't think we run pytest yet in CI do we?

Tools/scripts/tempcal_IMU.py
```diff
@@ -1,27 +1,25 @@
-#!/usr/bin/env python
-'''
+#!/usr/bin/env python3
+"""
```
peterbarker, line 2:
```suggestion
"""
AP_FLAKE8_CLEAN
```

tridge:
@amilcarlucas I'd like to see it being used with SITL, using SIM_IMUT parameters to setup a known temperature curve then checking the tool produces the same result both with and without the --online option

peterbarker:
@amilcarlucas tridge asked for some specific testing on this PR.  Are you planning on doing that testing?

amilcarlucas:
@peterbarker do we have a pytest running CI job now?
If not, I can do the tests that tridge wanted and add that.

amilcarlucas:
I will do this at the conference, with some help from tridge.

PR #29425, opened 2025-03-04
DroneCAN: update libcanard for semaphore bug fix

tridge:
it was possible for a runtime allocated subscriber, client or server to trigger a call to a pure virtual handle_message function in the HandlerList class as the constructor ordering inserts the handler into the list before the vtable pointer is updated to the vtable for the child class. The same issue happens on destructor

this could happen in a couple of places in ArduPilot:

 - the MPPT battery driver does runtime allocation of a client on an R/C switch
 - the DroneCAN_Serial client is started after the thread starts so packet processing is started

thanks to @tpwrules and @bugobliterator 

does not need to be in 4.6.0 I think, but could be in 4.6.x once sufficiently tested in master


tpwrules (approved):
I did the PR for this so I'm a bit biased, but I did do some basic smoke tests that a DroneCAN GPS still works.

rmackay9:

PR #29428, opened 2025-03-04
AP_Mount: fix CADDX RC update rate

rmackay9:
This fixes an issue with the CADDX driver that causes the gimbal to move at only 1hz towards its target when the user is controlling the gimbal via RC input.

This has been lightly tested on real hardware

Thanks very much to @boa-pe for finding and reporting this issue!


peterbarker (approved):
LGTM

rmackay9:
Thanks, merged to 4.6!

PR #29432, opened 2025-03-04
Swap -f for --model in JSON read me instructions

ConorZAM:
The previous read me suggested using "-f json" to include the interface in SITL. This was difficult to work with as it replaced the vehicle's frame definition argument and prevented users from accessing the default parameters. By using "--model json" instead, the "-f" argument is left free for the vehicle's frame definition.


peterbarker (approved):
I'vefixed the commit message up to conf9orm to opur standards and will mark as MergeOnCIPass.

Thanks!

Issue #29384, opened 2025-02-27
EKF3 velocity divergence pre-arm


tridge:
We had a divergence of EKF3 velocity pre-arming on the ground which caused a roll over on takeoff of a quadplane.
GPS velocity looks fine, EKF3 velocity is very bad. Normalised innovations were low.
Replay log here:
http://uav.tridgell.net/tmp/00000082.BIN
replays nicely, no log packet loss
firmware is 4.6.0-beta4, but master replay shows the same issue.

old versions tested:
 - 4.5.7 gives same replay result
 - 4.4.4 gives the same replay result
 - 4.3.8 gives almost the same result (same general VE divergence)
 - 4.2.3 reproduces the issue
 - the 4.1.x replay doesn't work with the log

you can use replay with EKF2 on this log with this PR:
https://github.com/ArduPilot/ardupilot/pull/29405
and --force-ekf2
Using EKF2 the velocity does not diverge, so this looks to be an EKF3 only issue, and one that has been in EKF3 for a very long time

Apart from the divergence, the other worrying part of this log is that the scaled innovations presented to the EKF failsafe code and to the user over MAVLink do not show any health issues with the EKF. Same goes for the lane switching error scores. So no fail-safes, no GCS warnings, no arming check failures.

Other notes:
 - delaying the start of GPS fusion in EKF3 to be after 90s of stability in NavEKF3_core::calcGpsGoodToAlign() instead of 10s "fixes" the issue

![Image](https://github.com/user-attachments/assets/c3c9cfaa-f5d1-4805-ad76-77bf9883c417)

![Image](https://github.com/user-attachments/assets/82d19bf0-38aa-49a9-8d45-576162bef959)

update:
 - two more examples of a copter here: https://uav.tridgell.net/tmp/Lift/
 - it seems possible it is related to being on sloping ground pre-arm?

tridge:
V09 is diverging:

![Image](https://github.com/user-attachments/assets/b7211e06-e72c-4c63-9ed2-0bd6b9de4ed1)

tridge:
the V05 (variance of velocity east) is diverging badly:
![Image](https://github.com/user-attachments/assets/22a8ac8d-7346-4547-88e8-4689a163d762)
it is stopping at 1000.0, which comes from NavEKF3_core::ConstrainVariances()

priseborough:
[...]

PR #29434, opened 2025-03-05
AP_NavEKF3: added in a check for high velocity innovation

tridge:
if using GPS and our horizontal velocity innovation is high then mark the filter as unhealthy, which causes an arming failure

this helps with #29384

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3.cpp
@@ -1157,6 +1157,10 @@ bool NavEKF3::pre_arm_check(bool requires_position, char *failure_msg, uint8_t f
             }
             return false;
         }
+        // run per-core pre-arm checks
+        if (!core[i].pre_arm_check(requires_position, failure_msg, failure_msg_len)) {
+            return false;
+        }
     }
     return true;
 }
--- a/libraries/AP_NavEKF3/AP_NavEKF3_Outputs.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_Outputs.cpp
@@ -32,6 +32,31 @@ bool NavEKF3_core::healthy(void) const
     return true;
 }
 
+/*
+  per-core pre-arm checks. returns false if we fail arming checks, in
+  which case the buffer will be populated with a failure message
+  requires_position should be true if horizontal position configuration should be checked
+*/
+bool NavEKF3_core::pre_arm_check(bool requires_position, char *failure_msg, uint8_t failure_msg_len) const
+{
+    if (requires_position) {
+        // additional checks when position is required, used by pre-arm checks
+        const float max_vel_innovation = 2.0;
+        const float hvel_innovation = sqrtf(sq(innovVelPos[0])+sq(innovVelPos[1]));
+        if (onGround && PV_AidingMode == AID_ABSOLUTE &&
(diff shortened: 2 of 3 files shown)


peterbarker:
Looks right.  How was it tested?

tridge:
> Looks right. How was it tested?

replay on the log, with printf() for when it triggers

peterbarker:
@rmackay9 we have users moving from various non-GPS-navigation modes (eg. inside hangars) to using-GPS.  Will this adversely affect them?

I'm thinking of things like the `ahrs-source-extnav-optflow.lua` workflows.  I don't think we have any autotest for that one.

rmackay9:
Hi @peterbarker, I'm not sure.

TBH, I'm not sure if we should be changing the meaning of "healthy".  If we want the pre-arm to protect against arming with high innovations then I think we should instead add a specific check for that.

rishabsingh3003:
I think this would block arming in some valid use cases; for example, a user trying to take off manually in alt-hold mode from a location that is adversely affecting the GPS (hence causing high innovations) but will be fine once high up in the air. It should be a specific check, as Randy suggested.

peterbarker:
> I think this would block arming in some valid use cases; for example, a user trying to take off manually in alt-hold mode from a location that is adversely affecting the GPS (hence causing high innovations) but will be fine once high up in the air. It should be a specific check, as Randy suggested.

We have an arming bit for this specific situation - "always require position for arming"....

tridge:
@rmackay9 @rishabsingh3003 I've changed it so that the check is only done if position is required

rmackay9:
hmm, this essentially adds another type of healthy.  I still don't understand why we don't just add an arming check if we're trying to prevent arming with a high velocity innovation. That would surely be a more direct approach?

rishabsingh3003 (approved):
I like this change. However, checking just velocity seems a bit arbitrary.  We should probably be checking at least some of the other states as well for high innovations? 

For example, I probably wouldn't want to takeoff in guided if the vertical position innovations are super high. Might be scope creep considering this in response to a particular crash.

peterbarker (approved):
So this will give us a pre-arm check, and I think it's mergeable just for that.  Would have prevented a bad mishap, and is easily back-portable.

The current formulation doesn't give us any post-arming protection.  We could possibly look at *telling* the EKF we require it to have position, instead of passing that information in as part of the arming checks.  That would allow us to do this velocity check continuously rather than just at prearm.

tridge:
> For example, I probably wouldn't want to takeoff in guided if the vertical position innovations are super high. Might be scope creep considering this in response to a particular crash.

The position states will reset on high innovation, but the velocity states won't. We could add a check for how recently a reset has happened if we want to cover that. I was trying to get a minimal change in for this particular issue that we can put into 4.6.0.

rmackay9:
I think functionally this is fine but the way it's hidden in the ahrs::healthy() seems unintuitive for future developers to figure out.

I very much like Tridge's suggestion to, "I could add a NavEKF3_core::pre_arm_check()".  I think then it becomes very obvious what the check's purpose is.

tridge:
@rmackay9 I added the per-core pre-arm check function as discussed

peterbarker (approved):

rmackay9 (approved):
Great, thanks!

rmackay9:

PR #29435, opened 2025-03-05
AP_Mission: fixed option_is_set()

tridge:
bug introduced in PR #27654

all PRs should be tested ...

thanks to menschel for reporting this issue:
https://discuss.ardupilot.org/t/plane-4-6-0-beta/126358/42

### DIFF
--- a/libraries/AP_Mission/AP_Mission.h
+++ b/libraries/AP_Mission/AP_Mission.h
@@ -738,7 +738,7 @@ public:
         CONTINUE_AFTER_LAND      =  2,  // continue running mission (do not disarm) after land if takeoff is next waypoint
     };
     bool option_is_set(Option option) const {
-        return (_options.get() & (uint16_t)option) != 0;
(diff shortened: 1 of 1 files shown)


rmackay9:
I guess it was tested but only by setting the first bit :-)

rmackay9:
I think the issue could also be resolved by changing the enum.  This would make it consistent with [how AP_ExternalAHRS does it](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_ExternalAHRS/AP_ExternalAHRS.h#L177).

peterbarker (approved):
I do happen to agree with Randy's comment about being consistent about where we do the shifting, 'though.

rmackay9:
@peterbarker, let's be consistent then, it'll be easy!

rmackay9 (changes requested):
let me raise a PR with the alternative fix

tridge:
randy posted alternative

PR #29436, opened 2025-03-05
AP_Mission: fix option_is_set

rmackay9:
This is the alternative to https://github.com/ArduPilot/ardupilot/pull/29435

This has been lightly tested in SITL using the two options that Copter supports:

1. Clear On Boot ---> the mission was correctly cleared after SITL was restarted
2. Continue After Land ---> with this set (and only when this was set) the vehicle was able to fly, land and takeoff again in a single mission


peterbarker (approved):
LGTM

rmackay9:

PR #29444, opened 2025-03-06
AP_RangeFinder: add hexsoon 24G support

rmackay9:
This adds support for the Hexsoon 24G radar.

The approach is the same as for the Hexsoon Proximity driver (https://github.com/ArduPilot/ardupilot/pull/29235) in that a new RNGFNDx_TYPE enum is added but the CAN protocol is the shared with another driver.

I've tested this on real hardware and it works a-ok.


rmackay9:
I actually agree with CoPilot's comment above that we should consider adding "24G" somehow into the name.  Perhaps just in the user facing RNGFND1_TYPE parameter description.

Basically there are two radars provided by hexsoon:

- 77G proximity sensors which provide 3 sectors of distances
- 24G rangefinders which provide just a single distance

libraries/AP_RangeFinder/AP_RangeFinder_NRA24_CAN.cpp
```diff
@@ -1,6 +1,6 @@
 #include "AP_RangeFinder_config.h"
 
-#if AP_RANGEFINDER_NRA24_CAN_ENABLED
+#if AP_RANGEFINDER_NRA24_CAN_ENABLED || AP_RANGEFINDER_HEXSOONRADAR_ENABLED
```
peterbarker, line 3: This is really, really ugly.

```suggestion
#if AP_RANGEFINDER_NRA24_CAN_DRIVER_ENABLED
```

... and set that if either of AP_RANGEFINDER_NRA24_CAN_ENABLED or AP_RANGEFINDER_HEXSOONRADAR_ENABLED is set.

I think that should simplify AP_RangeFinder.cpp a lot

peterbarker (changes requested):

libraries/AP_RangeFinder/AP_RangeFinder_NRA24_CAN.cpp
peterbarker: ... would be nice to find something generic to take the palce of `AP_RANGEFINDER_NRA24_CAN_DRIVER_ENABLED` here, but this is better than starting to || in ever damned alias for the same device.

rmackay9: ok, thanks for the patch.  TBH, I'm not sure I agree it's any better but the discussion at least led me to go back and review again and I had missed another place where the NRA_CAN enum was being used (now fixed).

rmackay9:
Thanks @peterbarker, I've gone ahead and merged it then, hope that's OK with everyone.

rmackay9:

PR #29445, opened 2025-03-06
AP_ADSB: clarify ADSB_LIST_ALT docs

tridge:
alt is above our alt

### DIFF
--- a/libraries/AP_ADSB/AP_ADSB.cpp
+++ b/libraries/AP_ADSB/AP_ADSB.cpp
@@ -150,7 +150,7 @@ const AP_Param::GroupInfo AP_ADSB::var_info[] = {
 
     // @Param: LIST_ALT
     // @DisplayName: ADSB vehicle list altitude filter
-    // @Description: ADSB vehicle list altitude filter. Vehicles detected above this altitude will be completely ignored. They will not show up in the SRx_ADSB stream to the GCS and will not be considered in any avoidance calculations. A value of 0 will disable this filter.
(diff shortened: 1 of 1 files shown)


libraries/AP_ADSB/AP_ADSB.cpp
```diff
@@ -150,7 +150,7 @@ const AP_Param::GroupInfo AP_ADSB::var_info[] = {
 
     // @Param: LIST_ALT
     // @DisplayName: ADSB vehicle list altitude filter
-    // @Description: ADSB vehicle list altitude filter. Vehicles detected above this altitude will be completely ignored. They will not show up in the SRx_ADSB stream to the GCS and will not be considered in any avoidance calculations. A value of 0 will disable this filter.
+    // @Description: ADSB vehicle list altitude filter. Vehicles detected more than this altitude above our owm altitude will be completely ignored. They will not show up in the SRx_ADSB stream to the GCS and will not be considered in any avoidance calculations. A value of 0 will disable this filter.
```
Hwurzburg, line 153:
```suggestion
    // @Description: ADSB vehicle list altitude filter. Vehicles detected more than this altitude above our own altitude will be completely ignored. They will not show up in the SRx_ADSB stream to the GCS and will not be considered in any avoidance calculations. A value of 0 will disable this filter.
```

Hwurzburg (changes requested):

libraries/AP_ADSB/AP_ADSB.cpp
peterbarker: Fixed

PR #29446, opened 2025-03-06
AP_HAL_ChibiOS: define mav sys id 2 for CubeRedSecondary in hwdef

bugobliterator:
moves the correct setting for Mavlink system id for CubeRedSecondary from default.parm to hwdef. 

Thanks to @tpwrules for finding this out.


tpwrules (approved):
Tested that it builds and still shows a system ID of 2 after a fresh flash.

timtuxworth:
> Tested that it builds and still shows a system ID of 2 after a fresh flash.

What if you manually change it to something other than 2 (e.g. 17) - will it still show 17 after a fresh flash?

tpwrules:
> > Tested that it builds and still shows a system ID of 2 after a fresh flash.
> 
> What if you manually change it to something other than 2 (e.g. 17) - will it still show 17 after a fresh flash?

Yes, that define just sets the default for the parameter. By "fresh flash" here I meant a different vehicle so that the parameters are erased. So no, it would not remember the 17. But if you do a firmware upgrade then yes it will remember it.

PR #29448, opened 2025-03-06
AP_HAL_ESP32: init esp32s3m5stampfly

tpwrules:
This is a ready-to-flash board definition for the ready-to-fly 55mm M5Stack M5Stamp Fly quadcopter. Hopefully it can serve as an inexpensive reference design for the often-beleaguered ESP32 support. It flies great using the stock parameters and a joystick over MAVLink over WiFi.

The CPU and flash are still slow and I had to limit it to <1/3 of a stock copter's loop rate but for amateur pilots that is not noticeable. Current sensor suite is suitable for non-position-controlled modes (e.g. Stabilize and AltHold). Battery monitoring is also set up and works. Flight time is 4-5 minutes.

Work remains to get the compass driver working so that automated flight is possible with an inexpensive GPS. Then hopefully some day the LEDs, buzzer, optical flow sensor, and chip LiDARs can be made to work though those may end up not possible within ArduPilot's framework and given the hardware constraints.

### DIFF
--- a/libraries/AP_HAL_ESP32/boards/esp32s3m5stampfly.h
+++ b/libraries/AP_HAL_ESP32/boards/esp32s3m5stampfly.h
@@ -1,0 +1,145 @@
+/*
+ * This file is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This file is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+// GROVE EXPANSION
+//
+// The Stampfly has two Seeed Studio Grove connectors to attach peripherals. By
+// default, the left (red) connector is I2C (with no other devices on the bus)
+// and the right (black) connector is SERIAL3. However, the two signal pins in
+// each can be reassigned by editing this file.
+//
+// Grove Red (I2C)
+// 1 (toward front): GPIO_NUM_15, SCL (has 4.7kΩ pullup to 3.3V)
+// 2               : GPIO_NUM_13, SDA (has 4.7kΩ pullup to 3.3V)
+// 3               : +V (+5V if plugged into USB, else +Vbat)
+// 4 (toward rear) : GND
+//
+// Grove Black (SERIAL3)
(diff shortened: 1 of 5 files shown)


libraries/AP_HAL_ESP32/boards/esp32s3m5stampfly.h
```diff
@@ -0,0 +1,145 @@
...
+// log over MAVLink by default
+#define HAL_LOGGING_FILESYSTEM_ENABLED 0
+#define HAL_LOGGING_DATAFLASH_ENABLED 0
+#define HAL_LOGGING_MAVLINK_ENABLED 1
+#define HAL_LOGGING_BACKENDS_DEFAULT 2
+
+#define HAL_BOARD_LOG_DIRECTORY "/SDCARD/APM/LOGS"
+#define HAL_BOARD_STORAGE_DIRECTORY "/SDCARD/APM/STORAGE"
+#define HAL_BOARD_LOG_DIRECTORY "/SDCARD/APM/LOGS"
+#define HAL_BOARD_TERRAIN_DIRECTORY "/SDCARD/APM/TERRAIN"
```
peterbarker, line 140:
```suggestion
```

libraries/AP_HAL_ESP32/boards/esp32s3m5stampfly.h
```diff
@@ -0,0 +1,145 @@
...
+
+#define HAL_BARO_DEFAULT        HAL_BARO_BMP280_I2C
+#define HAL_BARO_BMP085_NAME        "bmp280"
+#define HAL_BARO_PROBE_LIST         PROBE_BARO_I2C(BMP280, 0, 0x76)
+
+#define AP_COMPASS_ENABLE_DEFAULT 0
+#define ALLOW_ARM_NO_COMPASS
+#define AP_AIRSPEED_ENABLED 0
+#define AP_AIRSPEED_ANALOG_ENABLED 0
+#define AP_AIRSPEED_BACKEND_DEFAULT_ENABLED 0
```
peterbarker, line 73:
```suggestion
```

peterbarker (approved):
LGEM.

tpwrules:
The defines you point out are definitely cargo culted but they don't seem to be wrong. Will clean them up later; eventually ESP32 would like hwdef.dat support too.

srmainwaring (approved):
@tpwrules - looks good to me - essentially the same as your `stampfly` branch which I have flight tested and used as as base for modifications to use GPS + ELRS.

The main difference I can see between this PR and the `stamplfy` branch is the latter includes a method to set the chip select in `SPIDevice`.

tpwrules:
Thanks for the review.

> The main difference I can see between this PR and the `stamplfy` branch is the latter includes a method to set the chip select in `SPIDevice`.

Those methods are required to make the Pixart driver work. I disabled that driver, and wasn't convinced the chip select functions were correct enough in operation overall to be worth upstreaming.

I did of course flight test this branch briefly, so the SPI IMU is good.

tridge (approved):

PR #29456, opened 2025-03-07
Extract Features by adding compile time source 

bugobliterator:


peterbarker:
I do certainly like many parts of this - using a Feature object in `build_options.py`, for example.

I also like the work done in here to make a symbol persist.

Things I don't like...
 - I don't really understand the expected workflow
 - the naming of extract features "v2", which doesn't really tell me what;s going on
 - the C++ parsing in Python
 - the inclusion of the header file in board_list.py; it's an additional step and I'm not sure why it is required as we could just grab *every* header!
 - the fact that the new features.txt won't match the result of `extract_features.py`

I'd actually like to use the "retain symbols" thing you've got in this to be able to mark features are present in a binary even when there's no vestige of them in the symbol table.  That would be able to get rid of some of the hackery in `extract_features.py` and a bunch in `test_build_options.py`.  (I did actually have a poke at doing this but couldn't find the compiler stuff to do it).

I've created https://github.com/ArduPilot/ardupilot/pull/29482 which does the double-build that you suggested a while back - build a binary from the features we extracted from it, make sure the result is the same as the original.

So back to the workflow thing.

What does this get us which isn't covered by `test_build_options.py` and potentially that double-build branch of mine?  I can see that it at least gets us something for a couple of features which aren't pulled out by the current system (the CAN-long-printf stuff), but, as above, it would be nice to have that in the original extract_features!.

tridge:
I'm also confused what this gains us over the existing system

tridge:
@bugobliterator is there an update on this? any chance to look at the comments?

peterbarker:
https://github.com/ardupilot/ardupilot/blob/master/Tools/autotest/test_build_options.py#L205

```
                feature_define_whitelist = set([
                    'AP_RANGEFINDER_ENABLED',  # only at vehicle level ATM
                    'HAL_PERIPH_SUPPORT_LONG_CAN_PRINTF',  # no symbol
                    'AP_PROXIMITY_HEXSOONRADAR_ENABLED',  # this shares symbols with AP_PROXIMITY_MR72_ENABLED
                    'AP_PROXIMITY_MR72_ENABLED',    # this shares symbols with AP_PROXIMITY_HEXSOONRADAR_ENABLED
                    'AP_RANGEFINDER_NRA24_CAN_ENABLED',
                    'AP_RANGEFINDER_HEXSOONRADAR_ENABLED',
                ])
```

PR #29461, opened 2025-03-08
Move RC protocol used for Aero from Linux HAL to AP_RCProtocol

peterbarker:
MAVProxy PR for testing is here: https://github.com/ArduPilot/MAVProxy/pull/1520

... near as I can tell both old and new code fails to lock against race conditions updating the RC data.


peterbarker:
Only compiler output change is aero:

```
Every 2.0s: ~/rc/ardupilot/Tools/scripts/filter_size_compare_branches_csv.py --hide /tmp/some.csv  ; wc /tmp/some.csv                                        threads: Sun Mar  9 18:39:51 2025

------------------------  -----  ------  -----  -----  -----  -----
Board                     blimp  copter  heli   plane  rover  sub
SITL_arm_linux_gnueabihf  0      0       0      0      0      0
SITL_x86_64_linux_gnu     0      0       0      0      0      0
aero                      66516  62280   66456  62248  62276  66564
------------------------  -----  ------  -----  -----  -----  -----
  361   361 10005 /tmp/some.csv
```

rmackay9:
I always say this kind of thing but I think the Intel Aero is long gone and it was never very popular.  If it's causing any kind of support overhead, I vote for removing it.

PR #29464, opened 2025-03-08
AP_Baro: Determine the result of the device data reading process

muramura:
Modify the description to clarify that no errors occur during data reading from the device.

### DIFF
--- a/libraries/AP_Baro/AP_Baro_BMP280.cpp
+++ b/libraries/AP_Baro/AP_Baro_BMP280.cpp
@@ -86,7 +86,9 @@ bool AP_Baro_BMP280::_init()
 
     // read the calibration data
     uint8_t buf[24];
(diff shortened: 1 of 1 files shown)


rmackay9:
This seems like a good change but we need to consider what happens when these reads fail.

Another possible change would be to at least initialise the buffers to zero.  I guess that the existing code could lead to uninitialised values being used.

peterbarker (approved):
LGTM.

Failing to register if we can't get the calibrations is entirely reasonable.

Using uninitialised value off the stack is a terrible response to not reading the reg data, so that's also a big improvement!

I have tested this on a SDModelH7v1 which has a BMP280 in it, and it's working just fine

peterbarker:
Merged, thanks!

rmackay9:

Issue #2716, opened 2015-08-19
Plane: rcin throttle jitter results in large throttle bumps


KIrill-ka:
That happens in FBWB mode, should be the same in CRUISE and AUTO with throttle nudge.
Input jitter is 1us. Output jitter is quite significant ~70us peak to peak which is almost 10% of TECS' throttle range.

After some digging into the problem I found two causes.
1. 1us jitter may result 1% jitter because the throttle is converted to 0-100 integer range.
2. TECS  reacts quite aggressively to a minor speed change fed to update_pitch_throttle.

Configuration info:
ArduPlane 3.3.0, PX4, s.bus connection to RX,
airspeed sensor enabled

Naterater:
@KIrill-ka, is this still an issue with newer firmware such as Plane 3.9? I have not witnessed the same behavior recently.

WickedShell:
@Naterater yes, it's due to internally quantizing all throttle on plane from 0-100 in steps of 1. This needs to change over time, but it's a large/slow change with a lot of implicit assumptions that might be missed.

KIrill-ka:
I'm sorry, I haven't tried the recent ardupilot versions. And I can't check it now since it's too cold for me to fly.

SlipstreamFPV:
What can be done to address this issue? It would seem RC filtering is a must particularly in auto modes.

It would also seem that the TECS controller is far too aggressive in chasing controller adjustments where it could have a slower response to controller changes while maintaining a fast response to external factors as required.

magicrub:
I actually have a fix for this on my fork but haven't brought it in yet. It's part of a zillion other changes that I need to get merged

SlipstreamFPV:
That's great news! I'm really looking forward to it being merged as I've struggled with this on a few planes. Thanks for all your hard work.

IamPete1:
The 100 steps on the output thing has been fixed by https://github.com/ArduPilot/ardupilot/pull/18706. We do have a similar issue on the input, https://github.com/ArduPilot/ardupilot/pull/18985

PR #18985, opened 2021-10-19
Increase manual throttle resolution by making scaled channels values float

shellixyz:
Complements #18706

PR #29465, opened 2025-03-08
RC_Channel: make pwm_to_angle_dz_trim and pwm_to_range_dz full resolution

andyp1per:
Before:
image

After:
image

### DIFF
--- a/libraries/RC_Channel/RC_Channel.cpp
+++ b/libraries/RC_Channel/RC_Channel.cpp
@@ -334,20 +334,20 @@ int16_t RC_Channel::get_control_mid() const
   return an "angle in centidegrees" (normally -4500 to 4500) from
   the current radio_in value using the specified dead_zone
  */
-int16_t RC_Channel::pwm_to_angle_dz_trim(uint16_t _dead_zone, uint16_t _trim) const
+float RC_Channel::pwm_to_angle_dz_trim(uint16_t _dead_zone, uint16_t _trim) const
 {
     int16_t radio_trim_high = _trim + _dead_zone;
     int16_t radio_trim_low  = _trim - _dead_zone;
 
-    int16_t reverse_mul = (reversed?-1:1);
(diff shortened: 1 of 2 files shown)


IamPete1:
This is a more complete version of this fix. It has lots of approvals, I'm not sure why we did not merge it. https://github.com/ArduPilot/ardupilot/pull/18985

This is a issue which this would resolve: https://github.com/ArduPilot/ardupilot/issues/2716

rmackay9:
I suspect the reason that https://github.com/ArduPilot/ardupilot/pull/18985 wasn't merged was simply that the user got frustrated that we dropped the ball on merging it and gave up.  That sometimes happens and while I think @peterbarker has been going through and reviving/closing some issues and PRs (at least he was for a time) we still miss some.

Perhaps I'll add a note to our [contributing code wiki page](https://ardupilot.org/dev/docs/contributing.html) to give advice for when PRs get stuck

tridge:
@andyp1per what is the theory of the regression? The reports seem to say that 4.5.7 isn't affected, and 4.5.7 uses integer maths for these range conversions

IamPete1:
> @andyp1per what is the theory of the regression? The reports seem to say that 4.5.7 isn't affected, and 4.5.7 uses integer maths for these range conversions

I don't think this is a regression. Because of the way the rounding works it can seem smooth if your throttle is say 50.5% because the noise is not enough to change the rounded result. But if your throttle is 50.1% then the noise means its bouncing between 49% and 50%.

andyp1per:
> I don't think this is a regression. Because of the way the rounding works it can seem smooth if your throttle is say 50.5% because the noise is not enough to change the rounded result. But if your throttle is 50.1% then the noise means its bouncing between 49% and 50%.

Right, this is pre-existing behaviour but it causes discontinuous spikes in the requested output

tridge (approved):

PR #29470, opened 2025-03-09
AP_Proximity: Group the drivers

muramura:
This is a supplement to commit e93594c4.

### DIFF
--- a/libraries/AP_Proximity/AP_Proximity.h
+++ b/libraries/AP_Proximity/AP_Proximity.h
@@ -87,11 +87,13 @@ public:
 #if AP_PROXIMITY_LD06_ENABLED
         LD06 = 16,
 #endif
+#if AP_PROXIMITY_MR72_DRIVER_ENABLED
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks for the PR.  I don't think we need this though.

AP_PROXIMITY_MR72_DRIVER_ENABLED is dependent upon the values of AP_PROXIMITY_MR72_ENABLED and AP_PROXIMITY_HEXSOONRADAR_ENABLED.

@peterbarker if you agree I think we can close this

peterbarker:
Agreed, we don't really need this one.

The defines are around this way because the options to be set on the custom build server are the hexsoon/nra defines, so stting the DRIVER define based on them makes sense.

We shouldn't merge this one because i(a) it disrupts the uniformity of that definition list (which is ugly enough as it is!), abd (b) it may lead to even nastier problems in the future if we add to this list (which we shouldn't...) and (c) the _DRIVER define is almost an internal detail and not fit for a header file here.

Thanks anyway, it was a reasonable observation.  Closing this one now.

PR #29474, opened 2025-03-09
Tools: drop Ubuntu Bionic and Debian Buster

tpwrules:
They have been past end of support for nearly one year and nearly two years respectively. Allows us to require at minimum Python 3.8.


peterbarker:
Users wanting these older releases can use older ArduPilot rleases to bring these environments up if required (and possible...)

PR #29475, opened 2025-03-09
SerialManager IOMCU fixups 

IamPete1:
Fix-ups for https://github.com/ArduPilot/ardupilot/pull/29094.

The IOMCU now shows up as a full serial port with parameters (CubeOrangePlus), so the param docs should be added so the user knows what 50 is. This also adds defines for the default baud and protocol so the params don't show up as changed by the user.

image

I'm not sure if the intention was that the params should show up at all on boards other than CubeRed?

There is also a now unused stats tracker that can be removed now the IOMCU is in the main serial array.


libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -1718,6 +1718,12 @@ def write_UART_config(self, f):
                 f.write(
                     '#define HAL_UART_IO_DRIVER constexpr ChibiOS::UARTDriver &uart_io = serial%sDriver;\n' % (index)
                 )
+                f.write(
```
peterbarker, line 1721: This makes
```
libraries/AP_Vehicle/AP_Vehicle.cpp:361:        serial_manager.set_protocol_and_baud(HAL_UART_IOMCU_IDX, AP_SerialManager::SerialProtocol_IOMCU, 0);
```

redundant - I think we should remove the code in AP_Vehicle.

I think it would be nice to do away with `HAL_UART_IOMCU_IDX` entirely.  I'd really like to make IOMCU "just another serial protocol"...

peterbarker (changes requested):

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
IamPete1: It seems dumb to expose the parameters tho. Under what circumstances would the user ever change them (except CubeRed).

libraries/AP_HAL_ChibiOS/hwdef/CubeRedPrimary/hwdef.dat
```diff
@@ -287,10 +287,6 @@ SERIAL_ORDER OTG1 USART2 USART6 USART3 UART4 UART8 OTG2 UART7
 EXT_FLASH_SIZE_MB 32
 INT_FLASH_PRIMARY 1
 
-# set protocol for UART7 to SerialProtocol_PPP
-define DEFAULT_SERIAL7_PROTOCOL SerialProtocol_PPP
-define DEFAULT_SERIAL7_BAUD 8000000
```
IamPete1, line 292: These conflict with setting UART7 to IOMCU on the next line.

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -184,7 +184,7 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
...
+    // @Values: -1:None, 1:MAVLink1, 2:MAVLink2, 3:Frsky D, 4:Frsky SPort, 5:GPS, 7:Alexmos Gimbal Serial, 8:Gimbal, 9:Rangefinder, 10:FrSky SPort Passthrough (OpenTX), 11:Lidar360, 13:Beacon, 14:Volz servo out, 15:SBus servo out, 16:ESC Telemetry, 17:Devo Telemetry, 18:OpticalFlow, 19:RobotisServo, 20:NMEA Output, 21:WindVane, 22:SLCAN, 23:RCIN, 24:EFI Serial, 25:LTM, 26:RunCam, 27:HottTelem, 28:Scripting, 29:Crossfire VTX, 30:Generator, 31:Winch, 32:MSP, 33:DJI FPV, 34:AirSpeed, 35:ADSB, 36:AHRS, 37:SmartAudio, 38:FETtecOneWire, 39:Torqeedo, 40:AIS, 41:CoDevESC, 42:DisplayPort, 43:MAVLink High Latency, 44:IRC Tramp, 45:DDS XRCE, 46:IMUDATA, 48:PPP, 49:i-BUS Telemetry, 50: IOMCU
```
tridge, line 187: serial manager does not control the IOMCU uart, so we should not have the parameter exposed. If the parameter is exposed it should do something

tridge:
@bugobliterator can you please check? did you deliberately make SERIAL7 appear in param list for CubeOrangePlus and other IOMCU boards? That is a bad user experience showing parameters that do nothing

bugobliterator:
> @bugobliterator can you please check? did you deliberately make SERIAL7 appear in param list for CubeOrangePlus and other IOMCU boards? That is a bad user experience showing parameters that do nothing

@tridge yes that unintentional. Here is a fix that should solve this issue. https://github.com/ArduPilot/ardupilot/pull/29493

libraries/AP_HAL_ChibiOS/hwdef/CubeRedPrimary/hwdef.dat
bugobliterator: No it doesn't this is a default parameter if IOMCU is disabled. This is a necessary change.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
bugobliterator: Again this is not correct for CubeRed. please check above comment for this. The parameter will no longer be exposed if not desired, fix for that is here https://github.com/ArduPilot/ardupilot/pull/29493

libraries/AP_SerialManager/AP_SerialManager.cpp
bugobliterator: fixed here https://github.com/ArduPilot/ardupilot/pull/29493

bugobliterator (changes requested):
the stats changes are correct, except that all the rest of the changes will break CubeRed workflow. 

The current behaviour is that the serial protocol is set to IOMCU for internal use only, it shouldn't have been exposed for other boards, that's oversight on my part. I have a fix for that here https://github.com/ArduPilot/ardupilot/pull/29493

For cubered this parameter is defaulted to PPP but if BRD_IO_ENABLE is set its changed to IO Protocol in AP_Vehicle, that is the desired behaviour. And the serial is visible in parameters as well, for users to modify as well.

IamPete1:
@bugobliterator What is the plan with the serial port params on CubeRed? Are the only valid values IOMCU or PPP? In which case I don't think you need to expose the params, the switch can be done on only board IO enable param? If the params are exposed setting board IO enable should set and default the parameters. My goal with adding the defaults is to make sure that parameters that the user did not change don't show up as not default values in the log.

libraries/AP_SerialManager/AP_SerialManager.cpp
Hwurzburg: so is this going to be removed? if not, put a WIKI NEEDED label on the PR please

bugobliterator:
> @bugobliterator What is the plan with the serial port params on CubeRed? Are the only valid values IOMCU or PPP? In which case I don't think you need to expose the params, the switch can be done on only board IO enable param? If the params are exposed setting board IO enable should set and default the parameters. My goal with adding the defaults is to make sure that parameters that the user did not change don't show up as not default values in the log.

Actually for CubeRed primary secondary comms can be DDS and mavlink as well, nothing is set in stone there, and would like to keep all the options open there. As far as the change to not show up as non-default value, I guess we can change the default value of the parameter instead of just setting the value itself.

libraries/AP_SerialManager/AP_SerialManager.cpp
IamPete1: Since the param is still exposed on CubeRed we need the value in the description.

libraries/AP_HAL_ChibiOS/hwdef/CubeRedPrimary/hwdef.dat
IamPete1: removed

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
IamPete1: removed

IamPete1:
@bugobliterator Thanks for the explination. I have added what I think is missing from https://github.com/ArduPilot/ardupilot/pull/29493. 

The only thing that remains is there is no way for the hardware report WebTool to know which logged UART instance is the IOMCU. I had it as a special case of 100 but that was removed in https://github.com/ArduPilot/ardupilot/pull/29094. For all the other UARTs I can lookup the params based on the index to find the protocol. After https://github.com/ArduPilot/ardupilot/pull/29493 that method won't work on the IOMCU anymore. We could hard coded it back to 100 for the instance which equals `HAL_UART_IOMCU_IDX` but that breaks the ability to look it up from the params on CubeRed. I guess we could work it out from the new `HAL_HIDE_SERIALx_PARAMS` and the existing `HAL_HAVE_SERIALx`. Any better ideas?

For reference this is the what the tool used to plot:
image

bugobliterator:
> @bugobliterator Thanks for the explination. I have added what I think is missing from #29493 and removed everything else.
> 
> The only thing that remains is there is no way for the hardware report WebTool to know which logged UART instance is the IOMCU. I had it as a special case of 100 but that was removed in #29094. For all the other UARTs I can lookup the params based on the index to find the protocol. After #29493 that method won't work on the IOMCU anymore. We could hard coded it back to 100 for the instance which equals `HAL_UART_IOMCU_IDX` but that breaks the ability to look it up from the params on CubeRed. I guess we could work it out from the new `HAL_HIDE_SERIALx_PARAMS` and the existing `HAL_HAVE_SERIALx`. Any better ideas?

@IamPete1 I think we can simply add protocol field to uart_info.txt. Where it can show which protocol its currently set to if available.

tridge (approved):

PR #29477, opened 2025-03-10
Vagrant: remove ancient machines from config

peterbarker:
users will need to be running older codebases to have these at all work, so having the old Vagrantfile from those releases is entirely reasonable


tpwrules:
Glad to see these go to. Related cleanup: https://github.com/ArduPilot/ardupilot/blob/1ca1c57016c38b7405ce7229daa5a10777a5853f/Tools/vagrant/initvagrant.sh#L71

peterbarker:
> Glad to see these go too. Related cleanup:
> 
> https://github.com/ArduPilot/ardupilot/blob/1ca1c57016c38b7405ce7229daa5a10777a5853f/Tools/vagrant/initvagrant.sh#L71

Removed it, thanks!

PR #29480, opened 2025-03-10
Plane: added fast attitude recovery for inverted flight in Q modes - rebased

priseborough:
Refer to https://github.com/ArduPilot/ardupilot/pull/28320 for discussion and review comments.

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -1937,6 +1937,18 @@ void QuadPlane::motors_output(bool run_rate_controller)
         if (now - last_att_control_ms > 100) {
             // relax if have been inactive
             relax_attitude_control();
+            /*
+              when starting the motors apply a limit to the target
+              thrust angle to allow us to rapidly recover from an
+              upset attitude.
+              We add 10% past our limit so we don't trigger this
+              recovery code too often for normal fixed wing flight
+             */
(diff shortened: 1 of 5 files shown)


IamPete1:
If I recall correctly I think we were happy with this as a implementation of one approach but we were not sure of the best approach. This still takes the shortest path to level, we may want to do some alternate path such as to prefer rolling to try and minimize aero loads. This also effectively bypasses the accel limits causing a "maximum effort" recovery, the point is that its faster, but it might also be so fast as to break the vehicle. 

I guess really we can rank methods as a trade off between aggressiveness and speed. 

Slowest and least aggressive.

- accel limited and aerodynamically aware

- accel limited shortest path (this is what we have now)

- maximum effort aerodynamically aware

- maximum effort shortest path (this is what is in this PR)

Fastest and most aggressive.

Which one is best depends on the vehicle and the situation.

tridge:
@priseborough @IamPete1 I've opened a different PR with a quite different approach which is simpler and produces very good results in SITL testing: https://github.com/ArduPilot/ardupilot/pull/29498

tridge:
@priseborough the test shows height loss of over 17m on recovery with this PR. It shows zero height loss with #29498

priseborough:
I prefer the https://github.com/ArduPilot/ardupilot/pull/29498 approach

PR #29489, opened 2025-03-10
Tools: tidy-up sitl_periph_xx defines assignment

shiv-tyagi:
Midway through renaming the HAL_PERIPH_ENABLE_XX defines in recent PRs, I realised that instead of setting these defines individually for every sitl_periph target, we can default them to 0 for all and explicitly set them to 1 wherever needed. This PR fixes them all.


peterbarker (approved):
LGTM.

PR #23287, opened 2023-03-22
waf: require python 3.8.0 or later

tridge:
for 4.5.x we no longer support py2

PR #29490, opened 2025-03-10
ardupilotwaf: require minimum Python 3.8

tpwrules:
Makes the comment honest.

Requires upgrading Cygwin to Python 3.9. Python 3.12 is available but it doesn't have all our packages. I think we will have to modernize a bit more.

### DIFF
--- a/wscript
+++ b/wscript
@@ -548,7 +548,7 @@ def configure(cfg):
 
     # require python 3.8.x or later
     cfg.load('python')
-    cfg.check_python_version(minver=(3,6,9))
+    cfg.check_python_version(minver=(3,8,0))
 
     cfg.load('ap_library')
 
--- a/.github/workflows/cygwin_build.yml
+++ b/.github/workflows/cygwin_build.yml
@@ -160,7 +160,7 @@ jobs:
 
       - uses: cygwin/cygwin-install-action@v5
         with:
-          packages: cygwin64 gcc-g++=10.2.0-1 ccache python37 python37-future python37-lxml python37-pip python37-setuptools python37-wheel git procps gettext
+          packages: cygwin64 gcc-g++=10.2.0-1 ccache python39 python39-future python39-lxml python39-pip python39-setuptools python39-wheel git procps gettext
           add-to-path: false
       # Put ccache into github cache for faster build
       - name: setup ccache
--- a/Tools/environment_install/install-prereqs-windows-andAPMSource.ps1
+++ b/Tools/environment_install/install-prereqs-windows-andAPMSource.ps1
@@ -14,10 +14,10 @@ Write-Output "Downloading ARM GCC Compiler 10-2020-Q4-Major (3/8)"
 Start-BitsTransfer -Source "https://firmware.ardupilot.org/Tools/STM32-tools/gcc-arm-none-eabi-10-2020-q4-major-win32.exe" -Destination "$PSScriptRoot\gcc-arm-none-eabi-10-2020-q4-major-win32.exe"
 
 Write-Output "Installing Cygwin x64 (4/8)"
-Start-Process -wait -FilePath $PSScriptRoot\setup-x86_64.exe -ArgumentList "--root=C:\cygwin64 --no-startmenu --local-package-dir=$env:USERPROFILE\Downloads --site=http://cygwin.mirror.constant.com --packages autoconf,automake,ccache,cygwin32-gcc-g++,gcc-g++=7.4.0-1,libgcc1=7.4.0.1,gcc-core=7.4.0-1,git,libtool,make,gawk,libexpat-devel,libxml2-devel,python37,python37-future,python37-lxml,python37-pip,libxslt-devel,python37-devel,procps-ng,zip,gdb,ddd --quiet-mode"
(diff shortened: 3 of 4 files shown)


peterbarker:
Why is it that we want to move forward with our minimum Python version?

Doing so is likely to break a bunch of people's environments.

tpwrules:
I'm mostly proposing that we make the comment correct. If we want to keep 3.6 the minimum I suppose that's okay, but it has not seen a release in nearly four years and is nearly ten years old.

We already tried the move a couple years ago (which is where the comment came from) but it had to be reverted because Ubuntu 18.04 was still on life support. Now that that's gone let's move.

tpwrules:
Here are your previous thoughts: https://github.com/ArduPilot/ardupilot/pull/23287#discussion_r1174275730 :)

timtuxworth:
> Why is it that we want to move forward with our minimum Python version?
> 
> Doing so is likely to break a bunch of people's environments.

I don't understand why this is even a question Peter, like my mother always used to say, if "a bunch of people"  jumped off a cliff would you follow them?

wscript
```diff
@@ -548,7 +548,7 @@ def configure(cfg):
 
     # require python 3.8.x or later
     cfg.load('python')
-    cfg.check_python_version(minver=(3,6,9))
+    cfg.check_python_version(minver=(3,8,0))
```
tridge, line 551: maybe if version_is_3.6, print big warning, sleep 5 seconds to user can see it

timtuxworth: < 3.8?

peterbarker:
Incidentally, if you can show we're currently broken on 3.6.9 / 3.7.0 we could probably merge this with the 3.8.0 requirement.

peterbarker:
Ping @tpwrules - testing requested here.

Hopefully you can show that we're just broken on <3.8 and we can just specify it as the minimum required version :-)

tridge does want the cygwin binary downloaded and tested, 'though.

khancyr:
ubuntu 20.04 that is the minimal we are supporting is python 3.8 so that would work as the basis, under this the pip install will probably failed to find the right packages version anyway

peterbarker:
pymavlink is now broken on Python 3.6.9 (eg. Bionic):
```
    bionic64:     Complete output from command python setup.py egg_info:
    bionic64:     Traceback (most recent call last):
    bionic64:       File "<string>", line 1, in <module>
    bionic64:       File "/tmp/pip-build-6wWD6q/pymavlink/setup.py", line 15, in <module>
    bionic64:         from Cython.Build import cythonize
    bionic64:     ImportError: No module named Cython.Build
```
```
vagrant@ubuntu-bionic:~$ python3 --version
Python 3.6.9
```

tpwrules:
The Linux build requires syntax features only in 3.8:

```
$ ./waf configure --board=linux
Setting top to                           : /home/tpwatson/ardusimstuff/ardupilot 
Setting out to                           : /home/tpwatson/ardusimstuff/ardupilot/build 
Autoconfiguration                        : enabled 
Checking for program 'python'            : /usr/bin/python3 
Checking for python version >= 3.6.9     : 3.6.9 
Setting board to                         : linux 
Traceback (most recent call last):
  File "/home/tpwatson/ardusimstuff/ardupilot/modules/waf/waflib/Scripting.py", line 159, in waf_entry_point
    run_commands()
  File "/home/tpwatson/ardusimstuff/ardupilot/modules/waf/waflib/Scripting.py", line 255, in run_commands
    ctx = run_command(cmd_name)
  File "/home/tpwatson/ardusimstuff/ardupilot/modules/waf/waflib/Scripting.py", line 239, in run_command
    ctx.execute()
  File "/home/tpwatson/ardusimstuff/ardupilot/modules/waf/waflib/Configure.py", line 159, in execute
    super(ConfigurationContext, self).execute()
  File "/home/tpwatson/ardusimstuff/ardupilot/modules/waf/waflib/Context.py", line 214, in execute
    self.recurse([os.path.dirname(g_module.root_path)])
  File "/home/tpwatson/ardusimstuff/ardupilot/modules/waf/waflib/Context.py", line 296, in recurse
    user_function(self)
  File "/home/tpwatson/ardusimstuff/ardupilot/wscript", line 579, in configure
    cfg.get_board().configure(cfg)
  File "Tools/ardupilotwaf/boards.py", line 1464, in configure
    cfg.load('linux')
  File "/home/tpwatson/ardusimstuff/ardupilot/modules/waf/waflib/Configure.py", line 257, in load
    module = Context.load_tool(tool, tooldir, ctx=self, with_sys_path=with_sys_path)
  File "/home/tpwatson/ardusimstuff/ardupilot/modules/waf/waflib/Context.py", line 735, in load_tool
    __import__(x % tool)
  File "Tools/ardupilotwaf/linux.py", line 17, in <module>
    import linux_hwdef  # noqa:501
  File "<fstring>", line 1
    (len(dev)=)
             ^
SyntaxError: invalid syntax
```

tpwrules:
But our wscript is rather gigantic and I'm not sure how to check before triggering a syntax error in the general case. Do we just have to keep the wscript compatible with any Python? Or accept that old Pythons might give syntax errors before we can complain about the version?

peterbarker (approved):
Need to test Windows binary

robertlong13:
Rebased to generate fresh artifacts to test

robertlong13:
Took a quadplane and a copter for a basic spin. Looks fine.

PR #29491, opened 2025-03-10
Allow E-Stop during motor test and allow changing values on plane

IamPete1:
Currently motor test by passes E-Stop. Copter will refuse to start a test if E-Stop is high but changing E-Stop state during a test will have no affect. Plane just ignores E-Stop.

Motors does stop outputting the test sequence once E-Stop goes high, but because it did not then output 0 throttle the last values were still getting sent to the motors. The fix is to output min if motors are not armed or the interlock is not set (this comes from E-Stop). That change means that the motor do stop if E-Stop is hit during a test.

Now motors do stop it is possible to have a ongoing motor test without noticing, so the `output_test_seq` method now returns false if it could not output that ends the motor test and sends "Motor Test: cancelled" message. 

Plane currently will not accept any motor test commands if the motors are armed, which they are during a motor test. This means that the "test all" and excitingly the "stop all" buttons in mission planner don't work. The fix is to allow new command is a existing motor test is ongoing, this is inline with copter.

image

Copter has lots of checks it runs before allowing motor test, it would be nice to bring those over to plane at some point.


rmackay9 (approved):
LGTM thanks.

I guess the message to the user, "Motor Test: cancelled", doesn't tell the user too much about why it was cancelled but it's probably OK.

PR #29494, opened 2025-03-11
Plane: Output rudder in VTOL modes

IamPete1:
Currently rudder is not output at all in VTOL modes. This means it is stuck at whatever the output was when the mode was entered. For example full rudder due to stick mixing in guided then switch to QRTL, rudder output remains even after pilot stick is centered.

image

This results in the vehicle crabbing all the way home.

image

With this PR the rudder it output as normal in QRTL. This includes the VTOL phases which is inline with what a nav land would do in AUTO.

image

For the other VTOL modes the rudder is now held centered. Using coordinated turn value seems a bit nonsense when we could be moving backwards or sideways.

### DIFF
--- a/ArduPlane/mode_qautotune.cpp
+++ b/ArduPlane/mode_qautotune.cpp
@@ -35,6 +35,9 @@ void ModeQAutotune::run()
     // Stabilize with fixed wing surfaces
     plane.stabilize_roll();
     plane.stabilize_pitch();
+
+    // Center rudder
+    output_rudder_and_steering(0.0);
 }
 
 void ModeQAutotune::_exit()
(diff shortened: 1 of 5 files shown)


IamPete1:
> What about QLand? (or Qacro not that I every fly it)

Q-Land inherits from Q-Loiter:

https://github.com/ArduPilot/ardupilot/blob/9fb1e2b9f7b746716397aa1d298943c201d9dd8a/ArduPlane/mode_qland.cpp#L32

Q-Acro inherits from Acro:
https://github.com/ArduPilot/ardupilot/blob/9fb1e2b9f7b746716397aa1d298943c201d9dd8a/ArduPlane/mode_qacro.cpp#L76

> What about Auto? (e.g. VTOL Takeoff, VTOL Land and I guess VTOL Transition)

Auto already outputs rudder, nothing changes as a result of this PR but we may want to change it in the future.

tridge (approved):

tridge:
I've added this to 4.6 as it can impact recovery using QHover/QLoiter after a stall

rmackay9:
this is included in 4.6.0-beta5 via PR https://github.com/rmackay9/rmackay9-ardupilot/pull/37

PR #29498, opened 2025-03-11
Plane: fast attitude recovery for quadplanes

tridge:
This is a rework of #28320 which uses the fixed wing attitude controller for recovery instead of modifying the VTOL controller. This gives much faster attitude recovery and also avoids an issue with the way the quaternion controller may try to pitch down while inverted, leading to a large amount of altitude loss
This approach builds on the way we already use the fixed wing controller during forward transitions, so it is very little code change, and testing shows it produces a very small amount of altitude loss (eg. in the autotest the alt loss is usually around 2m from fully inverted)

Update: this PR now makes the use of the forced fixed wing controller optional using Q_ASSIST_OPTIONS (defaults enabled) and adds spin recovery, also optional.

### DIFF
--- a/ArduPlane/Attitude.cpp
+++ b/ArduPlane/Attitude.cpp
@@ -378,6 +378,10 @@ void Plane::stabilize_yaw()
         SRV_Channels::set_output_scaled(SRV_Channel::k_steering, steering_output);
     }
 
+#if HAL_QUADPLANE_ENABLED
+    // possibly recover from a spin
+    quadplane.assist.output_spin_recovery();
+#endif
 }
 
 /*
--- a/ArduPlane/VTOL_Assist.cpp
+++ b/ArduPlane/VTOL_Assist.cpp
@@ -140,4 +140,94 @@ bool VTOL_Assist::should_assist(float aspeed, bool have_airspeed)
     return force_assist || speed_assist || alt_error.is_active() || angle_error.is_active();
 }
 
+/*
+  check if we are in VTOL recovery
+*/
+bool VTOL_Assist::check_VTOL_recovery(void)
+{
+    const bool allow_fw_recovery =
+        !option_is_set(OPTION::FW_FORCE_DISABLED) &&
+        !quadplane.tailsitter.enabled() &&
+        plane.control_mode != &plane.mode_qacro;
+    if (!allow_fw_recovery) {
+        quadplane.force_fw_control_recovery = false;
+        quadplane.in_spin_recovery = false;
+        return false;
+    }
+
+    // see if the attitude is outside twice the Q_ANGLE_MAX
+    const auto &ahrs = plane.ahrs;
+    const int16_t angle_max_cd = quadplane.aparm.angle_max;
+    if ((abs(ahrs.roll_sensor) > 2*angle_max_cd ||
+         abs(ahrs.pitch_sensor) > 2*angle_max_cd)) {
+            // we are 2x the angle limits trigger fw recovery
+        quadplane.force_fw_control_recovery = true;
+    }
+
+    if (quadplane.force_fw_control_recovery) {
+        // stop fixed wing recovery if inside Q_ANGLE_MAX
+        if ((abs(ahrs.roll_sensor) <= angle_max_cd &&
+             abs(ahrs.pitch_sensor) <= angle_max_cd)) {
+            quadplane.force_fw_control_recovery = false;
+            quadplane.attitude_control->reset_target_and_rate(false);
+
+            if (ahrs.groundspeed() > quadplane.wp_nav->get_default_speed_xy()*0.01) {
+                /* if moving at high speed also reset position
+                   controller and height controller
+
+                   this avoids an issue where the position
+                   controller may limit pitch after a strong
+                   acceleration event
+                */
+                quadplane.pos_control->init_z_controller();
+                quadplane.pos_control->init_xy_controller();
+            }
+        }
+    }
+
+    if (!option_is_set(OPTION::SPIN_DISABLED) &&
+        quadplane.force_fw_control_recovery) {
+        // additionally check for needing spin recovery
+        const auto &gyro = plane.ahrs.get_gyro();
+        quadplane.in_spin_recovery =
+            fabsf(gyro.z) > radians(10) &&
+            fabsf(gyro.x) > radians(30) &&
+            fabsf(gyro.y) > radians(30) &&
+            gyro.x * gyro.z < 0 &&
+            plane.ahrs.pitch_sensor < -4500;
+    } else {
+        quadplane.in_spin_recovery = false;
+    }
+    
+    return quadplane.force_fw_control_recovery;
+}
+
+
+/* if we are in a spin then counter with rudder and elevator
+
+   if roll rate and yaw rate are opposite and yaw rate is
+   significant then put in full rudder to counter the yaw rate
+   for spin recovery
+*/
+void VTOL_Assist::output_spin_recovery(void)
+{
+    if (!quadplane.in_spin_recovery) {
+        return;
+    }
+    if (quadplane.motors->get_desired_spool_state() !=
+        AP_Motors::DesiredSpoolState::THROTTLE_UNLIMITED) {
+        // if we and no longer running the VTOL motors we need to
+        // clear the spin flag
+        quadplane.in_spin_recovery = false;
+        return;
+    }
+    const Vector3f &gyro = plane.ahrs.get_gyro();
+
+    // put in opposite rudder to counter yaw, and neutral
+    // elevator until we're out of the spin
+    SRV_Channels::set_output_scaled(SRV_Channel::k_rudder, gyro.z > 0 ? -SERVO_MAX : SERVO_MAX);
+    SRV_Channels::set_output_scaled(SRV_Channel::k_elevator, 0);
+}
+
+
 #endif  // HAL_QUADPLANE_ENABLED
--- a/ArduPlane/VTOL_Assist.cpp
+++ b/ArduPlane/VTOL_Assist.cpp
@@ -158,16 +158,16 @@ bool VTOL_Assist::check_VTOL_recovery(void)
     // see if the attitude is outside twice the Q_ANGLE_MAX
     const auto &ahrs = plane.ahrs;
     const int16_t angle_max_cd = quadplane.aparm.angle_max;
-    if ((abs(ahrs.roll_sensor) > 2*angle_max_cd ||
-         abs(ahrs.pitch_sensor) > 2*angle_max_cd)) {
-            // we are 2x the angle limits trigger fw recovery
+    const float abs_angle_cd = fabsf(Vector2f{float(ahrs.roll_sensor), float(ahrs.pitch_sensor)}.length());
+
+    if (abs_angle_cd > 2*angle_max_cd) {
+        // we are 2x the angle limits, trigger fw recovery
         quadplane.force_fw_control_recovery = true;
     }
 
     if (quadplane.force_fw_control_recovery) {
         // stop fixed wing recovery if inside Q_ANGLE_MAX
-        if ((abs(ahrs.roll_sensor) <= angle_max_cd &&
-             abs(ahrs.pitch_sensor) <= angle_max_cd)) {
+        if (abs_angle_cd <= angle_max_cd) {
             quadplane.force_fw_control_recovery = false;
             quadplane.attitude_control->reset_target_and_rate(false);
 
--- a/ArduPlane/mode.cpp
+++ b/ArduPlane/mode.cpp
@@ -156,6 +156,12 @@ bool Mode::enter()
             bool have_airspeed = quadplane.ahrs.airspeed_estimate(aspeed);
             quadplane.assisted_flight = quadplane.assist.should_assist(aspeed, have_airspeed);
         }
+
+        if (is_vtol_mode() && !quadplane.tailsitter.enabled()) {
+            // if flying inverted and entering a VTOL mode cancel
+            // inverted flight
+            plane.inverted_flight = false;
+        }
 #endif
     }
 
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -875,7 +875,7 @@ void QuadPlane::run_esc_calibration(void)
  */
 void QuadPlane::multicopter_attitude_rate_update(float yaw_rate_cds)
 {
-    bool use_multicopter_control = in_vtol_mode() && !tailsitter.in_vtol_transition();
+    bool use_multicopter_control = in_vtol_mode() && !tailsitter.in_vtol_transition() && !force_fw_control_recovery;
     bool use_yaw_target = false;
 
     float yaw_target_cd = 0.0;
@@ -1954,7 +1954,40 @@ void QuadPlane::motors_output(bool run_rate_controller)
         if (now - last_att_control_ms > 100) {
             // relax if have been inactive
             relax_attitude_control();
+            /*
+              when starting the VTOL motors force use of the fixed
+              wing controller for attitude control if we are outside
+              of the Q_ANGLE_MAX limit. We keep it forced until we
+              have recovered to within the limit
+
+              This avoids an issue with the attitude controller
+              limiting the rate of recovery and also an issue with the
+              quaternion controller trying to fix the attitude in a
+              way that is bad for fixed wing aircraft (eg. putting
+              nose down when inverted, leading to rapid loss of
+              height)
+
+              we don't do this in QACRO or tailsitter modes, as this
+              type of attitude is expected
+            */
+            if (!tailsitter.enabled() &&
+                plane.control_mode != &plane.mode_qacro &&
+                (abs(ahrs.roll_sensor) > aparm.angle_max ||
+                 abs(ahrs.pitch_sensor) > aparm.angle_max)) {
+                force_fw_control_recovery = true;
+            }
         }
+        if (force_fw_control_recovery) {
+            // see if we have recovered attitude
+            if (abs(ahrs.roll_sensor) <= aparm.angle_max &&
+                abs(ahrs.pitch_sensor) <= aparm.angle_max) {
+                force_fw_control_recovery = false;
+                // reset the attitude target to the new attitude so
+                // the VTOL controller doesn't pull us back
+                attitude_control->reset_target_and_rate(false);
+            }
+        }
+
         // run low level rate controllers that only require IMU data and set loop time
         const float last_loop_time_s = AP::scheduler().get_last_loop_time_s();
         motors->set_dt(last_loop_time_s);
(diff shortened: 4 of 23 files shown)


ArduPlane/mode.cpp
```diff
@@ -153,6 +153,12 @@ bool Mode::enter()
             quadplane.assisted_flight = quadplane.assist.should_assist(aspeed, have_airspeed);
         }
 #endif
+
+        if (is_vtol_mode()) {
```
IamPete1, line 157: Tail-sitters deal with inverted flight correctly, if inverted they transition by putting the nose "down" rather than up.

ArduPlane/quadplane.cpp
```diff
@@ -1937,7 +1937,38 @@ void QuadPlane::motors_output(bool run_rate_controller)
...
+              nose down when inverted, leading to rapid loss of
+              height)
+
+              we don't do this in QACRO or tailsitter modes, as this
+              type of attitude is expected
+            */
+            if (!tailsitter.enabled() &&
+                plane.control_mode != &plane.mode_qacro &&
+                (abs(ahrs_view->roll_sensor) > aparm.angle_max ||
+                 abs(ahrs_view->pitch_sensor) > aparm.angle_max)) {
```
IamPete1, line 1959: I think this should use the larger of angle max and the fixedwing angle limits. Most vehicles have a larger roll limit in FW flight than they do in VTOL.

ArduPlane/quadplane.cpp
```diff
@@ -1937,7 +1937,38 @@ void QuadPlane::motors_output(bool run_rate_controller)
...
+                force_fw_control_recovery = true;
+            }
         }
+        if (force_fw_control_recovery) {
+            // see if we have recovered attitude
+            if (abs(ahrs_view->roll_sensor) <= aparm.angle_max &&
+                abs(ahrs_view->pitch_sensor) <= aparm.angle_max) {
+                force_fw_control_recovery = false;
+                relax_attitude_control();
```
IamPete1, line 1968: There is some stuff in here we would not want to do. For example this resets all the target/error/derivative filtering on the rate PIDs. Really we want to rest the attitude controller but not the rate controller. `attitude_control->reset_target_and_rate()`  does what we need I think.

ArduPlane/quadplane.cpp
```diff
@@ -1937,7 +1937,38 @@ void QuadPlane::motors_output(bool run_rate_controller)
...
+                plane.control_mode != &plane.mode_qacro &&
+                (abs(ahrs_view->roll_sensor) > aparm.angle_max ||
+                 abs(ahrs_view->pitch_sensor) > aparm.angle_max)) {
+                force_fw_control_recovery = true;
+            }
         }
+        if (force_fw_control_recovery) {
+            // see if we have recovered attitude
+            if (abs(ahrs_view->roll_sensor) <= aparm.angle_max &&
+                abs(ahrs_view->pitch_sensor) <= aparm.angle_max) {
```
IamPete1, line 1966: When using the fixed wing controller it is flying to the normal AHRS (not the view). So using the view to check for completion may never work. For example if I had a `Q_TRIM_PITCH` of 15 and a `Q_ANGLE_MAX` of 10 this check would not pass.

ArduPlane/quadplane.h
```diff
@@ -632,6 +632,9 @@ class QuadPlane
     // ignored unless OPTION_DELAY_ARMING or OPTION_TILT_DISARMED is set
     bool delay_arming;
 
+    // should we force use of fixed wing controller for attitude upset recovery?
+    bool force_fw_control_recovery;
```
IamPete1, line 636: I would like to see this in `QTUN`, it could be a new bit in the assistance bitmask.

https://github.com/ArduPilot/ardupilot/blob/9fb1e2b9f7b746716397aa1d298943c201d9dd8a/ArduPlane/quadplane.cpp#L3593-L3600

ArduPlane/quadplane.cpp
IamPete1: I wonder if we could re-use the angle assistance thresholds, the use case is the same really.

https://github.com/ArduPilot/ardupilot/blob/9fb1e2b9f7b746716397aa1d298943c201d9dd8a/ArduPlane/VTOL_Assist.cpp#L121-L133

Don't need the trigger delay of course in this case. We may not even have to check here, we could check if we were in angle assistance when the VTOL mode was entered.

IamPete1 (commented):
This effectively recovers as it would in Q-Assist in a fixed wing flight. One issues I have seen on a real vehicle is that the fixed wing angle controllers have a bad time unusual attitudes. A vehicle got into a inverted spin and was no able to recover in Q-Assist. The issue was three fold, it was oscillating about roll +-180 and could not decide which way to go. The fixed-wing controllers hard limit angle error so the error was not building the longer it was upside down (as it would in the VTOL controller), so it never tried that hard to recover. The VTOL rate controller I terms are also relaxed in assist so again it was never trying that hard. In that case we were able to switch to a VTOL mode and have it recover quickly (although not quite quick enough to prevent the crash).

One advantage of this method is that we can say that it is not anything that the vehicle would not do in Q-Assist so were not in any more risk of overloading the vehicle that we would be in that case. 

One disadvantage is that the user may be trying the VTOL mode because Q-Assist may have failed to recover, in which case it makes sense to try a different approach.

ArduPlane/quadplane.cpp
tridge: good catch, thanks, fixed

ArduPlane/quadplane.cpp
tridge: you're right, thanks, fixed

ArduPlane/quadplane.h
tridge: added, thanks

ArduPlane/quadplane.cpp
IamPete1: My thinking was just that we don't want a change in behavior from what we have now in the case that a user switches modes when the vehicle is within the expected attitude envelope for fixed-wing flight.

Is the goal that this should be a special case that kick in to recover from a unexpected attitude or should it kick in the course of a typical flight?

tridge:
> One issue I have seen on a real vehicle is that the fixed wing angle controllers have a bad time unusual attitudes. A vehicle got into a inverted spin and was no able to recover in Q-Assist. The issue was three fold, it was oscillating about roll +-180 and could not decide which way to go. The fixed-wing controllers hard limit angle error so the error was not building the longer it was upside down (as it would in the VTOL controller), so it never tried that hard to recover.

my preference is to apply this method with those limitations and work on improving the fixed wing attitude controller to better cope with these scenarios.

ArduPlane/mode.cpp
tridge: can you explain the scenario a bit more? the flag being reset is the one from the AUX switch. If this is dangerous then the use clicking off the inverted flight mode AUX switch would kill tailsitters

IamPete1: If the vehicle is flying inverted but the bit is cleared as the transition starts then it will try and reach level by pitching "up" and end up doing a 270 deg pitch (of course it won't make it because it will be trying to roll and the transition will force complete due to the roll error). With the inverted flag set it will try and reach level by pitching "down" and just to the 90 deg pitch. 

If you flip the switch in normal fixed wing flight then it works exactly as a normal plane. Once your in VTOL and the transition is done the switch has no affect.

Tailsitters could certainly clear the bit once the VTOL transition is done. Just it needs to still be set for the duration of the transition.

The support for this was added in https://github.com/ArduPilot/ardupilot/pull/17118. I though I had done a demo vid at the time, but I can't find it.

ArduPlane/quadplane.cpp
tridge: I actually do want a change in behaviour. When the two angle limits are similar it doesn't matter, but when the fixed wing roll limit is a lot larger than the VTOL limit then the vehicle starts flying outside the envelope where it has been tuned and I think it is a reasonable to expect the vehicle to come back within the VTOL angle limit as quickly as possible.

ArduPlane/mode.cpp
tridge: thanks for the explanation! I just tried it with the CatTailsitter and it matches your description. I've changed the code to clear the inverted_flight flag on tailsitters at the completion of the transition to VTOL

tridge:
@IamPete1 I've added in the attitude control method as discussed, plus the indecision change for the FW roll controller
in the new test added in this PR we now lose no altitude at all (usually gains about 0.1 or 0.2 m in recovery). Without these changes we lose around 25m or so

tridge:
need to do testing on some real quadplanes

ArduPlane/quadplane.cpp
```diff
@@ -4190,13 +4227,25 @@ bool QuadPlane::use_fw_attitude_controllers(void) const
...
+        if (in_vtol_mode()) {
+            // in VTOL modes always slave fixed wing to VTOL rate control
+            return false;
+        }
+
+        if (transition->use_multirotor_control_in_fwd_transition()) {
+            /*
+              special case for vectored yaw tiltrotors in forward
+              transition, keep multicopter control until we reach
+              target transition airspeed. This condition needs to
```
IamPete1, line 4243: This condition needs to.... ?

ArduPlane/tailsitter.cpp
```diff
@@ -562,6 +562,9 @@ bool Tailsitter::transition_vtol_complete(void) const
     const float trans_angle = get_transition_angle_vtol();
     if (labs(plane.ahrs.pitch_sensor) > trans_angle*100) {
         gcs().send_text(MAV_SEVERITY_INFO, "Transition VTOL done");
+        // clear inverted flight flag to make behaviour consistent
+        // with other quadplane types
+        plane.inverted_flight = false;
```
IamPete1, line 567: There are lots of other cases that return true from this function. You could put it here:

https://github.com/ArduPilot/ardupilot/blob/f328a659ef6a0acea80c7fb16c2c621fb1f3be15/ArduPlane/tailsitter.cpp#L900

I do think its OK to leave the inverted bit set, it should work as expected for tailsittters, just its inconsistent with normal quadplanes.

libraries/APM_Control/AP_RollController.cpp
```diff
@@ -202,13 +202,26 @@ float AP_RollController::get_servo_out(int32_t angle_err, float scaler, bool dis
...
+    }
+
     // Limit the demanded roll rate
     if (gains.rmax_pos && desired_rate < -gains.rmax_pos) {
         desired_rate = - gains.rmax_pos;
     } else if (gains.rmax_pos && desired_rate > gains.rmax_pos) {
         desired_rate = gains.rmax_pos;
     }
 
+    last_desired_rate = desired_rate;
```
IamPete1, line 223: This needs to be kept upto date in the `get_rate_out` method so it is still correct when we stop VTOL or acro. I think you may actually be able to remove it all together and use `pinfo.target`which is set here:

https://github.com/ArduPilot/ardupilot/blob/f328a659ef6a0acea80c7fb16c2c621fb1f3be15/libraries/APM_Control/AP_FW_Controller.cpp#L84

libraries/APM_Control/AP_RollController.cpp
```diff
@@ -202,13 +202,26 @@ float AP_RollController::get_servo_out(int32_t angle_err, float scaler, bool dis
...
     float desired_rate = angle_err_deg/ gains.tau;
 
+    /*
+      prevent indecision in the roll controller when target roll is
+      close to 180 degrees from the current roll
+     */
+    const float indecision_threshold_deg = 160;
+    if (fabsf(angle_err_deg) > indecision_threshold_deg) {
+        if (desired_rate * last_desired_rate < 0) {
+            desired_rate = -desired_rate;
```
IamPete1, line 212: I think this should recalculate the angle error. The problem is as the error passes 180 it will start getting smaller again while still trying to go the long way round. We can make it continue to get larger until it flips at 160.

I think something like this would work, but I have not tested. 
```
            desired_rate =  (360 - fabsf(angle_err_deg)) / gains.tau * is_positive(angle_err_deg) ? 1.0 : -1.0;
```

IamPete1 (commented):
I would like to see testing on a motor always on tailsitter like the swan H1 in realflight.

tridge:
need to test the HWing in RealFlight

tridge:
> I would like to see testing on a motor always on tailsitter like the swan H1 in realflight.

I have tested in RealFlight. Flies fine. Inverted mode doesn't work, but also doesn't work in master. I also found a bug where if you disarm in a fwd flight mode in an all motor tailsitter the motors don't stop! That bug is in master

ArduPlane/quadplane.cpp
tridge: fixed

ArduPlane/tailsitter.cpp
tridge: fixed

libraries/APM_Control/AP_RollController.cpp
tridge: fixed

libraries/APM_Control/AP_RollController.cpp
tridge: fixed with a different expression that I think is clearer

tridge:
@IamPete1 I have been testing recovery using the Griffin model in RealFlight. I'm getting some strange results. One is that when I switch to QLOITER the aircraft doesn't slow down. For example:
image
log here: http://uav.tridgell.net/GriffinTest/00000042.BIN
zooming in, we see the pitch staying below 2 degrees until it suddenly goes to 5 degrees, then some time later it does to 30 degrees, which is what I expected to happen to slow down from 50 m/s. 
Any idea what is limiting the pitch in QLOITER?

tridge:
@IamPete1 this is another one that worries me:
http://uav.tridgell.net/GriffinTest/00000039.BIN
it dropped like a stone inverted

IamPete1:
> Any idea what is limiting the pitch in QLOITER?

We do have the transition pitch limit stuff, it works in all modes but is only likely to be noticeable in Q_LOITER, the key function is `set_VTOL_roll_pitch_limit` there is a time parameter `Q_BACKTRANS_MS`. If were still using the FW controllers we need to delay the init of loiter navigation and delay the start of that pitch up limit. We should be able to use the same code paths as tailsitters. 

It does concern me for 4.6 that this is getting more and more complex.

IamPete1:
I had another thought, we should update `show_vtol_view` to account for the delay in switching controllers so it is correctly reflected in the logs and GCS.

tridge:
> We do have the transition pitch limit stuff, it works in all modes but is only likely to be noticeable in Q_LOITER, the key function is `set_VTOL_roll_pitch_limit` there is a time parameter `Q_BACKTRANS_MS`

I looked at that, but it was set to 3000 ms, yet, the pitch was limited to under 2 degrees for more than 25 seconds.



> I had another thought, we should update `show_vtol_view` to account for the delay in switching controllers so it is correctly reflected in the logs and GCS

yes, I have that change pending

tridge:
@IamPete1 the limiting of pitch is happening here:
image

ArduPlane/quadplane.cpp
```diff
@@ -1955,6 +1964,10 @@ void QuadPlane::motors_output(bool run_rate_controller)
             // relax if have been inactive
             relax_attitude_control();
         }
+
+        // see if we need to be in VTOL recovery
+        assist.check_VTOL_recovery();
```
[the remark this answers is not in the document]
tridge: yes, it is just a "set a flag" function

ArduPlane/Attitude.cpp
```diff
@@ -378,6 +378,10 @@ void Plane::stabilize_yaw()
         SRV_Channels::set_output_scaled(SRV_Channel::k_steering, steering_output);
     }
 
+#if HAL_QUADPLANE_ENABLED
+    // possibly recover from a spin
+    quadplane.assist.output_spin_recovery();
```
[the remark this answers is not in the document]
tridge: it is cleared on mode entry, plus the flag doesn't do anything unless output_spin_recovery() is called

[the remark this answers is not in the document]
tridge: hmm, I think you're right, it could in theory remain true in transition completion, I'll add a check for motors still being in UNLIMITED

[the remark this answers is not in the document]
tridge: i've also changed spin check to only kick it if pitch < -45 degrees. I suspect the recovery method won't work for a flat spin?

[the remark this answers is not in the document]
IamPete1: Right, the current detection won't trip in a flat spin because we would expect low X and Y rotation rates, only high about Z. 

The overrides here are no sufficient for flat spin recovery in any-case, we would also need to zero the forward throttle and ailerons. Hopefully TECS would push the nose down for us at least.

ArduPlane/VTOL_Assist.cpp
```diff
@@ -140,4 +140,94 @@ bool VTOL_Assist::should_assist(float aspeed, bool have_airspeed)
...
+        quadplane.force_fw_control_recovery = false;
+        quadplane.in_spin_recovery = false;
+        return false;
+    }
+
+    // see if the attitude is outside twice the Q_ANGLE_MAX
+    const auto &ahrs = plane.ahrs;
+    const int16_t angle_max_cd = quadplane.aparm.angle_max;
+    if ((abs(ahrs.roll_sensor) > 2*angle_max_cd ||
+         abs(ahrs.pitch_sensor) > 2*angle_max_cd)) {
```
IamPete1, line 162: May want to consider a circular limit. That is how Q angle max works. 

```(Vector2f{AB_diff[1], -AB_diff[0]}).length() > 2*angle_max_cd```

Will only make a difference in cases with high error in both roll and pitch of course.

File: ArduPlane/quadplane.cpp [hunk 13]
```diff
@@ -238,6 +238,14 @@ const AP_Param::GroupInfo QuadPlane::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("ASSIST_ANGLE", 45, QuadPlane, assist.angle, 30),
 
+    // @Param: ASSIST_OPTIONS
+    // @DisplayName: Quadplane assistance options
+    // @Description: Options for special QAssist features
+    // @Bitmask: 0: Disable force fixed wing controller recovery
+    // @Bitmask: 1: Disable quadplane spin recovery
+    // @User: Standard
+    AP_GROUPINFO("ASSIST_OPTIONS", 47, QuadPlane, assist.options, 0),
```
IamPete1, line 247: 47 was tilt type per the comment on 249. We are still doing the conversion so we probably cant re-use it yet.

tridge: we can re-use as it is a different type. In AP_Param::scan() we do a type check so reusing is fine as long as the type is different. The type was AP_int8 for tilt type, this is AP_Int16

IamPete1: Ah, OK, it would be worth updating the comment on 249.

tridge: added

ArduPlane/VTOL_Assist.cpp
tridge: changed to circular limit, thanks

rmackay9:
this is included in 4.6.0-beta5 via PR https://github.com/rmackay9/rmackay9-ardupilot/pull/37

PR #29505, opened 2025-03-11
AP_InertialSensor: Remove the variable _accel_max_abs_offsets its is …

JulioCesarMatias:
…not more used


peterbarker (approved):
LGTM.

Good find, merging this :-)

rmackay9 (changes requested):
Hi @JulioCesarMatias,

Thanks very much for this.  It seems that there is a merge commit in this PR.  Could you rebase on master to get rid of this?

PR #29508, opened 2025-03-12
AP_SerialManager: support baud rate of "12500"

rmackay9:
This PR adds support for SERIALx_BAUD = 12500 (which is mapped to 12,500,000).  This is important with AP's improved Ethernet support because it is the recommended baud rate to use with PPP (aka serial<->ethernet).  [See the DroneNet setup page](https://ardupilot.org/copter/docs/common-botblox-dronenet.html#autopilot-configuration) for an example of this.

I've tested this change on real hardware to confirm that if SERIALx_BAUD = 12500 the autopilot could connect over PPP

### DIFF
--- a/libraries/AP_SerialManager/AP_SerialManager.cpp
+++ b/libraries/AP_SerialManager/AP_SerialManager.cpp
@@ -166,7 +166,7 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
     // @Param: 0_BAUD
     // @DisplayName: Serial0 baud rate
     // @Description: The baud rate used on the USB console. Most stm32-based boards can support rates of up to 1500. If you setup a rate you cannot support and then can't connect to your board you should load a firmware from a different vehicle type. That will reset all your parameters to defaults.
-    // @Values: 1:1200,2:2400,4:4800,9:9600,19:19200,38:38400,57:57600,111:111100,115:115200,230:230400,256:256000,460:460800,500:500000,921:921600,1500:1500000,2000:2000000
+    // @Values: 1:1200,2:2400,4:4800,9:9600,19:19200,38:38400,57:57600,111:111100,115:115200,230:230400,256:256000,460:460800,500:500000,921:921600,1500:1500000,2000:2000000,12.5MBaud:12500000
     // @User: Standard
     AP_GROUPINFO("0_BAUD",  0, AP_SerialManager, state[0].baud, DEFAULT_SERIAL0_BAUD/1000),
 
@@ -192,7 +192,7 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
     // @Param: 1_BAUD
     // @DisplayName: Telem1 Baud Rate
     // @Description: The baud rate used on the Telem1 port. Most stm32-based boards can support rates of up to 1500. If you setup a rate you cannot support and then can't connect to your board you should load a firmware from a different vehicle type. That will reset all your parameters to defaults.
-    // @Values: 1:1200,2:2400,4:4800,9:9600,19:19200,38:38400,57:57600,111:111100,115:115200,230:230400,256:256000,460:460800,500:500000,921:921600,1500:1500000,2000:2000000
+    // @Values: 1:1200,2:2400,4:4800,9:9600,19:19200,38:38400,57:57600,111:111100,115:115200,230:230400,256:256000,460:460800,500:500000,921:921600,1500:1500000,2000:2000000,12.5MBaud:12500000
     // @User: Standard
     AP_GROUPINFO("1_BAUD", 2, AP_SerialManager, state[1].baud, DEFAULT_SERIAL1_BAUD),
 #endif


Hwurzburg (dismissed):
should also get rid of default protocols in metadata and wrong value for allowable baud rates

timtuxworth:
Thanks for doing this @rmackay9  - this is just what I need!

rmackay9:
Hi @Hwurzburg,

Thanks for the feedback.  Sorry, I don't know what, "default protocols in metadata and wrong value for allowable baud rates" refers to.  Could you give some more background?

rmackay9:
I was missing a "0" which is fixed now.  I've also confirmed that this works by testing a connection between my PC and the autopilot on a local ethernet network

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -181,7 +181,7 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
 
 #if HAL_HAVE_SERIAL1
     // @Param: 1_PROTOCOL
-    // @DisplayName: Telem1 protocol selection
+    // @DisplayName: Serial1 protocol selection
```
muramura, line 184: PIXHAWK and CUBE systems have TELEM1 and TELEM2 labeled on their cases.
I think it would be better to match the display accordingly.

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -190,38 +190,38 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
...
+    // @Values: 1:1200,2:2400,4:4800,9:9600,19:19200,38:38400,57:57600,111:111100,115:115200,230:230400,256:256000,460:460800,500:500000,921:921600,1500:1500000,2000:2000000,12500:12500000
     // @User: Standard
     AP_GROUPINFO("1_BAUD", 2, AP_SerialManager, state[1].baud, DEFAULT_SERIAL1_BAUD),
 #endif
 
 #if HAL_HAVE_SERIAL2
     // @Param: 2_PROTOCOL
     // @CopyFieldsFrom: SERIAL1_PROTOCOL
-    // @DisplayName: Telemetry 2 protocol selection
+    // @DisplayName: Serial2 protocol selection
```
muramura, line 203: PIXHAWK and CUBE systems have TELEM1 and TELEM2 labeled on their cases.
I think it would be better to match the display accordingly.

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -190,38 +190,38 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
     AP_GROUPINFO("1_PROTOCOL",  1, AP_SerialManager, state[1].protocol, DEFAULT_SERIAL1_PROTOCOL),
 
     // @Param: 1_BAUD
-    // @DisplayName: Telem1 Baud Rate
+    // @DisplayName: Serial1 Baud Rate
```
muramura, line 193: PIXHAWK and CUBE systems have TELEM1 and TELEM2 labeled on their cases.
I think it would be better to match the display accordingly.

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -190,38 +190,38 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
...
     // @CopyFieldsFrom: SERIAL1_PROTOCOL
-    // @DisplayName: Telemetry 2 protocol selection
+    // @DisplayName: Serial2 protocol selection
     // @Description: Control what protocol to use on the Telem2 port. Note that the Frsky options require external converter hardware. See the wiki for details.
     AP_GROUPINFO("2_PROTOCOL",  3, AP_SerialManager, state[2].protocol, DEFAULT_SERIAL2_PROTOCOL),
 
     // @Param: 2_BAUD
     // @CopyFieldsFrom: SERIAL1_BAUD
-    // @DisplayName: Telemetry 2 Baud Rate
+    // @DisplayName: Serial2 Baud Rate
```
muramura, line 209: PIXHAWK and CUBE systems have TELEM1 and TELEM2 labeled on their cases.
I think it would be better to match the display accordingly.

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -190,38 +190,38 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
...
     // @Description: The baud rate of the Telem2 port. Most stm32-based boards can support rates of up to 1500. If you setup a rate you cannot support and then can't connect to your board you should load a firmware from a different vehicle type. That will reset all your parameters to defaults.
     AP_GROUPINFO("2_BAUD", 4, AP_SerialManager, state[2].baud, DEFAULT_SERIAL2_BAUD),
 #endif
 
 #if HAL_HAVE_SERIAL3
     // @Param: 3_PROTOCOL
     // @CopyFieldsFrom: SERIAL1_PROTOCOL
-    // @DisplayName: Serial 3 (GPS) protocol selection
-    // @Description: Control what protocol Serial 3 (GPS) should be used for. Note that the Frsky options require external converter hardware. See the wiki for details.
+    // @DisplayName: Serial3 (GPS) protocol selection
```
muramura, line 217: PIXHAWK and CUBE systems have GPS labeled on their cases.
I think it would be better to match the display accordingly.

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -190,38 +190,38 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
...
+    // @Description: Control what protocol Serial3 (GPS) should be used for. Note that the Frsky options require external converter hardware. See the wiki for details.
     AP_GROUPINFO("3_PROTOCOL",  5, AP_SerialManager, state[3].protocol, DEFAULT_SERIAL3_PROTOCOL),
 
     // @Param: 3_BAUD
     // @CopyFieldsFrom: SERIAL1_BAUD
-    // @DisplayName: Serial 3 (GPS) Baud Rate
-    // @Description: The baud rate used for the Serial 3 (GPS). Most stm32-based boards can support rates of up to 1500. If you setup a rate you cannot support and then can't connect to your board you should load a firmware from a different vehicle type. That will reset all your parameters to defaults.
+    // @DisplayName: Serial3 (GPS) Baud Rate
```
muramura, line 223: PIXHAWK and CUBE systems have GPS labeled on their cases.
I think it would be better to match the display accordingly.

libraries/AP_SerialManager/AP_SerialManager.cpp
Hwurzburg: Many , many other autopilots have a different default protocol....

rmackay9: To avoid discussion and thus speed this change into master I've removed the changes to the parameter "DisplayName" and just left in the core change to add support for the 12500 baud rate which is really the key change that users need.

libraries/AP_SerialManager/AP_SerialManager.cpp
rmackay9: closing because change removed

libraries/AP_SerialManager/AP_SerialManager.cpp
rmackay9: marking as resolved because change removed

libraries/AP_SerialManager/AP_SerialManager.cpp
rmackay9: marking as resolved because change removed

libraries/AP_SerialManager/AP_SerialManager.cpp
rmackay9: marking as resolved because change removed

libraries/AP_SerialManager/AP_SerialManager.cpp
rmackay9: marking as resolved because change removed

timtuxworth:
Do you think we can get this back ported to 4.6 @rmackay9 ?

tpwrules:
This would be zero cost if you updated the metadata so the value is selected by itself. Would also avoid any compatibility concerns in the unlikely case somebody actually needs 12500 baud.

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -166,7 +166,7 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
     // @Param: 0_BAUD
     // @DisplayName: Serial0 baud rate
     // @Description: The baud rate used on the USB console. Most stm32-based boards can support rates of up to 1500. If you setup a rate you cannot support and then can't connect to your board you should load a firmware from a different vehicle type. That will reset all your parameters to defaults.
-    // @Values: 1:1200,2:2400,4:4800,9:9600,19:19200,38:38400,57:57600,111:111100,115:115200,230:230400,256:256000,460:460800,500:500000,921:921600,1500:1500000,2000:2000000
+    // @Values: 1:1200,2:2400,4:4800,9:9600,19:19200,38:38400,57:57600,111:111100,115:115200,230:230400,256:256000,460:460800,500:500000,921:921600,1500:1500000,2000:2000000,12500:12500000
```
tridge, line 169: I'd prefer 12.5MBaud as the name

tridge (changes requested):

libraries/AP_SerialManager/AP_SerialManager.cpp
rmackay9: done, thanks for the review

rmackay9:

PR #29509, opened 2025-03-12
SCurve: Add Z component to speed and accel constraints before starting next SCurve

MattKear:
I do not think we were correctly accounting for the z-component in the turning point comparison with speed and accel constraints.  This leads to the behaviour whereby we can temporarily exceed the speed up or down limits in WP nav when transiting through a fast waypoints.

**Behaviour Before Proposed Fix:**

Using a mission with waypoints that only vary in height.  My example here:
[s_curves_test_mission_pure_z.txt](https://github.com/user-attachments/files/19214752/s_curves_test_mission_pure_z.txt)

In this example WPNAV_SPEED_UP = 250 cm/s and WPNAV_SPEED_DN = 150 cm/s.

On the ascent:
image
Desired climb rate temporarily exceeds WPNAV_SPEED_UP as the vehicle transits through the waypoint.

On the descent:
image
Desired climb rate temporarily exceeds WPNAV_SPEED_DN as the vehicle transits through the waypoint.

Comparing this to the behaviour in XY, I used a mission with waypoints only varying in North:
[s_curves_test_mission_pure_x.txt](https://github.com/user-attachments/files/19214932/s_curves_test_mission_pure_x.txt)

This behaviour cannot be seen when transiting through WP (WPNAV_SPEED = 1000 cm/s): 
image

Having added some hacky logging to a working branch you can see the contributions of the SCurves which highlight the issue when doing the purely vertical mission:

image

**Behaviour With Proposed Fix:**

Vertical only mission:

I have checked and XY behaviour remains unchanged.
image

Only testing SITL thus far.

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -12416,12 +12416,12 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
         self.set_rc(1, 1200)
         self.delay_sim_time(1)  # build up some pilot desired stuff
         self.change_mode('AUTO')
-        self.wait_waypoint(2, 2)
+        self.wait_waypoint(2, 2, max_dist=3)
         self.set_parameters({
             'SIM_RC_FAIL': 1,
         })
 #        self.set_rc(1, 1500)  # note we are still in RC fail!
-        self.wait_waypoint(3, 3)
+        self.wait_waypoint(3, 3, max_dist=3)
         self.assert_mode_is('AUTO')
         self.change_mode('LOITER')
         self.wait_groundspeed(0, 0.1, minimum_duration=30, timeout=450)
--- a/libraries/AP_Math/SCurve.cpp
+++ b/libraries/AP_Math/SCurve.cpp
@@ -505,8 +505,8 @@ bool SCurve::advance_target_along_track(SCurve &prev_leg, SCurve &next_leg, floa
         next_leg.move_from_time_pos_vel_accel(time_to_destination * 0.5f, turn_pos, turn_vel, turn_accel);
         const float speed_min = MIN(get_speed_along_track(), next_leg.get_speed_along_track());
         if ((get_time_remaining() < next_leg.time_end() * 0.5f) && (turn_pos.length() < wp_radius) &&
-             (Vector2f{turn_vel.x, turn_vel.y}.length() < speed_min) &&
-             (Vector2f{turn_accel.x, turn_accel.y}.length() < accel_corner)) {
+             (turn_vel.length() < speed_min) &&
+             (turn_accel.length() < accel_corner)) {
             next_leg.move_from_pos_vel_accel(dt, target_pos, target_vel, target_accel);
         }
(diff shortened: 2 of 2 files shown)


IamPete1:
This is failing `AutoContinueOnRCFailsafe`  copter failsafe test. Not because the functionality it tests is broken but because there is a slight difference in the path that results in the test missing passing waypoint 3. I think this is just due to the fact it is using the mavlink nav controller output message at a low frequency. 

To tell any difference I had to plot the two logs together. 
Positions:
image

image

image

Velocity's:

image

image

image

Accels:
image

image

image

So there is a slight difference in the path. This change cannot change the SCurves them-selfs, it just changes the point at which we move onto the next SCurve

MattKear:
Thanks @IamPete1 for looking into that.  There being a small difference in the desired does make sense, because if there is a z-component to the mission then the point at which the fast way point gets accepted will change slightly.  Hence the next leg s-curve will be added in at a slightly different position/time along the mission.

On that basis, I should think it would be fine to modify this test so that it will pass with this fix to s-curves.  What do you think?

IamPete1:
> if there is a z-component to the mission

The problem is that there is not in this case. All the same altitude, pure north waypoints.

https://github.com/ArduPilot/ardupilot/blob/bfb93411727f43affd6cb517059b5590081acddb/Tools/autotest/arducopter.py#L12341-L12347

The Z portion of the mission matches well, just a difference in the north that is only clear in velocity and accel. 

I'm fairly sure the change is still correct, just it would be nice to understand what exactly is going on.

IamPete1:
> The problem is that there is not in this case. All the same altitude, pure north waypoints.

Although clearly my plots do show some change in altitude.

IamPete1:
> The problem is that there is not in this case. All the same altitude, pure north waypoints.

> Although clearly my plots do show some change in altitude.

To answer my own question here. The vehicle takes off to 30 meters before entering auto. The first auto way-point is a takeoff which is skipped because the vehicle is already above the set 10m takeoff altitude. That puts the vehicle onto waypoint 10. That is a 20 north and 20 meters down from the 30m takeoff altitude. The test is then looking for the vehicle to reach way-point 3. 

The key thing here is that first leg 20 meters down and 20 meters across. This means its constrained by the vertical speed so this patch makes a larger difference. Once waypoint 2 is reached waypoint 3 is another 20m north at the same altitude so the vehicle is no longer constrained by the vertical speed. Its that changeover between waypoint 2 and 3 that causes the change in behavior here.

IamPete1:
I have added a patch to increase the waypoint tolerance on the failing test.

rmackay9:
We agreed that after @lthall reviews we can merge this

lthall (commented):
I love the work but it doesn't quiet address the problems.
Old Problem:
image

This PR:
image

I have added a PR here: #29601 
image

MattKear:
Closing in favour of https://github.com/ArduPilot/ardupilot/pull/29601

PR #29510, opened 2025-03-12
Sub: allow stabilized modes to reach more aggressive attitudes

Williangalvani:
This increases ROLL_PITCH_INPUT_MAX (which is used elsewhere to constrain angle_max)  and changes the joystick code to consume the now-relaxed angle_max parameter instead of a hardcoded `200`.

### DIFF
--- a/ArduSub/joystick.cpp
+++ b/ArduSub/joystick.cpp
@@ -150,6 +150,7 @@ void Sub::transform_manual_control_to_rc_override(int16_t x, int16_t y, int16_t 
 
 void Sub::handle_jsbutton_press(uint8_t _button, bool shift, bool held)
 {
+    float angle_max = aparm.angle_max/10;  // convert from deg*100 to deg*10
     // Act based on the function assigned to this button
     switch (get_button(_button)->function(shift)) {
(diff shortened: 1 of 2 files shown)


ArduSub/joystick.cpp
```diff
@@ -150,6 +150,7 @@ void Sub::transform_manual_control_to_rc_override(int16_t x, int16_t y, int16_t
 
 void Sub::handle_jsbutton_press(uint8_t _button, bool shift, bool held)
 {
+    float angle_max = aparm.angle_max/10;  // convert from deg*100 to deg*10
```
peterbarker, line 153: Is this *really* in decidegrees?!  That's not a unit we've really used in ArduPilot.

I mean, I can see that it's multiplying a centdegrees by 10, which would give decidegrees, but isn't this really "your maximum trim angle is 10% of your maximum angle *in degrees*

peterbarker:
```suggestion
    const float trim_angle_max = aparm.angle_max * 0.1;  // trim angle is 10% of the maximum angle (degrees)
```

Williangalvani:
I'll close this as I'm planning do rewrite a significant portion of this code

PR #29512, opened 2025-03-12
HWDEF: Fix IMU rotation for Stellar H7V2

marchuks:
We need to change the default IMU rotation due to minor differences between the production version and the development sample.


MattKear:
Are there a significant number of the dev samples out in the wild?  Will changing this default catch any users out?

marchuks:
> Are there a significant number of the dev samples out in the wild? Will changing this default catch any users out?

Nothing in the wild. Only a few items in our developers.

MattKear (approved):
LGTM

andyp1per:
@marchuks for next time I think "hwdef" in the commit message should be lower case, but not too bothered.

PR #29513, opened 2025-03-13
autotest: change uses of recv_match to assert_receive_message

peterbarker:
we prefer assert_receive_message as it's less verbose than using recv_match all the time, and more consistent in that if you don't receive a message you will get a good error message, as opposed to a lot of places in here which would give you an attribute-on-None exception in case a timeout occurs.  Also means that you *always* get a timeout so won't hang the test in case the message doesn'tcome

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -155,7 +155,7 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
 
     def wait_landed_and_disarmed(self, min_alt=6, timeout=60):
         """Wait to be landed and disarmed"""
-        m = self.mav.recv_match(type='GLOBAL_POSITION_INT', blocking=True)
+        m = self.assert_receive_message('GLOBAL_POSITION_INT')
         alt = m.relative_alt / 1000.0 # mm -> m
         if alt > min_alt:
             self.wait_altitude(min_alt-1, min_alt+5, relative=True, timeout=timeout)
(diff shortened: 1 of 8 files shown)


Tools/autotest/arducopter.py
```diff
@@ -3848,39 +3840,31 @@ def RangeFinder(self):
...
         if not flags & mavutil.mavlink.MAV_SYS_STATUS_SENSOR_LASER_POSITION:
             raise NotAchievedException("Laser not enabled in SYS_STATUS")
 
         self.takeoff(10, mode="LOITER")
 
-        m_r = self.mav.recv_match(type='RANGEFINDER',
-                                  blocking=True)
-        m_p = self.mav.recv_match(type='GLOBAL_POSITION_INT',
-                                  blocking=True)
+        m_r = self.assert_receive_message('RANGEFINDER')
```
MattKear, line 3866: I'm not sure how much it matters but the default timeout on recv_match was None and by moving to assert_receive_message the default is 1.  That is probably for the best but may lead to a few new flapping tests if 1 s is a bit of a tight tolerance.  Just an observation though.

MattKear (approved):
LGTM, Just need to get it to pass the failing test, which looks unrelated to this PR.

Tools/autotest/arducopter.py
peterbarker: > I'm not sure how much it matters but the default timeout on recv_match was None and by moving to assert_receive_message the default is 1. That is probably for the best but may lead to a few new flapping tests if 1 s is a bit of a tight tolerance. Just an observation though.

My reading was that if it comes in as `None` then it will get defaulted to `1` explicitly further down in teh call train.  One of us is wrong :-)

peterbarker:
> LGTM, Just need to get it to pass the failing test, which looks unrelated to this PR.

I have *zero* idea why this PR tickles what appears to be a real, significant and long-standing bug in log transfer: https://github.com/ArduPilot/ardupilot/pull/29532

peterbarker:
@MattKear I did work out what the regression here was.  recv_match() times out in wall-clock time, `assert_receive_message` in simulated time!

So when we asserted we weren't receiving a message with recv_match we were doing so for maybe 60 simulated seconds.  With  assert_receive_match 2 simulated seconds.  After the change we were hitting a protection feature which doesn't let you start new logs too rapidly in Arduilot.

I've reverted that bit in this PR now.

PR #29518, opened 2025-03-13
AP_Motors: fixed motor knocking issue on tilt quadplanes

tridge:
the motor output was happening twice per loop with different values for tilt quadplanes using motor matrix

not tested yet

### DIFF
--- a/libraries/AP_Motors/AP_MotorsMatrix.cpp
+++ b/libraries/AP_Motors/AP_MotorsMatrix.cpp
@@ -176,7 +176,7 @@ void AP_MotorsMatrix::output_to_motors()
 
     // convert output to PWM and send to each motor
     for (i = 0; i < AP_MOTORS_MAX_NUM_MOTORS; i++) {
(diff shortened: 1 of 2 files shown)


IamPete1:
We do output PWM twice, but its after a cork and before a push, so I don't understand how its getting out to the motors.

IamPete1:
I have added the same fix for the tri-copter backend.

IamPete1:
Apparently this does fix, but I still don't understand the cork push thing.

IamPete1:
To be clear, I don't think we should merge this without understanding it. I also think there may be is a issue with how `motor_enabled_mask` is set, because we do the normal output first then the throttle override it must be 0 on the first call. We should test the tailsitter case too which is the other user of the `motor_enabled_mask` stuff.

tridge:
@IamPete1 I think this is a better fix:
https://github.com/ArduPilot/ardupilot/pull/29521

tridge:
went with cork/push fix

PR #29519, opened 2025-03-13
CAN: mailboxes never timeout

hshakula:
CANIface::discardTimedOutTxMailboxes does not drop mailboxes unless they have been `setup`. CANFDIface handles that correctly.

### DIFF
--- a/libraries/AP_HAL_ChibiOS/CanIface.cpp
+++ b/libraries/AP_HAL_ChibiOS/CanIface.cpp
@@ -371,6 +371,8 @@ int16_t CANIface::send(const AP_HAL::CANFrame& frame, uint64_t tx_deadline,
         txi.loopback       = (flags & Loopback) != 0;
         txi.abort_on_error = (flags & AbortOnError) != 0;
         // setup frame initial state
+        txi.aborted        = false;
+        txi.setup          = true;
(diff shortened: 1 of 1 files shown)


tpwrules:
How did you discover this issue? This logic is pretty questionable anyway, nobody ever clears the `setup` state. Are you actually seeing timeouts and jamming up the mailboxes?

I think in both drivers we could condense those three state variables `aborted`, `setup`, and `pushed` down to two or even one. Do you want to take on that work?

tpwrules:
Good to know.

I think the effort testing this change would be better served cleaning up the logic and fixing the related issues.

libraries/AP_HAL_ChibiOS/CanIface.cpp
```diff
@@ -371,6 +371,8 @@ int16_t CANIface::send(const AP_HAL::CANFrame& frame, uint64_t tx_deadline,
         txi.loopback       = (flags & Loopback) != 0;
         txi.abort_on_error = (flags & AbortOnError) != 0;
         // setup frame initial state
+        txi.aborted        = false;
+        txi.setup          = true;
```
tridge, line 375: probably should also reset setup flag in CANIface::handleTxMailboxInterrupt()

PR #29521, opened 2025-03-13
HAL_ChibiOS: fixed cork/push

tridge:
ensure that period[] array is not updated until we push
this fixes the stuttering motors on tilt quadplanes, and may fix other issues
this is an alternative to #29518

### DIFF
--- a/libraries/AP_HAL_ChibiOS/RCOutput.cpp
+++ b/libraries/AP_HAL_ChibiOS/RCOutput.cpp
@@ -737,7 +737,13 @@ void RCOutput::write(uint8_t chan, uint16_t period_us)
 
     chan -= chan_offset;
 
-    period[chan] = period_us;
+    if (corked) {
+        // when corked we put the updated period in a separate array which is
+        // copied to period[] when we push
+        period_corked[chan] = period_us;
+    } else {
+        period[chan] = period_us;
+    }
 
     if (chan < num_fmu_channels) {
         active_fmu_channels = MAX(chan+1, active_fmu_channels);
@@ -1344,6 +1350,7 @@ void RCOutput::push(void)
         INTERNAL_ERROR(AP_InternalError::error_t::flow_of_control);
     }
     corked = false;
+    memcpy(period, period_corked, sizeof(period));
     push_local();
 #if HAL_WITH_IO_MCU
     if (iomcu_enabled) {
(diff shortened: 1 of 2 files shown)


libraries/AP_HAL_ChibiOS/RCOutput.cpp
```diff
@@ -737,7 +737,13 @@ void RCOutput::write(uint8_t chan, uint16_t period_us)
 
     chan -= chan_offset;
 
-    period[chan] = period_us;
+    if (corked) {
+        // when corked we put the updated period in a separate array which is
+        // copied to period[] when we push
+        period_corked[chan] = period_us;
```
andyp1per, line 743: I mean shouldn't this always be the case? Couldn't you just assert on the non-corked case?

tridge: there could be special cases where we write non-corked, for something going into 4.6.0 I wouldn't want to rule out non-corked output one some weird vehicle setup
one example would be ReplayGyroFFT
another in toy_mode test code
for 4.7.0 I'd be happy to enforce corked only, starting with SITL

andyp1per (approved):

tpwrules:
So the problem is that the period[] array was read by the dshot_send function and values were sent out before the push?

andyp1per:
> So the problem is that the period[] array was read by the dshot_send function and values were sent out before the push? Does the memcpy create any tearing problems?
> 
> Should read see the pre-push or post-push value? It wasn't involved in the issue was it?

Once corked, no updates should be seen until push has been called. so read should see the post-pushed or pre-corked values

IamPete1:
We could also remove the output here: https://github.com/ArduPilot/ardupilot/blob/83b2a255c5be196b3cc2a2b80ce1d8379cf93cc5/libraries/SRV_Channel/SRV_Channel_aux.cpp#L351

So the value is not written to the hal.

andyp1per:
This incurs about a 0.1% CPU cost @2k and 0.2% @4k

andyp1per:
Briefly tested on my fast rate quad without issue

peterbarker (approved):

IamPete1 (approved):
LGTM.

rmackay9:

PR #29524, opened 2025-03-14
SITL: Add gausian noise and rate limit to VICON sim

rishabsingh3003:
Helps with testing non-gps stuff in SITL


libraries/SITL/SITL.cpp
```diff
@@ -724,6 +724,27 @@ const AP_Param::GroupInfo SIM::var_info3[] = {
     AP_SUBGROUPINFO(volz_sim, "VOLZ_", 55, SIM, Volz),
 #endif
 
+    // @Param: VICON_P_SD
```
tridge, line 727: should be a subgroup, ideally move the params into SIM_Vicon.cpp

tridge (changes requested):

tridge (approved):

Issue #29515, opened 2025-03-13
LittleFS: Unknown SPI Flash 0x00000000


be-ska:
## Bug report

**Issue details**

Using a custom hardware with W25N01GV flash chip, the LittleFS filesystem _sometimes_ can't be mounted, MAVProxy report:

```
Unknown SPI Flash 0x00000000
FlashMemory_LittleFS: dead
SDCard failed to start
```
The same hardware works flawlessly with the previous batch logging functionality.

I could fix the issue editing `mount_filesystem()`, calling `init_flash()` **before** `find_block_size_and_count()`.

https://github.com/ArduPilot/ardupilot/blob/d4389d1279d72287d94be0515f23ecac7b268959/libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp#L879-L891

It actually seems reasonable to reset and init the flash chip before requesting the JEDEC ID, I could open a PR with this simple modification if needed.

**Version**
4.7.0-dev

**Platform**
[x] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Hardware type**
Custom hardware with H753 and W25N01GV SPI flash chip

tpwrules:
This seems reasonable, please send a PR.

peterbarker:
OK, so this is an acknowledged problem, but the first PR didn't work out.  @be-ska do you have another idea yet?

be-ska:
This has been a weird issue to debug. Out of the five flight controllers I tested with LittleFS, only one exhibited this problem, and it was a recurring issue. Using the patch in the first pull request the issue seemed to be resolved, but after some flights (I’m not sure exactly how many, since I can’t even read the logs), the problem reappeared.
I also tried calling init_flash() in a loop for 10 seconds or until it returned true, and added some delays. This seemed to help at first, but after a while, the issue came back. I stopped testing at this point because I need to focus on other projects.
It’s worth noting that with the old but gold batch logging method I never experienced any failures. My idea was to compare the two implementations to identify any differences, but unfortunately I don’t have the time to do that at the moment.

PR #29525, opened 2025-03-14
AP_Filesystem: LittleFS: call init_flash() before reading JEDEC ID

be-ska:
Fix issue #29515

### DIFF
--- a/libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
+++ b/libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
@@ -870,6 +870,11 @@ bool AP_Filesystem_FlashMemory_LittleFS::mount_filesystem() {
     }
 
     dev_sem = dev->get_semaphore();
+
+    if (!init_flash()) {
+        mark_dead();
+        return false;
+    }
 #else
     fs_cfg.read = lfs_filebd_read;
     fs_cfg.prog = lfs_filebd_prog;
@@ -883,12 +888,7 @@ bool AP_Filesystem_FlashMemory_LittleFS::mount_filesystem() {
         mark_dead();
         return false;
     }
-#if CONFIG_HAL_BOARD != HAL_BOARD_SITL
-    if (!init_flash()) {
-        mark_dead();
-        return false;
-    }
-#endif
+
     if (lfs_mount(&fs, &fs_cfg) < 0) {
         /* maybe not formatted? try formatting it */
         GCS_SEND_TEXT(MAV_SEVERITY_NOTICE, "Formatting flash");


libraries/AP_Filesystem/AP_Filesystem_FlashMemory_LittleFS.cpp
```diff
@@ -870,6 +870,11 @@ bool AP_Filesystem_FlashMemory_LittleFS::mount_filesystem() {
     }
 
     dev_sem = dev->get_semaphore();
+
```
tpwrules, line 873:
```suggestion
// some devices/situations read a bogus ID and apparently
// need a reset beforehand, though they should be able
// to accept an ID read if they will accept a reset...
```

tpwrules (commented):
This could create a chicken and egg problem if we ever decide to determine the flash driver type (W25Nxx vs JEDEC NOR) at runtime but for now it's solving a real problem.

I wonder if the chip is busy with a command and your board starts up very quickly? Or you are somehow violating the notes in the datasheet about powering up with /CS high.

Do you have the capability to get logic analyzer traces of the failure? It's unexpected that the chip should accept a reset command but not an ID command, so maybe the reset is ignored and instead clears out whatever bad state there is. Can you test if simply reading the ID twice fixes it (e.g. duplicating line 730)? 

Tested that my KakuteH7Mini-Nand and KakuteH7Mini still mount LittleFS consistently so this PR isn't wrong per se, but I'm less sure about merging it.

be-ska:
Ok I think I've found the issue: KakuteH7Mini defined the spi dataflash as
```
SPIDEV dataflash SPI1 DEVID1 FLASH_CS   MODE3 32*MHZ 32*MHZ
```
but I've defined my hardware like Mamba H743v4, Matek H7A3 and [other boards](https://github.com/search?q=repo:ArduPilot/ardupilot+104*MHZ&type=code)
```
SPIDEV dataflash SPI2 DEVID1 FLASH_CS   MODE3 104*MHZ  104*MHZ
```

Changing SPI clock frequency from 104 MHz to 32 MHz actually fix the issue, without requiring additional modification to LittleFS code.

andyp1per:
> Ok I think I've found the issue: KakuteH7Mini defined the spi dataflash as
> 
> ```
> SPIDEV dataflash SPI1 DEVID1 FLASH_CS   MODE3 32*MHZ 32*MHZ
> ```
> 
> but I've defined my hardware like Mamba H743v4, Matek H7A3 and [other boards](https://github.com/search?q=repo:ArduPilot/ardupilot+%22104*MHZ%22+AND+%22dataflash%22&type=code)
> 
> ```
> SPIDEV dataflash SPI2 DEVID1 FLASH_CS   MODE3 104*MHZ  104*MHZ
> ```
> 
> Changing SPI clock frequency from 104 MHz to 32 MHz actually fix the issue, without requiring additional modification to LittleFS code.

Did you check the datasheet? These devices have quite strict startup conditions whereby they are not ready for some period after power is applied. I suspect it could just be that you are reading too soon.

be-ska:
> Did you check the datasheet? These devices have quite strict startup conditions whereby they are not ready for some period after power is applied. I suspect it could just be that you are reading too soon.

Yes, I've checked the datasheet, the CS and VCC pins are high for more than 6 seconds before the chip reading attempt.

Screenshot 2025-03-17 at 16 53 06

I also can't debug with a logic analyzer, because switching to 32 MHz SPI clock actually fixed the issue and my logic analyzer sampling rate is too low to catch 104 MHz signals.

andyp1per:
It may simply be that the traces are too long

be-ska:
> It may simply be that the traces are too long

I don't think that's the issue: traces are short and using the old batch log functionality it always works, only on master with LittleFS I can reproduce the issue. By the way the solution proposed on this PR clearly is not the right one, so I close it, thank you for your help debugging the issue @tpwrules @andyp1per

PR #29526, opened 2025-03-14
AC_Fence: do not store radius as integers if they look like integers

peterbarker:
... and a plan to get rid of this compatability code

This is why we have these two types (copied from the code):
```
// CIRCLE_INCLUSION_INT stores the radius an a 32-bit integer in
// metres.  This was a bug, and CIRCLE_INCLUSION was created to store
// as a 32-bit float instead.  We save as _INT in the case that the
// radius looks like an integer as a backwards-compatibility measure.
// For 4.2 we might consider only loading _INT and always saving as
// float, and in 4.3 considering _INT invalid
```

I don't think this is a good candidate for build_options.py - we should just auto-fix the thing in storage for a couple of versions.


peterbarker:
I've made sure going from master to this PR brings the fence in correct.

I think did a "fence save /tmp/some.fence" and loaded from that and got the same fence back.

andyp1per (approved):
LGTM

rmackay9 (approved):

PR #29528, opened 2025-03-14
Tools:add build option for baro thrust correction

Hwurzburg:
Don't see the point that only one board has this, and any user would need to build locally a new board def and firmware to use it...if its worth having in the code, we should let users access it

### DIFF
--- a/Tools/autotest/test_build_options.py
+++ b/Tools/autotest/test_build_options.py
@@ -268,6 +268,7 @@ class TestBuildOptions(object):
             'AP_PLANE_BLACKBOX_LOGGING',  # entirely elided if no user
             'AP_COMPASS_AK8963_ENABLED',  # probed on a board-by-board basis, not on CubeOrange for example
             'AP_COMPASS_LSM303D_ENABLED',  # probed on a board-by-board basis, not on CubeOrange for example
+            'AP_BARO_THST_COMP_ENABLED',  # compiler is optimising this symbol away
         ])
         if target.lower() != "copter":
             feature_define_whitelist.add('MODE_ZIGZAG_ENABLED')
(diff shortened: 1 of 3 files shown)


andyp1per (approved):
Thanks @Hwurzburg !

peterbarker (changes requested):
The nominated symbol is not present unless you are building for Plane or Sub.

Need to add this to this whitelist so we don't care if the symbol isn't present on Rover/Tracker/Sub etc: https://github.com/ardupilot/ardupilot/blob/master/Tools/autotest/test_build_options.py#L311


```
Build directory: /home/pbarker/rc/ardupilot/build/CubeOrange
Target     Text (B)  Data (B)  BSS (B)  Total Flash Used (B)  Free Flash (B)  Ex
ternal Flash Used (B)
--------------------------------------------------------------------------------
---------------------
bin/blimp   1391320      2284   259992               1393604          572464  No
t Applicable

'blimp' finished successfully (8.586s)
Traceback (most recent call last):
  File "/home/pbarker/rc/ardupilot/./Tools/autotest/test_build_options.py", line
 716, in <module>
    tbo.run()
  File "/home/pbarker/rc/ardupilot/./Tools/autotest/test_build_options.py", line 655, in run
    self.run_enable_in_turn()
  File "/home/pbarker/rc/ardupilot/./Tools/autotest/test_build_options.py", line 571, in run_enable_in_turn
    self.test_enable_feature(feature, options)
  File "/home/pbarker/rc/ardupilot/./Tools/autotest/test_build_options.py", line 226, in test_enable_feature
    self.assert_feature_in_code(defines, feature)
  File "/home/pbarker/rc/ardupilot/./Tools/autotest/test_build_options.py", line 402, in assert_feature_in_code
    raise ValueError(error)
ValueError: feature gated by AP_BARO_THST_COMP_ENABLED not compiled into (copter); extract_features.py bug?
```

peterbarker:
... actually, that's saying the symbol isn't present in Copter.  Strnage.

This is the test that's failing:
```
time ./Tools/autotest/test_build_options.py  --define-match=AP_BARO_THST_COMP_ENABLED
```

peterbarker (approved):

PR #29531, opened 2025-03-14
Plane: proposed changes for 4.6.0-beta5

tridge:
This rolls up some key PRs that are being considered for beta5.
 - https://github.com/ArduPilot/ardupilot/pull/29521
 - https://github.com/ArduPilot/ardupilot/pull/29498
 - https://github.com/ArduPilot/ardupilot/pull/28406
 - https://github.com/ArduPilot/ardupilot/pull/29494

I have test flown the 3rd change myself. The other two have only been SITL or bench tested so far. 
The fast attitude recovery PR needs more consideration as it will impact VTOL control during a forward transition. It fixes an issue where VTOL rate control does not match fixed wing control due to the input shaping in the VTOL control. Before considering this for beta5 we need some flight tests to look at any possible impacts

I will create a custom build tag for this PR to help users do testing

### DIFF
--- a/ArduPlane/ArduPlane.cpp
+++ b/ArduPlane/ArduPlane.cpp
@@ -871,7 +871,7 @@ bool Plane::set_target_location(const Location &target_loc)
         return false;
     }
     // add home alt if needed
-    if (loc.relative_alt) {
+    if (loc.relative_alt && !loc.terrain_alt) {
         loc.alt += plane.home.alt;
         loc.relative_alt = 0;
     }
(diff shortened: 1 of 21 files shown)


IamPete1:
Hasn't made it to the call yet but I would suggest https://github.com/ArduPilot/ardupilot/pull/29494 too. 

Cork push should be fine, will only affect DShot I think? Copter has the extra complication of the fast rate stuff of course, that will need testing. 

VTOL recovery one make me abit nervous, removing that rate liming could result in vehicles crashing in normal operations if there FW time constant results in a higher gain than the VTOL angle P. A halfway step might be to fix the reset to measured altitude to be reset to FW target, then they both start out in the same place and the accel limit will have much less impact. 

Guided changes look fine.

tridge:
> A halfway step might be to fix the reset to measured altitude to be reset to FW target, then they both start out in the same place and the accel limit will have much less impact.

I don't understand what you mean, sorry. I'm guessing you mean attitude not altitude, but when are you thinking we do this? you mean extra calls to attitude_control->reset_target_and_rate() ?
I've added the rudder fix, thanks!

rmackay9:
I've removed this from the 4.6.0 project because -beta5 is already out.  not quite sure what happened with this PR but I think other PRs covered it in the end

PR #29532, opened 2025-03-14
AP_Logger: do not end log transfer at end of log/block-fulfilled 

peterbarker:
this closes a race condition between satisfying the last request in a log (or a request for a random block in a log which is completely satisfied by the request, eg. a client filling gaps in a log) and "start_new_log()" magically restarting logging after a transfer.

This can cause transfers to fail as opening the log for reading can fail as we've left "recent_open_error" set.


tpwrules:
This has the potential to speed up QGC nicely since it has a funny habit of downloading only 512 log blocks at a time. But right now it seems broken.

peterbarker:
> This has the potential to speed up QGC nicely since it has a funny habit of downloading only 512 log blocks at a time. But right now it seems broken.

This is being surprisingly damaging to MAVProxy as well, particularly the "allow current transferer to transfer new block" bit, at least over 433MHz telemetry radio.  MAVProxy tiems out from a transfer if it doesn't get something within 0.7 seconds, and that happens a lot over 433 radios as we *also* try to shove way too much stuff into the pipe and corrupt it a lot.  The current formulation has ArduPilot ignore the fresh range request while transfering the old, the new formulation would nuke the old request for the entire file and start going block-to-block, stopping the streaming-of-whole-log which is what we'd prefer to be doing.

PR #29533, opened 2025-03-15
AP_DDS: Publish airspeed when unhealthy at a low rate

Ryanf55:
# Purpose

If airspeed is unhealthy, publish it anyways at a low rate. The message doesn't include a health status to show it's unhealthy (yet), so instead we populate NAN's.


tizianofiorenzani (approved):
LGTM

[edited] there are some change requests.

libraries/AP_DDS/AP_DDS_config.h
```diff
@@ -82,8 +82,12 @@
...
 #endif
 
-#ifndef AP_DDS_DELAY_AIRSPEED_TOPIC_MS
-#define AP_DDS_DELAY_AIRSPEED_TOPIC_MS 33
+#ifndef AP_DDS_DELAY_AIRSPEED_TOPIC_HEALTHY_MS
+#define AP_DDS_DELAY_AIRSPEED_TOPIC_HEALTHY_MS 33
+#endif
+
+#ifndef AP_DDS_DELAY_AIRSPEED_TOPIC_UNHEALTHY_MS
+#define AP_DDS_DELAY_AIRSPEED_TOPIC_UNHEALTHY_MS 33
```
tizianofiorenzani, line 90: If timing is the same, why not just publishing at constant rate and just put a NAN?

Ryanf55:
```suggestion
#define AP_DDS_DELAY_AIRSPEED_TOPIC_UNHEALTHY_MS 1000
```

tizianofiorenzani:
It seems there is still a reference to `AP_DDS_DELAY_AIRSPEED_TOPIC_MS`

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -1555,6 +1559,7 @@ void AP_DDS_Client::write_tx_local_velocity_topic()
 void AP_DDS_Client::write_tx_local_airspeed_topic()
 {
     WITH_SEMAPHORE(csem);
+    last_airspeed_time_ms = cur_time_ms;
```
tizianofiorenzani, line 1562: Is `cur_time_ms` not defined locally?

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -1712,10 +1717,13 @@ void AP_DDS_Client::update()
     }
 #endif // AP_DDS_LOCAL_VEL_PUB_ENABLED
 #if AP_DDS_AIRSPEED_PUB_ENABLED
-    if (cur_time_ms - last_airspeed_time_ms > DELAY_AIRSPEED_TOPIC_MS) {
-        last_airspeed_time_ms = cur_time_ms;
+    if (cur_time_ms - last_airspeed_time_ms > DELAY_AIRSPEED_TOPIC_HEALTHY_MS) {
```
tizianofiorenzani, line 1720: those names are wrong

tizianofiorenzani (changes requested):
It seems that there are wrong variable names and reference to local constants.

PR #29534, opened 2025-03-15
Plane: Add tkoff. and land. to landed_state method

rubenp02:
Updated the landed_state method in the MAVLink GCS module to accurately reflect the vehicle's status during takeoff and landing in the MAV_LANDED_STATE field of EXTENDED_SYS_STATE. The method now returns MAV_LANDED_STATE_TAKEOFF if the vehicle is taking off and MAV_LANDED_STATE_LANDING if it is landing, matching the ArduCopter and Blimp implementations.

In combination with mavlink/qgroundcontrol#12567, this makes it possible to abort landings in QGroundControl with ArduPilot aircraft.


IamPete1 (approved):
LGTM. This is a improvement, although i'm not sure we populate the methods correctly for all modes. @Hwurzburg looks like we might need to add the `is_landing` method to the new autoland mode.

Hwurzburg:
#29553

tridge (approved):

PR #29535, opened 2025-03-15
AP_Baro: Add support for AUAV airspeed sensor absolute channel

IamPete1:
The AUAV airspeed sensor has a second I2C address which provides a absolute reading for use as a barometer. The reading and calculation method is mostly the same so this adds a helper class that is shared between both airspeed and baro. (still need to workout the best way to put defines around that for the case where airspeed is not compiled in)

~~The updates work correctly for me with the airspeed, however my test sensor always returns busy when trying to read the extended calibration coefficients for the absolute channel. If I skip the extended calibration it works fine.~~ - fixed by slowing down the calibration reads.


IamPete1:
image
This is what is going on. The sensor should reply to the coefficient read with status `0x40`. Since it does not the driver just gets stuck trying to read the coefficients.

IamPete1:
I now have this working. The problem was that the absolute channel is clocked much slower than the differential channel so we have to be much more careful with the read timings.

IamPete1:
Marked for dev-call to test the waters on how we feel about a baro driver that depends on a airspeed one.

tridge (changes requested):
need to run a build all I think, not just CI builds (@peterbarker has a good setup for that, or ssh into autotest.ardupilot.org)

IamPete1:
> Hey @IamPete1 @tridge I'm not able to get Abs pressure working via periph. Any suggestions?

This should now be fixed.

IamPete1:
I have built everything for all targets with `./Tools/scripts/configure_all.py --build  --no-bl` (modified to build all vehicles)

The result is:
```
Failed builds:
  build: ARKV6X
  build: Airvolute-DCS2
  build: CubeRedSecondary-IO
  build: H757I_EVAL_intf
  build: KakuteF7-bdshot
  build: iomcu-dshot
  build: iomcu-f103
  build: iomcu-f103-8MHz-dshot
  build: iomcu-f103-dshot
  build: iomcu_f103_8MHz
  build: skyviper-f412-rev1
  build: skyviper-journey
  build: skyviper-v2450
```
ARKV6X, Airvolute-DCS2, H757I_EVAL_intf, fail trying to build periph with a vehicle hwdef.

CubeRedSecondary-IO, iomcu-dshot, iomcu-f103, iomcu-f103-8MHz-dshot, iomcu-f103-dshot, iomcu_f103_8MHz, fail because the script trys to build vehicles for it but the hwdef is for iofirmware.

KakuteF7-bdshot runs out of flash on plane its build sizes says its very close, master also runs out of flash for me.
image

skyviper-f412-rev1, skyviper-journey, skyviper-v2450 error out on heli because of toy mode

So, in summary I think everything that failed is because the script is trying to build things the boards were not means for.

tridge (approved):

PR #29537, opened 2025-03-16
AP_NavEKF3: Check if rangefinder is in range (if in use by EK3_SRC1_POSZ)

rishabsingh3003:
It is currently possible to takeoff in Alt-Hold, have EK3_SRC1_POSZ set to "RANGEFINDER", and have the rangefinder out of range (can be verified in SITL).
This can cause the EK3 height estimates to grow unconstrained till the rangefinder is back in view. 
I recently saw a log where this caused the copter to almost crash.


libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -1143,6 +1144,34 @@ bool NavEKF3::pre_arm_check(bool requires_position, char *failure_msg, uint8_t f
...
+#if AP_RANGEFINDER_ENABLED
+    if (requires_position || requires_height) {
+        const AP_NavEKF_Source::SourceZ height_source = sources.getPosZSource();
+        if (height_source == AP_NavEKF_Source::SourceZ::RANGEFINDER) {
+            const auto *_rng = dal.rangefinder();
+            if (_rng == nullptr) {
+                dal.snprintf(failure_msg, failure_msg_len, "Rangefinder not enabled");
+                return false;
+            }
+            bool rangefinder_healthy = false;
```
tridge, line 1157: can't we just check rngValidMeaTime_ms ?

tridge (changes requested):

libraries/AP_NavEKF3/AP_NavEKF3.cpp
tridge: should be in the new pre_arm check in the backends, so we check all lanes, check it is fusing

rmackay9:
@rishabsingh3003,

I'm pretty sure that I have a bug from a user here in Japan that shows this same issue

tridge:
@rishabsingh3003 do you have time to test the change I suggested?

rishabsingh3003:
@tridge I did get to try your suggestions, however it did not work as expected.
The problem is that there is code that "fakes" a rangefinder and forces RNGFNDx_GNDCLR reading when on "ground"  even if the rangefinder is unhealthy. 
So to the core it looks like there is healthy rangefinder being fused in which will go away as soon as the vehicle arms (at least for Copter).

This is the relevant line I am talking about: 
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp#L96

rishabsingh3003:
https://www.dropbox.com/scl/fi/a0xzssbfpvklh41saz7s4/00000066.BIN?rlkey=5ltbxc7qcn7kp6rf6qfaxr77o&e=1&st=zw4egz9t&dl=0

rmackay9:
Hi @rishabsingh3003,

I spent some time looking into this today and I may have a fix for the underlying issue https://github.com/ArduPilot/ardupilot/pull/30490.  I wonder though, have you ever been able to re-create the issue in SITL?

PR #29540, opened 2025-03-16
AP_NavEKF3: Add table to read EKF states and varainces limits

rishabsingh3003:
This is a NFC.
When reading logs from a vehicle with few diverged states/variances, it's very hard to know if the divergence stopped on its own or if it is hitting the constraints. I attempted to document them here.

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
@@ -1873,6 +1873,22 @@ void NavEKF3_core::ForceSymmetry()
 // if states are inactive, zero the corresponding off-diagonals
 void NavEKF3_core::ConstrainVariances()
 {
+    // Covariance constraints as of March 2025
+    // This table assumes normal operations, there are additional constraints for specific failure modes like "badIMUdata" and "inhibitDelAngBiasStates".
+    // +----------------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


libraries/AP_NavEKF3/AP_NavEKF3_core.cpp
```diff
@@ -1873,6 +1873,22 @@ void NavEKF3_core::ForceSymmetry()
 // if states are inactive, zero the corresponding off-diagonals
 void NavEKF3_core::ConstrainVariances()
 {
+    // Covariance constraints as of March 2025
```
peterbarker, line 1876: I suggest that we modify the log messages to contain this data.  It will be much more useful in the log analysis tools than a comment.

rishabsingh3003: @peterbarker Do you mean adding this data in brackets to the log message documentation?

peterbarker: > @peterbarker Do you mean adding this data in brackets to the log message documentation?

Yes, renaming the existing V00 fields to something more descriptive and adding all of the unit metadata etc etc.

Not to say we can't merge this as-is, but having all of this in the tools without having to go look at the code would be nice.

peterbarker:
Merged, thanks!

PR #29544, opened 2025-03-17
AP_Bootloader: only erase ecc corrupted page if error is detected in a reserved page

bugobliterator:
also fix the logic of num_pages, we don't decrement it in flash_init for flash_base_page, we only do it for reserved page.

### DIFF
--- a/Tools/AP_Bootloader/support.cpp
+++ b/Tools/AP_Bootloader/support.cpp
@@ -96,6 +96,7 @@ void flash_init(void)
            flash_base_page < num_pages) {
         reserved += stm32_flash_getpagesize(flash_base_page);
         flash_base_page++;
+        num_pages--;
     }
     /*
       reduce num_pages to account for FLASH_RESERVE_END_KB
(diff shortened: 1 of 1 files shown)


Tools/AP_Bootloader/support.cpp
```diff
@@ -96,6 +96,7 @@ void flash_init(void)
            flash_base_page < num_pages) {
         reserved += stm32_flash_getpagesize(flash_base_page);
         flash_base_page++;
+        num_pages--;
```
tridge, line 99: i think this break flash_func_sector_size() ?

Tools/AP_Bootloader/support.cpp
```diff
@@ -584,9 +585,28 @@ void check_ecc_errors(void)
     if (ofs < BOARD_FLASH_SIZE*1024) {
         // we must have ECC errors in flash
         flash_set_keep_unlocked(true);
+#if defined(FLASH_RESERVE_END_KB) && FLASH_RESERVE_END_KB > 0
```
tridge, line 588: maybe needs to goto start, or loop?

tridge (changes requested):

Issue #28736, opened 2024-11-25
Add support for signing configuration for separate MAVLink channels


IdoBuchman:
## Feature request

I want to activate signing on my MAVlink communication channel between the FCU and the Mission Planner. When I activate signing, all the mavlink ports get activated, including the mavlink port that communicates with my companion computer. Usually, there's no need to protect a wired Point to Point serial communication, and it requires ongoing support to keep the companion computer signing-key updated (If we want to change it from time to time).

### Solution Proposal
I suggest adding a new bitmask parameter that states which of the mavlink serial ports should work with signing. For keeping backward compatibility, we will make sure the default is 11111 (All ports activated).

Specifically, I found this code snippet under `libraries/GCS_MAVLink/GCS_Signing.cpp` lines 93-100, in updated master (commit 4b040eef93e):
```c
    // activate it immediately on all links:
    for (uint8_t i=0; i<MAVLINK_COMM_NUM_BUFFERS; i++) {
        GCS_MAVLINK *backend = gcs().chan(i);
        if (backend == nullptr) {
            return;
        }
        backend->load_signing_key();
    }
```
I assume that adding here a simple condition can do the job :smile: .

### Alternatives
* Using another existing parameter with a similar meaning to this I want to add, and use it instead of introducing a new one.
  * I searched through all the parameters in the copter documentation and did not find a good enough candidate.
* Adding a way to easily update my secret key on my companion computer
  * I found that it makes the operations and ongoing support inconvenient, and would be happy to have a better option.

**Platform**
[X] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

IdoBuchman:
@peterbarker I’d love to hear your thoughts :)

LupusTheCanine:
I would default to 0 and use bitmask to disable signing.

PR #29546, opened 2025-03-17
New MAVx_OPTIONS parameter to enable channel-specific signing

ben-sembira-1:
**Implements issue #28736** (opened by @IdoBuchman).

Instead of using a bitmask as initially suggested in the issue, this implementation integrates the new parameter into the existing GCS_MAVLINK_Parameters group, which I believe is the most suitable place for it. 😊

### DIFF
--- a/libraries/GCS_MAVLink/GCS.h
+++ b/libraries/GCS_MAVLink/GCS.h
@@ -512,6 +512,14 @@ protected:
     virtual bool persist_streamrates() const { return false; }
     void handle_request_data_stream(const mavlink_message_t &msg);
 
+    AP_Int16 options;
+    enum class Option : uint16_t {
+        MAVLINK2_SIGNING_DISABLED = (1U << 0),
+    };
+    bool option_enabled(Option option) const {
+        return options & static_cast<uint16_t>(option);
+    }
+
     virtual void handle_command_ack(const mavlink_message_t &msg);
     void handle_set_mode(const mavlink_message_t &msg);
     void handle_command_int(const mavlink_message_t &msg);
--- a/libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
+++ b/libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
@@ -202,6 +202,18 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
     // @RebootRequired: True
     // @User: Advanced
     AP_GROUPINFO("_ADSB",   10, GCS_MAVLINK, streamRates[GCS_MAVLINK::STREAM_ADSB], DRATE(GCS_MAVLINK::STREAM_ADSB)),
+
+    // ------------
+    // IMPORTANT: Add new stream rates *before* the _OPTIONS parameter.
+    // ------------
+
+    // @Param: _OPTIONS
+    // @DisplayName: Bitmask for configuring this telemetry channel
+    // @Description: Bitmask for configuring this telemetry channel. For having effect on all channels, set the relevant mask in all MAVx_OPTIONS parameters. Keep in mind that part of the flags may require a reboot to take action.
+    // @RebootRequired: True
+    // @User: Standard
+    // @Bitmask: 0:Accept unsigned MAVLink2 messages
+    AP_GROUPINFO("_OPTIONS",   20, GCS_MAVLINK, options, 0),
     AP_GROUPEND
 };
 #undef DRATE
--- a/libraries/GCS_MAVLink/GCS_Signing.cpp
+++ b/libraries/GCS_MAVLink/GCS_Signing.cpp
@@ -133,7 +133,7 @@ static bool accept_unsigned_callback(const mavlink_status_t *status, uint32_t ms
 void GCS_MAVLINK::load_signing_key(void)
 {
     struct SigningKey key;
-    if (!signing_key_load(key)) {
+    if (option_enabled(Option::MAVLINK2_SIGNING_DISABLED) || !signing_key_load(key)) {
         return;
     }
     memcpy(signing.secret_key, key.secret_key, 32);


olliw42:
IMHO it would make more sense to have an SRx_OPTIONS parameter with this as a bitmask

it's not related to a message, and an _OPTIONS bitmaks would allow other options

ben-sembira-1:
> IMHO it would make more sense to have an SRx_OPTIONS parameter with this as a bitmask
> 
> it's not related to a message, and an _OPTIONS bitmaks would allow other options

Thanks, @olliw42! How many bits should be reserved in the mask for this use case, in your opinion? I noticed that the code includes bitmasks using AP_Int8, AP_Int16 and AP_Int32.

ben-sembira-1:
@olliw42 I've updated the implementation to use a bitmask with an AP_Int16 variable.
Could you take another look and let me know if it’s ready to be merged?

I've verified that it works correctly on my physical setup.

peterbarker (changes requested):
Many conflicts to resolve.  Sorry, while I did make this PR a lot simpler to do (only one place to add the new parameter), it does mean a lot of the copy/paste stuff you did is no longer required.

You *do* need to give a good reason why we need this new option bit.

*Why* would you want to disable mavlink signing on a channel-by-channel basis?

ben-sembira-1:
> Many conflicts to resolve. Sorry, while I did make this PR a lot simpler to do (only one place to add the new parameter), it does mean a lot of the copy/paste stuff you did is no longer required.
> 
> You _do_ need to give a good reason why we need this new option bit.
> 
> _Why_ would you want to disable mavlink signing on a channel-by-channel basis?

Thank you for your response @peterbarker !

When we activate signing, all the mavlink ports get activated, including the mavlink port that communicates with my companion computer. Usually, there's no need to protect a wired Point to Point serial communication, and it requires ongoing support to keep the companion computer signing-key updated (If we want to change it from time to time).

peterbarker:
Closes https://github.com/ardupilot/ardupilot/issues/28736

ben-sembira-oz:
> Ping @ben-sembira-1 - rebase?
> 
> I can do the rebase for you if you like. I've had practise.

Wow, I was planning to start fresh from the current master branch this Wednesday—but if you’re able to take care of that, it would be fantastic! Either way, I’m fully committed to reworking this PR and testing it on my physical setup this Wednesday.

ben-sembira-1:
> Ping @ben-sembira-1 - rebase?
> 
> I can do the rebase for you if you like. I've had practise.

Wow, I was planning to start fresh from the current master branch this Wednesday—but if you’re able to take care of that, it would be fantastic! Either way, I’m fully committed to reworking this PR and testing it on my physical setup this Wednesday.

libraries/GCS_MAVLink/GCS.h
```diff
@@ -106,6 +106,11 @@ bool check_payload_size(mavlink_channel_t chan, uint16_t max_payload_len);
 
 
 #define GCS_MAVLINK_NUM_STREAM_RATES 10
+
+// -----------------------------------------------------------
+// GCS_MAVLINK_Parameters::options options definition:
+#define SR_OPTIONS_ENABLE_MAVLINK2_SINGING (1 << 0)
```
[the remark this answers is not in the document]
ben-sembira-1: @peterbarker Do you mean here to make the "0" value of the options represent SIGNING DISABLED? the current state is that 0 represents ENABLED as it was before the change.

ben-sembira-1:
@peterbarker I’ve reapplied the changes on top of the updated master. I haven’t tested them on my setup yet, but it would be very helpful if you could take a quick look and confirm that the changes are what you intended.

Specifically:
1. The `MAVLINK2_SIGNING` bit logic has been inverted — `0` now means signing is **disabled**, and `1` means it’s **enabled**.
1. Unlike the others, `SRx_OPTIONS` is not of type `DRATE`. As a result, it doesn’t appear in the ordered structs that define messages, entries, and defaults. This differs from the rest, but I assumed it’s the correct approach since this parameter isn’t a _stream rate_ itself — rather, it configures the behavior of the stream rates.

libraries/GCS_MAVLink/GCS.h
```diff
@@ -163,6 +163,12 @@ class GCS_MAVLINK_InProgress
     static uint32_t last_check_ms;
 };
 
+enum class SROption : int16_t
+{
+    None = 0,
```
peterbarker, line 168:
```suggestion
```

libraries/GCS_MAVLink/GCS.h
```diff
@@ -512,6 +518,11 @@ class GCS_MAVLINK
     virtual bool persist_streamrates() const { return false; }
     void handle_request_data_stream(const mavlink_message_t &msg);
 
+    AP_Int16 options;
+    bool option_enabled(SROption option) const {
```
peterbarker, line 522:
```suggestion
    bool option_enabled(Option option) const {
```

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
```diff
@@ -45,6 +49,7 @@ const struct AP_Param::GroupInfo *GCS::_chan_var_info[MAVLINK_COMM_NUM_BUFFERS];
 #define AP_MAV_DEFAULT_STREAM_RATE_EXTRA3 1
 #define AP_MAV_DEFAULT_STREAM_RATE_PARAMS 10
 #define AP_MAV_DEFAULT_STREAM_RATE_ADSB 0
+#define AP_MAV_DEFAULT_STREAM_RATE_OPTIONS STANDARD_DEFAULT_STREAM_RATE_OPTIONS
```
peterbarker, line 52: We don't really need to modify this based on vehicles; just have one default somewhere.

The code blocks you're modifying here we'd actually like to go away!

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
```diff
@@ -202,6 +210,18 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
...
+    // IMPORTANT: Add new stream rates *before* the _OPTIONS parameter.
+    // ------------
+
+    // @Param: _OPTIONS
+    // @DisplayName: Bitmask for configuring this telemetry channel
+    // @Description: Bitmask for configuring this telemetry channel. For having effect on all channels, set the relevant mask in all SRx_OPTIONS parameters. Keep in mind that part of the flags may require a reboot to take action.
+    // @RebootRequired: True
+    // @User: Standard
+    // @Bitmask: 0:Enable MavLink2 signing (Reboot required)
+    AP_GROUPINFO("_OPTIONS",   11, GCS_MAVLINK, options, float(AP_MAV_DEFAULT_STREAM_RATE_OPTIONS)),
```
peterbarker, line 224:
```suggestion
    AP_GROUPINFO("_OPTIONS",   20, GCS_MAVLINK, options, 0),
```

... leave a gap

We can't really have non-zero defaults for bitmasks.  Or at least not fior the first bits we set when we create the parameter.  So we might as well be consistent - leave the default value as zero.

It sucks talking in double-negatives, but there you go.

libraries/GCS_MAVLink/GCS.h
```diff
@@ -512,6 +518,11 @@ class GCS_MAVLINK
...
+        return options & static_cast<int16_t>(option);
```
peterbarker, line 523:
```suggestion
        return options & static_cast<uint16_t>(option);
```

libraries/GCS_MAVLink/GCS.h
```diff
@@ -163,6 +163,12 @@ class GCS_MAVLINK_InProgress
     static uint32_t last_check_ms;
 };
 
+enum class SROption : int16_t
```
peterbarker, line 166:
```suggestion
enum class Option : int16_t
```

peterbarker (changes requested):

ben-sembira-1:
@peterbarker Can you give a look on the test that failed in the previous commit? Do you know what I did that caused it to fail?

This is the failure in the `build-test` job:

image

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
```diff
@@ -202,6 +202,18 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
...
     // @User: Advanced
     AP_GROUPINFO("_ADSB",   10, GCS_MAVLINK, streamRates[GCS_MAVLINK::STREAM_ADSB], DRATE(GCS_MAVLINK::STREAM_ADSB)),
+
+    // ------------
+    // IMPORTANT: Add new stream rates *before* the _OPTIONS parameter.
+    // ------------
+
+    // @Param: _OPTIONS
+    // @DisplayName: Bitmask for configuring this telemetry channel
+    // @Description: Bitmask for configuring this telemetry channel. For having effect on all channels, set the relevant mask in all SRx_OPTIONS parameters. Keep in mind that part of the flags may require a reboot to take action.
```
timtuxworth, line 212: Shouldn't this be MAVx_OPTIONS?

ben-sembira-1: Yeah, will change. Thank you

peterbarker (changes requested):

libraries/GCS_MAVLink/GCS_Signing.cpp
```diff
@@ -106,25 +106,45 @@ void GCS_MAVLINK::handle_setup_signing(const mavlink_message_t &msg) const
...
-    MAVLINK_MSG_ID_RADIO
-};
-    
-static bool accept_unsigned_callback(const mavlink_status_t *status, uint32_t msgId)
-{
-    if (status == mavlink_get_channel_status(MAVLINK_COMM_0)) {
-        // always accept channel 0, assumed to be secure channel. This
-        // is USB on ChibiOS boards
-        return true;
+    static const uint32_t accept_list[] = {
```
peterbarker, line 109: Please don't re-indent the code.  Whitespace wars suck.

libraries/GCS_MAVLink/GCS_Signing.cpp
```diff
@@ -133,7 +153,8 @@ static bool accept_unsigned_callback(const mavlink_status_t *status, uint32_t ms
 void GCS_MAVLINK::load_signing_key(void)
 {
     struct SigningKey key;
-    if (!signing_key_load(key)) {
+    if (!signing_key_load(key))
```
peterbarker, line 156: remove whitespace change

libraries/GCS_MAVLink/GCS_Signing.cpp
```diff
@@ -143,7 +164,7 @@ void GCS_MAVLINK::load_signing_key(void)
     // prevents a window for replay attacks
     signing.timestamp = key.timestamp + 60UL * 100UL * 1000UL;
     signing.flags = MAVLINK_SIGNING_FLAG_SIGN_OUTGOING;
-    signing.accept_unsigned_callback = accept_unsigned_callback;
+    signing.accept_unsigned_callback = generate_accept_unsigned_callback(option_enabled(Option::MAVLINK2_SIGNING_DISABLED));
```
peterbarker, line 167: I don't think you need the `generate_` method here...

Just have an if/else right here?

peterbarker (changes requested):
I think we missed the fact there was this static method involved, complicating things.

We *can* do closurey-type things if we really need to - but I don't think we do.  It would save reboots being required when the parameter changes, but I don't think we need to fuss that much!

libraries/GCS_MAVLink/GCS.h
```diff
@@ -512,6 +512,15 @@ class GCS_MAVLINK
     virtual bool persist_streamrates() const { return false; }
     void handle_request_data_stream(const mavlink_message_t &msg);
 
+    AP_Int16 options;
+    enum class Option : uint16_t
+    {
```
peterbarker, line 517:
```suggestion
    enum class Option : uint16_t {
```

libraries/GCS_MAVLink/GCS_Signing.cpp
```diff
@@ -125,6 +125,11 @@ static bool accept_unsigned_callback(const mavlink_status_t *status, uint32_t ms
     }
     return false;
 }
+
+static bool always_true(const mavlink_status_t *status, uint32_t msgId)
```
peterbarker, line 129:
```suggestion
static bool accept_unsigned_callback_accept_all(const mavlink_status_t *status, uint32_t msgId)
```

libraries/GCS_MAVLink/GCS_Signing.cpp
```diff
@@ -143,7 +148,9 @@ void GCS_MAVLINK::load_signing_key(void)
     // prevents a window for replay attacks
     signing.timestamp = key.timestamp + 60UL * 100UL * 1000UL;
     signing.flags = MAVLINK_SIGNING_FLAG_SIGN_OUTGOING;
-    signing.accept_unsigned_callback = accept_unsigned_callback;
+    signing.accept_unsigned_callback = option_enabled(Option::MAVLINK2_SIGNING_DISABLED)
+        ? always_true
+        : accept_unsigned_callback;
```
peterbarker, line 153:
```suggestion
    if (option_enabled(Option::MAVLINK2_SIGNING_DISABLED) || chan == MAVLINK_COMM_0) {
        signing.accept_unsigned_callback = always_true;
    } else {
        signing.accept_unsigned_callback = accept_unsigned_callback;
    }
```

... and remove the `if` in `accept_unsigned_callback` for COMM_0

peterbarker (changes requested):
So close now.

Please state what testing you've done on it.  We've changed a lot here, so need to know it's been re-tested and actually does what it claims to!

libraries/GCS_MAVLink/GCS_Signing.cpp
```diff
@@ -113,18 +113,18 @@ static const uint32_t accept_list[] = {
     
 static bool accept_unsigned_callback(const mavlink_status_t *status, uint32_t msgId)
 {
-    if (status == mavlink_get_channel_status(MAVLINK_COMM_0)) {
-        // always accept channel 0, assumed to be secure channel. This
-        // is USB on ChibiOS boards
```
peterbarker, line 118: Need to move and augment this comment rather than just losing it.

peterbarker (changes requested):
Apart from the comment, just the commit list to fix to conform to our standards,  just one commit, `GCS_MAVLink: add option to accept unsigned packets on any mavlink connection"

peterbarker:
> The checks that failed:
>     * https://github.com/ArduPilot/ardupilot/actions/runs/15021673634
>     * https://github.com/ArduPilot/ardupilot/actions/runs/15021673662

The colcon tests are bouncing; something about missing messages, I'm told.  It's a known problem and I believe someone is looking at it.

ben-sembira-1:
@peterbarker I reverted most of our changes to fit your comment.

@IdoBuchman Can you please check the 3 things @peterbarker asked for on the physical setup?

Here are the compiled binaries of the current version (hash 75addd): https://limewire.com/d/ciPdR#rp4Zibcyqv

tridge:
@ben-sembira-1 can you explain why it didn't work in accept_unsigned_callback() ?

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
```diff
@@ -202,6 +202,18 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
     // @RebootRequired: True
     // @User: Advanced
     AP_GROUPINFO("_ADSB",   10, GCS_MAVLINK, streamRates[GCS_MAVLINK::STREAM_ADSB], DRATE(GCS_MAVLINK::STREAM_ADSB)),
+
+    // ------------
+    // IMPORTANT: Add new stream rates *before* the _OPTIONS parameter.
+    // ------------
+
+    // @Param: _OPTIONS
+    // @DisplayName: Bitmask for configuring this telemetry channel
```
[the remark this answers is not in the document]
rmackay9: Thanks for the change by the way!

ben-sembira-1:
> @ben-sembira-1 can you explain why it didn't work in accept_unsigned_callback() ? What were the symptoms?

@tridge Hey! I don't know why it didn't work😅 To be honest I still lack in understanding the accept_unsigned_callback() function.

The test case was:

- 2 mission planner computers (one with signing enabled and the same signing key as the pixhawk, the other with signing disabled)
- one pixhawk CubeOrangePlus

We set MAV2_OPTIONS=1 MAV3_OPTIONS=0, and then:
1. Connect the ADT to telem1 in the pixhawk
1. Connect the GDT to the MP with the signing key - see telemetries and change flight modes.
1. Same for the MP without signing
1. Change the connection in the pixawk to telem2 and do it again.

The expected results are that the MP with the signing key will always success connecting, and the other MP will fail on telem2 but succeed on telem3 (I hope I didn't jumble the order but you get the idea)

In the current state - it worked, but in the previous state using the accept_unsigned_callback() function the MP without the signing key failed connecting in both telem ports.

BTW @peterbarker I don't know if this is intentional or not, but consider start the MAV indexes from 0 so it will be synced with the hardware that most of the times call the first telemetry link telem1 (which is called MAV2 currently). Before the new changes we had SR0 for the serial USB connection and then the numbers lined up well.

peterbarker:
> BTW @peterbarker I don't know if this is intentional or not, but consider starting the MAV indexes from 0 so it will be synced with the hardware that most of the times call the first telemetry link telem1 (which is called MAV2 currently). Before the new changes we had SR0 for the serial USB connection and then the numbers lined up well.

We discussed that at DevCall and decided to keep the same convention we use on all of our backends - instances count from 1.

It's important to keep the distinction between serial and mavlink connections clear - they are decoupled, and that's important because modern backends can correspond to network ports or CAN serial passthrough connections rather than UARTs!

PR #29547, opened 2025-03-17
Copter: LOG_BITMASK includes PID logging by default 

rmackay9:
This is an alternative to PR https://github.com/ArduPilot/ardupilot/pull/28774

The changes vs the original PR is just that we disable the PID logging during the ATTITUDE_FAST test.  This test was failing because the ANG log messages were not being logged at 400hz but instead between 280 and 320hz in SITL.  This seems to be a feature of SITL only though, on a CubeOrangePlus the ANG logs did appear at 400hz


Tools/autotest/arducopter.py
```diff
@@ -10599,10 +10599,13 @@ def ATTITUDE_FAST(self):
         self.context_push()
         ex = None
         try:
-            old = self.get_parameter('LOG_BITMASK')
-            new = int(old) | (1 << 0)  # see defines.h
+            log_bitmask_old = int(self.get_parameter('LOG_BITMASK'))
+
+            # enable fast logging (bit 1) and disable PID logging (bit 12)
+            log_bitmask_new = (log_bitmask_old | (1 << 0)) & ~(1 << 12)
```
peterbarker, line 10605: Need to add a big comment in here as to *why* we are doing it.  I don't think we *should* need to do it, so we should at least leave a note for someone in the future as to what's going on.

peterbarker (changes requested):

tridge (approved):

Tools/autotest/arducopter.py
rmackay9: comment added, txs

# disable PID logging (bit 12) to avoid slowdowns in attitude logging in SITL

rmackay9:

PR #29550, opened 2025-03-17
AP_RCProtocol: fix bug that at least two serial configs are needed

heeplr:
Previously it didn't work with ARRAY_SIZE(serial_config) == 1


peterbarker:
Huh.  Whoops.  How  did you hit this as being a problem?

heeplr:
I commented out all protocols but one, to try not having the baudrate changed.

peterbarker:
Merged, thanks @heeplr !

PR #29554, opened 2025-03-18
Plane: change loc alt frame in set_guided_WP rather than callers

peterbarker:
centralise this to reduce accesses to the attributes

This changes our logic around mode_LoiterAltQLand - if terrain is not enabled then we would set the altitude relatve-to-home in `next_WP_loc`.  There are many places in the code which assume that next_WP_loc is in either terrain-relative or absolute frames.

@IamPete1 do you see any problems if we instantly convert the home-relative next-wp-loc for loiteraltqland to the absolute frame rather than home-relative frame?


peterbarker:
This isn't going to work as-is.  For example, when handling a do-reposition we may pass in a home-relative-alt location and be unable to convert it within `set_guided_WP`.  Since the caller has already changed the mode, this leaves you in a poor position if you want to pass `false` back from `set_guided_WP`...

Issue #29530, opened 2025-03-14
AP_Camera_Backend::Write_CameraInfo does not handle uninitialized location


blaab2:
## Bug report

**Issue details**

This is a HAL_BOARD_SITL bug report. The panic function within _libraries/AP_Common/Location.cpp_ is triggered if a camera trigger is requested without the location being available.

**code details**

 This occurs as the 
https://github.com/ArduPilot/ardupilot/blob/7601c8388eee00650632e6b323f7101c80372002/libraries/AP_Common/Location.cpp#L109
 panic function will be raised by `AP_Camera_Backend::Write_CameraInfo` if camera trigger is requested (for example via mission planner) and location is not known yet. The unguarded call to `current_loc.get_alt_cm` raises this.

**fix**

https://github.com/ArduPilot/ardupilot/blob/7601c8388eee00650632e6b323f7101c80372002/libraries/AP_Camera/AP_Camera_Logging.cpp#L33
and
https://github.com/ArduPilot/ardupilot/blob/7601c8388eee00650632e6b323f7101c80372002/libraries/AP_Camera/AP_Camera_Logging.cpp#L37
should be
`if (current_loc.initialised() && !current_loc.get_alt_cm(Location::AltFrame::ABSOLUTE, altitude_cm)) {`
and
`if (current_loc.initialised() && !current_loc.get_alt_cm(Location::AltFrame::ABOVE_HOME, altitude_rel_cm)) {`

I could send a PR if requested.

**Version**
master branch
4.7.0-dev

**Platform**
[?] All
[  ] AntennaTracker
[X] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
_irrelevant_

**Hardware type**
HAL_BOARD_SITL

**Logs**
```
#5  0x000055ebb20e1175 in AP_HAL::panic (errormsg=0x55ebb2231600 "Should not be called on invalid location: Location cannot be (0, 0, 0)") at ../../libraries/AP_HAL_SITL/system.cpp:51
        ap = {{gp_offset = 8, fp_offset = 48, overflow_arg_area = 0x7ffcabb23be0, reg_save_area = 0x7ffcabb23b20}}
```

rmackay9:
Hi @blaab2,

[...]

PR #29558, opened 2025-03-18
AP_Camera: avoid SITL panic if pic taken with no pos estimate

rmackay9:
This resolves issue https://github.com/ArduPilot/ardupilot/issues/29530 by checking that the current_loc has been initialised before attempting to get the altitude from it.

This fix is thanks to @blaab2 and the commit is co-authored with him, thanks!

I've lightly tested this in SITL by performing a before/after test to confirm the issue occurs before this fix and does not afterwards.

BTW, the issue can be reproduced in SITL by doing this:

1. rm eeprom.bin file
2. start SITL
3. param set LOG_DISARMED 1
4. param set GPS1_TYPE 0
5. restart SITL
6. right-mouse-button click on MP's map and select, "Trigger Camera NOW"

Below is a screenshot of what appears in the LOGS which is a mostly zero message
image

### DIFF
--- a/libraries/AP_Camera/AP_Camera_Logging.cpp
+++ b/libraries/AP_Camera/AP_Camera_Logging.cpp
@@ -30,14 +30,12 @@ void AP_Camera_Backend::Write_CameraInfo(enum LogMessages msg, uint64_t timestam
     }
 
     int32_t altitude_cm = 0;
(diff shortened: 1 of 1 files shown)


peterbarker (approved):
LGTM

PR #29559, opened 2025-03-18
libraries: SRV_Channel: specify Sub values for SERVOn_FUNCTIONs

ES-Alexander:
Found an old private discussion with @Williangalvani about this, and realised I never followed through with it.

This is a pretty rough pass from memory of available features, without much checking of actual implementation in the source.
I think this is worth including as-is, but if robust values are required we'll need to check at least:

| Value | Label | PR Includes? |
|---|---|---|
| 10 | CameraTrigger | N |
| 12-15 | Mount2* | N |
| 28 | Gripper | Y |
| 61 | Video Switch | Y |
| 70 | Throttle | N |
| 88 | Winch | Y |
| 90-93 | Camera* | Y |
| 94-109 | Script* | Y |
| 120-123 | NeoPixel* | Y |
| 124-127 | Rate* | N |
| 129-132 | ProfiLED* | N |
| 133 | Winch Clutch | N |
| 138-139 | Alarm* | N |


Hwurzburg:
Aux Function wiki page needs rework for sub, as do many others

Williangalvani (approved):

Issue #29555, opened 2025-03-18
disarm in fwd flight on all motor tailsitter


tridge:
disarming in an all-motor tailsitter quadplane in a fwd flight mode like FBWA does not stop the motors

rmackay9:
I've removed this *issue* from the 4.6 project because the PR that resolves it has already been added

PR #26249, opened 2024-02-18
Plane: fixed ramp of motors on back transition in tilt quadplanes

tridge:
this removes the special case for SHUT_DOWN state of the motors in tilt quadplanes. The old code seemed to assume that the output_motor_mask() doesn't work in SHUT_DOWN state, but in fact it does, so we can get proper spoolup behavious on back transition just by removing the special case

this PR builds on https://github.com/ArduPilot/ardupilot/pull/26248 to give us a quadplane-tilt frame for testing

PR #29561, opened 2025-03-18
Plane: tailsitter: ensure none-masked motors are zeroed in forward flight

IamPete1:
fixes https://github.com/ArduPilot/ardupilot/issues/29555. Thanks to https://github.com/ArduPilot/ardupilot/pull/26249 we were leaving the motors not masked for forward flight to be smoothly ramped back to zero for tiltrotors. However we overlooked the tailsitter case. Because the normal motor output is not called the motors were not being zeroed. The fix is to call the output function.

### DIFF
--- a/ArduPlane/tailsitter.cpp
+++ b/ArduPlane/tailsitter.cpp
@@ -289,6 +289,12 @@ void Tailsitter::output(void)
         return;
     }
 
+    if (!hal.util->get_soft_armed() ||
+        SRV_Channels::get_emergency_stop()) {
+        // Ensure motors stop on disarm
+        motors->output_min();
+    }
+
     // throttle 0 to 1
     float throttle = SRV_Channels::get_output_scaled(SRV_Channel::k_throttle) * 0.01;
 


IamPete1:
Disarming in FBWA on the HWing in realflight. 

image

IamPete1:
There could be a argument for setting the desired spool state rather than forcing min, that would do a nice spool down, but we did not do that before and we won't spool them up nicely again afterwards without some larger changes. So I think this is OK.

tridge:
@IamPete1 looks good, I'd just need to do some testing

tridge (dismissed):
@IamPete1 I tested on CatTailsitter in RealFlight. When I transition to FBWA the fwd motors stop
image
log https://uav.tridgell.net/tmp/cattailsitter5.bin

tridge:
image

tridge:
@IamPete1 I have force pushed a fix

IamPete1:
> I have force pushed a fix

I don't think it will work in the E-Stop case. Sorry, I do have something in mind but have not had time yet.

tridge:
@IamPete1 I've added emergency stop check as discussed

IamPete1:
Sorry, I thought of another case where we might do the wrong thing still. That is a "copter tailsiitter" which does not use all motors for forward flight, I think the motors that are not used in forward flight will just stick at there last VTOL value until you either disarm and this change kills them or you re-enter VTOL and they pick back up again. The test would be a quad tailsitter with `Q_TAILSIT_MOTMX`< 15.

Having said that this patch is still a improvement on what we have now.

tridge:
> Having said that this patch is still a improvement on what we have now.

ok, merging this for 4.6.0-beta5

rmackay9:
this is included in 4.6.0-beta5 via PR https://github.com/rmackay9/rmackay9-ardupilot/pull/42

Issue #28505, opened 2024-10-30
Add suport for MAV_CMD_DO_SET_ROI_LOCATION in AP_Mission


bytesByHarsh:
## Feature request

Currently only in `GCS_Common.cpp` file `MAV_CMD_DO_SET_ROI_LOCATION` support is added, the command sent via Command Long Packet is parsed properly but when `MAV_CMD_DO_SET_ROI_LOCATION` is sent as a mission item it returns `MAV_RESULT_UNSUPPORTED` to GCS.

**Describe the solution you'd like**
A possible solution can be adding a switch statement in the `AP_mission.cpp` file for `MAV_CMD_DO_SET_ROI_LOCATION`.

**Platform**
[ X ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

If required I can create a MR for adding these changes in the code.

bytesByHarsh:
Also if possible can we add the support for `MAV_CMD_DO_SET_ROI_NONE` as well.

entone:
Would love to see support for this as well. 

Does Mission Planner use different commands that are supported to provide the same functionality?

rmackay9:
Hi @entone, AP supports the very similar DO_SET_ROI which can be seen in MP's Plan screen's command field's drop-down.  It's also documented [here on the wik](https://ardupilot.org/copter/docs/mission-command-list.html#do-set-roi)i

DonLakeFlyer:
Right but DO_SET_ROI was deprecated 7 years ago. QGC only supports the current ROI command set. These are used not only in missions, but in setting an ROI directly while flying as well as the ROI_WPNEXT_OFFSET variant when building structure scan missions. As it stands now because of this a decent sized pile of good user functionality in QGC is not available to Ardupilot users. I realize the majority of Ardupilot users might use Mission Planner. But there are still a huge pile that use QGC. As far as I see the new ROI commands are basically just a restructuring of the old ones. I don't think there is new functionality, just a rearranging of the deck chars. I don't think it should not be super hard to support the new set while still keeping support for the old ones.

DonLakeFlyer:
Something along these lines as a start:
```
don@Donalds-Mac-Pro ardupilot % git diff
diff --git a/ArduCopter/mode_auto.cpp b/ArduCopter/mode_auto.cpp
index c5be03c8b1..860424440a 100644
--- a/ArduCopter/mode_auto.cpp
[...]

Issue #29566, opened 2025-03-19
DO_SET_ROI_LOCATION and DO_SET_ROI_NONE "gimbal device id" fields should be supported


rmackay9:
AP supports the  [DO_SET_ROI_LOCATION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_LOCATION) and [DO_SET_ROI_NONE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_NONE) as "immediate" and mission commands but we do not correctly consume "param1" ("Gimbal device ID").  Instead it is always the first camera gimbal that is controlled.  We should consume this field and control the gimbal device specified

G1-k:
Hi @rmackay9, 
I'm currently working on this issue and have been exploring how to support the gimbal_device_id field for DO_SET_ROI_LOCATION and DO_SET_ROI_NONE. It looks like handle_command_do_gimbal_manager_pitchyaw in AP_Mount already makes use of the gimbal ID field. Should the ROI handling (set_roi_target) also use the same gimbal selection logic as handle_command_do_gimbal_manager_pitchyaw?
I'd appreciate any guidance you can provide.

rmackay9:
Hi @G1-k, thanks for looking into this.  yes, I think the do-set-roi-location and do-set-roi-none should consume the gimbal id field in the same way as do-gimbal-manage-pitchyaw handler.  thanks!

ahmedmsalah99:
Hello @rmackay9 
This is my first contribution and I have added and tested this feature. Can you please check ?

ahmedmsalah99:
@rmackay9 
I know it's a big push that's hard to review. Should we make this issue for the auto mode only ?
I can keep auto mode commits only for copter and plane which are well tested. 

For the other parts we can create separate issues to make them easier to test. 
What do you think ?

khanhanif1136-commits:
Mu kgjcbnb

khanhanif1136-commits:
Egg and car aur

bringert:
While it would be nice to have support for specifying the gimbal device id to set the ROI for, maybe a simpler first step would be to just support param1 == 0 correctly, that is, to set the ROI for **all gimbals**, instead of only for the first one? That would be enough for my current use case at least.

PR #29567, opened 2025-03-19
Copter, Rover, Plane, Sub, Mission: add DO_SET_ROI_LOCATION and DO_SET_ROI_NONE support

rmackay9:
This adds basic support for the [DO_SET_ROI_LOCATION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_LOCATION) and [DO_SET_ROI_NONE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_NONE) mission commands

The support is only "basic" because we don't properly consume "param1" ("Gimbal device ID") so regardless of this argument's value only the first camera gimbal will be controlled (see issue https://github.com/ArduPilot/ardupilot/issues/29566).

This has been lightly tested in SITL on Copter.  We should also test on Rover, Plane and Sub.

This resolves issue https://github.com/ArduPilot/ardupilot/issues/28505

Some ground stations updates are also required to handle the DO_SET_ROI_LOCATION in the same way as we handle DO_SET_ROI.  For example, below is a screen shot of MAVProxy.  The white lines make it appear as if the vehicle will travel to the ROI_LOCATION's location.

- Mission Planner PR https://github.com/ArduPilot/MissionPlanner/pull/3507
- MAVProxy Issue: https://github.com/ArduPilot/MAVProxy/issues/1531

### DIFF
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -766,8 +766,11 @@ bool ModeAuto::start_command(const AP_Mission::Mission_Command& cmd)
         do_set_home(cmd);
         break;
 
+    case MAV_CMD_DO_SET_ROI_LOCATION:       // 195
+    case MAV_CMD_DO_SET_ROI_NONE:           // 197
     case MAV_CMD_DO_SET_ROI:                // 201
         // point the copter and camera at a region of interest (ROI)
+        // ROI_NONE can be handled by the regular ROI handler because lat, lon, alt are always zero
         do_roi(cmd);
         break;
 
@@ -992,6 +995,8 @@ bool ModeAuto::verify_command(const AP_Mission::Mission_Command& cmd)
     // do commands (always return true)
     case MAV_CMD_DO_CHANGE_SPEED:
     case MAV_CMD_DO_SET_HOME:
+    case MAV_CMD_DO_SET_ROI_LOCATION:
+    case MAV_CMD_DO_SET_ROI_NONE:
     case MAV_CMD_DO_SET_ROI:
 #if HAL_MOUNT_ENABLED
     case MAV_CMD_DO_MOUNT_CONTROL:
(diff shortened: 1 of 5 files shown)


ArduCopter/mode_auto.cpp
```diff
@@ -992,6 +994,8 @@ bool ModeAuto::verify_command(const AP_Mission::Mission_Command& cmd)
     // do commands (always return true)
     case MAV_CMD_DO_CHANGE_SPEED:
     case MAV_CMD_DO_SET_HOME:
+    case MAV_CMD_DO_SET_ROI_LOCATION:
+    case MAV_CMD_DO_SET_ROI_NONE:
```
peterbarker, line 998: Is there any reason not to do this in `AP_Mission::verify_command()`?

rmackay9: Hi @peterbarker, thanks for the review.  Functionally that would be fine but I think it would be odd to have half the implementation in AP_Mission and half in the vehicle code.

ArduCopter/mode_auto.cpp
```diff
@@ -766,6 +766,8 @@ bool ModeAuto::start_command(const AP_Mission::Mission_Command& cmd)
         do_set_home(cmd);
         break;
 
+    case MAV_CMD_DO_SET_ROI_LOCATION:       // 195
+    case MAV_CMD_DO_SET_ROI_NONE:           // 197
     case MAV_CMD_DO_SET_ROI:                // 201
         // point the copter and camera at a region of interest (ROI)
```
peterbarker, line 772:
```suggestion
        // point the copter and camera at a region of interest (ROI)
        // in the case of NONE then lat/lng/alt will all be zero and that is handled especially in AutoYaw::set_roi
```

peterbarker (approved):

ArduCopter/mode_auto.cpp
rmackay9: comment added, thanks for the review!

rmackay9:
I've tested this on Rover and Plane now and it seems to be working. With Rover it is quite clear that it's working correctly, with Plane it's a little more difficult but surely it's fine.

I'll merge after it passes CI

rmackay9:

PR #29568, opened 2025-03-19
Create and use MAV_ parameter namespace

peterbarker:
I've only moved one parameter, but `SYSID_MYGCS`  is also an obvious one to move.

Entry is required in each vehicle as we subclass the GCS object for each vehicle (eg. `GCS_Plane`)

Past this we should sprout `MAVn_OPTIONS` so we can move things from `SERIALn_OPTIONS` (eg. "private").

... and also have `MAV1_SR_EXTRA` etc etc.

I have tested each vehicle's parameter upgrade individually.  The automatic testing script fails because the system ID changes!

### DIFF
--- a/Tools/AP_Periph/Parameters.h
+++ b/Tools/AP_Periph/Parameters.h
@@ -99,6 +99,7 @@ public:
         k_param_esc_extended_telem_rate,
         k_param_imu_sample_rate,
         k_param_imu,
+        k_param__gcs,
     };
 
     AP_Int16 format_version;


tridge:
it would be nice for the MAVn subgroups to be SUBGROUPPTR so they appear when allocated

peterbarker:
`MAV1_PORT==21` meaning a network port that's doing mavlink...

amilcarlucas:
Good idea!

timtuxworth:
The current "mapping" between NETx ports and SRx_ values when the net protocol is mavlink is ... well I have no idea how it works honestly, so fixing this is a good thing. 

I like the idea of doing away with the arbitrary limit of 7 mavlink ports (with the unclear mapping of SRx to SERIALx vs NETx), but if you are going to allow > 9, **please use MAVxx** not MAVx so that they will sort in order when accessed and displayed.

I like Andrew's suggestion of just simply going with a static mapping to keep it simple, a complex mapping scheme is just going to be hard to explain and easy for people to get wrong ... so if I understand it right you would have

SERIAL1 -> MAV01_SR...
SERIAL2 -> not a mavlink port so MAV02 parameters are ignored
...
SERIAL7 -> MAV07_SR..

then (and no lets not use MAV20 - it should be P1 = 21)
NET_P1.. -> MAV21_SR...
NET_P2.. -> maybe mapped to a Gimbal so MAV21 parameters are ignored.
NET_P3 -> MAV23_SR

and I guess the same thing at 30 for mavlink over CAN?

Is that the idea?

peterbarker:
> ```
> NET_P1.. -> MAV21_OPTIONS, MAV21_SR...
> NET_P2.. -> maybe mapped to a Gimbal so MAV21 parameters are ignored.
> NET_P3 -> MAV23_OPTIONS, MAV23_SR
> ```

That's not how this is going to work.  The mapping will essentially stay the same; The `SR4_` parameters will simply move to `MAV03_` parameters.  I like the zero-padding idea, I'm definitely going to try to use that!

We will add a "magic" parameter which will give you a number which can be mapped to an object.  If below 20 it will be a serial port number (ie. "telem2" would be "2").  If above 20 it will be a network port (so 22 will be `NET_P2`.  If above 30... yeah, mavlink-over-CAN.  This was Randy's idea.

Note that this PR only starts to pave the way for those new `MAVxx` parameter blocks.  This one is just moving the global parameters initially.

timtuxworth:
> We will add a "magic" parameter which will give you a number which can be mapped to an object. If below 20 it will be a serial port number (ie. "telem2" would be "2"). If above 20 it will be a network port (so 22 will be `NET_P2`. If above 30... yeah, mavlink-over-CAN. This was Randy's idea.

Where will you add this "magic" parameter? Just trying to understand where this is going.

Rover/Parameters.cpp
```diff
@@ -394,6 +389,12 @@ const AP_Param::Info Rover::var_info[] = {
     // @Path: ../libraries/AP_Vehicle/AP_Vehicle.cpp
     PARAM_VEHICLE_INFO,
 
+#if HAL_GCS_ENABLED
+    // @Group: MAV_
+    // @Path: ../libraries/GCS_MAVLink/GCS.cpp
+    GOBJECT(_gcs,           "MAV_",  GCS),
```
tridge, line 395: underscore in wrong spot

tridge (commented):
we need a plan for handling SR conversion before we merge, I think we don't need to do SR param conversion

PR #29573, opened 2025-03-19
AC_Fence: Fixed wording in error messages

Georacer:
In both of those messages, the wording contradicts the actual logical check. The text has been adapted to match.


andyp1per (approved):

Georacer:
Strange that we had this one wrong for so long and no-one noticed.

PR #29577, opened 2025-03-20
SITL: SIM_FlightAxis: use an objectbuffer for holding connections to FlightAxis

peterbarker:
So the existing code was doing a bang-up job of double-buffering... using three pthread mutexes.

We have an ObjectBuffer_TS we can use to get the same effect, but it's simpler and generalises to having more waiting connections.

This *also* doesn't use pthread constructions, meaning it's more portable.

There's no discernible performance difference between this and master.  *Perhaps* a little better when running with a three socket queue - but there's a *lot* of noise!


tridge (approved):
seems to be very similar speed for me

PR #29581, opened 2025-03-20
Deadreckon

fortisON:


rmackay9:
Hi @fortisON,

Maybe this wasn't meant to be raised as a PR?  This looks like a C++ implementation of [this Lua script](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/applets/copter-deadreckon-home.lua) with some extra changes.

Anyway, I'll go ahead and close this but I'll see replies so if you'd like to keep it open and try to get it into the main C++ code we can perhaps help with that.  Txs!

PR #29582, opened 2025-03-20
AP_Periph: use boolean for AP_PERIPH_HAVE_LED_WITHOUT_NOTIFY

peterbarker:
... not defined()


peterbarker:
This is a no-compiler-output-change

magicrub (approved):
I'm happy with the change, and it is indeed better,, just curious what instigated it.

peterbarker:
> I'm happy with the change, and it is indeed better,, just curious what instigated it.

We're changing all of the `#ifdef` things to `#if` things whereever possible - *especially* in AP_Periph.  This will allow us to add an option to custom.ardupilot.org to build AP_Periph firmwares.

PR #29587, opened 2025-03-21
Add support for Loweheiser generator

Georacer:
# Summary

This PR adds support for [Loweheiser-brand on-board generators](https://www.loweheiser.com/hybrid-drone-generators/).

# Details

Simulation files for this generator already existed in ArduPilot (and some autotests). This PR adds the actual driver.

The bulk of the logic was written by @peterbarker, for ArduPilot 4.2. I did the rebasing to `master`.

Many thanks to [Harris Aerial](https://harrisaerial.com/) who sponsored this work.

I do not have hands-on experience with the hardware, so any and all reviews are welcome.

# Testing

New autotests have been added to explore some of the functionality.

Harris Aerial have tested their rebased branch (along with other commits) and report that their aircraft carrying the generator works as before.

Loweheiser's people tested Harris' branch and report that their system works as expected.

# Known issues

## EFI - Generator interdependency

This driver has a common backend for the EFI and the generator, which belongs to `AP_Generator`. That means that the `AP_EFI_Loweheiser` define needs to take into account that the generator part is also defined. Otherwise, there are builds that include EFI, don't include Generator, and thus fail due to include dependency failures.
This seems to do the job:
```c++
#ifndef AP_EFI_LOWEHEISER_ENABLED
#define AP_EFI_LOWEHEISER_ENABLED AP_EFI_BACKEND_DEFAULT_ENABLED && AP_GENERATOR_LOWEHEISER_ENABLED
#endif
```
although AP_EFI now has a dependency on AP_Generator.

## Prearm behaviour

Harris Aerial reports that if the generator is hot (from a previous run), and is set in "run mode", then the UA will arm even though the generator is not actually running.

# Miscellaneous

The branch that @peterbarker worked on also contained some small improvements for the [Intelligent Energy 2.4kW fuel cell](https://ardupilot.org/copter/docs/common-ie24-fuelcell.html). These were sneaked in this PR as well.


Georacer:
The build fails seemingly due to .h include issues. I'm investigating.

libraries/AP_Generator/AP_Generator_config.h
```diff
@@ -25,3 +25,7 @@
 #ifndef AP_GENERATOR_RICHENPOWER_ENABLED
 #define AP_GENERATOR_RICHENPOWER_ENABLED AP_GENERATOR_BACKEND_DEFAULT_ENABLED && HAL_PROGRAM_SIZE_LIMIT_KB > 1024
 #endif
+
+#ifndef AP_GENERATOR_LOWEHEISER_ENABLED
+#define AP_GENERATOR_LOWEHEISER_ENABLED AP_GENERATOR_BACKEND_DEFAULT_ENABLED
```
tridge, line 30: && HAL_PROGRAM_SIZE_LIMIT_KB > 1024

tridge: for custom build only use HAL_PROGRAM_SIZE_LIMIT_KB > 2048
that would be my preference

tridge:
needs to be added into Tools/scripts/build_options.py

Hwurzburg (changes requested):
This needs IFDEF additions to make it an optional addition(defaulted as not being included since its a rare application) as well as entries in build_options and extract_features

libraries/AP_Generator/AP_Generator_config.h
Georacer: Done.

Georacer:
> This needs IFDEF additions to make it an optional addition(defaulted as not being included since its a rare application) as well as entries in build_options and extract_features

This has now been marked with `HAL_PROGRAM_SIZE_LIMIT_KB > 2048`, which means it will only get build in SITL by default (or at least on boards with woads of space).

Hwurzburg (approved):

tridge (approved):

Georacer:
Hi @alfmh,

As you mention, it's likely preferable to have users manually engage the starter, if it is present in their system (they can always crank the engine, right?).

I have another PR in the making, we can take the discussion there.

Thank you for drafting the instructions, probably the best course of action is to create a Wiki page and put it there.
More on that later, after we're happy with the functionality of the code.

alfmh:
Hi @Georacer,

> As you mention, it's likely preferable to have users manually engage the starter, if it is present in their system (they can always crank the engine, right?).

By the moment, yes. The starter works at testing but it isn't a in middle term ready. for test the code and the electric starter, we need in our branch the parameters (GEN_AUTO_START)  to set 0 (Manual)

> I have another PR in the making, we can take the discussion there.

Could you share the link for this PR?

> Thank you for drafting the instructions, probably the best course of action is to create a Wiki page and put it there. More on that later, after we're happy with the functionality of the code.

Let me know how we can do the Wiki and where is it.

Thanks again @Georacer , for your help.

PR #29588, opened 2025-03-22
4.6 Backport of quadplane system ID feature

bnsgeyer:
This PR backports the quadplane system ID feature to 4.6


rmackay9:
I'm planning on including this is 4.6.2-beta2 although I'm unsure of the outcome of this morning's discussion with @tridge on whether we wanted to include it or not

rmackay9:
After a discussion with @bnsgeyer, he's reminded me that there were no objections to including it so in it goes!

PR #29589, opened 2025-03-22
Add logging of data coming through uarts

peterbarker:
Roughly based on the raw GPS logging.

Notable difference is that the packing into chunks is done in the caller, not the thread loop.  This is because we log which direction the data travelled in, and we don't want to run two buffers!

Second notable difference is that the thread moves through a linked list of allocated doing-raw-logging structures, which the backends may allocate based on parameters.

Also includes a utility which translates from one of these capture files into a .tlog.

Tested in SITL and on Durandal.

```
CubeOrange-periph-heavy  -24               *                                                   
CubeRedPrimary                      -48    *           -80     -56               -80    -24    -136
Durandal                            -24    *           -24     -16               -32    -48    40
Hitec-Airspeed           0                 *                                                   
KakuteH7-bdshot                     -16    *           -8      -8                -16    -16    -8
MatekF405                           8      *           8       8                 8      8      8
OMNIBUSF7V2                         -8     *           -8      -8                -8     -8     -8
Pixhawk1-1M-bdshot                  8                  8       8                 8      8      8
f103-QiotekPeriph        0                 *                                                   
f303-Universal           -8                *                                                   
iomcu                                                                0                         
revo-mini                           8      *           8       8                 8      8      8
skyviper-journey                                       0                                       
skyviper-v2450                                         8                                       
```

... only the fix for the _last_options has effect on compiler output, when testing this PR vs just those two commits.

When the feature is enabled:
```
pbarker@threads:~$ ~/rc/ardupilot/Tools/scripts/filter_size_compare_branches_csv.py --hide /tmp/some.csv  
-----------------------  ---------  -----  ----------  ------  ----  ----------  -----  -----  ---
CubeOrange-periph-heavy  584               *
CubeRedPrimary                      632    *           624     632               712    592    512
Durandal                            648    *           664     640               640    656    656
Hitec-Airspeed           4560              *
KakuteH7-bdshot                     536    *           536     536               536    536    536
MatekF405                           552    *           552     552               552    552    552
OMNIBUSF7V2                         536    *           536     536               536    536    536
Pixhawk1-1M-bdshot                  552                552     552               552    552    552
f103-QiotekPeriph        5740              *
f303-Universal           5564              *
iomcu                                                                0
revo-mini                           576    *           576     576               576    576    584
skyviper-journey                                       584
skyviper-v2450                                         584
```

### DIFF
--- a/libraries/AP_HAL/UARTDriver.cpp
+++ b/libraries/AP_HAL/UARTDriver.cpp
@@ -4,6 +4,12 @@
 #include "AP_HAL.h"
 #include <AP_Logger/AP_Logger.h>
 
+#if HAL_UART_DEBUG_LOGGING_ENABLED
+#include <AP_Filesystem/AP_Filesystem.h>
+#endif // HAL_UART_DEBUG_LOGGING_ENABLED
+
+extern const AP_HAL::HAL& hal;
+
 void AP_HAL::UARTDriver::begin(uint32_t baud, uint16_t rxSpace, uint16_t txSpace)
 {
     if (lock_write_key != 0) {
@@ -226,3 +232,128 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
 }
 #endif // HAL_LOGGING_ENABLED
 #endif // HAL_UART_STATS_ENABLED
+
+
+/*
+ * Support for logging all incoming/outgoing data from a serial port
+ */
+
+#if HAL_UART_DEBUG_LOGGING_ENABLED
+
+/*
+  log some data for debugging
+*/
+void AP_HAL::UARTDriver::log_read_data(const uint8_t *data, uint16_t length)
+{
+    log_data(data, length, false);
+}
+void AP_HAL::UARTDriver::log_written_data(const uint8_t *data, uint16_t length)
+{
+    log_data(data, length, true);
+}
+void AP_HAL::UARTDriver::log_data(const uint8_t *data, uint16_t length, bool is_written_data)
+{
+    // first see if we actually do want to log:
+    if ((_last_options & OPTION_LOGGING) == 0) {
+        return;
+    }
+
+    // find logging instance corresponding to this UART:
+    if (loginfo == nullptr) {
+        loginfo = NEW_NOTHROW LogInfo();
+        if (loginfo == nullptr) {
+            return;
+        }
+        loginfo->instance = instance;
+        loginfo->next = backend_loginfos;
+        backend_loginfos = loginfo;
+    }
+    if (!log_thread_created) {
+        log_thread_created = true;
+        hal.scheduler->thread_create(FUNCTOR_BIND_MEMBER(&AP_HAL::UARTDriver::logging_start, void), "uart_log", 4096, AP_HAL::Scheduler::PRIORITY_IO, 0);
+    }
+
+    // write packet into buffer:
+    struct PACKED Header {
+        enum class Flag {
+            IS_WRITTEN_DATA = 1,
+        };
+        uint32_t magic = 0x7fe53b04U;
+        uint32_t time_ms;
+        uint16_t length;
+        uint8_t flags;
+    };
+
+    // note that we are assuming we can do this writing atomically.
+    // If there are multiple writers to the serial port that won't be
+    // the case.  We could reserve, write to and commit an iovec
+    // instead of this:
+
+    const auto space_required = sizeof(Header) + length;
+
+    if (loginfo->buf.space() < space_required) {
+        return;
+    }
+
+    Header header;
+    header.time_ms = AP_HAL::millis();
+    header.length = length;
+    if (is_written_data) {
+        header.flags |= uint8_t(Header::Flag::IS_WRITTEN_DATA);
+    }
+
+    loginfo->buf.write((uint8_t*)&header, sizeof(header));
+    loginfo->buf.write(data, length);
+}
+
+// a linked list of backends which are currently logging:
+AP_HAL::UARTDriver::LogInfo *AP_HAL::UARTDriver::backend_loginfos;
+bool AP_HAL::UARTDriver::log_thread_created;
+
+// logging loop, needs to be static to allow for re-alloc of GPS backends
+void AP_HAL::UARTDriver::logging_loop(void)
+{
+    while (true) {
+        hal.scheduler->delay(10);
+        for (auto *loginfo_ptr = backend_loginfos; loginfo_ptr != nullptr; loginfo_ptr = loginfo_ptr->next) {
+            auto &loginfo = *loginfo_ptr;
+            if (loginfo.fd == -2) {
+                // previously failed; ignore.
+                continue;
+            }
+            // check we have a log file open:
+            if (loginfo.fd == -1) {
+                char fname[] = "uartN_XXX.log";
+                fname[4] = '0' + loginfo.instance;
+                uint32_t lognum;
+                for (lognum=0; lognum<1000; lognum++) {
+                    struct stat st;
+                    hal.util->snprintf(&fname[6], 8, "%03u.log", (unsigned)lognum);
+                    if (AP::FS().stat(fname, &st) != 0) {
+                        break;
+                    }
+                }
+                loginfo.fd = AP::FS().open(fname, O_WRONLY|O_CREAT|O_APPEND);
+                if (loginfo.fd == -1) {
+                    // convert into a permanent failure:
+                    loginfo.fd = -2;
+                }
+            }
+            uint32_t n = 0;
+            const uint8_t *p;
+            while ((p = loginfo.buf.readptr(n)) != nullptr && n != 0) {
+                // short writes are unlikely and are ignored (only FS full errors)
+                AP::FS().write(loginfo.fd, p, n);
+                loginfo.buf.advance(n);
+                AP::FS().fsync(loginfo.fd);
+            }
+        }
+    }
+}
+
+// logging thread start, needs to be non-static for thread_create
+void AP_HAL::UARTDriver::logging_start(void)
+{
+    logging_loop();
+}
+#endif // HAL_UART_DEBUG_LOGGING_ENABLED
--- a/libraries/AP_HAL/UARTDriver.h
+++ b/libraries/AP_HAL/UARTDriver.h
@@ -263,9 +263,11 @@ protected:
     virtual uint32_t get_total_dropped_rx_bytes() const { return 0; }
 #endif
 
-private:
     // option bits for port
     uint16_t _last_options;
+
+private:
+    uint8_t index;
 
 #if AP_UART_MONITOR_ENABLED
     ByteBuffer *_monitor_read_buffer;
--- a/libraries/AP_HAL/UARTDriver.h
+++ b/libraries/AP_HAL/UARTDriver.h
@@ -5,6 +5,7 @@
 #include "AP_HAL_Namespace.h"
 #include "utility/BetterStream.h"
 #include <AP_Logger/AP_Logger_config.h>
+#include <AP_HAL/utility/RingBuffer.h>
 
 #ifndef HAL_UART_STATS_ENABLED
 #define HAL_UART_STATS_ENABLED AP_HAL_UARTDRIVER_ENABLED
@@ -13,6 +14,10 @@
 #ifndef AP_UART_MONITOR_ENABLED
 #define AP_UART_MONITOR_ENABLED 0
 #endif
+
+#ifndef HAL_UART_DEBUG_LOGGING_ENABLED
+#define HAL_UART_DEBUG_LOGGING_ENABLED 0
+#endif  // HAL_UART_DEBUG_LOGGING_ENABLED
 
 class ExpandingString;
 class ByteBuffer;
@@ -41,6 +46,10 @@ public:
       be used
      */
     void begin_locked(uint32_t baud, uint16_t rxSpace, uint16_t txSpace, uint32_t write_key);
+
+#if HAL_UART_DEBUG_LOGGING_ENABLED
+    void set_instance(uint8_t _instance) { instance = _instance; }
+#endif  // HAL_UART_DEBUG_LOGGING_ENABLED
 
     /*
       single and multi-byte write methods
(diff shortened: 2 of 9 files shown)


libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,128 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+            return;
+        }
+        loginfo->instance = instance;
+        loginfo->next = backend_loginfos;
+        backend_loginfos = loginfo;
+    }
+    if (!log_thread_created) {
+        log_thread_created = true;
+        hal.scheduler->thread_create(FUNCTOR_BIND_MEMBER(&AP_HAL::UARTDriver::logging_start, void), "uart_log", 4096, AP_HAL::Scheduler::PRIORITY_IO, 0);
+    }
```
tpwrules, line 274: Only need to run if `loginfo == nullptr`.

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,128 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+
+    // write packet into buffer:
+    struct PACKED Header {
+        enum class Flag {
+            IS_WRITTEN_DATA = 1,
+        };
+        uint32_t magic = 0x7fe53b04U;
```
tpwrules, line 281: The format is different to the files written by AP_GPS so the magic should be at least a bit different. Probably just `0x7fe53b05U` is fine.

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,128 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+        header.flags |= uint8_t(Header::Flag::IS_WRITTEN_DATA);
+    }
+
+    loginfo->buf.write((uint8_t*)&header, sizeof(header));
+    loginfo->buf.write(data, length);
+}
+
+// a linked list of backends which are currently logging:
+AP_HAL::UARTDriver::LogInfo *AP_HAL::UARTDriver::backend_loginfos;
+bool AP_HAL::UARTDriver::log_thread_created;
```
tpwrules, line 311: Needs a mutex to protect these. The changes in the logging loop will be a bit subtle.

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,128 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+                    struct stat st;
+                    hal.util->snprintf(&fname[6], 8, "%03u.log", (unsigned)lognum);
+                    if (AP::FS().stat(fname, &st) != 0) {
+                        break;
+                    }
+                }
+                loginfo.fd = AP::FS().open(fname, O_WRONLY|O_CREAT|O_APPEND);
+                if (loginfo.fd == -1) {
+                    // convert into a permanent failure:
+                    loginfo.fd = -2;
```
tpwrules, line 339:
```suggestion
                    loginfo.fd = -2;
                    continue;
```

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,128 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+                }
+            }
+            uint32_t n = 0;
+            const uint8_t *p;
+            while ((p = loginfo.buf.readptr(n)) != nullptr && n != 0) {
+                // short writes are unlikely and are ignored (only FS full errors)
+                AP::FS().write(loginfo.fd, p, n);
+                loginfo.buf.advance(n);
+                AP::FS().fsync(loginfo.fd);
```
tpwrules, line 348: Fsync should be outside the loop. Also will generally be bad for performance but is probably necessary...

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,128 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+
+    loginfo->buf.write((uint8_t*)&header, sizeof(header));
+    loginfo->buf.write(data, length);
+}
+
+// a linked list of backends which are currently logging:
+AP_HAL::UARTDriver::LogInfo *AP_HAL::UARTDriver::backend_loginfos;
+bool AP_HAL::UARTDriver::log_thread_created;
+
+// logging loop, needs to be static to allow for re-alloc of GPS backends
```
tpwrules, line 313: Is the GPS backend statement still accurate?

libraries/AP_HAL/UARTDriver.h
```diff
@@ -42,6 +47,10 @@ class AP_HAL::UARTDriver : public AP_HAL::BetterStream {
      */
     void begin_locked(uint32_t baud, uint16_t rxSpace, uint16_t txSpace, uint32_t write_key);
 
+#if HAL_UART_DEBUG_LOGGING_ENABLED
+    void set_instance(uint8_t _instance) { instance = _instance; }
+#endif  // HAL_UART_DEBUG_LOGGING_ENABLED
```
tpwrules, line 52: How does this work for e.g. network and CAN ports? Instance should be > 9, but it's unclear if they will even be set up properly or the functions called. Does anything bad happen if you try?

libraries/AP_HAL/UARTDriver.h
```diff
@@ -263,11 +273,38 @@ class AP_HAL::UARTDriver : public AP_HAL::BetterStream {
     virtual uint32_t get_total_dropped_rx_bytes() const { return 0; }
 #endif
 
-private:
+#if HAL_UART_DEBUG_LOGGING_ENABLED
+    void log_read_data(const uint8_t *data, uint16_t length);
+    void log_written_data(const uint8_t *data, uint16_t length);
+#endif  // HAL_UART_DEBUG_LOGGING_ENABLED
```
tpwrules, line 279: Could put method bodies here since they are so simple.

libraries/AP_HAL_ChibiOS/UARTDriver.cpp
```diff
@@ -763,6 +763,10 @@ ssize_t UARTDriver::_read(uint8_t *buffer, uint16_t count)
         update_rts_line();
     }
 
+#if HAL_UART_DEBUG_LOGGING_ENABLED
+    log_read_data(buffer, count);
+#endif
```
tpwrules, line 768: Have you noticed any problem for our drivers which read 1 byte at a time? That will result in a lot of overhead. Perhaps the logging thread could do some sort of linearization?

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,128 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+                char fname[] = "uartN_XXX.log";
+                fname[4] = '0' + loginfo.instance;
+                uint32_t lognum;
+                for (lognum=0; lognum<1000; lognum++) {
+                    struct stat st;
+                    hal.util->snprintf(&fname[6], 8, "%03u.log", (unsigned)lognum);
+                    if (AP::FS().stat(fname, &st) != 0) {
+                        break;
+                    }
+                }
```
tpwrules, line 335: Would be more straightforward to just snprintf the whole thing.

libraries/AP_HAL/UARTDriver.h
```diff
@@ -100,6 +109,7 @@ class AP_HAL::UARTDriver : public AP_HAL::BetterStream {
         OPTION_MAVLINK_NO_FORWARD = (1U<<10), // don't forward MAVLink data to or from this device
         OPTION_NOFIFO             = (1U<<11), // disable hardware FIFO
         OPTION_NOSTREAMOVERRIDE   = (1U<<12), // don't allow GCS to override streamrates
+        OPTION_LOGGING            = (1U<<13), // enable raw logging
```
tpwrules, line 112: Needs param metadata update?

tpwrules (changes requested):
I like the idea but I think the code needs to move a little further from the stuff in AP_GPS to fix error cases and race conditions.

What is the calculus on writing into a separate file vs a dataflash log message? I think the latter could be made simpler but it would require a postprocessing script. Do we remember why a separate file was chosen for the GPS logging?

peterbarker (commented):
I've added a tool to convert from the new format back into the SITL GPS format.

Tools/autotest/rover.py
```diff
@@ -6886,6 +6887,52 @@ def REQUIRE_LOCATION_FOR_ARMING(self):
...
+        self.set_parameter("SERIAL1_OPTIONS", 8192)
+        self.reboot_sitl()
+        if os.path.exists(expected_log_filepath):
+            raise NotAchievedException("Log started to exist when it shouldn't")
+        mavproxy = self.start_mavproxy()
+        self.delay_sim_time(10)
+        self.stop_mavproxy(mavproxy)
+
+        if not os.path.exists(expected_log_filepath):
+            raise NotAchievedException(f"{expected_log_filepath} does not exist")
```
tpwrules, line 6904: Isn't all this racey?

libraries/AP_HAL/UARTDriver.cpp
tpwrules: You don't need a semaphore for one variable, but there's an implied relation with the struct it points to which matters on non-embedded platforms.

I don't see how it would be benign, any thread could have started the log thread and so we'd end up with multiple randomly writing data to different files.

Thanks for adding it.

libraries/AP_HAL_ChibiOS/UARTDriver.cpp
tpwrules: To start with we would coalesce within the same millisecond and direction, that wouldn't create a meaningful change. But it could be deferred.

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,124 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+bool AP_HAL::UARTDriver::log_thread_created;
+
+// logging loop, needs to be static as it loops over backends created
+// by all all backends
+void AP_HAL::UARTDriver::logging_loop(void)
+{
+    while (true) {
+        hal.scheduler->delay(10);
+        for (auto *loginfo_ptr = backend_loginfos; loginfo_ptr != nullptr; loginfo_ptr = loginfo_ptr->next) {
+            auto &loginfo = *loginfo_ptr;
```
tpwrules, line 316:
```suggestion
        hal.scheduler->delay(10);
        AP_HAL::UARTDriver::LogInfo *loginfo_ptr;
        {
            WITH_SEMAPHORE(loginfo_creation_sem);
            loginfo_ptr = backend_loginfos;
        }
        for (; loginfo_ptr != nullptr; loginfo_ptr = loginfo_ptr->next) {
            auto &loginfo = *loginfo_ptr;
```

This is the correct way to do it for non-embedded platforms due to the relation with what the struct points to.

tpwrules (commented):
A couple more minor things, and I should probably test before officially approving...

Would still like your thoughts on why not to put this in the dataflash log. We could likely move it later though, I'm assuming we will characterize this as strictly a debugging feature.

Tools/autotest/rover.py
peterbarker: Yeah, any time we use MAVProxy things get a bit racy.  I've been generous with the delays :-)

libraries/AP_HAL/UARTDriver.cpp
peterbarker: What exactly are you trying to protect against here?

tpwrules: There's no guarantee that another thread will see the contents of the loginfo struct before it sees the pointer to it. It's rather unlikely here but I think ArduPilot is a bit sleazy with locking and atomics in general...

Tools/autotest/rover.py
tpwrules: Still, you set to enable the logging and then hope the vehicle reboots before the log gets created.

libraries/AP_HAL_ChibiOS/UARTDriver.cpp
```diff
@@ -773,6 +777,10 @@ size_t UARTDriver::_write(const uint8_t *buffer, size_t size)
 		return 0;
 	}
 
+#if HAL_UART_DEBUG_LOGGING_ENABLED
+    log_written_data(buffer, size);
```
tridge, line 781: I think these can in in the AP_HAL code

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,124 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+    }
+
+    // find logging instance corresponding to this UART:
+    if (loginfo == nullptr) {
+        loginfo = NEW_NOTHROW LogInfo();
+        if (loginfo == nullptr) {
+            return;
+        }
+
+        WITH_SEMAPHORE(loginfo_creation_sem);
```
tridge, line 260: needs to be above allocation

libraries/AP_HAL/UARTDriver.cpp
```diff
@@ -226,3 +232,124 @@ void AP_HAL::UARTDriver::log_stats(const uint8_t inst, StatsTracker &stats, cons
...
+                }
+            }
+            uint32_t n = 0;
+            const uint8_t *p;
+            while ((p = loginfo.buf.readptr(n)) != nullptr && n != 0) {
+                // short writes are unlikely and are ignored (only FS full errors)
+                AP::FS().write(loginfo.fd, p, n);
+                loginfo.buf.advance(n);
+            }
+            AP::FS().fsync(loginfo.fd);
```
tridge, line 345: fsync once a second?

tridge (changes requested):

tridge:
we should not have a write to the filesystem per write() or read() call

libraries/AP_HAL/UARTDriver.cpp
peterbarker: > There's no guarantee that another thread will see the contents of the loginfo struct before it sees the pointer to it. It's rather unlikely here but I think ArduPilot is a bit sleazy with locking and atomics in general and I would like to see it done better...

I can't see how the pointer is exposed to the other threads before the guts have been initialised....

And just generally, any time I need more nightmares on race conditions I go back to and read Paul McKenny's paper.  So I'm covered there.

peterbarker:
Hmmm.

I now recall the last time I tried to do this I put the code into the ByteBuffer code.

The key reason being that you're logging bytes *available* to the API (and associated timestamp), not the timestamp the API actually gets around to reading them.

PR #29597, opened 2025-03-23
Re-init uart with configured baudrate

DrTobe:
In `AP_MSP_Telem_Backend.cpp`, the uart is re-initialized with baudrate 0. I found this when I was investigating issue https://discuss.ardupilot.org/t/configuring-serial-port-as-msp-fails-on-linux-speed-0-baud/131984. In my case, the change in this PR fixes the MSP communication from the `AP_Periph` board to my Linux-based board.

### DIFF
--- a/libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
+++ b/libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
@@ -87,7 +87,8 @@ bool AP_MSP_Telem_Backend::init_uart()
 {
     if (_msp_port.uart != nullptr)  {
         // re-init port here for use in this thread
-        _msp_port.uart->begin(0);
+        const auto baudrate = _msp_port.uart->get_baud_rate();
+        _msp_port.uart->begin(baudrate);
         return true;
     }
     return false;


rmackay9:
Hi @DrTobe,

Thanks for this!  Could you update the title of the commit to start with "AP_MSP: "?  We do this on all our commits because it helps ease backporting to be able to immediately see what subsystem a change affects.  .. just to be clear, I mean the commit title, not the PR title.

andyp1per:
The semantics of begin are that if you pass 0 you get the previously configured board rate - so this is not a bug in the driver per-se. It's possible this is a bug in the linux hal.

andyp1per (dismissed):
If it is a bug it needs a different fix.

DrTobe:
I have found and fixed the issue in `AP_HAL_Linux/UARTDriver.cpp` and squashed the commits. With these changes, I can run this on my Linux-based board successfully, too.

andyp1per (commented):
I don't know enough about the Linux Hal to know whether this is the correct fix, but it looks good to me.

peterbarker:
@Williangalvani your view on this fix?

libraries/AP_HAL_Linux/UARTDriver.cpp
```diff
@@ -78,14 +78,11 @@ void UARTDriver::_begin(uint32_t b, uint16_t rxS, uint16_t txS)
...
-    _device->set_speed(b);
-
     bool clear_buffers = false;
-    if (b != 0) {
-        if (_baudrate != b && hal.console != this) {
-            clear_buffers = true;
-        }
+    if (b != 0 && _baudrate != b && hal.console != this) {
+        _device->set_speed(b);
         _baudrate = b;
```
Williangalvani, line 84: It seems that after this patch,  _baudrate is only set when `hal.console != this`, while it was set regardless in the previous code. could that be an issue?

From a quick glimpse at the code this seems like a harmless change. Thoughts @peterbarker ? 

Other than that this looks good to me.

libraries/AP_HAL_Linux/UARTDriver.cpp
```diff
@@ -78,14 +78,11 @@ void UARTDriver::_begin(uint32_t b, uint16_t rxS, uint16_t txS)
...
     while (_in_timer) hal.scheduler->delay(1);
 
-    _device->set_speed(b);
-
     bool clear_buffers = false;
-    if (b != 0) {
-        if (_baudrate != b && hal.console != this) {
-            clear_buffers = true;
-        }
+    if (b != 0 && _baudrate != b && hal.console != this) {
```
tridge, line 82: hal.console could be a real serial device and often is one on linux, so we do need to call set_speed() when b != 0 on the console

tridge (changes requested):

tridge:
@DrTobe nice find btw!

libraries/AP_HAL_Linux/UARTDriver.cpp
DrTobe: So this should be something like the following?
```cpp
    if (b != 0 && _baudrate != b) {
        _device->set_speed(b);
        _baudrate = b;
        if (hal.console != this) {
            clear_buffers = true;
        }
    }
```

DrTobe:
I have adjusted the fix as suggested above. I have tested it for my board and it still works.

Sorry for the delay though, I have been very busy the past weeks.

tridge (approved):

rmackay9:

PR #29598, opened 2025-03-24
AP_RCTelemetry: Fix for DMA detection issue in UARTDriver for Linux

massout:
- Since the `is_dma_enabled()` function, which is defined as a virtual function within the `AP_HAL::UARTDriver` class in `"libraries/AP_HAL/UARTDriver.h"`, is not implemented in `"libraries/AP_HAL_Linux/UARTDriver.cpp"` for Linux and its default initialization is set to `false`, Raspberry Pi encounters DMA errors in CRSF protocol and relevant Telemetries.
- To prevent this error, necessary modifications were made in `"libraries/AP_RCTelemetry/AP_RCTelem_CRSF.cpp"`, `"libraries/AP_RCTelemetry/AP_RCTelem_GHST.cpp"`, and `"libraries/AP_RCProtocol/AP_RCProtocol_CRSF.cpp"`. 
- Since all UART ports on Raspberry Pi series support DMA, there was no need to implement the `is_dma_enabled()` function in `"libraries/AP_HAL_Linux/UARTDriver.cpp"`. However, it may be required for other Linux boards. Which have no DMA at all UART ports.
- Addition was made to the preprocessor line at the beginning. However, since there was a different condition inside the preprocessor block's `if` statement, it was placed inside the `if` block.

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -944,7 +944,7 @@ bool Copter::get_wp_bearing_deg(float &bearing) const
 bool Copter::get_wp_crosstrack_error_m(float &xtrack_error) const
 {
     // see GCS_MAVLINK_Copter::send_nav_controller_output()
-    xtrack_error = flightmode->crosstrack_error() * 0.01;
+    xtrack_error = flightmode->crosstrack_error_m() * 0.01;
     return true;
 }
 
--- a/ArduCopter/GCS_MAVLink_Copter.cpp
+++ b/ArduCopter/GCS_MAVLink_Copter.cpp
@@ -217,9 +217,9 @@ void GCS_MAVLINK_Copter::send_nav_controller_output() const
         degrees(targets_rad.z),
         flightmode->wp_bearing_deg(),
         MIN(flightmode->wp_distance_m(), UINT16_MAX),
-        copter.pos_control->get_pos_error_U_cm() * 1.0e-2f,
+        copter.pos_control->get_pos_error_U_cm() * 0.01,
         0,
-        flightmode->crosstrack_error() * 1.0e-2f);
+        flightmode->crosstrack_error_m() * 0.01);
 }
 
 float GCS_MAVLINK_Copter::vfr_hud_airspeed() const
--- a/ArduCopter/mode.h
+++ b/ArduCopter/mode.h
@@ -170,7 +170,7 @@ public:
     virtual bool get_wp(Location &loc) const { return false; };
     virtual float wp_bearing_deg() const { return 0; }
     virtual float wp_distance_m() const { return 0.0f; }
-    virtual float crosstrack_error() const { return 0.0f;}
+    virtual float crosstrack_error_m() const { return 0.0f;}
 
     // functions to support MAV_CMD_DO_CHANGE_SPEED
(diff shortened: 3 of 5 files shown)


rmackay9:
Hi @massout,

thanks for this!  Could you squash the two commits together and update the commit title to start with "AP_RCProtocol_CRSF: "?  We do this to make backporting easier.  To be clear I mean the commit title, not the PR title.

libraries/AP_RCProtocol/AP_RCProtocol_CRSF.cpp
```diff
@@ -675,8 +675,8 @@ bool AP_RCProtocol_CRSF::change_baud_rate(uint32_t baudrate)
     if (uart == nullptr) {
         return false;
     }
-#if !defined(STM32H7)
-    if (baudrate > get_bootstrap_baud_rate() && !uart->is_dma_enabled()) {
+#if !defined(STM32H7) 
+    if (baudrate > get_bootstrap_baud_rate() && ((CONFIG_HAL_BOARD == HAL_BOARD_CHIBIOS) && !uart->is_dma_enabled())) {
```
andyp1per, line 679: this should go in a #if section

libraries/AP_RCTelemetry/AP_CRSF_Telem.cpp
```diff
@@ -216,7 +216,7 @@ bool AP_CRSF_Telem::process_rf_mode_changes()
     }
 #if !defined (STM32H7)
     // warn the user if their setup is sub-optimal, H7 does not need DMA on serial port
-    if (_telem_bootstrap_msg_pending && !uart->is_dma_enabled()) {
+    if (_telem_bootstrap_msg_pending && ((CONFIG_HAL_BOARD == HAL_BOARD_CHIBIOS) && !uart->is_dma_enabled())) {
```
andyp1per, line 219: ditto

libraries/AP_RCTelemetry/AP_GHST_Telem.cpp
```diff
@@ -125,7 +125,7 @@ bool AP_GHST_Telem::process_rf_mode_changes()
     }
 #if !defined (STM32H7)
     // warn the user if their setup is sub-optimal, H7 does not need DMA on serial port
-    if (_telem_bootstrap_msg_pending && !uart->is_dma_enabled()) {
+    if (_telem_bootstrap_msg_pending && ((CONFIG_HAL_BOARD == HAL_BOARD_CHIBIOS) && !uart->is_dma_enabled())) {
```
andyp1per, line 128: ditto

andyp1per (changes requested):
Just some style/formatting changes required

massout (commented):
As I mentioned, all UART outputs with fixes should be DMA-supported for Linux devices; a more comprehensive solution should be found for devices that do not meet this criterion.

andyp1per (changes requested):
Sorry I was not clear. Like this:

```
#if !defined(STM32H7) && (CONFIG_HAL_BOARD == HAL_BOARD_CHIBIOS)
    if (baudrate > get_bootstrap_baud_rate() && !uart->is_dma_enabled())) {
        return false;
    }
#endif
```

PR #29599, opened 2025-03-24
AP_Mount: XFRobot driver

rmackay9:
This adds basic support for the [XFRobot serial of camera gimbals ](https://www.allxianfei.com/en/uav-payloads/) including the Z-1 mini which was used to bench test this driver.

To test this driver:

1. Build and install the firmware on an autopilot with at least 2MB of flash
2. Set the following parameters:
    - MNT1_TYPE = 14
    - CAM1_TYPE = 4
    - SERIALx_PROTOCOL = 8
    - SERIALx_BAUD = 115 (these baud rates will also likely work: 250000, 500000 and 1000000)
4. connect the camera gimbal to the autopilot as shown below.  This uses the SERIAL connection but ethernet should also work (but this hasn't been tested yet)
xfrobot-z1mini-autopilot-serial

This has been tested on real hardware including the following:

1. ensured roll, pitch and yaw control worked including both body-frame and earth-frame angles and rates
2. tested camera functions including take picture, record video, zoom.  "focus" also reported success but it was difficult to be sure if it actually worked
3. real-time video streaming over ethernet using VLC and video URL of rtsp://192.168.144.108
4. checked the vehicle location appeared correctly on the OSD and in the EXIF of pictures

Corresponding wiki PR is https://github.com/ArduPilot/ardupilot_wiki/pull/6716

image


libraries/AP_Mount/AP_Mount.cpp
```diff
@@ -178,6 +179,15 @@ void AP_Mount::init()
             serial_instance++;
             break;
 #endif // HAL_MOUNT_CADDX_ENABLED
+
+#if HAL_MOUNT_XFROBOT_ENABLED
+        // check for XFRobot gimbal
+        case Type::XFRobot:
+            _backends[instance] = NEW_NOTHROW AP_Mount_XFRobot(*this, _params[instance], instance, serial_instance);
```
tridge, line 186: i think the num_instances shouldn't increment on alloc failure, pattern is to do it outside of the switch

tridge (changes requested):

tridge:
need to save some flash space, this will overflow

libraries/AP_Mount/AP_Mount_XFRobot.cpp
```diff
@@ -0,0 +1,594 @@
...
+void AP_Mount_XFRobot::process_packet()
+{
+    // sanity check size of packet received
+    if (msg_buff_len < sizeof(GCUSimpleReply)) {
+        return;
+    }
+
+    // copy msg_buff to receive packet for easier access
+    GCUSimpleReply recv_packet;
+    memcpy(recv_packet.bytes, msg_buff, sizeof(recv_packet));
```
peterbarker, line 376: Better done as a union.

libraries/AP_Mount/AP_Mount_XFRobot.cpp
```diff
@@ -0,0 +1,594 @@
...
+    // byte 14~15: absolute pitch angle of vehicle (int16, -9000 ~ +9000)
+    // byte 16~17: absolute yaw angle of vehicle (uint16, 0 ~ 36000)
+    set_attitude_packet.content.main.roll_abs = htole16(constrain_int16(degrees(AP::ahrs().get_roll()) * 100, -18000, 18000));
+    set_attitude_packet.content.main.pitch_abs = htole16(constrain_int16(degrees(AP::ahrs().get_pitch()) * 100, -9000, 9000));
+    set_attitude_packet.content.main.yaw_abs = htole16(constrain_int16(degrees(wrap_PI(AP::ahrs().get_yaw())) * 100, -18000, 18000));
+
+    // byte 18~19: North acceleration of vehicle (int16, cm/s/s)
+    // byte 20~21: East acceleration of vehicle (int16, cm/s/s)
+    // byte 22~23: Upward acceleration of vehicle (int16, cm/s/s)
+    Vector3f accel_ef = AP::ahrs().get_accel_ef();
```
peterbarker, line 450:
```suggestion
    const Vector3f &accel_ef = AP::ahrs().get_accel_ef();
```

rmackay9: thanks for the review, changed to const reference

libraries/AP_Mount/AP_Mount_XFRobot.cpp
peterbarker:
```
--- a/libraries/AP_Mount/AP_Mount_XFRobot.cpp
+++ b/libraries/AP_Mount/AP_Mount_XFRobot.cpp
@@ -371,9 +371,8 @@ void AP_Mount_XFRobot::process_packet()
         return;
     }
 
-    // copy msg_buff to receive packet for easier access
-    GCUSimpleReply recv_packet;
-    memcpy(recv_packet.bytes, msg_buff, sizeof(recv_packet));
+    // cast msg_buff to receive packet for easier access
+    const GCUSimpleReply &recv_packet = *((GCUSimpleReply*)&msg_buff);
 
     // extract latest attitude
     attitude_latest = {
```

... but better would be to not have `msg_buffer` at all.  Instead have a `GCUSimpleReply recv_packet;`  member in your XFRobot object, and do all of your parsing inside that object.  That's really just changing all instances of (eg.) `crc16_ccitt(msg_buff, ...` to `crc16_ccitt(recv_packet.bytes, ...`

rmackay9: Hi @peterbarker,

OK, I understand now.  I think I've done what you've suggested.  The incoming byte buffer (aka "msg_buf") is now a union with the GCUSimpleReply so bytes are placed in the buffer and then when we go to extract items we use the structure.

peterbarker (dismissed):

libraries/AP_Mount/AP_Mount_config.h
```diff
@@ -58,6 +58,10 @@
 #define HAL_MOUNT_CADDX_ENABLED AP_MOUNT_BACKEND_DEFAULT_ENABLED && HAL_PROGRAM_SIZE_LIMIT_KB > 1024
 #endif
 
+#ifndef HAL_MOUNT_XFROBOT_ENABLED
+#define HAL_MOUNT_XFROBOT_ENABLED AP_MOUNT_BACKEND_DEFAULT_ENABLED && HAL_PROGRAM_SIZE_LIMIT_KB > 1024
```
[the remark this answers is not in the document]
rmackay9: thanks again for the review.  I've added a commit at the end to only enable the camera gimbal for >2MB boards (which is currently only the CubeRed)

libraries/AP_Mount/AP_Mount_XFRobot.cpp
```diff
@@ -0,0 +1,590 @@
...
+    // sanity check size of packet received
+    if (msg_buff_len < sizeof(GCUSimpleReply)) {
+        return;
+    }
+
+    // extract latest attitude
+    attitude_latest = {
+        .roll_ef_deg = msg_buff.simple_reply.main.roll_abs_cd * 0.01f,
+        .pitch_ef_deg = msg_buff.simple_reply.main.pitch_abs_cd * 0.01f,
+        .yaw_bf_deg = msg_buff.simple_reply.main.angle_z * 0.01f,
```
peterbarker, line 378: I'm *assuming* this one is also in cd

peterbarker (approved):
LGTM.  I'm a bit surprised that we can assume that this is a SimpleReply, but it doesn't look like there's an ID in the packet, so I guess "protocol version" must cover that.

libraries/AP_Mount/AP_Mount_XFRobot.cpp
rmackay9: Hi @peterbarker,

Yes, it's centi-degrees in the datasheet and I've also tested that the angles appear correctly.
image

The reason we use "angle_z" (which is a body-frame yaw) instead of "yaw_abs_cd" is just that that's how we tend to report camera gimbal angles in AP.

rmackay9:
@peterbarker

Yes, the way this gimbal works is that the first 68 bytes are always the same but the 69th byte is the "pod function" (some times called "order") which is essentially a message id.  There are about 26 possible values but all but 7 of them use this "SimpleReply" format.  We don't use or support those 7 message ids yet.

PR #29601, opened 2025-03-24
SCurve: Fix speed and accel checks to include z component when transitioning fast waypoints

lthall:
This addresses the issue in #29509.

### DIFF
--- a/libraries/AP_Math/SCurve.cpp
+++ b/libraries/AP_Math/SCurve.cpp
@@ -30,11 +30,10 @@ extern const AP_HAL::HAL &hal;
 
 #define SEG_INIT                0
 #define SEG_ACCEL_MAX           4
-#define SEG_TURN_IN             7
 #define SEG_ACCEL_END           7
 #define SEG_SPEED_CHANGE_END    14
 #define SEG_CONST               15
-#define SEG_TURN_OUT            15
+#define SEG_DECEL_START         15
 #define SEG_DECEL_END           22
 
 // constructor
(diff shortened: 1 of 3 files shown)


libraries/AP_Math/SCurve.cpp
```diff
@@ -30,7 +30,7 @@ extern const AP_HAL::HAL &hal;
 
 #define SEG_INIT                0
 #define SEG_ACCEL_MAX           4
-#define SEG_TURN_IN             4
+#define SEG_TURN_IN             7
```
MattKear, line 33: This makes so much more sense to me now.  I was confused by this when it = 4.
Not a blocker for this PR as it is a nit pick but in a future PR I think it would be nice to remove the SEG_TURN_IN define and just use the SEG_ACCEL_END define.  We already have the time_turn_in() function so moving to the SEG_ACCEL_END define in that function makes the code easier to follow IMHO.

MattKear (approved):
This is the correct fix and the changes to the if statements make this much easier to follow the logic.

MattKear:
As an aside, I have been working on an S-Curves kinematic web tool (this is how I found the issue in the first place).  I have applied this fix to that tool and it also demonstrates that this PR fixes the problem.

The below plots show a simple mission with 4 WP in a vertical line, aircraft targets are descending through the WPs.  The position is coloured by velocity magnitude.

**Before this PR:**

The velocity can be seen to exceed the max down velocity set in WPNAV_SPEED_DN:
image

Same image zoomed in on the WP to make the vel magnitude clear:
image


**With this PR applied:**

The velocity is constant at WPNAV_SPEED_DN:

image
image

libraries/AP_Math/SCurve.cpp
lthall: I will do it for this PR I think. I was thinking the same thing as I looked at it.

rmackay9 (approved):
LGTM, thanks!

rmackay9:

PR #29604, opened 2025-03-24
Tools: fix building environment on macos

hfahimiasl:
Fix invalid option error in cd command by removing unsupported -q flag.


hfahimiasl:
> Hi @hfahimiasl,
> 
> thanks very much for this. Could you update the commit title to start with "Tools: "? We add the subsystem to all our commits so that we can more easily backport fixes. It makes it easier because we can look at a commit and instantly know what part of the code is affected. To be clear, I mean the commit title, not the PR title.
> 
> Thanks!

Hi @rmackay9 
Thanks for the feedback! I’ve updated the commit title to start with "Tools: " as requested. Let me know if there’s anything else you’d like me to adjust!

PR #29607, opened 2025-03-25
Fix shadowed-variable problem in ChibiOS HAL

peterbarker:
We have two variables named `_last_options`, meaning updates in the base class wouldn't be seen in the ChibiOS HAL.


tpwrules (approved):
Fortunately the only methods that update this in AP_HAL are virtual so there's no actual API difference in this change. Best to not be confusing though.

PR #29608, opened 2025-03-25
Add support for System ID on Aux Switch in Plane

bnsgeyer:
This PR adds the support to run System ID from the Aux Switch in plane.  To this point I have implemented the input, rate, and mixer SID_AXIS for pitch and roll.  Because plane doesn't use a command model, I am thinking that I would just use the same method for the Recovery SID_AXIS inputs.  

I will need help with how to get the log data.  I see in the log file that AETR has the servo outputs and CTUN has the desired and actual attitudes.  The data I am missing is the desired rate and actual rate which I think I can get from PIDP and PIDR for the pitch and roll axes.  Just not sure how I should go about this.

What other modes should be included? Loiter?


ArduPlane/mode_stabilize.cpp
```diff
@@ -9,6 +9,16 @@ void ModeStabilize::update()
 
 void ModeStabilize::run()
 {
+    // note this is actually in deg/s for some SID_AXIS values for yaw
+    Vector3f offset_deg;
+#if AP_PLANE_SYSTEMID_ENABLED
```
tridge, line 14: expand #if

ArduPlane/systemid.cpp
```diff
@@ -155,15 +153,17 @@ void AP_SystemID::stop()
         attitude_offset_deg.zero();
         throttle_offset = 0;
 
-        auto *attitude_control = plane.quadplane.attitude_control;
-        attitude_control->bf_feedforward(restore.att_bf_feedforward);
-        attitude_control->rate_bf_roll_sysid(0);
-        attitude_control->rate_bf_pitch_sysid(0);
-        attitude_control->rate_bf_yaw_sysid(0);
-        plane.quadplane.pos_control->set_xy_control_scale_factor(1);
+        if (plane.quadplane.enabled()) {
```
tridge, line 156: we'll need some ifdef magic if we want users to be able to run sysid on builds with quadplane not compiled in. Not sure it is worth it?

tridge (changes requested):
very quick review, no testing done

ArduPlane/mode_fbwa.cpp
```diff
@@ -38,6 +38,15 @@ void ModeFBWA::update()
...
 {
+    // note this is actually in deg/s for some SID_AXIS values for yaw
+    Vector3f offset_deg;
+#if AP_PLANE_SYSTEMID_ENABLED
+    auto &systemid = plane.g2.systemid;
+    systemid.update();
+    offset_deg = systemid.get_attitude_offset_deg();
+#endif
+    plane.nav_roll_cd += offset_deg.x * 100.0f;
+    plane.nav_pitch_cd += offset_deg.y * 100.0f;
```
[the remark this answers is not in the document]
IamPete1: `supports_fw_sysid` could also check that quadplane assistance is not kicking in. I wonder if we should also do a min speed check, autotune checks airspeed is above min for example. We would have to be a bit more subtle here and stop the sweep rather than just not applying the offset for a few loops.

libraries/APM_Control/AP_RollController.cpp
```diff
@@ -208,8 +208,16 @@ float AP_RollController::get_servo_out(int32_t angle_err, float scaler, bool dis
     } else if (gains.rmax_pos && desired_rate > gains.rmax_pos) {
         desired_rate = gains.rmax_pos;
     }
+    desired_rate += _sysid_ang_vel_body;
```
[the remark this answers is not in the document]
bnsgeyer: I think it is cleaner doing it here.  Not saying this is right but Copter has it implemented this way and I followed that approach since it got merged that way.  Plus it seems fairly ambiguous where to put this for yaw.  It appeared like there was a lot going on that I couldn't follow.

tridge:
@bnsgeyer Peter and I flew a bixler with this PR in FBWA today. Log here: https://uav.tridgell.net/tmp/plane-sysid-bixler-11.bin

libraries/APM_Control/AP_RollController.cpp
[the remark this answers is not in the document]
tridge: @bnsgeyer most users don't have a yaw controller at all, the rudder is just a proportion of aileron (KFF_RDDRMIX). Users flying ACRO or aerobatics may have configured a full yaw controller, but it is pretty rare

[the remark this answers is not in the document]
bnsgeyer: @tridge Are you suggesting we don't allow System ID to run on the Yaw axis?  I think that would be a big shortcoming especially to those trying to identify a model for an aircraft.  It is important to have the yaw axis identified to determine dynamic modes like dutch roll which is a coupling between yaw and roll.  It is important to be able to excite these axes separately.

ArduPlane/systemid.cpp
```diff
@@ -204,42 +204,80 @@ void AP_SystemID::update()
             attitude_offset_deg.z = waveform_sample;
             break;
         case AxisType::RECOVER_ROLL:
-            attitude_offset_deg.x = waveform_sample;
-            attitude_control->bf_feedforward(false);
+            if (plane.quadplane.enabled() && plane.quadplane.in_vtol_mode()) {
+                attitude_offset_deg.x = waveform_sample;
+                attitude_control->bf_feedforward(false);
+            } else {
+
```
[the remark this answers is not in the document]
bnsgeyer: done. Thanks

ArduPlane/systemid.cpp
```diff
@@ -204,42 +204,80 @@ void AP_SystemID::update()
...
+            }
             break;
         case AxisType::RECOVER_PITCH:
-            attitude_offset_deg.y = waveform_sample;
-            attitude_control->bf_feedforward(false);
+            if (plane.quadplane.enabled() && plane.quadplane.in_vtol_mode()) {
+                attitude_offset_deg.y = waveform_sample;
+                attitude_control->bf_feedforward(false);
+            } else {
+
```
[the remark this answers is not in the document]
bnsgeyer: done. thanks.

libraries/APM_Control/AP_RollController.cpp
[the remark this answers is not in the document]
bnsgeyer: @tridge  I like how this is done in the controller.  It guarantees as the main vehicle code changes that the input and the output to the controller is always captured.  I don't have to guess where in the main code this is being called.

ArduPlane/systemid.cpp
```diff
@@ -204,42 +215,100 @@ void AP_SystemID::update()
             attitude_offset_deg.z = waveform_sample;
             break;
         case AxisType::RECOVER_ROLL:
```
IamPete1, line 217: I wonder if we should add new axis for this. It would just make the tools easier as we would not have to check flight mode too.

bnsgeyer: @IamPete1 Not sure what you mean?  for the recover_XXX cases?  please explain further

IamPete1: Just if we should have a `FIXEDWING_RECOVER_ROLL` to differentiate from the existing `RECOVER_ROLL` that is used for the VTOL controllers. We log the axis type in the `SIDD` log although I'm not sure if that log is still needed in this case.

ArduPlane/systemid.cpp
bnsgeyer: Done.

libraries/APM_Control/AP_RollController.cpp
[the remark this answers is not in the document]
bnsgeyer: Moved back to plane codebase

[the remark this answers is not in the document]
bnsgeyer: @tridge realized that by moving this back to the plane codebase, rate inputs can not be accomplished. The input has to be done at the controller level. If that is fine then I will remove that as an option for the SID_AXIS for plane.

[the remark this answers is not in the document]
tridge: hmm, good point, how useful do you think rate input vs attitude input will be? we could add a rate_offset float param to the APIs in APM_Controller, default value 0 if we need it

ArduPlane/Attitude.cpp
```diff
@@ -360,7 +379,16 @@ void Plane::stabilize_yaw()
...
      */
-    const float rudder_output = calc_nav_yaw_coordinated();
+    float rudder_output = calc_nav_yaw_coordinated();
+
+#if AP_PLANE_SYSTEMID_ENABLED
+    if (control_mode->supports_fw_systemid()) {
+        Vector3f offset;
+        auto &systemid = plane.g2.systemid;
+        offset = systemid.get_output_offset();
+        rudder_output += offset.z * 4500.0f;
```
tridge, line 389: for most users their rudder will already receive the aileron signal scaled by KFF_RDDRMIX (default 0.5). So this will mean we are doubling up that offset signal when doing systemID on the roll axis. I think until we think through a good plan for yaw that we not try to include rudder in systemID for fixed wing

ArduPlane/mode.h
```diff
@@ -526,6 +529,11 @@ class ModeStabilize : public Mode
 
     void run() override;
 
+#if AP_PLANE_SYSTEMID_ENABLED
+    // does this mode support fixed wing systemid?
+    bool supports_fw_systemid() const override { return true; }
```
tridge, line 534: I'd recommend not having STABILIZE included, the way it does attitude targets would make systemID a mess. I recommend FBWA, CRUISE and LOITER as the initial set of modes we support fixed wing systemID in

ArduPlane/systemid.cpp
```diff
@@ -298,8 +378,63 @@ void AP_SystemID::log_data() const
...
+                                    "ss-zkkkooooooooo", "F---------------", "Qfffffffffffffff",
+                                    AP_HAL::micros64(),
+                                    waveform_time, waveform_sample, waveform_freq_rads / (2 * M_PI),
+                                    degrees(delta_angle.x * dt_ang_inv),
+                                    degrees(delta_angle.y * dt_ang_inv),
+                                    degrees(delta_angle.z * dt_ang_inv),
+                                    delta_velocity.x * dt_vel_inv,
+                                    delta_velocity.y * dt_vel_inv,
+                                    delta_velocity.z * dt_vel_inv,
+                                    plane.nav_roll_cd / 100.0f,
```
tridge, line 429: minor nitpick, * 0.01 is preferred over / 100.0

tridge (changes requested):

ArduPlane/Attitude.cpp
```diff
@@ -119,8 +119,17 @@ void Plane::stabilize_roll()
         nav_roll_cd += 18000;
         if (ahrs.roll_sensor < 0) nav_roll_cd -= 36000;
     }
+float roll_out = stabilize_roll_get_roll_out();
+
+#if AP_PLANE_SYSTEMID_ENABLED
+    if (control_mode->supports_fw_systemid()) {
+        Vector3f offset;
+        auto &systemid = plane.g2.systemid;
+        offset = systemid.get_output_offset();
```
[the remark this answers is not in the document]
bnsgeyer: This is the mixer input and not the pilot input. That is why I’ve shown it as an output offset rather than the attitude offset. This is meant to bypass all stabilization controllers and make inputs directly to the servo or mixing unit so as to be able to determine stability of the aircraft control system.

ArduPlane/systemid.cpp
```diff
@@ -188,61 +200,124 @@ void AP_SystemID::update()
...
     auto *attitude_control = plane.quadplane.attitude_control;
+#endif
 
     switch (start_axis) {
         case AxisType::NONE:
             // not possible, see start()
             break;
         case AxisType::INPUT_ROLL:
-            attitude_offset_deg.x = waveform_sample;
+            plane.nav_roll_cd += waveform_sample * 100.0f;
```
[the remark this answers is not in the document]
bnsgeyer: I had to apply the offset to the NAV roll and NAV pitch variables here because by doing it in the stabilize methods, the logging would not show the input by the system ID attitude offset.

ArduPlane/Attitude.cpp
bnsgeyer: Not sure why this is an issue. It make yaw inputs directly to the servo. This would be one way of isolating the yaw response.  If you are firm on no including yaw at this point then I will remove it.

ArduPlane/mode.h
bnsgeyer: Not sure what you mean by this. I will have to dig into the code to see what stabilize does.

bnsgeyer:
@tridge @peterbarker i have tested this in SITL in FBWA and stabilize for all of the inputs other than throttle. I think I need help to figure out where that should go. I would appreciate if you could test it IRL.  Thanks

ArduPlane/systemid.cpp
bnsgeyer: @IamPete1 Starting to see the wisdom in your comment now that I am trying to code up the web tool. This would make that much easier

tridge:
@bnsgeyer I've force pushed a fix for the merge conflicts

peterbarker:
This was tested at SV and logs have been sent to Bill for analysis.

ArduPlane/systemid.cpp
bnsgeyer: Made new SID_AXIS values for plane SID

ArduPlane/systemid.cpp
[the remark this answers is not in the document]
bnsgeyer: No longer valid comment

ArduPlane/systemid.cpp
```diff
@@ -188,61 +200,124 @@ void AP_SystemID::update()
...
             break;
         case AxisType::INPUT_YAW:
-            attitude_offset_deg.z = waveform_sample;
-            break;
+            // Not incorporated into plane
+        break;
         case AxisType::RECOVER_ROLL:
-            attitude_offset_deg.x = waveform_sample;
-            attitude_control->bf_feedforward(false);
+            if (plane.control_mode->supports_vtol_systemid()) {
```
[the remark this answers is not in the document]
bnsgeyer: No longer valid as Fixed wing non-VTOL sweeps are broken out into separate SID_AXIS values.

ArduPlane/systemid.cpp
bnsgeyer: fixed. thanks!

ArduPlane/Attitude.cpp
bnsgeyer: Yaw and throttle inputs are removed.  We can address this again after this feature gets implemented.  but It will require another log message to collect the necessary data.

ArduPlane/mode.h
bnsgeyer: Removing stabilize as one of the flight modes for system ID feature.  You were right in that stabilize did not give good data that agreed with the predictions.

ArduPlane/Attitude.cpp
```diff
@@ -123,8 +123,17 @@ void Plane::stabilize_roll()
         nav_roll_cd += 18000;
         if (ahrs.roll_sensor < 0) nav_roll_cd -= 36000;
     }
+float roll_out = stabilize_roll_get_roll_out();
```
IamPete1:
```suggestion
    float roll_out = stabilize_roll_get_roll_out();
```

ArduPlane/Attitude.cpp
```diff
@@ -123,8 +123,17 @@ void Plane::stabilize_roll()
...
+
+#if AP_PLANE_SYSTEMID_ENABLED
+    if (control_mode->supports_fw_systemid()) {
```
IamPete1: It would make me a little happier if this also checked that system ID is actually running. It just makes it clear that is will have no effect if not running. Currently its just assuming the offset will be 0 in that case.

ArduPlane/Attitude.cpp
```diff
@@ -177,7 +186,17 @@ void Plane::stabilize_pitch()
         return;
     }
 
-    const float pitch_out = stabilize_pitch_get_pitch_out();
+    float pitch_out = stabilize_pitch_get_pitch_out();
+
+#if AP_PLANE_SYSTEMID_ENABLED
+    if (control_mode->supports_fw_systemid()) {
```
IamPete1: Per roll, It would make me a little happier if this also checked that system ID is actually running. It just makes it clear that is will have no effect if not running. Currently its just assuming the offset will be 0 in that case.

ArduPlane/mode.h
```diff
@@ -288,6 +291,11 @@ class ModeAuto : public Mode
     bool in_pullup() const { return pullup.in_pullup(); }
 #endif
 
+#if AP_PLANE_SYSTEMID_ENABLED
+    // does this mode support fixed wing systemid?
+    bool supports_fw_systemid() const override { return true; }
```
IamPete1, line 299: Should we also check some auto state here? For example not in VTOL and not taking off ect.

ArduPlane/mode.h
```diff
@@ -373,6 +381,11 @@ class ModeGuided : public Mode
 
     void update_target_altitude() override;
 
+#if AP_PLANE_SYSTEMID_ENABLED
+    // does this mode support fixed wing systemid?
+    bool supports_fw_systemid() const override { return true; }
```
IamPete1, line 389: Again it might be worth checking flight stage here too. We could be in VTOL flight.

ArduPlane/systemid.cpp
```diff
@@ -103,23 +103,54 @@ void AP_SystemID::start()
...
+    // Exits if quadplane or plane attempting to run System ID VTOL SID Axis in fixed wing flight mode
+    if (plane.control_mode->supports_fw_systemid() && (start_axis == AxisType::INPUT_ROLL 
+                                                        || start_axis == AxisType::INPUT_PITCH
+                                                        || start_axis == AxisType::INPUT_YAW
+                                                        || start_axis == AxisType::RECOVER_ROLL 
+                                                        || start_axis == AxisType::RECOVER_PITCH
+                                                        || start_axis == AxisType::RECOVER_YAW
+                                                        || start_axis == AxisType::RATE_ROLL
+                                                        || start_axis == AxisType::RATE_PITCH
+                                                        || start_axis == AxisType::RATE_YAW                                                        || start_axis == AxisType::RATE_ROLL
```
IamPete1:
```suggestion
                                                        || start_axis == AxisType::RATE_YAW
                                                        || start_axis == AxisType::RATE_ROLL
```

ArduPlane/systemid.cpp
```diff
@@ -103,23 +103,54 @@ void AP_SystemID::start()
...
+                                                        || start_axis == AxisType::MIX_ROLL
+                                                        || start_axis == AxisType::MIX_PITCH                                                        || start_axis == AxisType::RATE_ROLL
```
IamPete1:
```suggestion
                                                        || start_axis == AxisType::MIX_PITCH
                                                        || start_axis == AxisType::RATE_ROLL
```

ArduPlane/systemid.cpp
```diff
@@ -103,23 +103,54 @@ void AP_SystemID::start()
...
     }
-    if (!plane.quadplane.enabled()) {
-        gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: only for quadplane");
+    if (!plane.control_mode->supports_fw_systemid() && !plane.control_mode->supports_vtol_systemid()) {
+        gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: Not supported in mode %s", plane.control_mode->name());
         return;
     }
-    if (!plane.control_mode->supports_systemid()) {
-        gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: Not supported in mode %s", plane.control_mode->name());
+    // Exits if quadplane or plane attempting to run System ID VTOL SID Axis in fixed wing flight mode
```
IamPete1: I think this checks would be better done as a switch on start axis.
```

    // check if enabled
    if (!plane.control_mode->supports_fw_systemid() && !plane.control_mode->supports_vtol_systemid()) {
        gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: Not supported in mode %s", plane.control_mode->name());
        return;
    }

    switch (start_axis) {
        case AxisType::NONE:
            gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: No axis selected");
            return;

        case AxisType::INPUT_ROLL:
        case AxisType::INPUT_PITCH:
        case AxisType::INPUT_YAW:
        case AxisType::RECOVER_ROLL:
        case AxisType::RECOVER_PITCH;
        case AxisType::RECOVER_YAW:
        case AxisType::RATE_ROLL:
        case AxisType::RATE_PITCH:
        case AxisType::RATE_YAW:
        case AxisType::MIX_ROLL:
        case AxisType::MIX_PITCH:
        case AxisType::MIX_YAW:
        case AxisType::MIX_THROTTLE: {
            if (!plane.control_mode->supports_vtol_systemid()) {
#if HAL_QUADPLANE_ENABLED
                gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: Axis not supported for this flight mode");
#else
                gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: Axis not supported in Plane");
#endif
                return;
            }
            break;
        }

        case AxisType::FW_INPUT_ROLL:
        case AxisType::FW_INPUT_PITCH:
        case AxisType::FW_MIX_ROLL:
        case AxisType::FW_MIX_PITCH: {
            if (!plane.control_mode->supports_fw_systemid()) {
                gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: Axis not supported for this flight mode");
                return;
            }
            break;
        }
    }

    if (!hal.util->get_soft_armed()) {
        gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: must be armed");
        return;
    }
```

ArduPlane/systemid.cpp
```diff
@@ -103,23 +103,54 @@ void AP_SystemID::start()
...
         gcs().send_text(MAV_SEVERITY_WARNING, "SystemID: must be armed");
         return;
     }
 
     attitude_offset_deg.zero();
     throttle_offset = 0;
 
-    restore.att_bf_feedforward = plane.quadplane.attitude_control->get_bf_feedforward();
+#if HAL_QUADPLANE_ENABLED
+    if (plane.quadplane.enabled()) {
```
IamPete1:
```suggestion
    if (plane.quadplane.available()) {
```
Enabled will go true when the param is changed. `available` checks that init has been done and `attitude_control` is not nulprt.

ArduPlane/systemid.cpp
```diff
@@ -155,25 +186,29 @@ void AP_SystemID::stop()
...
-        attitude_control->rate_bf_roll_sysid_rads(0);
-        attitude_control->rate_bf_pitch_sysid_rads(0);
-        attitude_control->rate_bf_yaw_sysid_rads(0);
-        plane.quadplane.pos_control->set_NE_control_scale_factor(1);
-
-        // re-initialise the XY controller so we take current position as target
-        plane.quadplane.pos_control->init_NE_controller();
-
+#if HAL_QUADPLANE_ENABLED
+        if (plane.quadplane.enabled()) {
```
IamPete1:
```suggestion
        if (plane.quadplane.available()) {
```

ArduPlane/systemid.cpp
```diff
@@ -236,13 +271,77 @@ void AP_SystemID::update()
...
+            output_offset.x = waveform_sample;
+            break;
+        case AxisType::FW_MIX_PITCH:
+            output_offset.y = waveform_sample;
+            break;
+        default:
+            break;
+    }
+
+#if HAL_QUADPLANE_ENABLED
```
IamPete1: Surprised to see `HAL_QUADPLANE_ENABLED` here in the `fw_update` is that just because of the higher expected loop rate?

ArduPlane/Attitude.cpp
bnsgeyer: done. Thanks!

ArduPlane/Attitude.cpp
bnsgeyer: done. Thanks!

ArduPlane/Attitude.cpp
bnsgeyer: done. Thanks!

ArduPlane/systemid.cpp
bnsgeyer: done. thanks!

ArduPlane/systemid.cpp
bnsgeyer: done. thanks!

ArduPlane/systemid.cpp
bnsgeyer: done. thanks!

ArduPlane/systemid.cpp
bnsgeyer: Yes.  I didn't give Fixed Wing users the option of subsample log rates since their loop rate is only 50 hz.  I guess if we want to keep this consistent between fixed wing and vtol modes then I guess I shouldn't distinguish.  I will check to see if the default LOG_BITMASK settings guarantee full rate logging for plane.

bnsgeyer: the default is full rate logging so I made plane logging the same and quadplane which allow subsample logging.

ArduPlane/mode.h
bnsgeyer: @IamPete1 Please check the latest commit and let me know if that is the right approach for restricting System ID during takeoff and landing?

IamPete1:
I have done a PR against this branch to add in the additional checks I would like to see. In particular the chirp is stopped if VTOL assist kick in. 

https://github.com/bnsgeyer/ardupilot/pull/12

bnsgeyer:
@IamPete1 I incorporated your commit and verified that it did stop the system ID if the aircraft was given the land command in plane and it would not start if it was still in the Takeoff command.  I think it is ready to merge if you are happy with it.  I rebased to bring it up to current master

IamPete1:
I have rebased to see if we can get passed the failing linux check.

IamPete1 (approved):

bnsgeyer:
@tridge please review the latest changes that Pete made. Thanks!

PR #29613, opened 2025-03-25
AC_Avoidance: avoid use of location.alt in logging

peterbarker:
Changed to using floats.

Inspected lgos for test.Rover.PolyFenceObjectAvoidanceBendyRulerEasierGuided - altitudes are appropriate after this change considering what was in there beforehand.  Which appears to be garbage....

### DIFF
--- a/libraries/AC_Avoidance/AC_Avoidance_Logging.cpp
+++ b/libraries/AC_Avoidance/AC_Avoidance_Logging.cpp
@@ -12,10 +12,17 @@
 #if AP_OAPATHPLANNER_BENDYRULER_ENABLED
 void AP_OABendyRuler::Write_OABendyRuler(const uint8_t type, const bool active, const float target_yaw, const float target_pitch, const bool resist_chg, const float margin, const Location &final_dest, const Location &oa_dest) const
 {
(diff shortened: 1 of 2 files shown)


rmackay9:
Hi @peterbarker,

Thanks for this.  Can you clarify what was incorrect before?

The change in behaviour looks like we're now putting NaN in to the altitude logging if we can't get the alt above origin while before it would have just put whatever alt was in there regardless of the frame.

peterbarker:
> Thanks for this. Can you clarify what was incorrect before?

The direct inspection of the `.alt` field in the Location object was the problem.  We want to make this field private to avoid all of the sorts of bugs currently plaguing the Plane code.

> The change in behaviour looks like we're now putting NaN in to the altitude logging if we can't get the alt above origin while before it would have just put whatever alt was in there regardless of the frame.

Well, that and it's in metres not centimetres!

> I think the only place you'd see a change in behaviour is if bendy ruler was using alt-above-terrain. I must admit I'm not sure if we support that.

The numbers in this field for Rover are the same before/after adjusted for the different scale.  I don't think they're *good* values, but at least they're the same!

rmackay9 (approved):
ok, thanks!

PR #29614, opened 2025-03-25
SPI Full Duplex

andyp1per:
This PR makes the use of the full duplex SPI API with DMA optimisation explicit to avoid contortions in expected behaviour by the underlying HALs and in the caller.


libraries/AP_HAL/Device.h
```diff
@@ -124,6 +124,16 @@ class AP_HAL::Device {
     virtual bool transfer(const uint8_t *send, uint32_t send_len,
                           uint8_t *recv, uint32_t recv_len) = 0;
 
+    /*
+     * Core transfer function. This does a single bus transaction which
+     * sends len bytes and receives len bytes back from the slave.
+     * DMA-optimized on some implementations
+     *
+     * Return: true on a successful transfer, false on failure.
+     */
```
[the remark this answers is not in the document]
andyp1per: Thanks, fixed

tpwrules (changes requested):
Good news is this works and flies great on the Stampfly, which has a BMI270. I previously had to update the ESP32 HAL to account for the quirk and make that IMU actually work.

But I think I found one more place that needs changing to `transfer_fullduplex`: https://github.com/ArduPilot/ardupilot/blob/dfbbef4cb82a8185e58ff2f77cb5f16f689e9aa7/libraries/AP_OSD/AP_OSD_MAX7456.cpp#L205

libraries/AP_HAL_ChibiOS/SPIDevice.cpp
```diff
@@ -292,7 +292,7 @@ bool SPIDevice::transfer(const uint8_t *send, uint32_t send_len,
     if (!bus.semaphore.check_owner()) {
         return false;
     }
-    if ((send_len == recv_len && send == recv) || !send || !recv) {
```
tridge, line 295: this change is wrong, will break any out of tree device drivers

tridge (changes requested):

andyp1per:
Tested on Matek H7A3

tpwrules (approved):
I think we do need to somehow deprecate the old semantics in ChibiOS. I worry a bit about upstreaming broken drivers.

libraries/AP_HAL_ESP32/SPIDevice.cpp
```diff
@@ -112,7 +112,7 @@ bool SPIDevice::transfer(const uint8_t *send, uint32_t send_len,
 #ifdef SPIDEBUG
     printf("%s:%d \n", __PRETTY_FUNCTION__, __LINE__);
 #endif
-    if ((send_len == recv_len && send == recv) || !send || !recv) {
+    if (!send || !recv) {
```
tridge, line 115: I thought you reverted this change?

tridge (changes requested):

tridge (approved):

libraries/AP_HAL/SPIDevice.h
```diff
@@ -45,6 +45,14 @@ class SPIDevice : public Device {
     virtual bool transfer_fullduplex(const uint8_t *send, uint8_t *recv,
                                      uint32_t len) = 0;
 
+    /*
+     * Like #transfer(), but len bytes are both transmitted and received from/into @send_recv.
+     * This function is optimized for DMA-enabled multi-transfers without buffer copying
+     */
+    bool transfer_fullduplex(uint8_t *send_recv, uint32_t len) override {
```
tridge, line 52: this is a copy of the one in Device.h, why is this here?

tridge (changes requested):

libraries/AP_HAL/SPIDevice.h
tpwrules: Removed, hopefully it's not necessary for Linux or something :)

tpwrules (approved):
Tested again and it works on the bench on ESP32 and Cube Orange. Removed the redundant function. Should be ready for merge!

andyp1per:
@tpwrules the function you removed is necessary to make everything build correctly - I spent quite a long time fixing this - please don't make changes to other people's PR's unless you know it's correct!

libraries/AP_HAL/SPIDevice.h
andyp1per: It can't be removed or loads of hals do not build

tridge: which HAL doesn't build without the transfer_fullduplex() in AP_HAL/SPIDevice.h?

tpwrules: QURT at least had a build failure on CI. It seems to be a Clang warning that you have to override all the variations with different arguments of a particular function name to avoid ambiguity: https://stackoverflow.com/a/9995567

```
 [10/933] Compiling libraries/AC_AttitudeControl/AC_CommandModel.cpp
In file included from ../../libraries/AC_AttitudeControl/AC_CommandModel.cpp:1:
In file included from ../../libraries/AC_AttitudeControl/AC_CommandModel.h:6:
In file included from ../../libraries/AP_Param/AP_Param.h:27:
In file included from ../../libraries/AP_HAL/AP_HAL.h:8:
In file included from ../../libraries/AP_HAL/AP_HAL_Main.h:19:
In file included from ../../libraries/AP_HAL/HAL.h:11:
../../libraries/AP_HAL/SPIDevice.h:45:18: fatal error: 'AP_HAL::SPIDevice::transfer_fullduplex' hides overloaded virtual function [-Woverloaded-virtual]
    virtual bool transfer_fullduplex(const uint8_t *send, uint8_t *recv,
                 ^
../../libraries/AP_HAL/Device.h:137:18: note: hidden overloaded virtual function 'AP_HAL::Device::transfer_fullduplex' declared here: different number of parameters (2 vs 3)
    virtual bool transfer_fullduplex(uint8_t *send_recv, uint32_t len) {
                 ^
1 error generated.
```

I personally would just mark it as pure virtual `= 0` instead of duplicating the implementation. More C++ learning!

PR #29616, opened 2025-03-25
Scripting: added sim:set_pose() binding and example

tridge:
This allows scripting in SITL to override the vehicle position, velocity and attitude. It can be used to reproduce tricky issues in attitude recovery


rmackay9:
I haven't really reviewed this much but re naming of the lua script, maybe it should start with "sitl"?  We've already got the [sitl_standby_sim.lua](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/examples/sitl_standby_sim.lua) script that seems to be SITL/SIM related and its prefixed with "sitl"

tridge:
> I haven't really reviewed this much but re naming of the lua script, maybe it should start with "sitl"?

the reason I chose "sim_" is it applies to sim-on-hw too, @peterbarker has been trying to get me to refer to SIM instead of SITL when talking about something that applies to both

PR #29618, opened 2025-03-26
autotest: exempt AP_PROXIMITY_HEXSOONRADAR_ENABLED from sanity checks

peterbarker:
AP_PROXIMITY_MR72_ENABLED too

These share a symbol, so we can't leave one of them out of the code and see the symbol disappear!


rmackay9 (approved):
thanks!

Issue #29621, opened 2025-03-26
LOG_FILE_BUFSIZE


fanman2014:
**IF YOU DON'T REMOVE THESE FOUR LINES, THEN YOUR REPORT WILL BE CLOSED AUTOMATICALLY**
Questions and user problems should be directed at the forum (http://discuss.ardupilot.org)
_**Please be very sure you have found a bug when opening this issue**_
If there was a previous discussion in the forum, link to it

## Bug report

**Issue details**

_Please describe the problem_

**Version**
_What version was the issue encountered with_

**Platform**
[X  ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
_What type of airframe (flying wing, glider, hex, Y6, octa etc)_

**Hardware type**
_What autopilot hardware was used? (Pixhawk, Cube, Pixracer, Navio2, etc)_

**Logs**
_Please provide a link to any relevant logs that show the issue_

Wiki is confusing regarding base units for LOG_FILE_BUFSIZE. The default is 200 but wiki says max of 64kb?

LOG_FILE_BUFSIZE: Maximum AP_Logger File and Block Backend buffer size (in kilobytes)[¶](https://ardupilot.org/copter/docs/parameters.html#log-file-bufsize-maximum-ap-logger-file-and-block-backend-buffer-size-in-kilobytes)

The File and Block backends use a buffer to store data before writing to the block device. Raising this value may reduce “gaps” in your SD card logging. This buffer size may be reduced depending on available memory. PixHawk requires at least 4 kilobytes. Maximum value available here is 64 kilobytes.

rmackay9:
The fix has been merged to master now, thanks for the report!

PR #29622, opened 2025-03-26
AP_Logger: improve LOG_FILE_BUFSIZE param docs

rmackay9:
This improves the LOG_FILE_BUFSIZE parameter description a little bit by:

1. adding units
2. adding the range of 4~200 which is taken from the comments and the defines at the top of the file
3. removed a reference to Pixhawk which is now just one of more than 200 boards that we support

This resolves issue https://github.com/ArduPilot/ardupilot/issues/29621 which was also discussed here https://discuss.ardupilot.org/t/log-file-bufsize/132358


peterbarker (approved):
LGTM

rmackay9:
Normally I'd wait for a second approval but because it's a non-functional change (documentation only) I've gone ahead and merged this.

PR #29623, opened 2025-03-26
AP_NavEKF3: Manual Lane switching 

rishabsingh3003:
While doing EKF-based tests, this is helpful. 
When the EK3_OPTIONS second bit is enabled, automatic lane switching will be disabled (unless the core becomes "unhealthy", which is very extreme and we should never be flying with that). And EK3_PRIMIARY core will be enforced. This also means we can select the active core from a Lua Script mid flight.  

Might also have some use cases when lane switches are happening often because of small timing difference in the core. Example, flying with a single baro on a windy day and seeing lane switches isn't uncommon. 

I personally also have a usecase for this PR where I am working on another branch that allows different types of sensors to be used in different lanes, for example GPS on lane 1, Ext Nav on lane 2. For testing Ext Nav, it'll be helpful to disable lane switching.

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3.cpp
@@ -739,8 +739,8 @@ const AP_Param::GroupInfo NavEKF3::var_info2[] = {
 
     // @Param: OPTIONS
     // @DisplayName: Optional EKF behaviour
-    // @Description: This controls optional EKF behaviour. Setting JammingExpected will change the EKF nehaviour such that if dead reckoning navigation is possible it will require the preflight alignment GPS quality checks controlled by EK3_GPS_CHECK and EK3_CHECK_SCALE to pass before resuming GPS use if GPS lock is lost for more than 2 seconds to prevent bad
-    // @Bitmask: 0:JammingExpected
+    // @Description: EKF optional behaviour. Bit 0 (JammingExpected): Setting JammingExpected will change the EKF behaviour such that if dead reckoning navigation is possible it will require the preflight alignment GPS quality checks controlled by EK3_GPS_CHECK and EK3_CHECK_SCALE to pass before resuming GPS use if GPS lock is lost for more than 2 seconds to prevent bad position estimate. Bit 1 (Manual lane switching): DANGEROUS – If enabled, this disables automatic lane switching. If the active lane becomes unhealthy, no automatic switching will occur. Users must manually set EK3_PRIMARY to change lanes. No health checks will be performed on the selected lane. Use with extreme caution.
+    // @Bitmask: 0:JammingExpected, 1: ManualLaneSwitching
     // @User: Advanced
     AP_GROUPINFO("OPTIONS",  11, NavEKF3, _options, 0),
 
@@ -941,10 +941,20 @@ void NavEKF3::UpdateFilter(void)
         runCoreSelection = (imuSampleTime_us - lastUnhealthyTime_us) > 1E7;
     }
 
+    const uint8_t user_primary = uint8_t(_primary_core) < num_cores? _primary_core : 0;
+    bool lane_switching_enabled = true;
+    if (option_is_enabled(Option::ManualLaneSwitch)) {
+        lane_switching_enabled = false;
+        if (primary != user_primary) {
+            // switch back to the user-selected core
+            switchLane(user_primary);
+        }
+    }
+
     const bool armed  = dal.get_armed();
 
     // core selection is only available after the vehicle is armed, else forced to lane 0 if its healthy
-    if (runCoreSelection && armed) {
+    if (runCoreSelection && armed && lane_switching_enabled) {
         // update this instance's error scores for all active cores and get the primary core's error score
         float primaryErrorScore = updateCoreErrorScores();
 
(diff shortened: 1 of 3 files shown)


libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -941,10 +941,27 @@ void NavEKF3::UpdateFilter(void)
         runCoreSelection = (imuSampleTime_us - lastUnhealthyTime_us) > 1E7;
     }
 
+    const uint8_t user_primary = uint8_t(_primary_core) < num_cores? _primary_core : 0;
+    bool laneSwitchEnabled = true;
```
rmackay9, line 945: nitpick: I'd vote for this variable being called "lane_switching_enabled"

libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -941,10 +941,27 @@ void NavEKF3::UpdateFilter(void)
...
+            laneSwitchEnabled = true;
+        } else {
+            // current primary is healthy
+            laneSwitchEnabled = false;
+
+            if (primary != user_primary && core[user_primary].healthy()) {
+                // switch back to the user-selected core if it's healthy
+                primary = user_primary;
+            }
+    }
```
rmackay9, line 958: I think the brackets are not quite lined up..

libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -740,7 +740,7 @@ const AP_Param::GroupInfo NavEKF3::var_info2[] = {
     // @Param: OPTIONS
     // @DisplayName: Optional EKF behaviour
     // @Description: This controls optional EKF behaviour. Setting JammingExpected will change the EKF nehaviour such that if dead reckoning navigation is possible it will require the preflight alignment GPS quality checks controlled by EK3_GPS_CHECK and EK3_CHECK_SCALE to pass before resuming GPS use if GPS lock is lost for more than 2 seconds to prevent bad
-    // @Bitmask: 0:JammingExpected
+    // @Bitmask: 0:JammingExpected, 1: Disable lane switching
```
tridge, line 743: needs a better description, and maybe "Manual lane switching" ?

tridge (changes requested):
I think it should be manual-only, no health check

libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -739,8 +739,8 @@ const AP_Param::GroupInfo NavEKF3::var_info2[] = {
 
     // @Param: OPTIONS
     // @DisplayName: Optional EKF behaviour
-    // @Description: This controls optional EKF behaviour. Setting JammingExpected will change the EKF nehaviour such that if dead reckoning navigation is possible it will require the preflight alignment GPS quality checks controlled by EK3_GPS_CHECK and EK3_CHECK_SCALE to pass before resuming GPS use if GPS lock is lost for more than 2 seconds to prevent bad
-    // @Bitmask: 0:JammingExpected
+    // @Description: Bit 0 (JammingExpected): This controls optional EKF behaviour. Setting JammingExpected will change the EKF nehaviour such that if dead reckoning navigation is possible it will require the preflight alignment GPS quality checks controlled by EK3_GPS_CHECK and EK3_CHECK_SCALE to pass before resuming GPS use if GPS lock is lost for more than 2 seconds to prevent bad. Bit 1 (Manual lane switching): DANGEROUS – If enabled, this disables automatic lane switching. If the active lane becomes unhealthy, no automatic switching will occur. Users must manually set EK3_PRIMARY to change lanes. No health checks will be performed on the selected lane. Use with extreme caution.
```
rmackay9, line 742: looks like a little typo in the Description with "Bit 0" added to the front.  As a side note, parameter descriptions don't need to have a "This controls.." in the description.  Instead we could just have "EKF optional behaviour."

rmackay9: a preexisting issue but the existing description seems incomplete.  It ends with just, "more than 2 seconds to prevent bad".  bad what?

libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -941,10 +941,20 @@ void NavEKF3::UpdateFilter(void)
         runCoreSelection = (imuSampleTime_us - lastUnhealthyTime_us) > 1E7;
     }
 
+    const uint8_t user_primary = uint8_t(_primary_core) < num_cores? _primary_core : 0;
+    bool lane_switching_enabled = true;
+    if (option_is_enabled(Option::DisableLaneSwitch) ) {
```
rmackay9, line 946: nitpick: little space between the final round brackets

libraries/AP_NavEKF3/AP_NavEKF3.h
```diff
@@ -455,6 +458,7 @@ class NavEKF3 {
     // enum for processing options
     enum class Option {
         JammingExpected     = (1<<0),
+        DisableLaneSwitch   = (1<<1),
```
rmackay9, line 461: let's keep the naming consistent?  So "ManualLaneSwitch" might be better

libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -1049,12 +1057,22 @@ void NavEKF3::checkLaneSwitch(void)
...
-        updateLaneSwitchYawResetData(newPrimaryIndex, primary);
-        updateLaneSwitchPosResetData(newPrimaryIndex, primary);
-        updateLaneSwitchPosDownResetData(newPrimaryIndex, primary);
-        primary = newPrimaryIndex;
-        lastLaneSwitch_ms = now;
+    switchLane(newPrimaryIndex);
+}
+
+// switch to a new lane
+void NavEKF3::switchLane(uint8_t new_lane_index)
```
rmackay9, line 1064: is it lane or core?

rishabsingh3003: I have kept it as lane to keep it consistent with the nearby code

Issue #29600, opened 2025-03-24
SIM_WIND_TURB cannot be float


abeyn:
## Bug report

When I set param SIM_WIND_TURB to 0.99 in SITL nothing changes.

The problem in incorrect type cast in line: libraries/AP_HAL_SITL/SITL_State.cpp:371
input.wind.turbulence = _sitl?_sitl->wind_turbulance:0;
It cast float parameter to integer because ":0"
If change this code to _sitl?_sitl->wind_turbulance:0.0; problem is solved.

**Version**
4.7.0

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[X] Plane
[  ] Rover
[  ] Submarine

peterbarker:
@Safe-bet fixed this one for us - thanks all.

PR #29624, opened 2025-03-26
Changed default type of ternary operation from integer 0 to float 0.0f

Safe-bet:
This PR fixes the float issue by making the default type in ternary operation to be float from integer. I hope it doesn't cause any further complications in the code.


peterbarker:
> This PR fixes the float issue by making the default type in ternary operation to be float from integer. I hope it doesn't cause any further complications in the code.

Which "float issue" are you referring to here?  The value would ordinarily be promoted to a floating point value.  Were you seeing compiler warnings or equivalent?

Please note that the commit message (NOT the PR title) should be something like `AP_HAL_SITL: correct compiler warning promoting integer to float`.  Commit message requirements are in our style guide in our Wiki.

Safe-bet:
> > This PR fixes the float issue by making the default type in ternary operation to be float from integer. I hope it doesn't cause any further complications in the code.
> 
> Which "float issue" are you referring to here? The value would ordinarily be promoted to a floating point value. Were you seeing compiler warnings or equivalent?
> 
> Please note that the commit message (NOT the PR title) should be something like `AP_HAL_SITL: correct compiler warning promoting integer to float`. Commit message requirements are in our style guide in our Wiki.

Actually, there was an issue recently raised "SIM_WIND_TURB cannot be float #29600", due to one of the values being 0, the whole ternary operation output would be type casted to INT, so the issue mentioned would occur. I have just converted that INT 0 to float 0.0f.

rmackay9:
Hi @Safe-bet,

Thanks for the PR!  By the way, we always prefix our commits with the subsystem affected in order to make backporting easier.  Assuming this PR goes ahead, could you modify the commit title to start with "AP_HAL_SITL: "?  To be clear I mean the commit title, not the PR title.

libraries/AP_HAL_SITL/SITL_State.cpp
```diff
@@ -368,7 +368,7 @@ void SITL_State::_simulator_servos(struct sitl_input &input)
 
     input.wind.speed = wind_speed;
     input.wind.direction = wind_direction;
-    input.wind.turbulence = _sitl?_sitl->wind_turbulance:0;
+    input.wind.turbulence = _sitl?_sitl->wind_turbulance:0.0f;
```
peterbarker, line 371:
```suggestion
    input.wind.turbulence = _sitl->wind_turbulance;
```

Fixes the problem and now there's less code to have bugs in :-)

You can see at the top of this method we check for nullptr, and the same pointer is used unprotected elsewhere in here.

peterbarker (changes requested):
Holy type-inference failure, Batman!

This is really quite a gnarly problem, and I'm concerned we might have the same problem elsewhere.  I can't see how to automatically find those.

peterbarker:
Fixes https://github.com/ArduPilot/ardupilot/issues/29600

libraries/AP_HAL_SITL/SITL_State.cpp
Safe-bet: Hey Peter,
Why did you remove ternary operator here, wasn't it there to protect from nullptr, also can you tell me where this sitl struct was originally defined.
Thanks

peterbarker (approved):
LGTM

libraries/AP_HAL_SITL/SITL_State.cpp
peterbarker: > Hey Peter, Why did you remove ternary operator here, wasn't it there to protect from nullptr, also can you tell me where this sitl struct was originally defined. Thanks

I explained why it was safe to assume `_sitl` wasn't null in my earlier comment.

`_sitl` is initialised in `sitl_setup` from a singleton object.  The declaration of the variable is done in the base class for `SITL_State`, `SITL_State_Common.`

peterbarker:
I've squashed the wo commits together and tweaked the commit message a little.

tridge (approved):

PR #29625, opened 2025-03-26
Stop using location.alt field in AHRS logging

peterbarker:
I've checked the logs before/after this patch and the same data is present.

DCM doesn't seem to do well with altitude - the discretisation here is weird.

image

That log was created by right-clicking on a ~600m-high hill to the north-west of CMAC (the quarry) and "set home with height".

Looks like we may not be resetting the barometer zero datum correctly on a set-home.

### DIFF
--- a/libraries/AP_AHRS/AP_AHRS_Logging.cpp
+++ b/libraries/AP_AHRS/AP_AHRS_Logging.cpp
@@ -18,13 +18,17 @@ void AP_AHRS::Write_AHRS2() const
     if (!get_secondary_attitude(euler) || !get_secondary_position(loc) || !get_secondary_quaternion(quat)) {
         return;
     }
+    float alt;
+    if (!loc.initialised() || !loc.get_alt_m(Location::AltFrame::ABSOLUTE, alt)) {
+        alt = AP::logger().quiet_nanf();
+    }
     const struct log_AHRS pkt{
         LOG_PACKET_HEADER_INIT(LOG_AHR2_MSG),
         time_us : AP_HAL::micros64(),
         roll  : (int16_t)(degrees(euler.x)*100),
         pitch : (int16_t)(degrees(euler.y)*100),
         yaw   : (uint16_t)(wrap_360_cd(degrees(euler.z)*100)),
-        alt   : loc.alt*1.0e-2f,
+        alt   : alt,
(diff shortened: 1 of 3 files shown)


libraries/AP_AHRS/AP_AHRS_Logging.cpp
```diff
@@ -85,16 +93,27 @@ void AP_AHRS::Write_POS() const
...
     const struct log_POS pkt{
         LOG_PACKET_HEADER_INIT(LOG_POS_MSG),
         time_us        : AP_HAL::micros64(),
         lat            : loc.lat,
         lng            : loc.lng,
-        alt            : loc.alt*1.0e-2f,
-        rel_home_alt   : -home,
-        rel_origin_alt : get_relative_position_D_origin(origin) ? -origin : AP::logger().quiet_nanf(),
+        alt            : alt,
+        rel_home_alt   : -home_pos_down,   // down->up
```
IamPete1, line 115: Will negating a nan result in internal errors? I'm fairly sure its fine we only check on things like constrain.

peterbarker: Good question.

I've restructured the code to avoid the question :-)

libraries/AP_AHRS/AP_AHRS.cpp
```diff
@@ -1871,13 +1871,15 @@ bool AP_AHRS::get_relative_position_D_origin(float &posD) const
...
+bool AP_AHRS::get_relative_position_D_home(float &posD) const
 {
     if (!_home_is_set) {
         // fall back to an altitude derived from barometric pressure
         // differences vs a calibrated ground pressure:
         posD = -AP::baro().get_altitude();
-        return;
+        // note that the Baro library currently lies about its success
+        // in providing an altitude.  Not all vehicles have barometers.
+        return true;
```
tridge, line 1882: all returns are true, so bool pointless (in fact misleading)

tridge (dismissed):

libraries/AP_AHRS/AP_AHRS.cpp
peterbarker: I've removed this change entirely - we continue to `void` now.

PR #29626, opened 2025-03-27
AP_BattMonitor: use state rather than params for type

peterbarker:
@andyp1per you might like this one after having to play around with the nasty casting stuff!

Type has to live in Params as Params can't see the AP_BattMonitor class to copy the enum from.

edit: now fixes an issue where we were using the configured type rather than allocated type.  The use of the enumeration made it much clearer what was going on.

### DIFF
--- a/libraries/AP_BattMonitor/AP_BattMonitor_Analog.cpp
+++ b/libraries/AP_BattMonitor/AP_BattMonitor_Analog.cpp
@@ -88,14 +88,14 @@ AP_BattMonitor_Analog::AP_BattMonitor_Analog(AP_BattMonitor &mon,
 #endif
     _state.var_info = var_info;
     
-    if ((AP_BattMonitor::Type)_params._type.get() != AP_BattMonitor::Type::ANALOG_CURRENT_ONLY) {
+    if (_params._type != AP_BattMonitor::Type::ANALOG_CURRENT_ONLY) {
         _volt_pin_analog_source = hal.analogin->channel(_volt_pin);
         if (_volt_pin_analog_source == nullptr) {
             AP_BoardConfig::config_error("No analog voltage channel for battery %d", mon_state.instance);
         }
     }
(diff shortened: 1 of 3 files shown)


andyp1per (approved):

peterbarker:
@andyp1per sorry, I patched another couple of lines in here to look at the allocated type rather than the configured type.  I'm not sure might happen in the rest of the code if `has_current` returns true when you actually don't.  Nothing good, I'll wager.

libraries/AP_BattMonitor/AP_BattMonitor_Analog.cpp
```diff
@@ -108,7 +108,7 @@ AP_BattMonitor_Analog::AP_BattMonitor_Analog(AP_BattMonitor &mon,
 void
 AP_BattMonitor_Analog::read()
 {
-    if ((AP_BattMonitor::Type)_params._type.get() != AP_BattMonitor::Type::ANALOG_CURRENT_ONLY) {
+    if (_state.type != AP_BattMonitor::Type::ANALOG_CURRENT_ONLY) {
```
andyp1per, line 111: this is a change in behaviour, ditto below

andyp1per (changes requested):

libraries/AP_BattMonitor/AP_BattMonitor_Analog.cpp
peterbarker: Yes, this is why I was bringing it to your attention.

We had some nasty bugs in the CAN battery backend because we looked at the configured type rather than the sort of backend we *actually* allocated.

In the `read()` method here we *must* only look at the allocated type, *never* the configured-via-parameters type.  This patch makes sure we do that.  Whatever problems this might introduce are probably benign in this case, but we don't want the pattern in the code.

andyp1per: Ok, need to change the title then :)

peterbarker: > Ok, need to change the title then :)

Fair call, done

andyp1per (approved):

PR #29627, opened 2025-03-27
Scripting: added gyro to sim:set_pose()

tridge:
useful for some test scenarios and gives better repeatability


peterbarker (approved):
LGTM

PR #29636, opened 2025-03-28
AP_ExternalAHRS: Protect against uninitialized location and origin

Ryanf55:
* Add protections in AP_ExternalAHRS against improperly initialized EAHRS drivers
* Without this, you get faults deep inside DCM that are much less obvious
* An uninitialed origin or location, just after the `have_origin` or `have_location`, is likely a bug

Continuation of https://github.com/ArduPilot/ardupilot/pull/29628

### DIFF
--- a/libraries/AP_ExternalAHRS/AP_ExternalAHRS.cpp
+++ b/libraries/AP_ExternalAHRS/AP_ExternalAHRS.cpp
@@ -174,6 +174,11 @@ bool AP_ExternalAHRS::get_origin(Location &loc)
 {
     if (state.have_origin) {
         WITH_SEMAPHORE(state.sem);
+#if CONFIG_HAL_BOARD == HAL_BOARD_SITL
+        if (!state.origin.initialised()) {
+            AP_HAL::panic("Uninitialized origin in AP_ExternalAHRS.");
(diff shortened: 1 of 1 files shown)


peterbarker (approved):

libraries/AP_ExternalAHRS/AP_ExternalAHRS.cpp
```diff
@@ -174,6 +174,11 @@ bool AP_ExternalAHRS::get_origin(Location &loc)
 {
     if (state.have_origin) {
         WITH_SEMAPHORE(state.sem);
+#if CONFIG_HAL_BOARD == HAL_BOARD_SITL
+        if (!loc.initialised()) {
+            AP_HAL::panic("Uninitialized origin in AP_ExternalAHRS.");
+        }
+#endif
         loc = state.origin;
```
Ryanf55, line 182:
```suggestion
#if CONFIG_HAL_BOARD == HAL_BOARD_SITL
        if (!state.origin.initialised()) {
            AP_HAL::panic("Uninitialized origin in AP_ExternalAHRS.");
        }
#endif
        loc = state.origin;
```

PR #29639, opened 2025-03-28
AP_Compass: Added support for MMC5603 compass

pawelplsi:
This adds a driver for MMC5603 Memsic compass devices. Code is based on existing drivers for other MMC devices. Verified to work reliably.


peterbarker:
This implementation looks really quite close the the existing 5XX3 implementation.

Obviously the registers have moved around a bit, but I haven't spotted anything in here which says that we can't adjust the 5XX3 drive to handle this device as well as 5983.

We have to be *really* careful with the amount of flash space we take up with compass (and baro) drivers.  The problem is that because these are installed on external devices like compasses we generally need support for them on all our boards (so the end user doesn't need a custom compilation for their device).  Many of our boards are very short on space.  So if we can re-use the MMC5XX3 code we really should!

peterbarker:
Ping @pawelplsi - still want to chase this one?

peterbarker:
OK, no answer from the author.

Please re-open if you want to pursue this.

PR #29640, opened 2025-03-29
AP_GPS: Add protection against bad EAHRS GPS

Ryanf55:
* If an EARHS GPS is reporting 0 lat/lon, it's likely a bug
* At this point, the data should be initialized
* Protects against multipling by 1E7 instead of 1E-7 in lat-lon conversion
* Follow up to https://github.com/ArduPilot/ardupilot/pull/29636 and https://github.com/ArduPilot/ardupilot/pull/29628


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -1297,6 +1297,11 @@ bool AP_GPS::get_first_external_instance(uint8_t& instance) const
 void AP_GPS::handle_external(const AP_ExternalAHRS::gps_data_message_t &pkt, const uint8_t instance)
 {
     if (get_type(instance) == GPS_TYPE_EXTERNAL_AHRS && drivers[instance] != nullptr) {
+#if CONFIG_HAL_BOARD == HAL_BOARD_SITL
+    if (pkt.longitude == 0 && pkt.latitude == 0) {
+        AP_HAL::panic("Invalid location passed to AP_GPS::handle_external");
+    }
```
peterbarker, line 1303: Indenting is incorrect here

peterbarker (changes requested):
I'm happy to merge this once the indenting is fixed.

libraries/AP_GPS/AP_GPS.cpp
Ryanf55: Fixed!

Ryanf55:
Sorry, accidentally bumped the submodule. Indent is now fixed, ready for final review.

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -1297,6 +1297,11 @@ bool AP_GPS::get_first_external_instance(uint8_t& instance) const
 void AP_GPS::handle_external(const AP_ExternalAHRS::gps_data_message_t &pkt, const uint8_t instance)
 {
     if (get_type(instance) == GPS_TYPE_EXTERNAL_AHRS && drivers[instance] != nullptr) {
+#if CONFIG_HAL_BOARD == HAL_BOARD_SITL
+        if (pkt.longitude == 0 && pkt.latitude == 0) {
```
peterbarker, line 1301: Any reason you're not checking altitude here?  It's traditional.... and non-zero alt is suffucient to make a location valid as far as ArduPilot's concerned...

Ryanf55: Because the bug I found had altitude set, but lat/lon were not set due to a division/multiplication error of 1E7.

PR #29642, opened 2025-03-29
AP_GPS: Protect against nullptr port

Ryanf55:
Screenshot from 2025-03-28 23-36-01

Saw this when mucking with NET_* ports and GPS. I would prefer to protect against this earlier in the stack, but we can't because it's a switch/case where the backend gets allocated. The nullptr port check in the frontend is currently here: https://github.com/ArduPilot/ardupilot/blob/2a5bc6223ebbec935789b9cabf608dd3469a3afc/libraries/AP_GPS/AP_GPS.cpp#L680

No all GPS's need a port.

### DIFF
--- a/libraries/AP_GPS/AP_GPS_GSOF.cpp
+++ b/libraries/AP_GPS/AP_GPS_GSOF.cpp
@@ -47,6 +47,11 @@ AP_GPS_GSOF::AP_GPS_GSOF(AP_GPS &_gps,
     AP_GPS_Backend(_gps, _params, _state, _port)
 {
 
+    if (port == nullptr) {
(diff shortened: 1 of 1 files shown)


peterbarker:
I've created an alternate PR to this one here: https://github.com/ArduPilot/ardupilot/pull/29647

rmackay9:
I did not include this in 4.6.0-beta6 due to a merge conflict.  Perhaps @Ryanf55 could create a backport PR to get this in?

Ryanf55:
> I did not include this in 4.6.0-beta6 due to a merge conflict. Perhaps @Ryanf55 could create a backport PR to get this in?

Done!

rmackay9:
this is included in 4.6.0-beta6 via #29892

PR #29643, opened 2025-03-29
Factor MS56XX driver out into MS5607, MS5611, MS5637 and MS5837 backends

peterbarker:
```
CubeOrange-periph-heavy  -448              *                                                    
Durandal                            -424   *           -416    -424               -416   -416   360
KakuteH7-bdshot                     -416   *           -416    -416               -416   -416   328
MatekF405                           *      *           *       *                  *      *      *
Pixhawk1-1M-bdshot                  -1096              -1096   -1096              -1096  -1096  -264
f303-Universal           -408              *                                                    
revo-mini                           -1088  *           -1088   -1088              -1088  -1088  -264
```


peterbarker:
CubeOrangePlus retains both of its MS5611 sensors.

peterbarker:
Disco retains the symbols for the MS5607

peterbarker:
mini-pix retains the symbols for MS5611 and MS5607 (for the fallback option) but doesn't have the other 2 backends.

peterbarker:
@Williangalvani when do you think you might be able to test this on the 5837?  I think I've done what testing I can do here on this one.

tridge:
@peterbarker unfortunately vendors do swap parts around on boards within the MS family (eg. 5607 vs 5611). I think this would break that?

Williangalvani:
> @Williangalvani when do you think you might be able to test this on the 5837? I think I've done what testing I can do here on this one.

just tested. working normally after this patch

peterbarker:
> @peterbarker unfortunately vendors do swap parts around on boards within the MS family (eg. 5607 vs 5611). I think this would break that?

No, that behaviour is explicitly retained.  Note my comment about MiniPix retaining symbols for both 5607 and 5611.  It does that because there's still the `BARO_OPTIONS` bit to force interpretation as a 5607

tridge (approved):

PR #29645, opened 2025-03-30
Fixes for enabling/disbling various build options

peterbarker:
- more hexsoon/nanoradar fixes
 - many fixes around Plane systemid stuff
 - exempt follow library for most checks as it's problematic.
 

`test_build_options.py` runs successfully again past these patches.


rmackay9 (approved):
sorry about missing that hexsoon one

PR #29647, opened 2025-03-30
AP_GPS: correct _port array-overread in EAHRS case

peterbarker:
Alternative to https://github.com/ArduPilot/ardupilot/pull/29642/files

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -332,7 +332,7 @@ void AP_GPS::init()
     // search for serial ports with gps protocol
     const auto &serial_manager = AP::serialmanager();
     uint8_t uart_idx = 0;
-    for (uint8_t i=0; i<ARRAY_SIZE(params); i++) {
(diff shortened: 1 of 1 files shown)


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -671,13 +673,13 @@ AP_GPS_Backend *AP_GPS::_detect_instance(uint8_t instance)
 #if AP_GPS_GSOF_ENABLED
     case GPS_TYPE_GSOF:
         dstate->auto_detected_baud = false; // specified, not detected
-        return NEW_NOTHROW AP_GPS_GSOF(*this, params[instance], state[instance], _port[instance]);
```
Ryanf55, line 674: Is `port` now guarateed to be non-null?

tridge (approved):

libraries/AP_GPS/AP_GPS.cpp
peterbarker: No :-)

PR #29650, opened 2025-03-31
AP_HAL_ESP32: esp32s3devkit - rename ADC defines

nkruzan:
Rename ADC defines that were broken at some point. There was a previous pr that was closed, this replaces it.


davidbuzz:
lets get this in.

tridge (approved):

PR #29653, opened 2025-03-31
AP_Terrain: Protect against division by 0 when TERRAIN_SPACING is less then 1

### LINKED ISSUE #29392: Setting TERRAIN_SPACING Too Small Causes Instant Crash Instead of Running Out of Storage
## Bug report

**Issue details**
Hi~
When the parameter TERRAIN_SPACING is set to a value less than 1, the system crashes, causing the GCS to lose connection.
For example, param set TERRAIN_SPACING 0 or param set TERRAIN_SPACING 0.5
(But there are no errors when TERRAIN_SPACING =1 or 1.5)

According to the parameter [description](https://ardupilot.org/copter/docs/parameters.html#terrain-spacing-terrain-grid-spacing):
"Grid spacings lower than 100 meters waste SD card space if the GCS cannot provide that resolution."

If the value is set below 100, it should only result in wasted storage space (eventually leading to an SD card running out of space), but it should not cause an instant crash. Therefore, I think this is a bug.

The error message from debugging is as follows:
Thread 1 "arducopter" received signal SIGFPE, Arithmetic exception.
0x00005555556979a4 in AP_Terrain::calculate_grid_info(Location const&, AP_Terrain::grid_info&) const ()

**Version**
SITL, copter, 4.7.0-dev

**Platform**
[  ] All
[  ] AntennaTracker
[*] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

tiagoc-santos:
This fixes #29392. In the calculate_grid_info function and the east_blocks function of AP_Terrain, the TERRAIN_SPACING parameter is used to determine grid spacing and how many blocks are needed in a stride for a given location. 

When this parameter is set to a value less than 1, a division by zero occurs, leading to an arithmetic exception. This fix   ensures that grid_spacing is properly handled when its value is below 1. Specifically, if grid_spacing is less than 1, the system defaults to using a value of 1 instead.

This fix has been tested in SITL environment and seems to be working fine.

### DIFF
--- a/libraries/AP_Terrain/AP_Terrain.cpp
+++ b/libraries/AP_Terrain/AP_Terrain.cpp
@@ -441,6 +441,10 @@ bool AP_Terrain::pre_arm_checks(char *failure_msg, uint8_t failure_msg_len) cons
         hal.util->snprintf(failure_msg, failure_msg_len, "waiting for terrain data");
         return false;
     }
+    if (grid_spacing <= 0) {
+        hal.util->snprintf(failure_msg, failure_msg_len, "TERRAIN_SPACING can't be <= 0");
+        return false;
+    }
 
     return true;
 }


peterbarker:
This probably needs to move to being a float at some stage to allow for sub-metre terrain grids.

When we do that we'd have to find all of these places again, which would be unfortunate.

Did you consider throwing a `config_error()` instead of constraining the value like you've done here?  ie. force the user to choose a valid value....

tiagoc-santos:
Hi @peterbarker.

I've been digging into the suggestion you made and I have a few questions.

Should I do the verification (and subsequently throw the `config_error()` ) inside the "problematic" functions? If so, do they need to return? Or is there a better place to make this check?

Thanks!

peterbarker:
> Should I do the verification (and subsequently throw the `config_error()` ) inside the "problematic" functions?

Hmm.  Doesn't look like terrain has a `.init()` function in it.  That's where it should go.

So I think not with `config_error`.

Since this should be short-term until we change the spacing to be float, we could probably add a pre-arm check for it.  So that's just an additional clause in `AP_Terrain::pre_arm_check`

It's entirely possible when we take this solution to DevCall that the response will be, "no, just make it a float".

Did you want me to take this to DevCall before you do the work instead of after?

tiagoc-santos:
> Hmm. Doesn't look like terrain has a `.init()` function in it. That's where it should go.
> 
> So I think not with `config_error`.
> 
> Since this should be short-term until we change the spacing to be float, we could probably add a pre-arm check for it. So that's just an additional clause in `AP_Terrain::pre_arm_check`
> 
> It's entirely possible when we take this solution to DevCall that the response will be, "no, just make it a float".
> 
> Did you want me to take this to DevCall before you do the work instead of after?

I don’t mind doing the pre-arm check, but I agree it might be worth bringing it to DevCall first to see what the consensus is. If you’re okay with that, feel free to raise it there before I dive into the patch.

tridge:
@tiagoc-santos please change it so that AP_Terrain::pre_arm_checks() return false with an error message if spacing < 1

peterbarker:
> > @tiagoc-santos please change it so that AP_Terrain::pre_arm_checks() return false with an error message if spacing <= 0

Yep, that's the ticket.
 
> Done! Let me know if you need me to rebase and/or squash the commits.

I've just done that (before reading your offer here - sorry!)

Marking as MergeOnCIPass

peterbarker (approved):

peterbarker:
Merged, thanks!

PR #29655, opened 2025-03-31
Tools: Added INA2xx parameters to ModalAI Sentinel M0065 frame params

katzfey:
The voltage and current readings usually come from the ESC in VOXL 2 based designs. But in the case where the IO board is being used for PWM then voltage and current readings need to come from the external VOXL PM power module. This PR adds the configuration parameters needed to use the VOXL PM with the IO board to the frame params file.


peterbarker:
Merged it!

PR #29657, opened 2025-03-31
AP_Scripting: scripted arming checks for plane for now

timtuxworth:
Parameterized and/or customizable arming checks for Plane (for now).

There's an included list of standard checks that can be turned on/off (or set to warn only) by parameters, or you can add your own.

Based on some earlier work, from a year ago. I've been flying this for a while.

### DIFF
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -7232,6 +7232,119 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
             if abs(m.Pos - m.PosCmd) > 20:
                 break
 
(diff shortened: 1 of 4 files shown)


peterbarker:
Should his be an applet or an example?

applets are expectd to be, "install and you're good to go", but your comment about "prbably need to customise" in here kind of hints this is not that...

timtuxworth:
> Should his be an applet or an example?
> 
> applets are expectd to be, "install and you're good to go", but your comment about "prbably need to customise" in here kind of hints this is not that...

It's both - so not sure, it can be used "as is" with the existing suite of arming checks by just changing the parameters for each existing check (if necessary) or ...

can be customized to add new checks of your own.

PR #29661, opened 2025-04-01
Add code to help retire old parameter storage entries

peterbarker:
```
Durandal                            *      *           *       *                 176    *      *
KakuteH7-bdshot                     *      *           *       *                 152    *      *
MatekF405                           *      *           *       *                 144    *      *
Pixhawk1-1M-bdshot                  *                  *       *                 152    *      *
revo-mini                           *      *           *       *                 152    *      *
```

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1589,4 +1589,10 @@ void Plane::load_parameters(void)
     };
 
     AP_Param::convert_toplevel_objects(toplevel_conversions, ARRAY_SIZE(toplevel_conversions));
+
+    static const AP_Param::ConversionInfo stale_parameters[] {
+        // unused since 2018 / Plane-3.9.0.  Config error if set in ArduPilot-4.7
+        { Parameters::k_param_software_type, 0, AP_PARAM_INT8, nullptr }
+    };
+    AP_Param::assert_not_in_storage(stale_parameters, ARRAY_SIZE(stale_parameters));
 }


tpwrules:
Is there a possibility in the format to simply clear out the specific entries instead of requiring the user to do something? Maybe add an "IGNORE_ME" type or flag?

peterbarker:
> Is there a possibility in the format to simply clear out the specific entries instead of requiring the user to do something? Maybe add an "IGNORE_ME" type or flag?

User may want to go backwards in time, 'thjough?

tpwrules:
Surely there will be no backward to go if you instead make them erase all the params? Why not just erase the relevant part automatically?

peterbarker:
> Surely there will be no backward to go if you instead make them erase all the params? Why not just erase the relevant part automatically?

Fair call.

In answer to your question, yes, we can mark parameters as unused.  I've done this in a previous PR by creating a magic value like
```
    static const uint16_t       _sentinal_unused_key   = 0x1FE;
```
next to
```
    static const uint16_t       _sentinal_key   = 0x1FF;
```

tpwrules:
I don't know much about the param format but my idea was extending `ap_var_type` with a type meaning ignore me. If that's used to determine the payload size we may have to OR with 0x80 or something. I don't know which would be more backwards compatible.

libraries/AP_Param/AP_Param.cpp
```diff
@@ -2037,6 +2037,18 @@ void AP_Param::convert_old_parameters_scaled(const struct ConversionInfo *conver
...
 }
 
+void AP_Param::assert_not_in_storage(const ConversionInfo *table, uint8_t table_size)
+{
+    for (uint8_t i=0; i<table_size; i++) {
+        const auto &info = table[i];
+        uint8_t old_value[type_size(info.type)];
+        AP_Param *ap = (AP_Param *)&old_value[0];
+        if (find_old_parameter(&info, ap)) {
+            AP_BoardConfig::config_error("Stale parameters in storage; wipe storage and reload");
```
andyp1per, line 2047: This is just unfriendly. I have lots of vehicles that have been upgraded from old versions of copter, forcing people to wipe is forcing people to potentially make a mistake for our own convenience. Much better to automatically remove the stale parameters if possible.

PR #29662, opened 2025-04-01
AP_HAL_Chibios:add hwdef files for PixSurveyA2-IND

mikefenghao:
add hwdef files for PixSurveyA2-IND.
makeflyeasy has done several tests on PixSurveyA2-IND and works well.
We tested it using a STRIVER mini aircraft, both autonomously and manually violently flown.


Hwurzburg (changes requested):
I made some changes/additions to the readme to better match hwdef and allow easy move to wiki when time comes...

Also, the Battery monitor setup is confusing and questionable...you should setup only one monitor by default BATT_MONITOR....it can be analog or CAN, but you need to decide which...the readme will need changing as well as the hwdef when you decide.

Finally, I think @andyp1per may have an issue with the hedef BIDIR tags

mikefenghao:
> I made some changes/additions to the readme to better match hwdef and allow easy move to wiki when time comes...
> 
> Also, the Battery monitor setup is confusing and questionable...you should setup only one monitor by default BATT_MONITOR....it can be analog or CAN, but you need to decide which...the readme will need changing as well as the hwdef when you decide.
> 
> Finally, I think @andyp1per may have an issue with the hwdef BIDIR tags

In PixSurveyA2-IND, usually we will access two PMUs, the first one is CAN PMU and the second one is analog PMU. Do you mean why there are two Battery monitors in hwdef? Because we want to be compatible with the customers of PixSurvey A1-IND, because the carrier board of PixSurvey A1-IND only has two analog interfaces, so whether A1 or A2, they can use the carrier board with each other.

libraries/AP_HAL_ChibiOS/hwdef/PixSurveyA2-IND/hwdef.dat
```diff
@@ -0,0 +1,237 @@
...
+OSCILLATOR_HZ 24000000
+
+# board ID for firmware load
+APJ_BOARD_ID 6104
+
+FLASH_SIZE_KB 2048
+
+
+# with 2M flash we can afford to optimize for speed
+env OPTIMIZE -O2
```
andyp1per, line 16: Should remove as we are removing this from all boards

andyp1per (approved):
Looks correct

Hwurzburg (dismissed):
so if plugged into old carrier board, which pins go to PWR1 and which to PWR2
and in new carrier bd, which set of analog pins go to the PWR2 ADC connector?

libraries/AP_HAL_ChibiOS/hwdef/PixSurveyA2-IND/hwdef.dat
mikefenghao: > Should remove as we are removing this from all boards

Thanks, I've deleted it.

mikefenghao:
> so if plugged into old carrier board, which pins go to PWR1 and which to PWR2 and in new carrier bd, which set of analog pins go to the PWR2 ADC connector?

Thank you very much. I have read your email carefully. You're right. I've corrected it. Now BATT1 uses CAN and BATT2 uses analog.

Hwurzburg:
@mikefenghao I have cleaned up some things, and pushed:
- removed changes to A1 readme...should be a separate PR
- removed module changes included
- cleaned up board_types
- reformatted and added to README so conversion to wiki page in the future is easier
which brings up a question....so far we have not added PixSurvey autopilots to the wiki assuming they are only used in turnkey systems...if this is not correct and you sell some to non MFE vehicle users, we should add them to the wiki...might take a while, though, since the READMEs were not reviewed by me for inclusion in the wiki and will need additional work to create the page...

Hwurzburg (approved):
if you approve of the changes I made, I will mark for Devcall and merge

PR #29667, opened 2025-04-01
AHRS: fixed initial attitude for non-zero AHRS_ORIENTATION

tridge:
we need to get the orientation right before any data is read

### DIFF
--- a/libraries/AP_AHRS/AP_AHRS.cpp
+++ b/libraries/AP_AHRS/AP_AHRS.cpp
@@ -229,6 +229,8 @@ AP_AHRS::AP_AHRS(uint8_t flags) :
 // init sets up INS board orientation
 void AP_AHRS::init()
 {
+    update_orientation();
+
     // EKF1 is no longer supported - handle case where it is selected
     if (_ekf_type.get() == 1) {
         AP_BoardConfig::config_error("EKF1 not available");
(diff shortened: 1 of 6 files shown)


peterbarker (approved):
LGTM

rmackay9:

dallinbriggs:
I would like to open this issue back up. I can confirm that the issue persists in this commit, at least with a Cube Orange Plus. 
Test:
- AHRS_orientation = 8
- Cube orange booted while inverted and held still for the entire boot process

Results:
- The vehicle's orientation could be seen flipping over after boot. 

Here is a log of it happening after boot. I have also confirmed that cherry picking this commit back to 4.5.7 also gives the same results. 
[inverted_bootup.zip](https://github.com/user-attachments/files/20094287/inverted_bootup.zip)

rmackay9:
I've reviewed @dallinbriggs's log and also reproduced it on my own CubeOrange.  I've created an issue to capture the problem here https://github.com/ArduPilot/ardupilot/issues/30064

This is what I see with 4.6.0-beta6 and 4.7.0-dev using Copter:
image

Strangely 4.5.7 has very similar behaviour but it cannot be captured in the logs, perhaps because logging starts later?

Dallin's log shows slightly different behaviour with the roll-over to the correct attitude taking much longer.  Maybe this is because DCM is the active attitude estimator in Plane?
image

PR #29676, opened 2025-04-02
AP_LandingGear: stop dup retract messages, clean up code

Hwurzburg:
While understanding recent landing gear changes in Plane for wiki docs, I found that the code can send multiple retract texts while already retracted and that the **_have_changed**  variable is unnecessary..also cleaned up the deploy and retract height code a bit

### DIFF
--- a/libraries/AP_LandingGear/AP_LandingGear.cpp
+++ b/libraries/AP_LandingGear/AP_LandingGear.cpp
@@ -163,19 +163,17 @@ void AP_LandingGear::deploy()
         return;
     }
 
-    // set servo PWM to deployed position
-    SRV_Channels::set_output_limit(SRV_Channel::k_landing_gear_control, SRV_Channel::Limit::MAX);
-
-    // send message only if output has been configured
+    // set servo and send message only if output has been configured and not already deployed
     if (!_deployed &&
         SRV_Channels::function_assigned(SRV_Channel::k_landing_gear_control)) {
         GCS_SEND_TEXT(MAV_SEVERITY_INFO, "LandingGear: DEPLOY");
+        // set deployed flag
+        _deployed = true;
+        _have_changed = true;
+        LOGGER_WRITE_EVENT(LogEvent::LANDING_GEAR_DEPLOYED);
+        // set servo PWM to deployed position
+        SRV_Channels::set_output_limit(SRV_Channel::k_landing_gear_control, SRV_Channel::Limit::MAX);
     }
-
-    // set deployed flag
-    _deployed = true;
-    _have_changed = true;
-    LOGGER_WRITE_EVENT(LogEvent::LANDING_GEAR_DEPLOYED);
 }
 
 /// retract - retract landing gear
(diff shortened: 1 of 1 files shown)


libraries/AP_LandingGear/AP_LandingGear.cpp
```diff
@@ -286,18 +281,19 @@ void AP_LandingGear::update(float height_above_ground_m)
...
             deploy();
-        }
-        if ((_deployed || !_have_changed) &&
-            _retract_alt > 0 &&
-            _retract_alt >= _deploy_alt &&
-            alt_m >= _retract_alt &&
-            _last_height_above_ground < _retract_alt) {
-            retract();
+        } else {
+            if (_deployed  &&
```
peterbarker, line 290:
```suggestion
        } else if (_deployed  &&
```

peterbarker (commented):
How have you tested this?

Hwurzburg:
> How have you tested this?

SITL.....using AUX sw, auto alt deploy/retract in FBWA, and Mode takeoff to mode autoland

libraries/AP_LandingGear/AP_LandingGear.cpp
```diff
@@ -286,18 +281,19 @@ void AP_LandingGear::update(float height_above_ground_m)
 
     if (hal.util->get_soft_armed()) {
         // only do height based triggering when armed
-        if ((!_deployed || !_have_changed) &&
+        if (!_deployed  &&
```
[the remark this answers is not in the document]
Hwurzburg: my observation also....variable does nothing

libraries/AP_LandingGear/AP_LandingGear.cpp
Hwurzburg: okay

libraries/AP_LandingGear/AP_LandingGear.cpp
```diff
@@ -184,18 +181,16 @@ void AP_LandingGear::retract()
...
-    // reset deployed flag
-    _deployed = false;
-    _have_changed = true;
-    LOGGER_WRITE_EVENT(LogEvent::LANDING_GEAR_RETRACTED);
-
-    // send message only if output has been configured
-    if (SRV_Channels::function_assigned(SRV_Channel::k_landing_gear_control)) {
+    
+    // set servo and send message only if output has been configured and already deployed
+    if (_deployed &&
```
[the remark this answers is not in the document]
IamPete1: `_have_changed` certainly looks suspect, its never set false. I'm not sure what it was supposed to do, but whatever it was it seems to me that it would only work the first time.

IamPete1 (changes requested):
Change looks good but we have there is a DroneCAN module change in here too.

Hwurzburg:
@IamPete1 duh, fixed

peterbarker (changes requested):
commit message needs fixing

libraries/AP_LandingGear/AP_LandingGear.cpp
```diff
@@ -286,18 +281,17 @@ void AP_LandingGear::update(float height_above_ground_m)
...
             _last_height_above_ground > _deploy_alt) {
             deploy();
-        }
-        if ((_deployed || !_have_changed) &&
-            _retract_alt > 0 &&
-            _retract_alt >= _deploy_alt &&
-            alt_m >= _retract_alt &&
-            _last_height_above_ground < _retract_alt) {
-            retract();
+        } else if (_deployed  &&
```
tridge, line 289: this will break the case of DEPLOY_ALT=0 and RETRACT_ALT>0, and no aux function used, it will no longer retract on first takeoff

tridge (changes requested):

libraries/AP_LandingGear/AP_LandingGear.cpp
Hwurzburg: it does in SITL...servo 5 is the retract servo
LGR_DEPLOY_ALT   0.000000
LGR_DEPLOY_PIN   -1.000000 # Disabled
LGR_DEPLOY_POL   0.000000 # Low
LGR_ENABLE       1.000000 # Enabled
LGR_OPTIONS      2.000000 # Deploy during Land
LGR_RETRACT_ALT  20.000000
LGR_STARTUP      2.000000 # Deploy
LGR_WOW_PIN      -1.000000 # Disabled
LGR_WOW_POL      0.000000 # Low

gear
Also @IamPete1 concerns about it not retracting again on a touch and go don't occur

Hwurzburg: tthe only way behaviour diverges from master if LGR_STARTUP = 0 and the pilot does NOT deploy the gear before  setting it on the runway for takeoff via MAVLink with no AUX switch....but this is almost a certain crash, since the landing gear servo will be getting 0 PWM and have no torque...the 0 case , as Iunderstand it, was to prevent landing gear from changing position on the bench when powered up unexpectedly...even if the gear is down, it will potentially collapse on takeoff roll or sooner if the gear servo is not getting PWM...so this is a non-issue in my opinion...

Hwurzburg: added back have_changed

tridge (approved):

PR #29678, opened 2025-04-02
Tools: flake8 fixes for global keyword

peterbarker:
flake8 became more strict

eg.

Tools/autotest/sim_vehicle.py:911:9: F824 `global tracker_serial0` is unused: name is never assigned in scope
        global tracker_serial0


peterbarker:
sim-vehicle.py still works, and autotest is tested in CI

amilcarlucas (approved):

PR #29685, opened 2025-04-04
Plane: fixed terrain guided target intermediate alt handling

tridge:
introduces a new set_target_altitude_proportion_terrain() which handles the case of a terrain target destination with a source as absolute alt or terrain alt
This also removes (and bans) the use of terrain_alt=1 with relative_alt=0. That combination is not handled by the Location object but was used by plane. 

The fundamental problem is that in the Location object, if you have terrain_alt=1 and relative_alt=0 then the get_alt_m() function will get an altitude that is the height of terrain at that location plus the alt field. Plane has always assumed that in this case the altitude is the height of terrain at that location plus the alt field plus the home altitude. This is because we have a rule that when relative_alt=0 we assume that home.alt has been incorporated into the alt field. To square the circle we need to make this combination impossible, which is a major change in the plane code.


ArduPlane/altitude.cpp
```diff
@@ -321,6 +311,46 @@ void Plane::set_target_altitude_proportion(const Location &loc, float proportion
     }
 }
 
+#if AP_TERRAIN_AVAILABLE
+/*
+  change target altitude along a path between two locations
+  (prev_WP_loc and next_WP_loc) where the second location is a terrain
```
timtuxworth, line 317: What about the reverse? What if the prev_WP_loc is terrain and next_WP_loc is not?

timtuxworth:
Have you seen this @Ryanf55 ? Any thoughts?

ArduPlane/altitude.cpp
tridge: we handle that with the existing code. It won't follow terrain to get to the destination

ArduPlane/altitude.cpp
```diff
@@ -907,3 +937,24 @@ float Plane::get_landing_height(bool &rangefinder_active)
...
 }
+
+/*
+  if a terrain location doesn't have the relative_alt flag set
+  then fix the alt and trigger a flow of control error
+ */
+void Plane::fix_terrain_WP(Location &loc, uint32_t linenum)
+{
+    if (loc.terrain_alt && !loc.relative_alt) {
+        AP::internalerror().error(AP_InternalError::error_t::flow_of_control, linenum);
```
IamPete1, line 948: Probably worth a new enum value.

ArduPlane/altitude.cpp
```diff
@@ -907,3 +937,24 @@ float Plane::get_landing_height(bool &rangefinder_active)
 
     return height;
 }
+
+/*
+  if a terrain location doesn't have the relative_alt flag set
+  then fix the alt and trigger a flow of control error
+ */
+void Plane::fix_terrain_WP(Location &loc, uint32_t linenum)
```
IamPete1, line 945: I wonder if this should be on the location object? It can fetch home itself.

libraries/AP_AHRS/AP_AHRS_DCM.cpp
```diff
@@ -1064,14 +1064,14 @@ bool AP_AHRS_DCM::get_location(Location &loc) const
...
     const auto &baro = AP::baro();
     const auto &gps = AP::gps();
+    int32_t alt_cm;
     if (_gps_use == GPSUse::EnableWithHeight &&
         gps.status() >= AP_GPS::GPS_OK_FIX_3D) {
-        loc.alt = gps.location().alt;
+        alt_cm = gps.location().alt;
     } else {
-        loc.alt = baro.get_altitude() * 100 + AP::ahrs().get_home().alt;
+        alt_cm = baro.get_altitude() * 100 + AP::ahrs().get_home().alt;
```
IamPete1, line 1072: I know this is existing code, but I wonder if it should be `baro.get_altitude_AMSL()` The difference is home altitude vs the set field elevation.

ArduPlane/altitude.cpp
```diff
@@ -321,6 +311,46 @@ void Plane::set_target_altitude_proportion(const Location &loc, float proportion
...
+                                     prev_WP_height_terrain)) {
+        return false;
+    }
+    // and next_WP_loc alt as terrain
+    float next_WP_height_terrain;
+    if (!plane.next_WP_loc.get_alt_m(Location::AltFrame::ABOVE_TERRAIN,
+                                     next_WP_height_terrain)) {
+        return false;
+    }
+    const float p = constrain_float(plane.auto_state.wp_proportion, 0, 1);
```
IamPete1, line 341: `linear_interpolate` does the constrain for you too.

IamPete1 (commented):
Looks OK, this altitude stuff is always tricky, I the key thing is just to test each case.

tridge:
need to fix RockBlock.lua

ArduPlane/altitude.cpp
tridge: I think flow_of_control is right, we haven't yet had an issue finding the line, and I don't expect these will ever trigger

ArduPlane/altitude.cpp
tridge: I don't want to pollute the non-Plane code with this special case logic for plane

libraries/AP_AHRS/AP_AHRS_DCM.cpp
tridge: maybe in another PR, this shouldn't be changing behaviour

ArduPlane/altitude.cpp
tridge: thanks, changed

peterbarker (approved):
We flight-tested this today

Ryanf55:
> Have you seen this @Ryanf55 ? Any thoughts?

Id like to understand how it was flight tested, and whether there are more cases we want to test. It's not clear from Peter's comment what they tested, and Pete said we want to test each case. Does Peter want us to assume that he tested each case in flight? How do we know if we don't know what was tested.

If we have time: 
* Enumerate the test cases we want to test
* Run through them manually in SITl
* Or better yet, implement  the test cases in auto test. 

I'm happy to help with any of these in a week. 

I really like the premise of the PR, but I feel entirely unqualified to approve it without doing any of the above.

tridge:
@Ryanf55 it has had extensive testing in:
 - SITL with manual testing
 - autotest (see all the new tests added)
 - quadplane testing on a real vehicle, trying to trigger issues

The key cases are:
 - flying to a terrain alt target from a non-terrain target
 - flying between two terrain alt targets
 - flying between targets with increasing alt between
 - flying between targets with decreasing alt between
 - flying in a variety of modes (AUTO, LOITER, RTL, CRUISE etc)

PR #29687, opened 2025-04-04
Add missing RELAYn_FUNCTION values for VTX-contianing hwdefs

peterbarker:
... and the tool made to find and fix it, run like this:
```
reset; for f in `find libraries/AP_HAL_ChibiOS/hwdef -name hwdef.dat`; do echo "f=$f"; ./Tools/scripts/hack_hwdef_dat.py $f; done
```


Tools/scripts/hack_hwdef_dat.py
```diff
@@ -0,0 +1,85 @@
...
+
+class HackHWDef():
+    def __init__(self, filepath):
+        self.filepath = filepath
+
+    def run(self):
+        f = open(self.filepath, "r")
+        lines = f.readlines()
+
+        VTX_PWR_pin = None
```
andyp1per, line 20: CAM_SW is the same, needs to be included

libraries/AP_HAL_ChibiOS/hwdef/BROTHERHOBBYH743/hwdef.dat
```diff
@@ -158,6 +158,7 @@ PD2 SDMMC1_CMD SDMMC1
 # VTX Power control - should be high at startup to ensure power
 PD10 VTX_PWR OUTPUT HIGH GPIO(81)
 define RELAY2_PIN_DEFAULT 81
+define RELAY2_FUNCTION 1
```
andyp1per, line 161: My reading of the code is that this doesn't work, either needs changes to AP_Relay or needs to go in defaults.parm

andyp1per (changes requested):

Hwurzburg:
I am not a python guy, but there can be GPIOs defined without the VTX_PWR or CAM_SW labels but still have RELAY_PIN X assigned in the hwdef and used for power control or camera switching....I don't think the label is required in the hwdef to define a relay switchable GPIO

peterbarker:
@Hwurzburg has taken over this task with his own PR.

PR #29688, opened 2025-04-04
AP_InertialSensor: Fix temperature conversion for SCHA63T

tatsuy:
This change fixes the temperature conversion issue in the SCHA63T driver.
Previously, the temperature value was calculated using an unsigned data type.
However, since a sensor output of 0 corresponds to 35.5°C, values below that should yield a negative sensor output.
As a result, the temperature values were interpreted as large positive numbers (e.g., 1127.5°C).
Changing the data type from uint16_t to int16_t now correctly processes these negative sensor outputs.

Before
image

After
image


andyp1per (approved):
LGTM

PR #29693, opened 2025-04-04
Update install-prereqs-ubuntu.sh

dbstf:
Fix "python-argparse" to "python**3**-argparse" (note the "**3**")

### DIFF
--- a/Tools/environment_install/install-prereqs-ubuntu.sh
+++ b/Tools/environment_install/install-prereqs-ubuntu.sh
@@ -286,7 +286,11 @@ elif [ ${RELEASE_CODENAME} == 'lunar' ]; then
 elif [ ${RELEASE_CODENAME} != 'mantic' ] &&
      [ ${RELEASE_CODENAME} != 'noble' ] && 
      [ ${RELEASE_CODENAME} != 'oracular' ]; then
-    SITL_PKGS+=" python-argparse"
+    if apt-cache search python-argparse | grep argp; then
+        SITL_PKGS+=" python-argparse"
+    elif apt-cache search python3-argparse | grep argp; then
+        SITL_PKGS+=" python3-argparse"
+    fi
 fi
 
 # Check for graphical package for MAVProxy


Tools/environment_install/install-prereqs-ubuntu.sh
```diff
@@ -286,7 +286,7 @@ elif [ ${RELEASE_CODENAME} == 'lunar' ]; then
 elif [ ${RELEASE_CODENAME} != 'mantic' ] &&
      [ ${RELEASE_CODENAME} != 'noble' ] && 
      [ ${RELEASE_CODENAME} != 'oracular' ]; then
-    SITL_PKGS+=" python-argparse"
+    SITL_PKGS+=" python3-argparse"
```
peterbarker, line 289: I'm not sure this is something we want to be doing.  This code has worked on previous Ubuntu versions, and changing the package name might break that.

If this is breaking on more recent Ubuntus we probably want to add a clause for that version of Ubuntu.

dbstf: I installed the toolchain on Debian 12 (Bookworm), there the installation failed with "python-argparse" but worked with "python**3**-argparse". I know the script is geared towards Ubuntu and not Debian, but mostly they are "close enough".  Looking at the rest of the script (most packages starting with "python**3**-..."), this looks like an oversight, also mentioned in the script itself:
https://github.com/ArduPilot/ardupilot/blob/feff9dcf8e31cac7b0a1cbf41f3d743957fae432/Tools/environment_install/install-prereqs-ubuntu.sh#L82

dbstf: I added "apt-cache search" for python-argparse / python3-argparse. This should ensure wider compatibility instead of hard-coding it.

peterbarker (approved):
LGTM - thanks for changing to the search-for-package approach!

I've fixed the commit message and folded everything into one commit.

peterbarker:
Merged, thanks!

dbstf:
Happy tp help, thanks for merging!

PR #29695, opened 2025-04-04
SITL: removed PARA_ from whitelist

omarzayedbeshir:
I removed the PARA_ parameters from the whitelist since they were already documented.


peterbarker (approved):
LGTM, merging it - thanks!

PR #29696, opened 2025-04-04
Fix FW transition throttle glitch for tilting rotor quadplanes

priseborough:
This is a fix for the instant forward motor(s) throttle drop that can occur when a quadplane with tilting rotors completes the transition from VTOL to FW flight. This issue was reported here https://github.com/ArduPilot/ardupilot/issues/28957

This fix:

Extends the functionality of the set_throttle_min function such that the rate that the lower limit releases can be controlled by a TECS_THR_ERATE parameter and the throttle out previous value used by the TECS throttle rate limiter can also be optionally reset. If TECS_THR_ERATE is <=0 the legacy behaviour will be observed, ie the minimum throttle will be held for one TECS frame after the function is called.

Adds a function to the motors library that enables the raw normalised throttle demand to be obtained for a specified motor index.

Adds a function to the SRV_Channels library that enables the slew rate limiter state to be set for a specified channel function. 

Adds a function to the tiltrotor class that provides the mean throttle of all tilting motors.

When the transition into FW flight is ready to complete it checks for a throttle demand from tilting rotors or if not available the forward motor throttle demand. The TECS minimum throttle and SRV_Channels forward throttle are both set to this value.

Testing on a multirotor  VTOL with forrward tilting motors shows that the TECS throttle lower limit is set and releases at the specified 0.2 rate.

Screenshot 2025-04-04 at 9 58 10 PM

Screenshot 2025-04-04 at 9 57 40 PM

Screenshot 2025-04-04 at 9 57 00 PM

Here's a before and after for a SITL test using the -v Plane -f quadplane-tilthvec option

Before - the throttles of the tilting motors step to 0 before increasing
Screenshot 2025-04-07 at 8 06 41 AM

After - the throttles of the tilting motors ramp down smoothly
Screenshot 2025-04-07 at 8 04 35 AM

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -1478,6 +1478,14 @@ void SLT_Transition::update()
     // the tilt will decrease rapidly)
     if (quadplane.tiltrotor.fully_fwd() && transition_state != TRANSITION_AIRSPEED_WAIT) {
         if (transition_state == TRANSITION_TIMER) {
+            float throttle;
+            if (plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                // Reset the TECS minimum throttle to match throttle of forward thrust motors
+                // and set the throttle channel slew rate limiter to prevent a sudden drop in throttle
+                plane.TECS_controller.set_throttle_min(throttle, true);
+                SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle * 100);
+                SRV_Channels::set_output_scaled(SRV_Channel::k_throttle, throttle * 100);
+            }
             gcs().send_text(MAV_SEVERITY_INFO, "Transition FW done");
         }
         transition_state = TRANSITION_DONE;
@@ -1584,15 +1592,24 @@ void SLT_Transition::update()
         
     case TRANSITION_TIMER: {
         quadplane.set_desired_spool_state(AP_Motors::DesiredSpoolState::THROTTLE_UNLIMITED);
-        // after airspeed is reached we degrade throttle over the
-        // transition time, but continue to stabilize
+        // after airspeed is reached we degrade throttle over the transition time, but continue
+        // to stabilize and wait for any required forward tilt to complete and the timer to expire
         const uint32_t transition_timer_ms = now - transition_low_airspeed_ms;
         const float trans_time_ms = constrain_float(quadplane.transition_time_ms,500,30000);
-        if (transition_timer_ms > unsigned(trans_time_ms)) {
+        const bool tilt_fwd_complete = !quadplane.tiltrotor.enabled() || quadplane.tiltrotor.tilt_angle_achieved();
+        if (transition_timer_ms > unsigned(trans_time_ms) && tilt_fwd_complete) {
             transition_state = TRANSITION_DONE;
             in_forced_transition = false;
             transition_start_ms = 0;
             transition_low_airspeed_ms = 0;
+            float throttle;
+            if (plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                // Reset the TECS minimum throttle to match throttle of forward thrust motors
+                // and set the throttle channel slew rate limiter to prevent a sudden drop in throttle
+                plane.TECS_controller.set_throttle_min(throttle, true);
+                SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle * 100);
+                SRV_Channels::set_output_scaled(SRV_Channel::k_throttle, throttle * 100);
+            }
             gcs().send_text(MAV_SEVERITY_INFO, "Transition done");
         }
 
--- a/ArduPlane/tiltrotor.cpp
+++ b/ArduPlane/tiltrotor.cpp
@@ -782,4 +782,32 @@ bool Tiltrotor::tilt_over_max_angle(void) const
     return (current_tilt > MIN(tilt_threshold, get_forward_flight_tilt()));
 }
 
+// throttle of forward flight motors including any tilting motors
+bool Tiltrotor::get_forward_throttle(float &throttle) const
+{
+    if (!enabled() || !_is_vectored) {
+        return false;
+    }
+    const float throttle_range = motors->thr_lin.get_spin_max() - motors->thr_lin.get_spin_min();
+    if (!is_positive(throttle_range)) {
+        return false;
+    }
+    float throttle_sum = 0.0f;
+    uint8_t num_vectored_motors = 0;
+    for (uint8_t i = 0; i < AP_MOTORS_MAX_NUM_MOTORS; ++i) {
+        if (is_motor_tilting(i)) {
+            float thrust;
+            if (motors->get_thrust(i, thrust)) {
+                throttle_sum += (motors->thr_lin.thrust_to_actuator(thrust) - motors->thr_lin.get_spin_min()) / throttle_range;
+                num_vectored_motors ++;
+            }
+        }
+    }
+    if (num_vectored_motors > 0) {
+        throttle = throttle_sum / (float)num_vectored_motors;
+        return true;
+    }
+    return false;
+}
+
 #endif  // HAL_QUADPLANE_ENABLED
--- a/libraries/AP_TECS/AP_TECS.cpp
+++ b/libraries/AP_TECS/AP_TECS.cpp
@@ -264,7 +264,15 @@ const AP_Param::GroupInfo AP_TECS::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("PTCH_FF_K", 30, AP_TECS, _pitch_ff_k, 0.0),
 
-    // 31 previously used by TECS_LAND_PTRIM
+    // 31 previously used by AP_Int8 TECS_LAND_PTRIM which was removed in November 2022
+
+    // @Param: THR_ERATE
+    // @DisplayName: Forward throttle external limit slew rate
+    // @Description: The externally set forward throttle lower limit applied within TECS will be reduced by this many percentage points per second after being set. Set to a non positive value to hold the lower limit for one frame only.
+    // @Units: %/s
+    // @Range: 0 100
+    // @User: Advanced
+    AP_GROUPINFO("THR_ERATE", 31, AP_TECS, _thr_min_pct_ext_rate_lim, 20),
 
     // @Param: FLARE_HGT
     // @DisplayName: Flare holdoff height
@@ -1378,11 +1386,16 @@ void AP_TECS::update_pitch_throttle(int32_t hgt_dem_cm,
 }
 
 // set minimum throttle override, [-1, -1] range
-// it is applicable for one control cycle only
-void AP_TECS::set_throttle_min(const float thr_min) {
+// if reset_output is true the output slew limiter will also be reset to respect the lower limit
+// its decay is controlled by TECS_THR_ERATE
+void AP_TECS::set_throttle_min(const float thr_min, bool reset_output) {
     // Don't change the limit if it is already covered.
     if (thr_min > _THRminf_ext) {
         _THRminf_ext = thr_min;
+        if (reset_output) {
+            _last_throttle_dem = MAX(_last_throttle_dem, _THRminf_ext);
+            _throttle_dem = _last_throttle_dem;
+        }
     }
 }
 
@@ -1417,7 +1430,12 @@ void AP_TECS::_update_throttle_limits() {
     
     // Reset the external throttle limits.
     // Caller will have to reset them in the next iteration.
-    _THRminf_ext = -1.0f;
+    if (_thr_min_pct_ext_rate_lim > 0) {
+        _THRminf_ext -= 0.01f * float(_thr_min_pct_ext_rate_lim) * _DT;
+        _THRminf_ext = MAX(_THRminf_ext, -1.0f);
+    } else {
+        _THRminf_ext = -1.0f;
+    }
     _THRmaxf_ext = 1.0f;
 }
 
--- a/libraries/SRV_Channel/SRV_Channel_aux.cpp
+++ b/libraries/SRV_Channel/SRV_Channel_aux.cpp
@@ -822,6 +822,18 @@ void SRV_Channels::set_slew_rate(SRV_Channel::Function function, float slew_rate
     _slew = new_slew;
 }
 
+// update channels last_scaled_output to match value
+void SRV_Channels::set_slew_last_scaled_output(SRV_Channel::Function function, float value)
+{
+    for (slew_list *slew = _slew; slew; slew = slew->next) {
+        if (slew->func == function) {
+            // found existing item, update slew limiter value
+            slew->last_scaled_output = value;
+            return;
+        }
+    }
+}
+
 // call set_angle() on matching channels
 void SRV_Channels::set_angle(SRV_Channel::Function function, uint16_t angle)
 {
--- a/ArduPlane/tiltrotor.h
+++ b/ArduPlane/tiltrotor.h
@@ -70,6 +70,9 @@ public:
     // always return true if not enabled or not a continuous type
     bool tilt_angle_achieved() const { return !enabled() || (type != TILT_TYPE_CONTINUOUS) || angle_achieved; }
 
+    // throttle of tilting motors used for forward flight
+    bool get_forward_throttle(float &throttle) const;
+
     // Write tiltrotor specific log
     void write_log();
 
--- a/libraries/AP_Motors/AP_MotorsMulticopter.cpp
+++ b/libraries/AP_Motors/AP_MotorsMulticopter.cpp
@@ -830,6 +830,17 @@ bool AP_MotorsMulticopter::arming_checks(size_t buflen, char *buffer) const
     return true;
 }
 
+// return raw throttle out fraction for motor motor_num, returns true if value is valid, false otherwise
+bool AP_MotorsMulticopter::get_raw_motor_throttle(uint8_t motor_num, float& thr_out) const
+{
+    if (motor_num >= AP_MOTORS_MAX_NUM_MOTORS || !motor_enabled[motor_num]) {
+        return false;
+    }
+
+    thr_out = constrain_float(_actuator[motor_num], 0.0f, 1.0f);
+    return true;
+}
+
 #if APM_BUILD_TYPE(APM_BUILD_UNKNOWN)
 // Getters for AP_Motors example, not used by vehicles
 float AP_MotorsMulticopter::get_throttle_avg_max() const
--- a/libraries/AP_Motors/AP_MotorsMulticopter.h
+++ b/libraries/AP_Motors/AP_MotorsMulticopter.h
@@ -92,6 +92,8 @@ public:
     // return thrust for motor motor_num, returns true if value is valid false otherwise
     bool                get_thrust(uint8_t motor_num, float& thr_out) const override;
 
+    bool                get_raw_motor_throttle(uint8_t motor_num, float& thr_out) const override;
+
 #if HAL_LOGGING_ENABLED
     // 10hz logging of voltage scaling and max trust
     void                Log_Write() override;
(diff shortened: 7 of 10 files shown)


priseborough:
@tridge - I need help with setting up compiler directives to manage flash space

Georacer (changes requested):

ArduPlane/tiltrotor.cpp
```diff
@@ -782,4 +782,27 @@ bool Tiltrotor::tilt_over_max_angle(void) const
...
+// throttle of forward flight motors including any tilting motors
+bool Tiltrotor::get_forward_throttle(float &throttle) const
+{
+    if (_is_vectored) {
+        float throttle_sum = 0.0f;
+        uint8_t num_vectored_motors = 0;
+        for (uint8_t i = 0; i < AP_MOTORS_MAX_NUM_MOTORS; ++i) {
+            if (is_motor_tilting(i)) {
+                float thr_out;
+                if (motors->get_raw_motor_throttle(i, thr_out)) {
```
tridge, line 794: I think we can avoid the addition of the get_raw_motor_throttle() function by adding a variable that adds up the thrust on tilting motors in Tiltrotor::tilt_compensate_angle()

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -264,7 +264,13 @@ const AP_Param::GroupInfo AP_TECS::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("PTCH_FF_K", 30, AP_TECS, _pitch_ff_k, 0.0),
 
-    // 31 previously used by TECS_LAND_PTRIM
```
tridge, line 267: re-using slots like this is OK when they change type, and TECS_LAND_PTRIM was an AP_Int8, so this works, but we should keep a comment to say it was an AP_Int8, probably with a comment that TECS_LAND_PTRIM was removed in November 2022.

tridge (changes requested):
thanks for this @priseborough !

tridge:
@priseborough also please note this PR which allows the existing throttle slew limit to apply to tilted motors:
https://github.com/ArduPilot/ardupilot/pull/28960

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -1358,11 +1364,15 @@ void AP_TECS::update_pitch_throttle(int32_t hgt_dem_cm,
...
-void AP_TECS::set_throttle_min(const float thr_min) {
+// if reset_output is true the output slew limiter will also be reset to respect the lower limit
+// its decay is controlled by TECS_THR_ERATE
+void AP_TECS::set_throttle_min(const float thr_min, bool reset_output) {
     // Don't change the limit if it is already covered.
     if (thr_min > _THRminf_ext) {
         _THRminf_ext = thr_min;
+        if (reset_output) {
+            _throttle_dem = MAX(_throttle_dem, _THRminf_ext);
+        }
```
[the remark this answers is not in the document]
priseborough: I meant to reset _last_throttle_dem

ArduPlane/tiltrotor.cpp
priseborough: I wanted the raw throttle value for each motor, not the thrust value because the throttle  value is equivalent to what the FW controller needs to demand to achieve the same PWM output whereas the the thrust value is related to PWM via a nonlinear function.

priseborough:
> @priseborough also please note this PR which allows the existing throttle slew limit to apply to tilted motors: #28960 we may want both mechanisms. The advantage of the other PR is it works for non-TECS modes

I think we need both. That PR deals with issues before transition completes, this one with what was happening immediately after for tiltrotor setups. Note that the reset of servo throttle channel slew limiters should also help for non TECS modes after transition completes.

libraries/AP_TECS/AP_TECS.cpp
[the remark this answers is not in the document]
Georacer: Ah, yes, that makes more sense. Thanks.

libraries/SRV_Channel/SRV_Channel_aux.cpp
```diff
@@ -819,6 +819,18 @@ void SRV_Channels::set_slew_rate(SRV_Channel::Function function, float slew_rate
     _slew = new_slew;
 }
 
+// update channels last_scaled_output to match value
+void SRV_Channels::set_slew_last_scaled_output(SRV_Channel::Function function, float value)
```
IamPete1, line 823: Maybe "reset_slew_limit", that is a little more inline with what we would call it for the PIDs. It would set the current and last output in one call.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -264,7 +264,15 @@ const AP_Param::GroupInfo AP_TECS::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("PTCH_FF_K", 30, AP_TECS, _pitch_ff_k, 0.0),
 
-    // 31 previously used by TECS_LAND_PTRIM
+    // 31 previously used by AP_Int8 TECS_LAND_PTRIM which was removed in November 2022
+
+    // @Param: THR_ERATE
```
IamPete1, line 269: Do we need a new param? Can't we use the existing plane throttle slew limit?

ArduPlane/tiltrotor.cpp
```diff
@@ -782,4 +782,27 @@ bool Tiltrotor::tilt_over_max_angle(void) const
     return (current_tilt > MIN(tilt_threshold, get_forward_flight_tilt()));
 }
 
+// throttle of forward flight motors including any tilting motors
+bool Tiltrotor::get_forward_throttle(float &throttle) const
+{
+    if (_is_vectored) {
```
IamPete1, line 788:
```suggestion
    if (!_is_vectored) {
        return false;
    }
```

ArduPlane/tiltrotor.cpp
IamPete1: We have the same issue for tailsitters there we apply the inverse of the linearisation. 

https://github.com/ArduPilot/ardupilot/blob/0ecdae05ba028a06cf939ad294f9d41d5df2dcb1/ArduPlane/tailsitter.cpp#L313-L314

The `actuator_to_thrust` and `thrust_to_actuator` functions allow you to go back and forth. However, you would still need to filter by tilting/not tilting.

IamPete1 (commented):
This looks good. https://github.com/ArduPilot/ardupilot/pull/29011 Would handle the smoothing of the throttle outputs. But you would still need to workout the average throttle for the TECS re-set.

libraries/AP_TECS/AP_TECS.cpp
priseborough: It needs to be slower based on flight testing where the tilting rotors were energising the front wing in a tandem wing configuration.

ArduPlane/tiltrotor.cpp
priseborough: I've pushed a patch that uses the thrust_to_actuator function and reverted the patch that added the additional get_raw_motor_throttle function which should save on some flash space.

ArduPlane/tiltrotor.cpp
priseborough: Done

libraries/AP_TECS/AP_TECS.cpp
priseborough: I've update comments to make prior use clear.

priseborough:
In the process of cleaning it up and updating in response to review comments a single frame glitch in the servo demands has been introduced.

priseborough:
The reason for the glitch in servo demand after incorporating review comments was  https://github.com/ArduPilot/ardupilot/pull/29696/commits/af73ffb173dbb3efcda00147402cb23d739d1a42

The current throttle state also has to be reset to avoid a glitch if the TECS throttle demand is fetched before the next TECS update cycle.

I've added a patch that adds the set of current throttle state.

priseborough:
@IamPete1 and @tridge using the conversion form thrust to throttle applied to thrust demand is not exactly the same as retrieving the raw throttle demand.

If I revert commits https://github.com/ArduPilot/ardupilot/pull/29696/commits/3ad8f324eaadf7935ebed589042aaea70ce40b52 and https://github.com/ArduPilot/ardupilot/pull/29696/commits/8d227d92d3e9717117cb88f5d05fa4e1627ff85e I get the following ESC demands for vectoring quad SITL test

Screenshot 2025-04-13 at 2 25 54 PM

Using the thrust to throttle scaling I get

Screenshot 2025-04-13 at 2 23 22 PM

so unless we are really up against it on flash space I would prefer to use the raw throttle.

priseborough:
It has been flight tested on a ARACE tiltrotor at commit https://github.com/ArduPilot/ardupilot/pull/29696/commits/1204cba00904654f967d6b4155df3586730ca72c

Here are a couple of the forward transitions from the flight:

Transition in AUTO - no sudden dips in forward throttle on completion
Screenshot 2025-04-24 at 5 31 00 PM
Screenshot 2025-04-24 at 5 31 06 PM

Transition from QLOITER to  CRUISE - no sudden dips in forward throttle on completion
Screenshot 2025-04-24 at 5 32 08 PM
Screenshot 2025-04-24 at 5 32 21 PM

ArduPlane/quadplane.cpp
```diff
@@ -1478,6 +1478,15 @@ void SLT_Transition::update()
     // the tilt will decrease rapidly)
     if (quadplane.tiltrotor.fully_fwd() && transition_state != TRANSITION_AIRSPEED_WAIT) {
         if (transition_state == TRANSITION_TIMER) {
+            float throttle;
+            if (!plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                throttle = 0.01f * (float)plane.aparm.throttle_cruise;
```
IamPete1, line 1483: 0.01 and then setting it back on 1489.

ArduPlane/quadplane.cpp
```diff
@@ -1584,15 +1593,25 @@ void SLT_Transition::update()
...
             transition_low_airspeed_ms = 0;
+            float throttle;
+            if (!plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                throttle = 0.01f * SRV_Channels::get_output_scaled(SRV_Channel::k_throttle);
+            }
+            // Reset the TECS minimum throttle to match throttle of forward thrust motors
+            // and set the throttle channel slew rate limiter to prevent a sudden drop in throttle
+            plane.TECS_controller.set_throttle_min(throttle, true);
+            SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle);
+            SRV_Channels::set_output_scaled(SRV_Channel::k_throttle, throttle);
```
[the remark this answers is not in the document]
priseborough: Not set for the same reason as before.

ArduPlane/tiltrotor.cpp
```diff
@@ -782,4 +782,29 @@ bool Tiltrotor::tilt_over_max_angle(void) const
...
+        return false;
+    }
+    float throttle_sum = 0.0f;
+    uint8_t num_vectored_motors = 0;
+    for (uint8_t i = 0; i < AP_MOTORS_MAX_NUM_MOTORS; ++i) {
+        if (is_motor_tilting(i)) {
+            float thrust;
+            const float throttle_range = motors->thr_lin.get_spin_max() - motors->thr_lin.get_spin_min();
+            if (motors->get_thrust(i, thrust) && is_positive(throttle_range)) {
+                throttle_sum += (motors->thr_lin.thrust_to_actuator(thrust) - motors->thr_lin.get_spin_min()) / throttle_range;
```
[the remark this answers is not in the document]
priseborough: I'm OK to do it this way. Its only calculated on one frame during completion of transition so the computational inefficiency is not an issue.

ArduPlane/quadplane.cpp
```diff
@@ -1584,15 +1593,25 @@ void SLT_Transition::update()
...
+        if (transition_timer_ms > unsigned(trans_time_ms) && tilt_fwd_complete) {
             transition_state = TRANSITION_DONE;
             in_forced_transition = false;
             transition_start_ms = 0;
             transition_low_airspeed_ms = 0;
+            float throttle;
+            if (!plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                throttle = 0.01f * SRV_Channels::get_output_scaled(SRV_Channel::k_throttle);
+            }
+            // Reset the TECS minimum throttle to match throttle of forward thrust motors
```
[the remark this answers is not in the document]
priseborough: It's pointless given the prior forward throttle is from TECS anyway so I will rework to only run that code path when it is a tiltrotor.

ArduPlane/tiltrotor.cpp
```diff
@@ -782,4 +782,29 @@ bool Tiltrotor::tilt_over_max_angle(void) const
...
+{
+    if (!_is_vectored) {
+        return false;
+    }
+    float throttle_sum = 0.0f;
+    uint8_t num_vectored_motors = 0;
+    for (uint8_t i = 0; i < AP_MOTORS_MAX_NUM_MOTORS; ++i) {
+        if (is_motor_tilting(i)) {
+            float thrust;
+            const float throttle_range = motors->thr_lin.get_spin_max() - motors->thr_lin.get_spin_min();
```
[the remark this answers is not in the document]
priseborough: Done.

ArduPlane/quadplane.cpp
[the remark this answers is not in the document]
priseborough: Done.

ArduPlane/quadplane.cpp
```diff
@@ -1478,6 +1478,15 @@ void SLT_Transition::update()
...
         if (transition_state == TRANSITION_TIMER) {
+            float throttle;
+            if (!plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                throttle = 0.01f * (float)plane.aparm.throttle_cruise;
+            }
+            // Reset the TECS minimum throttle to match throttle of forward thrust motors
+            // and set the throttle channel slew rate limiter to prevent a sudden drop in throttle
+            plane.TECS_controller.set_throttle_min(throttle, true);
+            SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle);
+            SRV_Channels::set_output_scaled(SRV_Channel::k_throttle, throttle);
```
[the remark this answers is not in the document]
IamPete1: Your correct for the `set_output_scaled` but not the `set_slew_last_scaled_output`. Setting it to just throttle would ignore rudder of course, but that might be better than what it does now.

priseborough:
I've rebased and fixed the conflicts by removing an unnecessary reverted logging commit in TECS.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -264,7 +264,15 @@ const AP_Param::GroupInfo AP_TECS::var_info[] = {
...
+    // @DisplayName: Forward throttle external limit slew rate
+    // @Description: The forward throttle lower limit controlled by TECS when set externally will be reduced at this rate. Set to a non positive value to hold the lower limit for one frame only
+    // @Units: ms
```
tridge, line 272: I don't think it is in milliseconds, it is proportion per second. So the default of 0.2 means decay over 5 seconds. I think we either need a better description, or make it in seconds and call it TECS_THR_DECTIME for "decay time".

tridge (changes requested):
looks good, thanks!
just a small update to the new TECS parameter. Let me know if you'd like me to make and test that change

libraries/AP_TECS/AP_TECS.cpp
priseborough: I'll change it to percentage per second and update the description.

IamPete1 (changes requested):
With the throttle scale corrected it seems to work as expected. This is Qloiter to FBWB.

image

Throttle min is set to the current throttle then it decays back at the default 20% per second. High throttle in this case so it takes nearly the full 5 seconds to get down to 0. In this case the output throttle was soon constrained on that limit,

tridge (approved):

ArduPlane/quadplane.cpp
```diff
@@ -1478,6 +1478,14 @@ void SLT_Transition::update()
     // the tilt will decrease rapidly)
     if (quadplane.tiltrotor.fully_fwd() && transition_state != TRANSITION_AIRSPEED_WAIT) {
         if (transition_state == TRANSITION_TIMER) {
+            float throttle;
+            if (plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                // Reset the TECS minimum throttle to match throttle of forward thrust motors
+                // and set the throttle channel slew rate limiter to prevent a sudden drop in throttle
+                plane.TECS_controller.set_throttle_min(throttle, true);
+                SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle);
+                SRV_Channels::set_output_scaled(SRV_Channel::k_throttle, throttle);
```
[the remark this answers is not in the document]
tridge: fixed

ArduPlane/quadplane.cpp
```diff
@@ -1584,15 +1592,24 @@ void SLT_Transition::update()
...
             in_forced_transition = false;
             transition_start_ms = 0;
             transition_low_airspeed_ms = 0;
+            float throttle;
+            if (plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                // Reset the TECS minimum throttle to match throttle of forward thrust motors
+                // and set the throttle channel slew rate limiter to prevent a sudden drop in throttle
+                plane.TECS_controller.set_throttle_min(throttle, true);
+                SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle);
+                SRV_Channels::set_output_scaled(SRV_Channel::k_throttle, throttle);
```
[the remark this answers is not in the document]
tridge: fixed

tridge:
@IamPete1 @priseborough I've pushed the fixes and squashed

tridge:
looks good in RealFlight testing

ArduPlane/quadplane.cpp
```diff
@@ -1584,15 +1592,24 @@ void SLT_Transition::update()
...
             transition_state = TRANSITION_DONE;
             in_forced_transition = false;
             transition_start_ms = 0;
             transition_low_airspeed_ms = 0;
+            float throttle;
+            if (plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                // Reset the TECS minimum throttle to match throttle of forward thrust motors
+                // and set the throttle channel slew rate limiter to prevent a sudden drop in throttle
+                plane.TECS_controller.set_throttle_min(throttle, true);
+                SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle);
```
IamPete1, line 1610:
```suggestion
                SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle * 100);
```

ArduPlane/quadplane.cpp
```diff
@@ -1478,6 +1478,14 @@ void SLT_Transition::update()
     // the tilt will decrease rapidly)
     if (quadplane.tiltrotor.fully_fwd() && transition_state != TRANSITION_AIRSPEED_WAIT) {
         if (transition_state == TRANSITION_TIMER) {
+            float throttle;
+            if (plane.quadplane.tiltrotor.get_forward_throttle(throttle)) {
+                // Reset the TECS minimum throttle to match throttle of forward thrust motors
+                // and set the throttle channel slew rate limiter to prevent a sudden drop in throttle
+                plane.TECS_controller.set_throttle_min(throttle, true);
+                SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle);
```
IamPete1, line 1486:
```suggestion
                SRV_Channels::set_slew_last_scaled_output(SRV_Channel::k_throttle, throttle * 100);
```

IamPete1 (changes requested):

ArduPlane/quadplane.cpp
tridge: fixed

ArduPlane/quadplane.cpp
tridge: fixed

IamPete1 (approved):
LGTM

PR #29698, opened 2025-04-05
SITL: removed TA_ from whitelist

omarzayedbeshir:
I removed a TA_ parameter from the whitelist since it was already documented.


peterbarker:
Merged, thanks!

We'd usually say `Tools: vehicle_test_suite: ...`, BTW.  Near enough in this case :-)

Issue #29660, opened 2025-04-01
type infererence to int for AP_xxx types error


tridge:
example PR:
https://github.com/ArduPilot/ardupilot/pull/29624

likely also affected:
 - get_minimum_board_voltage() and related fns or maybe get() makes it ok?

tpwrules:
Minimum reproducible example:
```c++
#include <cstdio>

template<typename T> class fakeval 
{
public:
    void set(const T &v) {
        _value = v;
    }
    operator const T &() const {
        return _value;
    }
protected:
    T _value;
};

int main(int argc, char** argv) {
    fakeval<float> ff;
    ff.set(0.9);

    float val = argc > 0 ? ff : 0;
    printf("hi %f\n", val);
}

```

tpwrules:
I have workshopped some solutions with some very kind folks but it will be a bit before I can clean up and put together a PR.

tpwrules:
This is the relatively complete set of affected float casts. I was able to detect them by triggering compiler errors using an older and much worse solution to the problem.

It's possible to cause the same problem with integers (`int16_t v = true ? int16_param : (int8_t)0;`) but I choose to believe it's much less likely to be a problem. Also it's more annoying to fix.

```diff

---
 ArduPlane/Attitude.cpp                       | 2 +-
 ArduPlane/commands_logic.cpp                 | 2 +-
 ArduPlane/pullup.cpp                         | 2 +-
 ArduPlane/quadplane.cpp                      | 6 +++---
 Rover/mode_smart_rtl.cpp                     | 2 +-
 libraries/AP_Camera/AP_Camera_Backend.h      | 4 ++--
 libraries/AP_ESC_Telem/AP_ESC_Telem_SITL.cpp | 2 +-
 libraries/AP_Quicktune/AP_Quicktune.cpp      | 2 +-
 8 files changed, 11 insertions(+), 11 deletions(-)

diff --git a/ArduPlane/Attitude.cpp b/ArduPlane/Attitude.cpp
index e7f5953bb4..349c3ff706 100644
--- a/ArduPlane/Attitude.cpp
+++ b/ArduPlane/Attitude.cpp
@@ -664,7 +664,7 @@ void Plane::update_load_factor(void)
 #endif
 
     float stall_airspeed_1g = is_positive(aparm.airspeed_stall)
-                                  ? aparm.airspeed_stall
+                                  ? (float)aparm.airspeed_stall
                                   : aparm.airspeed_min;
 
[...]

PR #29703, opened 2025-04-06
AP_Math: split `radians` into float and double variants

tpwrules:
Most `radians` uses give a float argument and use a float return type. On `HAL_WITH_EKF_DOUBLE` boards this results in a bunch of unnecessary conversions. Create separate double and float variants of the function to eliminate the impact. The `DEG_TO_RAD` constant is already a float, therefore there is no loss of precision in the operation.

We also need a variant that takes an `int` argument as these are supplied in several places, primarily as constants, and the `int` conversion is otherwise ambiguous.

Saves a few hundred bytes of flash on such boards.

Prerequisite for candidate fix of #29660 .

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -889,7 +889,7 @@ void Copter::update_super_simple_bearing(bool force_update)
     }
 
     super_simple_last_bearing = bearing;
-    const float angle_rad = radians((super_simple_last_bearing+18000)/100);
+    const float angle_rad = radians((super_simple_last_bearing+18000)*0.01f);
     super_simple_cos_yaw = cosf(angle_rad);
     super_simple_sin_yaw = sinf(angle_rad);
 }
--- a/libraries/AP_Math/AP_Math.h
+++ b/libraries/AP_Math/AP_Math.h
@@ -212,7 +212,17 @@ inline double constrain_double(const double amt, const double low, const double 
 }
 
 // degrees -> radians
-static inline constexpr ftype radians(ftype deg)
+static inline constexpr double radians(double deg)
+{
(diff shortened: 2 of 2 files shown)


libraries/AP_Math/AP_Math.h
```diff
@@ -212,7 +212,17 @@ inline double constrain_double(const double amt, const double low, const double
 }
 
 // degrees -> radians
-static inline constexpr ftype radians(ftype deg)
+static inline constexpr double radians(double deg)
```
Ryanf55, line 215:
```suggestion
static inline constexpr double radians(double deg) WARN_IF_UNUSED
```
Prevent bugs?

tpwrules: I don't understand the value, e.g. `sin` is not tagged this way and we have survived this far without that.

Ryanf55: No runtime cost and prevent bugs. There's no downside other than having to read it in the header.

ArduCopter/Copter.cpp
```diff
@@ -889,7 +889,7 @@ void Copter::update_super_simple_bearing(bool force_update)
     }
 
     super_simple_last_bearing = bearing;
-    const float angle_rad = radians((super_simple_last_bearing+18000)/100);
+    const float angle_rad = radians((float)((super_simple_last_bearing+18000)/100));
```
tridge, line 892: I think * 0.01 would be better and fix an actual rounding bug

tridge (changes requested):

ArduCopter/Copter.cpp
tpwrules: Done, thanks.

PR #29704, opened 2025-04-06
Added Linux Mint translations for Ubuntu Jammy and Noble

tejashah88:
As of checking Linux Mint's releases page, there were the following versions of Linux Mint missing translations, which I've appended onto the install script:

| Linux Mint      | Ubuntu        |
|-----------------|---------------|
| Xia (22.1)      | Noble (24.XX) |
| Vera (21.1)     | Jammy (22.XX) |
| Victoria (21.2) | Jammy (22.XX) |
| Virginia (21.3) | Jammy (22.XX) |

For context, I was on Linux Mint Xia, and not having this translation causes the script to assume the release codename of "bullseye", which yields the error "Unable to locate package python-argparse".

As a side-note, I think it would help to add some insight as to why it should default to that specific Debian release instead of fast-failing outright. I understand that there are many flavors of Linux out there but at least a warning or a way to specify a custom release codename would help with installation problems like above.

### DIFF
--- a/Tools/environment_install/install-prereqs-ubuntu.sh
+++ b/Tools/environment_install/install-prereqs-ubuntu.sh
@@ -62,10 +62,10 @@ RELEASE_CODENAME=$(lsb_release -c -s)
 
 # translate Mint-codenames to Ubuntu-codenames based on https://www.linuxmint.com/download_all.php
 case ${RELEASE_CODENAME} in
-    wilma)
+    wilma | xia)
         RELEASE_CODENAME='noble'
         ;;
-    vanessa)
+    vanessa | vera | victoria | virginia)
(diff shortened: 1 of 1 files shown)


peterbarker:
I've touched up the commit message and marked for merge, thanks!

Quite happy to entertain patches for fixing that defaulting you're seeing - but they do have to not break that Debian release :-)

peterbarker:
Merged this, thanks.  @tejashah88 just a note - you should create a "topic branch" for this sort of thing rather than using your master branch.  Our Wiki has stuff on that, as does google.

tejashah88:
Thanks @peterbarker! I'll raise an issue to discuss more about handling Linux variants.

For sure, I'll keep that in mind if I make more future contributions.

PR #29705, opened 2025-04-06
AP_Scripting: added qnh_alt example

tridge:
useful for BVLOS ops. The one in AFS is only available if AFS is enabled

### DIFF
--- a/libraries/AP_Scripting/examples/qnh_alt.lua
+++ b/libraries/AP_Scripting/examples/qnh_alt.lua
@@ -1,0 +1,51 @@
+--[[
+    example that sends QNH altitude to GCS as a NAMED_VALUE_FLOAT QNG_ALT_FT in feet
+
+    Note: operator must set QNH_PRESSURE in hPa before each flight!
+--]]
+
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+
+local PARAM_TABLE_KEY = 93
+local PARAM_TABLE_PREFIX = "QNH_"
+
+-- add a parameter and bind it to a variable
+function bind_add_param(name, idx, default_value)
+   assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), string.format('could not add param %s', name))
+   return Parameter(PARAM_TABLE_PREFIX .. name)
+end
+
+-- setup script specific parameters
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 8), 'could not add param table')
+
+--[[
+  // @Param: QNH_PRESSURE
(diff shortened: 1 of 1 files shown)


libraries/AP_Scripting/examples/qnh_alt.lua
```diff
@@ -0,0 +1,51 @@
+--[[
+    example that sends QNH altitude to GCS as a NAMED_VALUE_FLOAT QNG_ALT_FT in feet
+
+    Note: operator must be QNH_PRESSURE in hPa before each flight!
```
timtuxworth, line 4: must "set"

libraries/AP_Scripting/examples/qnh_alt.lua
```diff
@@ -0,0 +1,51 @@
...
+   return Parameter(PARAM_TABLE_PREFIX .. name)
+end
+
+-- setup script specific parameters
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 8), 'could not add param table')
+
+--[[
+  // @Param: QNH_PRESSURE
+  // @DisplayName: QNH pressure in hPa
+  // @Description: QNH pressure in hPa
```
timtuxworth, line 24: at the home/takeoff location?

libraries/AP_Scripting/examples/qnh_alt.lua
```diff
@@ -0,0 +1,51 @@
+--[[
+    example that sends QNH altitude to GCS as a NAMED_VALUE_FLOAT QNG_ALT_FT in feet
```
timtuxworth, line 2: why feet? maybe explain that since we usually use SI for everything.

timtuxworth (changes requested):
Nice clear example of sending a named float. A couple of suggestions re comments and a typo (in a comment).

libraries/AP_Scripting/examples/qnh_alt.lua
tridge: QNH is defined by aviation standards, and provided by the aviation regulator

libraries/AP_Scripting/examples/qnh_alt.lua
tridge: because this is for regulatory compliance, and aviation altitudes are given in feet

libraries/AP_Scripting/examples/qnh_alt.lua
timtuxworth: I was suggesting that you add that to the description.

libraries/AP_Scripting/examples/qnh_alt.lua
timtuxworth: s/be "Note: operator must **set** QNH_PRESSURE in hPa before each flight!"

tridge: fixed

peterbarker (approved):

PR #29709, opened 2025-04-06
hwdef: add relay function 1 to all defs with realyx_pin_default defined

Hwurzburg:
Add relay function = 1 to match relay pin defs in hwdefs so that they work by default as expected by user


IamPete1:
I can't replicate this, if I set the pin default on a clean build the function is set to 1 automatically. I don't have one of the boards changed here handy but I added the pin defines to a board I do have. 

This is the bit of code that saves the function: https://github.com/ArduPilot/ardupilot/blob/0ecdae05ba028a06cf939ad294f9d41d5df2dcb1/libraries/AP_Relay/AP_Relay.cpp#L284

The only way that is not done is if the function had already been configured or if the pin is < 0. On a clean build the function is not configured and the pin define means you get a pin > 0.

It is possible that some combination of updates makes it not work. Its also possible that `defualts.param` vs the define makes it not work. What are the exact steps to reproduce?

Hwurzburg:
I also have verified on a pristine autopilot that shipped with BF that when loaded with AP 4.7 firmware sets the FUNCTION to 1 if the default relay pin is set in hwdef....curiously, it displays 0 as its default value, but a reset to defaults command does not set it back to 0, this is probably due to the relay innit function changing it  to 1 if the pin is not -1...
closing this as unnecessary (but taught me a bunch on python search and capture)

PR #29711, opened 2025-04-07
AP_Param: fix unexpected truncation of AP_Float parameters in expressions

### LINKED ISSUE #29660: type infererence to int for AP_xxx types error

tpwrules:
Fixes #29660 for floats at least. An expression like `float v = true ? float_param : 0;` would cause `float_param` to be truncated to an integer to match the other type. Fix by only letting `AP_Float` be converted to a float. It is still possible to get truncation with integer parameter types, e.g. `int16_t v = true ? int16_param : (int8_t)0;` but I don't think it's a problem at this point.

There is a bunch of subtlety and some caveats, please see the commit contents and messages for details. The issue contains a list of areas of the code whose behavior will be changed by this, but problems are exclusive to the ternary operator and overloaded functions, not standard arithmetic operators.

Tested on Cube Orange that the `CAMERA_FOV_STATUS` message now does not truncate the FOV values specified in the corresponding parameters.

### DIFF
--- a/libraries/AP_NavEKF3/AP_NavEKF3_AirDataFusion.cpp
+++ b/libraries/AP_NavEKF3/AP_NavEKF3_AirDataFusion.cpp
@@ -475,8 +475,8 @@ void NavEKF3_core::FuseSideslip()
 void NavEKF3_core::FuseDragForces()
 {
     // drag model parameters
-    const ftype bcoef_x = frontend->_ballisticCoef_x;
-    const ftype bcoef_y = frontend->_ballisticCoef_y;
+    const ftype bcoef_x = frontend->_ballisticCoef_x.get();
+    const ftype bcoef_y = frontend->_ballisticCoef_y.get();
     const ftype mcoef = frontend->_momentumDragCoef.get();
     const bool using_bcoef_x = bcoef_x > 1.0f;
     const bool using_bcoef_y = bcoef_y > 1.0f;
--- a/ArduPlane/system.cpp
+++ b/ArduPlane/system.cpp
@@ -381,7 +381,7 @@ void Plane::check_long_failsafe()
             failsafe_long_on_event(FAILSAFE_GCS, ModeReason::GCS_FAILSAFE);
         }
     } else {
-        uint32_t timeout_seconds = g.fs_timeout_long;
+        float timeout_seconds = g.fs_timeout_long;
         if (g.fs_action_short != FS_ACTION_SHORT_DISABLED) {
             // avoid dropping back into short timeout
             timeout_seconds = g.fs_timeout_short;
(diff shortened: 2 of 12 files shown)


peterbarker:
@Safe-bet more of the same issues were found, so thanks for bringing that float-truncation problem to our attention!

ArduCopter/Copter.cpp
```diff
@@ -889,7 +889,7 @@ void Copter::update_super_simple_bearing(bool force_update)
     }
 
     super_simple_last_bearing = bearing;
-    const float angle_rad = radians((super_simple_last_bearing+18000)/100);
+    const float angle_rad = radians((float)((super_simple_last_bearing+18000)/100));
```
tridge, line 892: should be 0.01

Safe-bet:
> @Safe-bet more of the same issues were found, so thanks for bringing that float-truncation problem to our attention!

Thank you very much, @peterbarker. It feels good to be appreciated. It was my first PR, and I hope to contribute more in the future through Github and hopefully as a part of GSOC.

libraries/AP_NavEKF3/AP_NavEKF3_AirDataFusion.cpp
```diff
@@ -475,8 +475,8 @@ void NavEKF3_core::FuseSideslip()
 void NavEKF3_core::FuseDragForces()
 {
     // drag model parameters
-    const ftype bcoef_x = frontend->_ballisticCoef_x;
-    const ftype bcoef_y = frontend->_ballisticCoef_y;
+    const ftype bcoef_x = frontend->_ballisticCoef_x.get();
+    const ftype bcoef_y = frontend->_ballisticCoef_y.get();
```
peterbarker, line 479: I wonder if these would be clearer as casts-to-ftype rather than .get()

similarly for many of the other casts.

rmackay9: I think ".get()" is prettier than a cast..

tpwrules: It would also have to be two casts, first to `float` then to the desired type, as we probably don't wish to implement cast operators for all desired types.

tridge (approved):
thanks for all your work on this!
I am curious why "return this->_value;" instead of "return _value;" ?

tpwrules:
> I am curious why "return this->_value;" instead of "return _value;" ?

Some strange namespacing rule I did not fully comprehend and this was the simplest fix for. The latter does not compile.

peterbarker (approved):

PR #29712, opened 2025-04-07
ArduPlane: fix quadplane stopping speed equation

eppravitra:
I'm not sure if you are aware of this, @tridge, but I think this is the correct equation to calculate the stopping speed. From my test, I don't notice any difference. I only test using SITL with Gazebo though.

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -2530,7 +2530,7 @@ void QuadPlane::vtol_position_controller(void)
         // calculate speed we should be at to reach the position2
         // target speed at the position2 distance threshold, assuming
         // Q_TRANS_DECEL is correct
-        const float stopping_speed = safe_sqrt(MAX(0, distance-position2_dist_threshold) * 2 * transition_decel) + position2_target_speed;
+        const float stopping_speed = safe_sqrt(MAX(0, distance-position2_dist_threshold) * 2 * transition_decel + sq(position2_target_speed));
 
         float target_speed = stopping_speed;
 
--- a/Tools/autotest/quadplane.py
+++ b/Tools/autotest/quadplane.py
@@ -1565,6 +1565,7 @@ class AutoTestQuadPlane(vehicle_test_suite.TestSuite):
             "SCR_ENABLE": 1,
             "SIM_SHIP_ENABLE": 1,
             "SIM_SHIP_SPEED": 5,
+            "Q_WP_SPEED": 700,
             "SIM_SHIP_DSIZE": 10,
             "FOLL_ENABLE": 1,
             "FOLL_SYSID": 17,


IamPete1 (commented):
I have done a geogebra of the change.

https://www.geogebra.org/calculator/frtg2bry

The current equation gets to 0 speed at distance threshold, the new equation gets to target speed at the distance threshold.

image

The change certainly makes the equation do what the comment says it should.

lthall (approved):

tridge:
@eppravitra thanks! we just need to do some testing both in RealFlight and in some real vehicles, @peterbarker and I can test on 2 quadplanes at SpringValley on a weekend, likely in 2 weeks time (I will be away this weekend)

tridge:
@eppravitra I'd also note that this code is also bad if you'd like to fix it - should not depend on wp_speed
```
        // limit target speed to a the pos1 speed limit, which starts out at the initial speed
        // but is adjusted if we start putting our nose down. We always allow at least twice
        // the WP speed
        target_speed = MIN(MAX(poscontrol.pos1_speed_limit, 2*wp_speed), target_speed);
```

tridge:
tested on a MakeFlyEasy Hero, no issues

eppravitra:
Hi Andrew @tridge,

The ship landing test keeps failing. I wonder if it has to do with this change.

I quickly looked at plane_ship_landing.lua, and I still can't spot anything obvious.

Your help would be greatly appreciated.

peterbarker:
> Hi Andrew @tridge,
> 
> The ship landing test keeps failing. I wonder if it has to do with this change.
> 
> I quickly looked at plane_ship_landing.lua, and I still can't spot anything obvious.
> 
> Your help would be greatly appreciated.

It's absolutely a regression as far as the autotest is concerned.  I haven't seen the test fail in CI before this PR.

It's also trivially reproducible locally.

IamPete1:
I have had a look at the failing test but I'm not that familiar with ship landing. It looks to me that its failing/passing due to mostly luck. This code change can only affect POS1. The final stage of landing is not affected. 

This is QPOS distance on master after POS1:

image

For the failure:

image

Both show a "ring down" it happens that one lands at a low point and the other lands at a high point.

tridge:
@eppravitra I think your PR has inadvertently exposed a flaw in the ship landing lua script. It does not slow down the landing if it is outside an acceptance cone for the landing on the target. We will need to fix that to get this to pass CI

tridge:
it turns out the key problem is Q_WP_SPEED=500 and the ship is moving at 5m/s, which is odd as SIM_SHIP_SPEED=3

tridge (approved):

IamPete1 (approved):
Thanks!

PR #29718, opened 2025-04-07
AP_InertialSensor: Remove unnecessary processing

muramura:
The variable RES is unused. Remove the code related to the RES variable.


peterbarker (approved):
This looks right to me, but I'd like @andyp1per to look at it as that `transfer_fullduplex" might not be the right function.

I will also point out that copying from `cmd` into `buf` here is pointless - could just transfer `cmd`!

That "default:" clause also doesn't belong.

andyp1per (approved):
Function is correct I think

peterbarker:
Merged, thanks!

PR #29719, opened 2025-04-07
AP_Logger: Ensure that the units are specified accurately

muramura:
This value is converted to bytes using 1024 as the divisor. Therefore, the unit is KiB.

WIKI:
https://en.wikipedia.org/wiki/Kilobyte


libraries/AP_Logger/AP_Logger.cpp
```diff
@@ -103,7 +103,7 @@ const AP_Param::GroupInfo AP_Logger::var_info[] = {
     // @Param: _FILE_BUFSIZE
     // @DisplayName: Logging File and Block Backend buffer size max (in kilobytes)
```
tpwrules, line 104:
```suggestion
    // @DisplayName: Logging File and Block Backend buffer size max (in kibibytes)
```

tpwrules:
I like the idea though!

rmackay9:
Txs. It would be good to squash the commits together into a single commit.

peterbarker (changes requested):
@muramura please squash the commits together

peterbarker:
I've rebased this.

PR #29730, opened 2025-04-09
Plane: fixed fence re-enable after fence breach

tridge:
If after a fence breach the user commands a WP change to a WP not in the landing sequence in AUTO mode or disables and re-enables the fence then the fence should be re-activated

### DIFF
--- a/ArduPlane/fence.cpp
+++ b/ArduPlane/fence.cpp
@@ -52,7 +52,7 @@ void Plane::fence_check()
 
     if (plane.previous_mode_reason == ModeReason::FENCE_BREACHED) {
         if (!armed || ((fence_breaches.new_breaches == 0 && orig_breaches == 0) && !plane.is_flying())) {
-            plane.previous_mode_reason = ModeReason::UNKNOWN;
+            plane.previous_mode_reason = ModeReason::FENCE_REENABLE;
         }
     }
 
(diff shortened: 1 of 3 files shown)


andyp1per (approved):

IamPete1:
May as well update this to use the new enum value too.
https://github.com/ArduPilot/ardupilot/blob/5233716e595b8f2e1f6cdf7af165607d20e41525/ArduPlane/fence.cpp#L34

tridge:
> May as well update this to use the new enum value too.

done

ArduPlane/fence.cpp
```diff
@@ -193,6 +203,11 @@ bool Plane::fence_stickmixing(void) const
 
 bool Plane::in_fence_recovery() const
 {
+    if (control_mode == &mode_auto && !mission.get_in_landing_sequence_flag()) {
```
peterbarker, line 206: I was surprised to see that on master we have a separate boolean for `in_return_path()` - I think we might need a clause in here for that state.

peterbarker (changes requested):

ArduPlane/fence.cpp
tridge: it is tricky to see how that will work, the flag gets set even if we're not using RET_AUTOLAND=4
we'd need to check for the flag not being set

peterbarker: Perhaps we should say the current patches are a significant improvement and create an issue for this case, then?

tridge: yes, I think so

PR #29714, opened 2025-04-07
Fix typos discovered by codespell

cclauss:
```
codespell **/*.py  --write-changes \
  --ignore-words-list=afile,bodgy,chek,commitish,edn,empy,finis,fram,mot,ned,parm,parms,rin,ser,sitl,siz
```
https://pypi.org/project/codespell

PR #29733, opened 2025-04-09
DRAFT: Port Tools/gittools/git-subsystems-split to Python

cclauss:
As discussed at:
* https://github.com/ArduPilot/ardupilot/pull/29714#issuecomment-2783503280

Let's try to port [`Tools/gittools/git-subsystems-split`](https://github.com/ArduPilot/ardupilot/blob/master/Tools/gittools/git-subsystems-split) to Python.  This contains several building blocks but does not yet:
* [x] Enable the commit message editing functionality --> `edit_subsystems_split_message()`
* [ ] `get_subsystems()` does not properly yield all subsystems (e.g., `autotest`, etc.)
* [ ] Enable the creation of the multiple commits.

Basicly https://github.com/ArduPilot/ardupilot/blob/5233716e595b8f2e1f6cdf7af165607d20e41525/Tools/gittools/git-subsystems-split#L64-L87

### DIFF
--- a/Tools/gittools/git-subsystems-split.py
+++ b/Tools/gittools/git-subsystems-split.py
@@ -1,0 +1,300 @@
+#!/usr/bin/env python
+
+"""Rewrite Tools/gittools/git-subsystems-split in Python.
+
+git-subsystems-split.py [OPTIONS]
+
+Ardupilot's git extension.
+
+Split HEAD commit into commits separated by subsystems (vehicles, libraries and
+folders in the project's root). Basically, reset and call commit-subsystems.
+
+If neither --copy or --edit is passed, then subsystems-split will try to make
+the original commit's message into a template for commit-subsystems.
+
+Options:
+    --copy
+    Make all commits have exactly the same message as the HEAD commit.
+
+    --edit
+    Edit the commit message as a template for commit-subsystems.
+"""
+
+import argparse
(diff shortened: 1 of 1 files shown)


cclauss:
I see elsewhere that gawk is used instead of awk.  Would that be a quick fix to my macOS issue?

Are the desired subsystem prefixes documented somewhere?  What would the prefix be for files at the root (like #29734)?  If someone can document them that would move this forward.

All the logging can be stripped away but helps while this is still a draft.

I will work offline on the creation of multiple commits.

If anyone sees a more direct way to achieve this objective then I would be ok with closing this pr.

tridge:
@cclauss I'm delighted to see this work! I was thinking of doing this myself, but happy for you to do it.
just put an easy to edit dictionary in the script for the mappings for special directories
for files at root, the only one regulary edited is wscript, which I normally prefix as "waf: xxxx"

peterbarker:
.... edits to `Tools/ardupilotwaf/**`  also usually get labelled with `waf:`.

cclauss:
@tridge wrote:
> just put an easy to edit dictionary in the script for the mappings for special directories

Please read thru the `get_commit_prefix_from_filepath()` logic.  I harvested several of your open PRs to get a hint about prefix logic.

PR #29735, opened 2025-04-09
Correct default OSD types

Hwurzburg:
I misunderstood (or forgot) that osd type 5 does not automatically include the analog OSD
in fact, the analog OSD includes MSP type (3) when enabled

this corrects any hwdefs that have both analog OSD and DisplayPort defaults to set OSD type to 1
will also update their wiki pages (including pending for 4.7) to say to enable OSD_TYPE2 to 5

also noted that some bds with no analog OSD define OSD_TYPE 5 correctly in hwdef but also duplicate in their params file, fixed it here also

### DIFF
--- a/libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/defaults.parm
+++ b/libraries/AP_HAL_ChibiOS/hwdef/ARK_FPV/defaults.parm
@@ -1,5 +1,3 @@
 # CAN
 CAN_P1_DRIVER 1  # Enables the use of CAN
 
-# Onboard OSD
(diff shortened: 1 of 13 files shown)


andyp1per (approved):
Thanks @Hwurzburg !

tridge (approved):

rmackay9:
I did not include this in 4.6.0-beta6 due to a merge conflict.  Perhaps @Hwurzburg or another dev could create a backport PR to get this in?

rmackay9:
This is included in 4.6.0-beta6 via https://github.com/ArduPilot/ardupilot/pull/29901

PR #29736, opened 2025-04-09
AP_HAL_Linux: ease access to verbose output by outputting on stdout aswell

heeplr:
...because, why not? There's a lot of useful info that's not easily accessible (for obvious reasons). On linux, we got stdout & stderr with pipes and everything.

While this naive attempt is useful for my debugging purposes, it should be seen more as a RFC than as PR.

At least the `#if CONFIG_HAL_BOARD == HAL_BOARD_LINUX` should probably be replaced by something like `CONFIG_USE_STDOUT` and set for boards that support it. But I'm not sure of an optimal solution and would kindly ask for advice from ppl with deeper knowledge of the codebase.

Maybe this could be done cleanly and at more places that output status messages, since I think it's *really* useful for debugging and it's basically free.

### DIFF
--- a/libraries/AP_HAL/utility/BetterStream.cpp
+++ b/libraries/AP_HAL/utility/BetterStream.cpp
@@ -1,6 +1,11 @@
 #include "BetterStream.h"
+#include <AP_HAL/AP_HAL.h>
 
+#if CONFIG_HAL_BOARD == HAL_BOARD_LINUX
+#include <cstdio>
+#endif
 #include "print_vprintf.h"
+
 
 void AP_HAL::BetterStream::printf(const char *fmt, ...)
 {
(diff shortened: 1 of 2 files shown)


peterbarker:
Isn't stdout mavlink by default at the moment?

heeplr:
> Isn't stdout mavlink by default at the moment?

You're right. By default ardupilot outputs all raw mavlink traffic to `stdout` causing the screen to garble with special characters.

This patch only handles text messages and also works if  `--serial0` is set but a client (or cable) is not connected.

It would probably be better if something like this patch used `stderr` so `stdout` can be silenced in the default case (to not garble the screen) by redirecting to `/dev/null` and not losing all the verbose messages (?)

Williangalvani:
Is there a particular reason we'd want stdout to be mavlink?

heeplr:
I hope someone knowledgeable can chime in. As for me, I can only guess: 

I could imagine two possibilites:
1. mavlink needs to have some output for ardupilot to work, even if none is defined by the user and it's just is the most sane default since stdout is always guaranteed to exist on *nix platforms.
2. stdout is very versatile. So when platform == linux, one can netcat, redirect to file, grep, less etc. (for debugging and/or production)

PR #29740, opened 2025-04-09
Remove vestiges of far-from-origin checks

peterbarker:
Unused defines


rmackay9 (approved):
surely if it compiles it must be fine, txs

PR #29744, opened 2025-04-10
Missing commits from fence margin check code.

andyp1per:
It seems I did not push the latest changes for https://github.com/ArduPilot/ardupilot/pull/28840 and it got merged without these. I am not quite sure what happened.

### DIFF
--- a/libraries/AC_Fence/AC_Fence.cpp
+++ b/libraries/AC_Fence/AC_Fence.cpp
@@ -42,6 +42,8 @@ extern const AP_HAL::HAL& hal;
 #define AC_FENCE_OPTIONS_DEFAULT                   0
 #endif
 
+#define AC_FENCE_NOTIFY_DEFAULT                     1.0f    // default to notifications at 1Hz
+
 //#define AC_FENCE_DEBUG
 
 const AP_Param::GroupInfo AC_Fence::var_info[] = {
@@ -145,6 +147,13 @@ const AP_Param::GroupInfo AC_Fence::var_info[] = {
     // @Bitmask: 0:Disable mode change following fence action until fence breach is cleared, 1:Allow union of inclusion areas, 2:Notify on margin breaches
     // @User: Standard
     AP_GROUPINFO_FRAME("OPTIONS", 11, AC_Fence, _options, static_cast<uint16_t>(AC_FENCE_OPTIONS_DEFAULT), AP_PARAM_FRAME_PLANE | AP_PARAM_FRAME_COPTER | AP_PARAM_FRAME_TRICOPTER | AP_PARAM_FRAME_HELI),
+
+    // @Param{Plane, Copter}: NTF_FREQ
+    // @DisplayName: Fence margin notification frequency in hz
+    // @Description: When bit 2 of FENCE_OPTIONS is set this parameter controls the frequency of margin breach notifications. If set to 0 only new margin breaches are notified.
+    // @Range: 0 10
+    // @User: Advanced
+    AP_GROUPINFO_FRAME("NTF_FREQ", 12, AC_Fence, _notify_freq, static_cast<float>(AC_FENCE_NOTIFY_DEFAULT), AP_PARAM_FRAME_PLANE | AP_PARAM_FRAME_COPTER | AP_PARAM_FRAME_TRICOPTER | AP_PARAM_FRAME_HELI),
 
     AP_GROUPEND
 };
@@ -909,6 +918,17 @@ void AC_Fence::record_margin_breach(uint8_t fence_type)
     // if we haven't already breached a margin limit, update the breach time
     const uint32_t now = AP_HAL::millis();
 
+    bool notify_margin_breach = false;
+
+    if (option_enabled(OPTIONS::NOTIFY_MARGIN_BREACH)) {
+        if ((!is_zero(_notify_freq)
+            && AP_HAL::timeout_expired(_last_margin_breach_notify_sent_ms, now,
+                                       uint32_t(roundf(1000.0f / _notify_freq))))
+            || !(_breached_fence_margins & fence_type)) {
+            notify_margin_breach = true;
+        }
+    }
+
     if (!(_breached_fence_margins & fence_type)) {
         _margin_breach_time = now;
     }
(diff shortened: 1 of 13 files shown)


andyp1per:
Flew 8 batteries on this - all worked as expected and the fence distance countdown was really nice!

peterbarker (approved):
LGTM, but we should fly it at SV

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -907,20 +907,24 @@ void AC_Fence::record_breach(uint8_t fence_type)
...
 
     // update bitmask
     _breached_fence_margins |= fence_type;
+
+    // emit a message indicated we're newly-breached, but not too often
+    if (option_enabled(OPTIONS::NOTIFY_MARGIN_BREACH)
+        && AP_HAL::timeout_expired(_last_margin_breach_notify_sent_ms, now, 1000U)) {
+        _last_margin_breach_notify_sent_ms = now;
+        char msg[MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN+1];
+        ExpandingString e(msg, MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN+1);
```
rmackay9, line 924: I've almost never seen this ExpandingString used.  @peterbarker you're OK with this?

rmackay9 (approved):
The renaming looks good to me, thanks!

peterbarker (changes requested):
We flew this at SV and tridge really didn't like the statustxt message load.  I was on the sticks so it didn't bother me :-)

@andyp1per probably want to discuss at DevCallEU?

andyp1per:
> We flew this at SV and tridge really didn't like the statustxt message load. I was on the sticks so it didn't bother me :-)

Ok sounds like we need a parameter to control the rate. For me the frequency is (a) actually what we originally agreed - i.e. supposed to be spammy - and (b) the bare minimum when you are moving fast - but I can see how this would be subjective. I think we should ditch the options bit and just have a parameter.

andyp1per:
I added an interval @peterbarker

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -146,6 +148,13 @@ const AP_Param::GroupInfo AC_Fence::var_info[] = {
     // @User: Standard
     AP_GROUPINFO_FRAME("OPTIONS", 11, AC_Fence, _options, static_cast<uint16_t>(AC_FENCE_OPTIONS_DEFAULT), AP_PARAM_FRAME_PLANE | AP_PARAM_FRAME_COPTER | AP_PARAM_FRAME_TRICOPTER | AP_PARAM_FRAME_HELI),
 
+    // @Param{Plane, Copter}: NTF_FREQ
+    // @DisplayName: Fence margin notification frequency in hz
+    // @Description: When bit 2 of FENCE_OPTIONS is set this parameter controls the frequency of margin breach notifications. If set to 0 only new margin breaches are notified.
+    // @Range: 0 10
```
rmackay9, line 154: we could put an @Units: Hz in here i think.

libraries/AC_Fence/AC_Fence.h
```diff
@@ -251,6 +251,7 @@ class AC_Fence
     AP_Int8         _ret_rally;             // return to fence return point or rally point/home
     AP_Int16        _ret_altitude;          // return to this altitude
     AP_Int16        _options;               // options bitmask, see OPTIONS enum
+    AP_Float        _notify_freq;    // margin notification frequency
```
rmackay9, line 254: personally I'd probably make it an AP_Int8 but not a blocker

andyp1per: The notify interval can't be an int as it needs to go lower than 1 hz - alternatives are changing to a ms interval, but I don't know which is more acceptable (or deci-hertz - ug)

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -146,6 +148,13 @@ const AP_Param::GroupInfo AC_Fence::var_info[] = {
     // @User: Standard
     AP_GROUPINFO_FRAME("OPTIONS", 11, AC_Fence, _options, static_cast<uint16_t>(AC_FENCE_OPTIONS_DEFAULT), AP_PARAM_FRAME_PLANE | AP_PARAM_FRAME_COPTER | AP_PARAM_FRAME_TRICOPTER | AP_PARAM_FRAME_HELI),
 
+    // @Param{Plane, Copter}: NTF_FREQ
```
tridge, line 151: we need to get rid of the _FRAME specific stuff in AC_Fence.cpp, poor old rovers are missing out on some options

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -146,6 +148,13 @@ const AP_Param::GroupInfo AC_Fence::var_info[] = {
...
+    // @DisplayName: Fence margin notification frequency in hz
+    // @Description: When bit 2 of FENCE_OPTIONS is set this parameter controls the frequency of margin breach notifications. If set to 0 only new margin breaches are notified.
+    // @Range: 0 10
+    // @User: Advanced
+    AP_GROUPINFO_FRAME("NTF_FREQ", 12, AC_Fence, _notify_freq, static_cast<float>(AC_FENCE_NOTIFY_DEFAULT), AP_PARAM_FRAME_PLANE | AP_PARAM_FRAME_COPTER | AP_PARAM_FRAME_TRICOPTER | AP_PARAM_FRAME_HELI),
```
tridge, line 156: poor little rover, left out again, can you remove this from the params where it doesn't make sense?

tridge (changes requested):

libraries/AC_Fence/AC_Fence.cpp
andyp1per: Fixed

libraries/AC_Fence/AC_Fence.cpp
andyp1per: Fixed

peterbarker (approved):

PR #29749, opened 2025-04-10
Fast rate should be constrained by actual loop rate

andyp1per:
Dynamic scaling of the rate loop rate should never go below the main loop rate, however the main loop rate can be lower than 400 Hz. This PR fixes this issue.

Have tested on my Chimera

### DIFF
--- a/ArduCopter/rate_thread.cpp
+++ b/ArduCopter/rate_thread.cpp
@@ -166,7 +166,8 @@ static inline bool run_decimated_callback(uint8_t decimation_rate, uint8_t& deci
 */
 void Copter::rate_controller_thread()
 {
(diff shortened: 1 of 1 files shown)


rmackay9:
Txs for this.  This is probably pretty obvious but I think all PRs should have:

1. a description
2. a mention of what testing has been done

Sorry to be annoying and pedantic but it will make the review process go faster

rmackay9:
Thanks for this.  You've probably already considered this but I wonder if the actual main loop rate or the configured main loop rate should be used.

andyp1per:
> Thanks for this. You've probably already considered this but I wonder if the actual main loop rate or the configured main loop rate should be used.

That's a reasonable suggestion - thanks @rmackay9 !

tridge (approved):

PR #29213, opened 2025-02-03
AP_NavEKF3: Allow extnav to work without vertical velocity

rishabsingh3003:
The added autotest fails in master.
The problem is that EK3 will fuse in ExtNav Vertical Velocity even if it has been disabled in EK3_SRCx_VELZ which can cause a spectacular crash.

PR #29750, opened 2025-04-10
AP_NavEKF: Allow EK3 cores to run different sources

rishabsingh3003:
By setting EK3_SRC_OPTIONS, this PR allows different EK3 cores to have different sensor sources.
As an example, when the correct bit is set,
First core will use the sensor set corresponding to EK3_SRC1
Second core will use sensor set corresponding to EK3_SRC2
And so on..

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -2867,6 +2867,76 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
         raise NotAchievedException("AUTOTUNE failed (%u seconds)" %
                                    (self.get_sim_time() - tstart))
 
+    def EKF3SRCPerCore(self):
+        '''Check SP/SH values in XKF4 for GPS (core 0) vs VICON (core 1) with targeted glitches.'''
+        self.set_parameters({
+            "EK3_ENABLE": 1,
+            "AHRS_EKF_TYPE": 3,
+            "VISO_TYPE": 1,
+            "SERIAL5_PROTOCOL": 2,
+            "EK3_SRC2_POSXY": 6,
+            "EK3_SRC2_VELXY": 6,
+            "EK3_SRC2_POSZ": 6,
+            "EK3_SRC2_VELZ": 6,
+            "EK3_SRC_OPTIONS": 8  # bitmask: enable multiple EKF cores
+        })
+
+        self.customise_SITL_commandline(["--serial5=sim:vicon"])
+        self.reboot_sitl()
+
+        self.wait_ready_to_arm()
+        self.takeoff(3)
+        self.delay_sim_time(5)
+
+        self.progress("Injecting VICON glitch")
+        self.set_parameter("SIM_VICON_GLIT_X", 100)
+        self.set_parameter("SIM_VICON_GLIT_Y", 100)
+        self.delay_sim_time(5)
+        self.set_parameter("SIM_VICON_GLIT_X", 0)
+        self.set_parameter("SIM_VICON_GLIT_Y", 0)
+
+        self.progress("Injecting BARO glitch")
+        self.set_parameter("SIM_BARO_GLITCH", 10)
+        self.delay_sim_time(5)
+        self.set_parameter("SIM_BARO_GLITCH", 0)
+
+        self.do_RTL()
+        self.wait_disarmed(timeout=100)
+
+        dfreader = self.dfreader_for_current_onboard_log()
+
+        max_SP_core0 = 0
+        max_SP_core1 = 0
+        max_SH_core0 = 0
+        max_SH_core1 = 0
+
+        while True:
+            m = dfreader.recv_match(type="XKF4")
+            if m is None:
+                break
+            if not hasattr(m, "C"):
+                continue
+            if m.C == 0:
+                max_SP_core0 = max(max_SP_core0, m.SP)
+                max_SH_core0 = max(max_SH_core0, m.SH)
+            elif m.C == 1:
+                max_SP_core1 = max(max_SP_core1, m.SP)
+                max_SH_core1 = max(max_SH_core1, m.SH)
+
+        self.progress(f"Core 0 (GPS): SP={max_SP_core0:.2f}, SH={max_SH_core0:.2f}")
+        self.progress(f"Core 1 (VICON): SP={max_SP_core1:.2f}, SH={max_SH_core1:.2f}")
+
+        # Validate: horizontal glitch only on core 1, vertical glitch only on core 0
+        if max_SP_core0 > 0.1:
+            raise NotAchievedException("Unexpected high SP in core 0 (GPS) during VICON glitch")
+        if max_SP_core1 < 0.9:
+            raise NotAchievedException("VICON glitch did not raise SP in core 1")
+
+        if max_SH_core1 > 0.5:
+            raise NotAchievedException("Unexpected high SH in core 1 (VICON) during BARO glitch")
+        if max_SH_core0 < 0.9:
+            raise NotAchievedException("BARO glitch did not raise SH in core 0")
+
     def AutoTuneSwitch(self):
         """Test autotune on a switch with gains being saved"""
 
@@ -13413,6 +13483,7 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
             self.CompassLearnCopyFromEKF,
             self.AHRSAutoTrim,
             self.Ch6TuningLoitMaxXYSpeed,
+            self.EKF3SRCPerCore,
         ])
         return ret
 
--- a/libraries/AP_NavEKF/AP_NavEKF_Source.cpp
+++ b/libraries/AP_NavEKF/AP_NavEKF_Source.cpp
@@ -135,7 +135,7 @@ const AP_Param::GroupInfo AP_NavEKF_Source::var_info[] = {
     // @Param: _OPTIONS
     // @DisplayName: EKF Source Options
     // @Description: EKF Source Options
-    // @Bitmask: 0:FuseAllVelocities, 1:AlignExtNavPosWhenUsingOptFlow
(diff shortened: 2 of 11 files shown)


libraries/AP_NavEKF/AP_NavEKF_Source.h
```diff
@@ -60,35 +61,48 @@ class AP_NavEKF_Source
     // initialisation
     void init();
 
+    uint8_t getActiveSourceSet(uint8_t core_number) const {
+        // check if we are using a separate source set for each core
+        if (_options.get() & int16_t(SourceOptions::SRC_PER_CORE)) {
```
tridge, line 66: please add const fn option_is_enabled()

libraries/AP_NavEKF/AP_NavEKF_Source.cpp
```diff
@@ -135,7 +135,7 @@ const AP_Param::GroupInfo AP_NavEKF_Source::var_info[] = {
     // @Param: _OPTIONS
     // @DisplayName: EKF Source Options
     // @Description: EKF Source Options
-    // @Bitmask: 0:FuseAllVelocities, 1:AlignExtNavPosWhenUsingOptFlow
+    // @Bitmask: 0:FuseAllVelocities, 1:AlignExtNavPosWhenUsingOptFlow, 2: Reserved , 3: Use SRC per core
```
tridge, line 138: needs better description.

tridge: remove reserved

tridge (changes requested):

tridge:
@rmackay9 please review

libraries/AP_NavEKF/AP_NavEKF_Source.h
rishabsingh3003: Already existed, I am using it now

libraries/AP_NavEKF/AP_NavEKF_Source.cpp
```diff
@@ -165,14 +165,14 @@ void AP_NavEKF_Source::setPosVelYawSourceSet(AP_NavEKF_Source::SourceSetSelectio
 }
 
 // true/false of whether velocity source should be used
-bool AP_NavEKF_Source::useVelXYSource(SourceXY velxy_source) const
+bool AP_NavEKF_Source::useVelXYSource(SourceXY velxy_source, uint8_t core_number) const
```
rmackay9, line 168: I think we use "core_index" instead of "core_number"

libraries/AP_NavEKF/AP_NavEKF_Source.h
```diff
@@ -60,35 +62,48 @@ class AP_NavEKF_Source
     // initialisation
     void init();
 
+    uint8_t getActiveSourceSet(uint8_t core_number) const {
```
rmackay9, line 65: let's put a comment above this declaration?

libraries/AP_NavEKF/AP_NavEKF_Source.h
```diff
@@ -60,35 +62,48 @@ class AP_NavEKF_Source
...
+        // check if we are using a separate source set for each core
+        if (option_is_set(SourceOptions::SRC_PER_CORE)) {
+            if (core_number >= AP_NAKEKF_SOURCE_SET_MAX) {
+                // this is a bit of a hack, but we need to return a valid source set
```
rmackay9, line 69: maybe remove the hack, just say what we're doing which is "invalid core index, return active source set"

libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -1013,7 +1013,9 @@ void NavEKF3::UpdateFilter(void)
     }
 
     // align position of inactive sources to ahrs
-    sources.align_inactive_sources();
+    for (uint8_t i=0; i<num_cores; i++) {
+        sources.align_inactive_sources(i);
```
rmackay9, line 1017: it doesn't matter functionally but we might as well move this loop into the align_inactive_sources() method

rmackay9:
This looks great, but what happens when a Lua script attempts to change the source set?  we could just ignore these requests or we could change the lane

libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -1548,7 +1552,12 @@ bool NavEKF3::using_extnav_for_yaw() const
 bool NavEKF3::configuredToUseGPSForPosXY(void) const
 {
     // 0 = use 3D velocity, 1 = use 2D velocity, 2 = use no velocity, 3 = do not use GPS
-    return  (sources.getPosXYSource() == AP_NavEKF_Source::SourceXY::GPS);
+    for (uint8_t i=0; i<num_cores; i++) {
+        if (sources.getPosXYSource(i) == AP_NavEKF_Source::SourceXY::GPS) {
```
tridge, line 1556: i think this should check only the active lane? otherwise AHRS will be very unhappy

tridge (changes requested):
configuredToUseGPSForPosXY() looks wrong

tridge (approved):

tridge:
@rmackay9 will review alignment code

libraries/AP_NavEKF/AP_NavEKF_Source.cpp
```diff
@@ -224,81 +225,114 @@ bool AP_NavEKF_Source::haveVelZSource() const
...
+        if (pos_xy_source == SourceXY::EXTNAV) {
+            // ExtNav is actively being used, do not align XY
+            allow_align_posxy = false;
+            break;
+        }
+
+        const bool valid_xy =
+            (pos_xy_source == SourceXY::GPS) ||
+            (pos_xy_source == SourceXY::BEACON) ||
+            ((pos_xy_source == SourceXY::OPTFLOW) &&
```
rmackay9: oops, optical flow is a velocity source..

rmackay9:
Let's test that the existing functionality works in SITL (e.g. without the new option being set)

1. SITL test switching between GPS and Vicon and ensure the existing bit leads to a jump when switching from Vicon->GPS but not GPS->Vicon.  [See dev wiki here](https://ardupilot.org/dev/docs/adding_simulated_devices.html#testing-vicon-aka-vision-positioning).
2. Confirm there is no jump when switching from Opticalflow->Vicon

BTW, this enhancement means that if the new bit is set and one EKF core/lane is using GPS and the other is using visual odometry, visual odometry will never be reset to match GPS even if the core using visual odometry is not the active core.  I suspect we will get requests from users to align the visual odometry to GPS if the core using visual odometry is not the active core.

rmackay9:
Hi @rishabsingh3003, looks like we need to rebase this on master now that your other PR https://github.com/ArduPilot/ardupilot/pull/29213 has gone in

rmackay9:
I've created a rebased version here with a small fix to the alignment logic when using optical flow https://github.com/rmackay9/rmackay9-ardupilot/tree/rishabh-ek3-src-mix

rishabsingh3003:
@rmackay9, could you force push on top of my branch? I'll give it a test. 
Thanks a lot for the help.

rmackay9:
Hi @rishabsingh3003, ok, pushed over top of this PR.  I'm also testing to see if I can find any issues

libraries/AP_NavEKF/AP_NavEKF_Source.cpp
rmackay9: fixed with a follow-up commit

rmackay9 (approved):
I've tested this fairly thoroughly in SITL and it seems to work as expected.

My testing was mostly focus on ensuring that the existing functionality still works as expected but I did also test the new source-per-lane functionality and it also worked.

Testing included:

1. replay tests to show that there were no changes when using GPS or GPS+Vicon when using the existing functionality (e.g. not using source-per-lane).
2. replay tests to show there were changes when us using GPS+Vicon with new source-per-lane feature enabled
3. tested reset logic between GPS + Vicon and OpticalFlow+Vicon using the existing functionality (e.g. not using source-per-lane).
4. light testing of source-per-lane using GPS + Vicon and triggered a change with the existing RC Option, "EKF lane switch attempt" (aka EKF_LANE_SWITCH).

So I think this is safe to merge once it passes CI.  We should probably squash the last two commits into earlier commits as well.

As discussed during the initial review some point it would be nice to have an RC option to force the active EKF lane but not a blocker of course.

Thanks for this great new feature!

rmackay9:
I've squashed the commits and rebased on master

Tools/autotest/arducopter.py
```diff
@@ -2970,6 +2970,77 @@ def AutoTuneYawD(self):
             if ogains[g] != ngains[g]:
                 raise NotAchievedException(f"AUTOTUNE gains not discarded {ogains=} {ngains=}")
 
+    def EKF3SRCPerCore(self):
```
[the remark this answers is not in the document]
peterbarker: > [nitpick] Autotest methods are typically prefixed with `test_` to make them consistent and discoverable. Consider renaming this to `test_EKF3SRCPerCore`.

They're not in ArduPilot-land.  At least not yet.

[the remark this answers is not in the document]
rmackay9: @peterbarker, indeed, co-pilot gets it right about half the time maybe.  If you look at one of it's earlier reviews it did find a real bug though.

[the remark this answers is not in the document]
peterbarker: > @peterbarker, indeed, co-pilot gets it right about half the time maybe. If you look at one of it's earlier reviews it did find a real bug though.

It's about 50:50 in my experience - and likely to get better.

Even here it's probably correct - we probably should conform to whatever Python standard thing this is!  But we don't want to mix styles, soo...

PR #29752, opened 2025-04-11
[4.6] AP_Camera: fix truncation of FOV stored in parameters

tpwrules:
Due to the strange casting rules of AP_Float parameters, the parameter value was converted to an integer because the other argument was an integer. Change the other argument to a float to cause the correct float cast instead.

Tested on Cube Orange that the `CAMERA_FOV_STATUS` message now does not truncate the FOV values specified in the corresponding parameters.


andyp1per (approved):

rmackay9 (approved):

rmackay9:
Merged, thanks!

PR #29753, opened 2025-04-11
AP_HAL_ChibiOS: add hwdef files for NarinFC-H5

vololand:
AP_HAL_ChibiOS: add hwdef files for NarinFC-H5


Hwurzburg:
@vololand since this is just a minor change to the H7, it should be done via includes....I built a new bootloader with the new board id (you had the H7 one), changed the bd id, since it was already taken....

as to the wiki page, it should just be included on the H7 page as a variant.

the big problem is the name...STM32-Hxxx is a valid processor series...naming it an -H5 is misleading and will cause confusion....
its your product, it will cause questions however....the pattern with other autopilots is to use something like Narin-H7A,etc.
not sure if this will be an issue when Dev team considers this PR

libraries/AP_HAL_ChibiOS/hwdef/NarinFC-H5/hwdef.dat
```diff
@@ -0,0 +1,25 @@
...
+#new SPI CS
+PF10 ICM45686_CS1 CS SPEED_VERYLOW
+PG6  ICM45686_CS CS SPEED_VERYLOW
+
+#new DRDY
+PE7  DRDY1_ICM45686_a INPUT GPIO(93)
+PJ0  DRDY5_ICM45686_b INPUT
+
+#new SPI IMUs
+SPIDEV icm45686_a     SPI1 DEVID4  ICM45686_CS1    MODE3  2*MHZ  8*MHZ
```
andyp1per, line 21: You could make these 16MHz and I think it would be better

andyp1per (approved):

vololand:
> @vololand since this is just a minor change to the H7, it should be done via includes...which I have provided new hwdefs...I built a new bootloader with the new board id (you had the H7 one), changed the bd id, since it was already taken....
> 
> as to the wiki page, it should just be included on the H7 page as a variant.
> 
> the big problem is the name...STM32-Hxxx is a valid processor series...naming it an -H5 is misleading and will cause confusion.... its your product, it will cause questions however....the pattern with other autopilots is to use something like Narin-H7A,etc. not sure if this will be an issue when Dev team considers this PR

@Hwurzburg
I checked in update hwdef.dat.
Thank you for your advice.

libraries/AP_HAL_ChibiOS/hwdef/NarinFC-H5/hwdef.dat
vololand: @andyp1per
I modified it to 16MHZ.
Thank you for your advice.

peterbarker:
@vololand are you absolutely sure that you want to call this a "H5"?  This might turn consumers off as they want a *H7* processor and would consider a "H5" to be a "lesser device" than a H7.  The fact it's running a H7 processor will be completely lost!

Please confirm you *really* want to use "H5" - everyone on the DevCall I'm currently on thinks its a bad name!

rmackay9:
We decided to hold off on merging this for another week in order to revisit the name again.

vololand:
> @vololand are you absolutely sure that you want to call this a "H5"? This might turn consumers off as they want a _H7_ processor and would consider a "H5" to be a "lesser device" than a H7. The fact it's running a H7 processor will be completely lost!
> 
> Please confirm you _really_ want to use "H5" - everyone on the DevCall I'm currently on thinks its a bad name!

@peterbarker
Thank you for your feedback.
I fully understand your concern.
I will use NarinFC-H5 as it is.
Thank you.

vololand:
> We decided to hold off on merging this for another week in order to revisit the name again.

@rmackay9
Thank you for your consideration.
I fully understand your concern.
I hope it will be merged as soon as possible.
Thank you.

peterbarker:
Merged it!

PR #29760, opened 2025-04-11
Set mavlink system ID in GCS library on all vehicles

peterbarker:
Four of our six vehicles updated their mavlink system ID from the parameter at 1Hz.

Instead, do it for all vehicles at the rate GCS update() is called.  It's a cheap operation.

This is a change in behaviour for Copter and Blimp which previously required a reboot for the parameter change to take effect.

See also https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/development.xml#L176 (`MAV_CMD_DO_SET_SYS_CMP_ID`) for a standardised way of changing a vehicle's system ID, rather than setting a parameter.

### DIFF
--- a/AntennaTracker/Tracker.cpp
+++ b/AntennaTracker/Tracker.cpp
@@ -77,9 +77,6 @@ void Tracker::get_scheduler_tasks(const AP_Scheduler::Task *&tasks,
 
 void Tracker::one_second_loop()
 {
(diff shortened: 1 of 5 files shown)


peterbarker:
Tested in SITL on Plane and Copter.

rmackay9 (approved):

PR #29761, opened 2025-04-11
AC_Fence: added FENCE_ALT_TYPE

tridge:
this allows for fence altitudes to be above ground level, absolute, origin or above home (the default and current behaviour)

### DIFF
--- a/libraries/AC_Fence/AC_Fence.cpp
+++ b/libraries/AC_Fence/AC_Fence.cpp
@@ -14,6 +14,7 @@
 #include <AP_HAL/AP_HAL.h>
 #include <AP_Logger/AP_Logger.h>
 #include <GCS_MAVLink/GCS.h>
+#include <AP_Terrain/AP_Terrain.h>
 
 extern const AP_HAL::HAL& hal;
 
@@ -73,7 +74,7 @@ const AP_Param::GroupInfo AC_Fence::var_info[] = {
 
     // @Param{Copter, Plane, Sub}: ALT_MAX
     // @DisplayName: Fence Maximum Altitude
-    // @Description: Maximum altitude allowed before geofence triggers
+    // @Description: Maximum altitude allowed before geofence triggers. See FENCE_ALT_TYPE for reference frame.
     // @Units: m
     // @Range: 10 1000
     // @Increment: 1
@@ -105,7 +106,7 @@ const AP_Param::GroupInfo AC_Fence::var_info[] = {
 
     // @Param{Copter, Plane, Sub}: ALT_MIN
     // @DisplayName: Fence Minimum Altitude
-    // @Description: Minimum altitude allowed before geofence triggers
+    // @Description: Minimum altitude allowed before geofence triggers. See FENCE_ALT_TYPE for reference frame.
     // @Units: m
     // @Range: -100 100
     // @Increment: 1
@@ -146,6 +147,13 @@ const AP_Param::GroupInfo AC_Fence::var_info[] = {
     // @User: Standard
     AP_GROUPINFO_FRAME("OPTIONS", 11, AC_Fence, _options, static_cast<uint16_t>(AC_FENCE_OPTIONS_DEFAULT), AP_PARAM_FRAME_PLANE | AP_PARAM_FRAME_COPTER | AP_PARAM_FRAME_TRICOPTER | AP_PARAM_FRAME_HELI),
 
+    // @Param: ALT_FRAME
+    // @DisplayName: Altitude frame
+    // @Description: The altitude reference frame for all altitude fences. Can be AMSL, home-relative, origin relative or AGL. Applies to both FENCE_ALT_MAX and FENCE_ALT_MIN.
+    // @User: Advanced
+    // @Values: 0:Above sea level, 1:Above Home, 2: Above Origin, 3: Above Terrain
+    AP_GROUPINFO_FRAME("ALT_TYPE", 12, AC_Fence, _alt_type, float(Location::AltFrame::ABOVE_HOME), AP_PARAM_FRAME_COPTER | AP_PARAM_FRAME_SUB | AP_PARAM_FRAME_TRICOPTER | AP_PARAM_FRAME_HELI | AP_PARAM_FRAME_PLANE),
+    
     AP_GROUPEND
 };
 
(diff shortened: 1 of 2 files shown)


libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -146,6 +147,13 @@ const AP_Param::GroupInfo AC_Fence::var_info[] = {
     // @User: Standard
     AP_GROUPINFO_FRAME("OPTIONS", 11, AC_Fence, _options, static_cast<uint16_t>(AC_FENCE_OPTIONS_DEFAULT), AP_PARAM_FRAME_PLANE | AP_PARAM_FRAME_COPTER | AP_PARAM_FRAME_TRICOPTER | AP_PARAM_FRAME_HELI),
 
+    // @Param: ALT_FRAME
+    // @DisplayName: Altitude frame
+    // @Description: The altitude reference frame for all altitude fences. Can be AMSL, home-relative, origin relative or AGL. Applies to both FENCE_ALT_MAX and FENCE_ALT_MIN.
```
peterbarker, line 152: I was thinking one for min, one for max.

For those users that don't want to come within n metres of the surface but do have an actual AMSL flight level limit.

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -521,6 +529,63 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
     return true;
 }
 
+/*
+  get our altitude in the FENCE_ALT_TYPE frame
+  return false if not available
+ */
+bool AC_Fence::get_alt_in_frame(float &alt)
```
peterbarker, line 536: This is the sort of naming you'd expect if passing in a frame.

```suggestion
bool AC_Fence::get_max_alt(float &alt)
```

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -521,6 +529,63 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
...
+
+    ahrs.get_relative_position_D_home(alt);
+
+    // translate to +ve up
+    alt = -alt;
+
+    /*
+      convert to desired altitude frame
+     */
+    const auto alt_frame = (Location::AltFrame)_alt_type.get();
```
peterbarker, line 548: Why does this need a cast?

libraries/AC_Fence/AC_Fence.cpp
```diff
@@ -521,6 +529,63 @@ bool AC_Fence::pre_arm_check(char *failure_msg, const uint8_t failure_msg_len) c
...
+    switch (alt_frame) {
```
peterbarker, line 549: I'm not understanding why we need this.

Can't we just construct a location object and call `get_alt(alt_frame)`?

peterbarker (changes requested):

libraries/AC_Fence/AC_Fence.cpp
tridge: it needs to work with no location, eg. copter indoors

libraries/AC_Fence/AC_Fence.cpp
rmackay9: I think it would be "get_alt_max" wouldn't it?

libraries/AC_Fence/AC_Fence.cpp
rmackay9: If we're using Fence then Copters, Rovers, etc need to know their Location.  We require that the EKF origin has been set and once this is done, the vehicle knows its Location.

rmackay9: Ah, sorry, in the case where only the altitude fence is specified the vehicle can arm without a Location.  We should consider enhancing the Copter pre-arm check so that if the altitude fence is enabled and the alt-type is set to Terrain then we also require a Location

rmackay9:
Great to see this change, I'm sure many users flying in mountainous regions will appreciate this

timtuxworth:
Along with Peter's other PR cleaning up location altitudes, this even more makes me think it might make sense for Altitude to be its own class. A location could have an altitude object, but altitudes (with magnitude and frame) could also be manipulated as an independent thing. A lot of this code would become simpler and easier to maintain.

LupusTheCanine:
> Along with Peter's other PR cleaning up location altitudes, this even more makes me think it might make sense for Altitude to be its own class. A location could have an altitude object, but altitudes (with magnitude and frame) could also be manipulated as an independent thing. A lot of this code would become simpler and easier to maintain.

Structured way of holding altitude may be useful but IMHO raw altitude class object shouldn't be exposed from Location as in Location class conversion AGL to AMSL etc. is coordinates dependant.
PS classes or something to enforce altitude reference frames would be good. (We really only need two, one for absolute altitude and one for terrain, sadly mp-units (units and quantities library hopefully headed for C++ standard) would probably be overkill and require to much work in the codebase 😢)

andyp1per:
Rebased on https://github.com/ArduPilot/ardupilot/pull/31619

libraries/AC_Fence/AC_Fence.cpp
andyp1per: No compilee without:

```
../../libraries/AC_Fence/AC_Fence.cpp: In member function ‘bool AC_Fence::check_fence_alt_max()’:
../../libraries/AC_Fence/AC_Fence.cpp:636:61: error: cannot convert ‘const signed char’ to ‘Location::AltFrame’
  636 |     if (!get_alt_U_in_frame(_curr_alt_u_m, _alt_max_type.get())) {
      |                                            ~~~~~~~~~~~~~~~~~^~
      |                                                             |
      |                                                             const signed char
      ```

libraries/AC_Fence/AC_Fence.cpp
andyp1per: Fixed in https://github.com/ArduPilot/ardupilot/pull/31619

libraries/AC_Fence/AC_Fence.cpp
andyp1per: Passing in the frame now so the naming is correct

andyp1per:
I've addressed the review comments in https://github.com/ArduPilot/ardupilot/pull/31619

tridge:
closing in favor of #31619

PR #29762, opened 2025-04-12
add magic MAVn_DEVID

peterbarker:
allows a user to work backwards from a MAVn_SRx parameter set to the device using those parameters

### DIFF
--- a/libraries/AP_SerialManager/AP_SerialManager.h
+++ b/libraries/AP_SerialManager/AP_SerialManager.h
@@ -229,6 +229,18 @@ public:
 #if HAL_LOGGING_ENABLED && HAL_UART_STATS_ENABLED
         AP_HAL::UARTDriver::StatsTracker stats;
 #endif
+
+        // Enum for device type used in get_device_id
+        enum class DeviceType {
+            UNKNOWN,
+            UART,
+            NETWORKING,
+            CANBUS,
+            SCRIPTING,
+        };
+
+        // Return a device id for this port
+        uint32_t get_device_id() const;
     };
 
     // get a state from serial index
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -156,6 +156,10 @@ bool GCS_MAVLINK::init(uint8_t instance)
         return false;
     }
 
+    // fill in the device ID (a parameter which allows a user to
+    // understand what their MAVn_ parameters actually correspond to)
+    devid.set(uartstate->get_device_id());
+
     // PARAMETER_CONVERSION - Added: May-2025 for ArduPilot-4.7
     // convert parameters; we used to use bits in the UARTDriver to
     // remember whether the mavlink connection on that interface was
--- a/libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
+++ b/libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
@@ -221,6 +221,13 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
     // This allows one time conversion while allowing user to flash between versions with and without converted params
     AP_GROUPINFO_FLAGS("_OPTIONSCNV",   21, GCS_MAVLINK, options_were_converted, 0, AP_PARAM_FLAG_HIDDEN),
 
+    // @Param: _DEVID
+    // @DisplayName: DeviceID
+    // @Description: ID of device using this MAVLink channel
+    // @ReadOnly: True
+    // @User: Advanced
+    AP_GROUPINFO_FLAGS("_DEVID",   22, GCS_MAVLINK, devid, 0, AP_PARAM_FLAG_INTERNAL_USE_ONLY),
+
     AP_GROUPEND
 };
 #undef DRATE
--- a/libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
+++ b/libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
@@ -225,6 +225,30 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
     // @DisplayName: DeviceID
     // @Description: ID of device using this MAVLink channel
     // @ReadOnly: True
+    // @Values: 0:Unknown
+    // @Values: 65542:SERIAL0 (USB on ChibiOS)
+    // @Values: 65798:SERIAL1
+    // @Values: 66054:SERIAL2
+    // @Values: 66310:SERIAL3
+    // @Values: 66566:SERIAL4
+    // @Values: 66822:SERIAL5
+    // @Values: 67078:SERIAL6
+    // @Values: 67334:SERIAL7
+    // @Values: 67590:SERIAL8
+    // @Values: 67846:SERIAL9
+    // @Values: 131078:NET_P1
+    // @Values: 131334:NET_P2
+    // @Values: 131590:NET_P3
+    // @Values: 131846:NET_P4
+    // @Values: 196614:CAN_D1_UC_S1
+    // @Values: 196870:CAN_D1_UC_S2
+    // @Values: 197126:CAN_D1_UC_S3
+    // @Values: 196622:CAN_D2_UC_S1
+    // @Values: 196878:CAN_D2_UC_S2
+    // @Values: 197134:CAN_D2_UC_S3
+    // @Values: 262150:SCR_SDEV1
+    // @Values: 262406:SCR_SDEV2
+    // @Values: 262662:SCR_SDEV3
     // @User: Advanced
     AP_GROUPINFO_FLAGS("_DEVID",   22, GCS_MAVLINK, devid, 0, AP_PARAM_FLAG_INTERNAL_USE_ONLY),
 
--- a/Tools/scripts/decode_devid.py
+++ b/Tools/scripts/decode_devid.py
@@ -24,6 +24,7 @@ parser.add_option("-C", "--compass", action='store_true', help='decode compass I
 parser.add_option("-I", "--imu", action='store_true', help='decode IMU IDs')
 parser.add_option("-B", "--baro", action='store_true', help='decode barometer IDs')
 parser.add_option("-A", "--airspeed", action='store_true', help='decode airspeed IDs')
+parser.add_option("-M", "--mavlink", action='store_true', help='decode MAVLink channel IDs')
 
 opts, args = parser.parse_args()
 
@@ -160,6 +161,13 @@ airspeed_types = {
     0x0C : "DEVTYPE_AIRSPEED_SCRIPTING",
 }
 
+mavlink_types = {
+    0x01 : "DEVTYPE_MAVLINK_UART",
+    0x02 : "DEVTYPE_MAVLINK_NETWORKING",
+    0x03 : "DEVTYPE_MAVLINK_CAN",
+    0x04 : "DEVTYPE_MAVLINK_SCRIPTING",
+}
+
 decoded_devname = ""
 
 if opts.compass:
@@ -173,7 +181,10 @@ if opts.baro:
 
 if opts.airspeed:
     decoded_devname = airspeed_types.get(devtype, "UNKNOWN")
-    
+
+if opts.mavlink:
+    decoded_devname = mavlink_types.get(devtype, "UNKNOWN")
+
 if bus_type == 3:
     #dronecan devtype represents sensor_id
     print("bus_type:%s(%u)  bus:%u address:%u(0x%x) sensor_id:%u(0x%x) %s" % (
--- a/libraries/AP_SerialManager/AP_SerialManager.cpp
+++ b/libraries/AP_SerialManager/AP_SerialManager.cpp
@@ -908,6 +908,60 @@ void AP_SerialManager::registered_ports_log()
 
 #endif // AP_SERIALMANAGER_REGISTER_ENABLED
 
+// Return a device id for this port
+uint32_t AP_SerialManager::UARTState::get_device_id() const
+{
+#if AP_SERIALMANAGER_REGISTER_ENABLED
+    if (idx >= AP_SERIALMANAGER_SCR_PORT_1) {
+        // Scripting port
+        return AP_HAL::Device::make_bus_id(
+            AP_HAL::Device::BUS_TYPE_SERIAL,
+            0,
+            idx - AP_SERIALMANAGER_SCR_PORT_1,
+            uint8_t(DeviceType::SCRIPTING)
+        );
+    }
+
+    if (idx >= AP_SERIALMANAGER_CAN_D2_PORT_1) {
+        // CAN D2 port
+        return AP_HAL::Device::make_bus_id(
+            AP_HAL::Device::BUS_TYPE_SERIAL,
+            1,
+            idx - AP_SERIALMANAGER_CAN_D2_PORT_1,
+            uint8_t(DeviceType::CANBUS)
+        );
+    }
+
+    if (idx >= AP_SERIALMANAGER_CAN_D1_PORT_1) {
+        // CAN D1 port
+        return AP_HAL::Device::make_bus_id(
+            AP_HAL::Device::BUS_TYPE_SERIAL,
+            0,
+            idx - AP_SERIALMANAGER_CAN_D1_PORT_1,
+            uint8_t(DeviceType::CANBUS)
+        );
+    }
+
+    if (idx >= AP_SERIALMANAGER_NET_PORT_1) {
+        // Networking port
+        return AP_HAL::Device::make_bus_id(
+            AP_HAL::Device::BUS_TYPE_SERIAL,
+            0,
+            idx - AP_SERIALMANAGER_NET_PORT_1,
+            uint8_t(DeviceType::NETWORKING)
+        );
+    }
+#endif // AP_SERIALMANAGER_REGISTER_ENABLED
+
+    // Standard port
+    return AP_HAL::Device::make_bus_id(
+        AP_HAL::Device::BUS_TYPE_SERIAL,
+        0,
+        idx,
+        uint8_t(DeviceType::UART)
+    );
+}
+
 namespace AP {
 
 AP_SerialManager &serialmanager()
--- a/libraries/GCS_MAVLink/GCS.h
+++ b/libraries/GCS_MAVLink/GCS.h
@@ -531,6 +531,7 @@ protected:
 
     // saveable rate of each stream
     AP_Int16        streamRates[NUM_STREAMS];
+    AP_Int32        devid;  // ID for device using this mavlink channel
 
     void handle_heartbeat(const mavlink_message_t &msg);
 


libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -2771,6 +2771,8 @@ void GCS::create_gcs_mavlink_backend(AP_HAL::UARTDriver &uart)
         return;
     }
 
+    // fill in the device ID
+
```
timtuxworth, line 2775: It looks like you added the comment but not the code that actually sets the devid?

peterbarker: Note the "WIP" label.  People asked on Discord about where this was up to, I decided to show not tell.... nad sak if anyone else wanted to take it up before I came back to it.

peterbarker:
Looking for ideas around:

```
In file included from ../../libraries/AP_HAL/SPIDevice.h:22,
                 from ../../libraries/AP_HAL/HAL.h:11,
                 from ../../libraries/AP_HAL/AP_HAL_Main.h:19,
                 from ../../libraries/AP_HAL/AP_HAL.h:8,
                 from ../../libraries/AP_Notify/DroneCAN_RGB_LED.cpp:17:
../../libraries/AP_HAL/Device.h:374:22: warning: ‘AP_HAL::Device::DeviceStructure::bus_type’ is too small to hold all values of ‘enum AP_HAL::Device::BusType’
  374 |         enum BusType bus_type : 3;
```

peterbarker:
Since I won't be around on DevCall to talk to this - the problem with adding this magic parameter is that we're out of space for buses in our device-id stuff!

timtuxworth:
> Since I won't be around on DevCall to talk to this - the problem with adding this magic parameter is that we're out of space for buses in our device-id stuff!

Does this mean we are out of space after adding this one? Or there is no space to add this one? In the first case, it's not a problem (for this), in the second case - what can we do?

tpwrules:
Looks like we are out of space so we can't add this one. The field is three bits so 7 is the max: https://github.com/ArduPilot/ardupilot/blob/71e9765a658ce06363975eb229775e6234c6a4ac/libraries/AP_HAL/Device.h#L372

I wonder if we can just use SERIAL and stuff our internal serial ID into the address or some other fields. I don't know who all the players are or where this is used off the top of my head. I don't know if these ID codes are AP-specific either.

libraries/AP_HAL/Device.h
```diff
@@ -41,6 +41,8 @@ class AP_HAL::Device {
         BUS_TYPE_MSP     = 5,
         BUS_TYPE_SERIAL  = 6,
         BUS_TYPE_WSPI    = 7,
+        BUS_TYPE_NET_PORT= 8,
+        BUS_TYPE_UART    = 9,
```
IamPete1: We have `BUS_TYPE_SERIAL` already?

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
```diff
@@ -221,6 +221,13 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
     // This allows one time conversion while allowing user to flash between versions with and without converted params
     AP_GROUPINFO_FLAGS("_OPTIONSCNV",   21, GCS_MAVLINK, options_were_converted, 0, AP_PARAM_FLAG_HIDDEN),
 
+    // @Param: _DEVID
+    // @DisplayName: DeviceID
+    // @Description: ID of device using this MAVLink channel
+    // @ReadOnly: True
+    // @User: Advanced
+    AP_GROUPINFO("_DEVID",   22, GCS_MAVLINK, devid, 0),
```
IamPete1: We typically use `AP_PARAM_FLAG_INTERNAL_USE_ONLY` on dev IDs.

rmackay9:
We should update the python decoding script if this goes in.  Also let's raise an issue for MP and any other GCSs and webtools that decode the devid.

rmackay9:
@peterbarker can you explain how a user would work back from this DevID to the serial port number?  E.g. MAVx to SERIALy

On the dev call we thought perhaps it would be better to just put the MAVx index into a parameter instead of trying to use a DEVID

timtuxworth:
Per the dev call discussion, could we not make this so much simpler by just adding a MAVx_PORT (or something similar) that maps to the SERIAL or NET port, CAN port or SCR (scripting) port index that we currently use under the covers. 

I know there was previously some hesitation about this, but surely it's simpler and easier to understand and explain than the DEViD thing? Is it possible it will also work without requiring GCS changes?

tpwrules:
Yeah here is the internal list: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_SerialManager/AP_SerialManager_config.h#L67

I think it would be easier than DEVID and doesn't need decoding.

peterbarker:
New look branch:

```
MAV1_DEVID: bus_type:SERIAL(6)  bus:0 address:0(0x0) devtype:0(0x0) USB0 (6)
MAV2_DEVID: bus_type:SERIAL(6)  bus:1 address:0(0x0) devtype:0(0x0) SERIAL1 (14)
MAV3_DEVID: bus_type:SERIAL(6)  bus:2 address:0(0x0) devtype:0(0x0) SERIAL2 (22)
MAV4_DEVID: bus_type:SERIAL(6)  bus:5 address:0(0x0) devtype:0(0x0) SERIAL5 (46)
MAV5_DEVID: bus_type:SERIAL(6)  bus:21 address:0(0x0) devtype:0(0x0) NET_P1 (174)
```

```
STABILIZE> SERIAL0_PROTOCOL 2           # MAVLink2
SERIAL1_PROTOCOL 2           # MAVLink2
SERIAL2_PROTOCOL 2           # MAVLink2
SERIAL3_PROTOCOL 5           # GPS
SERIAL4_PROTOCOL 5           # GPS
SERIAL5_PROTOCOL 2           # MAVLink2
SERIAL6_PROTOCOL -1          # None
SERIAL7_PROTOCOL -1          # None
param show NET*PRO*
STABILIZE> NET_P1_PROTOCOL  2           # MAVLink2
```

Note that we are encoding into the bus number, not devtype, and the [MAVProxy PR](https://github.com/ArduPilot/MAVProxy/pull/1736) has special treatment for that.

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
rmackay9: Txs, I've added the flag in a follow-up commit

rmackay9 (approved):
I've lightly tested this on a CubeOrange and the MAVn_DEVID parameters appear with unique numbers for each serial port.  Once merged I think the GCSs will show the Values string associated with the number so that users can more easily figure out which port the MAVn parameters control.

Image

BTW I hit this issue personally while investigating [how to setup DroneBridge ESP32 telemetry links for use with Follow mode](https://github.com/ArduPilot/ardupilot_wiki/pull/7988).  I'm sure I lost close to an hour because I assumed that MAV1_OPTIONS was for Serial1 instead of USB.

FYI @timtuxworth - rejoice!

rmackay9:
MAVProxy PR: https://github.com/ArduPilot/MAVProxy/pull/1736
Mission Planner issue: https://github.com/ArduPilot/MissionPlanner/issues/3761

tpwrules (dismissed):
The bus number maxes out at 31, this doesn't actually work for CAN or scripting ports?

Again, why not directly list the port index instead of calling it a DEVID? Otherwise we need a different DEVID encoding.

rmackay9:
Hi @tpwrules,

Txs for the review.

I'm also happy to have it be a MAVn_SERID (e.g. serial id) instead of a device id but which ever way we go, I'd like us to get this kind of solution merged - e.g. a parameter that users can look at to figure out which serial port the MAVn_xxx parameters control.

peterbarker:
> The bus number maxes out at 31, this doesn't actually work for CAN or scripting ports?

Moved to using dev_type instead, both MAVProxy and ArduPilot PRs.

> Again, why not directly list the port index instead of calling it a DEVID? Otherwise we need a different DEVID encoding.

Because I'm trying to work with an existing shit system rather than creating *another* shit system.

timtuxworth:
Claude says this:

### `_DEVID` value doesn't match the documented `@Values` table

In `GCS_Common.cpp`:

```cpp
devid.set(AP_HAL::Device::make_bus_id(
    AP_HAL::Device::BUS_TYPE_SERIAL,
    0,               // bus number
    0,               // addr
    uartstate->idx   // devtype
));
```

`make_bus_id(bus_type, bus, address, devtype)` packs its four arguments into three separate bytes of the 24-bit id (`AP_HAL/Device.h`):

- byte 0: `bus_type` (3 bits) | `bus` (5 bits)
- byte 1: `address`
- byte 2: `devtype`

Since `idx` is passed as `devtype` (byte 2), each increment of `idx` adds **65536**, not a small step. So for a UART on SERIAL1 (`idx == 1`):

```
devid = BUS_TYPE_SERIAL (6) + (1 << 16) = 65542
```

But the `@Values` table added in `GCS_MAVLink_Parameters.cpp` documents:

```
6:SERIAL0
14:SERIAL1
22:SERIAL2
30:SERIAL3
...
78:SERIAL9
174:NET_P1
...
334:CAN_D1_UC_S1
...
```

These documented numbers increase by **8** per index, i.e. they assume `6 + 8*idx` — consistent with `AP_SERIALMANAGER_NET_PORT_1 = 21` → `6 + 8*21 = 174`, and `AP_SERIALMANAGER_CAN_D1_PORT_1 = 41` → `6 + 8*41 = 334`. That matches encoding `idx` into the 5-bit `bus` field (each unit = `1<<3` = 8), not the `devtype` byte (each unit = `1<<16` = 65536).

**Net effect:** the actual runtime value of `MAVn_DEVID` will not equal any entry in its own `@Values` dropdown for any port except SERIAL0 (idx 0, where the multiplier doesn't matter). A GCS reading the metadata will show the wrong description — or none at all — for the value it receives. This undermines the entire purpose of the parameter.

### It's not quite as simple as swapping the argument order, either

If `idx` is instead passed as `bus` (to get the desired `×8` step), note `bus` is only a 5-bit bitfield (`uint8_t bus : 5`), max value 31. The documented table needs `idx` up to at least 62 (`SCR_SDEV2` → `(502-6)/8 = 62`). A bitfield assignment silently truncates at 31, so e.g. `CAN_D2_UC_S1` (idx 51) and both `SCR_SDEV` entries (idx 61/62) would wrap into the 0–31 range and collide with earlier SERIAL/NET indices rather than producing their documented values.

So a real fix needs one of:
1. Compute the raw `devid` value directly as arithmetic (`BUS_TYPE_SERIAL + (idx << 3)`) rather than going through `make_bus_id()`'s bitfield struct, or
2. Widen whatever field is meant to carry `idx` so it can hold the full range used by the `@Values` table, or
3. Rework the encoding/table so the two actually agree, and document what field officially owns "port index" in `Device.h`.

### Is this fatal?

No — it's a fixable encoding bug, not a design flaw. The underlying idea (surface a read-only per-channel device id, decode it against a documented value table) is sound and worth keeping. But as currently written the value produced won't match the value documented, so this shouldn't merge until one of the above is applied and verified. Easy to check in SITL: bring up a MAVLink connection on SERIAL1, read back `MAV1_DEVID`, and confirm it equals `14` per the table, not `65542`.

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
```diff
@@ -221,6 +221,32 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
...
+    // @Values: 174:NET_P1
+    // @Values: 182:NET_P2
+    // @Values: 190:NET_P3
+    // @Values: 198:NET_P4
+    // @Values: 334:CAN_D1_UC_S1
+    // @Values: 414:CAN_D2_UC_S1
+    // @Values: 494:SCR_SDEV1
+    // @Values: 502:SCR_SDEV2
+    // @User: Advanced
+    AP_GROUPINFO_FLAGS("_DEVID",   22, GCS_MAVLINK, devid, 0, AP_PARAM_FLAG_INTERNAL_USE_ONLY),
```
IamPete1, line 253: Could we put it at the top of the table and mark it as a enable parameter? That would hide the unused stream rates.

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -156,6 +156,15 @@ bool GCS_MAVLINK::init(uint8_t instance)
         return false;
     }
 
+    // fill in the device ID (a parameter which allows a user to
+    // understand what their MAVn_ parameters actually correspond to)
+    devid.set(AP_HAL::Device::make_bus_id(
+        AP_HAL::Device::BUS_TYPE_SERIAL,
```
IamPete1: There could be a argument for marking DroneCAN serial tunnels as DroneCAN bus type.

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -156,6 +156,15 @@ bool GCS_MAVLINK::init(uint8_t instance)
...
+        0,               // bus number
+        0,               // addr
+        uartstate->idx   // devtype
```
IamPete1: Dev type and address are both 8 bytes, I would say that address is a better for index than dev type. Maybe the dev type could be used to show uart/ethernet/droncan/scripting.

libraries/GCS_MAVLink/GCS_Common.cpp
IamPete1: I decided it would be better to use the dev type feild.

libraries/GCS_MAVLink/GCS_Common.cpp
IamPete1: Infact we could go further and split the index back up again. The bus can be used for CAN 1 vs CAN 2 for example, then use the type for CAN and the address for the port number of the CAN 1 ports.

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
IamPete1: This is not needed because the params are loaded by pointer anyway.

IamPete1:
I have done a PR for how I was thinking it could look here: https://github.com/peterbarker/ardupilot/pull/52

libraries/GCS_MAVLink/GCS_Common.cpp
IamPete1: see https://github.com/peterbarker/ardupilot/pull/52

tridge:

Full report: https://uav.tridgell.net/DevCallReviews/DevCallTopic/devcall_pr_reviews.html

Reviewed at head `ce2cf6e7a3` for the DevCallTopic call. Verdict: **REQUEST CHANGES**, though the first thing to say is that this is blocked on a decision rather than on code.

### The decision comes first

@tpwrules' CHANGES_REQUESTED (2026-08-22) asking why the port index is dressed up as a DEVID was never withdrawn, and @IamPete1 has since opened a competing implementation (peterbarker/ardupilot#52) that re-encodes the id in `AP_SerialManager`, updates `decode_devid.py`, and drops the `@Values` table entirely. Two live designs, one standing CHANGES_REQUESTED, one APPROVE. Taking that decision also dissolves the bug below, so it's worth taking first.

### BUG — the `@Values` table doesn't match what the code produces, for every port except SERIAL0

`libraries/AP_HAL/Device.h` packs a devid as `bus_type:3 | bus:5 | address:8 | devtype:8`, and `libraries/GCS_MAVLink/GCS_Common.cpp:161-166` passes `uartstate->idx` in the **devtype** slot — so the runtime value is `6 + idx*65536`. The table at `libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp:228-246` lists `6 + idx*8`, which is the **bus**-field encoding:

```
idx   table    actual
  0       6         6   ✓
  1      14     65542
  2      22    131078
  9      78    589830
 21     174   1376262
```

Only idx 0 agrees — which is why the CubeOrange bench test looked right: "unique numbers per port" holds whichever field carries the index. A GCS reading the metadata therefore shows no label, or a wrong one, for every value it actually receives except USB.

**The diagnosis currently on the thread is the wrong way round and is still unanswered.** The analysis pasted on 2026-08-22 concluded the *code* was wrong and that a fix was hard because `bus` is only 5 bits. But the code was moved to devtype deliberately ("Moved to using dev_type instead", 04:54) precisely to escape that 5-bit limit; the table commit predates the move and was never regenerated. So the fix is to regenerate the table as `6 + idx*65536` — not to touch `GCS_Common.cpp`.

### ISSUE — the magic value collides with real device ids at idx 0

`MAVn_DEVID` for SERIAL0 is `6`, and four in-tree devices already emit exactly that: `AP_Airspeed_NMEA.cpp:42`, `AP_Airspeed_External.cpp:8`, `AP_Baro_ExternalAHRS.cpp:9`, `AP_Compass_ExternalAHRS.cpp:23`. Not a functional bug — different parameters, nothing cross-checks them — but it undercuts the argument used to reject "just publish the port index": the value is only unambiguous because the consumer keys off the parameter *name*, and the MAVProxy PR does exactly that special-casing. If a name-keyed decoder is needed anyway, the DEVID framing isn't buying anything. Worth noting the competing implementation is collision-free by construction, since it never emits devtype 0.

### ISSUE — omitted ports, and `decode_devid.py` still not updated

`AP_DRONECAN_SERIAL_NUM_PORTS` and `AP_SCRIPTING_SERIALDEVICE_NUM_PORTS` are both 3, so idx 42, 43, 52, 53 and 63 are reachable and undocumented. And @rmackay9 asked on 2025-10-27 for the python decoder to be updated if this lands — `Tools/scripts/decode_devid.py` has no MAVLink mode and isn't among the three files touched here.

### Checked and clean

Parameter name length is fine (`MAV16_DEVID`, 11 chars — note the existing `MAV16_OPTIONSCNV` is already at exactly 16). The repeated `@Values:` lines are legal and concatenate. `set()` rather than `set_and_save()` is correct, since `load_all()` runs before `GCS_MAVLINK::init()`. Nothing in-tree consumes the value, so the magic id can't leak into logging, pre-arm, sensor ordering or MAVLink reporting. The 2025 "out of bus-type space" blocker is dead — this head reuses the existing `BUS_TYPE_SERIAL` and adds no bus type.

Despite being 16 months old the branch is only 13 commits behind master, none of which touch `GCS_MAVLink`, `AP_SerialManager` or `AP_HAL/Device.h` — so there's no stale-context risk. CI fully green.

timtuxworth:
> it undercuts the argument used to reject "just publish the port index": the value is only unambiguous because the consumer keys off the parameter _name_, and the MAVProxy PR does exactly that special-casing. If a name-keyed decoder is needed anyway, the DEVID framing isn't buying anything. Worth noting the competing implementation is collision-free by construction, since it never emits devtype 0.

Just saying.

peterbarker:
MAVProxy PR: https://github.com/ArduPilot/MAVProxy/pull/1736

timtuxworth:
Thanks @peterbarker @IamPete1 @rmackay9 !

rmackay9 (approved):

libraries/AP_HAL/Device.h
rmackay9: this has been resolved by merging in commits from IamPete1's branch

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
rmackay9: AP_PARAM_FLAG_INTERNAL_USE_ONLY has been added, thanks!

libraries/GCS_MAVLink/GCS_Common.cpp
rmackay9: we've merged in PeterH's branch

peterbarker:
```
 ../../modules/ChibiOS/os/common/ext/ST/STM32L4xx/stm32l431xx.h:1068:29: error: expected identifier before '(' token
   1068 | #define CAN                 ((CAN_TypeDef *) CAN1_BASE)
        |                             ^
  ../../libraries/AP_SerialManager/AP_SerialManager.h:238:13: note: in expansion of macro 'CAN'
    238 |             CAN,
        |             ^~~
```

Brilliant.

Renaming the new enumeration entry to CANBUS

libraries/AP_SerialManager/AP_SerialManager.h
```diff
@@ -229,6 +229,18 @@ class AP_SerialManager {
...
 #endif
+
+        // Enum for device type used in get_device_id
+        enum class DeviceType {
+            UNKNOWN,
+            UART,
+            NETWORKING,
+            CANBUS,
+            SCRIPTING,
+        };
```
timtuxworth, line 240: It's a bit of a nitpick but the type of descriptor doesn't match. e.g. UART and CANBUS are nouns but NETWORKING and SCRIPTING are adjectives. maybe UART, NETWORK, CANBUS, SCRIPT?

peterbarker: > It's a bit of a nitpick but the type of descriptor doesn't match. e.g. UART and CANBUS are nouns but NETWORKING and SCRIPTING are adjectives. maybe UART, NETWORK, CANBUS, SCRIPT?

No, NETWORKING and SCRIPTING are nouns in this context.  They refer to parts of ArduPilot.

PR #29763, opened 2025-04-12
AP_Mission: reset loop counter when loop completed

tridge:
this is a change of behaviour, but I think more closely aligns with expected behaviour (ie. principle of least surprise)

Should we have a MIS_OPTIONS bit for this?

### DIFF
--- a/libraries/AP_Mission/AP_Mission.cpp
+++ b/libraries/AP_Mission/AP_Mission.cpp
@@ -39,8 +39,8 @@ const AP_Param::GroupInfo AP_Mission::var_info[] = {
 
     // @Param: OPTIONS
     // @DisplayName: Mission options bitmask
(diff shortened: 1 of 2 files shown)


rmackay9:
Thanks for this. I think we should go without the MIS_OPTIONS if possible.  So if someone complains during beta testing we add it back but otherwise we go without.

libraries/AP_Mission/AP_Mission.cpp
```diff
@@ -2220,6 +2220,19 @@ bool AP_Mission::get_next_cmd(uint16_t start_index, Mission_Command& cmd, bool i
                 // continue searching from jump target
                 cmd_index = temp_cmd.content.jump.target;
             } else {
+                if (increment_jump_num_times_if_found) {
```
tridge, line 2223: will add option after devcall discussion

PR #29764, opened 2025-04-12
add and use a Location::copy_alt_from

peterbarker:
removes external knowledge of internal Location shape

### DIFF
--- a/libraries/AP_Common/Location.cpp
+++ b/libraries/AP_Common/Location.cpp
@@ -86,6 +86,14 @@ bool Location::change_alt_frame(AltFrame desired_frame)
     return true;
 }
 
+void Location::copy_alt_from(const Location &other)
+{
+    alt = other.alt;
+    relative_alt = other.relative_alt;
+    terrain_alt = other.terrain_alt;
+    origin_alt = other.origin_alt;
+}
+
 // get altitude frame
 Location::AltFrame Location::get_alt_frame() const
 {
--- a/libraries/SITL/SIM_Scrimmage.cpp
+++ b/libraries/SITL/SIM_Scrimmage.cpp
@@ -99,11 +99,13 @@ void Scrimmage::recv_fdm(const struct sitl_input &input)
 
     velocity_ef = Vector3f(pkt.speedN, pkt.speedE, pkt.speedD);
 
-    location.lat = pkt.latitude * 1.0e7;
-    location.lng = pkt.longitude * 1.0e7;
-    location.alt = pkt.altitude * 1.0e2;
-    position.z = (home.alt - location.alt) * 1.0e-2;
-
+    location = {
+        int32_t(pkt.latitude * 1.0e7),
+        int32_t(pkt.longitude * 1.0e7),
+        int32_t(pkt.altitude * 1.0e2),
+        Location::AltFrame::ABSOLUTE
+    };
+    UNUSED_RESULT(home.get_height_above(location, position.z));
 
     // velocity relative to air mass, in earth frame TODO
     velocity_air_ef = velocity_ef - wind_ef;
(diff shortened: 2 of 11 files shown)


timtuxworth:
Might it be better to create an Altitude class with a magnitude and frame? Then we would get things like copy() methods as standard and it would be much easier simpler to pass them around etc.

peterbarker:
> Might it be better to create an Altitude class with a magnitude and frame? Then we would get things like copy() methods as standard and it would be much easier simpler to pass them around etc.

That's an interesting concept - and it would actually match a function I was looking at writing the other day.

OTOH, it's actually *very* rare that we use altitudes without also having an associated lat/lng.  Maybe a dozen places in the code.

I'm reluctant to divert from our current push to eliminate access to the `.alt` field at this time.  It's just bug after bug after bug, and we *do* have clear both (a) intent and (b) a clear path to doing that at the moment.

timtuxworth:
> OTOH, it's actually _very_ rare that we use altitudes without also having an associated lat/lng. Maybe a dozen places in the code.

Fence min/max altitude ... ;-)

libraries/AP_Common/Location.cpp
```diff
@@ -86,6 +86,14 @@ bool Location::change_alt_frame(AltFrame desired_frame)
     return true;
 }
 
+void Location::copy_alt_from(const Location &other)
```
IamPete1, line 89: Should we be copying terrain relative altitudes without locations? That is somehow loosing some of the information. I note there is only one user in any-case.

libraries/SITL/SIM_Scrimmage.cpp
```diff
@@ -99,11 +99,13 @@ void Scrimmage::recv_fdm(const struct sitl_input &input)
 
     velocity_ef = Vector3f(pkt.speedN, pkt.speedE, pkt.speedD);
 
-    location.lat = pkt.latitude * 1.0e7;
-    location.lng = pkt.longitude * 1.0e7;
-    location.alt = pkt.altitude * 1.0e2;
-    position.z = (home.alt - location.alt) * 1.0e-2;
```
tridge, line 105: why remove position.z set?

libraries/SITL/SIM_SilentWings.cpp
```diff
@@ -209,16 +212,19 @@ void SilentWings::process_packet()
...
+            curr_location.lng,
+            0,
+            Location::AltFrame::ABSOLUTE,
+        };
+        origin.lat = home.lat;
+        origin.lng = home.lng;
         position.x = 0;
         position.y = 0;
-        position.z = -curr_location.alt;
+        position.z = -alt_abs_m * 100;;
```
tridge, line 227: looks wrong, maybe existing bug? position vector is used in meters in other places

libraries/SITL/SIM_XPlane.cpp
```diff
@@ -532,14 +536,18 @@ bool XPlane::receive_data(void)
     accel_earth.z += GRAVITY_MSS;
     
     // the position may slowly deviate due to float accuracy and longitude scaling
-    if (loc.get_distance(location) > 4 || abs(loc.alt - location.alt)*0.01f > 2.0f) {
+    ftype delta_height;
```
tridge, line 539: needs to be test flown over a long distance in XPlane

tridge (changes requested):

libraries/AP_Common/Location.cpp
peterbarker: Good call, added an internal error

libraries/SITL/SIM_Scrimmage.cpp
peterbarker: It's still there :-)

libraries/SITL/SIM_SilentWings.cpp
peterbarker: Yes, I agree, it does look suspect!  Not here to fix bugs ~~, just create them~~

libraries/SITL/SIM_XPlane.cpp
peterbarker: Removed these XPlane patches.

libraries/SITL/SIM_SilentWings.cpp
peterbarker: ... well, we did just end up changing this blind

libraries/AP_Common/Location.cpp
peterbarker: ... which we then removed at DevCall :-)

PR #29765, opened 2025-04-12
Plane: fixed incorrect set of location alt fields

tridge:
this fixes several places that set the alt field directly. The only one that is a real bug is in CIRCLE where if we were terrain following and change into CIRCLE mode we would climb to current alt + home alt absolute alt
This removes all sets of a alt field in a Location object in plane
This CIRCLE bug is also in 4.5.x

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -3289,7 +3289,8 @@ bool QuadPlane::do_vtol_takeoff(const AP_Mission::Mission_Command& cmd)
             return false;
         }
     } else {
-        plane.next_WP_loc.alt = plane.current_loc.alt + cmd.content.location.alt;
+        plane.next_WP_loc.set_alt_cm(plane.current_loc.alt + cmd.content.location.alt,
+                                     Location::AltFrame::ABSOLUTE);
     }
     throttle_wait = false;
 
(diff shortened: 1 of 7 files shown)


ArduPlane/quadplane.cpp
```diff
@@ -3289,7 +3289,8 @@ bool QuadPlane::do_vtol_takeoff(const AP_Mission::Mission_Command& cmd)
             return false;
         }
     } else {
-        plane.next_WP_loc.alt = plane.current_loc.alt + cmd.content.location.alt;
```
peterbarker, line 3292: Pre-existing, but the check on line 3287 looks wrong - I think it's possible those alts could be in different frames.

ArduPlane/quadplane.cpp
```diff
@@ -3560,7 +3562,8 @@ bool QuadPlane::verify_vtol_land(void)
                 plane.set_next_WP(plane.mission.get_current_nav_cmd().content.location);
             } else {
                 plane.set_next_WP(plane.next_WP_loc);
-                plane.next_WP_loc.alt = ahrs.get_home().alt;
+                plane.next_WP_loc.set_alt_cm(ahrs.get_home().alt,
```
peterbarker, line 3565: Ideal place to use `copy_alt_from` (916bc6a06db8e83d3a32c4914ac29b2dec80eded in my repo)

... and there's several of those in this PR

peterbarker (approved):
LGTM

ArduPlane/quadplane.cpp
tridge: you're right, I've pushed a fix and tested in SITL. Terrain alt VTOL takeoff did not work

rmackay9:
this is included in 4.6.0-beta6 as part of #29771

PR #29768, opened 2025-04-12
MSP VTX Support

andyp1per:
### Summary

Control an MSP video transmitter (HDZero, ~~Walksnail~~, ELRS backpack) from the flight controller — band/channel/frequency, power and pitmode — and report status to HD goggles.

### Classification & Testing

- [x] Automated test(s) verify changes (autotest `MSPVTXConfig`, `MSPDisplayPortVTXConfig`)
- [x] Tested manually, description below (SITL)
- [x] Tested on HDZero hardware (also tested on Walksnail and DJI but these appear not to support control)

Tested on an HDZero air unit + goggles on a TBS_LUCID_H7: band/channel, power and pitmode are all settable from the FC and confirmed by retuning the VRX; goggle-side channel changes coexist (push-on-change).

### Description

Adds MSP as a VTX control transport alongside CRSF/SmartAudio/Tramp, selected per-transport via the new `VTX_TYPES` parameter.

Two MSP VTX models are handled:
- **ELRS-backpack style** — the FC sends `MSP_SET_VTX_CONFIG` to the VTX.
- **HDZero/betaflight style** — the VTX is the MSP master. It polls the FC and, on finding a `"BTFL"` FC reporting `deviceIsReady=0`, uploads its own config and power table, then stops polling and expects the FC to push `MSP_VTX_CONFIG` on change. The FC reports not-ready until that handshake completes, then pushes changes (repeated for delivery, as the air unit doesn't report state back to confirm).

The FC identifier now matches the selected OSD symbol set (`BTFL`/`INAV`/`ARDU`); selecting `DISPLAYPORT_BTFL_SYMBOLS` is what enables HDZero VTX control. `MSP_OSD_CANVAS` is answered so HD goggles size their grid, and an `MSP_OPTIONS` bit holds the VTX at high power when disarmed.


hendjoshsr71:
Andy, does this add the ability to use ELRS TX backpacks and google backpacks to control and set the VTX?  

If so, I’ll give it a test.

andyp1per:
@hendjoshsr71 no idea, but I doubt it. I am only testing on HDZero right now. What messages do those backpacks send?

andyp1per:
@hendjoshsr71 how do I get the ELRS admin tool setup? What hardware do I need?

hendjoshsr71:
> @hendjoshsr71 how do I get the ELRS admin tool setup? What hardware do I need?

Just elrs with the backpack enabled, hd zero backpack enabled, and the correct msp message.

I think it might be this one,
https://github.com/betaflight/betaflight/blob/8e40eed8063b9b48813894ae5ad2c14458004dd2/src/main/msp/msp.c#L2145-L2162

https://discuss.ardupilot.org/t/vtx-control-from-rc-using-erls-rx-and-vtx-admin-tool/118778/2

Linjieqiang:
How's the progress?

andyp1per:
Not made any further progress as yet - busy with other projects

hydra:
Additionally, OpenPixelOSD supports MSP VTX, here's a link to their code where you can see the commands it uses.

https://github.com/cvetaevvitaliy/OpenPixelOSD/blob/master/src/vtx_msp.c#L380-L392

thespova:
Is this still being looked at? 

HDZero is showing SA VTX control as no longer recommended https://hd-zero.github.io/hdzero-vtx-docs/latest/vtx_user_manual/#smartaudio-protocolnot-recommend

When searching for Ardupilot MSP VTX control this is one of the first things that comes up.  Is this already part of dev branch for us to be able to try out?

andyp1per:
I will probably get to this in the not too distant future

thespova:
Yeah, I appreciate this getting some action!

peterbarker:
Andy is going to make this default to off, and modify boards that have MSP and the videotx hardware turn it on.

Peter will add a hwdef sanity check that if you mention a protocol in a SERIALn_PROTOCOL_DEFAULT that the protocol is actually built into the firmware.


```
../../libraries/AP_MSP/AP_MSP_Telem_DisplayPort.cpp: In member function 'virtual bool AP_MSP_Telem_DisplayPort::is_scheduler_enabled() const':
../../libraries/AP_MSP/AP_MSP_Telem_DisplayPort.cpp:46:16: error: 'vtx' is not a member of 'AP'
   46 |     return AP::vtx().get_enabled();
```

andyp1per:
Shoul dnow all be fixed

thespova:
Oh nice, is this just waiting on review now before it gets merged with dev branch?

What is meant by boards with MSP? Isn't this just a protocol that can be used on any serial port? 

Is this going to be a custom built option that can be added to boards with less flash or just not possible at all?

As soon as this is part of dev branch I'm hoping to test it on an Atom RC F405 board I have in a heli.

libraries/AP_HAL/board/sitl.h
```diff
@@ -96,6 +96,10 @@
 #define AP_FILTER_ENABLED 1
 #endif
 
+#ifndef AP_MSP_VIDEOTX_ENABLED
+#define AP_MSP_VIDEOTX_ENABLED 1
+#endif
+
```
peterbarker:
```suggestion
```

libraries/AP_HAL_ChibiOS/hwdef/include/minimize_fpv_osd.inc
```diff
@@ -26,6 +26,9 @@ define AC_PRECLAND_ENABLED 0
 # force Tramp support even in face of normal 1MB limit:
 define AP_TRAMP_ENABLED AP_VIDEOTX_ENABLED && OSD_ENABLED
 
+# force MSP VTX support even in face of normal 1MB limit:
+define AP_MSP_VIDEOTX_ENABLED AP_VIDEOTX_ENABLED && HAL_MSP_ENABLED
```
peterbarker:
```suggestion
define AP_MSP_VIDEOTX_ENABLED (AP_VIDEOTX_ENABLED && HAL_MSP_ENABLED)
```

I won't insist, but I've lost lots of time on occasion because these were missing!

libraries/AP_MSP/AP_MSP.cpp
```diff
@@ -44,8 +44,8 @@ const AP_Param::GroupInfo AP_MSP::var_info[] = {
 
     // @Param: _OPTIONS
     // @DisplayName: MSP OSD Options
-    // @Description: A bitmask to set some MSP specific options: EnableTelemetryMode-allows "push" mode telemetry when only rx line of OSD ic connected to autopilot,  EnableBTFLFonts-uses indexes corresponding to Betaflight fonts if OSD uses those instead of ArduPilot fonts. EnableINAVFonts uses INAV fonts and overrides EnableBTFLFonts if that option is enabled.
-    // @Bitmask: 0:EnableTelemetryMode, 1: unused, 2:EnableBTFLFonts, 3:EnableINAVFonts
+    // @Description: A bitmask to set some MSP specific options: EnableTelemetryMode-allows "push" mode telemetry when only rx line of OSD ic connected to autopilot,  EnableBTFLFonts-uses indexes corresponding to Betaflight fonts if OSD uses those instead of ArduPilot fonts. EnableINAVFonts uses INAV fonts and overrides EnableBTFLFonts if that option is enabled. ForceVTXHighPower always reports the vehicle as armed so a VTX that drops to low power when disarmed stays at full power.
```
peterbarker, line 47: Oooh.... that's interesting.

I mean, this is opt-in so I don't think it's a blocker, but I think we really do need to make a bit more of an issue about this whole, "lie about being armed".  That will affect everything to do with MSP...

e.g..... are google displays based on this so they might show as armed when the vehicle is not, so now the user can't tell whether the vehicle is *actually* armed?

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
```diff
@@ -123,6 +127,10 @@ bool AP_MSP_Telem_Backend::is_packet_ready(uint8_t idx, bool queue_empty)
 #endif
     case RTC_DATETIME:      // RTC
         return true;
+#if AP_MSP_VIDEOTX_ENABLED
+    case VTX_PARAMETERS:    // VTX control
+        return AP::vtx().have_params_changed();
```
peterbarker, line 132: So this method could only ever have one caller, but it kind of sounds generic (as in, when I read it I think, "since when?!"

I suggest keeping a `vtx_params_change_time_ms` and `AP::vtx().param_change_time_ms()` instead

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
```diff
@@ -447,7 +474,7 @@ MSPCommandResult AP_MSP_Telem_Backend::msp_process_command(msp_packet_t *cmd, ms
     return ret;
 }
 
-MSPCommandResult AP_MSP_Telem_Backend::msp_process_out_command(uint16_t cmd_msp, sbuf_t *dst)
+MSPCommandResult AP_MSP_Telem_Backend::msp_process_out_command(uint16_t cmd_msp, sbuf_t *src, sbuf_t *dst)
```
peterbarker, line 477:
```suggestion
MSPCommandResult AP_MSP_Telem_Backend::msp_process_out_command(uint16_t cmd_msp, const sbuf_t *src, sbuf_t *dst)
```

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
```diff
@@ -606,7 +645,11 @@ uint32_t AP_MSP_Telem_Backend::get_osd_flight_mode_bitmask(void)
 {
     // Note: we only set the BOXARM bit (bit 0) which is the same for BF, INAV and DJI VTX
     // When armed we simply return 1 (1 == 1 << 0)
-    if (hal.util->get_soft_armed()) {
+    // VTXs that drop to low power when disarmed watch this bit, so the
+    // ForceVTXHighPower option keeps it set to hold the VTX at full power
+    const AP_MSP *msp = AP::msp();
+    if (hal.util->get_soft_armed()
+        || (msp != nullptr && msp->is_option_enabled(AP_MSP::Option::VTX_HIGH_POWER))) {
```
peterbarker, line 652: Do the other stacks do the same thing to get the same effect?  Say, "armed" to force the VTX on?

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
```diff
@@ -1090,6 +1161,273 @@ MSPCommandResult AP_MSP_Telem_Backend::msp_process_out_rc(sbuf_t *dst)
 }
 #endif  // AP_RC_CHANNEL_ENABLED
 
+#if AP_MSP_VIDEOTX_ENABLED
+// set the configured pitmode option without disturbing the other option bits
+void AP_MSP_Telem_Backend::msp_vtx_set_pitmode(bool pitmode)
```
peterbarker, line 1166: Could I suggest a `set_option_enabled` instead?

libraries/AP_MSP/AP_MSP_Telem_Backend.h
```diff
@@ -149,6 +156,26 @@ friend AP_MSP;
...
+    // last configured VTX settings pushed to the air unit, to detect changes
+    struct {
+        uint8_t band;
+        uint8_t channel;
+        uint16_t freq_mhz;
+        uint16_t power_mw;
+        bool pitmode;
+        bool valid;
+    } _vtx_pushed;
+#endif
```
peterbarker:
```suggestion
#endif  // AP_MSP_VIDEOTX_ENABLED
```

similarly elsewhere

libraries/AP_MSP/msp_sbuf.cpp
```diff
@@ -42,4 +42,22 @@ void MSP::sbuf_write_data(sbuf_t *dst, const void *data, int len)
     dst->ptr += len;
 }
 
+uint8_t MSP::sbuf_read_u8(sbuf_t *src)
+{
+    if (sbuf_bytes_remaining(src) < 1) {
+        return 0;
```
peterbarker, line 48: Bleh.

This really looks like it should be an internal error if it ever occurs - similarly on the u16 case

libraries/AP_MSP/msp_sbuf.h
```diff
@@ -20,6 +20,9 @@ uint16_t sbuf_bytes_remaining(const sbuf_t *buf);
 bool sbuf_check_bounds(const sbuf_t *buf, const uint8_t len);
 void sbuf_switch_to_reader(sbuf_t *buf, uint8_t *base);
 void sbuf_write_data(sbuf_t *dst, const void *data, int len);
+// little-endian reads; advance the read pointer, returning 0 if the buffer is exhausted
```
peterbarker: Alternatively, make it boolean with a reference return value

libraries/AP_VideoTX/AP_VideoTX.cpp
```diff
@@ -270,6 +277,67 @@ void AP_VideoTX::update_all_power_dbm(uint8_t nlevels, const uint8_t power[])
     }
 }
 
+// mark the level matching the given mW as supported, stashing a non-standard
+// value in the custom slot; does not change the currently selected power
+void AP_VideoTX::update_power_mw(uint16_t power_mw, PowerActive active)
+{
+    for (uint8_t i = 0; i < VTX_MAX_POWER_LEVELS; i++) {
```
peterbarker: `ARRAY_SIZE`

Similarly elsewhere

libraries/AP_VideoTX/AP_VideoTX.cpp
```diff
@@ -270,6 +277,67 @@ void AP_VideoTX::update_all_power_dbm(uint8_t nlevels, const uint8_t power[])
...
+    uint8_t count = 0;
+    for (uint8_t i = 0; i < VTX_MAX_POWER_LEVELS; i++) {
+        if (_power_levels[i].active == PowerActive::Active && ++count == index) {
+            return _power_levels[i].mw;
+        }
+    }
+    return 0;
+}
+
+// one based index into the supported levels for a mW value, 0 if not matched
```
peterbarker: Should this read "active power levels"?

Similarly elsewhere

peterbarker (dismissed):
Did you ask claude to check how much of the new code is covered by the new regression tests?

libraries/AP_MSP/AP_MSP_Telem_Backend.h
```diff
@@ -23,12 +23,13 @@
...
 #include <AP_SerialManager/AP_SerialManager.h>
 #include <AP_OSD/AP_OSD.h>
+#include <AP_VideoTX/AP_VideoTX.h>
 
 #include "msp.h"
 
 #include <time.h>
 
-#define MSP_TIME_SLOT_MAX 12
+#define MSP_TIME_SLOT_MAX 13
```
peterbarker, line 35: I think you might need to gate this on `AP_MSP_VIDEOTX_ENABLED`

libraries/AP_HAL/board/sitl.h
andyp1per: Not sure I follow @peterbarker - there are SITL tests that need this defined and its disabled by default so what is wrong with this?

libraries/AP_MSP/AP_MSP.cpp
andyp1per: the armed status is shown by the OSD which we control so I don't think this will affect much other than not cooking your VTX

libraries/AP_MSP/msp_sbuf.cpp
andyp1per: internal error on VTX problems is not user-friendly in my view

libraries/AP_HAL/board/sitl.h
peterbarker: Bleh.  Sorry, I stuffed this up.

I *meant* to add a suggestion in `AP_MSP_config.h`:

```
#ifndef AP_MSP_VIDEOTX_ENABLED
#define AP_MSP_VIDEOTX_ENABLED (HAL_PROGRAM_SIZE > 2048)
#endif
```
... meaning it would get into SITL, Linux, CubeRedPrimary, H7A3 boards and whatnot.

libraries/AP_MSP/msp_sbuf.cpp
peterbarker: Falling out of the sky is also not user friendly.  An internal error here might stop that happening in the long-term.  If a user ever sees it *something has gone wrong*.

Hwurzburg (changes requested):
This adds 1.5K to all the F405 boards....200 bytes for a little used FPV  feature in even the H7s is too much....should be a turned off by default build option in IMO

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
andyp1per: have_params_changed() is the existing AP_VideoTX predicate - the SmartAudio and Tramp backends already gate on it (AP_SmartAudio.cpp, AP_Tramp.cpp), so it is not a single-caller method. This backend just follows the same pattern. Renaming it to a param_change_time_ms() timestamp API would touch those two backends as well; happy to do that across all three as a separate change if you would prefer it.

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
andyp1per: src cannot be const here: this switch also dispatches MSP_SET_VTX_CONFIG and MSP_SET_VTXTABLE_POWERLEVEL, whose handlers read a variable-length payload with sbuf_read_* and so advance src. It stays non-const unless the SET handling is split out of msp_process_out_command into a separate in-path - happy to do that if you think it is worth it.

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
andyp1per: No - this armed-bit trick is specific to the betaflight MSP protocol, where the VTX watches BOXARM in MSP_STATUS to decide power. SmartAudio and Tramp set VTX power directly over their own protocols and handle pitmode/pulldown themselves, so they never spoof armed state.

tridge:
**Automated review note — AI-generated (Claude), cross-checked by a second model against the live diff.** Please sanity-check before acting.

**Verdict: REQUEST CHANGES** — updated after a second-model cross-check. My earlier comment on this PR raised a finding that turned out to be wrong; it's withdrawn below and replaced with two that are more serious.

**Withdrawn — the sbuf internal-error finding was a false positive**
I previously said `msp_sbuf.h` documents an internal error that `msp_sbuf.cpp` doesn't implement. That was wrong. Both `sbuf_read_u8()` and `sbuf_read_u16()` call `sbuf_check_bounds()`, which does `INTERNAL_ERROR(AP_InternalError::error_t::flow_of_control)` before returning false (`msp_sbuf.cpp:21-24`). Sorry for the noise — please don't go looking for that one.

**Bug — unvalidated band/channel indexes the channel table out of bounds**
The standalone MSP VTX fields carry arbitrary `uint8_t` band and channel values, and `AP_MSP_Telem_Backend.cpp` (~line 1222) passes them straight to `VIDEO_CHANNELS[band][channel]`, whose accessor does no bounds checking. A malformed MSP packet reads outside the table and can persist an invalid VTX configuration. This is reachable from the wire, so it wants a range check before use.

**Bug — clearing the MSP bit in `VTX_TYPES` doesn't stop MSP controlling the VTX**
The MSP input paths gate only on the global `get_enabled()`, then mutate and save VTX state and set `_vtx_config_received`. Only the later provider-registration call honours `is_type_enabled(MSP)`. So a user who deliberately clears the MSP type bit still has an MSP peer able to manage their VTX settings — the parameter doesn't do what its name implies.

**Note — `AP_MSP_VIDEOTX_ENABLED` gating doesn't match its comment**
It keys off program size only, though the comment says it also requires AP_VideoTX. A custom >2 MiB build with VideoTX explicitly disabled would compile MSP VTX code against classes hidden behind `AP_VIDEOTX_ENABLED`.

**Note on scope** — this PR was reviewed in the 2026-07-22 run at head `0dd299e21a`, which has since been force-pushed away and is no longer fetchable, so I couldn't diff old→new and re-reviewed the full current diff (53 files, +828/−72).

libraries/AP_MSP/msp_sbuf.cpp
tridge: added internal error

libraries/AP_MSP/AP_MSP_config.h
```diff
@@ -15,3 +15,9 @@
 #ifndef HAL_WITH_MSP_DISPLAYPORT
 #define HAL_WITH_MSP_DISPLAYPORT HAL_MSP_ENABLED
 #endif
+
+// MSP VTX control needs both the MSP layer and AP_VideoTX. It is non-essential
+// and off by default; boards with MSP and the video hardware opt in.
+#ifndef AP_MSP_VIDEOTX_ENABLED
+#define AP_MSP_VIDEOTX_ENABLED 0
```
tridge: needs to be at least SITL, should probably be on >2048 too

tridge (dismissed):

tridge:
looks generally good, needs to enable in SITL and pass CI :-)

libraries/AP_HAL/board/sitl.h
andyp1per: Fixed thanks

libraries/AP_MSP/AP_MSP_config.h
andyp1per: Fixed

libraries/AP_HAL_ChibiOS/hwdef/include/minimize_fpv_osd.inc
andyp1per: Removed

libraries/AP_MSP/AP_MSP_Telem_Backend.h
andyp1per: Fixed generally

libraries/AP_MSP/msp_sbuf.h
andyp1per: Internall error was added

libraries/AP_VideoTX/AP_VideoTX.cpp
andyp1per: Fixed

libraries/AP_VideoTX/AP_VideoTX.cpp
andyp1per: Fixed

libraries/AP_MSP/AP_MSP_Telem_Backend.h
andyp1per: Fixed

libraries/AP_MSP/AP_MSP_Telem_Backend.cpp
andyp1per: It doesn't really support options - pitmode is a specific thing in MSP

Hwurzburg (approved):

PR #29772, opened 2025-04-13
Plane: Remove the initialization of the stack variable

muramura:
Since the distance value is set when the result of the get_height_above method is true, it is not necessary to initialize the stack variable.

AFTER
BUILD SUMMARY
Build directory: /home/parallels/work2/ardupilot/build/fmuv3
Target         Text (B)  Data (B)  BSS (B)  Total Flash Used (B)  Free Flash (B)  External Flash Used (B)
---------------------------------------------------------------------------------------------------------
bin/arduplane   1568629      4248   192496               1572877          507884  Not Applicable 

BEFORE
BUILD SUMMARY
Build directory: /home/parallels/work2/ardupilot/build/fmuv3
Target         Text (B)  Data (B)  BSS (B)  Total Flash Used (B)  Free Flash (B)  External Flash Used (B)
---------------------------------------------------------------------------------------------------------
bin/arduplane   1568633      4248   192500               1572881          507876  Not Applicable


libraries/AP_Common/Location.cpp
```diff
@@ -282,10 +282,18 @@ ftype Location::get_distance(const Location &loc2) const
     return norm(dlat, dlng) * LOCATION_SCALING_FACTOR;
 }
 
-// return the altitude difference in meters taking into account alt
-// frame.  if loc2 is below this location then "distance" will be
-// positive.  ie. this method returns how far above loc2 this location
```
peterbarker, line 287: some text has done missing

peterbarker (changes requested):

libraries/AP_Common/Location.cpp
muramura: I have incorporated the review comments.

peterbarker:
I've extracted the documentation change into a separate PR.

peterbarker:
I'm disinclined to push for this one.

While it does save bytes it also does introduce a very small risk of somehow using uninitialised stack data if something goes wrong in the called-function implemenation.  If this was new code I'd definitely not initialise it, but this is existing code so we should probably not fiddle.

If you're really keen please re-open and I'll take it to DevCall to get a Plane mantainer to make their final call on it.

Closing this one now.

PR #29774, opened 2025-04-13
AP_Bootloader: add UAV-DEV GmbH FC-UM982-H7 FC-M10S-H7 UM982-H7

mirkix:
Add IDs for upcoming hardware


andyp1per (approved):
It's fine to do it this way, but the range is reserved for you anyway so you could do it as part of the actual board PR if you wanted

mirkix:
@andyp1per Thanks Andy, will to that with my next hwdef.

PR #29775, opened 2025-04-13
AC_Fence: Check the FENCE settings first

muramura:
Check first whether the fence is enabled or disabled.


peterbarker (commented):
This looks OK to me, but Andy should approve

andyp1per (changes requested):
I just don't see that this adds value and these kind of changes should be tested which means the cost does not seem worth it to me

muramura:
If the FENCE test passes in the CI test, that should be sufficient.
I believe it's better not to include unnecessary processing.
When reading the source code, I wonder why such redundant operations are being performed.

timtuxworth:
I've implemented something like this as part of my scripted arming checks. The advantage of this approach is that you can change each check to be mandatory, warning only or disabled.
[Caveat: the current version contains bugs, I have a new version that needs to be checked in, I will try to do it soon]

PR #29776, opened 2025-04-14
GCS_MAVLink: remove unused send_named_float method from GCS_MAVLINK

peterbarker:


peterbarker:
.... turns out it *is* used :-)

image

PR #29777, opened 2025-04-14
AP_InertialSensor: disable loop check on sitl

andyp1per:
The new loop rate check effectively makes it impossible to run anything other than 400Hz as the loop rate on SITL since the INS rate is fixed at 1Khz (or less). This disables the check on SITL

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor.cpp
@@ -1505,7 +1505,7 @@ bool AP_InertialSensor::gyro_calibrated_ok_all() const
 // Prearm check to verify that we have sane sched loop rates set based on Gyro backend rates
 bool AP_InertialSensor::pre_arm_check_gyro_backend_rate_hz(char* fail_msg, uint16_t fail_msg_len) const
 {
-#if AP_SCHEDULER_ENABLED
+#if AP_SCHEDULER_ENABLED && CONFIG_HAL_BOARD != HAL_BOARD_SITL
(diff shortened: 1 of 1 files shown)


libraries/AP_InertialSensor/AP_InertialSensor.cpp
```diff
@@ -1505,7 +1505,7 @@ bool AP_InertialSensor::gyro_calibrated_ok_all() const
 // Prearm check to verify that we have sane sched loop rates set based on Gyro backend rates
 bool AP_InertialSensor::pre_arm_check_gyro_backend_rate_hz(char* fail_msg, uint16_t fail_msg_len) const
 {
-#if AP_SCHEDULER_ENABLED
+#if AP_SCHEDULER_ENABLED && CONFIG_HAL_BOARD != HAL_BOARD_SITL
```
tridge, line 1508: can you try changing SIM_RATE_HZ? we should be able to get faster IMUs in SITL

tridge (changes requested):

libraries/AP_InertialSensor/AP_InertialSensor.cpp
andyp1per: @tridge this is in realflight so my understanding is the gyro rate matches the loop rate?

tridge: decided can do a return pass if AHRS_EKF_TYPE=10 in SITL

andyp1per:
I've found a different approach for realflight

PR #29779, opened 2025-04-14
Tools: add Networking and CAN Mcast to feature list

bugobliterator:
Tested working with vscode extension using extract_features to show CAN MCAST and enable it.


peterbarker (changes requested):
```
./Tools/autotest/test_build_options.py --no-disable-all --no-disable-none --no-disable-in-turn --build-targets=copter --define-match-glob AP_NETWORKING_CAN_MCAST_ENABLED
```

yields
```
../../libraries/AP_HAL/utility/Socket.cpp:37:10: fatal error: sys/ioctl.h: No such file or directory
   37 | #include <sys/ioctl.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
```

bugobliterator:
So the issue was that when we set AP_NETWORKING_ENABLED to 1 on hardware without mac we don't fallback to PPP on ChibiOS boards. I have pushed the fix that should solve that issue.

peterbarker:
Merged, thanks!

PR #29784, opened 2025-04-15
Add support for MAV_CMD_DO_CHANGE_ALTITUDE

peterbarker:
Intended to replace the `handle_guided_change_alt` garbage where we pass in a new altitude using a mission item with a "current" of 3.

This functionality (changing altitude for current waypoint) is used by MAVProxy's "changealt" command.

At some stage I will update that to try the command first and if that fails go back to the mission item way of doing things.  But we probably don't want that until this is merged...


tridge (approved):

peterbarker:
Marked for backport consideration at behest of partner.

robertlong13:
This cherry-picks cleanly to 4.6, but causes build errors due to its dependence on #29741

PR #29785, opened 2025-04-15
Update AP_VideoTX.cpp for new 2500mW models

kozinalexey:
Add support 2500mW

### DIFF
--- a/libraries/AP_VideoTX/AP_VideoTX.cpp
+++ b/libraries/AP_VideoTX/AP_VideoTX.cpp
@@ -118,7 +118,7 @@ AP_VideoTX::PowerLevel AP_VideoTX::_power_levels[VTX_MAX_POWER_LEVELS] = {
     { 0x12, 600,  28, 0xFF }, // Tramp lies above power levels and always returns 25/100/200/400/600
     { 3,    800,  29, 40   },
     { 0x13, 1000, 30, 0xFF }, // only in SA 2.1
-    { 0xFF, 0,    0,  0XFF, PowerActive::Inactive }  // slot reserved for a custom power level
+    { 0xFF, 2500, 34, 0XFF}  // 2500mW
 };
 
 AP_VideoTX::AP_VideoTX()


libraries/AP_VideoTX/AP_VideoTX.cpp
```diff
@@ -118,7 +118,7 @@ AP_VideoTX::PowerLevel AP_VideoTX::_power_levels[VTX_MAX_POWER_LEVELS] = {
     { 0x12, 600,  28, 0xFF }, // Tramp lies above power levels and always returns 25/100/200/400/600
     { 3,    800,  29, 40   },
     { 0x13, 1000, 30, 0xFF }, // only in SA 2.1
-    { 0xFF, 0,    0,  0XFF, PowerActive::Inactive }  // slot reserved for a custom power level
+    { 0xFF, 2500, 34, 0XFF}  // 2500mW
```
peterbarker, line 121: Can you explain why we can re-purpose a slot "reserved for a custom power level" to be 2500mW?

Shouldn't this just be adding a line?

peterbarker (changes requested):
Commit message (not PR title) should be `AP_VideoTX: add 2500mW power level`

libraries/AP_VideoTX/AP_VideoTX.cpp
kozinalexey: Hello Peter, I believe that both options will work, in this option I made a compilation and made sure that the value 34 is suitable and switches the transmitter to 2500 mW mode. To add a new line, it may also be necessary to change the array size, but unfortunately I can't test this, I ended up in the hospital with a stroke and they won't let me near the computer, I wish everyone good health, :)

peterbarker: > Hello Peter, I believe that both options will work, in this option I made a compilation and made sure that the value 34 is suitable and switches the transmitter to 2500 mW mode. To add a new line, it may also be necessary to 
change the array size, but unfortunately I can't test this,

No worries, let's see what @andyp1per thinks...

> I ended up in the hospital with a stroke and they won't let me near the computer, I wish everyone good health, :)

Wow, maybe 2500mW *is* too much :-)  Do get well!

kozinalexey:
pr reworked with https://github.com/ArduPilot/ardupilot/pull/30895

PR #29788, opened 2025-04-15
AP_Mount: Use logical operator instead of bitwise

joshanne:
afaict, bitwise and (`&`) will always evaluate both sides, logical and (`&&`) will be able to short-circuit evaluate.

we're checking conditions, not bits.


peterbarker (approved):
LGTM

No change in overall behaviour, just enables short-circuit evaluation and disables slight amounts of confusion.

rmackay9 (approved):

PR #29789, opened 2025-04-15
AP_HAL_ChibiOS: Fix using the incorrect pin after checking the correct pin is defined

joshanne:
Tested with PixC4-Jetson, commenting VDD_BRICK_VALID out causes build to break. This change fixes the build.

### DIFF
--- a/libraries/AP_HAL_ChibiOS/AnalogIn.cpp
+++ b/libraries/AP_HAL_ChibiOS/AnalogIn.cpp
@@ -814,7 +814,7 @@ void AnalogIn::update_power_flags(void)
       for backup power)
     */
 #if defined(HAL_GPIO_PIN_VDD_BRICK2_VALID)
-    if (palReadLine(HAL_GPIO_PIN_VDD_BRICK_VALID) == 1) {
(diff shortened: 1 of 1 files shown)


tridge (approved):

rmackay9:

rmackay9:
I struggle a bit on how to explain this change in the 4.6.0-beta6 release notes https://github.com/ArduPilot/ardupilot/pull/29888.  At this point I've written what's below but a better explanation would be greatly appreciated!

`- Power brick2 valid flag fixed on some boards`

joshanne:
> I struggle a bit on how to explain this change in the 4.6.0-beta6 release notes https://github.com/ArduPilot/ardupilot/pull/29888.  At this point I've written what's below but a better explanation would be greatly appreciated!
> 
> `- Power brick2 valid flag fixed on some boards`

It's essentially a copy paste bug, imo. The logic for brick 1 was copied, but not all reads of the pin were updated. 

The effect of this is: 
* Two bricks defined, both  bricks report brick 1 status. 
* Only brick 2 defined, failure to build.

I would try to reference the fact that both bricks were reporting brick 1 status. 

It looks like only 1 hwdef made use of the required define.

Perhaps something like: "fix power brick 2 incorrectly reporting brick 1 status"?

PR #29790, opened 2025-04-16
AC_WPNav: avoid use of Location alt attribute in AC_WPNAV_OA

peterbarker:
this is mostly a comment, as I am trying to preserve existing behaviour while pointing out something that does look quite right


rmackay9 (approved):
I've tested this and it seems fine, thanks!

Below is a screen shot of a SITL test in a mountainous area and I confirmed that bendy ruler used with terrain altitude seems to work fine
image

PR #29791, opened 2025-04-16
autotest: exemptions for AP_MSP_INAV_FONTS_ENABLED

peterbarker:
OSD not instaniated on some vehicles


andyp1per:
Shouldn't this be dependent on display port so that this is unnecessary?

peterbarker:
> Shouldn't this be dependent on display port so that this is unnecessary?

These ones don't care about dependencies.  It's pretty hack.  @shiv-tyagi is going to do better by putting the vheicle in the build options for us...

PR #29798, opened 2025-04-16
Deprecate changing height using MISSION_ITEM_INT.current=3 

peterbarker:


rmackay9:
Only affects Plane so not my call but I'd get rid of it more aggressively

peterbarker:
> Only affects Plane so not my call but I'd get rid of it more aggressively

.... that would kill a feature (possibly only present in MAVProxy!) of being able to change you altitude while your vehicle is loitering a particular place.  https://github.com/ArduPilot/MAVProxy/pull/1555 moves MAVProxy away from using this == 3 mechanism to using the command we added support for last week.

I can't see support in MissionPlanner for this feature.

QGC notionally understands the feature exists, but it is impossible (in the QGC master code) to use the functionality.

On the subject - the "current == 2" case is next on the chopping block, and *that* definitely affects Copter.  The "go to this location" function is amply handled by `MAV_CMD_DO_REPOSITION`, but DO_YAW needs checking to make sure we have a good replacement!

peterbarker:
@MichelleRos and I flew this on the weekend.

ArduPlane/GCS_MAVLink_Plane.h
```diff
@@ -54,7 +54,11 @@ class GCS_MAVLINK_Plane : public GCS_MAVLINK
 
     void handle_message(const mavlink_message_t &msg) override;
     bool handle_guided_request(AP_Mission::Mission_Command &cmd) override;
+#if AP_MAVLINK_CHANGEALT_MISS_CURR_3_ENABLED
     void handle_change_alt_request(Location &location) override;
+#else
+    void handle_change_alt_request(Location &location);
```
IamPete1, line 60: Why not `#if` it out of the cpp too?

peterbarker: The method is used elsewhere in Plane, so we sitll need the method to be around.  It becomes non-virtual, 'though.

tridge (commented):
i'm concerned this will cause warnings on missionplanner

PR #29801, opened 2025-04-16
GCS_MAVLink: Remove unnecessary decisions

muramura:
The only condition for entering the else block is current == 3.
Therefore, the conditional check can be removed.


peterbarker:
Your logic is correct here, but I'd prefer *not* to do this.

I'm trying to eliminate both current==2 and current==3 cases from the codebase, and I would need to re-insert this line (or something like it) if someone had the ==3 case in but not the ==2 case in.  The first PR adding defines around this stuff is at 
https://github.com/ArduPilot/ardupilot/pull/29798/files and you can clearly see this will conflict.

The ==3 - particular after my PR - also helps clarify the condition we're entering that block, so I think we should preserve it until we remove it!

muramura:
@peterbarker san

The following condition can be made explicit by using a SWITCH statement.
I will change it to a SWITCH statement.

peterbarker:
Still no switch statement; please create a fresh PR if you do want to pursue.  This code really should go away rather than be tidied, however.

PR #29802, opened 2025-04-16
AP_Proximity: Add more proximity backends

rishabsingh3003:
We have sensors like the MR72 now that can be put all around the vehicle for 360 coverage


rmackay9:
This change looks OK and is low risk, the question is really just about how many do most users need.  I personally think 5 is too many and that it would be nice if we could make it a build option but I'm sure we will discuss at the next dev call

rmackay9 (approved):

rmackay9:

PR #29803, opened 2025-04-16
GCS_MAVLink: Remove unnecessary variable assignments

muramura:
This variable is used for a conditional check.
You can use it directly in the condition without assigning it to a separate variable.
There is no need to create an unnecessary variable just for the condition.

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -1382,8 +1382,7 @@ int8_t GCS_MAVLINK::deferred_message_to_send_index(uint16_t now16_ms)
         return -1;
     }
 
-    const uint16_t ms_since_last_sent = now16_ms - deferred_message[next_deferred_message_to_send_cache].last_sent_ms;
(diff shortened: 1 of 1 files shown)


rmackay9:
we sometimes do this because it's easier to read and I suspect it makes no difference to the final compiler output

.. so I don't think we should merge this.

Thanks very much in any case though!

peterbarker (changes requested):
Oh wow, no.

This is a *really* tricky bit of code to get right.  It's structured like this to force the value into a 16 bit variable before we do further calculations with it.  If we don't squish it into a 16-bit value first we end up doing 32-bit maths and getting the wrong answer.  Generally when working with 16-bit timestamps we have to make these intermediate variable to avoid subtle bugs creeping in (and not-so-subtle-bugs given what the test suite looks like!)

What we *really need here is a comment saying why we are doing 16-bit maths.

@muramura think you could add a comment instead?

peterbarker:
I've created a PR to replace this which just adds a comment explaining why we're doing it.

PR #16813, opened 2021-03-04
AP_Scripting: Add thrust stand

Gone4Dirt:
This is a fun lua script that I have been working on for the past few months.  I wanted to add another tool to the AP arsenal in terms of component selection / vehicle design.  So I have designed a static thrust stand that uses scripting to take advantage of AP, its  libraries, and mavlink for using the GCS for control and live telemetry.  All of the CAD for the hardware will be open source and the little python post-processing tool I am making (still needs a lot of work though! :-D )

### About the Project

Here is the thrust stand I have made:
image
image

Here are some of the features of the script:
- Logging of throttle, thrust, torque, voltage, current, motor commutations (RPM), time stamp.
- If using a BL Heli ESC then logging of ESC current, ESC RPM, ESC temp, ESC Voltage is also available.
- Throttle ramp mode: Gradually raises throttle to maximum (set as a param), pausing at 10 increments to capture steady-state response.
- Throttle transient mode:  Ramps throttle to 'hover throttle' and then proceeds to run through a table of step inputs about the the hover throttle to capture transient behaviour.  (The implementation of this still needs work).
- Over-current protection:  Set a parameter value for max current.  When exceeded the throttle is zeroed and the system stops  to prevent damaging ESCs and motors on test, with the screen and lights informing the user.
- Addressable LEDs used for throttle bar.
- OLED display used for on-screen feedback of throttle, thrust, torque, current, and voltage.
- System arming and disarming, arms 'copter' to make use of the motor safety behaviour already in AP, (just in-case the script dies).
- Pendant 'dead-mans' button for operating the motor.

There will be a blog post going up on the forum soon where I will get more of the details out.

The other thing that I am quite please about it there is a lot of flexibility in this.  Not everyone will necessarily need/want torque data for example, so they can easily get one of the low cost thrust stands off of bang good and plumb the load cell from that into this system to get way more data and automation.

I think one of the biggest benefits to users will be the ability to more easily get thrust data for expo fitting on multi-copters and VTOL planes.  The post-processing tool I am working on will help with this.

The data captured by this script is dumped to a *.CSV file so the bar to entry for processing the data is very accessible.  For example, it is easy to use the output in the already existing google sheets expo fitting tool.

Here is an example of some of the data I am getting out:
**Throttle Ramp Tests:**
image
image
image

**Throttle Transients:**
image
image
image

Here is what the post-processor currently looks like.  As I say, needs a lot more work.
image

### About the code

[...]

PR #29807, opened 2025-04-18
AP_Scripting: examples: add NAU7802.lua

peterbarker:
Need to tidy this up so that a user can calibrate with a parameter and provide simple weight measurements.

Also need to ensure this code can be used from other LUA scripts via require(?)

Swipe code from here, too: https://github.com/ArduPilot/ardupilot/pull/16813/files


libraries/AP_Scripting/examples/NAU7802.lua
```diff
@@ -0,0 +1,288 @@
...
+  if MSB and MID and LSB then
+
+    local value = (MSB << 16) | (MID << 8) | LSB
+  -- Quite likely to be a signed vs unsigned error here !! need to check
+  
+  --  local sign = (value & (1 << 23)) > 0
+  --  if sign then
+  --    value = 16777215 ~ value -- 23 bit not
+  --   end
+  -- gcs:send_text(0,string.format("%u, %u, %u, %u",MSB,MID,LSB,value));
```
IamPete1, line 158:
```suggestion
    local sign = (value & (1 << 23)) > 0

    if sign then
        value = 16777215 ~ value -- 24 bit XOR to flip the bits for two's compliment
        value = (value * -1) - 1
    end
```

IamPete1: You could probably string pack and unpack again to get this too. The suggestion is from the thrust test script.

peterbarker: Applied this patch

PR #29462, opened 2025-03-08
Add Statespace model support to SITL

bnsgeyer:
This PR adds a separate model to SITL to support state space models.  It currently supports multirotor and helicopter models.  It also uses JSON files for model parameter inputs.  This allows anyone to use their own state space dynamic model without having to modify the SITL code.

PR #29809, opened 2025-04-18
SITL: Support JSON aero coefficients

Ryanf55:
# Purpose

For simulations, we want to run SITL with our own aero model without recompiling the code.
This makes it easier to maintain than forks with modified coefficients, one for each vehicle.

# Tests performed

First, run with the default model which is loaded as JSON now. Arm and takeoff and observe it flies the same.

```bash
./Tools/autotest/sim_vehicle.py -v Plane  --console -DG
# wait some time
arm throttle
mode takeoff
```

Now, specify a json file to load, which loads the same file as the default.

```bash
./Tools/autotest/sim_vehicle.py -v Plane --model plane:@ROMFS/models/skywalker_2013.json  --console -DG
```

Finally, add your own aero model by copying skywalker_2013.json to a new file, and change a value.
```bash
cp Tools/autotest/models/skywalker_2013.json  Tools/autotest/models/custom_plane_aero.json
```

Now, raise the stall angle `"alpha_stall": 0.8`.

Finally, rerun with the model, and set a breakpoint to ensure we changed the value.
```bash
./Tools/autotest/sim_vehicle.py -v Plane --model plane:@ROMFS/models/custom_plane_aero.json  --console -DGB SIM_Plane.cpp:41
```

Then, print the stall angle to see it's been updated.

image

# Further improvements

Reduce code duplication

# Related

* Inspired by https://github.com/ArduPilot/ardupilot/pull/29462

### DIFF
--- a/libraries/SITL/SIM_Plane.cpp
+++ b/libraries/SITL/SIM_Plane.cpp
@@ -21,12 +21,23 @@
 
 #include <stdio.h>
 #include <AP_Filesystem/AP_Filesystem_config.h>
+#include <AP_Filesystem/AP_Filesystem.h>
 
 using namespace SITL;
 
 Plane::Plane(const char *frame_str) :
     Aircraft(frame_str)
 {
+
+    const char *colon = strchr(frame_str, ':');
+    size_t slen = strlen(frame_str);
+    // The last 5 letters are ".json"
+    if (colon != nullptr && slen > 5 && strcmp(&frame_str[slen-5], ".json") == 0) {
+        load_coeffs(colon+1);
+    } else {
+        coefficient = default_coefficients;
+    }
+
     mass = 2.0f;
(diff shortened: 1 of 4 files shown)


libraries/SITL/SIM_Plane.cpp
```diff
@@ -106,6 +117,116 @@ Plane::Plane(const char *frame_str) :
     }
 }
 
+void Plane::load_coeffs(const char *model_json)
```
Ryanf55, line 120: This is very similar to SIM_Frame's Frame::load_frame_params

IamPete1: Could they be combined? Maybe you could pass in the json_search array?

tridge (approved):

PR #29814, opened 2025-04-19
AP_Logger: Enable LOG_FILE_DSMROT by default

Ryanf55:
* Most devs seem to agree this is nice for log analysis with multiple flights on the same battery
* Also remind users to disable LOG_FILE_DSMROT if doing multiple flights with replay logging

Discussed in discord here: https://discord.com/channels/674039678562861068/710994174299603055/1362802558221160598

### DIFF
--- a/libraries/AP_Logger/AP_Logger.cpp
+++ b/libraries/AP_Logger/AP_Logger.cpp
@@ -117,17 +117,17 @@ const AP_Param::GroupInfo AP_Logger::var_info[] = {
 
     // @Param: _REPLAY
     // @DisplayName: Enable logging of information needed for Replay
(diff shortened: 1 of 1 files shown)


andyp1per (approved):

MichelleRos (changes requested):
If we just change the default like this, doesn't it mean that for people who haven't changed the parameter it will change to the new default? Which means everyone with replay enabled will suddenly have a change in behaviour and non-working replay logging with zero warning?

Ryanf55:
> If we just change the default like this, doesn't it mean that for people who haven't changed the parameter it will change to the new default? Which means everyone with replay enabled will suddenly have a change in behaviour and non-working replay logging with zero warning?

Good question! Are you suspecting users would be hit with this when they upgrade from 4.6 to 4.7? I'm happy to test it.

Ryanf55:
We might lose replay logs. Can reopen once we have safer replay param checking/prearms/Lua etc.

PR #29816, opened 2025-04-19
Fix floating point math in AP_Declination get_mag_field_ef method

LordPikaChu:
When running a quadcopter in SITL it would crash right after startup with the following error:

```
Process 5645 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGTERM
    frame #0: 0x00007ff81236afae libsystem_kernel.dylib`__semwait_signal + 10
libsystem_kernel.dylib`:
->  0x7ff81236afae <+10>: jae    0x7ff81236afb8            ; <+20>
    0x7ff81236afb0 <+12>: movq   %rax, %rdi
    0x7ff81236afb3 <+15>: jmp    0x7ff8123699d4            ; cerror
    0x7ff81236afb8 <+20>: retq   
Target 0: (arducopter) stopped.
```

Not sure if this crash could also happen in non-SITL builds. Anyway, this PR simplifies the rounding logic to avoid a situation where, if the `latitude_deg` or `longitude_deg` produce a value outside of int32_t range, the cast would crash the app. The resulting logic should produce the same result.

### DIFF
--- a/libraries/AP_Declination/AP_Declination.cpp
+++ b/libraries/AP_Declination/AP_Declination.cpp
@@ -36,10 +36,9 @@ bool AP_Declination::get_mag_field_ef(float latitude_deg, float longitude_deg, f
 {
     bool valid_input_data = true;
 
-    /* round down to nearest sampling resolution. On some platforms (e.g. clang on macOS),
-        the behaviour of implicit casts from int32 to float can be undefined thus making it explicit here. */
-    float min_lat = float(static_cast<int32_t>(static_cast<int32_t>(floorf(latitude_deg / SAMPLING_RES)) * SAMPLING_RES));
-    float min_lon = float(static_cast<int32_t>(static_cast<int32_t>(floorf(longitude_deg / SAMPLING_RES)) * SAMPLING_RES));
+    /* round down to nearest sampling resolution */
+    float min_lat = floorf(latitude_deg / SAMPLING_RES) * SAMPLING_RES;
+    float min_lon = floorf(longitude_deg / SAMPLING_RES) * SAMPLING_RES;
 
     /* for the rare case of hitting the bounds exactly
      * the rounding logic wouldn't fit, so enforce it.


peterbarker:
@andyp1per you fixed this in 2022 to avoid... floating point exceptions.

Didn't realise MacOSX was *this* much of a moving target :-)

@LordPikaChu commit message needs to have `AP_Declination: ` prepended to it

LordPikaChu:
> @andyp1per you fixed this in 2022 to avoid... floating point exceptions.
> 
> Didn't realise MacOSX was _this_ much of a moving target :-)
> 
> @LordPikaChu commit message needs to have `AP_Declination: ` prepended to it

Done 🫡 

I figured that the typecast was there to fix floating point exceptions after the fact (of creating the PR). I'm actually getting many more exceptions caused by floating point divisions in other cases too. I'm on an Intel Mac so I'm now wondering if this would also happen on Arm 🤔 These other exceptions are usually thrown when the drone does weird things in the sim, like crashing or flipping in guided mode. I can probably work around them, I'll be using manual mode in sim anyway. But the one exception which this PR fixes happens 100% of the time on startup so there's no workaround.

peterbarker:
@andyp1per could you make sure this doesn't cause a regression for you, please?

@LordPikaChu we don't like seeing merge commits in our commit list - this should just be a single commit... use `rebase` to move onto master, not `git pull`

andyp1per:
@peterbarker I don't see how this can not cause a regression since it undoes my change ...

peterbarker:
@LordPikaChu still have that merge commit in there.

@LordPikaChu Which platform is this?

We may need to have an __DARWIN__ check or whatever in here to avoid the problems.

LordPikaChu:
> @LordPikaChu still have that merge commit in there.
> 
> @LordPikaChu Which platform is this?
> 
> We may need to have an **DARWIN** check or whatever in here to avoid the problems.

It's an Intel Mac on Ventura. I've now looked through the codebase and it looks that these float casts by @andyp1per are all over the place. So I'll close the PR as it doesn't make sense to undo just a single one of these. Plus undoing them will make the program crash for people with other system versions/processors 😢 

@andyp1per what setup are you running on specifically?

I think I'll need to buy a Linux machine anyway to run Ardupilot. Or maybe experiment with VMs, which I'd probably have to do anyway to run multiple instances of Ardupilot for ML training.

PR #29817, opened 2025-04-19
AP_HAL_ChibiOS: VTX power pin moved on FlywooF405HD

andyp1per:
Reported by a user who also verified that the fix works


rmackay9:

PR #29780, opened 2025-04-14
AP_HAL_ChibiOS: add new hardware uav-dev-fc-um982

mirkix:
Add new hardware UAV-DEV GmbH autopilot uav-dev-fc-um982

uav-dev-fc-um982

- 50 mm x 50 mm
- STM32H743
- ICM45686
- DPS310
- BMM150
- UM982 L1/L2/L5 RTK GNSS with GNSS-heading
- USB-C
- Ethernet
- RCIN
- 6 x PWM / GPIO / serial ESC telemetry in
- 5 x serial
- 2 x CAN FD
- microSD

PR #29821, opened 2025-04-19
AP_Scripting: add gps:inject_data binding and NTRIP applet

yuri-rage:
Makes `AP_GPS::inject_data()` public and adds a Lua binding to it.

`net-ntrip.lua` applet makes use of that binding via a direct connection to an unsecure (http) NTRIP caster.

Tested on CUAV 7-Nano and Pixhawk v6X.

EDIT:
Failing a few chibios build tasks - perhaps I need an `#ifdef` guard - just not sure which to employ.

### DIFF
--- a/libraries/AP_Scripting/applets/net-ntrip.lua
+++ b/libraries/AP_Scripting/applets/net-ntrip.lua
@@ -1,0 +1,294 @@
+--[[
+    Script for direct connection to an NTRIP caster for Ethernet enabled autopilots.
+    The connected LAN must have WAN (Internet) access.
+    The NTRIP caster must allow http connections (will not work over https).
+
+    Edit CASTER_IP, CASTER_PORT, MOUNTPOINT, USERNAME, and PASSWORD to match
+    your NTRIP caster and credentials.
+
+    Edit NTRIP_VERSION as needed for your caster:
+        - "Ntrip/2.0" for modern casters (RTK2Go supports this)
+        - "Ntrip/1.0" for legacy casters (uses ICY 200 OK response)
+
+    Edit GGA_RATE_HZ to set the rate of GGA messages to send to the caster.
+    Set to 0 to disable sending own position to NTRIP caster.
+    Many NTRIP v1.0 casters require GGA messages for authentication.
+
+    RTK2Go is available at 3.143.243.81 port 2101
+]]
+
+local CASTER_IP             = "3.143.243.81"
+local CASTER_PORT           = 2101
+local MOUNTPOINT            = "MOUNTPOINT"
+local USERNAME              = "user-at-email.com"
+local PASSWORD              = "none" -- "none" is ok for RTK2G0
+local NTRIP_VERSION         = 2      -- 1 or 2
+local GGA_RATE_HZ           = 0      -- set 1 or 2 Hz as required, 0 to disable
+
+local MAV_SEVERITY          = { EMERGENCY = 0, ALERT = 1, CRITICAL = 2, ERROR = 3, WARNING = 4, NOTICE = 5, INFO = 6, DEBUG = 7 }
+
+-- map MAVLink GPS status to NMEA quality code
+local GPS_QUAL              = { [0] = 0, [1] = 0, [2] = 1, [3] = 1, [4] = 2, [5] = 2, [6] = 2 }
+
+local BOOT_DELAY_MS         = 15000 -- wait 15 seconds for GPS and network to initialize
+local ACK_TIMEOUT_MS        = 1000
+local CONNECTION_TIMEOUT_MS = 5000
+local MAX_RETRIES           = 3
+local RESET_MS              = 120000 -- wait 2 minutes after max retries
+local POLL_RATE_HZ          = 10
+local RECV_BLOCK_SIZE       = 1024
+
+local tcp_socket            = nil
+local retries               = 0
+local rtcm_buffer           = ""
+local last_data_time        = uint32_t(0)
+local last_gga_time         = uint32_t(0)
+local msg_count             = {}
+local send_debug_msg_count  = false
+
+-- Base64 encode a string
+local function base64_encode(input)
+    local b = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
+    return ((input:gsub('.', function(x)
+        local r, b_char = '', x:byte()
+        for i = 8, 1, -1 do r = r .. (b_char % 2 ^ i - b_char % 2 ^ (i - 1) > 0 and '1' or '0') end
+        return r
+    end) .. '0000'):gsub('%d%d%d?%d?%d?%d?', function(x)
+        if (#x < 6) then return '' end
+        local c = 0
+        for i = 1, 6 do c = c + (x:sub(i, i) == '1' and 2 ^ (6 - i) or 0) end
+        return b:sub(c + 1, c + 1)
+    end) .. ({ '', '==', '=' })[#input % 3 + 1])
+end
+
+local function process_rtcm_buffer()
+    -- RTCM message format:
+    -- Byte 0: Preamble (0xD3)
+    -- Byte 1-2: Message length (10 bits) + reserved (6 bits)
+    -- Bytes 3-N: Message data
+    -- Last 3 bytes: CRC
+
+    while #rtcm_buffer >= 3 do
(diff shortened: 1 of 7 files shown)


Hwurzburg:
no the tag is to remind me to add to the applets list....one line...nothing big

mirkix:
@yuri-rage Great, I just checked that with my FC with build in RTK GNSS https://github.com/ArduPilot/ardupilot/pull/29780 . One thing, my NTRIP provider (SAPOS in Germany) respond with `ICY 200 OK`, so I added that. But it not start sending RTCM data before I have send my position (GGA NMEA MSG).

rmackay9:
@yuri-rage,

Thanks for this.  Please tell us when you're ready and we can add to a dev call for final review

yuri-rage:
@rmackay9 first just need a little guidance on the chibios build failures. I think it's just a matter of an #ifdef, but not sure which.

Will take a look at @mirkix's issues as well - those are likely easier to overcome for a more robust implementation overall.

peterbarker (commented):
Shouldn't we be attempting to re-use ArduPilot's existing RTCM parsing?

My pr/ublox-parser-rewrite branch refactors the existing RTCMParser in such a way that it would be re-usable in scripting.  This is the comment:
```
+    // returns 0 if there's no packet.
+    // returns positive packet length if packet is found
+    // returns a negative number containing number of bytes to discard if packet not found
+    int16_t find_packet(const uint8_t *buffer, uint8_t buffer_len) const;
```

I'm not sure we should block this PR based on those musings - if it works, it's just a script which can people run if they feel like it.

What's the CPU overhead look like when using this?

yuri-rage:
@peterbarker I haven't specifically monitored CPU load when running this script, but I have run it concurrently with 3 others, 2 of which do a fair amount of comparison and potentially command processing on each cycle, and it runs without issue in parallel.

Agree that your rewrite would be useful. Unless you expect to merge very soon, we shouldn't block this PR for it.

At a minimum, I do think NTRIP v1.0 and GGA handling should be added before merge. As such, this is a WIP, based on comments above.

EDIT: Fixed up code review mistake (missing `#endif`).

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -1270,4 +1271,28 @@ int lua_DroneCAN_get_FlexDebug(lua_State *L)
 }
 #endif // HAL_ENABLE_DRONECAN_DRIVERS
 
+#if AP_GPS_ENABLED
+int lua_gps_inject_data(lua_State *L)
+{
+    binding_argcheck(L, 3);       // self + 2 args
+    luaL_checkudata(L, 1, "gps"); // require GPS object as first argument
+
+    const char *data = luaL_checkstring(L, 2);
```
tpwrules, line 1280: Use `luaL_checklstring` and drop the second argument? Pretty sure neither can return `nullptr`.

libraries/AP_Scripting/applets/net-ntrip.lua
```diff
@@ -0,0 +1,209 @@
...
+        end
+
+        local complete_message = rtcm_buffer:sub(1, total_length)
+        rtcm_buffer = rtcm_buffer:sub(total_length + 1)
+
+        gps:inject_data(complete_message, #complete_message)
+
+        if #rtcm_buffer >= 3 then
+            process_rtcm_buffer() -- recurse
+        end
```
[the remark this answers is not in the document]
yuri-rage: Yes. Oversight on my part after iterating over the logic structure a few times and leaving a redundant call here.

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -1270,4 +1271,28 @@ int lua_DroneCAN_get_FlexDebug(lua_State *L)
...
+
+    if (data == nullptr)
+    {
+        return luaL_error(L, "No data to inject");
+    }
+
+    if (len > 0)
+    {
+        AP::gps().inject_data((const uint8_t *)data, len);
+    }
```
tpwrules:
```suggestion
    binding_argcheck(L, 2);       // self + 1 arg
    luaL_checkudata(L, 1, "gps"); // require GPS object as first argument

    size_t len;
    const char *data = luaL_checklstring(L, 2, &len);

    if (len > 0 && len <= UINT16_MAX)
    {
        AP::gps().inject_data((const uint8_t *)data, (uint16_t)len);
    }
```

Please do it this way. Otherwise the user can pass a large length and potentially crash the autopilot. Lua is safe to have zeros in strings so I can't think of a reason to make the length an extra argument. `luaL_checklstring` cannot return a null pointer either, it raises a Lua error instead.

libraries/AP_Scripting/generator/description/bindings.desc
```diff
@@ -154,6 +154,7 @@ singleton AP_GPS method get_antenna_offset Vector3f uint8_t 0 ud->num_sensors()
 singleton AP_GPS method first_unconfigured_gps boolean uint8_t'Null
 singleton AP_GPS method gps_yaw_deg boolean uint8_t 0 ud->num_sensors() float'Null float'Null uint32_t'Null
 singleton AP_GPS method time_epoch_usec uint64_t uint8_t 0 ud->num_sensors()
+singleton AP_GPS manual inject_data lua_gps_inject_data 2 0
```
tpwrules:
```suggestion
singleton AP_GPS manual inject_data lua_gps_inject_data 1 0
```

tpwrules (changes requested):

libraries/AP_Scripting/lua_bindings.cpp
yuri-rage: You beat me to it. I just rebase/squashed and probably should've held off on pushing. Literally wrote what you have above in a pending commit, nearly verbatim! Thanks for the sanity check.

yuri-rage:
@mirkix, NTRIP v1.0 support added + GGA messaging to the caster

@tpwrules, binding modified to use `checklstring` and limit packet size.

@peterbarker, quite willing to use native parser in a future update if your rewrite is merged. CPU load is pretty low at the rate that this script updates (less than 1% change between scripting stopped and scripting started on an H757).

@rmackay9, with your blessing, this is ready for a Dev Call tag.

EDIT: No idea why this is failing the SITL Plane test for SPort telemetry. The added features do not touch that part of the codebase. Happy to make a change if anyone can point me in the correct direction.

libraries/AP_Scripting/applets/net-ntrip.lua
```diff
@@ -0,0 +1,297 @@
+--[[
+    Script for direct connection to an NTRIP caster for Ethernet enabled autopilots.
+    One port must be configured as a scripting TCP client:
+        NET_Px_PROTOCOL	28
+        NET_Px_TYPE	3
```
tpwrules:
```suggestion
```

This is only for using `serial:find_serial` in Lua which you are not; you are just using raw sockets.

libraries/AP_Scripting/applets/net-ntrip.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Configuration:
+
+Configure the autopilot with [DHCP or a static IP, proper gateway, and netmask](https://ardupilot.org/copter/docs/common-network.html).
+
+One port must be configured as a scripting TCP client:
+```
+    NET_Px_PROTOCOL	28
+    NET_Px_TYPE	3
+```
```
tpwrules:
```suggestion
```

As above

tpwrules (changes requested):
I was able to flash this to my Cube Red (incl. enabling DHCP), fill in the blanks to connect to my state network (incl. GGA), share an internet connection from my laptop, and get some RTK happening! It worked great on the bench.

That shows the hookup and basic script is sound. I did not try to fly, nor abuse the script for error handling validation. But I'm quite happy to merge this after you fix the little documentation issue.

Thanks for your work! This will be neat with the modem support, though there could be socket related deadlocks... But it's a script, it can be iterated and the C++ changes are clearly good.

yuri-rage:
@tpwrules appreciate your time on this! Documentation changes applied, commits squashed.

I've got several hours on my mower with this script in action. I think the only hiccups I've had are related to network bandwidth (using LOS wifi bridges whose bandwidth shrinks rapidly at longer range). And even then, I usually get a successful reconnection without further intervention.

tpwrules (approved):

PR #29824, opened 2025-04-20
Avoid using Baro in SITL library (nullptr deref running examples)

peterbarker:
nullptr dereference in some examples.

We shouldn't be using it over there anyway.

Similarly for Scheduler

... and a couple of other issues raised in the ToshbibaLED example specifically

### DIFF
--- a/libraries/AP_HAL_SITL/SITL_State.cpp
+++ b/libraries/AP_HAL_SITL/SITL_State.cpp
@@ -73,7 +73,6 @@ void SITL_State::_sitl_setup()
 
     fprintf(stdout, "Starting SITL input\n");
 
-    // find the barometer object if it exists
     _sitl = AP::sitl();
 
     if (_sitl != nullptr) {
@@ -319,7 +318,6 @@ void SITL_State::_simulator_servos(struct sitl_input &input)
     uint32_t now = AP_HAL::micros();
     last_update_usec = now;
 
-    float altitude = AP::baro().get_altitude();
     float wind_speed = 0;
     float wind_direction = 0;
     float wind_dir_z = 0;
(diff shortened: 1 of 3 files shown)


libraries/AP_HAL_SITL/SITL_State.cpp
```diff
@@ -346,6 +344,7 @@ void SITL_State::_simulator_servos(struct sitl_input &input)
         wind_dir_z =     _sitl->wind_dir_z_active;
         
         // pass wind into simulators using different wind types via param SIM_WIND_T*.
+        const float altitude = _sitl->state.height_agl;
```
Georacer, line 347: `AP:baro().get_altitude()` will always return altitude above home.
But `_silt->state.height_agl` has the capacity to contain actual AGL, right?

So in terms of wind calculation, that's even better! Nice.

libraries/AP_Notify/examples/ToshibaLED_test/ToshibaLED_test.cpp
```diff
@@ -13,13 +14,18 @@ AP_Notify notify;
 
 static ToshibaLED_I2C toshiba_led(1);
 
+#if CONFIG_HAL_BOARD == HAL_BOARD_SITL
+#include <SITL/SITL.h>
+SITL::SIM sitl;
+#endif
+
```
Georacer, line 21: I don't see why this is necessary here?

Georacer (changes requested):

libraries/AP_Notify/examples/ToshibaLED_test/ToshibaLED_test.cpp
peterbarker: The sitl object doesn't get instantiated anywhere else, and is used in <some codepath I can't recall but seemed reasonable>

We're trying to use the simulated i2c stuff in SITL and that code all assumes that AP::sitl() works.

libraries/AP_Notify/examples/ToshibaLED_test/ToshibaLED_test.cpp
```diff
@@ -13,13 +14,18 @@ AP_Notify notify;
...
 void setup(void)
 {
     // display welcome message
     hal.console->printf("Toshiba LED test ver 0.1\n");
 
     // initialise LED
-    if (toshiba_led.init()) {
+    if (!toshiba_led.init()) {
```
Ryanf55, line 28: Drive-by?

Ryanf55 (approved):

PR #29826, opened 2025-04-20
Fix and move get_fwd_pitch_is_limited()

lthall:
This fixes a couple problems:

- get_fwd_pitch_is_limited can be stuck on true if it becomes unsaturated while being true,
- it is only used in quadplane but runs on every loop in Copter,
- it was added to the main update XY position controller loop.

This PR fixes the bug and lest us only run it when called. I also want to remove it at compile time from copter but need some help to do this.

### DIFF
--- a/libraries/AC_AttitudeControl/AC_PosControl.cpp
+++ b/libraries/AC_AttitudeControl/AC_PosControl.cpp
@@ -710,13 +710,6 @@ void AC_PosControl::update_xy_controller()
     if (!limit_accel_xy(_vel_desired.xy(), _accel_target.xy(), accel_max)) {
         // _accel_target was not limited so we can zero the xy limit vector
         _limit_vector.xy().zero();
-    } else {
-        // Check for pitch limiting in the forward direction
-        const float accel_fwd_unlimited = _limit_vector.x * _ahrs.cos_yaw() + _limit_vector.y * _ahrs.sin_yaw();
-        const float pitch_target_unlimited = accel_to_angle(- MIN(accel_fwd_unlimited, accel_max) * 0.01f) * 100;
-        const float accel_fwd_limited = _accel_target.x * _ahrs.cos_yaw() + _accel_target.y * _ahrs.sin_yaw();
-        const float pitch_target_limited = accel_to_angle(- accel_fwd_limited * 0.01f) * 100;
-        _fwd_pitch_is_limited = is_negative(pitch_target_unlimited) && pitch_target_unlimited < pitch_target_limited;
     }
 
     // update angle targets that will be passed to stabilize controller
@@ -1420,6 +1413,24 @@ float AC_PosControl::crosstrack_error() const
         return safe_sqrt(MAX(pos_error.length_squared() - sq(dot_error), 0.0));
     }
 }
+
+#if APM_BUILD_TYPE(APM_BUILD_ArduPlane)
+/// returns true when the forward pitch demand is limited by the maximum allowed tilt
+bool AC_PosControl::get_fwd_pitch_is_limited() const 
+{
+    if (_limit_vector.xy().is_zero()) {  
+        return false;  
+    }  
+    const float angle_max = MIN(_attitude_control.get_althold_lean_angle_max_cd(), get_lean_angle_max_cd());
+    const float accel_max = angle_to_accel(angle_max * 0.01) * 100;
+    // Check for pitch limiting in the forward direction
+    const float accel_fwd_unlimited = _limit_vector.x * _ahrs.cos_yaw() + _limit_vector.y * _ahrs.sin_yaw();
+    const float pitch_target_unlimited = accel_to_angle(- MIN(accel_fwd_unlimited, accel_max) * 0.01f) * 100;
+    const float accel_fwd_limited = _accel_target.x * _ahrs.cos_yaw() + _accel_target.y * _ahrs.sin_yaw();
+    const float pitch_target_limited = accel_to_angle(- accel_fwd_limited * 0.01f) * 100;
+    return is_negative(pitch_target_unlimited) && pitch_target_unlimited < pitch_target_limited;
+}
+#endif // APM_BUILD_TYPE(APM_BUILD_ArduPlane)
 
 ///
 /// private methods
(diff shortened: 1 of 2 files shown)


andyp1per:
@lthall you can't add the #if to the header file, but you should be able to add it to the cpp file safely to achieve the effect you want

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1421,6 +1414,25 @@ float AC_PosControl::crosstrack_error() const
     }
 }
 
+//#if APM_BUILD_TYPE(APM_BUILD_ArduPlane)
```
peterbarker, line 1417: Leave these in; may need a `#else`

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1421,6 +1414,25 @@ float AC_PosControl::crosstrack_error() const
...
+/// returns true when the forward pitch demand is limited by the maximum allowed tilt
+bool AC_PosControl::get_fwd_pitch_is_limited() const 
+{
+    bool fwd_pitch_is_limited = false;
+    if (!_limit_vector.xy().is_zero()) {
```
peterbarker, line 1422:
```suggestion
    if (_limit_vector.xy().is_zero()) {
        return false;
    }
```

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1421,6 +1414,25 @@ float AC_PosControl::crosstrack_error() const
...
+        float angle_max = MIN(_attitude_control.get_althold_lean_angle_max_cd(), get_lean_angle_max_cd());
+        float accel_max = angle_to_accel(angle_max * 0.01) * 100;
+        // Check for pitch limiting in the forward direction
+        const float accel_fwd_unlimited = _limit_vector.x * _ahrs.cos_yaw() + _limit_vector.y * _ahrs.sin_yaw();
+        const float pitch_target_unlimited = accel_to_angle(- MIN(accel_fwd_unlimited, accel_max) * 0.01f) * 100;
+        const float accel_fwd_limited = _accel_target.x * _ahrs.cos_yaw() + _accel_target.y * _ahrs.sin_yaw();
+        const float pitch_target_limited = accel_to_angle(- accel_fwd_limited * 0.01f) * 100;
+        fwd_pitch_is_limited = is_negative(pitch_target_unlimited) && pitch_target_unlimited < pitch_target_limited;
```
peterbarker, line 1430:
```suggestion
        return is_negative(pitch_target_unlimited) && pitch_target_unlimited < pitch_target_limited;
```

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1421,6 +1414,25 @@ float AC_PosControl::crosstrack_error() const
     }
 }
 
+//#if APM_BUILD_TYPE(APM_BUILD_ArduPlane)
+/// returns true when the forward pitch demand is limited by the maximum allowed tilt
+bool AC_PosControl::get_fwd_pitch_is_limited() const 
+{
+    bool fwd_pitch_is_limited = false;
+    if (!_limit_vector.xy().is_zero()) {
+        float angle_max = MIN(_attitude_control.get_althold_lean_angle_max_cd(), get_lean_angle_max_cd());
```
peterbarker, line 1423:
```suggestion
        const float angle_max = MIN(_attitude_control.get_althold_lean_angle_max_cd(), get_lean_angle_max_cd());
```

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1421,6 +1414,25 @@ float AC_PosControl::crosstrack_error() const
...
+        float accel_max = angle_to_accel(angle_max * 0.01) * 100;
```
peterbarker, line 1424:
```suggestion
        const float accel_max = angle_to_accel(angle_max * 0.01) * 100;
```

peterbarker (changes requested):

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -710,13 +710,6 @@ void AC_PosControl::update_xy_controller()
     if (!limit_accel_xy(_vel_desired.xy(), _accel_target.xy(), accel_max)) {
         // _accel_target was not limited so we can zero the xy limit vector
         _limit_vector.xy().zero();
```
IamPete1, line 712: If I understand the bug correctly it is the `_fwd_pitch_is_limited` is not set in this code path. A alternate fix would be to add `_fwd_pitch_is_limited = false;`  here?

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1421,6 +1414,24 @@ float AC_PosControl::crosstrack_error() const
     }
 }
 
+#if APM_BUILD_TYPE(APM_BUILD_ArduPlane)
```
IamPete1, line 1417: You shouldn't need this, if its not called copter will not include the code. Because the function is const calling it cannot mess-up anything in the pos controller.

IamPete1 (approved):
This is a exact like for like move of the logic. 

Currently we only call this in one spot in quadplane, so there is no loss to move it out of the update loop. There is quite a bit of maths in the function so if we found ourselves calling it more then it might be worth putting it back into the update loop.

libraries/AC_AttitudeControl/AC_PosControl.cpp
lthall: Leaving this in copter would not be a good thing in my humble opinion. It has nothing to do with copter.

IamPete1: Right, I agree with the move, just trying to understand the bugfix.

libraries/AC_AttitudeControl/AC_PosControl.cpp
peterbarker: Sure, but Leonard *really* doesn't want this code on Copter and this ensures that :-)

PR #29833, opened 2025-04-20
AP_DDS: Added attitude target topic and namespace representation for global constants

snktshrma:
This adds attitude target topic, /ap/experimental/target_attitude and updates global constant as per this PR: https://github.com/ArduPilot/Micro-XRCE-DDS-Gen/pull/4


Ryanf55:
Before merging this, we need to get upstream changes merged. I'd like the API-changing stuff to go in as a separate PR rather than bundling that with a new feature. Overall, this looks good otherwise. It will need some docs and tests to show how it works and ensure it doesn't break.

snktshrma:
Alright @Ryanf55 ! I'll add some tests and probably put up a new PR for the API related changes. Upstream and that API changes will be required for this attitude target to be merge

tizianofiorenzani:
@snktshrma do you plan to get this merged?

libraries/AP_DDS/AP_DDS_ExternalControl.cpp
```diff
@@ -44,6 +44,47 @@ bool AP_DDS_External_Control::handle_global_position_control(ardupilot_msgs_msg_
...
 }
 
+bool AP_DDS_External_Control::handle_attitude_control(ardupilot_msgs_msg_AttitudeTarget& cmd_att)
+{
+    auto *external_control = AP::externalcontrol();
+    if (external_control == nullptr) {
+        return false;
+    }
+
+    const bool roll_rate_ignore   = cmd_att.type_mask & AttitudeTarget::IGNORE_ROLL_RATE;
```
tizianofiorenzani: If the result of ignoring a rate is setting it to zero, why not demanding that in the first place?

Tools/ros2/ardupilot_msgs/msg/AttitudeTarget.msg
```diff
@@ -0,0 +1,14 @@
+# Control the attitude of the vehicle
+# Frame - Body
+std_msgs/Header header
+
+uint8 type_mask
+uint8 IGNORE_ROLL_RATE = 1	# body_rate.x
+uint8 IGNORE_PITCH_RATE = 2	# body_rate.y
+uint8 IGNORE_YAW_RATE = 4	# body_rate.z
```
tizianofiorenzani, line 8: This might be already defined in GlobalPosition.msg and result in a conflict. Also, could we just set those as individual bools? We don't have to replicate the Mavlink Message.

libraries/AP_DDS/AP_DDS_ExternalControl.cpp
```diff
@@ -44,6 +44,47 @@ bool AP_DDS_External_Control::handle_global_position_control(ardupilot_msgs_msg_
...
+        }
+    }
+
+    Vector3f ang_vel_body;
+    // Ensure the ang-vel vector is always populated
+    ang_vel_body.x = roll_rate_ignore ? 0.0f : cmd_att.body_rate.x;
+    ang_vel_body.y = pitch_rate_ignore ? 0.0f : cmd_att.body_rate.y;
+    ang_vel_body.z = yaw_rate_ignore ? 0.0f : cmd_att.body_rate.z;
+
+    if (roll_rate_ignore && pitch_rate_ignore && yaw_rate_ignore) {
```
tizianofiorenzani, line 120: Why are we forced to providing a rate? In the GCS_Mavlink.cpp, when handling `MAVLINK_MSG_ID_SET_ATTITUDE_TARGET`, this is not required. We might also get rid of those flags altogether, as they just result in setting the rate component to zero.

tizianofiorenzani (changes requested):
It works, but I have some comments

Tools/ros2/ardupilot_msgs/msg/AttitudeTarget.msg
```diff
@@ -0,0 +1,14 @@
+# Control the attitude of the vehicle
+# Frame - Body
```
tizianofiorenzani, line 2: We might remind that we are talking about F-L-U body frame and orientation is respect to E-N-U

libraries/AP_DDS/AP_DDS_ExternalControl.cpp
```diff
@@ -44,6 +44,47 @@ bool AP_DDS_External_Control::handle_global_position_control(ardupilot_msgs_msg_
...
+    // thrust field should not be ignored
+    if (throttle_ignore) {
+        return false;
+    }
+
+    Quaternion attitude_quat;
+    if (attitude_ignore) {
+        attitude_quat.zero();
+    } else {
+        attitude_quat = Quaternion(cmd_att.orientation.w,cmd_att.orientation.x,cmd_att.orientation.y,cmd_att.orientation.z);
```
tizianofiorenzani: Quaternion should be rotated to FRD/NED here

libraries/AP_DDS/AP_DDS_ExternalControl.cpp
```diff
@@ -44,6 +44,47 @@ bool AP_DDS_External_Control::handle_global_position_control(ardupilot_msgs_msg_
...
+        if (!attitude_quat.is_unit_length()) {
+            return false;
+        }
+    }
+
+    Vector3f ang_vel_body;
+    // Ensure the ang-vel vector is always populated
+    ang_vel_body.x = roll_rate_ignore ? 0.0f : cmd_att.body_rate.x;
```
tizianofiorenzani: Body rate is in FLU, to be rotated to FRD

Tools/ros2/ardupilot_msgs/msg/AttitudeTarget.msg
snktshrma: Hi, the conflict shouldn't be an issue. It has been resolved in this PR, adding namespaced represenstation: https://github.com/ArduPilot/Micro-XRCE-DDS-Gen/pull/5
Regarding bools, yep we can do so, just following the existing format of DDS messages. Do you suggest we should do to all existing typemasks as well? Let me know

libraries/AP_DDS/AP_DDS_ExternalControl.cpp
snktshrma: Thanks for suggesting. So I tried to follow the exiting standard. The ignore mask is following mavlink and existing typemask convention.This is mainly to keep intent explicit.

The existing mavlink interface where the case when an ignore flag is set and other is not, it considers it ambiguous. https://github.com/ArduPilot/ardupilot/blob/7897a74ba329d7a7a82c57774a5c3dca4220eebd/ArduCopter/GCS_MAVLink_Copter.cpp#L998
I tried to improve upon that by setting the rates as 0 instead of considering the command ambiguous and the controller handles that it seems. https://github.com/ArduPilot/ardupilot/blob/7897a74ba329d7a7a82c57774a5c3dca4220eebd/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp#L303

Though I have also not considered case where all flags are set which is also not a good case (both for mavlink as well as DDS interface)

Seems like this part be discussed (whole typemask stuff)

ZhongmouLi:
@snktshrma Hi, thanks so much for your PR and may I ask if you could allow me to push to your PR?

Tools/ros2/ardupilot_msgs/msg/AttitudeTarget.msg
ZhongmouLi: @snktshrma , I have solved the frame problems for the attitude and angular velocity fixing the frame transformation from ROS2 (ENU/FLU) to ArduPilot (NED/FRD)

Since your PR does not allow me to 'push', could you please pull my changes from git@github.com:ZhongmouLi/ardupilot.git.

snktshrma: Hi @ZhongmouLi ! Thanks for your contribution. I didn't get time to complete this. I’ll get your commit in this PR.
Thanks

Issue #29778, opened 2025-04-14
Plane: Pitch stick mixing doesn't conserve energy


rubenp02:
## Bug report

**Issue details**
Stick mixing operates by changing the attitude demand, which is fine for roll but can have really bad consequences for pitch, as the system is decoupled from the TECS. If a stick nudge is held long enough, an overspeed or a stall WILL occur. This could be worked around by weighting the mix with the flight energy, but ideally the pitch mixing logic should live in the TECS itself and operate by nudging the altitude rate instead of the pitch directly.

Since there's a removed "DIRECT" stick mixing mode that falls back to the FBW logic, this could become the way to use the current stick mixing logic with FBWA-style pitch control, and a new "FBWBMixing" mode that uses the TECS-based pitch mixing could replace the "FBWMixing" mode, and be the new default. The "DIRECT" mode would be renamed to "FBWAMixing".

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[X] Plane
[  ] Rover
[  ] Submarine

Georacer:
Ah, I got confused with servo mixers, my apologies.

I must admit I haven't used stick mixing much. But indeed currently I would expect them to act upon angle, not climb rate.
This is simpler, because the attitude controller doesn't hold a lot of state. Whereas it would be harder to reset TECS back onto the original trajectory.

I agree that an 'FBWBMixer" would be the safer option.

LupusTheCanine:
1. We probably need TECS rework to support more "edge of envelope" though automatically finding max flight path angle velocity will likely be quite hard without a characterization flight and there would be need to update it if weight/performance changes.
2. I would do a combination ±0.9 is either linear or exponential altitude offset and edges of the range are maximum vertical speed climb/descent. If we exceed offset range switch to half range vertical speed control (unrestricted climb/descent) where midpoint means return to setpoint at maximum rate, ±0.5 maintains altitude and ±1 moves away from setpoint at maximum rate. Switchback to offset mode would occur when demanded offset would be greater than current offset (minding the sign of correction).    
I would provide option to allow or inhibit unrestricted climb/descent though I am not sure which would be better.

PS modern RC systems aren't that noisy.

peterbarker:
The current "damn the controllers, full speed [up]" approach is there quite on purpose.

[...]

PR #29842, opened 2025-04-21
Plane: Add roll-only stick mixing

rubenp02:
The current implementation of pitch stick mixing is unsafe, as it introduces an attitude override that is decoupled from the TECS, and also lacks any stall and overspeed prevention mechanisms. This means that if a pitch nudge is held long enough, an overspeed or a stall will occur.

This commit introduces a "FBW style (roll only)" setting to the STICK_MIXING parameter. It behaves like the current "FBWMixing" option for roll, but doesn't provide any pitch mixing. This is set as the default for improved safety.

Pitch stick mixing can be re-enabled by setting STICK_MIXING to "FBW-A style". This option replaces the removed "Direct" stick mixing mode in the enum. A warning about the lack of flight envelope protections in this mode has also been added to the description.

Related issue: https://github.com/ArduPilot/ardupilot/issues/29778

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -90,8 +90,8 @@ const AP_Param::Info Plane::var_info[] = {
 
     // @Param: STICK_MIXING
     // @DisplayName: Stick Mixing
-    // @Description: When enabled, this adds user stick input to the control surfaces in auto modes, allowing the user to have some degree of flight control without changing modes.  There are two types of stick mixing available. If you set STICK_MIXING to 1 then it will use "fly by wire" mixing, which controls the roll and pitch in the same way that the FBWA mode does. This is the safest option if you usually fly ArduPlane in FBWA or FBWB mode. If you set STICK_MIXING to 3 then it will apply to the yaw while in quadplane modes only, such as while doing an automatic VTOL takeoff or landing.
-    // @Values: 0:Disabled,1:FBWMixing,3:VTOL Yaw only
+    // @Description: When enabled, this adds user stick input to the control surfaces in auto modes, allowing the user to have some degree of flight control without changing modes. There are 3 types of stick mixing available. If you set STICK_MIXING to 1 or 4 then it will use "fly by wire" mixing. 4 will provide roll and yaw control, while 1 also provides FBW-A style pitch control. If you set STICK_MIXING to 3 then it will apply to the yaw while in quadplane modes only, such as while doing an automatic VTOL takeoff or landing. WARNING: FBW-A pitch control does not offer flight envelope protections. Prolonged pitch inputs in mode 1 can result in a stall or overspeed condition, and should be avoided.
+    // @Values: 0:Disabled,1:FBW style,3:VTOL Yaw only,4:FBW style (no pitch)
     // @User: Advanced
     GSCALAR(stick_mixing,           "STICK_MIXING",   uint8_t(StickMixing::FBW)),
 
--- a/ArduPlane/defines.h
+++ b/ArduPlane/defines.h
@@ -56,6 +56,7 @@ enum class StickMixing {
     FBW      = 1,
     DIRECT_REMOVED = 2,
     VTOL_YAW = 3,
+    FBW_NO_PITCH = 4,
 };
 
 // values for RTL_AUTOLAND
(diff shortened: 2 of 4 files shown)


ArduPlane/defines.h
```diff
@@ -53,8 +53,8 @@ enum failsafe_action_long {
 // type of stick mixing enabled
 enum class StickMixing {
     NONE     = 0,
-    FBW      = 1,
-    DIRECT_REMOVED = 2,
+    FBW_ROLL = 1,
+    FBWA     = 2,
```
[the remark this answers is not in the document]
IamPete1: I don't think we can change the existing enum values, users might still have them set making this a change in behavior.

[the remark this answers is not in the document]
rubenp02: Fixed

IamPete1 (approved):
LGTM

ArduPlane/mode.cpp
```diff
@@ -260,7 +260,9 @@ void Mode::run()
 {
     // Direct stick mixing functionality has been removed, so as not to remove all stick mixing from the user completely
     // the old direct option is now used to enable fbw mixing, this is easier than doing a param conversion.
-    if ((plane.g.stick_mixing == StickMixing::FBW) || (plane.g.stick_mixing == StickMixing::DIRECT_REMOVED)) {
+    if ((plane.g.stick_mixing == StickMixing::FBW) ||
```
peterbarker: `switch` (no default, have each value in there)

ArduPlane/defines.h
```diff
@@ -56,6 +56,7 @@ enum class StickMixing {
     FBW      = 1,
     DIRECT_REMOVED = 2,
     VTOL_YAW = 3,
+    FBW_ROLL = 4,
```
tridge: i think this is actually roll and yaw, not just roll

ArduPlane/Parameters.cpp
```diff
@@ -90,8 +90,8 @@ const AP_Param::Info Plane::var_info[] = {
...
-    // @Values: 0:Disabled,1:FBWMixing,3:VTOL Yaw only
+    // @Description: When enabled, this adds user stick input to the control surfaces in auto modes, allowing the user to have some degree of flight control without changing modes. There are 3 types of stick mixing available. If you set STICK_MIXING to 1 or 4 then it will use "fly by wire" mixing. 4 will only provide roll control, while 1 also provides FBW-A style pitch control. If you set STICK_MIXING to 3 then it will apply to the yaw while in quadplane modes only, such as while doing an automatic VTOL takeoff or landing. WARNING: FBW-A pitch control does not offer flight envelope protections. Prolonged pitch inputs in mode 1 can result in a stall or overspeed condition, and should be avoided.
+    // @Values: 0:Disabled,1:FBW style,3:VTOL Yaw only,4:FBW style (roll only)
```
tridge: roll and yaw

tridge (changes requested):
it gives rudder as well, so roll and rudder

peterbarker (changes requested):

ArduPlane/mode.cpp
rubenp02: Just to be sure, this is what it would look like (if I understood correctly):

```
void Mode::run()
{
    // Direct stick mixing functionality has been removed, so as not to remove all stick mixing from the user completely
    // the old direct option is now used to enable fbw mixing, this is easier than doing a param conversion.
    switch ((StickMixing)plane.g.stick_mixing) {
        case StickMixing::FBW:
        case StickMixing::FBW_ROLL:
        case StickMixing::DIRECT_REMOVED:
            plane.stabilize_stick_mixing_fbw();
            break;
        case StickMixing::NONE:
        case StickMixing::VTOL_YAW:
            break;
    }
    plane.stabilize_roll();
    plane.stabilize_pitch();
    plane.stabilize_yaw();
}
```

If this is what you wanted I can push it. I have to say that it seems a less idiomatic to me.

Edit: Nevermind, just pushed it. If you end up not liking it I'll change it back.

rubenp02:
If this gets merged, maybe you would also like to look into #29841, which fixes a bug in the pitch stick mixing.

peterbarker:
Merged, thanks!

PR #29843, opened 2025-04-21
Plane: Use accurate roll limits if airspd. sensor

rubenp02:
Aerodynamic load factor-based roll demand limits were only applied down to ±25º to keep the aircraft maneuverable with a bad airspeed estimate. However, with the addition of the AIRSPEED_STALL parameter, this is unsafe if the aircraft is actually flying very close to its stall speed, and can result in a spin (this has actually happened to us in a real flight flying with a bit of wind and turbulence with the airspeed target set to AIRSPEED_MIN).

This PR makes it possible for the roll demand limits to go all the way down to 0º if an airspeed sensor is fitted and in use and the AIRSPEED_STALL parameter is set, as under these conditions, the airspeed estimate is accurate and there is very little bank angle overhead to prevent a stall.

### DIFF
--- a/ArduPlane/Attitude.cpp
+++ b/ArduPlane/Attitude.cpp
@@ -685,34 +685,34 @@ void Plane::apply_load_factor_roll_limits(void)
         flight_option_enabled(FlightOptions::ENABLE_FULL_AERO_LF_ROLL_LIMITS) &&
         ahrs.using_airspeed_sensor() && is_positive(aparm.airspeed_stall);
 
-    const int32_t level_roll_limit_cd = g.level_roll_limit * 100;
-    int32_t lf_roll_limit_cd = aparm.roll_limit * 100;
+    const float level_roll_limit_deg = g.level_roll_limit;
+    float lf_roll_limit_deg = aparm.roll_limit;
     if (max_load_factor <= 1) {
         if (enforce_full_roll_limit) {
-            lf_roll_limit_cd = level_roll_limit_cd;
+            lf_roll_limit_deg = level_roll_limit_deg;
         } else {
             // 25° limit to ensure maneuverability if airspeed estimate is wrong
-            lf_roll_limit_cd = 2500;
+            lf_roll_limit_deg = 25;
         }
     } else if (max_load_factor < aerodynamic_load_factor) {
         // the demanded nav_roll would take us past the aerodynamic
         // load limit. Limit our roll to a bank angle that will keep
         // the load within what the airframe can handle.
-        lf_roll_limit_cd = degrees(acosf(1.0f / max_load_factor))*100;
+        lf_roll_limit_deg = degrees(acosf(1.0f / max_load_factor));
 
         // unless enforcing full limits, allow at least 25 degrees of roll to
         // ensure the aircraft can be manoeuvered with a bad airspeed estimate.
         // At 25 degrees the load factor is 1.1 (10%)
-        if (!enforce_full_roll_limit && lf_roll_limit_cd < 2500) {
-            lf_roll_limit_cd = 2500;
+        if (!enforce_full_roll_limit && lf_roll_limit_deg < 25) {
+            lf_roll_limit_deg = 25;
         }
 
         // always allow at least the wings level threshold to prevent flyaways
-        if (lf_roll_limit_cd < level_roll_limit_cd) {
-            lf_roll_limit_cd = level_roll_limit_cd;
+        if (lf_roll_limit_deg < level_roll_limit_deg) {
+            lf_roll_limit_deg = level_roll_limit_deg;
         }
     }
 
-    nav_roll_cd = constrain_int32(nav_roll_cd, -lf_roll_limit_cd, lf_roll_limit_cd);
-    roll_limit_cd = MIN(roll_limit_cd, lf_roll_limit_cd);
+    nav_roll_cd = constrain_int32(nav_roll_cd, -lf_roll_limit_deg * 100, lf_roll_limit_deg * 100);
+    roll_limit_cd = MIN(roll_limit_cd, lf_roll_limit_deg * 100);
 }
--- a/ArduPlane/Attitude.cpp
+++ b/ArduPlane/Attitude.cpp
@@ -671,23 +671,41 @@ void Plane::update_load_factor(void)
     float max_load_factor =
         sq(smoothed_airspeed / MAX(stall_airspeed_1g, 1));
 
+    // allow limiting roll command down to wings-level when necessary if
+    // airspeed is accurate and airspeed stall is set (implying low load factor
+    // overhead)
+    const bool enforce_full_roll_limit =
+        flight_option_enabled(FlightOptions::ENABLE_FULL_AERO_LF_ROLL_LIMITS) &&
+        ahrs.using_airspeed_sensor() && is_positive(aparm.airspeed_stall);
+
+    const int32_t level_roll_limit_cd = g.level_roll_limit * 100;
+    int32_t lf_roll_limit_cd = aparm.roll_limit * 100;
     if (max_load_factor <= 1) {
-        // our airspeed is below the minimum airspeed. Limit roll to
-        // 25 degrees
-        nav_roll_cd = constrain_int32(nav_roll_cd, -2500, 2500);
-        roll_limit_cd = MIN(roll_limit_cd, 2500);
+        if (enforce_full_roll_limit) {
+            lf_roll_limit_cd = level_roll_limit_cd;
+        } else {
+            // 25° limit to ensure maneuverability if airspeed estimate is wrong
+            lf_roll_limit_cd = 2500;
+        }
     } else if (max_load_factor < aerodynamic_load_factor) {
         // the demanded nav_roll would take us past the aerodynamic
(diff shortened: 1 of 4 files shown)


IamPete1:
In general limiting to 0 is a bad idea because it will result in no navigation and thus a flyaway. We use `LEVEL_ROLL_LIMIT` for this in in a few places (takeoff, landing, RTL initial climb). The idea being that there is still some navigation so the vehicle will at least make some progress in the desired direction and give the user time to figure out what is going on (eg a airspeed failure).

LupusTheCanine:
AIRSPEED_MIN should be set to >1.25 V_Stall, this gives you 56% margin and 25° bank requires 10% (1.1 load factor).

See https://ardupilot.org/plane/docs/airspeed-parameters-setup.html.

Maybe we should add a note to the description that the parameter MUST NOT be set to less than V_Stall*1.05+appropriate gust resilience margin).

LupusTheCanine:
> I disagree a bit. Firstly, IMO it's handy for AIRSPEED_MIN/MAX to simply be the airspeed range for navigation, instead of also having performance/flight envelope considerations.

AIRSPEED_MIN/MAX is primarily TECS envelope configuration parameter. IIRC the behaviour is that unless AIRSPEED_STALL is set, AIRSPEED_MIN is used as a base for roll limit so with stricter stall protection and AIRSPEED_STALL not set you would get no navigation at AIRSPEED_MIN (which is really bad). We can't assume users are so stupid to set AIRSPEED_MIN below actual stall speed.

> There's AIRSPEED_STALL for that. Secondly, only relying on speed safety margins to always keep a 25º roll margin isn't enough, because if there's an issue with speed control and the aircraft goes below AIRSPEED_MIN and into AIRSPEED_STALL, then it definitely should not command any roll because it will stall. There should be safety margins, but also defensive logic.

Defensive logic should be such that we never knowingly decelerate below AIRSPEED_MIN in airspeed controlled modes. Though that would cause undesirable unconstrained descent in the case of certain airspeed sensor failures, rectifiable only by switching to FBWA and lower or disabling airspeed sensor. 
I don't think we have enough incident data to identify more likely failure mode between propulsion and airspeed sensor which would be necessary to decide which protection is more important, altitude or airspeed.

ArduPlane/Attitude.cpp
```diff
@@ -670,23 +670,40 @@ void Plane::update_load_factor(void)
     float max_load_factor =
         sq(smoothed_airspeed / MAX(stall_airspeed_1g, 1));
 
+    // allow limiting roll command down to wings-level when necessary if
+    // airspeed is accurate and airspeed stall is set (implying low load factor
+    // overhead)
+    bool enforce_full_roll_limit =
+        ahrs.using_airspeed_sensor() && is_positive(aparm.airspeed_stall);
+
+    int32_t lf_roll_limit_cd = aparm.roll_limit * 100;
```
peterbarker: The new stuff here could probably be done as float degrees rather that int32 floats.

ArduPlane/Attitude.cpp
```diff
@@ -670,23 +670,40 @@ void Plane::update_load_factor(void)
     float max_load_factor =
         sq(smoothed_airspeed / MAX(stall_airspeed_1g, 1));
 
+    // allow limiting roll command down to wings-level when necessary if
+    // airspeed is accurate and airspeed stall is set (implying low load factor
+    // overhead)
+    bool enforce_full_roll_limit =
```
peterbarker:
```suggestion
    const bool enforce_full_roll_limit =
```

tridge (changes requested):
I think this would have to be a new FLIGHT_OPTIONS bit, as otherwise we are sure to get some upset users who find that couldn't turn the aircraft to avoid a tree/hill/building/person when they could previously.
I'd also like it to be refactored a bit, to split out the "2500" in many places to be a single limit value at the top, which is either 25 degrees or level_roll_limit.
I'd also prefer this code be converted to degrees instead of centi-degrees, but I won't insist on that for this PR if you don't want to do that

peterbarker (changes requested):
Maybe calculate the limits separately to the load factor?  Straight afterwards?

ArduPlane/Attitude.cpp
```diff
@@ -670,23 +670,40 @@ void Plane::update_load_factor(void)
...
         // the demanded nav_roll would take us past the aerodynamic
         // load limit. Limit our roll to a bank angle that will keep
-        // the load within what the airframe can handle. We always
-        // allow at least 25 degrees of roll however, to ensure the
-        // aircraft can be manoeuvered with a bad airspeed estimate. At
-        // 25 degrees the load factor is 1.1 (10%)
-        int32_t roll_limit = degrees(acosf(1.0f / max_load_factor))*100;
-        if (roll_limit < 2500) {
-            roll_limit = 2500;
+        // the load within what the airframe can handle.
```
tridge, line 700: a separate function for the roll limit, called from the load_factor() function

rubenp02:
Just added the FLIGHT_OPTION and refactored so that roll limits are applied in a different method, as discussed in the dev call.

ArduPlane/Attitude.cpp
rubenp02: Normally this isn't done in other parts of the codebase, no? Is there a particular reason to use const here?

ArduPlane/Attitude.cpp
rubenp02: I thought this made sense on the dev call but looking at it closer, floats are already being used where it matters, and the end result must be a cdeg int, so why use floats?

ArduPlane/Attitude.cpp
rubenp02: Oh, I added it as an entirely separate method and called it after update_load_factor in all the callsites. Thought the idea was to make both operations independent. If you prefer it to be called from update_load_factor I can change it back to that.

ArduPlane/Attitude.cpp
peterbarker: `const` anything everywhere, all at once.

We might not use best practises everywhere, but we can move towards that goal :-)

ArduPlane/Attitude.cpp
peterbarker: We'll be changing the other things to floats relatively soon, probably in the next 6 months or so.  Introducing more stuff to change isn't right at this point.

rubenp02: Ah, great, I'll change it then. Out of curiosity, why the change to floats? I would personally think integers are better and more reliable for this application (and similar others in the rest of the codebase), despite being a bit worse in terms of usability.

peterbarker: > Ah, great, I'll change it then. Out of curiosity, why the change to floats? I would personally think integers are better and more reliable for this application (and similar others in the rest of the codebase), despite being a bit worse in terms of usability.

Nobody likes centidegrees.  It really is a usability thing for the most part (I vaguely recall there being problems around discretization of values and bias introduction, but I think that was on something else using integers where we moved to floats.

ArduPlane/Attitude.cpp
rubenp02: Done, also made level_roll_limit_deg const

ArduPlane/Attitude.cpp
rubenp02: Changed to floats

ArduPlane/mode_fbwb.cpp
```diff
@@ -18,6 +18,7 @@ void ModeFBWB::update()
     // Thanks to Yury MonZon for the altitude limit code!
     plane.nav_roll_cd = plane.channel_roll->norm_input() * plane.roll_limit_cd;
     plane.update_load_factor();
+    plane.apply_load_factor_roll_limits();
```
tridge: better to call from end of update_load_factor()

ArduPlane/Parameters.cpp
```diff
@@ -1082,6 +1082,7 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Bitmask: 13: Indicate takeoff waiting for neutral rudder with flight control surfaces
     // @Bitmask: 14: In AUTO - climb to next waypoint altitude immediately instead of linear climb
     // @Bitmask: 15: Use minimum of target and actual speed for flap setting
+    // @Bitmask: 16: Enable full aerodynamic load factor-based roll limits. This should be enabled when an accurate and well calibrated airspeed sensor is used to impose the correct FBW roll limits when the aircraft is turning. Requires an enabled airspeed sensor and AIRSPEED_STALL set. WARNING - If your airspeed sensor is very inaccurate or fails, having this enabled may result in almost no roll maneuverability in FBW modes.
```
tridge: i suspect this won't display in MissionPlanner, too long. Better for the detail to go into the wiki

tridge (approved):

rmackay9:
@rubenp02, the title of this PR makes it difficult to know how it should appear in the 4.7 release notes.

rubenp02:
> @rubenp02, the title of this PR makes it difficult to know how it should appear in the 4.7 release notes.

Hi Randy, how about "Added optional full roll limiting near stall using airspeed sensor"?

PR #23503, opened 2023-04-15
AP_Periph: add node locate function

mirkix:
This add a new function to AP_Periph. If a heartbeat LED is defined, a new DroneCAN parameter `locate` will appear. When set `locate` to 1 the heartbeat LED starts flashing rapidly. This can be used to locate the node in your hardware setup if you have more than one DroneCan nodes in your setup.

PR #29844, opened 2025-04-21
AP_Periph: add node visual locate function

mirkix:
This add a new function to AP_Periph. If a heartbeat LED is defined, with set Bit 3 in AP_Periph DEBUG parameter a fast LED flash pattern will make it possible to visual locate the node quick.This can be used to locate the node in your hardware setup if you have more than one DroneCAN node in your setup.
Related to my old PR which I have closed now: https://github.com/ArduPilot/ardupilot/pull/23503

### DIFF
--- a/Tools/AP_Periph/AP_Periph.cpp
+++ b/Tools/AP_Periph/AP_Periph.cpp
@@ -416,15 +416,29 @@ void AP_Periph_FW::update()
     node_stats.update();
 #endif
 
-    static uint32_t last_led_ms;
     uint32_t now = AP_HAL::millis();
-    if (now - last_led_ms > 1000) {
-        last_led_ms = now;
+
 #ifdef HAL_GPIO_PIN_LED
+    static uint32_t last_led_ms;
+    static uint32_t next_led_ms = 1000;
+    if (now - last_led_ms > next_led_ms) {
+        last_led_ms = now;
         if (!no_iface_finished_dna) {
+            static uint8_t led_phase = 0;
+            if (debug_option_is_set(DebugOptions::FIND_NODE)) {
+                next_led_ms = (led_phase < 10) ? 100 : 500;
+                led_phase = (led_phase + 1) % 12;
+            } else {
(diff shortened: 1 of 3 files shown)


Tools/AP_Periph/AP_Periph.cpp
```diff
@@ -416,15 +416,29 @@ void AP_Periph_FW::update()
     node_stats.update();
 #endif
 
-    static uint32_t last_led_ms;
     uint32_t now = AP_HAL::millis();
-    if (now - last_led_ms > 1000) {
-        last_led_ms = now;
+
 #ifdef HAL_GPIO_PIN_LED
+    static uint32_t last_led_ms;
```
peterbarker, line 422: Could we drop this into a `flash_led` method, please?

Tools/AP_Periph/AP_Periph.cpp
```diff
@@ -416,15 +416,29 @@ void AP_Periph_FW::update()
...
+    static uint32_t next_led_ms = 1000;
+    if (now - last_led_ms > next_led_ms) {
+        last_led_ms = now;
         if (!no_iface_finished_dna) {
```
peterbarker, line 426: Doesn't this mean the code will be non-functional, depending on whether it has dna'd or not?

Tools/AP_Periph/AP_Periph.cpp
```diff
@@ -416,15 +416,29 @@ void AP_Periph_FW::update()
...
 #endif
 
-    static uint32_t last_led_ms;
     uint32_t now = AP_HAL::millis();
-    if (now - last_led_ms > 1000) {
-        last_led_ms = now;
+
 #ifdef HAL_GPIO_PIN_LED
+    static uint32_t last_led_ms;
+    static uint32_t next_led_ms = 1000;
```
peterbarker, line 423:
```suggestion
    static uint32_t current_led_interval_ms = 1000;
```

PR #29846, opened 2025-04-21
All vehicles: Fix non fatal rc failsafe bug

Hwurzburg:
I discovered this while working on adding RC to Sub...easy to reproduce

Sim vehicle with defaults...happens disarmed or in the air
- SIM_RC_FAIL =0    you get an RC failsafe, prearm or action
- Disable the failsafe  ie FS_THR_ENABLE or THR_FAILSAFE = 0
- SIM_RC_FAIL =1    you enabled RC, but failsafe state is still not cleared...no GCS message...prearms if disarmed on ground
wont clear until reboot or resetting failsafe enable

non fatal but really annoying if you run into it...
hope its doesn't make CI uhappy

### DIFF
--- a/ArduPlane/radio.cpp
+++ b/ArduPlane/radio.cpp
@@ -285,8 +285,9 @@ void Plane::control_failsafe()
 
     const bool allow_failsafe_bypass = !arming.is_armed() && !is_flying() && (rc().enabled_protocols() != 0);
     const bool has_had_input = rc().has_had_rc_receiver() || rc().has_had_rc_override();
-    if ((g.throttle_fs_enabled != ThrFailsafe::Enabled) || (allow_failsafe_bypass && !has_had_input)) {
-        // If not flying and disarmed don't trigger failsafe until RC has been received for the fist time
+    if ((g.throttle_fs_enabled != ThrFailsafe::Enabled && !failsafe.rc_failsafe) || (allow_failsafe_bypass && !has_had_input)) {
+        // If throttle fs not enabled and not in failsafe, or 
+        // not flying and disarmed, don't trigger failsafe check until RC has been received for the fist time  
         return;
     }
 
--- a/Rover/radio.cpp
+++ b/Rover/radio.cpp
@@ -146,6 +146,7 @@ void Rover::radio_failsafe_check(uint16_t pwm)
 {
     if (!g.fs_throttle_enabled) {
         // radio failsafe disabled
+        AP_Notify::flags.failsafe_radio = false;
         return;
     }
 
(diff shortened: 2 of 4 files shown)


Rover/radio.cpp
```diff
@@ -146,6 +146,9 @@ void Rover::radio_failsafe_check(uint16_t pwm)
 {
     if (!g.fs_throttle_enabled) {
         // radio failsafe disabled
+        if (notify.flags.failsafe_radio) {
+        AP_Notify::flags.failsafe_radio = false;
```
rmackay9, line 150: thanks for this.  we can remove the "if" part of this and simply just always set the flag.

ArduPlane/radio.cpp
```diff
@@ -402,7 +402,7 @@ bool Plane::rc_throttle_value_ok(void) const
  */
 bool Plane::rc_failsafe_active(void) const
 {
-    if (!rc_throttle_value_ok()) {
+    if (!rc_throttle_value_ok() || (ThrFailsafe(g.throttle_fs_enabled.get()) == ThrFailsafe::Disabled)) {
```
rmackay9, line 405: this is a bit unintuitive for someone reading the code because I think it means that the RC failsafe is active when it's actually been disabled.

tridge: i think this is already covered in Plane::rc_throttle_value_ok() implementation?

tridge (changes requested):
the change in plane radio.cpp doesn't look right to me

ArduPlane/radio.cpp
Hwurzburg: it is not....rc_throttle_value_ok()  early returns, yes, but ALWAYS when throttle failsafe is disabled, even when failsafe flag is still set ....rc_failsafe_active will return true and never return false after RC sis restored

Hwurzburg: @rmackay9 and that is exactly the problem...this function returns true even if rc failsafe has been disabled otherwise...I guess it could be handled somehow in the rc_throttle_okay function...but I don't think it really belongs there either

Rover/radio.cpp
Hwurzburg: will do

ArduPlane/radio.cpp
```diff
@@ -402,6 +399,10 @@ bool Plane::rc_throttle_value_ok(void) const
  */
 bool Plane::rc_failsafe_active(void) const
 {
+    if ( ThrFailsafe(g.throttle_fs_enabled.get()) == ThrFailsafe::Disabled) {
```
IamPete1, line 402: I think this means that we will start using invalid RC inputs if the fail safe is disabled.

IamPete1 (commented):
I'm slightly reluctant to change anything here, its a critical part of the code.

ArduPlane/radio.cpp
Hwurzburg: I should move it to the bottom of this method....will do

rmackay9 (approved):
I'm happy with the copter and rover changes, thanks!

tridge (approved):

PR #29853, opened 2025-04-21
AP_WheelEncoder: add support for single pin pulse encoders

heeplr:
This PR enables use of single pin pulse encoders. Direction is determined by the sign of thrust.
(Works for me™ but barely tested)

### DIFF
--- a/libraries/AP_WheelEncoder/WheelEncoder_Pulse.cpp
+++ b/libraries/AP_WheelEncoder/WheelEncoder_Pulse.cpp
@@ -1,0 +1,160 @@
+/*
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <AP_AHRS/AP_AHRS.h>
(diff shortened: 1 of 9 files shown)


libraries/AP_WheelEncoder/WheelEncoder_Pulse.cpp
```diff
@@ -0,0 +1,160 @@
...
+    // Vector3f forward_dir = rot.colx();  // body X-axis in NED
+    // forward_dir.z = 0;                  // only horizontal
+    // forward_dir.normalize();           // ensure unit vector
+
+    // // project velocity onto forward direction to get signed ground speed
+    // float signed_speed = velocity.dot(forward_dir) * 100.0f; // cm/s
+
+    // get current throttle
+    float steering = 0, throttle = 0;
+    if(!AP::vehicle()->get_steering_and_throttle(steering, throttle)) {
```
rmackay9, line 129: Txs for the PR.  I don't think we should have a driver reaching back up into the vehicle code.  We'd call this a "layering violation". Instead it would be better to pass down an expected direction into the wheel encoder library.

This makes me wonder what we would do for wheel that are not facing forward for example on a 3 wheeled vehicle or a vehicle that moves laterally.  In any case, this is likely an existing issue so you don't need to worry about it.

heeplr: > Txs for the PR. I don't think we should have a driver reaching back up into the vehicle code. We'd call this a "layering violation". Instead it would be better to pass down an expected direction into the wheel encoder library.
>

You mean code that sets some throttle property in `WheelEncoder_Backend` on every throttle change?
Or creating a new layer that accesses `AP_Vehicle` and provides a getter?

> This makes me wonder what we would do for wheel that are not facing forward for example on a 3 wheeled vehicle or a vehicle that moves laterally. In any case, this is likely an existing issue so you don't need to worry about it.

I let the code that uses the IMU in there commented out. I guess one could calculate the exact movement direction and combine it with a registered heading of the encoded wheel. But we probably have the angle that the encoded wheel is heading configured by parameters.

PR #29855, opened 2025-04-22
Copter: tidy radio throttle failsafe code

peterbarker:


rmackay9:
Thanks for this.  This is fine, an alternative change would be 

`if (failsafe.radio || !rc().has_ever_seen_rc_input() || !motors->armed()) {`

peterbarker:
> Thanks for this. This is fine, an alternative change would be
> 
> `if (failsafe.radio || !rc().has_ever_seen_rc_input() || !motors->armed()) {`

It actually wasn't a fine change, we were both misled by the comment.

The existing RHS of the expression is `!(A||B)` and that is equivalent to `(!A && !B)`, so my restructure was incorrect.

Even the comment I added was actually a hint that this was entirely wrong!  The comment was `  // do not failsafe if motors not armed`.  I think we can both see that is completely ludicrous - Copter is *supposed* to failsafe when motors are disarmed, just take no actions when doing so.

The existing-in-master logic is bogus because the first clause is almost always true - we almost always `have_ever_seen_rc_input`, so we never look at the motor state at all.

The old comment in there says, "pass the throttle through - but that's absolutely not what we should be doing at all!

PR #29857, opened 2025-04-22
AP_BattMonitor: handle_scripting nullptr check

robertlong13:
Don't want a bug in a lua script to cause a hard fault.


peterbarker (approved):
LGTM

PR #29863, opened 2025-04-22
Sub: turn lights and video_switch into aux functions

Williangalvani:
This decouples some of the "hard links" we have between input rc channels and functionality


peterbarker (approved):
Nice!

Hwurzburg:
can you remove the defaults Servo 13, 14, and 9 now from RCIN to these new functions? why were there two servos for lights?

Hwurzburg (approved):

PR #29868, opened 2025-04-23
Create hwdef-bl.dat

Aedrox-UAx:
Aedrox H7 FC hwdef-bl file added


andyp1per:
Replaced by https://github.com/ArduPilot/ardupilot/pull/29915

PR #29870, opened 2025-04-23
WP_Nav update axis, units and comments

lthall:
This is the continued work to update axis and units.

```
SCB: HACK: Removing (build/CubeOrange/modules/DroneCAN/libcanard/dsdlc_generated)
SCB: Running (./waf configure --board CubeOrange --consistent-builds --bootloader) in (.)
SCB: Running (./waf bootloader) in (.)
SCB: Running (rsync -ap build/ /tmp/tmpl7cvn4cm/out-branch-CubeOrange) in (.)
CubeOrange,*,*,*,*,*,*,*
```

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -474,13 +474,13 @@ AP_Vehicle::custom_mode_state* Copter::register_custom_mode(const uint8_t num, c
 // circle mode controls
 bool Copter::get_circle_radius(float &radius_m)
 {
-    radius_m = circle_nav->get_radius() * 0.01f;
+    radius_m = circle_nav->get_radius_cm() * 0.01f;
     return true;
 }
 
-bool Copter::set_circle_rate(float rate_dps)
+bool Copter::set_circle_rate(float rate_degs)
 {
-    circle_nav->set_rate(rate_dps);
+    circle_nav->set_rate_degs(rate_degs);
     return true;
 }
 #endif
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -39,7 +39,7 @@ bool ModeAuto::init(bool ignore_checks)
         }
 
         // initialise waypoint and spline controller
-        wp_nav->wp_and_spline_init();
+        wp_nav->wp_and_spline_init_cm();
 
         // initialise desired speed overrides
         desired_speed_override = {0, 0, 0};
@@ -341,10 +341,10 @@ bool ModeAuto::loiter_start()
 
     // calculate stopping point
     Vector3f stopping_point;
-    wp_nav->get_wp_stopping_point(stopping_point);
+    wp_nav->get_wp_stopping_point_NEU_cm(stopping_point);
 
     // initialise waypoint controller target to stopping point
-    wp_nav->set_wp_destination(stopping_point);
+    wp_nav->set_wp_destination_NEU_cm(stopping_point);
 
     // hold yaw at current heading
     auto_yaw.set_mode(AutoYaw::Mode::HOLD);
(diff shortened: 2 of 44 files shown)


ArduCopter/mode_auto.cpp
```diff
@@ -857,23 +857,23 @@ uint32_t ModeAuto::wp_distance() const
...
 {
     switch (_mode) {
     case SubMode::CIRCLE:
-        return copter.circle_nav->get_bearing_to_target();
+        return copter.circle_nav->get_bearing_to_target_cd();
     case SubMode::WP:
     case SubMode::CIRCLE_MOVE_TO_EDGE:
     default:
-        return wp_nav->get_wp_bearing_to_destination();
+        return wp_nav->get_wp_bearing_to_destination_cdeg();
```
peterbarker, line 876: We see to have both `cd` and `cdeg` - should we choose one?

ArduCopter/mode_auto.cpp
```diff
@@ -614,17 +614,17 @@ void PayloadPlace::start_descent()
...
 
     // initialise the vertical position controller
     if (!pos_control->is_active_NE()) {
         pos_control->init_NE_controller();
     }
 
     // set vertical speed and acceleration limits
-    pos_control->set_max_speed_accel_U_cm(wp_nav->get_default_speed_down(), wp_nav->get_default_speed_up(), wp_nav->get_accel_z());
-    pos_control->set_correction_speed_accel_U_cmss(wp_nav->get_default_speed_down(), wp_nav->get_default_speed_up(), wp_nav->get_accel_z());
+    pos_control->set_max_speed_accel_U_cm(wp_nav->get_default_speed_down_cms(), wp_nav->get_default_speed_up_cms(), wp_nav->get_accel_U_cmss());
```
IamPete1, line 626: Using both `up` and `U` is a little confusing.

IamPete1 (commented):
I guess the goal would be to get back to names without units once we get back to SI? This global renaming allows us to make that conversion bit by bit one function at a time.

ArduCopter/mode_auto.cpp
lthall: I don't think it is too bad here. Up and Down are explicitly the maximum speeds in those directions where "U" is the axis.

As you say, in the longer term I would like to have a set of defaults at the top of each libary file that says the defaults are NED and m, radians. All variables using alternatives will be labled _deg for example.

I am leaning towards radians being the default internal unit for angle and degrees being the default parameter unit.

peterbarker:
> I guess the goal would be to get back to names without units once we get back to SI? This global renaming allows us to make that conversion bit by bit one function at a time.

The concept is `void my_function(x_cm);` -> `void my_function_cm(x_cm)` -> `void my_function_cm(x_cm); void my_function_m(x);` -> `void my_function_m(x);`.  Several years down the track we might be able to remove `_m` from the name, but can't until screwing over rebased out-of-tree code is going to be an anomaly rather than a certainty...

ArduCopter/Copter.cpp
```diff
@@ -474,13 +474,13 @@ AP_Vehicle::custom_mode_state* Copter::register_custom_mode(const uint8_t num, c
 // circle mode controls
 bool Copter::get_circle_radius(float &radius_m)
 {
-    radius_m = circle_nav->get_radius() * 0.01f;
+    radius_m = circle_nav->get_radius_cm() * 0.01f;
     return true;
 }
 
 bool Copter::set_circle_rate(float rate_dps)
```
rmackay9, line 481: maybe we should rename Copter::set_circle_rate to Copter::set_circle_rate_degs

rmackay9: this "rate_dps" should be "rate_degs"

peterbarker (changes requested):

ArduCopter/mode_auto.cpp
rmackay9: we decided that it should be "cd"

ArduCopter/mode_auto.cpp
lthall: My impression is that People were happy with this on the dev call.

ArduCopter/mode_guided.cpp
```diff
@@ -1132,7 +1132,7 @@ int32_t ModeGuided::wp_bearing() const
 {
     switch(guided_mode) {
     case SubMode::WP:
-        return wp_nav->get_wp_bearing_to_destination();
+        return wp_nav->get_wp_bearing_to_destination_cdeg();
```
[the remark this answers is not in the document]
lthall: Fixed

lthall:
I think I have addressed all comments:

SCB: Running (rsync -ap build/ /tmp/tmp4x269y57/out-branch-CubeOrange) in (.)
CubeOrange,*,*,*,*,*,*,*

peterbarker:
Merged, thanks!

PR #16091, opened 2020-12-17
Move rudder arming up into RC_Channel library

peterbarker:
This will make rudder-arming consistent across our vehicles.

One known behaviour change is that the rover arm interval has moved from 2 to 3 seconds (to be like Copter and Plane).

Other known changes in behaviour:
 - Copter and Rover will try to arm if your throttle is non-zero, and you will get a warning that your throttle is non-zero.  In master we don't try to arm.  I think this new behaviour is far preferable from a user's standpoint as they're *told* why they can't arm right now in the case they meant to arm!  The new code has the same behaviour as Plane in master

~~We can't move the rudder-specific arming checks into the RC_Channels library as ToyMode also arms with a reason of "rudder" and presumably wants the same checks~~  We did some surgery to fix this

PR #29871, opened 2025-04-23
Copter: Enable the disarm operation until the disarm is completed

muramura:
The processing is different when the operation is continued for both arm and disarm operations. 
The disarm operation will remain active until the disarm is completed.

### DIFF
--- a/ArduCopter/motors.cpp
+++ b/ArduCopter/motors.cpp
@@ -58,7 +58,7 @@ void Copter::arm_motors_check()
         }
 
         // increase the counter to a maximum of 1 beyond the disarm delay
-        if (arming_counter <= DISARM_DELAY) {
+        if (arming_counter < DISARM_DELAY) {
             arming_counter++;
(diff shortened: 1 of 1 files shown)


ArduCopter/motors.cpp
```diff
@@ -58,7 +58,7 @@ void Copter::arm_motors_check()
         }
 
         // increase the counter to a maximum of 1 beyond the disarm delay
```
peterbarker, line 60: I'd rather not merge this as it conflicts with my efforts to getting this up into the parent class (https://github.com/ArduPilot/ardupilot/pull/16091)

muramura: Does it include my changes?
If not, please include them first.
After that, I think it's fine to resolve the conflict with the previous PR.

peterbarker:
This code no longer exists; please patch the AP_Arming code if you want to pursue this change.

PR #29873, opened 2025-04-23
AP_Proximity: PROXIMITY is optional, so the number of devices can be specified

muramura:
PROXIMITY devices are optional. 
Recently, the number of devices increased from 4 to 5. 
Therefore, the number of devices will be configurable.

AFTER
Screenshot from 2025-04-24 02-19-03

BEFORE
Screenshot from 2025-04-24 02-08-05

### DIFF
--- a/libraries/AP_Proximity/AP_Proximity.h
+++ b/libraries/AP_Proximity/AP_Proximity.h
@@ -28,7 +28,9 @@
 
 #include <AP_HAL/Semaphores.h>
 
-#define PROXIMITY_MAX_INSTANCES             5   // Maximum number of proximity sensor instances available on this platform
+#ifndef AP_PROXIMITY_MAX_INSTANCES 
+  #define AP_PROXIMITY_MAX_INSTANCES             5   // Maximum number of proximity sensor instances available on this platform
+#endif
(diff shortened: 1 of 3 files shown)


libraries/AP_Proximity/AP_Proximity.h
```diff
@@ -28,7 +28,9 @@
 
 #include <AP_HAL/Semaphores.h>
 
-#define PROXIMITY_MAX_INSTANCES             5   // Maximum number of proximity sensor instances available on this platform
+#ifndef PROXIMITY_MAX_INSTANCES
```
peterbarker, line 31: On second thoughts, I think we really should rename this before making it something someone can reasonably put in their hwdefs:

```suggestion
#ifndef AP_PROXIMITY_MAX_INSTANCES 
```

What do you think, @muramura ?

rishabsingh3003 (approved):

libraries/AP_Proximity/AP_Proximity.h
muramura: I added the prefix "AP_" at the beginning.

peterbarker (approved):
Great, thanks!

PR #29874, opened 2025-04-23
DroneCAN: determine based on the defined value

muramura:
The defined value is inverted. 
By inverting the variable value, it can be determined based on the defined value.


peterbarker (approved):
LGTM

tpwrules:
No flash change at least on Cube Orange.

Issue #29710, opened 2025-04-06
Feature request: Better handling of default behaviour when installing dependencies for Ubuntu/Debian-variants


tejashah88:
## Feature request

**Is your feature request related to a problem? Please describe.**

I was trying to install ArduPilot on Linux Mint "xia", which is the equivalent of Ubuntu "noble", but in the `Tools/environment_install/install-prereqs-ubuntu.sh` file, the translation for that codename was missing and the script assumed the release codename to be "bullseye". This led into a rabbit hole of trying to figure out why the script failed to `apt install python-argparse`, and thus the submission of #29704. 

My main issue is that there was no warning regarding the default selection of "bullseye", and it was not obvious as to why the script was trying to install a package that doesn't existing in the apt repositories. For context, Debian "bullseye" was first released in 2021 and while it's still maintained, there are clearly some incompatibilities with any modern Ubuntu-based Linux flavors, as seen in the script.

**Describe the solution you'd like**
I would prefer a mechanism in the script to warn the user of this default and/or the option to specify your own release codename if it  wasn't recognized from `lsb_release -c -s`. I wouldn't want to change the default behaviour in case people still rely on it but I personally prefer explict warnings over silent defaults in these situations.

**Describe alternatives you've considered**
N/A

**Platform**
[X] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

peterbarker:
@tejashah88 I've created a PR which forces the mapping.  Your review and/or testing would be appreciated as I'm not set up for testing Mint here.

PR #23260, opened 2023-03-20
Tools: add jolnir on install script

nkruzan:
Allow to install dev environment on distros that have 'jolnir' as release name

PR #29875, opened 2025-04-24
Tools: environment_install: make sure Linux Mint dist is always mapped

peterbarker:
... to an Uubntu release.

Saves confusion on newer Mint releases as thje install will fail with a clear message rather than get confused on packages

Closes https://github.com/ardupilot/ardupilot/issues/29710

### DIFF
--- a/Tools/environment_install/install-prereqs-ubuntu.sh
+++ b/Tools/environment_install/install-prereqs-ubuntu.sh
@@ -60,22 +60,37 @@ fi
 # Checking Ubuntu release to adapt software version to install
 RELEASE_CODENAME=$(lsb_release -c -s)
 
-# translate Mint-codenames to Ubuntu-codenames based on https://www.linuxmint.com/download_all.php
-case ${RELEASE_CODENAME} in
-    wilma | xia)
-        RELEASE_CODENAME='noble'
+RELEASE_DISTRIBUTOR=$(lsb_release -i -s)
+case ${RELEASE_DISTRIBUTOR} in
+    elementary)
+        case ${RELEASE_CODENAME} in
+            jolnir)
+                RELEASE_CODENAME='focal'
+                ;;
+        esac
         ;;
-    vanessa | vera | victoria | virginia)
-        RELEASE_CODENAME='jammy'
-        ;;
-    una | uma | ulyssa | ulyana | jolnir)
-        RELEASE_CODENAME='focal'
-        ;;
-    tricia | tina | tessa | tara)
-        RELEASE_CODENAME='bionic'
-        ;;
-    elsie)
-        RELEASE_CODENAME='bullseye'
+    LinuxMint)
+        # translate Mint-codenames to Ubuntu-codenames based on https://www.linuxmint.com/download_all.php
+        case ${RELEASE_CODENAME} in
+            wilma | xia)
+                RELEASE_CODENAME='noble'
+                ;;
+            vanessa | vera | victoria | virginia)
+                RELEASE_CODENAME='jammy'
+                ;;
+            una | uma | ulyssa | ulyana)
+                RELEASE_CODENAME='focal'
+                ;;
+            tricia | tina | tessa | tara)
+                RELEASE_CODENAME='bionic'
+                ;;
+            elsie)
+                RELEASE_CODENAME='bullseye'
+                ;;
+            *)
+                echo "Unable to map ${RELEASE_CODENAME} to an Ubuntu release.  Please patch this script and submit a pull request, or report at https://github.com/ArduPilot/ardupilot/issues"
+                exit 1
+        esac
         ;;
 esac
 


Tools/environment_install/install-prereqs-ubuntu.sh
```diff
@@ -60,22 +60,30 @@ fi
...
+                RELEASE_CODENAME='focal'
+                ;;
+            tricia | tina | tessa | tara)
+                RELEASE_CODENAME='bionic'
+                ;;
+            elsie)
+                RELEASE_CODENAME='bullseye'
+                ;;
+            *)
+                echo "Unable to map ${RELEASE_CODENAME} to an Ubuntu release"
```
tpwrules, line 84:
```suggestion
                echo "Unable to map '${RELEASE_CODENAME}' to an Ubuntu release. Please report this as a bug at https://github.com/ArduPilot/ardupilot/issues".
```

Suggest giving a confused user a path forward.

peterbarker: ITYM, "please create a PR to correct this problem <here>"? :-)

tejashah88:
@peterbarker I'll try testing this later this week, but 1 things I noticed from earlier. "jolnir" is not part of the Linux Mint ecosystem, but rather it seems to be added from #23260, which relates to a release from Elementary OS 6.1. You would need to add another conditional check for when `lsb_release -i -s` equals the equivalent distributor ID from Elementary OS for "jolnir" to work properly. I can see this becoming a maintenance "thorn in the side" if this list grows to other distributions outside Elementary OS.

tejashah88:
Another concern is that this change only stops _unknown Linux Mint OS distributions_, nothing outside that ecosystem. In other words, if someone installed ardupilot with Pop! OS or an unknown Elementary OS distribution and that user's release codename is not recognizable from the install script, it would be the same problem over again with "bullseye". I can see two ways to deal with this.

## Option A: Set a hard filter against unknown OS distributions
This would be as simple as checking `RELEASE_DISTRIBUTOR` if it's from `Ubuntu` or `LinuxMint` exclusively. While this could be a "breaking change", this would guarantee compatibility against what has been well-tested. Users can raise issues for their specific OS as needed (which can be easy to add back support).

## Option B: Implement checks for popular Ubuntu-variants and all their releases
This would involve having to check the `RELEASE_CODENAME` (`lsb_release -c -s`) and `RELEASE_DISTRIBUTOR` (`lsb_release -i -s`) for say the top 5 Ubuntu-based OS flavors and checking the release mapping to the Ubuntu counterpart, which seems over-engineered for a potentially non-existent userbase.

### My Recommendation
I'd recommend Option A since it's simple to implement and adding support for specific releases can involve simply asking the user the output of `lsb_release -c -s` and `lsb_release -i -s` and checking against the ubuntu equivalent codenames.

Tools/environment_install/install-prereqs-ubuntu.sh
peterbarker: I've added words.

peterbarker:
> Another concern is that this change only stops _unknown Linux Mint OS distributions_, nothing outside that ecosystem. In other words, if someone installed ardupilot with Pop! OS or an unknown Elementary OS distribution and that user's release codename is not recognizable from the install script, it would be the same problem over again with "bullseye". I can see two ways to deal with this.
> ## Option A: Set a hard filter against unknown OS distributions
> 
> This would be as simple as checking `RELEASE_DISTRIBUTOR` if it's from `Ubuntu` or `LinuxMint` exclusively. While this could be a "breaking change", this would guarantee compatibility against what has been well-tested. Users can raise issues for their specific OS as needed (which can be easy to add back support).
> ## Option B: Implement checks for popular Ubuntu-variants and all their releases
> 
> This would involve having to check the `RELEASE_CODENAME` (`lsb_release -c -s`) and `RELEASE_DISTRIBUTOR` (`lsb_release -i -s`) for say the top 5 Ubuntu-based OS flavors and checking the release mapping to the Ubuntu counterpart, which seems over-engineered for a potentially non-existent userbase.
> ### My Recommendation
> 
> I'd recommend Option A since it's simple to implement and adding support for specific releases can involve simply asking the user the output of `lsb_release -c -s` and `lsb_release -i -s` and checking against the ubuntu equivalent codenames.

I don't have a huge amount of interest in adding support for the less-popular Linux distributions - the lines you're thinking of with "over-engineering".

Someone with a bit more time and inclination could come in and do enough work that we could become strict on the distribution accepted, but I thought a little bit of work for Mint (given its popularity) was worthwhile, thus this PR.

peterbarker:
> @peterbarker I'll try testing this later this week, but 1 things I noticed from earlier. "jolnir" is not part of the Linux Mint ecosystem, but rather it seems to be added from #23260, which relates to a release from Elementary OS 6.1. You would need to add another 

Done

tejashah88:
That's acceptable with me. I also tested this PR on my machine and it does install the dependencies as expected. For this PR, I think the problem has been adressed well enough. LGTM 👍

peterbarker:
Thanks @tejashah88 , merged it!

PR #29877, opened 2025-04-24
AP_TECS: stop setting badDescent during transition

robertlong13:
`badDescent` prevents an unreasonable airspeed command from causing a dive. This is not needed during transition. If an unreasonable airspeed command is set, you'll eventually fail to transition, not nose-dive.

Additionally, 3 of these conditions are guaranteed to be true through transition, and toward the latter half of transition, when KEdot is small, any noise on KEdot can cause a zero-crossing. Looking back at old logs, I seem to get a `badDescent` flag, and subsequent throttle-back when the commanded airspeed drops to min, somewhere on every outbound transition (on 4.5.7 firmware, in case that's relevant). I don't want my pusher throttling back on outbound transitions.

It seems to me that `badDescent` should get the same treatment as `underspeed`, which is disabled in VTOL.

### DIFF
--- a/libraries/AP_TECS/AP_TECS.cpp
+++ b/libraries/AP_TECS/AP_TECS.cpp
@@ -916,20 +916,23 @@ void AP_TECS::_update_throttle_without_airspeed(int16_t throttle_nudge, float pi
 
 void AP_TECS::_detect_bad_descent(void)
 {
+    // Don't detect bad descents when gliding, transitioning, or when underspeed.
+    if (_flags.is_gliding || _flight_stage == AP_FixedWing::FlightStage::VTOL || _flags.underspeed) {
+        _flags.badDescent = false;
+        return;
+    }
+
     // Detect a demanded airspeed too high for the aircraft to achieve. This will be
     // evident by the following conditions:
-    // 1) Underspeed protection not active
-    // 2) Specific total energy error > 200 (greater than ~20m height error)
-    // 3) Specific total energy reducing
-    // 4) throttle demand > 90%
-    // If these four conditions exist simultaneously, then the protection
-    // mode will be activated.
-    // Once active, the following condition are required to stay in the mode
-    // 1) Underspeed protection not active
-    // 2) Specific total energy error > 0
-    // This mode will produce an undulating speed and height response as it cuts in and out but will prevent the aircraft from descending into the ground if an unachievable speed demand is set
+    // 1) Specific total energy error > 200 (greater than ~20m height error)
+    // 2) Specific total energy reducing
+    // 3) throttle demand > 90%
+    // If these conditions exist simultaneously, then the protection mode will be activated.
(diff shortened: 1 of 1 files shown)


libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -929,7 +935,7 @@ void AP_TECS::_detect_bad_descent(void)
     // 2) Specific total energy error > 0
     // This mode will produce an undulating speed and height response as it cuts in and out but will prevent the aircraft from descending into the ground if an unachievable speed demand is set
     float STEdot = _SPEdot + _SKEdot;
-    if (((!_flags.underspeed && (_STE_error > 200.0f) && (STEdot < 0.0f) && (_throttle_dem >= _THRmaxf * 0.9f)) || (_flags.badDescent && !_flags.underspeed && (_STE_error > 0.0f))) && !_flags.is_gliding) {
+    if ((!_flags.underspeed && (_STE_error > 200.0f) && (STEdot < 0.0f) && (_throttle_dem >= _THRmaxf * 0.9f)) || (_flags.badDescent && !_flags.underspeed && (_STE_error > 0.0f))) {
```
priseborough, line 938: The  _flags.underspeed true will falsify both  inputs to the OR statement so I would move it to line 920

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -916,6 +916,12 @@ void AP_TECS::_update_throttle_without_airspeed(int16_t throttle_nudge, float pi
 
 void AP_TECS::_detect_bad_descent(void)
 {
+    // Don't detect bad descents when gliding or transitioning
```
priseborough, line 919: I would move the _flags.underspeed check to this line, eg

```
// Don't detect bad descents when gliding or transitioning or when recovering from underspeed
if (_flags.is_gliding || _flight_stage == AP_FixedWing::FlightStage::VTOL || _flags.underspeed) {
```

libraries/AP_TECS/AP_TECS.cpp
robertlong13: Great call. That makes this monster if-statement way more readable.

libraries/AP_TECS/AP_TECS.cpp
robertlong13: Done

tridge (approved):
looks good, thanks!

priseborough (approved):

robertlong13:
Tested in SITL in RealFlight (injected a little noise on KED to reproduce). badDescent still works to prevent descent during unachievable airspeed and does not trip in transition. Tomorrow, I'll test on a real aircraft, but I'll only be testing the latter part (not tripping during transition).

robertlong13:
Flight tested on Ottano, much cleaner transitions now.

rmackay9:

PR #29882, opened 2025-04-24
Tools: fix Starling2Max param file

rmackay9:
This fixes a few issues with the parameter file for the [ModalAI Starling2Max quadcopter](https://www.modalai.com/en-jp/products/starling-2-max?variant=48172375310640)

I've tested this on a real vehicle and they seem OK to me


peterbarker:
@katzfey

katzfey:
Looks good to me

PR #29884, opened 2025-04-24
Fix potential vulnerable cloned functions: Bug: Lua can generate wrong code when _ENV is <const>

npt-1707:
Description
This PR fixes a potential vulnerability in singlevar() that was cloned from lua but did not receive the security patch. The original issue was reported and fixed under https://github.com/lua/lua/commit/1f3c6f4534c6411313361697d98d1145a1f030fa.
This PR applies the same patch to eliminate the vulnerability.

References
https://nvd.nist.gov/vuln/detail/cve-2022-28805
https://github.com/lua/lua/commit/1f3c6f4534c6411313361697d98d1145a1f030fa


tpwrules:
We have already rejected this patch in https://github.com/ArduPilot/ardupilot/pull/29366 . Please stop spamming vulnerability reports.

PR #29742, opened 2025-04-10
Add rc 2 sub

Hwurzburg:
SITL tested:
RC received and controls servos if assigned to RC
AUX switches work
Mode Switch works (default ch is 9 so as not to interfere with existing channel assignments)
RC failsafe activates and deactivates on pulse loss or throttle position
RC failsafe actions occur

Will test on CubeOrange Monday

Need Willian to test

PR #29885, opened 2025-04-24
RC_Channel:add Sub Aux Sw metadata

Hwurzburg:


peterbarker:
Shouldn't this be a patch in #29742 ?

Hwurzburg:
Thought it was going to be merged by now...I can combine, if Willian wants, since its still not merged

Williangalvani (approved):

PR #29887, opened 2025-04-25
AP_Follow axis and unit naming

lthall:
Work from before my main AP_Follow updates

```
SCB: HACK: Removing (build/CubeOrange/modules/DroneCAN/libcanard/dsdlc_generated)
SCB: Running (./waf configure --board CubeOrange --consistent-builds --bootloader) in (.)
SCB: Running (./waf bootloader) in (.)
SCB: Running (rsync -ap build/ /tmp/tmpn3s6xjeg/out-branch-CubeOrange) in (.)
CubeOrange,*,*,*,*,*,*,*
```

### DIFF
--- a/libraries/AP_Follow/AP_Follow.cpp
+++ b/libraries/AP_Follow/AP_Follow.cpp
@@ -74,7 +74,7 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
     // @Units: m
     // @Range: 1 1000
     // @User: Standard
-    AP_GROUPINFO("_DIST_MAX", 5, AP_Follow, _dist_max, 100),
+    AP_GROUPINFO("_DIST_MAX", 5, AP_Follow, _dist_max_m, 100),
 
     // @Param: _OFS_TYPE
     // @DisplayName: Follow offset type
@@ -106,7 +106,7 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
     // @Units: m
     // @Increment: 1
     // @User: Standard
-    AP_GROUPINFO("_OFS", 7, AP_Follow, _offset, 0),
+    AP_GROUPINFO("_OFS", 7, AP_Follow, _offset_ned_m, 0),
 
 #if !(APM_BUILD_TYPE(APM_BUILD_Rover))
     // @Param: _YAW_BEHAVE
@@ -160,23 +160,23 @@ AP_Follow::AP_Follow() :
 void AP_Follow::clear_offsets_if_required()
 {
     if (_offsets_were_zero) {
-        _offset.set(Vector3f());
+        _offset_ned_m.set(Vector3f());
         _offsets_were_zero = false;
     }
 }
 
 // get target's estimated location
-bool AP_Follow::get_target_location_and_velocity(Location &loc, Vector3f &vel_ned) const
+bool AP_Follow::get_target_location_and_velocity(Location &loc, Vector3f &vel_ned_ms) const
 {
-    Vector3p pos_ned;
-    Vector3f local_vel_ned;  // so we either change both return parameters or neither
-    if (!get_target_position_and_velocity(pos_ned, local_vel_ned)) {
+    Vector3p pos_ned_m;
+    Vector3f local_vel_ned_m;  // so we either change both return parameters or neither
+    if (!get_target_position_and_velocity_NED_m(pos_ned_m, local_vel_ned_m)) {
         return false;
     }
-    if (!AP::ahrs().get_location_from_origin_offset_NED(loc, pos_ned)) {
+    if (!AP::ahrs().get_location_from_origin_offset_NED(loc, pos_ned_m)) {
         return false;
     }
-    vel_ned = local_vel_ned;
+    vel_ned_ms = local_vel_ned_m;
 
     if (_alt_type == AP_FOLLOW_ALTITUDE_TYPE_RELATIVE) {
         loc.change_alt_frame(Location::AltFrame::ABOVE_HOME);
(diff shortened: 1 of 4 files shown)


peterbarker:
Past https://github.com/ArduPilot/ardupilot/pull/29907 this is binary-output-identical on our Linux and stm32 targets.

peterbarker (approved):
LGTM

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -74,39 +74,39 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
     // @Range: 1 1000
     // @User: Standard
-    AP_GROUPINFO("_DIST_MAX", 5, AP_Follow, _dist_max, 100),
+    AP_GROUPINFO("_DIST_MAX", 5, AP_Follow, _dist_max_m, 100),
 
     // @Param: _OFS_TYPE
     // @DisplayName: Follow offset type
     // @Description: Follow offset type
-    // @Values: 0:North-East-Down, 1:Relative to lead vehicle heading
+    // @Values: 0:North-East-Down, 1:Relative to lead vehicle heading_deg
```
IamPete1, line 82: I don't think we need this?

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -74,39 +74,39 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
     // @User: Standard
     AP_GROUPINFO("_OFS_TYPE", 6, AP_Follow, _offset_type, AP_FOLLOW_OFFSET_TYPE_NED),
 
     // @Param: _OFS_X
-    // @DisplayName: Follow offsets in meters north/forward
-    // @Description: Follow offsets in meters north/forward.  If positive, this vehicle fly ahead or north of lead vehicle.  Depends on FOLL_OFS_TYPE
+    // @DisplayName: Follow offsets_ned_m in meters north/forward
+    // @Description: Follow offsets_ned_m in meters north/forward.  If positive, this vehicle fly ahead or north of lead vehicle.  Depends on FOLL_OFS_TYPE
```
IamPete1, line 88: This makes the user facing docs harder to read IMHO.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -74,39 +74,39 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
     // @Range: -100 100
     // @Units: m
     // @Increment: 1
     // @User: Standard
 
     // @Param: _OFS_Y
-    // @DisplayName: Follow offsets in meters east/right
-    // @Description: Follow offsets in meters east/right.  If positive, this vehicle will fly to the right or east of lead vehicle.  Depends on FOLL_OFS_TYPE
+    // @DisplayName: Follow offsets_ned_m in meters east/right
+    // @Description: Follow offsets_ned_m in meters east/right.  If positive, this vehicle will fly to the right or east of lead vehicle.  Depends on FOLL_OFS_TYPE
```
IamPete1, line 96: This makes the user facing docs harder to read IMHO.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -74,39 +74,39 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
     // @Range: -100 100
     // @Units: m
     // @Increment: 1
     // @User: Standard
 
     // @Param: _OFS_Z
-    // @DisplayName: Follow offsets in meters down
-    // @Description: Follow offsets in meters down.  If positive, this vehicle will fly below the lead vehicle
+    // @DisplayName: Follow offsets_ned_m in meters down
+    // @Description: Follow offsets_ned_m in meters down.  If positive, this vehicle will fly below the lead vehicle
```
IamPete1, line 104: This makes the user facing docs harder to read IMHO.

IamPete1 (commented):
These are mostly in SI units already, do we need to add the units to the names in that case?

libraries/AP_Follow/AP_Follow.cpp
lthall: I agree, this was accidental.

libraries/AP_Follow/AP_Follow.cpp
lthall: I agree, this was accidental.

libraries/AP_Follow/AP_Follow.cpp
lthall: I agree, this was accidental.

libraries/AP_Follow/AP_Follow.cpp
lthall: I agree, this was accidental.

lthall:
> These are mostly in SI units already, do we need to add the units to the names in that case?

I did consider that but we are still mixing units a lot here. I am rewriting this libary and am happy to take responsibility for taking this through the whole process to end up where we all want to go.

You are right but for my own piece of mind I would like to keep this consistent with the other libaries I am working through. I plan to do a family of PR's dealing with each area of the code. Once they are all at the same state I will look for guidance as to what the next step will be but I am thinking:
- Naming update - units and axis
- Internal algorithums - unit and axis conversion (my consern here is size will probably increase before it drops back to the current level)
- Vehicle units and axis - I can see how I can do copter but the other aircraft will need to use the old unit conversion functions I provide.
- Parameters - Lots of help requied

Currently I think I can do:
Copter
- AC_AttitudeControl / AC_PosControl
- AC_AutoTune
- AC_WPNav
- AP_Follow
- AP_Math (control portions - should be limited or no changes)
- AP_Motors (should be limited or no changes)

Some others may come into the scope as I work through things. AC_PrecLand, AP_Avoidance, related libaries like that.

tridge (approved):

lthall:
```
SCB: HACK: Removing (build/CubeOrange/modules/DroneCAN/libcanard/dsdlc_generated)
SCB: Running (./waf configure --board CubeOrange --consistent-builds --bootloader) in (.)
SCB: Running (./waf bootloader) in (.)
SCB: Running (rsync -ap build/ /tmp/tmp32bgtzwe/out-branch-CubeOrange) in (.)
CubeOrange,*,*,*,*,*,*,*
```

Still all good.

peterbarker:
Zero binary output changes

PR #29889, opened 2025-04-25
Fixup lua language server CI

IamPete1:
Our lua language server CI has been incorrectly passing thanks to some recent changes to the language server. This updates so it correctly fails (and fixes the issues that crept in)

I can't get the problem matcher to work again, so no nice annotations, `fileLocation` seems to not work. It works fine for the luacheck problem matcher, so I have no idea.

A alternate fix would be to add `--check_format json` to the call to the language server. But the new native error printing is nicer than what I had done in python by parsing the json. Just we loose the problem match.

### DIFF
--- a/libraries/AP_Scripting/applets/RockBlock.lua
+++ b/libraries/AP_Scripting/applets/RockBlock.lua
@@ -760,6 +760,7 @@ function HLSatcom()
     --- check if GCS telemetry has been lost for RCK_TIMEOUT sec (if param enabled)
     if RCK_FORCEHL:get() == 2 then
         -- link lost time = boot time - GCS last seen time
+        ---@diagnostic disable-next-line cast-local-type
         link_lost_for = millis():toint() - gcs:last_seen()
         -- gcs:last_seen() is set to millis() during boot (on plane). 0 on rover/copter
         -- So if it's less than 10000 assume no GCS packet received since boot
--- a/.github/problem-matchers/Lua.json
+++ b/.github/problem-matchers/Lua.json
@@ -15,16 +15,14 @@
         },
         {
             "owner": "Lua-language-server-problem-matcher",
+            "fileLocation": [ "relative", "${GITHUB_WORKSPACE}" ],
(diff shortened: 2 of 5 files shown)


peterbarker:
@stephendade FYI

libraries/AP_Scripting/applets/RockBlock.lua
```diff
@@ -760,7 +760,7 @@ function HLSatcom()
     --- check if GCS telemetry has been lost for RCK_TIMEOUT sec (if param enabled)
     if RCK_FORCEHL:get() == 2 then
         -- link lost time = boot time - GCS last seen time
-        link_lost_for = millis():toint() - gcs:last_seen()
+        link_lost_for = (millis()- gcs:last_seen()):toint()
```
IamPete1, line 763: The subtraction was casting back to a uint32 so the result was not a int.

stephendade:
> @stephendade FYI

I'll run some checks in the next day or two and confirm the changes are OK (they most likely are)

peterbarker:
Bumped off DevCallTopic so @stephendade can run his tests.  Please ping once you're done, @stephendade :-)

IamPete1:
> Bumped off DevCallTopic so @stephendade can run his tests. Please ping once you're done, @stephendade :-)

I could alternately add a error ignore for just that file, then Stephen can take his time.

IamPete1:
> I could alternately add a error ignore for just that file, then Stephen can take his time.

I did that.

peterbarker (approved):

peterbarker:
@stephendade could you test the original PR to make sure it works, or create a patch which grabs the changes from the original PR and removes the new ignore lines, please?

stephendade:
> @stephendade could you test the original PR to make sure it works, or create a patch which grabs the changes from the original PR and removes the new ignore lines, please?

Sure. I've now had a chance to test @IamPete1's original changes, and they worked fine, so I'll create a new PR to re-include them

PR #29895, opened 2025-04-25
SRV_Channel: Consolidate the processing

muramura:
The processing can be consolidated. 
By doing so, the available space in flash memory will increase.


peterbarker (commented):
This is a no-compiler-output change:
```
MatekF405-Wing,*,*,*,*,*,*,*
```
So it doesn't save flash.

This is also incorrect; type_setup is not a property of setting the high-output, but still belongs in the setting of `type_angle`.

Closing this one.

PR #29898, opened 2025-04-26
Follow kinematic shaping

lthall:
This pr provides a kinematic estimate of the target location and heading.

I have added time out's and resets based on the kinematic limits and errors.

We need to decide on how we choose between the different data message sources. Ideally I would like to change the name and extend the _ALT_TYPE parameter to be 0: GPI absolute, 1: GPI relative, 2: Follow Target (GPI = global position int).

We also need to decide what the options and default settings should be. We can leave the kinematic interpolation off by default, we may choose not to process acceleration to reduce noise.

### DIFF
--- a/ArduCopter/mode_follow.cpp
+++ b/ArduCopter/mode_follow.cpp
@@ -47,6 +47,10 @@ void ModeFollow::run()
         return;
     }
 
+    // set offsets to current relative position if not already set
+    // this is done here to prevent the vehicle coliding with the target vehicle
+    g2.follow.init_offsets_if_required();
+
     // set motors to full range
     motors->set_desired_spool_state(AP_Motors::DesiredSpoolState::THROTTLE_UNLIMITED);
 
--- a/libraries/AP_Follow/AP_Follow.cpp
+++ b/libraries/AP_Follow/AP_Follow.cpp
@@ -1,17 +1,27 @@
 /*
-   This program is free software: you can redistribute it and/or modify
-   it under the terms of the GNU General Public License as published by
-   the Free Software Foundation, either version 3 of the License, or
-   (at your option) any later version.
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
 
-   This program is distributed in the hope that it will be useful,
-   but WITHOUT ANY WARRANTY; without even the implied warranty of
-   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-   GNU General Public License for more details.
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
 
-   You should have received a copy of the GNU General Public License
-   along with this program.  If not, see <http://www.gnu.org/licenses/>.
- */
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+//==============================================================================
+// AP_Follow Library
+// Target Following Logic for ArduPilot Vehicles
+//==============================================================================
+
+
+//==============================================================================
+// Includes
+//==============================================================================
 
 #include "AP_Follow_config.h"
 
@@ -26,10 +36,16 @@
 #include <AP_Logger/AP_Logger.h>
 #include <GCS_MAVLink/GCS.h>
 #include <AP_Vehicle/AP_Vehicle_Type.h>
+#include <AP_Scheduler/AP_Scheduler.h>
 
 extern const AP_HAL::HAL& hal;
 
-#define AP_FOLLOW_TIMEOUT_MS    3000    // position estimate timeout after 1 second
+
+//==============================================================================
+// Constants and Definitions
+//==============================================================================
+
+#define AP_FOLLOW_TIMEOUT_MS    3000    // position estimate timeout
 #define AP_FOLLOW_SYSID_TIMEOUT_MS 10000 // forget sysid we are following if we have not heard from them in 10 seconds
 
 #define AP_FOLLOW_OFFSET_TYPE_NED       0   // offsets are in north-east-down frame
@@ -40,10 +56,17 @@ extern const AP_HAL::HAL& hal;
 #define AP_FOLLOW_POS_P_DEFAULT 0.1f    // position error gain default
 
 #if APM_BUILD_TYPE(APM_BUILD_ArduPlane)
-#define AP_FOLLOW_ALT_TYPE_DEFAULT 0
+ #define AP_FOLLOW_ALT_TYPE_DEFAULT 0
+ #define AP_FOLLOW_DIST_MAX_DEFAULT 0
 #else
-#define AP_FOLLOW_ALT_TYPE_DEFAULT AP_FOLLOW_ALTITUDE_TYPE_RELATIVE
+ #define AP_FOLLOW_ALT_TYPE_DEFAULT AP_FOLLOW_ALTITUDE_TYPE_RELATIVE
+ #define AP_FOLLOW_DIST_MAX_DEFAULT 100
 #endif
+
+
+//==============================================================================
+// Constructor
+//==============================================================================
 
 AP_Follow *AP_Follow::_singleton;
 
@@ -70,11 +93,11 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
 
     // @Param: _DIST_MAX
     // @DisplayName: Follow distance maximum
-    // @Description: Follow distance maximum.  targets further than this will be ignored
+    // @Description: Follow distance maximum. If exceeded, the follow estimate will be considered invalid.
     // @Units: m
     // @Range: 1 1000
     // @User: Standard
-    AP_GROUPINFO("_DIST_MAX", 5, AP_Follow, _dist_max_m, 100),
+    AP_GROUPINFO("_DIST_MAX", 5, AP_Follow, _dist_max_m, AP_FOLLOW_DIST_MAX_DEFAULT),
 
     // @Param: _OFS_TYPE
     // @DisplayName: Follow offset type
@@ -85,7 +108,7 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
 
     // @Param: _OFS_X
     // @DisplayName: Follow offsets in meters north/forward
-    // @Description: Follow offsets in meters north/forward.  If positive, this vehicle fly ahead or north of lead vehicle.  Depends on FOLL_OFS_TYPE
+    // @Description: Follow offsets in meters north/forward. If positive, this vehicle fly ahead or north of lead vehicle.  Depends on FOLL_OFS_TYPE
     // @Range: -100 100
     // @Units: m
     // @Increment: 1
@@ -93,7 +116,7 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
 
     // @Param: _OFS_Y
     // @DisplayName: Follow offsets in meters east/right
-    // @Description: Follow offsets in meters east/right.  If positive, this vehicle will fly to the right or east of lead vehicle.  Depends on FOLL_OFS_TYPE
+    // @Description: Follow offsets in meters east/right. If positive, this vehicle will fly to the right or east of lead vehicle.  Depends on FOLL_OFS_TYPE
     // @Range: -100 100
     // @Units: m
     // @Increment: 1
@@ -101,12 +124,12 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
 
     // @Param: _OFS_Z
     // @DisplayName: Follow offsets in meters down
-    // @Description: Follow offsets in meters down.  If positive, this vehicle will fly below the lead vehicle
+    // @Description: Follow offsets in meters down. If positive, this vehicle will fly below the lead vehicle
     // @Range: -100 100
     // @Units: m
     // @Increment: 1
     // @User: Standard
-    AP_GROUPINFO("_OFS", 7, AP_Follow, _offset_ned_m, 0),
+    AP_GROUPINFO("_OFS", 7, AP_Follow, _offset_m, 0),
 
 #if !(APM_BUILD_TYPE(APM_BUILD_Rover))
     // @Param: _YAW_BEHAVE
@@ -119,7 +142,7 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
 
     // @Param: _POS_P
     // @DisplayName: Follow position error P gain
-    // @Description: Follow position error P gain.  Converts the difference between desired vertical speed and actual speed into a desired acceleration that is passed to the throttle acceleration controller
+    // @Description: Follow position error P gain. Converts the difference between desired vertical speed and actual speed into a desired acceleration that is passed to the throttle acceleration controller
     // @Range: 0.01 1.00
     // @Increment: 0.01
     // @User: Standard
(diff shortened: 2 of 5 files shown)


timtuxworth (changes requested):
Some initial comments.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,638 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
     }
     }
 
     if (updated) {
+        // message provided updated position information
+        Vector3f offset_m = _offset_XXD_m.get();
+
+        // if using relative frame, rotate offsets based on estimated heading
+        if (_offset_type == AP_FOLLOW_OFFSET_TYPE_RELATIVE) {
+            offset_m.xy().rotate(_estimate_heading_rad);
```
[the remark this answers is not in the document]
lthall: Fixed

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,638 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+            // if unable to retrieve local position, clear distance/bearing info
+            clear_dist_and_bearing_to_target();
+        } else {
+            // convert vehicle position to NED meters (NEU -> NED and cm -> m)
+            current_position_ned_m.z = -current_position_ned_m.z; // NEU to NED
+            current_position_ned_m *= 0.01;  // convert cm to m
+
+            // calculate distance vectors to target, both with and without offsets
+            const Vector3p dist_vec_ned_m = _estimate_pos_ned_m - current_position_ned_m.topostype();
+            const Vector3p ofs_dist_vec = _ofs_estimate_pos_ned_m - current_position_ned_m.topostype();
```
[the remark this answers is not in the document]
lthall: fixed

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,638 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+    }
+    _target_pos_ned_m.z = -packet.alt + origin.alt * 0.01;
+
+    // decode velocity if available (bit 1 of est_capabilities)
+    if (packet.est_capabilities & (1<<1)) {
+        _target_vel_ned_ms.x = packet.vel[0]; // velocity north
+        _target_vel_ned_ms.y = packet.vel[1]; // velocity east
+        _target_vel_ned_ms.z = packet.vel[2]; // velocity down
+    } else {
+        _target_vel_ned_ms.zero();
```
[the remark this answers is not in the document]
lthall: We need to only use one of these messages. This is something we need to add.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,638 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+    }
+
+    // decode acceleration if available (bit 2 of est_capabilities)
+    if (packet.est_capabilities & (1 << 2)) {
+        _target_accel_ned_mss.x = packet.acc[0]; // acceleration north
+        _target_accel_ned_mss.y = packet.acc[1]; // acceleration east
+        _target_accel_ned_mss.z = packet.acc[2]; // acceleration down
+    } else {
+        _target_accel_ned_mss.zero();
```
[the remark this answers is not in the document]
lthall: We need to only use one of these messages. This is something we need to add.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -556,51 +796,93 @@ Vector3f AP_Follow::rotate_vector(const Vector3f &vec, float angle_deg) const
...
+}
+#endif  // HAL_LOGGING_ENABLED
+
+
+//==============================================================================
+// Accessors and Helpers
+//==============================================================================
 
-    vel_ned = local_vel_ned;
+// Computes the estimated velocity in the NED frame by applying acceleration over the given time delta.
```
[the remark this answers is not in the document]
lthall: Not used

libraries/AP_Follow/AP_Follow.cpp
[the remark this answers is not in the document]
timtuxworth: When I send FOLLOW_TARGET from a plane, it doesn't have FOLLOW_TARGET_CAPABILITIES.VEL and FOLLOW_TARGET_CAPABILITIES.ACCEL and this breaks the follow code, I've seen it in testing. These lines (710,719) need to go.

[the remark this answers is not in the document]
lthall: Sorry @timtuxworth but I must be missunderstanding what you are saying. We have to set these to zero if they are not provided by the lead aircraft.

This PR currently doesn't let you turn off the consumption of GLOBAL_POSITION_INT when you provide FOLLOW_TARGET so using both will cause problems if one provides more information than the other. This is something I need to take to the Dev Call to get some guidance on how to implement this. Maybe I should do it the way I think it should be done so at least this PR is fully functional.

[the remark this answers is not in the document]
timtuxworth: I don't know why using both will cause problems. It works for me if I comment out those two lines, and FOLLOW_TARGET provides _additional_ information - specifically attitude, which GLOBAL_POSITION_INT doesn't provide. 

BTW - for plane, I can get the same data with GLOBAL_POSITION_INT and ATTITUDE (existing messages) without needing FOLLOW_TARGET.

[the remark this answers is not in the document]
lthall: Absolutly not. Stale values are worse than zero. Stale valuse will a random cause a random offset error. Zeros will result in a sensible and predictable following error.

There is no point getting the same data from multiple message sources when we can simply select the prefered one. In my implementation of the FOLLOW_TARGET message the data will be different to the GLOBAL_POSITION_INT for example.

timtuxworth:
> We need to decide on how we choose between the different data message sources. Ideally I would like to change the name and extend the _ALT_TYPE parameter to be 0: GPI absolute, 1: GPI relative, 2: Follow Target (GPI = global position int).

I'd like to see the _ALT_TYPE parameter align with Location::AltFrame, and in particular, ABOVE_TERRAIN = 3. Following at terrain altitude is an important use case.

> We also need to decide what the options and default settings should be. We can leave the kinimatic interpolation off by default, we may choose not to process acceleration to reduce noise.

Why would we choose to leave kinematic off?

lthall:
> Following at terrain altitude is an important use case. I've already included this change in my Plane Follow PR.

Currently I have no plans to handle terain altitude. I am not sure how we would specify this.

> Why would we choose to leave kinematic off?

The main reason is to maintain the current behaviour.

timtuxworth:
> > Following at terrain altitude is an important use case. I've already included this change in my Plane Follow PR.
> 
> Currently I have no plans to handle terain altitude. I am not sure how we would specify this.

I do it one of two ways, depending on a parameter
1. Explicitly specify (via parameter) a fixed altitude value that the following vehicle will use for it's terrain altitude or
2. Use change_alt_frame to lookup the terrain altitude of the target before using this value as the target for the follow vehicle

lthall:
> > > Following at terrain altitude is an important use case. I've already included this change in my Plane Follow PR.
> > 
> > 
> > Currently I have no plans to handle terain altitude. I am not sure how we would specify this.
> 
> I do it one of two ways, depending on a parameter
> 
> 1. Explicitly specify (via parameter) a fixed altitude value that the following vehicle will use for it's terrain altitude or
> 2. Use change_alt_frame to lookup the terrain altitude of the target (with offsets) before using this value as the target altitude for the follow vehicle

I really don't understand why you would do this. What does this enable?

timtuxworth:
> > > > Following at terrain altitude is an important use case. I've already included this change in my Plane Follow PR.
> > > 
> > > 
> > > Currently I have no plans to handle terain altitude. I am not sure how we would specify this.
> > 
> > 
> > I do it one of two ways, depending on a parameter
> > 
> > 1. Explicitly specify (via parameter) a fixed altitude value that the following vehicle will use for it's terrain altitude or
> > 2. Use change_alt_frame to lookup the terrain altitude of the target (with offsets) before using this value as the target altitude for the follow vehicle
> 
> I really don't understand why you would do this. What does this enable?

I work with two clients who use fixed wing VTOLs and quadcopter for magnetometry surveys. For them it's very important that each vehicle to follow the terrain at a specific height, often around 40m AGL. For use case #1 the planes are typically flying over mountainous areas. The two planes fly with an Y offset of around 50m -75m and the terrain height can be quite different even only 50m apart. It would give invalid results if the two drones flew at the same AMSL altitude. For example if the two planes are flying along a valley, one must fly much higher to be 40m above the ground if it is flying further up the slope. To be clear as well, because the terrain undulates, the vehicle cannot fly AMSL as it will likely hit something and maintaining a consistent 40m above the ground is important for the correct data to be collected.

For the second use case, the client flies quadcopter drones over remote areas. They fly with magnetometers hanging below the drone and need (again) to maintain 40m AGL. The terrain is not as extreme as the first use case, but flying multiple quadcopters allows them to collect more data faster, but each quad must fly a consistent path above the terrain at it's own location. This client asked me to write a Lua script to overcome the existing Copter follow code to allow it to follow AGL, but it would be much better IMO if the c++ code knew how to follow the terrain, this script is really a bit of a hack. 
https://github.com/timtuxworth/ardupilot_scripts/blob/main/copter-follow-terrain.lua

lthall:
> > > > > Following at terrain altitude is an important use case. I've already included this change in my Plane Follow PR.
> > > > 
> > > > 
> > > > Currently I have no plans to handle terain altitude. I am not sure how we would specify this.
> > > 
> > > 
> > > I do it one of two ways, depending on a parameter
> > > 
> > > 1. Explicitly specify (via parameter) a fixed altitude value that the following vehicle will use for it's terrain altitude or
> > > 2. Use change_alt_frame to lookup the terrain altitude of the target (with offsets) before using this value as the target altitude for the follow vehicle
> > 
> > 
> > I really don't understand why you would do this. What does this enable?
> 
> I work with two clients who use fixed wing VTOLs and quadcopter for magnetometry surveys. For them it's very important that each vehicle to follow the terrain at a specific height, often around 40m AGL. For use case #1 the planes are typically flying over mountainous areas. The two planes fly with an Y offset of around 50m -75m and the terrain height can be quite different even only 50m apart. It would give invalid results if the two drones flew at the same AMSL altitude. For example if the two planes are flying along a valley, one must fly much higher to be 40m above the ground if it is flying further up the slope. To be clear as well, because the terrain undulates, the vehicle cannot fly AMSL as it will likely hit something and maintaining a consistent 40m above the ground is important for the correct data to be collected.
> 
> For the second use case, the client flies quadcopter drones over remote areas. They fly with magnetometers hanging below the drone and need (again) to maintain 40m AGL. The terrain is not as extreme as the first use case, but flying multiple quadcopters allows them to collect more data faster, but each quad must fly a consistent path above the terrain at it's own location. This client asked me to write a Lua script to overcome the existing Copter follow code to allow it to follow AGL, but it would be much better IMO if the c++ code knew how to follow the terrain, this script is really a bit of a hack. https://github.com/timtuxworth/ardupilot_scripts/blob/main/copter-follow-terrain.lua

Thanks for the application example. I can see how that could be important. However, I think we are getting out of the scope of the current follow messages. Neither GPI or FT messages provide terrain altitude. Terrain is also not subject to any kinimatic limitations so it poses additional (but managable) challenges.

As for follow mode terrain following in copter, that has some more obvious and elegent solutions. I have considered that it may be usefull for aircraft to only follow in the horzontal while having manual control of altitude. This could be extended to use terrain following or use a default altitude. All this is a seperate problem to this PR or the update to follow mode I have planned after this.

timtuxworth:
> As for follow mode terrain following in copter, that has some more obvious and elegent solutions. I have considered that it may be useful for aircraft to only follow in the horizontal while having manual control of altitude. This could be extended to use terrain following or use a default altitude. All this is a seperate problem to this PR or the update to follow mode I have planned after this.

Yes I agree it's outside the scope of this PR, just wanted to make the _ALT_TYPE parameter aligns with Location::AltFrame, which is how I'm using it.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,638 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+
+            // initialise offsets dynamically if needed
+            init_offsets_if_required(dist_vec_ned_m.tofloat());
+
+            // record distance and bearing to target for reporting/logging
+            if (ofs_dist_vec.xy().is_zero()) {
+                clear_dist_and_bearing_to_target();
+            } else {
+                _dist_to_target_m = ofs_dist_vec.xy().length();
+                _bearing_to_target_deg = degrees(ofs_dist_vec.xy().angle());
```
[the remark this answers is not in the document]
lthall: The original code was:
```
    // record distance and heading_deg for reporting purposes
    if (is_zero(dist_with_offs_m.x) && is_zero(dist_with_offs_m.y)) {
        clear_dist_and_bearing_to_target();
    } else {
        _dist_to_target_m = safe_sqrt(sq(dist_with_offs_m.x) + sq(dist_with_offs_m.y));
        _bearing_to_target_deg = degrees(atan2f(dist_with_offs_m.y, dist_with_offs_m.x));
    }
```

So it was the offset location.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,638 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+            const Vector3p ofs_dist_vec = _ofs_estimate_pos_ned_m - current_position_ned_m.topostype();
+
+            // initialise offsets dynamically if needed
+            init_offsets_if_required(dist_vec_ned_m.tofloat());
+
+            // record distance and bearing to target for reporting/logging
+            if (ofs_dist_vec.xy().is_zero()) {
+                clear_dist_and_bearing_to_target();
+            } else {
+                _dist_to_target_m = ofs_dist_vec.xy().length();
```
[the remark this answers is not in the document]
lthall: Fixed

lthall:
I have opened up the test for the QuadPlane.ShipLanding test because it has a fundermental accurasy error. The test in this PR has the following errors in the east direction:

- Follow estimation < 2mm
- Position Control error < 70mm
- LUA based error ~ 1m

I can't see how we get:
 ShipLanding (ship landing test) (Did not land within 1m of ship distance=2.297590104696072)

tridge:
I think the key reason for the incorrect landing point for the ship landing is the AP_Follow::init_offsets_if_required() function. The quadplane ShipLanding autotest gets the distance from the location of the ship and the aircraft, but the AP_Follow::init_offsets_if_required() function auto-initialises an offset, setting FOLLOFS{X,Y,Z}. The plane lands at 0.8 from the target if you add FOLL_OFS_X=0.01 to the set of parameters in the test, as this prevents the automatic offset code from running.

lthall:
> I think the key reason for the incorrect landing point for the ship landing is the AP_Follow::init_offsets_if_required() function. The quadplane ShipLanding autotest gets the distance from the location of the ship and the aircraft, but the AP_Follow::init_offsets_if_required() function auto-initialises an offset, setting FOLL_OFS_{X,Y,Z}. The plane lands at 0.8 from the target if you add FOLL_OFS_X=0.01 to the set of parameters in the test, as this prevents the automatic offset code from running.

I have confirmed this. I would have expected this to happen before but looking at the old code we only automaticly set the offsets if we were using get_target_dist_and_vel_NED_m. The shiplanding LUA script is using get_target_location_and_velocity_ofs so it isn't set automaticly if the offsets are zero.

The question is what do we do from here. This seems like a big inconsistancy in the libary. Do we:
1. Set it on all functions like I have here
2. Call the function in Follow Mode in copter (I assume plane doesn't use it in the same way)
3. Some other suggestion I can't think of???

rmackay9:
I haven't read everything above but in general Follow's behaviour of interpreting all zero offsets as meaning the user wants to follow at the current distance is not good and we could remove this defaulting.  It seemed like a good idea at the time but in case it helps we should change this behaviour to be used based on a parameter or or an argument in a mavlink message that initiates the follow behaviour

timtuxworth:
> I haven't read everything above but in general Follow's behaviour of interpreting all zero offsets as meaning the user wants to follow at the current distance is not good and we could remove this defaulting. It seemed like a good idea at the time but in case it helps we should change this behaviour to be used based on a parameter or or an argument in a mavlink message that initiates the follow behaviour

Makes sense @rmackay9, but in that case what should happen if the user doesn't set the offsets? Should it fail to follow? I could maybe add a pre-arm for this case to my scripted arming checks.

lthall:
> > I haven't read everything above but in general Follow's behaviour of interpreting all zero offsets as meaning the user wants to follow at the current distance is not good and we could remove this defaulting. It seemed like a good idea at the time but in case it helps we should change this behaviour to be used based on a parameter or or an argument in a mavlink message that initiates the follow behaviour
> 
> Makes sense @rmackay9, but in that case what should happen if the user doesn't set the offsets? Should it fail to follow? I could maybe add a pre-arm for this case to my scripted arming checks.

For this PR I have maintained the current behaviour but called it directly from the mode rather than automatically as part of only one interface.

libraries/AP_Math/control.cpp
```diff
@@ -313,7 +313,7 @@ void shape_pos_vel_accel(postype_t pos_input, float vel_input, float accel_input
     vel_target += vel_input;
 
     // limit velocity between vel_min and vel_max
-    if (limit_total) {
+    if (limit_total && (is_negative(vel_min) || is_positive(vel_max))) {
```
timtuxworth, line 316: This test would pass if vel_min > vel_max as long as vel_max was positive or if vel_min is negative and vel_max < vel_min. I don't know what the constrain would return in that case. Perhaps you mean || to be && or alternative you need to add && vel_min < vel_max.

timtuxworth (commented):
Looking forward to getting this in, but is there a problem with that if statement in control.cpp?

Tools/autotest/quadplane.py
```diff
@@ -1568,6 +1568,7 @@ def ShipLanding(self):
             "Q_WP_SPEED": 700,
             "SIM_SHIP_DSIZE": 10,
             "FOLL_ENABLE": 1,
+            "FOLL_DIST_MAX": 1000,
```
tridge, line 1571: we should discuss if we should require this. The test passes without it, but with a lot of "No target location/velocity" messages. Previously the APIs that ship landing used didn't depend on FOLL_DIST_MAX being set

ArduCopter/mode_follow.cpp
```diff
@@ -20,7 +20,7 @@ bool ModeFollow::init(const bool ignore_checks)
         gcs().send_text(MAV_SEVERITY_WARNING, "Set FOLL_ENABLE = 1");
         return false;
     }
-
+
```
IamPete1, line 23:
```suggestion

```

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,634 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
     // @Param: _OPTIONS
     // @DisplayName: Follow options
     // @Description: Follow options bitmask
     // @Values: 0:None,1: Mount Follows lead vehicle on mode enter
     // @User: Standard
     AP_GROUPINFO("_OPTIONS", 11, AP_Follow, _options, 0),
 
+    // @Param: _ACCEL_NE
+    // @DisplayName: Acceleration limit for the horizontal kinematic input shaping
+    // @Description: Acceleration limit of the horizontal kinematic path generation used to determine how quickly the ship varies in velocity
```
IamPete1, line 169:
```suggestion
    // @Description: Acceleration limit of the horizontal kinematic path generation used to determine how quickly the target varies in velocity
```
Not everyone is landing on ships ;)

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,634 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+    // @Range: 0 5
+    // @Units: m/s/s
+    // @User: Advanced
+    AP_GROUPINFO("_ACCEL_NE", 12, AP_Follow, _accel_max_ne_mss, 2.5),
+
+    // @Param: _JERK_NE
+    // @DisplayName: Jerk limit for the horizontal kinematic input shaping
+    // @Description: Jerk limit of the horizontal kinematic path generation used to determine how quickly the ship varies in acceleration
```
IamPete1, line 177: ship again

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,634 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+    // @Range: 0 20
+    // @Units: m/s/s/s
+    // @User: Advanced
+    AP_GROUPINFO("_JERK_NE", 13, AP_Follow, _jerk_max_ne_msss, 5.0),
+
+    // @Param: _ACCEL_D
+    // @DisplayName: Acceleration limit for the vertical kinematic input shaping
+    // @Description: Acceleration limit of the vertical kinematic path generation used to determine how quickly the ship varies in velocity
```
IamPete1, line 185: ship again

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,634 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+    // @Range: 0 2.5
+    // @Units: m/s/s
+    // @User: Advanced
+    AP_GROUPINFO("_ACCEL_D", 14, AP_Follow, _accel_max_d_mss, 2.5),
+
+    // @Param: _JERK_D
+    // @DisplayName: Jerk limit for the vertical kinematic input shaping
+    // @Description: Jerk limit of the vertical kinematic path generation used to determine how quickly the ship varies in acceleration
```
IamPete1, line 193: ship

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,634 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+    // @Range: 0 5
+    // @Units: m/s/s/s
+    // @User: Advanced
+    AP_GROUPINFO("_JERK_D", 15, AP_Follow, _jerk_max_d_msss, 5.0),
+
+    // @Param: _ACCEL_H
+    // @DisplayName: Angular acceleration limit for the heading kinematic input shaping
+    // @Description: Angular acceleration limit of the heading kinematic path generation used to determine how quickly the ship varies in angular velocity
```
IamPete1, line 201: ship

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,634 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+    // @Range: 0 90
+    // @Units: deg/s/s
+    // @User: Advanced
+    AP_GROUPINFO("_ACCEL_H", 16, AP_Follow, _accel_max_h_degss, 90.0),
+
+    // @Param: _JERK_H
+    // @DisplayName: Angular jerk limit for the heading kinematic input shaping
+    // @Description: Angular jerk limit of the heading kinematic path generation used to determine how quickly the ship varies in angular acceleration
```
IamPete1, line 209: ship

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -613,13 +910,16 @@ bool AP_Follow::have_target(void) const
...
+// Accessor for the AP_Follow singleton instance.
+namespace AP {
+    AP_Follow &follow()
+    {
+        return *AP_Follow::get_singleton();
+    }
 }
 
-#endif  // AP_FOLLOW_ENABLED
+#endif  // AP_FOLLOW_ENABLED
```
IamPete1, line 925:
```suggestion
#endif  // AP_FOLLOW_ENABLED

```

libraries/AP_Follow/AP_Follow.h
```diff
@@ -153,33 +186,71 @@ class AP_Follow
...
-    JitterCorrection _jitter{3000};
+    //==========================================================================
+    // Parameters
+    //==========================================================================
+
+    AP_Int8     _enabled;
+    AP_Int16    _sysid;
+    AP_Float    _dist_max_m;
+    AP_Int8     _offset_type;
+    AP_Vector3f _offset_XXD_m;
```
IamPete1, line 197: XXD? XYD?

libraries/AP_Math/control.cpp
```diff
@@ -352,13 +352,69 @@ void shape_pos_vel_accel_xy(const Vector2p& pos_input, const Vector2f& vel_input
...
+    angle_error += angle_vel * stopping_time;
+
+    // Calculate time constants and limits to ensure stable operation
+    // The negative acceleration limit is used here because the square root controller
+    // manages the approach to the setpoint. Therefore the acceleration is in the opposite
+    // direction to the position error.
+    float angle_accel_tc_max;
+    float KPv;
+    angle_accel_tc_max = 0.5 * angle_accel_max;
+    KPv = 0.5 * angle_jerk_max / angle_accel_max;
```
IamPete1, line 397:
```suggestion
    const float angle_accel_tc_max = 0.5 * angle_accel_max;
    const float KPv = 0.5 * angle_jerk_max / angle_accel_max;
```
Const if possible.

libraries/AP_Follow/AP_Follow.h
```diff
@@ -153,33 +186,71 @@ class AP_Follow
...
+    AP_Int8     _yaw_behave;
+    AP_Int8     _alt_type;
+    AC_P        _p_pos;
+    AP_Int16    _options;
+    AP_Float    _accel_max_ne_mss;
```
rmackay9, line 202: overall this looks pretty good but I wonder why we've removed all the comments to the right of the variables and parameters?

Tools/autotest/quadplane.py
tridge: maybe default FOLL_DIST_MAX to 0 on plane? and make 0 mean no limit?

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -134,419 +155,634 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
     AP_GROUPINFO("_ALT_TYPE", 10, AP_Follow, _alt_type, AP_FOLLOW_ALT_TYPE_DEFAULT),
 #endif
 
+// May add option to turn on input shaping, turn on acceleration due to rotation of offset
```
rmackay9, line 158: we should probably prefix, "ToDo:" on this comment

rmackay9 (commented):
LGTM, just a request from me to restore some comments. thanks!

tridge:
with defaulting FOLL_OFS_MAX to 0 on plane I'm happy, thanks!

libraries/AP_Follow/AP_Follow.cpp
lthall: Thanks, good catch!

libraries/AP_Follow/AP_Follow.h
lthall: Changed to _offset_m

rmackay9 (approved):

timtuxworth (approved):
I'm looking forward to rebasing Plane Follow on this.

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -141,412 +164,625 @@ const AP_Param::GroupInfo AP_Follow::var_info[] = {
...
+        // update heading angle separately to maintain proper wrapping [-PI, PI]
+        postype_t estimate_heading_rad = _estimate_heading_rad;
+        update_pos_vel_accel(estimate_heading_rad, _estimate_heading_rate_rads, _estimate_heading_accel_radss, e_dt, 0.0, 0.0, 0.0);
+        _estimate_heading_rad = wrap_PI(estimate_heading_rad);
+    } else {
+        // no valid estimate yet: initialise from latest target position
+        _estimate_pos_ned_m = _target_pos_ned_m + delta_pos.topostype();
+        _estimate_vel_ned_ms = _target_vel_ned_ms + delta_vel;
+        _estimate_accel_ned_mss = _target_accel_ned_mss;
+        _estimate_valid = true;
```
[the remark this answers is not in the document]
lthall: Great catch!

libraries/AP_Math/control.cpp
lthall: ` if (limit_total && (vel_max>vel_min) {`

PR #29899, opened 2025-04-26
Copter: Shorten the lifetime of stack variables

muramura:
I will shorten the lifetime of stack variables.
There is no advantage to prolonging the lifetime of stack variables.


peterbarker (approved):
LGTM

rmackay9 (approved):

PR #28566, opened 2024-11-09
MAVLink: add support for the `AVAILABLE_MODES` message

IamPete1:
This message allows the GCS to get the list of available modes at runtime. In the future this allows things like https://github.com/ArduPilot/ardupilot/pull/28110

This requires: https://github.com/ArduPilot/mavlink/pull/374

This message is supported by the QGC daily build and has been tested using it.

For example if you connect to a plane without Quadplane enabled it will only show the fixed wing modes:
image

If you enable quadplane and re-connect you will then see all modes:
image

Current master shows a preconfigured subset of modes:
image

Editing a mode name in the cpp proves that it is updated with the new message:
image

Future work (not for this PR):

- Fillout the message more completely:
  - Set `MAV_MODE_PROPERTY` bitmask for each mode.
  - Set `MAV_STANDARD_MODE` enum for each mode

- Report the current mode with the new `CURRENT_MODE` message

- Send `AVAILABLE_MODES_MONITOR` to allow the GCS to notice and re-request if the number of modes changes.

- Support `MAV_CMD_DO_SET_STANDARD_MODE`

- Addition of new modes from scripting and more.

PR #29903, opened 2025-04-26
Sub:fix bug in mode.h

Hwurzburg:
Wrong mode ENUM was used in #28566


peterbarker:
Commit message should be better, eg `Sub: correct mode enumeration value returned by ModeSurface`.

PR #29905, opened 2025-04-26
AP_GPS: DroneCAN: only trust ellipsoid height if different from MSL

tpwrules:
AP_Periph sends `ellipsoid == MSL` if the underlying GPS driver can't retrieve an ellipsoid height for whatever reason. Older versions of AP_Periph (including all Here3+ firmwares at the time of writing) always send `MSL == ellipsoid` and never report a valid ellipsoid height.

To accommodate this, only use the ellipsoid height reported over DroneCAN to compute an undulation if we've ever seen it different from the MSL height. This ensures we accurately represent the equipment capability, at least if the GPS is not changed at runtime.

Future work will establish an "unknown" ellipsoid height value in the DroneCAN standard and always ignore it.

Tested on Cube Orange and MatekL431-Periph nodes (running various code versions) connected to a u-blox GPS.

### DIFF
--- a/libraries/AP_GPS/AP_GPS_DroneCAN.cpp
+++ b/libraries/AP_GPS/AP_GPS_DroneCAN.cpp
@@ -381,8 +381,17 @@ void AP_GPS_DroneCAN::handle_fix2_msg(const uavcan_equipment_gnss_Fix2& msg, uin
         loc.lat = msg.latitude_deg_1e8 / 10;
         loc.lng = msg.longitude_deg_1e8 / 10;
         const int32_t alt_amsl_cm = msg.height_msl_mm / 10;
-        interim_state.have_undulation = true;
-        interim_state.undulation = (msg.height_msl_mm - msg.height_ellipsoid_mm) * 0.001;
+        // if ellipsoid height is not supported by the GPS driver (or always
+        // with older releases), AP_Periph reports height_ellipsoid_mm == height_msl_mm .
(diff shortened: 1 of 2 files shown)


libraries/AP_GPS/AP_GPS_DroneCAN.cpp
```diff
@@ -381,8 +381,17 @@ void AP_GPS_DroneCAN::handle_fix2_msg(const uavcan_equipment_gnss_Fix2& msg, uin
         loc.lat = msg.latitude_deg_1e8 / 10;
         loc.lng = msg.longitude_deg_1e8 / 10;
         const int32_t alt_amsl_cm = msg.height_msl_mm / 10;
-        interim_state.have_undulation = true;
-        interim_state.undulation = (msg.height_msl_mm - msg.height_ellipsoid_mm) * 0.001;
+        // if ellipsoid height is not supported by the GPS driver (or always
+        // with older releases), AP_Periph reports height_ellipsoid_mm == height_msl_mm .
```
IamPete1, line 385: Should we stop doing that? Although it looks like there is no way to report as not available in the message.

tpwrules: Yes, as mentioned future work is to define a "not available" value and make at least both sides of ArduPilot use it.

tridge (approved):

rmackay9:

PR #29908, opened 2025-04-27
AP_GPS: Define the message IDs for NOVA

muramura:
The URL of the PDF file is incorrect. 
Define the message IDs and use a SWITCH statement.

### DIFF
--- a/libraries/AP_GPS/AP_GPS_NOVA.cpp
+++ b/libraries/AP_GPS/AP_GPS_NOVA.cpp
@@ -15,7 +15,7 @@
 
 //  Novatel/Tersus/ComNav GPS driver for ArduPilot.
 //  Code by Michael Oborne
-//  Derived from http://www.novatel.com/assets/Documents/Manuals/om-20000129.pdf
+//  Derived from https://hexagondownloads.blob.core.windows.net/public/Novatel/assets/Documents/Manuals/om-20000129/om-20000129.pdf
 
 #include "AP_GPS.h"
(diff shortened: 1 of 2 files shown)


Ryanf55 (approved):
Nice, the switch is more efficient than the `if` statement too.

libraries/AP_GPS/AP_GPS_NOVA.h
```diff
@@ -89,6 +89,7 @@ class AP_GPS_NOVA : public AP_GPS_Backend
         uint16_t recvswver;
     };    
 
+    static const uint8_t NOVA_PSRDOP = 174;
```
peterbarker, line 92: Is there any reason not to create an `enum class` to hold these identifiers?

libraries/AP_GPS/AP_GPS_NOVA.cpp
```diff
@@ -272,16 +271,18 @@ AP_GPS_NOVA::process_message(void)
...
-    {
+    case NOVA_PSRDOP: {  // psrdop
         const psrdop &psrdopu = nova_msg.data.psrdopu;
 
         state.hdop = (uint16_t) (psrdopu.hdop*100);
         state.vdop = (uint16_t) (psrdopu.htdop*100);
         return false;
     }
+    default:
+        break;
```
peterbarker, line 284: Not needed?

peterbarker (approved):

peterbarker:
Merged, thanks!

PR #29910, opened 2025-04-27
autotest: add support for adding prefix while launching tmux new-window

bugobliterator:
This allows a method to launch gdbserver from vscode extension

### DIFF
--- a/Tools/autotest/run_in_terminal_window.sh
+++ b/Tools/autotest/run_in_terminal_window.sh
@@ -29,7 +29,7 @@ if [ -n "$SITL_RITW_TERMINAL" ]; then
   chmod +x "$FILEPATH"
   $SITL_RITW_TERMINAL "$FILEPATH" &
 elif [ -n "$TMUX" ]; then
-  tmux new-window -dn "$name" "$*"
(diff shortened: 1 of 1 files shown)


Tools/autotest/run_in_terminal_window.sh
```diff
@@ -28,6 +28,8 @@ if [ -n "$SITL_RITW_TERMINAL" ]; then
   printf "%q " "$@" >>"$FILEPATH"
   chmod +x "$FILEPATH"
   $SITL_RITW_TERMINAL "$FILEPATH" &
+elif [ -n "$TMUX_PREFIX" ]; then
+  tmux new-window -dn "$name" "$TMUX_PREFIX $*"
 elif [ -n "$TMUX" ]; then
   tmux new-window -dn "$name" "$*"
```
peterbarker, line 34: Could we put the prefix stuff inside the existing TMUX clause?

It means you do need to set two environment variables but will avoid any cross-product problems.

bugobliterator: done

peterbarker (approved):
LGTM

PR #29916, opened 2025-04-28
fix enum format

christianrauch:
The `fprintf` for the `LINUX_BOARD_TYPE` fails due to the format specifier that is incompatible with enums:
```
../../libraries/AP_HAL_Linux/RCInput_RPI.cpp: In member function ‘void Linux::RCInput_RPI::set_physical_addresses()’:
../../libraries/AP_HAL_Linux/RCInput_RPI.cpp:293:56: error: format ‘%i’ expects argument of type ‘int’, but argument 3 has type ‘Linux::LINUX_BOARD_TYPE’ [-Werror=format=]
  293 |         fprintf(stderr,"Unknown Linux Board version!: %i\n", _version);
      |                                                       ~^     ~~~~~~~~
      |                                                        |     |
      |                                                        int   Linux::LINUX_BOARD_TYPE
cc1plus: some warnings being treated as errors
```

This PR fixes the issue with an explicit cast.

### DIFF
--- a/libraries/AP_HAL_Linux/RCInput_RPI.cpp
+++ b/libraries/AP_HAL_Linux/RCInput_RPI.cpp
@@ -290,7 +290,7 @@ void RCInput_RPI::set_physical_addresses()
         clk_base = RCIN_RPI_RPI4_CLK_BASE;
         pcm_base = RCIN_RPI_RPI4_PCM_BASE;
     } else {
-        fprintf(stderr,"Unknown Linux Board version!\n");
+        fprintf(stderr,"Unknown Linux Board version: %i\n", int(_version));
         exit(-1);
     }
 }


christianrauch:
> Please remove the remove-trailing-whitespace commit. We don't enforce no-trailing-whitespace at a technical level in the repository, so removing it just screws up our git history when trying to track stuff down.

I see. But a lot of IDEs actually do this automatically, and most projects I work with have code style settings that enforce such things. I think it wouldn't hurt to also enforce code style here.

> So the substantive change here isn't what your opening comment suggests it is. It's more, "inform the user of unknown RPi version number". You're adding this functionality (and presumably discovered you needed to add a cast while you were there :-) )

I don't understand this comment. The code was already there and throws this error. This somehow slipped your CI, so I just fixed this in commit https://github.com/ArduPilot/ardupilot/pull/29916/commits/9cead776ae7cb0b2d6ab77736db29adc22ec0f05.

> This isn't a bad change, but the commit message (not the PR title) needs to conform to our standards (which you can find in our style guide). Probably `AP_HAL_Linux: inform user of bad RPi board version`.
> 
> Also, please lose the exclamation mark in the message... it just looks wonky :-)

Well, I can just repeat: The code with its exclamation mark was already there. So... do you want me to remove the exclamation mark in a separate commit (because it is unrelated to the compilation fix)?

peterbarker:
> > Please remove the remove-trailing-whitespace commit. We don't enforce no-trailing-whitespace at a technical level in the repository, so removing it just screws up our git history when trying to track stuff down.
> 
> I see. But a lot of IDEs actually do this automatically, and most projects I work with have code style settings that enforce such things. I think it wouldn't hurt to also enforce code style here.

At the risk of repeating myself, please remove the patch which changes the trailing whitespace.  Remove trailing whitespace when you're making changes to the lines (or the lines around it for whitespace-only lines), but occluding history in the codebase isn't worth the extra tidiness we get from such patches.

> > So the substantive change here isn't what your opening comment suggests it is. It's more, "inform the user of unknown RPi version number". You're adding this functionality (and presumably discovered you needed to add a cast while you were there :-) )
> 
> I don't understand this comment. The code was already there and throws this error. This somehow slipped your CI, so I just fixed this in commit [9cead77](https://github.com/ArduPilot/ardupilot/commit/9cead776ae7cb0b2d6ab77736db29adc22ec0f05).

No, the code wasn't there.  This is your change:
image

Your PR should consist of one patch titled "AP_HAL_Linux: supply bad board version number to user" - > Well, I can just repeat: The code with its exclamation mark was already there. So... do you want me to remove the exclamation mark in a separate commit (because it is unrelated to the compilation fix)?

I believe you introduced the compilation failure in the first place by omitting a cast.

Adding the erroneous board version number to this message is a good patch to add to ArduPilot - but I thought while you were adding that removing the exclamation mark was a good idea.

christianrauch:
Sorry, I misread your comments. Your comments about the "added functionality" and the "wonky exclamation mark" made it sound as if I added the full `Unknown Linux Board version!` printout. But I indeed just changed that printout.

I removed the style fix and renamed the commit according to your suggestion.

PR #29917, opened 2025-04-29
CI: revert "build CubeRed-EKF2 not CubeOrange-EKF2 in CI"

tpwrules:
Reverts ArduPilot/ardupilot#28128 now that CubeRed is out of flash and CubeOrange has loads free.


rmackay9:
I wonder if we could rename the commit to CI: Revert build CubeBlack-EKF2 not CubeOrange-EKF2 in CI

peterbarker (approved):

PR #29919, opened 2025-04-29
Replay: add option to show progress of log file processed

bugobliterator:
Adds a progress bar to show percentage of file processed through replay


peterbarker (approved):
LGTM

PR #29921, opened 2025-04-29
Ensure stable IMU timing on RealFlight

andyp1per:
The current RealFlight approach means that we get highly variable loops and inter-loop dt's. This seems to significantly impact attitude control under arduous conditions. This PR generates high rate synthetic gyro samples that are sync'ed to the main loop's requests for samples. This results in very even loop dts whilst still ensuring that the latest data is always being used and appears to significantly help with attitude control.

Here is the updated, technically robust description. It now accounts for both the **transport jitter** (variable arrival time) and the **variable decimation** (skipped frames/variable physics steps).

***

**The Core Argument:**
This PR upgrades the `SIM_FlightAxis` backend from a "sampling" model to an "integrating" model. It resolves the critical instability caused by RealFlight’s **variable frame decimation** and **transport jitter**, where the simulator might run at 1200Hz but deliver frames irregularly (e.g., ~600Hz, skipping intermediate steps).

### **1. The Problem: Variable Decimation & Time Aliasing**
The interface faces two distinct challenges that the legacy fixed-schedule IMU could not handle:
* **Irregular Decimation:** RealFlight updates physics at high rates (~1200Hz) but drops frames non-deterministically before sending. Sometimes we receive adjacent frames; other times we skip intermediates.
* **The Failure Mode:** The old fixed-rate IMU timer was "blind" to these skips. If a frame was skipped, the state would jump significantly between updates. The fixed-rate IMU would interpret this jump as an instantaneous high-G event or snap rotation, feeding massive noise into the EKF and control loops.

### **2. The Solution: Adaptive Upsampling (Elastic Time)**
The PR shifts to an event-driven model that reconstructs the missing data:
* **Slave to the Frame:** We wait for a frame to arrive (`wait_for_sample`).
* **Delta-Based Interpolation:** Instead of assuming a constant timestep, the code calculates the **true simulation `dt`** between the current frame and the last received frame.
    * If we receive adjacent frames, it generates a standard batch of synthetic samples.
    * If frames were skipped (larger `dt`), it automatically scales the number of synthetic samples to fill that larger gap smoothly.
* **Ignoring Arrival Jitter:** By relying on the *simulation* time delta rather than the *wall clock* arrival time, we effectively filter out the OS/Network latency jitter.

### **3. The Engineering Win**
This mechanism acts as a **smart interpolation layer**.
* It smooths over the "stutter" of skipped frames by generating a continuous stream of IMU data that matches the average velocity over the gap.
* It ensures the Autopilot always sees a coherent timeline, regardless of whether the simulator sent every frame or every other frame.

**Summary:**
We cannot control RealFlight's output variability, but we can control how we digest it. This PR allows ArduPilot to "ride" the variable stream of data without aliasing, ensuring the EKF receives a smooth, physically accurate timeline essential for precision tuning.

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor_SITL.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor_SITL.cpp
@@ -310,6 +310,12 @@ void AP_InertialSensor_SITL::generate_gyro()
 
 void AP_InertialSensor_SITL::timer_update(void)
 {
+    // on some simulations (RealFlight) the aircraft sim decides when a new sample is available.
+    if (sitl->state.flightaxis_imu_frame_num > 0) {
+        update_from_frame();
+        return;
+    }
+
     uint64_t now = AP_HAL::micros64();
 #if 0
     // insert a 1s pause in IMU data. This triggers a pause in EK2
@@ -322,16 +328,7 @@ void AP_InertialSensor_SITL::timer_update(void)
         return;
     }
 
-    // on some simulations (RealFlight) the aircraft sim decides when a new sample is available.
-    const bool sync_imus_to_frames = frame_num > 0;
-
-    if (sync_imus_to_frames && frame_num == sitl->state.frame_num) {
-        return;
-    }
-
-    frame_num = sitl->state.frame_num;
-
-    if (now >= next_accel_sample || sync_imus_to_frames) {
+    if (now >= next_accel_sample) {
         if (((1U << accel_instance) & sitl->accel_fail_mask) == 0) {
 #if AP_SIM_INS_FILE_ENABLED
             if (sitl->accel_file_rw == SITL::SIM::INSFileMode::INS_FILE_READ
@@ -349,7 +346,7 @@ void AP_InertialSensor_SITL::timer_update(void)
             }
         }
     }
-    if (now >= next_gyro_sample || sync_imus_to_frames) {
+    if (now >= next_gyro_sample) {
         if (((1U << gyro_instance) & sitl->gyro_fail_mask) == 0) {
 #if AP_SIM_INS_FILE_ENABLED
             if (sitl->gyro_file_rw == SITL::SIM::INSFileMode::INS_FILE_READ
@@ -367,6 +364,26 @@ void AP_InertialSensor_SITL::timer_update(void)
                 }
             }
         }
+    }
+}
+
+void AP_InertialSensor_SITL::update_from_frame(void)
+{
+    if (sitl == nullptr) {
+        return;
+    }
+
+    if (flightaxis_imu_frame_num == sitl->state.flightaxis_imu_frame_num) {
+        return;
+    }
+
+    flightaxis_imu_frame_num = sitl->state.flightaxis_imu_frame_num;
+
+    if (((1U << accel_instance) & sitl->accel_fail_mask) == 0) {
+        generate_accel();
+    }
+    if (((1U << gyro_instance) & sitl->gyro_fail_mask) == 0) {
+        generate_gyro();
     }
 }
 
--- a/libraries/AP_InertialSensor/AP_InertialSensor_SITL.h
+++ b/libraries/AP_InertialSensor/AP_InertialSensor_SITL.h
@@ -32,6 +32,7 @@ private:
     void generate_gyro();
     float get_temperature(void);
     void update_file();
+    void update_from_frame();
 #if AP_SIM_INS_FILE_ENABLED
     void read_gyro(const float* buf, uint8_t nsamples);
     void read_gyro_from_file();
@@ -47,7 +48,7 @@ private:
 
     uint64_t next_gyro_sample;
     uint64_t next_accel_sample;
-    uint64_t frame_num;
+    uint64_t flightaxis_imu_frame_num;
     float gyro_time;
     float accel_time;
     float gyro_motor_phase[32];


lbullard-c2:
I have been running this commit for a couple of weeks, while overall it's a big improvement, especially in challenging flight scenarios, I suspect 100us samples are a bit too fast, especially when running SITL and Realflight on separate computers. We're seeing some interesting networking issues, especially on startup of SITL. Reducing to `#define SAMPLE_INTERVAL_S 1.0e-3` has made a pretty big improvement, without really seeing any downsides (yet). Possibly something in between is ideal

andyp1per:
Thanks for testing @lbullard-c2 ! It probably works well on my machine because it is using RF and SITL co-located and it is a Ryzen 9 9950X with an RTX 5080OC graphics card - so on the limits of consumer hardware right now. I agree there is probably an ideal value, although in theory the sample interval should not affect the network performance, since the network round trips are going at whatever rate RF gives us and we are just subdividing to give even timing. 1e-3 has the disadvantage of being quite close to the other sampling intervals going on and I suspect there is therefore a decent amount of jitter. Maybe 5e-4 would be better as that is quite representative of the real world with gyros running at 2KHz.

lbullard-c2:
I'd say my particular network environment might be the culprit, SITL is running on a server with a few enterprise switches / firewalls etc in between that probably don't like the traffic

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -52,6 +52,10 @@ const AP_Param::GroupInfo FlightAxis::var_info[] = {
 // the asprintf() calls are not worth checking for SITL
 #pragma GCC diagnostic ignored "-Wunused-result"
 
+#define SAMPLE_INTERVAL_S 100.0e-6  // 100us
```
robertlong13: To address lbullard's concerns, we could add a `SIM_RF_SAMPLEHZ` parameter. That way, I could include that within defaults.parm for vehicles that don't need super-fast updates.

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -366,61 +397,129 @@ void FlightAxis::exchange_data(const struct sitl_input &input)
...
+        average_frame_time_s = average_frame_time_s * 0.98 + dt * 0.02;
+    }
+#if HAL_LOGGING_ENABLED
+    uint64_t time_now = uint64_t(state.m_currentPhysicsTime_SEC * 1.0e6);
+// @LoggerMessage: RF
+// @Description: RealFlight mode messages
+// @Field: TimeUS: Time since system startup
+// @Field: Dt: delta time between this frame and the previous frae=me
+// @Field: Fps: frames-per-second implied by the current delta time
+    AP::logger().WriteStreaming("RF", "TimeUS,Dt,Fps", "QdI", time_now, dt, uint32_t(roundf(1/dt)));
```
robertlong13: Great idea. Could we also put glitch count and socket error count? I think this PR has pretty much eliminated both of those entirely (for me anyway). I'm gonna let it run for several hours and log those, but it could be good to have those in there for the future too.

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -430,25 +529,6 @@ void FlightAxis::update(const struct sitl_input &input)
...
-        update_mag_field_bf();
-        extrapolated_s += delta_time;
-        report_FPS();
-        return;
-    }
-
-    extrapolated_s = 0;
     
     if (initial_time_s <= 0) {
         dt_seconds = 0.001f;
```
robertlong13, line 582: We need to add `time_now_us = 1` in this if-block to prevent time from going backward later. If `time_now_us` is not updated, then the call to `time_advance()` increments the time by `frame_time_us`, which starts at 4000. This means that time is increasing by larger increments than it should, so `time_now_us` just keeps getting further and further _ahead_ of `new_time_us`. Once it get's more than half a second ahead, line 631 triggers and we get back on track: `time_now_us` gets set equal to `new_time_us`, the next call to `time_advance()` doesn't increment (and thus doesn't over-increment), we finally reach the line that sets `frame_time_us` to the correct value.

Adding `time_now_us = 1` after line 455 fixes all that.

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -544,6 +624,7 @@ void FlightAxis::update(const struct sitl_input &input)
     time_advance();
     uint64_t new_time_us = (state.m_currentPhysicsTime_SEC - initial_time_s)*1.0e6;
     if (new_time_us < time_now_us) {
+        //printf("Time going backwards by %u\n", uint32_t(time_now_us - new_time_us));
         uint64_t dt_us = time_now_us - new_time_us;
         if (dt_us > 500000) {
```
robertlong13, line 677: A `time_now_us++` in here would also prevent the bug I mentioned above where `time_advance()` causes `time_now_us` to run away faster than `new_time_us`.

Also, maybe just deleting the call to `time_advance()` entirely? I'm not sure why it's actually needed in the FlightAxis backend.

robertlong13 (changes requested):
This is freaking awesome! I rebased my RealFlight autotest PR on this and it's working much more smoothly. I was getting pretty frequent socket timeouts before when running autotests through RealFlight.

Found an issue with time going backward. I had to get to the bottom of this because I kept getting an exception during DCM update right after time goes backward, and so I couldn't even test this PR. The time going backward is not new to your PR, but for some reason, it reliably leading to a floating-point exception in the DCM update is. Some subtle change in the timing I guess. I can reliably get it to happen running the Titan Cobra (params attached if you want to try to reproduce on your machine).
[realflight-titan-cobra.parm.txt](https://github.com/user-attachments/files/20691350/realflight-titan-cobra.parm.txt)

The time going backward causes a [gigantic dt in the Inertial Sensor backend](https://github.com/ArduPilot/ardupilot/blob/ce8745496930fad7b530d3e2f7ade02e27b9d12d/libraries/AP_InertialSensor/AP_InertialSensor_Backend.cpp#L304) and causes huge windup of the delta angle accumulator. This is arguably a bug in the InertialSensor library, since I think we could get the same issue if a sensor dies and comes back after a long time (we have a check for large dt, but it doesn't clear all the variables it needs to in order to prevent this).

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: Thanks, fixed. I suspect this is not really necessary since the limiting factor is the net roundtrip and the extra samples just increase SITL CPU load which should be in plentiful supply.

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: Fixed

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: Thanks fixed. A quick test on my setup seemed to have no issues with this.

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: Ok, I fixed the other way so will ignore.

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -430,27 +536,10 @@ void FlightAxis::update(const struct sitl_input &input)
...
-        extrapolated_s += delta_time;
-        report_FPS();
-        return;
-    }
-
-    extrapolated_s = 0;
     
+    // prevent time going backwards
     if (initial_time_s <= 0) {
+        time_now_us = 1;
```
robertlong13:
```suggestion
        time_now_us = 1; // prevents time from going backward
```

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -430,27 +536,10 @@ void FlightAxis::update(const struct sitl_input &input)
...
-        update_position();
-        update_mag_field_bf();
-        extrapolated_s += delta_time;
-        report_FPS();
-        return;
-    }
-
-    extrapolated_s = 0;
     
+    // prevent time going backwards
```
robertlong13:
```suggestion
    // initialize timer
```

libraries/SITL/SIM_FlightAxis.cpp
robertlong13: Ah damn, sorry. I was not paying enough attention. I didn't realize this logged at the full 5kHz rate. Go back to logging the bare minimum stuff you had in there and lock it behind an option bit (defaulted off IMO). I'll open a separate PR for the extra crap I want to log. Sorry again.

robertlong13 (commented):
I've left a couple more minor comments.

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -44,8 +44,15 @@ const AP_Param::GroupInfo FlightAxis::var_info[] = {
     // @Bitmask: 1: Swap first 4 and last 4 servos (for quadplane testing)
     // @Bitmask: 2: Demix heli servos and send roll/pitch/collective/yaw
     // @Bitmask: 3: Don't print frame rate stats
+    // @bitmask: 3: Don't log Dt stats
```
robertlong13: I believe these are case sensitive. Also, bit 4.
```suggestion
    // @Bitmask: 4: Don't log Dt stats
```

robertlong13 (commented):
I've added a comment on the bitmask metadata, and you should reword the last commit, since that commit is more about parameter additions than the backward-time drive-by-fix.

After that, LGTM. Would be good to hear back from one more person running separate machines (I think Tridge does). All my testing has been WSL. I'm also going to run some tests with Cygwin.

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -366,61 +400,133 @@ void FlightAxis::exchange_data(const struct sitl_input &input)
...
+    }
+#if HAL_LOGGING_ENABLED
+    if (!(option_is_set(Option::NoDtLog))) {
+        uint64_t time_now = uint64_t(state.m_currentPhysicsTime_SEC * 1.0e6);
+// @LoggerMessage: RF
+// @Description: RealFlight mode messages
+// @Field: TimeUS: Time since system startup
+// @Field: Dt: delta time between this frame and the previous frae=me
+// @Field: Fps: frames-per-second implied by the current delta time
+        AP::logger().WriteStreaming("RF", "TimeUS,Dt,Fps,Glitch,SockErr", "QdIII", time_now, dt, uint32_t(roundf(1/dt)));
```
robertlong13: CI failing because Glitch and SockErr are still in the format.

robertlong13:
Shoot, this does break the Cygwin SITL somehow. It freezes up right after linking up with RealFlight. I'll need to see if I can get to the bottom of it. This is important because this is how people run SITL from Mission Planner.

image

Update:
Interesting, I left it running for an hour and it must have briefly unfrozen a couple times.
image

andyp1per:
> Shoot, this does break the Cygwin SITL somehow. It freezes up right after linking up with RealFlight. I'll need to see if I can get to the bottom of it. This is important because this is how people run SITL from Mission Planner.
> 
> image
> 
> Update: Interesting, I left it running for an hour and it must have briefly unfrozen a couple times. image

It has to be something to do with the socket handling IMO, be good to get to the bottom of this!

andyp1per:
Ok rebased on Peter's change that was responsible for the slowdown and now everything looks good

libraries/AP_InertialSensor/AP_InertialSensor_SITL.cpp
```diff
@@ -321,7 +321,17 @@ void AP_InertialSensor_SITL::timer_update(void)
     if (sitl == nullptr) {
         return;
     }
-    if (now >= next_accel_sample) {
+
+    // on some simulations (RealFlight) the aircraft sim decides when a new sample is available.
+    const bool running_in_lockstep = frame_num > 0;
```
tridge: what does being past the first frame have to do with being in lock-step scheduling?

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -187,7 +195,6 @@ void FlightAxis::parse_reply(const char *reply)
     }
 }
 
-
```
peterbarker, line 190: whitespace

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -205,14 +212,19 @@ bool FlightAxis::soap_request_start(const char *action, const char *fmt, ...)
     vasprintf(&req1, fmt, ap);
     va_end(ap);
 
+    // consumer/producer pattern
     while (sock == nullptr) {
-        sockcond1.wait_blocking();
-
+        socks_outsem.wait_blocking();
```
peterbarker:
```suggestion
        sock_outsem.wait_blocking();
```

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -387,60 +421,133 @@ void FlightAxis::exchange_data(const struct sitl_input &input)
 </ExchangeData>
 </soap:Body>
 </soap:Envelope>)",
-                           valid_channels,
-                           scaled_servos[0],
-                           scaled_servos[1],
```
peterbarker, line 392: random whitespace change

peterbarker (commented):
Consider breaking out socket semaphore renames out to a separate PR, which I'd just merge as being NFC.

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -205,14 +212,19 @@ bool FlightAxis::soap_request_start(const char *action, const char *fmt, ...)
...
         sock = socknext;
         socknext = nullptr;
-        sockcond2.signal();
+        socks_insem.signal();
     }
 
+    if (sock == nullptr) {
+        return false;
+    }
+    socks_insem.signal();
```
peterbarker: Why do we need these?

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -239,6 +251,7 @@ char *FlightAxis::soap_request_end(uint32_t timeout_ms)
     }
     sock->set_blocking(true);
     ssize_t ret = sock->recv(replybuf, sizeof(replybuf)-1, 1000);
+
```
peterbarker: whitespace

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -288,41 +301,62 @@ char *FlightAxis::soap_request_end(uint32_t timeout_ms)
...
 <soap:Body>
 <RestoreOriginalControllerDevice><a>1</a><b>2</b></RestoreOriginalControllerDevice>
 </soap:Body>
 </soap:Envelope>)");
-        soap_request_end(1000);
-        if(option_is_set(Option::ResetPosition)) {
-            soap_request_start("ResetAircraft", R"(<?xml version='1.0' encoding='UTF-8'?>
+    soap_request_end(1000UL);
+    if(option_is_set(Option::ResetPosition)) {
+        soap_request_start("ResetAircraft", R"(<?xml version='1.0' encoding='UTF-8'?>
```
peterbarker, line 361: whitespace-only change?

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -288,41 +301,62 @@ char *FlightAxis::soap_request_end(uint32_t timeout_ms)
...
 <soap:Body>
 <ResetAircraft><a>1</a><b>2</b></ResetAircraft>
 </soap:Body>
 </soap:Envelope>)");
-            soap_request_end(1000);
-        }
-        soap_request_start("InjectUAVControllerInterface", R"(<?xml version='1.0' encoding='UTF-8'?>
+        soap_request_end(1000UL);
+    }
+    soap_request_start("InjectUAVControllerInterface", R"(<?xml version='1.0' encoding='UTF-8'?>
```
peterbarker, line 369: If you want to make these whitespace changes then please create a no-compiler-output change.  I'll merge it.

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: locking strategy, consumer/producer - derived from the fast rate code which found many associated bugs

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: Removed

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: renamed

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: Its not. The if statement is gone so the indentation must change with it

libraries/AP_InertialSensor/AP_InertialSensor_SITL.cpp
andyp1per: I renamed the variable - its that the IMUs synced with the simulation rather than running freely in a separate thead

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: Removed

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: No. The if statement has gone so reformatting is neccessary

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: Its not random. The logic has changed (new functions, if statements added and removed) and the formatting must change with it.

andyp1per:
Now working on cygwin - turns out cygwin's usleep() is fundamentally broken

libraries/AP_InertialSensor/AP_InertialSensor_SITL.cpp
```diff
@@ -321,7 +321,17 @@ void AP_InertialSensor_SITL::timer_update(void)
     if (sitl == nullptr) {
         return;
     }
-    if (now >= next_accel_sample) {
+
+    // on some simulations (RealFlight) the aircraft sim decides when a new sample is available.
+    const bool sync_imus_to_frames = frame_num > 0;
+
+    if (sync_imus_to_frames && frame_num == sitl->state.frame_num) {
```
tridge: I don't really understand this logic, what are we trying to do here? why does state frame_num being the same as driver frame num signify anything? Could this be re-factored to make it clearer?

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -387,60 +448,133 @@ void FlightAxis::exchange_data(const struct sitl_input &input)
...
-        socket_frame_counter++;
         free(reply);
+        return true;
     }
+    return false;
 }
 
+bool FlightAxis::wait_for_sample(const struct sitl_input &input)
+{
+    uint32_t frame_count = 0;
```
tridge: unused variable?

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -387,60 +448,133 @@ void FlightAxis::exchange_data(const struct sitl_input &input)
...
+    uint32_t frame_tries = 0;
```
tridge: unused?

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -205,14 +240,19 @@ bool FlightAxis::soap_request_start(const char *action, const char *fmt, ...)
...
         sock = socknext;
         socknext = nullptr;
-        sockcond2.signal();
+        sock_insem.signal();
     }
 
+    if (sock == nullptr) {
+        return false;
+    }
+    sock_insem.signal();
```
tridge: signalling twice?

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -387,60 +448,133 @@ void FlightAxis::exchange_data(const struct sitl_input &input)
...
+        }
+        average_frame_time_s = average_frame_time_s * 0.98 + dt * 0.02;
+    }
+#if HAL_LOGGING_ENABLED
+    if (!(option_is_set(Option::NoDtLog))) {
+        uint64_t time_now = uint64_t(state.m_currentPhysicsTime_SEC * 1.0e6);
+// @LoggerMessage: RF
+// @Description: RealFlight mode messages
+// @Field: TimeUS: Time since system startup
+// @Field: Dt: delta time between this frame and the previous frae=me
```
tridge: typo in "frame"

libraries/SITL/SIM_FlightAxis.cpp
```diff
@@ -44,8 +44,15 @@ const AP_Param::GroupInfo FlightAxis::var_info[] = {
...
     // @Bitmask: 3: Don't print frame rate stats
+    // @Bitmask: 4: Don't log Dt stats
     // @User: Advanced
     AP_GROUPINFO("OPTS", 1, FlightAxis, _options, uint32_t(Option::ResetPosition)),
+
+    // @Param: SAMPLEHZ
+    // @DisplayName: FlightAxis IMU synthetic sample rate
+    // @Description: FlightAxis IMU synthetic sample rate
+    // @User: Advanced
+    AP_GROUPINFO("SAMPLEHZ", 2, FlightAxis, _samplehz, 5000),
```
tridge: maybe default to 2000? We want this to be workable for people with slower computers (eg. laptop)

libraries/SITL/SIM_FlightAxis.h
```diff
@@ -29,6 +29,8 @@
 
 namespace SITL {
 
+#define SAMPLES_PER_FRAME 10
```
tridge: unused?

tridge (changes requested):

libraries/AP_InertialSensor/AP_InertialSensor_SITL.cpp
tridge: maybe add a bool to sitl->state called synchronised_IMU ?
set that to true in SIM_FlightAxis, check it here

andyp1per:
copilot:summary

libraries/AP_InertialSensor/AP_InertialSensor_SITL.cpp
andyp1per: changed the naming. still want the count as otherwise you have to keep setting and unsetting a flag on  the SITL side.

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: removed

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: removed

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: removed

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: fixed

libraries/SITL/SIM_FlightAxis.cpp
andyp1per: ok fixed, I retested and it seems ok with this

libraries/SITL/SIM_FlightAxis.h
andyp1per: removed

libraries/AP_InertialSensor/AP_InertialSensor_SITL.cpp
andyp1per: still needs to be a frame check to see if time has moved on

tridge (approved):

PR #22593, opened 2023-01-05
AP_Scripting:Add mission to Script_Controller and improve

Hwurzburg:
replaces #22582 

-adds optional mission file load on scripts switching..can also be used as a mission switcher by itself, if desired
-adds ability to select which aux function (301-307) is used for switching
-no RC channel need be set,ie use GCS only
-if no RC channel set, nothing occurs on restart or boot....script can be active always if desired with no impact if GCS switching is not used or subdirectories not setup
-minimally verbose

@timtuxworth you might be interested in this

PR #29925, opened 2025-04-29
AP_Mission: Include fence and rally point in stored_in_location

rishabsingh3003:
Old way of sending fence as MISSION_ITEM and not MISSION_ITEM_INT seems to be broken. This seems to fix it.

This PR orignally broke it: https://github.com/ArduPilot/ardupilot/pull/22593 I think

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -2038,6 +2038,72 @@ class AutoTestCopter(vehicle_test_suite.TestSuite):
             "Fence test failed to reach home (%fm distance) - "
             "timed out after %u seconds" % (home_distance, timeout,))
 
+    def FenceUpload_MissionItem(self, timeout=180):
+        '''Test MISSION_ITEM fence upload/download'''
+        self.set_parameters({
+            "FENCE_ENABLE": 1,
+            "FENCE_TYPE": 6,  # polygon and circle fences
+        })
+
+        self.poll_home_position(quiet=False)
+
+        home_loc = self.mav.location()
+
+        fence_loc = [
+            self.offset_location_ne(home_loc, -110, -110),
+            self.offset_location_ne(home_loc, 110, -110),
+            self.offset_location_ne(home_loc, 110, 110),
+            self.offset_location_ne(home_loc, -110, 110),
+        ]
+
+        seq = 0
+        items = []
+        mission_type = mavutil.mavlink.MAV_MISSION_TYPE_FENCE
+        count = len(fence_loc)
+        for loc in fence_loc:
+            item = self.mav.mav.mission_item_encode(
+                1,
+                1,
+                seq,
+                mavutil.mavlink.MAV_FRAME_GLOBAL,
+                mavutil.mavlink.MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION,
+                0, 0,
+                count, 0, 0, 0,
+                loc.lat, loc.lng, 33.0,
+                mission_type
+            )
+            items.append(item)
+            seq += 1
+
+        self.upload_using_mission_protocol(mission_type, items)
+        downloaded_items = self.download_using_mission_protocol(mission_type)
+
+        if len(downloaded_items) != len(items):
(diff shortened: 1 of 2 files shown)


peterbarker:
Sorry, which "old way" are you referring to?  There are lots of "old ways" to transfer items :-)

We shouldn't actually be receiving anything via mission_item any more, *everything* should be coming in via `MISSION_ITEM_INT`.  We are *removing* support for `MISSION_ITEM`...

Could you provide steps to reproduce, please?

rmackay9:
As a side note, I'd like us to add instructions to this wiki page on how to upload/download missions to the vehicle 
https://ardupilot.org/dev/docs/mavlink-mission-upload-download.html

peterbarker:
> As a side note, I'd like us to add instructions to this wiki page on how to upload/download missions to the vehicle https://ardupilot.org/dev/docs/mavlink-mission-upload-download.html
> 
> It's been on my list forever but I actually don't know what messages to use and the flow of them

https://mavlink.io/en/services/mission.html

Omits the fact that they really should go via FTP in ArduPilot now-adays :-)

rmackay9:
@peterbarker,

Yes, we have another page here that needs filling out on how to use MAVFTP https://ardupilot.org/dev/docs/mavlink-mavftp.html

Anyway, not trying to pressure anyone into this.  Helping community developers helps the project.  Software that only the internal dev team understands is wasteful in some ways.

rishabsingh3003:
@peterbarker I have added an autotest that fails without the "fix" in this PR. 
I understand that we shoudn't be using MISSION_ITEM anymore, however, there are GCS's out there still using MISSION_ITEM and fence upload isn't working on them.

muramura:
ArduPilot must respond with a MISSION_ITEM_INT message when it receives a MISSION_REQUEST.
It is required to satisfy this behavior.

https://mavlink.io/en/services/mission.html#command_message_type
Deprecated Types: MISSION_ITEM
The legacy version of the protocol also supported [MISSION_REQUEST](https://mavlink.io/en/messages/common.html#MISSION_REQUEST) for requesting that a mission be sent as a sequence of [MISSION_ITEM](https://mavlink.io/en/messages/common.html#MISSION_ITEM) messages.

Both MISSION_REQUEST and MISSION_ITEM messages are now deprecated, and should no longer be sent. If MISSION_REQUEST is recieved the system should instead respond with [MISSION_ITEM_INT](https://mavlink.io/en/services/mission.html#MISSION_ITEM_INT) items (as though it received [MISSION_REQUEST_INT](https://mavlink.io/en/services/mission.html#MISSION_REQUEST_INT)).

peterbarker:
@rishabsingh3003 as @muramura points out, `MISSION_ITEM` definitely should no longer be used.

We have a timeline for removal:
```
// GCS should be using MISSION_REQUEST_INT instead; this is a waste of
// flash.  MISSION_REQUEST was deprecated in June 2020.  We started
// sending warnings to the GCS in Sep 2022 if MISSION_REQUEST was used.
// Copter 4.4.0 sends this warning.
// CODE_REMOVAL
// ArduPilot 4.4 sends warnings if MISSION_ITEM used
// ArduPilot 4.8 stops compiling in MISSION_ITEM but still sends warnings
// ArduPilot 4.9 removes the code but sends message about MISSION_ITEM not supported
// ArduPilot 4.10 stops sending the warning
```

I would encourage the affected user to move ahead their GCS sooner rather than later.  It's worth pointing out to them that we didn't move to `MISSION_ITEM_INT` just for fun - the precision offered by the waypoints in the newer message make it a very good idea to use it.

However, this was definitely a regression, so this is definitely a merge-it thing.

rishabsingh3003:
Thanks Peter. Also, the regression is present in 4.5 as well

peterbarker (approved):

rmackay9:
I've included the flight code change in 4.6.2-beta1 but could not include the autotest change due to a merge conflict

rmackay9:
The autotest has now been included in 4.6.2-beta1 as well thanks to PR https://github.com/ArduPilot/ardupilot/pull/30465

PR #29926, opened 2025-04-29
Plane: fix Q_OPTION bit 18 docstring

magate:
This docstring is not correct.

https://github.com/ArduPilot/ardupilot/blob/8e9a568a55be2d4129e0b06f460c82bd7dec7953/ArduPlane/mode_auto.cpp#L152-L162

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -283,7 +283,7 @@ const AP_Param::GroupInfo QuadPlane::var_info[] = {
     // @Bitmask: 15: ThrLandControl-enable throttle stick control of landing rate
     // @Bitmask: 16: DisableApproach-disable use of approach and airbrake stages in VTOL landing
     // @Bitmask: 17: EnableLandResposition-enable pilot controlled repositioning in AUTO land.Descent will pause while repositioning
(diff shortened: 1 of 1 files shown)


ArduPlane/quadplane.cpp
```diff
@@ -283,7 +283,7 @@ const AP_Param::GroupInfo QuadPlane::var_info[] = {
     // @Bitmask: 15: ThrLandControl-enable throttle stick control of landing rate
     // @Bitmask: 16: DisableApproach-disable use of approach and airbrake stages in VTOL landing
     // @Bitmask: 17: EnableLandResposition-enable pilot controlled repositioning in AUTO land.Descent will pause while repositioning
-    // @Bitmask: 18: ARMVTOL-arm only in VTOL or AUTO modes
+    // @Bitmask: 18: ARMVTOL-arm only in VTOL or AUTO modes when current nav cmd is VTOL Takeoff
```
peterbarker, line 286:
```suggestion
    // @Bitmask: 18: ARMVTOL-arm only in VTOL modes (or AUTO mode when current nav cmd is VTOL Takeoff)
```

peterbarker (approved):

Issue #7817, opened 2018-03-01
Plane: An error in TECS::_update_throttle()


luweiagi:
Hi, in void AP_TECS::_update_throttle_with_airspeed(void)
https://github.com/ArduPilot/ardupilot/blob/ada7d700c2ed470700350e3f6e022ab8866af999/libraries/AP_TECS/AP_TECS.cpp#L630
, there is an error:
Calculation of cosPhi may be not right.
![1](https://wx4.sinaimg.cn/mw690/006f01Rgly1fox66j5cl3j30mm02sacs.jpg)
So based
~~~
float cosPhi = sqrtf((rotMat.a.y*rotMat.a.y) + (rotMat.b.y*rotMat.b.y));
~~~
We can get(note that rotMat is ![5](https://wx1.sinaimg.cn/mw690/006f01Rgly1fox6bliuh5j300r00o0fp.jpg))
![4](https://wx3.sinaimg.cn/mw690/006f01Rgly1fox69v4nxlj30it014jr9.jpg)
, and then we get
![2](https://wx4.sinaimg.cn/mw690/006f01Rgly1fox66j1vnvj30al03ejra.jpg)

peterbarker:
@luweiagi please check which ordering of that rotation matrix you are expecting.

A unit test seems to show that these calculations are correct:
```
diff --git a/libraries/AP_Math/tests/test_matrix3.cpp b/libraries/AP_Math/tests/test_matrix3.cpp
index 099368e68d1..e89fafb984d 100644
--- a/libraries/AP_Math/tests/test_matrix3.cpp
+++ b/libraries/AP_Math/tests/test_matrix3.cpp
@@ -103,6 +103,19 @@ TEST_P(Matrix3fTest, Inverses)
     }
 }
 
+// not just testing eulers here, but also calculation of Phi in TECS
+// from rotation matrix:
+TEST_P(Matrix3fTest, PhiTest)
+{
+    Matrix3f rotMat;
+    rotMat.from_euler312(radians(17.3), 0, 0);  // r, p, y
+
+    float cosPhi = sqrtf((rotMat.a.y*rotMat.a.y) + (rotMat.b.y*rotMat.b.y));
+
+    ASSERT_EQ(cosPhi, cosf(radians(17.3)));
+}
+
+
 INSTANTIATE_TEST_CASE_P(InvertibleMatrices,
                         Matrix3fTest,
                         ::testing::ValuesIn(invertible));
```

```
[...]

PR #29933, opened 2025-04-30
AP_TECS: avoid use of cosphi_squared dcm matrix trick

peterbarker:
Issue https://github.com/ArduPilot/ardupilot/issues/7817 pointed out that the order in which we are retrieving the euler angles from the transformation matrix is incorrect.

Testing shows that there is a discrepancy between these values and the values retrieved directly from the AHRS.

Since we are simply attempting to retrieve the roll angle, extract it directly from the AHRS.

Closes https://github.com/ArduPilot/ardupilot/issues/7817

Ping @luweiagi


peterbarker:
Flew this at SV, all good.

tridge (approved):

PR #29935, opened 2025-04-30
Copter: Change to a typed definition

muramura:
In C++, you can make typed declarations without allocating memory.

### DIFF
--- a/ArduCopter/radio.cpp
+++ b/ArduCopter/radio.cpp
@@ -123,7 +123,7 @@ void Copter::read_radio()
     set_failsafe_radio(true);
 }
 
-#define FS_COUNTER 3        // radio failsafe kicks in after 3 consecutive throttle values below failsafe_throttle_value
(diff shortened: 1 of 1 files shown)


peterbarker (approved):
type safety is good

This is a no-compiler-output change

rmackay9:
TBH, I'm not sure about this.  Does this means we're going to go through all drivers and change definitions to constexpr?  Just worried about massive code churn

peterbarker:
> TBH, I'm not sure about this. Does this means we're going to go through all drivers and change definitions to constexpr? Just worried about massive code churn

There's nothing stopping that from happening.

It is an improvement in code quality, but it *is* code churn.

There are also reasons to not merge this; if we wanted to make these configurable in a hwdef we'd what to do the `ifndef` thing, which fits in with a define - we'd probably reintroduce one ending in `_DEFAULT` or somethng.

So @rmackay9 's not keen, I'll close this one in a little while, I think.

peterbarker:
Sorry @muramura , seems we're not going with type-correctness here for code-churn reasons and the fact we may want ifdefs later.

PR #29937, opened 2025-05-01
autotest: allow EKF3 more time to learn wind in DeadreckoningNoAirspeed

peterbarker:
... also slip in a few new convenience methods.

If you give the estimators a really long time (by pushing that number out to 6000) weird things start to happen:

image

### DIFF
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -2029,7 +2029,7 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
             self.progress("Orbit with GPS and learn wind")
             # allow longer to learn wind if there is no airspeed sensor
             if disable_airspeed_sensor:
-                self.delay_sim_time(60)
+                self.delay_sim_time(180)
             else:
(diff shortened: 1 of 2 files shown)


tpwrules:
Any idea why this changed?

peterbarker:
> Any idea why this changed?

I have a guess.

The test has been flapping for a while, it was *right* on the edge.

A recent patch of mine moved us from using AP::baro altitude (height above home) for our simulated wind to using simulated height-above-terrain when calculating simulated wind.  That appears to have moved the flapping test from "bad" to "worse".

That being said, our EKF3 wind estimation is essentially broken AFAICS - at least in SITL.  DCM nails it, EKF3 gets about half what it should, which is why the DeadreckoningNoAirspeed test doesn't succeed unless you've got ridiculously high margins.

I did spend some time yesterday bisecting the problem, but wasn't able to come up with any conclusions.  I *think* there may actually be two step changes somewhere in our history.  Python3 problems make the bisection annoying.

PR #29940, opened 2025-05-01
waf: ensure if you're using clang then you're using clang++ and vice-versa

peterbarker:
Weird things happen if you specify clang++ for your CXX compiler but leave CC as GCC.


khancyr (approved):
LGTM

PR #29942, opened 2025-05-01
AP_InertialSensor: remove unused rotation member

peterbarker:
It's private and doesn't appear in the C++


IamPete1 (approved):
If it builds it must be good.

Issue #29679, opened 2025-04-03
AP_Logger: Add MAVLink SYSTEM_TIME like information to Data Flash Log


horiuchi0photo:
## Feature request

Currently there is no information about the UNIX time the vehicle is flying, so the GCS or other post-processing tools must infer it from the GPS messages in the log.

I want messages like SYSTEM_TIME to be written to the log when logging is started and when the time base is changed.

**Platform**
[X] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

peterbarker:
I've created https://github.com/ArduPilot/ardupilot/pull/29948 which would close this issue AFAICS.

Are you able to test, @horiuchi0photo ?

Note this may end up in a `SYS` log message or similar as we don't have huge numbers of IDs now-adays.

horiuchi0photo:
Thank you, this is what I wanted.

I tried the patch, when the time was set during log output, the output was Epoch=0 (it seems the output timing was a bit early).

I think it's fine to add it to another log message.

PR #29948, opened 2025-05-01
Add logging for Unix Epoch time

peterbarker:
Closes https://github.com/ardupilot/ardupilot/issues/29679


libraries/AP_RTC/AP_RTC.cpp
```diff
@@ -84,12 +79,14 @@ void AP_RTC::set_utc_usec(uint64_t time_utc_usec, source_type type)
...
     }
-    WITH_SEMAPHORE(rsem);
 
-#if DEBUG_RTC_SHIFT
-    uint64_t old_utc = 0;
-    UNUSED_RESULT(get_utc_usec(old_utc));
-#endif
+#if AP_RTC_LOGGING_ENABLED
+    if (rtc_source_type != type) {
+        AP::logger().Write_RTC();
```
Williangalvani, line 85: If we got two SYSTEM_TIME messages (with different times) on the same flight, would this log both?
Not that I expect this to happen, but in my head it makes more sense to log whenever rtc_shift changes, instead of per source type

peterbarker: Fair point, I've changed it to log if either the type or shift changes.

Williangalvani (approved):
Looks good to me and worked as expected on my systems

Williangalvani:
can we get this in for 4.7?

IamPete1:
Would it be worth logging the source?

peterbarker:
> Would it be worth logging the source?

Added:
```
2026-01-26 16:43:09.84: RTC {TimeUS : 8904770, Epoch : 0, SourceType : 3}
2026-01-26 16:43:10.20: RTC {TimeUS : 9264626, Epoch : 0, SourceType : 3}
2026-01-26 16:43:30.38: RTC {TimeUS : 29444884, Epoch : 1769406210416107, SourceType : 0}
```

(log message documentation looks correct)

PR #29953, opened 2025-05-02
AP_Scripting: added EFI_DLA64 driver

tridge:
new protocol for updated EFI engine from DLA


Gromitvt10:
The driver is working on my DLA 64 EFI engine.  I can see EFI RPM, both cylinder head temps and fuel consumed. In Mission Planner "EFI Head Temp" is the first cylinder temperature. "EFI Exhaust Temp" 2nd cylinder temperature. Fuel Consumed is measured in milliliters, Fuel Flow is measured in milliliters per min.

tpwrules:
Why did the temperature unit change? That is a problem in some other scripts too apparently: #30608

tridge:
> Why did the temperature unit change?

yep, we screwed up, fixed in 4.6. We should put the same fix in the other drivers, though hopefully most will be moving to 4.6.x

tpwrules:
Why did a switch happen in the first place? What do you mean by "screwed up" here?

tridge:
> Why did a switch happen in the first place? What do you mean by "screwed up" here?

the data structure is used to populate the mavlink. The mavlink msg is in degrees C. Were were sending kelvin.

PR #29955, opened 2025-05-02
waf: allow --embed-XXX on configure

tridge:
allows embedding of any lua driver or applet in the tree into ROMFS

### DIFF
--- a/Tools/ardupilotwaf/boards.py
+++ b/Tools/ardupilotwaf/boards.py
@@ -5,9 +5,10 @@ import re
 import sys, os
 import fnmatch
 import platform
+import glob
 
 import waflib
-from waflib import Utils
+from waflib import Utils, Context
 from waflib.Configure import conf
 import json
 _board_classes = {}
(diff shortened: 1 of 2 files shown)


tpwrules:
Is it possible to specify a path? I think we could do that, then have it automatically search in those dirs and add .lua so you just have to do `--embed-file repl` or so and we don't have to present 20 additional options. I guess we already do for build options so maybe it's not that bad of a problem.

There are also some drivers that require ancillary modules.

tridge:
we could look for require() calls in the scripts to auto-add the modules, or we could start adding some '-- @Description: XXX' and '-- @Requires:' lines to the scripts as meta-data ?

shiv-tyagi:
Great work @tridge!

> we could look for require() calls in the scripts to auto-add the modules, or we could start adding something like
> -- @Description: XXX
>  -- @Requires: modules/mavport.lua

This sounds good. Should we also add a _Category_ (or _Group_) field to this metadata? To be able to organise these options better on the custom build server?

shiv-tyagi:
> we could look for require() calls in the scripts to auto-add the modules

I would prefer that we don't do a lot of scanning at build-time. It involves reading all those files when someone asks for the list of options available for a commit on custom build server and adds to the response time. I think it is better to explicitly specify what a driver is dependent on (in a single file so that we can construct the dependency tree quickly).

We can have something similar to the build_options.py to list the drivers/applets available to be embedded to a particular commit (_version_). That file would contain information like the label, description, path?, and dependencies for a driver/applet.

We can have automation scripts to do the scanning based on something which you described above (like looking for the require() calls) and adding entries to that file to make the life of people easy. But all of that should be done while submitting PRs to add those drivers in the tree.

peterbarker (approved):

PR #29956, opened 2025-05-02
AP_Logger: correct extraction of DMS message metadata

peterbarker:
... the regex is pretty tight!


rmackay9:
I have not included this in 4.6.2-beta1 due to a merge conflict.  If someone could create a backport PR against the ArduPilot-4.6 branch that would be a great help!

PR #29959, opened 2025-05-02
Add Sub tests for GCS and RC

Hwurzburg:
Very simple CI tests, but better than nothing

### DIFF
--- a/Tools/autotest/ardusub.py
+++ b/Tools/autotest/ardusub.py
@@ -237,6 +237,40 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
                     "Distance not maintained: want %.2f (+/- %.2f) got=%.2f" %
                     (previous_distance, delta, m.distance))
 
+    def GCSFailsafe(self):
+        '''Test GCSFailsafe'''
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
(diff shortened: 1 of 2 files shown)


tridge (approved):

Tools/autotest/ardusub.py
```diff
@@ -237,6 +237,40 @@ def watch_distance_maintained(self, delta=0.3, timeout=5.0):
...
+        self.wait_statustext("GCS Failsafe Cleared", timeout=60)
+        self.progress("GSC Failsafe OK")
+        self.disarm_vehicle()
+        self.context_pop()
+
+    # Tests actions and logic behind the radio failsafe
+    def ThrottleFailsafe(self):
+        '''Test RC and RC Failsafe'''
+        # disable GCS and enable RC
+        self.context_push()
```
peterbarker, line 259: Don't need this (you get one for free at the start of the test.

Tools/autotest/ardusub.py
```diff
@@ -237,6 +237,40 @@ def watch_distance_maintained(self, delta=0.3, timeout=5.0):
...
+        self.set_parameter('FS_GCS_ENABLE', 0)
+        self.set_parameter('RC_PROTOCOLS', 1)
```
peterbarker, line 261: can set_parameters which is marginally faster

PR #29960, opened 2025-05-03
CAN: fixed race in SITL and Linux HALs

tridge:
With multiple threads sending CAN messages we can get a race with the send mutex. This issue was found in SITL with quadplane-can frame.


Huibean:
@tridge  I am trying to make SITL sending > 50hz uavcan.equipment.actuator.ArrayCommand in mcast:0, in master only can get 15hz when CAN_D1_UC_SRV_RT is 50, with this PR improved, but still only getting 28hz max, I have confirmed the network is health by ```
./examples/servo_control.py  mcast:0 --rate 100
``` under pydronecan

peterbarker (approved):
LGTM

rmackay9:

PR #29961, opened 2025-05-03
AP_AdvancedFailsafe: option to not go back to the loss comm mission item if we are already in the return path

eppravitra:
Currently, if there is a loss comm., the aircraft will always fly to the mission item specified with AFS_WP_COMMS.

I added an option so that if we have already passed AFS_WP_COMMS and got a loss. We should not go back to AFS_WP_COMMS. 

This is useful, for example, if the aircraft is already in the landing sequence. If there is a loss during a landing attempt, it should just land instead of starting the landing sequence again. 

I tested the change with Gazebo SITL.

### DIFF
--- a/libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
+++ b/libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
@@ -168,6 +168,7 @@ const AP_Param::GroupInfo AP_AdvancedFailsafe::var_info[] = {
     // @Description: See description for each bitmask bit description
     // @Bitmask: 0: Continue the mission even after comms are recovered (does not go to the mission item at the time comms were lost)
     // @Bitmask: 1: Enable AFS for all autonomous modes (not just AUTO) 
+    // @Bitmask: 2: Option to not jump to AFS_WP_COMMS if already in the return path 
     AP_GROUPINFO("OPTIONS", 21, AP_AdvancedFailsafe, options, 0),
 
     // @Param: GCS_TIMEOUT
@@ -251,13 +252,16 @@ AP_AdvancedFailsafe::check(uint32_t last_valid_rc_ms)
         if (!gcs_link_ok) {
             GCS_SEND_TEXT(MAV_SEVERITY_DEBUG, "AFS State: DATA_LINK_LOSS");
             _state = STATE_DATA_LINK_LOSS;
-            if (_wp_comms_hold) {
+            
+            if (should_use_comms_hold()) {
                 _saved_wp = mission.get_current_nav_cmd().index;
                 mission.set_current_cmd(_wp_comms_hold);
+
                 if (mode == AFS_AUTO && option_is_set(Option::GCS_FS_ALL_AUTONOMOUS_MODES)) {
                     set_mode_auto();
                 }
             }
+          
             // if two events happen within 30s we consider it to be part of the same event
             if (now - _last_comms_loss_ms > 30*1000UL) {
                 _comms_loss_count++;
@@ -414,6 +418,27 @@ AP_AdvancedFailsafe::check_altlimit(void)
     return false;
 }
 
+/*
+  return true if we should jump to _wp_comms_hold
+ */
+bool AP_AdvancedFailsafe::should_use_comms_hold(void){
+    
+    AP_Mission *_mission = AP::mission();
+
+    if (_mission == nullptr) {
+        return false;
+    }
+
+    if (_wp_comms_hold <= 0) {
+        return false;
+    }
+
+    if ((_mission->state() == AP_Mission::MISSION_RUNNING) && _mission->get_in_return_path_flag() && option_is_set(Option::CONTINUE_IF_ALREADY_IN_RETURN_PATH)) {
+        return false;
+    }
+
+    return true;
+}
 /*
   return true if we should crash the vehicle
  */
--- a/libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.h
+++ b/libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.h
@@ -82,6 +82,9 @@ public:
     // during sensor calibration
     void heartbeat(void);
 
+    //return true if we should jump to _wp_comms_hold
+    bool should_use_comms_hold(void);
+
     // return true if we are terminating (deliberately crashing the vehicle)
     bool should_crash_vehicle(void);
 
@@ -138,7 +141,7 @@ protected:
     bool _heartbeat_pin_value;
 
     // saved waypoint for resuming mission
-    uint8_t _saved_wp;
+    uint16_t _saved_wp;
     
     // number of times we've lost GPS
     uint8_t _gps_loss_count;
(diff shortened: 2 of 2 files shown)


libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
```diff
@@ -253,7 +253,16 @@ AP_AdvancedFailsafe::check(uint32_t last_valid_rc_ms)
             _state = STATE_DATA_LINK_LOSS;
             if (_wp_comms_hold) {
                 _saved_wp = mission.get_current_nav_cmd().index;
-                mission.set_current_cmd(_wp_comms_hold);
+
+                //option to not go back to the loss comm mission item if we already passed it
+                if (option_is_set(Option::CONTINUE_IF_PASSED_LOSS_COMM_WP) && 
+                   (mission.state() == AP_Mission::MISSION_RUNNING) &&
+                   (mission.get_current_nav_cmd().index > _wp_comms_hold)){
```
peterbarker, line 260:
```suggestion
                   (mission.get_in_landing_sequence())){
```

... as discussed elsewhere, checking for a numeric difference in waypoint numbers doesn't make sense when there are jumps involved.  I believe you said you didn't want to do the set-current-cmd when you were already on the way home, so just check directly.

peterbarker (changes requested):

eppravitra:
Thanks Peter. I have been super busy. I'll look into this.

eppravitra:
Hi @peterbarker , 

Please see the changes. I have a very specific case where I still require a direct comparison of mission item numbers. However, I have implemented what you suggested. There are now two options that can handle both cases.

Am I right that I should check for the return path instead of the landing sequence?

libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
```diff
@@ -253,7 +253,16 @@ AP_AdvancedFailsafe::check(uint32_t last_valid_rc_ms)
             _state = STATE_DATA_LINK_LOSS;
             if (_wp_comms_hold) {
                 _saved_wp = mission.get_current_nav_cmd().index;
```
peterbarker, line 255: This probably wants to be within your `else` clause.

libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
```diff
@@ -253,7 +253,16 @@ AP_AdvancedFailsafe::check(uint32_t last_valid_rc_ms)
...
-                mission.set_current_cmd(_wp_comms_hold);
+
+                //option to not go back to the loss comm mission item if we are already in return path or passed loss comm mission item
+                if ((mission.state() == AP_Mission::MISSION_RUNNING) &&
+                    ((option_is_set(Option::CONTINUE_IF_ALREADY_IN_RETURN_PATH) && mission.get_in_return_path_flag()) ||
+                     (option_is_set(Option::CONTINUE_IF_PASSED_LOSS_COMM_WP)    && (mission.get_current_nav_cmd().index > _wp_comms_hold)))){
```
peterbarker, line 260:
```suggestion
                if (should_use_comms_hold()) {
```

and

```
    bool AP_AdvancedFailsafe::should_use_comms_hold() const
    {
        if (mission.state() != AP_Mission::MISSION_RUNNING) {
            return false;
        }
        if (mission.get_in_return_path_flag() && option_is_set(Option::CONTINUE_IF_ALREADY_IN_RETURN_PATH))
            return false;
        }
        return true;
```

Please create a separate commit for adding this logic into the method using the same style:
```
                     (option_is_set(Option::CONTINUE_IF_PASSED_LOSS_COMM_WP)    && (mission.get_current_nav_cmd().index > _wp_comms_hold)))){
```

I'm not sure *that* patch will make it in easily, whereas I'm pretty confident of the return-path-start stuff.

libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
```diff
@@ -253,7 +253,16 @@ AP_AdvancedFailsafe::check(uint32_t last_valid_rc_ms)
...
+                    
+                }else {//nominal case, set mission item to _wp_comms_hold
+                    mission.set_current_cmd(_wp_comms_hold);
+                }
+               
                 if (mode == AFS_AUTO && option_is_set(Option::GCS_FS_ALL_AUTONOMOUS_MODES)) {
```
peterbarker, line 266: Should this also be within the `else?`

eppravitra:
@peterbarker ,
I did not follow everything exactly as you suggested, because it could lead to (in some cases) AFS not triggering in GUIDED

peterbarker:
This is looking pretty good.

The lack of an autotest for this is bad.

What testing has this had?  How have you tested it?

eppravitra:
> What testing has this had? How have you tested it?

I tested the change with  SITL Gazebo

libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
```diff
@@ -251,13 +252,16 @@ AP_AdvancedFailsafe::check(uint32_t last_valid_rc_ms)
         if (!gcs_link_ok) {
             GCS_SEND_TEXT(MAV_SEVERITY_DEBUG, "AFS State: DATA_LINK_LOSS");
             _state = STATE_DATA_LINK_LOSS;
-            if (_wp_comms_hold) {
-                _saved_wp = mission.get_current_nav_cmd().index;
+            _saved_wp = mission.get_current_nav_cmd().index;
```
IamPete1: `_saved_wp` should be a uint16.

IamPete1: Why move this our of the if?

libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
```diff
@@ -251,13 +252,16 @@ AP_AdvancedFailsafe::check(uint32_t last_valid_rc_ms)
...
+
+            if (should_use_comms_hold()){
```
IamPete1:
```suggestion
            if (should_use_comms_hold()) {
```

libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
```diff
@@ -414,6 +418,27 @@ AP_AdvancedFailsafe::check_altlimit(void)
...
+
+    if (_mission == nullptr) {
+        return false;
+    }
+
+    if (_wp_comms_hold <= 0) {
+        return false;
+    }
+
+    if ((_mission->state() == AP_Mission::MISSION_RUNNING) && _mission->get_in_return_path_flag() && option_is_set(Option::CONTINUE_IF_ALREADY_IN_RETURN_PATH)){
```
IamPete1:
```suggestion
    if ((_mission->state() == AP_Mission::MISSION_RUNNING) && _mission->get_in_return_path_flag() && option_is_set(Option::CONTINUE_IF_ALREADY_IN_RETURN_PATH)) {
```

libraries/AP_AdvancedFailsafe/AP_AdvancedFailsafe.cpp
eppravitra: > Why move this our of the if?

I moved it inside the if as requested.

I can't think of a case where this would matter. Either way should work.

eppravitra:
Hi @IamPete1,

All your suggestions have been included.

IamPete1 (approved):
LGTM

PR #29963, opened 2025-05-03
ArduSub: add EMERGENCY surface mode to Sub

Hwurzburg:
allows recovery without position or depth sensor

considered making it part of SURFACE, but Sub mode base module has so many checks for position and depth sensor before even switching modes to SURFACE that making an independent mode was cleaner, and easier to understand and clearly does not affect other modes or operation. Also allows easy hiding from user as normal mode.

This has NOT been added to the MAVLink available modes message, and we do not intend to add as XML or to pymavlink as a user mode...this is for failsafe use primarily....SITL and MAVLink users can still command it by mode number if desired...

### DIFF
--- a/ArduSub/mode_emergency.cpp
+++ b/ArduSub/mode_emergency.cpp
@@ -1,0 +1,55 @@
+#include "Sub.h"
+
+
+bool ModeEmergency::init(bool ignore_checks)
+{
+    has_surfaced = false;
+    // initialize vertical speeds and acceleration
+    position_control->set_max_speed_accel_U_cm(-sub.get_pilot_speed_dn(), g.pilot_speed_up, g.pilot_accel_z);
+    position_control->set_correction_speed_accel_U_cmss(-sub.get_pilot_speed_dn(), g.pilot_speed_up, g.pilot_accel_z);
+
+    // initialise position and desired velocity
+    position_control->init_U_controller();
+    GCS_SEND_TEXT(MAV_SEVERITY_CRITICAL,"Emergencey surface!");
+    return true;
+    
+    has_surfaced = false;
+
+}
+
+void ModeEmergency::run()
+{
+    // if not armed set throttle to zero and exit immediately
+    if (!motors.armed()) {
+        motors.output_min();
+        motors.set_desired_spool_state(AP_Motors::DesiredSpoolState::GROUND_IDLE);
+        attitude_control->set_throttle_out(.5,true,g.throttle_filt);
+        attitude_control->relax_attitude_controllers();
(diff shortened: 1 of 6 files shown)


ArduSub/mode_emergency.cpp
```diff
@@ -0,0 +1,55 @@
...
+        return;
+    }
+
+    // Already at surface, hold depth at surface
+    if (sub.ap.at_surface || has_surfaced) {
+        if (!has_surfaced) {
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING,"At Surface");
+            has_surfaced = true;
+        }
+        motors.set_throttle(.55);  //hold at surface
```
IamPete1, line 38: I don't know anything about sub, but hard coding throttle values seems a bit scary.

ArduSub/mode_emergency.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    // set target climb rate
+    float cmb_rate_cms = constrain_float(fabsf(sub.wp_nav.get_default_speed_up_cms()), 1, position_control->get_max_speed_up_cms());
+
+    // update altitude target and call position controller
+    position_control->set_pos_target_U_from_climb_rate_cm(cmb_rate_cms);
+    position_control->update_U_controller();
+
+    // maximum climb to surface
+    motors.set_throttle(1.1); //over 1 to clear current throttle limit on surface detection
```
IamPete1, line 54: Why set target climb rates if your going to override the throttle?

ArduSub/mode_emergency.cpp
```diff
@@ -0,0 +1,55 @@
...
+        attitude_control->relax_attitude_controllers();
+        position_control->init_U_controller();
+        return;
+    }
+
+    // Already at surface, hold depth at surface
+    if (sub.ap.at_surface || has_surfaced) {
+        if (!has_surfaced) {
+            GCS_SEND_TEXT(MAV_SEVERITY_WARNING,"At Surface");
+            has_surfaced = true;
```
peterbarker, line 36: This split-state thing is odd.  Perhaps we need a "was_at_surface" thing instead of `has_surfaced`?

peterbarker (changes requested):
Maybe make surface work without baro instead?

rmackay9:
We discussed on the dev call and @Williangalvani said he'll take a look and try to combine with the existing SURFACE mode.

We also decided that we should support subs with no barometer.

Let's add a parameter for the throttle in case of no barometer

Hwurzburg:
closing since Willian has changed SURFACE mode in a PR

PR #29966, opened 2025-05-04
Rover: Change the if statement to a switch statement

muramura:
The processing was changed based on differences in the value of the same variable.
I replaced the if statements with a switch statement.


peterbarker:
This is kind of terrible as on ArduPilot value 1 is supposed to be motor/compass interference calibration.  It's pre-existing, 'though.

rmackay9:
merged, thanks!

PR #29971, opened 2025-05-04
Copter: Handle the case where the arming delay time is set to 0 seconds

muramura:
The arming delay time can be set to 0 seconds.
I handle the case of 0 seconds using the preprocessor.

Screenshot from 2025-05-04 21-52-07


rmackay9:
Thanks for the PR but I think this makes the code harder to understand so let's not do this.

PR #29974, opened 2025-05-05
Global: Clarify the floating point value

muramura:
I don't think there is any benefit to omitting the integer part in a floating-point value.
By adding a zero before the decimal point, we can clearly indicate that it is a floating-point value.

### DIFF
--- a/libraries/AP_Math/tests/test_geodesic_grid.cpp
+++ b/libraries/AP_Math/tests/test_geodesic_grid.cpp
@@ -222,13 +222,13 @@ INSTANTIATE_TEST_CASE_P(GeneralVectors,
  */
 static TestParam hardcoded_vectors[] = {
     /* a + 2 * m_a + .5 * m_c for T_4 */
-    {{.25f * M_GOLDEN, -.25f * (13.0f * M_GOLDEN + 1.0f), - 1.25f}, 17},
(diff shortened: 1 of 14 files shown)


libraries/AP_Math/tests/test_geodesic_grid.cpp
```diff
@@ -226,7 +226,7 @@ static TestParam hardcoded_vectors[] = {
     /* 3 * m_a + 2 * m_b 0 * m_c for T_4 */
     {{M_GOLDEN, -4.0f * M_GOLDEN -1.0f, 1.0f}, -1, {16, 18, -1}},
     /* 2 * m_c + (1 / 3) * m_b + .1 * c for T_13 */
-    {{-.2667f, .1667f * M_GOLDEN, 2.2667f * M_GOLDEN + .1667f}, 55},
+    {{-.2667f, 0.1667f * M_GOLDEN, 2.2667f * M_GOLDEN + 0.1667f}, 55},
```
rmackay9, line 229: I think the first item here was missed.

rmackay9 (approved):
LGTM, thanks!

libraries/AP_Math/tests/test_geodesic_grid.cpp
muramura: @rmackay9 san
Thank you for the review.
I have incorporated your comments.

peterbarker:
Merged, thanks!

PR #29976, opened 2025-05-05
Rover: Change to GCS_SEND_TEXT

muramura:
Both GCS().SEND_TEXT and GCS_SEND_TEXT are being used.
I think it would be better to standardize the notation.


rmackay9 (approved):
Thanks very much!

PR #29978, opened 2025-05-05
Sub:remove SITL dependancies

Hwurzburg:
This makes SITL behave the same as a real board does...these must have been patches for older CI issues...

### DIFF
--- a/ArduSub/AP_Arming_Sub.cpp
+++ b/ArduSub/AP_Arming_Sub.cpp
@@ -120,9 +120,7 @@ bool AP_Arming_Sub::arm(AP_Arming::Method method, bool do_arming_checks)
         AP::notify().update();
     }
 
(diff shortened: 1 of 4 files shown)


Williangalvani (approved):
lgtm. I don't see any reason for those to remain in the code.

ArduSub/AP_Arming_Sub.cpp
```diff
@@ -120,9 +120,7 @@ bool AP_Arming_Sub::arm(AP_Arming::Method method, bool do_arming_checks)
         AP::notify().update();
     }
 
-#if CONFIG_HAL_BOARD == HAL_BOARD_SITL
     send_arm_disarm_statustext("Arming motors");
```
peterbarker, line 123: Could I suggest removing this line too?

PR #29981, opened 2025-05-05
AP_Logger: Refactor the process

muramura:
The processing can be consolidated.
Add a blank line to separate methods.

### DIFF
--- a/libraries/AP_Logger/AP_Logger_MAVLink.cpp
+++ b/libraries/AP_Logger/AP_Logger_MAVLink.cpp
@@ -120,10 +120,7 @@ bool AP_Logger_MAVLink::free_seqno_from_queue(uint32_t seqno, dm_block_queue_t &
 
 bool AP_Logger_MAVLink::WritesOK() const
 {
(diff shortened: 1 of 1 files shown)


libraries/AP_Logger/AP_Logger_MAVLink.cpp
```diff
@@ -354,10 +352,7 @@ void AP_Logger_MAVLink::Write_DMS(AP_Logger_MAVLink &logger_mav)
 
 void AP_Logger_MAVLink::stats_log()
 {
-    if (!_initialised) {
-        return;
-    }
-    if (stats.collection_count == 0) {
+    if (!_initialised || stats.collection_count == 0) {
         return;
```
peterbarker, line 356: Please don't consolidate this

We end up with clauses which are ridiculously difficult to work out what they do.

peterbarker (changes requested):

peterbarker:
Sorry, but the "early return" pattern is used for a reason.  Removing complex compound logical statements makes the code much easier to maintain.

PR #29986, opened 2025-05-06
Fix clang++ compilation error due to multiple values for define in one compilation

peterbarker:


tridge:
@peterbarker nice find! can you do a binaries changed check on this PR?

peterbarker:
> @peterbarker nice find! can you do a binaries changed check on this PR?

No binary output changes.

tridge (approved):

PR #29987, opened 2025-05-06
AP_Mount: Siyi ZT6 min cam version is 0.1.9

rmackay9:
This adds a check that the Siyi ZT6's camera firmware is at least version 0.1.9 which will encourage users to update the firmware on the camera gimbal and reduce issues like those reported in [this 4.6 beta testing thread](https://discuss.ardupilot.org/t/geotag-images-taken-by-siyi-gimbal-cameras/129664/11).

This has been lightly tested on real-hardware
image


rmackay9:

PR #29989, opened 2025-05-06
Add option to ignore parameter sets / LUA applet to handle parameter sets

peterbarker:
~~Add `MAV_OPTION` bit to ignore setting of parameters via MAVLink.~~

State is now ephemeral.

Add a LUA script which will honours `PARAM_SET` - based on a a whitelist of parameter names.

```
STABILIZE> Lua: State memory usage: 4824 + 10598
AP: Scripting: restarted
AP: param-set-filter script loaded
STABILIZE> 
STABILIZE> param set DISARM_DELAY 5
STABILIZE> AP: param-set-filter: param set denied (DISARM_DELAY)
Flight battery 100 percent
AP: param-set-filter: param set denied (DISARM_DELAY)
Failed to set DISARM_DELAY to 5 (invalid returned value 3.0)

STABILIZE> 
```

This work sponsored by Freespace Solutions

### DIFF
--- a/libraries/AP_Scripting/applets/param-set.lua
+++ b/libraries/AP_Scripting/applets/param-set.lua
@@ -1,0 +1,171 @@
+-- Inspect parameter sets received via MAVLink, determine action based
+-- on whitelist.
+
+-- When this script runs and ENABLE is true ArduPilot will stop
+--   processing parameter-sets via the GCS library.  Instead, this
+--   script becomes responsible for setting parameters, and it will
+--   only set parameters which are whitelisted.  Setting ENABLE to
+--   false will allow ArduPilot to set parameters normally.
+
+-- Setting SCR_ENABLE to false while this script is running in the
+--   ENABLE state is... not advised.
+
+-- How To Use
+-- 1. copy this script to the autopilot's "scripts" directory
+-- 2. set SCR_ENABLE to 1
+
+-- global definitions
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+local UPDATE_INTERVAL_MS = 10           -- update at about 100hz
+
+-- get a reference to MAV_SYSID parameter so we can filter messages to
+--   just those parameter-sets aimed at this vehicle:
+MAV_SYSID = Parameter("MAV_SYSID")
+
+-- prefix for all text messages:
+local TEXT_PREFIX_STR = "param-set"
+
+--
+-- parameter setup
+--
+local PARAM_TABLE_KEY = 92
+local PARAM_TABLE_PREFIX = "PARAM_SET_"
+assert(param:add_table(PARAM_TABLE_KEY, PARAM_TABLE_PREFIX, 7), 'could not add param table')
+
+-- add a parameter and bind it to a variable
+function bind_add_param(name, idx, default_value)
+    assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), string.format('could not add param %s', PARAM_TABLE_PREFIX .. name))
+    return Parameter(PARAM_TABLE_PREFIX .. name)
+end
+
+--[[
+  // @Param: PARAM_SET_ENABLE
+  // @DisplayName: Param Set enable
+  // @Description: Param Set enable
+  // @Values: 0:Disabled,1:Enabled
+  // @User: Standard
+--]]
+local PARAM_SET_ENABLE = bind_add_param("ENABLE", 1, 1)
+
+-- initialize MAVLink rx with buffer depth and number of rx message IDs to register
+mavlink:init(5, 1)
+
+-- register message id to receive
+local PARAM_SET_ID = 23
+mavlink:register_rx_msgid(PARAM_SET_ID)
+
+-- handle PARAM_SET message
+local parameters_which_can_be_set = {}
+parameters_which_can_be_set["MAV_OPTIONS"] = true
+parameters_which_can_be_set["PARAM_SET_ENABLE"] = true
+parameters_which_can_be_set["BATT_ARM_MAH"] = true
+parameters_which_can_be_set["BATT_ARM_VOLT"] = true
+parameters_which_can_be_set["BATT_CAPACITY"] = true
+parameters_which_can_be_set["BATT_CRT_MAH"] = true
+parameters_which_can_be_set["BATT_CRT_VOLT"] = true
+parameters_which_can_be_set["BATT_FS_CRT_ACT"] = true
+parameters_which_can_be_set["BATT_FS_LOW_ACT"] = true
+parameters_which_can_be_set["BATT_LOW_MAH"] = true
+parameters_which_can_be_set["BATT_LOW_VOLT"] = true
+parameters_which_can_be_set["BRD_OPTIONS"] = true
+parameters_which_can_be_set["COMPASS_USE3"] = true
+parameters_which_can_be_set["FENCE_ACTION"] = true
+parameters_which_can_be_set["FENCE_ALT_MAX"] = true
+parameters_which_can_be_set["FENCE_ENABLE"] = true
+parameters_which_can_be_set["FENCE_RADIUS"] = true
+parameters_which_can_be_set["FENCE_TYPE"] = true
+parameters_which_can_be_set["LIGHTS_ON"] = true
+parameters_which_can_be_set["LOG_BITMASK"] = true
+parameters_which_can_be_set["LOG_DISARMED"] = true
+parameters_which_can_be_set["LOG_FILE_DSRMROT"] = true
+parameters_which_can_be_set["RTL_ALT"] = true
+parameters_which_can_be_set["RTL_LOIT_TIME"] = true
+parameters_which_can_be_set["RTL_SPEED"] = true
+
+local function should_set_parameter_id(param_id)
+    if parameters_which_can_be_set[param_id] == nil then
+        return false
+    end
+    return parameters_which_can_be_set[param_id]
+end
+
+local function handle_param_set(name, value)
+    -- we will not receive packets in here for the wrong system ID /
+    --   component ID; this is handled by ArduPilot's MAVLink routing
+    --   code
+
+    -- check for this specific ID:
+    if not should_set_parameter_id(name) then
+        gcs:send_text(MAV_SEVERITY.WARNING, string.format("%s: param set denied (%s)", TEXT_PREFIX_STR, name))
+        return
+    end
+
+    param:set_and_save(name, value)
+    gcs:send_text(MAV_SEVERITY.WARNING, string.format("%s: param set applied", TEXT_PREFIX_STR))
+end
+
+-- display welcome message
+gcs:send_text(MAV_SEVERITY.INFO, "param-set script loaded")
+
+-- initialise our knowledge of the GCS's allow-set-parameters state.
+--   We do not want to fight over setting this GCS state via other
+--   mechanisms (eg. an auxiliary function), so we keep this state
+--   around to track what we last set:
+local gcs_allow_set = gcs:get_allow_param_set()
+
+-- update function to receive param_set messages and perhaps act on them
+local function update()
+    -- return immediately if not enabled
+    if (PARAM_SET_ENABLE:get() <= 0) then
+        -- this script is disabled, set allow-via-GCS (once):
+        if not gcs_allow_set then
+          gcs:set_allow_param_set(true)
+          gcs_allow_set = true
+        end
+        -- drain all mavlink messages to avoid processing them when enabled
+        while true do
+          local msg, _ = mavlink:receive_chan()
+          if msg == nil then
+            break
+          end
+        end
+        return
+    end
+
+    -- this script is enabled, disallow setting via normal means (once):
+    if gcs_allow_set then
+        gcs:set_allow_param_set(false)
+        gcs_allow_set = false
+    end
+
+    -- consume all available mavlink messages
+    while true do
+        local msg, _ = mavlink:receive_chan()
+        if msg == nil then
+            break
+        end
+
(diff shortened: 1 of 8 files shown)


libraries/AP_Param/AP_Param.h
```diff
@@ -568,7 +568,9 @@ class AP_Param
 
     // return the persistent top level key for the ParamToken key
     static uint16_t get_persistent_key(uint16_t key) { return var_info(key).key; }
-    
+
+    bool allow_set_via_mavlink(uint16_t flags) const;
```
rmackay9, line 572: maybe we can put a comment above this function too?

libraries/AP_Scripting/applets/param-set-filter.lua
```diff
@@ -0,0 +1,150 @@
...
+
+-- add a parameter and bind it to a variable
+function bind_add_param(name, idx, default_value)
+    assert(param:add_param(PARAM_TABLE_KEY, idx, name, default_value), string.format('could not add param %s', PARAM_TABLE_PREFIX .. name))
+    return Parameter(PARAM_TABLE_PREFIX .. name)
+end
+
+--[[
+  // @Param: PSF_ENABLE
+  // @DisplayName: Packet Set Filter enable
```
rmackay9, line 42: Maybe it's Param set instead of Packet set?

libraries/GCS_MAVLink/GCS.h
```diff
@@ -1236,6 +1236,7 @@ class GCS
 
     enum class Option {
       GCS_SYSID_ENFORCE = (1U << 0),
+      IGNORE_PARAM_SET  = (1U << 1),
```
rmackay9, line 1239: I guess there's a parameter description update to go with this..

libraries/AP_Scripting/applets/param-set-filter.lua
```diff
@@ -0,0 +1,150 @@
...
+  // @Values: 0:Disabled,1:Enabled
+  // @User: Standard
+--]]
+local PSF_ENABLE = bind_add_param("ENABLE", 1, 1)
+
+--[[
+  // @Param: PSF_ENFORCE
+  // @DisplayName: Packet Set Filter enforce
+  // @Description: Packet Set Filter enforce - actually disallow the sets
+  // @Values: 0:Disabled,1:Enabled
```
rmackay9, line 53: I wonder if we really need both an ENABLE and an ENFORCE?

rmackay9:
This is a really great feature.  I've put a few comments in there to hopefully help push it along

libraries/GCS_MAVLink/GCS.h
timtuxworth: If it is persistent don't you have to change the default for SCR_ENABLE to true? otherwise how do you set SCR_ENABLE = 1 on a new install?

libraries/AP_Param/AP_Param.h
peterbarker: Done in precursor PR which will be merged before this one.

libraries/AP_Scripting/applets/param-set-filter.lua
peterbarker: Yes, thanks.

... and this isn't really a "filter" now, as such, it's really an alternate setting mechanism.  So I'm probably going to have to rename this - I'm going to try dropping the "filter" bit.

libraries/GCS_MAVLink/GCS.h
peterbarker: > If it is persistent don't you have to change the default for SCR_ENABLE to true? otherwise how do you set SCR_ENABLE = 1 on a new install?

Well, there's lots of ways to get that effect when building for yourself... and also, you can set that state before installing the script :-)

I'm changing this to be ephemeral state on the next version of this PR, so no longer relevant problem anyway :-)

libraries/AP_Scripting/applets/param-set-filter.lua
peterbarker: Yeah, even I thought this was overkill.

The concept was that ENABLE does what it does in most scripts - stop the script doing *anything*, and ENFORCE meant that it would actually cause it to not set parameters it wasn't sure of (so you could enable the script and make sure it was working as intended before setting enforce, to reduce the chances of locking yourself out of your vehicle.

Since we're making the thing ephemeral anyway, I'll trim this back to just ENABLE - so once enabled it will set parameters for you and set the GCS state variable to force it to stop setting parameters.  And setting ENABLE to false will allow GCS to set parameters and the script will stop setting parameters.

peterbarker:
I have re-jigged this PR extensively following our discussion on DevCall - I think this is a much better result because of that, even 'though it's rather different to the feature we decided on at a *previous* DevCall!

The state is now ephemeral, so the script now actively tells the GCS to stop processing requests when it loads and ENABLE is set.  If ENABLE becomes false it will tell the GCS to resume processing.

... it occurs to me we haven't mentioned the fact that users can notionally now allow set/deny set based on the value of the parameter, not just the name.  That's not done in here, but we've been asked for better parameter value validation on many occasions....

libraries/AP_Scripting/applets/param-set.lua
```diff
@@ -0,0 +1,164 @@
...
+local msg_map = {}
+msg_map[PARAM_SET_ID] = "PARAM_SET"
+
+local TEXT_PREFIX_STR = "param-set"    -- prefix for all text messages
+
+--
+-- parameter setup
+--
+local PARAM_TABLE_KEY = 92
+local PARAM_TABLE_PREFIX = "PS_"
```
rmackay9, line 40: If we've got the space, how about we make "PS_" longer to add more help to the user?  Some ideas that fit within 16 characters:

- PARAM_SET_ENABLE
- PARMSET_ENABLE
- PSET_ENABLE

peterbarker: > If we've got the space, how about we make "PS_" longer to add more help to the user? Some ideas that fit within 16 characters:
> 
>     * PARAM_SET_ENABLE

Yep, went with this one.

libraries/AP_Scripting/applets/param-set.lua
```diff
@@ -0,0 +1,164 @@
...
+    -- our component ID is assumed to be 1:
+    if msg.target_component ~= 1 then
+        return
+    end
+    -- check for this specific ID:
+    if not should_set_parameter_id(msg.param_id) then
+        gcs:send_text(MAV_SEVERITY.WARNING, string.format("%s: param set denied (%s)", TEXT_PREFIX_STR, msg.param_id))
+        return
+    end
+    param:set(msg.param_id, msg.param_value)
```
IamPete1, line 115: This is set only, eg not save. The existing MAVLink checks for changes and possibly also saves.

IamPete1 (commented):
Some things I notice:

- Existing MAVLink saves not just set.
- Existing MAVLink logs the parameter change. 
- Existing MAVLink invalidates the param count for enable params so the GCS should notice and offer to re-fetch all for the user.
- This bypasses `@READONLY` in the firmware.
- Blocks all param sets for MAVFTP.

I wonder if we would be better to have a dedicated param buffer that the script could register. So rather than getting the MAVLink message you would get the name and desired value. The advantage would be that it could work via FTP and the existing `@READONLY` check could be done first. Then we would need a way for the script to trigger a log, we could add a new param method for that. The only remaining thing would be invalidating the count, it looks like the param save queue should do that already.

libraries/AP_Scripting/applets/param-set.lua
```diff
@@ -0,0 +1,164 @@
+-- Inspect parameter sets received via MAVLink, determine action based
```
tridge, line 1: needs a .md file describing how to use this

libraries/AP_Scripting/applets/param-set.lua
```diff
@@ -0,0 +1,164 @@
...
+        if parsed_msg.msgid == PARAM_SET_ID then
+            handle_param_set(parsed_msg)
+        end
+        ::continue::
+    end
+
+    return update, UPDATE_INTERVAL_MS
+end
+
+return update()
```
tridge, line 164: consider adding the pcall wrapper?

tridge (changes requested):
great feature!

libraries/AP_Scripting/applets/param-set.lua
```diff
@@ -0,0 +1,165 @@
...
+    while true do
+        local msg, _ = mavlink:receive_chan()
+        if msg == nil then
+            break
+        end
+
+        local param_value, target_system, target_component, param_id, _ = string.unpack("<fBBc16B", string.sub(msg, 13, 36))
+        param_id = string.gsub(param_id, string.char(0), "")
+
+        handle_param_set(target_system, target_component, param_id, param_value)
```
IamPete1, line 145:
```suggestion
        handle_param_set(param_id, param_value)
```

libraries/AP_Scripting/applets/param-set.lua
```diff
@@ -0,0 +1,165 @@
...
+--   We do not want to fight over setting this GCS state via other
+--   mechanisms (eg. an auxiliary function), so we keep this state
+--   around to track what we last set:
+local gcs_allow_set = gcs:get_allow_param_set()
+
+-- update function to receive param_set messages and perhaps act on them
+local function update()
+    -- return immediately if not enabled
+    if (PARAM_SET_ENABLE:get() <= 0) then
+        -- this script is disabled, set allow-via-GCS (once):
```
IamPete1, line 121: Should probably still drain the buffer in this case. Otherwise when you enable the earlier sets will suddenly go through.

libraries/AP_Scripting/applets/param-set.lua
IamPete1: That will probably result in unused variable warnings, so you will have to add underscores.

peterbarker: Done, thanks.  How on earth did this pass tests, I wonder?!

libraries/AP_Scripting/applets/param-set.lua
peterbarker: Added a loop to drain the messages.

libraries/AP_Scripting/applets/param-set.lua
IamPete1: Didn't pass the scripting checks ;)

libraries/AP_Scripting/applets/param-set.md
```diff
@@ -0,0 +1,77 @@
...
+- Intercepts MAVLink `PARAM_SET` messages targeted at the vehicle.
+- Validates that the parameter ID is in the whitelist.
+- Only whitelisted parameters will be updated.
+- Dynamically toggles ArduPilot’s `gcs:set_allow_param_set()` based on enable state.
+
+## Requirements
+
+- ArduPilot 4.7 or later with Lua scripting enabled (`SCR_ENABLE = 1`)
+- The script must be placed in appropriate directories:
+  - Script in: `/scripts/param_set_filter.lua`
```
timtuxworth, line 22: I think this should say /APM/scripts
people will get confused and put this in a scripts directory in root.
it says "scripts in", so should it specify the file name (maybe not)?

timtuxworth (commented):
A few suggestions for the lua / UX.

libraries/AP_Scripting/applets/param-set.lua
```diff
@@ -0,0 +1,172 @@
...
+parameters_which_can_be_set["LOG_FILE_DSRMROT"] = true
+parameters_which_can_be_set["RTL_ALT"] = true
+parameters_which_can_be_set["RTL_LOIT_TIME"] = true
+parameters_which_can_be_set["RTL_SPEED"] = true
+
+local function should_set_parameter_id(param_id)
+    if parameters_which_can_be_set[param_id] == nil then
+        return false
+    end
+    return true
```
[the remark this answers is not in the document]
peterbarker: Changed.

libraries/AP_Scripting/applets/param-set.lua
```diff
@@ -0,0 +1,172 @@
...
+end
+
+local function handle_param_set(name, value)
+    gcs:send_text(MAV_SEVERITY.WARNING, string.format("%s: param set received", TEXT_PREFIX_STR))
```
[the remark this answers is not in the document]
peterbarker: Done

libraries/AP_Scripting/applets/param-set.md
peterbarker: Changed it

libraries/AP_Scripting/applets/param-set.md
```diff
@@ -0,0 +1,77 @@
...
+
+- `MAV_OPTIONS`
+- `PARAM_SET_ENABLE`
+- Battery parameters (e.g., `BATT_ARM_MAH`, `BATT_FS_CRT_ACT`, etc.)
+- `BRD_OPTIONS`
+- `COMPASS_USE3`
+- Geofence and RTL parameters (e.g., `FENCE_TYPE`, `RTL_ALT`)
+- `LOG_*`, `LIGHTS_ON`
+
+(See the full script for the complete whitelist.)
```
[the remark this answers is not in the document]
peterbarker: I think that's kind of implicit.

libraries/AP_Scripting/applets/param-set.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Usage
+
+1. Copy the script into the SD card's `/scripts/` directory.
```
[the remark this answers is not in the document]
peterbarker: Fixed.

libraries/AP_Scripting/applets/param-set.md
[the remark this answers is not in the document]
timtuxworth: Yes - but making it explicit might help some users. :-)

rmackay9 (approved):
LGTM, thanks!

tridge (approved):

PR #29991, opened 2025-05-06
Remove unused baro code

peterbarker:
- the HAL_BARO_DEFAULT for the LPSx25 is unused and ancient (crazyflie specifies this driver in its hwdef, so it can be done else-wise)
 - HAL_BARO_* defines are unused, so comment them out
 - HAL_BARO_DEFAULT defaults to HAL_BARO_NONE in AP_HAL_Boards.h, no need for other definitions to this value
 - remove unused defines of various baro names in hwdef.h files
 - remove defines of HAL_BARO_DEFAULT to values that don't exist / were never checked against / were killed before esp32 stuff was merged into master (eg. icarous)
 - remove defines of baro variables that are the default anyway
 -


peterbarker:
Journey and skyviper-f412-rev1 are both binary-output-identical once the hwdef embedding is taken out of the equation.

tridge (approved):

PR #29994, opened 2025-05-06
Copter:update TUNE metadata

Hwurzburg:


ArduCopter/Parameters.cpp
```diff
@@ -300,8 +300,8 @@ const AP_Param::Info Copter::var_info[] = {
     GSCALAR(esc_calibrate, "ESC_CALIBRATION",       0),
 
     // @Param: TUNE
-    // @DisplayName: Channel 6 Tuning
-    // @Description: Controls which parameters (normally PID gains) are being tuned with transmitter's channel 6 knob
+    // @DisplayName: Parameter Tuning
```
rmackay9, line 303: Hi, thanks very much for this.  Could we change it to something like this?

DisplayName: Tuning Parameter
Description: Tuning parameter (normally a PID gain) that is being tuned with an RC transmitter's knob.  The RC input channel used is assigned by setting RCx_OPTION to 219

Hwurzburg: done

Hwurzburg:
@rmackay9 merge then?

PR #29995, opened 2025-05-06
Sub: allow limiting upwards throttle at surface

Williangalvani:
This solves an issue where the thrusters start sucking in air when at the surface, which increased noise and power consumption

### DIFF
--- a/libraries/AP_Motors/AP_Motors6DOF.h
+++ b/libraries/AP_Motors/AP_Motors6DOF.h
@@ -41,6 +41,8 @@ public:
     // output_to_motors - sends minimum values out to the motors
     void output_to_motors() override;
 
+    void set_max_throttle(float max_throttle) { _max_throttle = max_throttle; }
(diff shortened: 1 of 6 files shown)


rmackay9:
Hi @Williangalvani,

Sorry to be so picky recently but normally devs don't merge their own PRs to master before getting an approval from another developer

libraries/AP_Motors/AP_Motors6DOF.h
```diff
@@ -41,6 +41,8 @@ class AP_Motors6DOF : public AP_MotorsMatrix {
     // output_to_motors - sends minimum values out to the motors
     void output_to_motors() override;
 
+    void set_max_throttle(float max_throttle) { _max_throttle = max_throttle; }
```
rmackay9, line 44: It would be good to add a comment above this function. Sorry but I think we have comments above every other function

PR #29996, opened 2025-05-06
Sub: allow SURFACE mode to work with no altitude data

Williangalvani:
This fixes an issue where SURFACE mode doesn't work on vehicles with no baros for altitude sensor.

### DIFF
--- a/ArduSub/Parameters.cpp
+++ b/ArduSub/Parameters.cpp
@@ -775,6 +775,14 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("ORIGIN_ALT", 21, ParametersG2, backup_origin_alt, 0),
 
+    // @Param: SFC_NOBARO_THST
+    // @DisplayName: Surface mode throttle output when no barometer is available
+    // @Description: Surface mode throttle output when no borometer is available. 100% is full throttle. -100% is maximum throttle downwards
+    // @Units: %
+    // @User: Standard
+    // @Range: -100 100
+    AP_GROUPINFO("SFC_NOBARO_THST", 22, ParametersG2, surface_nobaro_thrust, 10),
+
     AP_GROUPEND
 };
 
--- a/ArduSub/Parameters.h
+++ b/ArduSub/Parameters.h
@@ -209,7 +209,7 @@ public:
         k_param_cam_tilt_center, // deprecated
         k_param_frame_configuration,
         k_param_surface_max_throttle,
-
+        k_param_surface_nobaro_thrust,
         // 200: flight modes
         k_param_flight_mode1 = 200,
         k_param_flight_mode2,
@@ -404,6 +404,7 @@ public:
     AP_Float backup_origin_lat;
     AP_Float backup_origin_lon;
     AP_Float backup_origin_alt;
+    AP_Float surface_nobaro_thrust;
 };
 
 extern const AP_Param::Info        var_info[];
--- a/ArduSub/mode.cpp
+++ b/ArduSub/mode.cpp
@@ -95,8 +95,8 @@ bool Sub::set_mode(Mode::Number mode, ModeReason reason)
 
     // check for valid altitude if old mode did not require it but new one does
     // we only want to stop changing modes if it could make things worse
-    if (flightmode->has_manual_throttle() &&
-        !new_flightmode->has_manual_throttle() &&
+    if (!flightmode->requires_altitude() &&
+        new_flightmode->requires_altitude() &&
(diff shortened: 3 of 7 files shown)


Hwurzburg:
LGTM...sims fine...did not think you would want to get rid of requires GPS in the mode....if you are happy, I am happy

Hwurzburg (approved):

ArduSub/Parameters.cpp
```diff
@@ -767,6 +767,13 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("ORIGIN_ALT", 21, ParametersG2, backup_origin_alt, 0),
 
+    // @Param: SURF_NOBARO_THST
+    // @DisplayName: Throttle output when surface mode is engaged and baro is not available
```
rmackay9, line 771: not a blocker of course but normally with Parameter descriptions I try to put the most important word first.  So in this case, I'd start it with "Surface mode".  So for example it could be, "Surface mode throttle output when no barometer is available"

ArduSub/Parameters.h
```diff
@@ -208,7 +208,7 @@ class Parameters {
         k_param_throttle_gain,
         k_param_cam_tilt_center, // deprecated
         k_param_frame_configuration,
-
+        k_param_surface_sensorless_thrust,
```
rmackay9, line 211: sorry I'm going to be pedantic but I wonder if we could be consistent about whether we use  "nobaro" vs "sensorless"?  It's one of those golden rules where we try not to use synonyms.  This helps both with searchability (e.g. searches for "nobaro" will turn up both) and it helps with non-native english speakers who may not know about the synonym.

ArduSub/mode.cpp
```diff
@@ -95,8 +95,8 @@ bool Sub::set_mode(Mode::Number mode, ModeReason reason)
 
     // check for valid altitude if old mode did not require it but new one does
     // we only want to stop changing modes if it could make things worse
-    if (flightmode->has_manual_throttle() &&
```
rmackay9, line 98: this rename makes it inconsistent with Copter.  That's not necessarily a blocker but as we try to move vehicles closer to each other, let's try not to move in the opposite direction unless it's really necessary

ArduSub/mode_surface.cpp
```diff
@@ -32,28 +30,32 @@ void ModeSurface::run()
...
 
-    // Already at surface, hold depth at surface
-    if (sub.ap.at_surface) {
-        set_mode(Mode::Number::ALT_HOLD, ModeReason::SURFACE_COMPLETE);
-    }
-
-    // convert pilot input to lean angles
-    // To-Do: convert sub.get_pilot_desired_lean_angles to return angles as floats
-    sub.get_pilot_desired_lean_angles(channel_roll->get_control_in(), channel_pitch->get_control_in(), target_roll, target_pitch, sub.aparm.angle_max);
+    if (sensorless_mode) {
```
rmackay9, line 33: maybe we could put a comment above this block of code?

rmackay9:
It's really great to see this!

I'm not the Sub maintainer so please take my comments with a grain of salt.  I'm just being as pedantic as I would be if the code were Copter or Rover.

ArduSub/Parameters.cpp
Williangalvani: thanks! updated

ArduSub/mode_surface.cpp
Williangalvani: done!

ArduSub/Parameters.cpp
```diff
@@ -775,6 +775,14 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @User: Standard
     AP_GROUPINFO("ORIGIN_ALT", 21, ParametersG2, backup_origin_alt, 0),
 
+    // @Param: SURF_NOBARO_THST
+    // @DisplayName: Surface mode throttle output when no barometer is available
+    // @Description: Surface mode throttle output when no borometer is available. 100% is full throttle. -100% is maximum throttle downwards
+    // @Units: %
+    // @User: Standard
+    // @Range: -100 100
+    AP_GROUPINFO("SURF_NOBARO_THST", 22, ParametersG2, surface_nobaro_thrust, 10),
```
peterbarker, line 784:
```suggestion
    AP_GROUPINFO("SFC_NOBARO_THST", 22, ParametersG2, surface_nobaro_thrust, 10),
```

We're using SURF as a prefix in the "surface tracking" library, which sub might want at some stage.

rmackay9 (approved):
looks fine, thanks!

peterbarker (changes requested):

PR #29997, opened 2025-05-06
Allow initializing Camera_Servo to trim.

Williangalvani:
this allows the user to decide the initial values for the camera focus/zoom by changing the trim parameters.

Is there a cleaner way to do this?

The main issue is we want a  linear range but starting at the _TRIM value.

### DIFF
--- a/libraries/AP_Camera/AP_Camera_Servo.cpp
+++ b/libraries/AP_Camera/AP_Camera_Servo.cpp
@@ -61,9 +61,13 @@ bool AP_Camera_Servo::trigger_pic()
 
 bool AP_Camera_Servo::set_zoom(ZoomType zoom_type, float zoom_value)
 {
-    if (zoom_type == ZoomType::RATE) {
-        zoom_current_rate = zoom_value;
-        return true;
+    switch (zoom_type) {
+        case ZoomType::RATE:
+            zoom_current_rate = zoom_value;
+            return true;
+        case ZoomType::PCT:
+            SRV_Channels::set_output_scaled(SRV_Channel::k_cam_zoom, zoom_value * 0.01);
+            return true;
     }
     return false;
 }
@@ -71,9 +75,15 @@ bool AP_Camera_Servo::set_zoom(ZoomType zoom_type, float zoom_value)
 // set focus specified as rate
 SetFocusResult AP_Camera_Servo::set_focus(FocusType focus_type, float focus_value)
 {
-    if (focus_type == FocusType::RATE) {
-        focus_current_rate = focus_value;
-        return SetFocusResult::ACCEPTED;
+    switch (focus_type) {
+        case FocusType::RATE:
+            focus_current_rate = focus_value;
+            return SetFocusResult::ACCEPTED;
+        case FocusType::PCT:
+            SRV_Channels::set_output_scaled(SRV_Channel::k_cam_focus, focus_value * 0.01);
+            return SetFocusResult::ACCEPTED;
+        case FocusType::AUTO:
+            return SetFocusResult::UNSUPPORTED;
     }
     return SetFocusResult::UNSUPPORTED;
 }


libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -61,21 +61,31 @@ bool AP_Camera_Servo::trigger_pic()
...
-        focus_current_rate = focus_value;
-        return SetFocusResult::ACCEPTED;
+    switch (focus_type) {
+        case FocusType::RATE:
+            focus_current_rate = focus_value;
+            return SetFocusResult::ACCEPTED;
+        case FocusType::PCT:
+            SRV_Channels::set_output_scaled(SRV_Channel::k_cam_focus, focus_value * 0.01);
+            return SetFocusResult::ACCEPTED;
+        default:
```
peterbarker, line 86: remove default, add all enumeration values

libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -61,21 +61,31 @@ bool AP_Camera_Servo::trigger_pic()
...
-        return true;
+    switch (zoom_type) {
+        case ZoomType::RATE:
+            zoom_current_rate = zoom_value;
+            return true;
+        case ZoomType::PCT:
+            SRV_Channels::set_output_scaled(SRV_Channel::k_cam_zoom, zoom_value * 0.01);
+            return true;
+        default:
+            return false;
```
peterbarker, line 72: remove default, add all enumeration values

peterbarker (changes requested):

libraries/AP_Camera/AP_Camera_Servo.cpp
Williangalvani: done!

libraries/SRV_Channel/SRV_Channel.cpp
```diff
@@ -194,6 +194,26 @@ uint16_t SRV_Channel::pwm_from_range(float scaled_value) const
     return servo_min + uint16_t( (scaled_value * (float)(servo_max - servo_min)) / (float)high_out );
 }
 
+// convert a pwm to a 0..1 value
+float SRV_Channel::range_from_pwm() const
```
IamPete1, line 198: It might be nice to call this "scaled from pwm" and have it work for angle type too. Don't necessarily need to do it now. It would just be nice to check that this is configured as a range type `type_setup && !type_angle`.

libraries/SRV_Channel/SRV_Channel.cpp
```diff
@@ -194,6 +194,26 @@ uint16_t SRV_Channel::pwm_from_range(float scaled_value) const
...
+{
+    if (servo_max <= servo_min || high_out == 0) {
+        return 0;
+    }
+    if (output_pwm < servo_min) {
+        return 0;
```
IamPete1, line 204:
```suggestion
        return reversed ? 1.0 : 0.0;
```

libraries/SRV_Channel/SRV_Channel.cpp
```diff
@@ -194,6 +194,26 @@ uint16_t SRV_Channel::pwm_from_range(float scaled_value) const
...
+    }
+    if (output_pwm > servo_max) {
+        return 1;
```
IamPete1, line 207: return reversed ? 0.0 : 1.0;

libraries/SRV_Channel/SRV_Channel.cpp
```diff
@@ -194,6 +194,26 @@ uint16_t SRV_Channel::pwm_from_range(float scaled_value) const
...
+    }
+    float full_range = servo_max - servo_min;
+    float value = (float)(output_pwm - servo_min) / full_range;
```
IamPete1, line 210: In fact I think you could just use `linear_interpolate(0.0, 1.0, output_pwm, servo_min, servo_max)`

libraries/SRV_Channel/SRV_Channel_aux.cpp
```diff
@@ -757,6 +757,16 @@ void SRV_Channels::set_output_to_trim(SRV_Channel::Function function)
     }
 }
 
+void SRV_Channels::set_output_scaled_to_trim(SRV_Channel::Function function)
+{
+    for (uint8_t i=0; i<NUM_SERVO_CHANNELS; i++) {
+        if (channels[i].function == function) {
+            channels[i].set_output_pwm(channels[i].servo_trim);
+            set_output_scaled(function, channels[i].range_from_pwm() * channels[i].high_out);
```
IamPete1, line 765: On problem here is that you could have several output channels set to a given function. However, there is only one scaled output for each function. So, you will end up with the scaled output that represents the last channel in this loop.

libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -10,8 +10,8 @@ extern const AP_HAL::HAL& hal;
 void AP_Camera_Servo::init()
 {
     // set the zoom and focus to the trim point
-    SRV_Channels::set_output_scaled(SRV_Channel::k_cam_zoom, 500);
-    SRV_Channels::set_output_scaled(SRV_Channel::k_cam_focus, 500);
+    SRV_Channels::set_output_scaled_to_trim(SRV_Channel::k_cam_zoom);
+    SRV_Channels::set_output_scaled_to_trim(SRV_Channel::k_cam_focus);
```
IamPete1, line 14: You could acheave this with a output of 0 and set the function to a angle type.

Williangalvani: That's what I thought at first... 

The issue is that in angle mode the range gets squished.

Say we have MIN=1000, TRIM = 1300, MAX=2000. this means the curve (input vs pwm) is a lot steeper on the lower side. 

On the edge case of "Let's always start with the minimum zoom, so we can see everything", so that SERVO_TRIM = SERVO_MIN. That meant all lower range (-1 to 0?) had the same output, which just felt like a huge deadzone.

Can you think of an alternative solution?

Williangalvani:
Tagging devcall to check if someone has a better approach for this...

tridge:
I think adding a TRIM_RANGE here would help: 
```
    // used to get min/max/trim limit value based on reverse
    enum class Limit {
        TRIM,
        MIN,
        MAX,
        ZERO_PWM
    };
```

PR #29998, opened 2025-05-06
Fix invalid escape sequence in chibios.py

ArthurPatriot:
- Error Message: `ardupilot/Tools/ardupilotwaf/chibios.py:105: SyntaxWarning: invalid escape sequence '\P' if not where_python or "\Python\Python" not in where_python or "python.exe" not in where_python:`

- Python version: 3.12

- Ardupilot: Plane 4.5.7

- OS: Windows 11 + WSL Ubuntu 24.04

### DIFF
--- a/Tools/ardupilotwaf/chibios.py
+++ b/Tools/ardupilotwaf/chibios.py
@@ -102,7 +102,7 @@ class upload_fw(Task.Task):
         except subprocess.CalledProcessError:
             #if where.exe can't find the file it returns a non-zero result which throws this exception
             where_python = ""
-        if not where_python or "\Python\Python" not in where_python or "python.exe" not in where_python:
+        if not where_python or r"\Python\Python" not in where_python or "python.exe" not in where_python:
             print(self.get_full_wsl2_error_msg("Windows python.exe not found"))
             return False
         return True


Tools/ardupilotwaf/chibios.py
```diff
@@ -102,7 +102,7 @@ def wsl2_prereq_checks(self):
         except subprocess.CalledProcessError:
             #if where.exe can't find the file it returns a non-zero result which throws this exception
             where_python = ""
-        if not where_python or "\Python\Python" not in where_python or "python.exe" not in where_python:
+        if not where_python or r"\Python\Python" not in where_python or "python.exe" not in where_python:
```
peterbarker, line 105: This does't look right; these aren't regular expressions, and that's what adding `r` to the front of a string indicates.

Just remove the first two clauses here:
```suggestion
        if "python.exe" not in where_python:
```

The first bit is redundant, the second bit we don't want (we don't care where python.exe exists, just that it exists.

peterbarker (changes requested):

peterbarker:
... also, PRs against master, backports if the patches are accepted, please.

peterbarker:
This string no longer exists in master - it's been fixed up in there now.

Thanks anyway!

magicrub:
@peterbarker should this get back ported?

PR #29999, opened 2025-05-07
hwdef: remove unused lines from JFB110 hwdef

peterbarker:
```
Board,copter
JFB110,*
```


rmackay9:
This board is owned by JAE so hopefully we can get feedback from @jfbblue0922

peterbarker:
@rmackay9 feel free to test - periodic testing of boards is a good idea anyway! - but it it binary-output identical :-)

rmackay9 (approved):
ok, if the binary output is identical then no need to test really, I'm sure it's fine, thanks!

PR #29957, opened 2025-05-02
Revert "Tools: ros2: Add airspeed test with plane fixtures"

peterbarker:
This reverts commit 4f666a1ae15951fba2d982e0afa3e46dad84bb84.

these tests seem to be failing in CI

PR #30002, opened 2025-05-07
Tools: Add airspeed test and SITL plane fixtures

Ryanf55:
Try again for https://github.com/ArduPilot/ardupilot/pull/29832 which broke after merge due to unrelated changes.
Originally reverted in https://github.com/ArduPilot/ardupilot/pull/29957

Tested 5x locally with no issues. I also checked the timeouts in the test report, and bumped this to 20 seconds since it's almost 10 seconds on each test to receive airspeed data from boot.


peterbarker:
Merged, thanks.

Might keep "ROS or DDS in the commit messages to not confuse things with the Python tests :-)

PR #30004, opened 2025-05-07
AP_HAL_ChibiOS: wrap `define AP_XX_ENABLED AP_PERIPH_YY_ENABLED` inside ifndef-endif

shiv-tyagi:
These options are listed on the custom build server hence we should be only defining these if they are not already defined. The build fails if the custom build server defines them first and then we redefine them here again.


peterbarker (approved):
LGTM

peterbarker (approved):
LGTM

PR #30006, opened 2025-05-07
AP_Mount: Adjust the version setting to match ZT6

muramura:
By setting the version to bdde2690 , it becomes a single line.
This makes the version easier to visualize.


rmackay9 (approved):
yes, thanks!

PR #30007, opened 2025-05-07
Global: Add units to the method that calculates the distance meter between waypoin ts

muramura:
The distance between WAYPOINTs is in centimeters.
I will add the unit to the method.

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -925,7 +925,7 @@ void Copter::update_altitude()
 bool Copter::get_wp_distance_m(float &distance) const
 {
     // see GCS_MAVLINK_Copter::send_nav_controller_output()
-    distance = flightmode->wp_distance() * 0.01;
+    distance = flightmode->wp_distance_m();
     return true;
 }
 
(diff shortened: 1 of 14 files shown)


rmackay9:
Thanks, I'm happy with this change but we should check that it doesn't interfere with @lthall and @peterbarker's ongoing work to rationalise the units of follow mode and the position controller

peterbarker:
This is a good patch, it's definitely in-line with what @lthall has been up to.

However, I would *also* suggest we could merge a PR which renames this method to add `_m` and starts to return metres rather than cm.

The reason we can do this is that you are changing *all* of the relevant code here, whereas the work Leonard is doing is going to lead to a bunch of subsequent PRs to start using base-SI-units.

@muramura could you put together an alternate PR which renames the current method and makes it return metres, and then change all callers to use the new method and cope with the metres return value, please?

muramura:
Convert centimeters to meters.

lthall (approved):
Nice work :)

Thanks for helping with this!!

peterbarker (approved):
LGTM

timtuxworth:
nice

PR #30009, opened 2025-05-08
Scripted Arming Checks

timtuxworth:
This adds a Lua Applet for scripted arming checks. Each check can be set as mandatory (generates a prearm fail), warning (displays a message, but allows arming) or disabled by setting a parameter.

There is a set of standard checks that I find useful, of course new ones can be added. Any suggestions appreciated. I've just added a check for Rally Point further than RALLY_LIMIT_KM from the home location.

This is designed to get around the controversy that often arises about adding arming checks to c++. To some people particular checks are essential, while to others they are unnecessary or annoying. 

This works for plane, copter and rover for now, but as a base could easily be extended to other vehicles.

### DIFF
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -7502,6 +7502,214 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
         self.set_rc(2, 1500)
         self.fly_home_land_and_disarm()
 
+    def ScriptedArmingChecksApplet(self):
+        """ Applet for Arming Checks will prevent a vehicle from arming based on scripted checks
+            """
+        self.start_subtest("Scripted Arming Checks Applet validation")
+        self.context_collect("STATUSTEXT")
+
+        """Initialize the FC"""
+        self.set_parameter("SCR_ENABLE", 1)
+        self.install_applet_script_context("arming-checks.lua")
+        self.reboot_sitl()
+        self.wait_ekf_happy()
+        self.wait_text("ArduPilot Ready", check_context=True)
+        self.wait_text("Arming Checks .* loaded", timeout=30, check_context=True, regex=True)
+
+        self.start_subsubtest("ArmCk: MAV_SYSID not set")
+        self.progress("Currently SYSID is %f" % self.get_parameter('MAV_SYSID'))
+        self.wait_text("ArmCk: warn: MAV_SYSID not set", timeout=30, check_context=True, regex=True)
+        """ disable the SYSID check, since autotest doesn't like changing the sysid"""
+        self.set_parameter("ARM_SYSID", -1)
+
+        ''' This check comes first since its part of the standard parameters - an invalid scaling speed'''
+        self.start_subsubtest("ArmCk: SCALING_SPEED close to AIRSPEED_CRUISE")
+        self.assert_prearm_failure("fail: Scaling spd", other_prearm_failures_fatal=False)
+        self.set_parameter("SCALING_SPEED", 22)
+        self.wait_text("clear: Scaling spd", check_context=True)
+
+        self.start_subsubtest("ArmCk: FOLL_SYSID must be set if FOLL_ENABLE = 1")
+        self.set_parameter("FOLL_ENABLE", 1)
+        self.set_parameter("FOLL_OFS_X", 10)
+        self.assert_prearm_failure("FOLL_SYSID not set", other_prearm_failures_fatal=False)
+        self.set_parameter("FOLL_SYSID", 3)
+        self.wait_text("clear: FOLL_SYSID not set", check_context=True)
+        self.set_parameter("FOLL_SYSID", -1)
+
+        self.start_subsubtest("ArmCk: FOLL_OFS_[XYZ] must be set if FOLL_ENABLE = 1")
+        self.set_parameter("FOLL_OFS_X", 0)
+        self.assert_prearm_failure("FOLL_OFS_[XYZ] = 0", other_prearm_failures_fatal=False)
+        self.set_parameter("FOLL_OFS_X", 10)
+        self.wait_text("clear: FOLL_OFS_[XYZ] = 0", check_context=True)
+        self.set_parameter("FOLL_OFS_X", 0)
+        self.assert_prearm_failure("FOLL_OFS_[XYZ] = 0", other_prearm_failures_fatal=False)
+        self.set_parameter("FOLL_OFS_Y", 10)
+        self.wait_text("clear: FOLL_OFS_[XYZ] = 0", check_context=True)
+        self.set_parameter("FOLL_OFS_Y", 0)
+        self.assert_prearm_failure("FOLL_OFS_[XYZ] = 0", other_prearm_failures_fatal=False)
+        self.set_parameter("FOLL_OFS_Z", 10)
+        self.wait_text("clear: FOLL_OFS_[XYZ] = 0", check_context=True)
+        """ clear these checks to make the context cleaner for subsequent tests """
+        self.set_parameters({
+            "ARM_SYSID": -1,
+            "ARM_FOLL_SYSID": -1,
+            "ARM_FOLL_SYSID_X": -1,
+            "ARM_FOLL_OFS_DEF": -1,
+        })
+
+        self.start_subsubtest("ArmCk: RTL_ALTITUDE must be legal")
+        self.progress("Currently ARM_P_RTL_ALT is %f" % self.get_parameter('ARM_P_RTL_ALT'))
+        self.set_parameter("RTL_ALTITUDE", 150)
+        self.assert_prearm_failure("ArmCk: fail: RTL_ALTITUDE too high", other_prearm_failures_fatal=False)
+        self.set_parameter("RTL_ALTITUDE", 120)
+        self.wait_text("clear: RTL_ALT", check_context=True)
+
+        self.start_subsubtest("ArmCk: RTL_CLIMB_MIN must be legal")
+        self.set_parameter("RTL_CLIMB_MIN", 150)
+        self.wait_text("ArmCk: warn: RTL_CLIMB_MIN too high", check_context=True)
+        self.set_parameter("RTL_CLIMB_MIN", 120)
+        self.wait_text("clear: RTL_CLIMB_MIN too high", check_context=True)
+
+        self.start_subsubtest("ArmCk: AIRSPEED stall < min < cruise < max")
+        ''' Airspeed parameter start out as
+        AIRSPEED_STALL = 0
+        AIRSPEED_MIN = 10
+        AIRSPEED_CRUISE = 22
+        AIRSPEED_MAX = 30
+        '''
+        self.start_subsubtest("ArmCk: AIRSPEED max < others")
+        self.set_parameter("AIRSPEED_MAX", 5)
+        self.assert_prearm_failure("ArmCk: fail: stall < min", other_prearm_failures_fatal=False)
+        self.set_parameter("AIRSPEED_MAX", 30)
+        self.wait_text("clear: stall < min", check_context=True)
+        self.start_subsubtest("ArmCk: AIRSPEED cruise < min")
+        self.set_parameter("AIRSPEED_MIN", 25)
+        self.assert_prearm_failure("ArmCk: fail: stall < min", other_prearm_failures_fatal=False)
+        self.set_parameter("AIRSPEED_MIN", 10)
+        self.wait_text("clear: stall < min", check_context=True)
+        self.start_subsubtest("ArmCk: AIRSPEED cruise > max")
+        self.set_parameter("AIRSPEED_CRUISE", 40)
+        self.set_parameter("SCALING_SPEED", 40)
+        self.assert_prearm_failure("ArmCk: fail: stall < min", other_prearm_failures_fatal=False)
+        self.set_parameter("AIRSPEED_CRUISE", 22)
+        self.set_parameter("SCALING_SPEED", 22)
+        self.wait_text("clear: stall < min", check_context=True)
+
+        self.start_subsubtest("ArmCk: AIRSPEED_MIN must be > AIRSPEED_STALL")
+        ''' only applies if AIRSPEED_STALL is non zero'''
+        self.set_parameter("AIRSPEED_STALL", 2)
+        self.wait_text("ArmCk: info: Min Speed not", check_context=True)
+        self.set_parameter("AIRSPEED_STALL", 8)
+
+        self.start_subsubtest("ArmCk: Mount SYSID must not match FOLL_SYSID")
+        self.set_parameter("ARM_SYSID", -1)
+        ''' to fail the check MNTx_SYSID_DFLT must be non zero but not= FOLL_SYSID'''
+        self.set_parameter("MNT1_SYSID_DFLT", 1)
+        ''' Need a healthy camera mount defined for this to work '''
+        self.setup_servo_mount()
+        self.progress("rebooting to enable MNT1")
+        self.reboot_sitl() # to handle MNT_TYPE changing
+        self.wait_ekf_happy()
+        self.wait_text("ArduPilot Ready", check_context=True)
+        self.wait_text("Arming Checks .* loaded", timeout=30, check_context=True, regex=True)
+        self.progress("Currently FOLL_SYSID is %f" % self.get_parameter('FOLL_SYSID'))
+        self.progress("Currently MNT1_SYSID_DFLT is %f" % self.get_parameter('MNT1_SYSID_DFLT'))
+        self.set_parameter("ARM_SYSID", -1)
+        self.progress("Currently FOLL_SYSID is %f" % self.get_parameter('FOLL_SYSID'))
+        self.progress("Currently MNT1_SYSID_DFLT is %f" % self.get_parameter('MNT1_SYSID_DFLT'))
+        self.wait_text("warn: MNTx_SYSID != FOLL", check_context=True)
+
+        self.start_subsubtest("ArmCk: Fence must be enabled or autoenabled (warning)")
+        self.reboot_sitl()
+        self.wait_ekf_happy()
+        self.wait_text("ArduPilot Ready", check_context=True)
+        self.wait_text("Arming Checks .* loaded", timeout=30, check_context=True, regex=True)
+        self.load_fence("CMAC-fence.txt")
+        self.wait_statustext("warn: Fence not enabled", check_context=True)
+        self.set_parameter("FENCE_ENABLE", 1)
+        self.wait_statustext("clear: Fence not enabled", check_context=True)
+        self.set_parameter("FENCE_ENABLE", 0)
+        self.wait_statustext("warn: Fence not enabled", check_context=True)
+        self.set_parameter("FENCE_AUTOENABLE", 1)
+        self.wait_text("clear: Fence not enabled", check_context=True)
+        self.set_parameter("FENCE_AUTOENABLE", 0)
+        self.wait_text("warn: Fence not enabled", check_context=True)
+
+    def ScriptedArmingChecksAppletEStop(self):
+        """ Applet for Arming Checks will prevent a vehicle from arming based on scripted checks
+            """
+        self.start_subtest("Scripted Arming Checks Applet validation")
+        self.context_collect("STATUSTEXT")
+
+        """Initialize the FC"""
+        self.set_parameter("SCR_ENABLE", 1)
+        self.install_applet_script_context("arming-checks.lua")
+        self.reboot_sitl()
+        self.wait_ekf_happy()
+        self.wait_text("ArduPilot Ready", check_context=True)
+        self.wait_text("Arming Checks .* loaded", timeout=30, check_context=True, regex=True)
+        self.set_parameters({
+            "ARM_SYSID": -1,
+            "ARM_FOLL_SYSID": -1,
+            "ARM_FOLL_SYSID_X": -1,
+            "ARM_FOLL_OFS_DEF": -1,
+            "ARM_P_SCALING": -1,
+        })
+
+        self.start_subsubtest("ArmCk: Cannot arm while motors estopped")
+        self.set_parameter("RC6_OPTION", 165)
+        self.progress("rebooting to enable RC channel")
+        self.reboot_sitl() # to handle RC option changing
+        self.wait_ekf_happy()
+        self.wait_text("ArduPilot Ready", check_context=True)
+        self.wait_text("Arming Checks .* loaded", timeout=30, check_context=True, regex=True)
+        self.set_parameters({
+            "ARM_SYSID": -1,
+            "ARM_FOLL_SYSID": -1,
+            "ARM_FOLL_SYSID_X": -1,
+            "ARM_FOLL_OFS_DEF": -1,
+            "ARM_P_SCALING": -1,
+        })
+        self.set_rc(6, 1000)
+        self.assert_prearm_failure("ArmCk: fail: Motors EStopped", other_prearm_failures_fatal=False)
+        self.set_rc(6, 2000)
+        self.wait_text("clear: Motors EStopped", timeout=30, check_context=True, regex=True)
+        self.wait_ready_to_arm()
+
+    def ScriptedArmingChecksAppletRally(self):
+        """ Applet for Arming Checks will prevent a vehicle from arming based on scripted checks
+            """
+        self.start_subtest("Scripted Arming Checks Applet validation")
+        self.context_collect("STATUSTEXT")
+
+        """Initialize the FC"""
+        self.set_parameter("SCR_ENABLE", 1)
+        self.install_applet_script_context("arming-checks.lua")
+        self.reboot_sitl()
+        self.wait_ekf_happy()
+        self.wait_text("ArduPilot Ready", check_context=True)
+        self.wait_text("Arming Checks .* loaded", timeout=30, check_context=True, regex=True)
+
+        self.start_subsubtest("ArmCk: MAV_SYSID not set")
+        self.progress("Currently SYSID is %f" % self.get_parameter('MAV_SYSID'))
+        self.wait_text("ArmCk: warn: MAV_SYSID not set", timeout=30, check_context=True, regex=True)
+        """ disable the SYSID check, since autotest doesn't like changing the sysid"""
+        self.set_parameters({
+            "ARM_SYSID": -1,
+            "ARM_FOLL_SYSID": -1,
+            "ARM_FOLL_SYSID_X": -1,
+            "ARM_FOLL_OFS_DEF": -1,
+            "ARM_P_SCALING": -1,
+        })
+
+        self.start_subsubtest("ArmCk: Rally Point must be < ARM_V_RALLY_MAX meters away")
+        self.progress("Currently RALLY_LIMIT_KM is %f" % self.get_parameter('RALLY_LIMIT_KM'))
+        loc = self.home_relative_loc_ne(6500, -50)
+        self.upload_rally_points_from_locations([loc])
+        self.wait_text("warn: Rally too far", check_context=True)
+        self.set_parameter("RALLY_LIMIT_KM", 7)
+        self.wait_text("clear: Rally too far", check_context=True)
+
     def tests(self):
         '''return list of all tests'''
         ret = []
@@ -7674,6 +7882,9 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
             self.RudderArmingWithArmingChecksZero,
             self.TerrainLoiterToCircle,
             self.FenceDoubleBreach,
+            self.ScriptedArmingChecksApplet,
+            self.ScriptedArmingChecksAppletEStop,
+            self.ScriptedArmingChecksAppletRally,
         ]
 
     def disabled_tests(self):
--- a/Tools/autotest/quadplane.py
+++ b/Tools/autotest/quadplane.py
@@ -2752,6 +2752,36 @@ class AutoTestQuadPlane(vehicle_test_suite.TestSuite):
         self.set_rc(4, 1500)
         self.disarm_vehicle()
 
+    def ScriptedArmingChecksApplet(self):
+        """ Applet for Arming Checks will prevent a vehicle from arming based on scripted checks
+            """
+        self.start_subtest("Scripted Arming Checks Applet validation")
+        self.context_collect("STATUSTEXT")
+
+        applet_script = "arming-checks.lua"
+        """Initialize the FC"""
+        self.set_parameter("SCR_ENABLE", 1)
+        self.install_applet_script_context(applet_script)
+        self.reboot_sitl()
+        self.wait_ekf_happy()
+        self.wait_text("ArduPilot Ready", check_context=True)
+        self.wait_text("Arming Checks .* loaded", timeout=30, check_context=True, regex=True)
+        '''self.install_messageprinter_handlers_context(['PARAM_VALUE'])'''
+
+        self.start_subsubtest("ArmCk: Q_RTL_ALT must be legal")
+        self.set_parameter("SCALING_SPEED", 22)
+        self.set_parameter("Q_RTL_ALT", 150)
+        self.assert_prearm_failure("ArmCk: fail: Q_RTL_ALT too high", other_prearm_failures_fatal=False)
+        self.set_parameter("Q_RTL_ALT", 120)
+        self.wait_text("clear: Q_RTL_ALT", check_context=True)
+
+        self.start_subsubtest("ArmCk: Q_RTL vs QLand")
+        ''' this is only a warning'''
+        self.set_parameter("Q_OPTIONS", 33)
+        self.wait_text("ArmCk: note: Q will RTL", check_context=True)
+        self.set_parameter("Q_OPTIONS", 1)
+        self.wait_text("ArmCk: note: Q will land", check_context=True)
+
     def tests(self):
         '''return list of all tests'''
 
@@ -2811,5 +2841,6 @@ class AutoTestQuadPlane(vehicle_test_suite.TestSuite):
             self.CruiseRecovery,
             self.RudderArmedTakeoffRequiresNeutralThrottle,
             self.RudderArmingWithARMING_CHECK_THROTTLEUnset,
+            self.ScriptedArmingChecksApplet,
         ])
         return ret
--- a/libraries/AP_Scripting/applets/arming-checks.lua
+++ b/libraries/AP_Scripting/applets/arming-checks.lua
@@ -1,0 +1,633 @@
+-- This script runs arming checks for validations that are important
+-- to some, but might need to behave differently depending on the vehicle or use case
+-- so we don't want to bake them into the firmware. 
+-- The severity of each check can be set by simply setting the value of parameters.
+-- from warnings, which still allow arming to to errors failsafe which prevent arming 
+-- Set the value of the matching parameter to the MAV_SEVERITY of the error level you want.
+--
+-- New checks can optionally be added, by adding a true/false function (false = fail)
+-- and adding a check to the arming_checks table. 
+--
+-- Thanks to @yuri_rage and Peter Barker for help with the Lua and Autotests
+
+SCRIPT_VERSION = "4.7.0-021"
+SCRIPT_NAME = "Arming Checks"
+SCRIPT_NAME_SHORT = "ArmCk"
+
+REFRESH_DELAY = 500      -- wait this many milliseconds between each update loop (while not armed)
+INITIAL_DELAY = 20000   -- wait 20 seconds for the AP to settle down before starting to show messages
+
+GLOBAL_PARAM_TABLE_BASE = 170
+VALUES_PARAM_TABLE_KEY = GLOBAL_PARAM_TABLE_BASE + 9
+VALUES_PARAM_TABLE_PREFIX = "ARM_V_"
+
+FIRMWARE = {GLOBAL = 0, ROVER = 1, COPTER = 2, PLANE = 3, ANTENNA = 4, SUB = 7, BLIMP = 12, HELI = 13 }
+VEHICLE = {
+    -- these table key comments are so that a grep on PARAM_TABLE_KEY will find all the keys used by this script
+    -- PARAM_TABLE_KEY = 170
+    [FIRMWARE.GLOBAL]   = {prefix = "ARM_", key = GLOBAL_PARAM_TABLE_BASE},
+    -- PARAM_TABLE_KEY = 171
+    [FIRMWARE.ROVER]    = {prefix = "ARM_R_", key = GLOBAL_PARAM_TABLE_BASE + 1},
+    -- PARAM_TABLE_KEY = 172
+    [FIRMWARE.COPTER]   = {prefix = "ARM_C_", key = GLOBAL_PARAM_TABLE_BASE + 2},
+    -- PARAM_TABLE_KEY = 173
+    [FIRMWARE.PLANE]    = {prefix = "ARM_P_", key = GLOBAL_PARAM_TABLE_BASE + 3},
+    -- PARAM_TABLE_KEY = 174
+    [FIRMWARE.ANTENNA]  = {prefix = "ARM_A_", key = GLOBAL_PARAM_TABLE_BASE + 4},
+    -- PARAM_TABLE_KEY = 175
+    [FIRMWARE.BLIMP]    = {prefix = "ARM_B_", key = GLOBAL_PARAM_TABLE_BASE + 5},
+    -- PARAM_TABLE_KEY = 176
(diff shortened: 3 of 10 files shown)


libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,441 @@
...
+
+-- Arming checks can be deleted if not required, or set the parameter to MAV_SEVERITY.NONE to avoid changing the script
+-- or new arming checks added 
+local arming_checks = {
+    -- __call = function(cls, func, param_name, text, severity, passed, changed) -- constructor
+--[[
+    // @Param: ZAR_SYSID
+    // @DisplayName: MAV_SYSID must be set
+    // @Description: Check that MAV_SYSID (or SYDID_THISMAV) has been set. 3 or less to prevent arming. -1 to disable.
+    // @Values: 0:Emergency(PreArm),1:Alert(PreArm),2:Critical(PreArm),3:Error(PreArm),4:Warning,5:Notice,6:Info,7:Debug,-1:Disabled
```
rmackay9, line 255: maybe we should put -1 at the front of the list so it's in numerical order?

rmackay9 (changes requested):
Thanks for this, just one thing, I guess this is Plane specific?  If "yes", could the script be renamed to be prefixed with "plane-"?

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: Well I am hoping to simply extend it to do other vehicles, rather than having different scripts. I'm expecting there will be a lot of common checks. That's my theory anyway.

timtuxworth:
> Thanks for this, just one thing, I guess this is Plane specific? If "yes", could the script be renamed to be prefixed with "plane-"?

Well I am hoping to simply extend it to do other vehicles, rather than having different scripts. I'm expecting there will be a lot of common checks. That's my theory anyway.

rmackay9:
Hi @timtuxworth,

OK great, if it's meant to work on all vehicles, could you test this by running it on Copter (and perhaps Rover if you have time)?  I suspect the calls to get the Q_ENABLE parameter will mean the script errors out on copter

Tools/autotest/arduplane.py
```diff
@@ -7232,6 +7232,121 @@ def Volz(self):
...
+            self.wait_text("Cleared: Scaling spd not close", check_context=True)
+
+            self .start_subsubtest("ArmCk: FOLL_SYSID must be set if FOLL_ENABLE = 1")
+            self.set_parameter("FOLL_ENABLE", 1)
+            self.assert_prearm_failure("FOLL_SYSID not set", other_prearm_failures_fatal=False)
+            self.set_parameter("FOLL_SYSID", 3)
+            self.wait_text("Cleared: FOLL_SYSID not set", check_context=True)
+
+            ''' Need a healthy camera mount defined for this to work '''
+            self .start_subsubtest("ArmCk: Mount SYSID must be match FOLL_SYSID")
```
Ryanf55, line 7283:
```suggestion
            self.start_subsubtest("ArmCk: Mount SYSID must be match FOLL_SYSID")
```
Extra space?

Ryanf55:
Very nice Tim!

Tools/autotest/arduplane.py
timtuxworth: fixed

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,440 @@
...
+    SCALING_SPEED = Parameter("SCALING_SPEED")
+end
+
+-- this is the pre 4.7 name for this parameter
+local function try_sysid_thismav()
+    MAV_SYSID = Parameter("SYSID_THISMAV")
+end
+if not pcall(try_sysid_thismav) then
+    MAV_SYSID = Parameter("MAV_SYSID")
+end
```
IamPete1, line 95: If you init without a name you don't get the error if the param is not found, so you don't need the pcall. 
```
MAV_SYSID = Parameter()
if not MAV_SYSID:init("MAV_SYSID") then
    assert(MAV_SYSID:init("SYSID_THISMAV"), "SYSID not found")
end
```

Result is the same, just this might be a little neater.

timtuxworth: done - thanks, I didn't know that.

Tools/autotest/arduplane.py
```diff
@@ -7232,6 +7232,136 @@ def Volz(self):
...
+            "RC6_OPTION": 213,  # MOUNT1_PITCH
+            "SERVO%u_FUNCTION" % roll_servo: 8, # roll
+            "SERVO%u_FUNCTION" % pitch_servo: 7, # pitch
+            "SERVO%u_FUNCTION" % yaw_servo: 6, # yaw
+        })
+
+    def ScriptedArmingChecksApplet(self):
+        """ Applet for Arming Checks will prevent a vehicle from arming based on scripted checks
+            """
+        self .start_subtest("Scripted Arming Checks Applet validation")
```
peterbarker, line 7251:
```suggestion
        self.start_subtest("Scripted Arming Checks Applet validation")
```

A few places like this

libraries/AP_Scripting/applets/arming-checks.md
```diff
@@ -0,0 +1,64 @@
...
+
+The SCALING_SPEED should be close to AIRSPEED_CRUISE. Its quite easy to remember to set AIRSPEED_CRUISE correctly
+when setting up an aircraft, but not to reset (and possibly retune) SCALING_SPEED. This gives a warning if the values are very different.
+
+# Operation
+
+Copy this file to the APM/scripts folder on your SD card and make sure 
+
+SCR_ENABLE = 1
+SCR_HEAPSIZE = 250000 (minimum)
```
tridge, line 62: no way it is 250k

Tools/autotest/arduplane.py
```diff
@@ -7232,6 +7232,136 @@ def Volz(self):
...
+        self.context_push()
+        self.context_collect("STATUSTEXT")
+        ex = None
+
+        applet_script = "arming-checks.lua"
+        try:
```
peterbarker, line 7257: Don't need this try block

Tools/autotest/quadplane.py
```diff
@@ -2633,6 +2633,50 @@ def terrain_height_range(mav, m):
...
 
+    def ScriptedArmingChecksApplet(self):
+        """ Applet for Arming Checks will prevent a vehicle from arming based on scripted checks
+            """
+        self .start_subtest("Scripted Arming Checks Applet validation")
+        self.context_push()
+        self.context_collect("STATUSTEXT")
+        ex = None
+
+        applet_script = "arming-checks.lua"
```
peterbarker, line 2644: Don't need this try block

Tools/autotest/quadplane.py
```diff
@@ -2633,6 +2633,50 @@ def terrain_height_range(mav, m):
...
+        try:
+            """Initialize the FC"""
+            self.set_parameter("SCR_ENABLE", 1)
+            self.install_applet_script(applet_script)
```
peterbarker, line 2648:
```suggestion
            self.install_applet_script_context(applet_script)
```

Tools/autotest/quadplane.py
```diff
@@ -2633,6 +2633,50 @@ def terrain_height_range(mav, m):
...
+            self.wait_text("Cleared: RTL_ALT", check_context=True)
+
+            self.start_subsubtest("ArmCk: Q_RTL vs QLand")
+            ''' this is only a warning'''
+            self.set_parameter("Q_OPTIONS", 33)
+            self.wait_text("ArmCk: Q will RTL", check_context=True)
+            self.set_parameter("Q_OPTIONS", 1)
+            self.wait_text("ArmCk: Q will land", check_context=True)
+
+            self .start_subsubtest("Scripted Arming Checks complete")
```
peterbarker, line 2669: Yeah, not a subtest :-)

Tools/autotest/arduplane.py
```diff
@@ -7232,6 +7232,136 @@ def Volz(self):
...
+            "SERVO%u_FUNCTION" % roll_servo: 8, # roll
+            "SERVO%u_FUNCTION" % pitch_servo: 7, # pitch
+            "SERVO%u_FUNCTION" % yaw_servo: 6, # yaw
+        })
+
+    def ScriptedArmingChecksApplet(self):
+        """ Applet for Arming Checks will prevent a vehicle from arming based on scripted checks
+            """
+        self .start_subtest("Scripted Arming Checks Applet validation")
+        self.context_push()
```
peterbarker, line 7252: If you only ever pop this off at the end of the test then you don't need this; you get an implicit context for your test.

Tools/autotest/arduplane.py
```diff
@@ -7232,6 +7232,136 @@ def Volz(self):
...
+        self.context_collect("STATUSTEXT")
+        ex = None
+
+        applet_script = "arming-checks.lua"
+        try:
+            """Initialize the FC"""
+            self.set_parameter("SCR_ENABLE", 1)
+            self.install_applet_script(applet_script)
```
peterbarker, line 7260:
```suggestion
            self.install_applet_script_context(applet_script)
```

peterbarker (changes requested):

rmackay9 (approved):
I think it's good that we have something like this especially considering @timtuxworth is planning to maintain it

timtuxworth:
Oh @peterbarker this is failing FRSky SPort autotest which has nothing to do with the changes in this PR. Can you help with that?

Tools/autotest/arduplane.py
timtuxworth: OIC I didn't know about that - also in quadplane.py - making the change.

Tools/autotest/quadplane.py
timtuxworth: True but it makes it much easier to find things in the autotest output. :-) Ok removed.

Tools/autotest/quadplane.py
timtuxworth: removed

Tools/autotest/arduplane.py
timtuxworth: fixed

libraries/AP_Scripting/applets/arming-checks.md
timtuxworth: I don't know if it needs the minimum changed. the autotests don't change this value and they pass. so I've just removed this line.

Tools/autotest/arduplane.py
timtuxworth: I copied this from somewhere else. Removed. thanks.

Georacer:
This is a very interesting pattern @timtuxworth !

I think that might be the most "meta" Lua I've seen in ArduPilot.
Perhaps a bit overkill just to check for the arming checks, but very scalable.

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,462 @@
...
+                        check_passed = true
+                    end
+                end
+                validated = validated and check_passed
+            else -- if nothing changed we still want to fail on errors
+                if not check_passed then -- it seems like logically "and" will work but it doesn't
+                    if param_severity ~= MAV_SEVERITY.NONE and param_severity <= MAV_SEVERITY.ERROR then
+                        arming:set_aux_auth_failed(arm_auth_id, string.format('%s: fail: %s', SCRIPT_NAME_SHORT, check.text))
+                        validated = false
+                    end
```
tridge: not a blocker, but 6 levels of if nesting is an indication of poor structure

libraries/AP_Scripting/applets/arming-checks.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Parameters
+
+There is a single parameter defined for each arming check. The error or warning level can be changed for each arming check by setting the MAV_SEVERITY of the arming check. An arming check can be disabled by setting the parameter value to MAV_SEVERITY.NONE or -1.
+
+All arming check parameters are prefixed with ZAR_
+Z = Scripting Parameter
+AR = Arming Checks
+
+# ZAR_SYSID
```
tridge: I'm not a fan of the ZAR prefix, maybe ACHK_ ?

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,462 @@
...
+-- Thanks to @yuri_rage and Peter Barker for help with the Lua and Autotests
+
+SCRIPT_VERSION = "4.7.0-008"
+SCRIPT_NAME = "Arming Checks"
+SCRIPT_NAME_SHORT = "ArmCk"
+
+REFRESH_RATE = 1000
+INITIAL_DELAY = 20000   -- wait 20 seconds for the AP to settle down before starting to show messages
+
+ALTITUDE_LEGAL_MAX = 120 -- max legal altitude is 120m by default, change if different limits apply
```
peterbarker: parameter?

timtuxworth: Good idea

libraries/AP_Scripting/applets/arming-checks.md
timtuxworth: I like it - since you aren't requesting a change ...

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: Maybe for V2?

Tools/autotest/arduplane.py
```diff
@@ -7303,6 +7303,147 @@ def MAV_CMD_NAV_LOITER_TURNS_zero_turn(self):
...
+        self.set_parameters({
+            "MNT1_TYPE": 1,
+            "MNT1_PITCH_MIN": -45,
+            "MNT1_PITCH_MAX": 45,
+            "RC6_OPTION": 213,  # MOUNT1_PITCH
+            "SERVO%u_FUNCTION" % roll_servo: 8, # roll
+            "SERVO%u_FUNCTION" % pitch_servo: 7, # pitch
+            "SERVO%u_FUNCTION" % yaw_servo: 6, # yaw
+        })
+
```
peterbarker: This method was just copied over from Copter.

That's not the right way to share this code between Copter and Plane tests - it should be moved into the parent class in vehicle_test_suite.py.

Not going to block the PR based on this, but the fact I'll go and create a PR to move the code is annoying.

Tools/autotest/arduplane.py
```diff
@@ -7303,6 +7303,147 @@ def MAV_CMD_NAV_LOITER_TURNS_zero_turn(self):
...
+    def ScriptedArmingChecksApplet(self):
+        """ Applet for Arming Checks will prevent a vehicle from arming based on scripted checks
+            """
+        self.start_subtest("Scripted Arming Checks Applet validation")
+        self.context_collect("STATUSTEXT")
+
+        applet_script = "arming-checks.lua"
```
peterbarker: You don't really need a local variable for this - you're only using the thing once

Tools/autotest/arduplane.py
```diff
@@ -7303,6 +7303,147 @@ def MAV_CMD_NAV_LOITER_TURNS_zero_turn(self):
...
+        self.reboot_sitl()
+        self.wait_ekf_happy()
+        self.wait_text("ArduPilot Ready", check_context=True)
+        self.wait_text("Arming Checks .* loaded", timeout=30, check_context=True, regex=True)
+
+        self.start_subsubtest("ArmCk: MAV_SYSID not set")
+        self.progress("Currently SYSID is %f" % self.get_parameter('MAV_SYSID'))
+        self.wait_text("ArmCk: warn: MAV_SYSID not set", timeout=30, check_context=True, regex=True)
+        """ disable the SYSID check, since autotest doesn't like changing the sysid"""
+        self.set_parameter("ZAR_SYSID", -1)
```
peterbarker: I'm not really happy with "ZAR" as a parameter prefix in an applet.

Tools/autotest/arduplane.py
```diff
@@ -7303,6 +7303,147 @@ def MAV_CMD_NAV_LOITER_TURNS_zero_turn(self):
...
+        self.set_parameter("FOLL_OFS_Y", 10)
+        self.wait_text("clear: FOLL_OFS_[XYZ] = 0", check_context=True)
+        self.set_parameter("FOLL_OFS_Y", 0)
+        self.assert_prearm_failure("FOLL_OFS_[XYZ] = 0", other_prearm_failures_fatal=False)
+        self.set_parameter("FOLL_OFS_Z", 10)
+        self.wait_text("clear: FOLL_OFS_[XYZ] = 0", check_context=True)
+        """ clear these checks to make the context cleaner for subsequent tests """
+        self.set_parameter("ZAR_FOLL_SYSID", -1)
+        self.set_parameter("ZAR_FOLL_SYSID_X", -1)
+        self.set_parameter("ZAR_FOLL_OFS_DEF", -1)
```
peterbarker: You can combine these with `set_parameters` which is marginally faster

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,496 @@
...
+    SCALING_SPEED = Parameter("SCALING_SPEED")
+end
+
+-- this is the pre 4.7 name for this parameter
+MAV_SYSID = Parameter()
+if not MAV_SYSID:init("MAV_SYSID") then
+    assert(MAV_SYSID:init("SYSID_THISMAV"), string.format("%s SYSID not found", SCRIPT_NAME_SHORT ))
+end
+-- check that the MAVLink SYSID has been set for this vehicle
+local function sysid_not_set()
```
peterbarker: Comment describing the function does not match the function name.

Don't invert the function name - call this `sysid_set` and used `not` in the caller, so
```
if not sysid_set():
```

Similarly on other methods

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,496 @@
...
+end
+-- check that if Follow is enabled, the FOLL_SYSID is set
+local function foll_sysid_not_set()
+    local foll_enable = FOLL_ENABLE:get()
+    if foll_enable == 1 then
+        local foll_sysid = FOLL_SYSID:get()
+        if foll_sysid == 0 then
+            return false
+        end
+    end
```
peterbarker: Avoid this nesting - just early-return when you can come up with an answer

Same applies to many functions in here

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,496 @@
...
+            if mnt2_sysid > 0 and foll_sysid ~= mnt2_sysid then
+                return false
+            end
+        end
+    end
+    return true
+end
+
+-- is the RTL altitude legal? This defaults to 120m (400')
+local function rtl_altitude_legal()
```
peterbarker, line 220: This function uses a mixture of early-return and the deeply-nested structures.  Unroll the rest of it :-)

libraries/AP_Scripting/applets/arming-checks.md
```diff
@@ -0,0 +1,63 @@
+# arming_checks
```
peterbarker: dash or underscore?

peterbarker:
```
10:51 AM]Peter Barker: Sorry, not a quick merge from me.  We're supposed to hold applets to a reasonably high standard, and I think a little more polishing would be a good idea in that one.  It's the sort of thing that partners are going to want to customise, so getting the structure just a little bit better is warranted IMO.  Someone else may disagree and just merge it, of course.
```

Tools/autotest/arduplane.py
timtuxworth: I heard in the devcall that although there was grumbling, this would be accepted. I don't want to change it, I think it's a very good idea which is why I did it.

Tools/autotest/arduplane.py
timtuxworth: Makes sense - done.

Tools/autotest/arduplane.py
timtuxworth: II just copied this, fixed

Tools/autotest/arduplane.py
timtuxworth: thanks

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: Yes makes sense.

Tools/autotest/arduplane.py
IamPete1: The problem is that is not a pattern we can apply over all our scripts. Why no prefix with "ARMING" so these params show up with all the other arming related params.

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: fixed

libraries/AP_Scripting/applets/arming-checks.md
timtuxworth: fixed

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: Still getting used to using early returns, back in the day we were taught it was bad practice, but I know times have changed. Generally fixed.

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: I get it, but I don't want to break it by refactoring it. I'm hoping that can be fixed "later".

Tools/autotest/arduplane.py
timtuxworth: It solves a problem for me. When managing parameters I **really** want to know which parameters are scripting and which are c++, this way I can just type "Z" into the MP search and bingo - all scripting parameters show up together, not just this script - all my scripts. I use it for all my scripts and I find it works great.

I find it confusing and difficult to manage when scripting parameters show up mixed in with c++ parameters, and I would not want these parameters mixed in with other "ARMING" parameters.

Also a practical problem - if you add a scripted parameter with "ARMING" then later someone else uses the same parameter name for a c++ parameter, it will work great for people who don't use scripts, but will break when someone loads the script on. Using "Z" guarantees no clashes now or ever with c++ parameters.

I don't know why we couldn't use it as a pattern, that's exactly what I would do.

IamPete1: > "Z" guarantees no clashes now or ever with c++ parameters.

Except all the `ZIGZ_`... parameters in copter for zig-zag mode.

It certainly would have been nice to have a pattern from the start, but unfortunately we didn't do that. Currently I think this just makes it more confusing for users because they don't expect there arming check parameters to start with a `Z`. Surely the point of this applet is to seamlessly integrate with the existing cpp functionality?

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,504 @@
...
+                        check_passed = true
+                    end
+                end
+                validated = validated and check_passed
+            else -- if nothing changed we still want to fail on errors
+                if not check_passed then -- it seems like logically "and" will work but it doesn't
+                    if param_severity ~= MAV_SEVERITY.NONE and param_severity <= MAV_SEVERITY.ERROR then
+                        arming:set_aux_auth_failed(arm_auth_id, string.format('%s: fail: %s', SCRIPT_NAME_SHORT, check.text))
+                        validated = false
+                    end
```
peterbarker: Still too deeply nested.

Tools/autotest/arduplane.py
timtuxworth: > It certainly would have been nice to have a pattern from the start, but unfortunately we didn't do that.

There's no time like the present :-)

> Surely the point of this applet is to seamlessly integrate with the existing cpp functionality?

Not really. The point is to give additional capability to selectively add arming checks that aren't included in the c++, to be able to dynamically turn them on and off or have them display as warnings and for it to be easy to manage.  In fact having them segregated from other arming options makes it easier to manage, based on having used this script on my planes and copters for around 2 years already..

timtuxworth: Intesesting to note that @Hwurzburg is recommending changing a scripted "AUTOLAND" prefix in a Lua script so it's different from the c++ AUTOLAND prefixes. Having AUTOLAND for c++ and ALAND for Lua is just creating more confusion IMO.

If ZAL_ (or ZZAL) was the Lua prefix this would not be necessary.

timtuxworth: Changed.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4002,6 +4002,19 @@ function fence:get_margin_breaches() end
...
+
+-- Returns the count of the number of Rally points
+---@return integer
+function rally:get_rally_total() end
+
+-- Returns a specfic rally by index as a Location (i.e NOT as a RallyLocation, a regular Location)
+---@param index integer
+---@param rally_location Location_ud
+---@return boolean
+function rally:get_rally_location_with_index(index, rally_location) end
```
IamPete1:
```suggestion
---@return rally_location Location_ud|nil
function rally:get_rally_location_with_index(index) end
```
`Location'Null` in the bindings means the location is returned if the function returns true, else nil is returned.

timtuxworth: Thanks @IamPete1  - Almost! the return can't have a name so its

`---@return Location_ud|nil`

Which works

timtuxworth:
I think this is now "good" - @peterbarker I've removed the dynamic generation of parameter keys.

It's failing a couple of checks, but I can't see that they are down to my code, can you help me to understand if it's something I've done?

timtuxworth:
This seems to be failing a MultiGPS test in Copter @rmackay9 - I'm pretty sure I didn't change this but of course I could be wrong. I wonder if you could take a look?

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4002,6 +4002,18 @@ function fence:get_margin_breaches() end
 ---@return number -- distance
 function fence:get_breach_distance(fence_type) end
 
+-- Rally library
+rally = {}
+
+-- Returns the count of the number of Rally points
+---@return integer
+function rally:get_rally_total() end
```
IamPete1: Do you need this binding? Or could you just call get `get_rally_location_with_index` until it returns nil?

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4002,6 +4002,18 @@ function fence:get_margin_breaches() end
...
+
+-- Returns a specfic rally by index as a Location (i.e NOT as a RallyLocation, a regular Location)
+---@param index integer
+---@return Location_ud|nil
+function rally:get_rally_location_with_index(index) end
```
IamPete1:
```suggestion
function rally:get_rally_location(index) end
```
I don't think you need the `with_index` in the function name.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4002,6 +4002,18 @@ function fence:get_margin_breaches() end
...
 
+-- Rally library
+rally = {}
+
+-- Returns the count of the number of Rally points
+---@return integer
+function rally:get_rally_total() end
+
+-- Returns a specfic rally by index as a Location (i.e NOT as a RallyLocation, a regular Location)
+---@param index integer
```
IamPete1:
```suggestion
---@param index integer -- 0 indexed
```
I guess it starts at 0?

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+local function motors_emergency_stopped()
+    return not SRV_Channels:get_emergency_stop()
+end
+
+-- Logic provided by Peter Barker 
+-- Fences present if 
+-- a. there are basic fences (assume this includes min(8)/max(1) and home circle(2) fences) or 
+-- b. there are polygon_fences (fence type bit 2 = 4) with at least one side 
+local function fence_present()
+    local enabled_fences = (FENCE_TYPE:get() or 0)
```
IamPete1: Don't need the `or x` on parameters.

libraries/AP_Scripting/applets/arming-checks.md
IamPete1: I think these have all been changed now?

libraries/AP_Scripting/docs/docs.lua
IamPete1: You can rename it in the scripting bindings and keep the c++ name the same.

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+assert(param:add_table(VEHICLE[FIRMWARE.PLANE].key, VEHICLE[FIRMWARE.PLANE].prefix, 10), string.format('could not add param table: (%d) %s ', VEHICLE[FIRMWARE.PLANE].key, VEHICLE[FIRMWARE.PLANE].prefix))
+
+-- create parameter table for ARM_V_ VALUE parameters which are "reference values" used by the parameter methods
+assert(param:add_table(VALUES_PARAM_TABLE_KEY, VALUES_PARAM_TABLE_PREFIX, 15),  string.format('could not add param table: (%d) %s ', VALUES_PARAM_TABLE_KEY, VALUES_PARAM_TABLE_PREFIX))
+
+local arm_auth_id = arming:get_aux_auth_id()
+if arm_auth_id == nil then
+    gcs:send_text(MAV_SEVERITY.ERROR, string.format("%s %s Failed", SCRIPT_NAME, SCRIPT_VERSION) )
+    gcs:send_text(MAV_SEVERITY.ERROR, string.format("%s: get_aux_auth_id failed", SCRIPT_NAME_SHORT))
+    return -- don't execute the script if we can't get an auth_id
```
IamPete1, line 72: You should probably throw a error here so the scripting arming check will still prevent arming. Otherwise these checks are not run and the user can still arm I think.

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+        self.changed = false
+    else
+        self.passed = passed_new
+        self.changed = true
+    end
+    return passed_new
+end
+
+-- parameters that store values used by the validation methods, rather than severities of the individual arming checks
+ARM_V_ALT_LEGAL = Parameter()
```
IamPete1, line 107: This is never init?

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+local alt_legal_max = 120.0
+ARM_V_RALLY_MAX = Parameter()
```
IamPete1: This is never used?

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+
+    if mav_sysid == nil or mav_sysid <= 1 then
+        return false
+    end
+    return true
+end
+
+-- check that if Follow is enabled, the FOLL_SYSID is set
+local function foll_sysid_set()
+    local foll_enable = FOLL_ENABLE:get() or 0
```
IamPete1:
```suggestion
    local foll_enable = FOLL_ENABLE:get()
```

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+        end
+    end
+    return true
+end
+
+-- is the RTL altitude legal? This defaults to 120m (400') as set in ARM_ALT_LEGAL
+local function rtl_altitude_legal()
+    -- plane uses RTL_ALTITUDE in m
+    -- RTL_ALTITUDE will be correct for either plane or copter
+    if (RTL_ALTITUDE ~= nil and (RTL_ALTITUDE:get() or 300) > alt_legal_max) then
```
IamPete1:
```suggestion
    if (RTL_ALTITUDE ~= nil and RTL_ALTITUDE:get() > alt_legal_max) then
```

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+        return false
+    end
+    return true
+end
+
+-- is the Q RTL altitude (quadplane) legal? This defaults to 120m (400') as set in ARM_ALT_LEGAL
+local function q_rtl_alt_legal()
+    -- Plane specific Parameters
+    if (Q_ENABLE:get() or 0) ~= 1 then
```
IamPete1:
```suggestion
    if Q_ENABLE:get() ~= 1 then
```

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+        return true
+    end
+    local q_options = Parameter("Q_OPTIONS"):get() or 0
+    local q_land_option = ((q_options & (1 << 5)) > 0)
+    local q_RTL_option = ((q_options & (1 << 20)) > 0)
+    if q_land_option or q_RTL_option then
+        return false
+    end
+    return true -- always ok if not a QuadPlane
```
IamPete1: Comment is out of date, QP early return is now at the top.

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+end
+
+-- A single check to ensure AIRSPEED_STALL < AIRSPEED_MIN < AIRSPEED_CRUISE < AIRSPEED_MAX
+local function airspeed_check()
+    local airspeed_stall = AIRSPEED_STALL:get() or -1
+    local airspeed_min = AIRSPEED_MIN:get() or -1
+    local airspeed_cruise = AIRSPEED_CRUISE:get() or -1
+    local airspeed_max = AIRSPEED_MAX:get() or -1
```
IamPete1: Don't need the `or`

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+        if airspeed_min < airspeed_stall then
+            return false
+        end
+    end
+    if airspeed_cruise < airspeed_min then
+        return false
+    end
+    if airspeed_max < airspeed_cruise then
+        return false
+    end
```
IamPete1, line 296: I thought we had plane checks for this stuff, if not then this might be worth adding to the c++.

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,620 @@
...
+        if airspeed_min > (airspeed_stall * 1.25) then
+            return false
+        end
+    end
+
+    return true
+end
+
+-- From the wiki: This is the center of the speed scaling range and should be set close to the normal cruising speed of the vehicle.
+local function scaling_speed_check()
```
IamPete1, line 315: I'm not a fan of this one. The wiki is correct, however if you change the scaling speed you have to re-tune. We don't want people to start using this script, get the arming check and then just change it in the field without knowing they have to retune.

IamPete1 (changes requested):

libraries/AP_Scripting/docs/docs.lua
timtuxworth: I didn't know you could do that, that makes the Lua methods much nicer.

libraries/AP_Scripting/docs/docs.lua
timtuxworth: indeed

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: Not sure how this will work. If you didn't get a aux_auth_id() the script will just fail, but it can't prevent arming. It does display an error (using send_text).

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: correct -fixed.

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: ok

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: No current checks for this, interesting that this throws errors with default parameters (specifically SCALING_SPEED). could be moved to c++ later I guess.

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: Well right now the default parameters give you a tune for an invalid SCALING_SPEED anyway. At least it should prompt the user to read the wiki. I've updated the .md file to say "The aircraft should be retuned if SCALING_SPEED is changed"

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: fixed

libraries/AP_Scripting/applets/arming-checks.lua
IamPete1: If the script throws a lua error then the scripting checks will prevent arming due to a script error. EG add:
```
error("Failed to get auth ID")
```
before the return.

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: removed

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: Yes but it fixes the language check complaining about the possible Nil return.

libraries/AP_Scripting/docs/docs.lua
timtuxworth: removed

libraries/AP_Scripting/applets/arming-checks.lua
IamPete1: If you use `Parameter(name)` then it should not complain about Nil. The complaint should only be if you do `Parameter():init(name)`

timtuxworth: Got it - fixed.

libraries/AP_Scripting/docs/docs.lua
```diff
@@ -4007,6 +4007,13 @@ function fence:get_margin_breaches() end
 ---@return number -- distance
 function fence:get_breach_distance(fence_type) end
 
+-- Rally library
+rally = {}
+-- Returns a specfic rally by index as a Location (i.e NOT as a RallyLocation, a regular Location)
```
IamPete1:
```suggestion

-- Returns a specific rally by index as a Location
```
Script writer does not care about RallyLocation if its not available.

libraries/AP_Scripting/applets/arming-checks.lua
```diff
@@ -0,0 +1,624 @@
...
+-- Thanks to @yuri_rage and Peter Barker for help with the Lua and Autotests
+
+SCRIPT_VERSION = "4.7.0-021"
+SCRIPT_NAME = "Arming Checks"
+SCRIPT_NAME_SHORT = "ArmCk"
+
+REFRESH_DELAY = 500      -- wait this many milliseconds between each update loop (while not armed)
+INITIAL_DELAY = 20000   -- wait 20 seconds for the AP to settle down before starting to show messages
+
+GLOBAL_PARAM_TABLE_BASE = 170
```
peterbarker, line 20: Let's make sure people can find IDs used by searching for the table key string

libraries/AP_Scripting/applets/arming-checks.lua
IamPete1: Its still here?

timtuxworth: > This is never init?

wow - no it's not!

libraries/AP_Scripting/docs/docs.lua
timtuxworth: changed.

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: Added PARAM_TABLE_KEY = comments for each key used.

libraries/AP_Scripting/applets/arming-checks.lua
timtuxworth: fixed

peterbarker (approved):

timtuxworth:
What does it mean "this job failed" with no details @peterbarker ?

peterbarker:
> What does it mean "this job failed" with no details @peterbarker ?

Wish I know.  Vexing when it happens.  I haven't noticed it until recently, so maybe a change on github?

peterbarker:
@IamPete1 just noticed you have requested changes on this one!

IamPete1 (approved):

PR #30011, opened 2025-05-08
ArduSub: improve turn counting

ES-Alexander:
- cast output to float, for extra precision that is already part of the tracking
- handle an incorrect edge-case (`radians(90)` would incorrectly register as state 3 instead of 1)
- clean up the state tracking code


ArduSub/turn_counter.cpp
```diff
@@ -5,12 +5,12 @@
 // Count total vehicle turns to avoid tangling tether
 void Sub::update_turn_counter()
 {
-    // Determine state
```
Williangalvani: can you get rid of this whitespace change?

Williangalvani (changes requested):
tested and works fine

Issue #29786, opened 2025-04-15
Rover: Follow mode odd circling


stephendade:
## Bug report

When in Follow mode, if the leader vehicle is stationary for some time (20-120sec ... it can vary) the following vehicles will begin circling their offset point:

![Image](https://github.com/user-attachments/assets/52e04ac7-7e16-4a31-9a3a-e03410a40095)

Can be replicated in SITL with ``sim_vehicle.py --vehicle=Rover --console --map --count=4 --mavproxy-args='--cmd="set fwdpos True"' --auto-sysid``

Then setting for vehicles 2-4:
```
FOLL_ENABLE=1
FOLL_SYSID=1
FOLL_OFS_X=<any>
FOLL_OFS_Y=<any>
```

**Version**
Current master, SITL

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  X] Rover
[  ] Submarine

peterbarker:
This feels like the Rovers trying to get somewhere but can't turn that tightly?

rmackay9:
I imagine that it's I-term build-up so it might be resolved by perhaps not trying to move if the target is less than the turning radius from the vehicle.

timtuxworth:
If it's working the way I think it is, the code is just dropping a WP for the follow vehicle to fly to, it keeps dropping new "pins" head as the target moves, so it never catches up, but if the target stops moving the pin will stop moving and standard "loiter at WP" logic kicks in, which is a circle around the point.

On Plane, I found this behavior so annoying that I stopped using it altogether and I'm now directing the follow vehicle using heading, not WP.

stephendade:
Thanks for the ideas!

This had slipped off my radar, but I should be able to investigate further next week.

PR #30012, opened 2025-05-08
Rover: Better handling of stopped leader in Follow mode

stephendade:
Fix for #29786.

Changes to the behaviour of Follow mode to ignore small velocities.

Tested in SITL.

### DIFF
--- a/Rover/mode_follow.cpp
+++ b/Rover/mode_follow.cpp
@@ -48,8 +48,15 @@ void ModeFollow::update()
     desired_velocity_ne.x = vel_of_target.x + (dist_vec_offs.x * kp);
     desired_velocity_ne.y = vel_of_target.y + (dist_vec_offs.y * kp);
 
-    // if desired velocity is zero stop vehicle
-    if (is_zero(desired_velocity_ne.x) && is_zero(desired_velocity_ne.y)) {
+    // if the desired velocity is less than 3cm/sec, stop vehicle
+    if (desired_velocity_ne.length() < 0.03f) {
+        _reached_destination = true;
+        stop_vehicle();
+        return;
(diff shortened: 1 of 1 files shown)


dxkruse:
Hey @stephendade, thanks again for looking into this! I have yet to test it in the field, but I've noticed that even with an RTK fix my groundspeed variation can be greater than 0.03 m/s. I could see value in making this a parameter such as FOLL_SPEED_MIN. I'll try to run a build and test this change in the next couple of days.

peterbarker:
> Hey @stephendade, thanks again for looking into this! I have yet to test it in the field, but I've noticed that even with an RTK fix my groundspeed variation can be greater than 0.03 m/s. I could see value in making this a parameter such as FOLL_SPEED_MIN. I'll try to run a build and test this change in the next couple of days.

Ping - did you get any testing done on this one?

dxkruse:
@peterbarker Apologies, its only myself and my co-founder and other things ended up taking priority. This is back at the top of my board now, I just ran a build with this implementation today and will be testing on vehicle tomorrow. Thanks for following up!

dxkruse:
@peterbarker I was able to test these changes on my skid steer rover today. Both vehicles using dual GPS for yaw with RTK Fixes. Previously, when it would reach the offset distance, it would pivot turn in circles. I have seen velocity variance higher than the 3cm/s threshold in the code, and that may result in some small, instantaneous hiccups every once in a while. I can't confirm this without some further testing, but in general the changes solved my problem. I think having this value as a parameter would be extremely beneficial for testing or operating with less than an RTK Fix. Thank you @stephendade!

rmackay9 (approved):

PR #30014, opened 2025-05-08
RC_Channel:add Lowheiser metadata to aux functions

Hwurzburg:
omitted in original scripting PR

### DIFF
--- a/libraries/RC_Channel/RC_Channel.cpp
+++ b/libraries/RC_Channel/RC_Channel.cpp
@@ -211,6 +211,7 @@ const AP_Param::GroupInfo RC_Channel::var_info[] = {
     // @Values{Plane}: 108:QRTL Mode
     // @Values{Copter}: 109:use Custom Controller
     // @Values{Copter, Rover, Plane, Blimp, Sub}:  110:KillIMU3
+    // @Values{Copter, Rover, Plane, Blimp, Sub}:  111:Loweheiser starter
     // @Values{Copter,Plane,Rover,Blimp,Sub,Tracker}: 112:SwitchExternalAHRS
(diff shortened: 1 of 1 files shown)


libraries/RC_Channel/RC_Channel.cpp
```diff
@@ -211,6 +211,7 @@ const AP_Param::GroupInfo RC_Channel::var_info[] = {
     // @Values{Plane}: 108:QRTL Mode
     // @Values{Copter}: 109:use Custom Controller
     // @Values{Copter, Rover, Plane, Blimp, Sub}:  110:KillIMU3
+    // @Values{Copter, Rover, Plane, Blimp, Sub}:  111:Loweheiser starter
```
peterbarker, line 214: Sub and Blimp seem awfully unlikely :-)

peterbarker (changes requested):
Commit message needs some help

libraries/RC_Channel/RC_Channel.cpp
Hwurzburg: true, but there should be compilation guards added in that case, correct?this reflects current code capabilities

Hwurzburg:
what commit message would you like to see?

peterbarker:
> what commit message would you like to see?

Same one but without the leading colon.

One of the big disappointments in the github interface - not able to suggest correct commit messages!

libraries/RC_Channel/RC_Channel.cpp
Georacer: Check, this is the correct value.

Georacer (approved):

PR #30016, opened 2025-05-08
Bugfix/ld 19 rebased master

boa-pe:
LD06 / LD19 driver from 4.5.7 fixed with transplant from RPLidar.
* Distance samples averaging
* Ignoring measurements with low confidence

### DIFF
--- a/libraries/AP_Proximity/AP_Proximity_LD06.cpp
+++ b/libraries/AP_Proximity/AP_Proximity_LD06.cpp
@@ -23,197 +23,189 @@
  *
  */
 
-#include "AP_Proximity_config.h"
+ #include "AP_Proximity_config.h"
 
-#if AP_PROXIMITY_LD06_ENABLED
-#include "AP_Proximity_LD06.h"
-
-#define LD_START_CHAR       0x54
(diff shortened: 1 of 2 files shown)


peterbarker:
Sadly, this is a significant step backwards in behaviour for me, master vs this PR.

In master I've got the sectors stable at appropriate distances for what the sensor is seeing.  It all looks very nice.

In this PR the sectors still seem to get the correct maximum value, but the readings are unstable; the sectors are jumping between multiple different values on a sub-second (ie consecutive readings) time-frame.

I think that means there must be a logic issue here; if it was just average-vs-longest I would still expect to see stable values, just shorter than on current master?

I can create a pair of short videos to illustrate if that would help.

boa-pe:
Hmm... As I remember that was behavior of old driver without this fix. Perhaps I missed something during rebase. I need to re-test this on hw.

rmackay9:
I've created another alternative solution here https://github.com/ArduPilot/ardupilot/pull/30050.  I haven't tested the code yet though :-)

peterbarker:
We have commits in master to fix the bright-light issues, and those will appear in 4.6.1.

Thanks for working through this with us!

PR #30017, opened 2025-05-09
29836 dds gliding service

Ryanf55:


ArduPlane/AP_ExternalControl_Plane.cpp
```diff
@@ -19,4 +19,37 @@ bool AP_ExternalControl_Plane::set_global_position(const Location& loc)
...
+    Request gliding
+*/
+bool AP_ExternalControl_Plane::request_gliding(bool gliding_requested)
+{
+    // The command is only valid in guided mode.
+    if (plane.control_mode != &plane.mode_guided) {
+        return false;
+    }
+
+    plane.TECS_controller.set_gliding_requested_flag(gliding_requested);
```
peterbarker, line 51: Does this get reverted somewhere on leaving mode-guided?  You mentioned doing in it mode_guided on Discord.

Ryanf55: I have added that now (unconditionally). I don't see any reason we would persist that flag over mode switch, so it seems like a safe addition to ModeGuided.

ArduPlane/AP_ExternalControl_Plane.cpp
```diff
@@ -38,4 +38,18 @@ bool AP_ExternalControl_Plane::set_airspeed(const float airspeed)
 #endif
 }
 
+/*
+    Request gliding
+*/
+bool AP_ExternalControl_Plane::request_gliding(bool gliding_requested)
+{
+    // The command is only valid in guided mode.
```
IamPete1, line 46: Should we also check were in fixed wing flight (ie not VTOL) and that we are currently flying?

ArduPlane/AP_ExternalControl_Plane.cpp
IamPete1: Likewise what happens if we switch to a type of guided where TECS is not used?

Ryanf55: That would potentially be bad. Do you have any recommendations on how to detect that given there isn't a clear interface in ModeGuided to know if TECS is being used?

ArduPlane/AP_ExternalControl_Plane.cpp
Ryanf55: I now check quadplane is assisting. Can you give it a look?

PR #30023, opened 2025-05-09
ChibiOS: UARTDriver: init dma buffer

pshavoc:
this change initializes the UART DMA buffers properly and also respects the OPTION_NODMA_RX and OPTION_NODMA_TX options.

### DIFF
--- a/libraries/AP_HAL_ChibiOS/UARTDriver.cpp
+++ b/libraries/AP_HAL_ChibiOS/UARTDriver.cpp
@@ -107,6 +107,9 @@ _baudrate(57600)
 {
     osalDbgAssert(serial_num < UART_MAX_DRIVERS, "too many SERIALn drivers");
     serial_drivers[serial_num] = this;
+    rx_bounce_buf[0] = nullptr;
+    rx_bounce_buf[1] = nullptr;
+    tx_bounce_buf = nullptr;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


tpwrules:
Thanks for the contribution. Are you seeing an issue caused by this? ArduPilot zeros memory on allocation so member variables are assumed initialized.

pshavoc:
I was finding that the driver was always using DMA for receiving data, even when I set the `OPTION_NODMA_RX` option flag. I traced this down to [this line](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_HAL_ChibiOS/UARTDriver.cpp#L332) and found that `rx_dma_enabled` was always true. I then added the code to initialize `rx_bounce_buf` to nullptr and that fixed the problem.

tpwrules:
What board and compiler?

pshavoc:
CubeOrangePlus, arm-none-eabi-gcc (GNU Arm Embedded Toolchain 10-2020-q4-major) 10.2.1 20201103 (release)

libraries/AP_HAL_ChibiOS/UARTDriver.cpp
```diff
@@ -107,6 +107,9 @@ _baudrate(57600)
 {
     osalDbgAssert(serial_num < UART_MAX_DRIVERS, "too many SERIALn drivers");
     serial_drivers[serial_num] = this;
+    rx_bounce_buf[0] = nullptr;
```
andyp1per, line 110: Please remove these, they are not necessary

andyp1per (commented):
Good catch! Only the second change is required I believe

libraries/AP_HAL_ChibiOS/UARTDriver.cpp
pshavoc: I believe it is necessary for correctly handling the case when the user intends to use DMA. Without it, `rx_bounce_buf[0]` could be non-null. [This](https://github.com/ArduPilot/ardupilot/pull/30023/files#diff-4861edb3976315311702def7694c2084c5a027e2bdc6b730ea6dd93cb2b19ce7R322) if case would fail and therefor the `malloc_type` would be skipped. Thats def bad.

peterbarker: Unless someone is creating one of these UARTDriver objects on the stack, all member fields are guaranteed zero in the constructor.  We override `new` to get that effect.

If someone is creating one of these on the stack... well...

peterbarker (dismissed):
This looks like a useful fix.  Please remove the lines which shouldn't be needed.  If you find they *are* needed we have serious problems :-)

peterbarker:
Ping @andyp1per - we're down to the single logical change now, and I think you're rather more qualified on this one than I am :-)

rmackay9:

PR #30026, opened 2025-05-10
RC_Channel: Assign ARM to a non-latching button

muramura:
A game controller has non-latching buttons.
Assign the ARM function to a non-latching button.


LupusTheCanine:
You can use button capability of mission planner to send arm command through Mavlink.

PR #30028, opened 2025-05-10
Plane: Convert the ENUM definition into a class

muramura:
The ENUM variables have grouping names prefixed.
In this case, I think it's better to convert them into a class.


ArduPlane/Plane.h
```diff
@@ -1228,31 +1228,31 @@ class Plane : public AP_Vehicle {
     bool nav_scripting_enable(uint8_t mode) override;
 #endif
  
-    enum Failsafe_Action {
-        Failsafe_Action_None      = 0,
-        Failsafe_Action_RTL       = 1,
-        Failsafe_Action_Land      = 2,
-        Failsafe_Action_Terminate = 3,
+    enum class Failsafe_Action : int8_t {
```
peterbarker, line 1231:
```suggestion
    enum class FailsafeAction : int8_t {
```

peterbarker (changes requested):
I like it

ArduPlane/Plane.h
muramura: I agreed. I made the change.

peterbarker (approved):
No compiler-output change.

ArduPlane/Plane.h
```diff
@@ -1228,31 +1228,31 @@ class Plane : public AP_Vehicle {
...
-    enum Failsafe_Action {
-        Failsafe_Action_None      = 0,
-        Failsafe_Action_RTL       = 1,
-        Failsafe_Action_Land      = 2,
-        Failsafe_Action_Terminate = 3,
+    enum class FailsafeAction : int8_t {
+        None      = 0,
+        RTL       = 1,
+        Land      = 2,
+        Terminate = 3,
```
peterbarker, line 1235:
```suggestion
        NONE      = 0,
        RTL       = 1,
        LAND      = 2,
        TERMINATE = 3,
```

similarly for the other enumeration values.

We should be consistent with the upper-casing of the enumeration value entries, and since we are changing everything here we should do it as part of this PR.

peterbarker (changes requested):
Sorry, just noticed we should make this additional change while we're here.

Issue #29500, opened 2025-03-11
Custom Build Server failure if everything disabled


rmackay9:
While I noticed this using the custom build server, I suspect it's an AP flight code include issue

../build_src/libraries/AP_Compass/AP_Compass.cpp: In member function 'void Compass::probe_i2c_spi_compasses()':
./hwdef.h:683:37: error: 'DRIVER_LSM303D' was not declared in this scope
  683 | #define HAL_MAG_PROBE1  ADD_BACKEND(DRIVER_LSM303D, AP_Compass_LSM303D::probe(hal.spi->get_device("lsm9ds0_ext_am"),ROTATION_YAW_270))
      |                                     ^~~~~~~~~~~~~~
../build_src/libraries/AP_Compass/AP_Compass.cpp:1100:30: note: in definition of macro 'ADD_BACKEND'
 1100 |     do { if (_driver_enabled(driver_type)) { _add_backend(backend); } \
      |                              ^~~~~~~~~~~
./hwdef.h:687:28: note: in expansion of macro 'HAL_MAG_PROBE1'
  687 | #define HAL_MAG_PROBE_LIST HAL_MAG_PROBE1;HAL_MAG_PROBE2;HAL_MAG_PROBE3;HAL_MAG_PROBE4
      |                            ^~~~~~~~~~~~~~
../build_src/libraries/AP_Compass/AP_Compass.cpp:1412:5: note: in expansion of macro 'HAL_MAG_PROBE_LIST'
 1412 |     HAL_MAG_PROBE_LIST;
      |     ^~~~~~~~~~~~~~~~~~
compilation terminated due to -Wfatal-errors.

peterbarker:
@rmackay9 this comes about because the hwdef is saying, "there is a sensor of this type on this bus at this address".  So when we create the hwdef.h for the board we add lines saying, "probe using this sensor driver backend at this bus and address".

We do not check that that sensor driver backend is actually compiled into the code - we just attempt to call the method.

We could relatively easily wrap the call to the probe method in `#ifdef`s.  We have not done that to date because it is considered unlikely that the user would *not* want to compile in support for the compasses which are physcially soldered to their autopilot's main board.  So instead we throw ugly compiler errors at them.  We have special carve-outs in various sanity-checking tools because we don't have these #ifdefs (https://github.com/ardupilot/ardupilot/blob/master/Tools/autotest/test_build_options.py#L79)

[...]

PR #30030, opened 2025-05-11
hwdef: make using custom build a bit easier

tridge:
This fixes two issues users struggle with on the custom build server, where they end up with a failing build due to incomplete dependencies

### DIFF
--- a/libraries/AP_HAL/hwdef/scripts/hwdef.py
+++ b/libraries/AP_HAL/hwdef/scripts/hwdef.py
@@ -345,6 +345,7 @@ class HWDef:
             f.write(
                 '#define HAL_MAG_PROBE%u %s ADD_BACKEND(DRIVER_%s, AP_Compass_%s::%s(%s))\n'
                 % (n, wrapper, driver, driver, probe, ','.join(dev[1:])))
+            f.write(f"#undef AP_COMPASS_{driver}_ENABLED\n#define AP_COMPASS_{driver}_ENABLED 1\n")
         if len(devlist) > 0:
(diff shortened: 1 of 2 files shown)


peterbarker:
I actually was preparing to jump the other way on this - if you disable the driver then we compile but you obviously don't get that device.  This is less magic and won't make the interface misbehave in the custom build server as this one will.

Closes https://github.com/ardupilot/ardupilot/issues/29500

peterbarker:
... not compiling it in is a matter of changing the generation in hwdef.py to inlcude `#ifdefs` around the driver.

tridge:
> if you disable the driver then we don't compile it in but you obviously don't get that device

that isn't very user friendly. In this example the user would get an apj file, but it wouldn't boot (no baros):
image

peterbarker:
> > if you disable the driver then we don't compile it in but you obviously don't get that device
> 
> that isn't very user friendly. In this example the user would get an apj file, but it wouldn't boot (no baros): image

We have an option to allow booting with no baros.

I didn't say it was a good idea, but it's still better than more magic.

tridge:
> I didn't say it was a good idea, but it's still better than more magic.

it really isn't. The problem is we lack a way to have a dependency to the driver lines in the hwdef.dat and no easy way to fix that. Making it 'magically' get it right for users is what software development is for. Treating that as a bad thing is not good

magicrub (approved):
I'm not a fan of the undef but I understand the reason. I've dealt with magical drivers "just working" and I prefer that over throwing pages and pages of compiler errors. Users don't know how to fix those and they shouldn't need to.

magicrub:
Once the build server gets smart enough we can remove the undef stuff.

PR #30031, opened 2025-05-11
Don't erase firmware on ECC error using parameter area for storage

andyp1per:


tridge (approved):

rmackay9 (approved):
I think we should merge this because it's the minimal change and allows us to get 4.6 out the door

rmackay9:
I've included this in 4.6.0 (official)

PR #30032, opened 2025-05-11
SITL: Fixing arming errors in Real Flight 9.5

muramura:
I ran SITL by connecting a hexacopter in RealFlight with Mission Planner.
In SITL, a 'GYRO 0' pre-arm check error is occurring.
When I attempt to arm, a GPS health error also appears.
I want to disable arming checks for SITL.


IamPete1:
It should be possible to get realflight to pass checks, although you might have to use AHRS type 10. https://github.com/ArduPilot/ardupilot/pull/29921 should help with that.

I don't think we should turn these checks of for all of SITL as it means they cannot no longer be tested in CI.

andyp1per:
This is the same change I made that tridge didn't like. My RF gyro update fixes both of these issues - https://github.com/ArduPilot/ardupilot/pull/29921

peterbarker:
Sorry, not the right fix; we should be able to get these passing.

PR #30033, opened 2025-05-11
AP_Generator: rename GEN message to RICH message

peterbarker:


Georacer:
Could you mention, for posterity, what is your intention for this change?

Based on the code, I understand this is just for the logging part, which is in line with that we do in Loweheiser, e.g.
That sounds good.

Loweheiser also declares new `GEN_*` parameters while Richenpower doesn't have any new ones.
In your opinion, should we rename the Loweheiser- and IE-specific parameters from `GEN_` to something else?

peterbarker:
> Could you mention, for posterity, what is your intention for this change?

Yeah, just to make the message rather more specific to the relevant device.

Worst case is you have two generators and a conflict :-)


> Loweheiser also declares new `GEN_*` parameters while Richenpower doesn't have any new ones. In your opinion, should we rename the Loweheiser- and IE-specific parameters from `GEN_` to something else?

Depends on who owns the parameters - if AP_Generator owns them then `GEN_` is right.

IamPete1 (approved):
LGTM

robertlong13:
@peterbarker , I think this broke CI. Renaming the "RICH" message effectively removed it from the logger documentation greylist.

PR #30034, opened 2025-05-11
Copter: Add a cancel function for LAND mode under GNSS conditions

muramura:
Sometimes I want to increase altitude while descending in LAND mode.
I will implement the cancel function, which is available in non-GNSS environments, for GNSS environments as well.


ArduCopter/mode_land.cpp
```diff
@@ -82,8 +82,16 @@ void ModeLand::gps_run()
...
         }
 
-        // run normal landing or precision landing (if enabled)
-        land_run_normal_or_precland(land_pause);
+        if (!copter.failsafe.radio && (g.throttle_behavior & THR_BEHAVE_HIGH_THROTTLE_CANCELS_LAND) != 0 && copter.rc_throttle_control_in_filter.get() > LAND_CANCEL_TRIGGER_THR){
+            LOGGER_WRITE_EVENT(LogEvent::LAND_CANCELLED_BY_PILOT);
+            // exit land if throttle is high
+            copter.set_mode(Mode::Number::LOITER, ModeReason::THROTTLE_LAND_ESCAPE);
+
+        } else {
```
rmackay9, line 90: Thanks for this, I think we should remove this else.  I'm just slightly concerned that there could possibly be a reason that the vehicle can't enter Loiter mode but we keep trying over and over and end up never running the regular land_run_normal_or_preclan() method

rmackay9 (approved):

rmackay9:
@muramura, this has been tested in SITL of course?

muramura:
@rmackay9 san

I have confirmed the operation using SITL with RealFlight.

peterbarker (commented):
A regression test would have been good.

muramura:
@peterbarker san. 
What kind of tests should I perform?

I want to support flight controllers that do not run Lua scripts.

ArduCopter/mode_land.cpp
```diff
@@ -82,6 +82,14 @@ void ModeLand::gps_run()
             land_pause = false;
         }
 
+        if (!copter.failsafe.radio && (g.throttle_behavior & THR_BEHAVE_HIGH_THROTTLE_CANCELS_LAND) != 0 && copter.rc_throttle_control_in_filter.get() > LAND_CANCEL_TRIGGER_THR){
```
peterbarker, line 85:
```suggestion
        if (rc().has_valid_input() &&
            (g.throttle_behavior & THR_BEHAVE_HIGH_THROTTLE_CANCELS_LAND) != 0  &&
            copter.rc_throttle_control_in_filter.get() > LAND_CANCEL_TRIGGER_THR) {
```
We *must* use has_valid_input here - otherwise a receiver going bad could trigger this code and the vehicle might not land

peterbarker:
> @peterbarker san. What kind of tests should I perform?

I'd like an autotest that denies GPS to the vehicle, starts to land but then pushed the throttle up to cancel the land.  We have all of the different pieces of that in arducopter.py already, and you can just look for the cancelled landing text.

PR #30035, opened 2025-05-11
autotest: add ratchet for log message documentation

peterbarker:
this should ensure that when a log message is added then it must be documented as well

... and hopefully we can reduce the length of these lists over time

### DIFF
--- a/Tools/autotest/vehicle_test_suite.py
+++ b/Tools/autotest/vehicle_test_suite.py
@@ -3087,6 +3087,109 @@ class TestSuite(ABC):
 
         return ids
 
+    def LoggerDocumentation_greylist(self):
+        '''returns a set of messages should should be documented but
+        are currently known as undocumented'''
+        return set([
+            "FENC",  # fence
+            "FTN3",  # gyrofft
+            "GEN",   # generator
+            "IE24",  # generator
+            "IEFC",  # generator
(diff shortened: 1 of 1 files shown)


Tools/autotest/vehicle_test_suite.py
```diff
@@ -3087,6 +3087,109 @@ def all_log_format_ids(self):
...
+
+            "SMVZ",  # Sim-Volz
+
+            "SORC",  # Soaring
+            "VAR",   # soaring
+
+            "TCLR",  # tempcal
+            "TEMP",  # temperature sensor library
+
+            "ARHS",  # autorotation
```
rmackay9, line 3135: It's off topic but do we really have ARHS and AHRS?  I wonder how much we thought about this

peterbarker: > It's off topic but do we really have ARHS and AHRS? I wonder how much we thought about this

:-)  It was like that when I got here :-)

That's be Auto Rotation Head Speed, at a guess.

rmackay9: @peterbarker, yeah, I just had to say something.. marking this as "resolved"

IamPete1 (approved):
Nice!

PR #30037, opened 2025-05-11
Remove INAV from Position Control

lthall:
This PR removes INAV from the position controller and replaces many functions in the main code to directly reference the position controller state where appropriate.

```
CubeOrange,608,*,1008,1024,688,760,648
```


libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1323,14 +1322,32 @@ void AC_PosControl::get_stopping_point_U_cm(postype_t &stopping_point_u_cm) cons
...
 ///
 /// System methods
 ///
 
+
+void AC_PosControl::update_estimates(bool high_vibes)
+{
+    if (!AP::ahrs().get_relative_position_NED_origin(_pos_estimate_ned_m)) {
+        float posD;
+        if (AP::ahrs().get_relative_position_D_origin_float(posD)) {
```
IamPete1, line 1338: Isn't this the same as the pos D that returned in the vector from the previous call?

IamPete1:
On plane this make me very nervous that we will forget to call `update_estimates`, I think the current plane changes don't get them all, and going forward it will be hard to not introduce bugs.

What is the goal or removing inav? It is still used for loiter and circle so this change makes it possible this users are have a different value the pos controller.

Why not add the call `pos_control.update_estimates` where `inertial_nav.update` is called now? It would be less code, and remove the possibility of forgetting to call it, the disadvantage would be a little more computation as it would be called when its not used.

lthall:
> On plane this make me very nervous that we will forget to call `update_estimates`, I think the current plane changes don't get them all, and going forward it will be hard to not introduce bugs.
> 
> What is the goal or removing inav? It is still used for loiter and circle so this change makes it possible this users are have a different value the pos controller.
> 
> Why not add the call `pos_control.update_estimates` where `inertial_nav.update` is called now? It would be less code, and remove the possibility of forgetting to call it, the disadvantage would be a little more computation as it would be called when its not used.

You are completly correct. Doing it now.

libraries/AC_AttitudeControl/AC_PosControl.cpp
lthall: No it isn't, it has an internal redundancy.
This is something that we may be able to improve.

libraries/AC_AttitudeControl/AC_PosControl.h
```diff
@@ -53,6 +52,8 @@ class AC_PosControl
     void set_dt(float dt) { _dt = dt; }
     float get_dt() const { return _dt; }
 
+    void update_estimates(bool high_vibes = false);
```
rmackay9, line 55: we should probably add a comment above this declaration, we might need the same thing in the .cpp file as well

libraries/AC_PID/AC_P_2D.cpp
```diff
@@ -21,16 +21,15 @@ AC_P_2D::AC_P_2D(float initial_p) :
 }
 
 // update_all - set target and measured inputs to P controller and calculate outputs
-Vector2f AC_P_2D::update_all(postype_t &target_x, postype_t &target_y, const Vector2f &measurement)
+Vector2f AC_P_2D::update_all(Vector2p target, const Vector2p &measurement)
```
rmackay9, line 24: Should this be a reference like, "Vector2p& target"?  I think we need this so that we can update the target when it's dragged by the current estimate

ArduSub/mode.cpp
```diff
@@ -155,6 +155,7 @@ bool Sub::set_mode(const uint8_t new_mode, const ModeReason reason)
 // called at 100hz or more
 void Sub::update_flight_mode()
 {
+    pos_control.update_estimates();
```
rmackay9, line 158: maybe we should do the same here as we've done for plane which is update this at the same time as inertial_nav.update() is called.  Same for blimp.

lthall: I have added this as a seperate commit so both options are there for discussion.

libraries/AC_PID/AC_P_2D.cpp
lthall: Fixed.

libraries/AC_AttitudeControl/AC_PosControl.h
lthall: Done.

libraries/AC_AttitudeControl/AC_PosControl.cpp
lthall: I have tried to explain this in the comment:

// Updates internal position and velocity estimates in the NED frame.
// Falls back to vertical-only estimates if full NED data is unavailable.
// When high_vibes is true, forces use of vertical fallback for velocity.

lthall:
> > On plane this make me very nervous that we will forget to call `update_estimates`, I think the current plane changes don't get them all, and going forward it will be hard to not introduce bugs.
> > What is the goal or removing inav? It is still used for loiter and circle so this change makes it possible this users are have a different value the pos controller.
> > Why not add the call `pos_control.update_estimates` where `inertial_nav.update` is called now? It would be less code, and remove the possibility of forgetting to call it, the disadvantage would be a little more computation as it would be called when its not used.
> 
> You are completly correct. Doing it now.

This has resulted in a crash dump but I don't understand why.

ArduPlane/Plane.cpp
```diff
@@ -204,6 +204,7 @@ void Plane::ahrs_update()
 
     // update inertial_nav for quadplane
     quadplane.inertial_nav.update();
+    quadplane.pos_control->update_estimates();
```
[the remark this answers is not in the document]
lthall: Legend, Thanks. I figured it would be something like this.

lthall:
Is this the right name?

get_pos_NEU_cm
get_pos_current_NEU_cm
get_pos_estimate_NEU_cm

I think I like get_pos_estimate_NEU_cm

ArduCopter/mode.cpp
```diff
@@ -405,6 +405,7 @@ bool Copter::set_mode(const uint8_t new_mode, const ModeReason reason)
 // called at 100hz or more
 void Copter::update_flight_mode()
 {
+    pos_control->update_estimates(vibration_check.high_vibes);
```
IamPete1, line 408: It would make me slightly happier if this was exactly where the old inav update was, just so its clear there is no change. We could then move it in a future PR if we need to.

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1402,7 +1425,7 @@ void AC_PosControl::write_log()
 /// crosstrack_error - returns horizontal error to the closest point to the current track
 float AC_PosControl::crosstrack_error() const
 {
-    const Vector2f pos_error = _inav.get_position_xy_cm() - (_pos_target_neu_cm.xy()).tofloat();
+    const Vector2f pos_error = (_pos_target_neu_cm.xy() - _pos_estimate_neu_cm.xy()).tofloat();
```
IamPete1, line 1428: I think target and measurement are reversed here so result is the other sign?

libraries/AP_AHRS/AP_AHRS.h
```diff
@@ -269,17 +269,41 @@ class AP_AHRS {
     // return the relative position NED from either home or origin
     // return true if the estimate is valid
     bool get_relative_position_NED_home(Vector3f &vec) const WARN_IF_UNUSED;
-    bool get_relative_position_NED_origin(Vector3f &vec) const WARN_IF_UNUSED;
+    bool get_relative_position_NED_origin(Vector3p &vec) const WARN_IF_UNUSED;
+    bool get_relative_position_NED_origin_float(Vector3f &vec) const WARN_IF_UNUSED {
```
IamPete1, line 273: Should probably be in the cpp.

libraries/AP_AHRS/AP_AHRS.h
```diff
@@ -269,17 +269,41 @@ class AP_AHRS {
...
+        vec = tmp_posNED.tofloat();
+        return true;
+    }
 
     // return the relative position NE from home or origin
     // return true if the estimate is valid
     bool get_relative_position_NE_home(Vector2f &posNE) const WARN_IF_UNUSED;
-    bool get_relative_position_NE_origin(Vector2f &posNE) const WARN_IF_UNUSED;
+    bool get_relative_position_NE_origin(Vector2p &posNE) const WARN_IF_UNUSED;
+    bool get_relative_position_NE_origin_float(Vector2f &posNE) const WARN_IF_UNUSED {
```
IamPete1, line 286: Should probably be in the cpp.

libraries/AP_AHRS/AP_AHRS.h
```diff
@@ -269,17 +269,41 @@ class AP_AHRS {
...
+        posNE = tmp_posNE.tofloat();
+        return true;
+    }
 
     // return the relative position down from home or origin
     // baro will be used for the _home relative one if the EKF isn't
     void get_relative_position_D_home(float &posD) const;
-    bool get_relative_position_D_origin(float &posD) const WARN_IF_UNUSED;
+    bool get_relative_position_D_origin(postype_t &posD) const WARN_IF_UNUSED;
+    bool get_relative_position_D_origin_float(float &posD) const WARN_IF_UNUSED {
```
IamPete1, line 299: Should probably be in the cpp.

IamPete1 (changes requested):
In copter there are lots of things changes over to the new methods, looks like there are also a dozen or so in plane, should they be moved across too?

Is there anything using inertial_nav left in copter? Can we remove if from copter.h (and the wscript)?

ArduCopter/mode_rtl.cpp
```diff
@@ -309,7 +309,7 @@ void ModeRTL::descent_run()
     attitude_control->input_thrust_vector_heading_cd(pos_control->get_thrust_vector(), auto_yaw.get_heading());
 
     // check if we've reached within 20cm of final altitude
-    _state_complete = labs(rtl_path.descent_target.alt - copter.inertial_nav.get_position_z_up_cm()) < 20;
+    _state_complete = labs(rtl_path.descent_target.alt - pos_control->get_pos_estimate_NEU_cm().tofloat().z) < 20;
```
tridge, line 312: probably don't need tofloat() here?

rmackay9 (approved):
LGTM after the two changes from PeterH are addressed, thanks!

ArduCopter/mode_rtl.cpp
lthall: I will check if the compiler complains.

libraries/AP_AHRS/AP_AHRS.h
tridge: yep

libraries/AC_AttitudeControl/AC_PosControl.cpp
lthall: You are correct Pete. However this was incorrect before. Error should be target - current. The reason it didn't matter before is we only use the length.

lthall:
```
/// crosstrack_error - returns horizontal error to the closest point to the current track
float AC_PosControl::crosstrack_error() const
{
    const Vector2f pos_error = (_pos_target_neu_cm.xy() - _pos_estimate_neu_cm.xy()).tofloat();
```
Vector is in the opposite direction.
```
    if (is_zero(_vel_desired_neu_cms.xy().length_squared())) {
        // crosstrack is the horizontal distance to target when stationary
        return pos_error.length();
```
The vector is the same length.
```
    } else {
        // crosstrack is the horizontal distance to the closest point to the current track
        const Vector2f vel_unit = _vel_desired_neu_cms.xy().normalized();
        const float dot_error = pos_error * vel_unit;
```
The dot_error has changed sign.
```
        // todo: remove MAX of zero when safe_sqrt fixed
        return safe_sqrt(MAX(pos_error.length_squared() - sq(dot_error), 0.0));
```
We use only the pos_error length so always positive and we square dot_error so the sign doesn't matter.
```
    }
}
```

ArduCopter/mode_rtl.cpp
lthall: There were a number of places that this was done. All fixed.

ArduCopter/mode.cpp
lthall: I have done this.

libraries/AC_WPNav/AC_WPNav.cpp
```diff
@@ -409,12 +408,9 @@ void AC_WPNav::shift_wp_origin_and_destination_to_current_pos_NE()
...
     _pos_control.init_NE_controller();
 
-    // get current and target locations
-    const Vector2f& curr_pos_neu_cm = _inav.get_position_xy_cm();
-
     // shift origin and destination_neu_cm horizontally
-    _origin_neu_cm.xy() = curr_pos_neu_cm;
-    _destination_neu_cm.xy() = curr_pos_neu_cm;
+    _origin_neu_cm.xy() = _pos_control.get_pos_estimate_NEU_cm().tofloat().xy();
+    _destination_neu_cm.xy() = _pos_control.get_pos_estimate_NEU_cm().tofloat().xy();
```
IamPete1, line 413: `tofloat().xy()` -> `xy().tofloat()`

libraries/AC_WPNav/AC_WPNav.cpp
```diff
@@ -578,13 +574,13 @@ void AC_WPNav::update_track_with_speed_accel_limits()
 /// get_wp_distance_to_destination - get horizontal distance to destination_neu_cm in cm
 float AC_WPNav::get_wp_distance_to_destination_cm() const
 {
-    return get_horizontal_distance_cm(_inav.get_position_xy_cm(), _destination_neu_cm.xy());
+    return get_horizontal_distance_cm(_pos_control.get_pos_estimate_NEU_cm().tofloat().xy(), _destination_neu_cm.xy());
```
IamPete1, line 577: `tofloat().xy()` -> `xy().tofloat()`

libraries/AC_WPNav/AC_WPNav.cpp
```diff
@@ -578,13 +574,13 @@ void AC_WPNav::update_track_with_speed_accel_limits()
...
 }
 
 /// get_wp_bearing_to_destination_cd - get bearing to next waypoint in centi-degrees
 int32_t AC_WPNav::get_wp_bearing_to_destination_cd() const
 {
-    return get_bearing_cd(_inav.get_position_xy_cm(), _destination_neu_cm.xy());
+    return get_bearing_cd(_pos_control.get_pos_estimate_NEU_cm().tofloat().xy(), _destination_neu_cm.xy());
```
IamPete1, line 583: `tofloat().xy()` -> `xy().tofloat()`

libraries/AC_WPNav/AC_WPNav_OA.cpp
```diff
@@ -48,7 +48,7 @@ float AC_WPNav_OA::get_wp_distance_to_destination_cm() const
         return AC_WPNav::get_wp_distance_to_destination_cm();
     }
 
-    return get_horizontal_distance_cm(_inav.get_position_xy_cm(), _destination_oabak_neu_cm.xy());
+    return get_horizontal_distance_cm(_pos_control.get_pos_estimate_NEU_cm().tofloat().xy(), _destination_oabak_neu_cm.xy());
```
IamPete1, line 51: `tofloat().xy()` -> `xy().tofloat()`

libraries/AC_WPNav/AC_WPNav_OA.cpp
```diff
@@ -59,7 +59,7 @@ int32_t AC_WPNav_OA::get_wp_bearing_to_destination_cd() const
         return AC_WPNav::get_wp_bearing_to_destination_cd();
     }
 
-    return get_bearing_cd(_inav.get_position_xy_cm(), _destination_oabak_neu_cm.xy());
+    return get_bearing_cd(_pos_control.get_pos_estimate_NEU_cm().tofloat().xy(), _destination_oabak_neu_cm.xy());
```
IamPete1, line 62: `tofloat().xy()` -> `xy().tofloat()`

IamPete1 (approved):
I think this looks good. 

The `tofloat().xy()` -> `xy().tofloat()` thing could be a future PR, we have the same issue in a few other places not touched in this PR.

ArduCopter/mode_guided.cpp
```diff
@@ -1120,7 +1120,7 @@ float ModeGuided::wp_distance_m() const
     case SubMode::WP:
         return wp_nav->get_wp_distance_to_destination_cm() * 0.01f;
     case SubMode::Pos:
-        return get_horizontal_distance_cm(inertial_nav.get_position_xy_cm(), guided_pos_target_cm.tofloat().xy()) * 0.01f;
+        return get_horizontal_distance_cm(pos_control->get_pos_estimate_NEU_cm().tofloat().xy(), guided_pos_target_cm.tofloat().xy()) * 0.01f;
```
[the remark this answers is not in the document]
lthall: Thanks, I did wonder at the time. I have fixed this where ever I have found it but I think you pointed out most of them below.

rmackay9:
I've successfully tested this on an EDU450 with CubeOrangePlus in stabilize, AltHold, Loiter and Auto modes without any issues

rmackay9:
I plan to merge this within 24hrs, please tell me if anyone objects!

PR #25203, opened 2023-10-07
Copter: choose RC channel for Transmitter Based Tuning

be-ska:
This PR add the ability to choose which RC channel controls the Transmitter Based Tuning, instead of the hard coded channel 6.
If a user want to use channel x for transmitter based tuning, the parameter RCx_OPTION should be set to 177 (Transmitter Tuning), as discussed in issue #25186.

The changes has been tested in SITL after setting parameters:
- `TUNE` 21
- `TUNE_MIN` 0
- `TUNE_MAX` 0.03
- `RC9_OPTION` 177

Changing the PWM value of CH8 results in different `ATC_RAT_RLL_D` values.

PR #30038, opened 2025-05-11
Copter: correct setting of deadzone on RC tuning channel 

peterbarker:
We lost this when moving from fixed-CH_6 to RC_OPTION=219 in https://github.com/ArduPilot/ardupilot/pull/25203/files

Without this the channel may have a deadzone, meaning the tuning knob won't be entirely linear across its range.  From a user perspectiv they might push the value up and up, then find it doesn't move for a while... and then move a lot.  Impatient users might blast past the deadzone when nothing is happening...

Behavioural change here: making a channel transmitter-tuning now requires a reboot for it to take effect.  This is so we set the deadzone in just one place.  Live-resetting the DZ on any channel you set to 219 seems like it could be unfortunate, and Copter doesn't update channel except on reboot (unlike Plane and Rover)


ArduCopter/tuning.cpp
```diff
@@ -9,8 +9,6 @@
 //  should be called at 3.3hz
 void Copter::tuning()
 {
-    const RC_Channel *rc_tuning = rc().find_channel_for_option(RC_Channel::AUX_FUNC::TRANSMITTER_TUNING);
```
rmackay9, line 12: I guess a simpler solution might have been to add a call here to set_default_dead_zone() but that would be inconsistent with how we do it for other channels

peterbarker: > I guess a simpler solution might have been to add a call here to set_default_dead_zone() but that would be inconsistent with how we do it for other channels

Yep, didn't want to answer the question, "why didn't you put it into `set_default_deadzones, that seems the right place?"...

As mentioned elsewhere, having the user need to reboot isn't a bad thing IMO.  If the user were to accidentally set the channel option to their roll channel and then away again then we'd now have a zero-default-deadzone on their roll channel.... and they could arm/takeoff in that state.

ArduCopter/Parameters.cpp
```diff
@@ -864,13 +864,13 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
...
     // @DisplayName: Tuning minimum
-    // @Description: Minimum value that the parameter currently being tuned with the transmitter's channel 6 knob will be set to
+    // @Description: Minimum value that the parameter currently being tuned with the transmitter's TRANSMITTER_TUNING knob will be set to
     // @User: Standard
     AP_GROUPINFO("TUNE_MIN", 31, ParametersG2, tuning_min, 0),
 
     // @Param: TUNE_MAX
     // @DisplayName: Tuning maximum
-    // @Description: Maximum value that the parameter currently being tuned with the transmitter's channel 6 knob will be set to
+    // @Description: Maximum value that the parameter currently being tuned with the transmitter's TRANSMITTER_TUNING knob will be set to
```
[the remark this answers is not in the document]
peterbarker: I was just doing more of what was in the originally accepted PR when we moved off CH6, I think.

I've added a commit using your words here.

tridge:
@peterbarker why not use get_control_in_zero_dz() ?

peterbarker:
> @peterbarker why not use get_control_in_zero_dz() ?

Changed to use that method.

ArduCopter/radio.cpp
```diff
@@ -32,6 +31,11 @@ void Copter::init_rc_in()
     channel_yaw->set_angle(ROLL_PITCH_YAW_INPUT_MAX);
     channel_throttle->set_range(1000);
 
+    rc_tuning = rc().find_channel_for_option(RC_Channel::AUX_FUNC::TRANSMITTER_TUNING);
+    if (rc_tuning != nullptr) {
+        rc_tuning->set_range(1);
+    }
```
peterbarker:
```suggestion
```

ArduCopter/tuning.cpp
```diff
@@ -26,9 +24,9 @@ void Copter::tuning()
         return;
     }
 
-    const uint16_t radio_in = rc_tuning->get_radio_in();
-    float tuning_value = linear_interpolate(g2.tuning_min, g2.tuning_max, radio_in, rc_tuning->get_radio_min(), rc_tuning->get_radio_max());
-    
+    const float control_in = rc_tuning->get_control_in_zero_dz();
```
peterbarker:
```suggestion
    const float control_in = rc_tuning->get_control_in_zero_dz();
```
norm_input_ignore_trim

rmackay9 (approved):

PR #30041, opened 2025-05-12
AP_Landing: fixed assumption of AMSL WPs

tridge:
the AP_Landing library assumed AMSL waypoints for both prev_WP_loc and next_WP_loc. This led to an incorrect glide slope for a NAV_LAND fixed wing landing when the landing WP is terrain relative.

This change fixes the code to handle either an AGL or AMSL location, which are the only two location types used in plane

fixes an issue in 4.6.0-beta reported here: https://discuss.ardupilot.org/t/plane-4-6-0-beta/126358/183

tested in SITL

### DIFF
--- a/libraries/AP_Landing/AP_Landing.cpp
+++ b/libraries/AP_Landing/AP_Landing.cpp
@@ -21,6 +21,7 @@
 #include <GCS_MAVLink/GCS.h>
 #include <AP_AHRS/AP_AHRS.h>
 #include <AC_Fence/AC_Fence.h>
+#include <AP_InternalError/AP_InternalError.h>
 
 // table of user settable parameters
 const AP_Param::GroupInfo AP_Landing::var_info[] = {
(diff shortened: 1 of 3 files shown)


libraries/AP_Landing/AP_Landing.cpp
```diff
@@ -207,6 +208,28 @@ AP_Landing::AP_Landing(AP_Mission &_mission, AP_AHRS &_ahrs, AP_TECS *_tecs_Cont
...
+{
+    int32_t alt_cm;
+    // try first with full conversion
+    if (loc.get_alt_cm(Location::AltFrame::ABSOLUTE, alt_cm)) {
+        return alt_cm;
+    }
+    if (loc.get_alt_frame() == Location::AltFrame::ABOVE_TERRAIN) {
+        // if we can't get true terrain then assume flat terrain
+        // around home
+        return loc.alt + ahrs.get_home().alt;
```
peterbarker, line 225: Would be nice to not reference the private-in-the-future `.alt` field directly in new code.

peterbarker (approved):
Looks right.

I don't think this is a regression in 4.6, is it?

... but really should have an autotest.

tridge:
@peterbarker it is a regression from 4.5.x

rmackay9:
This is included in 4.6.0 (stable) as part of PR https://github.com/ArduPilot/ardupilot/pull/30045

Issue #2125, opened 2015-04-15
Copter: Smarter CH6 tuning


ghost:
For in flight tuning it is often essential to change two parameters simultaneously. For example:
- Rate Roll P AND Rate Roll I.
- Rate Pitch P AND Rate Pitch I.
- Throttle Accel P AND Throttle Accel I (i.e. "Climb Sensitivity" from Basic Tuning).

I think, it would be great to tune both of these values simultaneously via CH6. Perhaps with two different MIN/MAX values.

Actually this is already possible for some values: Rate Roll P and Rate Pitch P (or I, or D), for example. So why not offer more of these "connected" CH6 options?

peterbarker:
This isn't a great issue as it boils down to "do things better" - there's no way this issue can be closed as "complete".

I've created https://github.com/ArduPilot/ardupilot/pull/30042 which will add the flexibility to add each of the separate tuning values onto different knobs on the transmitter, allowing for simultaneous (albeit more-manual) tuning of different values.

I intend to close this PR - if there are concrete suggestions of parameter-pairs that should be tuned on the same knob then please open fresh PRs with them nominated.

rmackay9:
Thanks for the follow-up @peterbarker and I've commented on the PR but the regular transmitter interface is very limited and I'm not sure it's worth trying to extend the RC interface with more features like this

PR #30042, opened 2025-05-12
Add support for second transmitter-tuning control channel

peterbarker:
~~There are several pre-cursor PRs to merge before this one.~~

Closes https://github.com/ardupilot/ardupilot/issues/2125


libraries/RC_Channel/RC_Channel.h
```diff
@@ -283,6 +283,8 @@ class RC_Channel {
         MOUNT2_YAW =         217, // mount4 yaw input
         LOWEHEISER_THROTTLE= 218, // allows for throttle on slider
         TRANSMITTER_TUNING = 219, // use a transmitter knob or slider for in-flight tuning
+        TRANSMITTER_TUNING2 = 220, // use another transmitter knob or slider for in-flight tuning
+        TRANSMITTER_TUNING3 = 221, // use yet another transmitter knob or slider for in-flight tuning
```
rmackay9, line 287: I guess we don't need TRANSMITTER_TUNING3?

peterbarker: > I guess we don't need TRANSMITTER_TUNING3?

I decided to reserve an additional value just to keep things consecutive if someone *did* want another one.

rmackay9: .. until muramura comes along in 5 minutes and removes it :-).  We have a golden rule in AP that we don't think about the future too much because it's hard to predict so let's remove it.  users don't care about the numbers much anyway now that we alphabetise the param values that are shown in the GCSs

peterbarker:
> Txs, the issue linked is from 2015 though so I wonder if perhaps things haven't moved on so much as to make this unnecessary. Back in those days we had limited interface options for most users (e.g. less users of telemetry, no lua scripting, less sophisticated GCSs).
> 
> I'm not blocking this but I think we should carefully consider whether we really want this

I believe people still do actually use these.

image

libraries/RC_Channel/RC_Channel.h
peterbarker: > .. until muramura comes along in 5 minutes and removes it :-). We have a golden rule in AP that we don't think about the future too much because it's hard to predict so let's remove it. users don't care about the numbers much anyway now that we alphabetise the param values that are shown in the GCSs

Fair enough, I've removed it.

Note I'm not wedded to adding this second channel, but it's always vexed me that there was only one (and that it was fixed at CH_6, of course ;-) )

rmackay9: @peterbarker, thanks for your efforts.  If you really want it then I'm happy to go along with that.  Can I ask though when was the last time that you (or anyone else on the CanberraUAV team) actually wanted to manually tune a copter from the transmitter?  I personally never tune using transmitter knobs anymore, I always just do the updates from the GCS.

peterbarker: > @peterbarker, thanks for your efforts. If you really want it then I'm happy to go along with that. Can I ask though when was the last time that you (or anyone else on the CanberraUAV team) actually wanted to manually tune a copter from the transmitter? I personally never tune using transmitter knobs anymore, I always just do the updates from the GCS.

Weelll.... I don't think the CanberraUAV people are typical fliers :-)

I don't think I've personally used this stuff in a *very* long time.  Mostly because I *do* usually have a friendly GCS operator to change values while I'm flying.

I have seen tridge use transmitter-based tuning within the last couple of years - but not this sort.  The sort that's only present in Plane and allows you to rotate through PID values using a switch.

If we want to start to ease this out of the code instead we can do that.  I think this would make sense as a build-server option anyway.  Still useful on 1MB boards but not vital.

This PR costs 136 bytes on CubeOrange.

rmackay9: ok, so the short answer is that even CanberraUAV won't use this feature.  If we have time, let's ask at the dev call what other devs think.

peterbarker: ... compiling the feature out saves 1,400 bytes.  I've added defines for that in this PR now.

ArduCopter/tuning.cpp
```diff
@@ -9,29 +10,51 @@
...
-    // exit immediately if the tuning function is not set or min and max are both zero
-    if ((g.rc_tuning_param <= 0) || (is_zero(g2.tuning_min.get()) && is_zero(g2.tuning_max.get()))) {
+
+    if (is_equal(tuning_min, tuning_max)) {
+        // endpoints are equal, there is no input range to tune across
         return;
     }
 
-    // exit immediately when radio failsafe is invoked or transmitter has not been turned on
-    if (!rc().has_valid_input() || rc_tuning->get_radio_in() == 0) {
```
peterbarker, line 23: The zero-input-value-check has been lost here and is a little nasty if we're using the normalised-input check

peterbarker: I've fixed this so we check that the normalised input value is valid before we use it.

PR #30046, opened 2025-05-12
Backport-4.6: disallow setting of readonly/internal parameters via mavftp

peterbarker:


rmackay9:
Sorry for whatever reason I think we have two versions of this PR and the other has already been merged.   https://github.com/ArduPilot/ardupilot/pull/30464

PR #30050, opened 2025-05-12
AP_Proximity: LD06 reports averages distances

rmackay9:
This modifies the LD06/LD19 proximity library so that average distances are reported instead of the shortest distance.

This will hopefully resolve the, "LD19 average change makes it sensitive to sunlight" issue from the [4.6 issues list](https://github.com/ArduPilot/ardupilot/issues/28612).

This is completely untested so I'd like help from @peterbarker and others to verify the code and test it.  I believe there is a SITL simulator so perhaps I can use that for testing

Some references:

- [beta testing discussion](https://discuss.ardupilot.org/t/looking-to-get-firmware-release-4-6-0-beta2-11-dec-2024/133422)
- [original PR that replaced the average distances with shortest distances](https://github.com/ArduPilot/ardupilot/pull/27045)
- [alternative PR](https://github.com/ArduPilot/ardupilot/pull/30016)

### DIFF
--- a/libraries/AP_Proximity/AP_Proximity_LD06.cpp
+++ b/libraries/AP_Proximity/AP_Proximity_LD06.cpp
@@ -178,41 +178,52 @@ void AP_Proximity_LD06::parse_response_data()
             continue;
         }
 
-        uint16_t a2d = (int)(angle_deg / 2.0) * 2;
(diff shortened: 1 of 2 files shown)


rmackay9:
closed because the mode filter solution works better

PR #30054, opened 2025-05-12
AP_Scripting: Add wind decoder for Trisonica LI-550

Ryanf55:
Bin logger for https://www.licor.com/products/trisonica/LI-550-mini

Dynamically writes log depending on what data outputs you enable.

Here's a uvw velocity plot.
image

Follow up: 
Fold this data into ArduPilot, share it via MAVLink or DDS outside AP, and use it as the airspeed source for the EKF.

### DIFF
--- a/libraries/AP_Scripting/examples/trisonica-mini.lua
+++ b/libraries/AP_Scripting/examples/trisonica-mini.lua
@@ -1,0 +1,127 @@
+-- Script decodes and logs wind sensor data for Trisonica LI-550 mini
+-- https://www.licor.com/products/trisonica/LI-550-mini
+
+-- Parameters:
+-- SCR_ENABLE 1
+-- SERIAL5_PROTOCOL 28
+
+-- In SITL, you can enable serial ports to connect to the real device.
+-- https://ardupilot.org/dev/docs/learning-ardupilot-uarts-and-the-console.html#the-8-uarts
+-- ./Tools/autotest/sim_vehicle.py -v Plane --console --map -A "--serial5=uart:/dev/ttyUSB0" -D
+
+-- Remember to change baud to 230k in the sensor setup and enable the fields you want.
+-- Also, enable 10Hz output instead of the default 5Hz.
+
+-- Example data string (excluding quotes, excluding the carriage return line feed ending)
+-- "S  00.08 S2  00.07 D  245 DV  033 U  00.06 V  00.03 W  00.05 T  55889220.00 C  346.68 H  17.92 DP  03.68 P -099.70 AD  0.0000000 AX  -2913 AY  -3408 AZ -16600 PI  011.4 RO  009.8 MX   -619 MY    845 MZ    782 MD  337 TD  337"
+
+-- Log severities
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+
+-- find the serial first (0) scripting serial port instance
+local port = serial:find_serial(0)
+
+if not port then
+    gcs:send_text(MAV_SEVERITY.EMERGENCY, "no trisonica scripting port")
+    return
+end
+
+-- begin the serial port
+port:begin(230400)
+port:set_flow_control(0)
+
+local function parse_wind_data(buf)
+    -- Split the string up into key and values splitting on the default space delimiter.
+
+    local parsed_values = {}
+
+    -- Match any key-value pair where key is a string and value is a number
+    for key, value in buf:gmatch("(%a+)%s*([%-%.%d]+)") do
+        parsed_values[key] = tonumber(value) -- Store key-value pair in the table
+    end
+
+    return parsed_values
+end
+
+local tag_ids = nil
+local tag_id_str = ""
+local value_format = ""
+local last_keys = nil
+
+local function log_wind_data(parsed)
+    -- Given a table of parsed data, log it.
+
+    -- Collect keys from parsed and store as a list
+    local current_keys = {}
+    for k in pairs(parsed) do
+        table.insert(current_keys, k)
+    end
+    table.sort(current_keys)
+
+    -- First packet, we must set last_keys.
+    if not last_keys then
+        last_keys = current_keys
+        return
+    end
+
+    -- Bail early if the first packet parsed only contained some of the keys. 
+    -- After the 2nd packet, it will stabilize.
+    if #current_keys ~= #last_keys then
+        last_keys = current_keys
+        return
+    end
+
+    for i = 1, #current_keys do
+        if current_keys[i] ~= last_keys[i] then
+            last_keys = current_keys
+            return
+        end
+    end
+
+    -- Keys are now stable, so store them as tag_ids
+    if not tag_ids then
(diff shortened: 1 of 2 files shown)


libraries/AP_Scripting/applets/trisonica-mini.lua
```diff
@@ -0,0 +1,100 @@
...
+    -- Build up the logger list of dynamic tag ID's.
+    for tag_id, v in pairs(parsed) do
+        table.insert(tag_ids, tag_id)
+        table.insert(values, v)
+    end
+    local tag_id_str = table.concat(tag_ids, ',')
+    local value_format = string.rep('f', #tag_ids)
+    
+    assert(#tag_ids < 15, "#tag_ids=" .. #tag_ids)
+    logger.write('W3D', tag_id_str, value_format,
```
IamPete1, line 62:
```suggestion
    logger:write('W3D', tag_id_str, value_format,
```
to pass CI.

IamPete1: Potentially dynamic log items will cause logging to fail if they ever change. Once you register a log name you must keep the same format.

One fix would be to log the name value pairs in a instance message with the tag as the instance.

Ryanf55: That's ok, these don't change at runtime unless you are reconfiguring the sensor with the tool, which you would disconnect it and reconnect to the autopilot.

libraries/AP_Scripting/applets/trisonica-mini.lua
```diff
@@ -0,0 +1,100 @@
...
+-- SERIAL5_PROTOCOL 28
+
+-- In SITL, you can enable serial ports to connect to the real device.
+-- https://ardupilot.org/dev/docs/learning-ardupilot-uarts-and-the-console.html#the-8-uarts
+-- ./Tools/autotest/sim_vehicle.py -v Plane --console --map -A "--serial5=uart:/dev/ttyUSB0" -D
+
+-- Remember to change baud to 230k in the sensor setup and enable the fields you want.
+-- Also, enable 10Hz output instead of the default 5Hz.
+
+-- Example data string (excluding quotes, including the carriage return line feed ending)
```
tpwrules, line 15: What precisely is the ending? As you describe it's "\r\n" but the code doesn't match that and no endings are in the sample.

libraries/AP_Scripting/applets/trisonica-mini.lua
```diff
@@ -0,0 +1,100 @@
...
+
+-- Log severities
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+
+-- find the serial first (0) scripting serial port instance
+local port = serial:find_serial(0)
+
+if not port then
+    gcs:send_text(MAV_SEVERITY.EMERGENCY, "No Scripting Serial Port")
+    return
```
tpwrules, line 26: Personal choice: `error("no trisonica scripting port")` modulo message wording.

libraries/AP_Scripting/applets/trisonica-mini.lua
```diff
@@ -0,0 +1,100 @@
...
+local buffer = ""
+
+local function parse_wind_data(buf)
+    -- Split the string up into key and values splitting on the default space delimiter.
+
+    local parsed_values = {}
+
+    -- Match any key-value pair where key is a string and value is a number
+    for key, value in buf:gmatch("(%a+)%s*([%-%.%d]+)") do
+        parsed_values[key] = tonumber(value) -- Store key-value pair in the table
```
tpwrules, line 42: Storing as a map is not great, it should be a list. A map's order is not defined so the log fields could be different each time depending on memory allocation etc.

Maybe a check if the log keys have been defined and if not build it separately so the code only has to run once anyway, as Pete correctly noted it can't be changed anyway

libraries/AP_Scripting/applets/trisonica-mini.lua
```diff
@@ -0,0 +1,100 @@
...
+end
+
+
+-- the main update function that is used to read in data from serial port
+local function update()
+
+    if not port then
+        gcs:send_text(0, "no Scripting Serial Port")
+        return update, 100
+    end
```
tpwrules, line 73:
```suggestion
```

You already bail above, no reason to have this.

libraries/AP_Scripting/applets/trisonica-mini.lua
```diff
@@ -0,0 +1,100 @@
...
+
+    -- gcs:send_text(MAV_SEVERITY.INFO, "update()")
+
+    local n_bytes = port:available()
+    while n_bytes > 0 do
+        local byte = port:read()
+
+        if byte then
```
tpwrules, line 81:
```suggestion
        if byte > -1 then
```

Only `nil` and `false` are falsy in Lua. You also don't need the outer loop. Or you could ditch this check and assume nobody is stealing bytes from under your nose.

tpwrules (changes requested):

libraries/AP_Scripting/applets/trisonica-mini.lua
Ryanf55:
```suggestion
    gcs:send_text(MAV_SEVERITY.EMERGENCY, "no trisonica scripting port")
    return
```

IamPete1: You could also assert the `find_serial` for a even more annoying warning.

libraries/AP_Scripting/applets/trisonica-mini.lua
Ryanf55: Although table order is not guaranteed, my testing on hardware shows that it works fine. I do like the idea of build the fields only once. I'll add that.

libraries/AP_Scripting/applets/trisonica-mini.lua
tpwrules: To be clear I don't care too much about the message wording, I just think it would be simpler and easier to just call `error` or `assert` instead of sending an emergency message and returning.

libraries/AP_Scripting/applets/trisonica-mini.lua
Ryanf55: `\r\n`. I just matched `\r` because it was easy.

libraries/AP_Scripting/applets/trisonica-mini.lua
Ryanf55: Done! I also only assemble the log structure string once now too.

tpwrules (approved):

IamPete1 (approved):

libraries/AP_Scripting/applets/trisonica-mini.lua
```diff
@@ -0,0 +1,127 @@
+-- Script decodes and logs wind sensor data for Trisonica LI-550 mini
+-- https://www.licor.com/products/trisonica/LI-550-mini
+
```
tridge, line 3: maybe in examples? later will be a driver

tridge (changes requested):

PR #30066, opened 2025-05-13
hwdef: arkv6x: add definitions for analog voltage and current monitoring

dakejahl:
Enables ADIO ADCs for analog current and voltage monitoring. The HAL_BATT_MONITOR_DEFAULT remains set to the INA226 by default.

### DIFF
--- a/libraries/AP_HAL_ChibiOS/hwdef/ARKV6X/hwdef.dat
+++ b/libraries/AP_HAL_ChibiOS/hwdef/ARKV6X/hwdef.dat
@@ -116,12 +116,6 @@ PF12 SCALED4_V3V3 ADC1 SCALE(2)
 
 PB1 VDD_5V_SENS  ADC1 SCALE(2)
 
-# pin7 on AD&IO, analog 12
-PC2 ADC1_6V6 ADC1 SCALE(2)
-
-# pin6 on AD&IO, analog 13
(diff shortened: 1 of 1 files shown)


peterbarker:
What testing has this had?

How are these pins exposed on the device?  Are you going to break people's existing setups by setting these defaults?

dakejahl:
These pins are ADC1_3V3 and ADC1_6V6 on the ADIO connector of the ARKV6X carrier. 
image

I opened this PR for a customer to test and marked it as a draft PR because I have not tested it. A customer is integrating a MAUCH 031 PDB which reports analog voltage and current and will be testing this PR and reporting back.

> Are you going to break people's existing setups by setting these defaults?

I don't think so, AFAIK this will have no effect since HAL_BATT_MONITOR_DEFAULT is set to 21 (INA2XX). I would only expect this to work if the type was changed to 4 (ANALOG_VOLTAGE_AND_CURRENT). Please correct me if I'm wrong.

peterbarker:
> I opened this PR for a customer to test and marked it as a draft PR because I have not tested it. A customer is integrating a MAUCH 031 PDB which reports analog voltage and current and will be testing this PR and reporting back.

Has the customer gotten back to you with results yet? :-)

dakejahl:
I reached out

PR #30069, opened 2025-05-14
AP_Common: tests: remove un-needed vehicle object

peterbarker:
This is a unit test - we run it in CI.

Runs locally.


khancyr (approved):
Lgtm.
Pass the test.
Probably a leftover from previous way we got this access

tridge (approved):

PR #30071, opened 2025-05-14
Copter: avoid updating simple-mode values when RC input is invalid

peterbarker:
if we are at bind-time values then we still get new radio frames, they're just not something we should use

Tested in SITL by setting a breakpoint within the new clause.  When we're in there having set SIM_RC_FAIL to 2 the channels are definitely using bind-time values.

### DIFF
--- a/ArduCopter/Copter.cpp
+++ b/ArduCopter/Copter.cpp
@@ -853,6 +853,11 @@ void Copter::update_simple_mode(void)
     // mark radio frame as consumed
     ap.new_radio_frame = false;
 
+    // avoid processing bind-time RC values:
+    if (!rc().has_valid_input()) {
+        return;
+    }
+
     if (simple_mode == SimpleMode::SIMPLE) {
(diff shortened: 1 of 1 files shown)


rmackay9 (approved):
thanks for this.  I think this is OK.

My only question was whether we should put this new check above or below the "ap.new_radio_frame = false;" line.  I think it makes sense to put it afterwards as this PR has done though because we are "consuming" the frame it's just that we've also decided that we don't want to use it for anything.

rmackay9:
as a side note, the ap.new_radio_frame is now only used by simple mode so we could give it a more specific name to make that clear.  This doesn't really matter though for now.

peterbarker:
> as a side note, the ap.new_radio_frame is now only used by simple mode so we could give it a more specific name to make that clear. This doesn't really matter though for now.

I'm looking forward to having an `rc().last_valid_input_ms()` call we can make.  When we've updated these simple-mode values we note down the timestamp for the values we just munged, and only munge them again if the timestamp has changed.

PR #30079, opened 2025-05-15
Rover: add 'Loiter or Hold' as a failsafe action

Williangalvani:
This is especially useful for boats, as we usually don't want it to drift away with the current once if failsafes

### DIFF
--- a/Rover/Rover.h
+++ b/Rover/Rover.h
@@ -416,7 +416,8 @@ private:
         Hold          = 2,
         SmartRTL      = 3,
         SmartRTL_Hold = 4,
-        Terminate     = 5
+        Terminate     = 5,
+        Loiter_Hold   = 6,
     };
 
     enum class Failsafe_Options : uint32_t {
--- a/Rover/fence.cpp
+++ b/Rover/fence.cpp
@@ -67,6 +67,11 @@ void Rover::fence_check()
                         set_mode(mode_hold, ModeReason::FENCE_BREACHED);
                     }
                     break;
+                case FailsafeAction::Loiter_Hold:
+                    if (!set_mode(mode_loiter, ModeReason::FENCE_BREACHED)) {
+                        set_mode(mode_hold, ModeReason::FENCE_BREACHED);
+                    }
+                    break;
                 case FailsafeAction::Terminate:
                     arming.disarm(AP_Arming::Method::FENCEBREACH);
                     break;
--- a/Tools/autotest/rover.py
+++ b/Tools/autotest/rover.py
@@ -174,55 +174,32 @@ class AutoTestRover(vehicle_test_suite.TestSuite):
         self.set_rc(3, 1500)
         self.progress("Circuit complete")
 
-    # def test_throttle_failsafe(self, home, distance_min=10, side=60,
-    #                            timeout=300):
-    #     """Fly east, Failsafe, return, land."""
-    #
-    #     self.mavproxy.send('switch 6\n')  # manual mode
-    #     self.wait_mode('MANUAL')
-    #     self.mavproxy.send("param set FS_ACTION 1\n")
-    #
-    #     # first aim east
-    #     self.progress("turn east")
-    #     if not self.reach_heading_manual(135):
-    #         return False
-    #
-    #     # fly east 60 meters
-    #     self.progress("# Going forward %u meters" % side)
-    #     if not self.reach_distance_manual(side):
-    #         return False
-    #
-    #     # pull throttle low
-    #     self.progress("# Enter Failsafe")
-    #     self.mavproxy.send('rc 3 900\n')
-    #
-    #     tstart = self.get_sim_time()
-    #     success = False
-    #     while self.get_sim_time() < tstart + timeout and not success:
-    #         m = self.mav.recv_match(type='VFR_HUD', blocking=True)
-    #         pos = self.mav.location()
-    #         home_distance = self.get_distance(home, pos)
-    #         self.progress("Alt: %u  HomeDistance: %.0f" %
-    #                       (m.alt, home_distance))
-    #         # check if we've reached home
-    #         if home_distance <= distance_min:
-    #             self.progress("RTL Complete")
-    #             success = True
-    #
-    #     # reduce throttle
-    #     self.mavproxy.send('rc 3 1500\n')
-    #     self.mavproxy.expect('AP: Failsafe ended')
-    #     self.mavproxy.send('switch 2\n')  # manual mode
-    #     self.wait_heartbeat()
-    #     self.wait_mode('MANUAL')
-    #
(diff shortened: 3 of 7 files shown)


rmackay9 (approved):
Hi @Williangalvani,

Thanks very much for this!  It looks good, I just wanted to check that it's been tested in SITL?

rmackay9:
@Williangalvani,

Tell me when you're ready and I think we can merge this

Williangalvani:
> @Williangalvani,
> 
> Tell me when you're ready and I think we can merge this

I think it is ready! I just added a test. Can you take a look? There was an older, commented out code, so I just replaced that.

Tools/autotest/rover.py
```diff
@@ -174,55 +174,38 @@ def drive_left_circuit(self):
...
-    #         "timed out after %u seconds" % timeout)
-    #         return False
+    def ThrottleFailsafe(self):
+        """Trigger throttle failsafes"""
+        self.progress("Testing throttle failsafe")
+        # set failsafe to Loiter_or_Hold
+        self.set_parameters({
+            "FS_THR_ENABLE": 1,
+            "FS_ACTION": 6,
+            "FS_THR_VALUE": 1000,
```
rmackay9, line 184: Perhaps @peterbarker knows better but for me, instead of setting FS_THR_VALUE and "self.set_rc(3, 900)" I would use the SIM_RC_FAIL parameter

self.set_parameter("SIM_RC_FAIL", 1)

and then to restore RC use:

self.set_parameter("SIM_RC_FAIL", 0)

functionally this makes no difference though

Tools/autotest/rover.py
```diff
@@ -174,55 +174,38 @@ def drive_left_circuit(self):
...
+        self.set_rc(3, 1000)
+        self.change_mode("MANUAL")
+        self.set_parameter("SIM_GPS1_JAM", 1)
+
+        self.progress("Testing failsafe to hold")
+        self.set_rc(3, 900)
+        self.wait_mode("HOLD", timeout=10)
+        self.progress("Loiter or Hold as throttle failsafe OK")
+
+        self.set_rc(3, 1000)
```
rmackay9, line 204: I think you can remove this stuff at the bottom because we have push and restore now before and after each test

Rover/Rover.h
```diff
@@ -416,7 +416,8 @@ class Rover : public AP_Vehicle {
         Hold          = 2,
         SmartRTL      = 3,
         SmartRTL_Hold = 4,
-        Terminate     = 5
+        Terminate     = 5,
+        Loiter_Hold   = 6
```
peterbarker, line 420:
```suggestion
        Loiter_Hold   = 6,
```

Tools/autotest/rover.py
Williangalvani: nice! updated

Tools/autotest/rover.py
Williangalvani: ok, I only needed to keep the disarm() for the autotest to stay happy

Tools/autotest/rover.py
```diff
@@ -174,55 +174,34 @@ def drive_left_circuit(self):
...
+        self.set_parameters({
+            "FS_THR_ENABLE": 1,
+            "FS_ACTION": 6,
+        })
+        self.change_mode("MANUAL")
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
+
+        self.progress("Testing failsafe to loiter")
+        self.set_parameter("SIM_RC_FAIL", 1)
```
peterbarker, line 190: So this is different to what was here a moment ago.

This was setting throttle to a bind-time value, now you are setting SITL to be "no pulses on RC input" instead.

So you're not really "testing throttle failsafe" here now, you are testing *RC* failsafes.

This is fine - but I *am* planning on renaming these parameters :-)

```suggestion
        self.set_parameter("SIM_RC_FAIL", 1)  # no-pulses failsafe
```

Tools/autotest/rover.py
```diff
@@ -174,55 +174,34 @@ def drive_left_circuit(self):
...
+        self.wait_mode("LOITER", timeout=10)
+
+        # restore RC and prepare to test hold fallback
+        self.set_parameter("SIM_RC_FAIL", 0)
+        self.set_parameter("SIM_GPS1_JAM", 1)
```
peterbarker, line 195: Usually we would fail the GPSs by disabling rather than jam them.... but this apparently works!

Tools/autotest/rover.py
```diff
@@ -174,55 +174,34 @@ def drive_left_circuit(self):
...
+        self.change_mode("MANUAL")
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
+
+        self.progress("Testing failsafe to loiter")
+        self.set_parameter("SIM_RC_FAIL", 1)
+        self.wait_mode("LOITER", timeout=10)
+
+        # restore RC and prepare to test hold fallback
+        self.set_parameter("SIM_RC_FAIL", 0)
```
peterbarker, line 194: Maybe use set_parameters here?  It's marginalyl faster

peterbarker (approved):

Tools/autotest/rover.py
rmackay9: So for me, throttle failsafe and RC failsafe are the same thing.  They may be triggered by slightly different signals sent by the RC receiver (e.g. futaba receivers send a pulse below 900us, other receivers send no pulses at all)... but they're essentially the same thing

Tools/autotest/rover.py
Williangalvani: disabling sounds better. updated.

Tools/autotest/rover.py
Williangalvani: done!

Williangalvani: actually now that I switched to changing SIM_GPS1_ENABLE, it is passing without changing GPS2. I wonder if that was just flaky because I was using JAM instead

peterbarker (approved):

Rover/fence.cpp
```diff
@@ -67,6 +67,11 @@ void Rover::fence_check()
                         set_mode(mode_hold, ModeReason::FENCE_BREACHED);
                     }
                     break;
+                case FailsafeAction::Loiter_Hold:
```
IamPete1, line 70: I think you need the docs for this one up in AC_Fence too.

IamPete1: https://github.com/ArduPilot/ardupilot/blob/7c0f01bd8d4c335a3fb962cfce925640cc1d4307/libraries/AC_Fence/AC_Fence.cpp#L70

Williangalvani: thanks! that was missing "terminate" too

Issue #30076, opened 2025-05-15
Fixing ExternalNav + GPS


eufrizz:
## Bug report
(previously discussed on Discord under the partner-theseus channel)
**Issue details**
Ardupilot is fusing external nav and GPS data when only external nav is selected as the input source.
From my understanding, only one of the two sources should be used for position at a time, as defined by EK3_SRC_POSXY and EK3_SRC_VELXY. However, if external nav is selected and GPS data is available, the filter incorrectly fuses this data too. This prevents flying on external nav with the ability to switch to GPS as a backup.
It manifests as an unstable position with high variances and resetting behaviour, if the two position estimates diverge - which will be the case with sources like VIO.
The difference can be demonstrated running a replay log with and without GPS data:

![Image](https://github.com/user-attachments/assets/081b7e71-1f6e-4498-878f-d1f117d8246d)
(notice the jagged position and very high innovations with resets)
![Image](https://github.com/user-attachments/assets/a8eff5b9-3618-4594-8405-340dce3be129)
(notice the smooth position and much lower innovations - though still a bit high)

Upon inspecting AP_NavEKF3_PosVelFusion.cpp, it seems the if statements in FuseVelPosNED are just not correctly written to separate external nav from GPS. What happens is that FuseVelPosNED is called with `extNavVelToFuse == gpsDataToFuse == fuseVelData == fusePosData == 0`, but with `fuseHgtData == 1` (presumably due to only having baro data). Thus when an if statement like the following is encountered:
```
if (extNavVelToFuse) {
                obs_data_chk = sq(constrain_ftype(extNavVelDelayed.err, 0.05f, 5.0f)) + sq(frontend->extNavVelVarAccScale * accNavMag);
            } else
#endif
            {
                obs_data_chk = sq(constrain_ftype(frontend->_gpsHorizVelNoise, 0.05f, 5.0f)) + sq(frontend->gpsNEVelVarAccScale * accNavMag);
            }
```

The else statement is triggered, incorrectly, incorporating GPS data when in fact external nav is selected. There needs to be an extra check for gpsDataToFuse or something like that.

~~Have fixed it in a branch, will add a pull request tomorrow, I published this issue prematurely by accident lol.~~
PR here: https://github.com/ArduPilot/ardupilot/pull/30080

**Version**
Ardupilot 4.6.0 beta 1 and 2

**Platform**
[  ] All
[  ] AntennaTracker
[ x ] Copter
[ x ] Plane
[  ] Rover
[...]

PR #30080, opened 2025-05-15
External nav+gps fix

eufrizz:
Resolves https://github.com/ArduPilot/ardupilot/issues/30076#issuecomment-2883484485

Basically, the case where FusePosVelNED function is called for only a baro update, caused GPS data to be inadvertently fused with the external nav estimate. This created two conflicting sources of position, and an unstable filter with resetting behaviour as external nav and GPS diverged (e.g. if external nav is VIO).
This PR allows testing of external nav with GPS as a backup. Works cleanly in the case of:
```
EK3_SRC1_POSXY=6
EK3_SRC1_VELXY=6
```
and 
```
EK3_SRC1_POSXY=3
EK3_SRC1_VELXY=3
```
Have not tested other permutations of sources too much. Runs with POSZ=3 but I don't have any data where baro and GPS alt are differentiated.

Restructuring the if statements to deal with GPS, external nav and other data separately, e.g.
```
if (extNavUsedForVel) {
   ...
} else {
   ...
}
```

becomes 
```
if (extNavUsedForVel) {
   ...
} else if (gpsDataToFuse) {
   ...
}   
else {
  // TODO: deal with vel/pos sources that aren't external nav or GPS
  ...
}
```
As you can see, I'm not entirely sure what to add in the else case as I don't have experience with other sensor modalities - these are:
Position: 
- Beacon
Velocity:
- Beacon
- OpticalFlow
- WheelEncoder

So would appreciate some guidance/help there. We can just set them to be the same as the GPS case but it seems like they should be treated separately?

I also broke up
```
if (fuseVelData || fusePosData || fuseHgtData) {
```

into separate if statements for each of the three conditions, as the code made more sense that way.

The other important change was:
```
if ((posTimeout || posVarianceIsTooLarge) && (!velAiding || gpsGoodToAlign))
```
to
```
if ((posTimeout || posVarianceIsTooLarge) && (!velAiding || (gpsGoodToAlign && gpsDataToFuse))) {
```
To prevent this clause falsely triggering when using external nav. Similar things are done with `gpsDataToFuse && frontend->_gpsGlitchRadiusMax <= 0` that are less critical but prevent the glitch radius parameter interfering with external nav operation.

The result is that


rmackay9:
Hi @eufrizz,

thanks very much for this.  It looks like this is failing some autotests for valid reasons.  Perhaps the issue is it fails when some defines are changed.

> error: 'extNavVelToFuse' was not declared in this scope


By the way, could you modify the commits so that they start with "AP_NavEKF3:"? We do this on all our commits to allow for easier understanding of what the commits affect and it helps with backporting.  To be clear, I mean the commit titles, not the PR titles

eufrizz:
> Hi @eufrizz,
> 
> thanks very much for this. It looks like this is failing some autotests for valid reasons. Perhaps the issue is it fails when some defines are changed.
> 
> > error: 'extNavVelToFuse' was not declared in this scope
> 
> By the way, could you modify the commits so that they start with "AP_NavEKF3:"? We do this on all our commits to allow for easier understanding of what the commits affect and it helps with backporting. To be clear, I mean the commit titles, not the PR titles

@rmackay9  Fixed!

Also took it for a test flight on an X500 quadcopter yesterday and it performed admirably, was able to switch between GPS and external nav no issues. Here is the [log](https://drive.google.com/file/d/1oY5Uv9pRjoBxhe-XEPVTxf_cVO8bUco0/view?usp=sharing) - it starts of with some low flying, switching between the external nav and GPS (shown below), then an auto mission where the external nav fails, innovation goes high, EKF failsafe occurs, and eventually we are able to switch it back to RTL using GPS.

Also, would it be possible to get this into the 4.6.0 release? Happy to join the Monday dev call or help out in some other way. Will hopefully be testing on a longer plane flight soon.

rishabsingh3003 (changes requested):
Hi! Thanks for this PR. I’m still a bit unclear on this part:

> "The case where FuseVelPosNED() is called for only a baro update caused GPS data to be inadvertently fused with the ExternalNav estimate."

I’ve been going over the logic and I don’t see how that would happen. From what I can tell, the actual fusion doesn’t occur until here:
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_NavEKF3/AP_NavEKF3_PosVelFusion.cpp#L983

At that point, fusion is strictly guarded by flags like fusePosData and fuseVelData, which are reset each timestep and only set under specific source conditions. So unless I am overlooking something, I don’t see a clear path for GPS data to be fused while EXTNAV is selected.

That said, I do agree the surrounding code, particularly where the measurement variances are set up (which is the top part of the function that you are making changes to), can be tricky. I’ve fixed a few issues there myself recently. However, with the volume of changes in the PR, it’s difficult to distinguish formatting updates from actual logic fixes.

As a general suggestion: this part of the EKF is extremely sensitive and quite complex. A small change can cause extreme behaviour. I’d strongly recommend breaking code changes into smaller PRs, each focused on a specific bug fix. If you would like to do formatting changes, I would suggest opening a PR that ONLY does that with no logical changes. That would make the review process much more manageable. Also, for anything touching this fusion logic, a REPLAY log should be included so that the fix is backed by a clear before/after proof of behavior.

Thanks again!

eufrizz:
Hi @rishabsingh3003, thanks for the review - I left the TODOs in various places to draw attention to places where I wasn't sure what do to, and was hoping for your input there. Currently I've only written it to deal with GPS and external nav but there are other input sources that need handling. E.g. maybe you can advise on how to handle range beacon input - I'm unfamiliar with the sensor and have no data to test on.

rmackay9:
Hi @eufrizz,

We have a virtual beacon system in SITL which can be [setup as described here](https://ardupilot.org/dev/docs/adding_simulated_devices.html#adding-range-beacons).

peterbarker:
Ping @eufrizz - requests have been left on this PR for you.

Is this still something you are interested in chasing?

eufrizz:
Hi @peterbarker , don't think I'm going to have the time to see this through sorry

PR #30081, opened 2025-05-16
AP_Scripting: correct use of pcall in examples and applets

peterbarker:
... tested in another script which uses the same block....


tpwrules:
Yeah I had kind of assumed those were commented out for a reason...

tridge (approved):

PR #30083, opened 2025-05-16
AntennaTracker NAV_CONTROLLER_OUTPUT.target_bearing = realy target bearing in guided mode

python36:
In guided mode we don't know the target bearing. Also, the only way to confirm acceptance of the SET_ATTITUDE_TARGET command is STATUSTEXT, which is not normal.


Ryanf55:
Can you add instructions how I can test this in SITL? 

Sorry, SET_ATTITUDE_TARGET interface does not have a good feedback mechanism.

python36:
Due to the fact that it was not possible to make a confirmation of acceptance of the SET_ATTITUDE_TARGET command, since the accuracy of the NAV_CONTROLLER_OUTPUT.targaet_bearing field is not sufficient - it needs to be in cDeg, then I made support for ATTITUDE_TARGET packages (#30548).

PR #30084, opened 2025-05-16
ArduPlane: Fixed transition to forward flight mode when transitioning to VTOL

tizianofiorenzani:
For Arduplane Tailsitter.

This patch fixes a problem with the Tailsitter state machine, which remains stuck in VTOL mode if we command a forward flight mode while transitioning to VTOL.

I don't know if we send this state out in any message, I just found out through sending it through DDS. During transition to QRTL I commanded CRUISE mode. Vehicle correctly switched to CRUISE but the state machine kept reporting "VTOL"

### DIFF
--- a/ArduPlane/tailsitter.cpp
+++ b/ArduPlane/tailsitter.cpp
@@ -872,7 +872,9 @@ void Tailsitter_Transition::update()
     }
 
     case TRANSITION_ANGLE_WAIT_VTOL:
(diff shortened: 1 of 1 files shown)


IamPete1:
We do have a existing PR, I thinks for the same issue?

https://github.com/ArduPilot/ardupilot/pull/29675

tizianofiorenzani:
Oh, great, I guess I can close this onw then

PR #30087, opened 2025-05-17
AP_Scripting: fix non-string-induced null pointer dereferences

tpwrules:
Incorrect use of the Lua string conversion functions could cause null pointer dereferences and possible autopilot faults. The simplest example is `error()`, which is guaranteed to cause a dereference if it occurs at the top level of a script, but there are some bindings susceptible to problems too. 

This is just the patches for 4.6 for easy backporting. The `error()` issue is also in 4.5 and probably earlier; that commit could be straightforwardly backported on its own. (The affected bindings are not in 4.5)

Tested on CubeOrange that the affected functions don't dereference null pointers anymore and appropriate errors are given.

### DIFF
--- a/libraries/AP_Scripting/lua_bindings.cpp
+++ b/libraries/AP_Scripting/lua_bindings.cpp
@@ -813,7 +813,7 @@ int lua_serial_writestring(lua_State *L)
 
     // get the bytes the user wants to write, along with their length
     size_t req_bytes;
-    const char *data = lua_tolstring(L, 2, &req_bytes);
+    const char *data = luaL_checklstring(L, 2, &req_bytes);
 
     // write up to that number of bytes
     const uint32_t written_bytes = port->write((const uint8_t*)data, req_bytes);
--- a/libraries/AP_Scripting/lua_scripts.cpp
+++ b/libraries/AP_Scripting/lua_scripts.cpp
@@ -40,6 +40,15 @@ uint32_t lua_scripts::loaded_checksum;
 uint32_t lua_scripts::running_checksum;
 HAL_Semaphore lua_scripts::crc_sem;
 
+// return string error message for error object at top of stack
+static const char *get_error_object_message(lua_State *L) {
+    const char *m = lua_tostring(L, -1);
+    if (!m) { // error object is not stringifiable
+        return "<error>";
+    }
+    return m;
+}
+
 lua_scripts::lua_scripts(const AP_Int32 &vm_steps, const AP_Int32 &heap_size, AP_Int8 &debug_options)
     : _vm_steps(vm_steps),
       _debug_options(debug_options)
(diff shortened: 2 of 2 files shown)


libraries/AP_Scripting/lua_scripts.cpp
```diff
@@ -40,6 +40,15 @@ uint32_t lua_scripts::loaded_checksum;
 uint32_t lua_scripts::running_checksum;
 HAL_Semaphore lua_scripts::crc_sem;
 
+// return string error message for error object at top of stack
+static const char *get_error_object_message(lua_State *L) {
+    const char *m = lua_tostring(L, -1);
+    if (!m) { // error object is not stringifiable
+        return "<error>";
```
timtuxworth, line 47: Should this be (char *)"<error>" - I'm pretty sure this won't compile on my Mac as is.

tpwrules: The macOS SITL CI build passed, and the `char[]` literal type will decay to a const pointer just fine?

IamPete1:
If i understand correctly the key difference is that `checklstring` will long jump to error so the nulptr is never used. Currently we would not error and end up trying to use it.

libraries/AP_Scripting/lua_bindings.cpp
```diff
@@ -661,7 +661,7 @@ int AP_HAL__I2CDevice_transfer(lua_State *L) {
 
     // Parse string of bytes to send
     size_t send_len;
-    const uint8_t* send_data = (const uint8_t*)(lua_tolstring(L, 2, &send_len));
+    const uint8_t* send_data = (const uint8_t*)(luaL_checklstring(L, 2, &send_len));
```
IamPete1, line 664: Its valid to pass nulptr to `transfer`, it is required by some devices. grep for `transfer(nullptr`.

So it that does work as expected from lua I think its fine to leave this one? The lus would be something like:

`AP_HAL__I2CDevice_ud:transfer(nil, 2)`

If that is safe we should add a comment here explaining why its OK and update the docs that you can pass nil for the first argument to only read.

libraries/AP_Scripting/lua_scripts.cpp
IamPete1: Yeah, I think its fine.

libraries/AP_Scripting/lua_bindings.cpp
IamPete1: I guess we could add a check that if it not a string it must be exactly nil, so passing other random stuff will error out.

tpwrules: Does it matter what the pointer is if the length argument is 0? A user could pass an empty string to get a zero length.

IamPete1: Hum, the chibiOS bounce buffer stuff checks for nulprt so it might matter. I guess you could always pass nulptr if the length is zero.

tpwrules: I like that option more, I will update to do that.

IamPete1 (approved):
LGTM, thanks!

rmackay9:

PR #30088, opened 2025-05-17
 AP_Scripting: fix more non-string-induced null pointer dereferences

tpwrules:
#30087 for master-specific code

### DIFF
--- a/libraries/AP_Scripting/AP_Scripting_helpers.cpp
+++ b/libraries/AP_Scripting/AP_Scripting_helpers.cpp
@@ -213,7 +213,7 @@ int DroneCAN_Handle::broadcast(lua_State *L)
     DroneCAN_Handle *h = check_DroneCAN_Handle(L, 1);
 
     size_t data_length;
-    const void *data = lua_tolstring(L, 2, &data_length);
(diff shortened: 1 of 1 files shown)


IamPete1:
Are there any DroneCAN messages that have no payload? I guess in that case we would just have the user pass a empty string?

tpwrules:
Yes, the request for GetNodeInfo is one. I checked that an empty string works properly for it.

IamPete1 (approved):
LGTM

rmackay9:
@tpwrules, @tridge 

Do we want to include this PR in 4.6.2-beta1?  This other very similar PR has been included: https://github.com/ArduPilot/ardupilot/pull/30087

tpwrules:
@rmackay9 thanks for your diligence! As I very briefly explained in the two PR messages, that PR is for code that’s in 4.6 but this one is for code in master only. I split it to make your job a little simpler.

If we backport the dronecan-from-Lua feature then we will want to backport this too. But right now that feature is not in 4.6 so backporting these fixes to it will fail!

rmackay9:
@tpwrules,  great stuff, thanks for the explanation!

PR #30090, opened 2025-05-17
autotest: add test for wp yaw behaviour type 1 

peterbarker:


Ryanf55:
Would be good to add a test for pitch up that it never stalls.

andyp1per (approved):

PR #30091, opened 2025-05-18
autotest: add test for TECS throttling up in auto

peterbarker:


Ryanf55:
Would be good to add a test for pitch up that it never stalls or goes below AIRSPEED_MIN

peterbarker:
This is the fundamental problem:
image

TECS internal energy is dropping as the plane descends, so it wants to increase the energy in the system by throttling up.  

TECS is driving the vehicle well past its configured `AIRSPEED_MAX`:
image

peterbarker:
Ping @rubenp02 - you've been annoyed by our stick-mixing lately, and have proposed a roll-only stick-mixing option which might be a useful intermediate workaround.

Ping @Georacer - can you think of a nice means of constraining TECS's behaviour without actually notifying that stick mixing is happening?

peterbarker:
Ping @Ryanf55 can you put the log from the real vehicle up somewhere, please?

Ryanf55:
> Ping @Ryanf55 can you put the log from the real vehicle up somewhere, please?

https://drive.google.com/file/d/1obewD2xbEDU6N7iaLra6dkCM2RwqdW0X/view?usp=drive_link

peterbarker:
Discussed on DevCall.

We need patches that stop TECS from over-speeding the vehicle

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -670,6 +676,15 @@ void AP_TECS::_detect_underspeed(void)
     }
 }
 
+void AP_TECS::_detect_overspeed(void)
+{
+    if (_flags.overspeed) {
+        _flags.overspeed = _TAS_state > _TASmax * 0.5;
```
Georacer, line 682: I don't think this is a good way to unset the overspeed flag.
Consider the following configuration:
- `AIRSPEED_MAX` = 25
- `AIRSPEED_MIN` = 20
- and cruise somewhere in between.

So basically you need to stall in order to clear your overspeed flag.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -637,6 +637,12 @@ void AP_TECS::_update_height_demand(void)
         // correct for offset between height above ground and height above datum used by control loops
         _hgt_dem += (_hgt_afe - _height);
     }
+
+    // avoid increasing the error between our current energy and
+    // demanded energy, to avoid pushing the throttle through the roof
+    if (_flags.overspeed) {
+        // _hgt_dem = _height;
+    }
```
Georacer, line 645: Pin so that we don't forget to amend this.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -746,6 +761,8 @@ void AP_TECS::_update_throttle_with_airspeed(void)
         _throttle_dem = 1.0f;
     } else if (_flags.is_gliding) {
         _throttle_dem = 0.0f;
+    // } else if (_flags.overspeed) {
+    //        _throttle_dem = 0.0f;
```
Georacer, line 765: Pin so that we don't forget to amend this.

Georacer (changes requested):
I'll give it some thought, Peter.
I have to understand how pitch nudging works in AUTO first.

libraries/AP_TECS/AP_TECS.cpp
peterbarker: > I don't think this is a good way to unset the overspeed flag. Consider the following configuration:
> 
>     * `AIRSPEED_MAX` = 25
> 
>     * `AIRSPEED_MIN` = 20
> 
>     * and cruise somewhere in between.
> 
> 
> So basically you need to stall in order to clear your overspeed flag.

Yes, thanks.  This was Ryan, Bob, Michelle and myself mucking around to see if we could get the thing to not over-speed :-)  May have gotten a little over-enthusiastic there...

Clamping into airspeed max yields oscillation, as you might imagine.  0.9 was no better.  0.5 moved things along nicely ;-)

libraries/AP_TECS/AP_TECS.cpp
peterbarker: > Pin so that we don't forget to amend this.

Hoping these are awful enough that someone is so revolted they come up with the correct solution.  We looked at the input side and the output side and discussed clamping the energy difference somewhere, but couldn't find something we liked.

libraries/AP_TECS/AP_TECS.cpp
```diff
@@ -828,22 +845,44 @@ void AP_TECS::_update_throttle_with_airspeed(void)
...
-                                        (double)STEdot_error,
-                                        (double)ff_throttle,
-                                        (double)integ_min,
-                                        (double)integ_max,
-                                        (double)_integTHR_state,
-                                        (double)SPE_err_min,
-                                        (double)SPE_err_max);
+            AP::logger().WriteStreaming(
+                "TEC3",
+                "TimeUS," "KEDD," "PEDD," "TEE," "TEDE," "FFT," "Imin," "Imax," "I," "Emin," "Emax," "P," "K",
```
Georacer, line 850: @peterbarker it also looks to me that you've mis-labeled the names here.
E.g. `_SKEdot` is `KED`, not `KEDD`.
And `SPE_err_max` is not `K`.

Georacer (changes requested):

Georacer:
How about something like this? https://github.com/Georacer/ardupilot/tree/pr/tecs-throttle-up-problem
The `_flag_pitch_forced` isn't really used, I was experimenting.

For some weird reason it seems to work up to half-way, then it reverts back to high throttle.

image

libraries/AP_TECS/AP_TECS.cpp
peterbarker: Fixed.  The reformatting here was just a drive-by when we were trying to wedge the energies into here.

PR #30094, opened 2025-05-18
Move option bits from SERIALn_OPTIONS to MAVn_OPTIONS

peterbarker:
This is built on top of https://github.com/ArduPilot/ardupilot/pull/29546


libraries/GCS_MAVLink/GCS.h
```diff
@@ -512,6 +512,23 @@ class GCS_MAVLINK
     virtual bool persist_streamrates() const { return false; }
     void handle_request_data_stream(const mavlink_message_t &msg);
 
+    AP_Int16 options;
+    enum class Option : uint16_t {
+        MAVLINK2_SIGNING_DISABLED = (1U << 0),
+        NO_FORWARD                = (1U << 1),  // don't forward MAVLink data to or from this device
+        NOSTREAMOVERRIDE          = (1U << 2),  // don't allow GCS to override streamrates
```
timtuxworth, line 519: It's not just GCS is it? I think this would also disable stream rate requests from another vehicle, a thing I have done.

libraries/GCS_MAVLink/GCS_Common.cpp
```diff
@@ -152,11 +152,25 @@ bool GCS_MAVLINK::init(uint8_t instance)
         return false;
     }
 
+    // PARAMETER_CONVERSION - Added: May-2025 for ArduPilot-4.7
+    // convert parameters; we used to use bits in the UARTDriver to
+    // remember whether the mavlink connection on that interface was
+    // "private" or not, and whether to ignore streamrate sets GCSs.
```
timtuxworth, line 158: Again - not sure if this is just stream rate sets from a GCS only.

timtuxworth (commented):
Just two minor points about the comments. I know it's possible to request stream rate changes from another vehicle. I'm pretty sure the disable stream rate request would apply to those requests too, not just requests from a GCS.

libraries/GCS_MAVLink/GCS.h
peterbarker: Reworded, thanks!

libraries/GCS_MAVLink/GCS_Common.cpp
peterbarker: Reworded this too - thanks!

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
```diff
@@ -202,6 +202,24 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
     // @RebootRequired: True
     // @User: Advanced
     AP_GROUPINFO("_ADSB",   10, GCS_MAVLINK, streamRates[GCS_MAVLINK::STREAM_ADSB], DRATE(GCS_MAVLINK::STREAM_ADSB)),
+
+    // ------------
+    // IMPORTANT: Add new stream rates *before* the _OPTIONS parameter.
```
IamPete1, line 207: Why?

IamPete1 (commented):
I think this basically looks good.

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
peterbarker: Just to group them together.  I know it's not *required*, but it's nicer when you're looking for things.

libraries/GCS_MAVLink/GCS_MAVLink_Parameters.cpp
```diff
@@ -202,6 +202,24 @@ const AP_Param::GroupInfo GCS_MAVLINK::var_info[] = {
...
+    // @Description: Bitmask for configuring this telemetry channel. For having effect on all channels, set the relevant mask in all MAVx_OPTIONS parameters. Keep in mind that part of the flags may require a reboot to take action.
+    // @RebootRequired: True
+    // @User: Standard
+    // @Bitmask: 0:Accept unsigned MAVLink2 messages
+    AP_GROUPINFO("_OPTIONS",   20, GCS_MAVLINK, options, 0),
+
+    // PARAMETER_CONVERSION - Added: May-2025 for ArduPilot-4.7
+    // Hidden param used as a flag for param conversion
+    // This allows one time conversion while allowing user to flash between versions with and without converted params
+    AP_GROUPINFO_FLAGS("_OPTIONSCNV",   20, GCS_MAVLINK, options_were_converted, 0, AP_PARAM_FLAG_HIDDEN),
```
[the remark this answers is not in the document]
peterbarker: We do not call `AP_Param::check_group_info` on objects loaded via `load_object_from_eeprom` - so only validation on the base structures.

It looks like actually calling `check_group_info` would require us passing more information into `load_object_from_eeprom` - the current parameter-name-prefix length and the currently shift value - so adding this sanity check is a non-trivial task!

Further complicating things is that these problems may arise because a script is trying to define parameters.  Our current `FATAL("...")` approach for invalid structures isn't suitable in this case!

As to why Everything Worked: they both get shoved into eeprom, differentiated by parameter type:
```
00d2: type 1 (INT8) key 258 group_element 1292 size 1 value 1
 01
00d7: type 2 (INT16) key 258 group_element 1292 size 2 value 0
 00 00
```

Probably not something to rely on... I've moved `_OPTIONSCNV` to parameter 21 and re-tested parameter upgrade

tridge (changes requested):
missing change to SERIALn_OPTIONS meta-data

tridge:
@rmackay9 I'd like you to look at this one with regards to user support, and the old SERIALn_OPTIONS parameter still being there, but two bits have moved. Normally users know to look for a different parameter when the old parameter disappears, in this case they need to notice the docs

peterbarker:
OK.  So we now move to 32 bits for `SERIALn_OPTIONS` as part of this PR.

Older firmwares will continue to look at the 16-bit parameters, newer firmwares the 32-bit parameter.  That allows us to zero out the 2 bits in SERIALn_OPTIONS as we upgrade the parameters, and raise a pre-arm check if we ever see them non-zero.

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -1581,6 +1581,13 @@ bool AP_Arming::serial_protocol_checks(bool display_failure)
        check_failed(display_failure, "Multiple SERIAL ports configured for RC input");
        return false;
     }
+    char failure_msg[100] = {};
+    if (!AP::serialmanager().pre_arm_checks(failure_msg, ARRAY_SIZE(failure_msg))) {
+        if (failure_msg[0] != '\0') {
```
IamPete1: Why this check? Looks like it should always fill in the message if it returns false.

peterbarker: It came along from where I swiped this from :-)

Removed it, thanks!

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -272,8 +272,8 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
...
-    // @Bitmask: 0:InvertRX, 1:InvertTX, 2:HalfDuplex, 3:SwapTXRX, 4: RX_PullDown, 5: RX_PullUp, 6: TX_PullDown, 7: TX_PullUp, 8: RX_NoDMA, 9: TX_NoDMA, 10: Don't forward mavlink to/from, 11: DisableFIFO, 12: Ignore Streamrate
+    // @Description: Control over UART options. The InvertRX option controls invert of the receive pin. The InvertTX option controls invert of the transmit pin. The HalfDuplex option controls half-duplex (onewire) mode, where both transmit and receive is done on the transmit wire. The Swap option allows the RX and TX pins to be swapped on STM32F7 based boards.  NOTE that two bits have moved from this parameter into MAVn_OPTIONS!
+    // @Bitmask: 0:InvertRX, 1:InvertTX, 2:HalfDuplex, 3:SwapTXRX, 4: RX_PullDown, 5: RX_PullUp, 6: TX_PullDown, 7: TX_PullUp, 8: RX_NoDMA, 9: TX_NoDMA, 10: UNUSED OPTION (was Don't forward mavlink to/from); see MAVn_OPTIONS, 11: DisableFIFO, 12: UNUSED OPTION (was Ignore Streamrate); see MAVn_OPTIONS
```
rmackay9: normally we would leave the option's old name and add (deprecated for 4.x)

rmackay9: Ignore Streamrate (deprecated >4.7)

tridge (approved):

PR #30098, opened 2025-05-18
Dds quadplane takeoff

Ryanf55:
# Purpose

Allow VTOL takeoff on quadplane using DDS

# Instructions
In ros workspace
```
ros2 launch ardupilot_sitl micro_ros_agent.launch.py
```
In ardupilot
```
./Tools/autotest/sim_vehicle.py -v plane -f quadplane --console --map --enable-dds -DG
```
In a CLI, switch to guided, arm, takeoff, and send a goal location showing switch to forward flight
```
ros2 service call /ap/mode_switch ardupilot_msgs/srv/ModeSwitch "{mode: 15}" # GUIDED
ros2 service call /ap/arm_motors ardupilot_msgs/srv/ArmMotors "{arm: true}"
ros2 service call /ap/experimental/takeoff ardupilot_msgs/srv/Takeoff "{alt: 20.0}"
ros2 topic pub /ap/cmd_gps_pose ardupilot_msgs/msg/GlobalPosition "{header: {frame_id: map}, latitude: -35.36078618, longitude: 149.16231842, altitude: 600.0, coordinate_frame: 5}"
```

### DIFF
--- a/ArduPlane/Plane.cpp
+++ b/ArduPlane/Plane.cpp
@@ -1009,6 +1009,14 @@ bool Plane::is_taking_off() const
     return control_mode->is_taking_off();
 }
 
+bool Plane::start_takeoff(const float alt) {
+#if HAL_QUADPLANE_ENABLED
+    return quadplane.do_user_takeoff(alt);
+#else
+    return false;
+#endif
+}
+
 // correct AHRS pitch for PTCH_TRIM_DEG in non-VTOL modes, and return VTOL view in VTOL
 void Plane::get_osd_roll_pitch_rad(float &roll, float &pitch) const
 {
--- a/ArduPlane/Plane.h
+++ b/ArduPlane/Plane.h
@@ -1003,6 +1003,8 @@ private:
     bool verify_nav_wp(const AP_Mission::Mission_Command& cmd);
 #if HAL_QUADPLANE_ENABLED
     bool verify_landing_vtol_approach(const AP_Mission::Mission_Command& cmd);
+    // vtol takeoff from AP_Vehicle for quadplane.
+    bool start_takeoff(const float alt) override;
 #endif
     void do_wait_delay(const AP_Mission::Mission_Command& cmd);
     void do_within_distance(const AP_Mission::Mission_Command& cmd);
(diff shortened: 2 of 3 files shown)


timtuxworth:
It seems like this actually adds a more generic "plane_takeoff" method for DDS, but initially only implements VTOL, but could implement fixed wing takeoff at some point in the future. I guess my gripe is that the PR says it's only implementing quadplane takeoff, but it's actually more than that, which is actually a good thing.

Any reason not to do it now @Ryanf55 ? Would it not need to simply switch to TAKEOFF mode?

ArduPlane/Plane.h
```diff
@@ -1003,6 +1003,8 @@ class Plane : public AP_Vehicle {
     bool verify_nav_wp(const AP_Mission::Mission_Command& cmd);
 #if HAL_QUADPLANE_ENABLED
     bool verify_landing_vtol_approach(const AP_Mission::Mission_Command& cmd);
+    // vtol takeoff from AP_Vehicle for quadplane.
+    bool start_takeoff(const float alt) override;
```
timtuxworth, line 1007: The method as implemented doesn't need to be limited to QUADPLANE_ENABLED.

timtuxworth (commented):
This seems to be a nicely generic function that could be fleshed out to work for fixed wing too.

ArduPlane/Plane.h
Ryanf55: For fixed wing takeoff, we already recommend using TAKEOFF mode via the ModeSwitch service, for now, and everyone seems happy with it.

ArduPlane/Plane.cpp
```diff
@@ -1009,6 +1009,14 @@ bool Plane::is_taking_off() const
     return control_mode->is_taking_off();
 }
 
+bool Plane::start_takeoff(const float alt) {
+#if HAL_QUADPLANE_ENABLED
+    return quadplane.do_user_takeoff(alt);
```
IamPete1, line 1014:
```suggestion
    return plane.quadplane.available() && quadplane.do_user_takeoff(alt);
```
`do_user_takeoff` doesn't check if quadplane is enabled. It might be a better fix to do the check in that function rather than here.

Ryanf55: Thanks. I applied here for now, we can discuss at dev call which class is responsible for checking if quadplane is available.

tridge (approved):

PR #30101, opened 2025-05-18
Copter: Mode Follow: Update to use input_pos_vel_accel commands

lthall:
This PR builds on recent updates to the Follow library. It replaces the original velocity-follow implementation—which used a P-term applied to a distance vector—with a more modern approach using the input_pos_vel_accel() interface to the position controller.

Key Improvements:
- The aircraft now follows an intercept path rather than simply chasing the target.
- No tuning required — motion shaping and aggressiveness are determined by the controller's internal constraints (accel/jerk limits).
- Enables smoother, more optimal convergence, especially during dynamic movement.


rmackay9 (approved):

tridge (commented):
we have a race condition (not caused by this PR) in update_estimates(), need a follow-up PR to fix this

PR #30109, opened 2025-05-19
Tools: install tornado for mavcesium

Ryanf55:


peterbarker:
I'm not 100% this is a quick-commit.

This package is only used for MAVProxy, not the ArduPilot project.  While it's small, it's also something else to go wrong with ArduPilot's infrastructure now.

Should we instead vastly improve MAVProxy's diagnostics so that it's made very clear which package the user has to install?

peterbarker:
.... fixing it in MAVProxy-land also means it's fixed for all users, not just Ubuntu users.

Ryanf55:
> .... fixing it in MAVProxy-land also means it's fixed for all users, not just Ubuntu users.

This is a pattern we have done for [wsproto](https://github.com/ArduPilot/ardupilot/pull/28560)? Do you want to change that?

peterbarker:
.... doing this in ArduPilot rather than MAVProxy means it works for people not developing ArduPilot but just running MAVProxy as a GCS.

khancyr:
I kind of agree with Peter. 
On medium term we would better start adding install dependency correctly on pymavlink and mavproxy to make dependencies easier on ardupilot. 

Another point to take into account is that the install scripts are generally run only once for initial install so it won't help user much if the dependency isn't install/updated in same time as mavproxy or pymavlink. 

And this should be probably restricted to venv install as this will mostly break all other install that depends on Tornado (e.g. [rosbridge_suite](https://github.com/RobotWebTools/rosbridge_suite)) 

Otherwise go for it. It shouldn't be much blocking and we can handle the mess if we break some environment until we build up everything

Ryanf55:
> I kind of agree with Peter. On medium term we would better start adding install dependency correctly on pymavlink and mavproxy to make dependencies easier on ardupilot.
> 
> Another point to take into account is that the install scripts are generally run only once for initial install so it won't help user much if the dependency isn't install/updated in same time as mavproxy or pymavlink.
> 
> And this should be probably restricted to venv install as this will mostly break all other install that depends on Tornado (e.g. [rosbridge_suite](https://github.com/RobotWebTools/rosbridge_suite))
> 
> Otherwise go for it. It shouldn't be much blocking and we can handle the mess if we break some environment until we build up everything

the setup scripts already breaks ROS development on ubuntu 22. AP should always go in a venv, unless a user passes a flag they want to pollute their host system.

tridge:
discussion still ongoing, won't merge for now

peterbarker:
OK.... I wonder where the discussion is happening.  Could someone summarise said discussion here, please?  :-)

PR #30112, opened 2025-05-19
Tracker w/o GPS

landswellsong:
The documentation implies that Tracker can work without a GPS unit installed by picking its location from paramters/EEPROM/home. When the tracker is actually set up like this, the `COMPASS_AUTODEC` code is never reached and `AP_Arming` spams the console with the complaints about home not being set.

This PR introduces the following:
1. Compass initial position can now be set explicitly and the flag indicating if it's set or not can be cleared from other code.
2. If Tracker has a start-up location (parameters/EEPROM) it calls set the aforementioned method which sets the declination off that location (provided `COMPASS_AUTODEC` is set etc).
3. Subsequent changes to the Tracker location while it's not using the GPS call the same method again to refresh the declination.
4. If the Tracker receives a real GPS update, it invalidates the previously set initial location flag which should restore the previous compass code path on the next update.
5. As a quality of life, we no longer complain about AHRS home in Tracker. Note that the Tracker does not have an accessible `ARMING_CHECK` parameter so it wasn't possible for the user to silence this message by just unsetting the corresponding flag.

### DIFF
--- a/AntennaTracker/system.cpp
+++ b/AntennaTracker/system.cpp
@@ -1,3 +1,4 @@
+#include "AP_Compass/AP_Compass.h"
 #include "Tracker.h"
 
 // mission storage
@@ -65,6 +66,12 @@ void Tracker::init_ardupilot()
     // see if EEPROM has a default location as well
     if (current_loc.lat == 0 && current_loc.lng == 0) {
         get_home_eeprom(current_loc);
+    }
+
+    // force compass declination update if starting coordinates are given
+    if (current_loc.lat != 0 || current_loc.lng != 0) {
+        AP::compass().set_initial_location((float)current_loc.lat / 10000000,
+                                           (float)current_loc.lng / 10000000);
     }
 
     hal.scheduler->delay(1000); // Why????
@@ -137,6 +144,9 @@ bool Tracker::set_home(const Location &temp, bool lock)
 
     current_loc = temp;
 
+    // update compass declination with the new location
+    AP::compass().set_initial_location((float)current_loc.lat / 10000000,
+                                       (float)current_loc.lng / 10000000);
     return true;
 }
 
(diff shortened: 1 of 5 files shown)


IamPete1:
I wonder if we should set the EKF origin with the initial location, that might be enough to get everything else working. 

> 5\. As a quality of life, we no longer complain about AHRS home in Tracker. Note that the Tracker does not have an accessible `ARMING_CHECK` parameter so it wasn't possible for the user to silence this message by just unsetting the corresponding flag.

As far as I can tell tracker never runs the arming checks, the only way would be if the GCS asks for a arming check to be run. Its certainly possible that Mission planner does that be default.

IamPete1:
> Unfortunately this isn't the case. Reproduction scenario, start a Tracker SITL with a param file like this:

Your right, its coming from the vehicle arming update. In any-case as your screen grab shows tracker thinks its armed, so it should not be running pre-arm checks. The problem is that tracker is not bringing the arming lib along for the ride when it arms.

This is the "arming" functions:
https://github.com/ArduPilot/ardupilot/blob/2565aa549313675212fa8520023b41d0113f0bdb/AntennaTracker/system.cpp#L143-L157

 There is no real way that we can just tell arming that the vehicle is armed, it always wants to run some checks. Ultimately I think we have to either re-work tracker to arm in the same way that the other vehicles do and use the arming library correctly or we should remove arming from tracker and keep it the way it is. 

We should be able to set the `AP_ARMING_ENABLED` define false and remove all of AP Arming.

IamPete1:
It would be nice to move towards the other vehicles. AP Arming just recently got a "Arm when ready option" for rover tracker could use something like that (https://github.com/ArduPilot/ardupilot/pull/29066). However, that might be a bit of a bigger job that just removing AP Arming altogether. I think we would accept either, although we should consult @peterbarker  too.

landswellsong:
Rebased over `master`. @peterbarker can I ask you to review this?

PR #30115, opened 2025-05-20
Document IE24 and IEFC log messages 

peterbarker:


libraries/AP_Generator/AP_Generator_IE_2400.cpp
```diff
@@ -480,6 +480,15 @@ void AP_Generator_IE_2400::log_write()
...
         case ProtocolVersion::LEGACY:
+            // @LoggerMessage: IE24
+            // @Description: Intelligent Energy Fuel Cell generator (legacy protocol)
+            // @URL: https://ardupilot.org/copter/docs/common-ie24-fuelcell.html
+            // @Field: TimeUS: Time since system startup
+            // @Field: FUEL: Fuel remaining
+            // @Field: SPMPWR: stack power module power draw
+            // @Field: POUT: output power
+            // @Field: ERR: error codes
+            // @FieldValueEnum: ERR: AP_Generator_IE_2400::ErrorCode
```
Georacer, line 491: This would be even better if it contained the individual values.

Georacer: Or does it populate automatically from the type name?

Georacer (approved):

libraries/AP_Generator/AP_Generator_IE_2400.cpp
peterbarker: > Or does it populate automatically from the type name?

This

PR #30116, opened 2025-05-20
AP_Scripting: param-set.lua: trim out unused MAV_SYSID parameter reference

peterbarker:
We discovered we didn't need to filter to messages just to this system ID, the bindings do that


Georacer (approved):
I haven't verified whether the script receives only its own SYSID param changes, but the variable is indeed unused.

timtuxworth (approved):
Makes sense.

PR #30075, opened 2025-05-15
Disable ECC flash erase on bootloaders using flash for parameter storage

andyp1per:
Just rebuilt all the H7 bootloaders that use flash for parameter storage

PR #30118, opened 2025-05-20
AP 4.6.0 stable release

rmackay9:
This is the 4.6.0 official release PR for Plane, Copter, Rover and Tracker

Currently this only includes these two changes:

- https://github.com/ArduPilot/ardupilot/pull/30031
- https://github.com/ArduPilot/ardupilot/pull/30045

Could someone remind me if we wanted to include other changes from [the 4.6 Project](https://github.com/orgs/ArduPilot/projects/14/views/2)?  Maybe this one?

- [ ] https://github.com/ArduPilot/ardupilot/pull/30087


andyp1per:
@rmackay9 I think we want this one - https://github.com/ArduPilot/ardupilot/pull/30075 - currently all the bootloaders are messed up. @tridge I don't agree with waiting for @tpwrules investigation - the whole point of not doing a proper fix was because we couldn't wait. Now we have the worst of all worlds.

tridge:
@andyp1per what we discussed on the dev call was to just update the bootloader binary for the TBS board from master, but not apply the other patches. I know that leaves us with a bad ECC system, but we've had it for quite a while and only hit the issue for this one board. We will hopefully be able to fix it properly for 4.6.1, but we don't want to delay the release any more

rmackay9:
@andyp1per @tridge,

thanks for that feedback, I've updated the PR to only include the TBS bootloader binary update.

ArduPlane/ReleaseNotes.txt
```diff
@@ -1,5 +1,14 @@
 ArduPilot Plane Release Notes:
 ------------------------------------------------------------------
+Release 4.6.0 21-May-2025
+
+Changes from 4.6.0-beta6
+
+1) Bug Fixes
+
+- Firmware not erased on ECC error on boards using parameter area for storage
```
tridge, line 9: should say only the TBS board changed

tridge (changes requested):

ArduPlane/ReleaseNotes.txt
rmackay9: Thanks for catching that, release notes fixed!

tridge (approved):

andyp1per:
> @andyp1per what we discussed on the dev call was to just update the bootloader binary for the TBS board from master, but not apply the other patches. I know that leaves us with a bad ECC system, but we've had it for quite a while and only hit the issue for this one board. We will hopefully be able to fix it properly for 4.6.1, but we don't want to delay the release any more

Understand, but IMO the reason for that is that virtually no-one who is beta testing 4.6 will update their bootloader - so its a bear trap that will only snowball after release.

PR #29783, opened 2025-04-15
GCS_MAVLink: remove 'No ap_message for mavlink id' warning

peterbarker:
Even just the first response to a probing message is annoying to the user

PR #30124, opened 2025-05-20
GCS_MAVLink: set message interval fail msg only sent on SITL

rmackay9:
This follow-up to https://github.com/ArduPilot/ardupilot/pull/29783 stops another "No ap_message for mavlink id (xxx)" status text message from being sent when a set-message-interval request fails except for on SITL where we continue to send it.

Thanks to @Hwurzburg for finding this

This has been tested in SITL and on real hardware to ensure that the message appears before the change but does not appear afterwards.  For that testing I used Mission Planner's Ctrl-F screen to pull up the "Message Interval" screen and then chose an unsupported message (e.g. ACTUATOR_CONTROL_TARGET) and set the interval to 3hz
image

Before this change we see the message appear, after the change it does not
image


peterbarker (approved):

rmackay9:

PR #30126, opened 2025-05-21
Add pre-arm check that system is initialised

peterbarker:
Not really because we care about initialised, just stops us running code when we don't want to when a prearm check is forced from the GCS.

### DIFF
--- a/Rover/AP_Arming_Rover.cpp
+++ b/Rover/AP_Arming_Rover.cpp
@@ -78,6 +78,11 @@ bool AP_Arming_Rover::pre_arm_checks(bool report)
         return true;
     }
 
+    if (!hal.scheduler->is_system_initialized()) {
+        check_failed(report, "System not initialised");
+        return false;
+    }
+
     //are arming checks disabled?
     if (checks_to_perform == 0) {
         return mandatory_checks(report);


peterbarker:
```
AP: Calibrating barometer
Got COMMAND_ACK: RUN_PREARM_CHECKS: ACCEPTED
AP: PreArm: System not initialised
Got COMMAND_ACK: RUN_PREARM_CHECKS: ACCEPTED
MANUAL> Mode MANUAL
AP: Barometer 1 calibration complete
```

rmackay9:
Looks ok.  Whenever we add a new arming check we should add it to this page (in alphabetical order) https://ardupilot.org/copter/docs/common-prearm-safety-checks.html

ArduCopter/AP_Arming_Copter.cpp
```diff
@@ -21,6 +21,11 @@ bool AP_Arming_Copter::run_pre_arm_checks(bool display_failure)
         return true;
     }
 
+    if (!hal.scheduler->is_system_initialized()) {
+        check_failed(display_failure, "System not initialised");
```
rmackay9, line 25: I'd change the wording to, "system initialising"

peterbarker:
I've marked this for backport; one of our partners hit it on 4.6.2

PR #30128, opened 2025-05-21
AP_Arming:expand metadata on ARMING_REQUIRE

Hwurzburg:


libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -115,9 +115,9 @@ const AP_Param::GroupInfo AP_Arming::var_info[] = {
...
+    // @Description: Arming disabled until some requirements are met. If 0, there are no requirements (arm immediately).  If 1, sends the minimum throttle PWM value to the throttle channel when disarmed. If 2, send 0 PWM (no signal) to throttle channel when disarmed. IF 3 (Rover only) Arm immendiately once pre-arm/arm checks are satisfied, but only one time per boot up. On planes with ICE enabled and the throttle while disarmed option set in ICE_OPTIONS, the motor will always get THR_MIN when disarmed. Arming will occur using either rudder stick arming (if enabled) or GCS command when all mandatory and ARMING_CHECK items are satisfied. Note, when setting this parameter to 0, a reboot is required to immediately arm the plane.
```
peterbarker, line 118:
```suggestion
    // @Description: Arming disabled until some requirements are met. If 0, there are no requirements (arm immediately).  If 1, sends the minimum throttle PWM value to the throttle channel when disarmed. If 2, send 0 PWM (no signal) to throttle channel when disarmed. If 3 (Rover only) Arm immediately once pre-arm/arm checks are satisfied, but only one time per boot up. On planes with ICE enabled and the throttle while disarmed option set in ICE_OPTIONS, the motor will always get THR_MIN when disarmed. Arming will occur using either rudder stick arming (if enabled) or GCS command when all mandatory and ARMING_CHECK items are satisfied. Note, when setting this parameter to 0, a reboot is required to immediately arm the plane.
```

Note that we *can* do per-vehicle @Description and it is looking like that might be a good idea as there's both Plane and Rover-specific stuff in here now.

peterbarker (changes requested):

libraries/AP_Arming/AP_Arming.cpp
Hwurzburg: I have split them up..thanks

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -115,9 +115,10 @@ const AP_Param::GroupInfo AP_Arming::var_info[] = {
...
+    // @Description{Plane}: Arming disabled until some requirements are met. If 0, there are no requirements (arm immediately).  If 1, sends the minimum throttle PWM value to the throttle channel when disarmed. If 2, send 0 PWM (no signal) to throttle channel when disarmed. On planes with ICE enabled and the throttle while disarmed option set in ICE_OPTIONS, the motor will always get THR_MIN when disarmed. Arming will occur using either rudder stick arming (if enabled) or GCS command when all mandatory and ARMING_CHECK items are satisfied. Note, when setting this parameter to 0, a reboot is required to immediately arm the plane.
+    // @Description{Plane, Rover}: Arming disabled until some requirements are met. If 0, there are no requirements (arm immediately).  If 1, sends the minimum throttle PWM value to the throttle channel when disarmed. If 3, Arm immediately once pre-arm/arm checks are satisfied, but only one time per boot up. Arming will occur using either rudder stick arming (if enabled) or GCS command when all mandatory and ARMING_CHECK items are satisfied. Note, a reboot is required when setting to 0 or 3 to automatically arm the vehicle.
```
peterbarker, line 119:
```suggestion
    // @Description{Rover}: Arming disabled until some requirements are met. If 0, there are no requirements (arm immediately).  If 1, all checks specified by ARMING_CHECKS must pass before the vehicle can be armed (for example, via rudder stick or GCS command).  If 3, Arm immediately once pre-arm/arm checks are satisfied, but only one time per boot up.  Note that a reboot is NOT required when setting to 0 but IS require when setting to 3.
```

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -115,9 +115,10 @@ const AP_Param::GroupInfo AP_Arming::var_info[] = {
...
+    // @Description{Plane}: Arming disabled until some requirements are met. If 0, there are no requirements (arm immediately).  If 1, sends the minimum throttle PWM value to the throttle channel when disarmed. If 2, send 0 PWM (no signal) to throttle channel when disarmed. On planes with ICE enabled and the throttle while disarmed option set in ICE_OPTIONS, the motor will always get THR_MIN when disarmed. Arming will occur using either rudder stick arming (if enabled) or GCS command when all mandatory and ARMING_CHECK items are satisfied. Note, when setting this parameter to 0, a reboot is required to immediately arm the plane.
```
peterbarker, line 118:
```suggestion
    // @Description{Plane}: Arming disabled until some requirements are met. If 0, there are no requirements (arm immediately).  If 1, sends the minimum throttle PWM value to the throttle channel when disarmed. If 2, send 0 PWM (no signal) to throttle channel when disarmed. On planes with ICE enabled and the throttle while disarmed option set in ICE_OPTIONS, the motor will always get THR_MIN when disarmed. Arming will be blocked until all mandatory and ARMING_CHECK items are satisfied; arming can then be accomplished via (eg.) rudder gesture or GCS command.
```

That bit is actively dangerous!

peterbarker (changes requested):

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -115,9 +115,10 @@ const AP_Param::GroupInfo AP_Arming::var_info[] = {
...
+    // @Description{Rover}: Arming disabled until some requirements are met. If 0, there are no requirements (arm immediately).  If 1, all checks specified by ARMING_CHECKS must pass before the vehicle can be armed (for example, via rudder stick or GCS command).  If 3, Arm immediately once pre-arm/arm checks are satisfied, but only one time per boot up.  Note that a reboot is NOT required when setting to 0 but IS require when setting to 3.
     // @Values{Plane}: 0:Disabled,1:minimum PWM when disarmed,2:0 PWM when disarmed
-    // @Values{Rover}: 0:No,1:Yes,3:AutoArmOnce
+    // @Values{Rover}: 0:No,1:minimum PWM when disarmed,3:AutoArmOnce
```
rmackay9, line 121: how about, "1:Yes (min PWM when disarmed)"

Hwurzburg: increased verbosity on all values

rmackay9:
Merged, thanks!

PR #30129, opened 2025-05-21
Tools: add -Wno-format-security to littlefs waf build flags

deej-io:
The littlefs module failed to build with some compilers as the -Wformat-security flag, which is enabled globally, requires the -Wformat flag, which was disabled specifically for this module.


andyp1per (approved):
Thanks!

PR #30130, opened 2025-05-21
Plane: add support for NAV_DELAY if flying auto mission in VTOL mode

jeff567:
NAV_DELAY is currently only supported if the plane is disarmed.  This pull request allows for NAV_DELAY to be used if flying an auto mission with a quadplane in VTOL mode.

### DIFF
--- a/ArduPlane/commands_logic.cpp
+++ b/ArduPlane/commands_logic.cpp
@@ -893,8 +893,8 @@ bool ModeAuto::verify_altitude_wait(const AP_Mission::Mission_Command &cmd)
 // verify_nav_delay - check if we have waited long enough
 bool ModeAuto::verify_nav_delay(const AP_Mission::Mission_Command& cmd)
 {
-    if (AP::arming().is_armed_and_safety_off()) {
-        // don't delay while armed, we need a nav controller running
+    if (AP::arming().is_armed_and_safety_off() && !plane.quadplane.in_vtol_mode()) {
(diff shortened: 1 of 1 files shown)


peterbarker:
Could you work up an autotest for this, please?

jeff567:
sure, just added an autotest in quadplane.  It performs an auto mission with two NAV_DELAY commands.  The first one happens when the aircraft is in VTOL mode (so shouldn't be ignored).  The aircraft then transitions to fixed-wing mode before the reaching the second NAV_DELAY command, which should be skipped.

peterbarker:
> sure, just added an autotest in quadplane. It performs an auto mission with two NAV_DELAY commands. The first one happens when the aircraft is in VTOL mode (so shouldn't be ignored). The aircraft then transitions to fixed-wing mode before the reaching the second NAV_DELAY command, which should be skipped.

Thanks for that!

I've increasingly taken to putting the missions in-line in the test - we have new-ish function to upload simple missions - keeps things a little more cohesive.  What you've got is serviceable, however.

jeff567:
> @jeff567 need to compile when quadplane is disabled, @jeff567 :-)

Sorry, I'll fix that in a bit.  That's to say we should only check plane.quadplane.in_vtol_mode() if HAL_QUADPLANE_ENABLED is true, right?

peterbarker:
Ping @jeff567 - still working on this?

jeff567:
apologies for the delay, how does this look?

peterbarker (approved):
LGTM

PR #30132, opened 2025-05-22
Document simulated Blimp log messages

peterbarker:


peterbarker:
> The SRT2 one is a bit odd - zero? But essentially all of these messages are actually to try to debug what's wrong with the 

Yeah, I don't know what is supposed to be in thre, and while I was trying to work it out I did see that we don't assign values in vector before using it.  So "zero" is accurate.  No idea what the *intent* for those values was :-)

> pendulum stability code (for the ones under the #if 0 sections I would think autotest wouldn't pick it up?). I just left them there for either myself/whoever comes next to try to fix this.

These are all picked up by the sanity checker.

> Thus, not sure if we really need the documentation for it? It does make the code a bit harder to read... Maybe I should just change the #if LOGGING_ENABLED to an #if 0 as well so it doesn't bother autotest? 'Cause the messages are really only there for debugging if anyone's trying to improve this vehicle's simulation, not really for general use.

Yes, so I can kill the exception list :-)

These should also be a help for the next person who works on this - it's more documentation and quite good at that as it's less likely to drift than other comments!

libraries/SITL/SIM_Blimp.cpp
```diff
@@ -143,6 +197,12 @@ void Blimp::calculate_forces(const struct sitl_input &input, Vector3f &body_acc,
 
 #if 0 //"Wobble" attempt
   rot_T.y = fin[0].Fz * radius + fin[1].Fz * radius;
+  // @LoggerMessage: SRT2
+  // @Description: Transformed Simulated Blimp Rotational forces
+  // @Field: TimeUS: Time since system startup
+  // @Field: rtx: zero
+  // @Field: rty: zero
```
MichelleRos, line 204: _rty_ is the wobble rotational forces - set in the line right above this addition.

peterbarker: Oops!  Fixed, thanks!

MichelleRos (approved):
LGTM

PR #30137, opened 2025-05-22
hwdef:revert osd type change

Hwurzburg:
this autopilot does not have an onboard OSD chip and should not have had the correction


peterbarker (approved):
LGTM

rmackay9:

PR #30140, opened 2025-05-22
Fix AP_DDS init on ChibiOS with GPS lock

### LINKED ISSUE #25535: AP_DDS: MicroROS agent doesn't publish topics when GPS automatically gets GPS lock.
## Bug report
AP_DDS: MicroROS agent doesn't publish topics when GPS automatically gets GPS lock.

We have Ardupilot running on a PX4 with a CubeOrange and a Here3+ GPS. We followed the documentation in the AP_DDS library to setup DDS/MicroROS: https://github.com/ArduPilot/ardupilot/tree/master/libraries/AP_DDS

We got it all working, where the data from the Here3+ could be published as topics to a ROS domain, however this is only the case when the Here3+ starts in a flashing blue "Disarmed with no GPS lock" state. When it is outside it automatically achieves GPS lock and flashes green "Disarmed (ready to arm), GPS lock acquired". See documentation here: https://docs.cubepilot.org/user-guides/here-3/here-3-manual

When the GPS starts without GPS lock the console output looks like the following:
Untitled

When the GPS starts with GPS lock the console output looks like this:
image

**Version**
4.4 dev version

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[ X ] Rover
[  ] Submarine

**Hardware type**
Pixhawk4 with CubeOrange and a Here3+

**Logs**
Untitled
imageimage

Ryanf55:
If the RTC doesn't have a source, AP_DDS initialization will currently fail. This patch blocks AP_DDS from starting till the RTC has a source.

A longer term improvement would be to lock AP_DDS onto an RTC source once it starts and never switch. If the RTC is expected to never have a source, AP_DDS will just use the millis() for the entire duration.

This is easily tested in SITL. Start sitl with AP_DDS, 
```
param set SIM_GPS1_ENABLE 0
reboot
```

Wait   a while, observe AP_DDS blocked from initialization, and then set SIM_GPS1_ENABLE  to 1 to observe it boot.

Tested this RTC patch on a vehicle running 4.5.

Closes #25535

Needs backport

Sponsored by AeroVironment, Inc.

### DIFF
--- a/libraries/AP_DDS/AP_DDS_Client.cpp
+++ b/libraries/AP_DDS/AP_DDS_Client.cpp
@@ -720,18 +720,18 @@ bool AP_DDS_Client::update_topic(ardupilot_msgs_msg_Status& msg)
     uint8_t fs_iter = 0;
     msg.failsafe_size = 0;
     if (rc().in_rc_failsafe()) {
-        msg.failsafe[fs_iter++] = FS_RADIO;
+        msg.failsafe[fs_iter++] = ardupilot_msgs_msg_Status_FS_RADIO;
     }
     if (battery.has_failsafed()) {
-        msg.failsafe[fs_iter++] = FS_BATTERY;
+        msg.failsafe[fs_iter++] = ardupilot_msgs_msg_Status_FS_BATTERY;
     }
     // TODO: replace flag with function.
     if (AP_Notify::flags.failsafe_gcs) {
-        msg.failsafe[fs_iter++] = FS_GCS;
+        msg.failsafe[fs_iter++] = ardupilot_msgs_msg_Status_FS_GCS;
     }
     // TODO: replace flag with function.
     if (AP_Notify::flags.failsafe_ekf) {
-        msg.failsafe[fs_iter++] = FS_EKF;
+        msg.failsafe[fs_iter++] = ardupilot_msgs_msg_Status_FS_EKF;
     }
     msg.failsafe_size = fs_iter;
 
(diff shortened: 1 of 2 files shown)


Ryanf55:
We need this in for 4.7. id like to work with Peter on the correct approach.

libraries/AP_DDS/AP_DDS_Client.cpp
```diff
@@ -720,18 +720,18 @@ bool AP_DDS_Client::update_topic(ardupilot_msgs_msg_Status& msg)
     uint8_t fs_iter = 0;
     msg.failsafe_size = 0;
     if (rc().in_rc_failsafe()) {
-        msg.failsafe[fs_iter++] = FS_RADIO;
+        msg.failsafe[fs_iter++] = ardupilot_msgs_msg_Status_FS_RADIO;
```
peterbarker, line 723: These look wrong?

peterbarker:
> If the RTC doesn't have a source, AP_DDS initialization will currently fail. This patch blocks AP_DDS from starting till the RTC has a source.

Seems OK.  What aspect of init fails?

> A longer term improvement would be to lock AP_DDS onto an RTC source once it starts and never switch. If the RTC is expected to never have a source, AP_DDS will just use the millis() for the entire duration.

That doesn't sound great; we have "levels" of goodness in AP_RTC for a reason, and it would be unfortunate to be locked into using a poor time source....

peterbarker:
Ping @Ryanf55 - review was left.  Planning on chasing this?

PR #30141, opened 2025-05-23
AC_Autotune: Fix gain load problem when mode set to FAILED

lthall:
This PR fixes a serious bug in the Multirotor Autotune where we break the primary rule "Gains are only changed from default (I excepted) during the twitch".

The code is set up such that when a test `step = ABORT;` is changed the gains are not changed immediately but instead on the next loop. When we also set `mode = FAILED;` we don't run the next loop.

This is where we set mode to FAILED:
https://github.com/ArduPilot/ardupilot/blob/d229c8c74655e2e9b9d42b95b212a0c286c8066a/libraries/AC_AutoTune/AC_AutoTune_Multi.cpp#L580

This is why the loop doesn't run again:
https://github.com/ArduPilot/ardupilot/blob/d229c8c74655e2e9b9d42b95b212a0c286c8066a/libraries/AC_AutoTune/AC_AutoTune.cpp#L303

### DIFF
--- a/libraries/AC_AutoTune/AC_AutoTune.cpp
+++ b/libraries/AC_AutoTune/AC_AutoTune.cpp
@@ -183,6 +183,9 @@ void AC_AutoTune::send_step_string()
     case TESTING:
         GCS_SEND_TEXT(MAV_SEVERITY_INFO, "AutoTune: Testing");
         return;
+    case COMPLETE:
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "AutoTune: Complete");
+        break;
     }
     GCS_SEND_TEXT(MAV_SEVERITY_INFO, "AutoTune: unknown step");
 }
@@ -385,11 +388,6 @@ void AC_AutoTune::control_attitude()
             step = TESTING;
             step_start_time_ms = now;
             step_time_limit_ms = get_testing_step_timeout_ms();
-            // set gains to their to-be-tested values
-            load_gains(GAIN_TEST);
-        } else {
-            // when waiting for level we use the intra-test gains
-            load_gains(GAIN_INTRA_TEST);
         }
 
         // Initialize test-specific variables
@@ -417,9 +415,6 @@ void AC_AutoTune::control_attitude()
 
     case TESTING: {
         // Run the twitching step
-        load_gains(GAIN_TEST);
-
-        // run the test
         test_run(axis, direction_sign);
 
         // Check for failure causing reverse response
@@ -567,9 +562,7 @@ void AC_AutoTune::control_attitude()
                     update_gcs(AUTOTUNE_MESSAGE_SUCCESS);
                     LOGGER_WRITE_EVENT(LogEvent::AUTOTUNE_SUCCESS);
                     AP_Notify::events.autotune_complete = true;
-
-                    // Return to original gains for landing
-                    load_gains(GainType::GAIN_ORIGINAL);
+                    step = COMPLETE;
                 } else {
                     AP_Notify::events.autotune_next_axis = true;
                     reset_update_gain_variables();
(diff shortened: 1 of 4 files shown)


libraries/AC_AutoTune/AC_AutoTune.cpp
```diff
@@ -586,16 +579,39 @@ void AC_AutoTune::control_attitude()
...
+
+    case COMPLETE:
+        // we don't need to do anything
+        break;
+    }
+
+    // Load the appropriate gains
+    switch (step) {
+    case WAITING_FOR_LEVEL:
+            // when waiting for level we use the intra-test gains
```
rmackay9, line 598: nitpick: we should probably reduce the whitespace by a few characters

tridge:
@lthall thanks for the angle P limit reduction for quadplanes! I do think it should be a parameter though. We've seen copters gets hugely excessive angle P which causes issues

libraries/AC_AutoTune/AC_AutoTune.h
```diff
@@ -208,7 +208,8 @@ class AC_AutoTune
         WAITING_FOR_LEVEL = 0,    // autotune is waiting for vehicle to return to level before beginning the next twitch
         TESTING           = 1,    // autotune has begun a test and is watching the resulting vehicle movement
         UPDATE_GAINS      = 2,    // autotune has completed a test and is updating the gains based on the results
-        ABORT             = 3     // load normal gains and return to WAITING_FOR_LEVEL
+        ABORT             = 3,    // load normal gains and return to WAITING_FOR_LEVEL
+        COMPLETE          = 4     // load original gains
```
peterbarker, line 212:
```suggestion
        COMPLETE          = 4,     // load original gains
```

libraries/AC_AutoTune/AC_AutoTune_Multi.cpp
```diff
@@ -58,7 +59,13 @@
 #define AUTOTUNE_FLTE_MIN                  2.5     // minimum Rate Yaw error filter value
 #define AUTOTUNE_RP_MIN                   0.01     // minimum Rate P value
 #define AUTOTUNE_RP_MAX                    2.0     // maximum Rate P value
-#define AUTOTUNE_SP_MAX                   40.0     // maximum Stab P value
+
+#if APM_BUILD_TYPE(APM_BUILD_ArduPlane)
```
andyp1per, line 63: I also feel a parameter for all vehicle types would be more appropriate here

lthall: Adding a parameter is beyond the scope of this pr. If reducing the default for plane isn't acceptable I am happy to remove it.

As it is I added it because Tridge requested it. However if this is something that is going to get in the way of an important bug fix, we can leave that to a separate PR.

libraries/AC_AutoTune/AC_AutoTune.h
```diff
@@ -207,7 +207,8 @@ class AC_AutoTune
         WAITING_FOR_LEVEL = 0,    // autotune is waiting for vehicle to return to level before beginning the next twitch
         TESTING           = 1,    // autotune has begun a test and is watching the resulting vehicle movement
         UPDATE_GAINS      = 2,    // autotune has completed a test and is updating the gains based on the results
-        ABORT             = 3     // load normal gains and return to WAITING_FOR_LEVEL
+        ABORT             = 3,    // load normal gains and return to WAITING_FOR_LEVEL
+        COMPLETE          = 4,     // load original gains
```
rmackay9, line 211: tiny whitespace issue here, an extra " "

rmackay9:
FYI, I've removed this from the 4.6 backports because it wasn't merged it seems

PR #30145, opened 2025-05-23
AP_RCTelemetry: rename sensor_status_flags to unhealthy_sensors

peterbarker:
the current name doesn't tell you much, and definitely doesn't hint that it's returning you which sensors are present and enabled but not healthy

No compiler output change
```
Board,plane
CubeOrangePlus,*
```

### DIFF
--- a/libraries/AP_RCTelemetry/AP_RCTelemetry.cpp
+++ b/libraries/AP_RCTelemetry/AP_RCTelemetry.cpp
@@ -99,7 +99,7 @@ uint8_t AP_RCTelemetry::run_wfq_scheduler(const bool use_shaper)
     float delay = 0;
     bool packet_ready = false;
 
-    // build message queue for sensor_status_flags
+    // build message queue for unhealthy_sensors()
     check_sensor_status_flags();
     // build message queue for ekf_status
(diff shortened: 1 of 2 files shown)


andyp1per (approved):

libraries/AP_RCTelemetry/AP_RCTelemetry.cpp
```diff
@@ -99,7 +99,7 @@ uint8_t AP_RCTelemetry::run_wfq_scheduler(const bool use_shaper)
     float delay = 0;
     bool packet_ready = false;
 
-    // build message queue for sensor_status_flags
```
rmackay9, line 102: this seems like a search replace mistake perhaps?

rmackay9 (changes requested):
sorry, I don't want to hold up progress, but I wonder slightly if this is an improvement.  No saying it need to come to a dev call but let's just think about it one more time.

peterbarker:
> sorry, I don't want to hold up progress, but I wonder slightly if this is an improvement. No saying it need to come to a dev call but let's just think about it one more time.

Well, I still think this is an improvement.

Issue #30142, opened 2025-05-23
Bogus Telemetry message about missing RC reciever


Hwurzburg:
Plane 4.6, we send occasional FRSKY/CRSF telem warnings about "No RC Receiver" when its obviously there and working from the logs....it does not appear in MAVLink texts to the GCS...only RC telem messages (it only apepars on TX Yapuu display, not in OSD )

peterbarker:
"No RC Receiver" is incorrect; this will show if we've seen an RC receiver but it is now unhealthy.

timtuxworth:
> "No RC Receiver" is incorrect; this will show if we've seen an RC receiver but it is now unhealthy.

I've seen this many times. I don't know what it means or what should I do about it. Usually I ignore it and everything seems fine, the plane flies, the RC works so I really don't understand what this message is doing for me.

rmackay9:
I've added this to the [4.6 issues list](https://github.com/ArduPilot/ardupilot/issues/28612)

PR #30147, opened 2025-05-23
GCS_MAVLink: protect control sensor flag bitmasks with a semaphore

peterbarker:
nothing says get_sensor_status_flags can't be called from multiple threads.

If it is called from multiple threads then there's a reasonably obvious race where the mask can be cleared while some other process is already updating the flags.  That can cause a caller to report something as unhealthy when it isn't.


peterbarker:
... perhaps processing pulses via `RCININT` could have us getting these flags in AP_RCTelemetry.  Certainly the board this was seen on can get its RC in this way.

It is unclear why *this* bit would fail multiple times but other bits in the same section wouldn't.

May close https://github.com/ArduPilot/ardupilot/issues/30142

tridge (approved):

rmackay9:

PR #30146, opened 2025-05-23
AP_TECS: Convert option to enum class

Ryanf55:
* And access with option_is_set helper

PR #30150, opened 2025-05-24
ArduPlane: change OPTION case to match convention

Ryanf55:
Same as requested here: https://github.com/ArduPilot/ardupilot/pull/30146/files


F11ff:
hasona1991

في الاثنين، ٧ يوليو ٢٠٢٥ ٦:٤٥ ص Peter Barker ***@***.***> كتب:

PR #30153, opened 2025-05-24
Remove INAV from Copter

lthall:
This is the final PR to remove INAV from Copter completly.

CubeOrange,0,*,32,56,*,*,0

### DIFF
--- a/ArduCopter/AP_Arming_Copter.cpp
+++ b/ArduCopter/AP_Arming_Copter.cpp
@@ -138,8 +138,10 @@ bool AP_Arming_Copter::barometer_checks(bool display_failure)
         // that may differ from the baro height due to baro drift.
         const auto &ahrs = AP::ahrs();
         const bool using_baro_ref = !ahrs.has_status(AP_AHRS::Status::PRED_HORIZ_POS_REL) && ahrs.has_status(AP_AHRS::Status::PRED_HORIZ_POS_ABS);
+        float pos_d_m = 0;
+        UNUSED_RESULT(AP::ahrs().get_relative_position_D_origin_float(pos_d_m));
         if (using_baro_ref) {
-            if (fabsf(copter.inertial_nav.get_position_z_up_cm() - copter.baro_alt) > PREARM_MAX_ALT_DISPARITY_CM) {
+            if (fabsf(-pos_d_m * 100.0 - copter.baro_alt) > PREARM_MAX_ALT_DISPARITY_CM) {
                 check_failed(Check::BARO, display_failure, "Altitude disparity");
                 ret = false;
             }
@@ -732,8 +734,10 @@ bool AP_Arming_Copter::arm(const AP_Arming::Method method, const bool do_arming_
             // ignore failure
         }
 
-        // remember the height when we armed
-        copter.arming_altitude_m = copter.inertial_nav.get_position_z_up_cm() * 0.01;
+        // remember the height when we armed (ignore failures)
+        float pos_d_m = 0;
+        UNUSED_RESULT(AP::ahrs().get_relative_position_D_origin_float(pos_d_m));
+        copter.arming_altitude_m = -pos_d_m;
     }
     copter.update_super_simple_bearing(false);
 
--- a/ArduCopter/GCS_MAVLink_Copter.cpp
+++ b/ArduCopter/GCS_MAVLink_Copter.cpp
@@ -948,291 +948,296 @@ bool GCS_MAVLINK_Copter::sane_vel_or_acc_vector(const Vector3f &vec) const
 #if MODE_GUIDED_ENABLED
 void GCS_MAVLINK_Copter::handle_message_set_attitude_target(const mavlink_message_t &msg)
 {
-        // decode packet
-        mavlink_set_attitude_target_t packet;
-        mavlink_msg_set_attitude_target_decode(&msg, &packet);
+    // decode packet
+    mavlink_set_attitude_target_t packet;
+    mavlink_msg_set_attitude_target_decode(&msg, &packet);
 
-        // exit if vehicle is not in Guided mode or Auto-Guided mode
-        if (!copter.flightmode->in_guided_mode()) {
-            return;
-        }
+    // exit if vehicle is not in Guided mode or Auto-Guided mode
+    if (!copter.flightmode->in_guided_mode()) {
+        return;
+    }
 
-        const bool roll_rate_ignore   = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_BODY_ROLL_RATE_IGNORE;
-        const bool pitch_rate_ignore  = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_BODY_PITCH_RATE_IGNORE;
-        const bool yaw_rate_ignore    = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_BODY_YAW_RATE_IGNORE;
-        const bool throttle_ignore    = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_THROTTLE_IGNORE;
-        const bool attitude_ignore    = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_ATTITUDE_IGNORE;
+    const bool roll_rate_ignore   = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_BODY_ROLL_RATE_IGNORE;
+    const bool pitch_rate_ignore  = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_BODY_PITCH_RATE_IGNORE;
+    const bool yaw_rate_ignore    = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_BODY_YAW_RATE_IGNORE;
+    const bool throttle_ignore    = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_THROTTLE_IGNORE;
+    const bool attitude_ignore    = packet.type_mask & ATTITUDE_TARGET_TYPEMASK_ATTITUDE_IGNORE;
 
-        // ensure thrust field is not ignored
-        if (throttle_ignore) {
-            // The throttle input is not defined
-            return;
-        }
+    // ensure thrust field is not ignored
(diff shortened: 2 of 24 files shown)


lthall:
> You should also be able to remove it from here:
> 
> https://github.com/ArduPilot/ardupilot/blob/fb4e5cefeb1ee5e4d9d323cc1db0e5b5048aa138/ArduCopter/wscript#L18

Thanks, I have done this.

ArduCopter/GCS_MAVLink_Copter.cpp
```diff
@@ -1069,7 +1069,13 @@ void GCS_MAVLINK_Copter::handle_message_set_position_target_local_ned(const mavl
             if (packet.coordinate_frame == MAV_FRAME_LOCAL_OFFSET_NED ||
                 packet.coordinate_frame == MAV_FRAME_BODY_NED ||
                 packet.coordinate_frame == MAV_FRAME_BODY_OFFSET_NED) {
-                pos_vector += copter.inertial_nav.get_position_neu_cm();
+                Vector3f pos_ned_m;
+                if (!AP::ahrs().get_relative_position_NED_origin_float(pos_ned_m)) {
+                    // need position estimate to calculate target position
+                    return;
```
IamPete1, line 1075: some places in this function call `copter.mode_guided.init(true);` when they return. Although 1055 does not....

ArduCopter/autoyaw.cpp
```diff
@@ -5,17 +5,23 @@ Mode::AutoYaw Mode::auto_yaw;
 // roi_yaw - returns heading towards location held in roi
 float Mode::AutoYaw::roi_yaw() const
 {
-    return get_bearing_cd(copter.inertial_nav.get_position_xy_cm(), roi.xy());
+    Vector2f pos_ne_m;
+    if (AP::ahrs().get_relative_position_NE_origin_float(pos_ne_m)){
```
IamPete1, line 9:
```suggestion
    if (AP::ahrs().get_relative_position_NE_origin_float(pos_ne_m)) {
```

ArduCopter/autoyaw.cpp
```diff
@@ -5,17 +5,23 @@ Mode::AutoYaw Mode::auto_yaw;
...
+        return get_bearing_cd(pos_ne_m * 100.0, roi.xy());
+    }
+    return copter.attitude_control->get_att_target_euler_cd().z;
```
IamPete1, line 12: I wonder if we should change out of auto yaw ROI mode in this case.

ArduCopter/events.cpp
```diff
@@ -322,7 +323,9 @@ void Copter::failsafe_deadreckon_check()
     const char* dr_prefix_str = "Dead Reckoning";
 
     // get EKF filter status
-    bool ekf_dead_reckoning = inertial_nav.get_filter_status().flags.dead_reckoning;
+    nav_filter_status filt_status {};
+    AP::ahrs().get_filter_status(filt_status);
+    bool ekf_dead_reckoning = filt_status.flags.dead_reckoning;
```
peterbarker, line 328:
```suggestion
    const bool ekf_dead_reckoning = AP::ahrs().has_status(SomeNamespace::DEAD_RECKONING);
```

IamPete1: What about being able to do `AP::ahrs().has_status(SomeNamespace::HORIZ_VEL | SomeNamespace::VERT_VEL)` to check that both horizontal and vertical available.

ArduCopter/AP_Arming_Copter.cpp
```diff
@@ -138,8 +138,10 @@ bool AP_Arming_Copter::barometer_checks(bool display_failure)
         // that may differ from the baro height due to baro drift.
         const auto &ahrs = AP::ahrs();
         const bool using_baro_ref = !ahrs.has_status(AP_AHRS::Status::PRED_HORIZ_POS_REL) && ahrs.has_status(AP_AHRS::Status::PRED_HORIZ_POS_ABS);
+        float pos_d_m = 0;
+        UNUSED_RESULT(AP::ahrs().get_relative_position_D_origin_float(pos_d_m));
         if (using_baro_ref) {
-            if (fabsf(copter.inertial_nav.get_position_z_up_cm() - copter.baro_alt) > PREARM_MAX_ALT_DISPARITY_CM) {
+            if (fabsf(pos_d_m) > PREARM_MAX_ALT_DISPARITY_CM) {
```
peterbarker, line 144: Unit failure - m vs cm

peterbarker: ... also, this is not the same check - you're asking the  AHRS for an altitude, not the baro library now, where the comment is definitely wanting you to be looking at the baro...

peterbarker (changes requested):

ArduCopter/toy_mode.cpp
```diff
@@ -846,15 +846,18 @@ void ToyMode::throttle_adjust(float &throttle_control)
...
+        float height = -pos_d_m - copter.arming_altitude_m;
+        if (throttle_control < 500 &&
+            height < TOY_DESCENT_SLOW_HEIGHT + TOY_DESCENT_SLOW_RAMP &&
+            copter.motors->armed() && !copter.ap.land_complete) {
+            float limit = linear_interpolate(TOY_DESCENT_SLOW_MIN, 0, height,
+                                            TOY_DESCENT_SLOW_HEIGHT, TOY_DESCENT_SLOW_HEIGHT+TOY_DESCENT_SLOW_RAMP);
+            if (throttle_control < limit) {
+                // limit descent rate close to the ground
+                throttle_control = limit;
+            }
```
peterbarker, line 860:
```suggestion
    float pos_d_m;
    if (AP::ahrs().get_relative_position_D_origin_float(pos_d_m)) {
        // we do not know how high we are, let alone where the ground is
        return;
    }
    if (throttle_control >= 500) {
        // user wants to climb - don't play with throttle
        return;
    }
    const float height = -pos_d_m - copter.arming_altitude_m;
    if (height >= TOY_DESCENT_SLOW_HEIGHT + TOY_DESCENT_SLOW_RAMP) {
        // vehicle is above ramp-adjusted slow-down height
        return;
    }
    if (!copter.motors->armed()) {
        // vehicle isn't armed, so don't play with throttle
        return;
    }
    if (copter.ap.land_complete) {
        // we're landed, don't play with throttle
        return;
    }
    const float limit = linear_interpolate(TOY_DESCENT_SLOW_MIN, 0, height,
                                            TOY_DESCENT_SLOW_HEIGHT,
                                            TOY_DESCENT_SLOW_HEIGHT+TOY_DESCENT_SLOW_RAMP);
    if (throttle_control >= limit) {
        // vehicle is above the now-calculated throttle-decrease height
        return;
    }
    // limit descent rate close to the ground
    throttle_control = limit;
```

... and the ordering of these should be fixed...

ArduCopter/autoyaw.cpp
```diff
@@ -5,17 +5,23 @@ Mode::AutoYaw Mode::auto_yaw;
...
 {
-    const Vector3f& vel = copter.inertial_nav.get_velocity_neu_cms();
-    const float speed_sq = vel.xy().length_squared();
     // Commanded Yaw to automatically look ahead.
-    if (copter.position_ok() && (speed_sq > (YAW_LOOK_AHEAD_MIN_SPEED * YAW_LOOK_AHEAD_MIN_SPEED))) {
-        _look_ahead_yaw = degrees(atan2f(vel.y,vel.x));
+    Vector3f vel_ned_ms;
+    if (copter.position_ok() && AP::ahrs().get_velocity_NED(vel_ned_ms)) {
+        const float speed_sq = vel_ned_ms.xy().length_squared() * 100.0;
+        if (speed_sq > (YAW_LOOK_AHEAD_MIN_SPEED * YAW_LOOK_AHEAD_MIN_SPEED)) {
```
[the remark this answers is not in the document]
lthall: That should be 10000.0 or 100^2

define YAW_LOOK_AHEAD_MIN_SPEED  100             // minimum ground speed in cm/s required before copter is aimed at ground course

ArduCopter/baro_ground_effect.cpp
```diff
@@ -39,29 +32,34 @@ void Copter::update_ground_effect_detector(void)
...
+    if (gndeffect_state.takeoff_expected && (tnow_ms-gndeffect_state.takeoff_time_ms > 5000 || (-pos_d_m * 100.0 - gndeffect_state.takeoff_alt_cm) > 50.0f)) {
         gndeffect_state.takeoff_expected = false;
     }
 
     // landing logic
     Vector3f angle_target_rad = attitude_control->get_att_target_euler_rad();
     bool small_angle_request = cosf(angle_target_rad.x)*cosf(angle_target_rad.y) > cosf(radians(7.5f));
-    bool xy_speed_low = (position_ok() || ekf_has_relative_position()) && xy_speed_cms <= 125.0f;
+    Vector3f vel_ned_ms;
+    bool xy_speed_low = AP::ahrs().get_velocity_NED(vel_ned_ms) && (vel_ned_ms.xy().length() < 0.125);
```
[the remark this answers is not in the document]
lthall: Yep Looks like it.

ArduCopter/GCS_MAVLink_Copter.cpp
lthall: Thanks Pete!

ArduCopter/crash_check.cpp
```diff
@@ -143,7 +143,9 @@ void Copter::thrust_loss_check()
     }
 
     // check for descent
-    if (!is_negative(inertial_nav.get_velocity_z_up_cms())) {
+    float vel_d_ms = 0;
+    if (!AP::ahrs().get_velocity_D(vel_d_ms, vibration_check.high_vibes) || !is_positive(vel_d_ms)) {
```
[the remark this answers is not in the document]
lthall: !is_positive isn't the same as is_negative. The zero is important.

rmackay9:
I found, fixed an issue with my changes to AP_SurfaceDistance.  I've also tested vs master to confirm that the new and old code produces the same output.

I think this PR should be OK to merge now

PR #30155, opened 2025-05-24
Ensure atomic access to turtle mode state from rate thread

andyp1per:
This is similar in nature to the bug fix required for motor test in the fast rate loop (FYI @rmackay9). Races between the threads can cause uncontrolled motor output when switching modes.

There is still a bug with arming on a switch that I need to debug

### DIFF
--- a/ArduCopter/mode.h
+++ b/ArduCopter/mode.h
@@ -3,6 +3,7 @@
 #include "Copter.h"
 #include <AP_Math/chirp.h>
 #include <AP_ExternalControl/AP_ExternalControl_config.h> // TODO why is this needed if Copter.h includes this
+#include <AP_HAL/Semaphores.h>
 
 #if AP_COPTER_ADVANCED_FAILSAFE_ENABLED
 #include "afs_copter.h"
@@ -1857,6 +1858,10 @@ private:
     float motors_output;
     Vector2f motors_input;
     uint32_t last_throttle_warning_output_ms;
+
+    // Semaphore to protect the motors from the arming state
+    HAL_Semaphore msem;
+    bool shutdown;
 };
 #endif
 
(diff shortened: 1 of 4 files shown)


ArduCopter/mode.h
```diff
@@ -1853,6 +1854,9 @@ class ModeTurtle : public Mode {
     float motors_output;
     Vector2f motors_input;
     uint32_t last_throttle_warning_output_ms;
+
+    HAL_Semaphore msem;
```
peterbarker, line 1863: Could we get a comment on what this semaphore protects, please?  I don't mind the shorter name, but an indication that this mutexes the motor outputs or whatever would be good

tridge:
needs review by @rmackay9

ArduCopter/mode_turtle.cpp
```diff
@@ -45,11 +48,13 @@ void ModeTurtle::arm_motors()
...
     change_motor_direction(true);
 
-    // disable throttle and gps failsafe
+    // disable throttle and gcs failsafe
     g.failsafe_throttle.set(FS_THR_DISABLED);
     g.failsafe_gcs.set(FS_GCS_DISABLED);
     g.fs_ekf_action.set(0);
 
+    // ensure the arming library is aware of our meddling
+    AP::arming().AP_Arming::arm(AP_Arming::Method::TURTLE_MODE, false);
```
peterbarker, line 58: I have concerns around what happens if this fails... or the associated disarm...

You might still be in the state you're trying to avoid if this fails?

rmackay9 (approved):
once the comment improvement is added, let's merge, thanks!

ArduCopter/mode.h
andyp1per: Fixed

ArduCopter/mode_turtle.cpp
andyp1per: Even if it fails we must ignore and go on - the copter has crashed, this is the only way out, if we prevent arming because the sdcard has failed then all is lost

Issue #29416, opened 2025-03-03
Rover: add support for CONDITION_YAW mission command


rmackay9:
Skid steering rovers and boats have the ability to control their heading so we could add support for the CONDITION_YAW command while in Auto and Guided mode.  This would allow users to more easily point their vehicle in a specified direction while they're not moving.

jrcacador:
Hi! Following the suggestion on [discord](https://discord.com/channels/674039678562861068/674039678982422579/1364797839309537320), I'm currently working on this issue

rmackay9:
One tricky thing about handling this issue while the vehicle is in Auto mode is that condition-yaw is a "do" command instead of a navigation command.  This means that the command will likely need to be placed after a waypoint command that includes a long delay.  For guided mode this isn't an issue.

PR #30157, opened 2025-05-24
Rover: CONDITION_YAW command to skid-steering Rover

jrcacador:
Related to #29416
Added initial handling logic for the CONDITION_YAW command. An associated autotest was created to validate the new functionality. However, the test currently fails, indicating that further debugging and refinement of the yaw control logic are needed.

### DIFF
--- a/Rover/GCS_MAVLink_Rover.cpp
+++ b/Rover/GCS_MAVLink_Rover.cpp
@@ -506,6 +506,9 @@ MAV_RESULT GCS_MAVLINK_Rover::handle_command_int_packet(const mavlink_command_in
 {
     switch (packet.command) {
 
+    case MAV_CMD_CONDITION_YAW:
+        return handle_command_condition_yaw(packet);
+
     case MAV_CMD_DO_CHANGE_SPEED:
         // param1 : type
         // param2 : new speed in m/s
@@ -593,6 +596,43 @@ MAV_RESULT GCS_MAVLINK_Rover::handle_command_nav_set_yaw_speed(const mavlink_com
         return MAV_RESULT_ACCEPTED;
 }
 #endif
+
+MAV_RESULT GCS_MAVLINK_Rover::handle_command_condition_yaw(const mavlink_command_int_t &packet)
+{
+    // param1 : target angle [0-360]
+    // param2 : speed during change [deg per second]
+    // param3 : direction (-1:ccw, +1:cw, 0:quickest direction to target heading)
+    // param4 : relative offset (1) or absolute angle (0)
+
+    if (!rover.g2.motors.have_skid_steering()) {
+        // not yet implemented for other vehicle types
+        return MAV_RESULT_UNSUPPORTED;
+    }
+
+    // exit if vehicle is not in Guided mode
+    if (!rover.control_mode->in_guided_mode()) {
+        return MAV_RESULT_FAILED;
+    }
+
+    // Check valid direction parameter
+    const int8_t direction = static_cast<int8_t>(packet.param3);
+    if (abs(direction) > 1) {  // only 0, -1 and 1 are valid direction values
+        return MAV_RESULT_DENIED;
+    }
+
+    // get final angle, 1 = Relative, 0 = Absolute
+    switch ((int) packet.param4) {
+    case 0: // absolute angle
+        rover.mode_guided.set_desired_turn_rate_and_heading(packet.param1 * 100.0f, packet.param2 * 100.0f, direction);
+        break;
+    case 1: // relative angle
+        rover.mode_guided.set_desired_turn_rate_and_heading_delta(packet.param1 * 100.0f, packet.param2 * 100.0f, direction);
+        break;
+    default:
+        return MAV_RESULT_DENIED;
+    }
+    return MAV_RESULT_ACCEPTED;
+}
 
 MAV_RESULT GCS_MAVLINK_Rover::handle_command_int_do_reposition(const mavlink_command_int_t &packet)
 {
--- a/Rover/mode.h
+++ b/Rover/mode.h
@@ -175,6 +175,10 @@ protected:
     // rate_max is a maximum turn rate in deg/s.  set to zero to use default turn rate limits
     void calc_steering_to_heading(float desired_heading_cd, float rate_max_degs = 0.0f);
 
+    // calculate steering output to drive towards desired heading
+    // rate_max is a maximum turn rate in deg/s.  set to zero to use default turn rate limits
+    void calc_directional_steering_to_heading(float desired_heading_cd, float rate_max_degs = 0.0f);
+
     // calculates the amount of throttle that should be output based
     // on things like proximity to corners and current speed
     virtual void calc_throttle(float target_speed, bool avoidance_enabled);
@@ -551,6 +555,10 @@ public:
     void set_desired_heading_delta_and_speed(float yaw_delta_cd, float target_speed);
     void set_desired_turn_rate_and_speed(float turn_rate_cds, float target_speed);
 
+    // set desired turn rate and heading
+    void set_desired_turn_rate_and_heading(float yaw_angle_cd, float turn_rate_cds, int8_t direction);
+    void set_desired_turn_rate_and_heading_delta(float yaw_delta_cd, float turn_rate_cds, int8_t direction);
+
     // set steering and throttle (-1 to +1).  Only called from scripts
     void set_steering_and_throttle(float steering, float throttle);
 
@@ -574,6 +582,7 @@ protected:
         TurnRateAndSpeed,
         Loiter,
         SteeringAndThrottle,
+        TurnRateAndHeading,
         Stop
     };
 
--- a/Rover/mode_guided.cpp
+++ b/Rover/mode_guided.cpp
@@ -101,6 +101,30 @@ void ModeGuided::update()
             break;
         }
 
+        case SubMode::TurnRateAndHeading:
+        {
+            // stop vehicle if target not updated within 3 seconds
+            if (have_attitude_target && (millis() - _des_att_time_ms) > 3000) {
+                GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "target not received last 3secs, stopping");
+                have_attitude_target = false;
+                break;
+            }
+            if (have_attitude_target) {
+                // run steering and throttle controllers
+                calc_directional_steering_to_heading(_desired_yaw_cd, _desired_yaw_rate_cds * 0.01f);
+                break;
+            }
+            // we have reached the destination so stay here
+            if (rover.is_boat()) {
+                if (!start_loiter()) {
+                    stop_vehicle();
+                }
+                break;
+            }
+            stop_vehicle();
+            break;
+        }
+
         case SubMode::Loiter:
         {
             rover.mode_loiter.update();
@@ -149,6 +173,7 @@ float ModeGuided::wp_bearing() const
         return g2.wp_nav.wp_bearing_cd() * 0.01f;
     case SubMode::HeadingAndSpeed:
     case SubMode::TurnRateAndSpeed:
+    case SubMode::TurnRateAndHeading:
         return 0.0f;
     case SubMode::Loiter:
         return rover.mode_loiter.wp_bearing();
@@ -169,6 +194,8 @@ float ModeGuided::nav_bearing() const
     case SubMode::HeadingAndSpeed:
     case SubMode::TurnRateAndSpeed:
         return 0.0f;
+    case SubMode::TurnRateAndHeading:
+        return _desired_yaw_cd * 0.01f;
     case SubMode::Loiter:
         return rover.mode_loiter.nav_bearing();
     case SubMode::SteeringAndThrottle:
@@ -187,6 +214,7 @@ float ModeGuided::crosstrack_error() const
         return g2.wp_nav.crosstrack_error();
     case SubMode::HeadingAndSpeed:
     case SubMode::TurnRateAndSpeed:
+    case SubMode::TurnRateAndHeading:
         return 0.0f;
     case SubMode::Loiter:
         return rover.mode_loiter.crosstrack_error();
@@ -206,6 +234,7 @@ float ModeGuided::get_desired_lat_accel() const
         return g2.wp_nav.get_lat_accel();
     case SubMode::HeadingAndSpeed:
     case SubMode::TurnRateAndSpeed:
+    case SubMode::TurnRateAndHeading:
         return 0.0f;
     case SubMode::Loiter:
         return rover.mode_loiter.get_desired_lat_accel();
@@ -226,6 +255,7 @@ float ModeGuided::get_distance_to_destination() const
         return _distance_to_destination;
     case SubMode::HeadingAndSpeed:
     case SubMode::TurnRateAndSpeed:
+    case SubMode::TurnRateAndHeading:
         return 0.0f;
     case SubMode::Loiter:
         return rover.mode_loiter.get_distance_to_destination();
@@ -246,6 +276,7 @@ bool ModeGuided::reached_destination() const
         return g2.wp_nav.reached_destination();
     case SubMode::HeadingAndSpeed:
     case SubMode::TurnRateAndSpeed:
+    case SubMode::TurnRateAndHeading:
     case SubMode::Loiter:
     case SubMode::SteeringAndThrottle:
     case SubMode::Stop:
@@ -269,6 +300,7 @@ bool ModeGuided::set_desired_speed(float speed)
     case SubMode::Loiter:
         return rover.mode_loiter.set_desired_speed(speed);
     case SubMode::SteeringAndThrottle:
+    case SubMode::TurnRateAndHeading:
     case SubMode::Stop:
         // no speed control
         return false;
@@ -288,6 +320,7 @@ bool ModeGuided::get_desired_location(Location& destination) const
         return false;
     case SubMode::HeadingAndSpeed:
     case SubMode::TurnRateAndSpeed:
+    case SubMode::TurnRateAndHeading:
         // not supported in these submodes
         return false;
     case SubMode::Loiter:
@@ -372,6 +405,36 @@ void ModeGuided::set_desired_turn_rate_and_speed(float turn_rate_cds, float targ
     // log new target
     rover.Log_Write_GuidedTarget((uint8_t)_guided_mode, Vector3f(_desired_yaw_rate_cds, 0.0f, 0.0f), Vector3f(_desired_speed, 0.0f, 0.0f));
 #endif
+}
+
+// set desired attitude
+void ModeGuided::set_desired_turn_rate_and_heading(float yaw_angle_cd, float turn_rate_cds, int8_t direction)
+{
+    // initialisation and logging
+    _guided_mode = SubMode::TurnRateAndHeading;
+    _des_att_time_ms = AP_HAL::millis();
+
+    // record targets
+    have_attitude_target = true;
+    _desired_yaw_rate_cds = turn_rate_cds; 
+    _desired_yaw_cd = wrap_180_cd(yaw_angle_cd);
+
+    // force clockwise or counter-clockwise movement according to direction
+    if (direction == -1 && is_positive(_desired_yaw_cd)) {
+        _desired_yaw_cd -= 36000.0;
+    } else if (direction == 1 && is_negative(_desired_yaw_cd)) {
+        _desired_yaw_cd += 36000.0;
+    }
+}
+
+// set desired attitude
+void ModeGuided::set_desired_turn_rate_and_heading_delta(float yaw_delta_cd, float turn_rate_cds, int8_t direction)
+{
+    /// handle initialisation
+    _guided_mode = SubMode::TurnRateAndHeading;
+    float current_yaw_cd = ahrs.yaw_sensor;
+
+    set_desired_turn_rate_and_heading(current_yaw_cd + yaw_delta_cd, turn_rate_cds, direction);
 }
 
 // set steering and throttle (both in the range -1 to +1)
--- a/Tools/autotest/rover.py
+++ b/Tools/autotest/rover.py
@@ -6361,6 +6361,60 @@ Brakes have negligible effect (with=%0.2fm without=%0.2fm delta=%0.2fm)
         self.wait_distance_to_home(0, 5, timeout=30)
         self.disarm_vehicle()
 
+    def MAV_CMD_CONDITION_YAW_ABSOLUTE(self):
+        '''simple test for CONDITION_YAW command with absolute angles'''
+        model = "rover-skid"
+        self.customise_SITL_commandline([],
+                                        model=model,
+                                        defaults_filepath=self.model_defaults_filepath(model))
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
+        self.change_mode('GUIDED')
+
+        for (heading, intermediate_heading, direction) in (170, 90, 1), (300, 90, -1), (200, 250, 0), (300, 250, 0):
+
+            def send_cmd(x, y):
+                self.run_cmd(
+                    mavutil.mavlink.MAV_CMD_CONDITION_YAW,
+                    p1=heading,     # target angle
+                    p2=10,          # degrees/second
+                    p3=direction,   # -1 is counter-clockwise, 1 clockwise
+                    p4=0,           # 1 for relative, 0 for absolute
+                )
+            """check if rotating in correct direction"""
+            self.wait_heading(intermediate_heading, 5, called_function=send_cmd)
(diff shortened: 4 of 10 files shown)


peterbarker (changes requested):
```
AT-0002.7: AP: target not received last 3secs, stopping
```

Will need to cope with that problem in the autotest, you can add a parameter to `wait_heading` to send the attitude request through until the condition is met (or the the call fails....)


The semantics of this in guided mode will need to be determined by @rmackay9 .

*do* we come to zero speed after completing this command or can you compose this with other guided-mode commands that set your vehicle speed?

If we don't compose, *should* we loiter after completing this given we haven't actually given a position command?

Just generally, this is really good work - the structure is correct.

Rover/mode.h
```diff
@@ -574,7 +578,8 @@ class ModeGuided : public Mode
         TurnRateAndSpeed,
         Loiter,
         SteeringAndThrottle,
-        Stop
+        Stop,
+        TurnRateAndHeading
```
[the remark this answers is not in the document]
jrcacador: Done

Rover/mode_guided.cpp
```diff
@@ -101,6 +101,29 @@ void ModeGuided::update()
...
+        {
+            // stop vehicle if target not updated within 3 seconds
+            if (have_attitude_target && (millis() - _des_att_time_ms) > 3000) {
+                gcs().send_text(MAV_SEVERITY_WARNING, "target not received last 3secs, stopping");
+                have_attitude_target = false;
+            }
+            if (have_attitude_target) {
+                // run steering and throttle controllers
+                calc_steering_to_heading(_desired_yaw_cd, _desired_yaw_rate_cds * 0.01f);
+            } else {
```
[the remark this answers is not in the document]
jrcacador: Done

Rover/mode_guided.cpp
```diff
@@ -101,6 +101,29 @@ void ModeGuided::update()
...
+                // we have reached the destination so stay here
```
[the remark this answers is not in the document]
jrcacador: Done

Rover/mode_guided.cpp
```diff
@@ -101,6 +101,29 @@ void ModeGuided::update()
...
+                if (rover.is_boat()) {
+                    if (!start_loiter()) {
+                        stop_vehicle();
+                    }
+                } else {
```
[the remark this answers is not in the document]
jrcacador: Done

jrcacador:
> Will need to cope with that problem in the autotest, you can add a parameter to wait_heading to send the attitude request through until the condition is met (or the the call fails....)

I don't really understand what I have to do for this, as I was not able to find anything similiar. Am I supposed to create a new parameter for this function? I couldn't really find a way to achieve the intended behaviour with the existing ones.

peterbarker:
> > Will need to cope with that problem in the autotest, you can add a parameter to wait_heading to send the attitude request through until the condition is met (or the the call fails....)
> 
> I don't really understand what I have to do for this, as I was not able to find anything similiar. Am I supposed to create a new parameter for this function? I couldn't really find a way to achieve the intended behaviour with the existing ones.

OK, you need to continuously send through or guided thinks you've given up and gone home for the day - that takes 3 seconds.  See `SET_ATTITUDE_TARGET_heading_test_target` - you want to pass in `called_function`

Tools/autotest/rover.py
```diff
@@ -6315,6 +6315,21 @@ def MAV_CMD_DO_CHANGE_SPEED(self):
...
+        self.arm_vehicle()
+        self.change_mode('GUIDED')
+        self.run_cmd(
+            mavutil.mavlink.MAV_CMD_CONDITION_YAW,
+            p1=60,  # target angle
+            p2=10,  # degrees/second
+            p3=1,   # -1 is counter-clockwise, 1 clockwise
+            p4=0,   # 1 for relative, 0 for absolute
+        )
+        self.wait_heading(60, called_function=self.SET_ATTITUDE_TARGET_heading_test_target(60, None, 1))
```
peterbarker, line 6330:
```suggestion
        def send_cmd(x, y):
            self.run_cmd(
                mavutil.mavlink.MAV_CMD_CONDITION_YAW,
                p1=60,  # target angle
                p2=10,  # degrees/second
                p3=1,   # -1 is counter-clockwise, 1 clockwise
                p4=0,   # 1 for relative, 0 for absolute
            )
        self.wait_heading(60, called_function=send_cmd)
```

Rover/mode_guided.cpp
```diff
@@ -374,6 +406,31 @@ void ModeGuided::set_desired_turn_rate_and_speed(float turn_rate_cds, float targ
...
+    _desired_yaw_rate_cds = turn_rate_cds;
+    have_attitude_target = true;
+}
+
+// set desired attitude
+void ModeGuided::set_desired_turn_rate_and_heading_delta(float yaw_delta_cd, float turn_rate_cds, int8_t direction)
+{
+    /// handle initialisation
+    if (_guided_mode != SubMode::HeadingAndSpeed) {
+        _guided_mode = SubMode::HeadingAndSpeed;
```
peterbarker, line 427: Surely this should be using your new SubMode?

Rover/mode_guided.cpp
```diff
@@ -374,6 +406,31 @@ void ModeGuided::set_desired_turn_rate_and_speed(float turn_rate_cds, float targ
...
+        _desired_yaw_cd = ahrs.yaw_sensor;
```
peterbarker, line 428: ... this probably shouldn't be setting _desired_yaw_cd here; use a local to hold the variable you will pass through to `set _desired_turn_rate_and_heading`.

... also, not clear this should be within the `else`

peterbarker (changes requested):

Rover/mode_guided.cpp
```diff
@@ -264,6 +294,7 @@ bool ModeGuided::set_desired_speed(float speed)
         return g2.wp_nav.set_speed_max(speed);
     case SubMode::HeadingAndSpeed:
     case SubMode::TurnRateAndSpeed:
+    case SubMode::TurnRateAndHeading:
```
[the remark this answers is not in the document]
jrcacador: Should I put it next to the ones without speed control for now then? If not, what speed do I set in this mode?

Tools/autotest/rover.py
jrcacador: Done

Rover/mode_guided.cpp
jrcacador: Done

Rover/mode_guided.cpp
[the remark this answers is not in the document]
jrcacador: After the changes, the test fails again. But only because I'm not changing the speed. I've tried changing the speed to test the command and it passes the test both for relative and absolute angles. 

That said, all that's left is to know how should this command affect the speed of the vehicle.

peterbarker (changes requested):

Rover/GCS_MAVLink_Rover.cpp
```diff
@@ -578,6 +581,29 @@ MAV_RESULT GCS_MAVLINK_Rover::handle_command_nav_set_yaw_speed(const mavlink_com
...
+        }
+
+        // get final angle, 1 = Relative, 0 = Absolute
+        if (packet.param4 > 0) {
+            // relative angle
+            rover.mode_guided.set_desired_turn_rate_and_heading_delta(packet.param1 * 100.0f, packet.param2 * 100.0f, packet.param3);
+        } else {
+            // absolute angle
+            rover.mode_guided.set_desired_turn_rate_and_heading(packet.param1 * 100.0f, packet.param2 * 100.0f, packet.param3);
+        }
```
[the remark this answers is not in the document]
jrcacador: I can't do a switch on a float value. So I'm casting it to int to be able to do it...

Rover/mode_guided.cpp
```diff
@@ -168,6 +193,7 @@ float ModeGuided::nav_bearing() const
         return g2.wp_nav.nav_bearing_cd() * 0.01f;
     case SubMode::HeadingAndSpeed:
     case SubMode::TurnRateAndSpeed:
+    case SubMode::TurnRateAndHeading:
```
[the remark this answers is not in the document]
jrcacador: Done

peterbarker (changes requested):
Looks good - but please squash all the commits together into a single commit.  Let me know if you're going to have a problem doing that.

jrcacador:
Okay, nice I can do that. The support to specify cw and ccw movement isn't implemented yet though. 
I need to do that part and improve the autotest to try all types of parameters for the command.

Should I still add this to this PR and then, when everything's fine, squash everything, or is it better to close this as it is and open a new PR for that part?

Rover/mode_guided.cpp
```diff
@@ -374,6 +407,32 @@ void ModeGuided::set_desired_turn_rate_and_speed(float turn_rate_cds, float targ
...
+void ModeGuided::set_desired_turn_rate_and_heading_delta(float yaw_delta_cd, float turn_rate_cds, int8_t direction)
+{
+    /// handle initialisation
+    float desired_yaw_cd = _desired_yaw_cd;
+    if (_guided_mode != SubMode::TurnRateAndHeading) {
+        _guided_mode = SubMode::TurnRateAndHeading;
+        desired_yaw_cd = ahrs.yaw_sensor;
+    }
+
+    set_desired_turn_rate_and_heading(wrap_180_cd(desired_yaw_cd + yaw_delta_cd), turn_rate_cds, direction);
```
IamPete1: Should this be relative to the current measured heading not the current target?

Rover/mode_guided.cpp
```diff
@@ -374,6 +407,32 @@ void ModeGuided::set_desired_turn_rate_and_speed(float turn_rate_cds, float targ
 #endif
 }
 
+// set desired attitude
+void ModeGuided::set_desired_turn_rate_and_heading(float yaw_angle_cd, float turn_rate_cds, int8_t direction)
```
IamPete1: `direction` is unused.

Rover/mode_guided.cpp
jrcacador: I believe that depends on the desired functionality for the command. I agree that basing it on the target heading rather than current sensor heading makes more sense. I just wrote it this way because it was the way implemented in similar functions like `set_desired_heading_delta_and_speed`.

Rover/mode_guided.cpp
jrcacador: Yes, I'm currently working on it.

jrcacador:
I've implemented support for specifying clockwise or counter-clockwise movement and made a simple test that expects the vehicle to turn to 60º counter-clockwise (performing a -300º rotation)

The test fails and I've tracked with gdb the problem. Further down the intruction flow, in the `AR_AttitudeControl` library, the function `get_turn_rate_from_heading` is called.
This function turns the received angle in radians to an angle in range [-PI, PI] at the beginning, which impossibilitates rotations over 180 degrees, no matter what I do. 

In order to get this to work, I would have to alter this function to instead turn the angle received into an angle in range [-2PI, 2PI]. I've tested this and it worked. However, this modification could break other code that relied on this function. 

So, essentially I'm asking what to do? Do I modify it, nonetheless? Do I create an identical function in the library to use in this scenario only and leave the existing one alone?

IamPete1:
> So, essentially I'm asking what to do? Do I modify it, nonetheless? Do I create an identical function in the library to use in this scenario only and leave the existing one alone?

We have a function to do the conversion for you `wrap_2PI`. It will take the -Pi to Pi and return 0 to 2Pi.

Rover/GCS_MAVLink_Rover.cpp
```diff
@@ -578,6 +585,38 @@ MAV_RESULT GCS_MAVLINK_Rover::handle_command_nav_set_yaw_speed(const mavlink_com
...
+        // param3 : direction (-1:ccw, +1:cw, 0:quickest direction to target heading)
+        // param4 : relative offset (1) or absolute angle (0)
+
+        // exit if vehicle is not in Guided mode
+        if (!rover.control_mode->in_guided_mode()) {
+            return MAV_RESULT_FAILED;
+        }
+
+        // Check valid direction parameter
+        int direction = (int) packet.param3;
```
rmackay9: we don't use int normally.  perhaps int16_h

libraries/AP_Math/AP_Math.cpp
```diff
@@ -257,10 +257,10 @@ ftype wrap_PI(const ftype radian)
     return res;
 }
 
-ftype wrap_2PI(const ftype radian)
+ftype wrap_2PI(const ftype radian, bool directional)
```
rmackay9: we should probably remove this directional argument because this is a very widely used function

Rover/GCS_MAVLink_Rover.cpp
```diff
@@ -578,6 +585,38 @@ MAV_RESULT GCS_MAVLINK_Rover::handle_command_nav_set_yaw_speed(const mavlink_com
...
 
+MAV_RESULT GCS_MAVLINK_Rover::handle_command_condition_yaw(const mavlink_command_int_t &packet)
+{
+        // param1 : target angle [0-360]
+        // param2 : speed during change [deg per second]
+        // param3 : direction (-1:ccw, +1:cw, 0:quickest direction to target heading)
+        // param4 : relative offset (1) or absolute angle (0)
+
+        // exit if vehicle is not in Guided mode
+        if (!rover.control_mode->in_guided_mode()) {
```
rmackay9: let's also reject the command immediately if this vehicle can't pivot (e.g. omni or skid-steering)

jrcacador:
```
case MAV_CMD_CONDITION_YAW:
    if (rover.g2.motors.have_skid_steering()) {
        return handle_command_condition_yaw(packet);
    }
    // not yet implemented for other vehicle types
    return MAV_RESULT_UNSUPPORTED;
```
Is this check upon receiving the command not enough? Or is it another thing I have to do? From what I understand this function already checks if the vehicle either has skid-steering or is omni.

libraries/AP_Math/AP_Math.cpp
jrcacador: Done

Rover/GCS_MAVLink_Rover.cpp
jrcacador: Done

jrcacador:
@rmackay9 Friendly reminder of this PR :)

libraries/APM_Control/AR_AttitudeControl.cpp
```diff
@@ -602,18 +602,18 @@ float AR_AttitudeControl::get_steering_out_lat_accel(float desired_accel, bool m
...
     // calculate the desired turn rate (in radians) from the angle error (also in radians)
-    float desired_rate = get_turn_rate_from_heading(heading_rad, rate_max_rads);
+    float desired_rate = get_turn_rate_from_heading(heading_rad, rate_max_rads, directional);
 
     return get_steering_out_rate(desired_rate, motor_limit_left, motor_limit_right, dt);
 }
 
 // return a desired turn-rate given a desired heading in radians
-float AR_AttitudeControl::get_turn_rate_from_heading(float heading_rad, float rate_max_rads) const
+float AR_AttitudeControl::get_turn_rate_from_heading(float heading_rad, float rate_max_rads, bool directional) const
```
IamPete1: I think you should pass a direction flag down to this function. The target heading being a whole rotation off will confuse logging and other things. 

maybe a new function `get_directional_turn_rate_from_heading`

jrcacador: Done

jrcacador:
@rmackay9 Is there anything else to change?

Rover/GCS_MAVLink_Rover.cpp
peterbarker: Move the check into this function - it's where Randy was expecting it :-)

Rover/GCS_MAVLink_Rover.cpp
```diff
@@ -500,6 +500,13 @@ MAV_RESULT GCS_MAVLINK_Rover::handle_command_int_packet(const mavlink_command_in
 {
     switch (packet.command) {
 
+    case MAV_CMD_CONDITION_YAW:
+        if (rover.g2.motors.have_skid_steering()) {
```
peterbarker: We probably should have a "can_pivot_turn()" here, but this is fine.  I don't think omnis should be returning "true" for have_skid_steering as there are many ways to do omnis that aren't skid steering

Not your problem to fix :-)

Rover/GCS_MAVLink_Rover.cpp
```diff
@@ -578,6 +585,38 @@ MAV_RESULT GCS_MAVLINK_Rover::handle_command_nav_set_yaw_speed(const mavlink_com
...
+        // param4 : relative offset (1) or absolute angle (0)
+
+        // exit if vehicle is not in Guided mode
+        if (!rover.control_mode->in_guided_mode()) {
+            return MAV_RESULT_FAILED;
+        }
+
+        // Check valid direction parameter
+        int16_t direction = static_cast<int16_t>(packet.param3);
+        if (direction != 0 && direction != 1 && direction != -1) {
```
peterbarker:
```suggestion
        if (abs(direction) > 1) {  // only 0, -1 and 1 are valid direction values
```

Very much optional change.

Rover/mode_guided.cpp
```diff
@@ -374,6 +407,38 @@ void ModeGuided::set_desired_turn_rate_and_speed(float turn_rate_cds, float targ
...
+    // force clockwise or counter-clockwise movement according to direction
+    if (direction == -1 && is_positive(_desired_yaw_cd)) {
+        _desired_yaw_cd -= 36000.0;
+    } else if (direction == 1 && is_negative(_desired_yaw_cd)) {
+        _desired_yaw_cd += 36000.0;
+    }
+}
+
+// set desired attitude
+void ModeGuided::set_desired_turn_rate_and_heading_delta(float yaw_delta_cd, float turn_rate_cds, int16_t direction)
```
peterbarker: Why `int16_t`?

Rover/mode_guided.cpp
```diff
@@ -374,6 +407,38 @@ void ModeGuided::set_desired_turn_rate_and_speed(float turn_rate_cds, float targ
...
+{
+    /// handle initialisation
+    float desired_yaw_cd = _desired_yaw_cd;
+    if (_guided_mode != SubMode::TurnRateAndHeading) {
+        _guided_mode = SubMode::TurnRateAndHeading;
+    }
```
peterbarker:
```suggestion
    _guided_mode = SubMode::TurnRateAndHeading;
```

Tools/autotest/rover.py
```diff
@@ -6292,6 +6292,33 @@ def MAV_CMD_DO_CHANGE_SPEED(self):
...
+
+        for (heading, intermediate_heading, direction) in (170, 90, 1), (300, 90, -1), (200, 250, 0), (300, 250, 0):
+
+            def send_cmd(x, y):
+                self.run_cmd(
+                    mavutil.mavlink.MAV_CMD_CONDITION_YAW,
+                    p1=heading,     # target angle
+                    p2=10,          # degrees/second
+                    p3=direction,   # -1 is counter-clockwise, 1 clockwise
+                    p4=0,           # 1 for relative, 0 for absolute
```
peterbarker, line 6382: Could we get some tests for relative too, please?

Rover/mode_guided.cpp
```diff
@@ -101,6 +101,30 @@ void ModeGuided::update()
             break;
         }
 
+        case SubMode::TurnRateAndHeading:
+        {
+            // stop vehicle if target not updated within 3 seconds
+            if (have_attitude_target && (millis() - _des_att_time_ms) > 3000) {
+                gcs().send_text(MAV_SEVERITY_WARNING, "target not received last 3secs, stopping");
```
peterbarker:
```suggestion
                GCS_SEND_TEXT(MAV_SEVERITY_WARNING, "target not received last 3secs, stopping");
```

Tools/autotest/rover.py
```diff
@@ -6292,6 +6292,33 @@ def MAV_CMD_DO_CHANGE_SPEED(self):
...
+                )
+            """check if rotating in correct direction"""
+            self.wait_heading(intermediate_heading-0.5, intermediate_heading+0.5, called_function=send_cmd)
+            """target angle"""
+            self.wait_heading(heading-0.5, heading+0.5, called_function=send_cmd)
```
peterbarker: Add a line here to make sure that we end up with a roughly zero rate at roughly our target heading...

peterbarker (changes requested):

Rover/mode_guided.cpp
jrcacador: I was using `int` previously but @rmackay9 said in a conversation above that `int` is not normally used and suggested the change to `int16_t`.

> we don't use int normally. perhaps int16_h

Tools/autotest/rover.py
jrcacador: The problem with relative tests is that I can't keep resending the command to prevent the rover from stopping after 3 seconds without new instructions, as I was doing with absolute angles. Resending would interfere with the target heading.
For now, I’ve done a test using small relative angles, which the rover can reach before the 3-second timeout.

Rover/mode_guided.cpp
jrcacador: Done

Tools/autotest/rover.py
jrcacador: Done

Rover/GCS_MAVLink_Rover.cpp
jrcacador: Done

Rover/GCS_MAVLink_Rover.cpp
jrcacador: Ok, so for now I have only moved the check into the function.

Rover/GCS_MAVLink_Rover.cpp
jrcacador: Way more elegant :) . Done

Rover/mode_guided.cpp
jrcacador: Done

libraries/AP_Math/AP_Math.cpp
```diff
@@ -266,6 +266,11 @@ ftype wrap_2PI(const ftype radian)
     return res;
 }
 
+ftype wrap_neg2PI_2PI(const ftype radian)
```
peterbarker: We like to have tests for all of our functions in AP_Math - it's the gtest stuff in libraries/AP_Math/tests and isn't hard to work with

This implementation should also go in the header to save some bytes

Rover/GCS_MAVLink_Rover.cpp
```diff
@@ -578,6 +581,43 @@ MAV_RESULT GCS_MAVLINK_Rover::handle_command_nav_set_yaw_speed(const mavlink_com
...
+        return MAV_RESULT_UNSUPPORTED;
+    }
+
+    // exit if vehicle is not in Guided mode
+    if (!rover.control_mode->in_guided_mode()) {
+        return MAV_RESULT_FAILED;
+    }
+
+    // Check valid direction parameter
+    int16_t direction = static_cast<int16_t>(packet.param3);
```
peterbarker:
```suggestion
    const int16_t direction = static_cast<int16_t>(packet.param3);
```

... 'though not sure why you went with int16_t

Rover/mode_guided.cpp
peterbarker: My question was more, "why not int8_t" :-)

It really doesn't matter.

peterbarker (approved):
LGTM

libraries/AP_Math/AP_Math.cpp
jrcacador: I've created the test and changed the implementation to the header file. I was not sure whether to use static inline or just inline for this function...

Rover/GCS_MAVLink_Rover.cpp
jrcacador: I agree it can be done with int8_t so I changed it to that :-)

peterbarker (changes requested):
I've just realised that the command you're using here isn't actually the right command!

See Copter's "within_distance" stuff to see what these are supposed to be for.

This is supposed to delay the state machine until you reach a yaw - not actually command the vehicle to achieve that yaw.  Copter also implements this command in the wrong way.

This should be done via `set_attitude_target`

rmackay9:
Hi @peterbarker, yes, indeed.  In Copter's defense though, there was no mission command that could command the yaw so we used this command to give us that ability.  Actually I think there still is no mission command to move just the yaw.

PR #30166, opened 2025-05-25
Add rssi to Sub

Hwurzburg:
discovered this missing on my first test swim


ArduSub/Parameters.h
```diff
@@ -366,7 +369,8 @@ class Parameters {
     AP_Int8         flight_mode6;
     AP_Int8         simple_modes;
     AP_Int8         flight_mode_chan;
-#endif    
+#endif
```
peterbarker, line 372: You're changing the whitespace here for no good reason.

If you were *fixing* the problem rather than just changing it I'd probably merge anyway :-)

peterbarker (changes requested):
What testing has this had?

ArduSub/Parameters.h
Hwurzburg: removed the line....line 372 got rid of the trailing whitespace since I had added the index k_param in the wrong spot and removed it and the incorrect whitespace at the end of the #endif got removed automatically...not sure why this is an issue...the added blank line I can understand...

Hwurzburg:
> What testing has this had?

SITL does not have RSSI without this...now it does, as does the board have the params....MP HUD displays it now

Williangalvani (approved):

PR #30169, opened 2025-05-26
Sub: correct setting of MAV_SYS_STATUS_SENSOR_ABSOLUTE_PRESSURE

peterbarker:
we only want to be present and enabled if there is actually a depth sensor on board.  Otherwise we can be present and enabled but never healthy when we don't have a depth sensor


peterbarker:
Needs @Williangalvani to approve

Williangalvani (approved):

PR #30171, opened 2025-05-26
AP_HAL_ChibiOS: Added support for X6_Pro flight controller

ZeroOne-Aero:
AP_HAL_ChibiOS: Added support for ZeroOne_Pro flight controller


Hwurzburg:
how does this differ from the X6? it seems it should be done via an include with undefs and defs or simply expanding the X6

ZeroOne-Aero:
Actually, we modified some hardware and used new sensors. I think we can submit a PR to update the hwdef file for the X6 to fit different versions. Also, how should we go about submitting our README and some product images in a standard way?

Hwurzburg:
on the PR modifying the X6 def to include the Pro, just also update its readme.....I will review as well as others....when merged I will update the X6 wikipage to be the X6/X6Pro page

ZeroOne-Aero:
Dear professors, the two sets of IIM42653 in X6Pro and the two sets of ICM45686 in X6 are pin-to-pin compatible. The hwdef files are compatible as well, so we only modified the readme file.

Hwurzburg (approved):

IamPete1 (dismissed):
I'm not qualified for hwdef stuff, but this does need a squash before merging.

peterbarker:
I've squashed this again and scaled the images as directed at DevCall.

Marking for MergeOnCIPass

@ZeroOne-Aero please don't use Merge commits - we don't allow them in ArduPilot.  You add those when you click the button in the github interface and don't choose "Update with Rebase".

ZeroOne-Aero:
> I've squashed this again and scaled the images as directed at DevCall.
> 
> Marking for MergeOnCIPass
> 
> @ZeroOne-Aero please don't use Merge commits - we don't allow them in ArduPilot. You add those when you click the button in the github interface and don't choose "Update with Rebase".

OK,thank you very much.

PR #30174, opened 2025-05-26
AC_Autotune: Fix gain load problem when mode set to FAILED

lthall:
This PR fixes a serious bug in the Multirotor Autotune where we break the primary rule "Gains are only changed from default (I excepted) during the twitch".

The code is set up such that when a test step = ABORT; is changed the gains are not changed immediately but instead on the next loop. When we also set mode = FAILED; we don't run the next loop.

I could not be confident that I could fix this problem with the structure we had in place so I first had to clean up Autotune a little to improve readability.

### DIFF
--- a/libraries/AC_AutoTune/AC_AutoTune_Heli.cpp
+++ b/libraries/AC_AutoTune/AC_AutoTune_Heli.cpp
@@ -291,7 +291,7 @@ void AC_AutoTune_Heli::test_run(AxisType test_axis, const float dir_sign)
             update_gcs(AUTOTUNE_MESSAGE_FAILED);
         } else if (tune_type == TuneType::TUNE_COMPLETE) {
             success_counter = AUTOTUNE_SUCCESS_COUNT;
-            step = StepType::UPDATE_GAINS;
+            step = Step::UPDATE_GAINS;
         }
         return;
     }
@@ -366,7 +366,7 @@ void AC_AutoTune_Heli::do_post_test_gcs_announcements() {
         break;
     }
 
-    if (step == StepType::UPDATE_GAINS) {
+    if (step == Step::UPDATE_GAINS) {
         switch (tune_type) {
         case TuneType::RATE_FF_UP:
         case TuneType::RATE_P_UP:
@@ -1023,7 +1023,7 @@ void AC_AutoTune_Heli::dwell_test_run(sweep_info &test_data)
         if (now_ms - step_start_time_ms >= sweep_time_ms + 200) {
             // we have passed the maximum stop time
             sweep_complete = true;
-            step = StepType::UPDATE_GAINS;
+            step = Step::UPDATE_GAINS;
         }
     } else {
         if (now_ms - step_start_time_ms >= step_timeout_ms || (freqresp_tgt.is_cycle_complete() && freqresp_mtr.is_cycle_complete())) {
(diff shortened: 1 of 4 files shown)


peterbarker (dismissed):

Tools/autotest/arducopter.py
```diff
@@ -3091,6 +3078,120 @@ def gains_same(gains1, gains2):
...
+            now = self.get_sim_time_cached()
+            m = self.mav.recv_match(type='STATUSTEXT',
+                                    blocking=True,
+                                    timeout=1)
+            if m is None:
+                continue
+            self.progress("STATUSTEXT (%u<%u): %s" % (now, deadline, m.text))
+            if "Determination Failed" in m.text:
+                break
+            if "AutoTune: Success" in m.text:
```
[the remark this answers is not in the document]
lthall: I couldn't get this to work. Happy to improve it further but I will close this for now.

libraries/AC_AutoTune/AC_AutoTune_Multi.h
```diff
@@ -28,172 +28,194 @@
...
-    // get_testing_step_timeout_ms accessor
+    // Timeout for a single twitch test
     uint32_t get_testing_step_timeout_ms() const override;
 
 private:
-    // twitch test functions for multicopter
-    void twitch_test_init();
-    void twitch_test_run(AxisType test_axis, const float dir_sign);
+    // Helpers for twitch-based test monitoring
+
```
rmackay9: white space is a bit odd from here on down.  a number of empty lines here and then lower down all the methods are all left justified

libraries/AC_AutoTune/AC_AutoTune_Heli.h
```diff
@@ -70,7 +70,7 @@ class AC_AutoTune_Heli : public AC_AutoTune
     // reset the update gain variables for heli
     void reset_update_gain_variables() override;
 
-    // initializes test
+    // Prepares all tuning state variables and target values for a new twitch test.
```
bnsgeyer:
```suggestion
    // Prepares all tuning state variables and target values for a new test.
```

rmackay9 (approved):

rmackay9:
Once @bnsgeyer and @lthall are happy we should merge this

libraries/AC_AutoTune/AC_AutoTune_Heli.cpp
```diff
@@ -520,6 +521,7 @@ void AC_AutoTune_Heli::load_intra_test_gains()
 // called by control_attitude() just before it beings testing a gain (i.e. just before it twitches)
 void AC_AutoTune_Heli::load_test_gains()
 {
+    attitude_control->use_sqrt_controller(false);
```
bnsgeyer: This should stay true for heli.

bnsgeyer (approved):
Tested for heli to ensure testing failure resulted in original gains being loaded.  Only able to test Max gain determination failure but that should cover the frequency exceeded failure as well since they are contained in the same conditional statement.  Everything check out fine.

rmackay9:
@lthall, it looks like there's a WIP PR there with only white space changes, if you're happy to remove that then I think we can merge this

rmackay9:
I have not included this in 4.6.2-beta1 due to a merge conflict.  If someone could create a backport PR against the ArduPilot-4.6 branch that would be a great help!

Issue #30152, opened 2025-05-24
ACRO_BAL_ROLL/ACRO_BAL_PITCH Parameter Value Range Not Checked Causes Crash/Takeoff Failure In Acro Mode


Lqs66:
## Bug report

**Issue details**

Ardupilot did not check the range of the parameters ACRO_BAL_ROLL and ACRO_BAL_PITCH. Entering values out of range (e.g. ACRO_BAL_ROLL=-5, ACRO_BAL_PITCH=-5) caused the drone to crash or fail to take off in acro mode.

**replay**

1. Start sitl and enter the following commands at the mavproxy console :
  1.1. `param set ACRO_BAL_PITCH -5.0` or `param set ACRO_BAL_ROLL -5.0`;
  1.2. `mode acro`;
  1.3. `arm throttle`;
  1.4. `rc 3 1500`.
The drone cannot take off.

2. Start sitl and enter the following commands at the mavproxy console :
 2.1. `mode acro`;
 2.2. `arm throttle`;
 2.3. `rc 3 1500`;
 2.4. `param set ACRO_BAL_ROLL -10.0`;
 2.5. `rc 1 1400`.
The drone will fall down.

**Version**
Copter-4.4

**Platform**
[  ] All
[  ] AntennaTracker
[x] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
SITL

**Hardware type**
None

**Logs**
None

PR #30176, opened 2025-05-27
Added range check for parameters ACRO_BAL_PITCH and ACRO_BAL_ROLL in ACRO mode

Lqs66:
This PR is for fixing issue #30152.


ArduCopter/mode_althold.cpp
```diff
@@ -40,7 +40,7 @@ void ModeAltHold::run()
 
     // get pilot desired climb rate
     float target_climb_rate = get_pilot_desired_climb_rate();
-    target_climb_rate = constrain_float(target_climb_rate, -get_pilot_speed_dn(), g.pilot_speed_up);
+    target_climb_rate = constrain_float(target_climb_rate, -get_pilot_speed_dn(), constrain_int16(g.pilot_speed_up, 50, 500));
```
peterbarker: Wait, this doesn't look right?

Why are we constraining pilot climb rate between 0.5m/s and 5m/s here?  Some vehicles and scenarios could be *well* outside these ranges!

ArduCopter/mode_acro.cpp
```diff
@@ -136,11 +136,13 @@ void ModeAcro::get_pilot_desired_angle_rates(float roll_in, float pitch_in, floa
 
         // Calculate trainer mode earth frame rate command for roll
         int32_t roll_angle = wrap_180_cd(att_target.x);
-        rate_ef_level_cd.x = -constrain_int32(roll_angle, -ACRO_LEVEL_MAX_ANGLE, ACRO_LEVEL_MAX_ANGLE) * g.acro_balance_roll;
+        float limited_acro_balance_roll = constrain_float(g.acro_balance_roll, 0.0f, 3.0f);
```
peterbarker: This limit seem arbitrary to me.  Even if the user goes outside this range the effects are probably not going to be particularly dire.

Hwurzburg:
Changing pilot set params, even if out of range, without at least a GCS warning is a big problem as far as I am concerned...any param should be able to be set "out of range", as long as it doesn't crash code (and THAT should be protected in the code) for any number of reasons...testing. unique vehicle,etc.
if this is going to be done it needs a GCS warning that its been done....more code bloat, however

lthall:
There is a problem highlighted here but I don't like this fix.

Best I can tell we only need to add a preflight check that acro_balance and climb rate are positive.

peterbarker:
@Lqs66 is this something you want to pursue?

If so, please remove changes related to pilot speed up/down and change to simply constrain to be >= 0.

Lqs66:
> @Lqs66 is this something you want to pursue?
> 
> If so, please remove changes related to pilot speed up/down and change to simply constrain to be >= 0.

I'm sorry for replying to you just now. I determined the parameter range according to the regulations in the ArduPilot documentation.
For the ACRO_BAL_PITCH parameter, the documentation specifies an acceptable input range of 0-3 (https://ardupilot.org/copter/docs/parameters.html#acro-bal-pitch-acro-balance-pitch%EF%BC%89).
For the ACRO_BAL_ROLL parameter, the documentation specifies an acceptable input range of 0-3 (https://ardupilot.org/copter/docs/parameters.html#acro-bal-roll-acro-balance-roll%EF%BC%89).
Is the "change to simply constrain to be >= 0" you mentioned referring to these two parameters?

ArduCopter/mode_althold.cpp
Lqs66: > Wait, this doesn't look right?
> 
> Why are we constraining pilot climb rate between 0.5m/s and 5m/s here? Some vehicles and scenarios could be _well_ outside these ranges!

For the PILOT_SPEED_UP parameter, I set the constraints according to the regulations in the documentation.
(https://ardupilot.org/copter/docs/parameters.html#pilot-speed-up-pilot-maximum-vertical-speed-ascending)

peterbarker:
> I'm sorry for replying to you just now. I determined the parameter range according to the regulations in the ArduPilot documentation. For the ACRO_BAL_PITCH parameter, the documentation specifies an acceptable input range of 0-3 

Yeah, those limits are *absolutely advisory*.  People frequently need to go outside of the recommended ranges.

> Is the "change to simply constrain to be >= 0" you mentioned referring to these two parameters?

Yes - just limit so that you can't specify a negative value, which should have been sufficient to avoid the crash (assuming you're capable of uprighting the vehicle yourself :-)

Lqs66:
> > I'm sorry for replying to you just now. I determined the parameter range according to the regulations in the ArduPilot documentation. For the ACRO_BAL_PITCH parameter, the documentation specifies an acceptable input range of 0-3
> 
> Yeah, those limits are _absolutely advisory_. People frequently need to go outside of the recommended ranges.
> 
> > Is the "change to simply constrain to be >= 0" you mentioned referring to these two parameters?
> 
> Yes - just limit so that you can't specify a negative value, which should have been sufficient to avoid the crash (assuming you're capable of uprighting the vehicle yourself :-)

I have addressed the comments and modified the PR :-)

peterbarker:
@Lqs66 thankyou for the patches here.

@lthall has noted the problems and we've merged [patches from him](https://github.com/ArduPilot/ardupilot/pull/31435) addressing the issue.

Apologies we didn't base the fix on your work here.

PR #30179, opened 2025-05-27
autotest: Create README file about the logmessage parser

shancock884:
This PR creates a README.md file in the Tools/autotest/logger_metadata folder, to give an overview of how the logger metadata processing works, such as which tags are used to generate the data and what files get created.

To support the README, there is also a minor update to the enum_parse.py script to allow users to dump out a list of all available enums, to help with completing the fields. (if this would be better as separate commit or seperate PR, let me know).
Output looks like this:
```
(venv-ardupilot) $ python enum_parse.py --verbose --vehicle Plane
EnumDocco.Enumeration: AP_Tuning_Plane::tuning_func [37 entries]
EnumDocco.Enumeration: AP_Tuning_Plane::tuning_sets [13 entries]
EnumDocco.Enumeration: QuadPlane::ThrustType [3 entries]
EnumDocco.Enumeration: QuadPlane::ACTION [2 entries]
EnumDocco.Enumeration: QuadPlane::RTL_MODE [4 entries]
...
```


Tools/autotest/logger_metadata/README.md
```diff
@@ -0,0 +1,100 @@
+# Logger Metadata: Automatic source code parser
+
+The scripts in this folder are used to automatically generate details
+of the log messages generated by Ardupilot in DataFlash log files (.BIN).
```
peterbarker, line 4:
```suggestion
of the log messages generated by ArduPilot in "DataFlash" log files (.BIN).
```

Tools/autotest/logger_metadata/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Execution
+
+The script is run as part of the automated testing, to allow the website to
+be updated automatically with any new changes. The latest files can be found [here](https://autotest.ardupilot.org/LogMessages/), with specific subfolders for each vehicle.
+
+To run the script manually for local testing, use the following command:
+```
+python parse.py --vehicle <vehicle>
```
peterbarker, line 13:
```suggestion
python3 parse.py --vehicle <vehicle>
```

Tools/autotest/logger_metadata/README.md
```diff
@@ -0,0 +1,100 @@
...
+```
+Where `<vehicle>` can be one of: Plane, Copter, Sub, Blimp, Rover or Tracker.
+
+For each vehicle type, the following files are generated:
+* **LogMessages.html** - HTML file (not used)
+* **LogMessages.md** - Markdown format file (not used)
+* **LogMessages.rst** - Rust format file - used for populating the Log Message
```
peterbarker, line 20: Not rust

Tools/autotest/logger_metadata/README.md
```diff
@@ -0,0 +1,100 @@
...
+or .h file, alongside the use of the message. The basic format is as follows:
+```
+// @LoggerMessage: <message name>
+// @Description: <message description>
+// @Field: <field name>: <field description>
+// @Field ...
+```
+If the same field set is used by multiple messages (for example the PIDx messages),
+then the `@LoggerMessage` and `@Description` lines may be repeated multiple times
+before the start of the fields.
```
peterbarker, line 37: Not understanding this bit

Tools/autotest/logger_metadata/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Enumerations and Bitmasks
+
+A `@Field` definition may be followed by one of the following, to link it to an
+enumeration or bitmask as follows:
+```
+// @FieldBits: <field name>: <bit 0 name>,<bit 1 name>,...
```
peterbarker, line 44: We should remove this one.  Only one user and FieldBitMaskEnum is much better

Tools/autotest/logger_metadata/README.md
```diff
@@ -0,0 +1,100 @@
...
+```
+Or for a more simple case:
+```
+enum <enum name> {
+    <enum entry name>,
+    ...
+}
+```
+
+Additionally, groups of `#define` statements can be extracted as an enumeration#
```
peterbarker, line 73: Is that # supposed to be on the end?

Tools/autotest/logger_metadata/README.md
```diff
@@ -0,0 +1,100 @@
...
+if surrounded by `@LoggerEnum` and `@LoggerEnumEnd` tags, as follows:
```
peterbarker, line 74: AuxFuncTrigger is actually missing `@LoggerEnumEnd`.... so this is wrong or (more likely) whatever's going on over in RC_CHannels is wrong..

Tools/autotest/logger_metadata/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+For each message, the format, unit and multiplier characters are extracted from
+either:
+* Calls to the Write, WriteStreaming or WriteCritical methods on APLogger.
+* LogStructure definitions.
+
+Note that the multiplier character is used to add the correct prefix to the unit,
+and is NOT used by log readers tools on the raw value itself. For example if a
+field has a unit of "A" and a multiplier of 0.001, this means the value is in mA,
+and will be shown to users as such in tools such as MAVExplorer.
```
peterbarker, line 100: ... Tools are encouraged to translate these units into the units the user may prefer.

peterbarker (changes requested):
Ping @Hwurzburg - lots of good information in here.

Tools/autotest/logger_metadata/README.md
shancock884: At present I don't think the one user can be replaced by an enum, as it comes from a struct with a number of 1-bit fields, which are not parsed. "field1:1; field2:1; etc".
You prefer to:
- keep functionality for this user, but don't document it, to avoid anyone else using it?
- remove from TECS, even if no replacement available?
- remove from TECS, but only after implementing 1-bit struct parsing to replace it?
(From memory there are other places using 1-bit struct fields, so doing this could help in other places).

Tools/autotest/logger_metadata/README.md
shancock884: Ah "reStructuredText" apparently! Learn something everyday...

Tools/autotest/logger_metadata/README.md
shancock884: Text updated. Let me know if better/good enough.

PR #30182, opened 2025-05-27
AP_Follow fix for lua calls and threading issue.

lthall:
This moves the update_estimates call to the main vehicle code and turns all calls to retrieve pos, vel, accel, heading ect, into const.

Still may need a "lock" or what ever that is called.

### DIFF
--- a/ArduPlane/Plane.cpp
+++ b/ArduPlane/Plane.cpp
@@ -206,7 +206,10 @@ void Plane::ahrs_update()
     quadplane.inertial_nav.update();
     if (quadplane.available()) {  
         quadplane.pos_control->update_estimates();  
-    }  
+    }
+#endif
+#if AP_SCRIPTING_ENABLED && AP_FOLLOW_ENABLED
+        g2.follow.update_estimates();
 #endif
 
 #if HAL_LOGGING_ENABLED
--- a/Rover/Rover.cpp
+++ b/Rover/Rover.cpp
@@ -323,6 +323,10 @@ void Rover::ahrs_update()
     } else if (gps.status() >= AP_GPS::GPS_OK_FIX_3D) {
         ground_speed = ahrs.groundspeed();
     }
+    
+#if AP_FOLLOW_ENABLED
+    g2.follow.update_estimates();
+#endif
 
 #if HAL_LOGGING_ENABLED
     if (should_log(MASK_LOG_ATTITUDE_FAST)) {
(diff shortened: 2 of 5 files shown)


ArduPlane/Plane.cpp
```diff
@@ -206,6 +206,9 @@ void Plane::ahrs_update()
     quadplane.inertial_nav.update();
     if (quadplane.available()) {  
         quadplane.pos_control->update_estimates();  
+#if AP_SCRIPTING_ENABLED && AP_FOLLOW_ENABLED
+        g2.follow.update_estimate();
```
peterbarker, line 210: Maybe pluralise to match pos_control

Rover/Rover.cpp
```diff
@@ -323,6 +323,10 @@ void Rover::ahrs_update()
     } else if (gps.status() >= AP_GPS::GPS_OK_FIX_3D) {
         ground_speed = ahrs.groundspeed();
     }
+    
+#if AP_FOLLOW_ENABLED
```
peterbarker, line 327: Maybe put this just below the update?

libraries/AP_Follow/AP_Follow.cpp
```diff
@@ -325,17 +324,15 @@ bool AP_Follow::update_estimate()
...
-    if (is_positive(_dist_max_m.get()) && (dist_vec_ned_m.length() > _dist_max_m)) {
-        return false;
+    if (AP::ahrs().get_relative_position_NED_origin(current_position_ned_m)) {
+        const Vector3p dist_vec_ned_m = _target_pos_ned_m - current_position_ned_m;
+        // If _dist_max_m is not positive, we don't check the distance
+        if (is_positive(_dist_max_m.get()) && (dist_vec_ned_m.length() > _dist_max_m)) {
+            _estimate_valid = false;
+        }
+    } else {
+        _estimate_valid = false;
```
peterbarker, line 334:
```suggestion
    if (!AP::ahrs().get_relative_position_NED_origin(current_position_ned_m)) {
        // no idea where we are; knowing where other things are won't help.
        _estimate_valid = false;
        return;
    }
    const Vector3p dist_vec_ned_m = _target_pos_ned_m - current_position_ned_m;
    // If _dist_max_m is not positive, we don't check the distance
    if (is_positive(_dist_max_m.get()) && (dist_vec_ned_m.length() > _dist_max_m)) {
        // target is too far away, mark the estimate invalid
        _estimate_valid = false;
    }
```

peterbarker (approved):

timtuxworth:
How does it get called from Plane? Doesn't it need a binding?

rmackay9:
I think the CI failures are because all calls (in copter) to the g2.follow object must be "bracketed" with lines like this:

```
#if MODE_FOLLOW_ENABLED
    g2.follow...
#end
```

Rover/Rover.cpp
rmackay9: I think it's OK where it is actually.. it makes sense..

PR #30185, opened 2025-05-28
AP_AIS: Populate OADatabase for avoidance

IamPete1:
This adds support for AIS vessels as a source for avoidance data. This allows a vehicle to navigate around them. It also does some minor fix-ups to both AIS and the OADatabase.

image

Features: 

- This converts the timeout parameter from a int8 to a int16, this allows longer timeouts. AIS messages can be quite low update-rates, anchored vessels are only required to update every 3 mins. So the timeout must be longer than 180 seconds, 127 was the maximum with the int8. A timeout of 10 minutes (600 seconds) might be a good value for AIS avoidance.
- The item refresh code now updates position, the current behavior is to just change the timeout, but not the position. This means the avoidance item will not update position until it is has changed by one radius, which here could be >100m.
- It is now possible for the data source to provide a radius for the item. In the case of AIS we can get the vessel dimensions so can provide the exact size.
- The max distance threshold now applies to the closest point not the center of the item, this means we don't ignore items with large radius.

Some shortcomings: 

- AIS is only supported in "path planning" avoidance. It is not included in simple avoidance. The other avoidance methods, proximity sensor and fences, are supported by both simple and "path planning" avoidance. We could add AIS to simple avoidance in the future.
- There is no way to ignore AIS from "path planning" but still use proximity. We could fix this by adding a sources bitmask to the OADatabase.
- The defaults don't work that well, everything is on a much larger scale. For example I set `OA_BR_LOOKAHEAD` to 500 (the default is 15) `OA_DB_RADIUS_MIN` to 10 (default of 0.01). This means would probably make it hard to find a set of params that would work well for AIS avoidance at sea and proximity avoidance in a harbor. 
- I'm not sure how well the avoidance would get on with large fast moving obstacles, we could plausibly have a 3m boat traveling at 3 knots trying to avoid a 300m vessel traveling at 30 knots. Such vessels won't change direction quickly, so some sort of look ahead or velocity projection could be used.

### DIFF
--- a/libraries/AC_Avoidance/AP_OADatabase.h
+++ b/libraries/AC_Avoidance/AP_OADatabase.h
@@ -31,16 +31,26 @@ public:
         Vector3f pos;           // position of the object as an offset in meters from the EKF origin
         uint32_t timestamp_ms;  // system time that object was last updated
         float radius;           // objects radius in meters
+
+        // Unique ID, assigned by the source, not used by all sources
+        uint32_t id;
+
         uint8_t send_to_gcs;    // bitmask of mavlink comports to which details of this object should be sent
         OA_DbItemImportance importance;
+
+        // Source of item, refresh treats sources differently
+        enum class Source {
+            proximity,
+            AIS
+        } source;
     };
 
     void init();
     void update();
 
     // Push an object into the database. Pos is the offset in meters from the EKF origin, measurement timestamp in ms, distance in meters, optional radius in meters
-    void queue_push(const Vector3f &pos, const uint32_t timestamp_ms, const float distance, float radius);
-    void queue_push(const Vector3f &pos, const uint32_t timestamp_ms, const float distance);
+    void queue_push(const Vector3f &pos, const uint32_t timestamp_ms, const float distance, float radius, const OA_DbItem::Source source, const uint32_t id = 0);
+    void queue_push(const Vector3f &pos, const uint32_t timestamp_ms, const float distance, const OA_DbItem::Source source, const uint32_t id = 0);
 
     // returns true if database is healthy
     bool healthy() const { return (_queue.items != nullptr) && (_database.items != nullptr); }
@@ -67,16 +77,16 @@ private:
 
     // database item management
     void database_item_add(const OA_DbItem &item);
-    void database_item_refresh(const uint16_t index, const OA_DbItem &item);
+    void database_item_refresh(OA_DbItem &current_item, const OA_DbItem &new_item) const;
     void database_item_remove(const uint16_t index);
     void database_items_remove_all_expired();
 
     // get bitmask of gcs channels item should be sent to based on its importance
     // returns 0xFF (send to all channels) if should be sent or 0 if it should not be sent
-    uint8_t get_send_to_gcs_flags(const OA_DbItemImportance importance);
+    uint8_t get_send_to_gcs_flags(const OA_DbItemImportance importance) const;
 
-    // returns true if database item "index" is close to "item"
-    bool is_close_to_item_in_database(const uint16_t index, const OA_DbItem &item) const;
+    // Return true if item A is likely the same as item B
+    bool item_match(const OA_DbItem& A, const OA_DbItem& B) const;
 
     // enum for use with _OUTPUT parameter
     enum class OutputLevel {
--- a/libraries/AC_Avoidance/AP_OADatabase.h
+++ b/libraries/AC_Avoidance/AP_OADatabase.h
@@ -66,7 +66,7 @@ private:
 
     // database item management
     void database_item_add(const OA_DbItem &item);
-    void database_item_refresh(const uint16_t index, const uint32_t timestamp_ms, const float radius);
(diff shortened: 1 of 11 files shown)


timtuxworth:
It's a good start, but we will need to add ways to deal with the lookahead and radius "depending on circumstances" - I have had to adjust them for plane when making BR work for plane (which works quite we'll so far). 

I'm thinking these values need to be variables, that can be set from parameters and also in Lua (and from c++)?

IamPete1:
Maybe we need to split parameters per source. It certainly makes sense to have a different timeout and min radius for proximity/AIS/ADSB. Beam width only really makes sense for proximity.

timtuxworth:
> Maybe we need to split parameters per source. It certainly makes sense to have a different timeout and min radius for proximity/AIS/ADSB. Beam width only really makes sense for proximity.

In the bendy ruler version for Plane, the radius can be different depending on the type of obstacle - larger for clouds or planes, smaller for birds or other UAVs.

libraries/SITL/SITL.cpp
```diff
@@ -1258,6 +1258,17 @@ const AP_Param::GroupInfo SIM::var_ins[] = {
     // @Description: If non-zero the vehicle will be clamped in position until the value on this servo channel passes 1800PWM
     AP_GROUPINFO("CLAMP_CH",     49, SIM, clamp_ch, 0),
 
+    // @Param: AIS_COUNT
+    // @DisplayName: Number of AIS vessels
+    // @Description: Total number of AIS simulated vessels
+    AP_GROUPINFO("AIS_COUNT", 50, SIM,  ais_vessel_count, -1),
```
tridge: I would prefer the parameters to be in SIM_AIS.cpp

libraries/SITL/SITL.h
```diff
@@ -598,6 +598,10 @@ class SIM {
     // clamp simulation - servo channel starting at offset 1 (usually ailerons)
     AP_Int8 clamp_ch;
 
+    // AIS vessel simulation
+    AP_Int8 ais_vessel_count;
```
tridge: should be in SIM_AIS.h

tridge (dismissed):

rmackay9:
If possible I think instead of moving the position within the OA Database we should pass down an id

libraries/SITL/SITL.cpp
IamPete1: The problem is that this serial driver lives in `AP_HAL_SITL`. So here in `SITL` we don't have a reference to get at the params by. 

I could still put them in there own table, but they I don't think they can go in the AIS class.

IamPete1:
> If possible I think instead of moving the position within the OA Database we should pass down an id

Added source and a ID to the database.

libraries/AC_Avoidance/AP_OADatabase.h
```diff
@@ -31,15 +31,26 @@ class AP_OADatabase {
         Vector3f pos;           // position of the object as an offset in meters from the EKF origin
         uint32_t timestamp_ms;  // system time that object was last updated
         float radius;           // objects radius in meters
+
+        // Unique ID, assigned by the source, not used by all sources
+        uint32_t id;
```
rmackay9, line 36: I guess this increases memory usage by about 1k bytes (5 bytes for each of 100 items).  I'm OK with this but FYI

rmackay9 (approved):
LGTM.  Ideally I'd like to see a bit of testing of the existing proximity sensor to show that the little circles continue to appear around barriers when using mission planner and ADSB messages are enabled.  [Dev wiki is here](so about an extra 1K of memory)... still, if PeterH is confident that it's all working then I'm OK with merging

libraries/AC_Avoidance/AP_OADatabase.h
IamPete1: I just found that by setting the type of importance the size stays the same. I think importance was defaulting to a int before and due to alignment that meant we skipped 3 bytes. Changing importance to a uint8 gets us back 6 bytes (4 + 3 - 1). This PR adds 5 bytes but we loose that last one to alignment.

So before and after were still on 28 bytes per item.

If we did the same optimization on the old struct we could have got it down to 24 bytes.

https://github.com/ArduPilot/ardupilot/pull/30185/commits/129634a7de7fa385c72d02487582d7d1dadd817b

IamPete1:
I have run copter SITL to test proximity items. These are the magic posts:

image

Bendy ruler still works on copter: 

image

rmackay9:
There's a request from Tridge that I'm not sure has been addressed or not.  Still, I think this can go in so I've put the "MergeOnCIPass" label on it

Sorry, I see there is discussion in the "code_review" channel so I've removed the "MergeOnCIPass".  I suspect that Tridge will say it's fine and we can merge but I guess we should confirm.

libraries/SITL/SITL.cpp
IamPete1: Done, it is a little odd in that you have to set the serial port protocol and enable AIS so begin is called on the port and the simulator is created and the params show up.

rmackay9:
Merged, thanks!

PR #30187, opened 2025-05-28
GCS_MAVLink: send dropped message count in SYS_STATUS.errors_count3

peterbarker:
for first backend instance

Useful for knowing if you are going to hav a good replay log / get all of the raw IMU data you want etc.

Tested using the following patch:
```
--- a/libraries/AP_Logger/AP_Logger_File.cpp
+++ b/libraries/AP_Logger/AP_Logger_File.cpp
@@ -454,6 +454,11 @@ bool AP_Logger_File::_WritePrioritisedBlock(const void *pBuffer, uint16_t size,
 
     uint32_t space = _writebuf.space();
 
+    if (rand() /float(RAND_MAX) < 0.01) {
+        _dropped++;
+        return false;
+    }
+
     if (_writing_startup_messages &&
         _startup_messagewriter->fmt_done()) {
         // the state machine has called us, and it has finished
```

image

### DIFF
--- a/libraries/GCS_MAVLink/GCS_Common.cpp
+++ b/libraries/GCS_MAVLink/GCS_Common.cpp
@@ -5901,6 +5901,12 @@ void GCS_MAVLINK::send_sys_status()
     const uint16_t errors2 = (errors>>16) & 0xffff;
     const uint16_t errors4 = AP::internalerror().count() & 0xffff;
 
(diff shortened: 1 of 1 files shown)


peterbarker:
This will conflict with https://github.com/ArduPilot/ardupilot/pull/20949

tridge (approved):

PR #30188, opened 2025-05-28
AR_Motors: added MOT_REV_DELAY

tridge:
adds an optional delay when reversing motors for ESCs that need it

this does not yet try to stop integrator windup while reversing. I suspect that is not needed for many ESCs as the required delay is quite small (eg. the ESCs I have need a 0.3s delay)
Also note that all ESCs actually do have a delay on reversal, so we already have the integrator issue, it is just that this makes the delay inherent in the signal sent to the ESC, rather than only being in the internals of the ESC

### DIFF
--- a/libraries/AR_Motors/AP_MotorsUGV.cpp
+++ b/libraries/AR_Motors/AP_MotorsUGV.cpp
@@ -118,6 +118,15 @@ const AP_Param::GroupInfo AP_MotorsUGV::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("THST_ASYM", 14, AP_MotorsUGV, _thrust_asymmetry, 1.0f),
 
+    // @Param: REV_DELAY
+    // @DisplayName: Motor reversal delay
+    // @Description: For reversible motors that need a delay before they can change direction. When greater than zero the throttle will go to zero for this amount of time before outputting the new throttle when the demanded motor direction changes.
+    // @Units: s
+    // @Range: 0.1 1.0
+    // @Increment: 0.1
+    // @User: Standard
+    AP_GROUPINFO("REV_DELAY", 15, AP_MotorsUGV, _reverse_delay, 0),
+    
     AP_GROUPEND
 };
 
@@ -1028,6 +1037,23 @@ void AP_MotorsUGV::output_throttle(SRV_Channel::Function function, float throttl
     }
 #endif  // AP_RELAY_ENABLED
 
+    if (_reverse_delay > 0) {
+        switch (function) {
+        case SRV_Channel::k_throttle:
+            rev_delay_throttle.output(function, throttle, _reverse_delay);
+            return;
+        case SRV_Channel::k_throttleLeft:
+            rev_delay_throttleLeft.output(function, throttle * 10, _reverse_delay);
+            return;
+        case SRV_Channel::k_throttleRight:
+            rev_delay_throttleRight.output(function, throttle * 10, _reverse_delay);
+            return;
+        default:
+            // fall through to other non-delayed outputs
+            break;
+        }
+    }
+
     // output to servo channel
     switch (function) {
         case SRV_Channel::k_throttle:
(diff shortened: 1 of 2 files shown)


libraries/AR_Motors/AP_MotorsUGV.cpp
```diff
@@ -1172,6 +1196,23 @@ bool AP_MotorsUGV::is_digital_pwm_type() const
     return false;
 }
 
+/*
+  handle delay on reversal for a throttle
+ */
+void AP_MotorsUGV::ReverseThrottle::output(SRV_Channel::Function function, float throttle, float delay)
+{
+    const uint32_t now_ms = AP_HAL::millis();
+    if (throttle * last_throttle < 0 &&
```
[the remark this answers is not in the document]
tridge: that suggestion is incorrect

libraries/AR_Motors/AP_MotorsUGV.cpp
```diff
@@ -1172,6 +1196,23 @@ bool AP_MotorsUGV::is_digital_pwm_type() const
...
+        now_ms - last_output_ms < delay*1000) {
+        // sign change, add pause
+        throttle = 0;
```
IamPete1, line 1208: You should set `limit.throttle_lower` or `limit.throttle_upper` in this case so we don't windup I. You can't do it here because you only need to set the steer limits for the left and right motors. Maybe you could make this a bool function that return true if the limit has been applied. Then you can set the limit that go with each function.

libraries/AR_Motors/AP_MotorsUGV.cpp
```diff
@@ -118,6 +118,13 @@ const AP_Param::GroupInfo AP_MotorsUGV::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("THST_ASYM", 14, AP_MotorsUGV, _thrust_asymmetry, 1.0f),
 
+    // @Param: REV_DELAY
+    // @DisplayName: Motor reversal delay
+    // @Description: For reversible motors that need a delay before they can reverse. When greater than zero the throttle will go to zero for this amount of time before outputting the new throttle.
```
IamPete1, line 123: I've no suggestions for wording, but I think it needs to be clear that is the motor reversing direction not just the vehicle going into reverse. That is to say, the delay will still apply when going from vehicle reverse to vehicle forwards.

libraries/AR_Motors/AP_MotorsUGV.cpp
peterbarker, line 1205: What happens if either one of these is zero?  eg. could we NOT pause when w need to because last_throttle is zero here?

IamPete1:
This also breaks slew limiting, the slew limit will not be aware of the zero hold, so when the hold is ended throttle may exceed the set slew limit.

libraries/AR_Motors/AP_MotorsUGV.cpp
tridge: fixed

tridge:
@IamPete1 we actually already have the jump and integrator build up in existing rovers and users don't seem to find it a problem. The slew limit is on the mixer input throttle, not the output throttle and there is no limit on steering rate of change, so we already jump when you steer.
Also, all ESCs have a delay when you change direction. It is an inherent part of how an ESC works. So the integrator build-up and other related issues already happens. If you set a really high MOT_REV_DELAY then it would be more noticible, but the actual values needed for ESCs is quite low. On the ESCs I did this for it is 0.3s. I suspect even if you could configure the ESC to not need the zero throttle to change then the ESC would still stop for around 0.3s. A graph of RPM vs throttle will not be a straight line through zero on any ESC.
So I'd vote for the lazy solution and see how well this works. If it turns out to be a real issue we can do something fancier

libraries/AR_Motors/AP_MotorsUGV.cpp
tridge: see comment below on why this already affects ESCs without this feature

libraries/AR_Motors/AP_MotorsUGV.cpp
tridge: fixed

IamPete1:
> So the integrator build-up and other related issues already happens. If you set a really high MOT_REV_DELAY then it would be more noticible, but the actual values needed for ESCs is quite low. On the ESCs I did this for it is 0.3s.

I think its fine to apply the lazy solution if you also constrain the timeout to a small number. 2 seconds maybe?

tridge:
> I think its fine to apply the lazy solution if you also constrain the timeout to a small number. 2 seconds maybe?

the documented range is 0.1 to 1.0 seconds. If the user goes outside that range then the GCS should warn. Putting an absolute restriction is I think unnecessary - I could imagine huge rovers (eg. something the size of a tank or barge) that needs a long time to reverse an engine, but where the time constant for control is so long that 2s is not an issue

rmackay9 (approved):

PR #30189, opened 2025-05-28
Copter: add FS_EKF_ACTION report only and notify in all cases

IamPete1:
This changes to always trigger the notify library and send a message in EKF fail-safes if armed, this should make it clear to the user that there is a issue. For example if flying in stabilize currently it would not be clear to the pilot that the vehicle is in a EKF failsafe, they would just find they can't switch to position control for some reason.

One disadvantage of this is that EKF failsafes will get more annoying which might lead to complaints.

It also adds a report only option for that will only trigger the notify state and a send text. In such a case the the pilot can decide what the best course of action is. The default value of the parameter is Land from position control modes.

### DIFF
--- a/ArduCopter/Parameters.cpp
+++ b/ArduCopter/Parameters.cpp
@@ -364,7 +364,7 @@ const AP_Param::Info Copter::var_info[] = {
     // @Param: FS_EKF_ACTION
     // @DisplayName: EKF Failsafe Action
     // @Description: Controls the action that will be taken when an EKF failsafe is invoked
-    // @Values: 1:Land, 2:AltHold, 3:Land even in Stabilize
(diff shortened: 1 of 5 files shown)


rmackay9:
Thanks for this.  In order to make it less annoying I wonder if it would help to only enable the notification after the EKF has calculated a good position estimate for a few seconds at least.

IamPete1:
> In order to make it less annoying I wonder if it would help to only enable the notification after the EKF has calculated a good position estimate for a few seconds at least.

It won't trigger if disarmed. Hopefully you can't arm until the EKF is happy.

IamPete1:
I have done some more testing. There key difference in notify behavior is that we will now play a tune, the `failsafe_ekf` flag in notify is only used for the tune. We do flash a color pattern in all cases already because of the `ekf_bad` flag.

The second change is that we now always get a "EKF Failsafe" message, there is already a "EKF variance" warning. So you now get both.

rmackay9:
I've approved it but I'd like to know what changes the user will see

PR #30190, opened 2025-05-28
Copter: guided `do_reposition` and `set_position_target` honor relative yaw

IamPete1:
Both the MAVLink and our internal APIs support relative yaw, we just need pass the flags down.


IamPete1:
Oops, I have forgotten the MAVLink changes that this needs. I will PR those first and then re-open.

PR #30199, opened 2025-05-29
AP_HAL_ChibiOS:add vimdrones ap_periph Battery Tag target

Huibean:
add vimdrones ap_periph Battery Tag target


libraries/AP_HAL_ChibiOS/hwdef/VM-L431-BatteryTag/defaults.parm
```diff
@@ -0,0 +1,3 @@
+FORMAT_VERSION 2
```
peterbarker, line 1: This shouldn't appear in here

peterbarker (changes requested):
There should be two commits here as you're also adding bootloaders to the `Tools` directory

libraries/AP_HAL_ChibiOS/hwdef/VM-L431-BatteryTag/defaults.parm
Huibean: updated, thanks!

PR #30209, opened 2025-05-31
Update mode_acro.cpp

Dyfflen:
Double #include "Sub.h"


andyp1per:
The fix is correct, but the commit message needs to start with "Sub:"

peterbarker (changes requested):
@Dyfflen @andyp1per meant the commit message, not the PR title.

peterbarker:
I've fixed up the commit messages.

PR #30210, opened 2025-06-01
Plane: Explanation of an FS_GCS_ENABL edge case

landswellsong:
Relevant code: https://github.com/ArduPilot/ardupilot/blob/3038c70637297d78ecdbbf59395f2b4449e9fce8/ArduPlane/system.cpp#L368-L376

Before `gcs_last_seen_ms` is set by a heartbeat from `SYSID_MYGCS` exactly, the conditions fail, ignoring all other connected GCSs. Learned the hard way 🙃 Probably useful to break it down in the documentation, it makes sense, but since it's not written anywhere it's hard to guess that the system would behave this way.

Unsure to which extent it applies to other vehicles as I don't have neither code not usage experience with anything outside Plane.


ArduPlane/Parameters.cpp
```diff
@@ -469,7 +469,7 @@ const AP_Param::Info Plane::var_info[] = {
...
...lsafe. Failsafe will trigger after FS_LONG_TIMEOUT seconds of no MAVLink heartbeat messages. There are three possible enabled settings. Setting FS_GCS_ENABL to 1 means that GCS failsafe will be triggered when the aircraft has not received a MAVLink HEARTBEAT message. Note that heartbeat tracking only becomes active after having received the first heartbeat from the SYSID_MYGCS system. Setting FS_GCS_ENABL to 2 means that GCS failsafe will be triggered on either a loss of HEARTBEAT messages, or a RADIO_STATUS message from a MAVLink enabled 3DR radio indicating that the ground station is not receiving status updates from the aircraft, which is indicated by the RADIO_STATUS.remrssi field being zero (this may happen if you have a one way link due to asymmetric noise on the ground station and aircraft radios).Setting FS_GCS_ENABL to 3 means that GCS failsafe will be triggered by Heartbeat(like option one), but only in AUTO mode. WARNING: Enabling this option opens up the possibility of your plane going into failsafe mode and running the motor on the ground it it loses contact with your ground station. If this option is enabled on an electric plane then you should enable ARMING_REQUIRED.
```
Hwurzburg:
```suggestion
    // @Description: Enable ground control station telemetry failsafe. Failsafe will trigger after FS_LONG_TIMEOUT seconds of no MAVLink heartbeat messages. There are three possible enabled settings. Setting FS_GCS_ENABL to 1 means that GCS failsafe will be triggered when the aircraft has not received a MAVLink HEARTBEAT message. Note that heartbeat tracking only becomes active after having received the first heartbeat from the MAV_GCS_SYSID primary GCS system. Setting FS_GCS_ENABL to 2 means that GCS failsafe will be triggered on either a loss of HEARTBEAT messages, or a RADIO_STATUS message from a MAVLink enabled telemetry adio indicating that the primary ground station is not receiving status updates from the aircraft, which is indicated by the RADIO_STATUS.remrssi field being zero (this may happen if you have a one way link due to asymmetric noise on the ground station and aircraft radios).Setting FS_GCS_ENABL to 3 means that GCS failsafe will be triggered by Heartbeat(like option one), but only in AUTO mode. WARNING: Enabling this option opens up the possibility of your plane going into failsafe mode and running the motor on the ground it it loses contact with your ground station. If this option is enabled on an electric plane then you should enable ARMING_REQUIRED.
```

Hwurzburg: I have update the wiki also

landswellsong: Amended to this version and force pushed

Hwurzburg (approved):

PR #30212, opened 2025-06-01
AP_SerialManager: allow for permanent passthrough

tridge:
it is sometimes useful to setup a serial passthrough as permanent, this allows for that by setting SERIAL_PASSTIMO to -1

### DIFF
--- a/libraries/AP_SerialManager/AP_SerialManager.cpp
+++ b/libraries/AP_SerialManager/AP_SerialManager.cpp
@@ -323,15 +323,15 @@ const AP_Param::GroupInfo AP_SerialManager::var_info[] = {
 
     // @Param: _PASS2
     // @DisplayName: Serial passthru second port
-    // @Description: This sets one side of pass-through between two serial ports. Once both sides are set then all data received on either port will be passed to the other port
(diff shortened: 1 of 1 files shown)


andyp1per (commented):
Although I think this is useful there is some danger - I have sometimes made my flight controllers inaccessible and the timeout has saved my bacon.

tridge:
> I have sometimes made my flight controllers inaccessible and the timeout has saved my bacon.

so don't enable the option if you don't want it permanent!
you can always install a different vehicle fw to reset
for me this was extremely useful in developing the driver for this PR:
https://github.com/ArduPilot/ardupilot/pull/30213

libraries/AP_SerialManager/AP_SerialManager.cpp
```diff
@@ -425,7 +425,9 @@ void AP_SerialManager::init_console()
 void AP_SerialManager::init()
 {
     // always reset passthru port2 on boot
```
IamPete1:
```suggestion
    // sometimes reset passthru port2 on boot
```

tridge: thanks, fixed

PR #30213, opened 2025-06-01
AP_Scripting: LTE driver using SIM7600 AT commands

tridge:
image

provides a MAVLink connection over LTE without needing a companion computer.
Tested on a SIM7600E-H modem from AliExpress

This is ideally paired with a port on the ArduPilot support server https://support.ardupilot.org. The flight controller will stream a mavlink connection onto the support server and the user can connect a GCS as a support engineer.

### DIFF
--- a/libraries/AP_Scripting/drivers/SIM7600.lua
+++ b/libraries/AP_Scripting/drivers/SIM7600.lua
@@ -1,0 +1,367 @@
+--[[
+    driver for SIM7600 LTE modems
+--]]
+
+local MAV_SEVERITY = {EMERGENCY=0, ALERT=1, CRITICAL=2, ERROR=3, WARNING=4, NOTICE=5, INFO=6, DEBUG=7}
+
+local PARAM_TABLE_KEY = 106
+local PARAM_TABLE_PREFIX = "LTE_"
+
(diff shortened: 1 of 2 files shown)


libraries/AP_Scripting/drivers/SIM7600.lua
```diff
@@ -0,0 +1,367 @@
...
+    Function to read from the UART and log the output
+    This function reads up to 512 bytes at a time and writes it to the log file
+    returns the string read or nil
+--]]
+local function uart_read()
+    local s = uart:readstring(512)
+    if s and #s > 0 and log_file then
+        log_file:write('[' .. s .. ']\n')
+        log_file:flush()
+    end
```
tpwrules, line 155: Should this be on by default? A bit concerned about overhead.

tridge: it only logs during the connection phase, not once connected, and is very useful for support
the file ends up quite small

tpwrules: Oh I see, that's fine then. It's not immediately obvious, perhaps a cleanup for the future.

tpwrules (approved):

PR #30214, opened 2025-06-01
AP_HAL_ChibiOS: add new hardware uav-dev-auav-g4

mirkix:
Add new hardware UAV-DEV GmbH baro and airspeed sensor `uav-dev-auav-g4`

image

- 34 mm x 34 mm
- STM32G474 512 Kbytes flash
- Redundant CAN FD
- AUAV baro and airspeed sensor


Hwurzburg:
@mirkix do you have link to this device for the wiki?....the UAV-DEV website does not show it

PR #30218, opened 2025-06-02
waf: make ROMFS/scripts/*.lua work on all boards

tridge:
this was only for SITL previously

### DIFF
--- a/Tools/ardupilotwaf/boards.py
+++ b/Tools/ardupilotwaf/boards.py
@@ -81,6 +81,12 @@ class Board:
                 AP_SCRIPTING_ENABLED = 0,
             )
 
+        # embed any scripts from ROMFS/scripts
+        if os.path.exists('ROMFS/scripts'):
+            for f in os.listdir('ROMFS/scripts'):
+                if fnmatch.fnmatch(f, "*.lua"):
+                    env.ROMFS_FILES += [('scripts/'+f,'ROMFS/scripts/'+f)]
+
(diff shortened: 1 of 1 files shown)


peterbarker:
What happens in the case of a conflict between scripts in `ROMFS` in the root directory of the checkout and those in the hwdef directory for the board?

tridge:
> What happens in the case of a conflict between scripts in `ROMFS` in the root directory of the checkout and those in the hwdef directory for the board?

if it is the same script name then I think you'll end up with the hwdef one, but I haven't tested. We could throw an error?

peterbarker:
> > What happens in the case of a conflict between scripts in `ROMFS` in the root directory of the checkout and those in the hwdef directory for the board?
> 
> if it is the same script name then I think you'll end up with the hwdef one, but I haven't tested. We could throw an error?

I'd prefer the hwdef to simply take precedence, I think.  But I'm really not fussed.  We just do kind of need to know :-)

tridge:
> I'd prefer the hwdef to simply take precedence, I think. But I'm really not fussed. We just do kind of need to know :-)

I've tested and it ends up embedding both: 
```
Embedding file scripts/land_hagl.lua:/data/APM/libraries/AP_HAL_ChibiOS/hwdef/CubeOrange/scripts/land_hagl.lua
Embedding file scripts/land_hagl.lua:ROMFS/scripts/land_hagl.lua
```

I've added a patch that throws an error if we have a duplicate

PR #30219, opened 2025-06-02
AP_DroneCAN: servo/esc don't output zero pwm

IamPete1:
Currently DroneCAN will try and output 0 PWM as a valid output. Our onboard outputs treat 0 as no pulses. Over DroneCAN for servos this will result in them going to min (unless direct PWM output option is set). For non reversing ESCs it will set to 0 output, for reversible ESCs this will result in full reverse.

### DIFF
--- a/libraries/AP_DroneCAN/AP_DroneCAN.cpp
+++ b/libraries/AP_DroneCAN/AP_DroneCAN.cpp
@@ -930,17 +930,23 @@ void AP_DroneCAN::SRV_push_servos()
 {
     WITH_SEMAPHORE(SRV_sem);
 
+    uint32_t non_zero_channels = 0;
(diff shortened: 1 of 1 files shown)


libraries/AP_DroneCAN/AP_DroneCAN.cpp
```diff
@@ -930,17 +930,23 @@ void AP_DroneCAN::SRV_push_servos()
 {
     WITH_SEMAPHORE(SRV_sem);
 
+    uint32_t none_zero_channels = 0;
```
nexton-winjeel:
```suggestion
    uint32_t non_zero_channels = 0;
```

IamPete1: fixed

tridge (approved):

rmackay9:

PR #30220, opened 2025-06-03
DroneCAN: opt to ignore ESC_OF on receive

robertlong13:
This is useful if you want ESC numbering in logs/MAVLink to be independent of servo numbering. For example, if your ESCs are on servos 17-24, but you want their telemetry to be reported to the GCS as ESCs 1-8, this allows you to do that. Tested (a 4.5 backport of this) on my OctoQuad VTOL.

If you are using two different busses, with different ESCs on each bus, and have your ESC_OF value set differently on each bus, this feature will result in clashes. Don't do that.

The more versitile way to handle this would be to add a separate receive offset for each bus (which defaults to -1, causing it to just match the transmit offset), but I abandoned this for two reasons:
1. Extra params most people don't need
2. Naming those new params is borderline impossible; I only get two characters left over to name this: "ESC_??"

### DIFF
--- a/libraries/AP_DroneCAN/AP_DroneCAN.cpp
+++ b/libraries/AP_DroneCAN/AP_DroneCAN.cpp
@@ -139,6 +139,7 @@ const AP_Param::GroupInfo AP_DroneCAN::var_info[] = {
     // @Bitmask: 7: Use HobbyWing ESC protocol
     // @Bitmask: 8: Enable DroneCAN statistics
     // @Bitmask: 9: Enable DroneCAN FlexDebug messages
+    // @Bitmask: 10: Ignore ESC offset on receive
     // @User: Advanced
     AP_GROUPINFO("OPTION", 5, AP_DroneCAN, _options, 0),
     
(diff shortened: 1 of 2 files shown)


andyp1per (approved):

IamPete1:
Seems a bit odd just to get the numbers to line up in log review. How about looking up the matching servo function in mission planner and displaying that as extra information like we do for RCOU and RCIN (https://github.com/ArduPilot/MissionPlanner/pull/2735).

robertlong13:
> Seems a bit odd just to get the numbers to line up in log review. How about looking up the matching servo function in mission planner and displaying that as extra information like we do for RCOU and RCIN ([ArduPilot/MissionPlanner#2735](https://github.com/ArduPilot/MissionPlanner/pull/2735)).

I'm not actually _that_ worried about the log review, it's mostly for lua bindings and the MAVLink telemetry for the Mission Planner status tab (right now, I can't even display ESCs above 16 without adding more variables in CurrentState.cs) and warning manager stuff. There are ways to handle it there, but they're a bit of a mess.

IamPete1:
I just think this will lead to more confusion. At least the current method its quite easy to lookup which ESC is which function by looking at the servo outputs. This means that you have to also go look at the DroneCAN options. This also allows you to clash with other ESC telemetry types, either from a second DroneCAN bus or via a different protocol.

> lua bindings

I think the script writer can work out what is going on relatively easily, You could add a little helper function to go from motor number to servo number for example.

> Mission Planner

We have a parameter for that already `ESC_TLM_MAV_OFS`, you would need to set it the same as `CAN_D1_UC_ESC_OF` to get the same behavior of the bit added here.

robertlong13:
> I think the script writer can work out what is going on relatively easily, You could add a little helper function to go from motor number to servo number for example.

The script writer needs to make sure to handle the fact that the [lua binding for that randomly spits out nils](https://github.com/ArduPilot/ardupilot/pull/29027). But yeah, I do take your point.

> We have a parameter for that already ESC_TLM_MAV_OFS

Well damn, that does exactly what I need there. I'm sure I must have seen that before, but I fully forgot about it. I think that's enough to convince me to close this.

PR #30223, opened 2025-06-03
Add revision of sensor set on Pixhawk6X

jyhminwang:
Since Pixhawk6X IMU3, ICM-20649, is EOL
Our Pixhawk6X board replaced it with another supported IMU, ICM-42688-P

Added FMUV6_BOARD_GODWIT_GA1 = 44 to distinguish this revision
And name this board type Pixhawk6X-GA1

Product can be found in https://www.accton-iot.com/godwit/index.html


andyp1per:
@jyhminwang I don't think we need need a new hardware def for this. Typically for new sensors we use the old hwdef and rely on the probing to find the right sensors. There is also support built-in specifically into the Pixhawk6X to allow this easily. LMK if you need help with this approach.

jyhminwang:
> @jyhminwang I don't think we need need a new hardware def for this. Typically for new sensors we use the old hwdef and rely on the probing to find the right sensors. There is also support built-in specifically into the Pixhawk6X to allow this easily. LMK if you need help with this approach.

Hi @andyp1per 
Yes, please. Appreciated and my pleasure. My only purpose is field customers can update official Ardupilot on this new board easily

PR #30225, opened 2025-06-03
AP_Baro: correct checked register allocation count in SPLO6

peterbarker:
Not enough checked registers for reg 0x09 on device 0x7601


peterbarker:
image

rmackay9:

PR #30226, opened 2025-06-03
AP_Serial: Provide bindings for parity and stop bits

ajboainain:
Adds Lua scripting support to configure UART parity and stop bits through the AP_Scripting_SerialAccess interface.

Changes
Added configure_parity(uint8_t) and set_stop_bits(uint8_t) methods to AP_Scripting_SerialAccess

Updated bindings.desc with proper type validation and range limits

Updated docs.lua for Lua API documentation

Needs further testing. If someone can test on my behalf I would be very grateful, as I am on vacation now.

### DIFF
--- a/libraries/AP_Scripting/AP_Scripting_SerialAccess.h
+++ b/libraries/AP_Scripting/AP_Scripting_SerialAccess.h
@@ -14,6 +14,9 @@ public:
 
     void begin(uint32_t baud);
 
+    void configure_parity(uint8_t parity);
+    void set_stop_bits(uint8_t stop_bits);
+
     size_t write(uint8_t c);
     size_t write(const uint8_t *buffer, size_t size);
 
--- a/libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
+++ b/libraries/AP_Scripting/AP_Scripting_SerialAccess.cpp
@@ -24,6 +24,18 @@ void AP_Scripting_SerialAccess::begin(uint32_t baud)
     }
 }
 
+void AP_Scripting_SerialAccess::configure_parity(uint8_t parity) {
+    if (!check_is_device_port()) {
+        stream->configure_parity(parity);
+    }
+}
+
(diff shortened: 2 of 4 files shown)


peterbarker:
Patches do look likely.

Should be a single commit with  title something like , `AP_Scripting: provide bindings for parity and stop bits`

@tpwrules is this likely to work as-is?

ajboainain:
> Patches do look likely.
> 
> 
> 
> Should be a single commit with  title something like , `AP_Scripting: provide bindings for parity and stop bits`
> 
> 
> 
> @tpwrules is this likely to work as-is?
> 
> 

Noted. I'll try to squash the commits and chance the message.

ajboainain:
OK. Squashed the commits into 1 commit and added correct format for commit message as requested.

libraries/AP_Scripting/AP_Scripting_SerialAccess.h
```diff
@@ -14,6 +14,9 @@ class AP_Scripting_SerialAccess {
 
     void begin(uint32_t baud);
 
+    bool configure_parity(uint8_t parity);
+    bool set_stop_bits(uint8_t stop_bits);
```
tpwrules:
```suggestion
    void configure_parity(uint8_t parity);
    void set_stop_bits(uint8_t stop_bits);
```

And remove the return statements from the implementations. The binding marks these functions as returning `void` and I can't think of a use for the return value so please completely remove it.

tpwrules (changes requested):
Looks good overall, tested on Cube Orange that at least all the parity settings work.

Please clean up the return values and then we can merge. Be sure to squash your fix into the original commit.

ajboainain:
> Looks good overall, tested on Cube Orange that at least all the parity settings work.
> 
> Please clean up the return values and then we can merge. Be sure to squash your fix into the original commit.

Noted. Changes have been made. Thank you so much for testing, @tpwrules !

tpwrules (approved):

PR #30229, opened 2025-06-03
SITL: Fix SIM_SPEEDUP -1

magicrub:
If you start a sim without setting param SIM_SPEEDUP then it will default to -1. That can't be good. Luckily, what's actually used in the SITL code is target_speedup that can be fetched via get_speedup() within the Airplane class and that set_speedup() at boot does a sanity check on the param. However, AP_Logger looks directly at the param value and sets a timeout depending on that. 

When it's negative it hangs the logio thread.

This PR does two things:
change SIM_SPEEDUP default from -1 to +1
sanity check sitl->speedup in logger. It's the only place in the code that uses it

### DIFF
--- a/libraries/AP_Logger/AP_Logger_File.cpp
+++ b/libraries/AP_Logger/AP_Logger_File.cpp
@@ -1057,7 +1057,7 @@ bool AP_Logger_File::io_thread_alive() const
     // disk.  Unfortunately these hardware devices do not obey our
     // SITL speedup options, so we allow for it here.
     SITL::SIM *sitl = AP::sitl();
-    if (sitl != nullptr) {
+    if (sitl != nullptr && sitl->speedup > 0) {
         timeout_ms *= sitl->speedup;
     }
 #endif
--- a/libraries/SITL/SITL.cpp
+++ b/libraries/SITL/SITL.cpp
@@ -228,7 +228,7 @@ const AP_Param::GroupInfo SIM::var_info[] = {
     // @Description: Runs the simulation at multiples of normal speed. Do not use if realtime physics, like RealFlight, is being used
     // @Range: 1 10
     // @User: Advanced
-    AP_GROUPINFO("SPEEDUP",       52, SIM,  speedup, -1),
+    AP_GROUPINFO("SPEEDUP",       52, SIM,  speedup, 1),
     // @Param: IMU_POS
     // @DisplayName: IMU Offsets
     // @Description: XYZ position of the IMU accelerometer relative to the body frame origin


timtuxworth:
Thanks Tom - I've found that using the default on Mac crashes SITL. I always set `--speedup  1` on the command line as a workaround, I guess it might be related to this.

libraries/AP_Logger/AP_Logger_File.cpp
```diff
@@ -1057,7 +1057,7 @@ bool AP_Logger_File::io_thread_alive() const
     // disk.  Unfortunately these hardware devices do not obey our
     // SITL speedup options, so we allow for it here.
     SITL::SIM *sitl = AP::sitl();
-    if (sitl != nullptr) {
+    if (sitl != nullptr && sitl->speedup > 0) {
```
peterbarker, line 1060: Shouldn't we be using `get_speedup()` here?

Tim pointed out that he uses speedup on the command-line; if he does that then the number used here would be incorrect?

libraries/SITL/SITL.cpp
```diff
@@ -228,7 +228,7 @@ const AP_Param::GroupInfo SIM::var_info[] = {
     // @Description: Runs the simulation at multiples of normal speed. Do not use if realtime physics, like RealFlight, is being used
     // @Range: 1 10
     // @User: Advanced    
-    AP_GROUPINFO("SPEEDUP",       52, SIM,  speedup, -1),
+    AP_GROUPINFO("SPEEDUP",       52, SIM,  speedup, 1),
```
peterbarker, line 231: This -1 is used elsewhere in the code as a flag value, eg.
```
    if (!is_equal(last_speedup, float(sitl->speedup)) && sitl->speedup > 0) {
```

maybe better to just use the method in the logger library and leave this thing sleep...
```suggestion
    AP_GROUPINFO("SPEEDUP",       52, SIM,  speedup, -1),
```

libraries/AP_Logger/AP_Logger_File.cpp
```diff
@@ -1057,7 +1057,7 @@ bool AP_Logger_File::io_thread_alive() const
...
         timeout_ms *= sitl->speedup;
```
peterbarker, line 1061:
```suggestion
    if (sitl != nullptr) {
        timeout_ms *= sitl->get_speedup();
```

peterbarker (changes requested):
Great find!  That MacOSX thing has been a pain since forever.

OTOH, I don't know why a really long timeout_ms here would necessarily kill MacOSX in the way that it does...

libraries/AP_Logger/AP_Logger_File.cpp
magicrub: Yes, using get_speedup() would be better but I didn't see any easy way for AP_Logger to reach down into a SIM::AIRCRAFT class.

libraries/SITL/SITL.cpp
magicrub: yeah.. I dunno. I saw that code there, it's literally the only other place looking at the param. Seems convoluted for a run-once scheme. I think using a negative number is just bad practice here. For example, this -1 value would only here here on the first run after a --wipe but not the second run.. so what is it accomplishing?

libraries/AP_Logger/AP_Logger_File.cpp
magicrub: doesn't compile.

peterbarker:
OK, correct fix is probably to move all the time handling code (`sync_frame_time`) out of SIM_Aircraft.

We should probably merge this this-century, however.

@timtuxworth are you able to test this PR?

magicrub:
I talked to @peterbarker about this and we both agree the better solution is to go through get_speedup() which I tried to use but we can't reqch that from our normal libraries so it requires larger infrastructure changes that start to become out of scope so this is good as-is for now until we get those larger infrastructure changes at a later time. Just waiting on a test confirmation from @timtuxworth to merge.

timtuxworth:
Might take me a couple of days, I'm on the road, but I'll test it as soon as I get back to Edmonton likely Monday/Tuesday.

timtuxworth:
This works. Tested in SITL on Mac M1 Pro. Without this fix SITL will crash unless you pass --speedup 1 on the command line (which sim_vehicle.py does by default, which is a workaround to this problem). 

Now running a SITL instance of ardupilot on a Mac doesn't crash.

peterbarker (approved):

peterbarker:
Merged, thanks!

magicrub:
@timtuxworth Thanks for testing!

PR #25249, opened 2023-10-13
Copter, AP_Mission: Send progress for individual mission items

nexton-winjeel:
A proposed change in MAVLink adds mission item progress to the `MISSION_CURRENT` message: https://github.com/mavlink/mavlink/pull/2029

This PR:
* adds the framework in `AP_Mission` for populating and transmitting the progress fields of `MISSION_CURRENT`; and
* updates Copter to send this information for a variety of mission items.

Associated QGroundControl PR (with images): https://github.com/mavlink/qgroundcontrol/pull/10828

**NOTE:** This PR won't build as-is, because it doesn't include the changes to the `MISSION_CURRENT` message definition.

PR #30231, opened 2025-06-04
Copter: announce number of completed circles in NAV_LOITER_TURNS

peterbarker:
image

Tested in SITL.

### DIFF
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -1721,6 +1721,8 @@ void ModeAuto::do_circle(const AP_Mission::Mission_Command& cmd)
 
     // move to edge of circle (verify_circle) will ensure we begin circling once we reach the edge
     circle_movetoedge_start(circle_center, circle_radius_m, circle_direction_ccw);
+
+    circle_last_num_complete = -1;
 }
 
 // do_loiter_time - initiate loitering at a point for a given time period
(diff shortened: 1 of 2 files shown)


ArduCopter/mode_auto.cpp
```diff
@@ -2266,8 +2268,14 @@ bool ModeAuto::verify_circle(const AP_Mission::Mission_Command& cmd)
...
 
+    const auto num_circles_completed = fabsf(copter.circle_nav->get_angle_total_rad()/float(M_2PI));
+    if (int(num_circles_completed) != int(circle_last_num_complete)) {
+        circle_last_num_complete = num_circles_completed;
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "Mission: %u Circle %u/%u", cmd.index, unsigned(num_circles_completed), unsigned(turns));
+    }
+
     // check if we have completed circling
-    return fabsf(copter.circle_nav->get_angle_total_rad()/float(M_2PI)) >= turns;
+    return num_circles_completed >= turns;
```
amilcarlucas, line 2278: Won't this change prevent stuff like 2.3 turns ?

peterbarker: How so?  I've simply created an intermediate variable to hold that first value here....

peterbarker: ... and yes, I did consider having the announcement in degrees and/or fractional turns :-)

amilcarlucas: Sorry, my bad. I see now that this is NFC.

peterbarker:
I'm wondering whether the output is correct for this one.  We announce how many we've completed, like we do for jumps.  But perhaps  this should be like waypoints instead where we announce which circle you're starting?

So it would be:
```
AP: Mission: 2 Starting circle 1/5
```

rmackay9:
thanks for this.  I think that 2 floating out at the front should perhaps have something before it so that it's clear what it is (it's the mission command index I think)

peterbarker:
> thanks for this. I think that 2 floating out at the front should perhaps have something before it so that it's clear what it is (it's the mission command index I think)

It is - it lines up with the waypoint announcement which it follows.  I'd be just as happy to remove it!

I would like feedback on whether we should do the "starting circle 2/5" thing rather than the current output which, when it says, "3/5" means you have just finished 3/5 circles.  I think I really do prefer the first option.

rmackay9:
Hi @peterbarker,

you're probably right that we should specify when we've started rather than finish.  In any case, I've approved

peterbarker:
Adjusted this to announce which circle it is starting:
image

nexton-winjeel:
Relates to #25249 and friends.

PR #30232, opened 2025-06-04
27033 gsof add eahrs driver

### LINKED ISSUE #27033: AP_GPS_GSOF: Initial EAHRS support
## Feature request

**Is your feature request related to a problem? Please describe.**

PX1 is a GNSS-INS, but currently is only treated as a GPS in ArduPilot.

**Describe the solution you'd like**

* Add an EAHRS driver for the PX-1 that supports ethernet

**Describe alternatives you've considered**
* n/a

**Platform**
[x] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Additional context**

* Factor out the parser logic from `AP_GPS/AP_GPS_GSOF.h` into a new `AP_GSOF/GSOF.h` (NFC)
* Add wiki entry how to configure ethernet transport for GSOF
* Create a new EAHRS driver for the PX-1 that enables GSOF 49/50 packets
* Add parsers for GSOF 49/50
* Populate EAHRS position/velocity data

Ryanf55:
# Purpose

Add support for Trimble  PX-1 GNSS-RTX-INS External AHRS.

# Testing Performed

I flew this on a copter with Ublox on GPS1, PX-1 as GPS2 and verified all EAHRS and GPS fields are in alignment in the dataflash logs.
I have not flown off the EAHRS solution yet, but plan to after the code has been reviewed and we verify replay works.

# Ticket

Closes #27033


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -664,7 +670,7 @@ AP_GPS_Backend *AP_GPS::_detect_instance(const uint8_t instance)
 #endif
 #if AP_EXTERNAL_AHRS_ENABLED
     case GPS_TYPE_EXTERNAL_AHRS:
-        if (AP::externalAHRS().get_port(AP_ExternalAHRS::AvailableSensor::GPS) >= 0) {
+        if (AP::externalAHRS().has_sensor(AP_ExternalAHRS::AvailableSensor::GPS)) {
```
Ryanf55: We might not want this change anymore.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+        ins_full_nav.latitude * 1E7,
+        ins_full_nav.longitude * 1E7,
+        (ins_full_nav.altitude - undulation) * 1E2,
+        Location::AltFrame::ABSOLUTE);
+    state.have_location = true;
+
+    state.quat.from_euler(
+        radians(ins_full_nav.roll_deg), 
+        radians(ins_full_nav.pitch_deg),
+        radians(wrap_360(ins_full_nav.heading_deg)));
```
Ryanf55: Well this is annoying, even with wrap_360, you still get this result:
image

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+                        gps_data.fix_type = AP_GPS_FixType::PPP;
+                        break;
+                    case AP_GSOF::GnssStatus::FIXED_RTK_MODE:
+                        gps_data.fix_type = AP_GPS_FixType::RTK_FIXED;
+                        break;
+                    case AP_GSOF::GnssStatus::FLOAT_RTK_MODE:
+                        gps_data.fix_type = AP_GPS_FixType::RTK_FLOAT;
+                        break;
+                    case AP_GSOF::GnssStatus::DR_MODE:
+                        gps_data.fix_type = AP_GPS_FixType::NONE;
```
Ryanf55: Note- if we are dead reckoning, we could set nav_filter_status.dead_reckoning to true. I need to understand how important all these flags are.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+                    // To reproduce this condition, disconnect the GPS antenna and reboot the receiver.
+                    post_filter();
+                }
+            }
+            if (parsed.get(AP_GSOF::INS_RMS)) {
+                last_ins_rms_ms = now;
+
+                gps_data.horizontal_pos_accuracy = Vector2d(ins_rms.pos_rms_n, ins_rms.pos_rms_e).length();
+                gps_data.vertical_pos_accuracy = ins_rms.pos_rms_d;
+                gps_data.horizontal_vel_accuracy = Vector2d(ins_rms.vel_rms_n, ins_rms.vel_rms_e).length();
```
Ryanf55: Despite populating these values, GPA[1].HAcc, GPA[1].SAcc, and GPA[1].VAcc are zero.
I don't see why because `AP_GPS_ExternalAHRS::handle_external` copies the data over.

Ryanf55: Resolved.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+// On most hardware, you must enable EAHRS:
+//     ./waf configure --board Pixhawk6X --enable-AHRS_EXT
+
+// GPS ride-along with EARHS as the 2nd GPS
+//      param set GPS_AUTO_SWITCH 0
+//      param set GPS2_TYPE 21
+//      param set GPS_PRIMARY 0
+
+#define ALLOW_DOUBLE_MATH_FUNCTIONS
```
Ryanf55: Somehow this causes some builds to fail even though the same pattern is used in other code.

libraries/AP_DroneCAN/AP_DroneCAN.cpp
```diff
@@ -1136,6 +1136,16 @@ void AP_DroneCAN::gnss_send_fix()
         pkt.mode = UAVCAN_EQUIPMENT_GNSS_FIX2_MODE_RTK;
         pkt.sub_mode = UAVCAN_EQUIPMENT_GNSS_FIX2_SUB_MODE_RTK_FIXED;
         break;
+    case AP_GPS::GPS_Status::GPS_OK_FIX_TYPE_STATIC:
+        pkt.status = UAVCAN_EQUIPMENT_GNSS_FIX2_STATUS_3D_FIX;
```
tridge: looks identical? can be folded together

libraries/AP_GPS/AP_GPS.h
```diff
@@ -154,6 +154,8 @@ class AP_GPS
         GPS_OK_FIX_3D_DGPS = 4,      ///< Receiving valid messages and 3D lock with differential improvements
         GPS_OK_FIX_3D_RTK_FLOAT = 5, ///< Receiving valid messages and 3D RTK Float
         GPS_OK_FIX_3D_RTK_FIXED = 6, ///< Receiving valid messages and 3D RTK Fixed
+        GPS_OK_FIX_TYPE_STATIC = 7,     ///< Receiving valid messages and static fixed, typically used for base stations
```
tridge, line 154: can you check what MissionPlanner does with these? it shows strings

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
tridge: can remove that wrap_360()
but we should normalise the others to match the ATT message

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    if (!hal.scheduler->thread_create(FUNCTOR_BIND_MEMBER(&AP_ExternalAHRS_GSOF::update_thread, void), "AHRS", 2048, AP_HAL::Scheduler::PRIORITY_SPI, 0)) {
+        AP_BoardConfig::allocation_error("GSOF ExternalAHRS failed to allocate ExternalAHRS update thread");
+    }
+
+    // Offer GPS even through it's a tightly coupled EKF.
+    set_default_sensors(uint16_t(AP_ExternalAHRS::AvailableSensor::GPS));
+
+    hal.scheduler->delay(5000);
```
tridge: what thread are we in? main thread? do we really need 5s here?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+            AP_GSOF::MsgTypes expected_gps;
+            expected_gps.set(AP_GSOF::POS_TIME);
+            expected_gps.set(AP_GSOF::INS_FULL_NAV);
+            expected_gps.set(AP_GSOF::INS_RMS);
+            expected_gps.set(AP_GSOF::LLH_MSL);
+            if (AP::gps().get_first_external_instance(instance) && parsed == expected_gps) {
+                AP::gps().handle_external(gps_data, instance);
+            }
+        }
+        hal.scheduler->delay_microseconds(100);
```
tridge: the delay should be at the top of the loop

libraries/AP_GPS/AP_GPS.h
Ryanf55: For 7 and 8, it displays the integer. Neither ArduPilot nor Mission Planner conform to the mavlink spec yet, so I will submit a patch to fix that.
image

peterbarker:
There are still comments on here that haven't been addressed, and it would be nice if more things in the CI suite passed :-)

libraries/AP_ExternalAHRS/AP_ExternalAHRS.h
```diff
@@ -79,6 +82,7 @@ class AP_ExternalAHRS {
     const char* get_name() const;
 
     enum class AvailableSensor {
+        NONE = 0, // No sensors available
```
peterbarker:
```suggestion
```

do not want these in bitmasks, it confuses absolute-value parameters with bitmasks.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+//      param set GPS_PRIMARY 0
+
+#define AP_MATH_ALLOW_DOUBLE_FUNCTIONS 1
+
+#include "AP_ExternalAHRS_config.h"
+
+#if AP_EXTERNAL_AHRS_GSOF_ENABLED
+
+#include "AP_ExternalAHRS_GSOF.h"
+#include "AP_Compass/AP_Compass_config.h"
```
peterbarker: Don't *really* need this one unless you're going to exclude the inclusion of the main compass header based on AP_COMPASS_ENABLED

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+#include <AP_HAL/utility/Socket.h>
+
+static const char* LOG_FMT = "%s ExternalAHRS: %s";
+
+extern const AP_HAL::HAL &hal;
+
+AP_ExternalAHRS_GSOF::AP_ExternalAHRS_GSOF(AP_ExternalAHRS *_frontend,
+        AP_ExternalAHRS::state_t &_state): AP_ExternalAHRS_backend(_frontend, _state)
+{
+
```
peterbarker:
```suggestion
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+    auto &sm = AP::serialmanager();
+    uart = sm.find_serial(AP_SerialManager::SerialProtocol_AHRS, 0);
+    if (!uart) {
```
peterbarker:
```suggestion
    if (uart == nullptr) {
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+        GCS_SEND_TEXT(MAV_SEVERITY_ERROR, LOG_FMT, get_name(), "no UART");
+        return;
+    }
+    baudrate = sm.find_baudrate(AP_SerialManager::SerialProtocol_AHRS, 0);
+    port_num = sm.find_portnum(AP_SerialManager::SerialProtocol_AHRS, 0);
+    pktbuf = NEW_NOTHROW uint8_t[AP_GSOF::MAX_PACKET_SIZE];
+
+    if (!pktbuf) {
```
peterbarker:
```suggestion
    if (pktbuf == nullptr) {
```

We're supposed to use type-correct stuff in new code

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    auto &sm = AP::serialmanager();
+    uart = sm.find_serial(AP_SerialManager::SerialProtocol_AHRS, 0);
+    if (!uart) {
+        GCS_SEND_TEXT(MAV_SEVERITY_ERROR, LOG_FMT, get_name(), "no UART");
+        return;
+    }
+    baudrate = sm.find_baudrate(AP_SerialManager::SerialProtocol_AHRS, 0);
+    port_num = sm.find_portnum(AP_SerialManager::SerialProtocol_AHRS, 0);
+    pktbuf = NEW_NOTHROW uint8_t[AP_GSOF::MAX_PACKET_SIZE];
```
peterbarker: Why are we doing it this way rather than just as a member?  The backend itself is `NEW_NOTHROW`d, so we don't gain anything doing another allocation here - except another possible runtime failure?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    if (!hal.scheduler->thread_create(FUNCTOR_BIND_MEMBER(&AP_ExternalAHRS_GSOF::update_thread, void), "AHRS", 2048, AP_HAL::Scheduler::PRIORITY_SPI, 0)) {
+        AP_BoardConfig::allocation_error("GSOF ExternalAHRS failed to allocate ExternalAHRS update thread");
+    }
+
+    // Offer GPS even through it's a tightly coupled EKF.
+    set_default_sensors(uint16_t(AP_ExternalAHRS::AvailableSensor::GPS));
+
+    hal.scheduler->delay(5000);
```
peterbarker: We do *not* want a 5 second delay in our boot process while you're working out whether the AHRS is talking to us.

Can we get away with a member variable indicating we've warned and some code in `update_thread`?

... we might *not* be able to get away with it, but if we can we should.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+    if (!initialised()) {
+        GCS_SEND_TEXT(MAV_SEVERITY_WARNING, LOG_FMT, get_name(), "missing data within 5s.");
+    }
+}
+
+// get serial port number for the uart
+int8_t AP_ExternalAHRS_GSOF::get_port(void) const
+{
+    if (!uart) {
```
peterbarker:
```suggestion
    if (uart == nullptr) {
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+    }
+    return port_num;
+};
+
+void AP_ExternalAHRS_GSOF::update_thread(void)
+{
+    // TODO configure receiver to output expected data.
+
+    auto last_debug = AP_HAL::millis();
+    size_t pps = 0;
```
peterbarker:
```suggestion
    uint32_t pps = 0;
```

.... 'though this also looks like code which will go away anyway

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    uart->begin(baudrate);
+    
+    while (true) {
+        // TODO should we ever call begin() again?
```
peterbarker: Possibly if you start magically configuring baud rates...

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+        while (uart->available() > 0) {
```
peterbarker: You are going to end up tight-looping in this thread at a ridiculously high priority while data isn't available on the UART.

Probably need a short sleep if there's no data to be read, meaning a change to this:

```suggestion
        const auto available = uart->available();
        if (available == 0) {
           .... sleep
            continue;
        }
        for (auto i=0; i<available; i++) {
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+            const uint8_t c = uart->read();
```
peterbarker:
```suggestion
            uint8_t c;
            if (!uart->read(c)) {
                break
            }
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+            }
+
+            if (parsed.get(AP_GSOF::INS_FULL_NAV)) {
+                last_ins_full_nav_ms = now;
+
+                gps_data.gps_week = ins_full_nav.gps_week;
+                gps_data.ms_tow = ins_full_nav.gps_time_ms;
+                gps_data.ned_vel_north = ins_full_nav.vel_n;
+                gps_data.ned_vel_east = ins_full_nav.vel_e;
+                gps_data.ned_vel_down = ins_full_nav.vel_d;
```
peterbarker: This doesn't look right - this isn't gps data, this is a position estimate based on sensor fusion!

So this data is what you should be returning as an AHRS backend, and `LLH_MSL` presumably is raw sensor data from the GPS attached to the Trimble so *should* be used to fill gps_data


... and if you want to argue the point that should be in the form of comments, because this is decidedly odd :-)

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+                gps_data.longitude = static_cast<int32_t>(llh_msl.longitude * 1E7);
+                gps_data.latitude = static_cast<int32_t>(llh_msl.latitude * 1E7);
+                gps_data.msl_altitude = static_cast<int32_t>(llh_msl.altitude_msl * 1E2);
+            }
+
+            if (parsed.get(AP_GSOF::LLH_MSL) && parsed.get(AP_GSOF::INS_FULL_NAV)) {
+                undulation = ins_full_nav.altitude - llh_msl.altitude_msl;
+            }
+
+            uint8_t instance;
```
peterbarker: Move to just above where it is filled in

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
peterbarker: Why was this comment resolved without action taken?

That action may have taken the form of a comment if tridge is wrong!

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+            }
+        }
+        hal.scheduler->delay_microseconds(100);
+        check_initialise_state();
+    }
+}
+
+void AP_ExternalAHRS_GSOF::check_initialise_state(void)
+{
+    const auto new_init_state = initialised();
```
peterbarker:
```suggestion
    const bool new_init_state = initialised();
```

You can take the use of `auto` too far :-)

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    return pos_time_healthy && ins_full_nav_healthy && ins_rms_healthy && llh_msl_healthy;
+}
+
+bool AP_ExternalAHRS_GSOF::filter_healthy() const
+{
+    // TODO get the right threshold from Trimble for arming vs in flight.
+    // Fow now, assume an aligned IMU is sufficient for flight.
+    auto const imu_alignment_healthy = (
+        ins_rms.imu_alignment_status == ImuAlignmentStatus::DEGRADED ||
```
peterbarker, line 371: degraded but not-dead-yet?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.h
```diff
@@ -0,0 +1,99 @@
...
+
+
+#pragma once
+
+#include "AP_ExternalAHRS_config.h"
+
+#if AP_EXTERNAL_AHRS_GSOF_ENABLED
+
+#include "AP_ExternalAHRS_backend.h"
+#include <AP_GPS/AP_GPS.h>
```
peterbarker:
```suggestion
```

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -113,12 +112,22 @@ AP_GSOF::process_message(MsgTypes& parsed_msgs)
                 parse_vel(a);
                 break;
             case DOP:
-                parse_dop(a);
+                parse_dop(a);
```
peterbarker: whitespace

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,254 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+    log_llh_msl();
+#endif // HAL_LOGGING_ENABLED
+}
+
+#if HAL_LOGGING_ENABLED
+void AP_GSOF::log_pos_time() const
+{
+
+    // @LoggerMessage: GSPT
+    // @Description: GPS Position Time Metadata from GSOF1
```
peterbarker:
```suggestion
    // @Description: GSOF GPS Position and Time Metadata
```

Put the cool stuff up the front

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,254 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+    // @Field: TimeUS: Time since system startup
+    // @Field: TOWms: Time of week in milliseconds
```
peterbarker:
```suggestion
    // @Field: TOWms: GPS Time of week in milliseconds
```

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,254 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+    // @Field: Week: GPS week number
+    // @Field: Sats: Number of satellites
+    // @Field: Flags1: Positioning flags byte 1
+    // @Field: Flags2: Positioning flags byte 2
```
peterbarker, line 277: You can get all sorts of useful stuff out of tools if you provide bitmasks for these fields

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,254 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+    // @Field: InitNum: Initialization count
```
peterbarker, line 278: Huh?

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,254 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+    AP::logger().WriteStreaming(
+        "GSPT",                                         // Message name
+        "TimeUS,TOWms,Week,Sats,Flags1,Flags2,InitNum", // Field names
+        "Q"    "I"   "H"  "B"  "B"    "B"     "B",      // Field types: Q=uint64_t, I=uint32_t, H=uint16_t, B=uint8_t
```
peterbarker: missing units and multipliers.

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,254 @@ void AP_GSOF::parse_vel(uint32_t a)
...
     // Skip pos_rms
     pos_sigma.sigma_east = be32tofloat_ptr(msg.data, a + 4);
     pos_sigma.sigma_north = be32tofloat_ptr(msg.data, a + 8);
     pos_sigma.sigma_up = be32tofloat_ptr(msg.data, a + 16);
 }
+
+void AP_GSOF::parse_ins_full_nav(uint32_t a)
+{
+    // https://receiverhelp.trimble.com/oem-gnss/gsof-messages-ins-full-nav.html
+    ins_full_nav.gps_week = be16toh_ptr(msg.data + a);
```
peterbarker, line 202: This is really not what you're supposed to be using a packed structure like `ins_full_nav` for.

Assuming msg.data contains a valid INS_FULL_NAV message, you would simply do a `memcpy` here to take a copy of the packet.  Users of the values would do their own conversion on the way out - and you would *only* do that on very-much-internal functions which are populating other structures (there will typically only be one place that does that)

... in fact, most drivers wouldn't even bother keeping around the packet at all - they would just directly extract into their internal data structures on receipt of the message (i.e. in here, after casting msg.data into ins_full_nav_t

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: The UART code with a buffer like this was copied from other parts of the codebase. I can change if you need.
Copied from AP_ExternalAHRS_VectorNav.cpp.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Ditto on the code quality. I'm all for it, let's update all the EAHRS drivers in another PR though.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: But this pattern is acceptable in AP_ExternalAHRS_MicroStrain.cpp?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Yea this is diagnostics to count how many packets we get per second. The ethernet and serial interfaces have different performance, and ethernet frequently goes unhealthy. I would hope to keep the debugging code; perhaps I should put it behind a compile option?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: I don't think so, so I'll remove the todo. Thanks for verifying.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Good point, the other approach Tridge suggested would not have solved this case.
Fixed in 0e3355a5991cf630ebe29c89e463e4224b3855d9

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Can we document the return value? UARTDriver.h does not say that 0 is a failure.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: I have added a descriptive comment in 1940467a2660c7e2e507b6560c6a4dfb23fbb725

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Both `instance` and `expected_gps` are used on the same line. I have moved it for your sanity regardless in 355deb17e8be588367d530de7070643d017639cb.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Correct. Degraded will be observed if you fly in significant multipath interface, or under trees. 
I have tested this in flight flying in an alley next to a tall building and the GNSS properly reported degraded health and AP started rejecting the GPS data.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.h
Ryanf55: Good find.

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: Fixed in b49b7887c367045cbf08ffe16ee1901f75c30fc5

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: So we can't memcpy because of the endianness. I was hoping ins_full_nav object already has the endianness conversion done so the values can be used as is. If you are proposing I instead do a memcpy on the buffer, and then pick through and convert endianness on all the fields, I can switch to do doing that, however, again, this is copying existing patterns. Do we need to change all the other AP_GSOF code in this PR?

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: Yea, I'm just being a bit lazy trying to get this merged before 4.7, and will polish it after.

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: The field documentation is copied directly from the ICD.
I think this is boot count, but ignore it. I'm just logging everythign so you don't have to send proprietary BIN logs from the PX-1 to Trimble to figure stuff out.

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: Added in 044784cd4e8a36aa696df6768375d33328fba984 along with other places that were GPS time.

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: Added in c178b74b14c0c14bf723ff69844d0d7d954fa872 but I wanted to keep GSOF1 to differentiate that from the other packets. It's now internally consistent.

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: Done in 044784cd4e8a36aa696df6768375d33328fba984

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55:
```suggestion
```

Ryanf55:
Need to fix logger metadata.
https://github.com/ArduPilot/ardupilot/actions/runs/19197230527/job/54879964205?pr=30232#step:4:93
```
  LoggerDocumentation (Test Onboard Logging Generation) (Command '['/__w/ardupilot/ardupilot/Tools/autotest/logger_metadata/parse.py', '--vehicle', 'Tracker'
  ```
  
  ```
  Unknown field (    //          0=GPS_ONLY
)
```


Which is observed in 
```
    // @LoggerMessage: GSN1
    // @Description: GSOF49 Full INS Navigation Solution 1
    // @Field: TimeUS: Time since system startup [µs]
    // @Field: GpsWeek: GPS week number since Jan 1980
    // @Field: GpsTimeMs: GPS time of week [ms]
    // @Field: ImuStat: IMU alignment status (enum)
    //          0=GPS_ONLY
    //          1=COARSE_LEVELING
    //          2=DEGRADED
    //          3=ALIGNED
    //          4=FULL_NAV
```

Bitfields are not documented in AP_Logger/README.md

Ryanf55:
> Need to fix logger metadata. https://github.com/ArduPilot/ardupilot/actions/runs/19197230527/job/54879964205?pr=30232#step:4:93
> 
> ```
>   LoggerDocumentation (Test Onboard Logging Generation) (Command '['/__w/ardupilot/ardupilot/Tools/autotest/logger_metadata/parse.py', '--vehicle', 'Tracker'
> ```
> 
> ```
> Unknown field (    //          0=GPS_ONLY
> )
> ```
> 
> Which is observed in
> 
> ```
>     // @LoggerMessage: GSN1
>     // @Description: GSOF49 Full INS Navigation Solution 1
>     // @Field: TimeUS: Time since system startup [µs]
>     // @Field: GpsWeek: GPS week number since Jan 1980
>     // @Field: GpsTimeMs: GPS time of week [ms]
>     // @Field: ImuStat: IMU alignment status (enum)
>     //          0=GPS_ONLY
>     //          1=COARSE_LEVELING
>     //          2=DEGRADED
>     //          3=ALIGNED
>     //          4=FULL_NAV
> ```
> 
> Bitfields are not documented in AP_Logger/README.md

Fixed! Since we don't use these enums yet, I just log the raw values for now.

libraries/AP_DroneCAN/AP_DroneCAN.cpp
peterbarker: @Ryanf55 please do not resolve PR comments without addressing them.

I believe in this case the only difference is a comment and that the compiler is very likely to fold the cases together.

But you should explain any reasoning you have for resolving things (including "resolved after discission with X" ('though if there was a question then a comment is probably warranted!

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
peterbarker: > The UART code with a buffer like this was copied from other parts of the codebase. I can change if you need. Copied from AP_ExternalAHRS_VectorNav.cpp.

VectorNav was the first externalahrs backend - and we probably worked up the current abstraction since then.  Please do change this.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
peterbarker: > But this pattern is acceptable in AP_ExternalAHRS_MicroStrain.cpp?

Not if we can get away without it!

libraries/AP_GSOF/AP_GSOF.cpp
peterbarker: It's a bit of a contradiction in terms you're creating here.  You're populating a PACKED structure which will never go on the wire, and named somehwat like you'd name something you'd put on the wire.

You're also probably doing unnecessary work here as I bet we're not using all of these fields!

I would suggest populating local member variables from this message as it comes off the wire *or* doing the memcpy (an assignment will probably do, actually) into the persistent packet structure and then do the conversion on the way out.  We do both of these things in ArduPilot, but the former is less prone to endianness bugs.

peterbarker:
> Can we document the return value? UARTDriver.h does not say that 0 is a failure.

0 is not failure - it really is `\0`.  `->read()` returns `int16_t` and you have to check for -1 before using the value.  Which many places to not, thus my goal of removing the damned call :-)

peterbarker:
> Bitfields are not documented in AP_Logger/README.md

The parameter documentation in general is not documented in there.

libraries/AP_DroneCAN/AP_DroneCAN.cpp
```diff
@@ -1145,6 +1145,16 @@ void AP_DroneCAN::gnss_send_fix()
...
+        pkt.mode = UAVCAN_EQUIPMENT_GNSS_FIX2_MODE_PPP; // Static is not representable in DroneCAN.
```
peterbarker: So we should add to the DroneCAN enumeration?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+    uart->begin(baudrate);
+    
+    while (true) {
+        const auto available = uart->available();
+        if (available == 0) {
+            hal.scheduler->delay_microseconds(100);
+            continue;
+        }
+        hal.scheduler->delay_microseconds(100);
```
peterbarker:
```suggestion
```

... pretty sure that would break this driver!

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,371 @@
...
+        const uint8_t c = uart->read();
```
peterbarker:
```suggestion
        uint8_t c;
        if (!uart->read(c)) {
                hal.scheduler->delay_microseconds(100);
                continue;
        }
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,371 @@
...
+        AP_GSOF::MsgTypes parsed;
+        const auto parse_res = parse(c, parsed);
+        if (parse_res != PARSED_GSOF_DATA) {
+            continue;
+        }
+
+        auto const now = AP_HAL::millis();
+        pps++;
+
+        if (now - last_debug > 1000) {
```
peterbarker, line 132:
```suggestion
#if AP_EXTERNALAHRS_GSOF_DEBUG_ENABLED
        if (now - last_debug > 1000) {
```

(you can probably get away with dropping `last_debug_ms` in here as a static to avoid clutter)

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,371 @@
...
+        if (parsed.get(AP_GSOF::POS_TIME)) {
+            last_pos_time_ms = now;
+
+            gps_data.satellites_in_view = pos_time.num_sats;
+        }
+
+        if (parsed.get(AP_GSOF::INS_FULL_NAV)) {
+            last_ins_full_nav_ms = now;
+
+            // The Trimble PX-1 is a tightly coupled GNSS-INS.
```
peterbarker, line 150: We really, really do not want to do this.  You can't feed position data which has been extrapolated using IMU sensor data into ArduPilot's EKF - because that's *exactly* what the EKF is about to do, so you will end up adding inertial measurements twice.

You can keep this if you only feed in actual position measurements, not an inertial nav solution.

It looks like you could populate this gps data from llh and 06 as raw measurements?

The nature of the beast means that it's only under dynamic flight you're likely to see problems with double-adding IMU data.

If all else fails, we could perhaps allow it but not by default and put some big warnings around its use in the Wiki...

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,371 @@
...
+                    break;
+                case AP_GSOF::GnssStatus::FLOAT_RTK_MODE:
+                    gps_data.fix_type = AP_GPS_FixType::RTK_FLOAT;
+                    break;
+                case AP_GSOF::GnssStatus::DR_MODE:
+                    gps_data.fix_type = AP_GPS_FixType::NONE;
+                    break;
+            }
+            if (ins_full_nav.gnss_status != AP_GSOF::GnssStatus::FIX_NOT_AVAILABLE) {
+                // If fix is unavailble, the lat/lon may be zero, so don't post the data.
```
peterbarker:
```suggestion
                // If fix is unavailable, the lat/lon may be zero, so don't post the data.
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,371 @@
...
+            gps_data.longitude = static_cast<int32_t>(llh_msl.longitude * 1E7);
+            gps_data.latitude = static_cast<int32_t>(llh_msl.latitude * 1E7);
+            gps_data.msl_altitude = static_cast<int32_t>(llh_msl.altitude_msl * 1E2);
+        }
+
+        if (parsed.get(AP_GSOF::LLH_MSL) && parsed.get(AP_GSOF::INS_FULL_NAV)) {
+            undulation = ins_full_nav.altitude - llh_msl.altitude_msl;
+        }
+
+        AP_GSOF::MsgTypes expected_gps;
```
peterbarker, line 210: Is this construct ignoring all of the timestamps (`last_llh_msl_ms`) before saying, "yes, we have the data"?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.h
```diff
@@ -0,0 +1,98 @@
...
+
+    AP_HAL::UARTDriver *uart;
+    int8_t port_num;
+    uint32_t baudrate;
+    uint8_t *pktbuf;
+
+    HAL_Semaphore sem;
+
+    // Used to monitor initialization state.
+    bool last_init_state = false;
```
peterbarker:
```suggestion
    bool last_init_state;
```

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.h
```diff
@@ -0,0 +1,98 @@
...
+    void post_filter() const;
+
+    AP_HAL::UARTDriver *uart;
+    int8_t port_num;
+    uint32_t baudrate;
+    uint8_t *pktbuf;
+
+    HAL_Semaphore sem;
+
+    // Used to monitor initialization state.
```
peterbarker, line 75: It's actually not; it's only used to announce whether the device has been initialised or not.

If the vehicle becomes un-iunitialised and reinitialised then you won't get another announcement AFAICS.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,371 @@
...
+{
+    const bool new_init_state = initialised();
+    // Only send the message after fully booted up, otherwise it gets dropped.
+    if (!last_init_state && new_init_state && AP_HAL::millis() > 5000) {
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, LOG_FMT, get_name(), "initialised.");
+        last_init_state = new_init_state;
+    }
+}
+
+void AP_ExternalAHRS_GSOF::post_filter() const
```
peterbarker, line 233: I don't think this separate method really buys anything except to disconnect the logic a bit.


and I'm really not sure what "post_filter" is supposed to mean, it's rather ambiguous.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
```diff
@@ -0,0 +1,371 @@
...
+    return last_pos_time_ms != 0 && last_ins_full_nav_ms != 0 && last_ins_rms_ms != 0 && last_llh_msl_ms != 0;
+}
+
+bool AP_ExternalAHRS_GSOF::pre_arm_check(char *failure_msg, uint8_t failure_msg_len) const
+{
+    if (!initialised()) {
+        hal.util->snprintf(failure_msg, failure_msg_len, LOG_FMT, get_name(), "not initialised");
+        return false;
+    }
+    if (!times_healthy()) {
```
peterbarker, line 288:
```suggestion
    if (data_is_stale()) {
```

libraries/AP_GSOF/AP_GSOF.cpp
peterbarker: ... having looked at this, we're making the situation worse in here in terms of mixing up on-the-wire structures and in-memory structures, and my suggestions above would make the code rather better (i.e. just `memcpy` and convert on the way out of the structure).

However, I can't insist on that in this PR - you're good to leave this as-is

libraries/AP_GSOF/AP_GSOF.cpp
peterbarker: I'd really rather get these documented before 4.7 (and before merge)

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,242 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+        pos_time.time_week_ms,                          // TOWms
+        pos_time.time_week,                             // Week
+        pos_time.num_sats,                              // Sats
+        pos_time.pos_flags1,                            // Flags1
+        pos_time.pos_flags2,                            // Flags2
+        pos_time.initialized_number                     // InitNum
+    );
+}
+
+void AP_GSOF::log_ins_full_nav() const
```
peterbarker, line 295: How fast do logs grow with all of this data?

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,242 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+    // @Field: VUN: North Velocity uncertainty RMS [m/s]
+    // @Field: VUE: East Velocity uncertainty RMS [m/s]
+    // @Field: VUD: Down Velocity uncertainty RMS [m/s]
+    // @Field: RollU: Roll RMS uncertainty [deg]
+    // @Field: PitchU: Pitch RMS uncertainty[deg]
+    // @Field: YawU: Yaw RMS uncertainty[deg]
+    AP::logger().WriteStreaming(
+        "GSR1",
+        "TUS,GpsTMs,PUN,PUE,PUD,VUN,VUE,VUD,RollU,PitchU,YawU",
+        "Q" "I"    "f"   "f"   "f"   "f"   "f"   "f"   "f"   "f"    "f",
```
peterbarker: missing units and multipliers

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,242 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+    // @Description: GSOF70 LLH with MSL Altitude
+    // @Field: TimeUS:    Time since system startup [µs]
+    // @Field: Lat:       Latitude [deg]
+    // @Field: Lng:       Longitude [deg]
+    // @Field: AltMSL:    Altitude above mean sea level [m]
+    // @Field: Model:     Geoid model used (null-terminated string)
+    AP::logger().WriteStreaming(
+        "GSLH",
+        "TimeUS,Lat,Lng,AltMSL,Model",
+        "Q"     "d"  "d"  "d"    "Z",
```
peterbarker: missing units and multipliers

peterbarker (changes requested):

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Removed in 3e8e282706c97d02c3104db6c7b8de86a2cfdc8a

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Removed in 3e8e282706c97d02c3104db6c7b8de86a2cfdc8a

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Added AP_EXTERNALAHRS_GSOF_DEBUG_ENABLED in ed10284663c4bf363c1fb24949289f328035a725

Will resolve once last_debug_ms is static.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: How else do we allow AP to accept a high accuracy EAHRS INS purely for position and heading fusion? Can we still use AP's EKF3 without relying on EAHRS's for the lower level data?

I can ask Trimble if [packet 60](https://receiverhelp.trimble.com/oem-gnss/gsof-messages-code-position-llh.html) is the GPS-only solution, but that only has LLH data, no accuracies, velocities and the usual suspects for a GPS message.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Yes, the stalenss of the data is handled in the `healthy()` call. Much like other drivers ignore staleness when parsing for content. We don't need to duplicate staleness checking here. For this statement below to be true, the data has to be sent in a single block (same packet) otherwise the parser resets. I've tested dropping rates of some packets in half and no GPS data is published into AP.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: This is from MicroStrain7. We can rename both if you have a better name. 
Would you rather have an 869 line long function called check_uart like InertialLabs that does everything?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Should we also change filter_healthy to filter_unhealthy and `healthy` to `unhealthy` to preserve parallel logic?

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.h
Ryanf55: Is that a hard requirement to implement un-initialization recovery? None of the other drivers handle that.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
peterbarker: > How else do we allow AP to accept a high accuracy EAHRS INS purely for position and heading fusion? Can we still use AP's EKF3 without relying on EAHRS's for the lower level data?

The sensor data you supply to the EKF should be straight-up sensor data (as corrected by RTCM or whatever).

The "high accuracy EAHRS INS" is why you're an AHRS backend!  The inertial-sensor-fused position from the external device is presented by the ExternalAHRS data as a canonical vehicle position - meaning we don't cecessarily know how that was really arrived at.... could be GPS, could be vision, at the AHRS level we shouldn't care (we do, kind of, but we shouldn't).

You can still use EKF3 without data from the EAHRS - but it will then need its own sensor suite (baro/GPS/compass/whatever) to derive its solution.

Feeding in the same GPS data that the ExternalAHRS is using to come up with its solution is fine - but it does need to be the same raw data that the external device is consuming, not the output of the fusion in the external device.
 
> I can ask Trimble if [packet 60](https://receiverhelp.trimble.com/oem-gnss/gsof-messages-code-position-llh.html) is the GPS-only solution, but that only has LLH data, no accuracies, velocities and the usual suspects for a GPS message.

There seemed to be packets floating around which did contain accuracies and velocities....

(apart from the mathematical suckitude which is passing in an inertial-sensor fused position to the EKF, there's also the fact that passing the raw data into the EKF gives you a second bite at coming up with a sane solution to your current position and attitude; if the ExternalAHRS goes insane you really don't want to be consuming its crap when you might be able to make do with the raw GPS data the external unit is having trouble with.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_config.h
```diff
@@ -36,5 +36,12 @@
...
 #ifndef AP_EXTERNAL_AHRS_SBG_ENABLED
 #define AP_EXTERNAL_AHRS_SBG_ENABLED AP_EXTERNAL_AHRS_BACKEND_DEFAULT_ENABLED
+#endif 
+
+#ifndef AP_EXTERNAL_AHRS_GSOF_ENABLED
+#define AP_EXTERNAL_AHRS_GSOF_ENABLED AP_EXTERNAL_AHRS_BACKEND_DEFAULT_ENABLED
 #endif
 
+#ifndef AP_EXTERNALAHRS_GSOF_DEBUG_ENABLED
+#define AP_EXTERNALAHRS_GSOF_DEBUG_ENABLED 1
```
tridge: do we really want this enabled in production use? it seems quite verbose

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
tridge: there are other downsides too - a major one being that this will defeat the dead-reckoning capabilities of ArduPilot. For example on fixed wing we can use side-slip fusion to get O(t) divergence. I don't see the GSOF driver checking the fly-forward flag, so it can't be doing side-slip fusion correctly, which means on loss of GPS it will diverge O(t^2). The same goes for the drag fusion on copters.

tridge (changes requested):

tridge:
@Ryanf55 are we using https://receiverhelp.trimble.com/oem-gnss/gsof-messages-code-position-llh.html ? that looks like the GNSS solution
also, does the receiver have a way to ask it to stop fusing GNSS? that would be what we would use to test it

libraries/AP_ExternalAHRS/AP_ExternalAHRS_config.h
Ryanf55:
```suggestion
#define AP_EXTERNALAHRS_GSOF_DEBUG_ENABLED 0
```

tridge (changes requested):
some changes needed:
 - add in get_variances() method to AHRS to allow for EKF failsafe
 - fix up the handling of gnss_status
 - provide docs on susceptibility to magnetic interference
 - we should work on fallback to EKF3, we need a way to make that decsion
 - need flags or param within the GSOF EAHRS to determine when it should declare itself unhealthy (eg. if a ublox is fitted, then user may decide that EAHRS should declare unhealthy when gnss_status is not RTK)

tridge:
some todo for trimble, none of this is required for this PR, but nice for the future:
 - add a protocol element to disable use of GNSS for testing purposes
 - add protocol packet for GNSS solution that doesn't involve IMU or mag
 - add protocol packet for mag field

libraries/AP_ExternalAHRS/AP_ExternalAHRS_config.h
Ryanf55: Applied in 37d42bca0f2f6fc6be6e971692dfc6fbd1f041b4

Ryanf55:
> some changes needed:
> 
> * add in get_variances() method to AHRS to allow for EKF failsafe

done in c1793d53405db458bb7abb9ebde6bf238e7b8c0a

> * fix up the handling of gnss_status

done in 3819f3f6f8d16351eca6e4e4f29f28f548e80e65

> * provide docs on susceptibility to magnetic interference

It's tagged as wiki needed; I've got a branch with this in it.

> * we should work on fallback to EKF3, we need a way to make that decsion

Agree, but out of scope for this PR. Looking forward to it soon.

> * need flags or param within the GSOF EAHRS to determine when it should declare itself unhealthy (eg. if a ublox is fitted, then user may decide that EAHRS should declare unhealthy when gnss_status is not RTK)

Agree, this is a nice addition, but not required for initial merge when used as the only GNSS source.


Ready for final review. I'm doing another round of flight testing on the current branch.
If it's good, I'll squash and request merge.

libraries/AP_DroneCAN/AP_DroneCAN.cpp
Ryanf55: I folded the comments together to be sure. Once DroneCAN supports the new modes, those will split apart.

libraries/AP_DroneCAN/AP_DroneCAN.cpp
Ryanf55: yes, I can figure this out in a follow up PR since it's not required for the PX-1 to work since PX-1 is not a dronecan device.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: Action was taken. This is fixed.

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: They are now documented.

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: Created a follow up issue and assigned myself.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: We discussed this in depth with Trimble. There are no solutions given the current state of AP and the PX-1, and we will accept the degraded dead reckoning performance.

I have tested failure of the GNSS in flight on a copter, and AP seamlessly switched to EKF2. 
improvements to fall back to EKF3 will help, and for best safety, the wiki will say to use a dual GPS solution for the EKF safety.

libraries/AP_DroneCAN/AP_DroneCAN.cpp
Ryanf55: See https://github.com/ArduPilot/ardupilot/issues/32165

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
Ryanf55: pktbuf was actually unused, cruft from previous code. The packet buffer is handled inside AP_GSOF. I've removed it and squashed that commit.

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: We've been running lots of flights, and I'd like to keep the logs at full rate for now. It's proving very useful for Trimble.

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.h
peterbarker: No, it's not.  Would be a good idea, of course.

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,245 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+    // @Field: Week: GPS week number
+    // @Field: Sats: Number of satellites
+    // @Field: Flags1: Positioning flags byte 1
+    // @Field: Flags2: Positioning flags byte 2
+    // @Field: InitNum: Initialization count
+    AP::logger().WriteStreaming(
+        "GSPT",                                         // Message
+        "TimeUS,TOWms,Week,Sats,Flags1,Flags2,InitNum", // Labels
+        "ss-S---",                                      // Units
+        "FC00000",                                      // Multipliers
```
peterbarker: These should be in the same format as "Field Types", each column lined up

Similarly elsehwere

Ryanf55: Fixed in 7c73b73b08cedced3bc455ec72221bbadf643c2a

Ryanf55:
> LGTM
> 
> Please follow up with adjustments to the autotest suite that adds a method like this:
> 
> ```
>     def InertialLabsEAHRS(self):
>         '''Test InertialLabs EAHRS support'''
>         self.fly_external_AHRS("ILabs", 5, "ap1.txt")
> ```
> 
> and appropriate entry to run that test in CI.
> 
> .... because currently if this breaks it's all on you for not providing a regression suite :-)

as discussed in discord, we will not be doing autotests in this PR because it will require rework of AP_GPS_GSOF tests to reuse the parser here. Next feature :)

libraries/AP_ExternalAHRS/AP_ExternalAHRS_GSOF.cpp
tridge: ok, not ideal, but as long as you note this in the wiki page (the page I'm sure you're planning to do!) then we'll accept it

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -134,26 +143,29 @@ AP_GSOF::process_message(MsgTypes& parsed_msgs)
...
     position.latitude_rad = be64todouble_ptr(msg.data, a);
     position.longitude_rad = be64todouble_ptr(msg.data, a + 8);
+    // Altitude is "Height from WGS-84 datum" -> Likely ellipsoid
     position.altitude = be64todouble_ptr(msg.data, a + 16);
 }
 
 void AP_GSOF::parse_vel(uint32_t a)
 {
-    // https://receiverhelp.trimble.com/oem-gnss/index.html#GSOFmessages_Velocity.html?TocPath=Output%2520Messages%257CGSOF%2520Messages%257C_____32
-    vel.velocity_flags = msg.data[a];
```
tridge, line 156: velocity_flags is now dead? not set anywhere but checked in two places

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -169,18 +181,245 @@ void AP_GSOF::parse_vel(uint32_t a)
...
+    // @Field: PUD: Down Position uncertainty RMS [m]
+    // @Field: VUN: North Velocity uncertainty RMS [m/s]
+    // @Field: VUE: East Velocity uncertainty RMS [m/s]
+    // @Field: VUD: Down Velocity uncertainty RMS [m/s]
+    // @Field: RollU: Roll RMS uncertainty [deg]
+    // @Field: PitchU: Pitch RMS uncertainty[deg]
+    // @Field: YawU: Yaw RMS uncertainty[deg]
+    AP::logger().WriteStreaming(
+        "GSR1",
+        "TUS,GpsTMs,PUN,PUE,PUD,VUN,VUE,VUD,RollU,PitchU,YawU",
```
tridge, line 381: number of fields doesn't match passed arguments?

libraries/AP_ExternalAHRS/AP_ExternalAHRS.cpp
```diff
@@ -120,6 +121,12 @@ void AP_ExternalAHRS::init(void)
         return;
 #endif
 
+#if AP_EXTERNAL_AHRS_GSOF_ENABLED
+    case DevType::GSOF:
+        backend = new AP_ExternalAHRS_GSOF(this, state);
```
tridge: should use NEW_NOTHROW

Ryanf55: Fixed in f2444847bde96790e99dcfb0b2e5946c22b0dfaa

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: Fixed, I removed the extra imu and gnss flags because they are logged elsewhere.

libraries/AP_GSOF/AP_GSOF.cpp
Ryanf55: Fixed! I also created a follow up issue to add velocity to the autotest.

peterbarker:
@Ryanf55 finally got this merged!  Thanks for your patience.

This will *probably* go back cleanly, but I suggest you might create a backport PR to make sure it goes into 4.7beta2 cleanly (I've added it to the 4.7 project)

rmackay9:
this is included in 4.7.0-beta2

rmackay9:
@Ryanf55,

Co-pilot came up with some comments when it reviewed this PR as part of the 4.7.0-beta2 release (https://github.com/ArduPilot/ardupilot/pull/32500).  Could you have a look?

libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -85,17 +83,18 @@ AP_GSOF::process_message(MsgTypes& parsed_msgs)
...
+#if CONFIG_HAL_BOARD == HAL_BOARD_SITL
+                AP_HAL::panic("Invalid output type.");
+#else
+                return false;
+#endif
+            }
+
             parsed_msgs.set(output_type);
             a++;
             const uint8_t output_length = msg.data[a];
```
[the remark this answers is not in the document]
Ryanf55: I wrote a test against the parser that generated corrupt packets and it's safe. AI is not correct.

PR #30235, opened 2025-06-04
hwdef: add CUAV flight controller OpenDroneID enabled build

cuav-chen2:
Add OpenDroneID enabled build for CUAV flight controllers.


Tools/AP_Bootloader/board_types.txt
```diff
@@ -487,6 +487,11 @@ AP_HW_AEROFOX_H7                     7110
 # OpenDroneID enabled boards. Use 10000 + the base board ID
 AP_HW_CubeOrange_ODID               10140
 AP_HW_Pixhawk6_ODID                 10053
+AP_HW_CUAVv5Nano_ODID               10050
+AP_HW_CUAVv5_ODID                   10050
```
peterbarker: Duplicate IDs; this doesn't look right.

Tools/AP_Bootloader/board_types.txt
```diff
@@ -487,6 +487,11 @@ AP_HW_AEROFOX_H7                     7110
...
+AP_HW_CUAV-Nora_ODID                11009
+AP_HW_CUAV-X7_ODID                  11010
```
peterbarker: _PRO in here somewhere?

peterbarker (changes requested):

Tools/AP_Bootloader/board_types.txt
cuav-chen2: Both of these boards are based on fmuv5, so the IDs are the same. Should we only allocate a fmuv5 ID to replace these?

Tools/AP_Bootloader/board_types.txt
cuav-chen2: modified.

Tools/AP_Bootloader/board_types.txt
```diff
@@ -487,6 +487,10 @@ AP_HW_AEROFOX_H7                     7110
 # OpenDroneID enabled boards. Use 10000 + the base board ID
 AP_HW_CubeOrange_ODID               10140
 AP_HW_Pixhawk6_ODID                 10053
+AP_HW_FMU_V5_ODID                   10050
```
peterbarker, line 490: I don't think we want a generic FMU_V5 ODID.  It's at odds with the "vendor is validating firmware" thing that ODID is about.

Tools/AP_Bootloader/board_types.txt
```diff
@@ -487,6 +487,10 @@ AP_HW_AEROFOX_H7                     7110
...
+AP_HW_CUAV_NORA_ODID                11009
+AP_HW_CUAV_X7_PRO_ODID              11010
+AP_HW_CUAV-7-NANO_ODID              17000
```
peterbarker, line 493: These look correct.

peterbarker (changes requested):
Could you split off the FMUv5 stuff into a different PR, please?  I don't think it actually works to have an ODID firmware using the fmuv5 board ID as a base.  To do so would thwart one of the protections that OpenDroneID has - that the people allowed to create firmwares for a board is limited to the people who created the bootloader on the board (leaving aside the "signed bootloader" stuff).  I'd be happy to merge all of the other boards, just not the fmuv5 stuff without more thought, at least.

cuav-chen2:
Thanks for merge, now close this PR.

PR #30238, opened 2025-06-04
AC_Fence: ensure that margin breaches are reset on landing

andyp1per:
We were not clearing margin breaches in landing even though the actual breach (alt min) had been cleared


rmackay9:
Hi @andyp1per I've added this to the 4.6 backports but please tell me if we don't need that

andyp1per:
@rmackay9 I think that's good - thanks

andyp1per:
@rmackay9 I have removed this from 4.6 since the margin change is only in 4.7

peterbarker (approved):

PR #30239, opened 2025-06-04
AP_Scripting: Add GPS switch script

Ryanf55:
* Use this script and an Aux switch to change the primary GPS for the EKF
* Tested in SITL, will fly soon


libraries/AP_Scripting/examples/gps_switch.lua
```diff
@@ -0,0 +1,60 @@
...
+
+-- Constants
+local GPS_FIRST = 0
+local GPS_SECOND = 1
+
+function update()
+   -- Read switch position (0=low, 1=mid, 2=high)
+   local sw_pos = rc:get_aux_cached(AUX_FUNCTION_NUM)
+   if sw_pos == nil then
+      gcs:send_text(1, "Waiting for RC input...")
```
tridge: this doesn't mean it is waiting for R/C input, it just means the user hasn't hit the switch yet (for example AUX functions can be buttons on the GCS)
should just return with no msg

tridge (changes requested):

libraries/AP_Scripting/examples/gps_switch.lua
tridge: in mavproxy use this: 
```
@alias add aux long DO_AUX_FUNCTION
```
and "aux 0", "aux 1", "aux 2"

Ryanf55: Done.

Ryanf55:
This has been tested in flight now.

peterbarker:
Merged!

PR #30240, opened 2025-06-04
AC_Fence: only provide distances to fences that are actually present

andyp1per:


rmackay9:
Txs again, I've added this to the 4.6 backports

andyp1per:
@rmackay9 I have removed this from 4.6 since the margin change is only in 4.7

PR #30245, opened 2025-06-05
AP_Scripting: enable PPP in SIM7600 modem driver

tridge:
this allows for the modem to be used as a PPP server, allowing the full range of ArduPilot networking capabilities, which means multiple TCP and UDP connections, scripting sockets etc
The driver now supports either a single TCP connection or a full PPP connection. 
I've also tested at up to 3686400 baud (3.6 MBaud) which makes for a very fast link

thanks to @Ryanf55 for the suggestion to try PPP !


libraries/AP_Scripting/drivers/SIM7600.md
```diff
@@ -21,6 +21,20 @@ The script uses the following parameters:
...
 
+## LTE_PROTOCOL
+
+This controls if a PPP connection will be used or a raw TCP connection
+with MAVLink2.
+
+Set LTE_PROTOCOL to 48 for PPP and enable ArduPilot networking with
+the NET_ENABLE=1 parameter.
+
+Set LTE*PROTOCOL to 2 for MAVLink2 and enable set the LTE_SERVER
```
Ryanf55:
```suggestion
Set LTE_PROTOCOL to 2 for MAVLink2 and enable set the LTE_SERVER
```

libraries/AP_Scripting/drivers/SIM7600.md
```diff
@@ -38,38 +52,41 @@ serial devices.
...
 This is the first octet of the server IP address to connect to. The
 full IP address is constructed from LTE_SERVER_IP0 through
-LTE_SERVER_IP3. Range: 0-255.
+LTE_SERVER_IP3. Range: 0-255. This is not used with PPP.
 
 ## LTE_SERVER_IP1
 
 This is the second octet of the server IP address to connect to.
-Range: 0-255.
+Range: 0-255. This is not used with PPP.
```
Ryanf55, line 60: Can you hide LTE_SERVER_IP* params if you set LTE_PROTOCOL to 48?

libraries/AP_Scripting/drivers/SIM7600.md
```diff
@@ -38,38 +52,41 @@ serial devices.
...
-the same baud rate. Range: 9600-115200. Default: 115200.
+the same baud rate. Range: 9600-3686400. Default: 115200.
 
 If using something other than 115200 you need to connect to the modem
 with a terminal program and use AT+IPREX=BAUD to set the baud rate and
-save with AT&W
+then save with AT&W. If the modem is wired to a flight controller then
+you can use the SERIAL_PASS parameters to give temporary control of
+the modem to a USB port so you can use a terminal protocol to
+configure the modem.
```
Ryanf55, line 89: Why can't the Lua script automatically just set the baud? Future update?

If the Lua script iterated through default baud and the baud you have configured to establish connection, then change the baud to the desired baud if it's not the same, that would make setup a lot easier for users.

Ryanf55 (approved):

libraries/AP_Scripting/drivers/SIM7600.md
tridge: yes, we could do that, bit more complex, but should be possible

libraries/AP_Scripting/drivers/SIM7600.md
tridge: fixed

libraries/AP_Scripting/drivers/SIM7600.md
tridge: yes, should be possible, just complex for the recovery with +++

PR #30246, opened 2025-06-05
AP_InertialSensor: support 3 harmonic notch filters

tridge:
some vehicles need 3, for example:

 - one for an EFI RPM (or generator)
 - one for ESCs
 - one for a frame resonance

I've also left open the possibility of 4, enabled in hwdef.dat
tested in SITL only


andyp1per (approved):
Nice!

PR #30249, opened 2025-06-05
AP_GPS: correct asymmetric calloc/free vs new/delete

peterbarker:
we calloc this, so delete is not correct

Picked up by CodeQL here: https://github.com/peterbarker/ardupilot/security/code-scanning/1


tpwrules:
Nobody but you seems to have permissions to view the link unfortunately so I can’t see any context from the scanner.

rmackay9:
LGTM.  BTW how dangerous is this mistake by the way?

peterbarker:
> LGTM. BTW how dangerous is this mistake by the way?

I think it comes down to the compiler.  I've read that it can be very bad on Windows.

peterbarker:
> Nobody but you seems to have permissions to view the link unfortunately so I can’t see any context from the scanner.

Sorry, I was actually wondering about that bit too :-)  We have a PR which points people at the "Security" tab in github, so I was just playing around with how it worked.

image

tpwrules (approved):
It shouldn’t matter to us since we override these functions to use the same allocator.

PR #30250, opened 2025-06-05
Potential fix for code scanning alert no. 11: Workflow does not contain permissions

peterbarker:
Potential fix for [https://github.com/peterbarker/ardupilot/security/code-scanning/11](https://github.com/peterbarker/ardupilot/security/code-scanning/11)

To fix the issue, we will add a `permissions` block at the root level of the workflow. This block will define the minimal permissions required for the workflow to function correctly. Based on the tasks performed in the workflow, the following permissions are appropriate:
- `contents: read` for accessing the repository's contents.
- `actions: write` for managing artifacts (e.g., uploading build artifacts).

The `permissions` block will be added at the top of the workflow, immediately after the `name` field.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


amilcarlucas:
Yes, this is a way to fix it. But usually you only give write permissions to a single task.
Take a look at https://github.com/ArduPilot/MethodicConfigurator/tree/master/.github/workflows for inspiration.

peterbarker:
> Yes, this is a way to fix it. But usually you only give write permissions to a single task. Take a look at https://github.com/ArduPilot/MethodicConfigurator/tree/master/.github/workflows for inspiration.

Doesn't this task in particular - the cygwin build - definitely need the write permissions for the artifacts it is creating?  I may as well do that experiment and se if it breaks!

peterbarker:
> Yes, this is a way to fix it. But usually you only give write permissions to a single task. Take a look at https://github.com/ArduPilot/MethodicConfigurator/tree/master/.github/workflows for inspiration.

Ah, I see.  permissions can be used further in.

So now there's a problem; if I just add the lines at the top then it gives a falsish-sense of security in that the permissions could be tighter.  OTOH, adding permissions into each individual task is a big job - especially since most of our tasks will probably need write permissions because of ccache.

Hmmm.

peterbarker:
@amilcarlucas so this appears to pass.

It is just changing the permission on "content" to "read", and not granting write permissions to anything.

Does this look correct to you?!

amilcarlucas:
Yes, that looks correct. Now you only need to sprinkle at job level some:

```
    permissions:
      contents: write       # for creating branches and commits
```
or

```
  permissions:
      pull-requests: write  # for creating PRs
```

In some jobs that create branches, commits or PRs.

amilcarlucas:
https://github.com/ArduPilot/MethodicConfigurator/tree/master/.github/workflows/pytest.yml is an example

peterbarker:
> ```
>   permissions:
>       pull-requests: write  # for creating PRs
> ```
> 
> In some jobs that create branches, commits or PRs.

I think the fact that no job failed means that I don't need to do that right now, at least?

peterbarker:
... how is this uploading artifacts?!  The text I opened this PR seems to indicate it shouldn't be able to unless I grant it extra permissions!?

peterbarker:
@amilcarlucas so it's this:
image

... so everything is permitted by default.

peterbarker:
I've now added sufficient `permission:` entries to the workflow files to pass when we move to the second option in the picture above, "Read repository contents and packages permissions"

robertlong13 (changes requested):
There are more instances of `/cygdrive/d/` that need to be fixed too.

peterbarker:
Probably going to chose this one.

We couldn't find any place that this actually improves security, and is maintenance overhead.

Someone who knows more about github actions needs to chime in and tell us why we need to do this one - what it protects us against.

peterbarker:
```
UTC1158 - https://github.com/ArduPilot/ardupilot/pull/30250
Put permission: into our workflows
We can’t see where this increases security
Need someone who knows what they’re doing with the workflows to tell us why this improves things
https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token seems to say actions is read-only in public repos, but don’t you need write there to create artifacts?
```

amilcarlucas:
This follows the principle of as few permissions as necessary. And the tests do pass now.

peterbarker:
> This follows the principle of as few permissions as necessary. And the tests do pass now.

I made that argument on the call and it was apparently not convincing enough to get this merged.

I'm not even sure this increases or decreases security as it is - are we actually granting permissions we don't need somehow?!  Since we tend to run all checks on Pull Request and they pass, why do we need to grant "write" to some job if we are putting "read" on the job?

I think someone with much better knowledge of github workflows should probably put this forward.  I was hoping to be able to enable the CodeQL static analysis functionality, but on further thinking that's probably not a great idea.  It picked up one real bug, but it also false-positived on a bunch of stuff which makes it look like more trouble than it is worth.

I'm going to close this now - someone else can chase it if they really feel strongly about it.

Issue #30241, opened 2025-06-04
DroneCAN GPS: Satellite count (num_sats) zeroed when no fix received


fsaulo:
## Bug report

**Issue details**

`satellites_visible` in `GPS2_RAW` is always zero for DroneCAN GPS2.

When using a DroneCAN device for GPS2, the `satellites_visible` field in the `GPS2_RAW` MAVLink topic consistently reports zero, even when a non-zero satellite count is provided in the simulated `gnss.Fix2` message.

**Version**
ArduCopter 4.5.7 (it is also present upstream)

**Steps to Reproduce:**

1. Configure the secondary GPS as a DroneCAN device in SITL.
2. Simulate a `gnss.Fix2` message with a non-zero `sats_used` value.
3. Send the simulated `gnss.Fix2` message through a virtual CAN device to the DroneCAN GPS2.
4. Observe the `satellites_visible` field of the `GPS2_RAW` MAVLink topic.

**Expected Behavior:**

The `satellites_visible` field in `GPS2_RAW` should reflect the `sats_used` value from the received `gnss.Fix2` message, even if a full fix is not yet available.

**Observed Behavior:**
The `satellites_visible` field in `GPS2_RAW` remains zero.

**Root Cause Analysis:**
A preliminary analysis of `AP_GPS/AP_GPS_DroneCAN.cpp` ([line 336](https://github.com/ArduPilot/ardupilot/blob/9eed018bd77efc34cc48bb6c320170174ab88223/libraries/AP_GPS/AP_GPS_DroneCAN.cpp#L336)) indicates that `gnss.Fix2` messages are not processed unless a valid fix is present. This results in the `num_sats` variable being zeroed out.

**Impact:**
This behavior specifically affects the DroneCAN GPS driver and does not apply to other GPS input methods, which correctly report satellite counts even without a full fix. This can hinder debugging and accurate satellite visibility monitoring when using DroneCAN GPS.

peterbarker:
@fsaulo just to be sure here - you are aware that Fix2 is used by *all* GPSs, not just the second one, right?  It's not "fix for second GPS", it's "second generation of DroneCAN message to convey GPS data because we completely screwed up the first generation message".

ie. however this is working for the second GPS should apply just the same for the first GPS.

fsaulo:
Hi @peterbarker, thanks for taking a look at this, I really appreciate it.

[...]

PR #30251, opened 2025-06-05
AP_GPS: Satellites count zeroed when no fix in gnss.Fix2 message

fsaulo:
Process `num_sats` for all incoming `gnss.Fix2` messages

See #30241.

### DIFF
--- a/libraries/AP_GPS/AP_GPS_DroneCAN.cpp
+++ b/libraries/AP_GPS/AP_GPS_DroneCAN.cpp
@@ -419,15 +419,14 @@ void AP_GPS_DroneCAN::handle_fix2_msg(const uavcan_equipment_gnss_Fix2& msg, uin
                 interim_state.have_speed_accuracy = false;
             }
         }
-
-        interim_state.num_sats = msg.sats_used;
     } else {
         interim_state.have_vertical_velocity = false;
         interim_state.have_vertical_accuracy = false;
         interim_state.have_horizontal_accuracy = false;
         interim_state.have_speed_accuracy = false;
-        interim_state.num_sats = 0;
     }
+
+    interim_state.num_sats = msg.sats_used;
 
     if (!seen_aux) {
         // if we haven't seen an Aux message then populate vdop and


libraries/AP_GPS/AP_GPS_DroneCAN.cpp
```diff
@@ -419,16 +419,15 @@ void AP_GPS_DroneCAN::handle_fix2_msg(const uavcan_equipment_gnss_Fix2& msg, uin
...
-        interim_state.num_sats = msg.sats_used;
     } else {
         interim_state.have_vertical_velocity = false;
         interim_state.have_vertical_accuracy = false;
         interim_state.have_horizontal_accuracy = false;
         interim_state.have_speed_accuracy = false;
-        interim_state.num_sats = 0;
     }
 
+    interim_state.num_sats = msg.sats_used;
```
peterbarker, line 429: Leaving this at zero is not going to be an accident.  It's been this way since 2018, and I've not traced it back further than that.

I don't understand *why* we want to leave num_sats at zero if we don't have a good fix, it does seem strange.  But to changesuch long-standing behaviour we would *really* need to know why it is like this in the first place!  That's probably a bunch of code archeology pre-2018 ('though I was unable to find where this code was moved from in the quick look I took), or finding the right person to ask.

@fsaulo what was your reasoning for making this change?  Just better visibility of what the GPS is up to in those frustrating minutes when you don't have a position?  Please note that the number of satellites is *not* a good indication of how good your lock actually is.  I mean, it can *act* as a comforter, but it's not a good number to use for, "how good is my fix?".

rmackay9: I think I agree with @fsaulo that we should display the number of satellites.

peterbarker: > I think I agree with @fsaulo that we should display the number of satellites.

So I don't know why we wouldn't.  It would seem to be a good idea to me.

But whoever set things up like this didn't do it for fun, it was *extra* coding to do it the way we currently do!  Changing the behaviour on a whim doesn't seem like a good idea - thus my suggestion of working out why it's like this in the first place and request for more information on why it's important to @fsaulo

fsaulo: @peterbarker We work with a GPS-like platform that uses the satellite number as a status indicator. This has been working out for other GPS input transport methods (e.g. NMEA and Mavlink with `GPS_INPUT`). 

The point is _why_ would only this implementation of the `gnss.Fix2` not show the number of satellites. I've looked around and couldn't find the same behavior in other platforms, and since I couldn't find any reasonable explanation, I assumed this was a _bug_.

peterbarker (approved):
After discussion with tridge, merging this one.

It's the zeroing of the accuracies which is important here.  We think.

peterbarker:
Merged, thanks!

PR #30253, opened 2025-06-06
SITL: correct accuracies supplied by VectorNav sim

peterbarker:
... set three different fields rather than the same field three times...

### DIFF
--- a/libraries/SITL/SIM_VectorNav.cpp
+++ b/libraries/SITL/SIM_VectorNav.cpp
@@ -211,12 +211,12 @@ void VectorNav::send_ins_gnss_packet(void)
     pkt.velNed1[2] = fdm.speedD;
 
     pkt.posU1[0] = 1;
-    pkt.posU1[0] = 1;
-    pkt.posU1[0] = 1.5;
+    pkt.posU1[1] = 1;
+    pkt.posU1[2] = 1.5;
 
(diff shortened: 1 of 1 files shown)


libraries/SITL/SIM_VectorNav.cpp
```diff
@@ -211,8 +211,8 @@ void VectorNav::send_ins_gnss_packet(void)
     pkt.velNed1[2] = fdm.speedD;
 
     pkt.posU1[0] = 1;
-    pkt.posU1[0] = 1;
-    pkt.posU1[0] = 1.5;
+    pkt.posU1[1] = 1;
+    pkt.posU1[2] = 1.5;
 
     pkt.velNed1[0] = 0.05;
     pkt.velNed1[0] = 0.05;
```
robertlong13: These three lines after are the same problem too right?

robertlong13 (approved):

libraries/SITL/SIM_VectorNav.cpp
Ryanf55: Fixed and force pushed.

peterbarker: LOL.  This is a very strange way to handle scope creep, but I like it.

Thanks both :-)

PR #30259, opened 2025-06-06
Copter: Change the variable type to INT

muramura:
A FLOAT value is being compared using the INT type. Therefore, it will be changed to INT type.


ArduCopter/mode.h
```diff
@@ -779,7 +779,7 @@ class ModeAuto : public Mode {
         float down;   // desired speed downwards in m/s. 0 if unset
     } desired_speed_override;
 
-    float circle_last_num_complete;
+    int circle_last_num_complete;
```
peterbarker, line 782: I did consider doing in this way.

I think there's scope for printing partial circles if you are circling slowly, which is why I went with float here.

muramura: The number of rotations must be an integer. For example, 0.5 rotations cannot be specified.

0.99 is 0
1.01 is 1
1.99 is 1

This condition returns TRUE when the specified number of times or more has been reached.

return num_circles_completed >= turns;

https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Mission/AP_Mission.cpp#L1109

peterbarker: > The number of rotations must be an integer. For example, 0.5 rotations cannot be specified.

Are you sure?  Have you tested that?  We are supposed to handle fractional turns:

```
        float get_loiter_turns(void) const {
            float turns = LOWBYTE(p1);
            if (type_specific_bits & (1U<<1)) {
                // special storage handling allows for fractional turns
                turns *= (1.0/256.0);
            }
            return turns;
        }
```

peterbarker:
I have tested this PR on top of https://github.com/ArduPilot/ardupilot/pull/31379 (including changing the number of circles desired to 0.5).

As suspected, this PR does make it impossible to do half-a-circle.

Closing this one now.

Issue #30265, opened 2025-06-07
DroneCAN Deadlock - ArduPlane 4.7


Ryanf55:
## Bug report

**Issue details**

I'm trying to configure AM32 DroneCAN for ArduPlane for a single motor output and hit a droneCAN deadlock while using the DroneCAN GUI.

https://discord.com/channels/674039678562861068/998765238465867876/1380766360761663518

Also discussed here: 
https://discord.com/channels/674039678562861068/1378926251489890374

According to @tpwrules , I ran into the same issue. To avoid this, don't use the DroneCAN GUI in flight to be safe.

```
AP: ArduPlane V4.7.0-dev (d6948690)
AP: ChibiOS: c6d0293e
AP: Pixhawk6X 00280021 31325101 39313731
AP: IOMCU: 410 2003 412FC231
AP: RCOut: PWM:1-16
AP: IMU0: fast hi-res sampling 2.0kHz/2.0kHz
AP: IMU2: normal sampling 1.6kHz/1.6kHz
```

with --enable-DDS if you want to recompile, here's the steps.

```
./waf configure --board Pixhawk6X --extra-hwdef=ardusoar-radian.dat --enable-DDS
nice ./waf plane --upload
```
And, `ardusoar-radian.dat`
```
# Disable UDP because we use serial
define AP_DDS_UDP_ENABLED 0

# Disable VisualOdom - this vehicle relies on GPS
define AP_DDS_VISUALODOM_ENABLED 0

# No experimental topics are needed
define AP_DDS_EXPERIMENTAL_ENABLED 0

# Make time a low rate
define AP_DDS_DELAY_TIME_TOPIC_MS 1000

# Only need global pose, so disable local
define AP_DDS_LOCAL_POSE_PUB_ENABLED 0

# No local vel
define AP_DDS_LOCAL_VEL_PUB_ENABLED 0

# Bump airspeed rate down
define AP_DDS_DELAY_AIRSPEED_TOPIC_MS 100

# No RC out needed
define AP_DDS_RC_PUB_ENABLED 0

# No clock needed
define AP_DDS_CLOCK_PUB_ENABLED 0

# Disable joy input for safety
define AP_DDS_JOY_SUB_ENABLED 0

# Don't allow ROS 2 arming
define AP_DDS_ARM_SERVER_ENABLED 0

# It's not a VTOL
define AP_DDS_VTOL_TAKEOFF_SERVER_ENABLED 0
```

I can't get crash dump (yet).
[...]

PR #30266, opened 2025-06-07
AP_CANManager: fixed mutex ordering deadlock in MAVCAN

tridge:
we had a deadlock between the mutex in HAL_ChibiOS CAN layer and the MAVLinkCAN object

the deadlock arose because the low level CAN layer holds the HAL CAN mutex before it calls up to the MAVLinkCAN frame callback

on the mavlink side, it took the MAVLinkCAN mutex before calling down to the CAN send layer, this meant it took the 2 mutexes in the opposite order

the symptoms were:

 - failure to boot if we have active MAVCAN during reboot as we deadlock during the scheduler delay callback

 - deadlock during heavy MAVCAN usage

### DIFF
--- a/libraries/AP_CANManager/AP_MAVLinkCAN.cpp
+++ b/libraries/AP_CANManager/AP_MAVLinkCAN.cpp
@@ -75,9 +75,9 @@ void AP_MAVLinkCAN::handle_can_frame(const mavlink_message_t &msg)
 {
     if (frame_buffer == nullptr) {
         // allocate frame buffer
-        WITH_SEMAPHORE(can_forward.sem);
         // 20 is good for firmware upload
         uint8_t buffer_size = 20;
+        WITH_SEMAPHORE(frame_buffer_sem);
         while (frame_buffer == nullptr && buffer_size > 0) {
             // we'd like 20 frames, but will live with less
             frame_buffer = NEW_NOTHROW ObjectBuffer<BufferFrame>(buffer_size);
@@ -107,8 +107,10 @@ void AP_MAVLinkCAN::handle_can_frame(const mavlink_message_t &msg)
             bus : p.bus,
             frame : AP_HAL::CANFrame(p.id, p.data, p.len)
         };
-        WITH_SEMAPHORE(can_forward.sem);
-        frame_buffer->push(frame);
+        {
+            WITH_SEMAPHORE(frame_buffer_sem);
+            frame_buffer->push(frame);
+        }
         break;
     }
 #if HAL_CANFD_SUPPORTED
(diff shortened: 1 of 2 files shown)


tpwrules:
This matches my analysis, but I think you typoed. One of the offending mutexes (mutices?) is in the AP_HAL CANIface; it's not specific to AP_HAL_ChibiOS and that doesn't seem to have a mutex.

Is the refactoring of `AP_MAVLinkCAN::can_frame_callback` supposed to address the reboot bug? Could you split into a separate commit? If it's just a cleanup, I'd prefer a separate PR. There are some other cleanups in here that need doing as well.

tpwrules:
Also should fix #30265

tridge:
> Is the refactoring of `AP_MAVLinkCAN::can_frame_callback` supposed to address the reboot bug?

it is needed to minimise the amount of code we hold the mutex over, so we don't hold can_forward.sem while we are sending mavlink messages. So it is all part of the same fix. View the code with ignore whitespace enabled and I think you'll find it clear

tpwrules:
I agree that it's a good change, I just don't think it actually changes behavior. Those mavlink functions can't call callbacks or acquire further mutexes can they?

tridge:
> Those mavlink functions can't call callbacks or acquire further mutexes can they?

I wouldn't guarantee it. They can do serial writes, and serial ports can be tunnelled either on networking or on CAN. I suspect it either doesn't cause an issue or it will be a very rare issue, but I think it should be fixed and is really part of the same general issue of holding a local data structure mutex while calling to external functions, so I think it is part of the same problem in this class

Ryanf55:
I only hit this once in 4 hours of mucking about on the bench. I have no idea how to reproduce it. 

I would note that I hit it with **light** usage of one AM32 ESC and one one airspeed sensor.

tridge:
@Ryanf55 I was able to reproduce by having the GUI tool ESC panel open and sending messages at high rate over MAVCAN when I reboot the flight controller

tpwrules:
> the same general issue of holding a local data structure mutex while calling to external functions 

The flipside of fixing like this is now there is no longer a guarantee that you won't get any further messages to a destination once the filter change function returns. This doesn't matter here, but I think in the callback functions it could.

tridge:
> The flipside of fixing like this is now there is no longer a guarantee that you won't get any further messages to a destination once the filter change function returns.

why would that worry us? Once filter change function returns then we should accept more messages, why would we want to block them?

tpwrules:
I think it's fine for the filter function. The problem is easier to illustrate with `callbacks.sem` in `AP_HAL::CANIface::receive`, in a hypothetical scenario where that is changed to drop the callback lock while the callback is executing.

1. You have a callback function registered which e.g. puts the packet in a buffer.
2. A packet comes in and the lock is taken and the callback is loaded
3. The lock is released and the callback is called
4. You unregister the callback on another thread and that goes through since the lock is released
5. You free the buffer the callback puts packets into cause you're not using it anymore
6. But whoops, the callback was still executing, use after free!

I don't think the change to the mavlink code here creates a problem like that, at worst we will send one more message than the receiver was expecting but it can't know when we will process its request so it can't know when to stop expecting messages.

tpwrules (approved):
Able to update a Here4 through a Cube Orange without any deadlocks (tried five times), where before this PR it would deadlock just about every time, sometimes multiple times per update.

Also sat down to look through the code logic and am reasonably convinced it's okay. My eyes are noticing more cleanup opportunities but I think this is good to go in. Do we need to backport to periph?

tridge:
@tpwrules I don't think we need a backport as AP_Periph usually doesn't do MAVCAN

PR #30267, opened 2025-06-07
Delete CarbonixL496

robertlong13:
The HolybroG4_GPS accidentally got assigned the same ID when it was introduced. The CarbonixL496 was obscure enough to delete entirely and reassign the ID to the HolybroG4_GPS.

I'm pretty sure no CarbonixL496s actually exist anymore, and if they do, they are my problem, so this is safe.


peterbarker (approved):
LGTM

PR #30273, opened 2025-06-08
hwdef: add new pixhawk fmu6x board, accton godwit ga1

jyhminwang:
Add a new board hwdef files
It is a pixhawk mfu v6x variant, use IMU set (BMI088, ICM-42688-P, ICM-42688-P)
And the Product can be found in https://www.accton-iot.com/godwit/index.html


jyhminwang:
Hi maintainers,
I'd updated to latest branch commit b252d89c854ac7cfebefe6403eff1301313c49ae
But the faild CI test seems from that change
Should I wait for next successful commit?
Or could please maintainers just put this PR to merge queue if your review is fine? 

image

Hwurzburg:
@andyp1per @tridge shouldn't this be a PR against the Pixhawk6X hwdef for this variation?

peterbarker:
> @andyp1per @tridge shouldn't this be a PR against the Pixhawk6X hwdef for this variation?

That, or my preference, a completely new board.  Allows the author to have their own branding on the board, avoids workaround for ... problematic....  specifications (such as `detect_fmuv6_variant` (and *that* is going to go to hell when the manufacturers git upon a combination of sensors which makes it impossible for us to distinguish the various boards....))

Having a separate board config like this allows the user to have better branding than they otherwise might.  Since we haven't done this for CUAV or Holybro, you're probably correct, @Hwurzburg .

My preference would be to drop CONFIG_ARCH_FMUV6 here and have it as a completely stand-alone board.  I'm not sure that's suitable for the user.

jyhminwang:
Hi Sirs
Actually we did feel confused with the board definition. But this PR could be a  balance approach … at least at this moment

Hwurzburg:
@jyhminwang so we discussed....we need to know if you are contractually obligated to ship this device with a PIX4 board id or not or if the board is not going to be shipped with ArduPilot and has not got a dfu button....if both of these are untrue, then we should create a unique board id, and make a targeted hwdef for only this board....otherwise, then we need to change this to being a revision of the Pixhawk6X hwdef files

jyhminwang:
Accton will request to reserve board ids
So, close this PR

PR #30274, opened 2025-06-08
AP_Networking: make PPP much more robust

tridge:
this fixes two key issues with PPP:

 - if we lost the link (eg. serial cable unplugged) then it did not properly restart the link when the cable is plugged back in

 - if using hardware flow control and the UART write buffer fills up then we would sleep with the TCPIP lock held, which means if the main thread tries to send an IPv4 packet then it would block on the TCPIP lock, and we would watchdog

### DIFF
--- a/libraries/AP_Networking/AP_Networking_PPP.cpp
+++ b/libraries/AP_Networking/AP_Networking_PPP.cpp
@@ -28,6 +28,11 @@ extern const AP_HAL::HAL& hal;
 #define PPP_DEBUG_TX 0
 #define PPP_DEBUG_RX 0
 
+// timeout for PPP link, if no packets in this time then restart the link
+#ifndef PPP_LINK_TIMEOUT_MS
+#define PPP_LINK_TIMEOUT_MS 5000U
+#endif
+
 /*
   output some data to the uart
  */
@@ -37,6 +42,12 @@ uint32_t AP_Networking_PPP::ppp_output_cb(ppp_pcb *pcb, const void *data, uint32
     LWIP_UNUSED_ARG(pcb);
     uint32_t remaining = len;
     const uint8_t *ptr = (const uint8_t *)data;
+
+    if (pcb->phase == PPP_PHASE_TERMINATE) {
+        // don't output while terminating
+        return 0;
+    }
+
 #if PPP_DEBUG_TX
     bool flag_end = false;
     if (ptr[len-1] == 0x7E) {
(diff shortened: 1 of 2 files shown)


libraries/AP_Networking/AP_Networking_PPP.cpp
```diff
@@ -169,6 +194,31 @@ bool AP_Networking_PPP::init()
...
+ */
+static void netif_promote(struct netif *iface)
+{
+    extern struct netif *netif_list;
+    if (netif_list == nullptr || netif_list == iface) {
+        // already first in the list
+        return;
+    }
+    for (struct netif *prev = netif_list;
+         prev != nullptr;
```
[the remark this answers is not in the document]
rmackay9: definitely don't want to use NULL

rmackay9:
Great to see this.  Sorry, it's just so easy to request a review by Co-pilot

timtuxworth:
I haven't been able to demonstrate problems repeatably, but I've definitely experienced reliability issues with PPP, especially in cases of companion computers which can take longer to boot than the FC. I'll be looking forward to trying this, I have a couple of vehicles that I can test this one that are flyable without any changes.

This is great thanks @tridge !

libraries/AP_Networking/AP_Networking_PPP.cpp
```diff
@@ -58,15 +69,20 @@ uint32_t AP_Networking_PPP::ppp_output_cb(ppp_pcb *pcb, const void *data, uint32
...
 #endif
-    while (remaining > 0) {
-        const auto n = driver.uart->write(ptr, remaining);
-        if (n > 0) {
-            remaining -= n;
-            ptr += n;
-        } else {
-            hal.scheduler->delay_microseconds(100);
-        }
+    if (driver.uart->txspace() < remaining) {
```
rmackay9, line 72: On the dev call the suggestion came up that we should add logging of this somehow

tridge: we could, but logging PPP stats should really be a separate PR I think. This is a minimal fix that we want for 4.6.x

Ryanf55:
Just pulled master and tested it with DDS, LGTM. Nice improvement.

rmackay9:
This is included in 4.6.2-beta1 via PR https://github.com/ArduPilot/ardupilot/pull/30432

PR #30276, opened 2025-06-08
Plane: support individual vtail and elevon mixing parameter

mduclehcm:
## Summary
This PR introduces support for individual mixing parameters for V-tail and elevon configurations in ArduPlane, providing more granular control over mixing behavior compared to the existing global mixing parameters.

- **Individual Mixing Control**: Added `MIX_INDV_EN` parameter to enable/disable individual mixing parameters
- **V-tail Specific Parameters**:
  - `MIX_VT_GAIN` - Individual gain control for V-tail mixing (Range: 0.0-1.0, Default: 0.5)
  - `MIX_VT_OFFSET` - Individual offset control for V-tail mixing (Range: -1000-1000, Default: 0)
- **Elevon Specific Parameters**:
  - `MIX_EL_GAIN` - Individual gain control for elevon mixing (Range: 0.0-1.0, Default: 0.5)
  - `MIX_EL_OFFSET` - Individual offset control for elevon mixing (Range: -1000-1000, Default: 0)

### Backward Compatibility
- When `MIX_INDV_EN` is disabled (default), the system uses existing global `MIXING_GAIN` and `MIXING_OFFSET` parameters
- When enabled, V-tail and elevon mixing use their respective individual parameters
- No breaking changes to existing configurations

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -1269,6 +1269,10 @@ const AP_Param::GroupInfo ParametersG2::var_info[] = {
     // @Path: systemid.cpp
     AP_SUBGROUPINFO(systemid, "SID", 38, ParametersG2, AP_SystemID),
 #endif
+
+    // @Group: MIX_
+    // @Path: mixing.cpp
+    AP_SUBGROUPINFO(mixing, "MIX_", 39, ParametersG2, AP_Mixing),
     
     AP_GROUPEND
 };
@@ -1281,6 +1285,7 @@ ParametersG2::ParametersG2(void) :
 #if HAL_SOARING_ENABLED
     ,soaring_controller(plane.TECS_controller, plane.aparm)
 #endif
+    ,mixing()
 {
     AP_Param::setup_object_defaults(this, var_info);
 }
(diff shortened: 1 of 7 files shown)


Hwurzburg:
@mduclehcm what vehicle uses both vtail and elevons? ie why the need for this?

timtuxworth:
I'm not sure what @mduclehcm is thinking @Hwurzburg , but maybe the new AtomRC Dolphin Pro? It seems to be a bit strange. I don't have one but just watched Andrew Newton and Painless360 videos.

mduclehcm:
> I'm not sure what @mduclehcm is thinking @Hwurzburg , but maybe the new AtomRC Dolphin Pro? It seems to be a bit strange. I don't have one but just watched Andrew Newton and Painless360 videos.

Yes, I made this change specifically for the AtomRC Dolphin Pro.
This plane is lack of pitch authority when using only the V-tail for pitch control.
To address this, I implemented a mix of V-tail and elevon for pitch. In this configuration, I need to maintained high pitch authority on the V-tail while reducing it on the elevons to preserve roll authority.
I completed several flights with the updated parameters, and the aircraft performed quite well

PR #30277, opened 2025-06-08
Use cygpath to get working directory for cygwin build job

peterbarker:
Apparently this can move around, so assume /c/d isn't great

Extracted from: https://github.com/ArduPilot/ardupilot/pull/30250/commits

### DIFF
--- a/.github/workflows/cygwin_build.yml
+++ b/.github/workflows/cygwin_build.yml
@@ -168,11 +168,12 @@ jobs:
           PATH: /usr/bin:$(cygpath ${SYSTEMROOT})/system32
         shell: C:\cygwin\bin\bash.exe -eo pipefail '{0}'
         run: >-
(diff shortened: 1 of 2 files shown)


robertlong13 (approved):
This looks good to me, and it's working well on forks where it was failing previously.

Do you know how to confirm that the ccache stuff is actually working? Obviously the CI is passing, but I don't know how to confirm that we're getting the speedup we're trying for.

Regardless, everything looks right and it's certainly less broken than before, so I'm going to throw my approval on it.

peterbarker:
> Do you know how to confirm that the ccache stuff is actually working? Obviously the CI is passing, but I don't know how to confirm that we're getting the speedup we're trying for.

Locally `ccache -C -z` and run twice, looking at  stats after each run

On github... custom workflow, I guess.

rmackay9:

PR #30272, opened 2025-06-08
Run flake8 on all files in-tree

peterbarker:

PR #30279, opened 2025-06-08
waf: use flake8 --extend-exclude to enable the linter to pass

cclauss:
### Before:
% `uv tool run flake8 . ` # Lots of errors

### After:
% `uv tool run flake8 . ` # No errors

An alternative approach to https://github.com/ArduPilot/ardupilot/pull/30272#issuecomment-2954073915


Ryanf55:
I am good with either this approach or the other one with adding noqa: Flake8, depends if we want devs to be able to know the file follows Flake8 in the file or not

.flake8
```diff
@@ -19,5 +19,161 @@ extend-exclude =
     build,
     modules,
     .git
+    libraries/AP_Camera/examples/gst_rtsp_server.py
+    libraries/AP_Camera/examples/gst_rtsp_to_wx.py
+    libraries/AP_Camera/examples/gst_udp_to_rtsp.py
+    libraries/AP_Camera/examples/gst_udp_to_wx.py
+    libraries/AP_Declination/generate/generate.py
```
peterbarker, line 26: I would prefer this markup in-file so someone editing the file knows straight-up it's not checked for flake8 correctness and might be tempted to fix it.

.flake8
```diff
@@ -19,5 +19,161 @@ extend-exclude =
...
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/dma_resolver.py
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/STM32F100xB.py
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/STM32F103xB.py
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/STM32F105xC.py
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/STM32F303xC.py
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/STM32F405xx.py
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/STM32F407xx.py
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/STM32F412Rx.py
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/STM32F413xx.py
+    libraries/AP_HAL_ChibiOS/hwdef/scripts/STM32F427xx.py
```
peterbarker, line 44: These are generated files; it might be possible to fix the generator to produce linted python.

peterbarker (changes requested):
I don't mind the concept of flipping to a must-be-marked-as-crap rather than must-be-marked-as-good model.

Commit message needs fixing - this has nothing to do with waf.

.flake8
cclauss: I will try to do that in a separate PR.

.flake8
cclauss: `markup in-file` also seems better because `pre-commit run flake8 --all-files` seems to ignore `--extend-exclude=`.

cclauss:
I thought that files in the root directory, such as `.flake8` and `.pre-commit-config.yaml`, were considered `waf` by `git-subsystems-split`.
https://github.com/ArduPilot/ardupilot/pull/29733/files#diff-0244bc85528ea956d0e0fba3a130a7577e1c6d0e2f737187a8c36e4f3e225ea1R80-R81

cclauss:
Closing in favor of
* #30288 and other PRs that will follow it.

.flake8
cclauss: Done in:
* #30340

.flake8
cclauss: @peterbarker Can you please point me to the code that generates these files?
* #30340

PR #30280, opened 2025-06-08
Generate README.md automatically from hwdef.dat

andyp1per:
For example:
```
./libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py ./libraries/AP_HAL_ChibiOS/hwdef/SequreH743/hwdef.dat --generate-readme
```
Will generate README.md for the SequreH743 using information in the hwdef.dat

### DIFF
--- a/libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
+++ b/libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
@@ -119,6 +119,9 @@ class ChibiOSHWDef(hwdef.HWDef):
         # list of shared up timers
         self.shared_up = []
 
+        # board name
+        self.board_mame = ""
+
     def get_mcu_lib(self, mcu):
         '''get library file for the chosen MCU'''
         import importlib
@@ -1677,8 +1680,7 @@ INCLUDE common.ld
             return default
         return p.extra_value(name, type=str, default=default)
 
-    def write_UART_config(self, f):
-        '''write UART config defines'''
+    def get_UART_list(self):
         serial_list = self.get_config('SERIAL_ORDER', required=False, aslist=True)
         hide_iomcu_uart = False
         if 'IOMCU_UART' in self.config:
@@ -1686,6 +1688,11 @@ INCLUDE common.ld
 
         if 'IOMCU_UART' in self.config and self.config['IOMCU_UART'][0] not in serial_list:
             serial_list.append(self.config['IOMCU_UART'][0])
+        return serial_list
+
+    def write_UART_config(self, f):
+        '''write UART config defines'''
+        serial_list = self.get_UART_list()
         if serial_list is None:
             return
         while len(serial_list) < 3: # enough ports for CrashCatcher UART discovery
@@ -2724,6 +2731,9 @@ Please run: Tools/scripts/build_bootloaders.py %s
 
     def write_processed_defaults_file(self, filepath):
         # see if board has a defaults.parm file or a --default-parameters file was specified
+        if not self.hwdef[0] or not args.params:
+            return
+
         defaults_filename = os.path.join(os.path.dirname(self.hwdef[0]), 'defaults.parm')
         defaults_path = os.path.join(os.path.dirname(self.hwdef[0]), args.params)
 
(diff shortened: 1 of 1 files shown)


libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -3107,6 +3117,294 @@ def write_default_parameters(self):
         self.romfs_add('defaults.parm', filepath)
         self.have_defaults_file = True
 
+    def readme_write_intro(self, f):
```
peterbarker, line 3112: Nice!  I actually attempted this here: https://github.com/ArduPilot/ardupilot/pull/27128

Could you pull this into a separate tool which instantiates `HWDef`, instead of in-lining it here?

Also, this new tool belongs in `libraries/AP_HAL/hwdef/scripts`, not in the ChibiOS directory.

peterbarker (changes requested):

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
andyp1per: @peterbarker it requires stuff that is only in the ChibiOS version so I don't think that is possible

MallikarjunSE:
> For example:
> ```
> ./libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py ./libraries/AP_HAL_ChibiOS/hwdef/SequreH743/hwdef.dat --generate-readme
> ```
> Will generate README.md for the SequreH743 using information in the hwdef.dat

Amazing!

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
peterbarker: Yes, it's possible, just requires structuring appropriately.

I've force-pushed my branch after fixing it up for the hwdef restructure.  It pulls information from the hwdef which is only present in the ChiibOS hwdef, but renders the baro, compass and IMU lists for both bbbmini and Pixhawk6X.

As part of this I remembered why I stopped working on this - I hit a roadblock with the current format of the hwdef; I will highlight it in your PR here.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
```diff
@@ -3107,6 +3117,294 @@ def write_default_parameters(self):
...
+        f.write('''
+## Features
+
+ - MCU - %s 32-bit processor running at 480 MHz\n''' % self.mcu_type)
+        count = 1
+        for dev in self.imu_list:
+            f.write(' - IMU%u - %s\n' % (count, dev[0]))
+            count = count+1
+        if self.baro_list:
+            f.write(' - Barometer - %s\n' % self.baro_list[0][0])
```
peterbarker, line 3128: This is a problem if the board has more than one barometer.

And the real problem is that I don't think we can distinguish between boards that have two barometers and boards that have *alternative* barometers, where a vendor has changed parts do to unavailability.

peterbarker:
This is what my tool currently produces for Pixhawk6X:
```
pbarker@crun:~/rc/ardupilot(pr/rst-from-hwdef)$ cat /tmp/some_html/Pixhawk6X.rst 

============
UART Mapping
============

 - SERIAL0 -> OTG1
 - SERIAL1 -> UART7 (RTS/CTS pins)
 - SERIAL2 -> UART5 (RTS/CTS pins)
 - SERIAL3 -> USART1
 - SERIAL4 -> UART8
 - SERIAL5 -> USART2 (RTS/CTS pins)
 - SERIAL6 -> UART4
 - SERIAL7 -> USART3
 - SERIAL8 -> OTG2


====
Baro
====

 - BMP388 * 2
 - ICP201XX * 2


=======
Compass
=======

 - BMM150
 - RM3100


===
IMU
===

 - Invensensev3 * 8
 - BMI088 * 2
 - Invensensev2 * 2
 - ADIS1647x


======
SDCard
======

 - microSD card slot (SDMMC2)
pbarker@crun:~/rc/ardupilot(pr/rst-from-hwdef)$ 
```

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
andyp1per: There are obviously limitations, but its a good starting point. I think ultimately we should force metadata to be put in hwdef that removes these issues, but that will have to evolve over time.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
andyp1per: It needs to be able to access functions and data in the chibios hwdef script. I do not want to refactor that as part of this, that can come later I also do not want to start copying the chibios hwdef parsing stuff, that's just make work for no compelling reason. I understand that you would like this all to be completely generic, but that is not the issue we are facing here - we have a steady stream of chibios based boards where getting the README done is a significant drain on my time.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
peterbarker: > There are obviously limitations, but its a good starting point. I think ultimately we should force metadata to be put in hwdef that removes these issues, but that will have to evolve over time.

Well, I put in the "serial purpose" stuff in to see what that might look like.

libraries/AP_HAL_ChibiOS/hwdef/scripts/chibios_hwdef.py
peterbarker: > It needs to be able to access functions and data in the chibios hwdef script. I do not want to refactor that as part of this, that can come later I also do not want to start copying the chibios hwdef parsing stuff, that's just make work for no compelling reason. I understand that you would like this all to be completely generic, but that is not the issue we are facing here - we have a steady stream of chibios based boards where getting the README done is a significant drain on my time.

It's not hard to make this work correctly.  And *not* within ChibiOSHWDef, in which it absolutely, categorically, does not belong.

I have provided an example where you can accomplish the same thing (a) generically and (b) outside of ChibiOSHWDef in a self-contained script.

I could move your existing implementation into such a structure if you're amenable.

Hwurzburg:
Would be great if we get this to generate readme files since I spend a lot of my time on getting them not only correct, but also in a format that allows me to generate a wiki page in less than one hour instead of three....hence not only the content must be complete and accurate, the format must be close to the wiki page....since I spend over 50% of my wiki time on review of new autopilot boards, I am very concerned that any readme generator conform to current pattern that reduces my conversion time...

andyp1per:
> Would be great if we get this to generate readme files since I spend a lot of my time on getting them not only correct, but also in a format that allows me to generate a wiki page in less than one hour instead of three....hence not only the content must be complete and accurate, the format must be close to the wiki page....since I spend over 50% of my wiki time on review of new autopilot boards, I am very concerned that any readme generator conform to current pattern that reduces my conversion time...

This is literally why I did this ;)

PR #30281, opened 2025-06-08
AP_AHRS: added sim as ahrs_ekf_type param desc

snktshrma:


peterbarker (changes requested):
Commit message is incorrect; this is `AP_AHRS:` not AP_NavEKK3:`

peterbarker:
Merged, thanks!

PR #30283, opened 2025-06-09
AP_Scripting: add CMUX to LTE modem driver

tridge:
This adds GSM 07.10 muxing to the LTE modem driver, allowing it to maintain separate AT command streams and data streams for PPP or transparent TCP
This allows us to query the signal strength and other information about the link while running PPP.
It also allows for auto-baudrate changing, so the user can set LTE_BAUD to a higher baud rate without having to pre-configure the modem.

### DIFF
--- a/libraries/AP_Scripting/drivers/LTE_modem.lua
+++ b/libraries/AP_Scripting/drivers/LTE_modem.lua
@@ -96,11 +96,11 @@ local LTE_SERVER_PORT = bind_add_param('SERVER_PORT',  8, 0)
 --[[
     // @Param: LTE_BAUD
     // @DisplayName: Serial Baud Rate
-    // @Description: Baud rate for the serial port to the LTE modem. If using something other than 115200 you need to connect to the modem and use AT+IPREX=BAUD to set the baud rate and save with AT&W. The fastest supported baudrate is 3686400.
+    // @Description: Baud rate for the serial port to the LTE modem when connected. Initial power on baudrate is in LTE_IBAUD
     // @Values: 19200:19200,38400:38400,57600:57600,115200:115200,230400:230400,460800:460800,921600:921600,3686400:3686400
     // @User: Standard
 --]]
-local LTE_BAUD        = bind_add_param('BAUD',  9, 115200)
+local LTE_BAUD        = bind_add_param('BAUD',  9, 921600)
 
 --[[
     // @Param: LTE_TIMEOUT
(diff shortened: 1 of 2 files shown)


libraries/AP_Scripting/drivers/LTE_modem.lua
```diff
@@ -94,11 +96,11 @@ local LTE_SERVER_PORT = bind_add_param('SERVER_PORT',  8, 0)
 --[[
     // @Param: LTE_BAUD
     // @DisplayName: Serial Baud Rate
-    // @Description: Baud rate for the serial port to the LTE modem. If using something other than 115200 you need to connect to the modem and use AT+IPREX=BAUD to set the baud rate and save with AT&W. The fastest supported baudrate is 3686400.
+    // @Description: Baud rate for the serial port to the LTE modem when connected. Initial power on baudrate is in LTE_IBAUD
     // @Values: 19200:19200,38400:38400,57600:57600,115200:115200,230400:230400,460800:460800,921600:921600,3686400:3686400
     // @User: Standard
 --]]
-local LTE_BAUD        = bind_add_param('BAUD',  9, 115200)
+local LTE_BAUD        = bind_add_param('BAUD',  9, 921600)
```
IamPete1, line 103: Why not pull from the serial baud rate params for the scripting port. (we really should make this easier.)

libraries/AP_Scripting/drivers/LTE_modem.lua
```diff
@@ -148,49 +195,294 @@ local log_file = io.open('SIM7600.log', 'w')
...
+local EA = 0x01
+local CR_SEND = 0x02
+
+-- CMUX buffer state
+local cmux = {}
+cmux.buffers = {[1] = "", [2] = ""} -- DLC1=AT, DLC2=DATA(PPP or TCP)
+
+--[[
+    FCS lookup table for polynomial x^8 + x^2 + x^1 + 1 (0x07)
+    This is the reverse of the standard CRC-8 table
```
IamPete1, line 227: Do we have this in CPP? I wonder it if would make sense to expose it. (not in this PR tho) .

libraries/AP_Scripting/drivers/LTE_modem.lua
```diff
@@ -148,49 +195,294 @@ local log_file = io.open('SIM7600.log', 'w')
...
-    end
+    log_data(s, '<<<')
+    stats.bytes_in = stats.bytes_in + #s
     return s
 end
 
+--[[
+    Function to write to the UART and log the command
+--]]
+local function uart_write(s)
```
IamPete1, line 206: BTW, you can use the same markup as the lua docs to get better type checking. Once scripts start getting complex its worth the effort. 

For example:
```suggestion
---@param s string
---@return number
local function uart_write(s)
```

IamPete1 (approved):
Only scripting and it passes checks, so I'm happy.

libraries/AP_Scripting/drivers/LTE_modem.lua
tridge: we could, but this saves a config step, as the modems come pre-configured at 115200

libraries/AP_Scripting/drivers/LTE_modem.lua
tridge: I don't know, but I want the script to work with 4.6.x anyway

libraries/AP_Scripting/drivers/LTE_modem.lua
tridge: ok, thanks for the note

libraries/AP_Scripting/drivers/LTE_modem.md
```diff
@@ -188,3 +194,11 @@ Common issues:
 
 If the connection fails or is lost, the driver will automatically attempt
 to reconnect by restarting the connection sequence.
+
+# Logging
+
+A LTE log message is saved in the onboard log. That has signal
+strength information and data transfer statistics.
+
+A NAMED_VALUE_FLOAT MAVLink message "LTE_RSSI" is sent with the RSSI
```
Ryanf55, line 203: Have you considered using this packet?
https://mavlink.io/en/messages/common.html#RADIO_STATUS

tridge: I did consider it, but it isn't a good fit

PR #30286, opened 2025-06-09
Adjust Ainstein LR-D1 driver to handle new packet format

peterbarker:
More information to come here, but a new protocol format allows the rangefinder to directly return its ranging status.

### DIFF
--- a/libraries/AP_RangeFinder/AP_RangeFinder_Ainstein_LR_D1.cpp
+++ b/libraries/AP_RangeFinder/AP_RangeFinder_Ainstein_LR_D1.cpp
@@ -82,7 +82,6 @@ bool AP_RangeFinder_Ainstein_LR_D1::get_reading(float &reading_m)
     if (u.packet.header_lsb != PACKET_HEADER_LSB ||
         u.packet.device_id != 0x00 ||
         u.packet.length != 28 ||
-        u.packet.objects_number != 1 ||
         u.calculate_checksum() != u.packet.checksum) {
         // sanity checks failed - discard and try again next time we're called:
         move_signature_in_buffer(1);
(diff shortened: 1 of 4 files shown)


libraries/AP_RangeFinder/AP_RangeFinder_Ainstein_LR_D1.cpp
```diff
@@ -24,62 +24,136 @@
...
 
-        const uint8_t malfunction_alert = buffer[1];
-        reading_m = UINT16_VALUE(buffer[3], buffer[4]) * 0.01;
-        const uint8_t snr = buffer[5];
+    // distinguish between old and new packet format by inspecting
+    // "byte 12" (offset 11); in the old format this will always be
+    // 0xff.  This was reserved for "offset count", now is altitude-valid.
+    const bool is_v19000 = (u.packet_v19000.altitude_valid != 0xff);
+
+    static uint8_t last_altitude_valid = -1;
```
rmackay9: normally we don't use static variables

libraries/AP_RangeFinder/AP_RangeFinder_Ainstein_LR_D1.cpp
```diff
@@ -24,62 +24,136 @@
...
+    if (u.packet_v19000.altitude_valid != last_altitude_valid) {
+        GCS_SEND_TEXT(MAV_SEVERITY_INFO, "out_of_range_field = %u", u.packet_v19000.altitude_valid);
```
rmackay9: won't this spam the user?

rmackay9:
This generally looks OK to me so after the comments are address I'm OK with merging.

peterbarker:
@rmackay9 's comments have been addressed.

Vendor has tested and is happy with it.

Only minor changes since Randy looked at this.  I have added a little bit more in the way of autotest as well.

Intending to merge this one soon.

PR #30289, opened 2025-06-09
CI: validate parameter files

robertlong13:
We don't automatically check our parameter files for correctness (name typos, param renames/removals, etc.). This adds the following, bare minimum checks (links take you to existing failures in the repo):
- [Every line in the parameter file has valid syntax](https://github.com/ArduPilot/ardupilot/blob/8fe927c1d3eb6be9da8544ac48f7ec210abf5789/Tools/autotest/default_params/motorboat.parm#L2)
- [Every parameter actually exists](https://github.com/ArduPilot/ardupilot/blob/8fe927c1d3eb6be9da8544ac48f7ec210abf5789/libraries/AP_HAL_ChibiOS/hwdef/SkystarsH7HD/defaults.parm#L7)
- [Every parameter appears in a file only once](https://github.com/ArduPilot/ardupilot/blob/8fe927c1d3eb6be9da8544ac48f7ec210abf5789/Tools/autotest/default_params/sub.parm#L5) (helps find copy/paste errors)
- [We don't try to set parameters that are ReadOnly](https://github.com/ArduPilot/ardupilot/blob/8fe927c1d3eb6be9da8544ac48f7ec210abf5789/libraries/AP_HAL_ChibiOS/hwdef/skyviper-v2450/defaults.parm#L181).

In Peter Barker style, I'm adding this test with a long whitelist for all the param files that are failing, and I'll chip away at fixing those files over the next few weeks as individual PRs.

### DIFF
--- a/Tools/scripts/board_list.py
+++ b/Tools/scripts/board_list.py
@@ -3,6 +3,7 @@
 import os
 import re
 import fnmatch
+from collections.abc import Collection
 
 '''
 list of boards for build_binaries.py and custom build server
@@ -27,10 +28,10 @@ class Board(object):
         ]
 
 
-def in_blacklist(blacklist, b):
-    '''return true if board b is in the blacklist, including wildcards'''
-    for bl in blacklist:
-        if fnmatch.fnmatch(b, bl):
+def in_boardlist(boards : Collection[str], board : str) -> bool:
+    '''return true if board is in a collection of wildcard patterns'''
+    for pattern in boards:
+        if fnmatch.fnmatch(board, pattern):
             return True
     return False
 
@@ -145,7 +146,7 @@ class BoardList(object):
                 ret += [line]
         return ret
 
-    def find_autobuild_boards(self, build_target=None):
+    def find_autobuild_boards(self, build_target=None, skip : Collection[str] = None):
         ret = []
         for board in self.boards:
             if board.is_ap_periph:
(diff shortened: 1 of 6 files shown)


Tools/scripts/board_list.py
```diff
@@ -145,7 +145,7 @@ def read_hwdef(self, filepath):
                 ret += [line]
         return ret
 
-    def find_autobuild_boards(self, build_target=None):
+    def find_autobuild_boards(self, build_target=None, blacklist=None):
```
cclauss: Probably more politically correct to switch to `blocklist` or `denylist` while making these changes.

Tools/scripts/param_check_all.py
```diff
@@ -0,0 +1,221 @@
...
+import sys
+import glob
+from board_list import BoardList
+from param_check import get_metadata, check_file, SkippedChecks
+
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../Tools', 'autotest'))
+from pysim.vehicleinfo import VehicleInfo # noqa: E402
+
+
+vehicle_whitelist = [
```
cclauss: allowlist or passlist?

Tools/scripts/param_check_unittests.py
```diff
@@ -0,0 +1,421 @@
+#!/usr/bin/env python
```
peterbarker:
```suggestion
#!/usr/bin/env python3
```

Tools/scripts/param_check_all.py
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env python
```
peterbarker:
```suggestion
#!/usr/bin/env python3
```

Tools/scripts/board_list.py
```diff
@@ -145,7 +145,7 @@ def read_hwdef(self, filepath):
                 ret += [line]
         return ret
 
-    def find_autobuild_boards(self, build_target=None):
+    def find_autobuild_boards(self, build_target=None, skiplist=None):
```
peterbarker:
```suggestion
    def find_autobuild_boards(self, build_target=None, skip : Container = None):
```

If we are going to move away from the instantly recognisable term we are currently using we should probably remove the redundant information in the variable name.  `skip` shouldn't even be a "list" here, and "skiplist" is definitely not an instantly recognisable term!

Tools/scripts/param_check_all.py
```diff
@@ -0,0 +1,221 @@
...
+import sys
+import glob
+from board_list import BoardList
+from param_check import get_metadata, check_file, SkippedChecks
+
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../Tools', 'autotest'))
+from pysim.vehicleinfo import VehicleInfo # noqa: E402
+
+
+vehicle_skiplist = [
```
peterbarker:
```suggestion
vehicles_to_skip = set([
```

Tools/scripts/board_list.py
robertlong13: Sounds good.

The whitelist/blacklist terminology I had in the first version of this was actually pretty confusing, which was part of the reason I switched it (in addition to Christian's review). In the context of my param checks, my list of vehicles to skip are a whitelist (accept them automatically without checking), but in the context of autobuild, it's a blacklist. Naming skip makes it a little clearer regardless of context.

robertlong13: @peterbarker You good with `Iterable[str]` instead of `Container`?

peterbarker: > @peterbarker You good with `Iterable[str]` instead of `Container`?

What's your thinking there?  I thought "Container" was the correct thing to use for something you might do "in" on.

robertlong13: We're actually checking wildcard matches, so we're not doing like `name in skip`, we're doing like `for pattern in skip: match(board, pattern)`. Since we're iterating over it, it felt right to use Iterable. I'm overthinking it though, as Container should still do the business just fine.

robertlong13: Fixed

cclauss:
Iterables do not need to have a `len()` while containers do.  Iterables handle containers plus generator expressions, such as functions that `yield` values instead of `returning` them.

The processing might be faster and require less code by using sets, but I doubt this is performance-critical code.

peterbarker (approved):
LGTM

peterbarker:
> Iterables do not need to have a `len()` while containers do. Iterables handle containers plus generator expressions, such as functions that `yield` values instead of `returning` them.

That all makes sense, thanks.

> The processing might be faster and require less code by using sets, but I doubt this is performance-critical code.

It looks like we could have gone with Iterable instead, allowing for generator expressions.... but I rather doubt that will be a problem for us, and we can fix it if it is :-)

PR #30270, opened 2025-06-07
Add pre-commit to GitHub Actions

cclauss:
In a GitHub Action, run `pre-commit run --all-files` on all pull requests.

Fixes: #24555 
* #24555

PR #30288, opened 2025-06-09
Tools: Add `# flake8: noqa` linter directive

cclauss:
Add the linter directive `# flake8: noqa` to the Python files in `/Tools` that currently fail `flake8` with the parameters in `.flake8`.  Add `flake8 Tools` to pre-commit so that we have ongoing flake8 testing of all Python files in `/Tools`.

As recommended at https://github.com/ArduPilot/ardupilot/pull/30279#discussion_r2134851607

### Before:
% `uv tool run flake8 Tools ` # Lots of errors

### After:
% `uv tool run flake8 Tools ` # No errors

@Ryanf55

PR #30290, opened 2025-06-09
Tools: Add pre-commit to GitHub Actions

cclauss:
A subset of #30270 that only does `chmod` on files in the Tools directory.
* #30270

Once this is landed and pre-commit is running on all pull requests, we can rebase #30270 to fix the `libraries` files.

@peterbarker


Ryanf55 (approved):

amilcarlucas (approved):
Not sure about Tools/autotest/param_metadata/xmlemit.py

All others look OK

.pre-commit-config.yaml
```diff
@@ -31,12 +31,9 @@ repos:
               )$
         -   id: check-added-large-files
         -   id: check-executables-have-shebangs
+            files: ^Tools/
```
khancyr, line 34: why restrict only on Tools ? we should fix permissions on all files and we got py script into libraries/ too (specially on AP_HAL_Chibios)

khancyr:
@peterbarker the command run is `pre-commit run --show-diff-on-failure --color=always ${{ inputs.extra_args }}` so it doesn't push thing on git , just show if there are issues

cclauss:
Three ci tools have been created by the author of pre-commit and flake8:
* https://github.com/pre-commit/action -- GitHub Action that does not allow/enable any auto fixes.
* https://pre-commit.ci -- Does not use GitHub Actions and allows/enables autofixes and periodic hook upgrade PRs.
* https://pre-commit.ci/lite.html -- GitHub Action that allows/enables auto fixes.

We want autofixes on localhost (the developer's machine), but not after the commit arrives in GitHub Actions.  Fortunately, the chosen Action does not allow/enable the latter.

A tool such as gh or https://github.com/stefanzweifel/git-auto-commit-action or https://github.com/peter-evans/create-pull-request is generally needed to add new changes to a commit that has been pushed to GitHub Actions.

.pre-commit-config.yaml
cclauss: https://github.com/ArduPilot/ardupilot/pull/30288#discussion_r2136752562

https://github.com/ArduPilot/ardupilot/pull/30270#issuecomment-2969707706

cclauss:
Closing.  The most interesting changes were hijacked into #30333

PR #30291, opened 2025-06-09
add dvl-gps integration tests

Williangalvani:
This checks switching from VICON (which uses the same API as DVLs) and GPS, and multiple configurations of them.

### DIFF
--- a/Tools/autotest/ardusub.py
+++ b/Tools/autotest/ardusub.py
@@ -1444,7 +1444,137 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
 
                 if alt_error > max_error:
                     max_error = alt_error
+        self.disarm_vehicle()
 
+    def IgnoreGPSDrift(self):
+        """ Test GPS-DVL integration. DVL (VISO) should be able to filter the GPS noise.
+        this tests switching over to VISO, running for a bit with no sensors, re-enabling viso,
+        and finally surfacing, where we expect the GPS position will be used to fix the drift
+        resulting of running with no positioning sensors."""
+
+        self.customise_SITL_commandline(["--serial5=sim:vicon:"])
+
+        # configure EKF to use external nav instead of GPS
+        self.set_parameters({
+            "EK3_SRC1_POSXY": 3,
+            "EK3_SRC1_VELXY": 6,
+            "EK3_SRC1_POSZ": 1,
+            "EK3_SRC1_VELZ": 0,
+
+            "VISO_TYPE": 1,
+            "SERIAL5_PROTOCOL": 1,
+            "SIM_VICON_TMASK": 8,  # send VISION_POSITION_DELTA
+            # "SIM_VICON_TMASK": 2,  # send VISION_SPEED_ESTIMATE
+            # "SIM_VICON_TMASK": 10,  # send VISION_SPEED_ESTIMATE AND VISION_POSITION_DELTA
+            # "EK3_VELNE_M_NSE": 0.001,
+
+            "GPS1_TYPE": 1,
+            "SIM_GPS1_DRFTALT": 3,
+            "SIM_GPS1_ACC": 0.3,
+            "SIM_GPS1_ENABLE": 1,
+            "SIM_GPS1_TYPE": 1,
+
+            "GPS2_TYPE": 0,
+
+            "EK3_POSNE_M_NSE": 100,
+        })
+        self.reboot_sitl()
+
+        self.wait_ready_to_arm()
+        self.arm_vehicle()
+        self.progress("Ensuring that VISO manages to filter the GPS noise")
+        self.change_mode('POSHOLD')
+        self.set_parameter("SIM_GPS1_HNSE", 2)
+        self.wait_location(
+            self.sim_location(),
+            location_source="SIMSTATE",
+            minimum_duration=10,
+            height_accuracy=10.0,
+            accuracy=1.0,
+            timeout=60.0,
+        )
+
+        self.progress("Diving to 10m")
+        self.set_rc(Joystick.Throttle, 1300)
+        self.wait_altitude(altitude_min=-10, altitude_max=-9, relative=False, timeout=60)
+        self.set_rc(Joystick.Throttle, 1500)
+
+        self.progress("Disabling GPS")
+        self.set_parameters({
+            "SIM_GPS1_ENABLE": 0,
+        })
+
+        self.progress("Driving GPS-less for 10m using VISO")
+        self.set_rc(Joystick.Forward, 1900)
+        self.wait_distance(10, timeout=60, location_source="SIMSTATE")
+        self.set_rc(Joystick.Forward, 1500)
+
+        self.progress("Surfacing")
+        self.set_rc(Joystick.Throttle, 1900)
+        self.wait_altitude(altitude_min=-0.5, altitude_max=1, relative=False, timeout=60)
+        self.set_rc(Joystick.Throttle, 1500)
+
+        self.progress("Re-enabling GPS, expecting position to converge within 2m")
+        self.set_parameters({
+            "SIM_GPS1_ENABLE": 1,
+        })
+        self.wait_location(self.mav.location(), minimum_duration=5, accuracy=2, timeout=60.0)
+
+        self.progress("Driving forward 10m with GPS+VISO")
+        self.set_rc(Joystick.Forward, 1900)
+        self.wait_distance(10, timeout=60, location_source="SIMSTATE")
+        self.set_rc(Joystick.Forward, 1500)
+
+        self.progress("Disabling GPS and diving to 10m for no-GPS+no-VISO test")
+        self.set_parameters({
+            "SIM_GPS1_ENABLE": 0,
+        })
+        self.set_rc(Joystick.Throttle, 1300)
+        self.wait_altitude(altitude_min=-10, altitude_max=-9, relative=False, timeout=60)
+        self.set_rc(Joystick.Throttle, 1500)
+
+        self.progress("Driving 10m on VISO only, verifying EKF has relative position")
+        self.set_rc(Joystick.Forward, 1900)
+        self.wait_distance(10, timeout=60, location_source="SIMSTATE")
+        self.wait_ekf_flags(
+            mavutil.mavlink.ESTIMATOR_POS_HORIZ_REL,
+            0,
+            timeout=15,
+        )
+
+        self.progress("Disabling VISO and waiting for EKF to lose position")
+        self.set_parameters({
+            "SIM_VICON_TMASK": 0,
+        })
+        self.wait_ekf_flags(
+            0,
+            mavutil.mavlink.ESTIMATOR_POS_HORIZ_REL | mavutil.mavlink.ESTIMATOR_POS_HORIZ_ABS,
+            timeout=15,
+        )
+
+        self.progress("Stopping to build position error while EKF has no aiding")
+        self.set_rc(Joystick.Forward, 1500)
+        self.delay_sim_time(10, reason="build position error while EKF has no aiding")
+
+        self.progress("Re-enabling VISO and driving forward")
+        self.set_rc(Joystick.Forward, 1900)
+        self.delay_sim_time(5, reason="drive forward with no aiding")
+        self.set_parameters({
+            "SIM_VICON_TMASK": 10,
+        })
+        self.delay_sim_time(10, reason="drive forward with VISO aiding")
+        self.set_rc(Joystick.Forward, 1500)
+
+        self.progress("Surfacing to simulate re-acquiring GPS")
+        self.set_rc(Joystick.Throttle, 1900)
+        self.wait_altitude(altitude_min=-0.5, altitude_max=1, relative=False, timeout=60)
+        self.set_rc(Joystick.Throttle, 1500)
+
+        self.progress("Re-enabling GPS, expecting position correction")
+        self.set_parameters({
+            "SIM_GPS1_ENABLE": 1,
+        })
+        self.wait_distance(10, accuracy=5, timeout=60)
         self.disarm_vehicle()
 
     def AutoTerrainRecover(self):
@@ -1523,6 +1653,7 @@ class AutoTestSub(vehicle_test_suite.TestSuite):
             self.UpsideDown,
             self.GuidedWP,
             self.AutoTerrainRecover,
+            self.IgnoreGPSDrift,
         ])
 
         return ret
--- a/libraries/SITL/SIM_GPS.cpp
+++ b/libraries/SITL/SIM_GPS.cpp
@@ -95,11 +95,11 @@ const AP_Param::GroupInfo SIM::GPSParms::var_info[] = {
     AP_GROUPINFO("POS",        9, GPSParms, pos_offset, 0),
 
     // @Param: NOISE
-    // @DisplayName: GPS Noise
-    // @Description: Amplitude of the GPS altitude error
+    // @DisplayName: GPS vertical noise
+    // @Description: Amplitude of the GPS vertical position error
     // @Units: m
     // @User: Advanced
-    AP_GROUPINFO("NOISE",     10, GPSParms, noise, 0),
+    AP_GROUPINFO("NOISE",     10, GPSParms, noise_vertical, 0),
 
     // @Param: LCKTIME
     // @DisplayName: GPS Lock Time
@@ -160,6 +160,13 @@ const AP_Param::GroupInfo SIM::GPSParms::var_info[] = {
     // @Values: 0:No GPS connected, 1:No Fix, 2:2D Fix, 3:3D Fix, 4:3D DGPS Fix, 5:3D RTK Float, 6:3D RTK Fixed
     // @User: Advanced
     AP_GROUPINFO("FIXTYPE", 19, GPSParms, fix_type, 6),
+
+    // @Param: HNSE
+    // @DisplayName: GPS horizontal noise
+    // @Description: Radius of the GPS horizontal position error in meters
+    // @Units: m
+    // @User: Advanced
+    AP_GROUPINFO("HNSE",       20, GPSParms, noise_horizontal, 0),
 
     AP_GROUPEND
 };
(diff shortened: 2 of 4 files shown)


Tools/autotest/ardusub.py
```diff
@@ -1133,6 +1160,150 @@ def SurfaceSensorless(self):
         self.disarm_vehicle()
         self.context_pop()
 
+    def wait_for_distance(self, distance, timeout=100):
```
peterbarker: Reimplement in terms of `wait_distance_to_location`

Tools/autotest/ardusub.py
```diff
@@ -1133,6 +1160,150 @@ def SurfaceSensorless(self):
...
+        original_location = self.mav.recv_match(type='GLOBAL_POSITION_INT', blocking=True)
```
peterbarker: avoid recv_match in new code, use assert_receive_message.  We also have a helper to get a location object somewhere...

Tools/autotest/ardusub.py
```diff
@@ -1133,6 +1160,150 @@ def SurfaceSensorless(self):
...
+            "GPS2_TYPE": 1,
+            "SIM_GPS2_DRFTALT": 3,
+            "SIM_GPS2_ACC": 0.3,
+            "SIM_GPS2_ENABLE": 1,
+            "SIM_GPS2_TYPE": 1,
+
+            "EK3_POSNE_M_NSE": 100,
+
+        })
+        self.disarm_vehicle()
```
peterbarker:
```suggestion
```

If you are armed here something seriously bad has happened

Tools/autotest/ardusub.py
```diff
@@ -1133,6 +1160,150 @@ def SurfaceSensorless(self):
...
+        self.wait_altitude(altitude_min=-10, altitude_max=-9, relative=False, timeout=60)
+        self.set_rc(Joystick.Throttle, 1500)
+
+        # drive for 10 meters
+        self.set_rc(Joystick.Forward, 1900)
+        self.wait_for_distance(10, timeout=60)
+        self.set_parameters({
+            "SIM_VICON_TMASK": 0,
+        })
+        self.progress("We should now be dead reckoning")
```
peterbarker: Can we check for that?  The EKF_STAUS should tell you.  This stops the test passing in future when it should fail...

peterbarker (changes requested):
Nice test :-)

Tools/autotest/ardusub.py
Williangalvani: Well, part of the issue is I cant get the damn flag to enable, hence the "_should_".
Probably some Sub edge case...
I'll try to go deeper (pun not intended) into the ekf to figure out why this is happening

Tools/autotest/ardusub.py
Williangalvani: so... our code REALLY likes GPS. [self.mav.location() just hangs if there's no GPS fix](https://github.com/ArduPilot/pymavlink/blob/master/mavutil.py#L824)

Tools/autotest/ardusub.py
```diff
@@ -1133,6 +1160,145 @@ def SurfaceSensorless(self):
...
+        """ Test GPS-DVL integration. DVL (VISO) should be able to filter the GPS noise.
+        this tests switching over to VISO, running for a bit with no sensors, re-enabling viso,
+        and finally surfacing, where we expect the GPS position will be used to fix the drift
+        resulting of running with no positioning sensors."""
+
+        self.customise_SITL_commandline(["--serial5=sim:vicon:"])
+
+        # configure EKF to use external nav instead of GPS
+        self.set_parameters({
+            "EK3_SRC1_POSXY": 3,
```
rmackay9, line 1459: is it intentional that the position comes from GPS?

libraries/SITL/SIM_GPS.cpp
```diff
@@ -545,8 +545,9 @@ void GPS::update()
     // Applying GPS glitch
     // Using first gps glitch
     Vector3f glitch_offsets = params.glitch;
-    d.latitude += glitch_offsets.x;
-    d.longitude += glitch_offsets.y;
+    // 1 meter at the equator is 1.113195e-5 degrees
+    d.latitude += glitch_offsets.x +  params.accuracy * 0.3 * sinf(now_ms * 0.0005f) * 1.113195e-5;
```
tridge: needs a tconst for this, default to disabled (eg. 0 tconst)

tridge (dismissed):

Tools/autotest/ardusub.py
rmackay9: as discussed on the call this is intentional so I'll mark this as resolved

Ryanf55:
Nice! If you want to improve the realism of the GPS wandering, I can get you in touch with Trimble who can connect you with references on how to improve this.

peterbarker:
@Williangalvani this one's still draft?  It's probably mergeable as-is, 'though?

peterbarker:
(well, commit messages need fixing...)

Tools/autotest/ardusub.py
```diff
@@ -155,6 +155,31 @@ def watch_altitude_maintained(self, delta=0.3, timeout=5.0):
                     "Altitude not maintained: want %.2f (+/- %.2f) got=%.2f" %
                     (previous_altitude, delta, m.alt))
 
+    def watch_position_maintained(self, delta=0.3, timeout=5.0):
+        """Watch and wait for the actual position to be maintained
+
+        Keyword Arguments:
+            delta {float} -- Maximum distance allows in meters
+            timeout {float} -- Timeout time in simulation seconds (default: {5.0})
```
hunt0r: This is a tiny detail and **completely optional**.
```suggestion
    def watch_position_maintained(self, delta=0.3, timeout=5.0):
        """Watch and wait for the actual position to be maintained

        Arguments:
            delta {float} -- Maximum distance allows in meters
            timeout {float} -- Timeout time in simulation seconds (default: {5.0})
```

If you meant these to be keyword args, use the asterisk to make them so?
```
    def watch_position_maintained(self, *, delta=0.3, timeout=5.0):
```

Ok to resolve anytime, with or without changes.

libraries/SITL/SIM_GPS.cpp
```diff
@@ -556,8 +562,10 @@ void GPS::update()
     // Applying GPS glitch
     // Using first gps glitch
     Vector3f glitch_offsets = params.glitch;
-    d.latitude += glitch_offsets.x;
-    d.longitude += glitch_offsets.y;
+    // 1 meter at the equator is 1.113195e-5 degrees
+    // TODO: scale longitude noise with latitude
+    d.latitude += glitch_offsets.x +  params.noise_radius * sinf(now_ms * 0.0005f) * 1.113195e-5;
+    d.longitude += glitch_offsets.y + params.noise_radius * cosf(now_ms * 0.0005f) * 1.113195e-5;
```
hunt0r: [optional] I was surprised by the multiply-by-0.0005 inside there. It looks like a "magic number".

From context, I am guessing this is an arbitrary scaling factor to achieve random-ish wandering behavior. (the wandering is a slowly evolving circle of radius noise_radius. It's period is... <checks AI> 12.6 sec.)

If I got that wrong, consider clarifying it somehow?
Assuming I got that right, it's proof _can_ be understood as is... but I recommend making it explicitly clear. A comment would work, but I like making the code self-evident by adding variables/names, such as:
```suggestion
    gps_wander_angle_rad = now_ms * 0.0005f  // choosing T=12.6 sec arbitrarily
    // TODO: scale longitude noise with latitude
    d.latitude += glitch_offsets.x +  params.noise_radius * sinf(gps_wander_angle_rad) * 1.113195e-5;
    d.longitude += glitch_offsets.y + params.noise_radius * cosf(gps_wander_angle_rad) * 1.113195e-5;
```

(By [optional] I mean it's fine to resolve this, with or without change, anytime.)

Tools/autotest/ardusub.py
```diff
@@ -155,6 +155,31 @@ def watch_altitude_maintained(self, delta=0.3, timeout=5.0):
...
+        self.progress('Position to be watched: (%f, %f)' % (previous_position.lat, previous_position.lng))
+        while True:
+            m = self.assert_receive_message('SIMSTATE')
+            if self.get_sim_time_cached() - tstart > timeout:
+                self.progress('Position hold done')
+                return
+            if self.get_distance_int(previous_position, m) > delta:
+                raise NotAchievedException(
+                    "Position not maintained: got %f meters too far from the previous position after %f seconds " %
+                    (self.get_distance_int(previous_position, m), self.get_sim_time_cached() - tstart))
```
hunt0r: [optional] As a matter of style, I think avoiding `while True` in favor of self-evidently terminating loops is good. Is this an equivalent refactor?
```suggestion
        while (self.get_sim_time_cached() - tstart) < timeout:
            m = self.assert_receive_message('SIMSTATE')
            if self.get_distance_int(previous_position, m) > delta:
                raise NotAchievedException(
                    "Position not maintained: got %f meters too far from the previous position after %f seconds " %
                    (self.get_distance_int(previous_position, m), self.get_sim_time_cached() - tstart))
                    
        self.progress('Position hold done')
```

(By [optional] I mean this is OK as is, feel free to resolve without reply if you prefer as-is.)

hunt0r: [optional] I generally find f-strings easier to understand. Consider this?
```suggestion
                raise NotAchievedException(
                    f"Position not maintained: got {self.get_distance_int(previous_position, m)} meters"
                     " too far from the previous position after {self.get_sim_time_cached() - tstart} seconds ")
```

(By [optional] I mean this is fine as-is, feel free to resolve without reply if you prefer it as-is.)

hunt0r: Oh, I just noticed my suggestion is very similar to a while loop you have below... except I did the negated one. If these could be made DRY via a helper function which is used 2x, that's my top choice. Otherwise, I support aligning one way or the other, and don't care which (return from inside loop or raise from inside loop).

hunt0r: I see other old-style string formats below. I won't comment those, but good to align both with the surrounding context and also across the PR for whatever is chosen.

peterbarker: We use the `while` as an idiom as the result of `get_sim_time_cached` is often used in the loop body too.

Tools/autotest/ardusub.py
peterbarker: Well... they're keyword arguments with or without the `*` :-)

libraries/SITL/SIM_GPS.cpp
peterbarker: This would be a good clarification - even better if it were to compile ;-)

Tools/autotest/ardusub.py
peterbarker: Using f-strings on code moving forward is definitely encouraged.

Tools/autotest/ardusub.py
peterbarker: Yeah, OK, that sucks.

Fix is here: https://github.com/ArduPilot/ardupilot/pull/32287

Tools/autotest/ardusub.py
```diff
@@ -378,7 +403,7 @@ def watch_true_distance_maintained(self, match_distance, delta=0.3, timeout=5.0,
 
         def get_true_distance():
             """Return the True distance from the simulated range finder"""
-            m_true = self.mav.recv_match(type='STATUSTEXT', blocking=True, timeout=3.0)
+            m_true = self.assert_receive_message(type='STATUSTEXT', timeout=3.0)
```
peterbarker:
```suggestion
            m_true = self.assert_receive_message('STATUSTEXT', timeout=3.0)
```

similarly elsewhere

Tools/autotest/ardusub.py
```diff
@@ -155,6 +155,31 @@ def watch_altitude_maintained(self, delta=0.3, timeout=5.0):
                     "Altitude not maintained: want %.2f (+/- %.2f) got=%.2f" %
                     (previous_altitude, delta, m.alt))
 
+    def watch_position_maintained(self, delta=0.3, timeout=5.0):
```
peterbarker: This should also not be needed after my PR fixing wait_location - it has a minimum_duration option

Tools/autotest/ardusub.py
peterbarker: Yet to be resolved....

peterbarker (changes requested):
One thing you can do to make the test a bit tighter is to install an instance of `ValidateIntPositionAgainstSimState` as a message hook.  So you can drive your sub around and every single position that comes in is checked to make sure it's aligned with the simulated position.  You have some "should glitch here" things, so you would have to just add it to a context you then pop off (or remove it with a call)

Tools/autotest/ardusub.py
hunt0r: Oops! You're absolutely right. I was thinking of "keyword _only_ args". I'll resolve this one myself without changes.

Tools/autotest/ardusub.py
hunt0r: Sounds like a prime time to use the walrus operator!?! 😄 

(I'm kidding: the walrus operator is rarely worth it, even if it would work. If the original `while True` is consistent with local style, good to resolve this comment without changes. And maybe even to switch the other one to using that same idiom. I'm happy with anything.)

libraries/SITL/SIM_GPS.cpp
hunt0r: 🤦 

Errr.. I mean... "Proper interpretation into valid syntax is left as an exercise for the reader." 🤓

Tools/autotest/ardusub.py
hunt0r: Update: I can't resolve it myself, sorry. Don't know why.

ES-Alexander: > I can't resolve it myself, sorry. Don't know why.

@hunt0r I've been running into this lately too, and discovered yesterday it's [an extremely frustrating (and old) GitHub limitation](https://github.com/orgs/community/discussions/8327). No repo write permissions or not PR author means no thread resolve permission (even if you started the comment thread, and/or have triage access), but the button is still there just to taunt you 🤦🏼‍♂️

hunt0r: Thank you!

Tools/autotest/ardusub.py
ES-Alexander: My personal preferred indefinite-`while` style is to use a string as the reason (e.g. `while "holding position":`), since non-empty strings are truthy, but not sure if that seems too magic for ArduPilot 🤷‍♂️

Tools/autotest/ardusub.py
Williangalvani: cool, I'll wait for that to go in. Thanks!

peterbarker: > wait for that

If you approve it I'll merge it :-)

libraries/SITL/SIM_GPS.cpp
hunt0r: @Williangalvani I noticed you applied this idea above... did you mean to apply it here?
(Totally fine if no! Just looks a little like a minor oversight.)

Tools/autotest/ardusub.py
peterbarker: It's been merged now

libraries/SITL/SIM_GPS.cpp
```diff
@@ -556,8 +563,10 @@ void GPS::update()
     // Applying GPS glitch
     // Using first gps glitch
     Vector3f glitch_offsets = params.glitch;
-    d.latitude += glitch_offsets.x;
-    d.longitude += glitch_offsets.y;
+    // 1 meter at the equator is 1.113195e-5 degrees
+    // TODO: scale longitude noise with latitude
+    d.latitude += glitch_offsets.x +  params.noise_radius * sinf(gps_wander_angle_rad) * 1.113195e-5;
+    d.longitude += glitch_offsets.y + params.noise_radius * cosf(gps_wander_angle_rad) * 1.113195e-5;
```
hunt0r: For fun, I solved the equations using a spherical Earth.
Also: https://www.tauday.com

Your (1.113195e-5 deg / m) implies an earth-radius R = (360 / 1.113195e-5 / tau) = 5146967 m.
(Estimates I see suggest your value is 20% low, in case you want to adjust.)

At an input latitude, the meters-per-degree-longitude = tau * R * cos(lat) / 360.
Its reciprocal is the number to plug into your code... but also I would structure a bit differently.

Here's my suggested lines of pseudocode, let me know if you like well enough to rewrite as compilable C++. I can push a commit to your PR.
```suggestion
    earth_radius_m = 6.371e6f
    lat_wander_m = params.noise_radius * sinf(gps_wander_angle_rad)
    lon_wander_m = params.noise_radius * cosf(gps_wander_angle_rad)
    deg_lat_per_m = 360 / (math.tau * earth_radius_m)
    highest_permissible_lat_deg = 89.0f
    cosine_of_lat = (absf(d.lat) < highest_permissible_lat_deg) ? cosd(d.latitude) : cosd(highest_permissible_lat_deg)
    deg_lon_per_m = deg_lat_per_m / cosine_of_lat
    d.latitude += glitch_offsets.x + lat_wander_m * deg_lat_per_m;
    d.longitude += glitch_offsets.y + lon_wander_m * deg_lon_per_m;
```

I did not carefully check this for mistakes. Happy to do that if you want me to contribute it to your effort, just ask!

hunt0r (commented):
Looking good! My suggestion is **completely optional**.

libraries/SITL/SIM_GPS.cpp
Williangalvani: I've integrated these changes

libraries/SITL/SIM_GPS.cpp
```diff
@@ -564,8 +572,14 @@ void GPS::update()
     // Applying GPS glitch
     // Using first gps glitch
     Vector3f glitch_offsets = params.glitch;
-    d.latitude += glitch_offsets.x;
-    d.longitude += glitch_offsets.y;
+    // apply a slow circular horizontal wander of radius noise_horizontal (in metres)
+    const double earth_rad_inv = 1.569612305760477e-7; // 1/authalic radius, rad/m
```
hunt0r: Minor: this can be `constexpr`. But I'm also ok with it as-is.

hunt0r (approved):

libraries/SITL/SIM_GPS.cpp
peterbarker: Maybe add something in definitions.h:
```
// radius of earth in meters
#define RADIUS_OF_EARTH 6378100
```

peterbarker (approved):
LGTM

PR #30296, opened 2025-06-10
Plane: remove use of roll_sensor from VTOL assist calculations

peterbarker:
... and augment the autotests to check pitch assist


IamPete1 (approved):
LGTM

Issue #28210, opened 2024-09-24
Flash savings ideas : features/drivers to disable/remove to save flash


rmackay9:
AP's flash size is constantly growing as new features and drivers are added but unfortunately the STM32 autopilot's CPUs flash size is not keeping up.  This means we need to disable or remove lesser used features from many boards.

This issue is to collect ideas (primarily from members of the core dev team) on features/drivers they think could be sacrificed.

Note we have a few ways to reduce flash:

1. move the feature/driver to be disabled by default thus forcing users to rely on the [custom build server](https://custom.ardupilot.org/)
2. migrate the feature to Lua forcing the user to load a Lua script
3. completely remove the feature or driver

Ryanf55:
* AP_GPS_GSOF once we have it available as an EAHRS.

timtuxworth:
Plane: suggesting all of these could be custom build server options only.
ICE
Precision Landing
Gimbals (except for Mount and Servo Gimbal)
VTX
Generator

amilcarlucas:
Winch
Exotic ESCs

peterbarker:
I've used some of these ideas in https://github.com/ArduPilot/ardupilot/pull/28123 - thanks!


edit: corrected URL -pb

peterbarker:
- [ ] bootloader flashing

Hwurzburg:
-[] Rangefinder on Plane

tridge:
@andyp1per suggested -Os or -O1 default on current 2M boards

peterbarker:
[disable-in-turn.csv](https://github.com/user-attachments/files/19495856/disable-in-turn.csv)

That's the output from `./Tools/autotest/test_build_options.py`'s "disable in turn" step.  It shows the number of bytes saved on CubeOrange when each feature is disabled.  This might help selection of things to boot from the standard build.

rmackay9:
Thanks @peterbarker.  I've extracted your list and sorted it for copter



<html>
<head>
</head>

<body>

Feature | copter
-- | --
AP_SCRIPTING_ENABLED | 182160
HAL_NAVEKF3_AVAILABLE | 171400
HAL_ENABLE_DRONECAN_DRIVERS | 96120
HAL_LOGGING_ENABLED | 95704
HAL_MOUNT_ENABLED | 67300
AP_FENCE_ENABLED | 50360
AP_RANGEFINDER_ENABLED | 48056
HAL_GYROFFT_ENABLED | 42308
[...]

PR #30298, opened 2025-06-10
Limit KDECAN and PiccoloCAN to custom build server / OEM setups / >2MB boards

peterbarker:
As suggested here: https://github.com/ardupilot/ardupilot/issues/28210

Oddly, KDECAN is checked in that issue but is still in the code for <=2MB boards.

Also changes PiccoloCAN to use standard defines


peterbarker:
```
----------  -----  ----------  ------  -----  -----  -----  -----
Board       blimp  bootloader  copter  heli   plane  rover  sub
CubeOrange  -8920  *           -8936   -8928  -9072  -8936  -9072
----------  -----  ----------  ------  -----  -----  -----  -----
```

rmackay9 (approved):
I also think that piccolo CAN is a rarely used feature but perhaps others will disagree
FYI @magicrub

peterbarker:
> I also think that piccolo CAN is a rarely used feature but perhaps others will disagree FYI @magicrub

I also think this is rare.

I think we do need a Plane maintainer to approve this one, thanks @rmackay9

peterbarker:
tridge objected at DevCall

peterbarker:
I've changed this PR to just be PiccoloCAN - split KDECAN into a different PR.

rmackay9:
We decided to discuss with Currawong to see if they're OK with this or not.  If they do agree we need to update the Currawong wiki pages to make it clear to users how to use the custom build server to enable PiccoloCAN

SchrodingersGat:
@rmackay9 we have no objections to this, as long as the docs are clear on how to optionally compile the feature in.

rmackay9:
Hi @SchrodingersGat, thanks for that.  I'll update the wiki shortly.  BTW, this won't apply to 4.6 so it will only be from 4.7 onwards.

rmackay9:
Here's the wiki PR to tell users that from AP-4.8 onwards they will need to use the custom build server https://github.com/ArduPilot/ardupilot_wiki/pull/7619

PR #30299, opened 2025-06-10
AP_Beacon: disable feature due to flash space constraints

peterbarker:
ArduPilot doesn't have room for every feature it can have, and hasn't for some time.


peterbarker:
```
CubeOrange,-8760,*,-13768,-13768,-8768,-13728,-8976
```

peterbarker:
> I think that sadly every time we remove a feature like this we need to update the wiki to tell users how to re-enable it with the custom build server

Sorry, yes, missed the label on this, and probably a few others... 

We need to call them out in the release notes as "pay attention here!", too.

PR #30300, opened 2025-06-10
RC_Channel: make RC dummy channel file-scoped to allow consting methods

peterbarker:
will allow more consts throughout code.

### DIFF
--- a/libraries/RC_Channel/RC_Channels.cpp
+++ b/libraries/RC_Channel/RC_Channels.cpp
@@ -309,7 +309,8 @@ void RC_Channels::set_aux_cached(RC_Channel::AUX_FUNC aux_fn, RC_Channel::AuxSwi
 // allow use of the pointer without checking it for null-ness.  If an
 // invalid option has been chosen somehow then the returned channel
 // will be a dummy channel.
-RC_Channel &RC_Channels::get_rcmap_channel_nonnull(uint8_t rcmap_number)
+static RC_Channel dummy_rcchannel;
+RC_Channel &RC_Channels::get_rcmap_channel_nonnull(uint8_t rcmap_number) const
 {
     RC_Channel *ret = RC_Channels::rc_channel(rcmap_number-1);
     if (ret != nullptr) {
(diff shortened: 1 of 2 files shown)


libraries/RC_Channel/RC_Channels.cpp
```diff
@@ -309,35 +309,36 @@ void RC_Channels::set_aux_cached(RC_Channel::AUX_FUNC aux_fn, RC_Channel::AuxSwi
 // allow use of the pointer without checking it for null-ness.  If an
 // invalid option has been chosen somehow then the returned channel
 // will be a dummy channel.
-RC_Channel &RC_Channels::get_rcmap_channel_nonnull(uint8_t rcmap_number)
+static RC_Channel dummy_rcchannel;
```
andyp1per, line 312: Can't you make this const as well?

peterbarker: > Can't you make this const as well?

I don't think so - the object has to be mutable as callers are expecting to call set methods and the like on it.

I just tried it now and I get:
```
../../libraries/RC_Channel/RC_Channels.cpp: In member function 'RC_Channel& RC_Channels::get_rcmap_channel_nonnull(uint8_t) const':
../../libraries/RC_Channel/RC_Channels.cpp:319:12: error: binding reference of type 'RC_Channel&' to 'const RC_Channel' discards qualifiers
  319 |     return dummy_rcchannel;
      |            ^~~~~~~~~~~~~~~
```

The `const` in the method is saying, "I don't modify the object you are calling me on".  You can hand back random non-const things for the caller to play with so long as they don't come from yourself (which is why I'm moving the object out in this PR).

peterbarker:
Well, this is weird:
```
CubeRedPrimary                 0      *           0       8                 -176   16     -16
Durandal                       0      *           32      16                0      -40    -16
Hitec-Airspeed      *                 *                                                   
KakuteH7-bdshot                8      *           -8      24                -16    -32    -8
MatekF405                      8      *           16      16                -464   8      -32
Pixhawk1-1M-bdshot             -112               56      32                -104   120    -80
f103-QiotekPeriph   *                 *                                                   
f303-Universal      *                 *                                                   
revo-mini                      8      *           0       -8                96     16     -32
skyviper-journey                                  -312                                    
skyviper-v2450                                    -16                                     
```

tridge (approved):

PR #30301, opened 2025-06-10
AP_DDS: change of case to option `--enable-DDS` in README and autotest

srmainwaring:
Couple of stragglers needing a case change to match the updated option.


Tools/autotest/sim_vehicle.py
```diff
@@ -1353,7 +1353,7 @@ def generate_frame_help():
                      type=str,
                      default="127.0.0.1",
                      help="IP address of the simulator. Defaults to localhost")
-group_sim.add_option("--enable-dds", action='store_true',
+group_sim.add_option("--enable-DDS", action='store_true',
```
peterbarker, line 1356: You know this is a sim_vehicle option and separate from the waf opt0on etc?

Changing this might break people's scripts they use to start sim_vehicle...

srmainwaring: @peterbarker  Did not appreciate that they were different (had thought it might just be passed through to the AP binary). 
Can remove the second commit if it's an issue, so maybe hang fire on the merge please

srmainwaring:
> People are trying the mixed-case version, so we may as well fix it

Ok - run with the change. It is bound to be confusing if they are different.

peterbarker:
> > People are trying the mixed-case version, so we may as well fix it
> 
> Ok - run with the change. It is bound to be confusing if they are different.

Noting the user on Discord, it's probably a good idea to be consistent.

srmainwaring:
Actually there is this with `sim_vehicle.py`

```bash
File "/Users/rhys/Code/ros2/jazzy/ros2-ardupilot/src/ardupilot/Tools/autotest/sim_vehicle.py", line 1589, in <module>
    do_build(cmd_opts, frame_infos)
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rhys/Code/ros2/jazzy/ros2-ardupilot/src/ardupilot/Tools/autotest/sim_vehicle.py", line 410, in do_build
    if opts.enable_dds:
       ^^^^^^^^^^^^^^^
AttributeError: 'Values' object has no attribute 'enable_dds'. Did you mean: 'enable_DDS'?
```

Whereas the `./waf build` is working when configured with `--enable-DDS`, so the `opts.enable_dds` needs changing too. Apologies - thought was automatic.

PR #30303, opened 2025-06-10
AP_GPS: UBLOX: `RELPOSNED` minor refactor and add high precision values

IamPete1:
A little bit of a drive by, this applies the high precision values to length and pos down. It also un-normalizes pos down if the normalized flag is set. This also re-orders things so `calculate_moving_base_yaw` doesn't have to do all the same calculations again.

### DIFF
--- a/libraries/AP_GPS/AP_GPS_UBLOX.cpp
+++ b/libraries/AP_GPS/AP_GPS_UBLOX.cpp
@@ -1641,18 +1641,34 @@ AP_GPS_UBLOX::_parse_gps(void)
             MB_Debug("RELPOSNED flags: %lx valid: %lx invalid: %lx\n", _buffer.relposned.flags, valid_mask, invalid_mask);
             if (((_buffer.relposned.flags & valid_mask) == valid_mask) &&
                 ((_buffer.relposned.flags & invalid_mask) == 0)) {
-                if (calculate_moving_base_yaw(_buffer.relposned.relPosHeading * 1e-5,
-                                          _buffer.relposned.relPosLength * 0.01,
-                                          _buffer.relposned.relPosD*0.01)) {
-                    state.have_gps_yaw_accuracy = true;
-                    state.gps_yaw_accuracy = _buffer.relposned.accHeading * 1e-5;
-                    _last_relposned_ms = AP_HAL::millis();
-                }
+
+                // Populte state
                 state.relPosHeading = _buffer.relposned.relPosHeading * 1e-5;
                 state.relPosLength  = _buffer.relposned.relPosLength * 0.01;
                 state.relPosD       = _buffer.relposned.relPosD * 0.01;
                 state.accHeading    = _buffer.relposned.accHeading * 1e-5;
                 state.relposheading_ts = AP_HAL::millis();
+
+                // Add on high presision components if avalable
+                if (_buffer.relposned.relPosHPLength > -100 && _buffer.relposned.relPosHPLength < 100) {
+                    // 0.1mm to m
+                    state.relPosLength += _buffer.relposned.relPosHPLength * 0.0001;
+                }
+                if (_buffer.relposned.relPosHPD > -100 && _buffer.relposned.relPosHPD < 100) {
+                    // 0.1mm to m
+                    state.relPosD += _buffer.relposned.relPosHPD * 0.0001;
+                }
+
+                // un-normalized pos down if necessary
+                if ((_buffer.relposned.flags & uint32_t(RELPOSNED::relPosNormalized)) != 0) {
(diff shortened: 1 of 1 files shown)


libraries/AP_GPS/AP_GPS_UBLOX.cpp
```diff
@@ -1641,18 +1641,34 @@ AP_GPS_UBLOX::_parse_gps(void)
...
                 state.relPosHeading = _buffer.relposned.relPosHeading * 1e-5;
                 state.relPosLength  = _buffer.relposned.relPosLength * 0.01;
                 state.relPosD       = _buffer.relposned.relPosD * 0.01;
                 state.accHeading    = _buffer.relposned.accHeading * 1e-5;
                 state.relposheading_ts = AP_HAL::millis();
+
+                // Add on high presision components if avalable
+                if (_buffer.relposned.relPosHPLength > -100 && _buffer.relposned.relPosHPLength < 100) {
+                    // 0.1mm to m
+                    state.relPosLength += _buffer.relposned.relPosHPLength * 0.0001;
```
IamPete1, line 1657: image

libraries/AP_GPS/AP_GPS_UBLOX.cpp
```diff
@@ -1641,18 +1641,34 @@ AP_GPS_UBLOX::_parse_gps(void)
...
+                }
+                if (_buffer.relposned.relPosHPD > -100 && _buffer.relposned.relPosHPD < 100) {
+                    // 0.1mm to m
+                    state.relPosD += _buffer.relposned.relPosHPD * 0.0001;
```
IamPete1, line 1661: image

libraries/AP_GPS/AP_GPS_UBLOX.cpp
```diff
@@ -1641,18 +1641,34 @@ AP_GPS_UBLOX::_parse_gps(void)
...
+                }
+
+                // un-normalized pos down if necessary
+                if ((_buffer.relposned.flags & uint32_t(RELPOSNED::relPosNormalized)) != 0) {
+                    state.relPosD *= state.relPosLength;
```
IamPete1: image

libraries/AP_GPS/AP_GPS_UBLOX.cpp
```diff
@@ -1641,18 +1641,34 @@ AP_GPS_UBLOX::_parse_gps(void)
...
+                    // 0.1mm to m
+                    state.relPosLength += _buffer.relposned.relPosHPLength * 0.0001;
+                }
+                if (_buffer.relposned.relPosHPD > -100 && _buffer.relposned.relPosHPD < 100) {
+                    // 0.1mm to m
+                    state.relPosD += _buffer.relposned.relPosHPD * 0.0001;
+                }
+
+                // un-normalized pos down if necessary
+                if ((_buffer.relposned.flags & uint32_t(RELPOSNED::relPosNormalized)) != 0) {
```
tridge: the datasheet doesn't say what it is normalized to. 1 meter? 1 cm? pi?
what testing have you done of this? do you have captures showing UBX data where it has set the normalization flag?

tridge (changes requested):
needs testing and we need to know how the normalization is done
I'd also love to know how you found this!

libraries/AP_GPS/AP_GPS_UBLOX.cpp
IamPete1: No testing, just based on reading the datasheet. I have had a deeper look and can't find any more info. I think this is what they probably mean, but in the absence of anything concrete I guess we could also just refuse to accept the message if the flag is set. Just ignoring the flag as we do currently seems like a bad approach.

IamPete1: I could drop this change out of this PR and move it to another, so this would just be the addition of the high precision fields.

IamPete1: Hum: https://content.u-blox.com/sites/default/files/documents/ZED-F9P-MovingBase_AppNote_UBX-19009093.pdf

page 20.

image

https://www.u-blox.com/sites/default/files/ZED-F9H_IntegrationManual_UBX-19030120.pdf

image

IamPete1: It seems like we would need a F9H and a F9P pair to get this behavior. I don't have the hardware to test that combo.

IamPete1: Nobody has heard of the F9H, so we don't need to support it. I have removed this change. We may want to check that this flag is not set in the future.

PR #30306, opened 2025-06-10
AP_GPS: cope with F9P configured to never send MSG_STATUS

tridge:
ensure we don't get stuck with RATE_STATUS as unconfigured


andyp1per (approved):
Confirmed by the vendor

rmackay9:

PR #21620, opened 2022-09-02
Plane: postition control QPOS1 and QPOS2

IamPete1:
This adds position control to the QPOS1 and QPOS2 VTOL approach states. It attempts to fly a straight line from the entry of QPOS1 to the destination. In QPOS1 the target position is simply the closest point on that line. This allows to to "free wheel" along the axis of the line while trying to minimize cross track. If the vehicle ever goes beyond the destination the target point becomes "locked" to the destination, this means that the postilion controller helps slow down in the case of overshoot. 

Once in QPOS2 that target point is advances along the line at WP speed until it reaches the destination, where it is again locked. The threshold for the switch to QPOS2 is now purely based on ground speed. 

In both QPOS1 and QPOS2 the nose of the vehicle is pointed into the apparent wind.

Some testing with 10m/s cross wind:
PR Normal approach:
image

Master Normal approach:
image

So for normal approach we see a very little difference, maybe slightly less overshoot.

PR "too close" approach:
image

Master "too close":
image

A significant improvement, PR stops much sooner thanks to the additional "pull" of the target potion.

We still suffer somewhat when the vehicle is not traveling towards the destination when entering QPOS1. This PR apply the deceleration against the direction between the start of QPOS1 and the destination. Master always applies it away from the destination.

I think this is a improvement, and it simplifys the code a little. I was tempted to combine QPOS1 and QPOS2, there is now very little difference. There is still more improvements possible especially for "too close" approaches.

Needs lots of edge case testing and testing with ship landing.

PR #30309, opened 2025-06-11
Plane: improve POSITION1 velocity control

robertlong13:
For a while, I've noticed an annoying behavior during POS1. The aircraft initially can't decelerate fast enough, but then as it gets slower, it massively undershoots QPOS.TSpd. This can be seen by the aircraft pitching up to max and then suddenly pitching back down below 0 right after. I can reproduce this in RealFlight using the Titan Cobra, but I've encountered this in real life on Volanti and Ottano.

backtransition_before

The POS1 controller target velocity is shown in purple. The actual velocity, in orange, is doing a terrible job following it. We can see that the undershoot is actually commanded by the underlying position controller's target velocity in brown (multiple brown lines, I know). The position controller's target velocity goes lower because we start winding up a position error (blue/brown lines). The pitch angle (green/brown) goes from 27 degrees pitched up to 4 degrees pitched down back-to-back.

Essentially what the controller is saying is "I'm at the right speed to reach the point X meters away, but I got here too early and I should slow down even more to get my position-vs-time curve to match up with expectations". It does a great job of getting the actual position to come back and follow that curve, but that's entirely unnecessary.

Instead, we should clear the position error when we're in POS1. The outer controller just wants the inner controller to handle velocity and acceleration.

Here's the same plot after this PR. I think this is much nicer. Now the actual velocity follows QPOS.TSpd quite well. And our attitude commands are much more reasonable. We only hit 17 degrees pitch up, and after, we relax to 3 degrees pitch up, and we don't drop the nose until we switch to POS2 (I could try to knock that out too, but I'm fine with that).

backtransition_after

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -2634,6 +2634,8 @@ void QuadPlane::vtol_position_controller(void)
 
         // use input shaping and abide by accel and jerk limits
         pos_control->input_vel_accel_NE_m(target_speed_ne_ms, target_accel_ne_mss);
+        // During POS1, we only want to control velocity and acceleration
+        pos_control->stop_pos_NE_stabilisation();
 
         // run horizontal velocity controller
         run_xy_controller(MAX(approach_accel_mss, transition_decel_mss)*1.5);


robertlong13:
[backtransition_before_and_after.zip](https://github.com/user-attachments/files/20683273/backtransition_before_and_after.zip)
Here are the logs from RealFlight, flown on the Titan Cobra at RF_AirStadium.

robertlong13:
I recently learned about https://github.com/ArduPilot/ardupilot/pull/21620. That PR also fixes the issue I'm trying to solve here (among other things). It's in need of an update (it's based on 4.4dev), but ~~I can confirm it stops the controller from trying to hold a distance-vs-time curve and instead correctly follows a speed-vs-distance curve~~ (update: only works on 4.4; rebasing to 4.7 broke it somehow):

newplot (28)

[backtransition_21620_4.4dev.zip](https://github.com/user-attachments/files/23177846/backtransition_21620_4.4dev.zip)

If Pete's planning to revive that that PR and get it in before 4.7, then we don't need mine. Otherwise, I think my smaller PR is worth doing in the meantime.

ArduPlane/quadplane.cpp
```diff
@@ -2634,6 +2634,8 @@ void QuadPlane::vtol_position_controller(void)
 
         // use input shaping and abide by accel and jerk limits
         pos_control->input_vel_accel_NE_m(target_speed_ne_ms, target_accel_ne_mss);
+        // During POS1, we only want to control velocity and acceleration
+        pos_control->stop_pos_NE_stabilisation();
```
IamPete1, line 2638: @lthall Do you see any problem with us using this combination of methods? It does seem a little odd that were calling `input_vel_accel_NE_m` which does integrate position and then zeroing the error in `stop_pos_NE_stabilisation` I wonder if we would be better to add a method that does not integrate position in the fist place.

lthall: No, I made this function available for this specific purpose. Here is where it is used in the same way in Copter Guided:

https://github.com/ArduPilot/ardupilot/blob/6581d3d0a8567c4d9d43a2d7dfe8023858fba1b5/ArduCopter/mode_guided.cpp#L840

IamPete1: Great, Thanks

timtuxworth (approved):
Tested in RealFlight with #31401 - a big improvement.

IamPete1 (approved):

PR #30310, opened 2025-06-11
Prevent firmware erasure on boards using flash for parameter storage

andyp1per:
Both of these boards have been reported as losing firmware in the field on 4.6. This reverts the issue.


tridge:
@andyp1per I've updated this PR to only rebuild the impacted boards (ie. H7 with param storage in flash)

PR #30312, opened 2025-06-11
Adding Astra Battery Monitor

anhepl:
Astra Smart Battery - Designed, Developed and Manufactured by HASTIN ENERGY PRIVATE LIMITED, INDIA.
The inhouse designed and developed **Battery Management System** is embedded within the Battery Pack Casing itself. Unlike other SMBus based battery monitors, Astra BMS is capable of sending 100+ parameters which can be useful for the pilots and fleet managers as well. 
Visit www.hastinenergy.com for more information.


Hwurzburg (changes requested):
needs build_options addition

anhepl:
Hello 
> needs build_options addition

Since it is similar to smbus_generic and uses the same SMBus files as reference, I don't think we need to update the builds_options addition.
I'm an amateur, please let me know if I'm wrong.

Also, thanks for the input. I'll update the wiki asap

Hwurzburg:
I beleive that if it has a specific enable name, it should be in build options to allow its removal or addition instead of always being included for any SMBUS batt mon setup

anhepl:
> I beleive that if it has a specific enable name, it should be in build options to allow its removal or addition instead of always being included for any SMBUS batt mon setup

I think we have already done that.
In AP_BattMonitor.cpp
image

In AP_BattMonitor_config.h
image

I have taken references of NeoDesign and made the required changes to the above files.
If you could specifically point out which file needs additional changes, that would be appreciated.

Thanks

mainak027 (approved):

Hwurzburg:
no, as it is now, anytime SMBUS monitors are enabled, its enabled....adding an entry in build_options. py allows a user to add smbus monitors but specifically remove the astra for space...conversely an F405 user tight on space cant add an smbus monitor without adding them all.....the generic/neodesign and SUI monitors need additions in the build_options but they arent there currently due to historical reasons(build server was done after they were added), but new ones should conform and have the ability to remove or add by user for space savings...this driver adds 700 bytes anytime SBUS monitors are enabled...at some point I may try to fix the others as build options myself

anhepl:
Understood, thanks for explaining in detail. I have made changes to build_option.py as recommended by you. 
Please review and if it is good to go, I myself can fix the others and contribute to the community.

Hwurzburg:
looks good to me...asked @peterbarker to review....will mark for devcall

Hwurzburg (approved):

Hwurzburg:
@anhepl if more changes are to be made, let me know I will remove the DEVCALL tag...I has assumed this was tested and finished

anhepl:
@Hwurzburg 
Hi Henry, 
Thanks for approving the changes at the earliest. The code is tested and finished. I have also updated the wiki and raised a pull request in the ardupilot_wiki repo.

IamPete1 (changes requested):
This seems to be a exact copy of the SMBus generic backend with the addition of a call to read `read_soh`? Is there any reason we can add state of heath support to all SMBus backends?

anhepl:
> This seems to be a exact copy of the SMBus generic backend with the addition of a call to read `read_soh`? Is there any reason we can add state of heath support to all SMBus backends?

Hi Peter, we are developing the Astra Battery Monitor library specifically for the Astra Smart Batteries. You are correct. This code is being referenced from SMBus-Generic. Instead of modifying the actual generic code just to serve our batteries is not right. It might take up more space as well for people who are not using our batteries. We are yet to add more code to this library to retrieve lot more data for custom GCS/ LUA scripts. Our SMART BATTERY can provide all the information that can be retrieved using the **_BATTERY_INFO (372)_** MAVLINK packet. As @Hwurzburg guided, we have added the option to select Astra Battery Monitor during firmware build. So it is not going to affect other users.

peterbarker:
Please subclass Generic - see the Maxell driver as an example.

peterbarker (changes requested):
subclass Generic

anhepl:
@peterbarker We understand that Astra Battery Monitor library has same code as SMBus-Generic. Unlike Maxell, our battery monitor can provide lot more parameter which are part of **_BATTERY_INFO (372) MAVLINK packet_**
As of now, the code contains parameters which are part of **_BATTERY_STATUS MAVLINK packet_** which can be extracted with the code that SMBus_Generic has and the backend works successfully with no bugs.
For the other part where we can extract parameters of **_BATTERY_INFO_**, the code backend is still under development phase. Once this development is completed, the code will be different than that of SMBus_Generic. So we believe, it won't make sense to refer it like Maxell driver library does.
Astra Smart Battery is a commercial product unlike other DIY products.

rmackay9:
Hi @anhepl,

It's a golden rule in ArduPilot that we don't add code that's not used.  Similarly we won't add duplicate code.  Once you've got the complete driver we can revisit but for now, we can't accept this PR as it is I'm afraid.

Hwurzburg:
I added a WIP label since this will need significant differentiation to be considered in the future...

PR #30317, opened 2025-06-11
Rename HAL_PROBE_EXTERNAL_I2C_COMPASSES to AP_COMPASS_PROBING_ENABLED

peterbarker:
This is a no-compiler-output change.

This PR also force it to have a value, not just be defined.

```
pbarker@threads:~/rc/ardupilot$ ~/rc/ardupilot/Tools/scripts/filter_size_compare_branches_csv.py --hide-binary-identical /tmp/some.csv  
-----  ---------  -----  ----------  ------  ----  ----------  -----  -----  ---
Board  AP_Periph  blimp  bootloader  copter  heli  iofirmware  plane  rover  sub
-----  ---------  -----  ----------  ------  ----  ----------  -----  -----  ---
pbarker@threads:~/rc/ardupilot$ 
```


peterbarker:
(no compiler output change when hwdef changes are excluded from the binary, I should say!)

andyp1per (changes requested):
Please can we delay on merging this. This makes all new boards defined in master not work in 4.6 and there are still a few boards in the works for 4.6, I think we should wait a couple more cycles before doing this.

peterbarker:
> Please can we delay on merging this. This makes all new boards defined in master not work in 4.6 and there are still a few boards in the works for 4.6, I think we should wait a couple more cycles before doing this.

Not repeating the same pain as the LED changes would be nice!

Could you evaluate ~~https://github.com/ArduPilot/ardupilot/compare/master...peterbarker:ardupilot:pr/catch-backport-problem?expand=1~~ https://github.com/ArduPilot/ardupilot/pull/30324 , please - as a patch to go back into 4.6 to avoid backporting errors.

andyp1per:
Not sure the other patch helps that much.

libraries/AP_Compass/AP_Compass_config.h
```diff
@@ -127,3 +127,14 @@
 #ifndef AP_COMPASS_RM3100_ENABLED
 #define AP_COMPASS_RM3100_ENABLED AP_COMPASS_I2C_BACKEND_DEFAULT_ENABLED
 #endif
+
+// compatability code added for ArduPilot-4.7, June 2025 to make life
+// easier for external hwdef owners
+#ifdef HAL_PROBE_EXTERNAL_I2C_COMPASSES
+#error HAL_PROBE_EXTERNAL_I2C_COMPASSES is no longer used; try defining AP_COMPASS_PROBING_ENABLED to 1 instead
```
tpwrules, line 134: Can we use the whiz-bang new [list](https://github.com/ArduPilot/ardupilot/blob/71e9765a658ce06363975eb229775e6234c6a4ac/libraries/AP_HAL/hwdef/scripts/hwdef.py#L244) ? Then we can be sure there are no misses.

tpwrules:
I think the above would be nice but it's not mandatory. I added it to the list and configured all and there were no errors.

PR #30318, opened 2025-06-11
hwdef: fixed defaults.parm for missing RELANn_FUNCTION

tridge:
when we define a relay pin in hwdef.dat we should default the function to RELAY


andyp1per:
It's fine with me, but I thought I had specifically asked @IamPete1 this question and he had concluded that it was not necessary?

Hwurzburg:
I carefully walked the code and its true....if the pin   is defined in hwdef, even though the default for RELAY_FUNCTION is zero and **shown** as such, it gets forced to RELAY_FUNCTION = 1 during init.
so this is unnecessary as @andyp1per states....sorry for the red herring

PR #30319, opened 2025-06-11
AP_BoardConfig: Delete the comments

muramura:
Delete this comment-out since it will never be enabled.

In other files, it has been deleted.
The handling is inconsistent.


peterbarker:
I'm leaving these in deliberately so they don't get re-assigned.

PR #30320, opened 2025-06-12
hwdef: reduce image sizes

tridge:
some of the images were much larger than they need to be
reduces the hwdef directory size by 90 MByte


Hwurzburg (approved):

tpwrules:
I'd prefer not to introduce JPEG artifacts on the diagrams.

Also the old files will be around forever, so this increases the repo size but decreases the checkout size.

peterbarker:
> I'd prefer not to introduce JPEG artifacts on the diagrams.
> 
> Also the old files will be around forever, so this increases the repo size but decreases the checkout size.

A strategic error putting this data into the repo, I think.  Should we change our strategy?

peterbarker (changes requested):
Thomas raised questions.

tridge:
> A strategic error putting this data into the repo, I think. Should we change our strategy?

maybe a CI test checking max image size, max 2MByte?
note that checkouts being smaller is still nice.

tridge:
also note that reducing the large files makes using the images nicer (faster download). We have one image that is 10340 x 14999, which is just silly

tpwrules:
Who or what uses these images? If they are for the wiki should we move them over there?

Hwurzburg:
unfortunately, many do not read the wiki...even so, it takes some time from the merge of a new board to get a wiki page done...especially if I am on vacation (smileyface emoticon here) so without the diagrams a new user of the board would be lost

tridge:
> Who or what uses these images? If they are for the wiki should we move them over there?

I certainly use them, I look in the hwdef directory for up to date pinout information

tpwrules:
I'd like to see a bit more of a human touch, i.e. resizing, cropping, etc. instead of just turning down JPEG quality to meet a threshold. I can do that if you like.

There are also PNG optimizers too, but it may not be worth the churn on smaller images. I agree the 20MB ones are absurd and adding an image 5% of their size would not be much different in the scheme of things. But all the little ones it's probably not worth.

davidbuzz:
i think right now, the best thing we can do is (a) merge this pr and (b) encourage --depth=1 use for casual user/s, and everything else we decide to do can be done after both thise steps ( eg CI )

tpwrules (changes requested):
`libraries/AP_HAL_ChibiOS/hwdef/ZeroOneX6_Air/ZeroOneX6_Air.jpg` got changed, it looks like there was an existing .jpg your script overwrote with the converted .png

davidbuzz:
discussed at dev call.

tpwrules (changes requested):
`libraries/AP_HAL_ChibiOS/hwdef/ZeroOneX6_Air/ZeroOneX6_Air.jpg` got changed, it looks like there was an existing .jpg your script overwrote with the converted .png

tpwrules:
Aside from that I think I'm okay... we can use CI to do better in the future. We should point out to contributors to resize images too. It looks like the incremental size to the repo is some 18M which is fine.

tpwrules (changes requested):
@davidbuzz the image is still broken, please stop dismissing my review

For the third time:
`libraries/AP_HAL_ChibiOS/hwdef/ZeroOneX6_Air/ZeroOneX6_Air.jpg` got changed, it looks like there was an existing .jpg your script overwrote with the converted .png

Hwurzburg:
> @davidbuzz the image is still broken, please stop dismissing my review
> 
> For the third time: `libraries/AP_HAL_ChibiOS/hwdef/ZeroOneX6_Air/ZeroOneX6_Air.jpg` got changed, it looks like there was an existing .jpg your script overwrote with the converted .png

FYI..that png should really be deleted, it was never used, turns out...only the jpg..

PR #30324, opened 2025-06-12
AP_Compass: add ifdef to catch backport error (AP_COMPASS_PROBING_ENABLED)

peterbarker:


peterbarker:
Partner to https://github.com/ArduPilot/ardupilot/pull/30317

andyp1per:
I mean I'm not sure this cuts it does it? Cherry-picks are not going to pick this up unless we build all boards - which we don't do in CI.

peterbarker:
> I mean I'm not sure this cuts it does it? Cherry-picks are not going to pick this up unless we build all boards - which we don't do in CI.

Are you seriously saying you would create a PR by cherry-picking the board hwdef back and then *not even trying to build it*?!

andyp1per:
> > I mean I'm not sure this cuts it does it? Cherry-picks are not going to pick this up unless we build all boards - which we don't do in CI.
> 
> Are you seriously saying you would create a PR by cherry-picking the board hwdef back and then _not even trying to build it_?!

It happens all the time

PR #30330, opened 2025-06-12
AP_Camera_Servo: allow setting absolute zoom/focus endpoints via mavlink

Williangalvani:


libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -61,19 +61,31 @@ bool AP_Camera_Servo::trigger_pic()
...
-    if (zoom_type == ZoomType::RATE) {
-        zoom_current_rate = zoom_value;
-        return true;
+    switch (zoom_type) {
+        case ZoomType::RATE:
+            zoom_current_rate = zoom_value;
+            return true;
+        case ZoomType::PCT:
+            // expects to receive a value between 0 and 100
+            SRV_Channels::set_output_scaled(SRV_Channel::k_cam_zoom, constrain_float(zoom_value * 10, 0, 1000));
```
ES-Alexander: This is unclear - if it's expecting a 0-100 value, why are you providing a 0-1000 constraint?

libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -61,19 +61,31 @@ bool AP_Camera_Servo::trigger_pic()
...
-    if (focus_type == FocusType::RATE) {
-        focus_current_rate = focus_value;
-        return SetFocusResult::ACCEPTED;
+    switch (focus_type) {
+        case FocusType::RATE:
+            focus_current_rate = focus_value;
+            return SetFocusResult::ACCEPTED;
+        case FocusType::PCT:
+            // expects to receive a value between 0 and 100
+            SRV_Channels::set_output_scaled(SRV_Channel::k_cam_focus, constrain_float(focus_value * 10, 0, 1000));
```
ES-Alexander: Same as above.

libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -61,19 +61,31 @@ bool AP_Camera_Servo::trigger_pic()
...
+            return SetFocusResult::ACCEPTED;
+        case FocusType::AUTO:
+            return SetFocusResult::UNSUPPORTED;
     }
     return SetFocusResult::UNSUPPORTED;
```
ES-Alexander, line 92:
```suggestion
        case FocusType::AUTO:
        default:
            return SetFocusResult::UNSUPPORTED;
    }
```

libraries/AP_Camera/AP_Camera_Servo.cpp
```diff
@@ -61,19 +61,31 @@ bool AP_Camera_Servo::trigger_pic()
...
+    switch (zoom_type) {
+        case ZoomType::RATE:
+            zoom_current_rate = zoom_value;
+            return true;
+        case ZoomType::PCT:
+            // expects to receive a value between 0 and 100
+            SRV_Channels::set_output_scaled(SRV_Channel::k_cam_zoom, constrain_float(zoom_value * 10, 0, 1000));
+            return true;
     }
     return false;
```
ES-Alexander, line 74:
```suggestion
        default:
            return false;
    }
```
Not sure what ArduPilot's convention is for function-ending `switch-case` blocks, but I think it's potentially clearer to have the final result handled in `default` instead of outside the block 🤷‍♂️

ES-Alexander (commented):
I know this is still a draft, but I had a look while looking for something else, so figured I'd share some thoughts 🤷‍♂️ :-)

libraries/AP_Camera/AP_Camera_Servo.cpp
nexton-winjeel: I can't speak for the dev team, but I prefer not to have a `default` and to handle unexpected values outside of the `switch`. That way, when a new `enum` value is added, the compiler will warn that not all cases are handled.

Williangalvani: Yes, I believe that is the general reasoning on Ardupilot.
Whenever a new enum entry is added, the compiler will make sure we don't forget of handling it anywhere.

libraries/AP_Camera/AP_Camera_Servo.cpp
Williangalvani: updated the comment

PR #30335, opened 2025-06-13
Enable dependencies for features enabled with `./waf --enable-FOO`

peterbarker:
image

Also removes the magic enabling of the dependency for PPP


libraries/AP_HAL_ChibiOS/hwdef/CubeOrangePlus/hwdef.dat
```diff
@@ -100,4 +100,5 @@ BOARD_VALIDATE $CHECK_IMU0_PRESENT $CHECK_IMU1_PRESENT $CHECK_IMU2_PRESENT $CHEC
 env BUILD_ABIN True
 define HAL_INS_HIGHRES_SAMPLE 6
 
+define AP_NETWORKING_ENABLED 1
```
tridge, line 103: this breaks --disable-networking

PR #30339, opened 2025-06-13
Add New Target DAKEFPVH743

engineer-dakefpv:


mop98915:
I have the board in hands if any testing is required

engineer-dakefpv:
As with DAKEFPVH743PRO, it has been revised in the light of comments.
hwdef.dat:
1.UART5 use DMA for RCIN.
2.USART2,3 NODMA.
3.USART6,7 use DMA.

defaults.parm:
1.setup for LEDs on chan14.

WiringDiagramTop.png:
1.updated the correct wiring diagram.

Hwurzburg:
Exactly how does this differ from the PRO version?

andyp1per:
Yeah as Henry says - if the two boards are sufficiently similar then one should be based on the other or they might even share the same firmware

engineer-dakefpv:
> Exactly how does this differ from the PRO version?

The H743 was designed before, but when applying for INAV firmware, the PD15 (Motor 8) pin could not initiate a DMA request, resulting in no PWM output on the INAV. So we reassigned the timer, designed the H743PRO, and added support for the CAN protocol. However, a lot of H743 flight controllers have been shipped before, and many customers have been asking our H743 to provide support for Ardupilot.

andyp1per:
OK so I think you do need separate definitions if the motor setup is different, but you should be able to base one substantially on the other

engineer-dakefpv:
Request review. @Hwurzburg

Hwurzburg (approved):

andyp1per (approved):

mop98915:
In mission planner occours error when flashing

PR #30346, opened 2025-06-13
AP_Scripting: loiter to autoland altitude

landswellsong:
Based on similar functionality in AUTOLAND mode with some Lua limitations. Meant to be used with stable 4.6 builds (can't do a PR versus that branch because the script history appeared in 4.7). 

Verified with 4.6.0 SITL, seems to work alright, with some differences:
- Since the mission is pre-generated at takeoff, there is no way of telling if we are left or right of the approach path, defaulting to right/cw loiter.
- Apparently no Lua binding for L1's `loiter_radius()` method so we aren't exiting at the final approach WP exactly, just close to it.

image

### DIFF
--- a/libraries/AP_Scripting/applets/UniversalAutoLand.lua
+++ b/libraries/AP_Scripting/applets/UniversalAutoLand.lua
@@ -48,6 +48,7 @@ NAV_WAYPOINT = 16
 NAV_TAKEOFF = 22
 NAV_LAND = 21
 DO_LAND_START = 189
+LOITER_TO_ALT = 31
 
 TAKEOFF_PITCH = 15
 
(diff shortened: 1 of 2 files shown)


Hwurzburg (changes requested):
aside from the scripting failures, we should change the param prefix to ALAND so as to not conflict with the AUTOLAND params in 4.7

timtuxworth:
> aside from the scripting failures, we should change the param prefix to ALAND so as to not conflict with the AUTOLAND params in 4.7

I'm suggesting that we prefix scripts with Z (or ZZ) with a two character code as a prefix so ZAL_ or ZZAL_

Hwurzburg:
@landswellsong we decided that an ALAND prefix was best....make that change and fix CI fails and we will merge....thanks

peterbarker:
Ping @landswellsong - general agreement this can go in, a small amount of work needed...

landswellsong:
Ok I'll take care of it and report

PR #30348, opened 2025-06-14
AP_HAL: hwdef: scripts: enforce use of board_types.txt - nix numeric IDs

peterbarker:
```
    f.write('#define APJ_BOARD_ID %s\n' % self.get_numeric_board_id())
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pbarker/rc/ardupilot/libraries/AP_HAL_ChibiOS/hwdef/scripts/../../../../libraries/AP_HAL/hwdef/scripts/hwdef.py", line 82, in get_numeric_board_id
    raise ValueError("Numeric board ID found.  Your APJ_BOARD_ID must not use a number.  Change the number to be the name of the board used in Tools/AP_Bootloader/board_types.txt")  # noqa:E501
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Numeric board ID found.  Your APJ_BOARD_ID must not use a number.  Change the number to be the name of the board used in Tools/AP_Bootloader/board_types.txt
Failed to process hwdef.dat ret=1
(complete log in /home/pbarker/rc/ardupilot/build/config.log)
pbarker@crun:~/rc/ardupilot(pr/no-numeric-board-ids)$ 
```


peterbarker:
This is a no-compiler-output change PR.

Except for CubeRedSecondary-IO, in which scripting has been removed.

libraries/AP_HAL_ChibiOS/hwdef/iomcu/hwdef.inc
```diff
@@ -5,7 +5,7 @@
 FLASH_RESERVE_START_KB 4
 
 # board ID. See Tools/AP_Bootloader/board_types.txt
-APJ_BOARD_ID 3
+APJ_BOARD_ID EXT_HW_RADIOLINK_MINI_PIX
```
peterbarker: We definitely won't be enforcing this value anywhere.

We should probably fix it to be an *actual*, *distinct* board ID, potentially also including different MCU types as different boards (and fixing CubeOrangeSecondary-IO to not use the same ID as an f103-based IOMCU!

Hwurzburg (approved):

libraries/AP_HAL/hwdef/scripts/hwdef.py
```diff
@@ -79,7 +79,7 @@ def get_numeric_board_id(self):
         number via board_list.txt'''
         some_id = self.get_config('APJ_BOARD_ID')
         if some_id.isnumeric():
-            return some_id
+            raise ValueError(f"Numeric board ID found ({some_id}).  Your APJ_BOARD_ID must not use a number.  Change the number to be the name of the board used in Tools/AP_Bootloader/board_types.txt")  # noqa:E501
```
tridge: enforce only in CI would be good

libraries/AP_HAL_ChibiOS/hwdef/iomcu/hwdef.inc
peterbarker: Allocate an ID for IOMCU

tridge: needs to be fixed

andyp1per: Yeah probably need a different value for different MCUs in IOMCU

andyp1per:
I like this

libraries/AP_HAL_ChibiOS/hwdef/iomcu/hwdef.inc
peterbarker: > Yeah probably need a different value for different MCUs in IOMCU

Thanks for mentioning this.  I brought up the same point on DevCall but nobody really responded to that.  I pointed out we use the same ID for Mini-Pix, f103-IOMCU, f303-IOMCU and CubeRedSecondary-IO!

It's not *that* critical as nothing is currently enforced.  But we should fix things.

libraries/AP_HAL/hwdef/scripts/hwdef.py
peterbarker: Now only enforce in CI.  Or those who choose to set the same environment variable that github actions does...

libraries/AP_HAL_ChibiOS/hwdef/iomcu/hwdef.inc
peterbarker: Fixed - allocated three IDs for various iomcu flavours

peterbarker: > Yeah probably need a different value for different MCUs in IOMCU

I think I've allocated and used sufficient now.

peterbarker:
Output from `size_compare_branches.py` is as expected:

```
pbarker@threads:~$ ~/rc/ardupilot/Tools/scripts/filter_size_compare_branches_csv.py --hide-binary-identical /tmp/some.csv  
----------------------------  -----  ----------  ------  ----  -----  -----  ---
Board                         blimp  bootloader  copter  heli  plane  rover  sub
ARK_FPV                       8      *           8       8     8      8      8
ATOMRCF405NAVI-Deluxe         16     *           8       8     8      8      8
AnyleafH7                     8      *           16      16    16     8      8
BROTHERHOBBYF405v3            16     *           16      16    16     16     16
CubeOrange                    8      *           8       8     8      8      8
CubeOrange-ODID               8      *           0       0     8      0      8
CubeOrange-SimOnHardWare                         8
CubeOrange-bdshot             8      *           8       8     8      8      8
CubeOrange-joey               0      *           0       0     0      0      0
CubeOrangePlus                -8     *           -8      -8    0      -8     0
CubeOrangePlus-ODID           -8     *           0       -8    -8     0      -8
CubeOrangePlus-SimOnHardWare                     0
CubeOrangePlus-bdshot         0      *           0       0     0      0      0
Pixhawk6C                     24     *           16      24    24     24     16
Pixhawk6C-bdshot              16     *           16      16    16     16     16
----------------------------  -----  ----------  ------  ----  -----  -----  ---
pbarker@threads:~$ 
```

... noting that we don't put the board ID anywhere in the IOMCU firmwares, they were unchanged.  Other changes here are just the hwdef.dat changes.

peterbarker:
I have verified that CI will fail PRs that don't use numeric IDs by killing the changes to the BrotherHobbies boards:
image

libraries/AP_HAL/hwdef/scripts/hwdef.py
```diff
@@ -92,6 +97,15 @@ def get_numeric_board_id(self):
             if m.group('name') == some_id:
                 return m.group('board_id')
 
+        # upstream's portion of the board_list.txt has a bunch of
+        # boards in a "reserved" section for no apparent reason.
+        # They're a mess to parse, so just hack it up:
+        custom_map = {
+            '"PX4 [BL] FMU v6C.x"': 56,
```
tridge: not a fan of this custom map :-)

tridge (changes requested):
I'd rather change to PX4_BL_FMU_V6C

peterbarker:
Made the changes requested at DevCall - did away with the mapping from weird string to board ID for Pixhawk6C, and just modified the board type.  I created a PR in the PX4-Bootloader repository to make the same change.

peterbarker:
I've again confirmed there's no change to the build products for the IOMCUs.

peterbarker:
configure-all still passes everything when I rebase this on master.  Merging as agreed at DevCall.

PR #30350, opened 2025-06-14
Plane: fix short failsafe

Jazhi:
Hi,
I tested the short failsafe feature, but found that it is currently not working as expected. It appears that others have encountered the same issue - short failsafe triggered immediately without `FS_SHORT_TIMEOUT`:
https://discuss.ardupilot.org/t/fs-short-timeout-not-changing-rc-failsafe-timeout/119625

This PR fixes the `FS_SHORT_TIMEOUT` handling logic to ensure that the failsafe trigger responds correctly.
It just uses the same mechanism that long failsafe did. And since [arm check](https://github.com/ArduPilot/ardupilot/blob/966588f7142919482efc342cceebf0927366e65c/ArduPlane/AP_Arming_Plane.cpp#L78) will ask `FS_SHORT_TIMEOUT` < `FS_LONG_TIMEOUT`, they are not treated as relatively timeout usage anymore. (We don't need to know when short failsafe happened for long failsafe)
In addition, I also cleaned up some minor terminology inconsistencies across related functions.
I’ve tested this in SITL and confirmed that the failsafe now behaves as intended.

Hope this helps!

### DIFF
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -437,19 +437,10 @@ const AP_Param::Info Plane::var_info[] = {
 
     // @Param: FS_SHORT_ACTN
     // @DisplayName: Short failsafe action
-    // @Description: The action to take on a short (FS_SHORT_TIMEOUT) failsafe event. A short failsafe event can be triggered either by loss of RC control (see THR_FS_VALUE) or by loss of GCS control (see FS_GCS_ENABL). If in CIRCLE or RTL mode this parameter is ignored. A short failsafe event in stabilization and manual modes will cause a change to CIRCLE mode if FS_SHORT_ACTN is 0 or 1, a change to FBWA mode with zero throttle if FS_SHORT_ACTN is 2, and a change to FBWB mode if FS_SHORT_ACTN is 4. In all other modes (AUTO, GUIDED and LOITER) a short failsafe event will cause no mode change if FS_SHORT_ACTN is set to 0, will cause a change to CIRCLE mode if set to 1, will change to FBWA mode with zero throttle if set to 2, or will change to FBWB if set to 4. Please see the documentation for FS_LONG_ACTN for the behaviour after FS_LONG_TIMEOUT seconds of failsafe. This parameter only applies to failsafes during fixed wing modes. Quadplane modes will switch to QLAND unless Q_OPTIONS bit 5(QRTL) or 20(RTL) are set.
+    // @Description: The action to take on a short failsafe event. A short failsafe event can be triggered instantly by loss of RC control or by throttle value (see THR_FS_VALUE). If in CIRCLE or RTL mode this parameter is ignored. A short failsafe event in stabilization and manual modes will cause a change to CIRCLE mode if FS_SHORT_ACTN is 0 or 1, a change to FBWA mode with zero throttle if FS_SHORT_ACTN is 2, and a change to FBWB mode if FS_SHORT_ACTN is 4. In all other modes (AUTO, GUIDED and LOITER) a short failsafe event will cause no mode change if FS_SHORT_ACTN is set to 0, will cause a change to CIRCLE mode if set to 1, will change to FBWA mode with zero throttle if set to 2, or will change to FBWB if set to 4. Please see the documentation for FS_LONG_ACTN for the behaviour after FS_LONG_TIMEOUT seconds of failsafe. This parameter only applies to failsafes during fixed wing modes. Quadplane modes will switch to QLAND unless Q_OPTIONS bit 5(QRTL) or 20(RTL) are set.
     // @Values: 0:CIRCLE/no change(if already in AUTO|GUIDED|LOITER),1:CIRCLE,2:FBWA at zero throttle,3:Disable,4:FBWB
     // @User: Standard
     GSCALAR(fs_action_short,        "FS_SHORT_ACTN",  FS_ACTION_SHORT_BESTGUESS),
-
-    // @Param: FS_SHORT_TIMEOUT
-    // @DisplayName: Short failsafe timeout
-    // @Description: The time in seconds that a failsafe condition has to persist before a short failsafe event will occur. This defaults to 1.5 seconds
-    // @Units: s
-    // @Range: 1 100
-    // @Increment: 0.5
-    // @User: Standard
-    GSCALAR(fs_timeout_short,        "FS_SHORT_TIMEOUT", 1.5f),
 
     // @Param: FS_LONG_ACTN
     // @DisplayName: Long failsafe action
--- a/ArduPlane/events.cpp
+++ b/ArduPlane/events.cpp
@@ -18,11 +18,10 @@ bool Plane::failsafe_in_landing_sequence() const
     return false;
 }
 
-void Plane::failsafe_short_on_event(enum failsafe_state fstype, ModeReason reason)
+void Plane::rc_failsafe_short_on_event()
 {
     // This is how to handle a short loss of control signal failsafe.
-    failsafe.state = fstype;
-    failsafe.short_timer_ms = millis();
+    failsafe.state = FAILSAFE_SHORT;
     failsafe.saved_mode_number = control_mode->mode_number();
     switch (control_mode->mode_number())
     {
@@ -35,15 +34,15 @@ void Plane::failsafe_short_on_event(enum failsafe_state fstype, ModeReason reaso
     case Mode::Number::CRUISE:
     case Mode::Number::TRAINING:  
         if(plane.emergency_landing) {
-            set_mode(mode_fbwa, reason); // emergency landing switch overrides normal action to allow out of range landing
+            set_mode(mode_fbwa, ModeReason::RADIO_FAILSAFE); // emergency landing switch overrides normal action to allow out of range landing
             break;
         }
         if(g.fs_action_short == FS_ACTION_SHORT_FBWA) {
-            set_mode(mode_fbwa, reason);
+            set_mode(mode_fbwa, ModeReason::RADIO_FAILSAFE);
         } else if (g.fs_action_short == FS_ACTION_SHORT_FBWB) {
-            set_mode(mode_fbwb, reason);
+            set_mode(mode_fbwb, ModeReason::RADIO_FAILSAFE);
         } else {
-            set_mode(mode_circle, reason); // circle if action = 0 or 1 
+            set_mode(mode_circle, ModeReason::RADIO_FAILSAFE); // circle if action = 0 or 1 
         }
         break;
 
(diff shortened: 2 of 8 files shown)


Hwurzburg (commented):
this is a behavior change but I believe its a correct one and matches our documentation
hard to believe no one has noticed this in 8 years!

ArduPlane/events.cpp
```diff
@@ -234,13 +233,13 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
 #endif
         break;
     }
-    gcs().send_text(MAV_SEVERITY_WARNING, "%s Failsafe On: %s", (reason == ModeReason:: GCS_FAILSAFE) ? "GCS" : "RC Long", control_mode->name());
+    gcs().send_text(MAV_SEVERITY_WARNING, "%s Failsafe On: switched to %s", (reason == ModeReason:: GCS_FAILSAFE) ? "GCS" : "RC Long", control_mode->name());
```
IamPete1, line 236: I don't think adding the mode name adds anything. Sometimes we don't switch modes and if we do the GCS user should see the mode change without having to come and look at the message.

ArduPlane/events.cpp
```diff
@@ -234,13 +233,13 @@ void Plane::failsafe_long_on_event(enum failsafe_state fstype, ModeReason reason
...
 }
 
 void Plane::failsafe_short_off_event(ModeReason reason)
```
IamPete1: I think it would make this easier to understand if the function name made it clear that it is only RC fail safes. 
eg:
`rc_failsafe_short_off_event`

ArduPlane/events.cpp
```diff
@@ -22,7 +22,6 @@ void Plane::failsafe_short_on_event(enum failsafe_state fstype, ModeReason reaso
 {
```
IamPete1, line 22: maybe `rc_failsafe_short_on_event`.

ArduPlane/system.cpp
```diff
@@ -399,13 +397,16 @@ void Plane::check_long_failsafe()
 
 void Plane::check_short_failsafe()
 {
+    const uint32_t rc_last_seen_ms = failsafe.last_valid_rc_ms;
```
IamPete1: maybe rename to `check_short_rc_failsafe`

ArduPlane/events.cpp
IamPete1: also don't need to pass the mode reason, its always `ModeReason::RADIO_FAILSAFE`

ArduPlane/events.cpp
IamPete1: also don't need to pass the mode reason or state, its always `FAILSAFE_SHORT` and `ModeReason::RADIO_FAILSAFE`

IamPete1 (commented):
Some suggestions to make this change a little more com comprehensive if your up for it. It would also be good to update the param descriptions for `FS_SHORT_ACTN` and `FS_SHORT_TIMEOUT` to make it clear that they only apply to loss of RC control.

ArduPlane/events.cpp
Hwurzburg: short failsafes are only a plane rc thing, not sure it would be confused with anything else...but name change does not hurt anything

IamPete1: > short failsafes are only a plane rc thing, not sure it would be confused with anything else...but name change does not hurt anything

The param description says it applies to GCS too. 

```
A short failsafe event can be triggered either by loss of RC control (see THR_FS_VALUE) or by loss of GCS control (see FS_GCS_ENABL).
```

https://ardupilot.org/plane/docs/parameters.html#fs-short-actn-short-failsafe-action

IamPete1: We should fix that too of course.

Hwurzburg: oops, indeed it does....my mistake

LupusTheCanine:
> this is a behavior change but I believe its a correct one and matches our documentation hard to believe no one has noticed this in 8 years!

I guess few people test RC failsafe thoroughly and behavior difference, especially on the bench, is minimal to the observer it isn't like the plane was doing a sharp turn sooner than expected.

I would be wary of changing the behavior, probably at least one user hasn't adjusted the timeout back to reasonable value after testing 😅 .

Hwurzburg:
since you are continuin gto push to this, I have removed the DEVCALL tag

Jazhi:
All checks passed.
Glad to help~

tridge:
we better think carefully about fixing this ... if someone is in MANUAL mode and you wait 1.5s before kicking in short failsafe then we will likely lose the vehicle

LupusTheCanine:
> we better think carefully about fixing this ... if someone is in MANUAL mode and you wait 1.5s before kicking in short failsafe then we will likely lose the vehicle

1.5s is likely to be salvageable unless the plane is about to land or doing acrobatics. Somebody might have set it to 10s when testing and seeing no effects and not set it back 😅.

Trainer mode sounds like good immediate failsafe for manual and acro could get some weak leveling.

Hwurzburg:
you could change the default to something like 0.5s to emulate current behavior but this PR will then at least make the timeout functional

LupusTheCanine:
> you could change the default to something like 0.5s to emulate current behavior but this PR will then at least make the timeout functional



> you could change the default to something like 0.5s to emulate current behavior but this PR will then at least make the timeout functional

Default change won't affect anybody who ever changed that parameter (including people who left unreasonable value after failed testing).

IMHO acro and manual should immediately go to Circle or at least level flight.

Hwurzburg:
if we dont correct this we should remove the timeout entirely....

Jazhi:
Would it be a problem if we add another parameter to emulate 0.5s timeout and remove the current one?
Like "RC_FS_SHORT_TIMEOUT"

IamPete1:
Adding a new parameter would get everyone back to defaults at least. I think we first need to take this to a devcall and decide what we want the behavior to be. 

@Jazhi  I guess you want this to work, so would like to set a short timeout. The users from the linked thread too. So that is some votes for keeping a parameter.

Jazhi:
@IamPete1  Right. I do think it looks helpful. We do have a long failsafe can RTL, which also doesn't support loiter and circle yet. And for example, RC runs out of battery, plane cannot stay somewhere as a drone (brake) till people solve the issue. I even personally wish it could work when GCS lost. Anyway, got it — happy to help further if needed!

tridge:
we could change the param ID number, but with the same name, thus forcing a revert to default, and make default 0.5s?

Jazhi:
If I didn't get it wrong, [read_radio](https://github.com/ArduPilot/ardupilot/blob/d11dabdb89c9cc5e5630c5bff0582e06c4f87e8b/ArduPlane/Plane.cpp#L68) update in 50Hz. After 10 [loops](https://github.com/ArduPilot/ardupilot/blob/d11dabdb89c9cc5e5630c5bff0582e06c4f87e8b/ArduPlane/radio.cpp#L298) of RC fail, it triggers [check_short_failsafe](https://github.com/ArduPilot/ardupilot/blob/d11dabdb89c9cc5e5630c5bff0582e06c4f87e8b/ArduPlane/system.cpp#L408C25-L408C26). So 0.2s probably will be the least delay time that people have experienced in these years.

peterbarker:
> we could change the param ID number, but with the same name, thus forcing a revert to default, and make default 0.5s?

Better to break people's `param load`s too, IMO

tridge (changes requested):
I think we either need to remove the parameter completely, or rename and renumber 
eg. FS_SHORT_RC_TIME, with default matching current behaviour
key thing is we must not change default behaviour
I'd be happy with just removing the parameter

Hwurzburg:
@Jazhi fyi, this was the consensus of the Dev team.....for this PR to move forward, one of Tridge's suggested change options must be made to the PR....most think that the param should just be removed...

Jazhi:
Got it. FS_SHORT_TIMEOUT probably didn’t add much value over time, and maybe RC link stability improved with it too. I’ll go ahead and remove it.

Hwurzburg:
I dont think you removed the parameter itself

Jazhi:
Sorry. Other things have caused me. I may need some time to fix it.

Hwurzburg (approved):

IamPete1 (approved):

IamPete1:
I have rebased so hopefully this will get past CI.

ArduPlane/Parameters.cpp
```diff
@@ -437,20 +437,11 @@ const AP_Param::Info Plane::var_info[] = {
...
+    // @Description: The action to take on a short failsafe event. A short failsafe event can be triggered instantly by loss of RC control (see THR_FS_VALUE). If in CIRCLE or RTL mode this parameter is ignored. A short failsafe event in stabilization and manual modes will cause a change to CIRCLE mode if FS_SHORT_ACTN is 0 or 1, a change to FBWA mode with zero throttle if FS_SHORT_ACTN is 2, and a change to FBWB mode if FS_SHORT_ACTN is 4. In all other modes (AUTO, GUIDED and LOITER) a short failsafe event will cause no mode change if FS_SHORT_ACTN is set to 0, will cause a change to CIRCLE mode if set to 1, will change to FBWA mode with zero throttle if set to 2, or will change to FBWB if set to 4. Please see the documentation for FS_LONG_ACTN for the behaviour after FS_LONG_TIMEOUT seconds of failsafe. This parameter only applies to failsafes during fixed wing modes. Quadplane modes will switch to QLAND unless Q_OPTIONS bit 5(QRTL) or 20(RTL) are set.
```
Hwurzburg: historical, but since we are changing it
```suggestion
    // @Description: The action to take on a short failsafe event. A short failsafe event can be triggered instantly by loss of RC control or by throttle value (see THR_FS_VALUE). If in CIRCLE or RTL mode this parameter is ignored. A short failsafe event in stabilization and manual modes will cause a change to CIRCLE mode if FS_SHORT_ACTN is 0 or 1, a change to FBWA mode with zero throttle if FS_SHORT_ACTN is 2, and a change to FBWB mode if FS_SHORT_ACTN is 4. In all other modes (AUTO, GUIDED and LOITER) a short failsafe event will cause no mode change if FS_SHORT_ACTN is set to 0, will cause a change to CIRCLE mode if set to 1, will change to FBWA mode with zero throttle if set to 2, or will change to FBWB if set to 4. Please see the documentation for FS_LONG_ACTN for the behaviour after FS_LONG_TIMEOUT seconds of failsafe. This parameter only applies to failsafes during fixed wing modes. Quadplane modes will switch to QLAND unless Q_OPTIONS bit 5(QRTL) or 20(RTL) are set.
```

Hwurzburg (changes requested):

Hwurzburg (approved):

Hwurzburg:
@Jazhi what were the changes you made in that last commit push? they not obvious...

Jazhi:
hi @Hwurzburg I simply rebased it to pass the checks.

Jazhi:
Rebased only

PR #30352, opened 2025-06-14
autotest: add ScriptingFlyVelocity autotest

andyp1per:
~~This demonstrates a problem with set_target_velocity_NED() where request in m/s are showing up as 100x in the log~~
this is an issue with GCS display, but the test is still a good one as there are no autotests currently for these scripting functions.

### DIFF
--- a/Tools/autotest/arducopter.py
+++ b/Tools/autotest/arducopter.py
@@ -13705,7 +13705,7 @@ RTL_ALT 111
         self.install_example_script_context('set-target-velocity.lua')
         self.reboot_sitl()
 
-        # Should be alowd to enter from alt hold
+        # Should be allowed to enter from alt hold
         self.change_mode("ALT_HOLD")
         self.wait_ready_to_arm()
         self.arm_vehicle()


peterbarker:
I can't fault the code here.

It's logging the demand correctly in cm/s - units are correct, multiplier is correct.

It's working just the same as other things in the same like `CTUN.CRt` , for example.

I think you might be expecting MissionPlanner to automatically scale into units you like?  It's absolutely possible for MissionPlanner to magically scale into units you want (eg. degrees from radians, metres/second from cm/second) - but I don't think it ever did that?

When you graph this in MissionPlanner, does it label the axes incorrectly or something?  eg. stating something is in metres when it is clearly logged in cm?

Tools/autotest/arducopter.py
```diff
@@ -13695,6 +13695,28 @@ def ScriptingFlipMode(self):
...
+        '''test flying a velocity vector using scripting'''
+
+        # enable scripting and install set-target-velocity script
+        self.set_parameters({
+            "SCR_ENABLE": 1,
+        })
+        self.install_example_script_context('set-target-velocity.lua')
+        self.reboot_sitl()
+
+        # Should be alowd to enter from alt hold
```
peterbarker:
```suggestion
        # Should be allowed to enter from alt hold
```

peterbarker (approved):
LGTM

PR #30355, opened 2025-06-14
Copter: Fix units in Log_Write_Guided_Position_Target

lthall:
This is a simple fix to include the conversion from cd to m in the logging function. This is the minimum change fix. I will convert this to all operate in the units of m when I convert the functions that use it to m.

Master:
image

After PR:

Both can be plotted on the same axis.

image

It appears that the F-BBB-BBBBBB has no impact as I have also changed that to F-000-000000


peterbarker (approved):
LGTM

PR #30354, opened 2025-06-14
Add unit extensions for variables in  PosHold and ZigZag units

lthall:
This is a zero compiler change:

```
Board,blimp,bootloader,copter,heli,plane,rover,sub
CubeOrange,*,*,*,*,*,*,*
```

PR #30358, opened 2025-06-15
Copter: Mode PosHold timing fix

lthall:
This moves the timing in PosHold to use AP_HAL::millis() instead of hard coded cycles that assume an update rate of 400 Hz.


ArduCopter/mode_poshold.cpp
```diff
@@ -67,6 +66,7 @@ bool ModePosHold::init(bool ignore_checks)
 // should be called at 100hz or more
 void ModePosHold::run()
 {
+    const uint32_t now = AP_HAL::millis();
```
peterbarker: now_ms

ArduCopter/mode_poshold.cpp
```diff
@@ -194,11 +194,11 @@ void ModePosHold::run()
...
                 // initialise BRAKE mode
-                roll_mode = RPMode::BRAKE;        // Set brake roll mode
-                brake.roll_cd = 0.0f;                  // initialise braking angle to zero
-                brake.angle_max_roll_cd = 0.0f;        // reset brake_angle_max so we can detect when vehicle begins to flatten out during braking
-                brake.timeout_roll = POSHOLD_BRAKE_TIME_ESTIMATE_MAX; // number of cycles the brake will be applied, updated during braking mode.
-                brake.time_updated_roll = false;   // flag the braking time can be re-estimated
+                roll_mode = RPMode::BRAKE;          // Set brake roll mode
+                brake.roll_cd = 0.0f;               // initialise braking angle to zero
+                brake.angle_max_roll_cd = 0.0f;     // reset brake_angle_max so we can detect when vehicle begins to flatten out during braking
+                brake.start_time_roll = now;           // the time that the brake will be applied, updated during braking mode.
```
peterbarker: start_time_roll_ms

ArduCopter/mode_poshold.cpp
```diff
@@ -207,34 +207,35 @@ void ModePosHold::run()
...
             // calculate brake.roll angle to counter-act velocity
             update_brake_angle_from_velocity(brake.roll_cd, vel_right_cms);
 
             // update braking time estimate
             if (!brake.time_updated_roll) {
                 // check if brake angle is increasing
-                if (fabsf(brake.roll_cd) >= brake.angle_max_roll_cd) {
-                    brake.angle_max_roll_cd = fabsf(brake.roll_cd);
+                if (fabsf(brake.pitch_cd) >= brake.angle_max_pitch_cd) {
+                    brake.angle_max_pitch_cd = fabsf(brake.pitch_cd);
```
peterbarker: change from roll to pitch?!

ArduCopter/mode_poshold.cpp
```diff
@@ -207,34 +207,35 @@ void ModePosHold::run()
...
                 } else {
                     // braking angle has started decreasing so re-estimate braking time
-                    brake.timeout_roll = 1+(uint16_t)(LOOP_RATE_FACTOR*15L*(int32_t)(fabsf(brake.roll_cd))/(10L*(int32_t)g.poshold_brake_rate));  // the 1.2 (12/10) factor has to be tuned in flight, here it means 120% of the "normal" time.
+                    brake.start_time_roll = now;
```
peterbarker: _ms all around

ArduCopter/mode_poshold.cpp
```diff
@@ -207,34 +207,35 @@ void ModePosHold::run()
...
 
             // update braking time estimate
             if (!brake.time_updated_roll) {
                 // check if brake angle is increasing
-                if (fabsf(brake.roll_cd) >= brake.angle_max_roll_cd) {
-                    brake.angle_max_roll_cd = fabsf(brake.roll_cd);
+                if (fabsf(brake.pitch_cd) >= brake.angle_max_pitch_cd) {
+                    brake.angle_max_pitch_cd = fabsf(brake.pitch_cd);
                 } else {
                     // braking angle has started decreasing so re-estimate braking time
```
peterbarker, line 222: No longer doing this, you're just resetting the time

ArduCopter/mode_poshold.cpp
```diff
@@ -207,34 +207,35 @@ void ModePosHold::run()
...
-                brake.timeout_roll--;
-            } else {
-                // indicate that we are ready to move to Loiter.
-                // Loiter will only actually be engaged once both roll_mode and pitch_mode are changed to RPMode::BRAKE_READY_TO_LOITER
-                //  logic for engaging loiter is handled below the roll and pitch mode switch statements
-                roll_mode = RPMode::BRAKE_READY_TO_LOITER;
+                // reduce braking timer
+                if (now - brake.start_time_roll > break_timeout_roll) {
+                    // indicate that we are ready to move to Loiter.
+                    // Loiter will only actually be engaged once both pitch_mode and pitch_mode are changed to RPMode::BRAKE_READY_TO_LOITER
```
peterbarker: Uh, whoops

peterbarker:
This PR is built on https://github.com/ArduPilot/ardupilot/pull/30354 - only the last patch is interesting here

ArduCopter/mode_poshold.cpp
lthall: >> I didn't see any recalculation of the actual time required to do the braking maneuver; when I saw it it was just resetting the start time of the maneuver.

Yes we only reset the time here.

The old code calculated the time required based on brake.roll on the one loop that it runs this code in. But it also records that value in brake.angle_max_roll_cd.

So to get arround the change in pattern to now_ms - bla I calculated break_timeout_roll based on brake.angle_max_roll_cd on each loop before checking it.

` const uint32_t break_timeout_roll = MIN(POSHOLD_BRAKE_TIME_ESTIMATE_MAX_MS, (1.5 * 1000 * 0.01) * brake.angle_max_roll_cd / g.poshold_brake_rate);`

ArduCopter/mode.h
```diff
@@ -1401,41 +1401,41 @@ class ModePosHold : public Mode {
...
-        float roll_cd;                      // target roll angle during braking periods
-        float pitch_cd;                     // target pitch angle during braking periods
-        int16_t timeout_roll;               // number of cycles allowed for the braking to complete, this timeout will be updated at half-braking
-        int16_t timeout_pitch;              // number of cycles allowed for the braking to complete, this timeout will be updated at half-braking
-        float angle_max_roll_cd;            // maximum lean angle achieved during braking. Used to determine when the vehicle has begun to flatten out so that we can re-estimate the braking time
-        float angle_max_pitch_cd;           // maximum lean angle achieved during braking. Used to determine when the vehicle has begun to flatten out so that we can re-estimate the braking time
-        int16_t to_loiter_timer;            // cycles to mix brake and loiter controls in POSHOLD_TO_LOITER
+        uint8_t time_updated_roll   : 1;        // true if braking timeout on roll axis has been re-estimated
+        uint8_t time_updated_pitch  : 1;        // true if braking timeout on pitch axis has been re-estimated
```
peterbarker: lose the `: 1` to save flash

ArduCopter/mode.h
```diff
@@ -1401,41 +1401,41 @@ class ModePosHold : public Mode {
...
     } brake;
 
+
     // loiter related variables
-    int16_t controller_to_pilot_timer_roll;     // cycles to mix controller and pilot controls in POSHOLD_CONTROLLER_TO_PILOT
-    int16_t controller_to_pilot_timer_pitch;    // cycles to mix controller and pilot controls in POSHOLD_CONTROLLER_TO_PILOT
-    float controller_final_roll_cd;             // final roll angle from controller as we exit brake or loiter mode (used for mixing with pilot input)
-    float controller_final_pitch_cd;            // final pitch angle from controller as we exit brake or loiter mode (used for mixing with pilot input)
+    uint32_t controller_to_pilot_start_time_roll;   // time (ms) when transition from controller to pilot roll input began
+    uint32_t controller_to_pilot_start_time_pitch;  // time (ms) when transition from controller to pilot pitch input began
```
peterbarker:
```suggestion
    uint32_t controller_to_pilot_start_time_roll;   // time (ms) when transition from controller to pilot roll input began
    uint32_t controller_to_pilot_start_time_pitch;  // time (ms) when transition from controller to pilot pitch input began
```
```suggestion
    uint32_t controller_to_pilot_start_time_roll_ms;   // time (ms) when transition from controller to pilot roll input began
    uint32_t controller_to_pilot_start_time_pitch_ms;  // time (ms) when transition from controller to pilot pitch input began
```

ArduCopter/mode.h
```diff
@@ -1401,41 +1401,41 @@ class ModePosHold : public Mode {
...
-        float angle_max_pitch_cd;           // maximum lean angle achieved during braking. Used to determine when the vehicle has begun to flatten out so that we can re-estimate the braking time
-        int16_t to_loiter_timer;            // cycles to mix brake and loiter controls in POSHOLD_TO_LOITER
+        uint8_t time_updated_roll   : 1;        // true if braking timeout on roll axis has been re-estimated
+        uint8_t time_updated_pitch  : 1;        // true if braking timeout on pitch axis has been re-estimated
+        float gain;                             // braking gain used to convert velocity to lean angle
+        float roll_cd;                          // braking roll angle in centidegrees
+        float pitch_cd;                         // braking pitch angle in centidegrees
+        uint32_t start_time_roll;               // system time (ms) when braking on roll axis began
+        uint32_t start_time_pitch;              // system time (ms) when braking on pitch axis began
```
peterbarker:
```suggestion
        uint32_t start_time_roll_ms;               // system time (ms) when braking on roll axis began
        uint32_t start_time_pitch_ms;              // system time (ms) when braking on pitch axis began
```

ArduCopter/mode.h
```diff
@@ -1401,41 +1401,41 @@ class ModePosHold : public Mode {
...
     // wind compensation related variables
-    Vector2f wind_comp_ef;                      // wind compensation in earth frame, filtered lean angles from position controller
-    float wind_comp_roll_cd;                    // roll angle to compensate for wind
-    float wind_comp_pitch_cd;                   // pitch angle to compensate for wind
-    uint16_t wind_comp_start_timer;             // counter to delay start of wind compensation for a short time after loiter is engaged
-    int8_t  wind_comp_timer;                    // counter to reduce wind comp roll/pitch lean angle calcs to 10hz
+    Vector2f wind_comp_ef;           // wind compensation acceleration vector (earth frame), low-pass filtered
+    float wind_comp_roll_cd;         // roll angle (deg x100) to counter wind based on earth-frame lean
+    float wind_comp_pitch_cd;        // pitch angle (deg x100) to counter wind based on earth-frame lean
+    uint32_t wind_comp_start_timer;  // system time (ms) when wind compensation updates are allowed to begin
```
peterbarker:
```suggestion
    uint32_t wind_comp_start_timer_ms;  // system time (ms) when wind compensation updates are allowed to begin
```

Is that comment accurate?  We don't usually store times into the future, and that's what this sounds like

ArduCopter/mode_poshold.cpp
```diff
@@ -194,11 +194,11 @@ void ModePosHold::run()
...
                 // initialise BRAKE mode
-                roll_mode = RPMode::BRAKE;        // Set brake roll mode
-                brake.roll_cd = 0.0f;                  // initialise braking angle to zero
-                brake.angle_max_roll_cd = 0.0f;        // reset brake_angle_max so we can detect when vehicle begins to flatten out during braking
-                brake.timeout_roll = POSHOLD_BRAKE_TIME_ESTIMATE_MAX; // number of cycles the brake will be applied, updated during braking mode.
-                brake.time_updated_roll = false;   // flag the braking time can be re-estimated
+                roll_mode = RPMode::BRAKE;          // Set brake roll mode
+                brake.roll_cd = 0.0f;               // initialise braking angle to zero
+                brake.angle_max_roll_cd = 0.0f;     // reset brake_angle_max so we can detect when vehicle begins to flatten out during braking
+                brake.start_time_roll = now_ms;           // the time that the brake will be applied, updated during braking mode.
```
peterbarker: Indenting.  Other people care more than I do.

peterbarker:
```suggestion
                brake.start_time_roll_ms = now_ms;           // the time that the brake will be applied, updated during braking mode.
```

ArduCopter/mode_poshold.cpp
```diff
@@ -207,6 +207,7 @@ void ModePosHold::run()
 
         case RPMode::BRAKE:
         case RPMode::BRAKE_READY_TO_LOITER:
+
```
peterbarker: bad whitespace change

ArduCopter/mode_poshold.cpp
peterbarker: > > > I didn't see any recalculation of the actual time required to do the braking maneuver; when I saw it it was just resetting the start time of the maneuver.
> 
> Yes we only reset the time here.

Great, I'm not going insane.  Is there some reason not to make a method call here to update the expected duration here?

ArduCopter/mode_poshold.cpp
```diff
@@ -217,24 +218,24 @@ void ModePosHold::run()
                     brake.angle_max_roll_cd = fabsf(brake.roll_cd);
                 } else {
                     // braking angle has started decreasing so re-estimate braking time
-                    brake.timeout_roll = 1+(uint16_t)(LOOP_RATE_FACTOR*15L*(int32_t)(fabsf(brake.roll_cd))/(10L*(int32_t)g.poshold_brake_rate));  // the 1.2 (12/10) factor has to be tuned in flight, here it means 120% of the "normal" time.
+                    brake.start_time_roll = now_ms;
                     brake.time_updated_roll = true;
```
peterbarker, line 224: Why don't we just call a subroutine here to do the calculation?

ArduCopter/mode_poshold.cpp
lthall: Just because I would need to make that variable permanent and I already have the information stored in  brake.angle_max_roll_cd and g.poshold_brake_rate. So I went with the lower memory option.

ArduCopter/mode_poshold.cpp
lthall: Not sure what you are suggesting.

ArduCopter/mode.h
lthall: Doesn't this change behaviour?

ArduCopter/mode.h
lthall: fixed

ArduCopter/mode.h
lthall: I think this is ok. There isn't anything that prevents this happening anyway.

ArduCopter/mode_poshold.cpp
```diff
@@ -392,17 +395,15 @@ void ModePosHold::run()
...
+                if (now_ms - brake.loiter_transition_start_time_ms > POSHOLD_BRAKE_TO_LOITER_TIME_MS) {
                     // progress to full loiter on next iteration
                     roll_mode = RPMode::LOITER;
                     pitch_mode = RPMode::LOITER;
                 }
 
                 // mix of brake and loiter controls.  0 = fully brake
                 // controls, 1 = fully loiter controls
-                const float brake_to_loiter_mix = (float)brake.to_loiter_timer / (float)POSHOLD_BRAKE_TO_LOITER_TIMER;
+                const float brake_to_loiter_mix = (float)(now_ms - brake.loiter_transition_start_time_ms) / (float)POSHOLD_BRAKE_TO_LOITER_TIME_MS;
```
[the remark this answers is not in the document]
lthall: This is done in mix_controls.

ArduCopter/mode_poshold.cpp
```diff
@@ -522,23 +524,18 @@ float ModePosHold::mix_controls(float mix_ratio, float first_control, float seco
...
 void ModePosHold::update_brake_angle_from_velocity(float &brake_angle_cd, float velocity_cms)
 {
     float lean_angle;
-    float brake_rate = g.poshold_brake_rate;
-
-    brake_rate /= (float)LOOP_RATE_FACTOR;
-    if (brake_rate <= 1.0f) {
-        brake_rate = 1.0f;
-    }
+    float brake_delta_cd = g.poshold_brake_rate * G_Dt;
```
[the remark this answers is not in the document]
lthall: Fixed

ArduCopter/mode.h
```diff
@@ -1401,41 +1401,41 @@ class ModePosHold : public Mode {
...
-        float roll_cd;                      // target roll angle during braking periods
-        float pitch_cd;                     // target pitch angle during braking periods
-        int16_t timeout_roll;               // number of cycles allowed for the braking to complete, this timeout will be updated at half-braking
-        int16_t timeout_pitch;              // number of cycles allowed for the braking to complete, this timeout will be updated at half-braking
-        float angle_max_roll_cd;            // maximum lean angle achieved during braking. Used to determine when the vehicle has begun to flatten out so that we can re-estimate the braking time
-        float angle_max_pitch_cd;           // maximum lean angle achieved during braking. Used to determine when the vehicle has begun to flatten out so that we can re-estimate the braking time
-        int16_t to_loiter_timer;            // cycles to mix brake and loiter controls in POSHOLD_TO_LOITER
+        uint8_t time_updated_roll;          // true if braking timeout on roll axis has been re-estimated
+        uint8_t time_updated_pitch;         // true if braking timeout on pitch axis has been re-estimated
```
peterbarker:
```suggestion
        bool time_updated_roll;          // true if braking timeout on roll axis has been re-estimated
        bool time_updated_pitch;         // true if braking timeout on pitch axis has been re-estimated
```

ArduCopter/mode.h
```diff
@@ -1401,41 +1401,41 @@ class ModePosHold : public Mode {
...
     // wind compensation related variables
-    Vector2f wind_comp_ef;                      // wind compensation in earth frame, filtered lean angles from position controller
-    float wind_comp_roll_cd;                    // roll angle to compensate for wind
-    float wind_comp_pitch_cd;                   // pitch angle to compensate for wind
-    uint16_t wind_comp_start_timer;             // counter to delay start of wind compensation for a short time after loiter is engaged
-    int8_t  wind_comp_timer;                    // counter to reduce wind comp roll/pitch lean angle calcs to 10hz
+    Vector2f wind_comp_ef;              // wind compensation acceleration vector (earth frame), low-pass filtered
+    float wind_comp_roll_cd;            // roll angle (deg x100) to counter wind based on earth-frame lean
+    float wind_comp_pitch_cd;           // pitch angle (deg x100) to counter wind based on earth-frame lean
+    uint32_t wind_comp_start_timer_ms;  // delay time (ms) when wind compensation updates are allowed to begin
```
peterbarker:
```suggestion
    uint32_t wind_comp_start_time_ms;  // delay time (ms) when wind compensation started
```

ArduCopter/mode_poshold.cpp
```diff
@@ -9,15 +9,13 @@
...
-#define POSHOLD_WIND_COMP_TIMER_10HZ            40      // counter value used to reduce wind compensation to 10hz
-#define LOOP_RATE_FACTOR                        4       // used to adapt PosHold params to loop_rate
-#define TC_WIND_COMP                            0.0025f // Time constant for poshold_update_wind_comp_estimate()
+#define POSHOLD_BRAKE_TIME_ESTIMATE_MAX_MS      6000    // Maximum duration (ms) allowed for braking before transitioning to loiter
+#define POSHOLD_BRAKE_TO_LOITER_TIME_MS         1500    // Duration (ms) over which braking is blended into loiter control during BRAKE_TO_LOITER phase
+#define POSHOLD_WIND_COMP_START_TIME_MS         1500    // Delay (ms) after entering loiter before wind compensation begins updating
+#define POSHOLD_CONTROLLER_TO_PILOT_MIX_TIME_MS 500     // Duration (ms) over which control is blended from autopilot to pilot input during CONTROLLER_TO_PILOT_OVERRIDE
+#define POSHOLD_SMOOTH_RATE_FACTOR              0.0125f // Low-pass filter factor for smoothing pilot roll/pitch input as it returns to center
+#define POSHOLD_WIND_COMP_TIMER_10HZ            40      // Loop counter interval for executing wind compensation updates at 10 Hz
```
peterbarker:
```suggestion
```

peterbarker (changes requested):

ArduCopter/mode_poshold.cpp
```diff
@@ -217,24 +216,28 @@ void ModePosHold::run()
...
-                roll_mode = RPMode::BRAKE_READY_TO_LOITER;
+                // scaling factors:
+                // 1.5 times the time taken to level the aircraft in ms
+                // 1000 to convert from seconds to ms
+                // 0.01 to convert angle_max_roll_cd to degrees
+                const uint32_t brake_timeout_roll_ms = MIN(POSHOLD_BRAKE_TIME_ESTIMATE_MAX_MS, (1.5 * 1000 * 0.01) * brake.angle_max_roll_cd / g.poshold_brake_rate_degs);
+
+                // if velocity is very low reduce braking time to 0.5seconds
+                if ((fabsf(vel_right_cms) <= POSHOLD_SPEED_0) && (now_ms - brake.start_time_roll_ms > 500)) {
+                    brake.start_time_roll_ms = now_ms - brake_timeout_roll_ms + 500;
```
peterbarker, line 235: This needs some work - can't ever make times in the future!

ArduCopter/mode_poshold.cpp
```diff
@@ -311,24 +312,28 @@ void ModePosHold::run()
...
-                pitch_mode = RPMode::BRAKE_READY_TO_LOITER;
+                // scaling factors:
+                // 1.5 times the time taken to level the aircraft in ms
+                // 1000 to convert from seconds to ms
+                // 0.01 to convert angle_max_pitch_cd to degrees
+                const uint32_t brake_timeout_pitch_ms = MIN(POSHOLD_BRAKE_TIME_ESTIMATE_MAX_MS, (1.5 * 1000 * 0.01) * brake.angle_max_pitch_cd / g.poshold_brake_rate_degs);
+
+                // if velocity is very low reduce braking time to 0.5seconds
+                if ((fabsf(vel_fw_cms) <= POSHOLD_SPEED_0) && (now_ms - brake.start_time_pitch_ms > 500)) {
+                    brake.start_time_pitch_ms = now_ms - brake_timeout_pitch_ms + 500;
```
peterbarker, line 330: again, needs work

lthall:
Flight test in sitl with wind all seems to work fine:

- Roll and pitch behave the same way,
- Compensation for wind works as expected,
- Aircraft drifts in Alt_Hold but then stops after switch to PosHold,
- PosHold does not drift with wind when moving cross wind,
- Breaking is happening when sticks are set to zero.

ArduCopter/Parameters.h
```diff
@@ -413,7 +413,7 @@ class Parameters {
     AP_Int8         wp_yaw_behavior;            // controls how the autopilot controls yaw during missions
 
 #if MODE_POSHOLD_ENABLED
-    AP_Int16        poshold_brake_rate;         // PosHold flight mode's rotation rate during braking in deg/sec
+    AP_Int16        poshold_brake_rate_degs;         // PosHold flight mode's rotation rate during braking in deg/sec
```
rmackay9: tiny white space issue here, not a real problem though

ArduCopter/mode_poshold.cpp
```diff
@@ -217,24 +215,28 @@ void ModePosHold::run()
...
             } else {
-                // indicate that we are ready to move to Loiter.
-                // Loiter will only actually be engaged once both roll_mode and pitch_mode are changed to RPMode::BRAKE_READY_TO_LOITER
-                //  logic for engaging loiter is handled below the roll and pitch mode switch statements
-                roll_mode = RPMode::BRAKE_READY_TO_LOITER;
+                // scaling factors:
+                // 1.5 times the time taken to level the aircraft in ms
+                // 1000 to convert from seconds to ms
+                // 0.01 to convert angle_max_roll_cd to degrees
+                const uint32_t brake_timeout_roll_ms = MIN(POSHOLD_BRAKE_TIME_ESTIMATE_MAX_MS, (1.5 * 1000 * 0.01) * brake.angle_max_roll_cd / g.poshold_brake_rate_degs);
```
rmackay9, line 231: It's an existing issue but a divide-by-zero is possible here because g.poshold_brake_rate_degs is a user defined parameter

ArduCopter/mode_poshold.cpp
```diff
@@ -217,24 +215,28 @@ void ModePosHold::run()
...
+
+                // if velocity is very low reduce braking time to 0.5seconds
+                if ((fabsf(vel_right_cms) <= POSHOLD_SPEED_0) && (now_ms - brake.start_time_roll_ms > 500) && (brake_timeout_roll_ms > 500)) {
+                    brake.start_time_roll_ms = now_ms - brake_timeout_roll_ms + 500;
```
rmackay9, line 235: I think it might be better to have a braking time variable that is set instead of changing the start_time_roll_ms value.

ArduCopter/mode_poshold.cpp
```diff
@@ -217,24 +215,28 @@ void ModePosHold::run()
...
+                }
+
+                // reduce braking timer
```
rmackay9: I guess this comment is out-of-date?  I don't think we are "reducing braking timer" here.  It looks instead as if we are checking if sufficient time has elapsed

ArduCopter/mode_poshold.cpp
```diff
@@ -311,24 +311,28 @@ void ModePosHold::run()
...
             } else {
-                // indicate that we are ready to move to Loiter.
-                // Loiter will only actually be engaged once both pitch_mode and pitch_mode are changed to RPMode::BRAKE_READY_TO_LOITER
-                //  logic for engaging loiter is handled below the pitch and pitch mode switch statements
-                pitch_mode = RPMode::BRAKE_READY_TO_LOITER;
+                // scaling factors:
+                // 1.5 times the time taken to level the aircraft in ms
+                // 1000 to convert from seconds to ms
+                // 0.01 to convert angle_max_pitch_cd to degrees
+                const uint32_t brake_timeout_pitch_ms = MIN(POSHOLD_BRAKE_TIME_ESTIMATE_MAX_MS, (1.5 * 1000 * 0.01) * brake.angle_max_pitch_cd / g.poshold_brake_rate_degs);
```
rmackay9, line 326: same comment here as for roll.  Perhaps we should constrain g.poshold_brake_rate_degs to some acceptable values somewhere

ArduCopter/mode_poshold.cpp
```diff
@@ -311,24 +311,28 @@ void ModePosHold::run()
...
+
+                // if velocity is very low reduce braking time to 0.5seconds
+                if ((fabsf(vel_fw_cms) <= POSHOLD_SPEED_0) && (now_ms - brake.start_time_pitch_ms > 500) && (brake_timeout_pitch_ms > 500)) {
+                    brake.start_time_pitch_ms = now_ms - brake_timeout_pitch_ms + 500;
+                }
+
+                // reduce braking timer
```
rmackay9: same nitpick here about the comment as made above in the roll section

rmackay9:
Thanks, this is definitely a nice step forward and makes the code style to be much closer to other modes.

ArduCopter/mode_poshold.cpp
lthall: I started out trying to do this but hit problems with int wrapping. This was the best I could come up with that mirrored the old behaviour.

ArduCopter/mode_poshold.cpp
lthall: Added this at the top:

   // enforce minimum allowed value for poshold_brake_rate_degs
    if (g.poshold_brake_rate_degs < POSHOLD_BRAKE_RATE_MIN) {
        g.poshold_brake_rate_degs.set_and_save(4);
    }

ArduCopter/mode_poshold.cpp
lthall: // enforce minimum allowed value for poshold_brake_rate_degs
    if (g.poshold_brake_rate_degs < POSHOLD_BRAKE_RATE_MIN) {
        g.poshold_brake_rate_degs.set_and_save(4);
    }

rmackay9 (approved):

ArduCopter/mode_poshold.cpp
```diff
@@ -67,10 +64,16 @@ bool ModePosHold::init(bool ignore_checks)
...
 void ModePosHold::run()
 {
+    const uint32_t now_ms = AP_HAL::millis();
     float controller_to_pilot_roll_mix; // mix of controller and pilot controls.  0 = fully last controller controls, 1 = fully pilot controls
     float controller_to_pilot_pitch_mix;    // mix of controller and pilot controls.  0 = fully last controller controls, 1 = fully pilot controls
     const Vector3f& vel_neu_cms = pos_control->get_vel_estimate_NEU_cms();
 
+    // enforce minimum allowed value for poshold_brake_rate_degs
+    if (g.poshold_brake_rate_degs < POSHOLD_BRAKE_RATE_MIN) {
+        g.poshold_brake_rate_degs.set_and_save(4);
```
rmackay9: this is fine but maybe we should change the set-and-save to be this?

g.poshold_brake_rate_degs.set_and_save(POSHOLD_BRAKE_RATE_MIN);

ArduCopter/mode_poshold.cpp
rmackay9: yup, ok

ArduCopter/mode_poshold.cpp
lthall: Oh #*#$

lthall: I added POSHOLD_BRAKE_RATE_MIN then didn't put it in the most important part!

Very sorry!

PR #30362, opened 2025-06-15
Tools: added name to GIT_Success.txt

Nvinh5148:
## What is this PR?

Add support for reading total voltage, current, and remaining capacity from JBD BMS over CAN using ID 0x100, as described in the provided protocol.

## Why is this needed?

To enable DroneCAN battery status reporting directly from BMS  JBD

## Tested on

- SITL (build success)
- Real BMS via STM32 + CAN interface (tested and verified)

## Notes

- Adds new file: `libraries/AP_BattMonitor/AP_BattMonitor_JBDCAN.cpp`
- Does not affect existing battery monitor functionality


peterbarker (changes requested):
This doesn't look right.

Did you *look* at the PR contents before hitting the "Create Pull Request" button?!

peterbarker (changes requested):
... still doesn't look right :-)

There's only one patch in here which matches the PR title.

The rest is the shell of something else.  The "something else" mentions "DroneCAN", but a battery monitor speaking DroneCAN should not need changes to ArduPilot to be supported.

... also changing random vscode stuff which it shouldn't.

peterbarker:
Ping @Nvinh5148 - still chasing this?

PR #30364, opened 2025-06-15
AP_RangeFinder: add special case to has_orientation

rishabsingh3003:
I am not sure if this is the correct fix. DroneCAN rangefinders are not added to backend till we receive a valid message. 
This means if the DroneCAN rangefinder hasn't initialised when the vehicle checks for an "enabled rangefinder": https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/sensors.cpp#L21 , it will remain disabled forever. 

This is very easy to reproduce on SITL. Add a DroneCAN rangefinder a few seconds after launching SITL, and monitor SYS_STATUS. Rangefinder remains disabled in the message.


rishabsingh3003:
Example script used to test this: https://github.com/dronecan/pydronecan/pull/84

peterbarker:
> I am not sure if this is the correct fix. DroneCAN rangefinders are not added to backend till we receive a valid message. This means if the DroneCAN rangefinder hasn't initialised when the vehicle checks for an "enabled rangefinder": https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/sensors.cpp#L21 , it will remain disabled forever.

Why is it that we don't instantiate the RangeFinder backend until we receive a DroneCAN message for it?  Has to be a good reason, right?

tridge (approved):

PR #30367, opened 2025-06-16
GCS_MAVLink: set message interval fail msg only sent on SITL

peterbarker:


rmackay9:

PR #30371, opened 2025-06-16
Copter: modes: get_alt_above_ground_cm to const

IamPete1:
Minor follow up to https://github.com/ArduPilot/ardupilot/pull/30195, the method can be const.

### DIFF
--- a/ArduCopter/mode.cpp
+++ b/ArduCopter/mode.cpp
@@ -600,7 +600,7 @@ void Mode::make_safe_ground_handling(bool force_throttle_unlimited)
 /*
   get a height above ground estimate for landing
  */
-int32_t Mode::get_alt_above_ground_cm(void)
+int32_t Mode::get_alt_above_ground_cm(void) const
 {
     int32_t alt_above_ground_cm;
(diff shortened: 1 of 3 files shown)


cclauss:
What about the 15 annotations at the bottom right of https://github.com/ArduPilot/ardupilot/actions/runs/15679190553 ?

IamPete1:
> What about the 15 annotations at the bottom right of https://github.com/ArduPilot/ardupilot/actions/runs/15679190553 ?

Some tests are known to be a bit unreliable so they have built in retries. The first attempt fails and you get the warning but it passes on a second attempt to the test as a whole passes.

cclauss:
`‘buff’ may be used uninitialized [-Wmaybe-uninitialized]` seems like a linter warning, not an intermittent test failure.

IamPete1:
> `‘buff’ may be used uninitialized [-Wmaybe-uninitialized]` seems like a linter warning, not an intermittent test failure.

Yep, I'm sure not as a result as this PR tho. Not a very useful warning, but if you can track it down a PR would be great!

PR #30375, opened 2025-06-17
AP_Camera: log angles in degrees rather than cd, check angle limit using degrees

peterbarker:
Augmented existing autotest to check the angles.

Interestingly the augmented test passed before/after because of the use of the magic, ancient and deprectated "C" and "c" logging field designators, which pymavlink magically applies to the values for you!

### DIFF
--- a/libraries/AP_Camera/AP_Camera_Backend.cpp
+++ b/libraries/AP_Camera/AP_Camera_Backend.cpp
@@ -85,7 +85,7 @@ void AP_Camera_Backend::update()
     }
 
     // check vehicle roll angle is less than configured maximum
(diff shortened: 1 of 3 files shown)


IamPete1:
I guess this might be a important message in a mapping workflow, I wonder if we should change the labels to purposely break peoples scripts. Currently they would probably still run but with wild results.

peterbarker:
> I guess this might be a important message in a mapping workflow, I wonder if we should change the labels to purposely break peoples scripts. Currently they would probably still run but with wild results.

Definitely worth considering.  We're at 60 of 64 characters so we could say "RD,PD,YD"...

PR #30377, opened 2025-06-17
AP_HAL_ChibiOS: chibios_hwdef.py: stop overriding process_file

peterbarker:
same as that in the base class


robertlong13 (approved):
LGTM

PR #30379, opened 2025-06-17
hwdef: remove un-needed lines

peterbarker:
These directives are either the defaults or inherited from include files.


peterbarker:
I've confirmed this is a no-compiler-output-change

PR #30380, opened 2025-06-17
Plane: reduce speed scaler in assisted flight at low airspeed

tridge:
when assisting with VTOL motors we want to ensure the speed scaler is low to prevent reversed ailerons during a VTOL recovery
In recovery we can get negative airspeeds, we need to lower speed scalers

### DIFF
--- a/ArduPlane/Attitude.cpp
+++ b/ArduPlane/Attitude.cpp
@@ -24,7 +24,7 @@ float Plane::calc_speed_scaler(void)
         speed_scaler = constrain_float(speed_scaler, scale_min, scale_max);
 
 #if HAL_QUADPLANE_ENABLED
-        if (quadplane.in_vtol_mode() && arming.is_armed_and_safety_off()) {
+        if ((quadplane.in_vtol_mode() || quadplane.in_assisted_flight()) && arming.is_armed_and_safety_off()) {
             // when in VTOL modes limit surface movement at low speed to prevent instability
             float threshold = airspeed_min * 0.5;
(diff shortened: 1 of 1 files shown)


IamPete1:
This also results in a forward flight integrator decay. In assisted flight the forward flight integratorors will be the only ones doing anything, if we decay them we will have no I terms.

tridge:
> This also results in a forward flight integrator decay. In assisted flight the forward flight integratorors will be the only ones doing anything, if we decay them we will have no I terms.

true, though this kicks in at airspeeds under half AIRSPEED_MIN. Do you think the integrators are doing much good at those speeds?

tridge:
ping @priseborough

IamPete1:
> true, though this kicks in at airspeeds under half AIRSPEED_MIN. Do you think the integrators are doing much good at those speeds?

They will be frozen, but not zeroed. The decay should be smooth, but it means when you do get back to forward flight you have to build up again. The freeze we do already should mean that you pick back up with the correct trim once you do get going again. Then again zeroing them means that a transition from assist will be the same as the first transition into forward flight where the interrogators have not been run at all yet,  

I don't necessarily think its a big deal, just something we should have a think about.

tridge:
@IamPete1 I think decaying I is the right thing to do. Otherwise if we are flying backwards we'll still end up with a big reversed control surface.

IamPete1 (approved):
Looks good.

rmackay9:
This is included in 4.6.2-beta2

Issue #28497, opened 2024-10-30
AP_DDS: Race condition generating DDS messages (microxrceddsgen is not thread safe)


Ryanf55:
## Bug report

**Issue details**

Very occasionally, microxrceddsgen fails to generate messages. This is likely because waf is writing to the same file from multiple threads.

**Version**
4.5
**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
all
**Hardware type**
sitl
**Logs**

**Details**

microxrceddsgen can take an argument for multiple files at the same time. We call `bld()` once for each file. It would be faster to do batch them per folder, or all together at the same time.

```
[...]

PR #30385, opened 2025-06-17
AP_DDS: fix race condition in waf DDS build

srmainwaring:
Reworked version of @Ryanf55 's patch (https://github.com/Ryanf55/ardupilot/tree/28497-fix-waf-dds-gen-race)

- Fixes https://github.com/ArduPilot/ardupilot/issues/28497


Ryanf55:
Huge thanks for finishing this out. Looks great Rhys.

PR #26090, opened 2024-01-28
HAL_ChibiOS: raise SDMMC clock limit on H7

tridge:
this allows for faster log download on ethernet.
On CubeRedPrimary log download using the web server on ethernet goes up from 2.4MByte/s to 4.0MByte/s

I don't think this should go straight into 4.5.x, I'd like to see if anyone notices any reliability issues with the faster clock first

PR #30387, opened 2025-06-18
AP_HAL_ChibiOS: JFB110 microSD speed increased

rmackay9:
For reasons that I do not understand, this resolves the JFB110 startup issue reported in this [4.6 discussion](https://discuss.ardupilot.org/t/jfb110-fc-mavftp-issue/134938)

I've tested before and after to confirm this resolves the problem

This issue started appearing after this PR was merged https://github.com/ArduPilot/ardupilot/pull/26090

This change was actually suggested by @jfbblue0922


rmackay9:

PR #30120, opened 2025-05-20
AC_AttitudeControl: Add conversions from cd to rad

lthall:
This breaks the attitude controller functions where they convert from centidegrees to radians. This will let us add additional conversions in for degrees while maintaining centidegrees.

PR #30391, opened 2025-06-18
Copter: Add and fix variable units

lthall:
No compiler change fixing units and comments.

To go in after #30120

`
CubeOrange,*,*,*,*,*,*,*
`

### DIFF
--- a/ArduCopter/mode.h
+++ b/ArduCopter/mode.h
@@ -427,7 +427,7 @@ public:
     bool is_autopilot() const override { return false; }
     bool init(bool ignore_checks) override;
     void exit() override;
-    // whether an air-mode aux switch has been toggled
+    // Called when air mode is enabled via AUX switch; prevents automatic reset to default air_mode state
     void air_mode_aux_changed();
     bool allows_save_trim() const override { return true; }
     bool allows_flip() const override { return true; }
@@ -438,9 +438,9 @@ protected:
     const char *name() const override { return "ACRO"; }
     const char *name4() const override { return "ACRO"; }
 
-    // get_pilot_desired_angle_rates - transform pilot's normalised roll pitch and yaw input into a desired lean angle rates
+    // get_pilot_desired_rates_cds - transform pilot's normalised roll pitch and yaw input into a desired lean angle rates
     // inputs are -1 to 1 and the function returns desired angle rates in centi-degrees-per-second
-    void get_pilot_desired_angle_rates(float roll_in, float pitch_in, float yaw_in, float &roll_out, float &pitch_out, float &yaw_out);
+    void get_pilot_desired_rates_cds(float roll_in_norm, float pitch_in_norm, float yaw_in_norm, float &roll_out_cds, float &pitch_out_cds, float &yaw_out_cds);
 
     float throttle_hover() const override;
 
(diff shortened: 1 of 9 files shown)


ArduCopter/mode.h
```diff
@@ -438,9 +438,9 @@ class ModeAcro : public Mode {
     const char *name() const override { return "ACRO"; }
     const char *name4() const override { return "ACRO"; }
 
-    // get_pilot_desired_angle_rates - transform pilot's normalised roll pitch and yaw input into a desired lean angle rates
+    // get_pilot_desired_rates_cds - transform pilot's normalised roll pitch and yaw input into a desired lean angle rates
     // inputs are -1 to 1 and the function returns desired angle rates in centi-degrees-per-second
-    void get_pilot_desired_angle_rates(float roll_in, float pitch_in, float yaw_in, float &roll_out, float &pitch_out, float &yaw_out);
+    void get_pilot_desired_rates_cds(float roll_norm_in, float pitch_norm_in, float yaw_norm_in, float &roll_out_cds, float &pitch_out_cds, float &yaw_out_cds);
```
peterbarker: Wondering about `norm` on the ends as a pseudo-unit

ArduCopter/mode_acro.cpp
```diff
@@ -10,8 +10,8 @@
 void ModeAcro::run()
 {
     // convert the input to the desired body frame rate
-    float target_roll, target_pitch, target_yaw;
-    get_pilot_desired_angle_rates(channel_roll->norm_input_dz(), channel_pitch->norm_input_dz(), channel_yaw->norm_input_dz(), target_roll, target_pitch, target_yaw);
+    float target_roll_cd, target_pitch_cd, target_yaw_cd;
+    get_pilot_desired_rates_cds(channel_roll->norm_input_dz(), channel_pitch->norm_input_dz(), channel_yaw->norm_input_dz(), target_roll_cd, target_pitch_cd, target_yaw_cd);
```
peterbarker: Kind of wondering why you removed the "angle" from the names

peterbarker (approved):
LGTM

ArduCopter/mode.h
rmackay9: yes, using "norm" as pseudo-units makes sense to me

ArduCopter/mode_acro.cpp
rmackay9: after a quick discussion with Leonard I think losing the "angle" is actually an improvement.  but he will add an "s" at the end.

ArduCopter/mode.h
lthall: I just moved it to the end.

ArduCopter/mode_acro.cpp
lthall: having the angle is inconsistent with what we do in the attitude controller and the name is getting quiet long with the units.

ArduCopter/mode.h
rmackay9: Looks like the arguments were renamed in the .cpp but the .h was missed.

lthall: Thanks and sorry!!!

PR #30393, opened 2025-06-18
Remove support for SP01 board

peterbarker:
there's no automatic setting for this board, so it would have been selected by parameter only.

This was a development board pre-PCNC01 - it wasn't widely distributed.

If we want this back we can create a full-on board for it...

### DIFF
--- a/libraries/AP_InertialSensor/AP_InertialSensor.cpp
+++ b/libraries/AP_InertialSensor/AP_InertialSensor.cpp
@@ -1264,12 +1264,6 @@ AP_InertialSensor::detect_backends(void)
                                                     hal.spi->get_device("bmi055_g"),
                                                     ROTATION_ROLL_180_YAW_90));
         break;
-        
-    case AP_BoardConfig::PX4_BOARD_SP01:
-        _fast_sampling_mask.set_default(1);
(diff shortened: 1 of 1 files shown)


peterbarker:
```
Durandal                            -24    *           -32     -24               -24    -24    -32
Pixhawk1-1M-bdshot                  0                  0       0                 0      0      0
```

robertlong13:
> it wasn't widely distributed.

I doubt anyone will mind, but just to clarify, does removing it help us at all? Most of these sections are compiled out if the board defines something like `HAL_***_PROBE_LIST`, so this only really helps the generic fmuv3 targets and the like, right?

peterbarker:
> > it wasn't widely distributed.
> 
> I doubt anyone will mind, but just to clarify, does removing it help us at all? Most of these sections are compiled out if the board defines something like `HAL_***_PROBE_LIST`, so this only really helps the generic fmuv3 targets and the like, right?

It's not for size, it's unused garbage cluttering up the code and pointlessly increasing maintenance overhead...

Those three blocks in the Compass, IMU and BARO I've modified at least three times myself, and I'm pretty sure others have also touched them.  All for a board which *nobody* is running.

robertlong13 (approved):

PR #26497, opened 2024-03-13
Move GPS instance parameters to a sub-object

peterbarker:
The new tool confirms correct transformation:
```
OK: GPS_TYPE=17.0 converted  to GPS_TYPE=17.0
OK: GPS_TYPE2=37.0 converted  to GPS2_TYPE=37.0
OK: GPS_GNSS_MODE=21.0 converted  to GPS_GNSS_MODE=21.0
OK: GPS_GNSS_MODE2=45.0 converted  to GPS2_GNSS_MODE=45.0
OK: GPS_RATE_MS=186.0 converted  to GPS_RATE_MS=186.0
OK: GPS_RATE_MS2=123.0 converted  to GPS2_RATE_MS=123.0
OK: GPS_POS1_X=3.75 converted  to GPS_POS_X=3.75
OK: GPS_POS2_X=6.9 converted  to GPS2_POS_X=6.9
OK: GPS_POS1_Y=2.75 converted  to GPS_POS_Y=2.75
OK: GPS_POS2_Y=5.9 converted  to GPS2_POS_Y=5.9
OK: GPS_POS1_Z=12.1 converted  to GPS_POS_Z=12.1
OK: GPS_POS2_Z=-6.9 converted  to GPS2_POS_Z=-6.9
OK: GPS_DELAY_MS=987.0 converted  to GPS_DELAY_MS=987.0
OK: GPS_DELAY_MS2=2345.0 converted  to GPS2_DELAY_MS=2345.0
OK: GPS_COM_PORT=19.0 converted  to GPS_COM_PORT=19.0
OK: GPS_COM_PORT2=100.0 converted  to GPS2_COM_PORT=100.0
OK: GPS_CAN_NODEID1=109.0 converted  to GPS_CAN_NODEID=109.0
OK: GPS_CAN_NODEID2=102.0 converted  to GPS2_CAN_NODEID=102.0
OK: GPS1_CAN_OVRIDE=34.0 converted  to GPS_CAN_OVRIDE=34.0
OK: GPS2_CAN_OVRIDE=67.0 converted  to GPS2_CAN_OVRIDE=67.0
OK: GPS_MB1_TYPE=1.0 converted  to GPS_MB_TYPE=1.0
OK: GPS_MB1_OFS_X=3.14 converted  to GPS_MB_OFS_X=3.14
OK: GPS_MB1_OFS_Y=2.18 converted  to GPS_MB_OFS_Y=2.18
[...]

PR #30397, opened 2025-06-19
hwdef: fix luminousbee5 parameters

robertlong13:
This fixes an errant rename of GPS_SBAS_TYPE to GPS1_SBAS_TYPE that occurred in https://github.com/ArduPilot/ardupilot/pull/26497.

I also removed the skybrush-fork parameters and added a blank include file for them to populate on their fork. This keeps our new CI checks (that I used to catch the above GPS mistake) from complaining without needing to keep a list of non-existent param patters to ignore. As a maintainer of some forks myself, I think this is an easier pattern. FYI @ntamas


ntamas (approved):
Looks good to me! One question: now that you fixed the param file, are you sure you want to remove `luminousbee5` from `param_check_all.py`? Wasn't the whole point of this PR to make the file pass the param checks?

robertlong13:
> Looks good to me! One question: now that you fixed the param file, are you sure you want to remove `luminousbee5` from `param_check_all.py`?

I've removed it from the list of boards to _skip_. When I added the check, I just dumped every failing board into that list so I could come back to them as separate PRs.

ntamas:
Ah okay then - sorry, it's 3am here so I kinda missed the point.

peterbarker (approved):

PR #30398, opened 2025-06-19
Plane/Copter/Rover/Tracker: 4.6.1 release

rmackay9:
this updates the Plane, Copter, Rover and Tracker release notes and version for the 4.6.1 official release

There is actually one code change that's been merged to the [ArduPilot-4.6 branch ](https://github.com/ArduPilot/ardupilot/commits/ArduPilot-4.6/) but I think it is non-functional so it's not in the release notes


peterbarker (approved):
LGTM

PR #30399, opened 2025-06-19
GCS_MAVLink: fix FTP burst complete indicator

ZachFleck42:
https://mavlink.io/en/services/ftp.html#reading-a-file-burstreadfile

Per the docs, `burst_complete` should be set to `1` on the last burst packet.
Currently, it's only being set on the 500th (hard-coded server-limit) packet. 
This change adjusts it so it's also set on the final burst packet if there are less than 500 response packets.

Note that an EOF NAK will still be sent afterwards, which isn't specified in the docs. This can be prevented by adding a `break;` to the check on line 533. However, I was hesitant to propose that in fear of breaking existing systems for detecting the end of a burst file read.

### DIFF
--- a/libraries/GCS_MAVLink/GCS_FTP.cpp
+++ b/libraries/GCS_MAVLink/GCS_FTP.cpp
@@ -525,7 +525,7 @@ void GCS_MAVLINK::ftp_worker(void) {
 
                             reply.opcode = FTP_OP::Ack;
                             reply.offset = request.offset + i * max_read;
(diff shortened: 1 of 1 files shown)


magicrub:
Nice find @ZachFleck42 ! I'll make sure to test this. However, if this does break anything then it's likely that CI will catch it for something like this. I'll check MAVProxy and mission planner

andyp1per (approved):
LGTM

magicrub:
Confirmed no change in behavior for MAVProxy and Mission Planner

PR #30401, opened 2025-06-19
Allow AP_MotorsTri to be compiled out of firmware

peterbarker:
```
###### Enabling TRI(AP_MOTORS_TRI_ENABLED) on copter costs -1760 bytes
###### Enabling TRI(AP_MOTORS_TRI_ENABLED) on plane costs -1704 bytes
###### Enabling TRI(AP_MOTORS_TRI_ENABLED) on rover costs 0 bytes
###### Enabling TRI(AP_MOTORS_TRI_ENABLED) on antennatracker costs 0 bytes
###### Enabling TRI(AP_MOTORS_TRI_ENABLED) on sub costs 8 bytes
###### Enabling TRI(AP_MOTORS_TRI_ENABLED) on blimp costs 0 bytes
```

No compiler output change unless the option is used:
```
Board                    AP_Periph  blimp  bootloader  copter  heli  iofirmware  plane  rover  sub
CubeOrange-periph-heavy  *                 *                                                   
Durandal                            *      *           *       *                 *      *      *
Hitec-Airspeed           *                 *                                                   
KakuteH7-bdshot                     *      *           *       *                 *      *      *
MatekF405                           *      *           *       *                 *      *      *
Pixhawk1-1M-bdshot                  *                  *       *                 *      *      *
f103-QiotekPeriph        *                 *                                                   
f303-Universal           *                 *                                                   
iomcu                                                                *                         
revo-mini                           *      *           *       *                 *      *      *
skyviper-v2450                                         *                                       
```

PR #30402, opened 2025-06-19
Pr/fix kakutef7 dshot

peterbarker:
This is a single commit on top of https://github.com/ArduPilot/ardupilot/pull/30401 which allows KakuteF7-bdshot to compile on Plane again.

It disables transmitter-based tuning and quadplane tricopter motors library.  Anybody running a tricopter frame with KakuteF7 will not be able to without a custom build

KakuteF7-bdshot ends up with about 900 bytes free after this.

### DIFF
--- a/Tools/scripts/build_options.py
+++ b/Tools/scripts/build_options.py
@@ -207,8 +207,8 @@ BUILD_OPTIONS = [
     Feature('Gimbal', 'XFROBOT', 'HAL_MOUNT_XFROBOT_ENABLED', 'Enable XFRobot gimbal', 0, "MOUNT"),
     Feature('Gimbal', 'VIEWPRO', 'HAL_MOUNT_VIEWPRO_ENABLED', 'Enable Viewpro gimbal', 0, "MOUNT"),
 
-    Feature('VTOL Type', 'TRI', 'AP_MOTORS_TRI_ENABLED', 'TriCopters', 0, None),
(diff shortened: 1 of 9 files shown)


IamPete1:
Tricopter is much more common on plane than copter, the ranger T1 for example. I would vote to remove hex and octas first (maybe they are already).

andyp1per:
> Tricopter is much more common on plane than copter, the ranger T1 for example. I would vote to remove hex and octas first (maybe they are already).

True, sorry I agree

Tools/scripts/build_options.py
```diff
@@ -207,6 +207,7 @@ def config_option(self):
     Feature('Gimbal', 'XFROBOT', 'HAL_MOUNT_XFROBOT_ENABLED', 'Enable XFRobot gimbal', 0, "MOUNT"),
     Feature('Gimbal', 'VIEWPRO', 'HAL_MOUNT_VIEWPRO_ENABLED', 'Enable Viewpro gimbal', 0, "MOUNT"),
 
+    Feature('VTOL Type', 'TRI', 'AP_MOTORS_TRI_ENABLED', 'TriCopters', 0, None),
```
tridge: should be VTOL Frame

rmackay9 (approved):

peterbarker:
Still a no-compiler-output-change:
```
Pixracer                            *      *           *       *                 *      *      *
```

Tools/scripts/build_options.py
peterbarker: Fixed

PR #30410, opened 2025-06-20
Remove HAL_BOARD_SUBTYPE_CHIBIOS_FMUV{3,5}

peterbarker:
This is the last of the `HAL_BOARD_SUBTYPE_CHIBIOS_*` defines.  Still have the `HAL_CHIBIOS_ARCH_FMUV*` defines yet.

Necessitated allowing boards to specify whether pins are analogue or digital for the purposes of the leak detector in their hwdefs.

I did consider something like `if (AP::hal().pin_is_analogue(pin)) {` instead of the arrangement here, but that may be rather complicated with the muxing available on H7, for example.  I decided not to over-think it.


peterbarker:
@Williangalvani this removes one of (what I consider to be) one of the biggest warts in the Sub (and soon to be Rover?!) codebase.  This change removes the knowledge about the shape of each board from the codebase and moves it to where it should be, in the hwdef file.

It also allows us to remove the remaining board subtypes meaning this sort of thing can't sneak in again (at least based on these defines...).

Williangalvani:
Nice! I just need to test it, but that's a nice cleanup!

peterbarker:
> Nice! I just need to test it, but that's a nice cleanup!

Thanks!  And also thanks for testing!

I tested this on a Cube and it seemed to work as expected.  I was going to get to Navigator in the next few days (it's on my desk, and I have instructions here somewhere :-) )

peterbarker:
> Nice! I just need to test it, but that's a nice cleanup!

Any progress on testing?

tridge:
needs approval by @Williangalvani too

Williangalvani:
sorry, we had a bad time testing it, for hw reasons.
I've sent a build to a colleague with a pixhawk, which is the target I'm concerned about right now. He should be able to test this tomorrow. I'm sorry for the hold up.

peterbarker:
Ping @Williangalvani :-)

peterbarker:
> Sorry for the delay, I finally got my hands on a good pixhawk to test. Everything working as expected here :]

Thanks!

Marked for MergeOnCIPass

PR #30411, opened 2025-06-20
AP_Scripting: docs_check.py: flake8 correctness

peterbarker:
... also avoid catching bare exception, handle the case where the regex isn't matching

```
pbarker@crun:~/rc/ardupilot(pr/docs-check-flake8)$ flake8 libraries/AP_Scripting/tests/docs_check.py
pbarker@crun:~/rc/ardupilot(pr/docs-check-flake8)$ 
```

Extracted from https://github.com/ArduPilot/ardupilot/pull/30272/files

### DIFF
--- a/libraries/AP_Scripting/tests/docs_check.py
+++ b/libraries/AP_Scripting/tests/docs_check.py
@@ -5,12 +5,15 @@
 '''
 Reads two lua docs files and checks for differences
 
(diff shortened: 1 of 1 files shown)


cclauss (approved):
Was the formatting done via psf/black?  LGTM.

peterbarker:
> Was the formatting done via psf/black? LGTM.

No, this was done by hand to ensure to be formatted the way we prefer (mostly just indenting stuff).  ie. if I'm changing a line to conform to our enforced rules I'll also change it to be as we like it.

PR #30412, opened 2025-06-20
Copter: Move to using radian setters in the AttitudeController

lthall:
To follow #30421

### DIFF
--- a/ArduCopter/mode.cpp
+++ b/ArduCopter/mode.cpp
@@ -461,23 +461,18 @@ void Copter::notify_flight_mode() {
 }
 
 // get_pilot_desired_angle - transform pilot's roll or pitch input into a desired lean angle
-// returns desired angle in centi-degrees
-void Mode::get_pilot_desired_lean_angles_cd(float &roll_out_cd, float &pitch_out_cd, float angle_max_cd, float angle_limit_cd) const
+// returns desired angle in radians
+void Mode::get_pilot_desired_lean_angles_rad(float &roll_out_rad, float &pitch_out_rad, float angle_max_rad, float angle_limit_rad) const
 {
+    // throttle failsafe check
     if (!rc().has_valid_input()) {
-        roll_out_cd = 0.0;
-        pitch_out_cd = 0.0;
+        roll_out_rad = 0.0;
+        pitch_out_rad = 0.0;
         return;
     }
 
     //transform pilot's normalised roll or pitch stick input into a roll and pitch euler angle command
-    float roll_out_deg;
-    float pitch_out_deg;
-    rc_input_to_roll_pitch(channel_roll->get_control_in()*(1.0/ROLL_PITCH_YAW_INPUT_MAX), channel_pitch->get_control_in()*(1.0/ROLL_PITCH_YAW_INPUT_MAX), angle_max_cd * 0.01,  angle_limit_cd * 0.01, roll_out_deg, pitch_out_deg);
-
-    // Convert to centi-degrees
-    roll_out_cd = roll_out_deg * 100.0;
-    pitch_out_cd = pitch_out_deg * 100.0;
+    rc_input_to_roll_pitch_rad(channel_roll->norm_input_dz(), channel_pitch->norm_input_dz(), angle_max_rad,  angle_limit_rad, roll_out_rad, pitch_out_rad);
 }
 
 // transform pilot's roll or pitch input into a desired velocity
@@ -489,13 +484,8 @@ Vector2f Mode::get_pilot_desired_velocity(float vel_max) const
         return vel;
     }
     // fetch roll and pitch inputs
-    float roll_out = channel_roll->get_control_in();
-    float pitch_out = channel_pitch->get_control_in();
-
-    // convert roll and pitch inputs to -1 to +1 range
-    float scaler = 1.0 / (float)ROLL_PITCH_YAW_INPUT_MAX;
-    roll_out *= scaler;
-    pitch_out *= scaler;
+    float roll_out = channel_roll->norm_input_dz();
+    float pitch_out = channel_pitch->norm_input_dz();
 
     // convert roll and pitch inputs into velocity in NE frame
     vel = Vector2f(-pitch_out, roll_out);
@@ -546,14 +536,14 @@ void Mode::zero_throttle_and_relax_ac(bool spool_up)
     } else {
         motors->set_desired_spool_state(AP_Motors::DesiredSpoolState::GROUND_IDLE);
     }
-    attitude_control->input_euler_angle_roll_pitch_euler_rate_yaw_cd(0.0f, 0.0f, 0.0f);
+    attitude_control->input_euler_angle_roll_pitch_euler_rate_yaw_rad(0.0f, 0.0f, 0.0f);
     attitude_control->set_throttle_out(0.0f, false, copter.g.throttle_filt);
 }
 
 void Mode::zero_throttle_and_hold_attitude()
 {
     // run attitude controller
-    attitude_control->input_rate_bf_roll_pitch_yaw_cds(0.0f, 0.0f, 0.0f);
+    attitude_control->input_rate_bf_roll_pitch_yaw_rads(0.0f, 0.0f, 0.0f);
     attitude_control->set_throttle_out(0.0f, false, copter.g.throttle_filt);
 }
 
@@ -594,7 +584,7 @@ void Mode::make_safe_ground_handling(bool force_throttle_unlimited)
     pos_control->relax_U_controller(0.0f);   // forces throttle output to decay to zero
     pos_control->update_U_controller();
     // we may need to move this out
-    attitude_control->input_euler_angle_roll_pitch_euler_rate_yaw_cd(0.0f, 0.0f, 0.0f);
+    attitude_control->input_euler_angle_roll_pitch_euler_rate_yaw_rad(0.0f, 0.0f, 0.0f);
 }
 
 /*
@@ -763,9 +753,9 @@ void Mode::land_run_horizontal_control()
         // there is any position estimate drift after touchdown. We
         // limit attitude to 7 degrees below this limit and linearly
         // interpolate for 1m above that
-        const float attitude_limit_cd = linear_interpolate(700, copter.aparm.angle_max, get_alt_above_ground_cm(),
+        const float attitude_limit_rad = linear_interpolate(radians(7), attitude_control->lean_angle_max_rad(), get_alt_above_ground_cm(),
                                                      g2.wp_navalt_min*100U, (g2.wp_navalt_min+1)*100U);
-        const float thrust_vector_max = sinf(cd_to_rad(attitude_limit_cd)) * GRAVITY_MSS * 100.0f;
+        const float thrust_vector_max = sinf(attitude_limit_rad) * GRAVITY_MSS * 100.0f;
         const float thrust_vector_mag = thrust_vector.xy().length();
         if (thrust_vector_mag > thrust_vector_max) {
             float ratio = thrust_vector_max / thrust_vector_mag;
@@ -818,13 +808,13 @@ void Mode::precland_retry_position(const Vector3f &retry_pos)
         // allow user to take control during repositioning. Note: copied from land_run_horizontal_control()
         // To-Do: this code exists at several different places in slightly different forms and that should be fixed
         if (g.land_repositioning) {
-            float target_roll_cd = 0.0f;
-            float target_pitch_cd = 0.0f;
+            float target_roll_rad = 0.0f;
+            float target_pitch_rad = 0.0f;
             // convert pilot input to lean angles
-            get_pilot_desired_lean_angles_cd(target_roll_cd, target_pitch_cd, loiter_nav->get_angle_max_cd(), attitude_control->get_althold_lean_angle_max_cd());
+            get_pilot_desired_lean_angles_rad(target_roll_rad, target_pitch_rad, loiter_nav->get_angle_max_rad(), attitude_control->get_althold_lean_angle_max_rad());
 
             // record if pilot has overridden roll or pitch
-            if (!is_zero(target_roll_cd) || !is_zero(target_pitch_cd)) {
+            if (!is_zero(target_roll_rad) || !is_zero(target_pitch_rad)) {
                 if (!copter.ap.land_repo_active) {
                     LOGGER_WRITE_EVENT(LogEvent::LAND_REPO_ACTIVE);
                 }
@@ -999,7 +989,7 @@ Mode::AltHoldModeState Mode::get_alt_hold_state(float target_climb_rate_cms)
 
 // transform pilot's yaw input into a desired yaw rate
 // returns desired yaw rate in centi-degrees per second
-float Mode::get_pilot_desired_yaw_rate_cds() const
+float Mode::get_pilot_desired_yaw_rate_rads() const
 {
     if (!rc().has_valid_input()) {
         return 0.0f;
@@ -1009,7 +999,7 @@ float Mode::get_pilot_desired_yaw_rate_cds() const
     const float yaw_in = channel_yaw->norm_input_dz();
 
     // convert pilot input to the desired yaw rate
-    return g2.command_model_pilot_y.get_rate() * 100.0 * input_expo(yaw_in, g2.command_model_pilot_y.get_expo());
+    return radians(g2.command_model_pilot_y.get_rate()) * input_expo(yaw_in, g2.command_model_pilot_y.get_expo());
 }
 
 // pass-through functions to reduce code churn on conversion;
--- a/ArduCopter/mode_acro.cpp
+++ b/ArduCopter/mode_acro.cpp
@@ -10,8 +10,8 @@
 void ModeAcro::run()
 {
     // convert the input to the desired body frame rate
-    float target_roll_cds, target_pitch_cds, target_yaw_cds;
-    get_pilot_desired_rates_cds(channel_roll->norm_input_dz(), channel_pitch->norm_input_dz(), channel_yaw->norm_input_dz(), target_roll_cds, target_pitch_cds, target_yaw_cds);
+    float target_roll_rads, target_pitch_rads, target_yaw_rads;
+    get_pilot_desired_rates_rads(channel_roll->norm_input_dz(), channel_pitch->norm_input_dz(), channel_yaw->norm_input_dz(), target_roll_rads, target_pitch_rads, target_yaw_rads);
 
     if (!motors->armed()) {
         // Motors should be Stopped
@@ -61,10 +61,10 @@ void ModeAcro::run()
     // run attitude controller
     if (g2.acro_options.get() & uint8_t(AcroOptions::RATE_LOOP_ONLY)) {
         // send rate commands to attitude controller (RATE_LOOP_ONLY bypasses full attitude stabilization)
-        attitude_control->input_rate_bf_roll_pitch_yaw_2_cds(target_roll_cds, target_pitch_cds, target_yaw_cds);
+        attitude_control->input_rate_bf_roll_pitch_yaw_2_rads(target_roll_rads, target_pitch_rads, target_yaw_rads);
     } else {
         // send rate commands to attitude controller with attitude stabilization
-        attitude_control->input_rate_bf_roll_pitch_yaw_cds(target_roll_cds, target_pitch_cds, target_yaw_cds);
+        attitude_control->input_rate_bf_roll_pitch_yaw_rads(target_roll_rads, target_pitch_rads, target_yaw_rads);
     }
 
     // output pilot's throttle without angle boost
@@ -103,10 +103,10 @@ float ModeAcro::throttle_hover() const
 }
 
 // Transform normalized pilot inputs (-1 to 1) into desired angular rates (centidegrees/second)
-void ModeAcro::get_pilot_desired_rates_cds(float roll_in_norm, float pitch_in_norm, float yaw_in_norm, float &roll_out_cds, float &pitch_out_cds, float &yaw_out_cds)
+void ModeAcro::get_pilot_desired_rates_rads(float roll_in_norm, float pitch_in_norm, float yaw_in_norm, float &roll_out_rads, float &pitch_out_rads, float &yaw_out_rads)
 {
-    float rate_delta_max_cds;
-    Vector3f rate_ef_level_cds, rate_bf_level_cds, rate_bf_request_cds;
+    float rate_delta_max_rads;
+    Vector3f rate_ef_level_rads, rate_bf_level_rads, rate_bf_request_rads;
 
     // apply circular limit to pitch and roll inputs
     float norm_in_length = norm(pitch_in_norm, roll_in_norm);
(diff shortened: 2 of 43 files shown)


ArduCopter/mode.cpp
```diff
@@ -461,23 +461,18 @@ void Copter::notify_flight_mode() {
...
-// returns desired angle in centi-degrees
-void Mode::get_pilot_desired_lean_angles_cd(float &roll_out_cd, float &pitch_out_cd, float angle_max_cd, float angle_limit_cd) const
+// returns desired angle in radians
+void Mode::get_pilot_desired_lean_angles_rad(float &roll_out_rad, float &pitch_out_rad, float angle_max_rad, float angle_limit_rad) const
 {
-    if (!rc().has_valid_input()) {
-        roll_out_cd = 0.0;
-        pitch_out_cd = 0.0;
+    // throttle failsafe check
+    if (copter.failsafe.radio || !rc().has_ever_seen_rc_input()) {
```
rmackay9: ah, this looks like a regression from a change that PeterB put in a few weeks ago

libraries/AC_AttitudeControl/AC_AttitudeControl.h
```diff
@@ -147,9 +147,8 @@ class AC_AttitudeControl {
     // get the slew yaw rate limit in deg/s
     float get_slew_yaw_max_degs() const;
 
-    // Returns the maximum allowed yaw rate for setpoint slewing (in rad/s).
-    // This limit applies in modes like Loiter or RTL to constrain sudden yaw changes.
-    float get_slew_yaw_max_rads() const { return radians(get_slew_yaw_max_degs()); }
+    // get the slew yaw rate limit in rad/s
+    float get_slew_yaw_max_rads();
```
rmackay9: this can probably be "const"

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1484,16 +1484,16 @@ void AC_PosControl::update_terrain()
...
+    const float accel_forward_mss = (accel_n_cmss * _ahrs.cos_yaw() + accel_e_cmss * _ahrs.sin_yaw()) * 0.01;
+    const float accel_right_mss = (-accel_n_cmss * _ahrs.sin_yaw() + accel_e_cmss * _ahrs.cos_yaw()) * 0.01;
 
     // update angle targets that will be passed to stabilize controller
-    pitch_target_cd = accel_mss_to_angle_deg(-accel_forward_cmss * 0.01) * 100;
-    float cos_pitch_target = cosf(pitch_target_cd * M_PI / 18000.0f);
-    roll_target_cd = accel_mss_to_angle_deg((accel_right_cmss * cos_pitch_target)*0.01) * 100;
+    pitch_target_rad = accel_mss_to_angle_rad(-accel_forward_mss);
+    float cos_pitch_target = cosf(pitch_target_rad);
+    roll_target_rad = accel_mss_to_angle_rad((accel_right_mss * cos_pitch_target));
```
rmackay9: we can remove the double brackets on this line

libraries/AC_AutoTune/AC_AutoTune.cpp
```diff
@@ -240,24 +240,24 @@ void AC_AutoTune::run()
         return;
     }
 
-    float desired_yaw_rate_cds;  // used during manual control
-    get_pilot_desired_rp_yrate_cd(desired_roll_cd, desired_pitch_cd, desired_yaw_rate_cds);
-
+    float desired_yaw_rate_rads = 0.0;   // used during manual control
+    get_pilot_desired_rp_yrate_rad(desired_roll_rad, desired_pitch_rad, desired_yaw_rate_rads);
+
```
rmackay9: tiny addition of whitespace

libraries/AC_AutoTune/AC_AutoTune.cpp
```diff
@@ -240,24 +240,24 @@ void AC_AutoTune::run()
         return;
     }
 
-    float desired_yaw_rate_cds;  // used during manual control
-    get_pilot_desired_rp_yrate_cd(desired_roll_cd, desired_pitch_cd, desired_yaw_rate_cds);
-
+    float desired_yaw_rate_rads = 0.0;   // used during manual control
```
rmackay9: I don't think we need to add the "= 0.0;" here because the line below will always set the variable to something

libraries/AC_AutoTune/AC_AutoTune.h
```diff
@@ -94,7 +94,7 @@ class AC_AutoTune
     virtual float get_pilot_desired_climb_rate_cms(void) const = 0;
 
     // get pilot input for designed roll and pitch, and yaw rate
-    virtual void get_pilot_desired_rp_yrate_cd(float &roll_cd, float &pitch_cd, float &yaw_rate_cds) = 0;
+    virtual void get_pilot_desired_rp_yrate_rad(float &roll_cd, float &pitch_cd, float &yaw_rate_cds) = 0;
```
rmackay9: need to fixup the argument's endings as well I think

libraries/AC_AutoTune/AC_AutoTune.cpp
```diff
@@ -750,18 +749,18 @@ void AC_AutoTune::get_poshold_attitude(float &roll_cd_out, float &pitch_cd_out,
       position. This ensures that autotune doesn't have to deal with
       more than 2.5 degrees of attitude on the axis it is tuning
      */
-    float target_yaw_cd = degrees(atan2f(pdiff.y, pdiff.x)) * 100;
+    float target_yaw_rad = atan2f(pos_error_neu_cm.y, pos_error_neu_cm.x);
     if (axis == AxisType::PITCH) {
         // for roll and yaw tuning we point along the wind, for pitch
         // we point across the wind
-        target_yaw_cd += 9000;
+        target_yaw_rad += cd_to_rad(9000);
```
rmackay9: we could use degrees(90) instead of cd_to_rad(9000) if we want to reduce the usage of centi-degrees in constants.  just an idea though

libraries/AC_Autorotation/AC_Autorotation.cpp
```diff
@@ -326,35 +326,34 @@ void AC_Autorotation::update_forward_speed_controller(float pilot_accel_norm)
...
     // Calculate roll and pitch targets from angles, negative accel for negative pitch (pitch forward)
-    Vector2f angle_target_deg = { accel_mss_to_angle_deg(-bf_accel_target_mss.x), // Pitch
-                              accel_mss_to_angle_deg(bf_accel_target_mss.y)}; // Roll
+    Vector2f angle_target_rad = { accel_mss_to_angle_rad(-bf_accel_target_mss.x), // Pitch
+                                  accel_mss_to_angle_rad(bf_accel_target_mss.y)}; // Roll
 
     // Ensure that the requested angles do not exceed angle max
-    // todo: FIX BUG: degrees limited in centidegrees
-    _limit_accel |= angle_target_deg.limit_length(_attitude_control->lean_angle_max_cd());
+    _limit_accel |= angle_target_rad.limit_length(_attitude_control->lean_angle_max_rad());
```
rmackay9, line 333: @Ferruccio1984, @MattKear FYI, we're fixing this angle limit here so this is a change in behaviour

libraries/AC_Avoidance/AC_Avoid.cpp
```diff
@@ -495,15 +495,15 @@ void AC_Avoid::adjust_velocity_z(float kP, float accel_cmss, float& climb_rate_c
 
 // adjust roll-pitch to push vehicle away from objects
 // roll and pitch value are in centi-degrees
```
rmackay9: need to update the comment

libraries/AP_Math/control.cpp
```diff
@@ -582,45 +582,52 @@ float input_expo(float input, float expo)
...
 float angle_deg_to_accel_mss(float angle_deg)
 {
-    return GRAVITY_MSS * tanf(radians(angle_deg));
+    return angle_rad_to_accel_mss(radians(angle_deg));
+}
+
+// accel_mss_to_angle_rad converts a maximum accel in m/s/s to a lean angle in radians
+float accel_mss_to_angle_rad(float accel_mss)
+{
+    return atanf((accel_mss/GRAVITY_MSS));
```
rmackay9: we can lose one pair of brackets here

libraries/AP_Math/control.h
```diff
@@ -150,15 +150,21 @@ float kinematic_limit(Vector3f direction, float max_xy, float max_z_pos, float m
...
-// returns roll and pitch euler angles in degrees
-void rc_input_to_roll_pitch(float roll_in_unit, float pitch_in_unit, float angle_max_deg, float angle_limit_deg, float &roll_out_deg, float &pitch_out_deg);
+float accel_mss_to_angle_deg(float accel_mss);
+
+// rc_input_to_roll_pitch_rad - transform pilot's normalised roll or pitch stick input into a roll and pitch euler angle command
+// roll_in_unit and pitch_in_unit - are normalised roll and pitch stick input
+// angle_max_rad - maximum lean angle from the z axis
+// angle_limit_rad - provides the ability to reduce the maximum output lean angle to less than angle_max_rad without changing the scaling
+// returns roll and pitch angle in radians
+void rc_input_to_roll_pitch_rad(float roll_in_unit, float pitch_in_unit, float angle_max_rad, float angle_limit_rad, float &roll_out_rad, float &pitch_out_rad);
```
rmackay9: don't need to change this but we're using "norm" these days it seems instead of "unit"

libraries/AP_Math/tests/test_rc_in_to_roll_pitch.cpp
```diff
@@ -19,10 +19,10 @@ TEST(RC2RPTest, Corners) {
     float pitch_val_deg[] = {-pc, pc, -pc, pc};
 
     for (uint i=0; i<ARRAY_SIZE(roll_val_deg); i++) {
-        rc_input_to_roll_pitch(roll_in[i], pitch_in[i], angle_max_deg, angle_limit_deg, roll_out_deg, pitch_out_deg);
+        rc_input_to_roll_pitch_rad(roll_in[i], pitch_in[i], angle_max_rad, angle_limit_rad, roll_out_rad, pitch_out_rad);
 
-        EXPECT_TRUE(fabsf(roll_out_deg  - roll_val_deg[i])  < 0.0005f);
-        EXPECT_TRUE(fabsf(pitch_out_deg - pitch_val_deg[i]) < 0.0005f);
+        EXPECT_TRUE(fabsf(roll_out_rad  - radians(roll_val_deg[i]))  < radians(0.001f));
```
rmackay9, line 24: it's a little funny that we've changed the constant from 0.0005 to 0.001

libraries/AP_Math/tests/test_rc_in_to_roll_pitch.cpp
```diff
@@ -31,25 +31,25 @@ TEST(RC2RPTest, Axes) {
...
                              0};
 
-    for (uint i=0; i<ARRAY_SIZE(roll_val_deg); i++) {
-        rc_input_to_roll_pitch(roll_in[i], pitch_in[i], angle_max_deg, angle_limit_deg, roll_out_deg, pitch_out_deg);
+    for (uint i=0; i<ARRAY_SIZE(roll_val_rad); i++) {
+        rc_input_to_roll_pitch_rad(roll_in[i], pitch_in[i], angle_max_rad, angle_limit_rad, roll_out_rad, pitch_out_rad);
 
-        EXPECT_TRUE(fabsf(roll_out_deg  - roll_val_deg[i])  < 0.00005f);
-        EXPECT_TRUE(fabsf(pitch_out_deg - pitch_val_deg[i]) < 0.00005f);
+        EXPECT_TRUE(fabsf(roll_out_rad  - roll_val_rad[i])  < radians(0.001f));
```
rmackay9, line 51: maybe here too we have an unintentional change to the check's constant?

libraries/AC_AutoTune/AC_AutoTune.cpp
lthall: radians(90)

easy mistake to make.

libraries/AP_Math/tests/test_rc_in_to_roll_pitch.cpp
lthall: Yeh, I think this is a little funny too. I assumed it was a floating point thing.

libraries/AP_Math/tests/test_rc_in_to_roll_pitch.cpp
lthall: Yeh, I think this is a little funny too. I assumed it was a floating point thing.

rmackay9: I think this test was failing because radians(0.00005) = 0.000000873 is hitting the limits of what we can represent using floats.  Let's leave the constant as 0.00005f

rmackay9: actually on second though, what you've done is very understandable so let's leave it like you've done it

libraries/AP_Math/tests/test_rc_in_to_roll_pitch.cpp
rmackay9: yes, this is fine, thanks!

libraries/AC_Avoidance/AC_Avoid.cpp
```diff
@@ -516,29 +516,29 @@ void AC_Avoid::adjust_roll_pitch(float &roll_cd, float &pitch_cd, float veh_angl
     get_proximity_roll_pitch_norm(roll_positive, roll_negative, pitch_positive, pitch_negative);
 
     // add maximum positive and negative percentages together for roll and pitch, convert to centi-degrees
```
rmackay9: we should probably fix this comment to "convert to radians"

libraries/AC_Autorotation/AC_Autorotation.cpp
MattKear: @rmackay9 thanks for the heads up and for fixing this.  I am happy with the change in behaviour.

libraries/AC_AutoTune/AC_AutoTune.cpp
```diff
@@ -750,18 +749,18 @@ void AC_AutoTune::get_poshold_attitude(float &roll_cd_out, float &pitch_cd_out,
...
     if (axis == AxisType::PITCH) {
         // for roll and yaw tuning we point along the wind, for pitch
         // we point across the wind
-        target_yaw_cd += 9000;
+        target_yaw_rad += radians(90);
     }
     // go to the nearest 180 degree mark, with 5 degree slop to prevent oscillation
-    if (fabsf(yaw_cd_out - target_yaw_cd) > 9500) {
-        target_yaw_cd += 18000;
+    if (fabsf(yaw_out_rad - target_yaw_rad) > cd_to_rad(9500)) {
```
rmackay9: "cd_to_rad(9500)" could be replaced with "radians(95)".  something similar one line lower could also be done.

MattKear (approved):

libraries/AC_AutoTune/AC_AutoTune.cpp
```diff
@@ -711,7 +710,7 @@ void AC_AutoTune::get_poshold_attitude(float &roll_cd_out, float &pitch_cd_out,
     }
 
     // don't go past 10 degrees, as autotune result would deteriorate too much
-    const float angle_max_cd = 1000;
+    const float angle_max_rad = cd_to_rad(1000);
```
rmackay9: this one could also become, "radians(10)"

ArduCopter/mode_systemid.cpp
```diff
@@ -268,24 +268,24 @@ void ModeSystemId::run()
                     gcs().send_text(MAV_SEVERITY_INFO, "SystemID Stopped: axis = 0");
                     break;
                 case AxisType::INPUT_ROLL:
-                    target_roll_cd += waveform_sample*100.0f;
+                    target_roll_rad += waveform_sample*100.0f;
```
rmackay9: I think we need to change this to, "target_roll_rad += radians(waveform_sample);".  below as well

ArduCopter/mode_poshold.cpp
```diff
@@ -82,11 +82,13 @@ void ModePosHold::run()
...
-    float target_roll_cd, target_pitch_cd;
-    get_pilot_desired_lean_angles_cd(target_roll_cd, target_pitch_cd, copter.aparm.angle_max, attitude_control->get_althold_lean_angle_max_cd());
+    float target_roll_rad, target_pitch_rad;
+    get_pilot_desired_lean_angles_rad(target_roll_rad, target_pitch_rad, attitude_control->lean_angle_max_rad(), attitude_control->get_althold_lean_angle_max_rad());
+    float target_roll_cd = rad_to_cd(target_roll_rad);
+    float target_pitch_cd = rad_to_cd(target_pitch_rad);
 
     // get pilot's desired yaw rate
-    float target_yaw_rate_cds = get_pilot_desired_yaw_rate_cds();
+    float target_yaw_rate_cds = rad_to_cd(get_pilot_desired_yaw_rate_rads());
```
rmackay9, line 91: should this be changed to "target_yaw_rate_rads" in poshold and sport modes to be consistent with Loiter?

ArduCopter/mode_follow.cpp
```diff
@@ -100,22 +100,22 @@ void ModeFollow::run()
...
+                yaw_rad = radians(target_heading_deg);
+                yaw_rate_rads = radians(target_heading_rate_degs);
                 break;
             }
 
             case AP_Follow::YAW_BEHAVE_DIR_OF_FLIGHT: {
                 // Face the direction of travel
                 if (vel_ofs_ne_cms.length_squared() > (100.0 * 100.0)) {
-                    yaw_cd = get_bearing_cd(Vector2f{}, vel_ofs_ne_cms);
+                    yaw_rad = vel_ofs_ne_cms.angle();
```
rmackay9, line 118: I don't think these are equivalent are they?  From a quick test I think they're off by 180deg

ArduCopter/mode_follow.cpp
```diff
@@ -100,22 +100,22 @@ void ModeFollow::run()
                 Vector3f accel_ned_mss;  // accel of lead vehicle
                 if (g2.follow.get_target_pos_vel_accel_NED_m(pos_ned_m, vel_ned_ms, accel_ned_mss))
                 if (pos_ned_m.xy().length_squared() > 1.0) {
-                    yaw_cd = get_bearing_cd(Vector2f{}, (pos_ned_m.xy() - pos_control->get_pos_target_NEU_cm().xy()).tofloat());
+                    yaw_rad = (pos_ned_m.xy() - pos_control->get_pos_target_NEU_cm().xy()).tofloat().angle();
```
rmackay9, line 103: similar to the comment below I think we should double check this calc to confirm there is no change in behaviour

ArduCopter/mode_flip.cpp
```diff
@@ -103,27 +103,27 @@ void ModeFlip::run()
...
+        flip_angle_rad = ahrs.get_pitch_rad() * pitch_dir;
     }
 
     // state machine
     switch (_state) {
 
     case FlipState::Start:
         // under 45 degrees request 400deg/sec roll or pitch
-        attitude_control->input_rate_bf_roll_pitch_yaw_cds(FLIP_ROTATION_RATE_CDS * roll_dir, FLIP_ROTATION_RATE_CDS * pitch_dir, 0.0);
+        attitude_control->input_rate_bf_roll_pitch_yaw_rads(cd_to_rad(FLIP_ROTATION_RATE_CDS) * roll_dir, cd_to_rad(FLIP_ROTATION_RATE_CDS) * pitch_dir, 0.0);
```
rmackay9: we convert FLIP_ROTATION_RATE_CDS to radians so many times it might be best to just change the defintion

ArduCopter/mode_acro_heli.cpp
```diff
@@ -127,29 +127,29 @@ void ModeAcro_Heli::run()
...
 // virtual_flybar - acts like a flybar by leaking target atttitude back to current attitude
-void ModeAcro_Heli::virtual_flybar( float &roll_out_cds, float &pitch_out_cds, float &yaw_out_cds, float pitch_leak, float roll_leak)
+void ModeAcro_Heli::virtual_flybar( float &roll_out_rads, float &pitch_out_rads, float &yaw_out_rads, float pitch_leak, float roll_leak)
 {
-    Vector3f rate_ef_level_cds, rate_bf_level_cds;
+    Vector3f rate_ef_level_rads, rate_bf_level_rads;
 
     // get attitude targets
-    const Vector3f att_target_cd = attitude_control->get_att_target_euler_cd();
+    const Vector3f att_target_rad = attitude_control->get_att_target_euler_rad();
```
rmackay9: it's an existing issue but this can be made a const reference.  so...

const Vector3f& att_target_rad = attitude_control->get_att_target_euler_rad();

ArduCopter/mode_acro.cpp
```diff
@@ -120,82 +120,82 @@ void ModeAcro::get_pilot_desired_rates_cds(float roll_in_norm, float pitch_in_no
...
 
         // get attitude targets
-        const Vector3f att_target_euler_cd = attitude_control->get_att_target_euler_cd();
+        const Vector3f att_target_euler_rad = attitude_control->get_att_target_euler_rad();
 
         // Calculate trainer mode earth frame rate command for roll
-        float roll_angle_cd = wrap_180_cd(att_target_euler_cd.x);
-        rate_ef_level_cds.x = -constrain_float(roll_angle_cd, -ACRO_LEVEL_MAX_ANGLE, ACRO_LEVEL_MAX_ANGLE) * g.acro_balance_roll;
+        float roll_angle_rad = wrap_PI(att_target_euler_rad.x);
+        rate_ef_level_rads.x = -constrain_float(roll_angle_rad, cd_to_rad(-ACRO_LEVEL_MAX_ANGLE), cd_to_rad(ACRO_LEVEL_MAX_ANGLE)) * g.acro_balance_roll;
```
rmackay9, line 140: similar to Flip mode we could consider changing the ACRO_LEVEL_MAX_ANGLE definition to be in radians.  The downside of doing this though is it would affect any external people/companies that have these definitions in their hwdef files... I've never heard of anyone overriding these values in an hwdef though

rmackay9:
I've done a 2nd review now and added some comments

I've added @andyp1per and @bnsgeyer because Acro and Heli are affected as well

ArduCopter/mode_poshold.cpp
lthall: Yes but I want to do sport and PosHold as a separate PR as they require more careful and focused work and reverse engineering.

ArduCopter/mode_acro.cpp
lthall: I would like to do this when we convert Sport mode.

ArduCopter/mode_follow.cpp
lthall: Yeh, this seems to be all sorts of messed up. The vector angle would normally be
`angle = atan2(y,x)`

But we seem to represent `atan2f(x, y)`.

Then we calculate the angle as
`M_PI_2 + atan2F(-x, y);`

It looks like we have got all this completly messed up.

lthall: It looks like:
`#define atan2F(x,y) atan2f(x,y)`
calls
`__MATHCALL_VEC (atan2,, (_Mdouble_ __y, _Mdouble_ __x));` (this is what I would expect)

if I read that correctly then we are doing the math correctly just swapping x and y labels. If that is the case this would be much easier to understand if we fix this.

lthall: `atan2(-x, y) = atan2(y, x)-pi/2`

therefore if we unwrap each side to -pi to pi

`unwrap(atan2(−x,y) + pi/2) = unwrap(atan2(y,x) − pi/2 +pi/2) = unwrap(atan2(y,x))`

But when we look at the range of each side
`atan2(−x,y) + pi/2`  has a range of -pi/2 to 3pi/2
`atan2(y,x)` has a range of -pi to pi

So the angle function is defined as:

// computes the angle of this vector in radians, from 0 to 2pi,
// from a unit vector(1,0); a (1,1) vector's angle is +M_PI/4
T angle(void) const;

This isn't correct, the range is -pi/2 to 3pi/2

I think the intent was to create an output that was 0 to 2pi but missed. The good thing is the angle is correct.

To fix this we would need to change:


```
    // computes the angle of this vector in radians, from -pi to pi,
    // from a unit vector(1,0); a (1,1) vector's angle is +M_PI/4
    T angle(void) const;
```
    
```
template <typename T>
T Vector2<T>::angle(void) const
{
    return atan2F(y, x);
}
```
  
And to make it clear what is happening 
#define atan2F(y,x) atan2f(y,x)

lthall: Now bearing is calculated:


// return bearing in centi-degrees between two positions
float get_bearing_cd(const Vector2f &origin, const Vector2f &destination)
{
    float bearing = rad_to_cd(atan2f(destination.y-origin.y, destination.x-origin.x));
    if (bearing < 0) {
        bearing += 36000.0f;
    }
    return bearing;
}

In radians from the origin this is
atan2f(y, x)

Same as above with a range of -pi to pi

lthall: So this is the same.

lthall: Fix and unit tests here: #30454

Ferruccio1984 (approved):
Thanks @lthall @rmackay9  for all this! I have left a little note in heli.cpp to update the dynamic flight check- not a blocker of course.

ArduCopter/mode.cpp
```diff
@@ -461,23 +461,18 @@ void Copter::notify_flight_mode() {
...
-        pitch_out_cd = 0.0;
+        roll_out_rad = 0.0;
+        pitch_out_rad = 0.0;
         return;
     }
 
     //transform pilot's normalised roll or pitch stick input into a roll and pitch euler angle command
-    float roll_out_deg;
-    float pitch_out_deg;
-    rc_input_to_roll_pitch(channel_roll->get_control_in()*(1.0/ROLL_PITCH_YAW_INPUT_MAX), channel_pitch->get_control_in()*(1.0/ROLL_PITCH_YAW_INPUT_MAX), angle_max_cd * 0.01,  angle_limit_cd * 0.01, roll_out_deg, pitch_out_deg);
```
andyp1per, line 476: The change to the input method makes this harder to review. Are these the same?

tridge:
I'm happy with the plane section of the PR

peterbarker:
```
// Acro Mode
#ifndef ACRO_LEVEL_MAX_ANGLE_DEG
 # define ACRO_LEVEL_MAX_ANGLE_DEG      30 // maximum lean angle in trainer mode measured in centidegrees
#endif
#ifdef ACRO_LEVEL_MAX_ANGLE
#error "Please update your hwdef to use ACRO_LEVEL_MAX_ANGLE_DEG, not ACRO_LEVEL_MAX_ANGLE"
#endif // ACRO_LEVEL_MAX_ANGLE

```

ArduCopter/mode_acro.cpp
```diff
@@ -120,82 +120,82 @@ void ModeAcro::get_pilot_desired_rates_cds(float roll_in_norm, float pitch_in_no
...
+            } else if (roll_angle_rad < -angle_max_rad) {
+                rate_ef_level_rads.x += sqrt_controller(-angle_max_rad - roll_angle_rad, radians(g2.command_model_acro_rp.get_rate()) / cd_to_rad(ACRO_LEVEL_MAX_OVERSHOOT), attitude_control->get_accel_roll_max_radss(), G_Dt);
             }
 
-            if (pitch_angle_cd > angle_max_cd) {
-                rate_ef_level_cds.y += sqrt_controller(angle_max_cd - pitch_angle_cd, g2.command_model_acro_rp.get_rate() * 100.0 / ACRO_LEVEL_MAX_OVERSHOOT, attitude_control->get_accel_pitch_max_cdss(), G_Dt);
-            } else if (pitch_angle_cd < -angle_max_cd) {
-                rate_ef_level_cds.y += sqrt_controller(-angle_max_cd - pitch_angle_cd, g2.command_model_acro_rp.get_rate() * 100.0 / ACRO_LEVEL_MAX_OVERSHOOT, attitude_control->get_accel_pitch_max_cdss(), G_Dt);
+            if (pitch_angle_rad > angle_max_rad) {
+                rate_ef_level_rads.y += sqrt_controller(angle_max_rad - pitch_angle_rad, radians(g2.command_model_acro_rp.get_rate()) / cd_to_rad(ACRO_LEVEL_MAX_OVERSHOOT), attitude_control->get_accel_roll_max_radss(), G_Dt);
```
andyp1per: Think this should be get_accel_pitch_max_radss()?

ArduCopter/mode_acro.cpp
```diff
@@ -120,82 +120,82 @@ void ModeAcro::get_pilot_desired_rates_cds(float roll_in_norm, float pitch_in_no
...
+            } else if (pitch_angle_rad < -angle_max_rad) {
+                rate_ef_level_rads.y += sqrt_controller(-angle_max_rad - pitch_angle_rad, radians(g2.command_model_acro_rp.get_rate()) / cd_to_rad(ACRO_LEVEL_MAX_OVERSHOOT), attitude_control->get_accel_roll_max_radss(), G_Dt);
```
andyp1per: ditto

andyp1per (changes requested):

ArduCopter/mode.cpp
lthall: Yes both Randy and I believe this is equivalent.

andyp1per (approved):

ArduCopter/mode_acro.cpp
rmackay9: very nice, txs for the review!

bnsgeyer (approved):
LGTM

peterbarker:
Flew this one yesterday on a 7" quad without incident.

rmackay9 (approved):
Thanks for all this.  I've gone through it all line-by-line for a 3rd time and with the latest fix I think it's ready to merge now

rmackay9:
Just for the record, despite all the reviews we managed to break simple mode (and super simple mode) with this PR.  The issue is here https://github.com/ArduPilot/ardupilot/issues/33936 and the fix is here: https://github.com/ArduPilot/ardupilot/pull/33947

PR #30369, opened 2025-06-16
colcon: Use current flake8 in GitHub Action

cclauss:
This GitHub Action fails occasionally, so let's upgrade its flake8 to see if that provides more stable runs.

https://pypi.org/project/flake8 --> v7.2.0

flake8: Both the local issue and the flake8 issue it points to were closed 4 or 5 years ago, so let's test with current flake8.

PR #30396, opened 2025-06-18
Tools: ros2: improve stability of DDS CI tests

srmainwaring:
Ongoing attempt to improve stability in the DDS `colcon` CI tests.

- The tests that run the `micro_ros_agent` with `serial` transport require a temporary directory for the virtual devices `tty0` and `tty1` managed by `socat`. This changes the fixture that creates the temp `dev` folder to have `function` rather than `module` scope. This ensures that the directory is torn down after every test.

- Other fixtures use `function` scope by default, this makes that explicit.

- Use `extras_require` in `setup.py` as `tests_require` is removed from newer versions of setuptools. See discussion and suggestions in: https://github.com/ros2/ros2_documentation/issues/4727

- Increase `pytest` verbosity and run stepwise.

- Increase the timeout while waiting for processes to start and synchronise

- Ensure that the `msg_event_object` in the test node callbacks is set last. The callback runs asynchronously and once the event is set the main fixture can proceed to check the event state and any additional flags, and then shutdown rclpy. Suspect this can result in an invalid context for the logger, and this exception may lead the fixture to not shut down cleanly, or misreport the state.

## Motivation

Example from: https://github.com/ArduPilot/ardupilot/actions/runs/15767553194/job/44448010803?pr=30409

The test fails when a SIGINT is sent to the SITL process:

```bash
[mavproxy.py -33] AP: EKF3 IMU1 is using GPS
[INFO] [dds_udp.parm --synthetic-clock -34]: sending signal 'SIGINT' to process[dds_udp.parm --synthetic-clock -34]
[INFO] [mavproxy.py -33]: sending signal 'SIGINT' to process[mavproxy.py -33]
```

At this point it's not clear whether the ROS 2 node `AirspeedTester` has been started (need to add further logging), however it appears that the test framework is attempting to kill the SITL process before the message checks have been run. 

Full log below

<details>

```bash
=================================== FAILURES ===================================
____________________ test_dds_udp_airspeed_msg_recv_copter _____________________
/opt/ros/humble/lib/python3.10/site-packages/launch_pytest/plugin.py:438: in inner
    on_shutdown()
/opt/ros/humble/lib/python3.10/site-packages/launch_pytest/fixture.py:36: in finalize_launch_service
    assert rc == 0, f"{eprefix} launch service failed when finishing, return code '{rc}'"
E   AssertionError: When running test test_dds_udp_airspeed_msg_recv_copter launch service failed when finishing, return code '1'
---------------------------- Captured stdout setup -----------------------------
[...]

PR #30414, opened 2025-06-20
colcon: Use current MAVProxy in GitHub Action

cclauss:
This GitHub Action fails occasionally, so let's upgrade its MAVProxy to see if that provides more stable runs.
* https://pypi.org/project/MAVProxy --> v1.8.71

Given the changes in MAVProxy, should we upgrade?
* https://github.com/ArduPilot/MAVProxy/releases
* https://github.com/ArduPilot/ardupilot/pull/30369#issuecomment-2975476125

---

`pip install -U` can be quite opaque to readers because `pip install` currently has six `--u*` options.

% `python3 -m pip install --help | grep "\-\-u"`
```
  --user                      Install to the Python user install directory for
  -U, --upgrade               Upgrade all specified packages to the newest
  --upgrade-strategy <upgrade_strategy>
  --use-pep517                Use PEP 517 for building source distributions
  --use-feature <feature>     Enable new functionality, that may be backward
  --use-deprecated <feature>  Enable deprecated functionality, that will be
```

### DIFF
--- a/.github/workflows/colcon.yml
+++ b/.github/workflows/colcon.yml
@@ -183,12 +183,12 @@ jobs:
       - name: Install MAVProxy
         run: |
           # Install this specific version just to prevent rolling updates.
-          # These are the latest available
-          python3 -m pip install MAVProxy==1.8.67
+          # See the latest available at https://pypi.org/project/MAVProxy
+          python3 -m pip install MAVProxy==1.8.71
       - name: Install local pymavlink
         working-directory: ./src/ardupilot/modules/mavlink/pymavlink
         run: |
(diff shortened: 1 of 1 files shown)


srmainwaring (approved):
LGTM. Though I don't think that the MAVProxy version is the reason for the CI failures in the DDS CI tests.

I suspect the issue is a timeout in one of the processes awaiting synchronisation managed by the ROS 2 `launch_pytest` package. (https://github.com/ArduPilot/ardupilot/pull/30396)

Ryanf55 (approved):
I like the switch from shorthand -U. Makes it clearer what's being done.

PR #22127, opened 2022-11-05
Plane: Quadplane: correct units used in rate_bf_yaw_target

IamPete1:
`get_gyro` is rad/sec `rate_bf_yaw_target` is expecting centi-deg/sec.

This is the original intent, but the current code is effectively setting to 0. Changing to that might be a better fix.  

Thanks to @skoch1s for finding the issue. https://github.com/ArduPilot/ardupilot/pull/22109 https://github.com/ArduPilot/ardupilot/issues/22103

PR #30164, opened 2025-05-25
20250525 fix radians

lthall:
This PR changes all 

`radians(variable * 0.01f);`

to

`variable * radians(0.01f);`

```
Board,blimp,bootloader,copter,heli,plane,rover,sub
CubeOrange,-56,*,-208,-248,-240,-120,-184
```

PR #30415, opened 2025-06-20
QuadPlane: fix transition yaw suppression

robertlong13:
This PR will need careful testing, because it is changing a long-standing (bad) behavior. I'll be deploying this on the Carbonix birds immediately, and I've extensively tested in RealFlight, but we should get some others.

During transition the VTOL yaw controller is _intended_ to be suppressed (except for vectored tiltrotors). This is to avoid wing twist since on quadcopters yaw is achieved by lifting more from one diagonal pair or another. As implemented, the yaw controller tries to hold zero yaw rate, which is not what was intended. This was due to a radian/centidegree unit confusion when the line was first added. Later, in #22127, the unit confusion was cleared up, but to avoid risking a behavior change, it was decided to explicitly set the rate command to 0. This can and does lead to large yaw outputs, even in normal straight-line transitions.

The problem is even worse if you don't use level-transition and turn before transition is complete, especially if you have a good/tight yaw controller (this is the case with Ottano, which has great yaw control and a very long time to transition). Also, with the current behavior, during any QAssist, the VTOL controller actively fights against any coordinated turn. If a control surface fails, and you try to limp home on force QAssist, things might go very badly.

This PR lets you select between two behaviors:
- VTOL controller actively assists with turn coordination
- VTOL yaw controller is (properly) suppressed

Neither is exactly the same as the current behavior, but I'd say the former is closer to the current behavior, so I chose that as the default. In a straight-line transition, assisting with turn coordination is the same as holding 0 yaw rate. And in a turning transition, it leads to significantly smaller yaw outputs and better flight. One could argue the default should be to suppress yaw, since that's safer for quadplanes with poor stiffness, but they've been managing with the current worse behavior for years with apparently no issue. I can go either way on the default, I just need the option.

Next I'll post some logs and graphs from my RealFlight testing.

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -1569,7 +1569,7 @@ void SLT_Transition::update()
             // is needed to maintain good control in forward
             // transitions
             quadplane.attitude_control->reset_yaw_target_and_rate();
-            quadplane.attitude_control->rate_bf_yaw_target(0.0);
+            quadplane.attitude_control->rate_bf_yaw_target(rad_to_cd(quadplane.ahrs.get_gyro().z));
         }
         if (quadplane.tiltrotor.enabled() && !quadplane.tiltrotor.has_fw_motor()) {
             // tilt rotors without dedicated fw motors do not have forward throttle output in this stage
@@ -1645,7 +1645,7 @@ void SLT_Transition::update()
         // yaw control throughout the transition
         if (!quadplane.tiltrotor.is_vectored()) {
             quadplane.attitude_control->reset_yaw_target_and_rate();
-            quadplane.attitude_control->rate_bf_yaw_target(0.0);
+            quadplane.attitude_control->rate_bf_yaw_target(rad_to_cd(quadplane.ahrs.get_gyro().z));
         }
         break;
     }
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -288,6 +288,7 @@ const AP_Param::GroupInfo QuadPlane::var_info[] = {
     // @Bitmask: 20: Force RTL mode-forces RTL mode on rc failsafe in VTOL modes overriding bit 5(USE_QRTL)
     // @Bitmask: 21: Tilt rotor-tilt motors up when disarmed in FW modes (except manual) to prevent ground strikes.
     // @Bitmask: 22: Scale FF by the ratio of VTOL to plane angle P gains in Position 1 phase of transition into VTOL flight as well as reducing VTOL angle P based on airspeed.
+    // @Bitmask: 23: Suppress VTOL yaw stabilization during transition (ignored on vectored-yaw tiltrotors)
     AP_GROUPINFO("OPTIONS", 58, QuadPlane, options, 0),
 
     AP_SUBGROUPEXTENSION("",59, QuadPlane, var_info2),
@@ -1561,7 +1562,7 @@ void SLT_Transition::update()
         }
         quadplane.hold_hover(climb_rate_cms);
 
-        if (!quadplane.tiltrotor.is_vectored()) {
+        if (!quadplane.tiltrotor.is_vectored() && quadplane.option_is_set(QuadPlane::OPTION::SUPPRESS_YAW_TRANSITION)) {
             // set desired yaw to current yaw in both desired angle
             // and rate request. This reduces wing twist in transition
             // due to multicopter yaw demands. This is disabled when
@@ -1643,7 +1644,7 @@ void SLT_Transition::update()
         // control surfaces at this stage.
         // We disable this for vectored yaw tilt rotors as they do need active
         // yaw control throughout the transition
-        if (!quadplane.tiltrotor.is_vectored()) {
+        if (!quadplane.tiltrotor.is_vectored() && quadplane.option_is_set(QuadPlane::OPTION::SUPPRESS_YAW_TRANSITION)) {
             quadplane.attitude_control->reset_yaw_target_and_rate();
             quadplane.attitude_control->rate_bf_yaw_target(rad_to_cd(quadplane.ahrs.get_gyro().z));
         }
--- a/ArduPlane/quadplane.h
+++ b/ArduPlane/quadplane.h
@@ -608,6 +608,7 @@ private:
         FS_RTL=(1<<20),
         DISARMED_TILT_UP=(1<<21),
         SCALE_FF_ANGLE_P=(1<<22),
+        SUPPRESS_YAW_TRANSITION=(1<<23),
     };
     bool option_is_set(OPTION option) const {
         return (options.get() & int32_t(option)) != 0;


robertlong13:
First I did 3 straight-line transitions on the Titan Cobra in RealFlight. This was interesting as I hadn't realized before that the large yaw/roll oscillation I get at the end of transition is a direct result of the yaw controller not being suppressed. The three tests:
1. Current behavior, holding 0 yaw rate
2. Intended behavior, suppressing yaw controller
3. Coordinated turn assistance

Logs: 
[StraightTransition.zip](https://github.com/user-attachments/files/20831901/StraightTransition.zip)
[TurningTransition_0.zip](https://github.com/user-attachments/files/20831916/TurningTransition_0.zip)
[TurningTransition_1.zip](https://github.com/user-attachments/files/20831917/TurningTransition_1.zip)
[TurningTransition_2.zip](https://github.com/user-attachments/files/20831922/TurningTransition_2.zip)


0_CurrentBehavior
1_IntendedBehavior
2_Coordination

As we can see, the coordinated turn option doesn't do any _worse_ than the current behavior. It's basically identical for straight transitions. The oscillation starts as soon as the VTOL control starts phasing out. I wasn't able to get this oscillation to go away other than reducing Q_TRANSITION_TIME to 500ms. The suppressed-yaw version looks great though.

Next, I did 3 more flights with a 180 transition, sending the aircraft to a waypoint directly behind it, as a bit of a torture test. Additionally, I did some figure-8s with force-qassist on, to demonstrate how bad the current behavior is during qassist. In the plots below, the transition is on the left, and the figure-8s are on the right.

0_CurrentBehavior
1_IntendedBehavior
2_Coordination

The current behavior is terrible. It leads to maxing out the yaw output. The turn coordination option has brief moments of large output, but nowhere near as much as the current behavior, and it's actually helping. Here is the ground track for the three tests too:
0_CurrentBehavior_Map
1_IntendedBehavior_Map
2_Coordination_Map

In the current behavior, the aircraft is max banked over, but refusing to turn until enough airspeed is reached to swamp out the VTOL controller.

ArduPlane/quadplane.h
```diff
@@ -608,6 +608,7 @@ class QuadPlane
         FS_RTL=(1<<20),
         DISARMED_TILT_UP=(1<<21),
         SCALE_FF_ANGLE_P=(1<<22),
+        SUPPRESS_YAW_TRANSITION=(1<<23),
```
IamPete1, line 611: I think bit 22 (23 bits) is the limit of the mantissa. The problem is float transport of parameters over MAVLink. I guess we could add a auto test for it. But I would expect at some combination of high bits the lower bits start getting lost.

ArduPlane/quadplane.cpp
```diff
@@ -1561,15 +1562,15 @@ void SLT_Transition::update()
...
+        if (!quadplane.tiltrotor.is_vectored() && quadplane.option_is_set(QuadPlane::OPTION::SUPPRESS_YAW_TRANSITION)) {
             // set desired yaw to current yaw in both desired angle
             // and rate request. This reduces wing twist in transition
             // due to multicopter yaw demands. This is disabled when
             // using vectored yaw for tilt-rotors as the yaw control
             // is needed to maintain good control in forward
             // transitions
             quadplane.attitude_control->reset_yaw_target_and_rate();
-            quadplane.attitude_control->rate_bf_yaw_target(0.0);
+            quadplane.attitude_control->rate_bf_yaw_target(rad_to_cd(quadplane.ahrs.get_gyro().z));
```
IamPete1, line 1573: I'm sure @lthall would be happy for us to change this to a native radians method (https://github.com/ArduPilot/ardupilot/pull/30164). Looks like the two cases updated here are the only users. The method immediately calls `cd_to_rad`.

IamPete1: Also it looks like the comment on the method in attitude control is wrong.

IamPete1 (commented):
This does effectively mean were going open loop in yaw. I can imagine some situations where this would be bad:

Large yaw offset required in hover due to imbalance/misalignment. In this case the vehicle would just spin. We have to deal with being in airspeed wait indefinitely (for example FBWA with 0 throttle)

Transition from hover with a tail wind. A user might have disabled weather vaining for take off and be hovering with the tail into the wind. Once transition is started the vehicle will be rapidly pushed round into the wind, effectively weather vaining despite that being disabled.

ArduPlane/quadplane.cpp
tridge, line 1573: this looks very dangerous to me. It means we chase the filtered yaw with no limit. This could lead to us trying to spin about the Z axis.
If you really want to disable the yaw control then we should also the AP_Motors controller to disable yaw control. We have a function motors->disable_yaw_torque() for that which we use for tilt vectored frames. 
I think what we need is this:
 - at the start of a transition aim for zero yaw rate as we do now
 - as airspeed increases decrease the yaw control at the AP_Motors level, with a scale factor, so that at the transition speed (or maybe half the transition speed?) it is zero. We could change disable_yaw_torque() to take a single loop float multiplier

I think we could add the option to use the coordinated turn yaw rate as well.

robertlong13:
> I can imagine some situations where this would be bad:

Those are good arguments for why the default should be what I've set it to. Though, I wouldn't totally oppose getting rid of the option altogether and just deleting the whole if-block.

ArduPlane/quadplane.cpp
robertlong13: > this looks very dangerous to me

This isn't some new idea of mine, this is what the line was before https://github.com/ArduPilot/ardupilot/pull/22127, just now with the correct units. All I'm doing is giving the option (defaulted to off) to do what this line originally intended, which I assume someone felt strongly was important, and figured others felt that way. If not, I'm open to just deleting this if-block.

> at the start of a transition aim for zero yaw rate as we do now

That is very dangerous in my opinion. The only option should be coordinated turn, not zero. You get a coordinated turn automatically by skipping this if-block (which is what this PR is doing by default). Holding zero yaw is only safe if wings are level. Rolling 45degrees left and saturating the yaw to try to stay straight is crazy.

robertlong13: To be clear, `quadplane.hold_hover(climb_rate_cms)`, a few lines up sets the target yaw rate to coordinate the turn. We throw it away here (and I think we shouldn't)

IamPete1: I don't think changing at the motors level is the correct thing to do, we could add a new attitude control method that ensures there is zero rate demand. That is what this is is trying to do. A method that always results in 0 yaw error will never ask motors to yaw.

robertlong13:
Splitting this into two PRs. I'll keep this one for trying to implement Tridge's recommendation of phasing out yaw output as you pick up airspeed. But for now, I actually think we should just delete these lines.

IamPete1:
I wonder if it would be sufficient to run the yaw controller as normal but decay the I terms. That should mean that no large output ever builds up. You could bring in the I limit from the param value to 0 as the transition progresses. Or do a leaky I term like heli that leaks down faster with airspeed.

kmk142789 (approved):

robertlong13:
Closing this as I'm really happy with the coordinated turns in https://github.com/ArduPilot/ardupilot/pull/30434 and I don't think we need the suppressed behavior anymore at all.

PR #30416, opened 2025-06-20
AP_GSOF: Protect against corrupt output_length

Ryanf55:
# Details

* If the serial stream has a two byte corruption that passes the summing checksum, and the corrupt byte is output_length or msg.length, it can cause a fault
* Added a test that isolates the issue

# Verification

Uncomment the GTEST_SKIP line
```c++
    // GTEST_SKIP() << "There is not yet a convention for loading in a data file in a cross-platform way in AP for unit tests";
```

Compile the tests.
```
./waf configure --board sitl --debug
./waf tests
```

set a breakpoint in the check, and verify we can detect corruption

```
ryan@ryan-thinkpad:~/Dev/ardupilot_ws/src/ardupilot$ gdb --args ./build/sitl/tests/test_gsof ---gtest_filter=AP_GSOF.packet1_corrupt_bytes
>>> b AP_GSOF:108
>>> r
```

image


libraries/AP_GSOF/AP_GSOF.cpp
```diff
@@ -100,7 +100,11 @@ AP_GSOF::process_message(MsgTypes& parsed_msgs)
             a++;
             const uint8_t output_length = msg.data[a];
             a++;
-            // TODO handle corruption on output_length causing buffer overrun?
+
+            if ((a + output_length) > msg.length) {
```
peterbarker, line 104: You also get a fun effect in here if output_length is zero.  I think we need an additional check...

```suggestion
            if (output_length == 0 or (a + output_length) > msg.length) {
```

PR #30417, opened 2025-06-20
Make sure CRSF bootstrap completes when not using passthrough

andyp1per:
If you are not using passthrough the version bootstrap never completes.


tridge (approved):

rmackay9:

PR #30425, opened 2025-06-22
Copter: Remove most of the remaining _cd() functions

lthall:
To follow #30412

```
CubeOrange,*,*,-16,16,40,*,40
```

### DIFF
--- a/ArduCopter/Attitude.cpp
+++ b/ArduCopter/Attitude.cpp
@@ -56,8 +56,8 @@ void Copter::update_throttle_hover()
     float throttle = motors->get_throttle();
 
     // calc average throttle if we are in a level hover.  accounts for heli hover roll trim
-    if (throttle > 0.0f && fabsf(vel_d_ms) < 0.6 &&
-        fabsf(ahrs.roll_sensor-attitude_control->get_roll_trim_cd()) < 500 && labs(ahrs.pitch_sensor) < 500) {
+    if ((throttle > 0.0f) && (fabsf(vel_d_ms) < 0.6) &&
+        (fabsf(ahrs.get_roll_rad() - attitude_control->get_roll_trim_rad()) < radians(5)) && (labs(ahrs.get_pitch_rad()) < radians(5))) {
         // Can we set the time constant automatically
         motors->update_throttle_hover(0.01f);
 #if HAL_GYROFFT_ENABLED
--- a/ArduCopter/RC_Channel_Copter.cpp
+++ b/ArduCopter/RC_Channel_Copter.cpp
@@ -793,13 +793,13 @@ void RC_Channels_Copter::auto_trim_run()
             return;
         }
         // calculate roll trim adjustment, divisor set subjectively to give same "feel" as previous RC input method
-        float roll_trim_adjustment = radians(copter.attitude_control->get_att_target_euler_cd().x / 2000.0f);
+        float roll_trim_adjustment_rad = copter.attitude_control->get_att_target_euler_rad().x / 20.0f;
 
         // calculate pitch trim adjustment, divisor set subjectively to give same "feel" as previous RC input method
-        float pitch_trim_adjustment = radians(copter.attitude_control->get_att_target_euler_cd().y / 2000.0f);
+        float pitch_trim_adjustment_rad = copter.attitude_control->get_att_target_euler_rad().y / 20.0f;
 
         // add trim to ahrs object, but do not save to permanent storage:
-        AP::ahrs().add_trim(roll_trim_adjustment, pitch_trim_adjustment, false);
+        AP::ahrs().add_trim(roll_trim_adjustment_rad, pitch_trim_adjustment_rad, false);
 }
 
 #endif  // AP_COPTER_AHRS_AUTO_TRIM_ENABLED
--- a/ArduCopter/crash_check.cpp
+++ b/ArduCopter/crash_check.cpp
@@ -124,8 +124,8 @@ void Copter::thrust_loss_check()
 
     // check for desired angle over 15 degrees
     // todo: add thrust angle to AC_AttitudeControl
-    const Vector3f angle_target = attitude_control->get_att_target_euler_cd();
-    if (sq(angle_target.x) + sq(angle_target.y) > sq(THRUST_LOSS_CHECK_ANGLE_DEVIATION_CD)) {
+    const Vector3f& angle_target_rad = attitude_control->get_att_target_euler_rad();
+    if (angle_target_rad.xy().length_squared() > sq(cd_to_rad(THRUST_LOSS_CHECK_ANGLE_DEVIATION_CD))) {
         thrust_loss_counter = 0;
         return;
     }
--- a/ArduCopter/land_detector.cpp
+++ b/ArduCopter/land_detector.cpp
@@ -111,8 +111,8 @@ void Copter::update_land_detector()
 #endif
 
         // check for aggressive flight requests - requested roll or pitch angle below 15 degrees
-        const Vector3f angle_target = attitude_control->get_att_target_euler_cd();
-        bool large_angle_request = angle_target.xy().length() > LAND_CHECK_LARGE_ANGLE_CD;
+        const Vector3f& angle_target_rad = attitude_control->get_att_target_euler_rad();
+        bool large_angle_request = angle_target_rad.xy().length() > cd_to_rad(LAND_CHECK_LARGE_ANGLE_CD);
         SET_LOG_FLAG(large_angle_request, LandDetectorLoggingFlag::LARGE_ANGLE_REQUEST);
 
         // check for large external disturbance - angle error over 30 degrees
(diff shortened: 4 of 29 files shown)


ArduCopter/Attitude.cpp
```diff
@@ -57,7 +57,7 @@ void Copter::update_throttle_hover()
 
     // calc average throttle if we are in a level hover.  accounts for heli hover roll trim
     if (throttle > 0.0f && fabsf(vel_d_ms) < 0.6 &&
-        fabsf(ahrs.roll_sensor-attitude_control->get_roll_trim_cd()) < 500 && labs(ahrs.pitch_sensor) < 500) {
+        fabsf(ahrs.get_roll_rad() - attitude_control->get_roll_trim_rad()) < radians(5) && labs(ahrs.pitch_sensor) < 500) {
```
rmackay9: I wonder if we should perhaps also change the pitch component to this?

fabs(ahrs.get_pitch_rad()) < radians(5)

optionally it might also be good to add some extra brackets before and after the "&&"

ArduCopter/RC_Channel_Copter.cpp
```diff
@@ -793,13 +793,13 @@ void RC_Channels_Copter::auto_trim_run()
             return;
         }
         // calculate roll trim adjustment, divisor set subjectively to give same "feel" as previous RC input method
-        float roll_trim_adjustment = radians(copter.attitude_control->get_att_target_euler_cd().x / 2000.0f);
+        float roll_trim_adjustment_rad = copter.attitude_control->get_att_target_euler_rad().x / 20.0f;
```
rmackay9, line 796: I think maybe the "20.0f" should be "radians(20.0f)"?

rmackay9: sorry, no, it's correct as is

ArduCopter/crash_check.cpp
```diff
@@ -124,8 +124,8 @@ void Copter::thrust_loss_check()
 
     // check for desired angle over 15 degrees
     // todo: add thrust angle to AC_AttitudeControl
-    const Vector3f angle_target = attitude_control->get_att_target_euler_cd();
-    if (sq(angle_target.x) + sq(angle_target.y) > sq(THRUST_LOSS_CHECK_ANGLE_DEVIATION_CD)) {
+    const Vector3f angle_target_rad = attitude_control->get_att_target_euler_rad();
```
rmackay9: it's an existing issue but we can make this a const reference I think.

const Vector3f& angle_target_rad = attitude_control->get_att_target_euler_rad();

ArduCopter/crash_check.cpp
```diff
@@ -124,8 +124,8 @@ void Copter::thrust_loss_check()
...
+    if (sq(angle_target_rad.x) + sq(angle_target_rad.y) > sq(cd_to_rad(THRUST_LOSS_CHECK_ANGLE_DEVIATION_CD))) {
```
rmackay9: It's fine as-is but we could optionally change this to use Vector2f::length_squared()

if (angle_target_rad.xy().length_squared() > sq(cd_to_rad(THRUST_LOSS_CHECK_ANGLE_DEVIATION_CD))) {

ArduCopter/land_detector.cpp
```diff
@@ -111,8 +111,8 @@ void Copter::update_land_detector()
 #endif
 
         // check for aggressive flight requests - requested roll or pitch angle below 15 degrees
-        const Vector3f angle_target = attitude_control->get_att_target_euler_cd();
-        bool large_angle_request = angle_target.xy().length() > LAND_CHECK_LARGE_ANGLE_CD;
+        const Vector3f angle_target_rad = attitude_control->get_att_target_euler_rad();
```
rmackay9: again an existing issue but we should use a const reference here

const Vector3f& angle_target_rad = attitude_control->get_att_target_euler_rad();

ArduCopter/land_detector.cpp
```diff
@@ -298,8 +298,8 @@ void Copter::update_throttle_mix()
         // autopilot controlled throttle
 
         // check for aggressive flight requests - requested roll or pitch angle below 15 degrees
-        const Vector3f angle_target = attitude_control->get_att_target_euler_cd();
-        bool large_angle_request = angle_target.xy().length() > LAND_CHECK_LARGE_ANGLE_CD;
+        const Vector3f angle_target_rad = attitude_control->get_att_target_euler_rad();
```
rmackay9: we could use a const reference here too

const Vector3f& angle_target_rad = attitude_control->get_att_target_euler_rad();

libraries/AC_WPNav/AC_WPNav.h
```diff
@@ -145,6 +145,9 @@ class AC_WPNav
     /// get_bearing_to_destination - get bearing to next waypoint in centi-degrees
     virtual int32_t get_wp_bearing_to_destination_cd() const;
 
+    /// get_bearing_to_destination - get bearing to next waypoint in radians
+    virtual float get_bearing_to_target_rad() const;
```
rmackay9: do we really want to use the word "target" here instead of "destination"?  I'm not sure we should introduce the word "target" should we?

libraries/AC_AttitudeControl/AC_AttitudeControl_Heli.h
```diff
@@ -72,7 +72,7 @@ class AC_AttitudeControl_Heli : public AC_AttitudeControl {
 
     // get_roll_trim - angle in centi-degrees to be added to roll angle for learn hover collective. Used by helicopter to counter tail rotor thrust in hover
     float get_roll_trim_cd() override;
-
+
```
rmackay9: little white space issue here I think

rmackay9:
This looks quite good to me.  We should definitely fix the const references mentioned above, the rest are optional.

libraries/AC_WPNav/AC_WPNav.h
lthall: I agree.

libraries/AC_AttitudeControl/AC_PosControl.cpp
```diff
@@ -1318,9 +1318,9 @@ void AC_PosControl::get_stopping_point_U_cm(postype_t &stopping_point_u_cm) cons
 }
 
 /// get_bearing_to_target_cd - get bearing to target position in centi-degrees
-int32_t AC_PosControl::get_bearing_to_target_cd() const
+float AC_PosControl::get_bearing_to_target_rad() const
 {
-    return get_bearing_cd(Vector2f{0.0, 0.0}, (_pos_target_neu_cm.xy() - _pos_estimate_neu_cm.xy()).tofloat());
+    return get_bearing_rad(Vector2f{0.0, 0.0}, (_pos_target_neu_cm.xy() - _pos_estimate_neu_cm.xy()).tofloat());
```
IamPete1, line 1323: I think this could just be 
```suggestion
    return (_pos_target_neu_cm.xy() - _pos_estimate_neu_cm.xy()).angle();
```

libraries/AC_WPNav/AC_WPNav.cpp
```diff
@@ -583,6 +583,12 @@ int32_t AC_WPNav::get_wp_bearing_to_destination_cd() const
     return get_bearing_cd(_pos_control.get_pos_estimate_NEU_cm().xy().tofloat(), _destination_neu_cm.xy());
 }
 
+/// get_wp_bearing_to_destination_cd - get bearing to next waypoint in centi-degrees
+float AC_WPNav::get_wp_bearing_to_destination_rad() const
+{
+    return get_bearing_rad(_pos_control.get_pos_estimate_NEU_cm().xy().tofloat(), _destination_neu_cm.xy());
```
IamPete1, line 589:
```suggestion
    return (_pos_control.get_pos_estimate_NEU_cm().xy(), _destination_neu_cm.xy()).angle();
```

libraries/AP_Math/location.cpp
```diff
@@ -23,14 +23,16 @@
 #include "AP_Math.h"
 #include "location.h"
 
-// return bearing in centi-degrees between two positions
+// return bearing_rad in radians between two positions
+float get_bearing_rad(const Vector2f &origin, const Vector2f &destination)
+{
+    return wrap_2PI(atan2f(destination.y-origin.y, destination.x-origin.x));
```
IamPete1, line 29: I don't think we need this helper at all, better to just use the `angle` method on the Vector2.

rmackay9: the difference between using this function and angle is just that angle will return -PI ~ +PI while this function returns 0 ~ 2xPI().  We probably handle that correctly but if we're using this function as a straight substitution for the existing get_bearing_cd() we should probably note change the output range (e.g. let's keep the function)

rmackay9 (approved):
LGTM.

BTW, there's an existing issue in the get_roll_trim_rad (and get_roll_trim_cd before it) could have been made const.

rmackay9:
@IamPete1 thanks very much for the review.  I general agree that we should make more use of the angle() function now that it's fixed but I think to avoid further delay we can stick with this new function which ensures there's no change in behaviour

I think I'll merge this now and @lthall can consider for future PRs whether we replace the new function is just angle().

rmackay9:
Just for accountability, there were a couple of bugs introduced in this PR which will be fixed by https://github.com/ArduPilot/ardupilot/pull/32016

PR #30427, opened 2025-06-22
Tools: Fix `test_ccache.py` to not confuse pytest discovery

cclauss:
`Tools/scripts/build_tests/test_ccache.py` has a filename that will make Pytest Discovery think that it should be tested.

This script also has code ___at global scope___ that will fail if certain command line arguments are not provided.

The proposed change is to put that code under `if __name__ == "__main__":` so it will not be run at test time.
* https://docs.python.org/3/library/__main__.html

Search for `test_ccache`:
* https://github.com/search?q=repo%3AArduPilot%2Fardupilot%20test_ccache&type=code

How was this tested?
1. `pytest Tools/scripts/build_tests/test_ccache.py`
2. `pytest Tools`

Both of the above find no tests instead of crashing with a long, confusing error.

### DIFF
--- a/Tools/scripts/build_tests/test_ccache.py
+++ b/Tools/scripts/build_tests/test_ccache.py
@@ -8,13 +8,6 @@ import argparse
 import sys
 import os
 
-parser = argparse.ArgumentParser(description='test ccache performance')
-parser.add_argument('--boards', default='MatekF405-bdshot,MatekF405-TE-bdshot', help='boards to test')
-parser.add_argument('--min-cache-pct', type=int, default=75, help='minimum acceptable ccache hit rate')
-parser.add_argument('--display', action='store_true', help='parse and show ccache stats')
-
-args = parser.parse_args()
-
 
 def ccache_stats():
     '''return hits/misses from ccache -s'''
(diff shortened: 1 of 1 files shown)


Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_virtual_ports.py
```diff
@@ -23,6 +23,7 @@
 
 WAIT_FOR_START_TIMEOUT = 5.0
 
+
```
tpwrules: Spurious change, please remove

Tools/scripts/build_tests/test_ccache.py
```diff
@@ -48,41 +41,49 @@ def build_board(boardname):
...
-        f.write(f"### ccache -s Output with {boards}\n")
-        f.write(f"```\n{result.stdout}\n```\n")
-        f.write(f"### ccache hit percentage (min {args.min_cache_pct})\n")
-        f.write("ccache hit percentage: %.1f%%  %s\n" % (hit_pct, post))
-if hit_pct < args.min_cache_pct:
-    print("ccache hits too low, need %d%%" % args.min_cache_pct)
-    sys.exit(1)
-else:
-    print("ccache hits good")
+if __name__ == "__main__":
```
tpwrules: Would suggest going further to the idiomatic style mentioned in the Python docs so we don't have a huge if statement.

Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_virtual_ports.py
cclauss: Running `git commit --no-verify` all the time is a bit scary.

% `git commit -am"Tools: Fix test_ccache.py to not confuse pytest discovery"`
```
Check line ending character (LF).........................................Passed
check for added large files..............................................Passed
check that executables have shebangs.................(no files to check)Skipped
check that scripts with shebangs are executable..........................Passed
check for merge conflicts................................................Passed
check xml............................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_virtual_ports.py

All done! ✨ 🍰 ✨
1 file reformatted.

flake8...................................................................Passed
```
% `git checkout master -- Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_virtual_ports.py`
% `git commit -am"Tools: Fix test_ccache.py to not confuse pytest discovery" --no-verify`

Tools/scripts/build_tests/test_ccache.py
```diff
@@ -48,41 +41,53 @@ def build_board(boardname):
...
+            f.write("ccache hit percentage: %.1f%%  %s\n" % (hit_pct, post))
+    if hit_pct < args.min_cache_pct:
+        print("ccache hits too low, need %d%%" % args.min_cache_pct)
+        sys.exit(1)
+    else:
+        print("ccache hits good")
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
tpwrules:
```suggestion
    main()
```

If you are going to do the exits in main instead of having it return a value.

Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_virtual_ports.py
tpwrules: Very few developers use the pre-commit hooks, but they should run in CI. I've asked the relevant people why this didn't get caught and they are investigating.

tpwrules (approved):
The ccache test passed and printed success.

PR #30428, opened 2025-06-22
ardupilotwaf: miscellaneous fixes

tpwrules:
Please see commits for details.

### DIFF
--- a/Tools/ardupilotwaf/mavgen.py
+++ b/Tools/ardupilotwaf/mavgen.py
@@ -9,7 +9,7 @@ The **mavgen.py** program is a code generator which creates mavlink header files
 
 from waflib import Logs, Task, Utils, Node
 from waflib.TaskGen import feature, before_method, extension
-import os, sys
+import sys
 import os.path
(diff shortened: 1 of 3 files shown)


khancyr:
hello, for the first commit, the test should be : 
install pymavlink with pip: try SITL build.
remove pymavlink from pip : try SITL build . It should still work by using the pymavlink into modules/mavlink/pymavlink. 

This allow some consistency on build and not to have to install pymavlink manual as we just need to generate the  mavlink headers (and previously, the C extension would require some deps not usefull for building)

tpwrules:
Yes, I did do that test and it worked properly.

tridge (approved):

tridge (dismissed):
need to work out why we are getting a binary change

tpwrules:
Rebuilding fixed it. In either case there was no checksum difference??

tpwrules:
I fixed the CI bugs showing spurious changes in #30554 and #30555 .

tridge (approved):

PR #30429, opened 2025-06-22
Tools: ros2: satisfy formatter

srmainwaring:
https://github.com/ArduPilot/ardupilot/pull/30396 includes a change with invalid `black` formatting. Was not reported in CI but is being picked up by precommit hook.

Further discussion here: 
- https://github.com/ArduPilot/ardupilot/pull/30427#discussion_r2160420068
- https://discord.com/channels/674039678562861068/850924523610963988/1386408054823719165

### DIFF
--- a/Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_virtual_ports.py
+++ b/Tools/ros2/ardupilot_dds_tests/test/ardupilot_dds_tests/test_virtual_ports.py
@@ -23,6 +23,7 @@ from launch_pytest.tools import process as process_tools
 
 WAIT_FOR_START_TIMEOUT = 5.0
 
+
 @launch_pytest.fixture
 def launch_description(virtual_ports):
     """Launch description fixture."""


srmainwaring:
It seems that we run the `ament_black` test against the CMake project `ardupilot_stil` which also includes a Python library. However we dropped the equivalent `test_flake8.py` from the pure Python package `ardupilot_dds_tests` because it conflicts with the formatting rules we are applying. The main different appears to be line-length:

```bash
E         ./test/ardupilot_dds_tests/test_time_msg_received.py:101:100: E501 line too long (102 > 99 characters)
```

Incidentally, including `test_flake8.py` in CI would have picked up this formatting error:

```bash
E         ./test/ardupilot_dds_tests/test_virtual_ports.py:26:1: E302 expected 2 blank lines, found 1
``` 

We need a way to have `test_flake8.py` use the repo precommit formatting rules.

cclauss:
[`psf/black`](https://pypi.org/project/black) is from the Python Software Foundation, which is trying to standardize Python formatting and ensure it is safe.

Each January, they publish a `stable style` for that year so that changes are robustly debated but rarely made.  `psf/black` and `ruff format` are both using stable style 2025, but I sense that [`ament_black`](https://pypi.org/project/ament-black) bundles a 2023 or 2024 version of `psf/black`.  Fortunately, the changes are few.

Another reason for an `ament-black` upgrade is
* botsandus/ament_black#15

The rules that [`ament-lint-pep257`](https://pypi.org/project/ament-lint-pep257) uses come from https://www.pydocstyle.org

PR #30430, opened 2025-06-22
Plane: Quadplane: coordinate turns in transition

robertlong13:
`hold_hover` and `hold_stabilize` set the commanded yaw rate to that of a coordinated turn during assisted flight (calling   `multicopter_attitude_rate_update(get_desired_yaw_rate_cds(false))`) That command gets thrown away here, ostensibly to suppress yaw output in transition, but it actually causes far more (see my analysis in #30415).

This is actually a well-tested code-path, since this is what vectored tilt-rotors always do.

Split out from #30415. Both Tridge and Pete responded skeptically about very idea yaw suppression. If yaw suppression is bad, then both the intended behavior and actual behavior of these lines is bad, and they should just be deleted.

Tridge suggested a possible idea of phasing out yaw command through the transition, and I'll change #30415 to implement that, but in the meantime, I'm going to split out this bit.

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -1561,16 +1561,6 @@ void SLT_Transition::update()
         }
         quadplane.hold_hover(climb_rate_cms);
 
-        if (!quadplane.tiltrotor.is_vectored()) {
-            // set desired yaw to current yaw in both desired angle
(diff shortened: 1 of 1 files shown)


robertlong13:
Closing so I can delete this branch. I did not know my new org role gave me access to push new branches to ArduPilot directly.

PR #30434, opened 2025-06-23
Plane: Quadplane: coordinate turns in transition

robertlong13:
`hold_hover` and `hold_stabilize` set the commanded yaw rate to that of a coordinated turn during assisted flight (calling   `multicopter_attitude_rate_update(get_desired_yaw_rate_cds(false))`) That command gets thrown away here, ostensibly to suppress yaw output in transition, but it actually causes far more (see my analysis in #30415).

This is actually a well-tested code-path, since this is what vectored tilt-rotors always do.

Split out from #30415. Both Tridge and Pete responded skeptically about very idea yaw suppression. If yaw suppression is bad, then both the intended behavior and actual behavior of these lines is bad, and they should just be deleted.

Tridge suggested a possible idea of phasing out yaw command through the transition, and I'll change #30415 to implement that, but in the meantime, I'm going to split out this bit.

### DIFF
--- a/ArduPlane/quadplane.cpp
+++ b/ArduPlane/quadplane.cpp
@@ -1561,16 +1561,6 @@ void SLT_Transition::update()
         }
         quadplane.hold_hover(climb_rate_cms);
 
-        if (!quadplane.tiltrotor.is_vectored()) {
-            // set desired yaw to current yaw in both desired angle
-            // and rate request. This reduces wing twist in transition
-            // due to multicopter yaw demands. This is disabled when
-            // using vectored yaw for tilt-rotors as the yaw control
-            // is needed to maintain good control in forward
-            // transitions
-            quadplane.attitude_control->reset_yaw_target_and_rate();
-            quadplane.attitude_control->rate_bf_yaw_target(0.0);
-        }
         if (quadplane.tiltrotor.enabled() && !quadplane.tiltrotor.has_fw_motor()) {
             // tilt rotors without dedicated fw motors do not have forward throttle output in this stage
             // prevent throttle I wind up
@@ -1637,16 +1627,6 @@ void SLT_Transition::update()
         quadplane.assisted_flight = true;
         quadplane.hold_stabilize(throttle_scaled);
 
-        // set desired yaw to current yaw in both desired angle and
-        // rate request while waiting for transition to
-        // complete. Navigation should be controlled by fixed wing
-        // control surfaces at this stage.
-        // We disable this for vectored yaw tilt rotors as they do need active
-        // yaw control throughout the transition
-        if (!quadplane.tiltrotor.is_vectored()) {
-            quadplane.attitude_control->reset_yaw_target_and_rate();
-            quadplane.attitude_control->rate_bf_yaw_target(0.0);
-        }
         break;
     }
 


IamPete1:
One side affect here is that this also results in allowing pilot yaw.

robertlong13:
@IamPete1, what about adding an early return here? Basically, this would mean that assisted flight should never use the pilot multirotor yaw stick input nor weathervaning. I believe this would also allow us to get rid of the "should weathervane" input, which seems like a nice cleanup. https://github.com/robertlong13/ardupilot/blob/7bb2b61203907fafb545a0117d68ca60f6e32a2b/ArduPlane/quadplane.cpp#L1335

I believe the other option proposed on the call would be to explicitly set the rate_bf_yaw_target to desired_auto_yaw_rate_cds after the hold_hover and hold_stabilize. I think that achieves the same result, but I like my above-solution a little more.

robertlong13:
In both cases, this means that in FBWA with zero throttle, you can't yaw the bird around where you want to go and then punch the throttle. You have to do that in QHover or QLoiter first. I think that's okay.

robertlong13:
I've decided to go the route of adding the if-block back and explicitly set the yaw rate to that of the coordinated turn instead of relying on the earlier calculations in `hold_hover` and `hold_stabilize`. This eliminates the risk of me changing something subtle about vectored-tiltrotors, since I haven't taken enough time to fully understand how they are handling yaw. This also bypasses the issue Pete raised (correctly) about causing pilot yaw to leak through. This also means I haven't fixed any pilot yaw leaks for vectored tilts, but that's an existing issue (maybe; I haven't actually tested; that may get handled somehow somewhere else).

I have also added a flag to `desired_auto_yaw_rate_cds` to decide whether we want the rate-of-change of the euler angle or the body-z rotation rate. It previously returned the euler rate exclusively, which is the correct rate for about half of the things this ends up in; I have not fixed the other half of the cases (existing issue and probably no big deal), but the change in this PR does need it to be body-z rate.

This should be a pretty safe minimal change now. Tested on the Cobra in RealFlight with worst-case 180deg transition and force-qassist figure-8s and it looks great. (I did not get a weather opportunity to test on Ottano yet unfortunately)

newplot (24)

ArduPlane/quadplane.cpp
```diff
@@ -1432,8 +1432,12 @@ float QuadPlane::desired_auto_yaw_rate_cds(void) const
     if (aspeed < 1) {
         aspeed = 1;
     }
-    float yaw_rate = degrees(GRAVITY_MSS * tanf(cd_to_rad(plane.nav_roll_cd))/aspeed) * 100;
-    return yaw_rate;
+    if (body_frame) {
+        return degrees(GRAVITY_MSS * sinf(cd_to_rad(plane.nav_roll_cd))/aspeed) * 100;
+    }
+    else {
+        return degrees(GRAVITY_MSS * tanf(cd_to_rad(plane.nav_roll_cd))/aspeed) * 100;
```
peterbarker:
```suggestion
    }
    return degrees(GRAVITY_MSS * tanf(cd_to_rad(plane.nav_roll_cd))/aspeed) * 100;
```

PR #30439, opened 2025-06-23
Plane: rework loiter radius correction and apply navigation roll limit to all nav. modes

rubenp02:
## Summary

Reworks Plane L1 loiter radius correction to use the fixed-wing turn radius formula and applies the navigation roll limit to all L1 navigation modes.

## Testing (more checks increases chance of being merged)

- [X] Checked by a human programmer
- [X] Tested in SITL
- [X] Tested on hardware
- [ ] Logs attached
- [X] Logs available on request
- [ ] Autotest included
- [X] Autotests updated

## Description

The implementation in `master` relies on loiter-specific radius compensation based on a sea-level scaling model. That approach was specific to loiters and depended on dedicated parameter setup to get the best behavior.

This PR updates `AP_L1_Control` so `NAVL1_ROLL_MAX` acts as an explicit navigation roll limit across waypoint, loiter, and heading-hold modes by constraining lateral acceleration demand before roll command output. When `NAVL1_ROLL_MAX` is zero, the controller falls back to the generic `ROLL_LIMIT_DEG`.

It also reworks loiter radius correction to calculate the minimum achievable turn radius from target airspeed and the active navigation roll limit, using the canonical fixed-wing turn radius formula instead of the old sea-level scaling model. In addition to this, the loiter radius correction method is renamed from loiter_radius to corrected_loiter_radius for readability. 

The previous explicit loiter radius correction inside `AP_L1_Control` is removed because altitude effects on turning performance are already captured by TECS, so a loiter-specific correction is unnecessary.

The correction model update and the removal of explicit loiter radius compensation are non-functional changes:

latitude_comparison

---

<details>
<summary>Screenshot comparison (all default params.)</summary>

### `master`

imagen

### This PR

imagen

### DIFF
--- a/Tools/autotest/arduplane.py
+++ b/Tools/autotest/arduplane.py
@@ -1154,7 +1154,8 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
             p7=x.alt_rel,    # alt
             frame=mavutil.mavlink.MAV_FRAME_GLOBAL_RELATIVE_ALT,
         )
-        expected_radius = 100
+        REALLY_BAD_FUDGE_FACTOR = 1.15  # FIXME
+        expected_radius = REALLY_BAD_FUDGE_FACTOR * self.get_parameter('WP_LOITER_RAD')
         self.wait_circling_point_with_radius(loc, expected_radius)
 
         self.context_collect('CAMERA_FEEDBACK')
@@ -1667,13 +1668,16 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
         self.do_fence_enable()
         self.assert_fence_enabled()
 
+        REALLY_BAD_FUDGE_FACTOR = 1.085 # FIXME
+        expected_radius = REALLY_BAD_FUDGE_FACTOR * self.get_parameter('WP_LOITER_RAD')
+
         self.takeoff(alt=50, alt_max=300)
         # Trigger fence breach, fly to rally location
         self.set_parameters({
             "FENCE_RET_RALLY": 1,
             "FENCE_ALT_MIN": 60,
         })
-        self.wait_circling_point_with_radius(rally_loc, return_radius)
+        self.wait_circling_point_with_radius(rally_loc, expected_radius)
         self.set_parameter("FENCE_ALT_MIN", 0) # Clear fence breach
 
         # 10 second fence min retrigger time
@@ -1688,7 +1692,7 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
         self.wait_altitude(altitude_min=return_alt-3,
                            altitude_max=return_alt+3,
                            relative=True)
-        self.wait_circling_point_with_radius(fence_loc, return_radius)
+        self.wait_circling_point_with_radius(fence_loc, expected_radius)
         self.do_fence_disable() # Disable fence so we can land
         self.fly_home_land_and_disarm() # Pack it up, we're going home.
 
@@ -2295,7 +2299,7 @@ class AutoTestPlane(vehicle_test_suite.TestSuite):
         self.wait_heading(290)
         self.wait_heading(300)
         dest = self.position_target_loc()
-        REALLY_BAD_FUDGE_FACTOR = 1.25  # FIXME
+        REALLY_BAD_FUDGE_FACTOR = 1.15  # FIXME
         expected_radius = REALLY_BAD_FUDGE_FACTOR * self.get_parameter('WP_LOITER_RAD')
         self.wait_circling_point_with_radius(dest, expected_radius)
 


libraries/AP_L1_Control/AP_L1_Control.cpp
```diff
@@ -30,14 +30,6 @@ const AP_Param::GroupInfo AP_L1_Control::var_info[] = {
...
     AP_GROUPINFO("XTRACK_I",   2, AP_L1_Control, _L1_xtrack_i_gain, 0.02),
 
-    // @Param: LIM_BANK
-    // @DisplayName: Loiter Radius Bank Angle Limit
-    // @Description: The sealevel bank angle limit for a continuous loiter. (Used to calculate airframe loading limits at higher altitudes). Setting to 0, will instead just scale the loiter radius directly
-    // @Units: deg
-    // @Range: 0 89
-    // @User: Advanced
-    AP_GROUPINFO("LIM_BANK",   3, AP_L1_Control, _loiter_bank_limit, 0.0f),
-
```
IamPete1, line 40: Leave a comment that index 3 was used by LIM_BANK. Otherwise someone might add a new param and find they get the old value from this param from eeprom.

ArduPlane/mode_autoland.cpp
```diff
@@ -145,7 +145,8 @@ bool ModeAutoLand::_enter()
 
     // corrected_loiter_radius is the radius the vehicle will actually fly, this gets larger as altitude increases.
     // Strictly this gets the loiter radius at the current altitude, really we want the loiter radius at final_wp_alt.
-    const float corrected_loiter_radius = plane.nav_controller->loiter_radius(loiter_radius);
+    const float corrected_loiter_radius =
+        plane.nav_controller->calc_corrected_loiter_radius(loiter_radius);
```
rubenp02:
```suggestion
        plane.nav_controller->corrected_loiter_radius(loiter_radius);
```

Maybe shorten to just `corrected_loiter_radius`.

IamPete1 (commented):
I think this is a good change. The key thing is that loiters will now happen at the expected radius until the bank angle limit is reached. This is a change in behavior.

tridge:
I'd like to see testing at high altitude, ensuring that we do scale radius with height when needed

IamPete1:
There was also the thought of keeping `LIM_BANK` to allow users to set a reduced "steady state" roll limit, this would means the radius would increase sooner which would allow some head room to the roll limit thus allowing a neater circle in wind. The key thing is that 0 would mean use the main roll limit so everyone gets the new behavior.

rubenp02:
> There was also the thought of keeping LIM_BANK to allow users to set a reduced "steady state" roll limit, this would means the radius would increase sooner which would allow some head room to the roll limit thus allowing a neater circle in wind.

I think there should be a nicer way to account for the wind and such, that one seems a bit weird to me since it would only apply to loiters (and not to waypoint navigation for instance, which may be unexpected for a parameter called NAVL1_LIM_BANK).

rubenp02:
> I'd like to see testing at high altitude, ensuring that we do scale radius with height when needed

@tridge [here](https://drive.google.com/file/d/1CJ_QhxSP3bKU-jgS_eZPjFmA5-LFngay/view?usp=sharing)'s a SITL log using these changes that shows the implicit loiter radius scaling. It uses the default SITL parameters except for ARSPD_USE, which is set to DoNotUse because the simulated airspeed sensor fails over a given altitude which affects the TECS and therefore the results.

The log shows the aircraft performing a LOITER_TO_ALT to 10,000m relative alt. The intended loiter for the radius is 100 m, and it holds up until around 4,000m AMSL:

imagen

IamPete1:
Looks like this might need some CI fixes for the new behavior.

rubenp02:
> Looks like this might need some CI fixes for the new behavior.

Should those go here or in another PR?

IamPete1:
> Should those go here or in another PR?

Here so we can get it past CI. If you click on the tests that are failing and wait for it to load and then scroll right to the bottom you can see its failing on: `FenceRetRally`, `ADSBResumeActionResumeLoiter` I guess the test has somehow hard-coded the old loiter radius.

rubenp02:
> > Should those go here or in another PR?
> 
> Here so we can get it past CI. If you click on the tests that are failing and wait for it to load and then scroll right to the bottom you can see its failing on: `FenceRetRally`, `ADSBResumeActionResumeLoiter` I guess the test has somehow hard-coded the old loiter radius.

Okay, I can see the issues. I'll probably fix them tomorrow.

rubenp02:
I have updated the REALLY_BAD_FUDGE_FACTORs to see if we can get CI to pass. I also needed to fix a small bug in a test (I thought that was the simplest solution, but if you prefer to split it into another PR I can do that).

When this is merged, #30438 should finally do away with all REALLY_BAD_FUDGE_FACTORs.

IamPete1 (approved):

libraries/AP_L1_Control/AP_L1_Control.cpp
```diff
@@ -30,13 +30,7 @@ const AP_Param::GroupInfo AP_L1_Control::var_info[] = {
...
     AP_GROUPINFO("XTRACK_I",   2, AP_L1_Control, _L1_xtrack_i_gain, 0.02),
 
-    // @Param: LIM_BANK
-    // @DisplayName: Loiter Radius Bank Angle Limit
-    // @Description: The sealevel bank angle limit for a continuous loiter. (Used to calculate airframe loading limits at higher altitudes). Setting to 0, will instead just scale the loiter radius directly
-    // @Units: deg
-    // @Range: 0 89
-    // @User: Advanced
-    AP_GROUPINFO("LIM_BANK",   3, AP_L1_Control, _loiter_bank_limit, 0.0f),
+    // Note: Index 3 was taken up by the removed parameter LIM_BANK and should't be reused.
```
tridge: some people have set this parameter, and not necessarily to the value of their roll limit, they may well want a lower limit in auto modes than in manual modes like FBWA.
I think we need to keep the parameter

rubenp02:
> I think we re-work so this calculates a roll limit, then uses that to calculate a loiter radius (when requested), and also to calculate a lateral accel limit, so it would apply in all autonomous flight, not just loitering

That's a substantial change. Maybe it's best if this gets merged, with the LIM_BANK parameter added back in but defaulting to ROLL_LIMIT_DEG when set to 0, and making it work outside loiter navigation is left to another PR?

Also, maybe I'm wrong but there seems to be a bit of confusion around why loiter radius increases with altitude. It's not because the max. sustainable bank angle gets any smaller, it's simply because we fly at a higher TAS to maintain the IAS at which the bank angle is sustainable. This is an entirely separate concern (it is handled in the TECS and in the aerodynamic load factor-based roll limiting in the attitude controller). In fact in the SITL log I attached it can be seen how the aircraft initially doesn't bank much to achieve the turn radius, and it needs to bank more and more to maintain it as it climbs until it reaches the ROLL_LIMIT_DEG, at which point it can only increase the turn radius.

So we don't need to calculate a roll limit here (we can use the generic one or the navigation-specific one), and the description of the NAVL1_LIM_BANK parameter ("The sealevel bank angle limit for a continuous loiter. ...") is inaccurate. The only reason why the `loiter_radius` method was replaced instead of removed is because some other parts of the code need to know if a substantially larger radius is going to be flown. So the only advantage of extending this to waypoint navigation is to provide a separate bank angle parameter for all L1 navigation.

libraries/AP_L1_Control/AP_L1_Control.cpp
rubenp02: The main thing is that the roll limiting and the loiter radius compensation are fully separate concerns. I can put the parameter back but there's probably a more elegant way to apply an auto mode-specific roll limit externally.

tridge:
@rubenp02 opened thread for this on discord here: https://discord.com/channels/674039678562861068/1392633137875521567

tridge:
how about this:
 - NAVL1_G_LIMIT parameter. if > 0 then use it. If == 0 then calculate based on roll limit, sea level and AIRSPEED_MAX, if < 0 then disable this limit feature
 - default value of 0 for principle of least surprise for existing users
 - apply this limit in all parts of L1 controller, so both loiter radius and for L1 WP navigation lateral accel limit

IamPete1:
I had forgotten we had not got this in. Now 4.7 has branched we could have another go at getting this in.

rubenp02:
> I had forgotten we had not got this in. Now 4.7 has branched we could have another go at getting this in.

It's a bit dusty! I updated it a few weeks ago (including getting a proper limiter in) but had to get on with something else before I got fully finished, so I didn't post it. I'll get to it ASAP.

rubenp02:
Updated!

@tridge:
> how about this:
> 
>     * NAVL1_G_LIMIT parameter. if > 0 then use it. If == 0 then calculate based on roll limit, sea level and AIRSPEED_MAX, if < 0 then disable this limit feature
> 
>     * default value of 0 for principle of least surprise for existing users
> 
>     * apply this limit in all parts of L1 controller, so both loiter radius and for L1 WP navigation lateral accel limit

I didn't end up doing this: I have implemented it as a navigation exclusive roll limit instead. I believe this is more intuitive for users, means we can reuse the old LIM_BANK parameter, and I also don't think track-lateral acceleration is typically measured in Gs? Which might have been confusing. Perhaps I'm wrong, please correct me if so!

I think the result is pretty good and matches the spirit of the suggestion though. But if you prefer I do this instead, just say the word. It's a pretty small change.

rubenp02:
Sorted out CI.

PR #30440, opened 2025-06-23
Plane: Generalize loiter radius compensation

rubenp02:
# Plane: Generalize loiter radius compensation

Description
-----------
This PR adds arbitrary speed and altitude turn radius calculations to the loiter radius compensation rework at #30439. It has been split from #29165 for ease of review.

## Changes

1. Made the baro module a dependency of the L1 navigation controller, necessary for getting the current altitude and calculating arbitrary altitude IAS to TAS conversions.
2. Updated the calc_corrected_loiter_radius public method to optionally accept altitude AMSL and indicated airspeed for a more flexible compensation.
3. Used this new functionality to fix an inaccuracy in the autoland implementation.


rubenp02:
This can be extended to check if the mission contains any unachievable loiters during pre-arm. A is_loiter_achievable method is implemented in the original PR this is split from, but since it isn't used here, I decided not to include it.

PR #30441, opened 2025-06-23
Sub: place Home at surface when setting it inflight

Williangalvani:
Fixes an issue where the home would be set to the ekf origin, which uses the GPS altitude, which is far, far more imprecise than our baros, and doesn't take tides into account.

Fixes the issue of "my rov got a gps lock at the surface and now thinks it is 8m underwater because the GPS is ~drunk~ innacurate on vertical position"

### DIFF
--- a/ArduSub/Sub.cpp
+++ b/ArduSub/Sub.cpp
@@ -431,13 +431,7 @@ float Sub::get_alt_rel() const
 
     // get relative position
     float posD;
-    if (ahrs.get_relative_position_D_origin_float(posD)) {
-        if (ahrs.home_is_set()) {
-            // adjust to the home position
-            auto home = ahrs.get_home();
-            posD -= static_cast<float>(home.alt) * 0.01f;
-        }
-    } else {
+    if (!ahrs.get_relative_position_D_origin_float(posD)) {
         // fall back to the barometer reading
         posD = -AP::baro().get_altitude();
     }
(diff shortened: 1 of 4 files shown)


rmackay9:
I think it might be better to instead always set the EKF origin to be at altitude zero.

This PR could lead to the user being unable to reset their relative altitude to zero.  Image this case:

1. the vehicle's altitude estimate is non-zero (e.g. reports 1m when actually at the surface)
2. user sets home to the current location but the home altitude will remain at zero and so the relative altitude will not become zero

Williangalvani:
> I think it might be better to instead always set the EKF origin to be at altitude zero.

but we can't set the ekf origin once it has already been set, can we?

 > the vehicle's altitude estimate is non-zero (e.g. reports 1m when actually at the surface)
user sets home to the current location but the home altitude will remain at zero and so the relative altitude will not become zero

In Sub, I think the best way to fix would actually be re-calibrating the baro.

This approach would also make us unable to handle missions with absolute altitude, right? as setting the home only affects the relative altitude reading, and the origin from gps can have a large vertical error.

I'm a bit worried about bodies of water at higher elevations though...

rmackay9:
Hi @Williangalvani,

I think an AHRS_OPTION or EK3_OPTION that forces the EKF origin to always be set to zero would work.  So the moment that something tries to set the EKF origin, that altitude is ignored and instead it's alway uses a zero altitude.

PR #30443, opened 2025-06-23
Tools: Fix `run_filter_test.py` to not confuse pytest discovery

cclauss:
`Tools/FilterTestTool/run_filter_test.py` has a filename that will make Pytest Discovery think that it should be tested.

This script also has code ___at global scope___ that will fail if certain command line arguments are not provided.

The proposed change is to put that code under `if __name__ == "__main__":` so it will not be run at test time.
* https://docs.python.org/3/library/__main__.html

Search for `run_filter_test`:
* https://github.com/search?q=repo%3AArduPilot%2Fardupilot%20run_filter_test&type=code

How was this tested?
1. `pytest Tools/FilterTestTool/run_filter_test.py`
2. `pytest Tools`

Both of the above find no tests instead of crashing with a confusing error.

### DIFF
--- a/Tools/FilterTestTool/run_filter_test.py
+++ b/Tools/FilterTestTool/run_filter_test.py
@@ -19,143 +19,13 @@ this program. If not, see <http://www.gnu.org/licenses/>.
 __author__ = "Guglielmo Cassinelli"
 __contact__ = "gdguglie@gmail.com"
 
-try: # Python 3.x
-    from tkinter import Tk
-    from tkinter.filedialog import askopenfilename
-except ImportError: # Python 2.x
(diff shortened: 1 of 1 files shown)


tpwrules:
This might be another candidate for aging out. We need to test a log through it to make sure it's not broken but there is for sure some old stuff in it... I agree with the making it python3-specific.

cclauss:
Closing because even with these changes, `.github/workflows/test_unit_tests.yml` does not `use: setup-python`, so the script fails because `tkinter` is not installed.

tpwrules:
Why don't we put pytest tests in an allowlisted folder?

PR #30446, opened 2025-06-24
AC_PrecLand; MAVLink backend supports LOCAL_FRD frame

rmackay9:
This enhances the AC_PrecLand MAVLink backend so that it can accept LANDING_TARGET messages with the LOCAL_FRD frame (previously it only supported the BODY_FRD frame).  This essentially turns off the roll/pitch compensation (but leaves yaw compensation in place) so that a vehicle's main camera gimbal can be used for precision landing.

The PrecisionLoiter autotest has been enhanced slightly to test sending messages using LOCAL_FRD

I hope to test [this BlueOS Extension](https://github.com/rmackay9/blueos-precision-landing) in the coming week to confirm this all works on a real vehicle


libraries/AC_PrecLand/AC_PrecLand.cpp
```diff
@@ -632,23 +638,42 @@ bool AC_PrecLand::retrieve_los_meas(Vector3f& target_vec_unit_body)
...
+        Vector3f target_vec_unit_ned;
+        switch (target_vec_frame) {
+        case VectorFrame::BODY_FRD:
+            // convert to NED
+            target_vec_unit_ned = _inertial_data_delayed->Tbn * target_vec_unit;
+            break;
+        case VectorFrame::LOCAL_FRD:
+            // rotate vector using delayed yaw
+            float roll_rad, pitch_rad, yaw_rad;
+            _inertial_data_delayed->Tbn.to_euler(&roll_rad, &pitch_rad, &yaw_rad);
```
rishabsingh3003, line 671: No big deal and not a requirement, but it would have been nice to have a "get_yaw" with Matrix3f to avoid having to unessarly do the trigonometry calculations for pitch and roll.

libraries/AC_PrecLand/AC_PrecLand_MAVLink.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+// retrieve updates from sensor
+void AC_PrecLand_MAVLink::update()
+{
+    _los_meas.valid = _los_meas.valid && AP_HAL::millis() - _los_meas.time_ms <= 1000;
+}
+
+void AC_PrecLand_MAVLink::handle_msg(const mavlink_landing_target_t &packet, uint32_t timestamp_ms)
+{
+    _distance_to_target = packet.distance;
```
rishabsingh3003: We should probably have a `_los_meas.valid = false` here, to prevent stale usage of `_los_meas.vec_unit` incase this function returns early? We check the timestamp of the last consumed packet to make sure we don't consume duplicates, so this is probably not a problem yet, but my worry is that someone may use this in the future by mistake elsewhere.

rishabsingh3003: I think the other backends may need this as well

rishabsingh3003 (approved):
@rmackay9 thanks!  Very requested feature. 
Two small optional comments, not a blocker.

libraries/AC_PrecLand/AC_PrecLand_MAVLink.cpp
rmackay9: Hi @rishabsingh3003,

Thanks very much for the review, very few others are familiar with this library.

Re adding specifically setting "_los_meas.valid = false", I've just done a light review and I don't think this is necessary because the target's position will only be updated when "_los_meas.time_ms" is changed by a backend and each backend's update() function includes a timeout.  Put another way, I don't think we need to invalidate the last known reading before the timeout does.

rmackay9: @rishabsingh3003,

Ah, I see your point re _distance_to_target though.  You're absolutely right that the driver would consume that field even if the LANDING_TARGET was using an invalid frame.  I've added a fix so that the field is only consumed when other fields are also consumed, thanks!

libraries/AC_PrecLand/AC_PrecLand.cpp
rmackay9: yes, the atan2 functions are quite expensive and we'd avoid doing one of these if we added a get_yaw().  In the interests of getting this in quickly, I think I might pass on adding this extra method for now though if that's OK.

PR #30449, opened 2025-06-24
pre-commit: Add a job that runs ament_pep257 Tools/ros2

cclauss:
[`Tools/ros2/ardupilot_dds_tests/test/test_pep257.py`](https://github.com/ArduPilot/ardupilot/blob/master/Tools/ros2/ardupilot_dds_tests/test/test_pep257.py) ensures [PEP 257 – Docstring Conventions](https://peps.python.org/pep-0257) compliance for all Python files in Tools/ros2.

Here are two solutions (we should pick just one) that replicate this functionality in pre-commit.
1. Use `flake8` plus `flake8-docstring` to run the 'ament-style` rules on Tools/ros2.
2. Run [`ament_pep257`](https://pypi.org/project/ament-lint-pep257) on Tools/ros2.

Quick equivalent commands:
```
uv tool run --with=flake8-docstrings flake8 --select=D \
    --ignore=D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404 Tools/ros2

uv tool run --from=ament-lint-pep257 ament_pep257 Tools/ros2  > /dev/null
```
How was this tested?
1. Run both equivalent commands to ensure no errors.
2. `pre-commit run flake8 --all-files` to ensure no errors.
3. `pre-commit run ament-pep257 --all-files` to ensure no errors.

Edit `Tools/ros2/ardupilot_sitl/src/ardupilot_sitl/utilities.py` and add a leading or trailing space inside the docstring.

Rerun the four commands to ensure that each raises a `D210: No whitespaces allowed surrounding docstring text` error.


cclauss:
It would be also interesting to add a copyright pre-commit hook:
* #30598

.pre-commit-config.yaml
```diff
@@ -55,6 +55,29 @@ repos:
       rev: 7.3.0
       hooks:
       -   id: flake8
+  
+  -   repo: https://github.com/pycqa/flake8
+      rev: 7.3.0
+      hooks:
+      -   id: flake8
+          name: flake8-docstrings with ament-style rules
```
Ryanf55: If you ran ament_flake8 as a pre-commit rule, would that replace the need to manually specify all the ignores on line 69?

Ryanf55 (approved):
Overall, a nice improvement, but perhaps theres a way to avoid duplicating the specific ament rules here.

Works on my system as intended.
image

cclauss:
Thanks for the review
> Here are two solutions (we should pick just one)

Do you prefer the hook that uses `flake8-docstrings` or the one that uses `ament_pep257`?

Ryanf55:
Works for

> Thanks for the review
> 
> > Here are two solutions (we should pick just one)
> 
> Do you prefer the hook that uses `flake8-docstrings` or the one that uses `ament_pep257`?

`ament_pep257` please

PR #30456, opened 2025-06-24
Rover: add AP_LeakDetector

### LINKED ISSUE #19103: Leak failsafe from ardusub for Arduboat
can the leak failsafe from ardusub be added to the regular rover firmware so it can be used on boats?

Williangalvani:
fix https://github.com/ArduPilot/ardupilot/issues/19103

### Summary

Adds a leak failsafe to Rover for boats, reusing the existing AP_LeakDetector library
from Sub, and makes that library compile-time optional. Fixes #19103

### Classification & Testing (check all that apply and add your own)

- [ ] Non-functional change
- [ ] No-binary change
- [ ] Infrastructure change (e.g. unit tests, helper scripts)
- [x] Automated test(s) verify changes (e.g. unit test, autotest)
- [x] Tested manually, description below (e.g. SITL)
- [ ] Tested on hardware
- [ ] Logs attached
- [ ] Logs available on request

New `Rover.LeakFailsafe` autotest drives a simulated digital detector pin and checks
that the failsafe warns, switches to the configured mode and clears again once dry.
Rover and Sub were built for SITL with the feature enabled (default), with
`AP_LEAKDETECTOR_ENABLED=0`, and with each of `AP_LEAKDETECTOR_ANALOG_ENABLED`,
`AP_LEAKDETECTOR_DIGITAL_ENABLED` and `AP_LEAKDETECTOR_BACKEND_DEFAULT_ENABLED`
set to 0. `Sub.DiveManual` still passes. Not tested on hardware yet.

### Description

Rover gains `FS_LEAK_ENABLE`, which selects the action taken when a leak is detected
using the same values as `FS_ACTION`; zero disables leak detection entirely. The
detector is polled at 5Hz, sets `AP_Notify::flags.leak_detected`, and reports through
the existing `failsafe_trigger()` path so `FS_TIMEOUT` and `FS_OPTIONS` apply as they
do for the other failsafes. Leaks reported by a remote sensor over MAVLink `SYS_STATUS`
are honoured as well, so a second hull can be monitored — same handler Sub already has.

AP_LeakDetector gains an `AP_LeakDetector_config.h` with `AP_LEAKDETECTOR_ENABLED` plus
per-backend defines, and a matching entry in `build_options.py` and `extract_features.py`.
It defaults to enabled, which only affects Sub and Rover since they are the only vehicles
that build the library. The ArduSub commit guards its uses of the detector so that build
also honours the define; there is no functional change to Sub.

This contribution was AI-assisted: I'm responsible for the original design but used AI for addressing some issues and reviews.


Rover/Parameters.cpp
```diff
@@ -344,6 +344,13 @@ const AP_Param::Info Rover::var_info[] = {
     GOBJECT(_gcs,           "MAV",  GCS),
 #endif
 
+    // @Param: FS_LEAK_ENABLE
+    // @DisplayName: Leak Failsafe Enable
+    // @Description: Controls what action to take if a leak is detected.
+    // @Values: 0:Warn only,1:RTL,2:Hold,3:SmartRTL,4:SmartRTL_Hold,5:Terminate,6:Loiter_Hold
```
rmackay9: Thanks for this.  For the Values it's OK to have spaces.  So we can make "SmartRTL_Hold" into "SmartRTL or Hold".  Similarly "Loiter or Hold"

rmackay9: I think maybe this parameters should also be enclosed with AP_LEAKDETECTOR_ENABLED checks?

Rover/Parameters.h
```diff
@@ -263,6 +264,7 @@ class Parameters {
     AP_Int8     fs_crash_check;
     AP_Int8     fs_ekf_action;
     AP_Float    fs_ekf_thresh;
+    AP_Int8     fs_leak_enabled;
```
rmackay9, line 268: we can enclose this with AP_LEAKDETECTOR_ENABLED as well I think

libraries/AP_LeakDetector/AP_LeakDetector.cpp
```diff
@@ -1,6 +1,8 @@
+
 #include "AP_LeakDetector.h"
 #include "AP_LeakDetector_Analog.h"
 #include "AP_LeakDetector_Digital.h"
+#if AP_LEAKDETECTOR_ENABLED
```
rmackay9: I'm not sure that these defines will work will they?  I think they need to be put at a lower level because libraries can't see the definitions from the vehicle level config.h file.  @peterbarker knows best

libraries/AP_LeakDetector/AP_LeakDetector.h
```diff
@@ -1,11 +1,16 @@
 #pragma once
 
+#ifndef AP_LEAKDETECTOR_ENABLED
+#define AP_LEAKDETECTOR_ENABLED 1
+#endif
```
peterbarker: Please create an `AP_LeakDetector_config.h` to hold this

libraries/AP_LeakDetector/AP_LeakDetector.h
```diff
@@ -1,11 +1,16 @@
...
+
 #include <AP_Param/AP_Param.h>
 
 #define LEAKDETECTOR_MAX_INSTANCES 3
 
 #define LEAKDETECTOR_COOLDOWN_MS 3000 // Status will return true for this long after last time leak was detected
 
+#if AP_LEAKDETECTOR_ENABLED
```
peterbarker: Move this to just below the include of `AP_LeakDetector_config.h`

ArduSub/config.h
```diff
@@ -232,3 +232,4 @@
 # define FLIGHT_MODE_6 Mode::Number::SURFACE
 #endif
 
+#define AP_LEAKDETECTOR_ENABLED 1
```
peterbarker:
```suggestion
```

libraries/AP_LeakDetector/AP_LeakDetector_Analog.cpp
```diff
@@ -1,5 +1,6 @@
 #include "AP_LeakDetector_Analog.h"
 #include <AP_HAL/AP_HAL.h>
+#if AP_LEAKDETECTOR_ENABLED
```
peterbarker:
```suggestion
#include "AP_LeakDetector_config.h"

#if AP_LEAKDETECTOR_ANALOG_ENABLED

#include "AP_LeakDetector_Analog.h"
#include <AP_HAL/AP_HAL.h>
```

similarly for the .cpp file and the digital files; `_config.h` specifies AP_LEAKDETECTOR_ANALOG_ENABLED true of `AP_LEAKDETECTOR_BACKEND_DEFAULT_ENABLED` is true - and it is true if `AP_LEAKDETECTOR_ENABLED`.  This gives this the correct modern "pattern" we use in AP

Rover/config.h
```diff
@@ -55,6 +55,9 @@
 # define MODE_FOLLOW_ENABLED AP_FOLLOW_ENABLED
 #endif
 
+#ifndef AP_LEAKDETECTOR_ENABLED
+#define AP_LEAKDETECTOR_ENABLED 1
+#endif
```
peterbarker:
```suggestion
```

Can't do this at the vehicle level.

If it is a problem on Plane and Copter then you have to "stub" things out in AP_LeakDetector.cpp if the vehicle doesn't want the feature.

Rover/Rover.h
```diff
@@ -57,6 +57,10 @@
 #include <AP_Scripting/AP_Scripting.h>
 #endif
 
+#ifdef AP_LEAKDETECTOR_ENABLED
+#include <AP_LeakDetector/AP_LeakDetector.h>
+#endif
+
```
peterbarker:
```suggestion
#include <AP_LeakDetector/AP_LeakDetector_config.h>
```

peterbarker (changes requested):

Hwurzburg (changes requested):
needs build wserver option and feature extraction

ES-Alexander:
We should `#if` guard [the remote detection too](https://github.com/ArduPilot/ardupilot/blame/2d30d80feaf6aa6756567b2dac9d9b0ad469c9b6/ArduSub/GCS_MAVLink_Sub.cpp#L739-L748), and potentially add that functionality to Rover (e.g. as another means of supporting leak detections from both hulls of a catamaran).

libraries/AP_LeakDetector/AP_LeakDetector_config.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <AP_HAL/AP_HAL_Boards.h>
+
+#ifndef AP_LEAKDETECTOR_ENABLED
+#define AP_LEAKDETECTOR_ENABLED 1
```
ES-Alexander, line 6: Does this enable it by default for firmware that don't specify the compile flag? Or only for ones that already try to include it as part of the build process, just with the option to manually disable it at compile time?

Tools/autotest/rover.py
```diff
@@ -204,6 +204,31 @@ def ThrottleFailsafe(self):
         self.disarm_vehicle()
         self.progress("Loiter or Hold as throttle failsafe OK")
 
+    def LeakFailsafe(self):
+        """Trigger the leak failsafe"""
+        pin = 8  # not used by the default SITL pin mask
```
ES-Alexander:
```suggestion
        leak_detector_pin = 8  # not used by the default SITL pin mask
```

ES-Alexander (commented):
Code looks generally reasonable to me, with just a minor comment on an ambiguous variable name.

The functionality seems to work (on my bench hardware setup), but if the radio failsafe is enabled then for some reason triggering the leak failsafe actually triggers the radio failsafe, despite me having valid joystick input to the GCS:

Image

If I disable the radio failsafe then triggering the leak failsafe works as expected (with the action set to `loiter or hold (6)` and no GPS connected, so loiter fails):

Image

---

It would be nice if the warnings were synchronous with how the trigger+action occur (e.g. 'leak failsafe', then 'mode change failed', then 'mode changed to HOLD' - currently the mode change is reported first), but not sure if that breaks some async code that's important for avoiding waiting around for statustext messages to send before being able to report more general status updates 🤷‍♂️

PR #18905, opened 2021-10-11
AC_Avoid: Dijkstras: don't put points in corners

IamPete1:
This makes a change not to put visgraph point in corners, this maintains full visibility of the polygon with significantly fewer points. In addition it adds logging of visgraph points, with a option bit. In combination with https://github.com/ArduPilot/ardupilot/pull/18892 you can make nice plots to see what is going on. 

A example graph with this PR, graph points in yellow:
image

The same on master:
image

So we go from 17 points to 8. Dijkstras is approx n^2, 289 >>> 64.

For convex hull inclusion polygons we will no longer add any points. Less of a improvement for exclusion zones as they will have mostly external corners.

PR #30457, opened 2025-06-24
AP_OADijkstra: Fix missing Wrap

lthall:
While checking for places #30454 output range change could mess things up I found that we were missing a wrap in Dijkstra.

This will mean that when either side of -90 degrees before #30454, and 180 degrees after #30454. we could miss detecting a corner.

### DIFF
--- a/libraries/AC_Avoidance/AP_OADijkstra.cpp
+++ b/libraries/AC_Avoidance/AP_OADijkstra.cpp
@@ -384,7 +384,7 @@ bool AP_OADijkstra::create_inclusion_polygon_with_margin(float margin_cm, AP_OAD
             }
 
             // don't add points in corners
-            if (fabsf(intermediate_pt.angle() - before_pt.angle()) < M_PI_2) {
+            if (fabsf(wrap_PI(intermediate_pt.angle() - before_pt.angle()) < M_PI_2)) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


libraries/AC_Avoidance/AP_OADijkstra.cpp
```diff
@@ -384,7 +384,7 @@ bool AP_OADijkstra::create_inclusion_polygon_with_margin(float margin_cm, AP_OAD
             }
 
             // don't add points in corners
-            if (fabsf(intermediate_pt.angle() - before_pt.angle()) < M_PI_2) {
+            if (fabsf(wrap_PI(intermediate_pt.angle() - before_pt.angle()) < M_PI_2)) {
```
IamPete1:
```suggestion
            if (fabsf(wrap_PI(intermediate_pt.angle() - before_pt.angle())) < M_PI_2) {
```

IamPete1 (commented):
I added this in https://github.com/ArduPilot/ardupilot/pull/18905. I'm surprised, I did quite a bit of testing at the time. This certainly looks like a correct fix tho.

rmackay9 (approved):
surely this is safe, thanks!

timtuxworth:
Could I ask you to approve @IamPete1 ? I'd like to test it in master, but not confident to approve it although it does seem to make sense.

rmackay9:
I've done some more testing with Dijkstras in SITL and it seems fine.  I actually couldn't reproduce the exact case where it goes wrong before
before-vs-after
image

Issue #29095, opened 2025-01-17
Lidar/Rangefinder used ONLY for VTOL takeoff and VTOL landing in Quadplane


Gromitvt10:
## Feature request

**Is your feature request related to a problem? Please describe.**
_When flying my quadplane with Lidar installed i enable rangefinder for automatic landing. The rangefinder will be used both on the landing approach and for final flare.  But, if the lidar, for any reason shows a low reading, it will trigger the alt_assist even when not on approach or landing. This in turn initiates the VTOL motors to start and creates a lot of drag. 
The problem is this aircraft can fly for 8 hours, what if it was flying miles away and the lidar randomly starting doing this, false low readings.  It will constantly be turning the VTOL motors on and off and would eventually drain its battery.  We can fix or replace this particular lidar unit, but that doesn't mean the new one won't eventually do the same thing. This issue surfaced only after many flights on the aircraft, who's to say it won't happen again with a new lidar unit after many flights.

**Describe the solution you'd like**
_We currently set the Alt_assist to zero, but that turns it off completely. I would like a way to use the lidar only for VTOL takeoff (transition height-use until transition is complete) and landing measurements (over landing spot all the way until touch down). I would like to use Baro for Alt-assist. I don't need the lidar when i'm flying at 1000ft, especially when it can only read up 300ft.  A setting or flag that lets you chose to "use lidar only for takeoff and landing" would be great.  If not checked then it uses lidar for everything, when checked it uses Baro for Alt_assist.

**Describe alternatives you've considered**
_I have considered not using lidar at all, but then why even have it. Discussed changing the unit or changing unit parameters, this may fix this particular instance but won't fix the overall issue of Ardupilot engaging Alt assist for false readings._

**Platform**
[  ] All
[  ] AntennaTracker
[  ] Copter
[ X ] Plane
[  ] Rover
[  ] Submarine

**Additional context**
_This issue has been brought up and discussed in my partner chat_

IamPete1:
We do have a Q_OPTION for airspeed assist:
"Disable speed based Qassist when using synthethic airspeed estimates". 

So it would make sense to add this in a similar way. However, I think we have run out of bits.....

timtuxworth:
> We do have a Q_OPTION for airspeed assist: "Disable speed based Qassist when using synthethic airspeed estimates".
> 
> So it would make sense to add this in a similar way. However, I think we have run out of bits.....

"You always need more!"

tridge:
[...]

PR #30458, opened 2025-06-25
Plane: make RNGFND_LANDING a bitmask

tridge:
this gives greater control of when the rangefinder will be used. A particularly important one is the ability to disable it for VTOL assist as false positives can be quite common and drains the battery
main question is if we should also rename to  RNGFND_USE or RNGFND_USE_MASK. I didn't for this PR, but could discuss

### DIFF
--- a/ArduPlane/GCS_Plane.cpp
+++ b/ArduPlane/GCS_Plane.cpp
@@ -116,7 +116,7 @@ void GCS_Plane::update_vehicle_sensor_status_flags(void)
     const RangeFinder *rangefinder = RangeFinder::get_singleton();
     if (rangefinder && rangefinder->has_orientation(plane.rangefinder_orientation())) {
         control_sensors_present |= MAV_SYS_STATUS_SENSOR_LASER_POSITION;
-        if (plane.g.rangefinder_landing) {
+        if (uint16_t(plane.g.rangefinder_landing.get()) != 0) {
             control_sensors_enabled |= MAV_SYS_STATUS_SENSOR_LASER_POSITION;
         }
         if (rangefinder->has_data_orient(plane.rangefinder_orientation())) {
--- a/ArduPlane/Parameters.cpp
+++ b/ArduPlane/Parameters.cpp
@@ -785,9 +785,9 @@ const AP_Param::Info Plane::var_info[] = {
     GOBJECT(rangefinder,            "RNGFND", RangeFinder),
 
     // @Param: RNGFND_LANDING
-    // @DisplayName: Enable rangefinder for landing
-    // @Description: This enables the use of a rangefinder for automatic landing. The rangefinder will be used both on the landing approach and for final flare
-    // @Values: 0:Disabled,1:Enabled
+    // @DisplayName: Enable use of rangefinder
+    // @Description: Sets the use of a rangefinder for automatic landing and other use cases. When enabled for landing and takeoff the rangefinder will be used both on the landing approach and for final flare as well as as VTOL landing and for takeoffs and throttle suppression when close to the ground. When enabled for assist the rangefinder will be used for VTOL assistance. When enabled for climb the rangefinder will be used for the initial climb in QRTL and AUTO. Set to 0 to disable use of the rangefinder.
+    // @Bitmask: 0:All, 1:TakeoffAndLanding, 2:Assist, 3:InitialClimb
     // @User: Standard
     GSCALAR(rangefinder_landing,    "RNGFND_LANDING",   0),
 #endif
--- a/ArduPlane/altitude.cpp
+++ b/ArduPlane/altitude.cpp
@@ -115,7 +115,7 @@ int32_t Plane::get_RTL_altitude_cm() const
   return relative altitude in meters (relative to terrain, if available,
   or home otherwise)
  */
-float Plane::relative_ground_altitude(bool use_rangefinder_if_available, bool use_terrain_if_available)
+float Plane::relative_ground_altitude(enum RangeFinderUse use_rangefinder, bool use_terrain_if_available)
 {
 #if AP_MAVLINK_MAV_CMD_SET_HAGL_ENABLED
    float height_AGL;
@@ -126,13 +126,13 @@ float Plane::relative_ground_altitude(bool use_rangefinder_if_available, bool us
 #endif // AP_MAVLINK_MAV_CMD_SET_HAGL_ENABLED
 
 #if AP_RANGEFINDER_ENABLED
-   if (use_rangefinder_if_available && rangefinder_state.in_range) {
+   if (rangefinder_use(use_rangefinder) && rangefinder_state.in_range) {
         return rangefinder_state.height_estimate;
    }
 #endif
 
 #if HAL_QUADPLANE_ENABLED && AP_RANGEFINDER_ENABLED
-   if (use_rangefinder_if_available && quadplane.in_vtol_land_final() &&
+   if (rangefinder_use(use_rangefinder) && quadplane.in_vtol_land_final() &&
        rangefinder.status_orient(rangefinder_orientation()) == RangeFinder::Status::OutOfRangeLow) {
        // a special case for quadplane landing when rangefinder goes
        // below minimum. Consider our height above ground to be zero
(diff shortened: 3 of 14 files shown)


timtuxworth:
This is excellent @tridge thanks. Is it scope creep to ask about adding a bit for regular flight/terrain following? Probably I guess, but there are probably people who might like that (in future).

tridge:
> Is it scope creep to ask about adding a bit for regular flight/terrain following?

yes, too much scope creep, sorry

ArduPlane/quadplane.cpp
```diff
@@ -1871,7 +1871,7 @@ void QuadPlane::update_throttle_suppression(void)
     }
 
     // if we are more than 5m from home altitude then allow motors to run
-    if (plane.relative_ground_altitude(plane.g.rangefinder_landing) > 5) {
+    if (plane.relative_ground_altitude(RangeFinderUse::TAKEOFF) > 5) {
```
IamPete1: Its not clear cut that takeoff is the only use case here. I think it right that it probably is, but its not 100% clear.

The whole throttle suppression stuff is very confusing.

ArduPlane/quadplane.cpp
```diff
@@ -3867,7 +3868,7 @@ float QuadPlane::get_weathervane_yaw_rate_cds(void)
     float wv_output;
     if (weathervane->get_yaw_out(wv_output,
                                      plane.channel_rudder->get_control_in(),
-                                     plane.relative_ground_altitude(plane.g.rangefinder_landing),
+                                     plane.relative_ground_altitude(RangeFinderUse::LANDING),
```
IamPete1: Weathervane certainly is not only used when landing. We need to check for landing and takeoff to pass in here too you could set it correctly in those cases.

ArduPlane/takeoff.cpp
```diff
@@ -381,7 +381,7 @@ int8_t Plane::takeoff_tail_hold(void)
  */
 void Plane::landing_gear_update(void)
 {
-    g2.landing_gear.update(relative_ground_altitude(g.rangefinder_landing));
+    g2.landing_gear.update(relative_ground_altitude(RangeFinderUse::LANDING));
```
IamPete1: This will result in landing gear deploying on landing but not retracting on takeoff.

ArduPlane/quadplane.cpp
IamPete1: You should be able to pass multiple values I think?

```
const bool is_landing = in_vtol_land_sequence();
RangeFinderUse useRangeFinder = RangeFinderUse::ALL;
if (is_takeoff) {
    useRangeFinder |= RangeFinderUse::TAKEOFF;
}
if (is_landing ) {
    useRangeFinder |= RangeFinderUse::LANDING;
}
```

ArduPlane/quadplane.cpp
tridge: if we're in here we are suppressed, so can't be flying, so takeoff is the only option

peterbarker:
Closes #29095

ArduPlane/quadplane.cpp
tridge: after thinking about use cases, I can't think of why a user would want it on for landing and not for takeoff, so I combined those two in a single bit, which simplifies things a lot

ArduPlane/takeoff.cpp
tridge: fixed

ArduPlane/Parameters.cpp
```diff
@@ -785,9 +785,9 @@ const AP_Param::Info Plane::var_info[] = {
     GOBJECT(rangefinder,            "RNGFND", RangeFinder),
 
     // @Param: RNGFND_LANDING
-    // @DisplayName: Enable rangefinder for landing
-    // @Description: This enables the use of a rangefinder for automatic landing. The rangefinder will be used both on the landing approach and for final flare
-    // @Values: 0:Disabled,1:Enabled
+    // @DisplayName: Enable use of rangefinder
+    // @Description: Sets the use of a rangefinder for automatic landing and other use cases. When enabled for landing and takeoff the rangefinder will be used both on the landing approach and for final flare as well as as VTOL landing and for takeoffs and throttle suppression when close to the ground. When enabled for assist the rangefinder will be used for VTOL assistance. When enabled for climb the rangefinder will be used for the initial climb in QRTL and AUTO. Set to 0 to disable use of the rangefinder.
+    // @Bitmask: 0:All, 1:TakeoffAndLanding, 2:Assist, 3:Climb
```
IamPete1:
```suggestion
    // @Bitmask: 0:All, 1:TakeoffAndLanding, 2:Assist, 3:InitialClimb
```

ArduPlane/altitude.cpp
```diff
@@ -163,13 +163,28 @@ float Plane::relative_ground_altitude(bool use_rangefinder_if_available, bool us
...
+/*
+  return true if we should use the rangefinder for a specific use case
+ */
+bool Plane::rangefinder_use(enum RangeFinderUse use_rangefinder) const
+{
+    const uint8_t use = uint8_t(g.rangefinder_landing.get());
+    if (use == uint8_t(RangeFinderUse::NONE)) {
+        return false;
+    }
+    if (use == uint8_t(RangeFinderUse::ALL)) {
```
IamPete1: Should this just check the first bit, rather than exact equality?

ArduPlane/Parameters.cpp
tridge: thanks, changed

ArduPlane/altitude.cpp
tridge: yes, it should! fixed

ArduPlane/GCS_Plane.cpp
```diff
@@ -116,7 +116,7 @@ void GCS_Plane::update_vehicle_sensor_status_flags(void)
     const RangeFinder *rangefinder = RangeFinder::get_singleton();
     if (rangefinder && rangefinder->has_orientation(plane.rangefinder_orientation())) {
         control_sensors_present |= MAV_SYS_STATUS_SENSOR_LASER_POSITION;
-        if (plane.g.rangefinder_landing) {
+        if (uint16_t(plane.g.rangefinder_landing.get()) != 0) {
```
IamPete1, line 119:
```suggestion
        if (uint16_t(plane.g.rangefinder_landing.get()) != uint16_t(RangeFinderUse::NONE)) {
```
I guess a alternate would be to add a enum value for "any" so you can use the `rangefinder_use` function.

ArduPlane/quadplane.cpp
```diff
@@ -3861,7 +3862,7 @@ float QuadPlane::get_weathervane_yaw_rate_cds(void)
     float wv_output;
     if (weathervane->get_yaw_out(wv_output,
                                      plane.channel_rudder->get_control_in(),
-                                     plane.relative_ground_altitude(plane.g.rangefinder_landing),
+                                     plane.relative_ground_altitude(RangeFinderUse::TAKEOFF_LANDING),
```
IamPete1, line 3865: Just a note that weathervane is not only use on take off and landing. It also works in QLoiter.

ArduPlane/quadplane.cpp
```diff
@@ -3820,7 +3821,7 @@ float QuadPlane::forward_throttle_pct()
         // If we are below alt_cutoff then scale down the effect until
         // it turns off at alt_cutoff and decay the integrator
         float alt_cutoff = MAX(0,vel_forward_alt_cutoff);
-        float height_above_ground = plane.relative_ground_altitude(plane.g.rangefinder_landing);
```
IamPete1, line 3823: This is another one that is also used Qstab, Qhover and Qloiter.

IamPete1:
I wonder if things like windvane, forward throttle and landing gear should have there own bits rather than being under takeoff/landing.

ArduPlane/quadplane.cpp
tridge: yes, though if you don't want it for takeoff/landing you are unlikely to want it for loiter. It's a trade off between complexity and usefulness

PR #30460, opened 2025-06-25
AP_Scripting: added BTAG_CUR_CYCLES

tridge:
this can be used by another battery estimation lua script to adjust for max number of cycles across all batteries


libraries/AP_Scripting/applets/BatteryTag.lua
```diff
@@ -69,7 +86,14 @@ local function check_batterytag()
         return
     end
 
-    highest_cycles = math.max(num_cycles, highest_cycles)
+    if num_cycles > highest_cycles then
+        highest_cycles = num_cycles
+        BTAG_CUR_CYCLES:set_and_save(highest_cycles)
+    end
+    if not node_cycles[nodeid] then
```
peterbarker, line 93: I wonder if we should report any time these change?  Long-running vehicles swapping redundant batteries might like that.

PR #30461, opened 2025-06-25
Fix non-Periph Network Bootloodaer

magicrub:
Fix non-Periph Network Bootlooder

Build fails on any FMU bootloader with networking

problems that this PR fixes:
- RAM_RESERVE_START is not defined
- CANARD_ENABLE_DEADLINE is set to 1 which is not implemented in the bootloader
- CANARD_MULTI_IFACE is set to 1 which is not implemented in the bootloader

### DIFF
--- a/Tools/ardupilotwaf/chibios.py
+++ b/Tools/ardupilotwaf/chibios.py
@@ -488,7 +488,7 @@ def setup_canmgr_build(cfg):
         ]
     env.CFLAGS += ['-DHAL_CAN_IFACES=2']
 
-    if not env.AP_PERIPH:
+    if not env.AP_PERIPH and not cfg.options.bootloader:
         env.DEFINES += [
             'DRONECAN_CXX_WRAPPERS=1',
             'USE_USER_HELPERS=1',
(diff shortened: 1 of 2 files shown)


magicrub:
It also needs AP_CHECK_FIRMWARE_ENABLED=1 but I didn't quite want to add that to this PR (yet)

magicrub:
In master, this is what's needed to get CubeRedPrimary bootloader to build. In this case, the other CAN settings don't seem to be required. Interesting!

```
# Ethernet
PC1  ETH_MDC            ETH1
PA2  ETH_MDIO           ETH1
PC4  ETH_RMII_RXD0      ETH1
PC5  ETH_RMII_RXD1      ETH1
PB12 ETH_RMII_TXD0      ETH1
PB13 ETH_RMII_TXD1      ETH1
PB11 ETH_RMII_TX_EN     ETH1
PA7  ETH_RMII_CRS_DV    ETH1
PA1  ETH_RMII_REF_CLK   ETH1

define BOARD_PHY_ID  MII_LAN8720_ID
define BOARD_PHY_RMII

# Refer to https://maclookup.app/vendors/cubepilot-pty-ltd
# Note, lower 3 bytes (ADDR3,4,5) will be replaced with the platform UUID
define AP_NETWORKING_DEFAULT_MAC_ADDR "A8:B0:28:00:00:00"

include ../include/network_bootloader.inc
undef AP_NETWORKING_CAN_MCAST_ENABLED
define AP_NETWORKING_CAN_MCAST_ENABLED 0
define AP_CHECK_FIRMWARE_ENABLED 1

```

peterbarker:
Does the bootloader work after building?

As it is we'd be changing behaviour on a lot of boards without any positive benefit ATM, I think.  We should probably make sure that the only boards with compiler output change are those that can actually get the benefits of the PR.

Would be awfully nice to have network-bootloader enablable on ethernet-capable boards!  When do you think you might get back to this PR, @magicrub ?

PR #30466, opened 2025-06-25
Backport: Plane: Fix MAV_CMD_NAV_LOITER_TURNS with zero turns skipping waypoint location

robertlong13:
4.6 backport of #28614


robertlong13:
And I've tested that the autotest correctly fails before the second commit is applied.

rmackay9:
This is included in 4.6.2-beta1, thanks!

PR #30467, opened 2025-06-25
autotest: correct bisect-helper instructions

peterbarker:


Tools/autotest/bisect-helper.py
```diff
@@ -21,10 +21,13 @@
 Work out who killed bebop:
 cp -a Tools/autotest/bisect-helper.py /tmp
 git bisect reset
-git bisect good a7647e77d9 &&
-  git bisect bad 153ad9539866f8d93a99e9998118bb090d2f747f &&
+BAD=master
+GOOD=master~256
```
andyp1per: FWIW I don't find this an improvement. I always use specific commits

peterbarker: Cool.  I've reverted that bit.  Now just uses a board that works and an example of a board failure string

tridge (approved):

PR #30471, opened 2025-06-25
Disable uncommon actuators on boards with less than 2MB of flash

peterbarker:
See "Exotic ESCs" here: https://github.com/ardupilot/ardupilot/issues/28210


amilcarlucas:
FETtecOne wire is not a Servo, It's a commonly used ESC. It's wrongly categorized in the CBS.
I do not think it is nice to remove it on F4 (small)  boards as these boards are the ones that use FETtec more often.

andyp1per (changes requested):
I agree not to remove fetec

peterbarker:
Updated for just removing Volz and Robotis:

```
Durandal                            -3216  *           -3552   -3520              -2984  -3480  -3352
KakuteH7-bdshot                     -3584  *           -3272   -3304              -4648  -3816  -3736
MatekF405                           *      *           *       *                  *      *      *
Pixhawk1-1M-bdshot                  *                  *       *                  *      *      *
SITL_x86_64_linux_gnu               *                  *       *                  *      *      *
revo-mini                           *      *           *       *                  *      *      *
```

amilcarlucas:
Those are very good savings. How about merging just 2 of the three commits for now?

amilcarlucas (changes requested):
Keep the FETtec as it's usage is widly spead on small FCs.

mduclehcm:
I think we should use hwdefine to disable a feature instead of hard-disabling it in the source code.
We can have multiple hwdefine files, for example: default, 1MB, and 2MB flash hwdefine files.
If we disable a feature using the default hwdefine, users just need to tweak an option in the custom build server to get it working when needed.
I used to struggle to enable the Nooploop lidar sensor for the Matek 405 because it was only enabled for boards with more than 1MB flash. But the driver only has about 100 lines of code.

peterbarker:
> I think we should use hwdefine to disable a feature instead of hard-disabling it in the source code. We can 

I'm not sure you understand what's going on here.  These defines are what the custom build server interface is based around.  This PR simply stop us building these drivers in on boards with 2MB of flash or less - they can still be used, you just need a custom build (eg. via the custom build server)

amilcarlucas:
@peterbarker drop the first commit and LGTM

peterbarker:
Well, I've dropped the first commit, so now it's just Volz and Robotis.

We're similarly unable to remove KDECan and PiccoloCAN because apparently these are super-special-nobody-can-do-without device drivers (https://github.com/ArduPilot/ardupilot/pull/30298)

I'd like to remove all of these less-used drivers to avoid the hunger-games shenanigans when we actually do start running short on flash again, but we're definitely hitting the "but my actuator is special" problems with making these things optional.

peterbarker:
> FETtecOne wire is not a Servo, It's a commonly used ESC. It's wrongly categorized in the CBS. I do not think it is nice to remove it on F4 (small) boards as these boards are the ones that use FETtec more often.

It's already unavailable by default on F405 board.  I'm not seeing a large number of new F427 boards.

andyp1per (approved):

PR #30472, opened 2025-06-25
allow AvoidADSB to be compiled out of the ArduCopter codebase

peterbarker:
See https://github.com/ardupilot/ardupilot/issues/28210

Note that this is a bit confusing as there's a *flightmode* which is responsible for navigation, but there's a bunch more code in `ArduCopter/avoidance_adsb.cpp` which I've gated with the same define.  Can only stomach so many different defines...

```
###### Enabling COPTER_MODE_AVOIDADSB(AP_COPTER_MODE_AVOIDADSB_ENABLED) on copter costs -4376 bytes
```

```
###### Enabling ADSB(HAL_ADSB_ENABLED) on copter costs -23680 bytes
```

### DIFF
--- a/Tools/scripts/build_options.py
+++ b/Tools/scripts/build_options.py
@@ -151,6 +151,7 @@ BUILD_OPTIONS = [
 
     Feature('Camera', 'RUNCAM', 'AP_CAMERA_RUNCAM_ENABLED', 'Enable RunCam control', 0, 'Camera'),
 
+    Feature('Copter', 'COPTER_MODE_AVOIDADSB', 'AP_COPTER_MODE_AVOIDADSB_ENABLED', 'Enable AvoidADSB', 0, 'ADSB'),
     Feature('Copter', 'MODE_ZIGZAG', 'MODE_ZIGZAG_ENABLED', 'Enable Mode ZigZag', 0, None),
     Feature('Copter', 'MODE_SYSTEMID', 'MODE_SYSTEMID_ENABLED', 'Enable Mode SystemID', 0, 'Logging'),
     Feature('Copter', 'MODE_SPORT', 'MODE_SPORT_ENABLED', 'Enable Mode Sport', 0, None),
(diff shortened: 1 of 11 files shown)


peterbarker:
Note that this doesn't remove the feature by default; it's just an option in the CBS

Tools/scripts/build_options.py
```diff
@@ -151,6 +151,7 @@ def config_option(self):
 
     Feature('Camera', 'RUNCAM', 'AP_CAMERA_RUNCAM_ENABLED', 'Enable RunCam control', 0, 'Camera'),
 
+    Feature('Copter', 'COPTER_MODE_AVOIDADSB', 'AP_COPTER_MODE_AVOIDADSB_ENABLED', 'Enable AvoidADSB', 0, 'ADSB'),
```
rmackay9, line 154: I think "COPTER_MODE_AVOIDADSB" should be just "MODE_AVOIDADSB"

rmackay9:
While reviewing this, I noticed we have two payload place definitions:

AP_MISSION_NAV_PAYLOAD_PLACE_ENABLED
AC_PAYLOAD_PLACE_ENABLED

I think we should get rid of the Copter specific one

Tools/scripts/build_options.py
peterbarker: > I think "COPTER_MODE_AVOIDADSB" should be just "MODE_AVOIDADSB"

Plane has one of these too... having same-names defines across multiple vehicles which might to subtly different things is what I' concerned about, which is why I've been putting COPTER/PLANE/ROVER into new things going into build_options.py

peterbarker: I wonder if I should just create a single define for `AP_AVOIDANCE_ENABLED` and use it everywhere instead?  In the same way that we have `AP_ARMING_ENABLED`.

rmackay9: Hi @peterbarker,

For copter and maybe rover, there are 3 different avoidance features:

1. adsb avoidance (which is silly and should be removed or be combined with path planning as PeterH has recently done for AIS)
2. simple avoidance (e.g. stopping)
3. path planning (dijkstras, bendy ruler)

So I think we need these defines:

a) ADSB_ENABLE which just enables/disables the ADSB sensors
b) MODE_ADSBAVOID_ENABLE which should definitely default to being off and we can move towards removing the mode completely
c) simple avoidance
d) path planning avoidance

peterbarker:
We merged an alternative to this.

PR #30473, opened 2025-06-25
AP_BattMonitor: compile error if sum-backend and <2 max-instances

peterbarker:


magicrub:
Sure.. but how about setting this in the config too
```
#ifndef AP_BATTERY_SUM_ENABLED
#define AP_BATTERY_SUM_ENABLED AP_BATTERY_BACKEND_DEFAULT_ENABLED && (AP_BATT_MONITOR_MAX_INSTANCES >= 3)
#endif

```

libraries/AP_BattMonitor/AP_BattMonitor.h
```diff
@@ -16,6 +16,14 @@
 #define AP_BATT_MONITOR_MAX_INSTANCES       9
 #endif
 
+// sanity checks for battery backends:
+#if AP_BATTERY_SUM_ENABLED
+#if AP_BATT_MONITOR_MAX_INSTANCES < 3
+#error "AP_BATT_MONITOR_MAX_INSTANCES must be at least three for AP_BATTERY_SUM_ENABLED"
+#endif  // AP_BATT_MONITOR_MAX_INSTANCES < 3
+#endif  // AP_BATTERY_SUM_ENABLED
```
magicrub:
```suggestion
#if AP_BATTERY_SUM_ENABLED && AP_BATT_MONITOR_MAX_INSTANCES < 3
#error "AP_BATT_MONITOR_MAX_INSTANCES must be at least three for AP_BATTERY_SUM_ENABLED"
#endif
```

magicrub:
@peterbarker I pushed another commit on here with a better solution

libraries/AP_BattMonitor/AP_BattMonitor.h
peterbarker: I've applied this one

peterbarker:
> @peterbarker I pushed another commit on here with a better solution

I've kept the reformatting - it does look better that way.

I put it in AP_BattMonitor.h to avoid the overhead of including it everywhere, which is what _config.h can do.  You did highlight that's the wrong place for it - I've moved it into the .cpp file for `AP_BattMonitor_Sum` instead.  We only need to look at that sanity check *once* during the entire compilation, so that's the right place for it.

The number-of-sensors define move is good (but shows the poor naming of same).

I have not left in the change to the activation of the feature for a few reasons:
 - it's a change in behaviour whereas this one is supposed to be NFC.
 - it may actually cause some peripherals not to build
 - it may not be wanted on some peripherals that just want to return raw data (like peripherals...)
 - we try not to get clever with these defines by turning features on - apart from making them dependent on things they won't compile without (eg. Sum requires Battery), and our checks for HAL_PROGRAM_SIZE.  And that last one isn't a gimme, either; it's convenient (keep stuff grouped) and traditional, but we could actually do the same thing with an include...

peterbarker:
Fun times :-)

```
2025-06-26T01:11:38.7903929Z ../../libraries/AP_BattMonitor/AP_BattMonitor_Sum.cpp:13:2: error: #error "AP_BATT_MONITOR_MAX_INSTANCES must be at least three for AP_BATTERY_SUM_ENABLED"
2025-06-26T01:11:38.7911372Z    13 | #error "AP_BATT_MONITOR_MAX_INSTANCES must be at least three for AP_BATTERY_SUM_ENABLED"
2025-06-26T01:11:38.7912062Z       |  ^~~~~
2025-06-26T01:11:38.7912439Z compilation terminated due to -Wfatal-errors.
2025-06-26T01:11:38.7912774Z 
2025-06-26T01:11:39.2676729Z Build failed
2025-06-26T01:11:39.2677146Z  -> task in 'objs/AP_BattMonitor' failed (exit status 1): 
2025-06-26T01:11:39.2677730Z 	{task 139658985274080: cxx AP_BattMonitor_Sum.cpp -> AP_BattMonitor_Sum.cpp.0.o}
2025-06-26T01:11:39.2678130Z  (run with -v to display more information)
2025-06-26T01:11:39.2678508Z Waf: Leaving directory `/__w/ardupilot/ardupilot/build/CubeOrange-periph'
```

libraries/AP_BattMonitor/AP_BattMonitor_Sum.cpp
```diff
@@ -9,6 +9,10 @@
 
 #include "AP_BattMonitor_Sum.h"
 
+#if AP_BATT_MONITOR_MAX_INSTANCES < 3
```
tridge: i think < 2 would be better

PR #30476, opened 2025-06-25
Plane: added an option to enable flaps by actual speed

tridge:
When the new FLIGHT_OPTIONS bit is enabled:
 - In manual throttle modes actual airspeed (or estimate) is used for flaps
 - in auto-throttle modes the minimum of target airspeed and actual airspeed is used

this allows for auto flaps in manual throttle modes, and when speed drops in an uncommanded fashion in auto-throttle modes

this needs to be used with caution and carefully tested on an aircraft as it could result in an oscillation, as adding flaps can change the airspeed

### DIFF
--- a/ArduPlane/servos.cpp
+++ b/ArduPlane/servos.cpp
@@ -685,10 +685,23 @@ void Plane::set_servos_flaps(void)
         manual_flap_percent = channel_flap->percent_input();
     }
 
-    if (control_mode->does_auto_throttle()) {
+    const auto flap_actual_speed = flight_option_enabled(FlightOptions::FLAP_ACTUAL_SPEED);
+    const bool has_target_airspeed = control_mode->does_auto_throttle();
+    if (has_target_airspeed || flap_actual_speed) {
         int16_t flapSpeedSource = 0;
-        if (ahrs.using_airspeed_sensor()) {
+        float est_airspeed;
+        bool have_airspeed = ahrs.airspeed_estimate(est_airspeed);
+        if (has_target_airspeed && ahrs.using_airspeed_sensor()) {
             flapSpeedSource = target_airspeed_cm * 0.01f;
+            if (flap_actual_speed) {
+                // if we have a target and also want to use actual
+                // speed then use the minimum of the two so we bring
+                // flaps in early when deliberately slowing down
+                flapSpeedSource = MIN(flapSpeedSource, est_airspeed);
+            }
+        } else if (flap_actual_speed && have_airspeed) {
+            // use actual speed directly
+            flapSpeedSource = est_airspeed;
         } else {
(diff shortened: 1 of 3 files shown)


IamPete1:
I guess we could add some hysteresis, or have separate flap deploy and flap retract airspeeds. If we wanted to be more "feature complete" we could also constrain the target speed based on flap setting. So you can't go fast until you have put the flaps up and you can't go slow until you have put them down.

tridge:
> I guess we could add some hysteresis, or have separate flap deploy and flap retract airspeeds

this PR as is seems to work well on the two aircraft we've tried it on (Bixler and ARACE Phoenix), if user feedback points us towards a fancier scheme we can do that, but for now this is simple and seems to work

ARACE-DEV:
We did bench-test and flight test the function, seems to work as expected.

ArduPlane/servos.cpp
```diff
@@ -685,10 +685,23 @@ void Plane::set_servos_flaps(void)
...
     }
 
-    if (control_mode->does_auto_throttle()) {
+    const auto flap_actual_speed = flight_option_enabled(FlightOptions::FLAP_ACTUAL_SPEED);
+    const bool has_target_airspeed = control_mode->does_auto_throttle();
+    if (has_target_airspeed || flap_actual_speed) {
         int16_t flapSpeedSource = 0;
-        if (ahrs.using_airspeed_sensor()) {
+        float est_airspeed;
+        bool have_airspeed = ahrs.airspeed_estimate(est_airspeed);
```
peterbarker, line 693:
```suggestion
        const bool have_airspeed = ahrs.airspeed_estimate(est_airspeed);
```

or could be moved into the one place it is used and save the call when we have a target and we're using a asensor

peterbarker (approved):
An autotest would be nice.  It's bolting a few things together to do its thing, so I could see unintended consequences in the future.

ArduPlane/servos.cpp
```diff
@@ -685,10 +685,23 @@ void Plane::set_servos_flaps(void)
...
             flapSpeedSource = target_airspeed_cm * 0.01f;
+            if (flap_actual_speed) {
+                // if we have a target and also want to use actual
+                // speed then use the minimum of the two so we bring
+                // flaps in early when deliberately slowing down
+                flapSpeedSource = MIN(flapSpeedSource, est_airspeed);
+            }
+        } else if (flap_actual_speed && have_airspeed) {
+            // use actual speed directly
+            flapSpeedSource = est_airspeed;
```
IamPete1, line 704: This case bypasses the `using_airspeed_sensor` check.

IamPete1: https://github.com/ArduPilot/ardupilot/pull/33995

PR #30478, opened 2025-06-26
AP_Networking: add options to control PPP timeouts

tridge:
some PPP daemons are sensitive to how we handle timeouts. This gives options to disable the data timeout and echo limit to maximise compatibility with a range of PPP daemon implementations
also changes the timeout (when enabled) to 15s
This also reverts a change where we lie about sending a PPP packet when we can't send it due to buffer space. I thought this was needed based on past testing, but I had been fooled by a problem with the LTE modem firmware. With the updated firmware it is fine to return 0 if we can't send a packet. Also tested with pppd on Linux.


timtuxworth:
I've just tested this with my vehicle that had been failing. It works fine if I use UDP to connect to the board over PPP, but using TCP I've found creates a slow and unreliable connection. Tridge has previously encouraged me to use UDP over TCP, this probably rises to the level of maybe it should be a note/warning in the wiki.

I have another vehicle that is using the BotBlox DroneNet router device that has a built in PPP server. This also works with this new code.

rmackay9 (approved):
I think this looks safe and it apparently fixes a regression ahead of 4.6.2-beta1 so I'm going to approve and merge.

We struggle to find reviewers for this bit of the code so any further feedback is greatly appreciated!

PR #30479, opened 2025-06-26
Add and use AP_ADSB_AVOIDANCE_ENABLED

peterbarker:
alternative to https://github.com/ArduPilot/ardupilot/pull/30472

```
MatekF405                           *      *           0       0                 *      *      *
Pixhawk1-1M-bdshot                  *                  0       0                 *      *      *
Pixracer                            *      *           *       *                 *      *      *
revo-mini                           *      *           0       0                 *      *      *
skyviper-journey                                       0                                       
skyviper-v2450                                         0                                       
```

The new define doesn't actually save as much as I expected: 
```
----------  -----  ----------  ------  -----  -----  -----  ----
Board       blimp  bootloader  copter  heli   plane  rover  sub
CubeOrange  -224   *           -4592   -4592  -4144  -208   -208
----------  -----  ----------  ------  -----  -----  -----  ----
```

This leaves the feature in place.  Silently removing safety features not high on my list of priorities.  Non-silently removing them slightly higher.  The trick to the latter is probably to move the existing enable parameter out to be compiled in only if the feature is not compield in - and for it to be an error for it to be set.  Then we have a new enable parameter (with parameter conversion from the old one).

### DIFF
--- a/Tools/scripts/build_options.py
+++ b/Tools/scripts/build_options.py
@@ -380,6 +380,7 @@ BUILD_OPTIONS = [
     Feature('Other', 'CUSTOM_ROTATIONS', 'AP_CUSTOMROTATIONS_ENABLED', 'Enable Custom  sensor rotations', 0, None),
     Feature('Other', 'PID_FILTERING', 'AP_FILTER_ENABLED', 'Enable PID filtering', 0, None),
     Feature('Other', 'POLYFENCE_CIRCLE_INT_SUPPORT', 'AC_POLYFENCE_CIRCLE_INT_SUPPORT_ENABLED', 'Fence circle compatability', 0, None),  # NOQA:E501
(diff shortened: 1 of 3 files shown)


Tools/scripts/build_options.py
```diff
@@ -380,6 +380,7 @@ def config_option(self):
     Feature('Other', 'CUSTOM_ROTATIONS', 'AP_CUSTOMROTATIONS_ENABLED', 'Enable Custom  sensor rotations', 0, None),
     Feature('Other', 'PID_FILTERING', 'AP_FILTER_ENABLED', 'Enable PID filtering', 0, None),
     Feature('Other', 'POLYFENCE_CIRCLE_INT_SUPPORT', 'AC_POLYFENCE_CIRCLE_INT_SUPPORT_ENABLED', 'Fence circle compatability', 0, None),  # NOQA:E501
+    Feature('Other', 'AP_AVOIDANCE', 'AP_ADSB_AVOIDANCE_ENABLED', 'Enable "ADSB" Avoidance', 0, 'ADSB'),
```
rmackay9, line 383: looking good!  nitpick: we can probably remove the double quotes around ADSB.

PR #30482, opened 2025-06-26
Tool: Fix Docker build on macOS Apple Silicon by handling GID conflicts

mduclehcm:
## Problem
Docker builds fail on macOS 15.5 due to group ID conflicts. The original Dockerfile assumes the specified GID (1000) is available, but on macOS systems, this GID may already be assigned to existing system groups, causing the `groupadd` command to fail.

## Solution
This PR adds conditional logic to handle GID conflicts gracefully

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,12 @@ ARG SKIP_AP_COV_ENV=1
 ARG SKIP_AP_GIT_CHECK=1
 ARG DO_AP_STM_ENV=1
 
-RUN groupadd ${USER_NAME} --gid ${USER_GID}\
+RUN if ! getent group ${USER_GID} > /dev/null 2>&1; then \
+        groupadd ${USER_NAME} --gid ${USER_GID}; \
+    else \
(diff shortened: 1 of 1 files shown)


khancyr:
Hello,

Thanks for reporting this but your diagnosis seems wrong.

What is the error you got ? The container shouldn't know the host user and groups

Can you share the commands you are using to build the container?

peterbarker:
@khancyr apparently the base image for MaxOSX had extra stuff in it not in the base image for e.g. x86_64?

Looking at https://askubuntu.com/questions/1513927/ubuntu-24-04-docker-images-now-includes-user-ubuntu-with-uid-gid-1000 it looks like this is something we're going to have to work on for x86_64 images when the Docker Ubuntu version is moved forward anyway, so we should look a bit harder at this PR, I think.

khancyr:
override by https://github.com/ArduPilot/ardupilot/pull/32484
Thanks anyway

